PR #4010, opened 2024-05-09
HOSTEDCP-1542: hcco: expose Azure federated token files

stevekuznetsov:
hcco: expose Azure federated token files

Expect components to use Azure workload identity, propagate those
secrets from the HyperShift Operator to each individual HostedCluster
namespace, then to the guest clusters themselves.

---

### DIFF
--- a/api/hypershift/v1beta1/hostedcluster_types.go
+++ b/api/hypershift/v1beta1/hostedcluster_types.go
@@ -1701,12 +1701,13 @@ type AWSServiceEndpoint struct {
 // would be pre-created and then their names would be used respectively in the ResourceGroupName, SubnetName, VnetName
 // fields of the Hosted Cluster CR. An existing cloud resource is expected to exist under the same SubscriptionID.
 type AzurePlatformSpec struct {
-	// Credentials is the object containing existing Azure credentials needed for creating and managing cloud
-	// infrastructure resources.
+	// TenantID is the id for the Azure tenant in which this Hosted Cluster will be created.
 	//
 	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="TenantID is immutable"
+	// +immutable
 	// +required
-	Credentials corev1.LocalObjectReference `json:"credentials"`
+	TenantID string `json:"tenantID,omitempty"`
 
 	// Cloud is the cloud environment identifier, valid values could be found here: https://github.com/Azure/go-autorest/blob/4c0e21ca2bbb3251fe7853e6f9df6397f53dd419/autorest/azure/environments.go#L33
 	//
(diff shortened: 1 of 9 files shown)


api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1701,12 +1701,13 @@ type AWSServiceEndpoint struct {
...
-	// Credentials is the object containing existing Azure credentials needed for creating and managing cloud
-	// infrastructure resources.
+	// TenantID is the id for the Azure tenant in which this Hosted Cluster will be created.
 	//
 	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="TenantID is immutable"
+	// +immutable
 	// +required
-	Credentials corev1.LocalObjectReference `json:"credentials"`
+	TenantID string `json:"tenantID,omitempty"`
```
stevekuznetsov, line 1710: FWIW a quick review of the forums, etc, shows that this (and the subscription ID) are not secret values.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -1451,6 +1413,17 @@ func (r *reconciler) reconcileCloudCredentialSecrets(ctx context.Context, hcp *h
 	return errs
 }
 
+func AzureFederatedTokenCredentialSecretData(name, infraID string, config *hyperv1.AzurePlatformSpec) map[string][]byte {
+	return map[string][]byte{
+		"azure_region":               []byte(config.Location),
+		"azure_resource_prefix":      []byte(name + "-" + infraID),
```
stevekuznetsov, line 1419: The conversation [here](https://github.com/openshift/cloud-credential-operator/blob/master/docs/azure_workload_identity.md) does not explicitly list the resource prefix and resource-group, but I can't imagine it will hurt?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1791,6 +1792,70 @@ type AzurePlatformSpec struct {
...
 	// +required
 	SecurityGroupID string `json:"securityGroupID,omitempty"`
+
+	// IngressClientID is the client ID appropriate for the Ingress Operator.
+	//
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="IngressClientID is immutable"
+	// +kubebuilder:validation:Required
+	// +immutable
+	// +required
+	IngressClientID string `json:"ingressClientID"`
```
stevekuznetsov, line 1802: All of these we should be able to get super easily by using `ccoctl` from the release payload, but the node-pool management and CPO are going to be something we need to figure out and encode into `CredentialsRequests` ourselves.

stevekuznetsov:
/hold

We should just use `ccoctl` and not do this.

PR #4018, opened 2024-05-10
HOSTEDCP-1542: cmd/cluster: refactor to remove example fixtures

stevekuznetsov:
The goal of this refactor is to reduce the complexity in the command-line tooling. Overall, the changes here remove duplicative structures that copied data around and co-locate the logic with the data, instead of first aggregating all data into one uber-structure and then conditionally acting on that structure. These refactors have a number of benefits:

 - locality of behavior: in the past, it was very difficult if not impossible to determine where a value was used, as it would be bound to a flag in one package, copied around between container structs a couple of times, then have some generic logic act on the presence or absence of the value to e.g. change a field on the HostedCluster. Simply reading the generic logic was often not enough to understand what was going on, as many of the conditional branches in the example fixture code could only ever trigger for one specific platform, and you'd never know unless you traced how the example options uber-struct had its fields set in every provider.
 - clear go-to-definition: as a knock-on effect of the above, now there's *one* structure that holds a command-line flag and it's trivial to use the LSP when determining where that flag is used and how
 - composability: as exemplified in the KubeVirt NodePool code, we are able to compose commands as necessary. When commands re-use the same arguments with the same flags and the same validation logic, there's no need to copy things around and re-implement anything; by localizing flag binding, validation and option completion, we gain small, composable parts that we can use to build larger commands with


cmd/nodepool/kubevirt/create.go
```diff
@@ -2,75 +2,118 @@ package kubevirt
...
+	flags.StringVar(&opts.Memory, "memory", opts.Memory, "The amount of memory which is visible inside the Guest OS (type BinarySI, e.g. 5Gi, 100Mi)")
+	flags.Uint32Var(&opts.Cores, "cores", opts.Cores, "The number of cores inside the vmi, Must be a value greater or equal 1")
+	flags.StringVar(&opts.RootVolumeStorageClass, "root-volume-storage-class", opts.RootVolumeStorageClass, "The storage class to use for machines in the NodePool")
+	flags.Uint32Var(&opts.RootVolumeSize, "root-volume-size", opts.RootVolumeSize, "The size of the root volume for machines in the NodePool in Gi")
+	flags.StringVar(&opts.RootVolumeAccessModes, "root-volume-access-modes", opts.RootVolumeAccessModes, "The access modes of the root volume to use for machines in the NodePool (comma-delimited list)")
+	flags.StringVar(&opts.RootVolumeVolumeMode, "root-volume-volume-mode", opts.RootVolumeVolumeMode, "The volume mode of the root volume to use for machines in the NodePool. Supported values are \"Block\", \"Filesystem\"")
+	flags.StringVar(&opts.ContainerDiskImage, "containerdisk", opts.ContainerDiskImage, "A reference to docker image with the embedded disk to be used to create the machines")
```
davidvossel, line 40: one oddity we have to deal with is that there are cli args exposed in the `hypershift` cli that we didn't want carried over to `hcp` cli. 

for kubevirt, the `--containerdisk` cli arg is one of those options. In our case, we haven't used that arg in our development flows for quite some time now, so it can be removed entirely. There might be other cases though where the dev cli (hypershift) and productized cli (hcp) differ slightly. I think that's part of the reason that the tools split the way they did.

stevekuznetsov: Perhaps we can use this as a good forcing function to make the decision about why we have it in `hypershift` if we don't have it in `hcp`! If it's intentional, we can factor a base set of options and then decorate each side with ancillary things as necessary. What we have now is implicit at best - what would you like me to do with these? I can delete this arg entirely, and are the others you speak of in other platforms?

bryan-cox: FWIW this same oddity applies to AWS as well. There are additional AWS flags in `hypershift` than in `hcp`. May be true for `Agent` platform as well.

stevekuznetsov: @bryan-cox can you speak to whether that's intended or emergent behavior?

stevekuznetsov: (a good number of the flags that differed in KubeVirt were accidental, and *should* have been in `hcp`, but were not)

bryan-cox: Yeah we went through an  exercise to figure out what flags from the `hypershift cli` should be supported officially in the `hcp cli`.

davidvossel: > a good number of the flags that differed in KubeVirt were accidental, and should have been in hcp

From what I remember, those missing kubevirt args were between the `hcp create cluster` and `hcp create nodepool` commands. Both those commands are "productized"

I what we're talking about here in this comment section is more about the differences between `hypershift` (dev cli) and `hcp` (productized cli).  There were some dev cli args (like the container disk one for KubeVirt) that weren't meant to be productized.


we made this kind of complicated :( Speaking for the KubeVirt platform, i'm willing to drop all discrepancies that exist between the hypershift and hcp cli... meaning if it's kubevirt related and only in hypershift cli, we can drop it entirely.

davidvossel (commented):
kubevirt parts look accurate

sjenning:
```
=== RUN   TestAutoscaling/ValidateHostedCluster/EnsureNoCrashingPods
    util.go:533: Container cloud-controller-manager in pod kubevirt-cloud-controller-manager-7cc4ccf9b9-644r2 has a restartCount > 0 (1)

=== RUN   TestCreateCluster/ValidateHostedCluster/EnsureNoCrashingPods
    util.go:533: Container cloud-controller-manager in pod kubevirt-cloud-controller-manager-6d48cd658d-d6k7n has a restartCount > 0 (1)
    util.go:533: Container oauth-apiserver in pod openshift-oauth-apiserver-5c7dd4b5dc-l2s5z has a restartCount > 0 (1)
```

Seems like a flake but some jobs are passing now (new development)

/retest-required

sjenning:
Not sure exactly what is happening here, but this PR _does_ seem to be causing `EnsureNoCrashingPods` to fail in kubevirt e2e.  Other kubevirt presubs do not show this flake and we hit this in the last 3 runs.  Junit not working for some reason as well make it even more annoying.

```
    util.go:533: Container cloud-controller-manager in pod kubevirt-cloud-controller-manager-84fb6f744c-9kljp has a restartCount > 0 (1)
    util.go:533: Container oauth-apiserver in pod openshift-oauth-apiserver-5f964f6b-dx7lm has a restartCount > 0 (1)
```

https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_hypershift/4018/pull-ci-openshift-hypershift-main-e2e-kubevirt-aws-ovn/1798984858463113216/artifacts/e2e-kubevirt-aws-ovn/run-e2e-local/artifacts/TestCreateCluster/namespaces/e2e-clusters-dgknb-example-mhl8d/core/pods/logs/kubevirt-cloud-controller-manager-84f6cd4458-pfjlp-cloud-controller-manager-previous.log

stevekuznetsov:
@sjenning that's ok - certainly possible it's broken, will look.

stevekuznetsov:
Kubevirt failing on

```
 === NAME  TestNodePool/HostedCluster0/Main/TestNodePoolReplaceUpgrade
    nodepool_upgrade_test.go:160: Validating all Nodes have the synced labels and taints
    nodepool_upgrade_test.go:167: Updating NodePool image. Image: registry.build05.ci.openshift.org/ci-op-z247rcb1/release@sha256:fc76bbd8dbe41816d167025346ad780301266bfda93cd95f214d89e74575e906
    util.go:428: Waiting for nodepool e2e-clusters-cd69f/example-6wrhv-test-replaceupgrade to report version 4.17.0-0.ci.test-2024-06-07-190031-ci-op-z247rcb1-latest (currently 4.17.0-0.ci-2024-06-07-041707) 
    util.go:435: Failed to get nodepool: client rate limiter Wait returned an error: context deadline exceeded
    util.go:440: 
        failed waiting for nodepool version
        Unexpected error:
            <context.deadlineExceededError>: 
            context deadline exceeded
            {}
        occurred 
```

:thinking:

PR #4020, opened 2024-05-12
CNF-12681: Added support for performance profile status

rbaturov:
This PR includes the implementation of this PR on the hypershift side:
 https://github.com/openshift/enhancements/pull/1619

Up until now, there was no defined resolution over how PerformanceProfile
status conditions with the status of the different components created
from the PerformanceProfile (MachineConfig, KubeletConfig, Tuned) are handled and exposed.
To resolve this, a custom ConfigMap is added and will be used as a middleware object.
It will hold the updated status populated by the performance profile controller
and will be watched by the NodePool controller, which will mirror the current status provided to the NodePool.status.condition.

The resolution described above consist of the following changes:
* Added a watch to the specific performance profile status ConfigMap object that will trigger the reconcile loop.
* Added the SetPerformanceProfileStatus function to mirror the performance profile status conditions to the nodepool conditions with a prefix.


stevekuznetsov:
Where does the node tuning operator usually put this information? Can the NodePool controller read it directly from there, instead of using a middleware object?

rbaturov:
> Where does the node tuning operator usually put this information? Can the NodePool controller read it directly from there, instead of using a middleware object?

In NTO on OCP, we have `performanceprofile` object and the performance profile controller is responsible for writing this information to the object status.
However, on hypershift `performanceprofiles` are applied as configmap. Thus, adjustments for this are required in the NTO.
Since NTO with these adustements can't change the ConfigMap and not the PerformanceProfile embedded in it since it controlled by the Hypershift operator which will override the changes immediately.
The Nodepool is the object which suppose to report about the status of the PerformanceProfile since it represents the current configuration of the nodes.
Since NTO cannot modify the Nodepool directly, a middle-ware object (ConfigMap) is used to write to it the status and the nodepool controller will read from it.

rbaturov:
> Sorry if these are obvious questions, but I am new - using proxy objects to funnel state in a distributed system can have a lot of downsides w.r.t. consistency, especially around outages, as etcd has no transactions and no way to keep objects in sync.
> 
> > However, on hypershift performanceprofiles are applied as configmaps. Thus, adjustments for this are required in the NTO.
> 
> Are `performanceprofiles` cluster-scoped in OCP core? Have we done a spike to see if we could namespace them and re-use that object validation for HCP?
> 

Yes performance profile are cluster scoped.
> > Since NTO cannot modify the Nodepool directly,
> 
> Why?

He doesn't have RBAC permissions for that

rbaturov:
@sjenning @csrwng I will kindly ask if you guys could review this please, as you are familiar this the enhancement proposal

support/util/performanceprofilestatus/performanceprofilestatus.go
```diff
@@ -0,0 +1,160 @@
...
+
+	return &performanceprofilev2.PerformanceProfileStatus{
+		Conditions:   conditions,
+		RuntimeClass: &runtimeClass,
+		Tuned:        &tuned,
+	}
+}
+
+func Degraded() *performanceprofilev2.PerformanceProfileStatus {
+	lastHeartbeatTime := "2024-04-18T06:55:45Z"
```
enxebre, line 99: is this a test stub? if so can we please move into its appropriate _test file?

support/util/performanceprofilestatus/performanceprofilestatus.go
```diff
@@ -0,0 +1,160 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	performanceprofilev2 "github.com/openshift/cluster-node-tuning-operator/pkg/apis/performanceprofile/v2"
+	conditionsv1 "github.com/openshift/custom-resource-status/conditions/v1"
+	"gopkg.in/yaml.v2"
+)
+
+func Available() *performanceprofilev2.PerformanceProfileStatus {
+	lastHeartbeatTime := "2024-04-18T06:55:45Z"
```
jparrill, line 15: Same as @enxebre mentioned before, for all regarding testing stub

support/util/performanceprofilestatusconfigmap/performanceprofilestatusconfigmap.go
```diff
@@ -0,0 +1,74 @@
...
+	cm := getTestPerformanceProfileStatusConfigMap(controlPlaneNamespace, nodePoolName)
+	for _, opt := range opts {
+		if err := opt(cm); err != nil {
+			return nil, err
+		}
+	}
+	return cm, nil
+}
+
+func getTestPerformanceProfileStatusConfigMap(controlPlaneNamespace, nodePoolName string) *corev1.ConfigMap {
```
jparrill, line 24: If this is part of a test component, I recommend to separate the `getTestPerformanceProfileStatusConfigMap` from `getPerformanceProfileStatusConfigMap` keeping the last one here and putting the tests things in the (as @enxebre mentioned before too (sorry to repeat myself again :/) to the appropriate `_test.go` file.

jparrill (changes requested):

support/util/performanceprofilestatus/performanceprofilestatus.go
rbaturov: Yes will do so.

support/util/performanceprofilestatus/performanceprofilestatus.go
rbaturov: @jparrill All these testing stubs are needed to create a mocked config map that will hold the performance profile status. These stubs used to create the following [3 config maps](https://github.com/rbaturov/hypershift/blob/add-pp-status-cm/hypershift-operator/controllers/nodepool/nodepool_controller_test.go#L2386)  in the `nodepool_controller_test.go`.

support/util/performanceprofilestatusconfigmap/performanceprofilestatusconfigmap.go
rbaturov: Do you mean to seperate `getTestPerformanceProfileStatusConfigMap` from the function `New`?

hypershift-operator/controllers/nodepool/nodepool_controller_test.go
```diff
@@ -2375,6 +2378,166 @@ func TestTaintsToJSON(t *testing.T) {
...
+		expectedConditions         map[string]hyperv1.NodePoolCondition
+	}{
+		{
+			name:                       "Performance profile is available",
+			PerformanceProfileStatusCM: availablePPstatusCM,
+			expectedConditions: map[string]hyperv1.NodePoolCondition{
+				performanceprofilestatus.AvailableConditionType: {
+					Type:    performanceprofilestatus.AvailableConditionType,
+					Status:  corev1.ConditionTrue,
+					Message: "cgroup=v1;",
```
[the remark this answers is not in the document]
rbaturov: That means the cluster is using cgroup v1. we used to list it in the message field of this condition. Now, I do see this is no longer the case, might need to update this.

support/util/performanceprofilestatusconfigmap/performanceprofilestatusconfigmap.go
rbaturov: most of the `performanceprofilestatusconfigmap` and `performanceprofilestatus` packages are used to create mocked configmaps (in a real cluster, generated by NTO) that will hold performance profile status.

jparrill: No, I mean, things like:

```
lastHeartbeatTime := "2024-04-18T06:55:45Z"
```

Are meant to be in a `xxx_test.go` file because they are not relevant for the logic of the code, but on testing circumstances.

rbaturov: @jparrill anyways I have changed the code please look

rbaturov:
@jparrill @enxebre I have made some notables changes. can you PTAL and review?

Tal-or:
/hold 
depends on #https://github.com/openshift/cluster-node-tuning-operator/pull/1089

hypershift-operator/controllers/nodepool/nodepool_controller.go
```diff
@@ -3366,3 +3376,64 @@ func (r *NodePoolReconciler) reconcileMirroredConfigs(ctx context.Context, logr
...
+// SetPerformanceProfileConditions checks for performance profile status updates, and reflects them in the nodepool status conditions
+func (r *NodePoolReconciler) SetPerformanceProfileConditions(ctx context.Context, logger logr.Logger, nodePool *hyperv1.NodePool, controlPlaneNamespace string, toDelete bool) error {
+	if toDelete {
+		ppConds := []string{
+			hyperv1.NodePoolPerformanceProfileTuningAvailableConditionType,
+			hyperv1.NodePoolPerformanceProfileTuningProgressingConditionType,
+			hyperv1.NodePoolPerformanceProfileTuningUpgradeableConditionType,
+			hyperv1.NodePoolPerformanceProfileTuningDegradedConditionType,
+		}
+		for _, ppCond := range ppConds {
```
jparrill, line 3389: could you avoid shortcuts like ppCond or ppConds? I would change by something like condition and performanceConditions

hypershift-operator/controllers/nodepool/nodepool_controller.go
```diff
@@ -3366,3 +3376,64 @@ func (r *NodePoolReconciler) reconcileMirroredConfigs(ctx context.Context, logr
...
+		LabelSelector: labels.SelectorFromSet(map[string]string{nodeTuningGeneratedPerformanceProfileStatusLabel: "true"}),
+		Namespace:     controlPlaneNamespace,
+	}); err != nil {
+		return err
+	}
+	if len(cmList.Items) > 1 {
+		return fmt.Errorf("there cannot be more than one PerformanceProfile ConfigMap status per NodePool. found: %d NodePool: %s", len(cmList.Items), nodePool.Name)
+	}
+	if len(cmList.Items) == 0 {
+		// Only log here and do not return an error because it might take sometime for NTO to
```
jparrill, line 3406: What happens if the CM takes forever to be created? (for whatever reason)

hypershift-operator/controllers/nodepool/nodepool_controller.go
```diff
@@ -3366,3 +3376,64 @@ func (r *NodePoolReconciler) reconcileMirroredConfigs(ctx context.Context, logr
...
+			Type:               fmt.Sprintf("%s/%s", hyperv1.NodePoolPerformanceProfileTuningConditionTypePrefix, ppCond.Type),
+			Status:             ppCond.Status,
+			Reason:             ppCond.Reason,
+			Message:            ppCond.Message,
+			ObservedGeneration: nodePool.Generation,
+		}
+		oldCondition := FindStatusCondition(nodePool.Status.Conditions, condition.Type)
+
+		// Will set the condition only if it was not set previously, or has changed
+		if oldCondition == nil || reflect.DeepEqual(oldCondition, condition) {
```
jparrill, line 3434: We try to avoid using reflect because of the performance, for that, you can use the field `observedGeneration` which basically tells you if the status has changed from last time.

hypershift-operator/controllers/nodepool/nodepool_controller_test.go
```diff
@@ -2622,6 +2622,309 @@ func TestTaintsToJSON(t *testing.T) {
...
+  lastheartbeattime: "2024-04-18T06:55:45Z"
+  lasttransitiontime: "2024-04-18T06:55:45Z"
+- type: Degraded
+  status: "False"
+  reason: ""
+  message: ""
+  lastheartbeattime: "2024-04-18T06:55:45Z"
+  lasttransitiontime: "2024-04-18T06:55:45Z"
+tuned: openshift-cluster-node-tuning-operator/openshift-node-performance-performance
+runtimeclass: performance-performance`,
```
jparrill, line 2690: In order to make the tests more readables, maybe you can create some helper functions to fill the tests data fields, it will be more readable.

hypershift-operator/controllers/nodepool/nodepool_controller_test.go
```diff
@@ -2622,6 +2622,309 @@ func TestTaintsToJSON(t *testing.T) {
...
+			// Therefore, we expect the nodepool conditions to have no performance profile conditions.
+			if !tc.hasPerformanceProfileApplied {
+				for _, NodePoolCond := range performanceProfileConditions {
+					cond := FindStatusCondition(nodePool.Status.Conditions, NodePoolCond)
+					g.Expect(cond).To(BeNil())
+				}
+				return
+			}
+
+			for _, NodePoolCond := range performanceProfileConditions {
```
jparrill, line 2915: s/NodePoolCond/NodePoolCondition/

hypershift-operator/controllers/nodepool/nodepool_controller_test.go
```diff
@@ -2622,6 +2622,309 @@ func TestTaintsToJSON(t *testing.T) {
...
+
+			ctx := context.Background()
+
+			// In case performance profile is applied, a config map holding the performance profile status is generated
+			// by NTO should exist on the hosted control plane namespace.
+			if tc.hasPerformanceProfileApplied {
+				r.Create(ctx, tc.PerformanceProfileStatusCM)
+			}
+			err := r.SetPerformanceProfileConditions(ctx, logr.Discard(), nodePool, controlPlaneNamespace, false)
+			g.Expect(err).ToNot(HaveOccurred())
```
jparrill, line 2903: Maybe you can add some additional info on error.

hypershift-operator/controllers/nodepool/nodepool_controller_test.go
```diff
@@ -2622,6 +2622,309 @@ func TestTaintsToJSON(t *testing.T) {
...
+
+			// In case there is no performance profile applied, no configmap with status is expected.
+			// Therefore, we expect the nodepool conditions to have no performance profile conditions.
+			if !tc.hasPerformanceProfileApplied {
+				for _, NodePoolCond := range performanceProfileConditions {
+					cond := FindStatusCondition(nodePool.Status.Conditions, NodePoolCond)
+					g.Expect(cond).To(BeNil())
```
jparrill, line 2910: Same as above

hypershift-operator/controllers/nodepool/nodepool_controller_test.go
```diff
@@ -2622,6 +2622,309 @@ func TestTaintsToJSON(t *testing.T) {
...
+				}
+				return
+			}
+
+			for _, NodePoolCond := range performanceProfileConditions {
+				cond := FindStatusCondition(nodePool.Status.Conditions, NodePoolCond)
+				expectCond := tc.expectedConditions[NodePoolCond]
+				g.Expect(cond).ToNot(BeNil())
```
jparrill, line 2918: More info is appreciated. Also for the other 4 bellow.

jparrill (changes requested):

hypershift-operator/controllers/nodepool/nodepool_controller.go
Tal-or: @jparrill thank you for the thorough review. 
I don't think that a missing ConfigMap that represents the PerformanceProfile status should cause the controller to throw an error, which eventually requires user intervention to make the controller happy again. 
Printing a log should be a good compromisation  IMO.
Moreover, a missing CM might imply for a more severe issue in NTO which will eventually failed at some point to apply/read the PerfromanceProfile/Tuned objects. For such case Hypershift controller will throw an error.

That being said, I'm completely open to changes and I would appreciate your thought about the topic.

Tal-or:
/hold cancel
#https://github.com/openshift/cluster-node-tuning-operator/pull/1089 got merged

Tal-or:
@jparrill All comments has been addressed, thank you!
PTAL again when possible.
About: https://github.com/openshift/hypershift/pull/4020#discussion_r1688270194 we can discuss about it, I'm open for any other solution

hypershift-operator/controllers/nodepool/nodepool_controller.go
jparrill: I would log it as an `Error` instead of `Info` maybe not forcing the controller to fail and get blocked, but coming back returning `nil` and continuing with the reconcile loop until the user adds the needed things to make the operator to work normally. WDYT?

jparrill:
Thank you for the changes @rbaturov I think we will need another 👀 to got it merge but looks good now.
/approve

hypershift-operator/controllers/nodepool/nodepool_controller.go
Tal-or: @jparrill works for me. 
Once the ConfigMap gets created it will trigger the reconciliation loop.

Tal-or: Done.

Tal-or:
> Thank you for the changes @rbaturov I think we will need another 👀 to got it merge but looks good now. /approve

Thank you @jparrill please add the relevant members for the review. 
We need to make sure this PR goes in before OCP CF

jparrill:
Sure, could someone PTAL? @sjenning @enxebre @csrwng

PR #4021, opened 2024-05-13
HOSTEDCP-1655: Don't Ignore SubnetID and Network Security Group ID flags when values are passed through

bryan-cox:
Doesn't ignore the SubnetID and Network Security Group ID flags when values are passed in

Fixes [HOSTEDCP-1655](https://issues.redhat.com/browse/HOSTEDCP-1655)

**Checklist**

### DIFF
--- a/cmd/infra/azure/create.go
+++ b/cmd/infra/azure/create.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	ctrl "sigs.k8s.io/controller-runtime"
 	"strings"
 	"time"
 
@@ -14,8 +15,6 @@ import (
 
 	"github.com/openshift/hypershift/cmd/log"
 	"github.com/openshift/hypershift/cmd/util"
-	"github.com/openshift/hypershift/support/azureutil"
-
 	utilrand "k8s.io/apimachinery/pkg/util/rand"
 	"k8s.io/utils/ptr"
 	"sigs.k8s.io/yaml"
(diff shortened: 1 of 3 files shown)


cmd/infra/azure/create.go
```diff
@@ -205,9 +191,37 @@ func (o *CreateInfraOptions) Run(ctx context.Context, l logr.Logger) (*CreateInf
...
+
+		subnetID = *vnet.Properties.Subnets[0].ID
+		l.Info("Successfully retrieved existing subnet", "ID", subnetID)
+	}
+	result.SubnetID = subnetID
+
+	// Extract network security group ID from the VNET ID if the network security group ID was not passed through the flag
+	networkSecurityGroupID := o.NetworkSecurityGroupID
+	if len(networkSecurityGroupID) <= 0 || len(result.SecurityGroupID) <= 0 {
+		vnet, err := azureutil.GetVnetInfoFromVnetID(ctx, result.VNetID, subscriptionID, azureCreds)
```
muraee, line 213: if both subnetID and NetworkSecurityGroupID are not set, we are going to call this `GetVnetInfoFromVnetID` API twice.
I think the check whether to call this API or not should should consider both, i.e.:
```go
if len(subnetID) == 0 || len(NetworkSecurityGroupID) == 0 {
  vnet, err := azureutil.GetVnetInfoFromVnetID(ctx, result.VNetID, subscriptionID, azureCreds)

  if len(subnetID) == 0 {
    subnetID = *vnet.Properties.Subnets[0].ID
  }

  if len(NetworkSecurityGroupID) == 0  {
    networkSecurityGroupID = *vnet.Properties.Subnets[0].Properties.NetworkSecurityGroup.ID
  }
}
```

PR #4027, opened 2024-05-13
HOSTEDCP-1561: Move HCP Product CLI to STS

Patryk-Stefanski:

Fixes #

**Checklist**


cmd/cluster/aws/create.go
```diff
@@ -252,8 +260,18 @@ func IsRequiredOption(flag string, value string) error {
 // not empty; validates if the credentials secret is not empty, that it can be retrieved
 func ValidateCreateCredentialInfo(opts *core.CreateOptions) error {
 	if len(opts.CredentialSecretName) == 0 {
-		if err := IsRequiredOption("aws-creds", opts.AWSPlatform.AWSCredentialsFile); err != nil {
-			return err
+		if opts.AWSPlatform.AWSCredentialsFile == "" {
```
stevekuznetsov, line 263: We should throw an error if both `--aws-creds` and `--sts-creds` are passed

cmd/cluster/aws/destroy.go
```diff
@@ -21,11 +21,15 @@ func NewDestroyCommand(opts *core.DestroyOptions) *cobra.Command {
 
 	opts.AWSPlatform = core.AWSPlatformDestroyOptions{
 		AWSCredentialsFile: "",
+		RoleArn:            "",
```
stevekuznetsov, line 24: `""` is the zero-value for a string, and these struct literals call out the places where we have meaningful defaults - so let's not put anything in there that is just the zero-value, as it takes away from the ability to grok where our flags have meaningful defaults

cmd/consolelogs/aws/getlogs.go
```diff
@@ -44,10 +50,18 @@ func NewCommand() *cobra.Command {
 	cmd.Flags().StringVarP(&opts.Namespace, "namespace", "n", opts.Namespace, "A cluster namespace")
 	cmd.Flags().StringVar(&opts.Name, "name", opts.Name, "A cluster name")
 	cmd.Flags().StringVar(&opts.AWSCredentialsFile, "aws-creds", opts.AWSCredentialsFile, "Path to an AWS credentials file (required)")
+	cmd.Flags().StringVar(&opts.RoleArn, "role-arn", opts.RoleArn, "The ARN of the role to assume when retrieving logs.")
+	cmd.Flags().StringVar(&opts.StsCredentialsFile, "sts-creds", opts.StsCredentialsFile, "Path to STS credentials file to use when assuming the role..")
 	cmd.Flags().StringVar(&opts.OutputDir, "output-dir", opts.OutputDir, "Directory where to place console logs (required)")
 
 	cmd.MarkFlagRequired("name")
-	cmd.MarkFlagRequired("aws-creds")
+	if opts.AWSCredentialsFile == "" {
```
stevekuznetsov, line 58: This code runs before the flags are parsed, so this will always be true. Prefer to validate the _data_ (in the options struct) over the flags.

cmd/consolelogs/aws/getlogs.go
```diff
@@ -74,7 +88,16 @@ func (o *ConsoleLogOpts) Run(ctx context.Context) error {
...
-	awsSession := awsutil.NewSession("cli-console-logs", o.AWSCredentialsFile, o.AWSKey, o.AWSSecretKey, region)
+	var awsSession *session.Session
+	if o.RoleArn != "" && o.StsCredentialsFile != "" {
+		var err error
+		awsSession, err = awsutil.NewStsSession("cli-console-logs", o.StsCredentialsFile, o.RoleArn, region)
+		if err != nil {
+			return fmt.Errorf("failed to create STS session: %w", err)
+		}
+	} else {
+		awsSession = awsutil.NewSession("cli-console-logs", o.AWSCredentialsFile, o.AWSKey, o.AWSSecretKey, region)
```
stevekuznetsov, line 99: I'm surprised this can't error - why?

cmd/infra/aws/create_cli_role.go
```diff
@@ -0,0 +1,246 @@
...
+	"github.com/aws/aws-sdk-go/service/iam"
+	"github.com/aws/aws-sdk-go/service/sts"
+	"github.com/spf13/cobra"
+
+	awsutil "github.com/openshift/hypershift/cmd/infra/aws/util"
+	"github.com/openshift/hypershift/cmd/log"
+)
+
+const (
+	cliRolePolicy = `{
```
stevekuznetsov, line 17: Didn't @muraee just add this?

cmd/infra/aws/create_cli_role.go
```diff
@@ -0,0 +1,246 @@
...
+	roleArn, err := createIAMRoleOpts.CreateRoleWithInlinePolicy(ctx, iamClient)
+	if err != nil {
+		return err
+	}
+
+	fmt.Printf("Sucessfully created/updated role %s, arn: %s\n", o.RoleName, roleArn)
+	return nil
+}
+
+func assumeRoleTrustPolicy(ctx context.Context, client *sts.STS) (string, error) {
```
stevekuznetsov, line 215: This looks like it has a lot in common with `iam.go` - let's refactor to share where we can, please.

cmd/infra/aws/create_cli_role.go
```diff
@@ -0,0 +1,246 @@
...
+				},
+				"Action": "sts:AssumeRole"
+			}
+		]
+	}`, *identity.Arn)
+
+	return assumeRolePolicy, nil
+}
+
+func (o *CreateCLIRoleOptions) ParseAdditionalTags() ([]*iam.Tag, error) {
```
stevekuznetsov, line 237: This looks like a copy-pasta from `iam.go` - please DRY it out

cmd/infra/aws/create_iam.go
```diff
@@ -155,7 +169,17 @@ func (o *CreateIAMOptions) CreateIAM(ctx context.Context, client crclient.Client
 		return nil, err
 	}
 
-	awsSession := awsutil.NewSession("cli-create-iam", o.AWSCredentialsFile, o.AWSKey, o.AWSSecretKey, o.Region)
+	var awsSession *session.Session
```
stevekuznetsov, line 172: The sheer number of places where you need to add these new flags and validation is too high to be tenable. Let's break out some

```go
type AWSCredentialOptions struct {
    RoleARN string
    STSCreds string
    // whatever else
}

func DefaultOptions() *AWSCredentialOptions {
    return &AWSCredentialOptions{/* place any meaningful defaults here */}
}

func BindOptions(opts *AWSCredentialOptions, flags *pflag.FlagSet) {
    flags.StringVar(&opts.RoleArn, "role-arn", opts.RoleArn, "The ARN of the role to assume when creating the iam.")
    flags.StringVar(&opts.StsCredentialsFile, "sts-creds", opts.StsCredentialsFile, "Path to STS credentials file to use when assuming the role.")
}

func (opts *AWSCredentialOptions) Validate() error {
    /* add validation here */
}
```

Then, embed this struct into the others where needed and call `BindOptions()` on it where flags are being added.

cmd/infra/aws/iam.go
```diff
@@ -686,6 +655,51 @@ func (o *CreateIAMOptions) CreateWorkerInstanceProfile(client iamiface.IAMAPI, p
 	return nil
 }
 
+type CreateIAMRoleOptions struct {
+	RoleName          string
+	TrustPolicy       string
+	PermissionsPolicy string
+
+	additionalIAMTags []*iam.Tag
```
stevekuznetsov, line 663: why is this not exported?

cmd/infra/aws/util/util.go
```diff
@@ -31,6 +33,50 @@ func NewSession(agent string, credentialsFile string, credKey string, credSecret
 	return awsSession
 }
 
+func NewStsSession(agent, credentialsFile, roleArn, region string) (*session.Session, error) {
```
stevekuznetsov, line 36: Looks like `NewSession()` calls `session.Must()` - either refactor `NewSession()` to return an error or follow its lead and panic in `NewStsSession()` - having parallel constructs that have similar DX helps a lot.

cmd/infra/aws/create_cli_role.go
Patryk-Stefanski: yes he did, This PR wasn't meant to be reviewed yet (but I always welcome an early review :) ) and I was using his work for testing. So this comment and the following comments should have been left on his pr. In this case Im more than happy to make any changes that @muraee approves.

cmd/infra/aws/create_cli_role.go
Patryk-Stefanski: same as https://github.com/openshift/hypershift/pull/4027#discussion_r1599624238

cmd/infra/aws/create_cli_role.go
Patryk-Stefanski: same as https://github.com/openshift/hypershift/pull/4027#discussion_r1599624238

cmd/infra/aws/iam.go
Patryk-Stefanski: same as https://github.com/openshift/hypershift/pull/4027#discussion_r1599624238

cmd/consolelogs/aws/getlogs.go
Patryk-Stefanski: because the original implementation was using a session.Must which doesn't return an error, not sure why this is.

cmd/infra/aws/create_cli_role.go
stevekuznetsov: OK. His PR already went in, let's get these following comments addressed in this PR then.

cmd/infra/aws/create_iam.go
Patryk-Stefanski: @stevekuznetsov because of the --secret-creds flag we sometimes need the roleArn without the need of stsCreds. I have created a new util validation function for validating the required flags here cmd/util/credentials_validation.go

stevekuznetsov: Let's find a factoring where we bind the flags once in the repo and validate them in one flow. If that means that the options need parameters on creation and/or validation, so be it. For all commands that want to authenticate to AWS, let's provide a package that lets them bind the flags they need and get an AWS session out the other end. A good litmus test IMHO will be that this PR adds only one line that binds `sts-creds` as a flag to a struct member.

cmd/infra/aws/util/util.go
```diff
@@ -31,6 +32,53 @@ func NewSession(agent string, credentialsFile string, credKey string, credSecret
...
 
+func NewStsSession(agent, credentialsFile, credKey, credSecretKey, sessionToken, roleArn, region string) (*session.Session, error) {
+	stsSessionOpts := session.Options{}
+
+	if credKey != "" && credSecretKey != "" && sessionToken != "" {
+		stsSessionOpts.Config.Credentials = credentials.NewStaticCredentials(credKey, credSecretKey, sessionToken)
+	}
+
+	if credentialsFile != "" {
+		stsSessionOpts.SharedConfigFiles = append(stsSessionOpts.SharedConfigFiles, credentialsFile)
```
muraee, line 43: this is expecting an aws credentials file, I think it should expect the output of `aws sts get-session-token --output json` instead and then parse it and extract necessary fields.
Otherwise, the use will have to do 1 extra step to format the sts-creds file before passing it to the cli. If we do this, we can validate that the session_token exist and throw an error/warning that the user is not using temporary credentials.

cmd/infra/aws/create_iam.go
Patryk-Stefanski: @stevekuznetsov Just reading over the PR and I agree with you but since I'm leaving for PTO today and this feature is time sensitive I think this would be more of nice to have in this scenario. @muraee I don't know if you have the time to look into this approach or maybe this can be a separate Jira in our backlog?

stevekuznetsov: I can't speak to timelines but I think trending in the direction I outline here will help us immensely with maintenance of the codebase.

csrwng: @Patryk-Stefanski given that you'll be out, is it ok if we get someone else to take over the PR after today? I don't think @stevekuznetsov's are that big, but agree they would help with the maintenance of the code.

Patryk-Stefanski: @csrwng I have added @muraee as contributor and I can add anyone else aswell just let me know.

cmd/infra/aws/util/sts.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/request"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"k8s.io/utils/ptr"
+
+	supportawsutil "github.com/openshift/hypershift/support/awsutil"
+)
+
+func NewStsSession(agent, rolenArn, region string, assumeRoleCreds *credentials.Credentials) (*session.Session, error) {
```
csrwng, line 17: nit: s/Sts/STS/

cmd/infra/aws/util/sts.go
```diff
@@ -0,0 +1,70 @@
...
+	})
+	return awsSession, nil
+}
+
+func ParseSTSCredentialsFile(credentialsFile string) (*credentials.Credentials, error) {
+	var stsCreds struct {
+		Credentials Credentials `json:"Credentials"`
+	}
+
+	rawStsCreds, err := os.ReadFile(credentialsFile)
```
csrwng, line 47: nit: s/Sts/STS/

cmd/infra/aws/util/util.go
```diff
@@ -1,18 +1,89 @@
...
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 )
 
-func NewSession(agent string, credentialsFile string, credKey string, credSecretKey string, region string) *session.Session {
+type AWSCredentialsOptions struct {
+	AWSCredentialsFile string
+
+	RoleArn            string
+	StsCredentialsFile string
```
csrwng, line 23: nit: s/Sts/STS/

csrwng (commented):
Just some naming nits, otherwise lgtm

cmd/cluster/aws/destroy.go
```diff
@@ -138,20 +138,26 @@ func DestroyCluster(ctx context.Context, o *core.DestroyOptions) error {
...
 			return err
 		}
-	} else {
-		// Check the secret exists now, otherwise stop
-		opts.Log.Info("Retrieving credentials secret", "namespace", opts.Namespace, "name", opts.CredentialSecretName)
-		if _, err := util.GetSecret(opts.CredentialSecretName, opts.Namespace); err != nil {
+		return nil
+	}
+
+	if opts.AWSPlatform.AWSCredentialsOpts.AWSCredentialsFile == "" {
```
csrwng, line 151: Is there a reason we're doing this check again? Wouldn't Validate above catch it?

muraee: Validate above is executed when the CredentialsSecret is not set.

If CredentialsSecret is set we want to only check that `role-arn` is required, but only if AWSCredentialsFile is not set. i.e. AWSCredentialsFile  has priority over credentialSecret (this is the current behavior, I can change it if it doesn't make sense)

cmd/cluster/core/create.go
```diff
@@ -167,7 +168,7 @@ type KubevirtPlatformCreateOptions struct {
 }
 
 type AWSPlatformOptions struct {
-	AWSCredentialsFile      string
+	AWSCredentialsOpts      awsutil.AWSCredentialsOptions
```
stevekuznetsov, line 171: You do not need to change this now, but FWIW if you had embedded the `awsutil.AWSCredentialsOptions` it would have required a much smaller diff (and a lot less typing!)

cmd/cluster/aws/create.go
```diff
@@ -48,11 +46,13 @@ func NewCreateCommand(opts *core.CreateOptions) *cobra.Command {
...
 	cmd.Flags().StringVar(&opts.AWSPlatform.EtcdKMSKeyARN, "kms-key-arn", opts.AWSPlatform.EtcdKMSKeyARN, "The ARN of the KMS key to use for Etcd encryption. If not supplied, etcd encryption will default to using a generated AESCBC key.")
 	cmd.Flags().BoolVar(&opts.AWSPlatform.EnableProxy, "enable-proxy", opts.AWSPlatform.EnableProxy, "If a proxy should be set up, rather than allowing direct internet access from the nodes")
-	cmd.Flags().StringVar(&opts.CredentialSecretName, "secret-creds", opts.CredentialSecretName, "A Kubernetes secret with needed AWS platform credentials: aws-creds, pull-secret, and a base-domain value. The secret must exist in the supplied \"--namespace\". If a value is provided through the flag '--pull-secret', that value will override the pull-secret value in 'secret-creds'.")
+	cmd.Flags().StringVar(&opts.CredentialSecretName, "secret-creds", opts.CredentialSecretName, "A Kubernetes secret with needed AWS platform credentials: sts-creds, pull-secret, and a base-domain value. The secret must exist in the supplied \"--namespace\". If a value is provided through the flag '--pull-secret', that value will override the pull-secret value in 'secret-creds'.")
```
stevekuznetsov, line 49: Changing what we expect in this secret is breaking for any existing users of the CLI that are using the flag and secret today - if we have such users, this is not safe to do.

stevekuznetsov (commented):
Looks good! Couple small comments

cmd/cluster/aws/create.go
muraee: this is used by ACM, we already communicated the new changes with them.

cmd/cluster/aws/destroy.go
```diff
@@ -138,20 +138,26 @@ func DestroyCluster(ctx context.Context, o *core.DestroyOptions) error {
 	return core.DestroyCluster(ctx, hostedCluster, o, destroyPlatformSpecifics)
 }
 
-// ValidateCredentialInfo validates if the credentials secret name is empty, the aws-creds is not empty; validates if
+// ValidateCredentialInfo validates if the credentials secret name is empty, the aws-creds or sts-creds mutually exclusive and are not empty; validates if
 // the credentials secret is not empty, that it can be retrieved.
 func ValidateCredentialInfo(opts *core.DestroyOptions) error {
```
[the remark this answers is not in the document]
muraee: `ValidateCreateCredentialInfo` also validates the pull-secret while this one doesn't, so I don't see how we can share the func easily.

cmd/infra/aws/util/util.go
```diff
@@ -1,18 +1,89 @@
...
+	}
+	if err := util.ValidateRequiredOption("role-arn", opts.RoleArn); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (opts *AWSCredentialsOptions) BindFlags(flags *pflag.FlagSet) {
+	flags.StringVar(&opts.AWSCredentialsFile, "aws-creds", opts.AWSCredentialsFile, "Path to an AWS credentials file")
+	flags.StringVar(&opts.RoleArn, "role-arn", opts.RoleArn, "The ARN of the role to assume.")
```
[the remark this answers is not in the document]
muraee: how would you do that? one marks the flags as required, the other one doesn't

cmd/cluster/aws/create.go
stevekuznetsov: What is the path for this not to be breaking, then? They will have to update the secret to contain both values before this merges and is rolled out?

cmd/cluster/aws/destroy.go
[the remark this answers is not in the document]
stevekuznetsov: What about

```go
func ValidateCreateCredentialInfo(opts *core.CreateOptions) error {
    if err := ValidateAWSCredentialInfo(opts.AWSPlatform.AWSCredentialsOpts, opts.CredentialSecretName); err != nil {
        return err
    } 
    
    if len(opts.CredentialSecretName) == 0 {
        if err := util.ValidateRequiredOption("pull-secret", opts.PullSecretFile); err != nil {
            return err
        }
    }
    return nil
}
```

cmd/infra/aws/util/util.go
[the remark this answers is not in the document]
stevekuznetsov: I would personally not bother with the `cobra.MarkFlagRequired` business - you need to have the correct logic in `Validate()` anyway and with the programmatic re-use of the `Options` struct in e2e, so it's not really buying us anything.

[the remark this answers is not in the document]
muraee: `cobra.MarkFlagRequired` also helps prioritize those fields with auto completion. But not very important.
Addressed.

cmd/cluster/aws/destroy.go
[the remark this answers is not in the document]
muraee: addressed

cmd/cluster/aws/create.go
muraee: we need this to merge today on our side. I don't how/when they roll out new versions.

stevekuznetsov (approved):
/lgtm

PR #4029, opened 2024-05-14
HOSTEDCP-1556: Support Priority based expander for cluster-autoscaler

muraee:

Fixes #[HOSTEDCP-1556](https://issues.redhat.com/browse/HOSTEDCP-1556)

**Checklist**

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/autoscaler/reconcile.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/autoscaler/reconcile.go
@@ -43,6 +43,7 @@ func ReconcileAutoscalerDeployment(deployment *appsv1.Deployment, hcp *hyperv1.H
 	}
 
 	args := []string{
+		"--expander=priority,least-waste",
 		"--cloud-provider=clusterapi",
 		"--node-group-auto-discovery=clusterapi:namespace=$(MY_NAMESPACE)",
 		"--kubeconfig=/mnt/kubeconfig/target-kubeconfig",


control-plane-operator/controllers/hostedcontrolplane/autoscaler/reconcile.go
```diff
@@ -43,6 +43,7 @@ func ReconcileAutoscalerDeployment(deployment *appsv1.Deployment, hcp *hyperv1.H
 	}
 
 	args := []string{
+		"--expander=priorty,least-waste",
```
enxebre, line 46: I'd expect this to be something we expose in hc.spec.ClusterAutoscaling and let it be a consumer choice.

muraee: this option will be ignored if the priority configMap is not present. So the consumer choice would be to either create the configMap to use this feature or not to use the default behavior

enxebre: there's a typo in "priorty".

>this option will be ignored if the priority configMap is not present. So the consumer choice would be to either create the configMap to use this feature or not to use the default behavior

Yes, this enables opt-in for priority expander and fallback to least-waste. However It doesn't let a HC consumer to choose different [expanders](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-are-expanders).
I'd expect us to expose expanders in the API now to be more flexible about consumer choices e.g [cao](https://github.com/openshift/cluster-autoscaler-operator/blob/master/pkg/apis/autoscaling/v1/clusterautoscaler_types.go#L87)
That said we can always open up later and let this be our default.

muraee: I see, makes sense.
I will create another ticket to add support for other expanders through the API.

> there's a typo in "priorty".

fixed, thanks!

enxebre:
known issue failing kubevirt
```
    --- FAIL: TestCreateCluster/Teardown (29.55s)
        --- FAIL: TestCreateCluster/Teardown/ValidateMetricsAreExposed (0.04s)
panic: prometheus service account not found [recovered]
	panic: prometheus service account not found
```

/hold cancel 
/cherry-pick release-4.16

PR #4034, opened 2024-05-14
OCPBUGS-33659: Improve resliency of size tagging when hostedcluster KAS down

csrwng:
When the kube apiserver of a hosted cluster is not available, the replica status of nodepools will not be accurate because the CAPI controllers can no longer get node counts from the API server. This commit improves the handling of this situation with 2 changes:
- Switches to use .spec.replicas to determine node count of nodepools that do not have autoscaling turned on.
- Once a hosted cluster has been tagged with a size, only if the kube apiserver of the hosted cluster is available is the hosted cluster allowed to move to a different size.
- 
Fixes #[OCPBUGS-33659](https://issues.redhat.com/browse/OCPBUGS-33659)

**Checklist**

### DIFF
--- a/hypershift-operator/controllers/hostedclustersizing/hostedclustersizing_controller.go
+++ b/hypershift-operator/controllers/hostedclustersizing/hostedclustersizing_controller.go
@@ -15,6 +15,7 @@ import (
 	hyperutil "github.com/openshift/hypershift/support/util"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -218,6 +219,12 @@ func (r *reconciler) reconcile(
 		return nil, fmt.Errorf("failed to determine if HCCO reports node count: %w", err)
 	}
 
+	// Determine if the Kube API Server is available to determine if we can trust the node count from nodepool.status.replicas
+	// If the Kube API Server is not available, we cannot trust the node count from nodepool.status.replicas
+	// Ref: kubernetes-sigs/cluster-api#10195
+	kasAvailableCondition := meta.FindStatusCondition(hostedCluster.Status.Conditions, string(hypershiftv1beta1.KubeAPIServerAvailable))
+	kasAvailable := kasAvailableCondition != nil && kasAvailableCondition.Status == metav1.ConditionTrue
+
 	// first, we figure out the node count for the hosted cluster
 	var nodeCount uint32
 	if hccoReportsNodeCount {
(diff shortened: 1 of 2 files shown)


hypershift-operator/controllers/hostedclustersizing/hostedclustersizing_controller_test.go
```diff
@@ -106,9 +106,9 @@ func TestSizingController_Reconcile(t *testing.T) {
 			},
 			nodePoolsForHostedCluster: func(_ context.Context, _ *hypershiftv1beta1.HostedCluster) (*hypershiftv1beta1.NodePoolList, error) {
 				return &hypershiftv1beta1.NodePoolList{Items: []hypershiftv1beta1.NodePool{
-					{Status: hypershiftv1beta1.NodePoolStatus{Replicas: 10}},
```
stevekuznetsov, line 109: I would expect to see a new unit test with autoscaling off, using spec, and this existing test would continue to use status and not have autoscaling on.

hypershift-operator/controllers/hostedclustersizing/hostedclustersizing_controller_test.go
```diff
@@ -106,9 +106,9 @@ func TestSizingController_Reconcile(t *testing.T) {
...
-					{Status: hypershiftv1beta1.NodePoolStatus{Replicas: 3}},
```
stevekuznetsov, line 110: I would also expect to see a new unit test for a no-op when we _want_ to transition but the KAS condition is not set, leading us to not transition

hypershift-operator/controllers/hostedclustersizing/hostedclustersizing_controller.go
```diff
@@ -216,7 +217,9 @@ func (r *reconciler) reconcile(
 
 	// first, we figure out the node count for the hosted cluster
 	var nodeCount uint32
+	nodeCountRequiresAPIServer := false
```
enxebre, line 220: Can we include a comment for future code readers? e.g. "When the node count comes from status.replicas we only trust it if the kas is available, otherwise we might get false 0s, https://github.com/kubernetes-sigs/cluster-api/issues/10195".

hypershift-operator/controllers/hostedclustersizing/hostedclustersizing_controller.go
```diff
@@ -216,7 +217,9 @@ func (r *reconciler) reconcile(
...
 	if hccoReportsNodeCount {
+		nodeCountRequiresAPIServer = true
```
enxebre, line 222: should we also let the HCCO to compute nodecount from the NodePools spec?
Even without facing this issue wouldn't that make sense from the autoscaling functionallity pov?
If so we could change existing hostedControlPlane.Status.NodeCount or maybe have a new one e.g. hostedControlPlane.Status.NodeCountIntent

hypershift-operator/controllers/hostedclustersizing/hostedclustersizing_controller.go
```diff
@@ -232,7 +235,25 @@ func (r *reconciler) reconcile(
 		}
 
 		for _, nodePool := range nodePools.Items {
-			nodeCount += uint32(nodePool.Status.Replicas)
+			var replicas uint32
+			// If autoscaling, the replicas should be returned from status
+			if nodePool.Spec.AutoScaling != nil {
+				nodeCountRequiresAPIServer = true
```
enxebre, line 241: Alternatively to use `nodeCountRequiresAPIServer = true` here and in l222

We could also fetch the condition first and return early when needed. E.g.
Then in both places we'd do
```
if !kasAvailable && replicas == 0 { 
  log
  return nil, nil
}
```
Just an alternative thought, feel free to ignore it.

hypershift-operator/controllers/hostedclustersizing/hostedclustersizing_controller.go
csrwng: Actually scratch that. In the cases where we don't have nodepools, but nodes exist (IBM Cloud), falling back on the nodepools will give us the wrong answer.

stevekuznetsov (commented):
/lgtm

stevekuznetsov (commented):
/lgtm

csrwng:
CI is currently flaking on cloud quota issues, and this PR does not impact any codepath tested by CI

PR #4036, opened 2024-05-15
HOSTEDCP-1542: cmd: add an option to render into a file, use it in e2e

stevekuznetsov:
The closest thing we have to testing the command-line is our end-to-end code that re-uses the options structs and runs them through the logic that the command-line does. We can capture a lot of the work done in the command-line by inspecting the output manifests, which is easiest to do by rendering them as a test artfact.

Helps to test https://github.com/openshift/hypershift/pull/4018


cmd/cluster/cluster.go
```diff
@@ -62,7 +62,8 @@ func NewCreateCommands() *cobra.Command {
 	cmd.PersistentFlags().StringVar(&opts.ReleaseStream, "release-stream", opts.ReleaseStream, "The OCP release stream for the cluster (e.g. 4.15.0-0.nightly), this flag is ignored if release-image is set")
 	cmd.PersistentFlags().StringVar(&opts.PullSecretFile, "pull-secret", opts.PullSecretFile, "File path to a pull secret.")
 	cmd.PersistentFlags().StringVar(&opts.ControlPlaneAvailabilityPolicy, "control-plane-availability-policy", opts.ControlPlaneAvailabilityPolicy, "Availability policy for hosted cluster components. Supported options: SingleReplica, HighlyAvailable")
-	cmd.PersistentFlags().BoolVar(&opts.Render, "render", opts.Render, "Render output as YAML to stdout instead of applying")
+	cmd.PersistentFlags().BoolVar(&opts.Render, "render", opts.Render, "Render output as YAML instead of applying")
+	cmd.PersistentFlags().StringVar(&opts.RenderInto, "render-into", opts.RenderInto, "Render output as YAML into this file instead of applying. If unset, YAMl will be output to stdout.")
```
bryan-cox, line 66:
```suggestion
	cmd.PersistentFlags().StringVar(&opts.RenderInto, "render-into", opts.RenderInto, "Render output as YAML into this file instead of applying. If unset, YAML will be output to stdout.")
```

bryan-cox: ^nit

product-cli/cmd/cluster/cluster.go
```diff
@@ -70,7 +70,8 @@ func NewCreateCommands() *cobra.Command {
 	cmd.PersistentFlags().StringVar(&opts.NetworkType, "network-type", opts.NetworkType, "Enum specifying the cluster SDN provider. Supports either Calico, OVNKubernetes, OpenShiftSDN or Other.")
 	cmd.PersistentFlags().StringVar(&opts.PullSecretFile, "pull-secret", opts.PullSecretFile, "Filepath to a pull secret.")
 	cmd.PersistentFlags().StringVar(&opts.ReleaseImage, "release-image", opts.ReleaseImage, "The OCP release image for the HostedCluster.")
-	cmd.PersistentFlags().BoolVar(&opts.Render, "render", opts.Render, "Renders the HostedCluster manifest output as YAML to stdout instead of automatically applying the manifests to the management cluster.")
+	cmd.PersistentFlags().BoolVar(&opts.Render, "render", opts.Render, "Render output as YAML instead of applying")
+	cmd.PersistentFlags().StringVar(&opts.RenderInto, "render-into", opts.RenderInto, "Render output as YAML into this file instead of applying. If unset, YAMl will be output to stdout.")
```
bryan-cox, line 74:
```suggestion
	cmd.PersistentFlags().StringVar(&opts.RenderInto, "render-into", opts.RenderInto, "Render output as YAML into this file instead of applying. If unset, YAML will be output to stdout.")
```

bryan-cox: ^nit

bryan-cox (commented):
lgtm, just had a few very minor nits

test/e2e/util/hypershift_framework.go
```diff
@@ -258,7 +260,23 @@ func (h *hypershiftTest) createHostedCluster(opts *core.CreateOptions, platform
...
+	// Dump the output from rendering the cluster objects for posterity
+	dumpDir := filepath.Join(artifactDir, artifactSubdirFor(h.T))
+	if err := os.MkdirAll(dumpDir, 0755); err != nil {
+		h.Errorf("failed to create dump directory: %v", err)
+	}
+
+	opts.Render = true
+	opts.RenderInto = filepath.Join(dumpDir, "manifests.yaml")
+	h.Logf("Dumping new cluster manifests to %s", opts.RenderInto)
+	if err := createCluster(h.ctx, hc, opts); err != nil {
```
muraee, line 272: `createCluster()` also creates infra, so calling it just to render the manifest, will create extra unneeded infra which will also not be cleaned up.

muraee:
why do we even need this? all resources created by the cli are already part of the dump. What is the added value of having them in a single manifest. You will mainly be looking only at the HostedCluster / NodePool resources, which are easily found in the dump,

stevekuznetsov:
> why do we even need this? 

I wrote a PR that changes how the CLI works. I want to test that, and to do that I wanted to capture what the CLI was going to do by dumping the objects it would make, then we can assert over that output. I don't want post-creation modifications, etc, to influence this. Can you think of another way to capture it?

test/e2e/util/hypershift_framework.go
stevekuznetsov: Interesting, I was hoping that it would be deterministic and re-use the infra the second time around.

muraee:
> I don't want post-creation modifications, etc, to influence this
the HO wouldn't modify the spec of the HostedCluster or NodePool, and the API defaults would only apply if the fields are not set by the CLI. So we would know if the CLI did something wrong.

test/e2e/util/hypershift_framework.go
stevekuznetsov: Updated

stevekuznetsov:
@muraee I want to test the CLI. While it's possible that nothing touches these manifests after the fact, having to a) know what the list of manifests is, b) collect them from the larger artifact dir, c) know which fields to ignore and when is much more difficult than simply asserting over the output of the CLI.

csrwng:

stevekuznetsov:
```
   * could not run steps: step [input:root] failed: failed to wait for importing imagestreamtags on ci-op-lbq4w3i0/pipeline: failed to reimport the tag ci-op-lbq4w3i0/pipeline:ocp_4.16_base-rhel9: unable to import tag ci-op-lbq4w3i0/pipeline:ocp_4.16_base-rhel9 with message Internal error occurred: [dockerimage.image.openshift.io "quay.io/openshift/ci:ci-op-lbq4w3i0_pipeline_ocp_4.16_base-rhel9" not found, dockerimage.image.openshift.io "quay-proxy.ci.openshift.org/openshift/ci:ci-op-lbq4w3i0_pipeline_ocp_4.16_base-rhel9" not found] on the image stream even after (6) imports: timed out waiting for the condition 
```

/retest

stevekuznetsov:
holy moly we did it

PR #4037, opened 2024-05-15
OCPBUGS-33729: Do not ignore single placeholder pods for a specific hosted cluster

csrwng:
The current logic that looks at placeholder pods ignores unpaired pods because in the case of generic placeholders, the pods are likely in the middle of getting rolled out by their respective deployment. However, unpaired pods that correspond to a specific hosted cluster (have a pair node selector) can exist by themselves if there's already an existing node for the cluster. This commit updates the logic to not ignore single placeholder pods for specific hosted clusters.

Fixes #[OCPBUGS-33729](https://issues.redhat.com/browse/OCPBUGS-33729)

**Checklist**

### DIFF
--- a/hypershift-operator/controllers/scheduler/autoscaler.go
+++ b/hypershift-operator/controllers/scheduler/autoscaler.go
@@ -540,8 +540,9 @@ type podPair struct {
 	p2 corev1.Pod
 }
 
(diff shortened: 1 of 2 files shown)


stevekuznetsov (commented):
/lgtm

PR #4045, opened 2024-05-16
OCPBUGS-33802: don't delay new cluster creation

stevekuznetsov:
hostedclustersizing: don't enforce delays on brand-new clusters

Before, we would make brand-new clusters sit through the transition
delays erroneously. Now, we allow them to take on their first size
without any delay.

---

hostedclustersizing: don't enforce concurrency on new clusters

Before, we would enforce global concurrency limits on brand new clusters
joining the fleet and being given a size. This is not tenable with an
SLA for cluster bringup, so we no longer do that.

---

### DIFF
--- a/hypershift-operator/controllers/hostedclustersizing/hostedclustersizing_controller.go
+++ b/hypershift-operator/controllers/hostedclustersizing/hostedclustersizing_controller.go
@@ -346,30 +346,35 @@ func (r *reconciler) reconcile(
 		}
 	}
 
-	hostedClusters, err := r.listHostedClusters(ctx)
-	if err != nil {
-		return nil, fmt.Errorf("failed to list hosted clusters when calculating concurrency: %w", err)
-	}
+	// For new clusters being added to the fleet, we have an SLA on creation time and can't afford to delay
+	// the first transition, as it is required for the control plane to schedule. For other clusters, though,
+	// we want to limit the amount of churn happening in order to promote the stability of the management plane.
+	if scheduled := hostedCluster.Annotations[hypershiftv1beta1.HostedClusterScheduledAnnotation]; scheduled == "true" {
+		hostedClusters, err := r.listHostedClusters(ctx)
+		if err != nil {
+			return nil, fmt.Errorf("failed to list hosted clusters when calculating concurrency: %w", err)
+		}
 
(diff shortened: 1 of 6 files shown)


hypershift-operator/controllers/hostedclustersizing/hostedclustersizing_controller.go
```diff
@@ -328,30 +328,35 @@ func (r *reconciler) reconcile(
...
 	}
 
-	hostedClusters, err := r.listHostedClusters(ctx)
-	if err != nil {
-		return nil, fmt.Errorf("failed to list hosted clusters when calculating concurrency: %w", err)
-	}
+	// For new clusters being added to the fleet, we have an SLA on creation time and can't afford to delay
+	// the first transition, as it is required for the control plane to schedule. For other clusters, though,
+	// we want to limit the amount of churn happening in order to promote the stability of the management plane.
+	if scheduled := hostedCluster.Annotations[hypershiftv1beta1.HostedClusterScheduledAnnotation]; sizeClassLabelPresent && scheduled == "true" {
```
csrwng, line 334: why `sizeClassLabelPresent`? Legacy clusters will be scheduled but not have a size label and we do want the concurrency limit to apply to them.

csrwng (commented):

hypershift-operator/controllers/hostedclustersizing/hostedclustersizing_controller.go
stevekuznetsov: oops I think this should have been an `||` ... or we can just omit `sizeClassLabelPresent`

csrwng:
Overriding as we don't test this in CI

stevekuznetsov:
hmm ...

```
$ time go test ./hypershift-operator/controllers/hostedclustersizing/... -count 100 -run "^TestSizingController_Reconcile/pending_transition,_hcco_doesn't_report_node_count$"
ok  	github.com/openshift/hypershift/hypershift-operator/controllers/hostedclustersizing	0.056s

real	0m1.304s
user	0m2.292s
sys	0m0.518s
```

Can't seem to reproduce the error.

stevekuznetsov:
Ah, needed to rebase on your KAS status PR.

PR #4051, opened 2024-05-17
TRT-1672: e2e: proceed with teardown even if dump fails

sjenning:
Currently, if `dumpCluster()` fails, the HostedCluster is not deleted, leaving behind all of its AWS resources.  This is a problem.


sjenning:
Presubs are still running again 4.16, where the ocm issue still exists
https://github.com/openshift/openshift-controller-manager/pull/309

However, that allowed us to confirm this fix, even though the job itself will fail.

/override ci/prow/e2e-aws
/override ci/prow/e2e-azure

PR #3714, opened 2024-03-08
WIP: make HC image cluster config authoritative

sjenning:

PR #4057, opened 2024-05-20
OCPBUGS-33377: Modify token secret MCS hash in place 

enxebre:
We are relying on .image and .proxy here to trigger upgrades and create a new tokenSecret https://github.com/openshift/hypershift/pull/3714/files#diff-c666a2a9dc48b0d3de33cc8fa3a054f853671f0428f9d56b263750dfc6d26e00R266
    but then we use the whole .config for the hash in the token and mcs secret https://github.com/openshift/hypershift/pull/3795/files#diff-681fe9efe998da46c20a41e51c975bac6a9b14e93b7705f40da204d30a0ddebcR49.
    Say a new NodePool is created or something triggers a rolling upgrade:
    - A new token secret gets created.
    - Before the ignition controllers watches it and can generate the payload, the hc.config changes to e.g. include an IDP.
    - The MCS secret hc-configuration-hash is updated accordingly.
    - The token secret hc-configuration-hash is not updated, because that only happens on token secret creation.
    - The ignition controller tries to generate the payload and it fails perpetually with https://github.com/openshift/hypershift/pull/3795/files#diff-7cc71524ad99a2a526cceabd327a67c2767dc0c49e5c4aa8629ca88b0d205072R131
    This will also be the case if the ignition server gets recreated by any reason (e.g. management cluster dataplane recycling), all the token secrets which e.g. hc.config.idp changed after they were created will fail to re-generate the payload forever.

This PR mitigates the discrepancy by letting the hash to be changed inplace in the token secret.
Follow up: Only set the hash based on the same input that triggers a new token secret creation.

Fixes #

**Checklist**

### DIFF
--- a/hypershift-operator/controllers/nodepool/nodepool_controller.go
+++ b/hypershift-operator/controllers/nodepool/nodepool_controller.go
@@ -1641,6 +1641,11 @@ func reconcileTokenSecret(tokenSecret *corev1.Secret, nodePool *hyperv1.NodePool
 		tokenSecret.Data[TokenSecretPullSecretHashKey] = []byte(supportutil.HashSimple(pullSecret))
 		tokenSecret.Data[TokenSecretHCConfigurationHashKey] = []byte(hcConfigurationHash)
 	}
+	// TODO (alberto): Only apply this on creation and change the hash generation to only use triggering upgrade fields.
+	// We let this change to happen inplace now as the tokenSecret and the mcs config use the whole spec.Config for the comparing hash.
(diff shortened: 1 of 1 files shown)


enxebre:
/hold
a bit

enxebre:
cc @sjenning

stevekuznetsov (commented):
/lgtm

jparrill:
Hey @enxebre will this be backported to 4.16, 4.15 and 4.14?

enxebre:
>Hey @enxebre will this be backported to 4.16, 4.15 and 4.14?

Posting here for visibility: This an HO change. It affects any ocp version.

PR #4060, opened 2024-05-20
HOSTEDCP-1687: Allow arm64 baremetal nodes to a nodepool in a multi-arch cluster

krhas7:
Expand supported platform to allo the addition of an arm64 baremetal node to a baremetal hosted cluster.

Fixes #

**Checklist**

### DIFF
--- a/hypershift-operator/controllers/nodepool/nodepool_controller.go
+++ b/hypershift-operator/controllers/nodepool/nodepool_controller.go
@@ -979,16 +979,12 @@ func isArchAndPlatformSupported(nodePool *hyperv1.NodePool) bool {
 		if nodePool.Spec.Arch == hyperv1.ArchitectureAMD64 || nodePool.Spec.Arch == hyperv1.ArchitectureARM64 {
 			supported = true
 		}
-	case hyperv1.NonePlatform:
-		if nodePool.Spec.Arch == hyperv1.ArchitectureAMD64 || nodePool.Spec.Arch == hyperv1.ArchitectureARM64 {
-			supported = true
-		}
 	case hyperv1.AzurePlatform, hyperv1.KubevirtPlatform:
 		if nodePool.Spec.Arch == hyperv1.ArchitectureAMD64 {
 			supported = true
 		}
 	case hyperv1.AgentPlatform:
-		if nodePool.Spec.Arch == hyperv1.ArchitectureAMD64 || nodePool.Spec.Arch == hyperv1.ArchitecturePPC64LE {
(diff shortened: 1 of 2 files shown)


hypershift-operator/controllers/nodepool/nodepool_controller_test.go
```diff
@@ -3097,6 +3097,18 @@ func TestIsArchAndPlatformSupported(t *testing.T) {
 			},
 			expect: true,
 		},
+		{
```
bryan-cox, line 3100: Can we add another case for arm for none platform and maybe an invalid case for the none platform please?

bryan-cox:
@krhas7 You'll need to link a Jira ticket to this PR as well so the right Jira label is added to get this merged.

hypershift-operator/controllers/nodepool/nodepool_controller_test.go
krhas7: Hi @bryan-cox, added the suggested tests. Please review. Thank you.

bryan-cox:
/retitle HOSTEDCP-1687: Allow arm64 baremetal nodes to a nodepool in a multi-arch cluster

bryan-cox:
/hold 

for HyperShift team review

hypershift-operator/controllers/nodepool/nodepool_controller.go
```diff
@@ -984,7 +984,7 @@ func isArchAndPlatformSupported(nodePool *hyperv1.NodePool) bool {
 			supported = true
 		}
 	case hyperv1.AgentPlatform:
```
jparrill, line 986: I would add also the `hyperv1.NonePlatform` as another case, even not being supported, I think it could be useful for dev envs.

hypershift-operator/controllers/nodepool/nodepool_controller_test.go
```diff
@@ -3097,6 +3097,18 @@ func TestIsArchAndPlatformSupported(t *testing.T) {
...
+			name: "supported platform with multiple arch baremetal - arm64",
```
jparrill, line 3101: If the upper suggestion is implemented, you will need to modify the tests to also include the NonePlatform.

jparrill (commented):
Dropped a comment, modify if makes sense, otherwise LGTM

hypershift-operator/controllers/nodepool/nodepool_controller_test.go
krhas7: Hi @jparrill added the suggested checks for None. Please review. Thanks.

PR #4069, opened 2024-05-21
[release-4.16] NO-JIRA: Bump library-go dependency

orenc1:
This is needed in order to adjust to the change in behavior releated to service accounts token secrets which is fixed https://github.com/openshift/library-go/pull/1697

Fixes #

**Checklist**


sjenning:
Looks like an actual failure in `TestCreateClusterCustomConfig` where nodes fail to join

sjenning:
KAS is failing to start with
```
E0521 18:01:49.021163       1 run.go:74] "command failed" err="error while building transformers: KMSv1 is deprecated and will only receive security updates going forward. Use KMSv2 instead.  Set --feature-gates=KMSv1=true to use the deprecated KMSv1 feature."
```

sjenning:
/hold

Hacking `KMSv1` feature gate in is not the way.  We need to figure out why the default feature set is not including it.

support/config/featuregates.go
```diff
@@ -13,9 +13,9 @@ func FeatureGates(fg *configv1.FeatureGateSelection) []string {
 		enabled = fg.CustomNoUpgrade.Enabled
 		disabled = fg.CustomNoUpgrade.Disabled
 	} else {
-		fs := configv1.FeatureSets[fg.FeatureSet]
-		for _, fgDescription := range fs.Enabled {
```
csrwng, line 17: This is wrong. The function still needs to iterate over enabled featuregates. As it is, it's only iterating over disabled ones.

orenc1: good catch.
thanks, fixed.

orenc1:
this run failed on Autoscaling test:
```
    util.go:238: 
        failed to ensure guest nodes became ready, ready: (3/2): 
        Unexpected error:
            <wait.errInterrupted>: 
            timed out waiting for the condition
            {
                cause: <*errors.errorString | 0xc000970190>{
                    s: "timed out waiting for the condition",
                },
            }
        occurred
    --- FAIL: TestAutoscaling/Main (2190.11s)
```
i think it's a flake this time.
/test e2e-aws

PR #4071, opened 2024-05-21
HOSTEDCP-1564: Ensure VNET, NSG, & Managed RG locations match

bryan-cox:
Adds validation to the HostedCluster controller to ensure the VNET resource group location, the network security resource group location, and the managed resource group location match.

Fixes [HOSTEDCP-1564](https://issues.redhat.com/browse/HOSTEDCP-1564)

**Checklist**

### DIFF
--- a/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
+++ b/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
@@ -93,6 +93,7 @@ import (
 	"github.com/openshift/hypershift/hypershift-operator/controllers/manifests/ignitionserver"
 	kvinfra "github.com/openshift/hypershift/kubevirtexternalinfra"
 	"github.com/openshift/hypershift/support/api"
+	"github.com/openshift/hypershift/support/azureutil"
 	"github.com/openshift/hypershift/support/capabilities"
 	"github.com/openshift/hypershift/support/certs"
 	"github.com/openshift/hypershift/support/config"
@@ -4220,10 +4221,12 @@ func (r *HostedClusterReconciler) validateAzureConfig(ctx context.Context, hc *h
 		return nil
 	}
 
+	// Verify the platform is at least initialized
 	if hc.Spec.Platform.Azure == nil {
 		return errors.New("azurecluster needs .spec.platform.azure to be filled")
 	}
 
+	// Verify the credentials secret contains the data fields we expect
 	credentialsSecret := &corev1.Secret{ObjectMeta: metav1.ObjectMeta{
 		Namespace: hc.Namespace,
 		Name:      hc.Spec.Platform.Azure.Credentials.Name,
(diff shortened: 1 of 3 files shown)


hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
```diff
@@ -1404,35 +1404,6 @@ func TestValidateConfigAndClusterCapabilities(t *testing.T) {
 			},
 			expectedResult: errors.New(`[credentials secret for cluster doesn't have required key AZURE_CLIENT_ID, credentials secret for cluster doesn't have required key AZURE_CLIENT_SECRET, credentials secret for cluster doesn't have required key AZURE_SUBSCRIPTION_ID, credentials secret for cluster doesn't have required key AZURE_TENANT_ID]`),
 		},
-		{
```
bryan-cox, line 1407: This would require actual Azure cloud resources setup now so I removed the test case.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -4239,7 +4242,17 @@ func (r *HostedClusterReconciler) validateAzureConfig(ctx context.Context, hc *h
 		}
 	}
 
-	return utilerrors.NewAggregate(errs)
+	if len(errs) > 0 {
```
csrwng, line 4245: why not include the error below in all the validation errors?

hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
csrwng: Not for this PR, but can we eventually use a mock implementation of the APIs to run the test?

csrwng (commented):
some comments

hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
bryan-cox: Agree, it's something we need to look into

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
bryan-cox: It looked to me that was particularly tied to the `for` loop on L4239. 

I don't have any issues adding it to that though.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -4239,7 +4242,17 @@ func (r *HostedClusterReconciler) validateAzureConfig(ctx context.Context, hc *h
...
 	}
 
-	return utilerrors.NewAggregate(errs)
+	// Verify the resource group locations match
+	err := azureutil.VerifyResourceGroupLocationsMatch(ctx, hc, credentialsSecret)
+	if err != nil {
+		errs = append(errs, err)
+	}
+
+	if len(errs) > 0 {
```
csrwng, line 4251: nit: this check is not necessary. If len(errs) == 0, NewAggregate returns nil

PR #4074, opened 2024-05-21
OCPBUGS-34095: Sync the autoscaler periodically

csrwng:
In some situations, the presence of a placeholder pod does not cause the autoscaler to reconcile. This adds a regular sync to the autoscaler regardless of pod events.

Fixes #[OCPBUGS-34095](https://issues.redhat.com/browse/OCPBUGS-34095)

**Checklist**


enxebre:
can you please elaborate on the reasons why reconciliation doesn't happen as expected?
Thoughts on the ticker vs using requeueAfter?
/lgtm

PR #4075, opened 2024-05-22
OCPBUGS-33977: Allow specifying machine health check timeout in cluster sizing config

csrwng:
Adds an optional machine healthcheck timeout field to the effects of a
size configuration. This allows setting a timeout for specific size of
clusters.

Introduces annotations to override machine health check timeout and max
unavailable values. The annotations can be set in either the
HostedCluster or a specific NodePool. With the one on the NodePool
taking precedence.

Fixes #[OCPBUGS-33977](https://issues.redhat.com/browse/OCPBUGS-33977)

**Checklist**


api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -286,6 +286,16 @@ const (
 
 	// ManagementPlatformAnnotation specifies the infrastructure platform of the underlying management cluster
 	ManagementPlatformAnnotation = "hypershift.openshift.io/management-platform"
+
+	// MachineHealthCheckTimeoutAnnotation allows overriding the default machine health check timeout for
+	// nodepools. The annotation can be set in either the HostedCluster or the NodePool. If set on both, the
+	// one on the NodePool takes precedence. The value is a go duration string with a number and a unit (ie. 8m, 1h, etc)
+	MachineHealthCheckTimeoutAnnotation = "hypershift.openshift.io/machine-health-check-timeout"
```
enxebre, line 293: Let's create a follow up jira to decide how to expose MHC in the NodePool API
https://issues.redhat.com/browse/HOSTEDCP-1680

hypershift-operator/controllers/nodepool/nodepool_controller.go
```diff
@@ -1876,13 +1876,41 @@ func taintsToJSON(taints []hyperv1.Taint) (string, error) {
...
 	// TODO (alberto): possibly expose this config at the nodePool API.
 	maxUnhealthy := intstr.FromInt(2)
+	timeOut := 8 * time.Minute
+
+	maxUnhealthyOverride := nodePool.Annotations[hyperv1.MachineHealthCheckMaxUnhealthyAnnotation]
+	if maxUnhealthyOverride == "" {
+		maxUnhealthyOverride = hc.Annotations[hyperv1.MachineHealthCheckMaxUnhealthyAnnotation]
+	}
+	if maxUnhealthyOverride != "" {
+		maxUnhealthy = intstr.FromString(maxUnhealthyOverride)
```
enxebre, line 1898: MHC will transparently fail if this value is not legit?

hypershift-operator/controllers/scheduler/dedicated_request_serving_nodes.go
```diff
@@ -640,6 +640,17 @@ func (r *DedicatedServingComponentSchedulerAndSizer) updateHostedCluster(ctx con
 	if sizeConfig.Effects != nil && sizeConfig.Effects.APICriticalPriorityClassName != nil {
 		hc.Annotations[hyperv1.APICriticalPriorityClass] = *sizeConfig.Effects.APICriticalPriorityClassName
 	}
+	if sizeConfig.Effects != nil && sizeConfig.Effects.MachineHealthCheckTimeout != nil {
+		hc.Annotations[hyperv1.MachineHealthCheckTimeoutAnnotation] = sizeConfig.Effects.MachineHealthCheckTimeout.Duration.String()
+	} else {
+		// If mhc timeout is configured for any size in the config, remove the annotation
+		// to fallback to the default
+		if configHasMHCTimeout(config) {
```
enxebre, line 648: why this check, should we just delete it?

hypershift-operator/controllers/scheduler/dedicated_request_serving_nodes.go
```diff
@@ -640,6 +640,17 @@ func (r *DedicatedServingComponentSchedulerAndSizer) updateHostedCluster(ctx con
...
+			if hc.Annotations[hyperv1.MachineHealthCheckTimeoutAnnotation] != "" {
```
enxebre, line 649: do we need the if here, will delete(hc.Annotations, hyperv1.MachineHealthCheckTimeoutAnnotation) panic without it?

hypershift-operator/controllers/scheduler/dedicated_request_serving_nodes.go
```diff
@@ -640,6 +640,17 @@ func (r *DedicatedServingComponentSchedulerAndSizer) updateHostedCluster(ctx con
...
+				delete(hc.Annotations, hyperv1.MachineHealthCheckTimeoutAnnotation)
+			}
```
enxebre, line 651: it'd be good to unit test this func

hypershift-operator/controllers/nodepool/nodepool_controller.go
csrwng: true, let me see if this can be validated first

hypershift-operator/controllers/scheduler/dedicated_request_serving_nodes.go
csrwng: The reasoning was that if you had configured a MHC timeout anywhere in your config, we would remove the annotation here. Otherwise, leave it alone in case (for whatever reason) you're simply annotating the HostedCluster.

hypershift-operator/controllers/scheduler/dedicated_request_serving_nodes.go
csrwng: delete will not panic, but the if as mentioned earlier is to determine if we should remove the annotation

enxebre: I mean the second if `if hc.Annotations[hyperv1.MachineHealthCheckTimeoutAnnotation] != "" {`
seems redundant. We could just

		if configHasMHCTimeout(config) {
				delete(hc.Annotations, hyperv1.MachineHealthCheckTimeoutAnnotation)
			}

csrwng: ahh got it
thx!

csrwng:
This code is not tested for kubevirt and has no impact on that platform

PR #4076, opened 2024-05-22
OCPBUGS-34071: Do not scale down legacy nodes

csrwng:
Legacy nodes (nodes without a size label) are managed by fleet manager. Therefore we should not try to scale their respective machinesets down to avoid continuously reverting fleet manager's changes.

Fixes #[OCPBUGS-34071](https://issues.redhat.com/browse/OCPBUGS-34071)

**Checklist**

### DIFF
--- a/hypershift-operator/controllers/scheduler/autoscaler.go
+++ b/hypershift-operator/controllers/scheduler/autoscaler.go
@@ -206,10 +206,12 @@ func nodeMachineSetsToScaleDown(node *corev1.Node, machineSets []machinev1beta1.
 	pairLabel := node.Labels[OSDFleetManagerPairedNodesLabel]
 	if pairLabel != "" {
 		nodesToScaleDown = filterNodes(nodes, func(n *corev1.Node) bool {
-			return n.Labels[OSDFleetManagerPairedNodesLabel] == pairLabel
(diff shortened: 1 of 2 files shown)


hypershift-operator/controllers/scheduler/autoscaler.go
```diff
@@ -206,10 +206,12 @@ func nodeMachineSetsToScaleDown(node *corev1.Node, machineSets []machinev1beta1.
 	pairLabel := node.Labels[OSDFleetManagerPairedNodesLabel]
 	if pairLabel != "" {
 		nodesToScaleDown = filterNodes(nodes, func(n *corev1.Node) bool {
-			return n.Labels[OSDFleetManagerPairedNodesLabel] == pairLabel
+			return n.Labels[OSDFleetManagerPairedNodesLabel] == pairLabel && n.Labels[hyperv1.NodeSizeLabel] != ""
```
enxebre, line 209: shouldn't we instead return early in the MachineSetDescaler.Reconcile if node.Labels[hyperv1.NodeSizeLabel] == ""?

csrwng:
/override ci/prow/e2e-kubevirt-azure-ovn
This is not tested for kubevirt and has no impact on that platform

PR #4085, opened 2024-05-23
OCPBUGS-34274: Generate default worker security group rules based on machineCIDR

mjlshen:
Previously, these rules were generated off of the containing VPC's main CIDR range. However AWS VPCs can have multiple, distinct CIDR ranges, and the main CIDR range may not be the same as the machine CIDR, where the worker nodes are. This results in security group rules being generated that do not allow the worker nodes to communicate with the hosted control plane via the PrivateLink VPC Endpoint.

Ref: https://aws.amazon.com/about-aws/whats-new/2017/08/amazon-virtual-private-cloud-vpc-now-allows-customers-to-expand-their-existing-vpcs/

[OCPBUGS-34274](https://issues.redhat.com/browse/OCPBUGS-34274)

**Checklist**

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
@@ -4702,7 +4702,7 @@ func (r *HostedControlPlaneReconciler) reconcileDefaultSecurityGroup(ctx context
 
 	originalHCP := hcp.DeepCopy()
 	var condition *metav1.Condition
-	sgID, creationErr := createAWSDefaultSecurityGroup(ctx, r.ec2Client, hcp.Spec.InfraID, hcp.Spec.Platform.AWS.CloudProviderConfig.VPC, hcp.Spec.Platform.AWS.ResourceTags)
+	sgID, creationErr := createAWSDefaultSecurityGroup(ctx, r.ec2Client, hcp)
 	if creationErr != nil {
 		condition = &metav1.Condition{
 			Type:    string(hyperv1.AWSDefaultSecurityGroupCreated),
@@ -4749,10 +4749,16 @@ func awsSecurityGroupName(infraID string) string {
 	return fmt.Sprintf("%s-default-sg", infraID)
 }
 
-func createAWSDefaultSecurityGroup(ctx context.Context, ec2Client ec2iface.EC2API, infraID, vpcID string, additionalTags []hyperv1.AWSResourceTag) (string, error) {
+func createAWSDefaultSecurityGroup(ctx context.Context, ec2Client ec2iface.EC2API, hcp *hyperv1.HostedControlPlane) (string, error) {
 	logger := ctrl.LoggerFrom(ctx)
 
-	// Determine VPC cidr
+	var (
+		vpcID          = hcp.Spec.Platform.AWS.CloudProviderConfig.VPC
+		infraID        = hcp.Spec.InfraID
+		additionalTags = hcp.Spec.Platform.AWS.ResourceTags
+	)
+
+	// Validate VPC exists
 	vpcResult, err := ec2Client.DescribeVpcsWithContext(ctx, &ec2.DescribeVpcsInput{
 		VpcIds: []*string{awssdk.String(vpcID)},
 	})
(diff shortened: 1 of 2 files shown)


control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -4762,7 +4768,18 @@ func createAWSDefaultSecurityGroup(ctx context.Context, ec2Client ec2iface.EC2AP
 	if len(vpcResult.Vpcs) == 0 {
 		return "", fmt.Errorf("vpc %s not found", vpcID)
 	}
-	vpcCIDR := awssdk.StringValue(vpcResult.Vpcs[0].CidrBlock)
+
+	// Validate the hostedcontrolplane defines a machine CIDR
+	if len(hcp.Spec.Networking.MachineNetwork) == 0 {
+		logger.Error(errors.New("hostedcontrolplane.spec.networking.machineNetwork length is 0"), "failed to extract machine CIDR while creating default security group")
```
enxebre, line 4774: why are logging in addition to the error return here?

enxebre:
note this won't reconciliate existing clusters as per check in l4694 of the controller

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
mjlshen: Ah I was trying to conform to other areas in this function - I can align them to not duplicate logging + returning errors

mjlshen:
>note this won't reconciliate existing clusters as per check in l4694 of the controller

This is ok for ROSA

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -4762,7 +4768,18 @@ func createAWSDefaultSecurityGroup(ctx context.Context, ec2Client ec2iface.EC2AP
 	if len(vpcResult.Vpcs) == 0 {
 		return "", fmt.Errorf("vpc %s not found", vpcID)
 	}
-	vpcCIDR := awssdk.StringValue(vpcResult.Vpcs[0].CidrBlock)
+
+	// Validate the hostedcontrolplane defines a machine CIDR
+	if len(hcp.Spec.Networking.MachineNetwork) == 0 {
```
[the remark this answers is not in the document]
mjlshen: I may not be fully understanding - is it? It doesn't seem to be defaulted from the CRD https://github.com/openshift/hypershift/blob/4c2f51c2e61dbe5408e78f48c90ff8cbe3ef5cc3/api/hypershift/v1beta1/hostedcluster_types.go#L690-L695

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -4748,24 +4748,38 @@ func awsSecurityGroupName(infraID string) string {
...
+		infraID        = hcp.Spec.InfraID
+		additionalTags = hcp.Spec.Platform.AWS.ResourceTags
+	)
+
+	// Validate VPC exists
 	vpcResult, err := ec2Client.DescribeVpcsWithContext(ctx, &ec2.DescribeVpcsInput{
 		VpcIds: []*string{awssdk.String(vpcID)},
 	})
 	if err != nil {
-		logger.Error(err, "Failed to describe vpc", "vpcID", vpcID)
```
csrwng, line 4759: I don't think we should delete this line. It's the only record we have of the full error. We only put the error code in status.

mjlshen: Makes sense, restored!

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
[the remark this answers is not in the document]
enxebre: It's here https://github.com/openshift/hypershift/blob/main/api/hypershift/v1beta1/hostedcluster_types.go#L388-L389

enxebre:
Thanks!
/lgtm

PR #4086, opened 2024-05-23
OCPBUGS-33952: Documented HCP service exposure

jparrill:
Added documentation detailing the various strategies available for exposing HostedControlPlane services. Additionally, included guidance on how to expose the data plane ingress using MetalLB. This information is now available in a new section called 'Recipes,' which will contain tested solutions for the multiple providers supported by hypershift.

- Fixed Contribute section of the documentation
- Added Service HCP publishing strategies
- Added a new section called Recipes to hold already tested useful recipes for customers

Fixes #[OCPBUGS-33952](https://issues.redhat.com/browse/OCPBUGS-33952)


docs/content/how-to/common/exposing-services-from-hcp.md
```diff
@@ -0,0 +1,89 @@
...
+        hostname: <URL>
+  - service: Konnectivity
+    servicePublishingStrate
+      type: Route
+  - service: Ignition
+    servicePublishingStrategy:
+      type: Route
+```
+
+If you wanna know more about how to expose the ingress service in the Data Plane side, please access the Recipes section to see how to do it with MetalLB.
```
bryan-cox, line 89: Could we add a link to the Recipes section for UX?

jparrill: Addressed :)

docs/content/recipes/common/exposing-dataplane-with-metallb.md
```diff
@@ -0,0 +1,119 @@
+## Configure MetalLB for HostedCluster's Data Plane
```
enxebre, line 1: can this just be in "how to" as well?

docs/content/how-to/common/exposing-services-from-hcp.md
```diff
@@ -0,0 +1,89 @@
+# Exposing the Hosted Control Plane Services
+
```
enxebre, line 2: can we start by introducing the different endpoints we expose and their motivations?

enxebre:
thanks a lot! some minor comments. lgtm otherwise

docs/content/recipes/common/exposing-dataplane-with-metallb.md
jparrill: It could, but I really think "recipes" section will give this documentation a real value. We can mix things like this, now we are mixing Disconnected + Agent topics, and it's a organisation problem to locate it in one place. Because of that, Toni commented that this recipes section could make much sense.

Another sample for this could be:
- Deploying disconnected operators based on tags instead of digests

This ^^ is a recurring topic I would like to document at some point, and I think the best place is the recipes. WDYT?

jparrill: From my PoV:

- Recipes include things non-related with HCP. (E.G MetalLB to configure services exposure)
- HowTo section implies how to configure HCP to reach a concrete end. (E.G How to configure TLS certificates for a disconnected environment)

In the first bullet point, we include external components to Openshift, I'm saying MetalLB but it could be w/e LoadBalancer you want to configure. 

Another Recipe could be "How to configure Disconnected environments to allow the HostedClsuter to pull images from tag"

enxebre: I'm not sure why this would match the expectation of an external reader. Seems to me quite arbitrary

enxebre:
I don't quite get the divergence of this categories as it seems arbitrary and I'm not sure how an external reader should choose to look at one or the other. But maybe is just me and also I don't want to block you and we can always follow up.
/lgtm

PR #4093, opened 2024-05-27
OCPBUGS-33987: Set controller concurrency to 1 and optimistic lock patching for DedicatedServingComponentScheduler*

enxebre:
DedicatedServingComponentScheduler* Operate over HCs as the driving resource. With current setup it might be possible for 2..10 reconciliations to happen conccurently resulting in those HC competing for Node pairs. This might result in Node pairs being split between different HCs silenty. Optimistic lock would make this discrepancy visible by making a patch request fail if one/both of the Nodes were changed. MaxConcurrentReconciles 1 would prevent the discrepancy from happening for most scenarios. Only the case where the intended transactional patch fails for one of the Nodes would potentially cause skew.

Fixes #

**Checklist**


enxebre:
/hold
a bit

enxebre:
/close
in favour of https://github.com/openshift/hypershift/pull/4103

PR #4095, opened 2024-05-27
TRT-1685: bump openshift/api

muraee:
**What this PR does / why we need it**:
fixes `ValidatingAdmissionPolicy` feature gate not enabled
Enabled in https://github.com/openshift/api/pull/1888
Needs https://github.com/openshift/release/pull/52491

**Which issue(s) this PR fixes** *(optional, use `fixes #<issue_number>(, fixes #<issue_number>, ...)` format, where issue_number might be a GitHub issue, or a Jira story*:
Fixes #

**Checklist**
- [ ] Subject and description added to both, commit and PR.
- [ ] Relevant issues have been referenced.
- [ ] This change includes docs. 
- [ ] This change includes unit tests.

PR #4099, opened 2024-05-28
Bump openshift/api dep

orenc1:
OpenShift 4.17 has bumped k8s to 1.30, which introduced a change in which ValidatingAdmissionPolicy was promoted to GA and will be enabled by default. This has caused HCP kube-apiservers to fail because the feature set we were using was setting ValidatingAdmissionPolicy=false, which is now invalid. Bumping openshift-api dependency should fix the issue by utilizing a valid set of feature gates and to be aligned with ocp 4.17 and k8s 1.30 APIs.

Fixes #

**Checklist**


orenc1:
closing in favor of #4095

PR #4101, opened 2024-05-28
CNV-39589: [Kubevirt] Add support for passing GPU devices to NodePools

orenc1:

Fixes #

**Checklist**


api/hypershift/v1beta1/nodepool_types.go
```diff
@@ -751,6 +754,20 @@ type KubevirtNetwork struct {
...
+
+	// Count is the number of instances the specified GPU device will be attached to each of the
+	// NodePool's nodes. Default is 1.
+	//
+	// +optional
+	// +kubebuilder:default=1
+	// +kubebuilder:validation:Minimum=1
+	Count int `json:"count,omitempty"`
+}
+
```
davidvossel, line 770: I don't think this will conflict with future usage of "Dynamic Resource Allocation". It's possible that this struct will get expanded in the future to allow us to express resourceClaims here as well.

The only thing i'm slightly questioning is whether we should call this struct `KubevirtHostDevices` or `KubevirtGPUs`. I'm starting a slack conversation in `#hypershift-kubevirt-platform` to talk through that possiblity

davidvossel: We talked about this out of band a bit already, but i wanted to leave the note on the PR so there's some history of this discussion.

I think we call this struct `KubevirtHostDevices` instead of `KubevirtGPU`. This gives us the ability to support more devices on the ndoepool outside of just GPUs in the future.

api/hypershift/v1alpha1/nodepool_types.go
```diff
@@ -740,17 +740,34 @@ type KubevirtNodePoolPlatform struct {
...
-// network interface interface
+// network interface
 type KubevirtNetwork struct {
 	// Name specify the network attached to the nodes
 	// it is a value with the format "[namespace]/[name]" to reference the
 	// multus network attachment definition
 	Name string `json:"name"`
 }
 
+type KubevirtGPU struct {
```
davidvossel, line 757: the v1alpha api needs to be update to use host device terminology

davidvossel (commented):
pr looks good. I made one minor comment about the alpha api.

next steps are 
* cli support for host devices
* e2e testing to verify host devices are passed through as expected when configured

cmd/nodepool/kubevirt/create.go
```diff
@@ -167,12 +200,43 @@ func (o *ValidatedKubevirtPlatformCreateOptions) Complete() (*KubevirtPlatformCr
...
+			}
+			if len(split) == 1 {
+				// count is not being used
+				kubevirtHostDevice = hyperv1.KubevirtHostDevice{
+					DeviceName: split[0],
+				}
+			}
+			hostDevices = append(hostDevices, kubevirtHostDevice)
+		}
+	}
```
davidvossel, line 231: I think this needs better error handling. As it's written here, someone could attempt to use an option not named `count` and we'd parse it like it was named count. There's also no verification that the count option has a value associated with it. We blindly try to access countSplit[1] without verifying the length of countSplit. 


Here's an example of how those two issues could be resolved. 

```
		for _, hostDevice := range o.HostDevices {
			split := strings.Split(hostDevice, ",")

			kubevirtHostDevice := hyperv1.KubevirtHostDevice{
				DeviceName: split[0],
			}

                        if len(spilt) == 1 {
                            continue
			} else if len(split) > 2 {
			    panic(fmt.Sprintf("invalid KubeVirt host device setting [%s]", hostDevice))
			}

		        // parse options (count is the only supported option right now)
			countSplit := strings.Split(split[1], ":")
                        if countSplit[0] != "count" || len countSplit != 2 {
                            panic(fmt.Sprintf("invalid KubeVirt host device setting [%s]", hostDevice))
                        }
			count, err := strconv.Atoi(countSplit[1])
			if err != nil {
			    panic(fmt.Sprintf("could not parse host device count: [%s]", hostDevice))
			}
			kubevirtHostDevice. Count = count
			
			hostDevices = append(hostDevices, kubevirtHostDevice)
		}
```

hypershift-operator/controllers/nodepool/kubevirt/kubevirt.go
```diff
@@ -271,6 +280,21 @@ func virtualMachineTemplateBase(nodePool *hyperv1.NodePool, bootImage BootImage)
 		template.Spec.Template.Spec.NodeSelector = kvPlatform.NodeSelector
 	}
 
+	if kvPlatform.KubevirtHostDevices != nil && len(kvPlatform.KubevirtHostDevices) > 0 {
+		hostDevices := []kubevirtv1.HostDevice{}
+		for _, hostDevice := range kvPlatform.KubevirtHostDevices {
+			for i := 1; i <= hostDevice.Count; i++ {
+				kvHostDevice := kubevirtv1.HostDevice{
+					Name:       "gpu" + strconv.Itoa(i),
```
davidvossel, line 288: In order to keep with the idea that this might be used for more generic devices in the future, maybe instead of calling this `gpu-` we should use `hostdevice-`

davidvossel (commented):
nice! I left a few minor comments. This looks really close

cmd/nodepool/kubevirt/create.go
orenc1: in the ` Validate()` method we're validating the argument passed with `--host-device-name` is of the correct format.
I separated the validation logic into `Validate()` and parsing and structuring logic into `Complete()`.
wouldn't that be enough to ensure only valid input be processed?
https://github.com/openshift/hypershift/pull/4101/files#diff-e9817ce7ac10df78b9412ce6725d2974cbb7f9262259f494fe74d7e0c987a69fR119-R140

hypershift-operator/controllers/nodepool/kubevirt/kubevirt.go
orenc1: done

cmd/nodepool/kubevirt/create.go
```diff
@@ -112,6 +115,30 @@ func (o *RawKubevirtPlatformCreateOptions) Validate() (*ValidatedKubevirtPlatfor
...
+			count, err := strconv.Atoi(countSplit[1])
+			if err != nil {
+				return nil, fmt.Errorf("invalid host device count format. unable to parse %s as integer: %w", countSplit[1], err)
+			}
+			if count < 1 {
+				return nil, fmt.Errorf("host device count must be greater than or equal to 1. received: %d", count)
+			}
+		}
+	}
+
```
davidvossel, line 141: there's a new unit test for this create command `cmd/cluster/kubevirt/create_test.go`

I think it would be good to exercise this host device cli arg there.

cmd/nodepool/kubevirt/create.go
davidvossel: > I separated the validation logic into Validate() and parsing and structuring logic into Complete()

both the Validate and Complete logic are kind of "half" parsing this arg in different ways.

I'd recommend consolidating this parsing logic all to one function.  You can call that function in both places if you really want, but i wouldn't parse this arg twice

cmd/nodepool/kubevirt/create.go
orenc1: a bunch of unit tests have been added for this CLI arg.

davidvossel (commented):
/lgtm
/approve

PR #4102, opened 2024-05-28
HOSTEDCP-1694: Automate HO Deployment Tag & Commit Updates for ROSA HCP

bryan-cox:

Fixes [HOSTEDCP-1694](https://issues.redhat.com/browse/HOSTEDCP-1694)

**Checklist**


hack/tools/deployment/deployment-update.go
```diff
@@ -0,0 +1,255 @@
...
+import (
+	"flag"
+	"fmt"
+	"log"
+	"os"
+	"strings"
+)
+
+var (
+	ValidEnvironments = []string{
```
enxebre, line 12: this could be a map[string][]string with the env as the hash key and sectors list as the value. Then there's no need to switch case below

enxebre:
One minor comment, otherwise looks like good starting point to me.

hack/tools/deployment/deployment-update.go
```diff
@@ -0,0 +1,255 @@
...
+	err = os.WriteFile(*file, []byte(output), 0644)
+	if err != nil {
+		log.Fatalf("Error: %v", err)
+	}
+
+}
+
+// updateDeploymentByEnvironment updates the sectors in a particular ROSA environment
+func updateDeploymentByEnvironment(env string, lines []string, oldTag, newTag, newCommit string) error {
+	switch env {
```
stevekuznetsov, line 131: You could factor this to reduce the repetition:

```go
func updateDeploymentByEnvironment(env string, lines []string, oldTag, newTag, newCommit string) error {
    for _, sectors := range map[string][]string{
        "int": Integration,
        // ...
    } {
        for _, sector := range sectors {
            if err := updateTagAndCommitSHA(&lines, sector, oldTag, newTag, newCommit); err != nil {
                return err
            }
        }
    }
}
```

bryan-cox: The repetition should be gone now.

hack/tools/deployment/deployment-update.go
```diff
@@ -0,0 +1,186 @@
...
+	if file != nil && len(*file) == 0 ||
+		env != nil && len(*env) == 0 ||
+		oldTag != nil && len(*oldTag) == 0 ||
+		newTag != nil && len(*newTag) == 0 ||
+		newCommit != nil && len(*newCommit) == 0 {
+		log.Fatalf("all flags are required")
+	}
+
+	// Verify the right env value was used
+	if validEnvironment := isValidEnvironment(*env); !validEnvironment {
```
stevekuznetsov, line 80:
```suggestion
	if !isValidEnvironment(*env) {
```

hack/tools/deployment/deployment-update.go
```diff
@@ -0,0 +1,186 @@
...
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// isValidEnvironment verifies the environment passed through the flag is valid
+func isValidEnvironment(env string) bool {
```
stevekuznetsov, line 125: might be simpler:

```go
func isValidEnvironment(env string) bool {
    _, valid := ROSAEnvironments[env]
    return valid
}

**Review by stevekuznetsov — COMMENTED** [review]
In general I'd discourage string parsing and walking a structured file, but this is a good start and we can improve it if we need to. A follow-up might be to code a bit more defensively since we're running roughshod over YAML - when we're searching for another sector, or transitioning between them, a more defensive approach would have us assert our assumptions - if we *don't* find a new sector but we think we should have, panic, etc. Fail explicitly vs falling through implicitly.

**Review by stevekuznetsov — APPROVED** [review]
/lgtm

File: hack/tools/deployment/deployment-update.go [hunk 5]
```diff
@@ -0,0 +1,295 @@
...
+	// Write the modified contents back to the file
+	err = os.WriteFile(*file, []byte(output), 0644)
+	if err != nil {
+		log.Fatalf("Error: %v", err)
+	}
+
+	if autoCommitChanges {
+		// Add the changes in git
+		log.Printf("Adding changes in git")
+		err = gitAdd(filepath.Dir(*file))
```
stevekuznetsov, line 137: Ideally create a context with `signal.Notify()` on SIGINT/os.Interrupt and pass in to child processes.

hack/tools/deployment/deployment-update.go
```diff
@@ -0,0 +1,295 @@
...
+		if err != nil {
+			log.Fatalf("Error pushing changes in git: %v", err)
+		}
+
+		// TODO automate the MR request
+		log.Printf("-------------Info for MR-------------")
+		log.Printf("MR Title: %s: Bump HyperShift in %s to %s / %s", *jiraTicket, *env, *newTag, *newCommit)
+
+		// Output the changes between tags for the merge request
+		cmd := exec.Command("go", "run", "../release/notes.go", "--from="+*oldTag, "--to="+*newTag)
```
stevekuznetsov, line 161: It can be super wonky to interrupt `go run` - can we get the release note code to be importable? If not easy, this is fine.

hack/tools/deployment/deployment-update.go
```diff
@@ -0,0 +1,295 @@
...
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("error creating branch: %s\n%s", err, output)
+	}
+	fmt.Printf("Output: %s\n", output)
+	return nil
+}
+
+func gitAdd(repoPath string) error {
+	cmd := exec.Command("git", "add", ".")
```
stevekuznetsov, line 265: Prefer to add specific files in automated tools to reduce the chances of accidental inclusion

hack/tools/deployment/deployment-update.go
```diff
@@ -0,0 +1,295 @@
...
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("error adding files: %s\n%s", err, output)
+	}
+	fmt.Printf("Output: %s\n", output)
+	return nil
+}
+
+func gitCommit(repoPath, message string) error {
+	cmd := exec.Command("git", "commit", "-m", message)
```
stevekuznetsov, line 276: nit: for folks that sign their commits you'll want `--signoff`

hack/tools/deployment/deployment-update.go
```diff
@@ -0,0 +1,295 @@
...
+		// Commit the changes in git
+		log.Printf("Committing changes in git")
+		err = gitCommit(filepath.Dir(*file), fmt.Sprintf("Bump HyperShift in %s to %s / %s", *env, *newTag, *newCommit))
+		if err != nil {
+			log.Fatalf("Error committing changes in git: %v", err)
+		}
+
+		// Push the changes up
+		log.Printf("Pushing changes in git")
+		err = gitPush(filepath.Dir(*file), "origin", branchName)
```
stevekuznetsov, line 151: nit: make the remote a parameter, I for instance do not use "origin"

hack/tools/deployment/deployment-update.go
```diff
@@ -0,0 +1,295 @@
...
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("error checking out branch: %s\n%s", err, output)
+	}
+	fmt.Printf("Output: %s\n", output)
+	return nil
+}
+
+func gitCreateBranch(repoPath, branch string) error {
+	cmd := exec.Command("git", "checkout", "-b", branch)
```
stevekuznetsov, line 254: nit: not sure if it matters but `git branch <name> <target>` sets `<target>` as the remote-tracking branch, useful for explicitly calling out what you're branching off of.

hack/tools/deployment/deployment-update.go
bryan-cox: Discussed privately in Slack. Really just looking for the quickest path to making this work. We will also have to refactor this file when the new way of rolling out HO is implemented so prefer to go with what we have now.

stevekuznetsov:
@bryan-cox i will review when you push the rebase!

bryan-cox:
@stevekuznetsov - it's been rebased now

hack/tools/deployment/deployment-update.go
```diff
@@ -0,0 +1,297 @@
...
+
+	if found && !modified {
+		log.Printf("did not find a line containing the old tag, %s, for sector '%s'", oldTag, sector)
+	}
+
+	return nil
+}
+
+func gitCheckout(repoPath, branch string) error {
+	cmd := exec.Command("git", "checkout", branch)
```
stevekuznetsov, line 245: I think the review comment on using CommandContext and os.Notify got lost - would be the only outstanding comment, rest looks great!

stevekuznetsov (changes requested):

stevekuznetsov (approved):

PR #4103, opened 2024-05-29
OCPBUGS-33987: Use configmaps to track hosted cluster pair labels and set controller concurrency

csrwng:

DedicatedServingComponentScheduler* Operate over HCs as the driving resource. With current setup it might be possible for 2..10 reconciliations to happen conccurently resulting in those HC competing for Node pairs. This might result in Node pairs being split between different HCs silenty. Optimistic lock would make this discrepancy visible by making a patch request fail if one/both of the Nodes were changed. MaxConcurrentReconciles 1 would prevent the discrepancy from happening for most scenarios. Only the case where the intended transactional patch fails for one of the Nodes would potentially cause skew.

Adds the use of configmaps as a lease for a particular fleet manager pair label for a given hosted cluster. In order to label nodes with a given fleet manager label, a configmap named with the pair label must exist and contain the hosted cluster's name and namespace.

Fixes #OCPBUGS-33987

**Checklist**

### DIFF
--- a/hypershift-operator/controllers/scheduler/dedicated_request_serving_nodes.go
+++ b/hypershift-operator/controllers/scheduler/dedicated_request_serving_nodes.go
@@ -8,6 +8,7 @@ import (
 
 	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
 	schedulingv1alpha1 "github.com/openshift/hypershift/api/scheduling/v1alpha1"
+	"github.com/openshift/hypershift/hypershift-operator/controllers/hostedcluster"
 	"github.com/openshift/hypershift/support/upsert"
 	"github.com/openshift/hypershift/support/util"
 	appsv1 "k8s.io/api/apps/v1"
@@ -25,6 +26,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
@@ -45,6 +47,11 @@ const (
 	PlaceholderLabel = "hypershift.openshift.io/placeholder"
 
 	autoSizerNamespace = placeholderNamespace
+
+	clusterNamespaceKey = "clusterNamespace"
+	clusterNameKey      = "clusterName"
+	pairLabelKey        = "hypershift.openshift.io/pairlabel"
+	schedulerFinalizer  = "hypershift.openshift.io/dedicated-request-serving-scheduler"
 )
 
 type DedicatedServingComponentNodeReaper struct {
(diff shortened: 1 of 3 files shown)


csrwng:
Supersedes #4093

hypershift-operator/controllers/scheduler/dedicated_request_serving_nodes.go
```diff
@@ -45,8 +48,67 @@ const (
...
+func (r *PairLabelConfigMapReaper) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	log := ctrl.LoggerFrom(ctx)
+	log.Info("Reconciling")
+	configMapList := &corev1.ConfigMapList{}
+	if err := r.List(ctx, configMapList, client.HasLabels{pairLabelKey}, client.InNamespace(placeholderNamespace)); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to list configmaps: %w", err)
+	}
+	for _, cm := range configMapList.Items {
+		hc := &hyperv1.HostedCluster{}
+		hc.Namespace = cm.Data[clusterNamespaceKey]
```
enxebre, line 92: Isn't all this reconcile implementing an ownerRef + garbage collection behaviour, i.e. wouldn't set an ownerRef in the ConfigMap to the HC be enough?

hypershift-operator/controllers/scheduler/dedicated_request_serving_nodes.go
```diff
@@ -473,7 +547,14 @@ func (r *DedicatedServingComponentSchedulerAndSizer) Reconcile(ctx context.Conte
 				return ctrl.Result{}, fmt.Errorf("failed to get nodes from placeholders: %w", err)
 			}
 			if len(candidateNodes) > 0 {
+				pairLabel = candidateNodes[0].Labels[OSDFleetManagerPairedNodesLabel]
+				if pairLabel == "" {
+					return ctrl.Result{}, fmt.Errorf("node %s has no pair label", candidateNodes[0].Name)
+				}
 				log.WithValues("pairLabel", candidateNodes[0].Labels[OSDFleetManagerPairedNodesLabel]).Info("claiming candidate nodes")
+				if err := r.ensurePairConfigMap(ctx, pairLabel, hc.Namespace, hc.Name); err != nil {
```
enxebre, line 555: I would expect this controller only ensures the configMap now (not ensureHostedClusterLabelAndTaint). Then we'd run a separate controller to watch those configMaps (e.g. caching by a particular label) and let it ensureHostedClusterLabelAndTaint triggered by the cm events.

hypershift-operator/controllers/scheduler/dedicated_request_serving_nodes.go
enxebre: I'd actually this reaper to be an "applier" https://github.com/openshift/hypershift/pull/4103/commits/cd2f799f41097b135b6a7f899c3294da3d7d0266#r1618385200 and let reaper be ownerRef builtin behaviour.

csrwng: cross-namespace owner refs [are not allowed](https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/), so the HC and configmap would need to be in the same namespace.

hypershift-operator/controllers/scheduler/dedicated_request_serving_nodes.go
csrwng: Hmm, yup I guess that would simplify things. As it is, we're doing too much. Let me rework this a bit.

hypershift-operator/controllers/scheduler/dedicated_request_serving_nodes.go
```diff
@@ -45,8 +48,67 @@ const (
...
+		defer ticker.Stop()
+		for range ticker.C {
+			cm := &corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Name: "ticker", Namespace: "none"}}
+			tickerChannel <- event.GenericEvent{Object: cm}
+		}
+	}()
+	return builder.Complete(r)
+}
+
+func (r *PairLabelConfigMapReaper) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
stevekuznetsov, line 83: Why not a finalizer?

hypershift-operator/controllers/scheduler/dedicated_request_serving_nodes.go
```diff
@@ -539,6 +633,49 @@ func (r *DedicatedServingComponentSchedulerAndSizer) Reconcile(ctx context.Conte
 	return ctrl.Result{}, nil
 }
 
+func (r *DedicatedServingComponentSchedulerAndSizer) ensurePairConfigMap(ctx context.Context, pairLabel, hcNamespace, hcName string) error {
+	cfgMap := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: placeholderNamespace,
+			Name:      pairLabel,
```
stevekuznetsov, line 640: Doesn't this leave open the issue of two reconcilers allocating the same cluster to two different pairs?

csrwng: > Doesn't this leave open the issue of two reconcilers allocating the same cluster to two different pairs?

Yes, but I would think the effect of that is not as bad as having 2 clusters with the same pair. We will eventually label/taint nodes in one of those pairs and that's what we will use.

stevekuznetsov (approved):

PR #1782, opened 2022-10-04
CNF-6516: NTO: Enable PerformanceProfile 

jlojosnegros:
**What this PR does / why we need it**:

As part of the phase-3 of Node Tuning Operator adaptation to run in Hypershift environments ( #1651 ), [Performance Profile controller](https://github.com/openshift/cluster-node-tuning-operator/blob/master/docs/performanceprofile/performance_controller.md) should also be adapted.

Performance Profile controller uses [`PerformanceProfile`](https://docs.openshift.com/container-platform/4.11/rest_api/node_apis/performanceprofile-performance-openshift-io-v2.html) as its API.
This PR modifies NodePool controller to allow Performance Profile controller to read `PerformanceProfile`s manifests using  the same procedure as NTO is already using with Tuned manifests

This PR contains the following changes, which are also outlined in this document openshift/enhancements#1244:

- NodePool controller changes:
  - Modify code to reconcile the spec.TunedConfig field so it would not only reconcile Tuned manifests but also PerformanceProfile manifests. For each NodePool, propagate any Tuned or PerformanceProfile manifests from the ConfigMaps referenced in the spec.TunedConfig field into ConfigMaps in the hosted control plane namespace where NTO, and so Performance Profile controller,  can read them.

**Which issue(s) this PR fixes** 
Fixes : [CNF-6291](https://issues.redhat.com/browse/CNF-6291)

**Checklist**
- [x] Subject and description added to both, commit and PR.
- [x] Relevant issues have been referenced.
- [x] This change includes docs. 
- [x] This change includes unit tests.

PR #4108, opened 2024-05-29
CNF-12951:hypershift:performanceprofile: associate profile name with user input

Tal-or:
When a `PerformanceProfile` should be mirrored into the HCP namespace, its given name which compose from a constant (perfprof) + the node pool name.

In case there are multiple `PerformanceProfiles`, one per hosted cluster, the approach mentioned above makes it hard to understand which `PreformanceProfile` in the clusters namespace is associate with each hostedcluster.

It requires to examine the node pool \`spec.tuningConfig\` to figure out the relation between user profile and the mirrored one.

This commit suggest to embed the configMap (that encapsulates the performanceProfile) name in the mirrored profile, to make the relation clear and observable on first sight.

Fixes # [CNF-12951](https://issues.redhat.com/browse/CNF-12951)

### DIFF
--- a/hypershift-operator/controllers/nodepool/manifests.go
+++ b/hypershift-operator/controllers/nodepool/manifests.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 
 	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
+	"github.com/openshift/hypershift/support/util"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	capiv1 "sigs.k8s.io/cluster-api/api/v1beta1"
@@ -12,6 +13,10 @@ import (
 const (
 	EC2VolumeDefaultSize int64  = 16
 	EC2VolumeDefaultType string = "gp3"
+
+	// QualifiedNameMaxLength is the maximal name length allowed for k8s object
+	// https://github.com/kubernetes/kubernetes/blob/957c9538670b5f7ead2c9ba9ceb9de081d66caa4/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L34
+	QualifiedNameMaxLength = 63
 )
 
 func machineDeployment(nodePool *hyperv1.NodePool, controlPlaneNamespace string) *capiv1.MachineDeployment {
@@ -72,11 +77,11 @@ func TunedConfigMap(namespace, name string) *corev1.ConfigMap {
 	}
 }
 
-func PerformanceProfileConfigMap(namespace, name string) *corev1.ConfigMap {
+func PerformanceProfileConfigMap(namespace, name, nodePoolName string) *corev1.ConfigMap {
 	return &corev1.ConfigMap{
 		ObjectMeta: metav1.ObjectMeta{
 			Namespace: namespace,
-			Name:      "perfprofile-" + name,
+			Name:      util.ShortenName(name, nodePoolName, QualifiedNameMaxLength),
 		},
 	}
 }
--- a/hypershift-operator/controllers/nodepool/nodepool_controller.go
+++ b/hypershift-operator/controllers/nodepool/nodepool_controller.go
@@ -45,6 +45,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	serializer "k8s.io/apimachinery/pkg/runtime/serializer/json"
@@ -682,7 +683,7 @@ func (r *NodePoolReconciler) reconcile(ctx context.Context, hcluster *hyperv1.Ho
 	}
 
 	// Validate tuningConfig input.
-	tunedConfig, performanceProfileConfig, err := r.getTuningConfig(ctx, nodePool)
+	tunedConfig, performanceProfileConfig, performanceProfileConfigMapName, err := r.getTuningConfig(ctx, nodePool)
 	if err != nil {
 		SetStatusCondition(&nodePool.Status.Conditions, hyperv1.NodePoolCondition{
 			Type:               hyperv1.NodePoolValidTuningConfigConditionType,
@@ -735,12 +736,14 @@ func (r *NodePoolReconciler) reconcile(ctx context.Context, hcluster *hyperv1.Ho
 		}
 	}
 
-	performanceProfileConfigMap := PerformanceProfileConfigMap(controlPlaneNamespace, nodePool.Name)
 	if performanceProfileConfig == "" {
-		if _, err := supportutil.DeleteIfNeeded(ctx, r.Client, performanceProfileConfigMap); err != nil {
+		// at this point in time, we no longer know the name of the ConfigMap in the HCP NS
+		// so, we remove it by listing by a label unique to PerformanceProfile
+		if err := deleteConfigByLabel(ctx, r.Client, map[string]string{PerformanceProfileConfigMapLabel: "true"}); err != nil {
 			return ctrl.Result{}, fmt.Errorf("failed to delete performanceprofileConfig ConfigMap: %w", err)
 		}
 	} else {
+		performanceProfileConfigMap := PerformanceProfileConfigMap(controlPlaneNamespace, performanceProfileConfigMapName, nodePool.Name)
 		if result, err := r.CreateOrUpdate(ctx, r.Client, performanceProfileConfigMap, func() error {
 			return reconcilePerformanceProfileConfigMap(performanceProfileConfigMap, nodePool, performanceProfileConfig)
 		}); err != nil {
@@ -2167,10 +2170,11 @@ func (r *NodePoolReconciler) getConfig(ctx context.Context,
 
 func (r *NodePoolReconciler) getTuningConfig(ctx context.Context,
 	nodePool *hyperv1.NodePool,
-) (string, string, error) {
+) (string, string, string, error) {
 	var (
 		configs                              []corev1.ConfigMap
 		tunedAllConfigPlainText              []string
+		performanceProfileConfigMapName      string
 		performanceProfileAllConfigPlainText []string
 		errors                               []error
 	)
@@ -2204,6 +2208,7 @@ func (r *NodePoolReconciler) getTuningConfig(ctx context.Context,
 			tunedAllConfigPlainText = append(tunedAllConfigPlainText, string(manifestTuned))
 		}
 		if manifestPerformanceProfile != nil {
+			performanceProfileConfigMapName = config.Name
 			performanceProfileAllConfigPlainText = append(performanceProfileAllConfigPlainText, string(manifestPerformanceProfile))
 		}
 	}
@@ -2216,7 +2221,7 @@ func (r *NodePoolReconciler) getTuningConfig(ctx context.Context,
 	sort.Strings(tunedAllConfigPlainText)
 	sort.Strings(performanceProfileAllConfigPlainText)
 
-	return strings.Join(tunedAllConfigPlainText, "\n---\n"), strings.Join(performanceProfileAllConfigPlainText, "\n---\n"), utilerrors.NewAggregate(errors)
+	return strings.Join(tunedAllConfigPlainText, "\n---\n"), strings.Join(performanceProfileAllConfigPlainText, "\n---\n"), performanceProfileConfigMapName, utilerrors.NewAggregate(errors)
 
 }
 
@@ -3223,4 +3228,20 @@ func globalConfigString(hcluster *hyperv1.HostedCluster) (string, error) {
 
 func payloadConfigHash(config, targetVersion, pullSecretName, globalConfig string) string {
 	return supportutil.HashSimple(config + targetVersion + pullSecretName + globalConfig)
+}
+
+func deleteConfigByLabel(ctx context.Context, c client.Client, lbl map[string]string) error {
+	cmList := &corev1.ConfigMapList{}
+	if err := c.List(ctx, cmList, &client.ListOptions{
+		LabelSelector: labels.SelectorFromSet(lbl),
+	}); err != nil {
+		return err
+	}
+	for i := range cmList.Items {
+		cm := &cmList.Items[i]
+		if _, err := supportutil.DeleteIfNeeded(ctx, c, cm); err != nil {
+			return err
+		}
+	}
+	return nil
 }
--- a/hypershift-operator/controllers/nodepool/nodepool_controller_test.go
+++ b/hypershift-operator/controllers/nodepool/nodepool_controller_test.go
@@ -1047,12 +1047,13 @@ status: {}
 
 	namespace := "test"
 	testCases := []struct {
-		name           string
-		nodePool       *hyperv1.NodePool
-		tuningConfig   []client.Object
-		tunedExpect    string
-		perfprofExpect string
-		error          bool
+		name               string
+		nodePool           *hyperv1.NodePool
+		tuningConfig       []client.Object
+		tunedExpect        string
+		perfprofExpect     string
+		perfProfNameExpect string
+		error              bool
 	}{
 		{
 			name: "gets a single valid TunedConfig",
@@ -1175,9 +1176,10 @@ status: {}
 					BinaryData: nil,
 				},
 			},
-			tunedExpect:    "",
-			perfprofExpect: perfprofOneDefaulted,
-			error:          false,
+			tunedExpect:        "",
+			perfprofExpect:     perfprofOneDefaulted,
+			perfProfNameExpect: "perfprofOne",
+			error:              false,
 		},
 		{
 			name: "Should be at most one PerformanceProfileConfig per NodePool",
@@ -1242,7 +1244,7 @@ status: {}
 			error:          true,
 		},
 		{
-			name: "PerformanceProfiles and Tuned Configs could cohexists",
+			name: "PerformanceProfiles and Tuned Configs could coexists",
 			nodePool: &hyperv1.NodePool{
 				ObjectMeta: metav1.ObjectMeta{
 					Namespace: namespace,
@@ -1291,9 +1293,10 @@ status: {}
 					},
 				},
 			},
-			tunedExpect:    tuned1Defaulted + "\n---\n" + tuned2Defaulted,
-			perfprofExpect: perfprofOneDefaulted,
-			error:          false,
+			tunedExpect:        tuned1Defaulted + "\n---\n" + tuned2Defaulted,
+			perfprofExpect:     perfprofOneDefaulted,
+			perfProfNameExpect: "perfprofOne",
+			error:              false,
 		},
 	}
 
@@ -1305,7 +1308,7 @@ status: {}
 				Client: fake.NewClientBuilder().WithObjects(tc.tuningConfig...).Build(),
 			}
 
-			td, pp, err := r.getTuningConfig(context.Background(), tc.nodePool)
+			td, pp, ppName, err := r.getTuningConfig(context.Background(), tc.nodePool)
 
 			if tc.error {
 				g.Expect(err).To(HaveOccurred())
@@ -1319,6 +1322,10 @@ status: {}
 			if diff := cmp.Diff(pp, tc.perfprofExpect); diff != "" {
 				t.Errorf("actual Performance Profile config differs from expected: %s", diff)
 				t.Logf("got:\n%s\n, expected:\n%s\n", pp, tc.perfprofExpect)
+			}
+			if diff := cmp.Diff(ppName, tc.perfProfNameExpect); diff != "" {
+				t.Errorf("Performance Profile config name differ from expected: %s", diff)
+				t.Logf("got:\n%s\n, expected:\n%s\n", ppName, tc.perfProfNameExpect)
 			}
 		})
 	}
--- a/support/util/route.go
+++ b/support/util/route.go
@@ -26,17 +26,17 @@ func ShortenRouteHostnameIfNeeded(name, namespace string, baseDomain string) str
 	if len(name)+len(namespace)+1 < validation.DNS1123LabelMaxLength {
 		return ""
 	} else {
-		return fmt.Sprintf("%s.%s", strings.TrimSuffix(shortenName(name+"-"+namespace, "", validation.DNS1123LabelMaxLength), "-"), baseDomain)
+		return fmt.Sprintf("%s.%s", strings.TrimSuffix(ShortenName(name+"-"+namespace, "", validation.DNS1123LabelMaxLength), "-"), baseDomain)
 	}
 }
 
-// shortenName returns a name given a base ("deployment-5") and a suffix ("deploy")
+// ShortenName returns a name given a base ("deployment-5") and a suffix ("deploy")
 // It will first attempt to join them with a dash. If the resulting name is longer
 // than maxLength: if the suffix is too long, it will truncate the base name and add
 // an 8-character hash of the [base]-[suffix] string.  If the suffix is not too long,
 // it will truncate the base, add the hash of the base and return [base]-[hash]-[suffix]
 // Source: openshift/origin v3.9.0 pkg/api/apihelpers/namer.go
-func shortenName(base, suffix string, maxLength int) string {
+func ShortenName(base, suffix string, maxLength int) string {
 	if maxLength <= 0 {
 		return ""
 	}
--- a/support/util/route_test.go
+++ b/support/util/route_test.go
@@ -55,7 +55,7 @@ func TestShortenName(t *testing.T) {
 
 		for j, test := range tests {
 			t.Run(fmt.Sprintf("test-%d-%d", i, j), func(t *testing.T) {
-				result := shortenName(test.base, test.suffix, kvalidation.DNS1123SubdomainMaxLength)
+				result := ShortenName(test.base, test.suffix, kvalidation.DNS1123SubdomainMaxLength)
 				if result != test.expected {
 					t.Errorf("Got unexpected result. Expected: %s Got: %s", test.expected, result)
 				}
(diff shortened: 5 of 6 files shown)


hypershift-operator/controllers/nodepool/manifests.go
```diff
@@ -69,11 +69,11 @@ func TunedConfigMap(namespace, name string) *corev1.ConfigMap {
 	}
 }
 
-func PerformanceProfileConfigMap(namespace, name string) *corev1.ConfigMap {
+func PerformanceProfileConfigMap(namespace, name, nodePoolName string) *corev1.ConfigMap {
 	return &corev1.ConfigMap{
 		ObjectMeta: metav1.ObjectMeta{
 			Namespace: namespace,
-			Name:      "perfprofile-" + name,
+			Name:      name + "-" + nodePoolName,
```
stevekuznetsov, line 76: This can overflow the allowed length for a Kubernetes object name. If we require that this name be human-readable, please truncate each section (name, node pool name) and concatenate. If we do not require, please base32(sha224(input)) to get a hash which we can use as the name for this `ConfigMap` and leave the identifying information in labels or annotations.

stevekuznetsov (changes requested):
We should be careful about deriving names from user input.

Tal-or:
> We should be careful about deriving names from user input.

@stevekuznetsov Thank you for you input. Any suggestion how to make it safer? the `name` part in this function equals to the name of the `ConfigMap` provided by the user, on which k8s already perform validation.

hypershift-operator/controllers/nodepool/manifests.go
Tal-or: Appreciate your comment. 
`name` and `nodePoolName` are valid qualified DNS subdomain name.
The only problem here is the length as you mentioned. I've fixed it in the last update, thanks!

hypershift-operator/controllers/nodepool/manifests.go
```diff
@@ -69,11 +69,17 @@ func TunedConfigMap(namespace, name string) *corev1.ConfigMap {
 	}
 }
 
-func PerformanceProfileConfigMap(namespace, name string) *corev1.ConfigMap {
+func PerformanceProfileConfigMap(namespace, name, nodePoolName string) *corev1.ConfigMap {
+	fullName := fmt.Sprintf("%s-%s", name, nodePoolName)
```
stevekuznetsov, line 73: I think this will have to be a bit more complex - since `name` is already a k8s object identifier and can be at the length limit itself, this `fullName` degrades to `name` in those cases and we lose the suffix / disambiguation you're trying to do here. 

You might choose to do some weighted prefixing where over-length issues get resolved by taking a prefix of both parts, if you want to keep a human-readable name. If you are not, base36(sha224(inputs)) is a good outlet. See [this example](https://github.com/openshift/hypershift/blob/921f254193afd0debb365fc96960bdba8129e970/control-plane-pki-operator/certificaterevocationcontroller/certificaterevocationcontroller.go#L447-L452) for an implementation.

Tal-or: > I think this will have to be a bit more complex - since name is already a k8s object identifier and can be at the length limit itself, this fullName degrades to name in those cases and we lose the suffix / disambiguation you're trying to do here.

I wouldn't over complicated it too much, but ok, no biggie to support it

Tal-or:
Hey @stevekuznetsov PTAL I addressed your comments

hypershift-operator/controllers/nodepool/manifests.go
```diff
@@ -69,11 +69,17 @@ func TunedConfigMap(namespace, name string) *corev1.ConfigMap {
...
+	// 63 is the qualifiedNameMaxLength allowed for k8s object
+	// https://github.com/kubernetes/kubernetes/blob/957c9538670b5f7ead2c9ba9ceb9de081d66caa4/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L34
+	if len(fullName) > 63 {
+		fullName = fmt.Sprintf("%s-%s", name[:32], nodePoolName[:31])
```
stevekuznetsov, line 77: You have not checked the length of `name` or `nodePoolName` so these slices may be out-of-bounds and panic.

Also, I would like to gently push back on your comment [here](https://github.com/openshift/hypershift/pull/4108/files#r1623367155) about "over complication" - if we do not write this code correctly, the controller will fail to function and human intervention will be required. I don't see a compelling reason to choose to write a method here that fails to correctly encode the name and hope that we don't hit those cases in production.

Tal-or: > You have not checked the length of name or nodePoolName so these slices may be out-of-bounds and panic.

My bad. Thank you for that

> Also, I would like to gently push back on your comment [here](https://github.com/openshift/hypershift/pull/4108/files#r1623367155) about "over complication" - if we do not write this code correctly, the controller will fail to function and human intervention will be required. I don't see a compelling reason to choose to write a method here that fails to correctly encode the name and hope that we don't hit those cases in production.

I apologies if my comment sound a bit rude, that's not what I meant.
What I meant is that having a weighted prefixes, where each name gets its prefix in proportion to its total length, is a bit complex because there are many edge cases and in some of them, the name still might be ambiguous.

BTW the previous implementation was problematic as well, because we are not checking the length of `perfprofile-+name`

hypershift-operator/controllers/nodepool/manifests.go
```diff
@@ -69,11 +69,50 @@ func TunedConfigMap(namespace, name string) *corev1.ConfigMap {
...
 			Namespace: namespace,
-			Name:      "perfprofile-" + name,
+			Name:      composeValidName(name, nodePoolName),
 		},
 	}
 }
+
+// composeValidName takes two qualified objects` names
+// and composes a new qualified name out them by performing weighted prefixing
+func composeValidName(name1, name2 string) string {
```
Tal-or, line 83: @stevekuznetsov Given the scope and the size of the function I suggest to move it into a central place and expose it as public, since it can be useful elsewhere.

stevekuznetsov: My personal rule is to do that sort of thing once the function needs to be re-used, instead of before-hand. In my opinion that helps to guard against premature optimizations or generalizations where we may be inferring possible future uses we don't know about yet. Once other users of the function exist, though, making it generic and a utility sounds good!

stevekuznetsov (commented):
Thanks for the updates, one last thing: could you please unit-test `composeValidName`? Alternatively, if you're interested, a [fuzzed test](https://go.dev/doc/tutorial/fuzz) would be a good match for this and would keep you from needing to think of every possible edge case to put into the unit test.

hypershift-operator/controllers/nodepool/manifests.go
stevekuznetsov: I see, no offense taken. Non-weighted approaches do exist, as well. I don't have any preference and the weighted approach is very reasonable.

hypershift-operator/controllers/nodepool/manifests.go
Tal-or: I completely agree, but i'll need it here: https://github.com/openshift/hypershift/pull/4150#discussion_r1628125124

Tal-or:
> Thanks for the updates, one last thing: could you please unit-test composeValidName? Alternatively, if you're interested, a [fuzzed test](https://go.dev/doc/tutorial/fuzz) would be a good match for this and would keep you from needing to think of every possible edge case to put into the unit test.

Good idea, added a fuzz test, hopefully I used it correctly.

hypershift-operator/controllers/nodepool/manifests.go
Tal-or: If it's fine by you, we can keep this function in this file for now, and move it to central package on  the second PR. this way it would be easier to explain the justification.

stevekuznetsov: Sounds good to me!

stevekuznetsov:
Nice! Looking good. Mind pasting the output of `go test -fuzz FuzzComposeValidName` for the approver?

/lgtm

Tal-or:
> Nice! Looking good. Mind pasting the output of `go test -fuzz FuzzComposeValidName` for the approver?
> 
> /lgtm

```
[titzhak@fedora nodepool]$ go test -fuzz FuzzComposeValidName
fuzz: elapsed: 0s, gathering baseline coverage: 0/7 completed
fuzz: elapsed: 0s, gathering baseline coverage: 7/7 completed, now fuzzing with 20 workers
fuzz: elapsed: 3s, execs: 37296 (12432/sec), new interesting: 0 (total: 7)
fuzz: elapsed: 6s, execs: 76520 (13073/sec), new interesting: 0 (total: 7)
fuzz: elapsed: 9s, execs: 114779 (12755/sec), new interesting: 0 (total: 7)
fuzz: elapsed: 12s, execs: 151121 (12111/sec), new interesting: 0 (total: 7)
fuzz: elapsed: 15s, execs: 186756 (11881/sec), new interesting: 0 (total: 7)
fuzz: elapsed: 18s, execs: 223024 (12089/sec), new interesting: 0 (total: 7)
fuzz: elapsed: 21s, execs: 257804 (11591/sec), new interesting: 0 (total: 7)
fuzz: elapsed: 24s, execs: 291400 (11202/sec), new interesting: 0 (total: 7)
fuzz: elapsed: 27s, execs: 327306 (11968/sec), new interesting: 0 (total: 7)
fuzz: elapsed: 30s, execs: 362923 (11870/sec), new interesting: 0 (total: 7)
fuzz: elapsed: 33s, execs: 398598 (11894/sec), new interesting: 0 (total: 7)
fuzz: elapsed: 36s, execs: 434157 (11853/sec), new interesting: 0 (total: 7)
fuzz: elapsed: 39s, execs: 469127 (11656/sec), new interesting: 0 (total: 7)
fuzz: elapsed: 42s, execs: 502549 (11140/sec), new interesting: 0 (total: 7)
fuzz: elapsed: 45s, execs: 536482 (11310/sec), new interesting: 0 (total: 7)
fuzz: elapsed: 48s, execs: 571028 (11517/sec), new interesting: 0 (total: 7)
fuzz: elapsed: 51s, execs: 605053 (11340/sec), new interesting: 0 (total: 7)
fuzz: elapsed: 54s, execs: 639684 (11543/sec), new interesting: 0 (total: 7)
fuzz: elapsed: 57s, execs: 674149 (11489/sec), new interesting: 0 (total: 7)
fuzz: elapsed: 1m0s, execs: 708181 (11345/sec), new interesting: 0 (total: 7)
fuzz: elapsed: 1m3s, execs: 740941 (10917/sec), new interesting: 0 (total: 7)
fuzz: elapsed: 1m6s, execs: 775150 (11406/sec), new interesting: 0 (total: 7)
^Cfuzz: elapsed: 1m9s, execs: 806914 (11409/sec), new interesting: 0 (total: 7)
PASS
ok      github.com/openshift/hypershift/hypershift-operator/controllers/nodepool        68.892s

```

Tal-or:
@sjenning @isco-rodriguez A gentle ping, could you please review?

Tal-or:
@stevekuznetsov Hello again, I had to make the ComposeValidName function public, since turns out there is an e2e test which relies on it. PTAL when possible

stevekuznetsov:
```
 vet: test/e2e/nodepool_nto_performanceprofile_test.go:119:39: invalid operation: performanceProfileConfigMap + nodePool.Name (mismatched types *"k8s.io/api/core/v1".ConfigMap and string) 
```

Looks like you are failing `go vet`.

Tal-or:
> ```
>  vet: test/e2e/nodepool_nto_performanceprofile_test.go:119:39: invalid operation: performanceProfileConfigMap + nodePool.Name (mismatched types *"k8s.io/api/core/v1".ConfigMap and string) 
> ```
> 
> Looks like you are failing `go vet`.

Yup, fixed it now

sjenning:
I hate to jump is so late here, but there seem to be much simpler solutions to this problem.

One is to just add an annotation to the mirrored configmap in the HCP namespace with the name of the source configmap in the HC namespace.

Another simple solution is to just reuse the configmap name for the mirrored version.  We'd likely have to prefix it to avoid potential collision, but that would still be simpler.

stevekuznetsov:
> Another simple solution is to just reuse the configmap name for the mirrored version. We'd likely have to prefix it to avoid potential collision, but that would still be simpler.

@sjenning isn't that exactly what is implemented here?

```
PerformanceProfileConfigMap(namespace, name, nodePoolName string)
```

`name` is the `ConfigMap` name.

sjenning:
A couple of things.

I don't think we want 129 new lines for fancy string concat.

Also we copy _tons_ for refs from the HC/NP specs into the HCP namespace and we don't do any fancy naming for them in order to map them back to the refs in the HC/NP namespace.

https://github.com/openshift/hypershift/blob/c371fd3af66778a4d15be63e1b7d859946080c11/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go#L1226-L1499

The generic configmap and secret ref copies at the end of that block are enumerated in the tests.

ConfigMap refs
https://github.com/openshift/hypershift/blob/c371fd3af66778a4d15be63e1b7d859946080c11/api/util/configrefs/refs_test.go#L16-L30

Secret refs
https://github.com/openshift/hypershift/blob/c371fd3af66778a4d15be63e1b7d859946080c11/api/util/configrefs/refs_test.go#L274-L293

Those refs actually retain there original names across the mirror, which could result in collision :grimacing: 

The point being, we don't want a fancy naming scheme for this particular case.  We don't have any direct mapping back to the refs on the HC/NP for any of these other refs, of which there are many.

If we do need, for whatever reason, a direct mapping back to the ref in the NP spec, we can annotation the configmap with the original ref name in the NP namespace and leave the original naming scheme in place.

```diff
diff --git a/hypershift-operator/controllers/nodepool/nodepool_controller.go b/hypershift-operator/controllers/nodepool/nodepool_controller.go
index 4514cf748..46b151346 100644
--- a/hypershift-operator/controllers/nodepool/nodepool_controller.go
+++ b/hypershift-operator/controllers/nodepool/nodepool_controller.go
@@ -106,6 +106,8 @@ const (
        nodeTuningGeneratedConfigLabel   = "hypershift.openshift.io/nto-generated-machine-config"
        performanceProfileConfigMapLabel = "hypershift.openshift.io/performanceprofile-config"
 
+       performanceProfileConfigMapNameAnnotation = "hypershift.openshift.io/performanceprofile-configmap-name"
+
        controlPlaneOperatorManagesDecompressAndDecodeConfig = "io.openshift.hypershift.control-plane-operator-manages.decompress-decode-config"
 
        controlPlaneOperatorCreatesDefaultAWSSecurityGroup = "io.openshift.hypershift.control-plane-operator-creates-aws-sg"
@@ -664,7 +666,7 @@ func (r *NodePoolReconciler) reconcile(ctx context.Context, hcluster *hyperv1.Ho
        }
 
        // Validate tuningConfig input.
-       tunedConfig, performanceProfileConfig, err := r.getTuningConfig(ctx, nodePool)
+       tunedConfig, performanceProfileConfig, performanceProfileConfigMapName, err := r.getTuningConfig(ctx, nodePool)
        if err != nil {
                SetStatusCondition(&nodePool.Status.Conditions, hyperv1.NodePoolCondition{
                        Type:               hyperv1.NodePoolValidTuningConfigConditionType,
@@ -724,7 +726,7 @@ func (r *NodePoolReconciler) reconcile(ctx context.Context, hcluster *hyperv1.Ho
                }
        } else {
                if result, err := r.CreateOrUpdate(ctx, r.Client, performanceProfileConfigMap, func() error {
-                       return reconcilePerformanceProfileConfigMap(performanceProfileConfigMap, nodePool, performanceProfileConfig)
+                       return reconcilePerformanceProfileConfigMap(performanceProfileConfigMap, nodePool, performanceProfileConfig, performanceProfileConfigMapName)
                }); err != nil {
                        return ctrl.Result{}, fmt.Errorf("failed to reconcile PerformanceProfile ConfigMap: %w", err)
                } else {
@@ -1618,11 +1620,15 @@ func reconcileTunedConfigMap(tunedConfigMap *corev1.ConfigMap, nodePool *hyperv1
 // reconcilePerformanceProfileConfigMap inserts the PerformanceProfile object manifest in performanceProfileConfig into ConfigMap performanceProfileConfigMap.
 // This is used to mirror the PerformanceProfile object manifest into the control plane namespace, for the Node
 // Tuning Operator to mirror and reconcile in the hosted cluster.
-func reconcilePerformanceProfileConfigMap(performanceProfileConfigMap *corev1.ConfigMap, nodePool *hyperv1.NodePool, performanceProfileConfig string) error {
+func reconcilePerformanceProfileConfigMap(performanceProfileConfigMap *corev1.ConfigMap, nodePool *hyperv1.NodePool, performanceProfileConfig, performanceProfileConfigMapName string) error {
        if err := reconcileNodeTuningConfigMap(performanceProfileConfigMap, nodePool, performanceProfileConfig); err != nil {
                return err
        }
        performanceProfileConfigMap.Labels[performanceProfileConfigMapLabel] = "true"
+       if performanceProfileConfigMap.Annotations == nil {
+               performanceProfileConfigMap.Annotations = make(map[string]string)
+       }
+       performanceProfileConfigMap.Annotations[performanceProfileConfigMapNameAnnotation] = performanceProfileConfigMapName
        return nil
 }
 
@@ -2144,10 +2150,11 @@ func (r *NodePoolReconciler) getConfig(ctx context.Context,
 
 func (r *NodePoolReconciler) getTuningConfig(ctx context.Context,
        nodePool *hyperv1.NodePool,
-) (string, string, error) {
+) (string, string, string, error) {
        var (
                configs                              []corev1.ConfigMap
                tunedAllConfigPlainText              []string
+               performanceProfileConfigMapName      string
                performanceProfileAllConfigPlainText []string
                errors                               []error
        )
@@ -2181,6 +2188,7 @@ func (r *NodePoolReconciler) getTuningConfig(ctx context.Context,
                        tunedAllConfigPlainText = append(tunedAllConfigPlainText, string(manifestTuned))
                }
                if manifestPerformanceProfile != nil {
+                       performanceProfileConfigMapName = config.Name
                        performanceProfileAllConfigPlainText = append(performanceProfileAllConfigPlainText, string(manifestPerformanceProfile))
                }
        }
@@ -2193,7 +2201,7 @@ func (r *NodePoolReconciler) getTuningConfig(ctx context.Context,
        sort.Strings(tunedAllConfigPlainText)
        sort.Strings(performanceProfileAllConfigPlainText)
 
-       return strings.Join(tunedAllConfigPlainText, "\n---\n"), strings.Join(performanceProfileAllConfigPlainText, "\n---\n"), utilerrors.NewAggregate(errors)
+       return strings.Join(tunedAllConfigPlainText, "\n---\n"), strings.Join(performanceProfileAllConfigPlainText, "\n---\n"), performanceProfileConfigMapName, utilerrors.NewAggregate(errors)
 
 }
```

stevekuznetsov:
@sjenning *if* we are to ever derive names of objects from other names of other objects, the "fancy string concat" is strictly required to be correct.

If the pattern is to use an annotation, sounds great, let's do that :)

Tal-or:
@sjenning The rational of this work was to be able to derive the association between the origin and mirrored objects just by peeking at  their names.
With the current approach, we need to inspect the `nodepool.spec.tuningconifg` in order to find the association between the origin and mirrored object.
The approach suggested in this PR can be useful for debugging (even on customer site) and for e2e testing, where you don't need to perform an API call to get the `nodepool`, just perform simple manipulation of strings.

Having an annotation will require inspecting the object as well, this is better than inspecting the `nodepool` but still not intuitive as suggested here.

BTW the original implementation had an hidden bug as @stevekuznetsov mentioned, the length of the name might ended up too long and the object would fail to be created.

sjenning:
That's fine.

I do see a functional issue in that it orphans the old mirrored configmap, the one starting with `perfprofile-`, in the HCP namespace.  We need to clean the up on upgrade if we are going to rename the resource.

stevekuznetsov:
@Tal-or if you do end up changing it to use an annotation, please know your hard work is not wasted. I would like to see your implementation used in `manifests.HostedControlPlaneNamespace(cluster.Namespace, cluster.Name)` - another place we derive one k8s name from many others. We will also have to think about backwards compat there.

Tal-or:
> That's fine.
> 
> I do see a functional issue in that it orphans the old mirrored configmap, the one starting with `perfprofile-`, in the HCP namespace. We need to clean the up on upgrade if we are going to rename the resource.

The feature is going to be announced only at 4.17 so we are good. It was never used just yet

Tal-or:
> @Tal-or if you do end up changing it to use an annotation, please know your hard work is not wasted. I would like to see your implementation used in `manifests.HostedControlPlaneNamespace(cluster.Namespace, cluster.Name)` - another place we derive one k8s name from many others. We will also have to think about backwards compat there.

Thanks @stevekuznetsov appreciate it. I'm waiting for @sjenning final call and I'll change accordingly.

sjenning:
> The feature is going to be announced only at 4.17 so we are good
This configmap was introduced in a commit that is contained in `release-4.16`

https://github.com/openshift/hypershift/pull/1782/files#diff-9cccba29261c004a08ff72e1e72f0fd2160ec0f31690dc94b8352701f5e9e4c8R76

Unless we get this in and backported before 4.16 GA (which is imminent and may already be locked out) there will be a released version of 4.16 out there that will create the old configmap.

Tal-or:
> > The feature is going to be announced only at 4.17 so we are good
> 
> This configmap was introduced in a commit that is contained in `release-4.16`
> 
> https://github.com/openshift/hypershift/pull/1782/files#diff-9cccba29261c004a08ff72e1e72f0fd2160ec0f31690dc94b8352701f5e9e4c8R76
> 
> Unless we get this in and backported before 4.16 GA (which is imminent and may already be locked out) there will be a released version of 4.16 out there that will create the old configmap.

I know it does and I agreed that theoretically there could be a chance for someone to create this performance profile configmap.
But as long as it's not documented and we didn't commit to support performance profile tunings on hypershift in 4.16 ( and we don't) it should be fine, isn't it?

The code on PAO which is actually doing something with this configmap only got merge recently on our master branch (future 4.17 )

sjenning:
TestNodePool/HostedCluster0/Main/TestNTOPerformanceProfile seems to be legit failing
```
    nodepool_test.go:353: incorrect condition: wanted AllMachinesReady=True, got AllMachinesReady=False: Draining(1 of 2 machines are not ready
        Machine example-4nnpt-test-ntoperformanceprofile-qpljn-l2d24: Draining
        )
    nodepool_test.go:353: correct condition: wanted AllNodesHealthy=True, got AllNodesHealthy=True: AsExpected(All is well)
    nodepool_test.go:353: incorrect condition: wanted AllMachinesReady=True, got AllMachinesReady=False: Deleted(1 of 2 machines are not ready
        Machine example-4nnpt-test-ntoperformanceprofile-qpljn-l2d24: Deleted
        )
    nodepool_test.go:353: incorrect condition: wanted AllNodesHealthy=True, got AllNodesHealthy=False: Deleting(Machine example-4nnpt-test-ntoperformanceprofile-qpljn-l2d24: Deleting
        )
    nodepool_test.go:362: Failed to validate NodePool conditions in 10m0s: timed out waiting for the condition
            --- FAIL: TestNodePool/HostedCluster0/Main/TestNTOPerformanceProfile (1365.46s)
```
Maybe the timeout is too short now? fyi @stevekuznetsov

Tal-or:
> TestNodePool/HostedCluster0/Main/TestNTOPerformanceProfile seems to be legit failing
> 
> ```
>     nodepool_test.go:353: incorrect condition: wanted AllMachinesReady=True, got AllMachinesReady=False: Draining(1 of 2 machines are not ready
>         Machine example-4nnpt-test-ntoperformanceprofile-qpljn-l2d24: Draining
>         )
>     nodepool_test.go:353: correct condition: wanted AllNodesHealthy=True, got AllNodesHealthy=True: AsExpected(All is well)
>     nodepool_test.go:353: incorrect condition: wanted AllMachinesReady=True, got AllMachinesReady=False: Deleted(1 of 2 machines are not ready
>         Machine example-4nnpt-test-ntoperformanceprofile-qpljn-l2d24: Deleted
>         )
>     nodepool_test.go:353: incorrect condition: wanted AllNodesHealthy=True, got AllNodesHealthy=False: Deleting(Machine example-4nnpt-test-ntoperformanceprofile-qpljn-l2d24: Deleting
>         )
>     nodepool_test.go:362: Failed to validate NodePool conditions in 10m0s: timed out waiting for the condition
>             --- FAIL: TestNodePool/HostedCluster0/Main/TestNTOPerformanceProfile (1365.46s)
> ```
> 
> Maybe the timeout is too short now? fyi @stevekuznetsov

This is weird, the test already passed in the past

Tal-or:
@stevekuznetsov I'm facing some issues with the e2e tests. They passed in the past I saw you made a serious refactor within this area.

stevekuznetsov:
The refactor should not have impacted any logic - I can't really make heads or tails of the output of the last test, best not to have spam messages, no?

Tal-or:
> The refactor should not have impacted any logic - I can't really make heads or tails of the output of the last test, best not to have spam messages, no?

@stevekuznetsov It was a desperate try to figure out what's going on and why the ConfigMap is not being removed. 
Didn't manage to figure out the root cause just yet.

Tal-or:
@stevekuznetsov @sjenning I found the issue, and it stems from the fact that when we need to delete the performanceProfile ConfigMap on the HCP NP,  we do not have the name of the ConfigMap that the user provided anymore (because it was just removed from the NodePool spec). So the `ComposeName` returns an invalid name and the deletion is falling because the object not found - https://github.com/Tal-or/hypershift/blob/pp_unique_name/hypershift-operator/controllers/nodepool/nodepool_controller.go#L741 

My suggestion is to change the delete function to list all ConfigMap's with the PerformanceProfile selector:
https://github.com/Tal-or/hypershift/blob/pp_unique_name/hypershift-operator/controllers/nodepool/nodepool_controller.go#L107 which would return a single object (because only one PerformanceProfile is allowed) and then remove it. 

WDYT?

csrwng (commented):
One comment

hypershift-operator/controllers/nodepool/manifests.go
```diff
@@ -72,11 +72,49 @@ func TunedConfigMap(namespace, name string) *corev1.ConfigMap {
...
 		},
 	}
 }
+
+// 63 is the qualifiedNameMaxLength allowed for k8s object
+// https://github.com/kubernetes/kubernetes/blob/957c9538670b5f7ead2c9ba9ceb9de081d66caa4/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L34
+const qualifiedNameMaxLength = 63
+
+// ComposeValidName takes two qualified objects` names
+func ComposeValidName(name1, name2 string) string {
```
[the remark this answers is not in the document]
Tal-or: @csrwng thank you for the review.
I understand your concern, but the shortenName function might returns an unreadable name like:
In my case I want the `PerformanceProfile` to be compose of The `ConfigMap` name + `NodePool` name
If i'll use the shortenName function, in some edge cases I might be getting something like:
[hash]-[NodePool name] 

While with this function we have at least some weight prefixing.

OTOH, your argument in the comment is completely right. 
This  PR became much more complex than I anticipated and I would close it since it doesn't worth the effort anymore, but the problem is that the existing implementation has a bug as well:
https://github.com/openshift/hypershift/pull/4108#discussion_r1627581711
> BTW the previous implementation was problematic as well, because we are not checking the length of "perfprofile-+name"

PR #4115, opened 2024-05-30
OCPBUGS-34640: Support for multiple name per component in releaseProvider

jparrill:
This PR adds support for components in the OCP payload that change their names across different OCP releases. Here's how it works: when you invoke the GetImages method, you should provide a slice of strings with the possible names the component could have in different OCP releases. The method will return the first matching name it finds

Fixes #[OCPBUGS-34640](https://issues.redhat.com/browse/OCPBUGS-34640)


enxebre:
thanks!
/lgtm

PR #4129, opened 2024-05-31
OCPBUGS-34807: Fixed ValidReleaseInfo condition

muraee:
Previously the condition was reconciled immediately before accessing any image in the playload. which means the condition will always be `True`

Fixes #

**Checklist**

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
@@ -716,26 +716,6 @@ func (r *HostedControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.R
 	if err != nil {
 		return reconcile.Result{}, fmt.Errorf("failed to look up release image metadata: %w", err)
 	}
-	releaseImageProvider := imageprovider.New(releaseImage)
-	{
-		if len(releaseImageProvider.GetMissingImages()) == 0 {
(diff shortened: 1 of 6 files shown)


control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -301,6 +301,19 @@ func (r *HostedControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.R
 	}
 
 	originalHostedControlPlane := hostedControlPlane.DeepCopy()
+	// Always update status based on the current state of the world.
+	defer func() {
```
enxebre, line 305: I'm not a huge fan of defer and patching, how about we keep the existing structure and have these missing images to be called and patched explicitly in e.g. l768?

enxebre:
This is reporting kuryr-cni and kuryr-controller from the cluster-network-operator deployment, @kyrtapz can we drop those references now?

kyrtapz:
> This is reporting kuryr-cni and kuryr-controller from the cluster-network-operator deployment, @kyrtapz can we drop those references now?

Yes, kuryr is no longer a supported CNI.

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
muraee: updated

muraee:
requires https://github.com/openshift/hypershift/pull/4164

stevekuznetsov (approved):

PR #1023, opened 2022-02-15
reconciliation race protection during upgrade with OPERATE_ON_RELEASE_IMAGE envvar

sjenning:
**What this PR does / why we need it**:
This PR reverts https://github.com/openshift/hypershift/pull/963 and uses a simplier approach to upgrade race protection.

The CPO and HCCO now have an optional `OPERATE_ON_RELEASE_IMAGE` envvar.  If the envvar is specified, the operator will only reconcile if the value of the flag matches `releaseImage` on the HCP.  If the envvar is not provided, the operator will reconcile regardless of `releaseImage` (development environments).

**Which issue(s) this PR fixes** *(optional, use `fixes #<issue_number>(, fixes #<issue_number>, ...)` format, where issue_number might be a GitHub issue, or a Jira story*:
Fixes #

**Checklist**
- [x] Subject and description added to both, commit and PR.
- [ ] Relevant issues have been referenced.
- [ ] This change includes docs. 
- [ ] This change includes unit tests.

PR #4138, opened 2024-06-03
OCPBUGS-35205: Add hypershift-cluster-version-operator image to release providers

muraee:
- allows CVO image to use registry overrides

Fixes #

**Checklist**

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/cvo/params.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/cvo/params.go
@@ -28,11 +28,17 @@ func NewCVOParams(hcp *hyperv1.HostedControlPlane, releaseImageProvider *imagepr
 		CLIImage:                releaseImageProvider.GetImage("cli"),
 		AvailabilityProberImage: releaseImageProvider.GetImage(util.AvailabilityProberImageName),
 		ControlPlaneImage:       util.HCPControlPlaneReleaseImage(hcp),
-		ReleaseImage:            hcp.Spec.ReleaseImage,
+		ReleaseImage:            releaseImageProvider.GetImage("cluster-version-operator"),
 		OwnerRef:                config.OwnerRefFrom(hcp),
 		ClusterID:               hcp.Spec.ClusterID,
 		PlatformType:            hcp.Spec.Platform.Type,
 	}
+	// fallback to hcp.Spec.ReleaseImage if "cluster-version-operator" image is not available.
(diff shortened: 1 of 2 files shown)


support/releaseinfo/static_provider.go
```diff
@@ -29,6 +29,14 @@ func (p *StaticProviderDecorator) Lookup(ctx context.Context, image string, pull
 	if err != nil {
 		return nil, err
 	}
+
+	releaseImage.Spec.Tags = append(releaseImage.Spec.Tags, imageapi.TagReference{
```
csrwng, line 33: why just not add an image [here](https://github.com/openshift/hypershift/blob/main/control-plane-operator/main.go#L357-L364)?

muraee: at that point we don't know what the release image is yet.

csrwng: it is available to the process: https://github.com/openshift/hypershift/blob/e6407b07158d3abf317f02021751c55e44d2166b/control-plane-operator/main.go#L417

muraee: nice, do you know since when we set this env variable? I am asking if we want backport this.

csrwng: Looks like long enough ago that it should be in all supported versions 
https://github.com/openshift/hypershift/pull/1023

muraee: updated

control-plane-operator/controllers/hostedcontrolplane/cvo/params.go
```diff
@@ -28,7 +28,7 @@ func NewCVOParams(hcp *hyperv1.HostedControlPlane, releaseImageProvider *imagepr
 		CLIImage:                releaseImageProvider.GetImage("cli"),
 		AvailabilityProberImage: releaseImageProvider.GetImage(util.AvailabilityProberImageName),
 		ControlPlaneImage:       util.HCPControlPlaneReleaseImage(hcp),
-		ReleaseImage:            hcp.Spec.ReleaseImage,
+		ReleaseImage:            releaseImageProvider.GetImage("hypershift-cluster-version-operator"),
```
csrwng, line 31: Just a couple more things:
- can we just rename this to "cluster-version-operator" ?
- could we fall back to hcp.Spec.ReleaseImage if we cannot get the one from the releaseImageProvider? (I'm thinking of local development use case where we may not be setting the env var)

muraee: updated.

PR #4142, opened 2024-06-03
HOSTEDCP-1709: test/e2e: constrain actions to context, timeout

stevekuznetsov:
test/e2e: constrain actions to context, timeout

Asynchronous actions that serve to make test assertions, etc, must be
constrained by the test context and/or a timeout - otherwise, even tests
that get interrupted will hang forever on these cases.

---

test/e2e: explicitly constrain pod streams

---

/assign @sjenning

### DIFF
--- a/test/e2e/nodepool_kv_advanced_multinet_test.go
+++ b/test/e2e/nodepool_kv_advanced_multinet_test.go
@@ -256,7 +256,7 @@ iptables -t nat -A PREROUTING -p tcp -d %[1]s -j DNAT --to-destination %[2]s
 	if err != nil {
 		return "", err
 	}
-	return e2eutil.RunCommandInPod(k.infra.Ctx(), infraClient, dnsmasqPod.Name, dnsmasqPod.Namespace, []string{"/bin/sh", "-c", command}, dnsmasqPod.Spec.Containers[0].Name)
+	return e2eutil.RunCommandInPod(k.infra.Ctx(), infraClient, dnsmasqPod.Name, dnsmasqPod.Namespace, []string{"/bin/sh", "-c", command}, dnsmasqPod.Spec.Containers[0].Name, 20*time.Minute)
 }
 
 func (k KubeVirtAdvancedMultinetTest) firstMachineAddress() (string, error) {
--- a/test/e2e/util/client.go
+++ b/test/e2e/util/client.go
@@ -2,6 +2,7 @@ package util
 
 import (
 	"fmt"
+	"time"
 
 	"k8s.io/client-go/rest"
 	cr "sigs.k8s.io/controller-runtime"
@@ -16,6 +17,7 @@ func GetConfig() (*rest.Config, error) {
 	}
 	cfg.QPS = 100
 	cfg.Burst = 100
+	cfg.Timeout = 5 * time.Minute
 	return cfg, nil
 }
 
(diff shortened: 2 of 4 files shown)


test/e2e/nodepool_kv_advanced_multinet_test.go
```diff
@@ -256,7 +256,7 @@ iptables -t nat -A PREROUTING -p tcp -d %[1]s -j DNAT --to-destination %[2]s
 	if err != nil {
 		return "", err
 	}
-	return e2eutil.RunCommandInPod(k.infra.Ctx(), infraClient, dnsmasqPod.Name, dnsmasqPod.Namespace, []string{"/bin/sh", "-c", command}, dnsmasqPod.Spec.Containers[0].Name)
+	return e2eutil.RunCommandInPod(k.infra.Ctx(), infraClient, dnsmasqPod.Name, dnsmasqPod.Namespace, []string{"/bin/sh", "-c", command}, dnsmasqPod.Spec.Containers[0].Name, 20*time.Minute)
```
stevekuznetsov, line 259: I have no idea if this takes a while or not, it's the only command we run that has any chance of doing so - figured I would be safe and not get bogged down in finding the perfect number in this PR.

test/e2e/util/client.go
```diff
@@ -16,6 +17,7 @@ func GetConfig() (*rest.Config, error) {
 	}
 	cfg.QPS = 100
 	cfg.Burst = 100
+	cfg.Timeout = 5 * time.Minute
```
stevekuznetsov, line 20: I don't *think* we are running any long-running watches or whatever ...

test/e2e/util/pod_exec.go
```diff
@@ -95,7 +101,7 @@ func (p *PodExecOptions) Run() error {
 		return err
 	}
 
-	pod, err := client.Pods(p.Namespace).Get(context.TODO(), p.PodName, metav1.GetOptions{})
```
stevekuznetsov, line 98: TBH `context.TODO()` should not be allowed in code review if we can afford to do it.

test/e2e/util/util.go
```diff
@@ -847,7 +847,7 @@ func EnsureNetworkPolicies(t *testing.T, ctx context.Context, c crclient.Client,
 			}
 
 			// Validate cluster-version-operator is not allowed to access management KAS.
-			_, err = RunCommandInPod(ctx, c, "cluster-version-operator", hcpNamespace, command, "cluster-version-operator")
+			_, err = RunCommandInPod(ctx, c, "cluster-version-operator", hcpNamespace, command, "cluster-version-operator", 0)
```
stevekuznetsov, line 850: Not super thrilled about how `RunCommandInPod` effectively needs to grow in order to expose all of the pod execution options that are otherwise so nicely captured in the executor struct, but whatever

test/e2e/util/pod_exec.go
```diff
@@ -84,7 +87,10 @@ func (p *PodExecOptions) Validate() error {
 }
 
 // Run executes a validated remote execution against a pod.
-func (p *PodExecOptions) Run() error {
+func (p *PodExecOptions) Run(ctx context.Context) error {
+	if p.Timeout == 0 {
```
sjenning, line 91: can we just specify a const `defaultPodExecTimeout` to 2m rather that have the user explicitly pass 0 and take that as "use default"?

stevekuznetsov:
/retest

```
 error: unable to iterate over layer sha256:019f3c6668f3f04466aa081183f5b6fee5b931b3daeae70e60fed011f249144f from registry.build02.ci.openshift.org/ci-op-l8b2jrhb/release:latest: stream error: stream ID 11; INTERNAL_ERROR; received from peer 
```

PR #4145, opened 2024-06-03
CNV-42626: Transition kubevirt vms to use topology spread constraints

davidvossel:
The descheduler has a policy that knows how to redistribute pods based on their TopologySpreadConstraints. by transitioning VMs from pod anti-affinity to use topo spread constraints, we can integrate better with descheduler logic that will accurately re-distribute VMs after a mgmt/infra cluster is updated.


davidvossel:
/hold

I need to test the update path to ensure this doesn't break updates in an unexpected way

stevekuznetsov (approved):
Logic here seems correct, but I can't speak to the implications on upgrade.

davidvossel:
@stevekuznetsov can i get another review when you get a chance. I had to rebase

PR #4160, opened 2024-06-05
HOSTEDCP-1680: Expose MHC details in NodePools autoRepair API

muraee:

Fixes #

**Checklist**


hypershift-operator/controllers/nodepool/nodepool_controller.go
```diff
@@ -1883,45 +1883,59 @@ func taintsToJSON(taints []hyperv1.Taint) (string, error) {
...
 		}
 	}
 
-	timeOutOverride := nodePool.Annotations[hyperv1.MachineHealthCheckTimeoutAnnotation]
-	if timeOutOverride == "" {
-		timeOutOverride = hc.Annotations[hyperv1.MachineHealthCheckTimeoutAnnotation]
-	}
-	if timeOutOverride != "" {
-		timeOutOverrideTime, err := time.ParseDuration(timeOutOverride)
+	timeoutOverride := hc.Annotations[hyperv1.MachineHealthCheckTimeoutAnnotation]
```
muraee, line 1913: do we need to keep those override annotations? @csrwng @enxebre

api/hypershift/v1beta1/nodepool_types.go
```diff
@@ -372,6 +372,41 @@ type NodePoolManagement struct {
...
 	// +kubebuilder:default=false
 	AutoRepair bool `json:"autoRepair"`
+
+	// AutoRepairSettings health checks settings for machines in the NodePool when AutoRepair is enabled.
+	//
+	// +optional
+	AutoRepairSettings *AutoRepairSettings `json:"autoRepairSettings,omitempty"`
+}
+
+type AutoRepairSettings struct {
```
enxebre, line 382: I think at most we would want to surface UnhealthyConditions and NodeStartupTimeout.
MaxUnhealthy / UnhealthyRange are coming from an initial design were the mhc implemented default remediation itself. Then we moved default remediation into the MachineSet and there's ongoing work upstream for extending remediation capabilities for e.g. limit inflight remediations.

hypershift-operator/controllers/nodepool/nodepool_controller.go
enxebre: yes, we need to keep it for now to enable a smooth transition in future.

api/hypershift/v1beta1/nodepool_types.go
muraee: I thought since Cesar introduced `MachineHealthCheckMaxUnhealthyAnnotation` annotation to override `MaxUnhealthy` there is a use-case already for users to set a different value. 
Do you mean that MaxUnhealthy / UnhealthyRange will be removed/deprecated upstream soon?

csrwng: @muraee I introduced it as an escape hatch for situations in which there's a lot of unhealthy nodes that should be recycled (this occurred once in rosa production already).

muraee: removed MaxUnhealthy / UnhealthyRange and added `UnhealthyConditions` field

csrwng: It might make sense to have our own copy of the type instead of referencing the CAPI type

api/hypershift/v1beta1/nodepool_types.go
```diff
@@ -372,6 +372,43 @@ type NodePoolManagement struct {
...
+type UnhealthyCondition struct {
+	// +kubebuilder:validation:Type=string
+	// +kubebuilder:validation:MinLength=1
+	Type corev1.NodeConditionType `json:"type"`
+
+	// +kubebuilder:validation:Type=string
+	// +kubebuilder:validation:MinLength=1
+	Status corev1.ConditionStatus `json:"status"`
+
+	Timeout metav1.Duration `json:"timeout"`
```
stevekuznetsov, line 411: This is probably the most interesting field since it's not part of a condition normally - document what it's doing.

hypershift-operator/controllers/nodepool/nodepool_controller.go
```diff
@@ -1907,20 +1908,54 @@ func (r *NodePoolReconciler) reconcileMachineHealthCheck(ctx context.Context,
 		if _, err := intstr.GetScaledValueFromIntOrPercent(&maxUnhealthyValue, 100, true); err != nil {
 			log.Error(err, "Cannot parse max unhealthy override duration", "value", maxUnhealthyOverride)
 		} else {
-			maxUnhealthy = maxUnhealthyValue
+			maxUnhealthy = &maxUnhealthyValue
 		}
 	}
 
-	timeOutOverride := nodePool.Annotations[hyperv1.MachineHealthCheckTimeoutAnnotation]
```
stevekuznetsov, line 1914: We want to keep these as annotations? Or add a field for this as well?

muraee: yes we are keeping them for now: https://github.com/openshift/hypershift/pull/4160#discussion_r1627667189
The timeout field is defined per Unhealthy Condition

stevekuznetsov: Er - right now you're `unhealthyConditions = append(unhealthyConditions` - are we requiring that the user remove the annotation *and* update the spec in the same UPDATE call? Perhaps logic to make sure that we don't end up with duplicates based on the current append logic is good.

muraee: the annotation will be ignored if the unhealthyConditions is specified.
I am resetting the array before the append `unhealthyConditions = []capiv1.UnhealthyCondition{}`

api/hypershift/v1beta1/nodepool_types.go
```diff
@@ -382,6 +382,45 @@ type NodePoolManagement struct {
...
+	AutoRepairSettings *AutoRepairSettings `json:"autoRepairSettings,omitempty"`
+}
+
+type AutoRepairSettings struct {
+	// Machines older than this duration without a node will be considered to have
+	// failed and will be remediated.
+	// If not set, this value is defaulted to 20 minutes.
+	// If you wish to disable this feature, set the value explicitly to 0.
+	//
+	// +optional
```
bryan-cox, line 398: Should we default this to in CEL?

enxebre: Why 20 min?
If we are defaulting in the backend, Let's rather default here explicitly.

muraee: 20min is the current value we use.

api/hypershift/v1alpha1/nodepool_types.go
```diff
@@ -378,6 +378,42 @@ type NodePoolManagement struct {
 	// +optional
 	// +kubebuilder:default=false
 	AutoRepair bool `json:"autoRepair"`
+
+	// AutoRepairSettings health checks settings for machines in the NodePool when AutoRepair is enabled.
+	//
+	// +optional
+	AutoRepairSettings *AutoRepairSettings `json:"autoRepairSettings,omitempty"`
```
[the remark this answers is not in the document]
muraee: how would you default `UnhealthyCondition`?

PR #4164, opened 2024-06-05
NO-JIRA: Delete references to the openshift-sdn image in CNO

danwinship:
openshift-sdn is gone in 4.17. There are still a bunch of references to it that need to be cleaned up, but for now, at least stop trying to resolve the image, since that would cause errors even in clusters that aren't trying to use openshift-sdn.

I am not sure that this is correct... it seems like maybe a given version of hypershift supports multiple target versions of openshift? In which case it would need to be able to refer to the 4.16 and earlier sdn images, but not the 4.17 and later sdn images?

Related to SDN-4896; there was no epic for hypershift because I didn't realize that hypershift also referred to the image directly until someone pointed it out...

**Checklist**

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/cno/clusternetworkoperator.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/cno/clusternetworkoperator.go
@@ -39,7 +39,6 @@ const (
 
 type Images struct {
 	NetworkOperator              string
-	SDN                          string
 	KubeProxy                    string
 	KubeRBACProxy                string
 	Multus                       string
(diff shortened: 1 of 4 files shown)


jparrill:
Hey @danwinship thanks for the PR!. Let me clarify the questions:

> I am not sure that this is correct... it seems like maybe a given version of hypershift supports multiple target versions of openshift? In which case it would need to be able to refer to the 4.16 and earlier sdn images, but not the 4.17 and later sdn images?

Yes, depending which version of Hypershift you have, you could deploy several OCP releases. AFAIK openshift-sdn bits were relevant in versions previous to 4.13, so I think these references are not relevant anymore.

jparrill:
/retest
/retitle NO-JIRA: Delete references to the openshift-snd image in CNO

danwinship:
/retest
I can't see the Konflux results so I don't know what it's complaining about.

danwinship:
@jparrill so can this get an lgtm then? Or if you do the "needs separate approve and lgtm" thing can you assign this to someone who can lgtm it?

jparrill:
Hey @danwinship I've pinged my colleges from Hypershift team to tag it. Thank you!

PR #4176, opened 2024-06-07
HOSTEDCP-1446: hack: make the e2e script generic

stevekuznetsov:

### DIFF
--- a/hack/ci-test-e2e.sh
+++ b/hack/ci-test-e2e.sh
@@ -41,9 +41,9 @@ if [[ -n "${REQUEST_SERVING_COMPONENT_TEST}" ]]; then
   --e2e.management-cluster-name=$(cat "${SHARED_DIR}/management_cluster_name")"
 fi
 
-bin/test-e2e \
-  -test.v \
-  -test.run=${CI_TESTS_RUN} \
+declare -a default_args=(
+  -test.v
+  -test.run=${CI_TESTS_RUN}
   -test.parallel=20 \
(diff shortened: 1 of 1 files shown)


stevekuznetsov:
/area testing

If we care strongly that the values of the flags are defined in the repo, as opposed to in the job, we can have separate entrypoints / `make` targets that all call the script.

stevekuznetsov:
```
 hack/ci-test-e2e.sh "" 
```

???

stevekuznetsov:
```
.PHONY: ci-test-e2e
ci-test-e2e:
	hack/ci-test-e2e.sh ${CI_TESTS_RUN}
```

sigh

hack/ci-test-e2e.sh
```diff
@@ -57,6 +57,14 @@ bin/test-e2e \
...
   --e2e.aws-endpoint-access=PublicAndPrivate \
   --e2e.external-dns-domain=service.ci.hypershift.devcluster.openshift.com \
-  ${REQUEST_SERVING_COMPONENT_PARAMS} | tee /tmp/test_out &
+  ${REQUEST_SERVING_COMPONENT_PARAMS}
+)
+
+declare -a args="$@"
+if [[ "$#" -lt 2 ]]; then
+  args="${default_args[@]}"
+fi
```
sjenning, line 66: This is cryptic.  Could do with a comment.

stevekuznetsov: Sure thing, Bash is not the best :)

csrwng:
The script modified by this PR does not impact kubevirt

PR #4182, opened 2024-06-10
OCPBUGS-33951: Documentation for disconnected MGMT Cluster

jparrill:
On the OCP Standalone product documentation, we explicitly says how to imageContentSources should be configured for IPI deployments, I'm following the same approach. Once you perform the oc-mirror stage, the output file with the ImageContentSources, should be pasted as IDMS/ICSP into the Management cluster, making possible to perform a disconnected deployment and explicitly matching the Metadata and Images OCP container images with the mirrored ones in the private registry

Fixes #[OCPBUGS-33951](https://issues.redhat.com/browse/OCPBUGS-33951)

### DIFF
--- a/docs/content/how-to/disconnected/idms-icsp-for-management-clusters.md
+++ b/docs/content/how-to/disconnected/idms-icsp-for-management-clusters.md
@@ -12,6 +12,93 @@ For the Control Plane, ICSP/IDMS objects are managed in the Management Cluster. 
 
 In summary, to work with disconnected registries in the HostedControlPlane, you first need to create the appropriate ICSP/IDMS in the Management Cluster. Then, to deploy disconnected workloads in the Data Plane, you need to add the desired entries into the ImageContentSources field inside the HostedCluster manifest.
 
+### Mirroring Prerequisites for the Management Cluster
+
+Deploying HostedControlPlanes in a disconnected environment involves a mirroring process that is straightforward, provided you have the necessary prerequisites:
+
+1. **Private Registry**: Ensure a private registry is deployed and operational.
+2. **Credentials**: Have a credentials file to pull from a public registry and push to your private registry.
+3. **oc-mirror Tool**: Install the `oc-mirror` tool.
+
+#### Setting Up the Private Registry
+
+For the private registry, robust platforms typically use large-scale solutions like Quay or Artifactory. However, for testing or development environments, you can set up a smaller registry by following these [steps](../../labs/IPv4/registry.md).
+
+#### Preparing the Credentials File
+
+Your credentials file, typically located at `${HOME}/.docker/config.json`, must include login credentials for at least two registries:
+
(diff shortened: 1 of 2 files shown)


docs/content/how-to/disconnected/idms-icsp-for-management-clusters.md
```diff
@@ -12,6 +12,93 @@ For the Control Plane, ICSP/IDMS objects are managed in the Management Cluster.
 
 In summary, to work with disconnected registries in the HostedControlPlane, you first need to create the appropriate ICSP/IDMS in the Management Cluster. Then, to deploy disconnected workloads in the Data Plane, you need to add the desired entries into the ImageContentSources field inside the HostedCluster manifest.
 
+### Mirroring Prerequisited for the Management Cluster
```
sjenning, line 15: s/Prerequisited/Prerequisites/

docs/content/how-to/disconnected/known-issues.md
```diff
@@ -190,7 +190,7 @@ for isName in ${IMAGESTREAMS}; do
 done
 ```
 
-## IDMS/ICSP with only root registry are not propagated to the registry-override flag under ignition server
+## IDMS/ICSP with only root registry are filling to the registry-override flag under ignition server
```
sjenning, line 193: It made more sense with `not propagated` to me.  Not sure what is meant by `filling`.

docs/content/how-to/disconnected/known-issues.md
```diff
@@ -202,7 +202,9 @@ failed to lookup release image: failed to extract release metadata: failed to ge
 ```
 
 **Root Cause**:
-Once the ICSP/IDMS are created in the Management cluster and they are only using in the "sources" side a root registry instead of pointing a registry namespace, the image-overrides are not properly propagated through all the deployments in the HostedControlPlane namespace. This is how looks like:
+Once the ICSP/IDMS are created in the Management cluster and they are only using in the "sources" side a root registry instead of pointing a registry namespace, the image-overrides are not filled with the explicit destination OCP Metadata and Release images, so the Hypeshift operator cannot infer the extact location of the images in the private registry.
+
+This is a sample of how it looks like:
```
sjenning, line 207: delete ` of how it looks like`

sjenning (commented):
In general, I think documentation is supposed to avoid the 2nd person you/your, but that can be done as part of a different effort.

docs/content/how-to/disconnected/known-issues.md
jparrill: With "filling" I was meant that the --registries-override field in the different deployments, is filled with the registries-override content only when an exact match happens with an IDMS/ICSP exists with that record in the Management Cluster.

bryan-cox (approved):
/lgtm

PR #4184, opened 2024-06-10
NO-JIRA: feat(olm): Set packageserver replicas to 2 for IBMCloudPlatform

hasueki:

Fixes #

**Checklist**

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/olm/params.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/olm/params.go
@@ -66,6 +66,10 @@ func NewOperatorLifecycleManagerParams(hcp *hyperv1.HostedControlPlane, releaseI
 	params.PackageServerConfig.SetRestartAnnotation(hcp.ObjectMeta)
 	params.PackageServerConfig.SetDefaultSecurityContext = setDefaultSecurityContext
 
+	if hcp.Spec.Platform.Type == hyperv1.IBMCloudPlatform && hcp.Spec.ControllerAvailabilityPolicy == hyperv1.HighlyAvailable {
+		params.PackageServerConfig.Replicas = 2
+	}
+
 	if hcp.Spec.OLMCatalogPlacement == "management" {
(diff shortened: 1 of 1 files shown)


control-plane-operator/controllers/hostedcontrolplane/olm/params.go
```diff
@@ -66,6 +66,10 @@ func NewOperatorLifecycleManagerParams(hcp *hyperv1.HostedControlPlane, releaseI
 	params.PackageServerConfig.SetRestartAnnotation(hcp.ObjectMeta)
 	params.PackageServerConfig.SetDefaultSecurityContext = setDefaultSecurityContext
 
+	if hcp.Spec.Platform.Type == hyperv1.IBMCloudPlatform && hcp.Spec.ControllerAvailabilityPolicy == hyperv1.HighlyAvailable {
```
rtheis, line 69: I don't think that this should be restricted to IBM cloud if standard OCP uses 2 replicas.

rtheis (changes requested):

control-plane-operator/controllers/hostedcontrolplane/olm/params.go
rtheis: I think https://github.com/openshift/operator-framework-olm/tree/release-4.16/pkg/package-server-manager controls the package server and uses 2 replicas in HA mode.  That would align with what we see in non-HyperShift clusters.

rtheis: After a quick test, I'm wrong and it appears that the OLM operator reconciles the packageserver deployment.

rtheis: And further testing indeed shows that the package server manager reconciles the packageserver csv and the csv controls the number of replicas.  So again, I think 2 replicas is correct.

hasueki: @csrwng @sjenning Do you have any opinions on this? If we think we should set the replicas down to 2 for HyperShift as a whole, I would be happy to make that adjustment.

rtheis (approved):
/lgtm

PR #4186, opened 2024-06-11
HOSTEDCP-1446: hack: remove old arguments and scripts

stevekuznetsov:
Consumers now pass arguments directly in the job, so there's no need to have them here.

Depends on https://github.com/openshift/release/pull/53047
/hold


jparrill:
Dependant PR merged.
/hold cancel
/lgtm
/retest

PR #4190, opened 2024-06-11
OCPBUGS-35252: vendor: bump openshift/api

stevekuznetsov:


csrwng:
/approve

@stevekuznetsov if we're going to create a 4.16 version of this PR we need to bump with the 4.16 branch there.

stevekuznetsov:
@csrwng do we not need to bump this in 4.17? My editor was showing the old content in vendor but the diff here isn't showing the update ...

PR #4191, opened 2024-06-11
HOSTEDCP-1671: Add NodePool API Support for Using Marketplace Images

bryan-cox:
Adds  Marketplace Image Info to Azure NodePool API and enables Azure VMs to be created from the Azure Marketplace 

Fixes [HOSTEDCP-1671](https://issues.redhat.com/browse/HOSTEDCP-1671)

**Checklist**


bryan-cox:
/hold

Holding so Steve doesn't have to keep rebasing https://github.com/openshift/hypershift/pull/4018. I am not ready to merge this anyway until I can retest with updated RHCOS images in the Azure Marketplace.

hypershift-operator/controllers/nodepool/azure.go
```diff
@@ -51,6 +54,24 @@ func azureMachineTemplateSpec(hcluster *hyperv1.HostedCluster, nodePool *hyperv1
 		FailureDomain:          failureDomain(nodePool),
 	}}}
 
+	if nodePool.Spec.Platform.Azure.ImageID != "" {
+		azureMachineTemplate.Template.Spec.Image = &capiazure.Image{ID: ptr.To(nodePool.Spec.Platform.Azure.ImageID)}
+	}
+
+	if nodePool.Spec.Platform.Azure.MarketplaceImageInfo.Publisher != "" {
```
Patryk-Stefanski, line 61: does this make image marketplace takes precedence over image ID ?

cmd/nodepool/azure/create.go
```diff
@@ -54,6 +64,21 @@ func (o *AzurePlatformCreateOptions) UpdateNodePool(_ context.Context, nodePool
...
+		"marketplace-sku":       o.MarketplaceSKU,
+		"marketplace-version":   o.MarketplaceVersion,
+	}
+	if marketplaceImageInfo["marketplace-publisher"] != "" || marketplaceImageInfo["marketplace-offer"] != "" || marketplaceImageInfo["marketplace-sku"] != "" || marketplaceImageInfo["marketplace-version"] != "" {
+		for flag, value := range marketplaceImageInfo {
+			if value == "" {
+				return fmt.Errorf("all marketplace flags (i.e. marketplace-publisher, marketplace-offer, marketplace-sku, marketplace-version) are required when using a marketplace image; the following flag was empty: %s", marketplaceImageInfo[flag])
+			}
+		}
+	}
```
Patryk-Stefanski, line 80: Could we move this to a separate function validateAzureOptions like we have for aws ? I feel like with the development for Azure we will be wanting more and more validation so makes sense to create a new func for it

bryan-cox:
/unhold

Removing explicit hold since - https://github.com/openshift/hypershift/pull/4018 merged

hypershift-operator/controllers/nodepool/azure.go
bryan-cox: No image ID seems to take precedence. See the API where I added this MarketplaceImageInfo. I believe I added the context there.

cmd/nodepool/azure/create.go
bryan-cox: Makes sense. Will update that.

hypershift-operator/controllers/nodepool/azure.go
Patryk-Stefanski: but if both are set then where will the azuremachinetemplate get the image.ID from if were overwriting it with the marketplaceImage down below?

api/hypershift/v1beta1/nodepool_types.go
```diff
@@ -879,15 +879,17 @@ type AzureNodePoolPlatform struct {
...
+	// ImageID is the id of the image to boot from. If both ImageID and MarketplaceImageInfo are provided, ImageID will
+	// take precedence according to CAPZ - https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/f49b5ac4fe91ac2d2439f202f9b7ce5bae17470e/azure/converters/image.go#L29.
 	//
 	// +optional
 	ImageID string `json:"imageID,omitempty"`
 
+	// MarketplaceImageInfo specifies an image to use from the Azure Marketplace.
+	//
+	// +optional
+	MarketplaceImageInfo MarketplaceImage `json:"marketplaceImageInfo,omitempty"`
```
enxebre, line 891: please make this a pointer

api/hypershift/v1beta1/nodepool_types.go
```diff
@@ -879,15 +879,17 @@ type AzureNodePoolPlatform struct {
 	// +required
 	VMSize string `json:"vmsize"`
 
-	// ImageID is the id of the image to boot from. If unset, the default image at the location below will be used and
-	// is expected to exist: subscription/<subscriptionID>/resourceGroups/<resourceGroupName>/providers/Microsoft.Compute/images/rhcos.x86_64.vhd.
-	// The <subscriptionID> and the <resourceGroupName> are expected to be the same resource group documented in the
-	// Hosted Cluster specification respectively, hcluster.Spec.Platform.Azure.SubscriptionID and
-	// hcluster.Spec.Platform.Azure.ResourceGroupName.
+	// ImageID is the id of the image to boot from. If both ImageID and MarketplaceImageInfo are provided, ImageID will
+	// take precedence according to CAPZ - https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/f49b5ac4fe91ac2d2439f202f9b7ce5bae17470e/azure/converters/image.go#L29.
```
enxebre, line 883: is there any value in having both specified? can't we otherwise have them in their own struct wit a union discriminator?

hypershift-operator/controllers/nodepool/azure.go
bryan-cox: This should be fixed now

api/hypershift/v1beta1/nodepool_types.go
bryan-cox: We would want ImageID for self-managed azure I would think.

bryan-cox:
/hold 

Holding for https://github.com/openshift/hypershift/pull/4212

api/hypershift/v1beta1/nodepool_types.go
```diff
@@ -872,21 +872,34 @@ type AgentNodePoolPlatform struct {
...
+
+const (
+	ImageID AzureVMImageType = iota
+	AzureMarketplace
+)
+
+// AzureVMImage represents the different types of image data that can be provided for an Azure VM
+type AzureVMImage struct {
+	Type                 AzureVMImageType  `json:"azureImageType"`
+	ImageID              *AzureImageInfo   `json:"imageID,omitempty"`
```
enxebre, line 885: what happens if you specify both? do we need a union discriminator? also this needs to be documented

api/hypershift/v1beta1/nodepool_types.go
enxebre: you also need the go markers see https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/1027-api-unions/README.md#design-details
See also https://github.com/search?q=repo%3Aopenshift%2Fapi%20unionDiscriminator&type=code for multiple examples

api/hypershift/v1alpha1/nodepool_types.go
```diff
@@ -869,12 +869,44 @@ type AgentNodePoolPlatform struct {
...
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Enum:=ImageID;AzureMarketplace
+	// +unionDiscriminator
+	Type AzureVMImageType `json:"azureImageType"`
+
+	// ImageID is the Azure resource ID of a VHD image to use to boot the Azure VMs from.
+	//
+	// +optional
+	// +unionMember, optional
+	ImageID *AzureImageInfo `json:"imageID,omitempty"`
```
muraee, line 894: can't this just be a `string` type, why do we need a wrapper struct around it?

api/hypershift/v1alpha1/nodepool_types.go
```diff
@@ -869,12 +869,44 @@ type AgentNodePoolPlatform struct {
...
 	VMSize string `json:"vmsize"`
-	// ImageID is the id of the image to boot from. If unset, the default image at the location below will be used:
-	// subscription/$subscriptionID/resourceGroups/$resourceGroupName/providers/Microsoft.Compute/images/rhcos.x86_64.vhd
-	// +optional
-	ImageID string `json:"imageID,omitempty"`
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf", message="Image is immutable"
+	// +kubebuilder:validation:Required
+	// +immutable
+	// +required
+	Image *AzureVMImage `json:"image,omitempty"`
```
muraee, line 909:
```suggestion
	Image AzureVMImage `json:"image"`
```
why is this a pointer with `omitempty` if its required?

api/hypershift/v1alpha1/nodepool_types.go
```diff
@@ -918,6 +950,45 @@ type AzureNodePoolPlatform struct {
...
+	ID *string `json:"imageID,omitempty"`
+}
+
+// MarketplaceImage specifies the information needed to create an Azure VM from an Azure Marketplace image. This struct
+// replicates the same fields found in CAPZ - https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/api/v1beta1/types.go.
+type MarketplaceImage struct {
+	// Publisher is the name of the organization that created the image
+	//
+	// +kubebuilder:validation:MinLength=1
+	Publisher string `json:"publisher,omitempty"`
```
muraee, line 964: `+kubebuilder:validation:MinLength=1` means this is required, right? we don't need `omitempty` then.

muraee: same comment for the rest of the fields in this struct.

api/hypershift/v1beta1/nodepool_types.go
```diff
@@ -872,21 +872,51 @@ type AgentNodePoolPlatform struct {
...
-	// The <subscriptionID> and the <resourceGroupName> are expected to be the same resource group documented in the
-	// Hosted Cluster specification respectively, hcluster.Spec.Platform.Azure.SubscriptionID and
-	// hcluster.Spec.Platform.Azure.ResourceGroupName.
+	// Image is the image to boot the VMs with
 	//
-	// +optional
-	ImageID string `json:"imageID,omitempty"`
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf", message="Image is immutable"
+	// +kubebuilder:validation:Required
+	// +immutable
```
enxebre, line 917: +immutable was a placeholder before CEL supported immutability. It has no effect, can we drop it?

api/hypershift/v1beta1/nodepool_types.go
```diff
@@ -872,21 +872,51 @@ type AgentNodePoolPlatform struct {
...
+	// +required
```
enxebre, line 918: everything is required unless explicitly made optional, can we drop the tag?

api/hypershift/v1beta1/nodepool_types.go
```diff
@@ -872,21 +872,51 @@ type AgentNodePoolPlatform struct {
...
+	Image *AzureVMImage `json:"image,omitempty"`
```
enxebre, line 919: why omitempty here?

cmd/cluster/azure/create.go
```diff
@@ -48,6 +49,11 @@ func bindCoreOptions(opts *RawCreateOptions, flags *pflag.FlagSet) {
 	flags.StringVar(&opts.DiskStorageAccountType, "disk-storage-account-type", opts.DiskStorageAccountType, "The disk storage account type for the OS disks for the VMs.")
 	flags.StringToStringVarP(&opts.ResourceGroupTags, "resource-group-tags", "t", opts.ResourceGroupTags, "Additional tags to apply to the resource group created (e.g. 'key1=value1,key2=value2')")
 	flags.StringVar(&opts.SubnetID, "subnet-id", opts.SubnetID, "The subnet ID where the VMs will be placed.")
+	flags.StringVar(&opts.MarketplacePublisher, "marketplace-publisher", opts.MarketplacePublisher, "The Azure Marketplace image publisher.")
```
enxebre, line 52: based on the code below if MarketplacePublisher != nil all the other new fields are required? where is that communicated?

cmd/cluster/azure/create.go
```diff
@@ -265,6 +276,27 @@ func credentialSecret(namespace, name string) *corev1.Secret {
 }
 
 func (o *CreateOptions) GenerateNodePools(constructor core.DefaultNodePoolConstructor) []*hyperv1.NodePool {
+	var vmImage *hyperv1.AzureVMImage
+	if o.MarketplacePublisher == "" {
```
enxebre, line 280: can we please make sure this assumptions are communicated, enforced and unit tested cc @stevekuznetsov ptal

cmd/nodepool/azure/create.go
```diff
@@ -54,6 +67,41 @@ func (o *AzurePlatformCreateOptions) UpdateNodePool(_ context.Context, nodePool
...
+	var vmImage *hyperv1.AzureVMImage
+	if o.ImageID != "" {
+		vmImage = &hyperv1.AzureVMImage{
+			Type: hyperv1.ImageID,
+			ImageID: &hyperv1.AzureImageInfo{
+				ID: ptr.To(o.ImageID),
+			},
+		}
+	} else {
+		// Validate publisher, offer, sku, and version flags are provided if one of them is not empty
```
enxebre, line 79: I'm missing where this is being enforced? same as https://github.com/openshift/hypershift/pull/4191/commits/55b2a197f07fe55f04f8df184288ef8f91e42f66#r1655066371

cmd/cluster/azure/create.go
bryan-cox: It's been added here - https://github.com/openshift/hypershift/blob/82c92662715c65de4e9b6c855c71e23a68ac88a2/cmd/cluster/azure/create.go#L114

cmd/nodepool/azure/create.go
bryan-cox: That's taken care of here - https://github.com/openshift/hypershift/blob/82c92662715c65de4e9b6c855c71e23a68ac88a2/cmd/nodepool/azure/create.go#L83

cmd/cluster/azure/create.go
```diff
@@ -100,6 +110,22 @@ func (o *RawCreateOptions) Validate(ctx context.Context, opts *core.CreateOption
 	if o.DiskEncryptionSetID != "" && o.ResourceGroupName == "" {
 		return nil, fmt.Errorf("--resource-group-name is required when using --disk-encryption-set-id")
 	}
+
+	// Validate publisher, offer, sku, and version flags are provided if one of them is not empty
+	marketplaceImageInfo := map[string]string{
+		"marketplace-publisher": o.MarketplacePublisher,
+		"marketplace-offer":     o.MarketplaceOffer,
```
enxebre, line 117: please include test cases for this, see TestCreateCluster

bryan-cox: unit test added

api/hypershift/v1alpha1/nodepool_types.go
```diff
@@ -869,16 +869,57 @@ type AgentNodePoolPlatform struct {
...
+	// Image is the image to boot the VMs with
+	//
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf", message="Image is immutable"
+	// +kubebuilder:validation:Required
+	Image AzureVMImage `json:"image"`
+
+	// DiskSizeGB is the size in GB to assign to the OS disk
+	// CAPZ default is 30GB, https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/b3708019a67ff19407b87d63c402af94ca4246f6/api/v1beta1/types.go#L599
+	//
+	// +kubebuilder:default:=30
```
enxebre, line 918: this default value change seems is sneaking in, what's the rationale?

cmd/cluster/azure/create.go
```diff
@@ -100,6 +110,22 @@ func (o *RawCreateOptions) Validate(ctx context.Context, opts *core.CreateOption
 	if o.DiskEncryptionSetID != "" && o.ResourceGroupName == "" {
 		return nil, fmt.Errorf("--resource-group-name is required when using --disk-encryption-set-id")
 	}
+
+	// Validate publisher, offer, sku, and version flags are provided if one of them is not empty
+	marketplaceImageInfo := map[string]string{
```
stevekuznetsov, line 115: You don't need to allocate a map for this, could just

```go
for flag, value := range map[string]*string{
    "marketplace-publisher": &o.MarketplacePublisher,
    /* rest */
} {
    if value == "" {
		return nil, fmt.Errorf("all marketplace flags (i.e. marketplace-publisher, marketplace-offer, marketplace-sku, marketplace-version) are required when using a marketplace image; the following flag was empty: %s", marketplaceImageInfo[flag])
    }
}
```

cmd/nodepool/azure/create.go
```diff
@@ -42,6 +48,12 @@ func NewCreateCommand(coreOpts *core.CreateNodePoolOptions) *cobra.Command {
 	cmd.Flags().BoolVar(&platformOpts.EnableEphemeralOSDisk, "enable-ephemeral-disk", platformOpts.EnableEphemeralOSDisk, "If enabled, the Azure VMs in the NodePool will be setup with ephemeral OS disks")
 	cmd.Flags().StringVar(&platformOpts.DiskStorageAccountType, "disk-storage-account-type", platformOpts.DiskStorageAccountType, "The disk storage account type for the OS disks for the VMs.")
 	cmd.Flags().StringVar(&platformOpts.SubnetID, "subnet-id", platformOpts.SubnetID, "The subnet id where the VMs will be placed.")
+	cmd.Flags().StringVar(&platformOpts.ImageID, "image-id", platformOpts.ImageID, "The Image ID to boot the VMs with.")
```
stevekuznetsov, line 51: Please create an `AzureMarketplaceOptions` struct, its own `BindOpts()` and use composition to embed it into the nodepool command. See how the KubeVirt node pool options are embedded into the KubeVirt cluster options for an example.

stevekuznetsov (changes requested):
We are set up now with good examples of how small options structs with their own flag binding and validation methods allow for composition. We should make sure we follow those patterns instead of duplicating code across commands as it's much easier to maintain.

api/hypershift/v1alpha1/nodepool_types.go
bryan-cox: I copied and pasted that in when I was updating the API fields so that the unit tests would pass. I set this back to what the default was before.

cmd/cluster/azure/create.go
bryan-cox: Ok, I updated this a bit more. It's okay if all the flags are empty; if one is set, the others need set as well.

cmd/nodepool/azure/create.go
bryan-cox: This should be resolved now.

cmd/util/azure.go
```diff
@@ -58,3 +58,28 @@ func ReadCredentials(path string) (*AzureCreds, error) {
 
 	return &result, nil
 }
+
+// ValidateMarketplaceFlags validates if any marketplace flag was used, all were set to a non-empty value
+func ValidateMarketplaceFlags(marketplaceFlags map[string]*string) error {
```
enxebre, line 63: fwiw there's also builtin MarkFlagsRequiredTogether cc @stevekuznetsov

bryan-cox: Unless there's a desire to do that, I'd rather leave what I have. 

It looks like I would need to pass in cmd in several different functions just to use it here.

enxebre:
/hold
until there's managed=aro e2e in place

api/hypershift/v1alpha1/nodepool_types.go
```diff
@@ -888,16 +888,55 @@ type AgentNodePoolPlatform struct {
...
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Enum:=ImageID;AzureMarketplace
+	// +unionDiscriminator
+	Type AzureVMImageType `json:"azureImageType"`
+
+	// ImageID is the Azure resource ID of a VHD image to use to boot the Azure VMs from.
+	//
 	// +optional
+	// +unionMember, optional
```
enxebre, line 912: This should be preferably a *string

api/hypershift/v1alpha1/nodepool_types.go
```diff
@@ -888,16 +888,55 @@ type AgentNodePoolPlatform struct {
...
+	//
+	// +optional
+	// +unionMember, optional
+	MarketplaceImageInfo *MarketplaceImage `json:"marketplaceImage,omitempty"`
+}
+
+type AzureNodePoolPlatform struct {
+	// VMSize is the Azure VM instance type to use for the nodes being created in the nodepool.
+	//
+	// +kubebuilder:validation:Required
```
enxebre, line 925: fwiw everything is required unless told otherwise. Shouldn't we have a len>0 validation here?

api/hypershift/v1alpha1/nodepool_types.go
```diff
@@ -888,16 +888,55 @@ type AgentNodePoolPlatform struct {
...
+	VMSize string `json:"vmsize"`
+
+	// Image is the image to boot the VMs with
+	//
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf", message="Image is immutable"
+	// +kubebuilder:validation:Required
```
enxebre, line 931: Shouldn't we have a len>0 validation here

hypershift-operator/controllers/nodepool/azure.go
```diff
@@ -33,11 +33,14 @@ func azureMachineTemplateSpec(hcluster *hyperv1.HostedCluster, nodePool *hyperv1
 		return nil, fmt.Errorf("failed to determine subnet name for Azure machine: %w", err)
 	}
 
+	if nodePool.Spec.Platform.Azure.Image.ImageID == "" && nodePool.Spec.Platform.Azure.Image.MarketplaceImageInfo == nil {
```
enxebre, line 36: This should never happen by design with the cel validation you put in place right? can we add a // comment stating that?
Also can the error indicate user action, e.g. "either ImageID or MarketplaceImageInfo need to be filled"

hypershift-operator/controllers/nodepool/azure.go
```diff
@@ -49,6 +52,25 @@ func azureMachineTemplateSpec(hcluster *hyperv1.HostedCluster, nodePool *hyperv1
 		FailureDomain: failureDomain(nodePool),
 	}}}
 
+	switch nodePool.Spec.Platform.Azure.Image.Type {
```
enxebre, line 55: can all this be a unit tested function?

api/hypershift/v1alpha1/nodepool_types.go
```diff
@@ -888,16 +888,55 @@ type AgentNodePoolPlatform struct {
...
+	// ImageID is the Azure resource ID of a VHD image to use to boot the Azure VMs from.
+	//
 	// +optional
+	// +unionMember, optional
 	ImageID string `json:"imageID,omitempty"`
+
+	// MarketplaceImageInfo contains the Azure Marketplace image info to use to boot the Azure VMs from.
+	//
+	// +optional
+	// +unionMember, optional
```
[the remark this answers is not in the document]
JoelSpeed: All union member fields must be `+optional` since they become required (you'll need CEL for this for now) when the discriminator matches their value.

The `+unionMember,optional`, means that this extra info is optional even when the discriminator points to it.

It's not typically used that often, but I suspect looking at this API, you don't want optional members.

You can see examples here to help, https://github.com/openshift/api/blob/465787efd0d69f12beaad9410a3812d823e78c2b/example/v1/types_stable.go#L108-L125

api/hypershift/v1alpha1/nodepool_types.go
```diff
@@ -888,16 +888,55 @@ type AgentNodePoolPlatform struct {
...
+	MarketplaceImageInfo *MarketplaceImage `json:"marketplaceImage,omitempty"`
```
JoelSpeed, line 919: This should match the discriminant value, so should be AzureMarketplace

cmd/cluster/azure/testdata/zz_fixture_TestCreateCluster_minimal_flags_necessary_to_render.yaml
```diff
@@ -131,7 +131,10 @@ spec:
   platform:
     azure:
       diskSizeGB: 120
-      imageID: fakeBootImageID
+      image:
```
enxebre, line 134: can we have a test case for the type marketplace?

api/hypershift/v1alpha1/nodepool_types.go
JoelSpeed: > fwiw everything is required unless told otherwise.

The rules for this are a little hairy and depend on package level markers and whether something is `omitempty` or not, so we normally recommend that every API field is marked explicitly as required or optional, we have had APIs accidentally switch required fields to optional in the past because of these non-obvious rules

cmd/cluster/azure/testdata/zz_fixture_TestCreateCluster_minimal_flags_necessary_to_render.yaml
enxebre: nvm I saw that's created in other commit

api/hypershift/v1alpha1/nodepool_types.go
JoelSpeed: Since this is a struct I don't think you need a length check

cmd/nodepool/azure/create.go
```diff
@@ -2,69 +2,183 @@ package azure
...
 	cmd := &cobra.Command{
 		Use:          "azure",
 		Short:        "Creates basic functional NodePool resources for Azure platform",
 		SilenceUsage: true,
 	}
+	BindDeveloperOptions(platformOpts, cmd.Flags())
+	cmd.RunE = func(cmd *cobra.Command, args []string) error {
+		validOpts, err := platformOpts.Validate()
+		if err != nil {
+			return err
```
enxebre, line 132: can we have test Data fixture for this same than we have for cluster create?

api/hypershift/v1alpha1/nodepool_types.go
```diff
@@ -911,32 +950,75 @@ type AzureNodePoolPlatform struct {
...
+	// new version becomes available.
+	//
+	// +kubebuilder:validation:MinLength=1
+	Version string `json:"version"`
+
+	// ThirdPartyImage indicates the image is published by a third party publisher and a Plan will be generated for it.
+	//
+	// +kubebuilder:default=false
+	// +optional
+	ThirdPartyImage bool `json:"thirdPartyImage"`
```
JoelSpeed, line 1019: Prefer not to use bools, but enums with meaningful string values, they evolve better

api/hypershift/v1alpha1/nodepool_types.go
```diff
@@ -911,32 +950,75 @@ type AzureNodePoolPlatform struct {
...
+	// +kubebuilder:validation:MinLength=1
+	SKU string `json:"sku"`
+
+	// Version specifies the version of an image sku. The allowed formats are Major.Minor.Build or 'latest'. Major,
+	// Minor, and Build are decimal numbers. Specify 'latest' to use the latest version of an image available at
+	// deployment time. Even if you use 'latest', the VM image will not automatically update after deploy time even if a
+	// new version becomes available.
+	//
+	// +kubebuilder:validation:MinLength=1
+	Version string `json:"version"`
```
JoelSpeed, line 1013: You should consider for each of these, what a sensible maximum length would be, any associated character sets that are or are not allowed, any patterns that should be matched

api/hypershift/v1alpha1/nodepool_types.go
enxebre: The criteria in this pkg it's been all is required unless mark as optional.
If we want to be explicit which I agree we should, let's follow up to do it consistently for all the API.

api/hypershift/v1alpha1/nodepool_types.go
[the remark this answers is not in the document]
bryan-cox: Thanks for the info! Yeah I don't think we want the optional part in `+unionMember,optional` given the statement in the example Joel provided `requiredMember is required when type is RequiredMember, and forbidden otherwise`

api/hypershift/v1alpha1/nodepool_types.go
bryan-cox: Created a story for our backlog - [HOSTEDCP-1825](https://issues.redhat.com/browse/HOSTEDCP-1825)

api/hypershift/v1alpha1/nodepool_types.go
bryan-cox: Ack, I just used the same CEL requirements CAPZ had. 

@bennerv would you happen to know if there are any such patterns, maximum lengths, etc. we should watch out for here?

api/hypershift/v1alpha1/nodepool_types.go
bryan-cox: Ack, I don't have a strong preference either way. @enxebre do you have a preference here? 

I implemented it as a bool since CAPZ has this as a bool value so I thought it would be easier to just copy the value over from our API when creating the Azure machine template.

cmd/nodepool/azure/create.go
bryan-cox: We have one for the new marketplace image unless you are referring to something else? cmd/cluster/azure/testdata/zz_fixture_TestCreateCluster_create_with_a_ure_marketplace_image.yaml

api/hypershift/v1alpha1/nodepool_types.go
enxebre: do we even need this field?

bryan-cox: I'll drop it. I talked with @bennerv and it doesn't seem like we need it - https://redhat-internal.slack.com/archives/C05Q232L29E/p1720706367758479?thread_ts=1718196807.708239&cid=C05Q232L29E.

api/hypershift/v1alpha1/nodepool_types.go
bennerv: https://learn.microsoft.com/en-us/partner-center/marketplace-offers/product-ingestion-api-vm#resource-api-reference

https://ingestionapi-swagger.azureedge.net/#/
```
$ az vm image list --publisher azureopenshift --offer aro4 --all  -otable
Architecture    Offer    Publisher       Sku      Urn                                          Version
--------------  -------  --------------  -------  -------------------------------------------  ---------------
x64             aro4     azureopenshift  411-v2   azureopenshift:aro4:411-v2:411.86.20220811   411.86.20220811
x64             aro4     azureopenshift  411-v2   azureopenshift:aro4:411-v2:411.86.20221004   411.86.20221004
```

- `publisher` I think this corresponds to externalID in the [product API](https://schema.mp.microsoft.com/schema/product/2022-03-01-preview3)
- "offer" corresponds to the [product API](https://schema.mp.microsoft.com/schema/product/2022-03-01-preview3)
    - This is hard to parse, I can't find it. 
- "sku" corresponds to the [virtual machine plan technical configuration](https://schema.mp.microsoft.com/schema/virtual-machine-plan-technical-configuration/2022-03-01-preview3)
    - pattern: `"^[a-z0-9-_]\u002B$"`
- version corresponds to the [same as above](https://schema.mp.microsoft.com/schema/virtual-machine-plan-technical-configuration/2022-03-01-preview3)
```  "versionNumber": {
          "minLength": 1,
          "maxLength": 32,
          "pattern": "^[0-9]\u002B\\.[0-9]\u002B\\.[0-9]\u002B$"
        },
```

hypershift-operator/controllers/nodepool/azure.go
bryan-cox: Yeah I added a unit test for the whole function with several scenarios.

api/hypershift/v1alpha1/nodepool_types.go
bryan-cox: Ok so publisher should have (to save ppl clicking a link):
```
            "pattern": "^[a-z0-9][a-z0-9-_]{2,49}$",
              "minLength": 3,
              "maxLength": 50,
```

api/hypershift/v1alpha1/nodepool_types.go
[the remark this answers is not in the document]
bryan-cox: This looks right to me but I keep getting an error when I try to apply the CR. Any suggestions?
```
// AzureVMImageType is a union discriminator for AzureVMImage
// +kubebuilder:validation:Enum:=ImageID;AzureMarketplace
type AzureVMImageType string

const (
 ImageID          AzureVMImageType = "ImageID"
 AzureMarketplace AzureVMImageType = "AzureMarketplace"
)

// AzureVMImage represents the different types of image data that can be provided for an Azure VM.
// +kubebuilder:validation:XValidation:rule="has(self.type) && self.type =='ImageID' ? has(self.imageID) : !has(self.imageID)",message="ImageID is required when type is ImageID, and forbidden otherwise"
// +kubebuilder:validation:XValidation:rule="has(self.type) && self.type == 'AzureMarketplace' ? has(self.azureMarketplace) : !has(self.azureMarketplace)",message="AzureMarketplace is required when type is AzureMarketplace, and forbidden otherwise"
// +union
type AzureVMImage struct {
 // Type is the type of image data that will be provided to the Azure VM. This can be either "ImageID" or
 // "AzureMarketplace".
 //
 // +unionDiscriminator
 Type AzureVMImageType `json:"azureImageType"`

 // ImageID is the Azure resource ID of a VHD image to use to boot the Azure VMs from.
 //
 // +optional
 // +unionMember
 ImageID *string `json:"imageID,omitempty"`

 // AzureMarketplace contains the Azure Marketplace image info to use to boot the Azure VMs from.
 //
 // +optional
 // +unionMember
 AzureMarketplace *MarketplaceImage `json:"azureMarketplace,omitempty"`
}
```

The error message
```
Error: CustomResourceDefinition.apiextensions.k8s.io "nodepools.hypershift.openshift.io" is invalid: [spec.versions[1].schema.openAPIV3Schema.properties[spec].properties[platform].properties[azure].properties[image].allOf[0].x-kubernetes-validations: Forbidden: must be empty to be structural, spec.versions[1].schema.openAPIV3Schema.properties[spec].properties[platform].properties[azure].properties[image].allOf[1].x-kubernetes-validations: Forbidden: must be empty to be structural]
CustomResourceDefinition.apiextensions.k8s.io "nodepools.hypershift.openshift.io" is invalid: [spec.versions[1].schema.openAPIV3Schema.properties[spec].properties[platform].properties[azure].properties[image].allOf[0].x-kubernetes-validations: Forbidden: must be empty to be structural, spec.versions[1].schema.openAPIV3Schema.properties[spec].properties[platform].properties[azure].properties[image].allOf[1].x-kubernetes-validations: Forbidden: must be empty to be structural]
```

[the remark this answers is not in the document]
JoelSpeed: Can you either push the changes or show the part of the generated openAPI schema that it's referring to?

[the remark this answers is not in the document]
bryan-cox: I pushed the changes in the latest commit - https://github.com/openshift/hypershift/pull/4191/commits/bcbffa511c8da47d16be3c421d6b18c640957490

api/hypershift/v1alpha1/nodepool_types.go
bryan-cox: Ok these should be added now. I had to sub out some of the unicode, like `\u002B`

api/hypershift/v1alpha1/nodepool_types.go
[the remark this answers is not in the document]
bryan-cox: Planning to follow up on this in a separate PR - https://github.com/openshift/hypershift/pull/4460

api/hypershift/v1alpha1/nodepool_types.go
```diff
@@ -888,16 +888,55 @@ type AgentNodePoolPlatform struct {
 	AgentLabelSelector *metav1.LabelSelector `json:"agentLabelSelector,omitempty"`
 }
 
+type AzureVMImageType string
```
enxebre, line 891: why are we adding stuff into alpha API?

api/hypershift/v1beta1/nodepool_types.go
```diff
@@ -924,8 +954,9 @@ type AzureNodePoolPlatform struct {
 	// Hosted Cluster specification respectively, HostedCluster.Spec.Platform.Azure.SubscriptionID and
 	// HostedCluster.Spec.Platform.Azure.ResourceGroupName.
 	//
-	// +optional
-	ImageID string `json:"imageID,omitempty"`
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf", message="Image is immutable"
```
enxebre, line 957: why is the image immutable?

api/hypershift/v1beta1/nodepool_types.go
```diff
@@ -997,6 +1026,39 @@ type AzureNodePoolPlatform struct {
 	MachineIdentityID string `json:"machineIdentityID,omitempty"`
 }
 
+// MarketplaceImage specifies the information needed to create an Azure VM from an Azure Marketplace image. This struct
+// replicates the same fields found in CAPZ - https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/api/v1beta1/types.go.
+type MarketplaceImage struct {
```
enxebre, line 1031: why does this need to be API facing? wouldn't all this be known information owned by us that would make it an impl detail?

bryan-cox: `AzureMachineTemplate` needs these fields for an Azure Marketplace image ([ref here](https://capz.sigs.k8s.io/topics/custom-images.html#using-azure-marketplace)).

enxebre: I think is fine to expose that in the API to enable compositional use cases. But I'd expect we will be able to infer the happy path Architecture/Offer/Publisher/Sku out of a given nodepool.spec.release. This happy path use case where we don't want to put the burden in the API consumer to choose the happy path default image values is what I'm not seeing captured in the API

api/hypershift/v1alpha1/nodepool_types.go
bryan-cox: If its not there, the unit tests are failing on `TestFuzzyConversion/for_NodePool`

api/hypershift/v1beta1/nodepool_types.go
enxebre: let's move forward the PR. Captured this follow up here https://issues.redhat.com/browse/HOSTEDCP-1858

PR #4193, opened 2024-06-11
OCPBUGS-35369: OCPBUGS-35252: Enforce Privileged Instead of Restricted PSA in 4.16

Evan-Reilly:
Align HyperShift with OpenShift's expectation that PSA is going back to enforcing privileged by default instead of restricted.

Fixes #
- https://issues.redhat.com/browse/OCPBUGS-35252

**Checklist**

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/clusterpolicy/config.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/clusterpolicy/config.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"path"
+	"strings"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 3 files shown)


jparrill:
/retitle OCPBUGS-35252: Enforce Privileged Instead of Restricted PSA in 4.16
/ok-to-test

jparrill:
is this part of this other effort? https://github.com/openshift/hypershift/pull/4190

jparrill:
Hey @Evan-Reilly  a couple of questions:

- is this part of this other effort? https://github.com/openshift/hypershift/pull/4190
- why this is not set against main first and then backported to 4.16?

stevekuznetsov:
/hold

I think the correct fix here is https://github.com/openshift/hypershift/pull/4197

stevekuznetsov:
@Evan-Reilly our understanding was the change in o/api would be sufficient as per #4197 - is that correct?

Evan-Reilly:
@stevekuznetsov I believe these changes would still be required as they were required in 4.15.

stevekuznetsov:
I guess I would have expected the `strings.HasPrefix()` to never fire since they moved the gate into a different class but if you can re-test and show it's working & required, let's merge this in.

rtheis:
@Evan-Reilly looks like you need to fix a PR test to go along with this change:

image

rtheis (approved):
/lgtm

bryan-cox (approved):
/approve

PR #4197, opened 2024-06-12
OCPBUGS-35369: OCPBUGS-35252: vendor: bump openshift/api

stevekuznetsov:


stevekuznetsov:
```
   * could not run steps: step src failed: error occurred handling build src-amd64: could not get build src-amd64: builds.build.openshift.io "src-amd64" not found 
```

:thinking: 

/test all

PR #4199, opened 2024-06-12
[release-4.15] NO-JIRA: hack: make the e2e script generic

stevekuznetsov:

Cherry-pick of https://github.com/openshift/hypershift/pull/4198

/assign @sjenning

### DIFF
--- a/hack/ci-test-e2e.sh
+++ b/hack/ci-test-e2e.sh
@@ -41,7 +41,7 @@ if [[ -n "${REQUEST_SERVING_COMPONENT_TEST}" ]]; then
   --e2e.management-cluster-name=$(cat "${SHARED_DIR}/management_cluster_name")"
 fi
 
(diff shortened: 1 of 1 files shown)


stevekuznetsov:
```
    util.go:436: Container csi-liveness-probe in pod kubevirt-csi-controller-77bb96bc78-86xgj has a restartCount > 0 (1)
W0613 16:12:21.932968       1 connection.go:183] Still connecting to unix:///csi/csi.sock
W0613 16:12:31.932872       1 connection.go:183] Still connecting to unix:///csi/csi.sock
W0613 16:12:41.933435       1 connection.go:183] Still connecting to unix:///csi/csi.sock
F0613 16:12:41.933503       1 main.go:146] failed to establish connection to CSI driver: context deadline exceeded
```

The e2e ran normally, however - @sjenning  I think we might be good

PR #4201, opened 2024-06-12
[release-4.14] NO-JIRA: hack: make the e2e script generic

stevekuznetsov:
Today, we duplicate the ci-test-e2e.sh script for Azure, and the KubeVirt tests don't even use it. This has a bunch of downsides, most notably that we don't even have jUnit for KubeVirt tests. We should aim to have only the generic, required, static flags in this script and allow each caller to simply pass their own as necessary.

Cherry-pick of https://github.com/openshift/hypershift/pull/4199

/assign @sjenning

PR #4202, opened 2024-06-12
[release-4.13] NO-JIRA: hack: make the e2e script generic

stevekuznetsov:

Cherry-pick of https://github.com/openshift/hypershift/pull/4201

/assign @sjenning


stevekuznetsov:
requires https://github.com/openshift/release/pull/53146

stevekuznetsov:
```
 flag provided but not defined: -e2e.management-parent-kubeconfig 
```

:thinking:

PR #4207, opened 2024-06-12
NO-JIRA: Fix the getting started page by adding export to BUCKET_NAME

marcbrown2024:
BUCKET_NAME is required to be exported to be accessed by the process that creates policy.json

**Checklist**


bryan-cox:
Thanks @marcbrown2024 , did you run into this while using a Mac? I don't think this issue exists when following those instructions on Linux. It doesn't hurt to export it either way though.

bryan-cox:
/retitle NO-JIRA: Fix the getting started page by adding export to BUCKET_NAME

marcbrown2024:
> Thanks @marcbrown2024 , did you run into this while using a Mac? I don't think this issue exists when following those instructions on Linux. It doesn't hurt to export it either way though.

I was using linux, fedora to be specific.

stevekuznetsov (approved):

PR #4210, opened 2024-06-13
HOSTEDCP-1542: Fixed infra-id not being defaulted first

muraee:
**What this PR does / why we need it**:
- Fixes `infra-id` not being defaulted first
- Fixes platform specific validations not being executed

follow-up of https://github.com/openshift/hypershift/pull/4018 

**Which issue(s) this PR fixes** *(optional, use `fixes #<issue_number>(, fixes #<issue_number>, ...)` format, where issue_number might be a GitHub issue, or a Jira story*:
Fixes #

**Checklist**
- [ ] Subject and description added to both, commit and PR.
- [ ] Relevant issues have been referenced.
- [ ] This change includes docs. 
- [ ] This change includes unit tests.

PR #4212, opened 2024-06-13
HOSTEDCP-1542: cmd: use validated and completed opts pattern

stevekuznetsov:
cmd: use validated and completed opts pattern

Humans make mistakes and the compiler is here to help. The pattern in
this commit is esoteric, but widespread in the Kubernetes ecosystem for
creating options structs that *must* be validated and completed before
being used. The type gymanstics are not ideal but the end experience is
not degraded since we're embedding everything.

In total, the pattern has a couple benefits:
 - enforcement of validation and completion
 - clear distinction between user input (via flags) and computed input
 - obvious flow for re-use in code consumers

---

Follow-up to #4210 

/assign @muraee @enxebre


stevekuznetsov:
```
    util.go:533: Container haproxy in pod ignition-server-proxy-5d456f8ddf-qz2lf has a restartCount > 0 (2)
[NOTICE]   (1) : haproxy version is 2.8.5-aaba8d0
[NOTICE]   (1) : path to executable is /usr/sbin/haproxy
[ALERT]    (1) : config : [/tmp/haproxy.conf:12] : 'server ignition_servers/ignition-server' : could not resolve address 'ignition-server'.
[ALERT]    (1) : config : Failed to initialize server(s) addr.
```

hm

/test e2e-aws

stevekuznetsov:
Kubevirt def broken

```
    hypershift_framework.go:290: failed to create cluster, tearing down: failed to validate core options: specified arch is not supported: 
```

will look in the morning

cmd/cluster/azure/create.go
```diff
@@ -76,54 +80,91 @@ type AzureEncryptionKey struct {
...
+	encryptionKey *AzureEncryptionKey
+	creds         util.AzureCreds
+}
+
+type CreateOptions struct {
+	// Embed a private pointer that cannot be instantiated outside of this package.
+	*completedCreateOptions
+}
+
+func (o *ValidatedCreateOptions) Complete(ctx context.Context, opts *core.CreateOptions) (core.Platform, error) {
```
enxebre, line 127: Follow up nit:
Any reason why Complete wouldn't be called Default()? not sure why we deviated from validate/default naming.

Also Can we make sure we // document all Public funcs?

enxebre:
lgtm overall. Need to do some testing

cmd/cluster/azure/create.go
stevekuznetsov: I haven't seen `Default()`, that's new to me - and in most cases it's *not* defaulting. For instance the cluster creation logic that reaches out and builds infra in the cloud, or reads the infrastructure JSON, that's not defaulting. The majority of what happens in `Complete()` is that kind of thing from what I've seen.

sjenning:

PR #4217, opened 2024-06-14
NO-JIRA: Don't run ValidateMetrics on teardown if test has already failed

muraee:
Don't run ValidateMetrics on teardown if test has already failed

Fixes #

**Checklist**

### DIFF
--- a/test/e2e/util/hypershift_framework.go
+++ b/test/e2e/util/hypershift_framework.go
@@ -160,6 +160,18 @@ func (h *hypershiftTest) teardown(hostedCluster *hyperv1.HostedCluster, opts *Pl
 	}
 
 	h.Run("Teardown", func(t *testing.T) {
+		if !h.Failed() {
+			ValidateMetrics(t, h.ctx, hostedCluster, []string{
+				hcmetrics.SilenceAlertsMetricName,
+				hcmetrics.LimitedSupportEnabledMetricName,
+				hcmetrics.ProxyMetricName,
+				hcmetrics.InvalidAwsCredsMetricName,
+				HypershiftOperatorInfoName,
+				npmetrics.SizeMetricName,
+				npmetrics.AvailableReplicasMetricName,
+			}, true)
+
+		}
 		teardownHostedCluster(t, h.ctx, hostedCluster, h.client, opts, artifactDir, cleanupPhase)
 	})
 }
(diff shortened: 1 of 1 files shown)


stevekuznetsov:
/hold

This won't run in cleanup phase. I think it's much easier to remove the `t.Run()` and not call `teardownHostedCluster()` twice rather than copying `validateMetrics()` to both places.

muraee:
@stevekuznetsov we don't want to run it in Cleanup phase anyway, cleanup phase means there was an uncaught panic somewhere.

/unhold

stevekuznetsov:
Why not? Seems super hard to reason about what is and is not run in these cases ...

stevekuznetsov:
cc @enxebre - we've already had some chats on the subtests for these parts. It seems to me like this is a clear-cut spot where _not_ using a subtest might be a good idea. Not only does it let us do logic on `t.Failed()` but it then also keeps us from having to have separate and somehow distinct execution paths through this (already fairly complex) set of logic.

From my past experience, keeping the framework execution as part of the outer test and sub-testing specific parts of the actual test logic (e.g. the PKI flow in CreateCluster, etc) really helps readability.

stevekuznetsov:
/lgtm cancel

Definitely seems weird to merge this in.

muraee:
> Why not? 

why would you run another validation test if the test has paniced (aka failed)? its pretty straightforward if you read the code calling `teardown`, we have the happy path:
- CreateCluster -> Validate -> Run the actual test -> Teardwon -> PostTeardown

And a failsafe case if there were a panic during the test execution:
- CreateCluster -> Validate -> Run the actual test -> panic -> registered cleanup func is triggered -> Teardwon

you are suggesting that we need to run the `ValidateMetrics` in the 2nd case, why would we want that?

The discussion of whether to run the happy path Teardown in a sub-test or not seems totally unrelated for this PR.

muraee:
the teardown logic is separated into 2 functions `(h *hypershiftTest) teardown()` and the helper func `teardownHostedCluster()` specifically so we don't pollute the actuall logic of tearing down a cluster via `teardownHostedCluster()` with unrelated sub-tests. So we don't have to use `t.Failed()` inside that func and keep it simple, which is what this PR fixes.

stevekuznetsov:
If you look at the code we run in the `defer`, we *already* (attempt to) run validations when there's been a panic, since we run `postTeardown()` in the deferred func. I'm not arguing for us to do something novel and different - I am simply arguing that we should be consistent. If we don't want to run validations after a panic, we should remove those across the board, rather than in this one place.

I'm not sure we derive value from the `teardown` -> `teardownHostedCluster` indirection, and, specifically, having to deal with "how many times have we torn down?" and "are we running in `t.Cleanup()` or not?" makes for a complex mental model of the flow for our tests. 

I propose we:
 - remove the call to `t.Cleanup()`
 - remove the need to track how many times we've done teardown
 - simplify the teardown code to remove the distinction between different cleanup phases
 - remove the sub-tests, which allows us to simply use `t.Failed()` to check whether or not a validation should occur

muraee:
`postTeardown()` will not run if there was a caught panic, because the test will be marked as failed with first and `postTeardown()` does check `h.Failed()` before running.

`t.Cleanup()` is always called, which is why we need to check if we have already torn down the cluster, because in the happy path, `teardown()` would have been already called and destroyed the cluster then we don't want to run it again, hence the check.

muraee:
Again, this a fairly small PR fixing a simple the issue that `ValidateMetrics` is executed in the wrong func. I don't believe this is the right place to discuss changes on how we structure our tests.

stevekuznetsov:
Removing the `t.Run()` would be just as small of a change - so respectfully I don't want to get down the rabbit hole of what is or is not the right forum. 

> will not run if there was a caught panic, because the test will be marked as failed first with 

I am pointing out that your argument for the refactor you made was that we did not need to bother running code in the post-panic handler, and I pointed out that your statement is inconsistent with the rest of the codebase, specifically because of what you write here - we already do that. If you feel strongly one way, let's make the code here consistent. The cleanup story in the existing end-to-end is pretty complex and I would appreciate us taking steps to making it less so, because I have trouble grok-ing it as it is.

PR #4254, opened 2024-06-21
OCPBUGS-35874: check mgmt cluster for route capability before DeleteIfNeeded for ovn sbdb route

sjenning:
xref https://issues.redhat.com//browse/OCPBUGS-25758

https://github.com/openshift/hypershift/pull/4033 introduced cleanup for the `ovnkube-sbdb Route, however, it did not check that the management cluster has the Route CRD present before attempting the delete leading to a regression on non-OCP mgmt clusters.


sjenning:
cc @rtheis

rtheis (approved):
/lgtm

I built and verified this fix with a ROKS version 4.16.0-rc.6 cluster.  Thanks.

PR #4256, opened 2024-06-21
HOSTEDCP-1447: test/e2e: tar.gz hosted cluster content

stevekuznetsov:
Our end-to-end tests take forever to load in Deck directly due to the number of files we have in the artifact directory. Generally, jobs will archive the results to improve this.

/assign @sjenning

### DIFF
--- a/test/e2e/util/hypershift_framework.go
+++ b/test/e2e/util/hypershift_framework.go
@@ -1,8 +1,11 @@
 package util
 
 import (
+	"archive/tar"
+	"compress/gzip"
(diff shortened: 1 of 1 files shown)


sjenning:
/lgtm
/hold
until we inspect e2e-aws artifacts and verify the hosted cluster artifacts are tar.gz'd as expected

stevekuznetsov:
Well, it works, but loading is still slow ... will have to do more digging into what is eating up so many files in GCS.

PR #4268, opened 2024-06-24
HOSTEDCP-1716: When running the HO locally it should no required a running pod

marcbrown2024:
This PR covers the steps to run the HyperShift Operator locally without needing access to a full production environment.

Fixes #[HOSTEDCP-1716](https://issues.redhat.com//browse/HOSTEDCP-1716)

Checklist


stevekuznetsov:
@marcbrown2024 thanks for the contributions! I would suggest a slightly different track:

 - running a binary locally gives some useful input to the validity of the operator, but does not give signal on the associated requirements for the deployment
 - helpful guides in Markdown go stale faster than we can keep up with them

In `test/integration` an automated process already exists for spinning up a `kind` cluster on your machine and using your local content to run the HyperShift components in the management plane. I'd love to collaborate with you to make sure that the approach for the integration tests works for your use-cases.  I think such an approach will prove to be more holistic in what it helps us verify locally and more durable to boot.

csrwng:
@stevekuznetsov I think we need both, since they're for different scenarios. Running the hypershift operator locally while pointing to a fully deployed management cluster (on AWS or Azure) allows you to test/debug things that you would only see in an environment like that (say nodepool controller AWS specifics, or even the scheduler/scaler for ROSA). The integration environment lets you quickly validate other things, and is definitely useful, so we should also document it.

stevekuznetsov:
@csrwng the integration scripts have no requirements on what they are pointed at - why not use them to hot-replace the `hypershift-operator` binary running in the cluster?

stevekuznetsov:
I guess I'd just say - let's put some value on executable, re-usable functionality here. I think everyone doing development on HyperShift has their own assemblage of shell scripts, hack docs, whatever. It's a lot. Find some semi-reusable thing and eat the cost of it not being perfectly optimal for your specific/special use-case, and reap the benefits to collaboration and onboarding of having a common platform.

hack/dev/hypershft-install.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+set -euo pipefail
+
+REGION="${HYPERSHIFT_REGION:-us-east-1}"
+BUCKET_NAME="${HYPERSHIFT_BUCKET_NAME:-hypershift-ci-oidc}"
+AWSCREDS="${HYPERSHIFT_AWS_CREDS:-$HOME/.aws/credentials}"
+
+# Install hypershift
```
enxebre, line 9: do we want to setup external dns https://hypershift-docs.netlify.app/how-to/aws/external-dns/
hypershift install --external-dns-provider=aws --external-dns-credentials=route53-aws-creds --external-dns-domain-filter=service-provider-domain.com ...?

csrwng: that would definitely be useful for AWS, but not for AKS. Should we have different scripts for the different environments where we may want to install? (we could start by renaming this one with `-aws`) and install external-dns, wdyt?

stevekuznetsov: One entrypoint that everyone uses, which knows how to validate the expected input based on the target you're deploying to? :D

enxebre: fwiw we do use external DNS atm in aks as well. But yes let's have a way to install the opinions per platform.

Makefile
```diff
@@ -286,3 +290,11 @@ ci-install-hypershift-private:
 regenerate-pki:
 	REGENERATE_PKI=1 $(GO) test ./control-plane-pki-operator/...
 	REGENERATE_PKI=1 $(GO) test ./test/e2e/... -run TestRegeneratePKI
+
+.PHONY: hypershift-install-aws
+hypershift-install-aws:
```
enxebre, line 295: hypershift-install-aws-dev?

docs/content/contribute/run-hypershift-operator-locally.md
```diff
@@ -0,0 +1,47 @@
...
+  ```
+
+!!! note 
+
+    `Consider setting HYPERSHIFT_REGION and HYPERSHIFT_BUCKET_NAME in your shell init script (e.g., $HOME/.bashrc).
+
+4. Install HyperShift in development mode which causes the operator deployment to be deployment scaled to zero so that it doesn't conflict with your local operator process (see [Prerequisites](../getting-started.md#prerequisites)):
+  
+  ```shell linenums="1"
+    HYPERSHIFT_REGION=us-east-2 HYPERSHIFT_BUCKET_NAME=my-bucket make hypershift-install-aws
```
enxebre, line 39: why do I need to export the variables and then pass them again?

enxebre: Also can we mention you can set them but there's defaults and `make whatever` should just work without exporting or setting anything

Makefile
marcbrown2024: Are you saying I should rename it to hypershift-install-aws-dev?

docs/content/contribute/run-hypershift-operator-locally.md
marcbrown2024: You are exporting the vars because in the script the default is us-east-1 and bucket is hypershift-ci-oidc. So you'll have to export the correct var if default is not what is needed

hack/dev/aws/run-operator-locally-aws-dev.sh
```diff
@@ -0,0 +1,25 @@
...
+  run \
+  --metrics-addr=0 \
+  --enable-ocp-cluster-monitoring=false \
+  --enable-ci-debug-output=true \
+  --private-platform=AWS \
+  --enable-ci-debug-output=true \
+  --oidc-storage-provider-s3-credentials "${AWSCREDS}" \
+  --oidc-storage-provider-s3-region "${REGION}" \
+  --oidc-storage-provider-s3-bucket-name "${BUCKET_NAME}" \
+  --control-plane-operator-image=quay.io/hypershift/hypershift:latest
```
enxebre, line 25: probably worth making --control-plane-operator-image a variable

PR #3969, opened 2024-05-01
OCPBUGS-32467: nodepool_controller: add a reconciler for cleanup

stevekuznetsov:
When we change the config for a NodePool or the way in which we hash the values, we leak token and user data secrets. However, since these secrets are annotated with the NodePool they were created for, it's simple to check that the Secret still matches what we expect and clean it up otherwise.

/assign @sjenning

PR #4269, opened 2024-06-25
OCPBUGS-35979: Keep old user data for aws < 4.16

enxebre:
We currently need to keep old userdata secrets in AWS < 4.16 to prevent machineDeployments rollout from failing to delete old Machines.
https://github.com/openshift/hypershift/blob/3efa23b932a59681346a7a432d349cfb6e44b13d/hypershift-operator/controllers/nodepool/nodepool_controller.go#L775-L777
https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3805

We regress that behaviour here https://github.com/openshift/hypershift/pull/3969

This PR fixes that by statically checking the hc release version.

Fixes #

**Checklist**

### DIFF
--- a/hypershift-operator/controllers/nodepool/nodepool_controller.go
+++ b/hypershift-operator/controllers/nodepool/nodepool_controller.go
@@ -3056,6 +3056,15 @@ func (r *secretJanitor) Reconcile(ctx context.Context, req reconcile.Request) (r
 		return ctrl.Result{}, err
 	}
 
+	shouldKeepOldUserData, err := r.shouldKeepOldUserData(ctx, hcluster)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	if shouldKeepOldUserData {
+		log.V(3).Info("Skipping secretJanitor reconciliation and keeping old user data secret")
+		return ctrl.Result{}, nil
+	}
+
 	controlPlaneNamespace := manifests.HostedControlPlaneNamespace(hcluster.Namespace, hcluster.Name)
 	expectedCoreConfigResources := expectedCoreConfigResourcesForHostedCluster(hcluster)
 	releaseImage, err := r.getReleaseImage(ctx, hcluster, nodePool.Status.Version, nodePool.Spec.Release.Image)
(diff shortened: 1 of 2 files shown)


enxebre:
/hold
a bit

hypershift-operator/controllers/nodepool/nodepool_controller.go
```diff
@@ -3128,6 +3137,44 @@ func (r *secretJanitor) Reconcile(ctx context.Context, req reconcile.Request) (r
...
+// shouldKeepOldUserData determines if the old user data should be kept.
+// For AWS < 4.16, we keep the old userdata Secret so old Machines during rolled out can be deleted.
+// Otherwise, deletion fails because of https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3805.
+// TODO (alberto): Drop this check when support for old versions without the fix is not needed anymore.
+func (r *NodePoolReconciler) shouldKeepOldUserData(ctx context.Context, hc *hyperv1.HostedCluster) (bool, error) {
+	if hc.Spec.Platform.Type != hyperv1.AWSPlatform {
+		return false, nil
+	}
+
+	// If there's a curren version in status, be conservative and assume that one is the one running CAPA.
```
andreadecorte, line 3149: NP
```suggestion
	// If there's a current version in status, be conservative and assume that one is the one running CAPA.
```

hypershift-operator/controllers/nodepool/nodepool_controller.go
```diff
@@ -3128,6 +3137,44 @@ func (r *secretJanitor) Reconcile(ctx context.Context, req reconcile.Request) (r
...
+	releaseInfo, err := r.ReleaseProvider.Lookup(ctx, releaseImage, pullSecretBytes)
+	if err != nil {
+		return true, fmt.Errorf("failed to lookup release image: %w", err)
+	}
+	hostedClusterVersion, err := semver.Parse(releaseInfo.Version())
+	if err != nil {
+		return true, err
+	}
+
+	if hostedClusterVersion.LT(semver.MustParse("4.16.0")) {
```
andreadecorte, line 3171: this will also exclude all the 4.16 nightlies, but I guess that it is expected/wanted

andreadecorte:
2 minor comments, but lgtm otherwise

hypershift-operator/controllers/nodepool/nodepool_controller.go
enxebre: this is saying less than 4.16

andreadecorte: yep. Just saying that this includes any 4.16 nightly, if it makes any difference. But we probably don't care. 
https://go.dev/play/p/YNZ4-qiy7Zt

enxebre: I see, yes that's desired as far as I can tell

PR #4272, opened 2024-06-25
HOSTEDCP-1778: Enable MultiArch flag by default

bryan-cox:
Enable MultiArch flag by default

Fixes [HOSTEDCP-1778](https://issues.redhat.com//browse/HOSTEDCP-1778)

**Checklist**

### DIFF
--- a/cmd/cluster/aws/create.go
+++ b/cmd/cluster/aws/create.go
@@ -44,6 +44,7 @@ type RawCreateOptions struct {
 	EnableProxy             bool
 	SingleNATGateway        bool
 	MultiArch               bool
+	SkipMultiArchImageCheck bool
 }
 
 // validatedCreateOptions is a private wrapper that enforces a call of Validate() before Complete() can be invoked.
@@ -359,6 +360,7 @@ func DefaultOptions() *RawCreateOptions {
 		RootVolumeType: "gp3",
 		RootVolumeSize: 120,
 		EndpointAccess: string(hyperv1.Public),
+		MultiArch:      true,
 	}
 }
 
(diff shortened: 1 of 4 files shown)


bryan-cox:
e2e-aws failed differently than the first time. Seems like a possible flake.

/test e2e-aws

bryan-cox:
@stevekuznetsov - do you know why the unit test is failing? I didn't make any changes to that one 🤔

stevekuznetsov:
Wild. My thought would be that when `multi-arch` is true, we do something with the pull secret. Right now the pull secret is entirely fake data, not even JSON:

https://github.com/openshift/hypershift/blob/54f1141b252fb664b76c061d0c6602e7dc36501f/cmd/cluster/aws/create_test.go#L198-L200

Reaching out to the registry during unit tests is a bad idea since the test will then take much longer and require the internet, etc. Is there a way to not reach out to the registry when multi-arch is true? If we need some extra data in order to not do that reach-out, perhaps we can add some flags for that and pass them in the test.

bryan-cox:
> Wild. My thought would be that when `multi-arch` is true, we do something with the pull secret. Right now the pull secret is entirely fake data, not even JSON:
> 
> https://github.com/openshift/hypershift/blob/54f1141b252fb664b76c061d0c6602e7dc36501f/cmd/cluster/aws/create_test.go#L198-L200
> 
> Reaching out to the registry during unit tests is a bad idea since the test will then take much longer and require the internet, etc. Is there a way to not reach out to the registry when multi-arch is true? If we need some extra data in order to not do that reach-out, perhaps we can add some flags for that and pass them in the test.

I added a specific flag for the unit test. I'm not sure how to get around the unit test other than using it. In order to see if the release image is manifest listed, we need to pull it and look at it to see.

PR #4218, opened 2024-06-14
OCPBUGS-34865: Add HTTP(s) konnectivity proxy for openshift-apiserver

csrwng:
**What this PR does / why we need it**:
Introduces a new command under cpo: konnectivity-https-proxy that
performs a similar job to konnectivity-socks-proxy except that it
exposes the http protocol. It allows passing user-configured proxy
configuration (http_proxy, https_proxy, no_proxy) as flag values and
uses them when tunneling traffic through konnectivity.

The new konnectivity HTTPS proxy allows setting the proxy to use for
data plane requests. This commit switches the proxy used by the
OpenShift APIServer to the new proxy and configures it with the proxy
configuration if specified in HostedCluster.spec.configuration.proxy

**Which issue(s) this PR fixes** *(optional, use `fixes #<issue_number>(, fixes #<issue_number>, ...)` format, where issue_number might be a GitHub issue, or a Jira story*:
Fixes #[OCPBUGS-34865](https://issues.redhat.com/browse/OCPBUGS-34865)

**Checklist**
- [x] Subject and description added to both, commit and PR.
- [x] Relevant issues have been referenced.
- [ ] This change includes docs. 
- [ ] This change includes unit tests.

PR #4273, opened 2024-06-26
OCPBUGS-37753: Let the CPO oidc check resolve through data plane

enxebre:
This enabled the oauth server traffic to flow through the dataplane to enable reaching private endpoints e.g ldap https://issues.redhat.com/browse/HOSTEDCP-421

This enabled fallback to the management cluster network so for public endpoints we are not blocking on having data plane, e.g. github https://issues.redhat.com/browse/OCPBUGS-8073

This PR enables the CPO oidc checks to flow through the data plane and fallback to the management side to satisfy both cases above https://issues.redhat.com/browse/RFE-5638

Fixes #

**Checklist**

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/oauth/idp_convert.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/oauth/idp_convert.go
@@ -15,17 +15,23 @@ import (
 
 	configv1 "github.com/openshift/api/config/v1"
 	osinv1 "github.com/openshift/api/osin/v1"
+	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
+	kas "github.com/openshift/hypershift/control-plane-operator/controllers/hostedcontrolplane/kas"
+	manifests "github.com/openshift/hypershift/control-plane-operator/controllers/hostedcontrolplane/manifests"
+	"github.com/openshift/hypershift/support/konnectivityproxy"
+	supportproxy "github.com/openshift/hypershift/support/proxy"
+	"github.com/openshift/hypershift/support/util"
+	"golang.org/x/net/http/httpproxy"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/cache"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/net"
+	clientcmd "k8s.io/client-go/tools/clientcmd"
 	"k8s.io/utils/pointer"
 	ctrl "sigs.k8s.io/controller-runtime"
 	crclient "sigs.k8s.io/controller-runtime/pkg/client"
-
-	"github.com/openshift/hypershift/support/util"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -2578,6 +2578,18 @@ func reconcileControlPlaneOperatorDeployment(
...
+						Command: []string{"/usr/bin/control-plane-operator", "konnectivity-socks5-proxy", "--resolve-from-guest-cluster-dns=true", "--resolve-from-management-cluster-dns=true"},
+						Args:    []string{"run"},
+						Resources: corev1.ResourceRequirements{
+							Requests: corev1.ResourceList{
+								corev1.ResourceCPU:    resource.MustParse("10m"),
+								corev1.ResourceMemory: resource.MustParse("10Mi"),
+							},
+						},
+					},
 				},
```
enxebre, line 2593: Accidental convenient side effects:
I think this should not need certs given it goes through https://github.com/openshift/hypershift/blob/main/konnectivity-socks5-proxy/main.go#L125. Otherwise we would have a chicken-egg issue to get the certs.

The kube client for dataplane dns resolution would pick the incluster config of the CPO and so trying to resolve actually via management svc network, fail and fallback https://github.com/openshift/hypershift/blob/main/konnectivity-socks5-proxy/main.go#L73

enxebre:
/hold
a bit

enxebre:
/hold
wants https://github.com/openshift/hypershift/pull/4218

control-plane-operator/controllers/hostedcontrolplane/oauth/idp_convert.go
```diff
@@ -596,12 +682,16 @@ func isValidURL(rawurl string, optional bool) bool {
 	return u.Scheme == "https" && len(u.Host) > 0 && len(u.Fragment) == 0
 }
 
-func transportForCARef(ctx context.Context, kclient crclient.Client, namespace, name, key string) (http.RoundTripper, error) {
+func transportForCARef(ctx context.Context, kclient crclient.Client, namespace, name, key string, dialer konnectivityproxy.ProxyDialer) (http.RoundTripper, error) {
```
csrwng, line 685: When constructing this transport, if we are using the data plane dialer, we need to take into consideration the user's proxy configuration in hcp.configuration.proxy
to configure the proxy on the transport, pls see:
https://github.com/openshift/hypershift/blob/4a34cb844b3703811787afa7f0298cea38f52719/konnectivity-https-proxy/cmd.go#L73-L78
https://github.com/openshift/hypershift/blob/4a34cb844b3703811787afa7f0298cea38f52719/konnectivity-https-proxy/cmd.go#L113-L115

If a ca bundle is specified for the proxy, that CA bundle also needs to be added to the trusted roots of the transport.

enxebre: addressed!

control-plane-operator/controllers/hostedcontrolplane/oauth/idp_convert.go
```diff
@@ -596,37 +668,93 @@ func isValidURL(rawurl string, optional bool) bool {
...
+		}
+		if len(hcpList.Items) != 1 {
+			return nil, fmt.Errorf("expected one hosted control plane, got %d", len(hcpList.Items))
+		}
+		hcp := hcpList.Items[0]
+		if proxy := hcp.Spec.Configuration.Proxy; proxy != nil {
+			userProxyConfig = &httpproxy.Config{
+				HTTPProxy:  proxy.HTTPProxy,
+				HTTPSProxy: proxy.HTTPSProxy,
+				NoProxy:    proxy.NoProxy,
```
csrwng, line 703: We should call https://github.com/openshift/hypershift/blob/ffe665196be2e2654d47fa3c798330625cdf493d/support/proxy/no_proxy.go#L15 to get the value for this.

enxebre: addressed!

control-plane-operator/controllers/hostedcontrolplane/oauth/idp_convert_test.go
```diff
@@ -185,3 +192,200 @@ func TestOpenIDProviderConversion(t *testing.T) {
 		})
 	}
 }
+
+func TestTransportForCARef(t *testing.T) {
```
csrwng, line 196: nice! ❤️

PR #4277, opened 2024-06-26
OCPBUGS-36230: Use a reconciled proxy config when generating nodepool userdata

csrwng:
The proxy configuration from the HostedCluster resource is not currently used when generating user data (an empty proxy config is used). This commit invokes the reconcile on the proxy config so that it's filled in with the values from the HostedCluster.

Fixes #[OCPBUGS-36230](https://issues.redhat.com/browse/OCPBUGS-36230)

**Checklist**


csrwng:
The change was introduced here: https://github.com/openshift/hypershift/pull/3969 
It didn't look intentional.

PR #4285, opened 2024-06-27
WIP: HOSTEDCP-1737: Use a fixed static IP for dataplane remote-kas HAProxy in shared-ingress mode

muraee:

The shared ingress controller now assigns a fixed static IP for each Hosted Cluster. This IP is used as the destination IP for requests sent from the data-plane through the HAproxy using the proxy protocol. This is then caught by the shared ingress router and used to discriminate which HostedCluster's KAS the request should be routed to.

Fixes #

**Checklist**

### DIFF
--- a/support/testutil/testutil.go
+++ b/support/testutil/testutil.go
@@ -74,6 +74,12 @@ type options struct {
 
 type option func(*options)
 
+func WithExtension(extension string) option {
+	return func(opts *options) {
+		opts.Extension = extension
+	}
+}
+
 // golden determines the golden file to use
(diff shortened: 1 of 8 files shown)


support/testutil/testutil.go
```diff
@@ -74,6 +74,12 @@ type options struct {
 
 type option func(*options)
 
+func WithExtension(extension string) option {
```
stevekuznetsov, line 77: Nice!

enxebre:
/close
in favour of https://github.com/openshift/hypershift/pull/4330

PR #4287, opened 2024-06-27
NO-JIRA: cmd: add support to configure machine networts via CLI

EmilienM:

Empty by default for backward compatibility, we can now configure the
Machine networks via the CLI.

This will become handy for the OpenStack platform, where when we let the
CAPI provider (namely CAPO) to handle the networks, we still need to
provide a machine network CIDR and using the CLI seems a consistent way
to do so.

### DIFF
--- a/cmd/cluster/core/create.go
+++ b/cmd/cluster/core/create.go
@@ -50,6 +50,7 @@ func DefaultOptions() *RawCreateOptions {
 		ControlPlaneAvailabilityPolicy: string(hyperv1.SingleReplica),
 		ServiceCIDR:                    []string{globalconfig.DefaultIPv4ServiceCIDR},
 		ClusterCIDR:                    []string{globalconfig.DefaultIPv4ClusterCIDR},
+		MachineCIDR:                    []string{},
 		Log:                            log.Log,
 		Arch:                           "amd64",
 		OLMCatalogPlacement:            hyperv1.ManagementOLMCatalogPlacement,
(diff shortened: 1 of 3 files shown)


cmd/cluster/cluster.go
```diff
@@ -27,6 +27,7 @@ func NewCreateCommands() *cobra.Command {
 
 	cmd.MarkFlagsMutuallyExclusive("service-cidr", "default-dual")
 	cmd.MarkFlagsMutuallyExclusive("cluster-cidr", "default-dual")
+	cmd.MarkFlagsMutuallyExclusive("machine-cidr", "default-dual")
```
stevekuznetsov, line 30: Please add these to the validation func on the `core.DefaultOptions()` - we have users of cluster creation outside of CLI and need to validate the actual options struct, not one implementation of how the data is passed (flags).

stevekuznetsov (commented):
Please surface this behavior in one of the `TestCreateCluster` unit tests, and if we don't already have it, unit tests for the options validation.

cmd/cluster/cluster.go
EmilienM: ack, I'll look at this asap.

EmilienM: In here? https://github.com/shiftstack/hypershift/blob/b68f5a62a23ea292c7363ad5413ce1f3d43034b0/cmd/cluster/core/create.go#L541-L542
I think I need a little bit of guidance here on what you expect the test to be, I'll be happy to write it.
Thanks

bryan-cox:
@EmilienM is this still needed?

mandre:
> > @EmilienM is this still needed?
> 
> yes, @mandre took over the PR.

Yeah, sorry, I haven't been explicit about the changes I recently made to the PR or the status. I have rebased the old PR to fix a conflict, and did 2 other things:
- remove the mutually exclusive property with "default-dual" as we don't do anything special for the MachineCIDR when default-dual is set.
- remove a validation that checked that `--machine-cidr`, `--cluster-cidr`, or `--service-cidr` isn't specified more than twice as I believed it didn't belong to this PR.

I believe this PR is now ready for review @bryan-cox when you get a chance.

cmd/cluster/core/create.go
```diff
@@ -406,6 +410,12 @@ func prototypeResources(opts *CreateOptions) (*resources, error) {
 	}
 	prototype.Cluster.Spec.Networking.ServiceNetwork = serviceNetworkEntries
 
+	var machineNetworkEntries []hyperv1.MachineNetworkEntry
+	for _, cidr := range opts.MachineCIDR {
+		machineNetworkEntries = append(machineNetworkEntries, hyperv1.MachineNetworkEntry{CIDR: *ipnet.MustParseCIDR(cidr)})
```
stevekuznetsov, line 415: The idea behind the validate/complete pattern is that the types coming out the other end to be used in product code don't need to do things like `ipnet.MustParseCIDR` - parse the value in `Complete()` and store the parsed CIDR as `net.IPNet` directly, so the code here can be certain it has the correct inputs. This has the nice side-effect of not writing code that can panic. 

A nice read on the thought behind it: https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/

mandre: Thanks for the link. We're now using the same pattern as ClusterCIDR and ServiceCIDR above.

PR #3849, opened 2024-04-06
OCPBUGS-31840: Make guest cluster components use the correct KAS port

fxierh:
Failed to provision a private HC on AWS, following the official docs https://hypershift-docs.netlify.app/how-to/aws/deploy-aws-private-clusters/. VMs are provisioned but unable to join the HC as Nodes. 

This is due to the fact that the kubelet on each HC VM tried to use api.fxie-hcp-2.hypershift.local:443 (instead of 6443 for KAS exposed via LB) to communicate with the KAS:
```Bash
# VM Journal
...
Apr 02 19:37:54 ip-10-0-142-1 kubenswrapper[2032]: E0402 19:37:54.769191    2032 kubelet_node_status.go:99] "Unable to register node with API server" err="Post \"https://api.fxie-hcp-2.hypershift.local:443/api/v1/nodes\": dial tcp 10.0.138.240:443: i/o timeout" node="ip-10-0-142-1.us-east-2.compute.internal"
...
```

Besides, the CNO also passes the wrong KAS port to Network components on the HC. 

Same for HA proxy configuration on the VMs:
```Bash
frontend local_apiserver
  bind 172.20.0.1:6443
  log global
  mode tcp
  option tcplog
  default_backend remote_apiserver

backend remote_apiserver
  mode tcp
  log global
  option httpchk GET /version
  option log-health-checks
  default-server inter 10s fall 3 rise 3
  server controlplane api.fxie-hcp-1.hypershift.local:443
```

PR #1007, opened 2022-02-12
AWS: Use availability prober in capi provider

alvaroaleman:
The AWS capi provider currently always errors for some time during
startup when a new cluster is created, because there is no token,
because the KAS is not up yet so the token minter sidecar can't mint
one. This change makes us use the availability prober init container to
avoid going into backoff due to this.

Using the availability prober required two new things to be available to
the provider:
* The HCP in case it uses a non-default apiserver port
* The Availability prober image

For the first, the existing provider interface was extended, as we only
have construct the hcp right before calling CAPIProviderDeploymentSpec.

For the static availability prober image we avoid extending the
interface for every provider for something that is only needed by AWS by
passing it into the provider at construction time.

**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, use `fixes #<issue_number>(, fixes #<issue_number>, ...)` format, where issue_number might be a GitHub issue, or a Jira story*:
Fixes #

**Checklist**
- [ ] Subject and description added to both, commit and PR.
- [ ] Relevant issues have been referenced.
- [ ] This change includes docs. 
- [ ] This change includes unit tests.

PR #4296, opened 2024-06-29
HOSTEDCP-1790: Expose the KAS services through port 443

fxierh:
Follows up https://github.com/openshift/hypershift/pull/3849. 

Expose the KAS services through port 443 instead of 6443 by default, regardless of whether Route KAS or LB KAS is used. 

Also, refer to the constants defined in constants.go instead of hardcoding them whenever possible.

### DIFF
--- a/support/constants/constants.go
+++ b/support/constants/constants.go
@@ -32,6 +32,7 @@ const (
 	KASSVCLBAzurePort           = 7443
 	KASSVCIBMCloudPort          = 2040
 	KASSVCPort                  = 443
+	KASSVCPortOld               = 6443
 	KASPodDefaultPort           = 6443
 	RouterSVCPort               = 443
 	DefaultServiceNodePortRange = "30000-32767"
--- a/support/constants/constants.go
+++ b/support/constants/constants.go
@@ -1,4 +1,4 @@
-package config
+package constants
 
 const (
 	// NeedManagementKASAccessLabel is used by network policies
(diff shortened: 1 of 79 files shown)


fxierh:
The failure of TestUpgradeControlPlane is expected due to the following test configuration:
- The hypershift-operator build includes changes from this PR
- The control-plane-operator build, sourced from the OCP release image, does not include changes from this PR

support/constants/constants.go
```diff
@@ -1,4 +1,4 @@
-package config
```
enxebre, line 1: any particular reason to move this file? To keep scope concise can we please keep it and only reconsider moving in a different PR?

fxierh: I encountered a circular dependency issue between support/util and support/config when attempting to reference the constants in the APIPortForLocalZone function. To resolve this, I moved the file.

Perhaps I could address the file relocation in a separate PR, which this current PR can then depend on ?

fxierh:
In order for TestUpgradeControlPlane (new HO + old CP + old NP -> new HO + new CP + old NP) to work, we need to select the KAS service port based on the HC version here::
https://github.com/openshift/hypershift/blob/4ac51c41d4c1b4af40f82374e2064ff8868de6dc/hypershift-operator/controllers/hostedcluster/internal/platform/aws/aws.go#L233

This is so the availability-prober initContainer within capi-provider deployment uses the correct KAS service port regardless of whether the hosted cluster is new or old.

fxierh:
Another option is to remove the availability prober from the capi-provider deployment on AWS.

Note:
- There is currently no availability prober for the capi-deployment on non-AWS platforms.
- The availability prober was added to the capi-deployment on AWS in [this PR](https://github.com/openshift/hypershift/pull/1007).

PR #4303, opened 2024-07-01
OCPBUGS-34816: Block data plane HC configuration requests

jparrill:
Added ValidatingAdmissionPolicies to the KAS's HostedCluster to block all configuration requests managed in the management cluster via the HostedCluster object.

Enabling the ValidatingAdmissionPolicy feature gate in the KAS allows us to block certain requests from the Data Plane side without using a webhook. To achieve this, we created the necessary resources (Certificate, RBAC, etc.) that allow HCCO to be identified with a specific user. This designated user will be the only one authorized to execute the required requests. All other users, including those using the admin-kubeconfig, will be blocked by the VAPs validation

Pull Request Content:

- Enabled ValidatingAdmissionPolicies (VAP) FeatureGate on the KAS
- Added a new user (via certificates) to HCCO
- Implemented a new reconciler on HCCO for VAPs managed by Hypershift
- Updated Kubernetes dependency to v0.30.2
- Added E2E test for VAPs configured by HCCO

Fixes #[OCPBUGS-34816](https://issues.redhat.com/browse/OCPBUGS-34816)

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/configoperator/reconcile.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/configoperator/reconcile.go
@@ -232,6 +232,7 @@ func ReconcileRole(role *rbacv1.Role, ownerRef config.OwnerRef, platform hyperv1
 			},
 		}...)
 	}
+	// TODO (jparrill): Add RBAC specific needs for Agent platform
 	return nil
 }
 
@@ -443,7 +444,7 @@ func buildHCCContainerMain(image, hcpName, openShiftVersion, kubeVersion string,
 
 func buildHCCVolumeKubeconfig(v *corev1.Volume) {
 	v.Secret = &corev1.SecretVolumeSource{
-		SecretName:  manifests.KASServiceKubeconfigSecret("").Name,
+		SecretName:  manifests.HCCOKubeconfigSecret("").Name,
 		DefaultMode: pointer.Int32(0640),
 	}
 }
--- a/control-plane-operator/hostedclusterconfigoperator/controllers/resources/kas/admissionpolicies.go
+++ b/control-plane-operator/hostedclusterconfigoperator/controllers/resources/kas/admissionpolicies.go
@@ -1,0 +1,170 @@
+package kas
+
+import (
+	"context"
+	"fmt"
+
+	configv1 "github.com/openshift/api/config/v1"
+	operatorv1alpha1 "github.com/openshift/api/operator/v1alpha1"
+	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
+	"github.com/openshift/hypershift/control-plane-operator/hostedclusterconfigoperator/controllers/resources/manifests"
+	"github.com/openshift/hypershift/support/config"
+	"github.com/openshift/hypershift/support/upsert"
+	k8sadmissionv1beta1 "k8s.io/api/admissionregistration/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type AdmissionPolicy struct {
+	Name             string
+	MatchConstraints *k8sadmissionv1beta1.MatchResources
+	Validations      []k8sadmissionv1beta1.Validation
+}
+
+const (
+	AdmissionPolicyNameConfig = "config"
+	AdmissionPolicyNameMirror = "mirror"
+	AdmissionPolicyNameICSP   = "icsp"
+)
+
+var (
+	allAdmissionPoliciesOperations = []k8sadmissionv1beta1.OperationType{"*"}
+	defaultMatchResourcesScope     = k8sadmissionv1beta1.ScopeType("*")
+	defaultMatchPolicyType         = k8sadmissionv1beta1.Equivalent
+	HCCOUserValidation             = k8sadmissionv1beta1.Validation{
+		Expression: fmt.Sprintf("request.userInfo.username == 'system:%s' || (has(object.spec) && has(oldObject.spec) && object.spec == oldObject.spec)", config.HCCOUser),
+		Message:    "This resource cannot be created, updated, or deleted. Please ask your administrator to modify the resource in the HostedCluster object.",
+		Reason:     ptr.To(metav1.StatusReasonInvalid),
+	}
+)
+
+// ReconcileKASValidatingAdmissionPolicies will create ValidatingAdmissionPolicies which block certain resources
+// from being updated/deleted from the DataPlane side.
+func ReconcileKASValidatingAdmissionPolicies(ctx context.Context, hcp *hyperv1.HostedControlPlane, client client.Client, createOrUpdate upsert.CreateOrUpdateFN) error {
+	log := ctrl.LoggerFrom(ctx)
+	log.Info("reconciling validating admission policies")
+
+	if err := reconcileConfigValidatingAdmissionPolicy(ctx, hcp, client, createOrUpdate); err != nil {
+		return fmt.Errorf("failed to reconcile Config Validating Admission Policy: %v", err)
+	}
+
+	if err := reconcileMirrorValidatingAdmissionPolicy(ctx, hcp, client, createOrUpdate); err != nil {
+		return fmt.Errorf("failed to reconcile Mirror Validating Admission Policies: %v", err)
+	}
+
+	return nil
+}
+
+func reconcileConfigValidatingAdmissionPolicy(ctx context.Context, hcp *hyperv1.HostedControlPlane, client client.Client, createOrUpdate upsert.CreateOrUpdateFN) error {
+	// Config AdmissionPolicy
+	configAdmissionPolicy := AdmissionPolicy{Name: AdmissionPolicyNameConfig}
+	configAPIVersion := []string{configv1.GroupVersion.Version}
+	configAPIGroup := []string{configv1.GroupVersion.Group}
+	configResources := []string{
+		"apiservers",
+		"authentications",
+		"featuregates",
+		"images",
+		"imagecontentpolicies",
+		"infrastructures",
+		"ingresses",
+		"proxies",
+		"schedulers",
+		"networks",
+		"oauths",
+	}
+
+	if hcp.Spec.OLMCatalogPlacement == hyperv1.ManagementOLMCatalogPlacement {
+		configResources = append(configResources, "operatorhubs")
+	}
+	configAdmissionPolicy.Validations = []k8sadmissionv1beta1.Validation{HCCOUserValidation}
+	configAdmissionPolicy.MatchConstraints = constructPolicyMatchConstraints(configResources, configAPIVersion, configAPIGroup, []k8sadmissionv1beta1.OperationType{"UPDATE", "DELETE"})
+	if err := configAdmissionPolicy.reconcileAdmissionPolicy(ctx, client, createOrUpdate); err != nil {
+		return fmt.Errorf("error reconciling Config Validating Admission Policy: %v", err)
+	}
+
+	return nil
+}
+
+func reconcileMirrorValidatingAdmissionPolicy(ctx context.Context, hcp *hyperv1.HostedControlPlane, client client.Client, createOrUpdate upsert.CreateOrUpdateFN) error {
+	// Mirroring AdmissionPolicies
+	mirrorAdmissionPolicy := AdmissionPolicy{Name: AdmissionPolicyNameMirror}
+	mirrorAPIVersion := []string{configv1.GroupVersion.Version}
+	mirrorAPIGroup := []string{configv1.GroupVersion.Group}
+	mirrorResources := []string{
+		"imagedigestmirrorsets",
+		"imagetagmirrorsets",
+	}
+
+	if hcp.Spec.OLMCatalogPlacement == hyperv1.ManagementOLMCatalogPlacement {
+		mirrorResources = append(mirrorResources, "operatorhubs")
+	}
+	mirrorAdmissionPolicy.Validations = []k8sadmissionv1beta1.Validation{HCCOUserValidation}
+	mirrorAdmissionPolicy.MatchConstraints = constructPolicyMatchConstraints(mirrorResources, mirrorAPIVersion, mirrorAPIGroup, allAdmissionPoliciesOperations)
+	if err := mirrorAdmissionPolicy.reconcileAdmissionPolicy(ctx, client, createOrUpdate); err != nil {
+		return fmt.Errorf("error reconciling Mirror Validating Admission Policy: %v", err)
+	}
+
+	// ICSP lives in other API, this is why we need to create another vap and vap-binding
+	icspAdmissionPolicy := AdmissionPolicy{Name: AdmissionPolicyNameICSP}
+	icspAPIVersion := []string{operatorv1alpha1.GroupVersion.Version}
+	icspAPIGroup := []string{operatorv1alpha1.GroupVersion.Group}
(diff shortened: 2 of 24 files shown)


sjenning:
fmt changes to `cmd/infra/aws/delegating_client.go` are causing `verify` to fail

control-plane-operator/hostedclusterconfigoperator/controllers/resources/kas/admissionpolicies.go
```diff
@@ -0,0 +1,128 @@
...
+
+const (
+	AdmissionPolicyNameConfig = "config"
+	AdmissionPolicyNameICSP   = "icsp"
+)
+
+var (
+	defaultAdmissionPoliciesOperations = []k8sadmissionv1beta1.OperationType{"UPDATE", "DELETE"}
+	HCCOUserValidation                 = k8sadmissionv1beta1.Validation{
+		Expression: fmt.Sprintf("request.userInfo.username.contains('system:%s') || (has(object.spec) && object.spec == oldObject.spec)", config.HCCOUser),
```
muraee, line 38: I don't think `has(object.spec)` is needed here. `spec` can never be nil.

jparrill: In the other tries, that. was failing because spec was nil, I think that's happened during 'CREATE'. Let me remove it.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/kas/admissionpolicies.go
```diff
@@ -0,0 +1,128 @@
...
+	AdmissionPolicies map[string]AdmissionPolicy
+}
+
+type AdmissionPolicy struct {
+	Name        string
+	APIGroup    []string
+	ApiVersion  []string
+	Resources   []string
+	Validations []k8sadmissionv1beta1.Validation
+}
```
muraee, line 28: why do we need those structs and all the helper methods? its kind hard to follow.
it would be good if we keep the same style we use for other resources reconciliation.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/kas/featuregates.go
```diff
@@ -0,0 +1,28 @@
...
+	if fg != nil {
+		return hyperconfig.FeatureGates(&fg.FeatureGateSelection)
+	} else {
+		return hyperconfig.FeatureGates(&configv1.FeatureGateSelection{
+			FeatureSet: configv1.Default,
+		})
+	}
+}
+
+func CheckDeployValidatingAdmissionPolicy(fg *configv1.FeatureGateSpec) bool {
```
muraee, line 18: where is this used? and why do we need it?

control-plane-operator/hostedclusterconfigoperator/controllers/resources/manifests/featuregates.go
```diff
@@ -0,0 +1,18 @@
+package manifests
+
+import (
+	configv1 "github.com/openshift/api/config/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func FeatureGates() *configv1.FeatureGate {
```
muraee, line 8: where is this used?

control-plane-operator/hostedclusterconfigoperator/controllers/resources/kas/admissionpolicies.go
jparrill: I've changed how it was structured to be more understandable. Please, let me know if there is any other concern.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/kas/admissionpolicies.go
```diff
@@ -0,0 +1,118 @@
...
+func (ap *AdmissionPolicy) reconcileAdmissionPolicy(ctx context.Context, client client.Client, createOrUpdate upsert.CreateOrUpdateFN) error {
+	vap := manifests.ValidatingAdmissionPolicy(ap.Name)
+	vap.Spec = k8sadmissionv1beta1.ValidatingAdmissionPolicySpec{
+		MatchConstraints: ap.MatchConstraints,
+		Validations:      ap.Validations,
+	}
+
+	if _, err := createOrUpdate(ctx, client, vap, func() error { return nil }); err != nil {
+		return fmt.Errorf("failed to create/update Validating Admission Policy with name %s: %v", ap.Name, err)
+	}
```
muraee, line 111: changes to the object should be done inside the mutate functions (last argument of `createOrUpdate`). Otherwise, out of band changes will not be kept and not reconciled over. 

```suggestion
	if _, err := createOrUpdate(ctx, client, vap, func() error { 
	  // update spec here
	  vap.Spec = ....
	  return nil 
	}); err != nil {
		return fmt.Errorf("failed to create/update Validating Admission Policy with name %s: %v", ap.Name, err)
	}
```

control-plane-operator/hostedclusterconfigoperator/controllers/resources/kas/admissionpolicies.go
```diff
@@ -0,0 +1,118 @@
...
+	policyBinding := manifests.ValidatingAdmissionPolicyBinding(fmt.Sprintf("%s-binding", ap.Name), ap.Name)
+	if _, err := createOrUpdate(ctx, client, policyBinding, func() error { return nil }); err != nil {
```
muraee, line 113: same comment as above, thje policyBinding spec should not be set in `manifests.ValidatingAdmissionPolicyBinding()` but directly here inside `createOrUpdate()`

control-plane-operator/hostedclusterconfigoperator/controllers/resources/manifests/admissionpolicies.go
```diff
@@ -0,0 +1,43 @@
...
+	return &k8sadmissionv1beta1.ValidatingAdmissionPolicyBinding{
+		ObjectMeta: metav1.ObjectMeta{Name: bindingName},
+		Spec: k8sadmissionv1beta1.ValidatingAdmissionPolicyBindingSpec{
+			PolicyName:        policyName,
+			ValidationActions: []k8sadmissionv1beta1.ValidationAction{k8sadmissionv1beta1.Deny},
+		},
+	}
+}
+
+func VAPolicyMatchConstraints(resources, apiVersion, apiGroup []string, operations []k8sadmissionv1beta1.OperationType) *k8sadmissionv1beta1.MatchResources {
```
muraee, line 28: I would make this a helper function in `admissionpolicies.go` file instead, e.g. `ConstructPolicyMatchConstraints()`. Since the returned object is not a standalone resource(manifest) that can be created.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -273,6 +273,14 @@ func (r *reconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Result
 		errs = append(errs, fmt.Errorf("failed to reconcile kubernetes.default endpoints and endpointslice: %w", err))
 	}
 
+	// Apply new ValidatingAdmissionPolicy to restrict the modification/creation of certain
+	// objects from the DataPlane which are already managed in the HostedCluster object
+	//
+	// TODO(jparrill): articulate the annotation and the force flag in the AdmissionPolicy.
```
muraee, line 279: nit
```suggestion
	// Apply new ValidatingAdmissionPolicy to restrict the modification/creation of certain
	// objects from the DataPlane which are managed by the HCCO
```

control-plane-operator/hostedclusterconfigoperator/controllers/resources/kas/featuregates.go
```diff
@@ -0,0 +1,16 @@
+package kas
+
+import (
+	configv1 "github.com/openshift/api/config/v1"
+	hyperconfig "github.com/openshift/hypershift/support/config"
+)
+
+func RecoverFeatureGates(fg *configv1.FeatureGateSpec) []string {
```
muraee, line 8: this is still not used as far as I can see, can we remove it?

jparrill:
canceling hold, lgtm label removed xD
/hold cancel

control-plane-operator/controllers/hostedcontrolplane/configoperator/reconcile.go
```diff
@@ -232,6 +232,7 @@ func ReconcileRole(role *rbacv1.Role, ownerRef config.OwnerRef, platform hyperv1
 			},
 		}...)
 	}
+	// TODO (jparrill): Add RBAC specific needs for Agent platform
```
enxebre, line 235: how is this related to this PR?

enxebre:
great feature!

control-plane-operator/hostedclusterconfigoperator/controllers/resources/kas/admissionpolicies.go
```diff
@@ -0,0 +1,150 @@
...
+	if err := reconcileICSPValidatingAdmissionPolicy(ctx, client, createOrUpdate); err != nil {
+		return fmt.Errorf("failed to reconcile ICSP Validating Admission Policy: %v", err)
+	}
+
+	return nil
+}
+
+func reconcileICSPValidatingAdmissionPolicy(ctx context.Context, client client.Client, createOrUpdate upsert.CreateOrUpdateFN) error {
+	icspAdmissionPolicy := AdmissionPolicy{Name: AdmissionPolicyNameICSP}
+	icspApiVersion := []string{"v1alpha1"}
```
csrwng, line 59: nit: s/Api/API/

You can also use the constants from the api itself:
```
import ( operatorv1alpha1 "github.com/openshift/api/operator/v1alpha1" )
...
icspAPIVersion := []string{operatorv1alpha1.GroupVersion.Version}
icspAPIGroup := []string{operatorv1alpha1.GroupVersion.Group}
```

control-plane-operator/hostedclusterconfigoperator/controllers/resources/kas/admissionpolicies.go
```diff
@@ -0,0 +1,150 @@
...
+	icspApiGroup := []string{"operator.openshift.io"}
```
csrwng, line 60: s/Api/API/

control-plane-operator/hostedclusterconfigoperator/controllers/resources/kas/admissionpolicies.go
```diff
@@ -0,0 +1,150 @@
...
+		return fmt.Errorf("error reconciling ICSP Validating Admission Policy: %v", err)
+	}
+
+	return nil
+}
+
+func reconcileConfigValidatingAdmissionPolicy(ctx context.Context, hcp *hyperv1.HostedControlPlane, client client.Client, createOrUpdate upsert.CreateOrUpdateFN) error {
+	// Config AdmissionPolicy
+	configAdmissionPolicy := AdmissionPolicy{Name: AdmissionPolicyNameConfig}
+	configApiVersion := []string{"v1"}
```
csrwng, line 75: See comment above about using api constants

csrwng: s/Api/API/

control-plane-operator/hostedclusterconfigoperator/controllers/resources/kas/admissionpolicies.go
```diff
@@ -0,0 +1,150 @@
...
+	configApiGroup := []string{"config.openshift.io"}
```
csrwng, line 76: s/Api/API/

csrwng (commented):
A couple of style nits

control-plane-operator/controllers/hostedcontrolplane/configoperator/reconcile.go
jparrill: I've seen this section when I was reviewing the new HCCO cert and account, and this is something I need add for the agent platform, not 100% related.

jparrill:
> /label tide/merge-method-squash

They was together in just 1 commit but Mulham mentioned in one of the reviews that it could be better to have both commits separated xD.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/kas/admissionpolicies.go
jparrill: makes sense

hasueki (changes requested):
/hold Need to evaluate impact on IBM Cloud

control-plane-operator/hostedclusterconfigoperator/controllers/resources/kas/admissionpolicies.go
```diff
@@ -0,0 +1,152 @@
...
+	AdmissionPolicyNameConfig = "config"
+	AdmissionPolicyNameICSP   = "icsp"
+)
+
+var (
+	defaultAdmissionPoliciesOperations = []k8sadmissionv1beta1.OperationType{"UPDATE", "DELETE"}
+	defaultMatchResourcesScope         = k8sadmissionv1beta1.ScopeType("*")
+	defaultMatchPolicyType             = k8sadmissionv1beta1.Equivalent
+	HCCOUserValidation                 = k8sadmissionv1beta1.Validation{
+		Expression: fmt.Sprintf("request.userInfo.username.contains('system:%s') || object.spec == oldObject.spec", config.HCCOUser),
```
muraee, line 36: why do we do a `contains` check here instead of an equality check?
With this, a user with the username `system:hosted-cluster-config-operator-fake` for example, would be allowed to edit these resources, no?

rtheis (commented):
What is the process for break glass access to make changes outside HCCO?

control-plane-operator/hostedclusterconfigoperator/controllers/resources/kas/admissionpolicies.go
jparrill: I would say yes, I can change it to be exactly that user.

jparrill:
> What is the process for break glass access to make changes outside HCCO?

Now we create a separated certificate for the HCCO user access so you will need to bake a new kubeconfig using the same HCCO certificates present in the HostedControlPlane namespace.

jparrill:
Fixing conformance tests in:

- https://github.com/openshift/origin/pull/28971

rtheis (commented):
Thank you.  Per our discussion, please ensure this supports `hypershift.openshift.io/disable-pki-reconciliation: "true"`.  Also, please only enable this for IBM Cloud starting with 4.17.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -277,6 +278,20 @@ func (r *reconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Result
 		errs = append(errs, fmt.Errorf("failed to reconcile kubernetes.default endpoints and endpointslice: %w", err))
 	}
 
+	releaseImageVersion, err := semver.Parse(releaseImage.Version())
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to parse release image version: %w", err)
+	}
+
+	// The exception for IBMCloudPlatform is due to the fact that the IBM will include new certificates for HCCO from 4.17 version
+	if !(hcp.Spec.Platform.Type == hyperv1.IBMCloudPlatform && (releaseImageVersion.Major == 4 && releaseImageVersion.Minor < 17)) {
```
jparrill, line 287: Hey @rtheis I think this is what you were mentioning, is that alright?
Regarding the DisablePKI annotation, yes, it's compatible.

EmilienM:
If someone backport this, please also include the commit in https://github.com/openshift/hypershift/pull/4609.
Thanks

PR #4306, opened 2024-07-02
NO-JIRA: Ensure VolumeSnapshots are deleted during cloud resource cleanup

davidvossel:
VolumeSnapshots are similar to PVCs in that they consume persistent cloud resources. If we're going to clean up PVCs, then it seems to make sense to cleanup VolumnSnapshots as well.

### DIFF
--- a/control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
+++ b/control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
@@ -9,6 +9,7 @@ import (
 	"sync"
 	"time"
 
+	snapshotv1 "github.com/kubernetes-csi/external-snapshotter/client/v6/apis/volumesnapshot/v1"
 	"github.com/openshift/hypershift/control-plane-operator/hostedclusterconfigoperator/api"
 	"github.com/openshift/hypershift/control-plane-operator/hostedclusterconfigoperator/controllers/resources/cco"
 	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
@@ -1857,6 +1858,17 @@ func (r *reconciler) ensureCloudResourcesDestroyed(ctx context.Context, hcp *hyp
 		log.Info("Persistent volumes are removed")
 	}
 
+	log.Info("Ensuring volume snapshots are removed")
+	removed, err = r.ensureVolumeSnapshotsRemoved(ctx)
+	if err != nil {
+		errs = append(errs, err)
+	}
+	if !removed {
+		remaining.Insert("volume-snapshots")
+	} else {
+		log.Info("Volume snapshots are removed")
+	}
+
 	return remaining, errors.NewAggregate(errs)
 }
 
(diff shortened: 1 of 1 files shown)


control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -1857,6 +1858,17 @@ func (r *reconciler) ensureCloudResourcesDestroyed(ctx context.Context, hcp *hyp
 		log.Info("Persistent volumes are removed")
 	}
 
+	log.Info("Ensuring volumn snapshots are removed")
+	removed, err = r.ensureVolumeSnapshotsRemoved(ctx)
```
jparrill, line 1862: I would pass the HCP Namespace by argument and put the `ensureVolumeSnapshotsRemoved` as a public function in a more common place to be reused (E.G support/utils (?)).

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -2131,6 +2143,22 @@ func (r *reconciler) ensureServiceLoadBalancersRemoved(ctx context.Context) (boo
 	return removed, nil
 }
 
+func (r *reconciler) ensureVolumeSnapshotsRemoved(ctx context.Context) (bool, error) {
+	log := ctrl.LoggerFrom(ctx)
+	vss := &snapshotv1.VolumeSnapshotList{}
+	if err := r.client.List(ctx, vss); err != nil {
```
jparrill, line 2149: You will need to add a `client.ListOptions` if you implement the suggestion from the comment above.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
davidvossel: > I would pass the HCP Namespace by argument 

This is cleaning up volume snapshots within the guest cluster. The hcp namespace would not be relevant here. 

> put the ensureVolumeSnapshotsRemoved as a public function in a more common place to be reused (E.G support/utils (?)).

That would be inconsistent with the current way this code is structured. I'm following the pattern used by this function for other resources that need to be cleaned up. 

examples a few lines up. 

```
	removed, err = r.ensureIngressControllersRemoved(ctx, hcp)
	if err != nil {
		errs = append(errs, err)
	}
	if !removed {
		remaining.Insert("ingress-controllers")
	} else {
		log.Info("Ingress controllers are removed")
	}
	log.Info("Ensuring load balancers are removed")
	removed, err = r.ensureServiceLoadBalancersRemoved(ctx)
	if err != nil {
		errs = append(errs, err)
	}
	if !removed {
		remaining.Insert("loadbalancers")
	} else {
		log.Info("Load balancers are removed")
	}
	log.Info("Ensuring persistent volumes are removed")
	removed, err = r.ensurePersistentVolumesRemoved(ctx)
```

davidvossel: @jparrill any more thoughts on this PR?

PR #4309, opened 2024-07-03
WIP: HOSTEDCP-1734: Drop extra haproxy

muraee:

This PR allows the shared-ingress controller to reconcile the HostedCluster.Spec.Networking.APIServer.AdvertiseAddress value to match the assigned static IP for the data-plane so that the kubenretes.svc endpoint in the guest cluster will match that IP. Requests origniating from the data-plane will now target that static IP directly which is intercepted by the HAProxy and forwarded to the shared-ingress router preserving the static IP as the destination IP.

depends on https://github.com/openshift/hypershift/pull/4285

Fixes #

**Checklist**


enxebre:
/close
in favour of https://github.com/openshift/hypershift/pull/4330

PR #4310, opened 2024-07-03
NO-JIRA: Let cmd unit tests run without a running Kubernetes cluster

muraee:

Currently the cmd unit tests require a valid Kubernetes cluster running to get its client and do some validations. This PR fixes that and drop this requirement.

Fixes #

**Checklist**


sjenning (commented):
lgtm, just one nit

support/util/util.go
```diff
@@ -369,7 +369,7 @@ func GetMgmtClusterCPUArch(ctx context.Context) (string, error) {
 	// Get the API version in JSON format
 	versionJSON, err := kc.RESTClient().Get().AbsPath("/version").DoRaw(ctx)
 	if err != nil {
-		return "", err
+		return "", fmt.Errorf("failed to check mgmt cluster CPU arch: %v", err)
```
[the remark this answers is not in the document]
muraee: done

cmd/util/client.go
```diff
@@ -33,6 +35,10 @@ func GetConfig() (*rest.Config, error) {
 
 // GetClient creates a controller-runtime client for Kubernetes
 func GetClient() (crclient.Client, error) {
+	if os.Getenv("FAKE_CLIENT") == "true" {
```
enxebre, line 38: I wonder if the client should be a param of createCluster instead?

muraee: that would require a bigger refactor, since GetClient() is called in multiple places across the cmd code. I can do that if you think that's better.

PR #4313, opened 2024-07-03
NO-JIRA: e2e tests: skip checking ValidKubeVirtInfraNetworkMTU if no workers

orenc1:
In case there are no worker nodes in the HostedCluster, the condition ValidKubeVirtInfraNetworkMTU is missing since the MTU prober hasn't been run, so no MTU is being reported by the Cluster Network Operator. In such case, we're removing that condition from the expectedConditions map.

Fixes #

**Checklist**


sjenning:

PR #4316, opened 2024-07-04
NO-JIRA: Default azure to run as managed-service=aro

enxebre:
There's a single valid known use case for azure atm. Let's default to it.
Also this drops rosa variable since there's no usage nor valid semantic for it atm.

**Checklist**

### DIFF
--- a/api/hypershift/v1beta1/hostedcluster_types.go
+++ b/api/hypershift/v1beta1/hostedcluster_types.go
@@ -275,9 +275,6 @@ const (
 	// AroHCP represents the ARO HCP managed service offering
 	AroHCP = "ARO-HCP"
 
-	// RosaHCP represents the ROSA HCP managed service offering
-	RosaHCP = "ROSA-HCP"
-
 	// HostedClusterSizeLabel is a label on HostedClusters indicating a size based on the number of nodes.
 	HostedClusterSizeLabel = "hypershift.openshift.io/hosted-cluster-size"
 
--- a/cmd/install/install.go
+++ b/cmd/install/install.go
@@ -152,7 +152,7 @@ func (o *Options) Validate() error {
 		errs = append(errs, fmt.Errorf("when invoking this command with the --rhobs-monitoring flag, the --enable-cvo-management-cluster-metrics-access flag is not supported "))
 	}
 
-	if len(o.ManagedService) > 0 && o.ManagedService != hyperv1.AroHCP && o.ManagedService != hyperv1.RosaHCP {
+	if len(o.ManagedService) > 0 && o.ManagedService != hyperv1.AroHCP {
 		errs = append(errs, fmt.Errorf("not a valid managed service type: %s", o.ManagedService))
 	}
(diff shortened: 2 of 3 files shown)


enxebre:
/hold
this needs the e2e to pass with the flag

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -275,9 +275,6 @@ const (
 	// AroHCP represents the ARO HCP managed service offering
 	AroHCP = "ARO-HCP"
 
-	// RosaHCP represents the ROSA HCP managed service offering
-	RosaHCP = "ROSA-HCP"
```
jparrill, line 279: Is this not needed anymore?

cmd/install/install.go
```diff
@@ -152,7 +152,7 @@ func (o *Options) Validate() error {
 		errs = append(errs, fmt.Errorf("when invoking this command with the --rhobs-monitoring flag, the --enable-cvo-management-cluster-metrics-access flag is not supported "))
 	}
 
-	if len(o.ManagedService) > 0 && o.ManagedService != hyperv1.AroHCP && o.ManagedService != hyperv1.RosaHCP {
```
jparrill, line 155: would we add another check for RosaHCP in any other place? or this is not useful anymore?

jparrill (commented):
Mostly questions. Otherwise LGTM

api/hypershift/v1beta1/hostedcluster_types.go
enxebre: AFAICT as mentioned in the PR desc: there's no usage nor valid semantic for it atm.
This should have never been added. It does nothing.

cmd/install/install.go
enxebre: AFAICT as mentioned in the PR desc: there's no usage nor valid semantic for it atm.
This should have never been added. It does nothing.

enxebre:
/hold
until we sort out azure e2e test

cmd/install/install.go
```diff
@@ -183,6 +183,7 @@ func NewCommand() *cobra.Command {
 	opts.ExternalDNSImage = ExternalDNSImage
 	opts.CertRotationScale = 24 * time.Hour
 	opts.EnableSizeTagging = false
+	opts.ManagedService = hyperv1.AroHCP
```
muraee, line 186: wouldn't this enable shared-ingress for all platforms?

cmd/install/install.go
jparrill: 👍 perfect thanks!

api/hypershift/v1beta1/hostedcluster_types.go
jparrill: 👍 perfect thanks!

cmd/install/install.go
enxebre: indeed. Fixed.

PR #4322, opened 2024-07-05
OCPBUGS-35905: E2E test to verify openshift-apiserver TLS certificates

jparrill:
Fixes #[OCPBUGS-35905](https://issues.redhat.com/browse/OCPBUGS-35905)

- Added E2E test to validate the proper presence of the TLS certificates into openshift-apiserver

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/oapi/deployment_test.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/oapi/deployment_test.go
@@ -87,16 +87,18 @@ func TestReconcileOpenshiftAPIServerDeploymentTrustBundle(t *testing.T) {
 	hcp.Name = "name"
 	hcp.Namespace = "namespace"
 	testCases := []struct {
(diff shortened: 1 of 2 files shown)


csrwng:
/hold
@jparrill I fixed this in this PR https://github.com/openshift/hypershift/pull/4218 which should now be ready.
The fix involves creating a separate mount per file from the projected volume.

jparrill:
@csrwng Ok, LGTM, I will keep this PR because of the E2E tests to avoid regressions. Once your PR got merged, I will remove the volumeMount part of the PR and rebase from `main`, only adding the E2E tests.

enxebre:
nice, thanks! is there a unit test that covers the same so we fail earlier?

jparrill:
> nice, thanks! is there a unit test that covers the same so we fail earlier?

Done, I've added the proxy test which it was missing one. Let me know if you have any other concerns @enxebre.

PR #4330, opened 2024-07-08
HOSTEDCP-1734, HOSTEDCP-1737: use proxy-protocol-v2 for shared ingress

muraee:
The use of proxy-protocol-v2 allows to send arbitrary data in the proxy header. This allows the data-plane HAProxy in this case to send the clusterID in the header. The shared-ingress HAproxy can then decode the attached clusterID from the proxy header and route the traffic accordingly to the correct KAS. Note that proxy-protocol-v2 is only supported in HAProxy v2.9+

Fixes #

**Checklist**


enxebre:
/hold
until managed=aro e2e are in place

support/testutil/testutil.go
```diff
@@ -74,6 +74,12 @@ type options struct {
 
 type option func(*options)
 
+func WithExtension(extension string) option {
```
enxebre, line 77: did this sneak in or is it intentional?

hypershift-operator/controllers/nodepool/haproxy.go
```diff
@@ -237,105 +223,60 @@ func MustAsset(name string) string {
...
+func apiServerProxyConfig(haProxyImage, cpoImage, clusterID,
+	externalAPIAddress, internalAPIAddress string,
 	externalAPIPort, internalAPIPort int32, proxyAddr, noProxy, serviceNetwork, clusterNetwork string) ([]byte, error) {
 	config := &ignitionapi.Config{}
 	config.Ignition.Version = ignitionapi.MaxVersion.String()
 
+	if sharedingress.UseSharedIngress() {
+		// TODO: get the image from somewhere else
+		// proxy protocol v2 requires haproxy v2.9+
+		haProxyImage = "docker.io/haproxy:2.9.9"
```
enxebre, line 240: wouldn't we hit docker pull rate limiting e.g in CI? can we please source this from quay?

Also we were using proxy protocol already, so I assume you mean TLV support?
Can add a link in the comment backing the statement, e.g. https://www.haproxy.com/blog/announcing-haproxy-2-9#proxy-protocol-tlv-fields

Also I created https://issues.redhat.com/browse/HOSTEDCP-1819 to track this.

hypershift-operator/controllers/sharedingress/router.go
```diff
@@ -172,7 +190,8 @@ func buildHCPRouterContainerMain() func(*corev1.Container) {
 			"haproxy",
 		}
 		// 4.16.0-rc.3 haproxy router image.
-		c.Image = "quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:06e8c4156888f50be0ac46bf21337ecda329befa0f9a6bb1df8bce078ae04aaa"
+		// c.Image = "quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:06e8c4156888f50be0ac46bf21337ecda329befa0f9a6bb1df8bce078ae04aaa"
```
enxebre, line 193: Please drop commented code, include a TODO here and incorporate same feedback than in https://github.com/openshift/hypershift/pull/4330/files#r1671941572

hypershift-operator/controllers/sharedingress/testdata/zz_fixture_TestGenerateConfig_When_there_s_two_HostedClusters_with_all_their_Routes_and_SVCs_it_should_generate_the_config_with_frontends_and_backends_for_both.cfg
```diff
@@ -19,9 +19,14 @@ defaults
...
   bind :::6443 v4v6 accept-proxy
   tcp-request inspect-delay 5s
+
+  # fc_pp_tlv(0x20) extracts and decodes the custom tlv proxy header sent from the data-pane HAProxy which contains the clusterID.
+  tcp-request content set-var(sess.cluster_id) fc_pp_tlv(0x20)
   tcp-request content accept if { req_ssl_hello_type 1 }
-  acl is_test-hc1-kube-apiserver dst 4.4.4.4
-  acl is_test-hc2-kube-apiserver dst 4.4.4.4
+
+  log-format "%{+Q}o\ custom_proxy_tlv_header = %[var(sess.cluster_id)], %ci:%cp -> %fi:%fp [%t] %ft %b/%s %Tw/%Tc/%Tt %B %ts %ac/%fc/%bc/%sc/%rc %sq/%bq"
```
enxebre, line 27: can we add a comment with logging example or a link to a source?

hypershift-operator/controllers/nodepool/haproxy.go
```diff
@@ -221,11 +208,10 @@ type fileToAdd struct {
 	source   func() ([]byte, error)
 	name     string
 	mode     int
-	params   map[string]string
+	params   map[string]any
```
enxebre, line 211: why does this need any now?

muraee: because of the new bool var `UseProxyProtocol`, passing it as a string "true" for example would require us to do string equality in the template instead of just an if statement. Also we don't have to parse the ports integers.

hypershift-operator/controllers/nodepool/haproxy.go
muraee: we were using proxy protocol v1, not v2. But yeah, TLV support is what we need. I will update it.

muraee: done

hypershift-operator/controllers/sharedingress/testdata/zz_fixture_TestGenerateConfig_When_there_s_two_HostedClusters_with_all_their_Routes_and_SVCs_it_should_generate_the_config_with_frontends_and_backends_for_both.cfg
muraee: added an example output with a link explaining the different logging vars.

enxebre:
cc @jharrington22

PR #4331, opened 2024-07-08
OCPBUGS-36680: ensure additionalTrustBundle propogates to workers

Patryk-Stefanski:
Adds a has for the additionalTrustBundle CM specified in the HC to the token secret, this will regenerate the payload and re rollout the nodes with the up to date additionalTrustBundle

Fixes # https://issues.redhat.com/browse/OCPBUGS-36680

**Checklist**

### DIFF
--- a/hypershift-operator/controllers/nodepool/nodepool_controller.go
+++ b/hypershift-operator/controllers/nodepool/nodepool_controller.go
@@ -103,6 +103,7 @@ const (
 	TokenSecretTokenKey                       = "token"
 	TokenSecretPullSecretHashKey              = "pull-secret-hash"
 	TokenSecretHCConfigurationHashKey         = "hc-configuration-hash"
+	TokenSecretAdditionalTrustBundleKey       = "additional-trust-bundle-hash"
 	TokenSecretConfigKey                      = "config"
 	TokenSecretAnnotation                     = "hypershift.openshift.io/ignition-config"
 	TokenSecretIgnitionReachedAnnotation      = "hypershift.openshift.io/ignition-reached"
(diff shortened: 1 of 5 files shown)


hypershift-operator/controllers/nodepool/nodepool_controller.go
```diff
@@ -835,8 +844,15 @@ func (r *NodePoolReconciler) reconcile(ctx context.Context, hcluster *hyperv1.Ho
 	if err != nil {
 		return ctrl.Result{}, fmt.Errorf("failed to hash HostedCluster configuration: %w", err)
 	}
+	additionalTrustBundle := ""
+	if hcluster.Spec.AdditionalTrustBundle != nil {
+		additionalTrustBundle, err = r.getAdditionalTrustBundle(ctx, hcluster)
+	}
+	if err != nil {
```
bryan-cox, line 851: Shouldn't this just go inside the if block and be with the err assignment on L849?

hypershift-operator/controllers/nodepool/nodepool_controller.go
```diff
@@ -835,8 +844,15 @@ func (r *NodePoolReconciler) reconcile(ctx context.Context, hcluster *hyperv1.Ho
 	if err != nil {
 		return ctrl.Result{}, fmt.Errorf("failed to hash HostedCluster configuration: %w", err)
 	}
+	additionalTrustBundle := ""
```
bryan-cox, line 847: Should this whole block be dropped? The same code is above at L632.

hypershift-operator/controllers/nodepool/nodepool_controller.go
```diff
@@ -3032,6 +3049,32 @@ func getPullSecretName(ctx context.Context, crclient client.Client, hostedCluste
 	return pullSecret.Name, nil
 }
 
+func (r *NodePoolReconciler) getAdditionalTrustBundle(ctx context.Context, hostedCluster *hyperv1.HostedCluster) (string, error) {
+	additionalTrustBundle := &corev1.ConfigMap{}
+	if err := r.Client.Get(ctx, client.ObjectKey{Namespace: hostedCluster.Namespace, Name: hostedCluster.Spec.AdditionalTrustBundle.Name}, additionalTrustBundle); err != nil {
+		return "", fmt.Errorf("cannot get additionalTrustBundle %s/%s: %w", hostedCluster.Namespace, hostedCluster.Spec.AdditionalTrustBundle.Name, err)
```
bryan-cox, line 3055: Should there be an additional check here to not return an error if it's just not found? This field is optional in the API.

Patryk-Stefanski: The only time this function gets called is when the additionalTrustBundle is specified in the HC.

hypershift-operator/controllers/nodepool/nodepool_controller.go
bryan-cox: Right, I mean just return the whole CM back as an object then just get the name and data when you need it from that object rather than retrieving it twice for different reasons.

PR #4333, opened 2024-07-08
OCPBUGS-34820: Delete IDMS in dataplane once HCP ICS field is removed

jparrill:
There was an issue in the reconciliation process where the IDMS entry in the DataPlane was only created or updated when the user modified the ImageContentSource. However, if this field was deleted, the IDMS entry was not removed. This PR ensures that the IDMS entry is properly deleted when appropriate.

Fixes #[OCPBUGS-34820](https://issues.redhat.com/browse/OCPBUGS-34820)


control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources_test.go
```diff
@@ -919,3 +982,70 @@ func TestReconcileClusterVersion(t *testing.T) {
...
+				g.Expect(errors.IsNotFound(err)).To(BeTrue(), "expecting a not-found error, received: %v", err)
+			}
+
+			if !tc.reconcileTwice && !tc.expectedToFail {
+				g.Expect(err).ToNot(HaveOccurred())
+				g.Expect(len(tc.hcp.Spec.ImageContentSources)).To(Equal(len(idms.Spec.ImageDigestMirrors)))
+				g.Expect(tc.expectedIDMS.Spec).To(BeEquivalentTo(idms.Spec))
+			}
+
+			if tc.reconcileTwice {
```
Patryk-Stefanski, line 1037: can we change the name from reconcileTwice to something more meaningful like removeICSAndReconcile

PR #4335, opened 2024-07-09
HOSTEDCP-1795,HOSTEDCP-1796: Customize the self-generated cert validity and rotation

jparrill:
Now via annotation the self-generated certificates validity and rotation could be customised. This will help to test the certificates regeneration

Fixes #[HOSTEDCP-1779](https://issues.redhat.com/browse/HOSTEDCP-1779)

- https://issues.redhat.com/browse/HOSTEDCP-1795
- https://issues.redhat.com/browse/HOSTEDCP-1796

Another more complete approach is implemented here: https://github.com/openshift/hypershift/pull/4318 for [HOSTEDCP-1795](https://issues.redhat.com/browse/HOSTEDCP-1795)

We decided to go with a much smaller version and not exposed to the API.

### DIFF
--- a/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
+++ b/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
@@ -2583,6 +2583,16 @@ func reconcileControlPlaneOperatorDeployment(
 		},
 	}
 
+	if hc.Annotations[certs.CertificateValidityAnnotation] != "" {
+		certValidity := hc.Annotations[certs.CertificateValidityAnnotation]
+		deployment.Spec.Template.Spec.Containers[0].Env = append(deployment.Spec.Template.Spec.Containers[0].Env,
+			corev1.EnvVar{
+				Name:  certs.CertificateValidityEnvVar,
+				Value: certValidity,
+			},
+		)
+	}
+
 	if openShiftTrustedCABundleConfigMapExists {
 		hyperutil.DeploymentAddOpenShiftTrustedCABundleConfigMap(deployment)
 	}
--- a/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
+++ b/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
@@ -2593,6 +2593,16 @@ func reconcileControlPlaneOperatorDeployment(
 		)
 	}
 
+	if hc.Annotations[certs.CertificateRenewalAnnotation] != "" {
+		certRenewalPercentage := hc.Annotations[certs.CertificateRenewalAnnotation]
+		deployment.Spec.Template.Spec.Containers[0].Env = append(deployment.Spec.Template.Spec.Containers[0].Env,
+			corev1.EnvVar{
+				Name:  certs.CertificateRenewalEnvVar,
+				Value: certRenewalPercentage,
+			},
+		)
+	}
+
 	if openShiftTrustedCABundleConfigMapExists {
 		hyperutil.DeploymentAddOpenShiftTrustedCABundleConfigMap(deployment)
 	}
--- a/support/certs/tls.go
+++ b/support/certs/tls.go
@@ -14,6 +14,7 @@ import (
 	"math"
 	"math/big"
 	"net"
+	"os"
 	"time"
 
 	"github.com/google/go-cmp/cmp"
@@ -45,6 +46,11 @@ const (
 	TLSSignerKeyMapKey = "tls.key"
 	// UserCABundleMapKeyis the key value in a user-provided CA configMap.
 	UserCABundleMapKey = "ca-bundle.crt"
+	// Custom certificate validity. The format of the annotation is a go duration string with a numeric component and unit.
+	// Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h"
+	CertificateValidityAnnotation = "hypershift.openshift.io/certificate-validity"
+	CertificateValidityEnvVar     = "CERTIFICATE_VALIDITY"
+	CertificateValidityDefault    = ValidityOneYear
 )
 
 // CertCfg contains all needed fields to configure a new certificate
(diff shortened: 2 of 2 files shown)


hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -2583,6 +2583,16 @@ func reconcileControlPlaneOperatorDeployment(
 		},
 	}
 
+	if hc.Annotations[certs.CertificateValidityAnnotation] != "" {
+		certValidity := hc.Annotations[certs.CertificateValidityAnnotation]
+		deployment.Spec.Template.Spec.Containers[0].Env = append(deployment.Spec.Template.Spec.Containers[0].Env,
+			corev1.EnvVar{
+				Name:  certs.CertificateValidityAnnotation,
```
csrwng, line 2590: s/CertificateValidityAnnotation/CertificateValidityEnvVar/

support/certs/tls.go
```diff
@@ -356,11 +369,16 @@ func ReconcileSignedCert(
 		secret.Data[caKey] = append([]byte(nil), ca.Data[opts.CASignerCertMapKey]...)
 	}
 
+	certValidity, err := time.ParseDuration(os.Getenv(CertificateValidityEnvVar))
```
csrwng, line 372: Can we check whether there's a value, and if not, then use the default. Otherwise try parsing. If there's an error parsing throw the error to prevent things from progressing if an invalid value is passed.

support/certs/tls.go
```diff
@@ -356,11 +369,20 @@ func ReconcileSignedCert(
...
 	}
 
+	var certValidity = ValidityOneYear
+
+	if os.Getenv(CertificateValidityEnvVar) != "" {
+		customCertValidityEnvVar, err := time.ParseDuration(os.Getenv(CertificateValidityEnvVar))
+		if err != nil {
+			certValidity = customCertValidityEnvVar
+		}
+	}
```
muraee, line 379:
```suggestion
	if value := os.Getenv(CertificateValidityEnvVar); value != "" {
		customCertValidityEnvVar, err := time.ParseDuration(value)
		if err != nil {
			return fmt.Errorf("invalid validity duration: %v", err)
		}
		certValidity = customCertValidityEnvVar
	}
```

support/certs/tls.go
```diff
@@ -356,11 +369,20 @@ func ReconcileSignedCert(
 		secret.Data[caKey] = append([]byte(nil), ca.Data[opts.CASignerCertMapKey]...)
 	}
 
+	var certValidity = ValidityOneYear
```
muraee, line 372: no need for the `var` keyword here
```suggestion
	certValidity := ValidityOneYear
```

support/certs/tls.go
jparrill: Talking with @csrwng we agree to avoid fail when the duration is not well set to avoid interfere with the normal workflow.

support/certs/tls.go
```diff
@@ -45,6 +46,10 @@ const (
 	TLSSignerKeyMapKey = "tls.key"
 	// UserCABundleMapKeyis the key value in a user-provided CA configMap.
 	UserCABundleMapKey = "ca-bundle.crt"
+	// Custom certificate validity. The format of the annotation is a go duration string with a numeric component and unit.
+	// Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h"
+	CertificateValidityAnnotation = "hypershift.openshift.io/certificate-validity"
```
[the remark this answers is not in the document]
jparrill: it's used in here: https://github.com/openshift/hypershift/pull/4335/files#diff-b5cde720842b19b53da3f454d827e27327c4559b2e4849906dac728586f51eb3R2586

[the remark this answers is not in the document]
muraee: oh didn't see it :)

support/certs/tls.go
csrwng: Please fail if we cannot parse the duration.

support/certs/tls.go
```diff
@@ -356,15 +373,36 @@ func ReconcileSignedCert(
...
 		KeyUsages:    x509.KeyUsageKeyEncipherment | x509.KeyUsageDigitalSignature,
 		ExtKeyUsages: extUsages,
-		Validity:     ValidityOneYear,
+		Validity:     certValidity,
 		DNSNames:     dnsNames,
 		IPAddresses:  ipAddresses,
 	}
-	if err := ValidateKeyPair(secret.Data[keyKey], secret.Data[crtKey], cfg, 30*ValidityOneDay); err == nil {
+
+	renewalDate := 30 * ValidityOneDay
```
csrwng, line 395: nit:
s/renewalDate/minimumRemainingValidity/

support/certs/tls.go
```diff
@@ -45,6 +47,13 @@ const (
 	TLSSignerKeyMapKey = "tls.key"
 	// UserCABundleMapKeyis the key value in a user-provided CA configMap.
 	UserCABundleMapKey = "ca-bundle.crt"
+	// Custom certificate validity. The format of the annotation is a go duration string with a numeric component and unit.
+	// Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h"
+	CertificateValidityAnnotation = "hypershift.openshift.io/certificate-validity"
+	CertificateValidityEnvVar     = "CERTIFICATE_VALIDITY"
+	// Custom certificate renewal percentage. The format of the annotation is a float64 value between 0 and 1.
```
csrwng, line 54: nit: "certificate renewal percentage" doesn't really explain well what the percentage is for. 
It might be easier to explain by an example. The certificate will renew when there is less than [this percentage] of the validity period left in a certificate's validity. For example... you set the validity to 100 days, and the percentage validity at "0.30".  It means that when there are less than  [ (100 days * .30) = 30 ] 30 days before the certificate expires, the certificate will renew.

PR #4337, opened 2024-07-09
WIP: Add hc.Spec.Tolerations to the hostedcluster API

tiraboschi:
**What this PR does / why we need it**:
Add hc.Spec.Tolerations to the hostedcluster API

Work In Progress

**Which issue(s) this PR fixes** *(optional, use `fixes #<issue_number>(, fixes #<issue_number>, ...)` format, where issue_number might be a GitHub issue, or a Jira story*:
Fixes https://issues.redhat.com/browse/CNV-39630

**Checklist**
- [ ] Subject and description added to both, commit and PR.
- [ ] Relevant issues have been referenced.
- [ ] This change includes docs. 
- [ ] This change includes unit tests.

PR #4339, opened 2024-07-09
 CNV-39630: Add ability to set tolerations on the HostedCluster 

davidvossel:

This PR adds the ability to specify tolerations on the HCP pods. This gives HCP administrators the ability to use HCP with custom taints in order to further influence HCP scheduling. 

The `hcp` cli has also been extended with the `--toleration` cli arg during cluster creation. This cli arg can be set multiple times to allow for multiple tolerations to be added to the hcp pods.

Example: `--tolerations="key=node-role.kubernetes.io/master,operator=Exists,effect=NoSchedule"` would allow the HCP pods to be scheduled on master nodes.


support/config/deployment.go
```diff
@@ -269,6 +269,8 @@ func (c *DeploymentConfig) setControlPlaneIsolation(hcp *hyperv1.HostedControlPl
 		})
 	}
 
+	c.Scheduling.Tolerations = append(c.Scheduling.Tolerations, hcp.Spec.Tolerations...)
```
tiraboschi, line 272: I was thinking if in terms of visibility it would be better to introduce a specific `c.setAdditionalTolerations(hcp)` and call it from `setLocation`.
Not 100% sure.

support/config/deployment.go
```diff
@@ -269,6 +269,8 @@ func (c *DeploymentConfig) setControlPlaneIsolation(hcp *hyperv1.HostedControlPl
...
+
```
tiraboschi, line 273: Can you please also add a unit test in support/config/deployment_test.go ?
You can take it from https://github.com/openshift/hypershift/pull/4337/files#diff-17ac58455ee34787c5ace14ebafe8b2e3ee34bb36ff47bc5dc73ea1b0835cc41

support/config/deployment.go
davidvossel: makes sense. I changed this

support/config/deployment.go
davidvossel: I cherry picked that part of your PR in as a separate commit and set you as the author

support/config/deployment_test.go
```diff
@@ -167,6 +167,52 @@ func TestSetControlPlaneIsolation(t *testing.T) {
...
+			Key:      "key2",
+			Operator: corev1.TolerationOpEqual,
+			Value:    "value2",
+			Effect:   corev1.TaintEffectNoSchedule,
+		},
+	}
+	hcp.Spec.Tolerations = append(hcp.Spec.Tolerations, customTolerations...)
+
+	cfg := &DeploymentConfig{}
+	cfg.setControlPlaneIsolation(hcp)
```
tiraboschi, line 191: I think you have now to call _`setAdditionalTolerations`_

davidvossel: fixed, thanks

davidvossel:
@tiraboschi when you get a chance, can you take another pass at this. I think i sorted out the verify and unit failures. For some reason an aws client was impacted by this change and I needed to update that for verify to pass

PR #4340, opened 2024-07-10
CPO/registry-operator: add trusted-ca-bundle-managed to mounts

EmilienM:

`trusted-ca-bundle-managed` is a ConfigMap that contains the trusted CA
and for the OpenStack platform we'll need this bundle to be mounted so
the cluster-image-registry-operator will be able to authenticate against
an OpenStack cloud that uses a self-signed certificate and that was
added to the additional trust bundle.


EmilienM:
/hold
I want to discuss this in office hours first.

EmilienM:
/close
I found another way to do it without it. We need to create specific credentials secrets.

PR #4342, opened 2024-07-10
HOSTEDCP-1818:fix(build): update to the fixed clamav image

celebdor:

The upcoming OpenShift Pipelines which will be deployed shortly has stricter validations on Pipelines and Task manifests. Clamav would fail the new validations. This commit should make us ready.
Fixes #[HOSTEDCP-1818](https://issues.redhat.com//browse/HOSTEDCP-1818)

**Checklist**


bryan-cox:
/retitle HOSTEDCP-1818:fix(build): update to the fixed clamav image

PR #4343, opened 2024-07-10
OCPBUGS-33934: Add newline after TLS certs referenced by image.config

jparrill:
Once you add more than one certificate to the ConfigMap referenced in the image.config.openshift.io spec field, the getImageRegistryCABundle processes the certificates properly. However, if you add these through the UI console, the user might not include the newline character, which results in overlapping certificates. This PR ensures a newline is always added after each certificate processed.

Fixes #[OCPBUGS-33934](https://issues.redhat.com/browse/OCPBUGS-33934)


CrystalChun (approved):
/lgtm

PR #4350, opened 2024-07-10
OSASINFRA-3525: Initial PR to start supporting OpenStack platform

EmilienM:

This PR is small enough to be reviewed but in its current state is a work in progress.
Our team has been working on that branch:

https://github.com/shiftstack/hypershift/commits/openstack-support/

And we plan to present squashed commits in relatively small PRs.
They don't expect to pass the OpenStack e2e yet, but at least they wouldn't break anything and
feedback can be provided by the Hypershift reviewers.

We hope to iterate that way.

Some TODO/FIXME in the code should be ignored unless red flags.


api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1823,6 +1833,12 @@ type AzurePlatformSpec struct {
 	SecurityGroupID string `json:"securityGroupID,omitempty"`
 }
 
+// OpenStackPlatformSpec specifies configuration for clusters running on OpenStack.
+type OpenStackPlatformSpec struct {
```
EmilienM, line 1837: Please don't review the spec here. We changed it later and this will be proposed in a new PR after this one.
Thanks

hypershift-operator/controllers/hostedcluster/internal/platform/openstack/openstack.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+}
+
+func (a OpenStack) ReconcileCAPIInfraCR(ctx context.Context, client client.Client, createOrUpdate upsert.CreateOrUpdateFN, hcluster *hyperv1.HostedCluster,
+	controlPlaneNamespace string, apiEndpoint hyperv1.APIEndpoint) (client.Object, error) {
+	return nil, nil
+}
+
+func (a OpenStack) CAPIProviderDeploymentSpec(hcluster *hyperv1.HostedCluster, _ *hyperv1.HostedControlPlane) (*appsv1.DeploymentSpec, error) {
+	// image := a.capiProviderImage
```
EmilienM, line 36: This will be uncommented and adjusted later when we enable the CAPI provider cluster. Please ignore the comments.

api/hypershift/v1beta1/hostedcluster_types.go
mdbooth: @EmilienM Just a warning: although we plan to change it and we don't expect anybody to actually use it...

We should still only commit backwards-compatible changes to Hypershift unless there are exceptional circumstances, as it can still cause grief.

Are these 2 fields in their expected final form? If not, lets remove them.

EmilienM: I'll be careful the next time; however I suggest we move forward here. I really don't want to burn unnecessary CI jobs here to iterate again on this one.
The next PR is going to update that.

PR #4349, opened 2024-07-10
OCPBUGS-34865: Revert "Add HTTP(s) konnectivity proxy and use it with OpenShift APIServer"

csrwng:
**What this PR does / why we need it**:
This reverts commit 556a45f5cdfb2e176f1a30191945ea6571491212.

Current code is failing conformance tests: https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-ci-openshift-hypershift-release-4.17-periodics-e2e-aws-ovn-conformance/1811055875415085056 

**Checklist**
- [ ] Subject and description added to both, commit and PR.
- [ ] Relevant issues have been referenced.
- [ ] This change includes docs. 
- [ ] This change includes unit tests.

PR #4351, opened 2024-07-11
OCPBUGS-34865: Fix name resolution for HTTPS konnectivity proxy

csrwng:
The new HTTPS konnectivity proxy requires that DNS names be resolved before generating a proper CONNECT string for the Konnectivity server. However the dialer was returning a nil IP if it didn't think it needed to resolve a name through the hosted cluster. This fix is to always resolve a name if using the HTTPS konnectivity proxy.

Fixes #[OCPBUGS-34865](https://issues.redhat.com/browse/OCPBUGS-34865)

**Checklist**


wking (approved):
Sounds convincing to me, patch is small, and we can't get much worse than #4349's revert, so:

/lgtm

PR #4161, opened 2024-06-05
OCPBUGS-24400: Recover the ETCD member on HostedCluster boot

jparrill:
Eventually, the ETCD deployment encountered issues booting when many HostedControlPlanes were being created. This caused one or more members to get stuck while trying to join the cluster but unable to do so. Once a pod is stuck in CrashLoopBackOff, the script will now create a snapshot and then remove the member folder, allowing the member to be properly formed and added to the cluster on the next attempt.

**Which issue(s) this PR fixes**:
Fixes #[OCPBUGS-24400](https://issues.redhat.com/browse/OCPBUGS-24400)

PR #4354, opened 2024-07-11
OCPBUGS-24400: Recreate the ETCD statefulSet on HostedCluster deployment when stays failing

jparrill:
Eventually, the ETCD deployment encountered issues booting when many HostedControlPlanes were being created. This caused one or more members to get stuck while trying to join the cluster but unable to do so. Once a pod is stuck in CrashLoopBackOff, after the time of etcdRecoveryPeriod a window will be opened to check if the ETCD is Ok. if the ETCD is not ok the stateful set and PVC inside will be updated with new labels forcing recreation in the next reconcile loop. After the time of etcdRecoveryTimeout the controller will not try to recover it again to avoid issues in other scenarios.

Which issue(s) this PR fixes:
Fixes #[OCPBUGS-24400](https://issues.redhat.com/browse/OCPBUGS-24400)

This is the second approach, the initial one is: https://github.com/openshift/hypershift/pull/4161

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
@@ -115,6 +115,9 @@ const (
 
 	hcpReadyRequeueInterval    = 1 * time.Minute
 	hcpNotReadyRequeueInterval = 15 * time.Second
+
+	etcdRecoveryPeriod  = 5 * time.Minute
+	etcdRecoveryTimeout = 10 * time.Minute
 )
 
 type InfrastructureStatus struct {
@@ -983,7 +986,8 @@ func (r *HostedControlPlaneReconciler) reconcile(ctx context.Context, hostedCont
 			return fmt.Errorf("failed to reconcile etcd: %w", err)
 		}
 		// Block until etcd is fully rolled out at the desired generation
-		if ready := util.IsStatefulSetReady(ctx, statefulSet); !ready {
+		ready := util.IsStatefulSetReady(ctx, statefulSet)
+		if !ready {
 			r.Log.Info("Waiting for etcd statefulset to become ready")
 			return nil
 		}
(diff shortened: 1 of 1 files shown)


enxebre:
do we know why etcd fails to reach quorum?

jparrill:
> do we know why etcd fails to reach quorum?

No, and I've tried to reproduce it many times :/. This only happens when there are a high load creating HC by perf team, this is the only hint we have.

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -2688,6 +2692,19 @@ func (r *HostedControlPlaneReconciler) reconcileManagedEtcd(ctx context.Context,
...
 
+	// Block to recover the ETCD cluster if it is not ready after etcdRecoveryPeriod
+	timePassed := time.Since(statefulSet.ObjectMeta.CreationTimestamp.Time)
+	ready := util.IsStatefulSetReady(ctx, statefulSet)
+	// If the etcd cluster is not ready after etcdRecoveryPeriod, we will attempt to recover it
+	// but we will stop trying it after etcdRecoveryTimeout to avoid loss of data in other scenarios.
+	if !ready && (timePassed > etcdRecoveryPeriod && timePassed <= etcdRecoveryTimeout) {
+		r.Log.Info("Recovering ETCD cluster")
+		// This will force the recreation of the ETCD pod
+		statefulSet.Spec.Template.Labels = map[string]string{"recovered": "true"}
```
Patryk-Stefanski, line 2703: Is it possible that the recreated etcd also fails? If so I think we should change the label to something else because it might be a bit misleading in a scenario where we attempt to recover but fail. Perhaps forcedRecreation: true

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -2688,6 +2692,19 @@ func (r *HostedControlPlaneReconciler) reconcileManagedEtcd(ctx context.Context,
 		}
 	}
 
+	// Block to recover the ETCD cluster if it is not ready after etcdRecoveryPeriod
+	timePassed := time.Since(statefulSet.ObjectMeta.CreationTimestamp.Time)
+	ready := util.IsStatefulSetReady(ctx, statefulSet)
+	// If the etcd cluster is not ready after etcdRecoveryPeriod, we will attempt to recover it
+	// but we will stop trying it after etcdRecoveryTimeout to avoid loss of data in other scenarios.
+	if !ready && (timePassed > etcdRecoveryPeriod && timePassed <= etcdRecoveryTimeout) {
```
enxebre, line 2700: @jparrill Is this possible:
- We have etcd quorum and HC is happy
- statefulSet goes not ready because one single replica failure. E.g. recycling management data plane
- The HC has live for less than etcdRecoveryTimeout
- We destroy etcd for a legit working cluster

I think this should be gated on the HC ever being available for the first time

enxebre: Following up here https://github.com/openshift/hypershift/pull/4419

PR #4359, opened 2024-07-12
HOSTEDCP-1820: Add snyk helper scripts to contrib

bryan-cox:
Adding some scripts I created to help me with searching for multiple CVE related programs in the RPMs.

Fixes [HOSTEDCP-1820](https://issues.redhat.com/browse/HOSTEDCP-1820)

**Checklist**

### DIFF
--- a/contrib/snyk/check-for-cve-in-cpo.sh
+++ b/contrib/snyk/check-for-cve-in-cpo.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+PWD=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
+
+IMAGE_TO_CHECK=$1
+SEARCH_TERMS=$2
+PULL_SECRETS_FILE=${PULL_SECRET:-$HOME/pull-secret.txt}
+GREP_PATTERN=$(echo "${SEARCH_TERMS}" | tr ',' '|')
+SHELL_CMD="rpm -qa | grep -Ei '${GREP_PATTERN}'"
+
+echo "Searching CPO image"
+HYPERSHIFT_IMAGE=$(oc adm release info "${IMAGE_TO_CHECK}" -a "${PULL_SECRETS_FILE}" --pullspecs | grep hypershift | awk '{print $2}')
+podman pull --tls-verify=false "${HYPERSHIFT_IMAGE}" --authfile "${PULL_SECRETS_FILE}" --quiet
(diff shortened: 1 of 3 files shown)


contrib/snyk/check-for-cve.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+PWD=/Users/user/.../hypershift/contrib/snyk
```
csrwng, line 3: You could just get the script directory
https://codefather.tech/blog/bash-get-script-directory/

contrib/snyk/check-for-cve.sh
```diff
@@ -0,0 +1,18 @@
...
+
+COMPONENT_TO_CHECK=$1
```
csrwng, line 5: Given that the component can only be HO or CPO, my pref would be to simply have 2 scripts that call the same code if necessary.

contrib/snyk/check-for-cve.sh
```diff
@@ -0,0 +1,18 @@
...
+IMAGE_TO_CHECK=$2
+PULL_SECRETS_FILE=$3
```
csrwng, line 7: I would default this to a path most folks could have and not necessarily expect it as an arg...
```bash
PULL_SECRETS_FILE=${PULL_SECRET:-$HOME/.pull-secret.txt}
```

contrib/snyk/search.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+rpm -qa | grep -i "wget\|cpython\|gnutls\|glib"
```
csrwng, line 2: The search string could be passed as a delimited string argument (like a , separated).

csrwng (commented):
Awesome stuff @bryan-cox 
Just some improvement suggestions, but that can definitely be done iteratively.

csrwng:
/lgtm
/hold

Feel free to remove the hold to merge as is.

contrib/snyk/check-for-cve-in-cpo.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+PWD=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
+
+IMAGE_TO_CHECK=$1
+SEARCH_TERMS=$2
+PULL_SECRETS_FILE=${PULL_SECRET:-$HOME/all-the-pull-secrets.json}
```
csrwng, line 7: It'd be good to use a consistent default in this script and the next

PR #4368, opened 2024-07-15
NO-JIRA: [release-4.16] test/e2e: remove api budget checks

csrwng:
presub e2e's a failing on this again and with no way to quickly diagnose the situation, we should not block PRs on this.

This isn't a great check as the number of mutating calls is somewhat outside our control with delays in cloud resource create/delete contirbuting to long test durations and more mutating calls.

We need a fundamentally better solution here based on KAS audit log analysis.

Manual backport of https://github.com/openshift/hypershift/pull/4286


sjenning:
Hit this here as well https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_hypershift/4385/pull-ci-openshift-hypershift-release-4.15-e2e-aws/1813949514063548416

/cherry-pick release-4.15

PR #4378, opened 2024-07-17
OCPBUGS-36897: fix(api): Nodepool CEL validation fix

a-dsouza:
**What this PR does / why we need it**: Fixes Nodepool CEL validation to allow neither replicas or autoScaling to be set

Fixes # https://issues.redhat.com/browse/OCPBUGS-36897

**Checklist**

### DIFF
--- a/api/hypershift/v1beta1/nodepool_types.go
+++ b/api/hypershift/v1beta1/nodepool_types.go
@@ -80,7 +80,7 @@ type NodePool struct {
 // NodePoolSpec is the desired behavior of a NodePool.
 // +kubebuilder:validation:XValidation:rule="!has(oldSelf.arch) || has(self.arch)", message="Arch is required once set"
 // +kubebuilder:validation:XValidation:rule="self.arch != 'arm64' || has(self.platform.aws)", message="Setting Arch to arm64 is only supported for AWS"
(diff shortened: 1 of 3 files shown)


hasueki (approved):
/lgtm

rtheis (approved):
/lgtm

muraee:
How about setting a default value of 0 on the `.replicas` field to mitigate the issue of old NodePools failing the validation, and preserve the previous behavior.

I don't think it's good to allow both `.replicas` or `.autoScaling` fields to not be set and rely on implicit defaulting.

enxebre:
>How about setting a default value of 0 on the .replicas field to mitigate the issue of old NodePools failing the validation, and preserve the previous behavior.
I don't think it's good to allow both .replicas or .autoScaling fields to not be set and rely on implicit defaulting.

In my mind 0 and nil here have different meanings. If I see replicas=0..N and autoscaler.spec set, I interpret a conflicting intent. So both unset make sense to me. Happy to hear other's thoughts

muraee:
> If I see replicas=0..N and autoscaler.spec set, I interpret a conflicting intent

agreed, this is prevented with the current CEL which I suggest we keep as is. If we add a default only for the replicas field, we satisfy the requirement for people wanting to leave the fields unset but make the behavior of that at least explicit on the API.
If the user decides to set spec.autoScaling later on, the current CEL will disallow it.

PR #4379, opened 2024-07-17
HOSTEDCP-1834: Fix Dependabot Issue By Updating Azure RHCOS Image Copy Process

bryan-cox:
The previous way of copying the RHCOS image from its blob to the blob in the HostedCluster's managed resource group needed to be updated since a previous dependency is no longer available. Specifically, github.com/tombuildsstuff/giovanni/storage/2019-12-12/blob/blobs was removed from GitHub. This process of copying the image also relied on external dependencies versus using Azure SDK functions to copy the RHCOS image. The new process uses only Azure SDK for Go functions to copy the RHCOS image.

Fixes [HOSTEDCP-1834](https://issues.redhat.com/browse/HOSTEDCP-1834)

**Checklist**


cmd/infra/azure/create.go
```diff
@@ -558,38 +555,46 @@ func createRhcosImages(ctx context.Context, l logr.Logger, o *CreateInfraOptions
...
 
-	blobClient := blobs.New()
-	blobClient.Authorizer = blobAuth
-	l.Info("Uploading rhcos image", "source", sourceURL)
-	input := blobs.CopyInput{
-		CopySource: sourceURL,
-		MetaData: map[string]string{
-			"source_uri": sourceURL,
-		},
+	imageBlobURL := "https://" + storageAccountName + ".blob.core.windows.net/" + "vhd" + "/"
```
csrwng, line 564: nit: you don't need to concat literals. It would also be cleaner with fmt.Sprintf

cmd/infra/azure/create.go
```diff
@@ -558,38 +555,46 @@ func createRhcosImages(ctx context.Context, l logr.Logger, o *CreateInfraOptions
 	if storageAccountKeyResult.Keys == nil || len(storageAccountKeyResult.Keys) == 0 || storageAccountKeyResult.Keys[0].Value == nil {
 		return "", errors.New("no storage account keys exist")
 	}
-	blobAuth, err := autorest.NewSharedKeyAuthorizer(storageAccountName, *storageAccountKeyResult.Keys[0].Value, autorest.SharedKey)
+
+	credential, err := container.NewSharedKeyCredential(storageAccountName, *storageAccountKeyResult.Keys[0].Value)
 	if err != nil {
-		return "", fmt.Errorf("failed to construct storage object authorizer: %w", err)
+		return "", err
```
csrwng, line 561: should we include a message with the error for consistency?

cmd/infra/azure/create.go
```diff
@@ -610,6 +615,53 @@ func createRhcosImages(ctx context.Context, l logr.Logger, o *CreateInfraOptions
...
+func copyImageAndWait(ctx context.Context, rhcosURL string, pageBlobClient *pageblob.Client) error {
+	_, err := pageBlobClient.CopyFromURL(ctx, rhcosURL, nil)
+	if err != nil {
+		return fmt.Errorf("failed to start the process to copy rhcos image to vhd blob: %w", err)
+	}
+
+	timeout := 10 * time.Minute
+	startTime := time.Now()
+
+	for {
```
csrwng, line 631: should this be done with https://github.com/openshift/hypershift/blob/main/vendor/k8s.io/apimachinery/pkg/util/wait/poll.go#L45

cmd/infra/azure/create.go
csrwng: also, should this variable be named `imageBlobURLPrefix` ?

csrwng (commented):
Some comments

PR #4381, opened 2024-07-17
OCPBUGS-36932: Add HTTP konnectivity proxy to OAuth server

csrwng:
The socks5 konnectivity proxy does not honor user-configured HTTP/S proxies when sending traffic through the dataplane. Most identity providers handled by the OAuth server do use HTTP/S for external communication and should honor any user configured proxy. This commit adds an additional container to the oauth server deployment with the HTTP konnectivity proxy. This proxy does honor the user-configured HTTP/S proxy for the HostedCluster when sending traffic through the data plane. The socks5 proxy is still used for the LDAP identity provider which does not use the proxy for its traffic. On the OAuth server, the HTTP_PROXY and HTTPS_PROXY environment variables point to the HTTP proxy while the ALL_PROXY variable still points to the socks5 proxy.

Fixes #[OCPBUGS-36932](https://issues.redhat.com/browse/OCPBUGS-36932)

**Checklist**


control-plane-operator/controllers/hostedcontrolplane/oauth/params.go
```diff
@@ -42,8 +43,8 @@ type OAuthServerParams struct {
 	LoginURLOverride        string
 	AvailabilityProberImage string `json:"availabilityProberImage"`
 	Availability            hyperv1.AvailabilityPolicy
-	Socks5ProxyImage        string
-	NoProxy                 []string
+	ProxyImage              string
```
enxebre, line 46: can we // document this new parameters?
should we group together all the proxy related params into their own struct?
that would also alleviate the reconcileDep signature which whereas is good to be explicit it's already a bit out of hand.

csrwng: Thx, updated.

PR #4382, opened 2024-07-17
OCPBUGS-37222: extract rhel9 MCO binaries for rhel8 based MCO images

sjenning:
Dynamically use rhel9 MCO binaries for rhel8 based MCO images

xref https://github.com/openshift/machine-config-operator/pull/4476

### DIFF
--- a/ignition-server/controllers/local_ignitionprovider.go
+++ b/ignition-server/controllers/local_ignitionprovider.go
@@ -1,6 +1,7 @@
 package controllers
 
 import (
+	"bytes"
 	"context"
 	"crypto/x509"
 	"crypto/x509/pkix"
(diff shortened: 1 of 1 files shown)


ignition-server/controllers/local_ignitionprovider.go
```diff
@@ -268,15 +269,56 @@ func (p *LocalIgnitionProvider) GetPayload(ctx context.Context, releaseImage str
 	err = func() error {
 		start := time.Now()
 		binaries := []string{"machine-config-operator", "machine-config-controller", "machine-config-server"}
+		suffix := ""
+
+		mcoOSReleaseBuf := &bytes.Buffer{}
```
enxebre, line 274: can we please follow up to unit test this?

sjenning: just to be clear, unit test the extraction of the RHEL major version from `/etc/os-release`?

PR #4386, opened 2024-07-18
NO-JIRA: Refactor Ensure... functions starting with EnsureNoCrashingPods

jparrill:
Refactored EnsureNoCrashingPods as a first step to use an standard interface for all Ensure functions

### DIFF
--- a/test/e2e/util/ensure.go
+++ b/test/e2e/util/ensure.go
@@ -1,0 +1,825 @@
+package util
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"os"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/google/go-cmp/cmp"
+	. "github.com/onsi/gomega"
+	routev1 "github.com/openshift/api/route/v1"
+	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
+	"github.com/openshift/hypershift/hypershift-operator/controllers/manifests"
+	suppconfig "github.com/openshift/hypershift/support/config"
+	"github.com/openshift/hypershift/support/util"
+	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	kapierror "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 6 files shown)


bryan-cox:
This looks like a large enough change that it should have its own Jira ticket IMO.

Is this related to some epic or feature about refactoring e2e's?

test/e2e/util/ensure.go
```diff
@@ -0,0 +1,825 @@
...
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	"k8s.io/utils/pointer"
+	capiv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	crclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type Ensure interface {
+	Ensure(t *testing.T, ctx context.Context, testParams *TestParams)
```
stevekuznetsov, line 34: This is an almost impossibly generic interface - what benefit are we gaining by abstracting this?

jparrill: If we need to add new variables it's matter off include them in the paramTest struct and map them in the proper tests at the beginning. It's more mechanical than difficult tbh.

jparrill:
> This looks like a large enough change that it should have its own Jira ticket IMO.
> 
> Is this related to some epic or feature about refactoring e2e's?

Yes, this is why I only put the first piece :). I don't think there is any Jira regarding this refactor but I think that the Ensure trigger function will help to make lighter the CI.

test/e2e/util/ensure.go
jparrill: I can compile them as you mention but I need to standardise them in the params terms and I don want to maintain them as separated code. I think it's valuable because 

- Delimites an standard for Ensure functions giving you an abstraction layer for the current functions and the future ones.
- Gives you the capability to trigger Ensure functions from CLI to test if you're doing ok on your E2E test.

It's really annoying to deal with E2E and fully delegate on CI to test a quick ensure function that you should be able to execute with just a small script.

This is totally IMHO, feel free @enxebre @csrwng @sjenning to comment you opinion on this :).

jparrill:
Looks like this will not be helpful at all for the team.
/close

PR #4388, opened 2024-07-19
NO-JIRA: Script to trigger Ensure functions from local

jparrill:
This script registers the EnsureXXX functions used usually in E2E tests and triggers it from local agains a Management or HostedCluster already created. The tipical EnsureXXX functions need to be moved to e2eutil.EnsureFuncs type in order to honor the interface contract. The PR https://github.com/openshift/hypershift/pull/4386 is the first step and the current PR depends on it.

- Depends on https://github.com/openshift/hypershift/pull/4386.

### DIFF
--- a/contrib/e2e-ensure-trigger/README.md
+++ b/contrib/e2e-ensure-trigger/README.md
@@ -1,0 +1,25 @@
+# E2E Ensure Function trigger
+
+This is a small script which takes care of the execution of a EnsureXXX function in the Management cluster you have up an running already. It's tedious sometimes to wait until the CI to finish the E2E test until you got feedback for just a quick execution of an function you're working on. This is why this script exists:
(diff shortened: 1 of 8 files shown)


bryan-cox:
This looks like a large enough change that it should have its own Jira ticket IMO.

Is this related to some epic or feature about refactoring e2e's?

jparrill:
> This looks like a large enough change that it should have its own Jira ticket IMO.
> 
> Is this related to some epic or feature about refactoring e2e's?

Not yet, was part of a PoC I've started on the Shift Week, so I think I can share with the rest of the team at some point and, if all agree and makes sense, create the epic a go with that.

jparrill:
Looks like this will not be helpful at all for the team.
/close

PR #4390, opened 2024-07-19
fix(build): update to the fixed clamav image

celebdor:
The upcoming OpenShift Pipelines which will be deployed shortly has stricter validations on Pipelines and Task manifests. Clamav would fail the new validations. This commit should make us ready.


celebdor:
closed in favor of #4392

PR #4396, opened 2024-07-19
OSASINFRA-3538: openstack: cluster CLI

EmilienM:

* Basic CLI options for cluster create & destroy. More options
  will come with Node pools and in the future with more usecases.
* Set a default Machine Network when creating the cluster with the CLI.
* Add the CLI options to e2e.

### DIFF
--- a/cmd/cluster/openstack/create.go
+++ b/cmd/cluster/openstack/create.go
@@ -1,0 +1,249 @@
+package openstack
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"gopkg.in/yaml.v3"
+
+	"github.com/openshift/hypershift/cmd/util"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
+	"github.com/openshift/hypershift/cmd/cluster/core"
+	corev1 "k8s.io/api/core/v1"
+)
+
+func DefaultOptions() *RawCreateOptions {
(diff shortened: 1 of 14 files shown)


cmd/cluster/openstack/create.go
```diff
@@ -0,0 +1,245 @@
...
+func NewCreateCommand(opts *core.RawCreateOptions) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:          "openstack",
+		Short:        "Creates basic functional HostedCluster resources on OpenStack platform",
+		SilenceUsage: true,
+	}
+
+	openstackOpts := DefaultOptions()
+	BindOptions(openstackOpts, cmd.Flags())
+	cmd.MarkPersistentFlagRequired("pull-secret")
```
bryan-cox, line 197: You may need to validate this flag is not empty in your validation function. We recently refactored how the CLI works and I just noticed this doesn't seem to be marking the flags as required - [OCPBUGS-37323](https://issues.redhat.com/browse/OCPBUGS-37323).

EmilienM:
sorry I had to fix unit tests and I missed something in the product CLI. This is ready for review now. Thanks!

bryan-cox:
Looks like there is a missing unit test for CreateCluster as well? AWS example - https://github.com/openshift/hypershift/blob/c40225a6155171291734f837f353fadf387bf463/cmd/cluster/aws/create_test.go#L123

cmd/cluster/openstack/create.go
EmilienM: nice suggestion, done. Thanks

EmilienM:
> Looks like there is a missing unit test for CreateCluster as well? AWS example -
> 
> https://github.com/openshift/hypershift/blob/c40225a6155171291734f837f353fadf387bf463/cmd/cluster/aws/create_test.go#L123

done!

PR #4405, opened 2024-07-22
HOSTEDCP-1798: Refactor multi-arch validation algorithm

bryan-cox:
Refactors the algorithm performing different multi-arch validations similar to [this](https://excalidraw.com/#json=QeumxPAwNmXPFxe4GjuWt,I_h08v5-INft4lOLYO5Mcg).

Fixes [HOSTEDCP-1778](https://issues.redhat.com/browse/HOSTEDCP-1778)

**Checklist**

### DIFF
--- a/cmd/cluster/aws/create.go
+++ b/cmd/cluster/aws/create.go
@@ -57,16 +57,11 @@ type ValidatedCreateOptions struct {
 }
 
 func (o *RawCreateOptions) Validate(ctx context.Context, opts *core.CreateOptions) (core.PlatformCompleter, error) {
-	// Validate if mgmt cluster and NodePool CPU arches don't match, a multi-arch release image or stream was used
-	// Exception for ppc64le arch since management cluster would be in x86 and node pools are going to be in ppc64le arch
-	if !o.MultiArch && !opts.Render && opts.Arch != hyperv1.ArchitecturePPC64LE {
-		mgmtClusterCPUArch, err := hyperutil.GetMgmtClusterCPUArch(ctx)
+	// Perform multi-arch validations except for render
+	if !opts.Render {
+		err := validateMultiArch(ctx, o, opts)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


bryan-cox:
I did not implement this block from the drawing as I don't recall why we originally needed it
```
 if multi-arch flag set 
        warn user multi-arch image not used,
        block hc creation
```

cmd/cluster/aws/create.go
```diff
@@ -504,3 +499,46 @@ func validateAWSOptions(ctx context.Context, opts *core.CreateOptions, awsOpts *
...
+		}
+
+		validMultiArchImage, err = registryclient.IsMultiArchManifestList(ctx, opts.ReleaseImage, pullSecret)
+		if err != nil {
+			return err
+		}
+	}
+
+	// Check if a release stream was provided instead and its multi-arch
+	if len(opts.ReleaseStream) > 0 && !strings.Contains(opts.ReleaseStream, "multi") {
```
muraee, line 520:
```suggestion
	if len(opts.ReleaseStream) > 0 && strings.Contains(opts.ReleaseStream, "multi") {
```
shouldn't this be a positive check or did I misunderstand it?

cmd/cluster/aws/create.go
```diff
@@ -504,3 +499,46 @@ func validateAWSOptions(ctx context.Context, opts *core.CreateOptions, awsOpts *
 
 	return nil
 }
+
+func validateMultiArch(ctx context.Context, o *RawCreateOptions, opts *core.CreateOptions) error {
+	var validMultiArchImage bool
```
muraee, line 504:
```suggestion
	validMultiArchImage := false
```
nit: make it more explicit

cmd/cluster/aws/create.go
bryan-cox: Yes 🤦🏻

PR #4406, opened 2024-07-23
HOSTEDCP-1791: dont render cloud-credentials secret

Patryk-Stefanski:

Fixes #

**Checklist**

### DIFF
--- a/cmd/cluster/core/create.go
+++ b/cmd/cluster/core/create.go
@@ -70,6 +70,7 @@ func bindCoreOptions(opts *RawCreateOptions, flags *pflag.FlagSet) {
 	flags.StringVar(&opts.ControlPlaneAvailabilityPolicy, "control-plane-availability-policy", opts.ControlPlaneAvailabilityPolicy, "Availability policy for hosted cluster components. Supported options: SingleReplica, HighlyAvailable")
 	flags.BoolVar(&opts.Render, "render", opts.Render, "Render output as YAML to stdout instead of applying")
 	flags.StringVar(&opts.RenderInto, "render-into", opts.RenderInto, "Render output as YAML into this file instead of applying. If unset, YAML will be output to stdout.")
+	flags.BoolVar(&opts.RenderSensitive, "render-sensitive", opts.RenderSensitive, "enables rendering of sensitive information in the output")
 	flags.StringVar(&opts.SSHKeyFile, "ssh-key", opts.SSHKeyFile, "Path to an SSH key file")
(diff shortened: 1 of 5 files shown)


test/e2e/util/fixture.go
```diff
@@ -183,6 +183,7 @@ func renderCreate(ctx context.Context, opts *core.RawCreateOptions, platformOpts
 
 	opts.Render = true
 	opts.RenderInto = outputFile
+	opts.RenderSensitive = false
```
fxierh, line 186: Will this get overridden in core.CreateCluster below ?

Patryk-Stefanski: It was by the complete function but it should anymore.

cmd/cluster/core/create.go
```diff
@@ -70,6 +70,7 @@ func bindCoreOptions(opts *RawCreateOptions, flags *pflag.FlagSet) {
 	flags.StringVar(&opts.ControlPlaneAvailabilityPolicy, "control-plane-availability-policy", opts.ControlPlaneAvailabilityPolicy, "Availability policy for hosted cluster components. Supported options: SingleReplica, HighlyAvailable")
 	flags.BoolVar(&opts.Render, "render", opts.Render, "Render output as YAML to stdout instead of applying")
 	flags.StringVar(&opts.RenderInto, "render-into", opts.RenderInto, "Render output as YAML into this file instead of applying. If unset, YAML will be output to stdout.")
+	flags.BoolVar(&opts.RenderSensitive, "render-sensitive", opts.RenderSensitive, "Render sensitive information in the output")
```
fxierh, line 73: It looks like RenderSensitive defaults to false now. Any chance of defaulting this to true so it won't break existing jobs which make use of the create - patch - apply logic ? 

Maybe we could negate the flag by making it `--no-render-sensitive`.

Patryk-Stefanski: Where does it default to false, from my understanding the is no default value and we only set it to false for azure e2e.

fxierh: It is a bool so its value will be false unless we set it to true

enxebre:
Please default to not rendering sensitive info once the relevant jobs have been updated.
fwiw this won't be an issue when moving to msi
/approve

bryan-cox:
/retest

for konflux

stevekuznetsov:
>  Please default to not rendering sensitive info ...

@enxebre wouldn't this break every consumer of `--render` that uses the output as a self-contained mechanism to install `HostedClusters` on a management plane?

PR #4408, opened 2024-07-23
HOSTEDCP-1813: Moves MachineIdentity from the HostedCluster API to the NodePool API

bryan-cox:
Moves `MachineIdentity` from the Azure section in the HostedCluster API to the Azure section in the NodePool API

Fixes [HOSTEDCP-1813](https://issues.redhat.com/browse/HOSTEDCP-1813)

**Checklist**

### DIFF
--- a/api/hypershift/v1alpha1/hostedcluster_types.go
+++ b/api/hypershift/v1alpha1/hostedcluster_types.go
@@ -1573,7 +1573,6 @@ type AzurePlatformSpec struct {
 	VnetID            string `json:"vnetID"`
 	SubnetID          string `json:"subnetID"`
 	SubscriptionID    string `json:"subscriptionID"`
-	MachineIdentityID string `json:"machineIdentityID"`
 	SecurityGroupID   string `json:"securityGroupID"`
 }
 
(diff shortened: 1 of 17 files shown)


api/hypershift/v1beta1/nodepool_types.go
```diff
@@ -964,6 +964,11 @@ type AzureNodePoolPlatform struct {
 	// If not specified then Boot diagnostics will be disabled.
 	// +optional
 	Diagnostics *Diagnostics `json:"diagnostics,omitempty"`
+
+	// MachineIdentityID is used as the user-assigned identity to be assigned to the VMs
+	//
+	// +optional
+	MachineIdentityID string `json:"machineIdentityID,omitempty"`
```
miguelsorianod, line 971: Could we document what is the behavior if `MachineIdentityID` is not provided?

Could we document if the resource associated to the ID needs to preexist beforehand in Azure?

Thanks

api/hypershift/v1alpha1/hostedcluster_types.go
```diff
@@ -1573,7 +1573,6 @@ type AzurePlatformSpec struct {
 	VnetID            string `json:"vnetID"`
 	SubnetID          string `json:"subnetID"`
 	SubscriptionID    string `json:"subscriptionID"`
-	MachineIdentityID string `json:"machineIdentityID"`
```
miguelsorianod, line 1576: So, machine identity is not needed at all during HostedCluster creation?

api/hypershift/v1beta1/nodepool_types.go
bryan-cox: Yeah I think we have discussed what happens if it is not provided a few times. 

I will add it explicitly here though, but this field could change as workload identity/MSI work progresses.

api/hypershift/v1alpha1/hostedcluster_types.go
bryan-cox: No

api/hypershift/v1beta1/nodepool_types.go
miguelsorianod: > Yeah I think we have discussed what happens if it is not provided a few times.

Yes, I am aware :+1: . It is more for consumers of the API and other people that might not be aware of it.

PR #4409, opened 2024-07-23
OSASINFRA-3312: Implements OpenStack Node pools

EmilienM:

This PR adds OpenStack node pools support.
It create a basic node pool with a flavor and image name (for now).
CLI options are updated.


EmilienM:
Note about e2e-openstack, it's going to fail for sure as the job is still not fully working yet but we're actively working on it.

EmilienM:
The openstack job failed because the job isn't ready yet and the IBM cloud jobs fail for a reason not related to this PR.

PR #4413, opened 2024-07-24
NO-JIRA: [release-4.15] test/e2e: remove api budget checks 

davidvossel:
presub e2e's a failing on this again and with no way to quickly diagnose the situation, we should not block PRs on this.

This isn't a great check as the number of mutating calls is somewhat outside our control with delays in cloud resource create/delete contirbuting to long test durations and more mutating calls.

We need a fundamentally better solution here based on KAS audit log analysis.

Manual backport of https://github.com/openshift/hypershift/pull/4368


davidvossel:
/override "Red Hat Konflux / hypershift-release-mce-25-on-pull-request"

PR #4415, opened 2024-07-24
OSASINFRA-3538: openstack: small fixes for CLI

EmilienM:

Small fixes for OpenStack CLI.


EmilienM:
the azure CI seems unstable but this isn't related to that PR. This can be safely merged I think.

PR #4417, opened 2024-07-25
HOSTEDCP-1815: feat(ignition): Add option to disable ignition server reconciliation

a-dsouza:
**What this PR does / why we need it**: Adds support to disable ignition server reconcilement 

Fixes # https://issues.redhat.com/browse/HOSTEDCP-1815

**Checklist**

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
@@ -1079,27 +1079,31 @@ func (r *HostedControlPlaneReconciler) reconcile(ctx context.Context, hostedCont
 		}
 	}
 
-	r.Log.Info("Reconciling ignition server")
-	if err := ignitionserver.ReconcileIgnitionServer(ctx,
-		r.Client,
-		createOrUpdate,
-		releaseImageProvider.Version(),
-		releaseImageProvider.GetImage(util.CPOImageName),
-		releaseImageProvider.ComponentImages(),
-		hostedControlPlane,
-		r.DefaultIngressDomain,
-		// The healthz handler was added before the CPO started to manage the ignition server, and it's the same binary,
-		// so we know it always exists here.
-		true,
-		r.ReleaseProvider.GetRegistryOverrides(),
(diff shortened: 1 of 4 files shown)


hasueki (commented):
/lgtm

rtheis (approved):
/lgtm

enxebre:
fwiw there's also reconcileMachineConfigServerConfig and reconcileCoreIgnitionConfig you might want to skip.
We might want to group them together under a self contained function that returns early if the annotation is set and that can be unit tested

a-dsouza:
> fwiw there's also reconcileMachineConfigServerConfig and reconcileCoreIgnitionConfig you might want to skip.
We might want to group them together under a self contained function that returns early if the annotation is set and that can be unit tested

@enxebre  Yup, I can take a stab at that. If the MCS config is only used by ignition, I'm assuming we can also skip the creation of the `MachineConfigServerCert`, right? 

Are we aware of any other deployments, config maps, certs, etc. solely utilized by the ignition server?

rtheis (approved):
/lgtm

hasueki (approved):
/lgtm

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -2525,12 +2527,14 @@ func (r *HostedControlPlaneReconciler) reconcilePKI(ctx context.Context, hcp *hy
...
 	}
 
-	if hcp.Spec.Platform.Type != hyperv1.IBMCloudPlatform {
-		ignitionServerCert := manifests.IgnitionServerCertSecret(hcp.Namespace)
-		if _, err := createOrUpdate(ctx, r, ignitionServerCert, func() error {
-			return pki.ReconcileIgnitionServerCertSecret(ignitionServerCert, rootCASecret, p.OwnerRef)
-		}); err != nil {
-			return fmt.Errorf("failed to reconcile ignition server cert: %w", err)
+	if _, exists := hcp.Annotations[hyperv1.DisableIgnitionServerAnnotation]; !exists {
+		if hcp.Spec.Platform.Type != hyperv1.IBMCloudPlatform {
```
csrwng, line 2531: Just a note... unrelated to this PR, but should fix. We should use the disable PKI annotation here instead of checking for the IBM Cloud platform.

PR #4423, opened 2024-07-26
OSASINFRA-3550: openstack: configure nodepool ports for Ingress

EmilienM:

* Add `CIDR` to `SubnetSpec` so when a user asks for CAPO to manage a subnet, they can provide the CIDR (they couldn't before, they could only provide the DNS nameserver and the DHCP allocation pool)
* Configure the OpenStackMachineTemplate with ports that allow Ingress traffic via allowed address pairs, where the IP of Ingress will be the 7th available IP in the subnet, like it's done in the Installer already.
* Add a few validations when managed subnets are in use

### DIFF
--- a/hypershift-operator/controllers/hostedcluster/internal/platform/openstack/openstack.go
+++ b/hypershift-operator/controllers/hostedcluster/internal/platform/openstack/openstack.go
@@ -3,6 +3,7 @@ package openstack
 import (
 	"context"
 	"fmt"
+	"net"
 	"os"
 
 	corev1 "k8s.io/api/core/v1"
@@ -22,11 +23,12 @@ import (
 	capiv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	"github.com/apparentlymart/go-cidr/cidr"
 	"github.com/openshift/hypershift/control-plane-operator/controllers/hostedcontrolplane/cloud/openstack"
+	openstackutil "github.com/openshift/hypershift/support/openstackutil"
 )
 
 const (
-	defaultCIDRBlock  = "10.0.0.0/16"
 	sgRuleDescription = "Managed by the Hypershift Control Plane Operator"
 )
 
@@ -63,7 +65,7 @@ func (a OpenStack) ReconcileCAPIInfraCR(ctx context.Context, client client.Clien
 	return openStackCluster, nil
 }
 
-func reconcileOpenStackClusterSpec(hcluster *hyperv1.HostedCluster, openStackClusterSpec *capo.OpenStackClusterSpec, apiEndpoint hyperv1.APIEndpoint) {
+func reconcileOpenStackClusterSpec(hcluster *hyperv1.HostedCluster, openStackClusterSpec *capo.OpenStackClusterSpec, apiEndpoint hyperv1.APIEndpoint) error {
 	openStackPlatform := hcluster.Spec.Platform.OpenStack
 
 	openStackClusterSpec.ControlPlaneEndpoint = &capiv1.APIEndpoint{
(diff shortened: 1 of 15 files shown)


hypershift-operator/controllers/hostedcluster/internal/platform/openstack/openstack.go
```diff
@@ -75,7 +76,12 @@ func reconcileOpenStackClusterSpec(hcluster *hyperv1.HostedCluster, openStackClu
 	// If no MachineNetwork is provided, use a default CIDR block.
 	// Note: The default is required for now because there is no CLI option to set the MachineNetwork.
 	// See https://github.com/openshift/hypershift/pull/4287
+	// When subnet is provided, machine network is required.
+	// Note: currently there is no validation that the provided subnets are within the machine network, it's up to the user to ensure that for now.
 	if hcluster.Spec.Networking.MachineNetwork == nil || len(hcluster.Spec.Networking.MachineNetwork) == 0 {
+		if openStackPlatform.ManagedSubnets != nil && len(openStackPlatform.ManagedSubnets) > 0 {
+			return fmt.Errorf("failed to reconcile OpenStack CAPI cluster, machine network is required when subnets are provided")
```
MaysaMacedo, line 83: Is the `MachineNetwork` now required because we need the CIDR in the `getIngressIP`?
If that is the case, I don't think we need to add a default CIDR, in the next line, anymore right?

hypershift-operator/controllers/hostedcluster/internal/platform/openstack/openstack.go
```diff
@@ -75,7 +76,12 @@ func reconcileOpenStackClusterSpec(hcluster *hyperv1.HostedCluster, openStackClu
 	// If no MachineNetwork is provided, use a default CIDR block.
 	// Note: The default is required for now because there is no CLI option to set the MachineNetwork.
 	// See https://github.com/openshift/hypershift/pull/4287
+	// When subnet is provided, machine network is required.
```
MaysaMacedo, line 79: You mean when CAPO will create the subnet, the machineNetwork is required, right?

hypershift-operator/controllers/nodepool/openstack/openstack.go
```diff
@@ -27,5 +28,29 @@ func MachineTemplateSpec(hcluster *hyperv1.HostedCluster, nodePool *hyperv1.Node
 		return nil, fmt.Errorf("image name is required")
 	}
 
+	port := capiopenstack.PortOpts{}
+	ingressIP, err := getIngressIP(hcluster.Spec.Networking.MachineNetwork[0])
```
MaysaMacedo, line 32: should this always be required then? regardless if BYON or CAPO managed subnet will be used

hypershift-operator/controllers/hostedcluster/internal/platform/openstack/openstack.go
EmilienM: this validation returns an error if MachineNetwork is not provided via CLI or YAML and if the user provides ManagedSubnets or Subnets are pre-created (I actually forgot a condition on the later, I'll add it now).
Does it make sense?

hypershift-operator/controllers/nodepool/openstack/openstack.go
MaysaMacedo: even if we have the default CIDR specified [here](https://github.com/openshift/hypershift/pull/4423/files#diff-0de67b02a8275f7a8227b3af5101786e70e5c651b852a87f6160f6563a9071d6R85) when no subnets are defined, we would get the machineNetwork empty here, right?

EmilienM: The line removal of `machineNetworks = []hyperv1.MachineNetworkEntry{{CIDR: *ipnet.MustParseCIDR(defaultCIDRBlock)}}` was an error, nice catch.

EmilienM: fixed, thanks

MaysaMacedo (commented):
/lgtm

EmilienM:
/test e2e-openstack
to make sure we don't regress

EmilienM:
/close
We found a way to not needing this PR, and use OpenStack Octavia for Ingress.

However we will probably re-open it or re-use some of it in the future when we'll add Ingress for HCP on OSP GA.

cc @MaysaMacedo

PR #4439, opened 2024-07-29
OSASINFRA-3550: docs/openstack: adds content for Ingress with MetalLB

EmilienM:

Adds content for Ingress on the OpenStack platform, with MetalLB.

### DIFF
--- a/docs/content/how-to/openstack/create-openstack-cluster.md
+++ b/docs/content/how-to/openstack/create-openstack-cluster.md
@@ -5,7 +5,8 @@ Install an OCP cluster running on VMs within a management OCP cluster
 ## Limitations
 
 * The HyperShift Operator with OpenStack support is currently in development and is not intended for production use.
-* Ingress and OpenStack CSI (Cinder and Manila) are not functional.
+* Enabling Ingress is a day 2 operation for now, using MetalLB.
+* OpenStack CSI (Cinder and Manila) are not functional.
 
 ## Prerequisites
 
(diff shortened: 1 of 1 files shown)


docs/content/how-to/openstack/create-openstack-cluster.md
```diff
@@ -192,6 +193,129 @@ NAME      VERSION       AVAILABLE   PROGRESSING   SINCE   STATUS
 version   4.17.0        True        False         5m39s   Cluster version is 4.17.0
 ```
 
+## Enabling Ingress
+
+Note: In 4.17, enabling Ingress is a day 2 operation and we tested it with MetalLB. This might change in the future.
+
+* Create a floating IP in your OpenStack cloud that will be the public IP for Ingress.
```
MaysaMacedo, line 200: The Floating IP needs to be routed to the machine Network, right?
It might be worth to add that as a requirement.

docs/content/how-to/openstack/create-openstack-cluster.md
```diff
@@ -192,6 +193,129 @@ NAME      VERSION       AVAILABLE   PROGRESSING   SINCE   STATUS
...
+* Update your DNS to create the wildcard for *apps.<cluster-name>.<base-domain> .
+* Create the OpenStack Neutron port for Ingress:
+
+```shell
+openstack port create --network k8s-clusterapi-cluster-<cluster-namespace>-<cluster-name>-<cluster-id> --fixed-ip ip-address=<internal-ingress-IP> hcp-ingress
+```
+
+Note: if CAPO handles the network resources, the internal ingress IP should be `10.0.0.7`.
```
MaysaMacedo, line 208: do we require this address because we're enforcing usage of one CIDR for the machines? is there a chance this won't be always true even when capo handles creation of resources?

docs/content/how-to/openstack/create-openstack-cluster.md
EmilienM: right but to me that goes without saying, since CAPO would connect the network to the external network via a router. Please suggest an addition if you feel like I missed something.

docs/content/how-to/openstack/create-openstack-cluster.md
EmilienM: in the current code as you know we use the `10.0.0.0/16` subnet if not provided by the user so I think it's mention here a valid IP that can be used for Ingress. I could maybe clarify that?

docs/content/how-to/openstack/create-openstack-cluster.md
MaysaMacedo:
```suggestion
* Create a floating IP using the external network configured on your cluster that will be the public IP for Ingress.
```

docs/content/how-to/openstack/create-openstack-cluster.md
MaysaMacedo: yes, I think it's a good idea to clarify it

davidvossel (commented):
/approve

add lgtm once you all are good with this

docs/content/how-to/openstack/create-openstack-cluster.md
EmilienM: I clarified it. Let me know what you think.

EmilienM:
Note about this doc, we'll still need to have https://github.com/openshift/hypershift/pull/4423 finished to have Ingress working but this will be sorted out. I think this can be merged anyway.

MaysaMacedo:
/lgtm
/hold

Feel free to cancel the hold if you think we don't need to wait for the other PR

PR #4445, opened 2024-07-30
OCPBUGS-37703: fix release stream flag/rhcos image for create cluste…

Patryk-Stefanski:
**What this PR does / why we need it**:  individually looks up OCP version based on the release stream so that RHCOS image isn't empty when creating an azure cluster.

Fixes # OCPBUGS-37703](https://issues.redhat.com/browse/OCPBUGS-37703)

**Checklist**

### DIFF
--- a/cmd/cluster/azure/create.go
+++ b/cmd/cluster/azure/create.go
@@ -3,6 +3,7 @@ package azure
 import (
 	"context"
 	"fmt"
+	"github.com/openshift/hypershift/cmd/version"
 	"net/url"
 	"os"
 	"strings"
@@ -349,7 +350,7 @@ func CreateInfraOptions(ctx context.Context, azureOpts *ValidatedCreateOptions, 
 	rhcosImage := azureOpts.RHCOSImage
 	if rhcosImage == "" {
 		var err error
-		rhcosImage, err = lookupRHCOSImage(ctx, opts.Arch, opts.ReleaseImage, opts.PullSecretFile)
+		rhcosImage, err = lookupRHCOSImage(ctx, opts.Arch, opts.ReleaseImage, opts.ReleaseStream, opts.PullSecretFile)
 		if err != nil {
 			return azureinfra.CreateInfraOptions{}, fmt.Errorf("failed to retrieve RHCOS image: %w", err)
 		}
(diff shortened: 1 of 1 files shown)


cmd/cluster/azure/create.go
```diff
@@ -371,7 +372,15 @@ func CreateInfraOptions(ctx context.Context, azureOpts *ValidatedCreateOptions,
 }
 
 // lookupRHCOSImage looks up a release image and extracts the RHCOS VHD image based on the nodepool arch
-func lookupRHCOSImage(ctx context.Context, arch string, image string, pullSecretFile string) (string, error) {
+func lookupRHCOSImage(ctx context.Context, arch, image, releaseStream, pullSecretFile string) (string, error) {
```
jparrill, line 375: It could be possible to cover the function with a unit test?

Patryk-Stefanski: This fix is blocking our azure e2e so I would prefer to add a unit test in a separate pr later on.

cmd/cluster/azure/create.go
```diff
@@ -371,7 +372,15 @@ func CreateInfraOptions(ctx context.Context, azureOpts *ValidatedCreateOptions,
...
+	if len(image) == 0 && len(releaseStream) != 0 {
+		defaultVersion, err := version.LookupDefaultOCPVersion(releaseStream)
+		if err != nil {
+			return "", fmt.Errorf("release image is required when unable to lookup default OCP version: %w", err)
```
bryan-cox, line 379: Should this say either a release image or a release stream should be provided?

bryan-cox, line 379:
```suggestion
			return "", fmt.Errorf("failed to lookup OCP image for release stream, %s: %w", releaseStream, err)
```

cmd/cluster/azure/create.go
sjenning: This is done in the core create
https://github.com/openshift/hypershift/blob/cbf39ff40a564fc4a99dfd7b0c4e1b335538d39d/cmd/cluster/core/create.go#L210-L216

It seems that Azure needs this at the infra create stage, which is why we need to do it here?

sjenning:
/test unit

`TestValidateKeyPairItempotency` flake

PR #4446, opened 2024-07-30
OSASINFRA-3487: e2e/openstack: skip cluster operator validations

EmilienM:

Ingress isn't yet implemented for HCP on OpenStack platform so the
console operator won't be Ready from CVO perspective, which makes an e2e
validation failing.
We skip it for now and we'll add it later.

### DIFF
--- a/docs/content/how-to/openstack/create-openstack-cluster.md
+++ b/docs/content/how-to/openstack/create-openstack-cluster.md
@@ -163,6 +163,9 @@ NAME            VERSION   KUBECONFIG                       PROGRESS   AVAILABLE 
 example         4.17.0    example-admin-kubeconfig         Completed  True        False         The hosted control plane is available
 ```
 
+Because Ingress is not yet supported on this platform, the Console operator won't report Ready.
+Once the user deploys Ingress on day 2 (e.g. via Metal LB), the console becomes available.
+
 ## Accessing the HostedCluster
 
 CLI access to the guest cluster is gained by retrieving the guest cluster's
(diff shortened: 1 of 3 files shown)


test/e2e/util/util.go
```diff
@@ -1312,7 +1312,11 @@ func ValidatePrivateCluster(t *testing.T, ctx context.Context, client crclient.C
 		g.Expect(hostedCluster.Status.ControlPlaneEndpoint.Host).ToNot(ContainSubstring("hypershift.local"))
 	}
 
-	validateHostedClusterConditions(t, ctx, client, hostedCluster, numNodes > 0, 10*time.Minute)
+	// On OpenStack, the console operator is not Ready because Ingress is not available yet.
+	// This makes this validation fail, so we skip it for now.
+	if hostedCluster.Spec.Platform.Type != hyperv1.OpenStackPlatform {
+		validateHostedClusterConditions(t, ctx, client, hostedCluster, numNodes > 0, 10*time.Minute)
```
EmilienM, line 1318: Note for myself: because we were waiting for 10 minutes here, I wonder if the validations after this one might fail because they're run too soon.

stevekuznetsov: `validateHostedClusterConditions` builds the set of conditions it needs to validate based on the `hostedCluster` - please amend the logic there, not here.

stevekuznetsov: eg https://github.com/openshift/hypershift/blob/c65722a26b5ace2b92eab2e41f5e2b18163b660f/support/conditions/conditions.go#L35-L72

EmilienM:
I'll probably close this PR because I'm trying to make Ingress working in our CI with MetalLB.
Thanks for the review though, @stevekuznetsov .

test/e2e/util/util.go
EmilienM: thanks, I'll take a look here.

EmilienM:
note for myself, I need to add destroy to e2e for platform openstack (in `e2e/util/fixture.go`)

EmilienM:
I think I'm going to give up this idea, this doesn't seem to work and anyway I didn't like from the beginning the idea of not validating CVO. On day 2 we can add Ingress with Metal LB and CVO will be happy. Let's close it for now.

/close

PR #4449, opened 2024-07-31
HOSTEDCP-1669: Make hostedcluster.spec.services immutable for non IBM platforms

enxebre:
There's many non supported input/day 2 changes we need to validate on the API for better UX. This is one of them. Any upcoming change needs to include a validation to the introducingwq check

Fixes #

**Checklist**

### DIFF
--- a/api/hypershift/v1beta1/hostedcluster_types.go
+++ b/api/hypershift/v1beta1/hostedcluster_types.go
@@ -321,6 +321,7 @@ const (
 )
 
 // HostedClusterSpec is the desired behavior of a HostedCluster.
+// +kubebuilder:validation:XValidation:rule="self.platform.type != 'IBMCloud' ? self.services == oldSelf.services : true", message="Services is immutable. Changes might result in unpredictable and disruptive behavior."
 type HostedClusterSpec struct {
 	// Release specifies the desired OCP release payload for the hosted cluster.
 	//
--- a/cmd/install/assets/hypershift-operator/hypershift.openshift.io_hostedclusters.yaml
+++ b/cmd/install/assets/hypershift-operator/hypershift.openshift.io_hostedclusters.yaml
@@ -9679,6 +9679,11 @@ spec:
             - services
             - sshKey
             type: object
+            x-kubernetes-validations:
+            - message: Services is immutable. Changes might result in unpredictable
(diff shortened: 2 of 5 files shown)


api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -426,6 +426,7 @@ type HostedClusterSpec struct {
 	//
 	// If a given service is not present in this list, it will be exposed publicly
 	// by default.
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf", message="Services is immutable"
```
zanetworker, line 429: I know its far cry, but if we can add some reasoning on why this is a good thing in the message, it would be great. 

"Serivices are immutable to/because ...."

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -426,6 +426,7 @@ type HostedClusterSpec struct {
 	//
 	// If a given service is not present in this list, it will be exposed publicly
 	// by default.
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf", message="Services is immutable. Changes might result in unpredictable and disruptive behavior."
```
bryan-cox, line 429: Should we go ahead and explicitly state whether this is required or not? ([HOSTEDCP-1825](https://issues.redhat.com/browse/HOSTEDCP-1825))

hasueki (changes requested):
/hold IBM Cloud alters these values after initial creation

rtheis (commented):
/hold Can we please discuss this and any other planned API changes in our interlock meeting tomorrow?

api/hypershift/v1beta1/hostedcluster_types.go
enxebre: every field is required unless specified otherwise. If we want to signal that explicitly consistently (which I think is a good thing), let's put a dedicated separate PR for it.

enxebre:
/retitle HOSTEDCP-1669: Make hostedcluster.spec.services immutable for non IBM platforms

rtheis (approved):
/lgtm

hasueki (approved):
/lgtm

@enxebre Thank you!

PR #4450, opened 2024-07-31
HOSTEDCP-1801: [CPO Refactor] Introduced an abstraction for CP Components 

muraee:
- introduces an abstraction that all control plane component should implement, allowing us to enforce opinionated defaults and settings for all components in one place
- streamline the development process by reducing some of the boilerplate code that needs to be copy/pasted when introducing new components.
- All Components reconciliation logic using the abstraction should be contained within the component pkg, reducing LOC of hostedcontrolplane_controller.go
- allow us to provide a default implementation for some common reconciliation logic such as RBAC to be reused by all components instead of reimplementing the same thing.

---
2 components refactored in this PR as an example:
- cluster-autoscaler
- openshift-router-controller-manger

Fixes #

**Checklist**


control-plane-operator/controllers/hostedcontrolplane/routecm/reconcile.go
```diff
@@ -0,0 +1,239 @@
...
+}
+
+// Name implements controlplanecomponent.DeploymentReconciler.
+func (r *RouteControllerManagerReconciler) Name() string {
+	return ComponentName
+}
+
+// ReconcileDeployment implements controlplanecomponent.DeploymentReconciler.
+func (r *RouteControllerManagerReconciler) ReconcileDeployment(cpContext component.ControlPlaneContext, deployment *appsv1.Deployment) error {
+	image := cpContext.ReleaseImageProvider.GetImage("route-controller-manager")
```
csrwng, line 70: For building the deployment reconciler, maybe we can use a builder pattern similar to what you have with the generic reconciler.

control-plane-operator/controllers/hostedcontrolplane/routecm/reconcile.go
```diff
@@ -0,0 +1,239 @@
...
+		return fmt.Errorf("failed to get openshift controller manager config: %w", err)
+	}
+
+	configBytes, ok := config.Data[configKey]
+	if !ok {
+		return fmt.Errorf("openshift controller manager configuration is not expected to be empty")
+	}
+	configHash := util.ComputeHash(configBytes)
+
+	maxSurge := intstr.FromInt(0)
```
csrwng, line 88: I don't see the need to set this in every reconcile function. It should be done generically. If we have different kinds of deployment strategy, we could maybe just specify which kind we need.

control-plane-operator/controllers/hostedcontrolplane/routecm/reconcile.go
```diff
@@ -0,0 +1,239 @@
...
+	maxUnavailable := intstr.FromInt(1)
+	deployment.Spec.Strategy.Type = appsv1.RollingUpdateDeploymentStrategyType
+	deployment.Spec.Strategy.RollingUpdate = &appsv1.RollingUpdateDeployment{
+		MaxSurge:       &maxSurge,
+		MaxUnavailable: &maxUnavailable,
+	}
+	if deployment.Spec.Selector == nil {
+		deployment.Spec.Selector = &metav1.LabelSelector{
+			MatchLabels: openShiftRouteControllerManagerLabels(),
```
csrwng, line 97: could we consolidate these with MultizoneSpreadLabels and not have to set them here?

control-plane-operator/controllers/hostedcontrolplane/routecm/reconcile.go
```diff
@@ -0,0 +1,239 @@
...
+		}
+	}
+	deployment.Spec.Template.ObjectMeta.Labels = openShiftRouteControllerManagerLabels()
+	deployment.Spec.Template.ObjectMeta.Annotations = map[string]string{
+		configHashAnnotation: configHash,
```
csrwng, line 102: It'd be nice if we could generalize the concept of resources that when changed should cause a rollout of this deployment. Like adding a DependsOn(resource1, resource2, ....) to the builder and then just hash all of their content in a generic hash annotation.

control-plane-operator/controllers/hostedcontrolplane/routecm/reconcile.go
```diff
@@ -0,0 +1,239 @@
...
+	}
+	deployment.Spec.Template.Spec.AutomountServiceAccountToken = ptr.To(false)
```
csrwng, line 104: We handle this generically via the RequiresMmgmtKAS setting

control-plane-operator/controllers/hostedcontrolplane/routecm/reconcile.go
```diff
@@ -0,0 +1,239 @@
...
+	deployment.Spec.Template.Spec.Containers = []corev1.Container{
+		util.BuildContainer(routeOCMContainerMain(), buildRouteOCMContainerMain(image)),
+	}
+
+	return nil
+}
+
+var (
+	volumes = component.Volumes{
+		routeOCMVolumeConfig().Name: component.Volume{
```
csrwng, line 114: given that we have a new struct for managing volumes, I don't think it makes sense to have functions that return volumes of which we only use the name. Maybe we can change those to const?

control-plane-operator/controllers/hostedcontrolplane/routecm/reconcile.go
```diff
@@ -0,0 +1,239 @@
...
+	deployment.Spec.Template.Spec.AutomountServiceAccountToken = ptr.To(false)
+	deployment.Spec.Template.Spec.Containers = []corev1.Container{
+		util.BuildContainer(routeOCMContainerMain(), buildRouteOCMContainerMain(image)),
+	}
+
+	return nil
+}
+
+var (
+	volumes = component.Volumes{
```
csrwng, line 113: If we can figure out a way to not have these be package level vars, that'd be awesome. For example, this can be inlined in the Volumes function below, and then when containers need to be built/reconciled, we automatically pass the volumes field to them.

control-plane-operator/controllers/hostedcontrolplane/routecm/reconcile.go
```diff
@@ -0,0 +1,239 @@
...
+}
+
+func routeOCMContainerMain() *corev1.Container {
+	return &corev1.Container{
+		Name: "route-controller-manager",
+	}
+}
+
+func buildRouteOCMContainerMain(image string) func(*corev1.Container) {
+	return func(c *corev1.Container) {
```
csrwng, line 160: This could be handled with a builder pattern as well

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -1665,16 +1663,6 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
 		return ctrl.Result{}, fmt.Errorf("failed to parse release image version: %w", err)
 	}
 
-	// In >= 4.11 We want to move most of the components reconciliation down to the CPO https://issues.redhat.com/browse/HOSTEDCP-375.
-	// For IBM existing clusters < 4.11 we need to stay consistent and keep deploying existing pods to satisfy validations.
-	// TODO (alberto): drop this after dropping < 4.11 support.
-	if !controlPlaneOperatorManagesMachineAutoscaler {
```
csrwng, line 1671: We should double check whether this can be removed

control-plane-operator/controllers/hostedcontrolplane/autoscaler/reconcile.go
```diff
@@ -1,46 +1,106 @@
...
+	}
+
+	// The deployment depends on the kubeconfig being reported.
+	if hcp.Status.KubeConfig == nil {
+		return false, nil
+	}
+	// Resolve the kubeconfig secret for CAPI which the autoscaler is deployed alongside of.
+	capiKubeConfigSecret := manifests.KASServiceCAPIKubeconfigSecret(hcp.Namespace, hcp.Spec.InfraID)
+	err := cpContext.Client.Get(cpContext, client.ObjectKeyFromObject(capiKubeConfigSecret), capiKubeConfigSecret)
+	if err != nil {
```
csrwng, line 83: should this return an error if it's just not found?

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -906,6 +916,30 @@ func (r *HostedControlPlaneReconciler) update(ctx context.Context, hostedControl
 		errs = append(errs, err)
 	}
 
+	// reconcile components only when kube apiserver is fully ready
+	// TOOD(Mulham): remove when dependcnies managment abstraction is implemented.
```
csrwng, line 920: s/TOOD/TODO/

csrwng (commented):
Some comments. Awesome stuff.

control-plane-operator/controllers/hostedcontrolplane/autoscaler/reconcile.go
muraee: fixed

control-plane-operator/controllers/hostedcontrolplane/routecm/reconcile.go
muraee: makes sense, done!

control-plane-operator/controllers/hostedcontrolplane/routecm/reconcile.go
muraee: yeah that would be better for sure. I tried to do that, but it turns out the configMap reconciler also needs the volumes.
https://github.com/openshift/hypershift/blob/5479c0ca8b0b59f0b9561a25d0955c293e4f47f6/control-plane-operator/controllers/hostedcontrolplane/routecm/config.go#L56

I have found the same pattern in multiple other components as well. we could pass the volumes to the genricReconciler reconciler function as well, but I am not sure if that makes sense.

control-plane-operator/controllers/hostedcontrolplane/routecm/reconcile.go
muraee: could you please elaborate more.

control-plane-operator/controllers/hostedcontrolplane/routecm/reconcile.go
muraee: seems we already do that and stomp whatever is set here in the deploymentConfig if replicas > 1
https://github.com/openshift/hypershift/blob/f7fddd1961a24b40fa0a20d30489e9a2f7284a53/support/config/deployment.go#L88-L97

so just removing this code here would be enough, right?

control-plane-operator/controllers/hostedcontrolplane/routecm/reconcile.go
muraee: afaics, not all components pass their labels as MultizoneSpreadLabels, I am not sure what the criteria is. How would you consolidate the labels with MultizoneSpreadLabels in this case?

control-plane-operator/controllers/hostedcontrolplane/routecm/reconcile.go
muraee: added!

control-plane-operator/controllers/hostedcontrolplane/routecm/reconcile.go
csrwng: > we could pass the volumes to the genricReconciler reconcile function too

I think that would work

control-plane-operator/controllers/hostedcontrolplane/routecm/reconcile.go
csrwng: Instead of having a generic reconcile function where we replicate some code across components, we could instead use a set of functions like 
container("name").Command("blah").Args("foo","bar").WithEnv(blah)

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
csrwng: that's what I'm not sure about. It may still be needed for ROKS or satellite

control-plane-operator/controllers/hostedcontrolplane/routecm/reconcile.go
csrwng: yes

muraee: removed

control-plane-operator/controllers/hostedcontrolplane/routecm/reconcile.go
muraee: done! let me know if that looks good to you.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
muraee: confirmed with Hide, they don't have any clusters < 4.11

muraee:
/close
superseded by https://github.com/openshift/hypershift/pull/4708

PR #4453, opened 2024-07-31
HOSTEDCP-1764: retrieve registryOverrides when ImageStream is not ava…

Patryk-Stefanski:
**What this PR does / why we need it**: Remove need for annotating images on create command when we are in aks

Fixes # [HOSTEDCP-1764](https://issues.redhat.com/browse/HOSTEDCP-1764)

**Checklist**

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
@@ -3760,12 +3760,12 @@ func (r *HostedControlPlaneReconciler) reconcileOperatorLifecycleManager(ctx con
 					}
 				}
 			} else {
-				for name, catalog := range olm.CatalogToImage {
-					pullSecret := common.PullSecret(hcp.Namespace)
-					if err := r.Get(ctx, client.ObjectKeyFromObject(pullSecret), pullSecret); err != nil {
-						return fmt.Errorf("failed to get pull secret for namespace %s: %w", hcp.Namespace, err)
-					}
+				pullSecret := common.PullSecret(hcp.Namespace)
+				if err := r.Get(ctx, client.ObjectKeyFromObject(pullSecret), pullSecret); err != nil {
+					return fmt.Errorf("failed to get pull secret for namespace %s: %w", hcp.Namespace, err)
+				}
 
+				for name, catalog := range olm.CatalogToImage {
 					parts := strings.Split(catalog, ":")
 					if len(parts) != 2 {
 						return fmt.Errorf("invalid catalog format %s, expected 'registry:tag'", catalog)
@@ -3802,7 +3802,7 @@ func (r *HostedControlPlaneReconciler) reconcileOperatorLifecycleManager(ctx con
 							*override = imageOverride
 						}
 					} else {
-						return fmt.Errorf("unknown catalog %s", name)
+						return fmt.Errorf("unknown catalog for catalog image override %s", name)
 					}
 				}
 			}
--- a/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
@@ -76,6 +76,7 @@ import (
 	"github.com/openshift/hypershift/support/proxy"
 	"github.com/openshift/hypershift/support/releaseinfo"
 	"github.com/openshift/hypershift/support/releaseinfo/registryclient"
+	"github.com/openshift/hypershift/support/thirdparty/library-go/pkg/image/reference"
 	"github.com/openshift/hypershift/support/upsert"
 	"github.com/openshift/hypershift/support/util"
 	prometheusoperatorv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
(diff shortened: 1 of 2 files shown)


control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -5,6 +5,8 @@ import (
 	crand "crypto/rand"
 	"errors"
 	"fmt"
+	"github.com/openshift/hypershift/support/releaseinfo/registryclient"
```
bryan-cox, line 8: nit: can you fix the imports here

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -3743,17 +3745,48 @@ func (r *HostedControlPlaneReconciler) reconcileOperatorLifecycleManager(ctx con
 				return fmt.Errorf("failed to reconcile catalogs: %w", err)
 			}
 
-			catalogsImageStream := manifests.CatalogsImageStream(hcp.Namespace)
-			if !overrideImages {
```
bryan-cox, line 3747: I don't know that we want to get rid of this. I wonder if IBM uses this or there are other use cases. 🤔

bryan-cox (commented):
Nice that you got this working!

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
Patryk-Stefanski: Were not getting rid of this were jus making sure it only runs if the mgmgt cluster has imageStreams capability

bryan-cox: Ah ok I see now

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -3743,17 +3745,48 @@ func (r *HostedControlPlaneReconciler) reconcileOperatorLifecycleManager(ctx con
...
+			} else {
+				for name, catalog := range olm.CatalogToImage {
+					pullSecret := common.PullSecret(hcp.Namespace)
+					if err := r.Get(ctx, client.ObjectKeyFromObject(pullSecret), pullSecret); err != nil {
+						return fmt.Errorf("failed to get pull secret: %w", err)
+					}
+
+					listDigest, err := registryclient.GetListDigest(ctx, catalog, pullSecret.Data[corev1.DockerConfigJsonKey])
+					if err != nil {
+						log.Fatalf("Failed to get manifest: %v", err)
```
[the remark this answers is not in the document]
bryan-cox: Yeah we should return back instead

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -3743,17 +3744,48 @@ func (r *HostedControlPlaneReconciler) reconcileOperatorLifecycleManager(ctx con
...
+				} else {
+					if _, err := util.DeleteIfNeeded(ctx, r, catalogsImageStream); err != nil {
+						return fmt.Errorf("failed to remove OLM Catalog ImageStream: %w", err)
+					}
 				}
-			} else if r.ManagementClusterCapabilities.Has(capabilities.CapabilityImageStream) {
-				if _, err := util.DeleteIfNeeded(ctx, r, catalogsImageStream); err != nil {
-					return fmt.Errorf("failed to remove OLM Catalog ImageStream: %w", err)
+			} else {
+				for name, catalog := range olm.CatalogToImage {
```
fxierh, line 3762: Looks like if the MC does not have IS cap, existing image overrides set via --annotations will not be respected, and p.RedHatOperatorsCatalogImageOverride will be set forcefully 🤔

Patryk-Stefanski: I added some logic to allow for annotations to work as well, I think there's no harm in providing this granularity.

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -3743,17 +3744,49 @@ func (r *HostedControlPlaneReconciler) reconcileOperatorLifecycleManager(ctx con
...
+
+					catalogOverrides := map[string]*string{
+						"redhat-operators":    &p.RedHatOperatorsCatalogImageOverride,
+						"certified-operators": &p.CertifiedOperatorsCatalogImageOverride,
+						"community-operators": &p.CommunityOperatorsCatalogImageOverride,
+						"redhat-marketplace":  &p.RedHatMarketplaceCatalogImageOverride,
+					}
+
+					if override, exists := catalogOverrides[name]; exists {
+						if *override == "" {
```
bryan-cox, line 3784: Should use ptr.Deref here

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -3743,17 +3744,66 @@ func (r *HostedControlPlaneReconciler) reconcileOperatorLifecycleManager(ctx con
...
+					if _, err := util.DeleteIfNeeded(ctx, r, catalogsImageStream); err != nil {
+						return fmt.Errorf("failed to remove OLM Catalog ImageStream: %w", err)
+					}
 				}
-			} else if r.ManagementClusterCapabilities.Has(capabilities.CapabilityImageStream) {
-				if _, err := util.DeleteIfNeeded(ctx, r, catalogsImageStream); err != nil {
-					return fmt.Errorf("failed to remove OLM Catalog ImageStream: %w", err)
+			} else {
+				for name, catalog := range olm.CatalogToImage {
+					pullSecret := common.PullSecret(hcp.Namespace)
```
bryan-cox, line 3764: Could you pull this block out before the for statement? That way it is not getting the pull secret over and over.

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -3743,17 +3744,66 @@ func (r *HostedControlPlaneReconciler) reconcileOperatorLifecycleManager(ctx con
...
+					if err := r.Get(ctx, client.ObjectKeyFromObject(pullSecret), pullSecret); err != nil {
+						return fmt.Errorf("failed to get pull secret for namespace %s: %w", hcp.Namespace, err)
+					}
+
+					parts := strings.Split(catalog, ":")
+					if len(parts) != 2 {
+						return fmt.Errorf("invalid catalog format %s, expected 'registry:tag'", catalog)
+					}
+					catalogImage, tag := parts[0], parts[1]
```
bryan-cox, line 3773: Is this always a tag and never a digest?

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -3743,17 +3744,66 @@ func (r *HostedControlPlaneReconciler) reconcileOperatorLifecycleManager(ctx con
...
+						"community-operators": &p.CommunityOperatorsCatalogImageOverride,
+						"redhat-marketplace":  &p.RedHatMarketplaceCatalogImageOverride,
+					}
+
+					if override, exists := catalogOverrides[name]; exists {
+						if ptr.Deref(override, "") == "" {
+							*override = imageOverride
+						}
+					} else {
+						return fmt.Errorf("unknown catalog %s", name)
```
bryan-cox, line 3805: Could you make this a little more verbose for debuggin purposes in the future? Something like "unknown catalog for catalog image override".

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
Patryk-Stefanski: this is retrieving it from the olm.CatalogImage var and from what I've seen its always a tag.

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -3743,17 +3744,66 @@ func (r *HostedControlPlaneReconciler) reconcileOperatorLifecycleManager(ctx con
...
+				for name, catalog := range olm.CatalogToImage {
+					parts := strings.Split(catalog, ":")
+					if len(parts) != 2 {
+						return fmt.Errorf("invalid catalog format %s, expected 'registry:tag'", catalog)
+					}
+					catalogImage, tag := parts[0], parts[1]
+
+					catalogImageParts := strings.Split(catalogImage, "/")
+					catalogRepo := catalogImageParts[len(catalogImageParts)-1]
+					catalogRegistry := strings.Join(catalogImageParts[:len(catalogImageParts)-1], "/")
```
csrwng, line 3777: Please use https://github.com/openshift/hypershift/blob/a11f0a1b12da74f345953f9441221e4418c314f4/support/thirdparty/library-go/pkg/image/reference/reference.go#L32 to parse the image ref

csrwng: and in general use the funcs there to build new refs

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -3743,17 +3744,66 @@ func (r *HostedControlPlaneReconciler) reconcileOperatorLifecycleManager(ctx con
...
+				} else {
+					if _, err := util.DeleteIfNeeded(ctx, r, catalogsImageStream); err != nil {
+						return fmt.Errorf("failed to remove OLM Catalog ImageStream: %w", err)
+					}
 				}
-			} else if r.ManagementClusterCapabilities.Has(capabilities.CapabilityImageStream) {
-				if _, err := util.DeleteIfNeeded(ctx, r, catalogsImageStream); err != nil {
-					return fmt.Errorf("failed to remove OLM Catalog ImageStream: %w", err)
+			} else {
+				pullSecret := common.PullSecret(hcp.Namespace)
```
csrwng, line 3763: should you not check here if overrideImages is set?

csrwng (commented):
a couple of comments

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
Patryk-Stefanski: No because this should also be run when there is no registry overrides

csrwng: But if overrideImages is set shouldn't we skip the logic of pulling from the registry?

Patryk-Stefanski: no we want to pull from the registry regardless of overrideImages. AKS has no ImageStream functionality which is what we use to set the catalog images so we need logic to do this on AKS regardless of the overrides. But if the overrides are provided then we should retrieve them catalog images from the mirror registry.

Patryk-Stefanski: Ah I gotcha now, was mixing it up with registryOverrides.

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -3743,17 +3745,66 @@ func (r *HostedControlPlaneReconciler) reconcileOperatorLifecycleManager(ctx con
...
-					return fmt.Errorf("failed to remove OLM Catalog ImageStream: %w", err)
+			} else {
+				if !overrideImages {
+					pullSecret := common.PullSecret(hcp.Namespace)
+					if err := r.Get(ctx, client.ObjectKeyFromObject(pullSecret), pullSecret); err != nil {
+						return fmt.Errorf("failed to get pull secret for namespace %s: %w", hcp.Namespace, err)
+					}
+
+					for name, catalog := range olm.CatalogToImage {
+						dockerImageRef, err := reference.Parse(catalog)
```
csrwng, line 3771: nit: s/dockerImageRef/imageRef/

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -3743,17 +3745,66 @@ func (r *HostedControlPlaneReconciler) reconcileOperatorLifecycleManager(ctx con
...
+						if err != nil {
+							return fmt.Errorf("failed to parse catalog image %s: %w", catalog, err)
+						}
+
+						if len(isImageRegistryOverrides) > 0 {
+							for registrySource, registryDest := range isImageRegistryOverrides {
+								if strings.Contains(dockerImageRef.Exact(), registrySource) {
```
csrwng, line 3778: why not just compare the registry ? 
if imageRef.Registry == registrySource

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -3743,17 +3745,66 @@ func (r *HostedControlPlaneReconciler) reconcileOperatorLifecycleManager(ctx con
...
+									dockerImageRef, err = reference.Parse(strings.Replace(dockerImageRef.Exact(), registrySource, registryDest[0], 1))
```
csrwng, line 3779: to construct the new ref, you can just copy the original and replace the .Registry field

csrwng (commented):
More comments

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
Patryk-Stefanski: thats what I had done originally but noticed that what we do in the code above is better because with the override we can point to a different registry, registry namespace and registry repo so it gives us more control.

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
Patryk-Stefanski: yes if we were only overriding the registry of the image, but for the same reason as [here](https://github.com/openshift/hypershift/pull/4453#discussion_r1709628865) Id like to keep it as is.

PR #4454, opened 2024-07-31
HOSTEDCP-1851: Use a multi-arch haproxy image for shared ingress

bryan-cox:
Use a multi-arch haproxy image for shared ingress

Fixes [HOSTEDCP-1851](https://issues.redhat.com/browse/HOSTEDCP-1851)

**Checklist**


hypershift-operator/controllers/nodepool/haproxy.go
```diff
@@ -237,7 +237,7 @@ func apiServerProxyConfig(haProxyImage, cpoImage, clusterID,
 	if sharedingress.UseSharedIngress() {
 		// proxy protocol v2 with TLV support (custom proxy protocol header) requires haproxy v2.9+, see: https://www.haproxy.com/blog/announcing-haproxy-2-9#proxy-protocol-tlv-fields
 		// TODO: get the image from the payload once available https://issues.redhat.com/browse/HOSTEDCP-1819
-		haProxyImage = "quay.io/mraee/haproxy:2.9"
+		haProxyImage = "quay.io/rh_ee_brcox/hypershift:haproxy2.9.9-multi"
```
jparrill, line 240: Do we have plan to host the image in our hypershift quay repo?

bryan-cox: No it will come from the release payload. The story for that is the comment right above this line.

muraee:
please also update 
https://github.com/openshift/hypershift/blob/ad54764956732525a14a60f811dab6e5a64ccb06/hypershift-operator/controllers/sharedingress/router.go#L195

bryan-cox:
> please also update
> 
> https://github.com/openshift/hypershift/blob/ad54764956732525a14a60f811dab6e5a64ccb06/hypershift-operator/controllers/sharedingress/router.go#L195

Ah thanks! I did not realize it was used in more than one spot.

PR #4456, opened 2024-07-31
OCPBUGS-37052: Use konnectivity HTTP proxy for ingress controller

csrwng:
Ingress operator canary checks require access to the internet since they probe the external canary routes. The socks5 proxy sidecar does not provide external access when a user proxy is required. This commit switches to using the konnectivity-https-proxy that does take into account the user's proxy configuration.

Fixes #OCPBUGS-37052

**Checklist**

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/ingressoperator/ingressoperator.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/ingressoperator/ingressoperator.go
@@ -3,6 +3,7 @@ package ingressoperator
 import (
 	"fmt"
 
+	configv1 "github.com/openshift/api/config/v1"
 	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
 	"github.com/openshift/hypershift/control-plane-operator/controllers/hostedcontrolplane/imageprovider"
 	"github.com/openshift/hypershift/control-plane-operator/controllers/hostedcontrolplane/kas"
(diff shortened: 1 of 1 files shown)


control-plane-operator/controllers/hostedcontrolplane/ingressoperator/ingressoperator.go
```diff
@@ -116,11 +124,11 @@ func ReconcileDeployment(dep *appsv1.Deployment, params Params, platformType hyp
 			{Name: "KUBECONFIG", Value: "/etc/kubernetes/kubeconfig"},
 			{
 				Name:  "HTTP_PROXY",
-				Value: fmt.Sprintf("socks5://127.0.0.1:%d", kas.KonnectivityServerLocalPort),
+				Value: fmt.Sprintf("http://127.0.0.1:%d", konnectivityProxyPort),
 			},
 			{
 				Name:  "HTTPS_PROXY",
-				Value: fmt.Sprintf("socks5://127.0.0.1:%d", kas.KonnectivityServerLocalPort),
+				Value: fmt.Sprintf("http://127.0.0.1:%d", konnectivityProxyPort),
```
jparrill, line 131: do we need to manage HTTPS env var using an https scheme on the value? 
If that so, do we need to handle the TLS certificates in that regard?

csrwng: > do we need to manage HTTPS env var using an https scheme on the value?
No, the client uses the HTTP proxy as a TCP CONNECT tunnel through which the HTTPS traffic is sent.

csrwng:
/retest-required
(failures don't seem related to this PR)

PR #4466, opened 2024-08-01
HOSTEDCP-1669: Add first azure services validation

enxebre:
This enforce all azure services to be route and have a hostname. This satisfies shared ingress expectation

This enforce all azure services to be route and have a hostname. This satisfies shared ingress expectation
This is the first of a series of PRs to improve API UX validation 

Fixes #

**Checklist**

### DIFF
--- a/api/hypershift/v1beta1/hostedcluster_types.go
+++ b/api/hypershift/v1beta1/hostedcluster_types.go
@@ -321,7 +321,12 @@ const (
 )
 
 // HostedClusterSpec is the desired behavior of a HostedCluster.
+
 // +kubebuilder:validation:XValidation:rule="self.platform.type != 'IBMCloud' ? self.services == oldSelf.services : true", message="Services is immutable. Changes might result in unpredictable and disruptive behavior."
+// +kubebuilder:validation:XValidation:rule="self.platform.type == 'Azure' ? self.services.exists(s, s.service == 'APIServer' && s.servicePublishingStrategy.type == 'Route' && s.servicePublishingStrategy.route.hostname != '') : true",message="Azure platform requires APIServer Route service with a hostname to be defined"
+// +kubebuilder:validation:XValidation:rule="self.platform.type == 'Azure' ? self.services.exists(s, s.service == 'OAuthServer' && s.servicePublishingStrategy.type == 'Route' && s.servicePublishingStrategy.route.hostname != '') : true",message="Azure platform requires OAuthServer Route service with a hostname to be defined"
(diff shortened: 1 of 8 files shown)


api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -321,6 +321,10 @@ const (
...
+// +kubebuilder:validation:XValidation:rule="self.platform.type == 'Azure' ? self.services.exists(s, s.service == 'OAuthServer' && s.servicePublishingStrategy.type == 'Route' && s.servicePublishingStrategy.route.hostname != '') : true",message="Azure platform requires OAuthServer Route service with a hostname to be defined"
+// +kubebuilder:validation:XValidation:rule="self.platform.type == 'Azure' ? self.services.exists(s, s.service == 'Konnectivity' && s.servicePublishingStrategy.type == 'Route' && s.servicePublishingStrategy.route.hostname != '') : true",message="Azure platform requires Konnectivity Route service with a hostname to be defined"
+// +kubebuilder:validation:XValidation:rule="self.platform.type == 'Azure' ? self.services.exists(s, s.service == 'Ignition' && s.servicePublishingStrategy.type == 'Route' && s.servicePublishingStrategy.route.hostname != '') : true",message="Azure platform requires Ignition Route service with a hostname to be defined"
```
muraee, line 327: maybe we can combine the 4 separate validations with `all` instead?:
```
// +kubebuilder:validation:XValidation:rule="self.platform.type == 'Azure' ? self.services.all(s, s.servicePublishingStrategy.type == 'Route' && s.servicePublishingStrategy.route.hostname != '') : true",message="Azure platform requires all services to use Route with a hostname defined"
```

and then maybe another one to check if len(services) >= 4 or smth if we want to make sure all services are defined.

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -321,6 +321,10 @@ const (
 )
 
 // HostedClusterSpec is the desired behavior of a HostedCluster.
+// +kubebuilder:validation:XValidation:rule="self.platform.type == 'Azure' ? self.services.exists(s, s.service == 'APIServer' && s.servicePublishingStrategy.type == 'Route' && s.servicePublishingStrategy.route.hostname != '') : true",message="Azure platform requires APIServer Route service with a hostname to be defined"
```
csrwng, line 324: I think this is true for all platforms, not only Azure. If you're using a route for APIserver you need to specify the hostname

enxebre: ack, once we get this in with agreement on the proposed testing pattern I'll follow up for increase/improve validations.

api/hypershift/v1beta1/hostedcluster_types.go
enxebre: ack, once we get this in with agreement on the proposed testing pattern I'll follow up for increase/improve validations.

PR #4474, opened 2024-08-01
OSASINFRA-3550: openstack: ingress API & octavia ingress provider

EmilienM:

Until now, we didn't have a working Ingress on an initial deployment, and we thought we would just rely on day 2 operations to setup Ingress with e.g. MetalLB.
It turns out that we could use OpenStack Octavia (Load Balancer as a service) as an optional provider for Ingress, at least for now in dev-preview.

With this PR, we don't expect the OpenStack cloud to have Octavia enabled which is why we disable the provider by default, which means Ingress won't be working on day 1 and it's up to the user to deploy something (e.g. MetalLB).

If Octavia is used, the floating IP and the DNS record for Ingress would have to be pre-created before deploying the hosted cluster.
The new CLI would be:
```bash
hypershift create cluster openstack --openstack-ingress-provider Octavia --openstack-ingress-floating-ip <-ingress-floating-IP> (...)
```

We have some validation that check that if a FIP is provided, it's only accepted with the Octavia provider for now.
The HCP.Spec has been updated as well to allow the Ingress FIP to be stored and the Ingress provider (default to `None` now).

In the future, we might change the default and even remove the Octavia provider if we find a better way (e.g. re-using on-prem LB, based on Keepalived).
But by keeping it "disabled" by default ensure some sort of backward compatibility even if OpenStack is dev-preview now.

### DIFF
--- a/api/hypershift/v1beta1/hostedcluster_types.go
+++ b/api/hypershift/v1beta1/hostedcluster_types.go
@@ -1918,6 +1918,48 @@ type OpenStackPlatformSpec struct {
 	// +listType=set
 	// +optional
 	Tags []string `json:"tags,omitempty"`
+
+	// IngressFloatingIP is an available floating IP in your OpenStack cluster that will
+	// be associated with the OpenShift ingress port.
+	//
+	// +optional
+	IngressFloatingIP string `json:"ingressFloatingIP,omitempty"`
+
+	// IngressProvider specifies the ingress provider to use for the cluster.
+	// Valid values are None and Octavia.
+	// The default value is None which means Ingress is not enabled.
+	// Octavia provider is experimental and might be replaced in the future.
+	// When Octavia is selected, a floating IP has to be provided in IngressFloatingIP.
+	//
+	// +default=None
+	// +kubebuilder:default=None
+	// +kubebuilder:validation:Enum=None;Octavia
+	// +optional
+	IngressProvider OpenStackIngressProvider `json:"ingressProvider,omitempty"`
+}
+
+// OpenStackIngressProvider specifies the ingress provider to use for the cluster.
+type OpenStackIngressProvider string
+
+const (
+	// OpenStackIngressProviderNone is the default value and means no ingress provider is used.
+	OpenStackIngressProviderNone OpenStackIngressProvider = OpenStackIngressProvider("None")
+
+	// OpenStackIngressProviderOctavia is the value to use when using Octavia as the ingress provider.
+	OpenStackIngressProviderOctavia OpenStackIngressProvider = OpenStackIngressProvider("Octavia")
+)
+
+func (p *OpenStackIngressProvider) String() string {
+	return string(*p)
+}
+
+func (p *OpenStackIngressProvider) Set(s string) error {
+	*p = OpenStackIngressProvider(s)
+	return nil
+}
+
+func (p *OpenStackIngressProvider) Type() string {
+	return "OpenStackIngressProvider"
 }
 
 // OpenStackIdentityReference is a reference to an infrastructure
--- a/control-plane-operator/hostedclusterconfigoperator/controllers/resources/ingress/reconcile.go
+++ b/control-plane-operator/hostedclusterconfigoperator/controllers/resources/ingress/reconcile.go
@@ -164,3 +164,57 @@ func ReconcileDefaultIngressPassthroughRoute(route *routev1.Route, cpService *co
 
 	return nil
 }
+
+// ReconcileDefaultIngressOctaviaService reconciles the default Ingress Service.
+func ReconcileDefaultIngressOctaviaService(ingressService *corev1.Service, ingressFloatingIP string) error {
+	if ingressFloatingIP == "" {
+		return fmt.Errorf("ingress floating IP is empty")
+	}
+
+	ingressService.Spec.Type = corev1.ServiceTypeLoadBalancer
+	ingressService.Spec.LoadBalancerIP = ingressFloatingIP
+
+	// Reconcile ports for http and https.
+	// We don't want to override the whole ServicePort, because
+	// after a reconcile loop, the spec will be updated with more fields.
+	var httpPortFound, httpsPortFound bool
+	for i := range ingressService.Spec.Ports {
+		switch ingressService.Spec.Ports[i].Name {
+		case "http":
+			ingressService.Spec.Ports[i].Protocol = corev1.ProtocolTCP
+			ingressService.Spec.Ports[i].Port = 80
+			ingressService.Spec.Ports[i].TargetPort = intstr.FromInt(80)
+			httpPortFound = true
+		case "https":
+			ingressService.Spec.Ports[i].Protocol = corev1.ProtocolTCP
+			ingressService.Spec.Ports[i].Port = 443
+			ingressService.Spec.Ports[i].TargetPort = intstr.FromInt(443)
+			httpsPortFound = true
+		}
+	}
+
+	if !httpPortFound {
+		ingressService.Spec.Ports = append(ingressService.Spec.Ports, corev1.ServicePort{
+			Name:       "http",
+			Protocol:   corev1.ProtocolTCP,
+			Port:       80,
+			TargetPort: intstr.FromInt(80),
+		})
+	}
+
+	if !httpsPortFound {
+		ingressService.Spec.Ports = append(ingressService.Spec.Ports, corev1.ServicePort{
+			Name:       "https",
+			Protocol:   corev1.ProtocolTCP,
+			Port:       443,
+			TargetPort: intstr.FromInt(443),
+		})
+	}
+
+	if ingressService.Spec.Selector == nil {
+		ingressService.Spec.Selector = map[string]string{}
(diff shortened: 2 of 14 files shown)


support/openstackutil/ingress.go
```diff
@@ -0,0 +1,26 @@
...
+	}
+	if ingressFloatingIP != "" && ingressProvider == "" {
+		return fmt.Errorf("cannot set floating IP without specifying ingress provider")
+	}
+	if ingressProvider != "" && ingressFloatingIP == "" {
+		return fmt.Errorf("cannot set ingress provider without specifying floating IP")
+	}
+	// For now, the floating IP can only be set when the ingress provider is "Octavia".
+	// This is because the floating IP is only used for the Octavia ingress provider.
+	if ingressProvider != "" && ingressProvider != hyperv1.OpenStackIngressProviderOctavia {
```
MaysaMacedo, line 22: Based on the comment above, shouldn't it be like this:
```suggestion
	if ingressFloatingIP != "" && ingressProvider != hyperv1.OpenStackIngressProviderOctavia {
```

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -930,6 +931,24 @@ func (r *reconciler) reconcileIngressController(ctx context.Context, hcp *hyperv
 		}
 	}
 
+	if hcp.Spec.Platform.Type == hyperv1.OpenStackPlatform {
+		ingressFloatingIP := hcp.Spec.Platform.OpenStack.IngressFloatingIP
+		ingressProvider := hcp.Spec.Platform.OpenStack.IngressProvider
+		if err := openstackutil.ValidateIngressOptions(ingressProvider, ingressFloatingIP); err != nil {
```
MaysaMacedo, line 937: why do we need to validate it again if it was already validate at `cmd/cluster/openstack/create.go`?
Is it because the hostedcluster could be created without cmdline?

EmilienM: yeah exactly, I thought it'll be nice to avoid controller errors in case someone creates the cluster via manifests.

support/openstackutil/ingress.go
EmilienM: at this point we know `ingressFloatingIP` is not empty, see check 1 and 3 before.

EmilienM:
/test e2e-azure
resting Azure but it seems unstable

EmilienM:
/test e2e-openstack
to ensure there is no regression on default deployments

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1911,6 +1911,44 @@ type OpenStackPlatformSpec struct {
...
 	// +optional
 	Tags []string `json:"tags,omitempty"`
+
+	// IngressFloatingIP is an available floating IP in your OpenStack cluster that will
+	// be associated with the OpenShift ingress port.
+	//
+	// +optional
+	IngressFloatingIP string `json:"ingressFloatingIP,omitempty"`
+
+	// IngressProvider specifies the ingress provider to use for the cluster.
```
EmilienM, line 1921: I guess I could have add more docs on this one.
I'll do it in a next PR if people don't mind, unless I'm asked to change something else, then I'll do it in batch.
Thanks

cmd/cluster/openstack/testdata/zz_fixture_TestCreateCluster_wrong_ingress_provider_name.yaml
```diff
@@ -0,0 +1,134 @@
+apiVersion: v1
```
EmilienM, line 1: that file is wrong I need to remove it

EmilienM:
/retest-required
/test e2e-openstack
CI glitch on the registry

mdbooth:
I'd prefer to see https://github.com/openshift/release/pull/55115 merged first so we can the e2e tests passing.

EmilienM:
> I'd prefer to see [openshift/release#55115](https://github.com/openshift/release/pull/55115) merged first so we can the e2e tests passing.

fair

EmilienM:
I confirm that we have no regression with that PR:

```
    util.go:324: ClusterVersionProgressing=True: ClusterOperatorNotAvailable(Unable to apply 4.17.0-0.ci.test-2024-08-02-142747-ci-op-xv0zc09t-latest: the cluster operator console is not available)
```
like before.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -930,6 +931,24 @@ func (r *reconciler) reconcileIngressController(ctx context.Context, hcp *hyperv
...
+			errs = append(errs, err)
```
csrwng, line 938: seems that should return here

csrwng (commented):
Just one comment

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
EmilienM: done

EmilienM:
/hold for CI
I'll hold cancel once we have positive results.

EmilienM:
/test e2e-openstack
With the recent changes in openshift/release, we should make progress to have Ingress working in the CI job.

EmilienM:
OpenStack CI will fail until we get https://github.com/openshift/release/pull/55160 (at least). let's wait and retest once ready.

EmilienM:
Status of this PR:
* I want to get the aws job green to avoid we don't break anything
* once https://github.com/openshift/release/pull/55196 has merged, we can run e2e-openstack and hopefully get good results. If not we'll iterate.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/ingress/reconcile.go
```diff
@@ -164,3 +164,32 @@ func ReconcileDefaultIngressPassthroughRoute(route *routev1.Route, cpService *co
...
 	return nil
 }
+
+// ReconcileDefaultIngressOctaviaService reconciles the default Ingress Service.
+func ReconcileDefaultIngressOctaviaService(ingressService *corev1.Service, ingressFloatingIP string) error {
+	if ingressFloatingIP == "" {
+		return fmt.Errorf("ingress floating IP is empty")
+	}
+
+	ingressService.Spec = corev1.ServiceSpec{}
```
sjenning, line 174: This is erasing the exist spec leading to a hot-reconcile loop
```
{"level":"info","ts":"2024-08-05T22:41:07Z","msg":"WARNING: Object got updated more than one time without a no-op update, this indicates hypershift incorrectly reverting defaulted values","type":"*v1.Service","name":"openshift-ingress/octavia-ingress","diff":"  &v1.Service{
  	TypeMeta:   {Kind: "Service", APIVersion: "v1"},
  	ObjectMeta: {Name: "octavia-ingress", Namespace: "openshift-ingress", UID: "47e7e65e-63d9-4291-b2f3-32fec7b5f5dd", ResourceVersion: "1306", ...},
  	Spec: v1.ServiceSpec{
  		Ports: []v1.ServicePort{
  			{
  				... // 3 identical fields
  				Port:       80,
  				TargetPort: {IntVal: 80},
- 				NodePort:   31330,
+ 				NodePort:   0,
  			},
  			{
  				... // 3 identical fields
  				Port:       443,
  				TargetPort: {IntVal: 443},
- 				NodePort:   30416,
+ 				NodePort:   0,
  			},
  		},
  		Selector:  {"ingresscontroller.operator.openshift.io/deployment-ingresscontroller": "default"},
  		ClusterIP: "172.31.230.21",
  		... // 5 identical fields
  		LoadBalancerSourceRanges:      nil,
  		ExternalName:                  "",
- 		ExternalTrafficPolicy:         "Cluster",
+ 		ExternalTrafficPolicy:         "",
  		HealthCheckNodePort:           0,
  		PublishNotReadyAddresses:      false,
  		SessionAffinityConfig:         nil,
  		IPFamilies:                    {"IPv4"},
  		IPFamilyPolicy:                &"SingleStack",
- 		AllocateLoadBalancerNodePorts: &true,
+ 		AllocateLoadBalancerNodePorts: nil,
  		LoadBalancerClass:             nil,
- 		InternalTrafficPolicy:         &"Cluster",
+ 		InternalTrafficPolicy:         nil,
  		TrafficDistribution:           nil,
  	},
  	Status: {},
  }
","semanticDeepEqual":false,"updateCount":10}
```
The spec should be reconciled field by field, not rebuild from scratch.

sjenning (changes requested):
/hold

EmilienM:
Note: for testing purpose, I cherry picked the commits from https://github.com/openshift/hypershift/pull/4492 until they are merged. This PR is still on hold for now, until we get green CI on e2e-openstack.

EmilienM:
@sjenning I removed the 2 commits from the other PR since I got better results, this PR can now land I think. Please add the LGTM again.
Thanks

/hold cancel

PR #4477, opened 2024-08-02
OSASINFRA-3554: OpenStack: Use external network ID in favor of Name

MaysaMacedo:
When the user provides an external network to hypershift that must be used by CCM to ensure that network is used when creating Floating IPs for the Load Balancers. CCM expects an external Network ID, right now we accept a Name from the user. This commit changes this behavior to consume an ID instead.

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/cloud/openstack/providerconfig.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/cloud/openstack/providerconfig.go
@@ -19,11 +19,11 @@ const (
 // ReconcileCloudConfigSecret reconciles the cloud config secret.
 // For some controllers (e.g. Manila CSI, CNCC, etc), the cloud config needs to be stored in a secret.
 // In the hosted cluster config operator, we create the secrets needed by these controllers.
-func ReconcileCloudConfigSecret(secret *corev1.Secret, cloudName string, credentialsSecret *corev1.Secret, caCertData []byte) error {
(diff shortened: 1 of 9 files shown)


control-plane-operator/controllers/hostedcontrolplane/cloud/openstack/providerconfig.go
```diff
@@ -19,11 +19,11 @@ const (
 // ReconcileCloudConfigSecret reconciles the cloud config secret.
 // For some controllers (e.g. Manila CSI, CNCC, etc), the cloud config needs to be stored in a secret.
 // In the hosted cluster config operator, we create the secrets needed by these controllers.
-func ReconcileCloudConfigSecret(secret *corev1.Secret, cloudName string, credentialsSecret *corev1.Secret, caCertData []byte) error {
+func ReconcileCloudConfigSecret(externalNetworkID *string, secret *corev1.Secret, cloudName string, credentialsSecret *corev1.Secret, caCertData []byte) error {
```
EmilienM, line 22: I'm fine passing the ID. An alternative could have to pass the whole spec like some other platforms do here; but again this is good :+1: .

EmilienM:
/lgtm
It's a necessary change for most clouds. Thanks

EmilienM:
you need to update `cmd/cluster/openstack/create_test.go`

EmilienM (approved):
/lgtm
Thanks!

PR #4488, opened 2024-08-05
HOSTEDCP-1850: Allow setting Kube APIServer maximum requests in flight

csrwng:
Adds support for overriding default maximum requests in flight for Kubernetes API server. This will allow service providers to rate limit the Kubernetes API server based on available resources.

Fixes #[HOSTEDCP-1850](https://issues.redhat.com/browse/HOSTEDCP-1850)

**Checklist**


csrwng:
Note: this is the CPO change. There will be a separate PR for the HO change. This one needs to be backported while the HO change does not.

control-plane-operator/controllers/hostedcontrolplane/kas/params.go
```diff
@@ -118,6 +121,12 @@ func NewKubeAPIServerParams(ctx context.Context, hcp *hyperv1.HostedControlPlane
 		params.Image = hcp.Spec.Configuration.Image
 		params.Scheduler = hcp.Spec.Configuration.Scheduler
 	}
+	if reqInflight := hcp.Annotations[hyperv1.KubeAPIServerMaximumRequestsInFlight]; reqInflight != "" {
+		params.MaxRequestsInflight = &reqInflight
```
sjenning, line 125: Do we want to do the defaulting at this level and pass `string` vs `*string`?  None of the other `KubeAPIServerConfigParams` fields are `*string`.

rtheis (approved):
/lgtm

Thank you.  I think this is a nice addition.

control-plane-operator/controllers/hostedcontrolplane/kas/params.go
csrwng: updated

rtheis (approved):
/lgtm

PR #4495, opened 2024-08-06
OCPBUGS-36261: label routes only when HCP router used

sjenning:
Right now, we use `hostname` being set on the `Route` publishing strategy as a proxy for the HCP router being enabled.  However, there are cases where the `hostname` is set by the user, when a non-default domain is used for example, when the Routes should be admitted by the mgmt cluster ingress but are not because they are labels for the HCP router.

This PR result in the Route only being label when we know the HCP router is in use.

If the cluster is `Private` or `PublicAndPrivate`, and the `oauth` service does not have a hostname specified, the internal (privatelink) oauth name is used for the oauth endpoint `infraStatus` and the KAS `oauthMetadata` config which is how the `console` gets the oauth endpoint.

https://github.com/openshift/hypershift/pull/4495/files#diff-3d4ffda0834e09a59a4d510523932e20b9234bcb9816e40258a0fb32a87a7006R2022

### DIFF
--- a/hypershift-operator/controllers/hostedcluster/ignitionserver/ignitionserver.go
+++ b/hypershift-operator/controllers/hostedcluster/ignitionserver/ignitionserver.go
@@ -97,7 +97,7 @@ func ReconcileIgnitionServer(ctx context.Context,
 				if serviceStrategy.Route != nil {
 					hostname = serviceStrategy.Route.Hostname
 				}
-				err := reconcileExternalRoute(ignitionServerRoute, ownerRef, routeServiceName, hostname, defaultIngressDomain)
+				err := reconcileExternalRoute(ignitionServerRoute, ownerRef, routeServiceName, hostname, defaultIngressDomain, hostname != "")
 				if err != nil {
 					return fmt.Errorf("failed to reconcile external route in ignition server: %w", err)
 				}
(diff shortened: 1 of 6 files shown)


hypershift-operator/controllers/hostedcluster/ignitionserver/ignitionserver.go
```diff
@@ -97,7 +97,7 @@ func ReconcileIgnitionServer(ctx context.Context,
 				if serviceStrategy.Route != nil {
 					hostname = serviceStrategy.Route.Hostname
 				}
-				err := reconcileExternalRoute(ignitionServerRoute, ownerRef, routeServiceName, hostname, defaultIngressDomain)
+				err := reconcileExternalRoute(ignitionServerRoute, ownerRef, routeServiceName, hostname, defaultIngressDomain, hostname != "")
```
csrwng, line 100: should this also use `useHCPRouter(hcp)` ?

sjenning: This in the HO in code that we don't use any more.  I figured I would just maintain the existing behavior here rather than copy/move the `useHCPRouter` function.

sjenning:
Something about this is causing `TestCreateClusterPrivate` to fail on `console` CO not `Available` and this in the console logs
```
2024-08-07T11:14:28.351457499Z F0807 11:14:28.351406       1 authoptions.go:329] Error initializing authenticator: failed to construct OAuth endpoint cache: failed to setup an async cache - caching func returned error: context deadline exceeded
```
Makes me think console is not using the privatelink dns for the oauth endpoint.

PR #4502, opened 2024-08-07
OCPBUGS-25937: Support KMS v2 on AWS

dgrisonnet:

This PR add support for KMS v2 on AWS. New AWS clusters will only be able to use the new KMS v2 API whilst existing clusters that are already using the v1 API will not be migrated to v2 and stay on v1.

Fixes #OCPBUGS-25937

**Checklist**

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/kas/kms/aws.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/kas/kms/aws.go
@@ -26,6 +26,7 @@ const (
 	backupAWSKMSUnixSocketFileName = "awskmsbackup.sock"
 	backupAWSKMSHealthPort         = 8081
 	awsKeyNamePrefix               = "awskmskey"
+	kmsAPIVersionV1                = "v1"
 )
 
 var (
@@ -78,7 +79,7 @@ func NewAWSKMSProvider(kmsSpec *hyperv1.AWSKMSSpec, kmsImage, tokenMinterImage s
 	}, nil
 }
 
-func (p *awsKMSProvider) GenerateKMSEncryptionConfig() (*v1.EncryptionConfiguration, error) {
+func (p *awsKMSProvider) GenerateKMSEncryptionConfig(apiVersion string) (*v1.EncryptionConfiguration, error) {
 	var providerConfiguration []v1.ProviderConfiguration
 	if len(p.activeKey.ARN) == 0 {
 		return nil, fmt.Errorf("active key metadata is nil")
@@ -90,10 +91,10 @@ func (p *awsKMSProvider) GenerateKMSEncryptionConfig() (*v1.EncryptionConfigurat
 	}
 	providerConfiguration = append(providerConfiguration, v1.ProviderConfiguration{
 		KMS: &v1.KMSConfiguration{
-			Name:      fmt.Sprintf("%s-%d", awsKeyNamePrefix, hasher.Sum32()),
-			Endpoint:  activeAWSKMSUnixSocket,
-			CacheSize: ptr.To[int32](100),
-			Timeout:   &metav1.Duration{Duration: 35 * time.Second},
+			APIVersion: apiVersion,
+			Name:       fmt.Sprintf("%s-%d", awsKeyNamePrefix, hasher.Sum32()),
+			Endpoint:   activeAWSKMSUnixSocket,
+			Timeout:    &metav1.Duration{Duration: 35 * time.Second},
 		},
 	})
 	if p.backupKey != nil && len(p.backupKey.ARN) > 0 {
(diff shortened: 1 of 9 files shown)


control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -2953,6 +2953,15 @@ func (r *HostedControlPlaneReconciler) reconcileKubeAPIServer(ctx context.Contex
 			if hcp.Spec.SecretEncryption.KMS == nil {
 				return fmt.Errorf("kms metadata not specified")
 			}
+
+			// If an EncryptionConfig exists, it is passed down when generating
+			// the new EncryptionConfig to make sure that we keep the original
+			// KMS API version.
+			err := r.Get(ctx, client.ObjectKeyFromObject(encryptionConfigFile), encryptionConfigFile)
+			if err != nil && !apierrors.IsNotFound(err) {
+				return fmt.Errorf("failed to get encryption config secret: %w", err)
```
muraee, line 2962: this is not needed, `createOrUpdate()` does that automatically before calling the Reconcile function

control-plane-operator/controllers/hostedcontrolplane/kas/kms/aws.go
```diff
@@ -158,6 +166,9 @@ func (p *awsKMSProvider) ApplyKMSConfig(podSpec *corev1.PodSpec) error {
 	}
 	container.VolumeMounts = append(container.VolumeMounts,
 		awsKMSVolumeMounts.ContainerMounts(KasMainContainerName)...)
+
+	container.Args = append(container.Args, "--encryption-provider-config-automatic-reload=false")
```
muraee, line 170: what does this flag do?

control-plane-operator/controllers/hostedcontrolplane/kas/kms/ibmcloud.go
```diff
@@ -58,7 +58,7 @@ func NewIBMCloudKMSProvider(ibmCloud *hyperv1.IBMCloudKMSSpec, kmsImage string)
 	}, nil
 }
 
-func (p *ibmCloudKMSProvider) GenerateKMSEncryptionConfig() (*v1.EncryptionConfiguration, error) {
+func (p *ibmCloudKMSProvider) GenerateKMSEncryptionConfig(apiVersion string) (*v1.EncryptionConfiguration, error) {
```
muraee, line 61:
```suggestion
func (p *ibmCloudKMSProvider) GenerateKMSEncryptionConfig(_ string) (*v1.EncryptionConfiguration, error) {
```

control-plane-operator/controllers/hostedcontrolplane/kas/kms/aws.go
dgrisonnet: It creates a file watcher in kube-apiserver that watches the encryption config file and dynamically apply changes without having to restart. It was recently introduced alongside kms v2, but if I understood the code correctly whenever the encryption type changes in Hypershift, we will rebuild the kube-apiserver anyway right? So my reasoning was to add  it as a safeguard to prevent changes to the file to be applied since they would most likely not be from the operator.

FWIW it was already added to the IBM cloud ksm code https://github.com/openshift/hypershift/blob/a11f0a1b12da74f345953f9441221e4418c314f4/control-plane-operator/controllers/hostedcontrolplane/kas/kms/ibmcloud.go#L307, but I am fine removing it here since it is kind of unrelated to the KMS v2 support.

control-plane-operator/controllers/hostedcontrolplane/kas/kms/ibmcloud.go
dgrisonnet: good catch! I thought I had already replaced it, but I must have missed it.

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
muraee: no, this is only to get the object key (namespaced name). when `createOrUpdate` is called, it does a get first and updates the obj pointer (similar to what you are doing here but implicitly) then calls the reconcile function.

control-plane-operator/controllers/hostedcontrolplane/kas/kms/aws.go
muraee: I see, I am fine adding it.

dgrisonnet:
/payload-job periodic-ci-openshift-hypershift-release-4.17-periodics-e2e-aws-ovn-conformance

dgrisonnet:
/payload-job periodic-ci-openshift-hypershift-release-4.17-periodics-e2e-aws-ovn

test/e2e/util/util.go
```diff
@@ -914,62 +914,78 @@ func EnsureAPIUX(t *testing.T, ctx context.Context, hostClient crclient.Client,
...
-		}
+		ensureSecretEncryptedUsingKMS(t, ctx, hostedCluster, guestClient, "k8s:enc:kms:")
+	})
+}
 
-		restConfig, err := GetConfig()
-		if err != nil {
-			t.Errorf("failed to get restConfig; %v", err)
-		}
+func EnsureSecretEncryptedUsingKMSV1(t *testing.T, ctx context.Context, hostedCluster *hyperv1.HostedCluster, guestClient crclient.Client) {
```
muraee, line 921: it seems we are not using this or `EnsureSecretEncryptedUsingKMS` anywhere, can we remove them?

dgrisonnet: I was considering keeping it around to write a test that would verify that existing KMS v1 installations are still using v1, but since I am low on time with the freeze, I was considering doing that in a separate PR.

PR #4503, opened 2024-08-07
OCPBUGS-24400: Reconcile over an ETCD failing member

jparrill:
Adds code to detect whether etcd needs fixing and to determine whether
it is in a state that can be recovered. If it can be recovered, it
performs recovery by adding back a missing member and deleting the
corresponding pod and pvc.

Adds a flag to the install command to disable etcd recovery for
consumers that manage their own etcd and do not need the additional
watches in the hypershift operator.

Fixes #[OCPBUGS-24400](https://issues.redhat.com/browse/OCPBUGS-24400)


hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -2325,6 +2332,85 @@ func (r *HostedClusterReconciler) reconcileAutoscaler(ctx context.Context, creat
...
+	}
+
+	if err := r.Delete(ctx, etcdPV); err != nil {
+		return fmt.Errorf("failed to delete etcd persistent volume: %w", err)
+	}
+
+	if err := r.Delete(ctx, &failingPods[0]); err != nil {
+		return fmt.Errorf("failed to delete failing etcd pod: %w", err)
+	}
+
```
devguyio, line 2410: how about creating an `Event` here with some summary of the deletion as a more durable trace of this operation taking place?

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -2325,6 +2332,85 @@ func (r *HostedClusterReconciler) reconcileAutoscaler(ctx context.Context, creat
...
+	etcdPVC := &corev1.PersistentVolumeClaim{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      etcdPV.Spec.ClaimRef.Name,
+			Namespace: hcpNS,
+		},
+	}
+
+	if err := r.Get(ctx, client.ObjectKeyFromObject(etcdPVC), etcdPVC); err != nil {
+		return fmt.Errorf("failed to get etcd persistent volume claim: %w", err)
```
devguyio, line 2396: I'm trying to think of there are cases when we fail to get the PVC or the PV, yet we would still like to proceed and attempt the pod deletion?

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -1076,6 +1077,12 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
 		return ctrl.Result{RequeueAfter: duration}, nil
 	}
 
+	if _, exists := hcluster.Annotations[ETCDRestoreFailingMemberAnnotation]; exists {
```
devguyio, line 1080: should we use the annotation value to signal the intention / state? 

e.g.:
```
"hypershift.openshift.io/etcd-restore-failing-member" : "true"  # SRE sets this to indicate the desire for the automation to kick in.
"hypershift.openshift.io/etcd-restore-failing-member" : "error"  # Set by the HO to indicate to SRE that this etcd recovery failed
"hypershift.openshift.io/etcd-restore-failing-member" : "done_20240803160405"  # Set by the HO to indicate that it succeeded at the shown timestamp (YYYYMMDDhhmmss)
```

I think this makes this operation more observable and traceable.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
devguyio: This would be still available even if SD removed the annotation

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -2325,6 +2325,67 @@ func (r *HostedClusterReconciler) reconcileAutoscaler(ctx context.Context, creat
 	return autoscaler.ReconcileAutoscaler(ctx, r.Client, hcp, clusterAutoscalerImage, utilitiesImage, createOrUpdate, r.SetDefaultSecurityContext, config.MutatingOwnerRefFromHCP(hcp, releaseVersion))
 }
 
+func (r *HostedClusterReconciler) reconcileETCDMemberRecovery(ctx context.Context, hcluster *hyperv1.HostedCluster, createOrUpdate upsert.CreateOrUpdateFN) error {
```
sjenning, line 2328: Why not do this in the CPO?  we really don't want the HO deploying things to the HCP beyond CAPI and CPO.  Also, this function isn't called anywhere (cause of the `verify` failure)

cmd/infra/aws/delegating_client.go
```diff
@@ -593,3 +594,5 @@ func (c *s3Client) PutObjectWithContext(ctx aws.Context, input *s3.PutObjectInpu
 func (c *s3Client) PutPublicAccessBlockWithContext(ctx aws.Context, input *s3.PutPublicAccessBlockInput, opts ...request.Option) (*s3.PutPublicAccessBlockOutput, error) {
 	return c.openshiftImageRegistry.s3Client.PutPublicAccessBlockWithContext(ctx, input, opts...)
 }
+
```
sjenning, line 597: Nothing but whitespace changes for this whole file

etcd-recovery/etcdrecovery.go
```diff
@@ -0,0 +1,397 @@
...
+
+	if failingEndpoint != nil || failingPod != nil {
+		return fmt.Errorf("etcd cluster is unhealthy: %v", failingEndpoint), failingEndpoint, failingPod, kubeClient
+	}
+
+	fmt.Println("no etcd pods are failing, and no etcd endpoints are in error state, skipping recovery procedure")
+	return nil, nil, nil, nil
+}
+
+func runRecovery(ctx context.Context, failingEndpoint endpointStatus, failingPod *corev1.Pod, k crclient.Client) error {
```
sjenning, line 162: This is a completely transparent wrapper for `recoverEtcd`

cmd/infra/aws/delegating_client.go
jparrill: This is my IDE's fault xD, sorry about that.

etcd-recovery/etcdrecovery.go
jparrill: Make sense, let me fix it.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
jparrill: Make sense the event or log entry thing or even maybe modify the ETCD status.condition.
I've changed the approach from a manual annotation to an automatic procedure with many verifications in the middle to try avoiding false positives.

etcd-recovery/etcdrecovery.go
```diff
@@ -0,0 +1,391 @@
...
+		return fmt.Errorf("failed to delete etcd persistent volume claim: %w", err)
+	}
+
+	if err := k.Delete(ctx, etcdPV); err != nil {
+		return fmt.Errorf("failed to delete etcd persistent volume: %w", err)
+	}
+
+	if err := k.Delete(ctx, &failingPod); err != nil {
+		return fmt.Errorf("failed to delete failing etcd pod: %w", err)
+	}
```
typeid, line 340: @jparrill as an FYI, we've seen cases where these steps are not enough to recover the etcd member of an HCP stuck in installing. A full re-creation of the statefulset/PVs for all 3 etcd /PVCs for all 3 etcd was needed.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -1455,6 +1460,13 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
 		}
 	}
 
+	// Reconcile the ETCD Recovery member
+	if hcluster.Spec.Etcd.ManagementType == hyperv1.Managed {
+		if err := r.reconcileETCDMemberRecovery(ctx, hcluster, createOrUpdate); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to reconcile etcd recovery member: %w", err)
+		}
+	}
```
typeid, line 1468: The PR description still mentions an `hypershift.openshift.io/etcd-restore-failing-member` annotation, did this change in this iteration of this PR? I don't see the annotation anymore :)

csrwng:
/retitle "OCPBUGS-24400: Reconcile over an ETCD failing member"

csrwng:
/retitle OCPBUGS-24400: Reconcile over an ETCD failing member

etcd-recovery/etcdrecovery.go
```diff
@@ -0,0 +1,391 @@
...
+	if failingEndpoint != nil || failingPod != nil {
+		return fmt.Errorf("etcd cluster is unhealthy: %v", failingEndpoint), failingEndpoint, failingPod, kubeClient
+	}
+
+	fmt.Println("no etcd pods are failing, and no etcd endpoints are in error state, skipping recovery procedure")
+	return nil, nil, nil, nil
+}
+
+func checkEtcdStatus(ctx context.Context, opts options, etcdClient *etcdclient.Client, k crclient.Client) (error, *endpointStatus, *corev1.Pod) {
+	//timeoutContext, cancel := context.WithTimeout(ctx, defaultEtcdClientTimeout)
```
bryan-cox: nit: these should be removed

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -2328,6 +2340,228 @@ func (r *HostedClusterReconciler) reconcileAutoscaler(ctx context.Context, creat
...
+	if err := r.cleanupEtcdRecoveryObjects(ctx, hcluster); err != nil {
+		return fmt.Errorf("failed to cleanup etcd recovery job: %w", err)
+	}
+
+	return nil
+}
+
+// checkEtcdRecoveryJobCompletion checks the status of the ETCD recovery job and returns 3 booleans:
+// - The first boolean indicates if the job exists or not.
+// - The second boolean indicates if the job has finished or not.
```
bryan-cox: Should this really be the job failed?

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -2328,6 +2340,228 @@ func (r *HostedClusterReconciler) reconcileAutoscaler(ctx context.Context, creat
...
+}
+
+func (r *HostedClusterReconciler) reconcileEtcdRecoveryJob(job *batchv1.Job, recoverySA *corev1.ServiceAccount, etcdEndpoints string) error {
+	job.Spec = batchv1.JobSpec{
+		Template: corev1.PodTemplateSpec{
+			Spec: corev1.PodSpec{
+				Containers: []corev1.Container{
+					{
+						Name:            "etcd-recovery",
+						Image:           "quay.io/jparrill/hypershift:OCPBUGS-24400v3-3",
```
bryan-cox: Should this be updated?

bryan-cox (commented):
Should we also not have upstream docs about this new functionality?

cmd/install/install.go
```diff
@@ -712,6 +714,7 @@ func hyperShiftOperatorManifests(opts Options) ([]crclient.Object, []crclient.Ob
 		EnableDedicatedRequestServingIsolation:  opts.EnableDedicatedRequestServingIsolation,
 		ManagedService:                          opts.ManagedService,
 		EnableSizeTagging:                       opts.EnableSizeTagging,
+		DisableEtcdRecovery:                     opts.DisableEtcdRecovery,
```
enxebre, line 717: All other flags are Enable* any reason to not be consistent?

etcd-recovery/etcdrecovery.go
```diff
@@ -0,0 +1,538 @@
...
+type options struct {
+	etcdClientCertFile string
+	etcdClientKeyFile  string
+	etcdCAFile         string
+	namespace          string
+}
+
+func NewRecoveryCommand() *cobra.Command {
+	opts := options{
+		etcdClientCertFile: "/etc/etcd/tls/client/etcd-client.crt",
```
enxebre, line 50: When would this change so they need to be configurable?

etcd-recovery/etcdrecovery.go
```diff
@@ -0,0 +1,538 @@
...
+	missingMember *string
+	recoverable   bool
+}
+
+func (s *etcdStatus) isUnhealthy() bool {
+	return s.missingMember != nil || s.failingPod != nil
+}
+
+func (s *etcdStatus) isRecoverable() bool {
+	recoverable := s.recoverable
```
enxebre, line 117: can we add a unit test?

etcd-recovery/etcdrecovery.go
```diff
@@ -0,0 +1,538 @@
...
+	if s.failingPod != nil && s.missingMember != nil {
+		if *s.failingPod != *s.missingMember {
+			recoverable = false
+		}
+	}
+	return recoverable
+}
+
+func (s *etcdStatus) memberToRecover() string {
```
enxebre, line 126: can we add a unit test?

etcd-recovery/etcdrecovery.go
```diff
@@ -0,0 +1,538 @@
...
+	if !s.isRecoverable() {
+		return ""
+	}
+	if s.failingPod != nil {
```
enxebre, line 130: why this one takes precedence?

etcd-recovery/etcdrecovery.go
```diff
@@ -0,0 +1,538 @@
...
+	timeOutCtx, cancel := context.WithTimeout(ctx, healthyClusterWait)
+	defer cancel()
+	return wait.PollUntilContextCancel(timeOutCtx, 30*time.Second, false, func(ctx context.Context) (done bool, err error) {
+		status, err := runStatus(ctx, opts)
+		if err != nil {
+			return false, err
+		}
+		isHealthy := !status.isUnhealthy()
+		if isHealthy {
+			log.Info("Cluster is now healthy, finished recovery job")
```
enxebre, line 213: really nit pick but logging "finished recovery job" seems out of scope of this func. It is making a non guaranteed assumption about the invoker

etcd-recovery/etcdrecovery.go
```diff
@@ -0,0 +1,538 @@
...
+
+	return checkEtcdStatus(ctx, opts, eclient, kclient)
+}
+
+func missingMember(members []*etcdserverpb.Member) *string {
+	names := sets.New[string]()
+	for _, m := range members {
+		names.Insert(m.Name)
+	}
+	for i := 0; i < 3; i++ {
```
enxebre, line 242: what happens when etcd is single replica?

etcd-recovery/etcdrecovery.go
```diff
@@ -0,0 +1,538 @@
...
+	// Determine member health by looking at their endpoint health
+	for _, member := range etcdMemberList.Members {
+		if len(member.ClientURLs) != 1 {
+			return nil, fmt.Errorf("member %s does not have an expected client URL: %v", member.Name, member.ClientURLs)
+		}
+		clientURL := member.ClientURLs[0]
+		memberHealth[member.Name] = isEndpointHealthy(ctx, opts, member.Name, clientURL)
+	}
+
+	switch len(memberHealth) {
```
enxebre, line 288: can we unit test the recoverable logic?

enxebre:
@csrwng can we update the PR desc to reflect current state?

cmd/install/install.go
```diff
@@ -230,6 +231,7 @@ func NewCommand() *cobra.Command {
 	cmd.PersistentFlags().StringVar(&opts.PullSecretFile, "pull-secret", opts.PullSecretFile, "File path to a pull secret.")
 	cmd.PersistentFlags().StringVar(&opts.ManagedService, "managed-service", opts.ManagedService, "The type of managed service the HyperShift Operator is installed on; this is used to configure different HostedCluster options depending on the managed service. Examples: ARO-HCP, ROSA-HCP")
 	cmd.PersistentFlags().BoolVar(&opts.EnableSizeTagging, "enable-size-tagging", opts.EnableSizeTagging, "If true, HyperShift will tag the HostedCluster with a size label corresponding to the number of worker nodes")
+	cmd.PersistentFlags().BoolVar(&opts.DisableEtcdRecovery, "disable-etcd-recovery", opts.DisableEtcdRecovery, "If true, HyperShift operator will not attempt to recover broken etcd clusters")
```
enxebre, line 234: nit: "will not attempt to recover broken etcd clusters" broken clusters makes me thing of cluster where quorum is broken, can we clarify that's not the case and this is only for failing members of working clusters in the desc of the flag?

enxebre: Why is this a install flag? How can I opt out for a single cluster? What about single replica clusters?

etcd-recovery/etcdrecovery.go
```diff
@@ -0,0 +1,538 @@
...
+func NewRunCommand(opts *options) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:          "run",
+		Short:        "Recover etcd cluster that has quorum",
+		SilenceUsage: true,
+		Run: func(cmd *cobra.Command, args []string) {
+			ctx, cancel := signal.NotifyContext(cmd.Context(), os.Interrupt, syscall.SIGINT)
+			defer cancel()
+			ctx = setupCmdContext(ctx)
+			err := runRecovery(ctx, *opts)
```
enxebre, line 87: run recovery might fail for transient issues or legitimately, in any case we fail here which kill the pod and hide logs?

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -2328,6 +2360,311 @@ func (r *HostedClusterReconciler) reconcileAutoscaler(ctx context.Context, creat
...
+	hcpNS := manifests.HostedControlPlaneNamespace(hcluster.Namespace, hcluster.Name)
+
+	// Preliminary checks to see if we need to proceed with the recovery procedure
+	if hcluster.Spec.ControllerAvailabilityPolicy != hyperv1.HighlyAvailable {
+		return nil, nil
+	}
+
+	// Check the recovery job
+	recoveryJob := etcdrecoverymanifests.EtcdRecoveryJob(hcpNS)
+	jobStatus, err := r.etcdRecoveryJobStatus(ctx, recoveryJob)
```
enxebre, line 2380: why is this a job vs a long running deployment which works in loop?

etcd-recovery/etcdrecovery.go
csrwng: The problem is that we have no good way of recovering from a missing member. The init container that re-adds the member only does it if the member already exists.  The new recovery code adds the member before deleting pod/pvc and that causes the recovery to succeed.

etcd-recovery/etcdrecovery.go
```diff
@@ -0,0 +1,538 @@
...
+
+	return nil
+}
+
+func waitForHealthyCluster(ctx context.Context, opts options) error {
+	log := ctrl.LoggerFrom(ctx)
+	log.Info("Waiting for etcd cluster to become healthy")
+	timeOutCtx, cancel := context.WithTimeout(ctx, healthyClusterWait)
+	defer cancel()
+	return wait.PollUntilContextCancel(timeOutCtx, 30*time.Second, false, func(ctx context.Context) (done bool, err error) {
```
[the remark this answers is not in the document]
csrwng: healthyClusterWait is set at 5min (interval is 30s). Yes, I think that should be enough time. If not, then we exit with an error and another job pod is launched which should see that all members are healthy and finish with success.

etcd-recovery/etcdrecovery.go
csrwng: It allows testing of the command independently from a local machine (which I had to do many times 😅)

etcd-recovery/etcdrecovery.go
csrwng: we don't run this recovery

cmd/install/install.go
csrwng: > How can I opt out for a single cluster to prefer manual intervention? What about single replica clusters?

I can introduce another knob (maybe an annotation). However, I couldn't think of a use case for it.

csrwng: > What about single replica clusters?

We only do the health check/recovery for ha clusters

etcd-recovery/etcdrecovery.go
csrwng: The job pod terminates and a new one is launched. All job pods remain around until we exhaust retries (10) or one succeeds.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
csrwng: The reason this is done from the HO is that we want to be able to recover even older clusters that do not necessarily have the latest CPO (it's a managed service requirement)
There's a few benefits for running a job:
- we're not adding a new permanent workload to control planes which can impose new resource requirements per cp
- a failed job and associated pods makes it easier to 1) understand that recovery was tried, 2) look at the logs of each attempt to understand why it failed. It's not as obvious from a controller to understand what happened unless you look at the controller logs.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -2328,6 +2360,311 @@ func (r *HostedClusterReconciler) reconcileAutoscaler(ctx context.Context, creat
...
+	}
+
+	if jobStatus.exists {
+		if !jobStatus.finished {
+			log.Info("waiting for etcd recovery job to complete")
+			return nil, nil
+		}
+
+		if !jobStatus.successful {
+			// TODO: create alert to trigger manual intervention in this case
```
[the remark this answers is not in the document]
csrwng: https://issues.redhat.com/browse/HOSTEDCP-1940

etcd-recovery/etcdrecovery.go
csrwng: we just need *a* name... at this point we know that if both failingPod and failingMember are set, they have to match (otherwise isRecoverable returned false).

hypershift-operator/controllers/hostedcluster/etcd_recovery.go
```diff
@@ -0,0 +1,330 @@
...
+	}
+
+	if jobStatus.exists {
+		if !jobStatus.finished {
+			log.Info("waiting for etcd recovery job to complete")
+			return nil, nil
+		}
+
+		if !jobStatus.successful {
+			// TODO: create alert to trigger manual intervention in this case
```
bryan-cox, line 52: nit: suggest to add HOSTEDCP-1940 here

bryan-cox:
lgtm; happy to tag it as so but I figured we might want to let others weigh again

jparrill:
Looks great to me. I've been testing in manually and it behaves as expected.
/lgtm

PR #4513, opened 2024-08-08
OSASINFRA-3553: relax PodAffinityTerm for zone spreading in HA

EmilienM:

We don't know whether an OpenStack cloud is deployed across multiple availability zones or not.
So we can't require the scheduling of HA pods across multiple zones.
For the OpenStack platform, let's relax that rule and allow to use the
`PodAntiAffinity.PreferredDuringSchedulingIgnoredDuringExecution` instead.
So now in HA mode, pods will be scheduled on multiple zones by preference but if it's not possible
they'll just be spread across nodes which is fine for us now.

Note: we also apply this change to the Kubevirt platform since this was explicitly asked in the review.

### DIFF
--- a/support/config/deployment.go
+++ b/support/config/deployment.go
@@ -174,7 +174,9 @@ func colocationLabelValue(hcp *hyperv1.HostedControlPlane) string {
 
 // setMultizoneSpread sets PodAntiAffinity with corev1.LabelTopologyZone as the topology key for a given set of labels.
 // This is useful to e.g ensure pods are spread across availavility zones.
-func (c *DeploymentConfig) setMultizoneSpread(labels map[string]string) {
+// If required is true, the rule is set as RequiredDuringSchedulingIgnoredDuringExecution, otherwise it is set as
+// PreferredDuringSchedulingIgnoredDuringExecution.
+func (c *DeploymentConfig) setMultizoneSpread(labels map[string]string, required bool) {
 	if labels == nil {
 		return
 	}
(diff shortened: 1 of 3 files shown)


support/config/deployment.go
```diff
@@ -184,14 +186,28 @@ func (c *DeploymentConfig) setMultizoneSpread(labels map[string]string) {
...
+				LabelSelector: &metav1.LabelSelector{
+					MatchLabels: labels,
+				},
 			},
-		},
-	)
+		)
+	} else {
+		c.Scheduling.Affinity.PodAntiAffinity.PreferredDuringSchedulingIgnoredDuringExecution = append(c.Scheduling.Affinity.PodAntiAffinity.PreferredDuringSchedulingIgnoredDuringExecution,
+			corev1.WeightedPodAffinityTerm{
```
EmilienM, line 200: This didn't work for me, I need to investigate further.

EmilienM: I made a typo that I fixed. This should be fine now.

EmilienM:
/cc davidvossel
for review, please

support/config/deployment.go
```diff
@@ -358,7 +374,12 @@ func (c *DeploymentConfig) setLocation(hcp *hyperv1.HostedControlPlane, multiZon
 	c.setColocation(hcp)
 	// TODO (alberto): pass labels with deployment hash and set this unconditionally so we don't skew setup.
 	if c.Replicas > 1 {
-		c.setMultizoneSpread(multiZoneSpreadLabels)
+		multiZoneRequired := true
+		// On OpenStack, we can't guarantee that the cloud is configured with multiple availability zones.
+		if hcp.Spec.Platform.Type == hyperv1.OpenStackPlatform {
+			multiZoneRequired = false
+		}
```
davidvossel, line 381: Let's add KubeVirt to this as well.

The reality is that the required zone spread has caused us (kubevirt platform) more trouble than it has benefited people. I think setting preferred during scheduling for zone labeling is a happy medium for us as well.

EmilienM:
@davidvossel updated. Let me know if you want me to add docs for Kubevirt as well.

davidvossel (commented):
/lgtm
/approve

EmilienM:
bot is stuck and doesn't merge this PR
/test e2e-aws e2e-kubevirt-aws-ovn-reduced

PR #4515, opened 2024-08-08
OSASINFRA-3565: openstack: additional ports

EmilienM:

Why:
On OpenStack, allow to create nodepools with additional ports to solve 4 use cases:
* Allow NFV nodepools, connected to the SRIOV/DPDK networks through vnic type = direct
* Allow to configure Address Pairs, for external loadbalancers like MetalLB
* Allow to connect the workloads on storage networks (e.g. Manila CSI in the future).
* Running workloads on provider networks (e.g. dual stack, etc).

How:
In Nodepool API, we now have the possibility to specify one or multiple additional ports that should be attached to the nodes.

Example:
```
--openstack-node-additional-port network-id:40a355cb-596d-495c-8766-419d98cadd57,vnic-type:direct
```
 The "network-id" field should point to an existing neutron network ID and the "vnic-type" is the type of the port to create, it can be specified multiple times to attach to multiple ports.
More examples are given in the doc.

It will control how the CAPI provider creates the Machine and add additional ports.

Note: 2 commits: one for API and one for the rest (controllers, doc, unit, e2e).


cmd/nodepool/openstack/create.go
```diff
@@ -117,8 +153,37 @@ func (o *OpenStackPlatformCreateOptions) Type() hyperv1.PlatformType {
...
+	}
+	return nil
+}
+
+func strToBool(s string) bool {
+	return s == "true"
+}
+
+func getAddressPairs(addressPairs string) []hyperv1.AddressPair {
+	separetedAddressPairs := strings.Split(addressPairs, "-")
```
EmilienM, line 183: this is the only separator that I could make work and thought of that wasn't that bad...

EmilienM:
We need https://github.com/openshift/release/pull/57410 to run the e2e tests that I started to write in this PR.

control-plane-operator/controllers/hostedcontrolplane/cloud/openstack/providerconfig.go
```diff
@@ -61,10 +63,21 @@ func getCloudConfig(externalNetworkID *string, cloudName string, credentialsSecr
 	if externalNetworkID != nil && *externalNetworkID != "" {
 		config += "floating-network-id = " + *externalNetworkID + "\n"
 	}
+	config += "\n[Networking]\n"
+	config += "address-sort-order = " + machineNetworksToList(machineNetwork) + "\n"
```
EmilienM, line 67: @MaysaMacedo @mdbooth this is the simplest way I found to do it without having to do surgery in CPO.

Matt I remember you wanted to change CPO to find out "primary" NIC with Neutron tags. While we can investigate that, I think for now doing address-sort-order in CPO isn't a bad thing given we're not even Tech Preview and it's something we can easily (IMO) iterate later.

cmd/nodepool/openstack/create.go
```diff
@@ -3,8 +3,10 @@ package openstack
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
+	"github.com/openshift/hypershift/cmd/cluster/kubevirt/params"
```
mdbooth, line 9: This suggests we should talk to the kubevirt folks about moving this package into a non-kubevirt library.

cc @davidvossel

cmd/nodepool/openstack/create.go
```diff
@@ -65,6 +95,10 @@ func (o *RawOpenStackPlatformCreateOptions) Validate() (*ValidatedOpenStackPlatf
 		return nil, fmt.Errorf("image name is required")
 	}
 
+	if err := validateAdditionalPorts(o.AdditionalPorts); err != nil {
+		return nil, err
+	}
```
mdbooth, line 100: Notice how by doing this both here and in Complete() we end up parsing the arguments twice.

Why don't we parse them just once, here, and replace AdditionalPorts in `ValidatedOpenStackPlatformCreateOptions` with a slice of parsed structs. Then we don't have to do it again in Complete(), we can just use the validated result.

cmd/cluster/openstack/create.go
```diff
@@ -172,6 +174,11 @@ func (o *RawCreateOptions) ApplyPlatformSpecifics(cluster *hyperv1.HostedCluster
 
 	cluster.Spec.Services = core.GetIngressServicePublishingStrategyMapping(cluster.Spec.Networking.NetworkType, false)
 
+	// MachineNetwork has no default in Hypershift but it's convenient to have one for OpenStack:
+	// * To specifiy the subnet that CAPO will manage.
+	// * To inform CCM (via cloud-provider-openstack) the prefered subnet for internal NodeIPs.
+	cluster.Spec.Networking.MachineNetwork = []hyperv1.MachineNetworkEntry{{CIDR: *ipnet.MustParseCIDR(openstackutil.DefaultMachineNetwork)}}
```
mdbooth, line 180: Presumably we don't want to overwrite this if the user set it?

control-plane-operator/controllers/hostedcontrolplane/cloud/openstack/providerconfig.go
mdbooth: Ack. As previously discussed I'm not a fan of this solution, but it's fine for now as you say.

hypershift-operator/controllers/hostedcluster/internal/platform/openstack/openstack.go
```diff
@@ -77,7 +77,7 @@ func reconcileOpenStackClusterSpec(hcluster *hyperv1.HostedCluster, openStackClu
 	// Note: The default is required for now because there is no CLI option to set the MachineNetwork.
 	// See https://github.com/openshift/hypershift/pull/4287
 	if len(hcluster.Spec.Networking.MachineNetwork) == 0 {
-		machineNetworks = []hyperv1.MachineNetworkEntry{{CIDR: *ipnet.MustParseCIDR(defaultCIDRBlock)}}
+		machineNetworks = []hyperv1.MachineNetworkEntry{{CIDR: *ipnet.MustParseCIDR(openstackutil.DefaultMachineNetwork)}}
```
mdbooth, line 80: Didn't we already default this elsewhere? Seems dangerous to do it twice.

test/e2e/nodepool_osp_multinet_test.go
```diff
@@ -0,0 +1,138 @@
...
+			return oms, err
+		},
+		[]e2eutil.Predicate[[]*capiopenstack.OpenStackMachine]{
+			func(machines []*capiopenstack.OpenStackMachine) (done bool, reasons string, err error) {
+				return len(machines) == int(*nodePool.Spec.Replicas), fmt.Sprintf("expected %d OpenStackMachines, got %d", *nodePool.Spec.Replicas, len(machines)), nil
+			},
+		},
+		[]e2eutil.Predicate[*capiopenstack.OpenStackMachine]{
+			func(machine *capiopenstack.OpenStackMachine) (done bool, reasons string, err error) {
+				// The number of ports for a machine is 1 + the number of additional ports in the nodepool.
```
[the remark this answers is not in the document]
mdbooth: I think this is fine? The ports should be copied to the OpenStackServer as required? If it's not fine it means we broken the CAPO api really badly.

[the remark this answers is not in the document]
EmilienM: ah you're right! https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/api/v1beta1/openstackmachine_types.go#L212

hypershift-operator/controllers/hostedcluster/internal/platform/openstack/openstack.go
EmilienM: Right, we do it in the CLI when creating the cluster. I'll delete it here then.

mdbooth: Hmm. I wonder if it would be safer to do it here The question is: do we want the defaulting behaviour to be the same when you're using the CLI as when you're not using the CLI? Probably?

EmilienM: * CLI: this PR will provide a default
* non-CLI: we assume that HostedCluster's YAML has it and if not we'll return an error.

I don't think it'll be accepted to provide a default directly in the API.

What do you think?

EmilienM:
/test e2e-openstack
just rebased on main.

EmilienM:
/test e2e-openstack
flake on upgrade? https://issues.redhat.com/browse/OCPBUGS-43802

EmilienM:
`e2e-openstack` job failed because of https://issues.redhat.com/browse/OCPBUGS-43802 but the new test passed.
This is ready for review.

EmilienM:
/test e2e-openstack
should pass now

hypershift-operator/controllers/nodepool/openstack/openstack.go
```diff
@@ -27,5 +28,47 @@ func MachineTemplateSpec(hcluster *hyperv1.HostedCluster, nodePool *hyperv1.Node
...
+			if port.Network != nil {
+				additionalPorts[i].Network = &capiopenstackv1beta1.NetworkParam{}
+				if port.Network.Filter != nil {
+					additionalPorts[i].Network.Filter = openstackutil.CreateCAPONetworkFilter(port.Network.Filter)
+				}
+				if port.Network.ID != nil {
+					additionalPorts[i].Network.ID = port.Network.ID
+				}
+			}
+			if port.Description != "" {
```
MaysaMacedo, line 53: nit port's description is always set at line 45, this check is not needed

EmilienM: thanks

hypershift-operator/controllers/nodepool/openstack/openstack.go
```diff
@@ -27,5 +28,40 @@ func MachineTemplateSpec(hcluster *hyperv1.HostedCluster, nodePool *hyperv1.Node
 		return nil, fmt.Errorf("image name is required")
 	}
 
+	// TODO: add support for BYO network/subnet
+	if len(hcluster.Spec.Platform.OpenStack.Subnets) == 0 && len(nodePool.Spec.Platform.OpenStack.AdditionalPorts) > 0 {
+		// Initialize the ports slice with an empty port which will be used as the primary port.
+		// CAPO will figure out the network and subnet for this port since tey are not provided.
+		ports := []capiopenstackv1beta1.PortOpts{{}}
```
MaysaMacedo, line 35: I gave a try to this changed, so it the following spec expected for the openstackmachine?
```
  spec:
    flavor: m1.xlarge
    image:
      filter:
        name: rhcos-4.17
    ports:
    - {}
    - description: Additional port for Hypershift node pool may
      network:
        id: 252c0fc3-7937-402d-81ee-dd515ee93edc
    providerID: openstack:///376a0ce0-a77a-42ee-a9ce-c114b40b2e32
```

I used the following command:

```
		hypershift create cluster openstack \
		--openstack-credentials-file foo \
		--openstack-external-network-id "14c15d33-175c-424e-88ba-361a875e0c5c" \
		--base-domain shiftstack-dev.devcluster.openshift.com \
		--name foo \
		--ssh-key foo \
		--pull-secret foo \
		--openstack-node-flavor m1.xlarge \
		--openstack-node-image-name rhcos-4.17 \
		--release-image "registry.build09.ci.openshift.org/ci-ln-q4sl582/release:latest" \
		--machine-cidr 192.168.25.0/24 \
		--node-pool-replicas 1 \
		--openstack-node-additional-port network-id:252c0fc3-7937-402d-81ee-dd515ee93edc
```

EmilienM: yeah, if you show the `OpenStackServer.Status` or the `OpenStackMachine.Status`, you would see the `ResolvedPortSpec` with the correct ports spec, including the network ID on the first port.
one improvement we could make in testing is checking that ^ around these lines: https://github.com/openshift/hypershift/pull/4515/files#diff-eceb97b077faccef260cf3513d16f8bfca128b139574e277e8dc97b4c19923ffR82

EmilienM: CI logs demonstrate it at least: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_hypershift/4515/pull-ci-openshift-hypershift-main-e2e-openstack/1853892824387293184/artifacts/e2e-openstack/hypershift-openstack-e2e-execute/artifacts/TestNodePool_HostedCluster0/namespaces/e2e-clusters-84xlc-example-p9wsj/infrastructure.cluster.x-k8s.io/openstackmachines/example-p9wsj-test-osp-multinet-2xwxj-mfx6r.yaml

EmilienM: Btw we do that because we can't really predict the network ID that CAPO will create, so we just put an empty slice and CAPO has code that for each port, it'll use the default network that it created in the port spec if this wasn't provided in the PortOpts.

support/openstackutil/constants.go
bryan-cox, line 1: We already have a constants file - /support/config/constants.go. Could you make a var() section for OpenShift constants there and remove this file?

support/openstackutil/conversion.go
bryan-cox, line 1: Could we get some unit tests for these functions?

bryan-cox (changes requested):

api/hypershift/v1beta1/openstack.go
```diff
@@ -333,3 +338,54 @@ type SubnetFilter struct {
...
+	// attached to. This is used to determine which mechanism driver(s) to
+	// be used to bind the port. The valid values are normal, macvtap,
+	// direct, baremetal, direct-physical, virtio-forwarder, smart-nic and
+	// remote-managed, although these values will not be validated in this
+	// API to ensure compatibility with future neutron changes or custom
+	// implementations. What type of vNIC is actually available depends on
+	// deployments. If not specified, the Neutron default value is used.
+	//
+	// +optional
+	VNICType string `json:"vnicType,omitempty"`
```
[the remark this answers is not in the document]
EmilienM: In the OpenStack API, this represents the type of vNIC which this port should be attached to. This is used to determine which mechanism driver(s) to be used to bind the port. The valid values are normal, macvtap, direct, baremetal, direct-physical, virtio-forwarder, smart-nic and remote-managed but the list is likely to be updated at some point and there is no validation for its value in our SDK (Gophercloud). So I don't think we want to add any here.

api/hypershift/v1beta1/openstack.go
```diff
@@ -333,3 +338,54 @@ type SubnetFilter struct {
...
+
+	// DisablePortSecurity enables or disables the port security when set.
+	// When not set, it takes the value of the corresponding field at the network level.
+	//
+	// +optional
+	DisablePortSecurity bool `json:"disablePortSecurity,omitempty"`
```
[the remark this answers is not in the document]
EmilienM: Yeah, I've seen it before :confused:. I've copied it from our CAPI provider:

```go
// DisablePortSecurity enables or disables the port security when set.
// When not set, it takes the value of the corresponding field at the network level.
// +optional
DisablePortSecurity *bool `json:"disablePortSecurity,omitempty"`
```

https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/7f7f0f0fd51307bd9e87925212d6c672f625dfbd/api/v1beta1/types.go#L377-L380

But if that's problematic for OpenShift, we can use enumeration of values.

cc @mdbooth for your opinion as well here.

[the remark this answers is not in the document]
mdbooth: It's fine. We can make this:
* Enabled
* Disabled
* OpenStackDefault (or nil)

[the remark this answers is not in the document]
EmilienM: Works for me, I'll make the change.

cmd/util/params.go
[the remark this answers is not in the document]
EmilienM: I'll push it in the next commit.

control-plane-operator/controllers/hostedcontrolplane/cloud/openstack/providerconfig.go
```diff
@@ -63,10 +65,21 @@ func getCloudConfig(externalNetworkID *string, cloudName string, credentialsSecr
...
 
 	return config
 }
 
+// machineNetworksToList converts a list of MachineNetworkEntry to a comma separated list of CIDRs.
+func machineNetworksToList(machineNetwork []hyperv1.MachineNetworkEntry) string {
```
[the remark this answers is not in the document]
EmilienM: I moved the function into util and added a test :+1:

control-plane-operator/controllers/hostedcontrolplane/cloud/openstack/providerconfig.go
```diff
@@ -48,7 +50,7 @@ func ReconcileCloudConfigConfigMap(externalNetworkID *string, cm *corev1.ConfigM
 }
 
 // getCloudConfig returns the cloud config.
-func getCloudConfig(externalNetworkID *string, cloudName string, credentialsSecret *corev1.Secret, caCertData []byte) string {
+func getCloudConfig(externalNetworkID *string, cloudName string, credentialsSecret *corev1.Secret, caCertData []byte, machineNetwork []hyperv1.MachineNetworkEntry) string {
```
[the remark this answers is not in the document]
EmilienM: :+1:

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -2703,14 +2703,14 @@ func (r *HostedControlPlaneReconciler) reconcileCloudProviderConfig(ctx context.
 
 		cfg := manifests.OpenStackProviderConfig(hcp.Namespace)
 		if _, err := createOrUpdate(ctx, r, cfg, func() error {
-			return openstack.ReconcileCloudConfigConfigMap(hcp.Spec.Platform.OpenStack.ExternalNetwork.ID, cfg, hcp.Spec.Platform.OpenStack.IdentityRef.CloudName, credentialsSecret, caCertData)
+			return openstack.ReconcileCloudConfigConfigMap(hcp.Spec.Platform.OpenStack.ExternalNetwork.ID, cfg, hcp.Spec.Platform.OpenStack.IdentityRef.CloudName, credentialsSecret, caCertData, hcp.Spec.Networking.MachineNetwork)
```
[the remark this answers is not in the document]
EmilienM: This has been done in v2:

https://github.com/openshift/hypershift/blob/17a958fe04edb8ca2ce45c84553ca95a773246df/control-plane-operator/controllers/hostedcontrolplane/v2/cloud_controller_manager/openstack/config.go#L73

Which is supposed to be transitioned in 4.19. I'm ok to make that change but this will involve other changes not related to this PR. WDYT?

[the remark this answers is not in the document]
EmilienM: This has been done in hostedcontrolplane v2 I believe:

https://github.com/openshift/hypershift/blob/17a958fe04edb8ca2ce45c84553ca95a773246df/control-plane-operator/controllers/hostedcontrolplane/v2/cloud_controller_manager/openstack/config.go#L73

Which is supposed to be transitioned in 4.19. I'm ok to make that change but this will involve other changes not related to this PR. WDYT?

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -2703,14 +2703,14 @@ func (r *HostedControlPlaneReconciler) reconcileCloudProviderConfig(ctx context.
...
 		}); err != nil {
 			return fmt.Errorf("failed to reconcile OpenStack cloud config: %w", err)
 		}
 
 		withSecrets := manifests.OpenStackProviderConfigWithCredentials(hcp.Namespace)
 		if _, err := createOrUpdate(ctx, r, withSecrets, func() error {
-			return openstack.ReconcileCloudConfigSecret(hcp.Spec.Platform.OpenStack.ExternalNetwork.ID, withSecrets, hcp.Spec.Platform.OpenStack.IdentityRef.CloudName, credentialsSecret, caCertData)
+			return openstack.ReconcileCloudConfigSecret(hcp.Spec.Platform.OpenStack.ExternalNetwork.ID, withSecrets, hcp.Spec.Platform.OpenStack.IdentityRef.CloudName, credentialsSecret, caCertData, hcp.Spec.Networking.MachineNetwork)
```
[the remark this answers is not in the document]
EmilienM: same comment as above.

support/openstackutil/conversion.go
EmilienM: :+1:

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -1542,7 +1542,7 @@ func (r *reconciler) reconcileCloudCredentialSecrets(ctx context.Context, hcp *h
 }
 
 // reconcileOpenStackCredentialsSecret is a wrapper used to reconcile the OpenStack cloud config secrets.
-func (r *reconciler) reconcileOpenStackCredentialsSecret(ctx context.Context, externalNetworkID *string, namespace, name string, credentialsSecret *corev1.Secret, cloudName string, caCertData []byte) error {
+func (r *reconciler) reconcileOpenStackCredentialsSecret(ctx context.Context, externalNetworkID *string, namespace, name string, credentialsSecret *corev1.Secret, cloudName string, caCertData []byte, machineNetwork []hyperv1.MachineNetworkEntry) error {
```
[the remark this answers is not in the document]
EmilienM: This function is a bit complicated to test TBH and `ReconcileCloudConfigSecret` is going away in v2. Are you fine if we leave it like this?

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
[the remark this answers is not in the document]
bryan-cox: That sounds reasonable to me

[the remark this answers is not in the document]
EmilienM: I ended up doing it on https://github.com/openshift/hypershift/pull/5091 because of https://issues.redhat.com/browse/OCPBUGS-44375

docs/content/how-to/openstack/create-openstack-cluster.md
```diff
@@ -247,24 +247,27 @@ example-twxns         Ready    worker   88s     v1.27.4+18eadca
...
 
 hcp create nodepool openstack \
   --cluster-name $CLUSTER_NAME \
   --name $NODEPOOL_NAME \
   --node-count $WORKER_COUNT \
   --openstack-node-image-name $IMAGE_NAME \
   --openstack-node-flavor $FLAVOR \
-  --openstack-node-availability-zone $AZ
+  --openstack-node-availability-zone $AZ \
+  --openstack-node-additional-port=network-id:$SRIOV_NEUTRON_NETWORK_ID,vnic-type:direct,address-pairs:$ADDRESS_PAIRS
```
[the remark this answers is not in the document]
EmilienM: done

EmilienM:
/hold
I'm going to wait for https://github.com/openshift/hypershift/pull/5095 to land first so we avoid another regression.

EmilienM:
/test e2e-openstack
I needed to update the e2e tests because now we need to check the OpenStackServer status, and not the OpenStackMachine anymore.

EmilienM:
/hold cancel
the openstack job passed

PR #4529, opened 2024-08-12
HOSTEDCP-1609: Support configuring Azure diagnostics via CLI flags

fxierh:
This is to facilitate automation which requires creating an HC/NP with Azure diagnostics enabled.

### DIFF
--- a/cmd/nodepool/azure/create.go
+++ b/cmd/nodepool/azure/create.go
@@ -2,6 +2,7 @@ package azure
 
 import (
 	"context"
+	"fmt"
 	"strings"
 
 	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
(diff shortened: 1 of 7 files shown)


cmd/nodepool/azure/create.go
```diff
@@ -56,6 +58,8 @@ func bindCoreOptions(opts *RawAzurePlatformCreateOptions, flags *pflag.FlagSet)
 	flags.StringVar(&opts.InstanceType, "instance-type", opts.InstanceType, "The instance type to use for the nodepool")
 	flags.Int32Var(&opts.DiskSize, "root-disk-size", opts.DiskSize, "The size of the root disk for machines in the NodePool (minimum 16)")
 	flags.StringVar(&opts.AvailabilityZone, "availability-zone", opts.AvailabilityZone, "The availabilityZone for the nodepool. Must be left unspecified if in a region that doesn't support AZs")
+	flags.StringVar(&opts.DiagnosticsStorageAccountType, "diagnostics-storage-account-type", opts.DiagnosticsStorageAccountType, "Specifies the type of storage account for storing diagnostics data. Supported values: Disabled, Managed, UserManaged.")
```
bryan-cox, line 61: We should add validation to enforce this

cmd/nodepool/azure/create.go
```diff
@@ -56,6 +58,8 @@ func bindCoreOptions(opts *RawAzurePlatformCreateOptions, flags *pflag.FlagSet)
...
+	flags.StringVar(&opts.DiagnosticsStorageAccountURI, "diagnostics-storage-account-uri", opts.DiagnosticsStorageAccountURI, "Specifies the URI of the storage account for diagnostics data. Applicable only if --diagnostics-storage-account-type is set to UserManaged.")
```
bryan-cox, line 62: We should add validation to enforce this

cmd/nodepool/azure/create.go
fxierh: Updated.

cmd/nodepool/azure/create.go
fxierh: Updated.

bryan-cox:
kubevirt will be blocked until https://github.com/openshift/hypershift/pull/4530

PR #4531, opened 2024-08-12
OCPBUGS-38357: openstack: add option to configure nodepool AZ

EmilienM:

In OpenStack NodePools, add the AZ option which will configure the Machine spec with the failure domain.
We need it because customers sometimes want to deploy their Nodepools in a given availability zone (e.g. east-1, west-1, etc). This patch enables that.

### DIFF
--- a/hypershift-operator/controllers/nodepool/capi.go
+++ b/hypershift-operator/controllers/nodepool/capi.go
@@ -428,6 +428,13 @@ func (c *CAPI) reconcileMachineDeployment(ctx context.Context, log logr.Logger,
 		},
 	}
 
+	// The CAPI provider for OpenStack uses the FailureDomain field to set the availability zone.
+	if c.nodePool.Spec.Platform.Type == hyperv1.OpenStackPlatform && c.nodePool.Spec.Platform.OpenStack != nil {
+		if c.nodePool.Spec.Platform.OpenStack.AvailabilityZone != "" {
+			machineDeployment.Spec.Template.Spec.FailureDomain = ptr.To(c.nodePool.Spec.Platform.OpenStack.AvailabilityZone)
+		}
+	}
+
 	// After a MachineDeployment is created we propagate label/taints directly into Machines.
 	// This is to avoid a NodePool label/taints to trigger a rolling upgrade.
 	// TODO(Alberto): drop this an rely on core in-place propagation once CAPI 1.4.0 https://github.com/kubernetes-sigs/cluster-api/releases comes through the payload.
--- a/api/hypershift/v1beta1/openstack.go
+++ b/api/hypershift/v1beta1/openstack.go
@@ -12,6 +12,19 @@ type OpenStackNodePoolPlatform struct {
 	//
 	// +optional
 	ImageName string `json:"imageName,omitempty"`
+
+	// availabilityZone is the nova availability zone in which the provider will create the VM.
+	// If not specified, the VM will be created in the default availability zone specified in the nova configuration.
+	// Availability zone names must NOT contain : since it is used by admin users to specify hosts where instances
+	// are launched in server creation. Also, it must not contain spaces otherwise it will lead to node that belongs
+	// to this availability zone register failure, see kubernetes/cloud-provider-openstack#1379 for further information.
+	// The maximum length of availability zone name is 63 as per labels limits.
+	//
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:Pattern=`^[^: ]*$`
+	// +kubebuilder:validation:MaxLength=63
+	// +optional
+	AvailabilityZone string `json:"availabilityZone,omitempty"`
 }
 
 // OpenStackPlatformSpec specifies configuration for clusters running on OpenStack.
--- a/api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/nodepools.hypershift.openshift.io/OpenStack.yaml
+++ b/api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/nodepools.hypershift.openshift.io/OpenStack.yaml
@@ -930,6 +930,18 @@ spec:
                     description: OpenStack specifies the configuration used when using
                       OpenStack platform.
                     properties:
+                      availabilityZone:
+                        description: |-
+                          availabilityZone is the nova availability zone in which the provider will create the VM.
+                          If not specified, the VM will be created in the default availability zone specified in the nova configuration.
+                          Availability zone names must NOT contain : since it is used by admin users to specify hosts where instances
+                          are launched in server creation. Also, it must not contain spaces otherwise it will lead to node that belongs
+                          to this availability zone register failure, see kubernetes/cloud-provider-openstack#1379 for further information.
+                          The maximum length of availability zone name is 63 as per labels limits.
+                        maxLength: 63
+                        minLength: 1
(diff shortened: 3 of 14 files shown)


api/hypershift/v1alpha1/nodepool_types.go
```diff
@@ -209,6 +209,10 @@ type NodePoolSpec struct {
 	// +kubebuilder:validation:Enum=arm64;amd64
 	// +optional
 	Arch string `json:"arch,omitempty"`
+
+	// FailureDomain is the preferred failure domain for the NodePool.
+	// +optional
+	FailureDomain *string `json:"failureDomain,omitempty"`
```
mdbooth, line 215: Adding this to v1alpha1 as well as v1beta1 doesn't look right to me. Is this normally done in HyperShift?

cmd/cluster/core/create.go
```diff
@@ -97,6 +97,7 @@ func bindCoreOptions(opts *RawCreateOptions, flags *pflag.FlagSet) {
 	flags.StringVar(&opts.Arch, "arch", opts.Arch, "The default processor architecture for the NodePool (e.g. arm64, amd64)")
 	flags.StringVar(&opts.PausedUntil, "pausedUntil", opts.PausedUntil, "If a date is provided in RFC3339 format, HostedCluster creation is paused until that date. If the boolean true is provided, HostedCluster creation is paused until the field is removed.")
 	flags.StringVar(&opts.ReleaseStream, "release-stream", opts.ReleaseStream, "The OCP release stream for the cluster (e.g. 4-stable-multi), this flag is ignored if release-image is set")
+	flags.StringVar(&opts.FailureDomain, "failure-domain", opts.FailureDomain, "The failure domain for the NodePool")
```
mdbooth, line 100: This is adding it to `create cluster`, right? Why would it be required here? I'd have thought it was only required for `create nodepool`.

mdbooth (changes requested):
Nodepool change looks good. It looks like the changes to cluster are a no-op? Also, modifying a legacy API version looks odd to me, but if somebody asked for it perhaps it's a convention here?

Incidentally, any idea why nobody's done this already? Is there some other way to do failure domains we don't know about?

api/hypershift/v1alpha1/nodepool_types.go
EmilienM: I asked myself this and looking at git history, this seems common practise here.

cmd/cluster/core/create.go
EmilienM: yeah I know, I'll have to look again but this is how other nodepool options were added too.

EmilienM:
@JoelSpeed I would like your feedback on this one. Let me know if someone else should look.

Thanks :pray:

api/hypershift/v1alpha1/nodepool_types.go
JoelSpeed: If there's no conversion webhook then this makes sense, I don't know if they have one or not

api/hypershift/v1beta1/nodepool_types.go
```diff
@@ -194,6 +194,10 @@ type NodePoolSpec struct {
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf", message="Arch is immutable"
 	// +optional
 	Arch string `json:"arch,omitempty"`
+
+	// FailureDomain is the preferred failure domain for the NodePool.
```
JoelSpeed, line 198: I would encourage expanding the description here.

Where does a user find out what the available and valid values are for this field for example?

Is it possible that a Nodepool could ever cover multiple failure domains?

What happens when the user doesn't specify a value here at all?

api/hypershift/v1beta1/nodepool_types.go
```diff
@@ -194,6 +194,10 @@ type NodePoolSpec struct {
...
+	// +optional
+	FailureDomain *string `json:"failureDomain,omitempty"`
```
JoelSpeed, line 200: Need this be a pointer? Given the empty string is not a valid failure domain name, I think it probably doesn't

enxebre:
Thanks! Intent for this is currently expressed via subnets for Azure and AWS and selectors for agent/kubevirt. What's the value of having a new top level input that impact all platforms? How would this play/conflict/UX when the matrix input when per platform subnet is also set? how do I known which input is valid here from a consumer pov? Would we do any validation against it? What happens when the value is invalid?
Unless there's a clear value on having a new top field, I'd expect this to be a new Openstack specific field.
Even if we agreed on the value, there would need to be a clear UX for the input matrix support when combined with per platform input.

EmilienM:
This is WIP I'm reworking it for the OpenStack platform only as it was requested.

hypershift-operator/controllers/nodepool/capi.go
```diff
@@ -428,6 +428,13 @@ func (c *CAPI) reconcileMachineDeployment(ctx context.Context, log logr.Logger,
 		},
 	}
 
+	// The CAPI provider for OpenStack uses the FailureDomain field to set the availability zone.
```
EmilienM, line 431: @mdbooth when time permits, I'll like you eyes on that.

Some platforms have FailureDomain e.g. AzureMachine.Spec.FailureDomain, we don't have that in OpenStackMachine. We rely on the CAPI Machine.Spec.FailureDomain.
That's why I did it this way.

api/hypershift/v1beta1/nodepool_types.go
```diff
@@ -931,6 +931,11 @@ type OpenStackNodePoolPlatform struct {
 	//
 	// +optional
 	ImageName string `json:"imageName,omitempty"`
+
+	// AvailabilityZone is the failure domain identifier where the VM should be attached to.
```
JoelSpeed, line 935: Do we have any restrictions we can apply here? A valid character set or pattern? A minimum or maximum length?

EmilienM: I'll copy paste the doc & validation from CPMS API :smiley:

api/hypershift/v1beta1/nodepool_types.go
```diff
@@ -931,6 +931,19 @@ type OpenStackNodePoolPlatform struct {
 	//
 	// +optional
 	ImageName string `json:"imageName,omitempty"`
+
+	// AvailabilityZone is the nova availability zone in which the provider will create the VM.
```
JoelSpeed, line 935: Nit
```suggestion
	// availabilityZone is the nova availability zone in which the provider will create the VM.
```

EmilienM: yes I wanted to do that but I followed other fields in HCP API...

EmilienM: I'll change it anyway :man_shrugging:

EmilienM:
I'm going to write a e2e nodepool test for this one.

EmilienM:
/test e2e-openstack
for regression testing

mdbooth (commented):
/lgtm

bryan-cox:
@EmilienM Would you please add the "why" to the commit description? Otherwise, seems fine to me.

EmilienM:
> @EmilienM Would you please add the "why" to the commit description? Otherwise, seems fine to me.

done

EmilienM:
rebased & commit message updated

mdbooth (commented):
/lgtm

PR #4532, opened 2024-08-12
HOSTEDCP-1483: update cloud-credential-operator in go.mod

Patryk-Stefanski:
**What this PR does / why we need it**: update cloud-credential-operator to newer version that doesn't use a deprecated import with a CVE

Fixes # [HOSTEDCP-1483](https://issues.redhat.com/browse/HOSTEDCP-1483)

**Checklist**


go.mod
```diff
@@ -68,23 +68,23 @@ require (
...
 	gopkg.in/yaml.v3 v3.0.1
-	k8s.io/api v0.30.1
-	k8s.io/apiextensions-apiserver v0.30.1
-	k8s.io/apimachinery v0.30.1
-	k8s.io/apiserver v0.30.1
+	k8s.io/api v0.30.2
+	k8s.io/apiextensions-apiserver v0.30.2
+	k8s.io/apimachinery v0.30.2
+	k8s.io/apiserver v0.30.2
 	k8s.io/cli-runtime v0.30.1
```
bryan-cox, line 83: We should bump the other k8s to the same version, cli-runtime, kube-scheduler, kubectl, pod-security-admission.

You'll probably need to bump the replaces at the bottom as well.

Patryk-Stefanski: seems like we do need it after all, bumped two of the versions there is no kubernetes v0.30.2 so I'm leaving it as is.

devguyio (commented):
/lgtm

PR #4536, opened 2024-08-12
[WIP] HOSTEDCP-1826: API review and feedback for Azure APIs

JoelSpeed:

This PR is intended to be used for conversation about improvements to the Azure APIs being added within HyperShift.
There are two commits so far, one which moves the existing Azure types to a new file, separate from the others, and the second commit is actually making some changes. Reviewing this commit wise is strongly advised.

So far, this has mostly been documentation, but I've added a bunch of TODOs throughout where I'm hoping others can help find the information that we need to document.

CC @enxebre 

Fixes #

**Checklist**


api/hypershift/v1beta1/azure_types.go
```diff
@@ -1,144 +1,177 @@
 package v1beta1
 
+// AzureVMImageType is used to specify the source of the Azure VM boot image.
+// Valid values are ImageID and AzureMarketplace.
+// +kubebuilder:validation:Enum:=ImageID;AzureMarketplace
```
JoelSpeed, line 5: Just moved this up to the type rather than changing it

api/hypershift/v1beta1/azure_types.go
```diff
@@ -1,144 +1,177 @@
...
+	ImageID AzureVMImageType = "ImageID"
+
+	// AzureMarketplace is used to specify the Azure Marketplace image info to use to boot the Azure VMs from.
 	AzureMarketplace AzureVMImageType = "AzureMarketplace"
 )
 
+// AzureNodePoolPlatform is the platform specific configuration for an Azure node pool.
 type AzureNodePoolPlatform struct {
-	// VMSize is the Azure VM instance type to use for the nodes being created in the nodepool.
+	// vmSize is the Azure VM instance type to use for the nodes being created in the nodepool.
```
JoelSpeed, line 18: Godocs for fields should start with the lower case version of the name, as this is what users interact with, serialized data

api/hypershift/v1beta1/azure_types.go
```diff
@@ -1,144 +1,177 @@
...
 	//
+	// TODO: This should be validated with a regex, based on https://learn.microsoft.com/en-us/azure/virtual-machines/vm-naming-conventions
```
JoelSpeed, line 20: I'm sure I've seen someone write a regex for this before, need to find where

api/hypershift/v1beta1/azure_types.go
```diff
@@ -1,144 +1,177 @@
...
-	VMSize string `json:"vmsize"`
+	VMSize string `json:"vmSize"`
 
-	// ImageID is the id of the image to boot from. If unset, the default image at the location below will be used and
+	// image is used to configure the VM boot image. If unset, the default image at the location below will be used and
 	// is expected to exist: subscription/<subscriptionID>/resourceGroups/<resourceGroupName>/providers/Microsoft.Compute/images/rhcos.x86_64.vhd.
 	// The <subscriptionID> and the <resourceGroupName> are expected to be the same resource group documented in the
 	// Hosted Cluster specification respectively, HostedCluster.Spec.Platform.Azure.SubscriptionID and
 	// HostedCluster.Spec.Platform.Azure.ResourceGroupName.
+	// The image is immutable once set.
```
JoelSpeed, line 29: Explain all validations to users

api/hypershift/v1beta1/azure_types.go
```diff
@@ -1,144 +1,177 @@
...
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf", message="image is immutable"
 	// +kubebuilder:validation:Required
 	Image AzureVMImage `json:"image"`
 
-	// DiskSizeGB is the size in GB to assign to the OS disk
-	// CAPZ default is 30GB, https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/b3708019a67ff19407b87d63c402af94ca4246f6/api/v1beta1/types.go#L599
+	// diskSizeGB is the size in GB to assign to the OS disk. This should be between 16 and <TODO find a limit>.
+	// When not set, this means no opinion and the platform is left to choose a reasonable default, which is subject to change over time.
+	// The current default is 30.
+	// TODO: Is this GB or GiB?
```
JoelSpeed, line 38: Do we know the answer to this one?

api/hypershift/v1beta1/azure_types.go
```diff
@@ -1,144 +1,177 @@
...
+	// The image is immutable once set.
 	//
-	// +kubebuilder:validation:XValidation:rule="self == oldSelf", message="Image is immutable"
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf", message="image is immutable"
 	// +kubebuilder:validation:Required
 	Image AzureVMImage `json:"image"`
 
-	// DiskSizeGB is the size in GB to assign to the OS disk
-	// CAPZ default is 30GB, https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/b3708019a67ff19407b87d63c402af94ca4246f6/api/v1beta1/types.go#L599
+	// diskSizeGB is the size in GB to assign to the OS disk. This should be between 16 and <TODO find a limit>.
```
JoelSpeed, line 35: Should set a limit, even if it is arbitrarily high

api/hypershift/v1beta1/azure_types.go
```diff
@@ -1,144 +1,177 @@
...
-	// DiskStorageAccountType is the disk storage account type to use. Valid values are:
-	// * Standard_LRS: HDD
-	// * StandardSSD_LRS: Standard SSD
-	// * Premium_LRS: Premium SDD
-	// * UltraSSD_LRS: Ultra SDD
-	//
-	// Defaults to Premium_LRS. For more details, visit the Azure documentation:
-	// https://docs.microsoft.com/en-us/azure/virtual-machines/disks-types#disk-type-comparison
+	// diskStorageAccountType is the disk storage account type to use.
+	// Valid values are Standard, StandardSSD, PremiumSSD and UltraSSD and omitted.
```
JoelSpeed, line 45: You don't need to use the same values that Azure APIs use, instead use something that is intuitive to the user, but meets conventions, so in this case, these are PascalCase, but should be intuitive, we know there are 4 tiers and the SSD versions are marked as SSDs

api/hypershift/v1beta1/azure_types.go
```diff
@@ -1,144 +1,177 @@
...
+	// Note that Standard means a HDD.
+	// The disk performance is tied to the disk type, please refer to the Azure documentation for further details
+	// https://docs.microsoft.com/en-us/azure/virtual-machines/disks-types#disk-type-comparison.
+	// When omitted this means no opinion and the platform is left to choose a reasonable default, which is subject to change over time.
```
JoelSpeed, line 49: For configuration APIs, default this on the controller side, and not the API side, else you have no way to change the default later should you wish to do so

api/hypershift/v1beta1/azure_types.go
```diff
@@ -1,144 +1,177 @@
...
+	// The current default is PremiumSSD.
 	//
-	// +kubebuilder:default:=Premium_LRS
-	// +kubebuilder:validation:Enum=Standard_LRS;StandardSSD_LRS;Premium_LRS;UltraSSD_LRS
+	// +kubebuilder:validation:Enum=Standard;StandardSSD;PremiumSSD;UltraSSD
 	// +optional
+	// TODO: Should all disk options come under a struct rather than be top level like this?
```
JoelSpeed, line 54: I don't know why we have several disk based options all prefixed with Disk, would be better to have a struct for this right?

api/hypershift/v1beta1/azure_types.go
```diff
@@ -1,144 +1,177 @@
...
 	DiskStorageAccountType string `json:"diskStorageAccountType,omitempty"`
 
-	// AvailabilityZone is the failure domain identifier where the VM should be attached to. This must not be specified
-	// for clusters in a location that does not support AvailabilityZone.
+	// availabilityZone is the failure domain identifier where the VM should be attached to.
+	// This must not be specified for clusters in a location that does not support AvailabilityZone because... TODO: why?
```
JoelSpeed, line 58: Don't just state this, explain why, else someone will ask

api/hypershift/v1beta1/azure_types.go
```diff
@@ -1,144 +1,177 @@
...
+	// Example values are TODO
+	// TODO: What is the valid character set for this field? What about minimum and maximum lengths?
```
JoelSpeed, line 60: Every string field should answer these questions, patterns, character sets, lengths are all important to think about

api/hypershift/v1beta1/azure_types.go
```diff
@@ -1,144 +1,177 @@
...
 	//
 	// +optional
 	AvailabilityZone string `json:"availabilityZone,omitempty"`
 
-	// DiskEncryptionSetID is the ID of the DiskEncryptionSet resource to use to encrypt the OS disks for the VMs. This
+	// diskEncryptionSetID is the ID of the DiskEncryptionSet resource to use to encrypt the OS disks for the VMs. This
 	// needs to exist in the same subscription id listed in the Hosted Cluster, HostedCluster.Spec.Platform.Azure.SubscriptionID.
 	// DiskEncryptionSetID should also exist in a resource group under the same subscription id and the same location
 	// listed in the Hosted Cluster, HostedCluster.Spec.Platform.Azure.Location.
+	// TODO: What does this do for a customer, why would they want to set it?
```
JoelSpeed, line 69: This godoc doesn't explain to the end user what they can achieve with this feature, we should give them the 20k foot view

api/hypershift/v1beta1/azure_types.go
```diff
@@ -1,144 +1,177 @@
...
+	// TODO: What is the valid character set for this field? What about minimum and maximum lengths?
 	//
 	// +optional
 	DiskEncryptionSetID string `json:"diskEncryptionSetID,omitempty"`
 
-	// EnableEphemeralOSDisk is a flag when set to true, will enable ephemeral OS disk.
+	// TODO: This shouldn't be a bool, but I don't know what the field is really for.
```
JoelSpeed, line 75: Bools are discourage upstream, and forbidden downstream. I haven't re-implemented this as an enum yet as I have no idea what it does, I was wondering if we create a `Disk` structure, should this be 
```
storageAccountMode: Persistent | Ephemeral
```

api/hypershift/v1beta1/azure_types.go
```diff
@@ -1,144 +1,177 @@
...
+// +kubebuilder:validation:XValidation:rule="has(self.type) && self.type == 'ImageID' ?  has(self.imageID) : !has(self.imageID)",message="imageID is required when type is ImageID, and forbidden otherwise"
+// +kubebuilder:validation:XValidation:rule="has(self.type) && self.type == 'AzureMarketplace' ?  has(self.azureMarketplace) : !has(self.azureMarketplace)",message="azureMarketplace is required when type is RequiredMember, and forbidden otherwise"
 // +union
 type AzureVMImage struct {
-	// Type is the type of image data that will be provided to the Azure VM. This can be either "ImageID" or
-	// "AzureMarketplace".
+	// type is the type of image data that will be provided to the Azure VM.
+	// Valid values are "ImageID" and "AzureMarketplace".
+	// ImageID means ... TODO
+	// AzureMarketplace means ... TODO
```
JoelSpeed, line 123: Help the user to choose the correct value by telling them what each one means

api/hypershift/v1beta1/azure_types.go
```diff
@@ -1,144 +1,177 @@
...
 	//
 	// +optional
 	// +unionMember
 	AzureMarketplace *MarketplaceImage `json:"azureMarketplace,omitempty"`
 }
 
-// MarketplaceImage specifies the information needed to create an Azure VM from an Azure Marketplace image. This struct
-// replicates the same fields found in CAPZ - https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/api/v1beta1/types.go.
+// MarketplaceImage specifies the information needed to create an Azure VM from an Azure Marketplace image.
+// + This struct replicates the same fields found in CAPZ - https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/api/v1beta1/types.go.
```
JoelSpeed, line 144: I don't think this information is supposed to end up in the kubectl explain output is it? Prefixing with a `+` skips it from ending up there

api/hypershift/v1beta1/azure_types.go
```diff
@@ -1,144 +1,177 @@
...
-	// Publisher is the name of the organization that created the image
+	// publisher is the name of the organization that created the image.
+	// It must be between 3 and 50 characters in length, and consist of only lowercase letters, numbers, and hyphens (-) and underscores (_).
+	// It must start with a lowercase letter or a number.
+	// TODO: Can we explain where a user might find this value, or provide an example of one they might want to use
 	//
 	// +kubebuilder:validation:Pattern=`^[a-z0-9][a-z0-9-_]{2,49}$`
 	// +kubebuilder:validation:MinLength=3
 	// +kubebuilder:validation:MaxLength=50
+	// +kubeubilder:validation:Required
```
JoelSpeed, line 154: Always be explicit about optional or required

api/hypershift/v1beta1/azure_types.go
```diff
@@ -1,144 +1,177 @@
...
+	// TODO: What about length limits?
 	//
 	// +kubebuilder:validation:Pattern=`^[a-z0-9-_]+$`
 	// +kubebuilder:validation:MinLength=1
 	SKU string `json:"sku"`
 
-	// Version specifies the version of an image sku. The allowed formats are Major.Minor.Build or 'latest'. Major,
-	// Minor, and Build are decimal numbers. Specify 'latest' to use the latest version of an image available at
+	// version specifies the version of an image sku. The allowed formats are Major.Minor.Build or 'latest'. Major,
+	// Minor, and Build are decimal numbers, e.g. '1.2.0'. Specify 'latest' to use the latest version of an image available at
```
JoelSpeed, line 174: Provide real world examples

api/hypershift/v1beta1/azure_types.go
```diff
@@ -149,21 +182,31 @@ type MarketplaceImage struct {
...
-	// +kubebuilder:default:=Disabled
-	StorageAccountType string `json:"storageAccountType,omitempty"`
-	// StorageAccountURI is the URI of the user-managed storage account.
+	// +kubebuilder:validation:Required
+	StorageAccountType string `json:"storageAccountType"`
+
+	// userManaged specifies the diagnostics settings for a virtual machine when the storage account is managed by the user.
+	// +optional
+	// +unionMember
+	UserManaged *UserManagedDiagnostics `json:"userManaged,omitempty"`
```
JoelSpeed, line 197: This struct looked like a union, but not quite

api/hypershift/v1beta1/azure_types.go
```diff
@@ -149,21 +182,31 @@ type MarketplaceImage struct {
...
+// UserManagedDiagnostics specifies the diagnostics settings for a virtual machine when the storage account is managed by the user.
+type UserManagedDiagnostics struct {
+	// storageAccountURI is the URI of the user-managed storage account.
 	// The URI typically will be `https://<mystorageaccountname>.blob.core.windows.net/`
 	// but may differ if you are using Azure DNS zone endpoints.
 	// You can find the correct endpoint by looking for the Blob Primary Endpoint in the
 	// endpoints tab in the Azure console or with the CLI by issuing
 	// `az storage account list --query='[].{name: name, "resource group": resourceGroup, "blob endpoint": primaryEndpoints.blob}'`.
-	// +kubebuilder:validation:Format=uri
+	// +kubebuilder:validation:XValidation:rule="isURL(self) && url(self).scheme() == 'https'", message="storageAccountURI must be a valid HTTPS URL"
```
JoelSpeed, line 208: CEL is better for this, and we can check that the value is secure as well

api/hypershift/v1beta1/azure_types.go
bennerv: https://learn.microsoft.com/en-us/azure/virtual-machines/disks-types

It's all GiB from what I can tell.  All standard HDD up to ultra SSD state their disk sizes in GiB.  But the [azure-sdk-for-go](https://raw.githubusercontent.com/Azure/azure-sdk-for-go/main/sdk/resourcemanager/compute/armcompute/models.go) uses GB.  :(

JoelSpeed: Right, so they've called it GB, but it's actually GiB, so we should call it GiB so that our users know for certain. Also should explain this in the godoc

JoelSpeed: Looking at the disk types table, that also has the maximum sizes, which I'll add in, do we support premium v2?

JoelSpeed: No, we do not support premium v2 right now

api/hypershift/v1beta1/azure_types.go
bryan-cox: 100 TiB? per https://learn.microsoft.com/en-us/azure/virtual-machines/disks-types#ultra-disk-size

JoelSpeed: I think the limits are set out here, https://learn.microsoft.com/en-us/azure/virtual-machines/disks-types, the 100TiB appears to be across your account, not for an individual disk

bryan-cox: Ah true. 65,536 then?

JoelSpeed: I think so, and we can use CEL to limit it for the smaller types too, I'll push up something to show this

api/hypershift/v1beta1/azure_types.go
bryan-cox: Hmm, I don't see one defined in [CAPZ](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/160a5c8891746b10b9ae5314d27073d5c76ea1f3/api/v1beta1/azuremachine_types.go#L39)

api/hypershift/v1beta1/azure_types.go
bryan-cox: I believe I set it to immutable because I thought changing it would result in a rollout of new nodes, but maybe that is not true or maybe its okay to allow that if the user choose to do so 🤔

api/hypershift/v1beta1/azure_types.go
```diff
@@ -0,0 +1,213 @@
...
+	// HostedCluster.Spec.Platform.Azure.ResourceGroupName.
+	// The image is immutable once set.
+	//
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf", message="image is immutable"
+	// +kubebuilder:validation:Required
+	Image AzureVMImage `json:"image"`
+
+	// diskSizeGiB is the size in GiB (1024^3 bytes) to assign to the OS disk. This should be between 16 and 65,536.
+	// When not set, this means no opinion and the platform is left to choose a reasonable default, which is subject to change over time.
+	// The current default is 30.
```
bryan-cox, line 38: Does this need to be removed?

bryan-cox: Or maybe updated to say CAPZ will default this to 30.

api/hypershift/v1beta1/azure_types.go
bryan-cox: That sounds fine to me

api/hypershift/v1beta1/azure_types.go
bryan-cox: This was to enable ephemeral OS disks - https://learn.microsoft.com/en-us/azure/virtual-machines/ephemeral-os-disks 

When its enabled we set the caching to read only and the diff disk settings to local - https://github.com/openshift/hypershift/blob/c467b683d6878a23bb8b08ab1a79f3d3e7a160b7/hypershift-operator/controllers/nodepool/azure.go#L90-L89

api/hypershift/v1beta1/azure_types.go
bryan-cox: source of info for azure marketplace images - https://techcommunity.microsoft.com/t5/marketplace-blog/exploring-virtual-machine-offers-in-the-azure-marketplace/ba-p/3779854

api/hypershift/v1beta1/azure_types.go
bryan-cox: cc: @Patryk-Stefanski - I thought you might want to take a stab at answering the TODOs for this one

api/hypershift/v1beta1/azure_types.go
```diff
@@ -0,0 +1,213 @@
...
+	//
+	// +optional
+	AvailabilityZone string `json:"availabilityZone,omitempty"`
+
+	// diskEncryptionSetID is the ID of the DiskEncryptionSet resource to use to encrypt the OS disks for the VMs. This
+	// needs to exist in the same subscription id listed in the Hosted Cluster, HostedCluster.Spec.Platform.Azure.SubscriptionID.
+	// DiskEncryptionSetID should also exist in a resource group under the same subscription id and the same location
+	// listed in the Hosted Cluster, HostedCluster.Spec.Platform.Azure.Location.
+	// TODO: What does this do for a customer, why would they want to set it?
+	// TODO: What is the valid character set for this field? What about minimum and maximum lengths?
```
bryan-cox, line 71: "Supported characters for the disk encryption set name are letters, numbers, underscores, hyphens, and periods. The name cannot begin with an underscore, hyphen, or period, and cannot end with a hyphen or period.

The value must be between 1 and 80 characters long."

api/hypershift/v1beta1/azure_types.go
```diff
@@ -0,0 +1,213 @@
...
+	//
+	// +optional
+	EnableEphemeralOSDisk bool `json:"enableEphemeralOSDisk,omitempty"`
+
+	// subnetID is the subnet ID of an existing subnet where the nodes in the nodepool will be created. This can be a
+	// different subnet than the one listed in the HostedCluster, HostedCluster.Spec.Platform.Azure.SubnetID, but must
+	// exist in the same network, HostedCluster.Spec.Platform.Azure.VnetID, and must exist under the same subscription ID,
+	// HostedCluster.Spec.Platform.Azure.SubscriptionID.
+	// subnetID is immutable once set.
+	// TODO: What is the valid character set for this field? What about minimum and maximum lengths?
```
bryan-cox, line 90: For the actual name...

"The name must begin with a letter or number, end with a letter, number or underscore, and may contain only letters, numbers, underscores, periods, or hyphens.

The name must be between 1 and 80 characters."

api/hypershift/v1beta1/azure_types.go
```diff
@@ -0,0 +1,213 @@
...
+	// machineIdentityID is a user-assigned identity assigned to the VMs used to authenticate with Azure services. The
+	// identify is expected to exist under the same resource group as HostedCluster.Spec.Platform.Azure.ResourceGroupName. This
+	// user assigned identity is expected to have the Contributor role assigned to it and scoped to the resource group
+	// under HostedCluster.Spec.Platform.Azure.ResourceGroupName.
+	//
+	// If this field is not supplied, the Service Principal credentials will be written to a file on the disk of each VM
+	// in order to be accessible by the cloud provider; the aforementioned credentials provided are the same ones as
+	// HostedCluster.Spec.Platform.Azure.Credentials. However, this is less secure than using a managed identity.
+	//
+	// TODO: What is the valid character set for this field? What about minimum and maximum lengths?
```
bryan-cox, line 110: So it follows this structure - https://github.com/openshift/hypershift/blob/b0a6b0af99a9120e77994907f0c702b5dd44980b/vendor/sigs.k8s.io/cluster-api-provider-azure/api/v1beta1/types.go#L568

From there its just checking each component in https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules I believe.

api/hypershift/v1beta1/azure_types.go
JoelSpeed: That's fine, whatever choice is made we need to tell them why it is immutable. If we can tell them the reason it's immutable that's even better

api/hypershift/v1beta1/azure_types.go
JoelSpeed: Yes something along those lines, need to confirm though, is this the key that encrypts the disk, or the key that encrypts the key that encrypts the disk, some clouds use different patterns here

api/hypershift/v1beta1/azure_types.go
JoelSpeed: Ok great, so we can explain the use case (lower latency to the disk), the risks, and the side effects, to help the end user understand what's going on with this option. I think we can get all of that from the links you've mentioned

api/hypershift/v1beta1/azure_types.go
JoelSpeed: No, typically we still say the default, it aligns with the previous sentence, and then the nodepool controller should be enforcing this default

api/hypershift/v1beta1/azure_types.go
JoelSpeed: When you say the actual name, is that not the value of this field?

api/hypershift/v1beta1/azure_types.go
JoelSpeed: Great, we can come up with a pattern to match for this

api/hypershift/v1beta1/azure_types.go
JoelSpeed: In this case, actually we may want this to be mutable, but warn users that if they do change it, it will roll out, WDYT?

JoelSpeed:
Posted some more updates, in particular around the VM size and disk options

api/hypershift/v1beta1/azure_types.go
```diff
@@ -0,0 +1,272 @@
...
+	// osDisk provides configuration for the OS disk for the nodepool.
+	// This can be used to configure the size, storage account type, encryption options and whether the disk is persistent or ephemeral.
+	// When not provided, the platform will choose reasonable defaults which are subject to change over time.
+	// Review the fields within the osDisk for more details.
+	OSDisk AzureNodePoolOSDisk `json:"osDisk"`
+
+	// availabilityZone is the failure domain identifier where the VM should be attached to.
+	// This must not be specified for clusters in a location that does not support AvailabilityZone because... TODO: why?
+	// Example values are TODO
+	// TODO: What is the valid character set for this field? What about minimum and maximum lengths?
```
JoelSpeed, line 54: IIUC correctly, AZs in azure are actually integers, but we represent them as strings? Typically `"1"`, `"2"` or `"3"`

api/hypershift/v1beta1/azure_types.go
JoelSpeed: Is this a complete azure resource ID, or just the subnet itself?

api/hypershift/v1beta1/azure_types.go
JoelSpeed: Others don't always start with the Azure prefix, this one definitely does? Should we prefix that ourselves to avoid the difference?

api/hypershift/v1beta1/azure_types.go
bryan-cox: sgtm

api/hypershift/v1beta1/azure_types.go
bryan-cox: A complete azure resource ID

api/hypershift/v1beta1/azure_types.go
enxebre: This needs to be mutable and trigger a rollout.
Let's use // +noderollout for fields which do so. That should be in the upper level NodePool struct (.NodePoolPlatform ) though since any change to AzureNodePoolPlatform triggers a rollout by design and if anything within it is immutable it needs to be enforced by CEL and documented here

api/hypershift/v1beta1/azure_types.go
enxebre: because the Machine will fail to boot?

api/hypershift/v1beta1/azure_types.go
```diff
@@ -0,0 +1,288 @@
...
+	// Availability zones are identified by numbers, either 1, 2 or 3.
+	//
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=3
+	// +optional
+	AvailabilityZone int32 `json:"availabilityZone,omitempty"`
+
+	// subnetID is the subnet ID of an existing subnet where the nodes in the nodepool will be created. This can be a
+	// different subnet than the one listed in the HostedCluster, HostedCluster.Spec.Platform.Azure.SubnetID, but must
+	// exist in the same network, HostedCluster.Spec.Platform.Azure.VnetID, and must exist under the same subscription ID,
```
enxebre, line 62: > but must // exist in the same network, HostedCluster.Spec.Platform.Azure.VnetID

who's enforcing this, capz?

enxebre: Is this exclusive with subnet id or any other field?

api/hypershift/v1beta1/azure_types.go
enxebre: >If this field is not supplied, the Service Principal credentials will be written to a file on the disk of each VM

why do we want to support this?

api/hypershift/v1beta1/azure_types.go
JoelSpeed: Is there any tooling looking for the `+noderollout` marker or is that just to jog memory?

api/hypershift/v1beta1/azure_types.go
JoelSpeed: We don't have to, but I think from the conversations we've had, this is what happens today. You could make this field required and ensure that the machineIdentity is provided

api/hypershift/v1beta1/azure_types.go
```diff
@@ -0,0 +1,328 @@
...
+	// The vnetName should be between 2 and 64 characters, consisting only of alphanumeric characters, hyphens, underscores and periods and must not end with either a period (.) or hyphen (-) character.
+	// The subnetName should be between 1 and 80 characters, consisting only of alphanumeric characters, hyphens and underscores and must start with an alphanumeric character and must not end with a period (.) or hyphen (-) character.
+	//
+	// +kubebuilder:validation:XValidation:rule="size(self.split('/')) == 11 && self.matches('^/subscriptions/.*/resourceGroups/.*/providers/Microsoft.Network/virtualNetworks/.*/subnets/.*$')",message="encryptionSetID must be in the format `/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Network/virtualNetworks/{vnetName}/subnets/{subnetName}`"
+	// +kubeubilder:validation:XValidation:rule="self.split('/')[2].matches('^[{]?[0-9a-fA-F]{8}-([0-9a-fA-F]{4}-){3}[0-9a-fA-F]{12}[}]?$')",message="the subscriptionId in the encryptionSetID must be a valid UUID. It should be 5 groups of hyphen separated hexadecimal characters in the form 8-4-4-4-12"
```
bryan-cox, line 82:
```suggestion
	// +kubeubilder:validation:XValidation:rule="self.split('/')[2].matches('^[{]?[0-9a-fA-F]{8}-([0-9a-fA-F]{4}-){3}[0-9a-fA-F]{12}[}]?$')",message="the subscriptionId in the subnetID must be a valid UUID. It should be 5 groups of hyphen separated hexadecimal characters in the form 8-4-4-4-12"
```

enxebre:
/close
following up here https://github.com/openshift/hypershift/pull/4963

PR #4538, opened 2024-08-12
HOSTEDCP-1871: Add multi-arch validation for Azure platform

bryan-cox:
Adds multi-arch validation for Azure platform

Fixes [HOSTEDCP-1871](https://issues.redhat.com/browse/HOSTEDCP-1871)

**Checklist**


api/hypershift/v1alpha1/hostedcluster_types.go
```diff
@@ -1574,6 +1574,12 @@ type AzurePlatformSpec struct {
 	SubnetID          string `json:"subnetID"`
 	SubscriptionID    string `json:"subscriptionID"`
 	SecurityGroupID   string `json:"securityGroupID"`
+
+	// MultiArch specifies whether the Hosted Cluster will be expected to support NodePools with different
+	// CPU architectures, i.e., supporting arm64 NodePools and supporting amd64 NodePools on the same Hosted Cluster.
+	// +kubebuilder:default=false
+	// +optional
+	MultiArch bool `json:"multiArch"`
```
JoelSpeed, line 1582: At any point, do you need to know which architectures the cluster should be supporting? Eg will you need to know if it's arm and amd vs arm and ppcle?

I wonder if this should be a list, where you can detect the available architectures and handle the multi-architecture case when the list has a length greater than 1.

If there's no benefit to understanding specific architectures, then perhaps an enum of `payloadArchitecture: Single | Multi` might be a good start?

bryan-cox: The architectures will be amd64, arm64. I don't see ppcle or s390x coming to Azure.

I would lean more towards "there's no benefit to understanding specific architectures".

JoelSpeed: I would probably go down the route of that enum then, in the future if we did need to know the difference, we could expand with another value `Specific` and then have the additional field for users to specify which specific architectures they want

api/hypershift/v1alpha1/hostedcluster_types.go
```diff
@@ -1574,6 +1574,26 @@ type AzurePlatformSpec struct {
 	SubnetID          string `json:"subnetID"`
 	SubscriptionID    string `json:"subscriptionID"`
 	SecurityGroupID   string `json:"securityGroupID"`
+
+	// payloadArchitecture specifies if a single CPU architecture payload or a multi-arch CPU payload is being used for
+	// the HostedCluster and HostedControlPlane. When this field is set to Multi, additional validation checks will be
+	// performed on the release image of the HostedCluster to confirm it is a multi-arch image.
+	//
+	// The default for this field is Multi since this allows both amd64 and arm64 NodePools to be created from day 0
```
JoelSpeed, line 1582: You've marked the field as required, so it cannot have a default?

api/hypershift/v1alpha1/hostedcluster_types.go
```diff
@@ -1574,6 +1574,26 @@ type AzurePlatformSpec struct {
...
+	// perspective avoiding any future HostedCluster and HostedControlPlane migrations from a single CPU architecture
+	// payload to a multi-arch CPU payload.
+	//
+	// Enum definitions:
+	// - Single means the release image or stream used for the HostedCluster and HostedControlPlane is related to a
+	// single, specific CPU type (i.e. amd64 or arm64).
+	// - Multi means the release image or stream used for the HostedCluster and HostedControlPlane is related to a
+	// payload with manifests for multiple CPU types (i.e. should include at least amd64 and arm64 for Azure).
```
JoelSpeed, line 1590: You'll need to write this in prose, when this gets rendered by `oc explain` it won't take into account your new line chars

api/hypershift/v1alpha1/hostedcluster_types.go
```diff
@@ -1574,6 +1574,26 @@ type AzurePlatformSpec struct {
...
+	//
+	// +kubebuilder:default=Multi
+	// +kubebuilder:validation:Enum=Single;Multi
+	// +kubebuilder:validation:Required
+	// +required
```
JoelSpeed, line 1595: If this API has shipped, you can't add a new required field. It would break all existing versions of it on upgrade.

I think you actually want this optional with a default

api/hypershift/v1alpha1/hostedcluster_types.go
```diff
@@ -1574,6 +1574,26 @@ type AzurePlatformSpec struct {
...
+	// perspective avoiding any future HostedCluster and HostedControlPlane migrations from a single CPU architecture
+	// payload to a multi-arch CPU payload.
+	//
+	// Enum definitions:
+	// - Single means the release image or stream used for the HostedCluster and HostedControlPlane is related to a
+	// single, specific CPU type (i.e. amd64 or arm64).
+	// - Multi means the release image or stream used for the HostedCluster and HostedControlPlane is related to a
+	// payload with manifests for multiple CPU types (i.e. should include at least amd64 and arm64 for Azure).
+	//
+	// +kubebuilder:default=Multi
```
JoelSpeed, line 1592: Are you ever likely to want to change this default?

We typically recommend for low volume APIs like this (configuration APIs), that you don't actually set defaults, but rather use a controller to implement the default, and state in your godoc that the value is subject to change over time

api/hypershift/v1alpha1/hostedcluster_types.go
bryan-cox: I don't understand. A field cannot be required and have a default value?

api/hypershift/v1alpha1/hostedcluster_types.go
bryan-cox: Is there an example you can point me to?

api/hypershift/v1alpha1/hostedcluster_types.go
JoelSpeed: Defaulting is applied after validation, so marking it required, would reject the request when the field is not populated, but marking it optional, would allow the object to be persisted, and then the default would be applied when being read

api/hypershift/v1alpha1/hostedcluster_types.go
bryan-cox: It's not shipped as in its in production. ARO HCP is still in development internally.

bryan-cox: No one should be using the alpha API

api/hypershift/v1alpha1/hostedcluster_types.go
JoelSpeed: What is the upgrade story here? If this API is added, can it accurately reflect the current state?

bryan-cox: There should be no HC using alpha API so there should be no upgrades happening today between alpha and beta API for Azure.

api/hypershift/v1alpha1/hostedcluster_types.go
bryan-cox: Ack, I thought you were referring to some special character or way of doing it 😅

cmd/cluster/aws/create.go
```diff
@@ -59,8 +57,9 @@ type ValidatedCreateOptions struct {
 
 func (o *RawCreateOptions) Validate(ctx context.Context, opts *core.CreateOptions) (core.PlatformCompleter, error) {
 	// Perform multi-arch validations except for render
+	var err error
 	if !opts.Render {
-		err := validateMultiArch(ctx, o, opts)
+		o.MultiArch, err = core.ValidateMultiArch(ctx, o.MultiArch, opts)
```
enxebre, line 62: why is this here and not in the core Validate?

cmd/cluster/azure/create.go
```diff
@@ -90,6 +93,31 @@ type ValidatedCreateOptions struct {
...
+
+	// Validate values used for PayloadArch
+	if o.PayloadArch != "Single" && o.PayloadArch != "Multi" {
+		return nil, fmt.Errorf("unsupported payload architecture type %s", o.PayloadArch)
+	}
+
+	// Perform multi-arch validations except for render
+	isMultiPayload := o.PayloadArch == "Multi"
+	if !opts.Render {
+		isMultiPayload, err = core.ValidateMultiArch(ctx, isMultiPayload, opts)
```
enxebre, line 106: why do we have ValidateMultiArch and ValidateMultiArchRelease? what's the difference?

cmd/cluster/azure/create.go
```diff
@@ -90,6 +93,31 @@ type ValidatedCreateOptions struct {
...
+		if err != nil {
+			return nil, err
+		}
+
+		if isMultiPayload {
```
enxebre, line 111: why is a flag being hardcoded here?

cmd/cluster/core/create.go
```diff
@@ -922,3 +926,73 @@ func parseTolerationString(str string) (*corev1.Toleration, error) {
...
+	}
+
+	// Check if a release stream was provided instead and its multi-arch
+	if len(opts.ReleaseStream) > 0 && strings.Contains(opts.ReleaseStream, "multi") {
+		validMultiArchImage = true
+	}
+
+	// The release image is not multi-arch so check the mgmt & nodepool cpu arches match
+	if !validMultiArchImage {
+		mgmtClusterCPUArch, err := hyperutil.GetMgmtClusterCPUArch(ctx)
```
enxebre, line 953: GetMgmtClusterCPUArch need to be unit tested

hypershift-operator/controllers/nodepool/nodepool_controller.go
```diff
@@ -355,6 +356,22 @@ func (r *NodePoolReconciler) reconcile(ctx context.Context, hcluster *hyperv1.Ho
 		return ctrl.Result{}, fmt.Errorf("failed to look up release image metadata: %w", err)
 	}
 
+	// Validate the NodePool and management cluster CPU architectures match if the HostedCluster does not support multi-arch NodePools
+	supportsMultiArchNodePoolCondition := meta.FindStatusCondition(hcluster.Status.Conditions, hyperv1.SupportsMultiArchNodePools)
+	if supportsMultiArchNodePoolCondition != nil && supportsMultiArchNodePoolCondition.Status == metav1.ConditionFalse {
+		// The HostedCluster does not support managing different CPU architecture NodePools. Validate the CPU
+		// architecture of the management cluster and the NodePool match.
+		mgmtClusterCPUArch, err := supportutil.GetMgmtClusterCPUArch(ctx)
```
enxebre, line 364: there's a client already built for this reconciler this should be using instead of let this func recreate one every time

hypershift-operator/controllers/nodepool/nodepool_controller.go
```diff
@@ -355,6 +356,22 @@ func (r *NodePoolReconciler) reconcile(ctx context.Context, hcluster *hyperv1.Ho
 		return ctrl.Result{}, fmt.Errorf("failed to look up release image metadata: %w", err)
 	}
 
+	// Validate the NodePool and management cluster CPU architectures match if the HostedCluster does not support multi-arch NodePools
```
enxebre, line 359: This logic needs to be unit tested

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -4078,38 +4103,29 @@ func (r *HostedClusterReconciler) validateReleaseImage(ctx context.Context, hc *
...
 
-		for _, nodePool := range nodePoolList.Items {
-			if err := hyperutil.DoesMgmtClusterAndNodePoolCPUArchMatch(mgmtClusterCPUArch, nodePool.Spec.Arch); err != nil {
-				return err
-			}
-		}
+	return supportedversion.IsValidReleaseVersion(&version, currentVersion, &supportedversion.LatestSupportedVersion, &minSupportedVersion, hc.Spec.Networking.NetworkType, hc.Spec.Platform.Type)
+}
 
+// isMultiArchNPCapable returns true if the HostedCluster release image has a multi-arch payload and false if it is a
```
enxebre, line 4111: based on what it does shouldn't be just named isMultiArchHC? or even isMultiArch?

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -1018,11 +1017,37 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
 		}
 	}
 
+	// Set SupportsMultiArchNodePools capable condition
+	{
+		condition := meta.FindStatusCondition(hcluster.Status.Conditions, hyperv1.SupportsMultiArchNodePools)
+
+		if condition == nil || condition.ObservedGeneration != hcluster.Generation || condition.Status != metav1.ConditionTrue {
```
enxebre, line 1024: why the check for condition.Status != metav1.ConditionTrue?

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -1018,11 +1017,37 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
...
+			condition := metav1.Condition{
```
enxebre, line 1025: Why is this a condition vs just doing all validating in flight? This is actually using bool state but then still need in flight validation to match the nodepool and HC arch when not multi arch.
If we choose persisting state why not use a status field with enum for cpus so there's no need for inflight operations at all?

cmd/cluster/aws/create.go
bryan-cox: This has been removed and now only core Validate does the check.

cmd/cluster/azure/create.go
```diff
@@ -48,6 +48,8 @@ func bindCoreOptions(opts *RawCreateOptions, flags *pflag.FlagSet) {
 	flags.StringVar(&opts.NetworkSecurityGroupID, "network-security-group-id", opts.NetworkSecurityGroupID, "The Network Security Group ID to use in the default NodePool.")
 	flags.StringToStringVarP(&opts.ResourceGroupTags, "resource-group-tags", "t", opts.ResourceGroupTags, "Additional tags to apply to the resource group created (e.g. 'key1=value1,key2=value2')")
 	flags.StringVar(&opts.SubnetID, "subnet-id", opts.SubnetID, "The subnet ID where the VMs will be placed.")
+	flags.StringVar(&opts.PayloadArch, "payload-arch", opts.PayloadArch, "Specifies if a single CPU architecture payload or a multi-arch CPU payload is being used for the HostedCluster and HostedControlPlane (e.g. Single, Multi)")
```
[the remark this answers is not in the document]
bryan-cox: That is leftover from the previous work where we were going to have the flag in the Azure API. It's been removed now.

cmd/cluster/azure/create.go
bryan-cox: One was a newer version written during the course of this PR. It's been removed and there is only one method now.

cmd/cluster/azure/create.go
bryan-cox: This has been removed

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
bryan-cox: That was a copy/paste leftover when I copied this from https://github.com/openshift/hypershift/blob/04e367810b6553909ea00e96c4ccb91e48f5dc5e/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go#L1012.

bryan-cox: It's been removed

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
bryan-cox: I have no problem renaming this. It's now `isMultiArchHC `.

hypershift-operator/controllers/nodepool/nodepool_controller.go
bryan-cox: Moved this into its own function to do so. Will add a unit test before merging PR.

cmd/cluster/core/create.go
bryan-cox: This method requires a valid mgmt cluster. Is there a way to unit test something like that?

cmd/cluster/core/create.go
```diff
@@ -930,3 +939,43 @@ func parseTolerationString(str string) (*corev1.Toleration, error) {
 
 	return &toleration, nil
 }
+
+// validateMgmtClusterAndNodePoolCPUArchitectures checks if a multi-arch release image or release stream was provided.
+// If none were provided, checks to make sure the NodePool CPU arch and the management cluster CPU arch; if they do not,
```
Patryk-Stefanski, line 944:
```suggestion
// If none were provided, checks to make sure the NodePool CPU arch and the management cluster CPU arch match; if they do not,
```

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -2595,6 +2597,12 @@ type HostedClusterStatus struct {
 	// +patchStrategy=merge
 	Conditions []metav1.Condition `json:"conditions,omitempty"`
 
+	// Payload represents the CPU architecture type of the HostedCluster.Spec.Release.Image
+	// +kubebuilder:validation:Enum=multi;arm64;amd64;ppc64le
```
JoelSpeed, line 2601: Enum values in Kube APIs should be pascal case
```suggestion
	// +kubebuilder:validation:Enum=Multi;ARM64;AMD64;PPC64LE
```

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -2595,6 +2597,12 @@ type HostedClusterStatus struct {
...
+	// +kubebuilder:validation:Optional
+	// +optional
```
JoelSpeed, line 2603: You only need one of these
```suggestion
	// +optional
```

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -2595,6 +2597,12 @@ type HostedClusterStatus struct {
 	// +patchStrategy=merge
 	Conditions []metav1.Condition `json:"conditions,omitempty"`
 
+	// Payload represents the CPU architecture type of the HostedCluster.Spec.Release.Image
```
JoelSpeed, line 2600: Should specify in the godoc which options are valid, users won't know the valid values based on the validations in the CRD unless we document them

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1338,6 +1338,8 @@ type AWSPlatformSpec struct {
 
 	// MultiArch specifies whether the Hosted Cluster will be expected to support NodePools with different
 	// CPU architectures, i.e., supporting arm64 NodePools and supporting amd64 NodePools on the same Hosted Cluster.
+	// Deprecated: This field is no longer used. The HyperShift Operator now performs multi-arch validations despite the
+	// platform type.
```
JoelSpeed, line 1342: What is the upgrade story for users already relying on this field?

bryan-cox: This flag was previously used to validate a multi-arch image was used for the HC when this flag was enabled. We perform multi-arch validation between the HC release image and the NodePool CPU in the NodePool controller now.

api/hypershift/v1alpha1/hostedcluster_types.go
```diff
@@ -2019,6 +2019,12 @@ type HostedClusterStatus struct {
 	// +patchStrategy=merge
 	Conditions []metav1.Condition `json:"conditions,omitempty"`
 
+	// Payload represents the CPU architecture type of the HostedCluster.Spec.Release.Image. The valid values are:
+	// Multi, ARM64, AMD64, or PPC64LE.
+	// +kubebuilder:validation:Enum=Multi;ARM64;AMD64;PPC64LE
+	// +optional
+	Payload string `json:"payload,omitempty"`
```
devguyio, line 2026: Would it make sense to use another name that represents that this is the arch? `PayloadArch` or `ReleaseArch` maybe?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -2595,6 +2597,12 @@ type HostedClusterStatus struct {
 	// +patchStrategy=merge
 	Conditions []metav1.Condition `json:"conditions,omitempty"`
 
+	// Payload represents the CPU architecture type of the HostedCluster.Spec.Release.Image. The valid values are:
+	// Multi, ARM64, AMD64, or PPC64LE.
+	// +kubebuilder:validation:Enum=Multi;ARM64;AMD64;PPC64LE
+	// +optional
+	Payload string `json:"payload,omitempty"`
```
devguyio, line 2604: Same comment as above

api/hypershift/v1alpha1/hostedcluster_types.go
devguyio: Why are we adding the field in the `v1alhpa1` if we already have a `v1beta1` API ?

devguyio (commented):
I just reviewed the proposed API change. I've a generic question though, why are we adding the field in the `v1alhpa1`  APIs if we already have a `v1beta1` API ?

api/hypershift/v1beta1/hostedcluster_types.go
bryan-cox: I don't have any strong opinions. If there is a consensus on changing this, I'm happy to do so.

bryan-cox:
> I just reviewed the proposed API change. I've a generic question though, why are we adding the field in the `v1alhpa1` APIs if we already have a `v1beta1` API ?

@devguyio - We fail unit tests at the moment if there isn't consistency between the two APIs for new fields like this.

api/hypershift/v1alpha1/hostedcluster_types.go
muraee: +1

hypershift-operator/controllers/nodepool/nodepool_controller.go
```diff
@@ -2916,3 +2929,16 @@ func deleteConfigByLabel(ctx context.Context, c client.Client, lbl map[string]st
 	}
 	return nil
 }
+
+// validateHCPayloadSupportsNodePoolCPUArch validates the HostedCluster payload can support the NodePool's CPU arch
+func validateHCPayloadSupportsNodePoolCPUArch(hc *hyperv1.HostedCluster, np *hyperv1.NodePool) error {
+	if strings.EqualFold(hc.Status.Payload, "multi") {
```
muraee, line 2935: maybe we should have a const for "multi"

hypershift-operator/controllers/nodepool/nodepool_controller.go
```diff
@@ -334,6 +334,19 @@ func (r *NodePoolReconciler) reconcile(ctx context.Context, hcluster *hyperv1.Ho
 		return ctrl.Result{}, fmt.Errorf("failed to look up release image metadata: %w", err)
 	}
 
+	// Validate the HostedCluster payload can support the NodePool's CPU arch
+	if err = validateHCPayloadSupportsNodePoolCPUArch(hcluster, nodePool); err != nil {
+		SetStatusCondition(&nodePool.Status.Conditions, hyperv1.NodePoolCondition{
+			Type:               hyperv1.NodePoolValidArchPlatform,
+			Status:             corev1.ConditionFalse,
+			Reason:             hyperv1.NodePoolInvalidArchPlatform,
+			Message:            fmt.Sprintf("NodePool CPU arch %s cannot be supported by the HostedCluster payload type: %s", nodePool.Spec.Arch, hcluster.Status.Payload),
```
muraee, line 343: you could just use `err.Error()` as the message here.
How about including the solution in the message? like `either change your NodePool arch or use a multi release image`

api/hypershift/v1beta1/hostedcluster_types.go
devguyio: Cool. I'd have questions though if the decision is to continue with this name 🙂

devguyio:
> > I just reviewed the proposed API change. I've a generic question though, why are we adding the field in the `v1alhpa1` APIs if we already have a `v1beta1` API ?
> 
> @devguyio - We fail unit tests at the moment if there isn't consistency between the two APIs for new fields like this.

This is a bit surprising, can you please elaborate more?

api/hypershift/v1alpha1/hostedcluster_types.go
bryan-cox: Updated

hypershift-operator/controllers/nodepool/nodepool_controller.go
bryan-cox: I don't think it matters in this case. The payload is `multi`. Joel said the enum in the API should be capitalized so that introduces a discrepancy between those two things.

hypershift-operator/controllers/nodepool/nodepool_controller.go
bryan-cox: Updated

hypershift-operator/controllers/nodepool/nodepool_controller.go
muraee: then the const could be "Multi".
just a nit, no need to change it.

docs/content/how-to/common/multi-arch-on-hcp.md
```diff
@@ -0,0 +1,41 @@
+# Multi-arch on Hosted Control Planes
+## General
+Several platforms now support multi-arch capable HostedClusters when a multi-arch release image or stream is used for 
+the HostedCluster. This means a HostedCluster can manage NodePools with different CPU architectures. 
+
+!!! note
+
+    An individual NodePool only supports one CPU architecture and cannot support multiple CPU archiectures within the 
+    same NodePool.
```
devguyio, line 9:
```suggestion
    An individual NodePool only supports one CPU architecture and cannot support multiple CPU architectures within the 
    same NodePool.
```

bryan-cox: I do have a PR that should catch things like this in the future 😅 - https://github.com/openshift/hypershift/pull/4700

bryan-cox: fixed

cmd/cluster/aws/create.go
```diff
@@ -376,6 +368,8 @@ func bindCoreOptions(opts *RawCreateOptions, flags *flag.FlagSet) {
 	flags.StringVar(&opts.CredentialSecretName, "secret-creds", opts.CredentialSecretName, "A Kubernetes secret with needed AWS platform credentials: sts-creds, pull-secret, and a base-domain value. The secret must exist in the supplied \"--namespace\". If a value is provided through the flag '--pull-secret', that value will override the pull-secret value in 'secret-creds'.")
 	flags.StringVar(&opts.IssuerURL, "oidc-issuer-url", "", "The OIDC provider issuer URL")
 	flags.BoolVar(&opts.MultiArch, "multi-arch", opts.MultiArch, "If true, this flag indicates the Hosted Cluster will support multi-arch NodePools and will perform additional validation checks to ensure a multi-arch release image or stream was used.")
+
+	_ = flags.MarkDeprecated("multi-arch", "Multi-arch validation is now performed automatically based on the release image.")
```
enxebre, line 372: And signaled in the HostedCluster.status

cmd/cluster/azure/create.go
```diff
@@ -101,8 +102,7 @@ func (o *RawCreateOptions) Validate(ctx context.Context, opts *core.CreateOption
 		},
 	}
 
-	var err error
-	validOpts.ValidatedAzurePlatformCreateOptions, err = o.NodePoolOpts.Validate()
+	validOpts.ValidatedAzurePlatformCreateOptions, err = o.NodePoolOpts.Validate(ctx, opts.Arch, opts.Name, opts.Namespace)
```
enxebre, line 105: can we instead pass down the whole opts *core.CreateOptions resource so we keep signatures consistent between platforms and we can then put an interface atop to keep implementations aligned and easier to follow and maintain

cmd/cluster/core/create.go
enxebre: As far as I can see it's a client making a request and parsing the response. So you can mock the client response and so document the expectation at the same time

cmd/cluster/core/create.go
```diff
@@ -943,3 +952,43 @@ func parseTolerationString(str string) (*corev1.Toleration, error) {
...
+	}
+
+	// If a release image/stream is not multi-arch, check the mgmt & NodePool CPU architectures match
+	if !validMultiArchImage {
+		mgmtClusterCPUArch, err := hyperutil.GetMgmtClusterCPUArch(ctx)
+		if err != nil {
+			return fmt.Errorf("failed to check mgmt cluster CPU arch: %v", err)
+		}
+
+		err = hyperutil.DoesMgmtClusterAndNodePoolCPUArchMatch(mgmtClusterCPUArch, opts.Arch)
```
enxebre, line 987: I think I would rather relax this req for now and not fail if opts.Arch is empty. This input is only leveraged to lookup an ami, which we could also infer by figuring the arch from the release image itself

cmd/nodepool/azure/create.go
```diff
@@ -99,7 +99,7 @@ type CompletedAzurePlatformCreateOptions struct {
 	*completetedAzurePlatformCreateOptions
 }
 
-func (o *RawAzurePlatformCreateOptions) Validate() (*ValidatedAzurePlatformCreateOptions, error) {
+func (o *RawAzurePlatformCreateOptions) Validate(ctx context.Context, arch, name, namespace string) (*ValidatedAzurePlatformCreateOptions, error) {
```
enxebre, line 102: See https://github.com/openshift/hypershift/pull/4538/files#r1764589754

cmd/nodepool/core/create.go
```diff
@@ -142,3 +144,29 @@ func (o *CreateNodePoolOptions) CreateNodePool(ctx context.Context, platformOpts
 	fmt.Printf("NodePool %s created\n", o.Name)
 	return nil
 }
+
+// ValidateMgmtClusterSupportsNodePoolCPUType validates the HostedCluster payload type can support the CPU architecture
+// of the NodePool.
```
enxebre, line 149: please add a unit test for this

cmd/nodepool/core/create.go
```diff
@@ -142,3 +144,29 @@ func (o *CreateNodePoolOptions) CreateNodePool(ctx context.Context, platformOpts
...
+	hc := &hyperv1.HostedCluster{}
+	err = client.Get(context.Background(), types.NamespacedName{Namespace: namespace, Name: name}, hc)
+	if err != nil {
+		// This is expected to happen when we create a cluster since there is no created HostedCluster CR to check the
+		// payload from.
+		logger.Info("WARNING: failed to get HostedCluster to check payload type")
+		return nil
+	}
+
+	if hc.Status.PayloadArch != "" && !strings.EqualFold(hc.Status.PayloadArch, "multi") && !strings.EqualFold(hc.Status.PayloadArch, arch) {
```
JoelSpeed, line 167: For the second part of this if condition, you could just compare against a const in the API, rather than a case insensitive match. Typically you'd create a string type alias for enums and have each valid value as an exported constant so that in code you can compare them.

Likewise where we have conversions from API casing to some other casing, this is normally done with a function that knows the correct conversions, with a switch statement, it doesn't matter too much here but for other places where it's not just a casing change that's what I'd suggest

cmd/cluster/azure/create.go
bryan-cox: If we want to do this, can we do it in a follow up ticket/PR?

The problem I ran into is the structs are different between cmd/cluster/azure/create.go (opts *core.CreateOptions) and when this same function is called in cmd/nodepool/azure/create.go (coreOpts *core.CreateNodePoolOptions). I think we would want to bring those two structs into alignment or just make one struct and use it in both places. To me, making that change is out of scope for this PR.

cmd/cluster/core/create.go
```diff
@@ -943,3 +952,43 @@ func parseTolerationString(str string) (*corev1.Toleration, error) {
 
 	return &toleration, nil
 }
+
+// validateMgmtClusterAndNodePoolCPUArchitectures checks if a multi-arch release image or release stream was provided.
```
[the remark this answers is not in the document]
bryan-cox: Yes, this is done a few lines down further - https://github.com/openshift/hypershift/blob/f508bfa6f378e0c2f39b2fd7aee791672639c385/cmd/cluster/core/create.go#L987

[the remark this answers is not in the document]
bryan-cox: Checking the release image matches the mgmt cluster CPU arch isn't a multi-arch thing. Can we follow up with that check in follow up ticket/PR?

cmd/cluster/core/create.go
bryan-cox: It's not only used for AMI for AWS. We also check the Arch is valid for a platform in `isArchAndPlatformSupported` in the NodePool Controller.

This field should never be empty since it gets a k8s default in the API and in the CLI.

We can't infer the arch from the release image when we are using multi-arch release images.

cmd/nodepool/core/create.go
bryan-cox: I'll add one. I was waiting to hear back on https://github.com/openshift/hypershift/pull/4538#discussion_r1764598254.

cmd/nodepool/core/create.go
bryan-cox: > you could just compare against a const in the API, rather than a case insensitive match.

It would still need to be a case insensitive check. The release payloads list the architectures in lowercase. You previously mentioned the enum in the API should have all the fields capitalized. So there's a discrepancy between those two things, which is why I think it would still need to be a case insensitive check.

cmd/nodepool/core/create.go
bryan-cox: Whoops wrong function

cmd/cluster/core/create.go
enxebre: I'm aware, I also think we should eventually drop this isArchAndPlatformSupported in favour of API validation and let that field have empty input. But anyways lets leave it as it's for now.

cmd/nodepool/core/create.go
enxebre: The enum in the status.payload is not consistent with the one in nodepool.spec.Arch one is using lowercase, other one is using uppercase
Also we already have declared consts in nodepool_types
	ArchitectureAMD64   = "amd64"
	ArchitectureS390X   = "s390x"
	ArchitecturePPC64LE = "ppc64le"
	ArchitectureARM64   = "arm64"

that can be leveraged for comparisons
Can we please also add multi to that list and use it consistently? I can see  several "multi" and "Multi" refs now

cmd/cluster/core/create.go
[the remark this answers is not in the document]
enxebre: can we please have a unit test for validateMgmtClusterAndNodePoolCPUArchitectures

cmd/cluster/core/create.go
bryan-cox: Ack

cmd/cluster/azure/create.go
enxebre: That can be solved with an interface that force both to implement GetClusterName, GetArch and GetNamespace. Can we at a minimum at a TODO and link a jira ticket capturing all this?

aside form this PR why isn't AzureMarketPlaceImageInfo a struct of AzurePlatformCreateOptions?

cmd/nodepool/core/create.go
JoelSpeed: I mean you would have a function, something along the lines of below, that makes explicit conversions rather than just relying on the assumption that casing is always comparable

```
func apiArchToManifestArch(apiArch apiv1.CPUArch) apiv1.ManifestArchitecture {
  switch apiArch {
  case apiv1.ArchitectureAMD64:
    return apiv1.ManifestArchitectureAMD64
  case apiv1.ArchitectureS390X:
    return apiv1.ManifestArchitectureS390X
  ...
  }
}
```

enxebre: Yes let's do that. We also have a similar ArchAliases for rhcos metadata.

bryan-cox: Yeah I will add that.

I will change the enums back to lowercase since that is what the OCP release image payload does. That makes the most sense to me.

bryan-cox: I think I got this all squared away in a new commit - [Add ArchitectureMulti to NodePool API](https://github.com/openshift/hypershift/pull/4538/commits/286e3e28e1bbee9a5f91e53f4c1c4083e985c21c). 

Please let me know if I missed something or misunderstood something 😄

api/hypershift/v1beta1/nodepool_types.go
```diff
@@ -14,10 +14,18 @@ import (
...
+	// See https://github.com/openshift/hypershift/pull/4538#discussion_r1765165827 for more details.
 	ArchitectureAMD64   = "amd64"
 	ArchitectureS390X   = "s390x"
 	ArchitecturePPC64LE = "ppc64le"
 	ArchitectureARM64   = "arm64"
+	ArchitectureMulti   = "multi"
+
+	ManifestArchitectureAMD64   = "AMD64"
+	ManifestArchitectureARM64   = "ARM64"
+	ManifestArchitecturePPC64LE = "PPC64L#"
```
JoelSpeed, line 27: Typo on this line

bryan-cox: 🤦🏻 - thanks, its fixed now

cmd/cluster/core/create_test.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	tests := []struct {
+		name        string
+		opts        *RawCreateOptions
+		expected    bool
+		expectError bool
+	}{
+		{
+			name: "Test case 1: Valid multi-arch image",
```
enxebre, line 36: Nice!
nit: no need to prefix with "test case n:"
Also can we rephrase to follow convention "When x it should y"? e.g. When a multi arch release image is given it should succeed with any NodePool Arch

cmd/cluster/core/create_test.go
```diff
@@ -0,0 +1,89 @@
...
+				Arch:           "arm64",
+			},
+			expectError: true,
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			g := NewWithT(t)
+			t.Setenv("FAKE_CLIENT", "true")
```
enxebre, line 80: why do you need to set this env variable? you decoupled the client now

support/util/util_test.go
```diff
@@ -485,3 +489,44 @@ func TestSanitizeIgnitionPayload(t *testing.T) {
...
+			Name:         "Nominal use case",
+			expectedArch: "amd64",
+			expectedErr:  false,
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.Name, func(t *testing.T) {
+			g := NewWithT(t)
+			t.Setenv("FAKE_CLIENT", "true")
```
enxebre, line 521: why do you need to set this env variable? you decoupled the client now

support/util/util.go
```diff
@@ -354,42 +355,39 @@ func ApplyAWSLoadBalancerSubnetsAnnotation(svc *corev1.Service, hcp *hyperv1.Hos
...
 
 	return nil
 }
 
-func GetMgmtClusterCPUArch(ctx context.Context) (string, error) {
-	cfg, err := util.GetConfig()
-	if err != nil {
-		return "", err
+func GetKubeClientSet() (kubeclient.Interface, error) {
+	if os.Getenv("FAKE_CLIENT") == "true" {
```
enxebre, line 367: why is this needed?

cmd/cluster/core/create_test.go
bryan-cox: This was left behind while I was debugging the unit tests. This has been removed now.

support/util/util.go
bryan-cox: This was left behind while I was debugging the unit tests. This has been removed now.

support/util/util_test.go
bryan-cox: This was left behind while I was debugging the unit tests. This has been removed now.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -4142,38 +4141,60 @@ func (r *HostedClusterReconciler) validateReleaseImage(ctx context.Context, hc *
...
+	if isMultiArchReleaseImage {
+		return "Multi", nil
+	}
 
-		for _, nodePool := range nodePoolList.Items {
-			if err := hyperutil.DoesMgmtClusterAndNodePoolCPUArchMatch(mgmtClusterCPUArch, nodePool.Spec.Arch); err != nil {
-				return err
-			}
-		}
+	arch, err := getImageArchitecture(ctx, hc.Spec.Release.Image, pullSecretBytes, imageMetadataProvider)
```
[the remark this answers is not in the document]
bryan-cox: Just because sometimes we don't have a release image in the CLI and we just have a release stream. In the CLI if it's a multi image or stream we don't query the mgmt cluster version.

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -2595,6 +2597,12 @@ type HostedClusterStatus struct {
 	// +patchStrategy=merge
 	Conditions []metav1.Condition `json:"conditions,omitempty"`
 
+	// payloadArch represents the CPU architecture type of the HostedCluster.Spec.Release.Image. The valid values are:
+	// Multi, ARM64, AMD64, or PPC64LE.
+	// +kubebuilder:validation:Enum=Multi;ARM64;AMD64;PPC64LE
+	// +optional
+	PayloadArch string `json:"payloadArch,omitempty"`
```
enxebre, line 2604: payloadArch should be type payloadArch which represents the enum

api/hypershift/v1beta1/nodepool_types.go
```diff
@@ -14,10 +14,18 @@ import (
...
 
 const (
+	// TODO - for the next API bump; the case on these needs to be changed.
+	// See https://github.com/openshift/hypershift/pull/4538#discussion_r1765165827 for more details.
 	ArchitectureAMD64   = "amd64"
 	ArchitectureS390X   = "s390x"
 	ArchitecturePPC64LE = "ppc64le"
 	ArchitectureARM64   = "arm64"
+	ArchitectureMulti   = "multi"
+
```
enxebre, line 24: where's the manfiest prefix coming from?

I'd expect we have something like
```
type payloadArch string

const (
	AMD64   payloadArch = "AMD64"
	PPC64LE payloadArch = "PPC64LE"
	ARM64   payloadArch = "ARM64"
	Multi   payloadArch = "Multi"
)

// toPayloadArch converts a string to payloadArch.
func toPayloadArch(arch string) payloadArch {
	switch strings.ToUpper(arch) {
	case string(AMD64):
		return AMD64
	case string(ARM64):
		return ARM64
	case string(PPC64LE):
		return PPC64LE
	case string("MULTI"):
		return Multi
	default:
		return ""
	}
}
```

Then any comparison must use toArch() or the consts
```
hc.Status.PayloadArch != Multi and hc.Status.PayloadArch != toArch(arch)
```

api/hypershift/v1beta1/hostedcluster_types.go
enxebre: https://github.com/openshift/hypershift/pull/4538/files#r1766918445

api/hypershift/v1beta1/nodepool_types.go
enxebre: Also why don't we have ArchitectureS390X in the new list?

api/hypershift/v1beta1/hostedcluster_types.go
bryan-cox: They don't have to set any value. It won't matter what value they set for this since the flag is not used in the code at all.

cmd/nodepool/core/create.go
```diff
@@ -142,3 +144,29 @@ func (o *CreateNodePoolOptions) CreateNodePool(ctx context.Context, platformOpts
 	fmt.Printf("NodePool %s created\n", o.Name)
 	return nil
 }
+
+// ValidateMgmtClusterSupportsNodePoolCPUType validates the HostedCluster payload type can support the CPU architecture
+// of the NodePool.
+func ValidateMgmtClusterSupportsNodePoolCPUType(ctx context.Context, name, namespace, arch string) error {
```
enxebre, line 150: we have
ValidateMgmtClusterSupportsNodePoolCPUType and validateMgmtClusterAndNodePoolCPUArchitectures
Then we have DoesMgmtClusterAndNodePoolCPUArchMatch

the latter is doing a plain comparison, so it could be just called archMatch or similar or is it even necessary?

The others two are confusing to me and seem unconsistent.

enxebre: also ValidateMgmtClusterSupportsNodePoolCPUType needs a unit too

api/hypershift/v1beta1/nodepool_types.go
bryan-cox: I dropped the previous list

cmd/nodepool/aws/create.go
```diff
@@ -60,6 +62,16 @@ func (o *AWSPlatformCreateOptions) UpdateNodePool(ctx context.Context, nodePool
 		}
 	}
 
+	client, err := util.GetClient()
+	if err != nil {
+		return err
+	}
+
+	// Validate HostedCluster payload can support the NodePool CPU type
```
enxebre, line 70: UpdateNodePool is not meant to run validations. Can you please be consistent with what you are doing in azure and the structure pattern introduced by
https://github.com/openshift/hypershift/pull/4018
https://github.com/openshift/hypershift/pull/4212

i.e in NewCreateCommand something like
```
	cmd.RunE = func(cmd *cobra.Command, args []string) error {
		// Validate HostedCluster payload can support the NodePool CPU type
		if err = core.ValidateHostedClusterPayloadSupportsNodePoolCPUType(ctx, client, hcluster.Name, hcluster.Namespace, nodePool.Spec.Arch); err != nil {
			return err
		}

		return coreOpts.CreateRunFunc(platformOpts)(cmd, args)
	}

	return cmd
```

After merging this (or any) PR the cli based code can only be better, introducing more tech debt is not acceptable

cmd/nodepool/core/create.go
enxebre: Why are we using a new term "cpu type" when we are using "arch" everywhere else?
This PR is using CPUArch cpuType and arch in multiple places to refer to the same thing, can we please just use arch consistently which also matches the api fields naming?

cmd/nodepool/aws/create.go
enxebre: Also why is this being added to this two particular platforms instead of to the core validation?

cmd/nodepool/core/create.go
```diff
@@ -142,3 +142,24 @@ func (o *CreateNodePoolOptions) CreateNodePool(ctx context.Context, platformOpts
...
+	hc := &hyperv1.HostedCluster{}
+	err := client.Get(context.Background(), types.NamespacedName{Namespace: namespace, Name: name}, hc)
+	if err != nil {
+		// This is expected to happen when we create a cluster since there is no created HostedCluster CR to check the
+		// payload from.
+		logger.Info("WARNING: failed to get HostedCluster to check payload type")
+		return nil
+	}
+
+	if hc.Status.PayloadArch != "" && hc.Status.PayloadArch != hyperv1.Multi && hc.Status.PayloadArch != hyperv1.ToPayloadArch(arch) {
```
enxebre, line 160: This has a race for when hc.Status.PayloadArch is not set yet. Can we account for that use case and let it succeed with a warning?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -2551,6 +2554,31 @@ type AESCBCSpec struct {
 	BackupKey *corev1.LocalObjectReference `json:"backupKey,omitempty"`
 }
 
+type PayloadArchType string
+
+const (
+	AMD64   PayloadArchType = "AMD64"
+	PPC64LE PayloadArchType = "PPC64LE"
```
enxebre, line 2561: is this intentionally missing S390X?

api/hypershift/v1beta1/hostedcluster_types.go
JoelSpeed: When the user has configured here to use a multi-arch payload, how do they represent that in the new API?

When the user has configured here _not_ to use a multi-arch payload, how do they represent that in the new API?

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -1036,11 +1035,19 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
 		}
 	}
 
+	// Set HostedCluster payload type
+	payloadType, err := r.determineHostedClusterPayloadType(ctx, hcluster, registryClientImageMetadataProvider)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	log.Info("HostedCluster payload type set to: " + string(payloadType))
```
enxebre, line 1044: This would be a verbosity logging every single reconcile, can we drop it?

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -1036,11 +1035,19 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
 		}
 	}
 
+	// Set HostedCluster payload type
+	payloadType, err := r.determineHostedClusterPayloadType(ctx, hcluster, registryClientImageMetadataProvider)
+	if err != nil {
+		return ctrl.Result{}, err
```
enxebre, line 1041: can we make this function bubble up in the validReleaseImageCondition above?

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -4142,38 +4141,60 @@ func (r *HostedClusterReconciler) validateReleaseImage(ctx context.Context, hc *
...
-		}
+	minSupportedVersion := supportedversion.GetMinSupportedVersion(hc)
 
-		if !isMultiArchReleaseImage {
-			return fmt.Errorf("release image is not a multi-arch image")
-		}
+	return supportedversion.IsValidReleaseVersion(&version, currentVersion, &supportedversion.LatestSupportedVersion, &minSupportedVersion, hc.Spec.Networking.NetworkType, hc.Spec.Platform.Type)
+}
+
+// determineHostedClusterPayloadType returns the HostedCluster payload's CPU architecture type
```
enxebre, line 4149: This uses PayloadType / CPU architecture and sets it to .payloadArch
can we please use a single term consistently across the base code?

hypershift-operator/controllers/nodepool/nodepool_controller.go
```diff
@@ -321,6 +321,19 @@ func (r *NodePoolReconciler) reconcile(ctx context.Context, hcluster *hyperv1.Ho
 		return ctrl.Result{}, fmt.Errorf("failed to look up release image metadata: %w", err)
 	}
 
+	// Validate the HostedCluster payload can support the NodePool's CPU arch
+	if err = validateHCPayloadSupportsNodePoolCPUArch(hcluster, nodePool); err != nil {
```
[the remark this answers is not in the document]
enxebre: Also this check needs to be grouped with one below for 	if !isArchAndPlatformSupported(nodePool) {

enxebre:
can we please include an e2e test for possibly in hypershiftTest.after() that ensures this is set as expected?

api/hypershift/v1beta1/hostedcluster_types.go
bryan-cox: It's been included now

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -1036,11 +1035,19 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
 		}
 	}
 
+	// Set HostedCluster payload type
+	payloadType, err := r.determineHostedClusterPayloadType(ctx, hcluster, registryClientImageMetadataProvider)
```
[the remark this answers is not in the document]
bryan-cox: Updated

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
bryan-cox: Removed

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
bryan-cox: Updated

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
bryan-cox: Updated

hypershift-operator/controllers/nodepool/nodepool_controller.go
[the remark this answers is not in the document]
bryan-cox: > This would temporarily short circuit every NodePool across the fleet until the field is set.
Wouldn't it only affect the particular NodePool that is reconciling here? 

Updated to set a status condition on success.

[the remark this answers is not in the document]
enxebre: >Wouldn't it only affect the particular NodePool that is reconciling here?

When you upgrade the image in a Deployment. It does a rolling upgrade of the Pods. The pod start running a the new controller with new code. As the controller starts it reconcile all existing CRs is watching.

cmd/nodepool/aws/create.go
bryan-cox: There was no core validation for NodePool. I added a function for this and removed the call to validateHostedClusterPayloadSupportsNodePoolCPUType in those two platforms.

cmd/nodepool/core/create.go
bryan-cox: Updated

cmd/nodepool/core/create.go
bryan-cox: Updated

support/util/util_test.go
```diff
@@ -485,3 +464,279 @@ func TestSanitizeIgnitionPayload(t *testing.T) {
...
+				g.Expect(err).To(HaveOccurred())
+			} else {
+				g.Expect(err).To(BeNil())
+				g.Expect(arch).To(Equal(tc.expectedArch))
+			}
+		})
+	}
+}
+
+func TestDetermineHostedClusterPayloadType(t *testing.T) {
```
enxebre, line 656: typo on this test name TestDetermineHostedClusterPayloadType/ TestDetermineHostedClusterPayloadArch/

cmd/nodepool/azure/create.go
```diff
@@ -144,7 +145,8 @@ func NewCreateCommand(coreOpts *core.CreateNodePoolOptions) *cobra.Command {
 	}
 	BindDeveloperOptions(platformOpts, cmd.Flags())
 	cmd.RunE = func(cmd *cobra.Command, args []string) error {
-		validOpts, err := platformOpts.Validate()
+		ctx := cmd.Context()
+		validOpts, err := platformOpts.Validate(ctx, coreOpts.Arch, coreOpts.ClusterName, coreOpts.Namespace)
```
enxebre, line 149: these params are unused now, can we drop them?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1338,6 +1339,8 @@ type AWSPlatformSpec struct {
 
 	// MultiArch specifies whether the Hosted Cluster will be expected to support NodePools with different
 	// CPU architectures, i.e., supporting arm64 NodePools and supporting amd64 NodePools on the same Hosted Cluster.
+	// Deprecated: This field is no longer used. The HyperShift Operator now performs multi-arch validations despite the
```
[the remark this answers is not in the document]
bryan-cox: Updated

bryan-cox:
The e2e's seem to be legitimately failing on `TestHAEtcdChaos ` and `TestNoneCreateCluster `:
```
TestHAEtcdChaos/EnsureHostedCluster expand_less | 0s
-- | --
{Failed  === RUN   TestHAEtcdChaos/EnsureHostedCluster     util.go:1873:          Expected             <v1beta1.PayloadArchType>: AMD64         to equal             <v1beta1.PayloadArchType>:      --- FAIL: TestHAEtcdChaos/EnsureHostedCluster (0.22s) }
```

bryan-cox:
When I check the TestHAEtcdChaos  and TestNoneCreateCluster HC, there is a `hc.status.payloadarch`. Not sure why these two tests are failing 🤔

test/e2e/util/util.go
```diff
@@ -1841,3 +1842,27 @@ func EnsureSATokenNotMountedUnlessNecessary(t *testing.T, ctx context.Context, c
 		}
 	})
 }
+
+func EnsurePayloadArchSetCorrectly(t *testing.T, ctx context.Context, client crclient.Client, hostedCluster *hyperv1.HostedCluster) {
+	EventuallyObject(t, ctx, fmt.Sprintf("HostedCluster %s/%s to have valid Status.Payload", hostedCluster.Namespace, hostedCluster.Name),
```
sjenning, line 1847: needs to be wrapped in a `t.Run()` like the other `Ensure*` assertions

sjenning (changes requested):

sjenning (approved):

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -2551,6 +2556,34 @@ type AESCBCSpec struct {
...
+	AMD64   PayloadArchType = "AMD64"
+	PPC64LE PayloadArchType = "PPC64LE"
+	S390X   PayloadArchType = "S390X"
+	ARM64   PayloadArchType = "ARM64"
+	Multi   PayloadArchType = "Multi"
+)
+
+// ToPayloadArch converts a string to payloadArch.
+func ToPayloadArch(arch string) PayloadArchType {
+	switch strings.ToUpper(arch) {
```
JoelSpeed, line 2571: This doesn't work for the Multi case does it? Multi isn't actually uppercased

cmd/cluster/aws/create.go
```diff
@@ -59,14 +57,9 @@ type ValidatedCreateOptions struct {
 
 func (o *RawCreateOptions) Validate(ctx context.Context, opts *core.CreateOptions) (core.PlatformCompleter, error) {
 	// Perform multi-arch validations except for render
-	if !opts.Render {
-		err := validateMultiArch(ctx, o, opts)
-		if err != nil {
-			return nil, err
-		}
-	}
+	var err error
```
JoelSpeed, line 60: Why not declare the error using the `:=` syntax on line 62?

cmd/cluster/core/create.go
```diff
@@ -597,6 +601,15 @@ func (opts *RawCreateOptions) Validate(ctx context.Context) (*ValidatedCreateOpt
 		} else if !apierrors.IsNotFound(err) {
 			return nil, fmt.Errorf("hostedcluster doesn't exist validation failed with error: %w", err)
 		}
+
+		// Validate multi-arch aspects
+		kc, err := hyperutil.GetKubeClientSet()
+		if err != nil {
+			return nil, err
+		}
+		if err = validateMgmtClusterAndNodePoolCPUArchitectures(ctx, opts, kc); err != nil {
```
JoelSpeed, line 610: Better to scope errors rather than reuse the parent scope
```suggestion
		if err := validateMgmtClusterAndNodePoolCPUArchitectures(ctx, opts, kc); err != nil {
```

cmd/cluster/core/create.go
```diff
@@ -597,6 +601,15 @@ func (opts *RawCreateOptions) Validate(ctx context.Context) (*ValidatedCreateOpt
 		} else if !apierrors.IsNotFound(err) {
 			return nil, fmt.Errorf("hostedcluster doesn't exist validation failed with error: %w", err)
 		}
+
+		// Validate multi-arch aspects
+		kc, err := hyperutil.GetKubeClientSet()
+		if err != nil {
+			return nil, err
```
JoelSpeed, line 608: Generally it's good practice to wrap errors in additional context rather than returning a naked error. It seems like you've done this in some places but not all right now?
```suggestion
			return nil, fmt.Errorf("could not retrieve kube clientset: %w", err)
```

api/hypershift/v1beta1/hostedcluster_types.go
bryan-cox: Good catch, will fix that.

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -2551,6 +2556,34 @@ type AESCBCSpec struct {
...
 }
 
+type PayloadArchType string
+
+const (
+	AMD64   PayloadArchType = "AMD64"
+	PPC64LE PayloadArchType = "PPC64LE"
+	S390X   PayloadArchType = "S390X"
+	ARM64   PayloadArchType = "ARM64"
+	Multi   PayloadArchType = "MULTI"
```
JoelSpeed, line 2566: This isn't correct now, this should be pascal case to match the API enum (which is currently correct)

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -2551,6 +2556,34 @@ type AESCBCSpec struct {
...
+)
+
+// ToPayloadArch converts a string to payloadArch.
+func ToPayloadArch(arch string) PayloadArchType {
+	switch strings.ToUpper(arch) {
```
JoelSpeed, line 2571: You can't rely on `strings.ToUpper` here, unless you also `strings.ToUpper` inside the case statements, but, realistically, this would be better to just be explicit, don't use a transform to avoid issues 

```
func ToPayloadArch(arch string) PayloadArchType {
  switch arch {
  case "amd64":
    return AMD64
  ...
  }
}
```

JoelSpeed (commented):
/hold constants and API no longer match, needs reverting to PascalCase

enxebre:
/retitle Add multi-arch validation for HC/NodePool compatibility

PR #4546, opened 2024-08-13
OCPBUGS-39525: Fix uses of MustParse* on non-constant input

mdbooth:
This PR systematically replaces all uses of MustParse functions with non-constant input with their Parse equivalent, except in one case where the containing function does not have an error return. We add a warning comment there instead.

2 of these in particular may allow a user who is able to create a HostedCluster to cause the HostedCluster controller to panic-loop. Specifically the following 2 fields are not validated by the API server on admission, and are parsed with MustParse:

* HostedCluster.Spec.Networking.APIServer.AdvertiseAddress
* HostedCluster.Spec.Networking.ClusterNetwork.CIDR

Fixes #

**Checklist**

### DIFF
--- a/cmd/nodepool/kubevirt/create.go
+++ b/cmd/nodepool/kubevirt/create.go
@@ -291,7 +291,8 @@ func (o *KubevirtPlatformCreateOptions) NodePoolPlatform() *hyperv1.KubevirtNode
 	}
 
 	if o.Memory != "" {
-		memory := apiresource.MustParse(o.Memory)
+		// TODO: add a debug trace for this error
+		memory, _ := apiresource.ParseQuantity(o.Memory)
 		platform.Compute.Memory = &memory
 	}
 	if o.Cores != 0 {
--- a/control-plane-operator/controllers/hostedcontrolplane/cno/clusternetworkoperator.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/cno/clusternetworkoperator.go
@@ -448,7 +448,11 @@ kubectl --kubeconfig $kc config use-context default`,
 		},
 	}
 
-	if semver.MustParse(params.ReleaseVersion).Minor == uint64(12) {
+	parsedReleaseVersion, err := semver.Parse(params.ReleaseVersion)
+	if err != nil {
+		return fmt.Errorf("parsing ReleaseVersion (%s): %w", params.ReleaseVersion, err)
+	}
+	if parsedReleaseVersion.Minor == uint64(12) {
 		dep.Spec.Template.Spec.InitContainers = append(dep.Spec.Template.Spec.InitContainers, corev1.Container{
 			Command: []string{"/bin/bash"},
 			Args: []string{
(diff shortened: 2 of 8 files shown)


cmd/nodepool/kubevirt/create.go
```diff
@@ -291,6 +291,7 @@ func (o *KubevirtPlatformCreateOptions) NodePoolPlatform() *hyperv1.KubevirtNode
 	}
 
 	if o.Memory != "" {
+		// NOTE(mdbooth): this use of MustParse is unsafe, but this method cannot return an error
```
jparrill, line 294: We are looking to someway add debug traces changing the log framework, so we can add this as a debug trace when the times come. Could you please rephrase with something like :

// TODO: add a debug trace for this error

control-plane-operator/controllers/hostedcontrolplane/cno/clusternetworkoperator.go
```diff
@@ -446,7 +446,11 @@ kubectl --kubeconfig $kc config use-context default`,
 		},
 	}
 
-	if semver.MustParse(params.ReleaseVersion).Minor == uint64(12) {
+	parsedReleaseVersion, err := semver.Parse(params.ReleaseVersion)
+	if err != nil {
+		return fmt.Errorf("parsing ReleaseVersion (%s): %w", params.ReleaseVersion, err)
+	}
+	if parsedReleaseVersion.Minor == uint64(12) {
```
jparrill, line 453: Hey @csrwng do we still need this check? AFAIK IBM moved from 4.12 some time ago in all regions.

jparrill (commented):
Hey @mdbooth thank you for the PR. Overall looks good, dropped a question for Cesar around a version exception that maybe is not necessary now and a small nit to rephrase a comment for a future refactor.

bryan-cox:
@mdbooth - would you also open an OCPBUG ticket for this and retitle this PR to include it please?

mdbooth:
> @mdbooth - would you also open an OCPBUG ticket for this and retitle this PR to include it please?

Sure. This was originally just going to be a drive-by, but I guess it's large enough for it now.

bryan-cox:
> > @mdbooth - would you also open an OCPBUG ticket for this and retitle this PR to include it please?
> 
> Sure. This was originally just going to be a drive-by, but I guess it's large enough for it now.

Thanks! Yeah unless its something really small, we require a Jira ticket ( 3. in https://hypershift-docs.netlify.app/contribute/#creating-a-pull-request)

jparrill:
@mdbooth Let me know when the Jira ticket was sent in order to continue with the process.

control-plane-operator/controllers/hostedcontrolplane/cno/clusternetworkoperator.go
mdbooth: Either way, I'd be inclined to separate a functional change to a different PR. This will make is easier to backport/revert them independently as required.

enxebre:
fwiw the IPNet net.IPNet MarshalJSON/UnmarshalJSON should prevent bad input from passing beyond API persistence. Still this is good, thanks!
/lgtm

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -4542,8 +4541,6 @@ func checkAdvertiseAddressOverlapping(hc *hyperv1.HostedCluster) field.ErrorList
 		return errs
 	}
 
-	advAddress = netip.MustParseAddr(advAddr.String())
```
mdbooth, line 4545: @enxebre To answer your question above about the change in findAdvertiseAddress, it was to remove the requirement to convert the output a second time with MustParseAddr() here. The code actually wants a netip.Addr, so it's simpler to return that in the first instance.

PR #4554, opened 2024-08-13
HOSTEDCP-1850: HyperShift operator: Allow setting Kube APIServer maximum requests inflight

csrwng:
Adds support to the HyperShift operator for overriding default maximum requests in flight for Kubernetes API server.
Adds fields to the ClusterSizingConfiguration CRD that allows setting values for these based on cluster size.

This is the hypershift operator half of https://github.com/openshift/hypershift/pull/4488

Fixes #[HOSTEDCP-1850](https://issues.redhat.com/browse/HOSTEDCP-1850)

**Checklist**


hypershift-operator/controllers/scheduler/dedicated_request_serving_nodes.go
```diff
@@ -771,6 +771,12 @@ func (r *DedicatedServingComponentSchedulerAndSizer) updateHostedCluster(ctx con
 			delete(hc.Annotations, hyperv1.MachineHealthCheckTimeoutAnnotation)
 		}
 	}
+	if sizeConfig.Effects != nil && sizeConfig.Effects.MaximumRequestsInflight != nil {
+		hc.Annotations[hyperv1.KubeAPIServerMaximumRequestsInFlight] = fmt.Sprint(*sizeConfig.Effects.MaximumRequestsInflight)
```
jparrill, line 775: It's there any kind of validation around the annotation content?

jparrill (commented):
Small question, otherwise lgtm

hypershift-operator/controllers/scheduler/dedicated_request_serving_nodes.go
csrwng: The CR ensures that it is a number. As far as limiting the values for that number, I'm not sure that we know just yet what those limits should be. This is an API that is meant to be consumed by the service provider (ie. fleet manager), so I'm less worried about validation here.

PR #4565, opened 2024-08-16
HOSTEDCP-1531: Implement Managed Identities in HyperShift

bryan-cox:
This PR implements the following components on the HCP to use MSI:
* Azure cloud provider
* Cluster API Azure (CAPZ)
* KMS
* Control Plane Operator (CPO)

Fixes [HOSTEDCP-1531](https://issues.redhat.com/browse/HOSTEDCP-1531)

**Checklist**


hypershift-operator/controllers/hostedcluster/internal/platform/azure/azure.go
```diff
@@ -254,10 +237,9 @@ func reconcileAzureClusterIdentity(ctx context.Context, c client.Client, hcluste
 	}
 
 	azureClusterIdentity.Spec = capiazure.AzureClusterIdentitySpec{
-		ClientID:     string(credentialsSecret.Data["AZURE_CLIENT_ID"]),
-		ClientSecret: corev1.SecretReference{Name: "azure-client-secret", Namespace: controlPlaneNamespace},
-		TenantID:     string(credentialsSecret.Data["AZURE_TENANT_ID"]),
-		Type:         capiazure.ServicePrincipal,
+		ClientID: hcluster.Spec.Platform.Azure.MSIClientIDs.ClusterAPIAzureMSIClientID,
+		TenantID: string(credentialsSecret.Data["AZURE_TENANT_ID"]),
```
bryan-cox, line 241: TODO - is this even needed?

bryan-cox: It isn't so I removed it.

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1850,6 +1851,43 @@ type AzurePlatformSpec struct {
...
+	// azureCloudProviderMSIClientID is the client ID of a pre-existing managed identity ID of that will be associated
+	// with the azure cloud provider, aka ccm. The managed identity is expected to be in
+	// AzurePlatformSpec.ResourceGroupName.
+	//
+	// +kubebuilder:validation:Required
+	// +required
+	AzureCloudProviderMSIClientID string `json:"azureCloudProviderMSIClientID"`
+
+	// clusterAPIAzureMSIClientID is the client ID of a pre-existing managed identity ID of that will be associated
+	// with cluster-api azure. The managed identity is expected to be in AzurePlatformSpec.ResourceGroupName.
```
miguelsorianod, line 1873: > The managed identity is expected to be in AzurePlatformSpec.ResourceGroupName.

I think we need to make this more flexible.  The `AzurePlatformSpec.ResourceGroupName.` value that we set to in the HostedCluster from CS is the Managed Resource Group Name. The end-user will precreate the Azure Managed Identities which means that they will *always* be outside of the managed resource group.

This comment also applies to all the other attributes where this restriction has been documented.

The same approach should be followed for the data plane identities too (which I understand is being worked on in another PR).

cc @bennerv

bryan-cox: Yeah, Ben and I have discussed they would be in a different resource group. The API will be updated before this PR is ready for review.

Further work on this PR is stalled until we receive the sidecar container.

miguelsorianod: :+1: thanks

bryan-cox:
/close
Most of this work is being integrated into https://github.com/openshift/hypershift/pull/4690

PR #4578, opened 2024-08-19
HOSTEDCP-1868: Pause CAPI cluster when HostedCluster is paused

muraee:
Pausing the CAPI cluster object ensures all machines reconcilation is paused during backup if the NodePool was not explicitly paused. 

Fixes #

**Checklist**

### DIFF
--- a/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
+++ b/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
@@ -1072,6 +1072,9 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
 		if err := pauseHostedControlPlane(ctx, r.Client, hcp, hcluster.Spec.PausedUntil); err != nil {
 			return ctrl.Result{}, err
 		}
+		if err := pauseCAPICluster(ctx, r.Client, hcp); err != nil {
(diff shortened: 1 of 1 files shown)


hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -3036,6 +3039,8 @@ func reconcilecontrolPlaneOperatorIngressOperatorRoleBinding(binding *rbacv1.Rol
 func reconcileCAPICluster(cluster *capiv1.Cluster, hcluster *hyperv1.HostedCluster, hcp *hyperv1.HostedControlPlane, infraCR client.Object) error {
 	// We only create this resource once and then let CAPI own it
 	if !cluster.CreationTimestamp.IsZero() {
+		// make sure cluster is not paused.
+		cluster.Spec.Paused = false
```
csrwng, line 3043: this would only unpause if the deletion timestamp is non-zero, I think we want the opposite

muraee: this is checking the creationTimestamp, or are you refering to something else?

csrwng: sorry I immediately thought it was the deletionTimestamp

PR #4580, opened 2024-08-20
HOSTEDCP-1565: external-dns: add proxy env vars if cluster proxy is configured

sjenning:
Currently external-dns doesn't work if the cluster on which is runs has a proxy configured

### DIFF
--- a/cmd/install/assets/hypershift_operator.go
+++ b/cmd/install/assets/hypershift_operator.go
@@ -11,6 +11,7 @@ import (
 	"github.com/openshift/hypershift/support/config"
 	"github.com/openshift/hypershift/support/images"
 	"github.com/openshift/hypershift/support/metrics"
+	"github.com/openshift/hypershift/support/proxy"
 	"github.com/openshift/hypershift/support/rhobsmonitoring"
 	"github.com/openshift/hypershift/support/util"
 
@@ -28,6 +29,8 @@ import (
 	"github.com/google/uuid"
 	prometheusoperatorv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	cdicore "kubevirt.io/containerized-data-importer-api/pkg/apis/core"
+
+	configv1 "github.com/openshift/api/config/v1"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


cmd/install/install.go
```diff
@@ -608,6 +609,19 @@ func hyperShiftOperatorManifests(opts Options) ([]crclient.Object, []crclient.Ob
 	objects = append(objects, trustedCABundle)
 
 	if len(opts.ExternalDNSProvider) > 0 {
+		// Setting the proxy for external-dns is best-effort, ignore errors
```
bryan-cox, line 612: What error is being ignored?

cmd/install/assets/hypershift_operator.go
```diff
@@ -316,6 +320,11 @@ func (o ExternalDNSDeployment) Build() *appsv1.Deployment {
 		)
 	}
 
+	// Add proxy settings if cluster has a proxy
+	if o.Proxy != nil {
```
bryan-cox, line 324: Should you check for Status too? Looks like it's optional - https://github.com/openshift/hypershift/blob/e93a5e12c668b997519220e14db5798f008bcfa3/vendor/github.com/openshift/api/config/v1/types_proxy.go#L34.

cmd/install/install.go
```diff
@@ -608,6 +609,19 @@ func hyperShiftOperatorManifests(opts Options) ([]crclient.Object, []crclient.Ob
...
+		proxy, _ := func() (*configv1.Proxy, error) {
+			proxy := &configv1.Proxy{}
+			client, err := util.GetClient()
+			if err != nil {
+				return nil, err
+			}
+			if err := client.Get(context.TODO(), crclient.ObjectKey{Name: "cluster"}, proxy); err != nil {
```
bryan-cox, line 619: It's always named cluster?

sjenning: It is always named `cluster`

cmd/install/install.go
sjenning: Any failure to get the Proxy cluster config.  No/bad kubeconfig, kubeconfig to non-OCP cluster, etc.

cmd/install/assets/hypershift_operator.go
sjenning: It is optional, but it can't be nil (not a pointer).  `SetEnvVarsTo` does the Right Thing if the fields are empty strings.

PR #4582, opened 2024-08-20
OCPBUGS-38755: Remove go workspaces from source control

csrwng:
    - Removes the go.work and go.work.sum files from the repository.
    - Updates the Makefile to not use go workspaces.
    - Updates the vendored staticcheck so it won't fail on new style range
    statements.
    - Makes minor fixes flagged by staticcheck
    - Adds instructions for setting up a Go workspace in HACKING doc

Fixes #[OCPBUGS-38755](https://issues.redhat.com/browse/OCPBUGS-38755)

**Checklist**


devguyio:
@csrwng can you try to explain again what was wrong with what you proposed previously? other than go.work and go.work.sum files, which other files will show up as modified that shouldn't be pushed to the repo?

csrwng:
> can you try to explain again what was wrong with what you proposed previously? other than go.work and go.work.sum files, which other files will show up as modified that shouldn't be pushed to the repo?

The problem is trying to use those go.work files. When you run operations like `go work sync` and `go work vendor` you end up with a dirty repository.

bryan-cox:
/retest

For konflux

PR #4583, opened 2024-08-20
WIP: support: add http proxy aware dialer

sjenning:
The `proxy.Dial()` only reads the `ALL_PROXY` environment variable and thus does not use `HTTP(S)_PROXY` like we intend.

https://github.com/openshift/hypershift/blob/36b7eb630bb4202c05d5e10b213debf70afd8c77/support/konnectivityproxy/dialer.go#L247-L250

Add a new dialer that supports these variables for http(s) schemes.


sjenning:
/close
in favor of #4671

PR #4586, opened 2024-08-21
OSASINFRA-3572: e2e/openstack: handle DNS for ingress using route53

EmilienM:
This change is only for our CI.
Instead of handling the DNS prior to deploy a HostedCluster, we now leverage AWS Route53 service, using the CI credentials to create the `*.apps.<cluster-name>.<base-domain>` DNS record dynamically.
With that work, we'll be able to not only simplify our CI workflow (that used to create a FIP, update DNS, etc), but also to deploy concurrent HostedClusters.

This PR leverages as much as possible the existing functions used by AWS infra to find DNS domains or create a DNS record.

### DIFF
--- a/test/e2e/util/util.go
+++ b/test/e2e/util/util.go
@@ -9,6 +9,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/aws/aws-sdk-go/service/route53"
 	"github.com/go-logr/logr"
 	"github.com/go-logr/zapr"
 	"github.com/google/go-cmp/cmp"
@@ -17,6 +18,9 @@ import (
 	routev1 "github.com/openshift/api/route/v1"
 	routev1client "github.com/openshift/client-go/route/clientset/versioned"
 	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
+	awsinfra "github.com/openshift/hypershift/cmd/infra/aws"
+	awsutil "github.com/openshift/hypershift/cmd/infra/aws/util"
+	awsprivatelink "github.com/openshift/hypershift/control-plane-operator/controllers/awsprivatelink"
 	hccokasvap "github.com/openshift/hypershift/control-plane-operator/hostedclusterconfigoperator/controllers/resources/kas"
 	hcmetrics "github.com/openshift/hypershift/hypershift-operator/controllers/hostedcluster/metrics"
 	"github.com/openshift/hypershift/hypershift-operator/controllers/manifests"
@@ -33,6 +37,7 @@ import (
 	authenticationv1 "k8s.io/api/authentication/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	kapierror "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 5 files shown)


test/e2e/util/util.go
```diff
@@ -1378,9 +1382,101 @@ func ValidateMetrics(t *testing.T, ctx context.Context, hc *hyperv1.HostedCluste
...
+		}
+		return true, nil
+	})
+	g.Expect(err).ToNot(HaveOccurred(), "failed to get router-default service")
+
+	routerDefaultIP := defaultIngressRouterService.Status.LoadBalancer.Ingress[0].IP
+	g.Expect(routerDefaultIP).ToNot(BeEmpty(), "router-default service does not have an IP")
+
+	// Just for CI testing
+	clusterOpts.AWSPlatform.CredentialSecretName = "/etc/hypershift-pool-aws-credentials/credentials"
```
EmilienM, line 1414: i'll need https://github.com/openshift/release/pull/55849 to test that

test/e2e/util/util.go
```diff
@@ -1378,9 +1381,93 @@ func ValidateMetrics(t *testing.T, ctx context.Context, hc *hyperv1.HostedCluste
...
+	g.Expect(routerDefaultIP).ToNot(BeEmpty(), "router-default service does not have an IP")
+
+	// Just for CI testing
+	clusterOpts.AWSPlatform.Credentials.AWSCredentialsFile = "/etc/hypershift-pool-aws-credentials/credentials"
+	region := "us-east-1"
+	awsSession, err := clusterOpts.AWSPlatform.Credentials.GetSession("e2e-openstack-dns-record-on-aws", nil, region)
+	g.Expect(err).ToNot(HaveOccurred(), "failed to create AWS session")
+
+	route53Client := route53.New(awsSession, awsutil.NewAWSRoute53Config())
+	g.Expect(route53Client).ToNot(BeNil(), "failed to create Route53 client")
```
davidvossel, line 1419: what destroys this route53 when the cluster goes away later on?

This seems like it's pretty similar to some of the infra setup stages for aws clusters https://github.com/openshift/hypershift/blob/06ab3404fda04032e097bdc15830c67adb39761e/test/e2e/util/fixture.go#L105

EmilienM: nothing yet, this is prototyping, I'll write destroy once I get an actual DNS record to be created.

> This seems like it's pretty similar to some of the infra setup stages for aws clusters
Exactly :confused: and I'm even thinking about using that command even if my HostedCluster is on OpenStack platform.

davidvossel: > Exactly 😕 and I'm even thinking about re-using that command even if my HostedCluster is on OpenStack platform.

yeah, if nothing else, you could probably figure out a way to refactor that command in a way that reuses the route53 logic alone.

EmilienM: that's exactly my plan :+1:

test/e2e/util/util.go
```diff
@@ -1378,9 +1383,114 @@ func ValidateMetrics(t *testing.T, ctx context.Context, hc *hyperv1.HostedCluste
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "router-default",
+			Namespace: "openshift-ingress",
+		},
+	}
+
+	err := wait.PollUntilContextTimeout(ctx, 10*time.Second, 30*time.Minute, true, func(ctx context.Context) (done bool, err error) {
+		err = guestClient.Get(ctx, crclient.ObjectKeyFromObject(defaultIngressRouterService), defaultIngressRouterService)
+		if err != nil {
+			return false, nil
```
MaysaMacedo, line 1402: shouldn't an error be returned here?

EmilienM: no, it returns "done == false", but no error. In fact in this function we never return an error. It seems to be a pattern that has been used elsewhere around here so I'll keep it unless this is problematic.

EmilienM: I addressed this comment and now we return an error.

EmilienM:
/test e2e-openstack
sounds like one of the 2 nodepools didn't come online.

EmilienM:
the OpenStack job passed, let's see if it's consistent or if we were lucky.

/test e2e-openstack

EmilienM:
/test e2e-openstack
infra issue, not related to any of this.

EmilienM:
backport: https://github.com/openshift/hypershift/pull/4644

mdbooth:
Nice!

@EmilienM Did we end up implementing the cleanup @davidvossel pointed out?

EmilienM:
Yes cleanup is done, you should see it in the code and in logs

EmilienM:
https://github.com/openshift/hypershift/pull/4586/files#diff-cebde6b83b4b372162b2c6a22710a4ab0b1838576120535593a66a3a9bdfc170R1449-R1481

PR #4589, opened 2024-08-22
OSASINFRA-3573: openstack: when using floating IP, don't create a dedicated service for Ingress

EmilienM:

Simplify the workflow where if a floating IP (FIP) is given, we attach it to the default router in Ingress and stop creating a dedicated service for it.
Therefore, we can remove the IngressProvider from the HostedCluster.Spec.Platform.OpenStack, since we don't need it.
This will simplify UX and also stop wasting a FIP in the cloud.

This is achieved by waiting for the router-default to exist and then update the load balancer IP with the provided floating IP. This will reconcile the object and at some point the router will be exposed through that floating IP.


EmilienM:
I'll put this one on hold for now as I'm prioritizing the work on https://github.com/openshift/hypershift/pull/4586

EmilienM:
I'm putting the commit into https://github.com/openshift/hypershift/pull/4586

EmilienM:
I got it working locally but there are a few bugs right now. I'll come back to it later.

EmilienM:
backport: https://github.com/openshift/hypershift/pull/4644

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -965,19 +965,21 @@ func (r *reconciler) reconcileIngressController(ctx context.Context, hcp *hyperv
...
+			err := r.client.Get(ctx, client.ObjectKeyFromObject(ingressDefaultRouterIngressService), ingressDefaultRouterIngressService)
+			if apierrors.IsNotFound(err) {
+				errs = append(errs, fmt.Errorf("ingress router-default service not found: %w", err))
+			} else if err != nil {
+				errs = append(errs, fmt.Errorf("failed to retrieve ingress router-default service: %w", err))
+			} else if ingressDefaultRouterIngressService.Spec.LoadBalancerIP != ingressFloatingIP {
+				if _, err := r.CreateOrUpdate(ctx, r.client, ingressDefaultRouterIngressService, func() error {
+					log.Info("updating ingress router-default service with floating IP", "ingressFloatingIP", ingressFloatingIP)
+					ingressDefaultRouterIngressService.Spec.LoadBalancerIP = ingressFloatingIP
+					return nil
```
mdbooth, line 979: Does cloud provider openstack support doing this? This looks like it would leak a floating IP at best. I'd bet it both leaks the old floating IP and doesn't update the port attachment, tbh. I don't think we should do this in any case: these fields are owned by cloud provider openstack; we shouldn't touch them.

I think we need to just remove support for specifying an IngressIP until we can add support for it to the ingress controller operator.

EmilienM: I've tested it and it worked fine.
I'll test it again (no CI coverage for now), but I think we don't give the time for another FIP to be allocated to the default-router, since we reconcile here once the object is found. So the object is updated really fast and CCM doesn't have the time to create a new FIP.

Yes I agree we should patch the ingress controller though.

EmilienM: And to be clear, I don't want to remove support for specifying an IngressIP because that's the only way to pre-configure the DNS record for Ingress.
I'm going to test this workflow now and watch for floating IPs being created/deleted in hwatch, making sure nothing leaks and will report results here today.

mdbooth: > I've tested it and it worked fine. I'll test it again (no CI coverage for now), but I think we don't give the time for another FIP to be allocated to the default-router, since we reconcile here once the object is found. So the object is updated really fast and CCM doesn't have the time to create a new FIP.
> 
> Yes I agree we should patch the ingress controller though.

This sounds like a race at best. We're presumably relying on setting this value before it has been actioned by CPO. I remain extremely uncomfortable with this, and I'd prefer simply not to claim that we support the feature until we can do it robustly.

EmilienM: There is definitely a race of course.
I've given it a try again today and I confirm that no FIP is leaked and `spec.LoadBalancerIP` is immediately populated after the `Service` object exists and there is really no time for another FIP to be created. This is not pretty I agree but I don't think it's being problematic now that we are dev preview.

mdbooth: It's a race. We will forget about this, it will eventually break and we'll be unable to fix it because ingress operator has no support for it.

I would very much prefer that we remove the feature until we're able to implement it properly. This is not safe, even for dev preview. It's easy to add features. It's really hard to remove them when we get found out.

PR #4606, opened 2024-08-26
OCPBUGS-38921: Set Image as mutable and trigger upgrades on Azure NodePool spec changes

bryan-cox:
This PR:
* lets Image be mutable in Azure NodePools. If Image is not mutable, then there is no way of upgrading the image for an Azure NodePool.
* Uses machineTemplateSpec for Azure in the NodePool controller so changes to the spec trigger NodePool upgrades.

Fixes [OCPBUGS-38921](https://issues.redhat.com/browse/OCPBUGS-38921)

**Checklist**

### DIFF
--- a/hypershift-operator/controllers/nodepool/azure.go
+++ b/hypershift-operator/controllers/nodepool/azure.go
@@ -16,18 +16,7 @@ import (
 	capiazure "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
 )
 
-func azureMachineTemplateSpec(hcluster *hyperv1.HostedCluster, nodePool *hyperv1.NodePool, existing capiazure.AzureMachineTemplateSpec) (*capiazure.AzureMachineTemplateSpec, error) {
-	// The azure api requires passing a public key. This key is randomly generated, the private portion is thrown away and the public key
-	// gets written to the template.
-	sshKey := existing.Template.Spec.SSHPublicKey
-	if sshKey == "" {
-		var err error
-		sshKey, err = generateSSHPubkey()
-		if err != nil {
-			return nil, fmt.Errorf("failed to generate a SSH key: %w", err)
-		}
-	}
-
+func azureMachineTemplateSpec(nodePool *hyperv1.NodePool) (*capiazure.AzureMachineTemplateSpec, error) {
 	subnetName, err := azureutil.GetSubnetNameFromSubnetID(nodePool.Spec.Platform.Azure.SubnetID)
 	if err != nil {
 		return nil, fmt.Errorf("failed to determine subnet name for Azure machine: %w", err)
@@ -49,7 +38,6 @@ func azureMachineTemplateSpec(hcluster *hyperv1.HostedCluster, nodePool *hyperv1
 		NetworkInterfaces: []capiazure.NetworkInterface{{
 			SubnetName: subnetName,
 		}},
-		SSHPublicKey:  sshKey,
 		FailureDomain: failureDomain(nodePool),
 	}}}
 
--- a/hypershift-operator/controllers/nodepool/nodepool_controller.go
+++ b/hypershift-operator/controllers/nodepool/nodepool_controller.go
@@ -2956,14 +2956,28 @@ func machineTemplateBuilders(hcluster *hyperv1.HostedCluster, nodePool *hyperv1.
 			return nil
 		}
 	case hyperv1.AzurePlatform:
+		var err error
 		template = &capiazure.AzureMachineTemplate{}
+		machineTemplateSpec, err = azureMachineTemplateSpec(nodePool)
+		if err != nil {
+			return nil, nil, "", err
+		}
 		mutateTemplate = func(object client.Object) error {
 			o, _ := object.(*capiazure.AzureMachineTemplate)
-			spec, err := azureMachineTemplateSpec(hcluster, nodePool, o.Spec)
-			if err != nil {
-				return err
+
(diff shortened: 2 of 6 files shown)


hypershift-operator/controllers/nodepool/nodepool_controller.go
```diff
@@ -2956,14 +2956,15 @@ func machineTemplateBuilders(hcluster *hyperv1.HostedCluster, nodePool *hyperv1.
 			return nil
 		}
 	case hyperv1.AzurePlatform:
+		var err error
 		template = &capiazure.AzureMachineTemplate{}
 		mutateTemplate = func(object client.Object) error {
 			o, _ := object.(*capiazure.AzureMachineTemplate)
-			spec, err := azureMachineTemplateSpec(hcluster, nodePool, o.Spec)
+			machineTemplateSpec, err = azureMachineTemplateSpec(nodePool, o.Spec)
```
fxierh, line 2963: The `mutateTemplate` function is defined but not immediately called, which means that `machineTemplateSpec` is left empty (its original value) by the time the AzureMachineTemplate’s name is computed.

fxierh (commented):
Thanks for the quick fix !

hypershift-operator/controllers/nodepool/azure.go
```diff
@@ -16,18 +16,7 @@ import (
 	capiazure "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
 )
 
-func azureMachineTemplateSpec(hcluster *hyperv1.HostedCluster, nodePool *hyperv1.NodePool, existing capiazure.AzureMachineTemplateSpec) (*capiazure.AzureMachineTemplateSpec, error) {
```
enxebre, line 19: why is this PR dropping this?

bryan-cox: Just dropping the sshKey part out of this function.

bryan-cox: I thought maybe we could drop the sshKey piece; not sure if we need this now. When I tried that, there is an upstream error where this field is not omitted and is initialized as "" which is invalid.

https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/5092

hypershift-operator/controllers/nodepool/azure.go
```diff
@@ -49,7 +38,6 @@ func azureMachineTemplateSpec(hcluster *hyperv1.HostedCluster, nodePool *hyperv1
 		NetworkInterfaces: []capiazure.NetworkInterface{{
 			SubnetName: subnetName,
 		}},
-		SSHPublicKey:  sshKey,
```
fxierh, line 52: Looks like updates to SSHPublicKey no longer triggers a rolling upgrade

fxierh: Ok it seems we use a different mechanism (i.e. MachineConfig) to propagate `hc.spec.sshKey` to the nodes, so AzureMachineTemplate.Template.Spec.SSHPublicKey is actually not used. This looks fine then

hypershift-operator/controllers/nodepool/nodepool_controller.go
```diff
@@ -2956,14 +2956,27 @@ func machineTemplateBuilders(hcluster *hyperv1.HostedCluster, nodePool *hyperv1.
...
 		mutateTemplate = func(object client.Object) error {
 			o, _ := object.(*capiazure.AzureMachineTemplate)
-			spec, err := azureMachineTemplateSpec(hcluster, nodePool, o.Spec)
-			if err != nil {
-				return err
+			o.Spec = *machineTemplateSpec.(*capiazure.AzureMachineTemplateSpec)
+
+			// The azure api requires passing a public key. This key is randomly generated, the private portion is thrown away and the public key
+			// gets written to the template.
+			sshKey := o.Spec.Template.Spec.SSHPublicKey
```
muraee, line 2971: `o.Spec.Template.Spec.SSHPublicKey` is already overwritten at this point, so you wouldn't get the existing value. You might wanna do that before:
```go
o.Spec = *machineTemplateSpec.(*capiazure.AzureMachineTemplateSpec)
```

bryan-cox: I tried this and it fails to even set a value for sshKey. I reverted this PR back to what I previously had.

muraee: what is the error you get?

bryan-cox: It doesn't seem to set the key at all - https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_hypershift/4606/pull-ci-openshift-hypershift-main-e2e-aks/1828828725832060928

hypershift-operator/controllers/nodepool/nodepool_controller.go
```diff
@@ -2956,14 +2956,29 @@ func machineTemplateBuilders(hcluster *hyperv1.HostedCluster, nodePool *hyperv1.
...
+			if sshKey == "" {
+				sshKey, err = generateSSHPubkey()
+				if err != nil {
+					return fmt.Errorf("failed to generate a SSH key: %w", err)
+				}
+				o.Spec.Template.Spec.SSHPublicKey = sshKey
```
muraee, line 2976: this line is not needed.

PR #4607, opened 2024-08-26
OCPBUGS-38878: Revert "Block data plane HC configuration requests"

EmilienM:

- **Revert "OCPBUGS-34816: Block data plane HC configuration requests"**
- **Revert "OCPBUGS-34816: Configure user for HCCO"**

This PR broke the OpenStack platform where CNO needs to update the Infrastructure config.
Until we find a fix that work for all platforms, let's revert it and unblock the OpenStack folks.


sjenning:
lets do https://github.com/openshift/hypershift/pull/4609 instead
/close

PR #4608, opened 2024-08-26
OCPBUGS-38936: Fixed NodePool version validation

muraee:

Currently the NodePool version validation doesn't respect LatestSupportedVersion const, which means NodePools can be upgraded to a version higher than the HO's LatestSupportedVersion, where the HostedCluster will fail on the same version.

Fixes #

**Checklist**


devguyio:
how do you think about adding a test that validates this?

sjenning:
> what do you think about adding a test that validates this?
We would need to take `LatestSupportedVersion` as an arg here to avoid needing to change the test on each branch bump.

Would be nice to have a test here since the function lends itself to it and it has lots of cases.

enxebre:
>Currently the NodePool version validation doesn't respect LatestSupportedVersion const, which means NodePools can be upgraded to a version higher than the HO's LatestSupportedVersion, where the HostedCluster will fail on the same version.

should we instead pick the HC dictating max vesion from status.version?

muraee:
> should we instead pick the HC dictating max vesion from status.version?

@enxebre thought about that, but then this will prevent parallel upgrades. NodePools will have to wait until the HostedCluster has completely finished the upgrade before they can start upgrading.

muraee:
@devguyio @sjenning added a unit test

devguyio (commented):
LGTM
will leave it for @sjenning or @enxebre to lgtm

enxebre:
>@enxebre thought about that, but then this will prevent parallel upgrades. NodePools will have to wait until the HostedCluster has completely finished the upgrade before they can start upgrading.

Technically that make sense though, you wouldn't want the NodePool to go ahead the cp as that's not supported.

hypershift-operator/controllers/nodepool/nodepool_controller.go
```diff
@@ -2545,16 +2546,35 @@ func (r *NodePoolReconciler) getReleaseImage(ctx context.Context, hostedCluster
 
 	minSupportedVersion := supportedversion.GetMinSupportedVersion(hostedCluster)
 
-	releaseInfo, err := r.ReleaseProvider.Lookup(ctx, hostedCluster.Spec.Release.Image, pullSecretBytes)
-	if err != nil {
-		return nil, fmt.Errorf("failed to lookup release image: %w", err)
+	var hostedClusterVersion *semver.Version
+	if hostedCluster.Status.Version != nil && len(hostedCluster.Status.Version.History) > 0 {
```
enxebre, line 2550: Can this chunk of code be a unit tested getCurrentVersion function?

PR #4609, opened 2024-08-26
OCPBUGS-38878: allow changes to infrastructures global config for Openstack

sjenning:
The CNO's `infrastructureconfig` controller modifies the global config for the platforms included here:

https://github.com/openshift/cluster-network-operator/blob/c21e13f626cfcb21fdbd42076b57d00cebb064ca/pkg/controller/infrastructureconfig/predicates.go#L23-L38

Openstack is the first of these platforms to be supported by hypershift. We need to allow changes from the CNO.

This PR allows changes to the infrastructure global config from any user, which could be further restricted to the service account of the CNO.  However, we just want to get Openstack back online asap for now.

### DIFF
--- a/control-plane-operator/hostedclusterconfigoperator/controllers/resources/kas/admissionpolicies.go
+++ b/control-plane-operator/hostedclusterconfigoperator/controllers/resources/kas/admissionpolicies.go
@@ -57,6 +57,16 @@ func ReconcileKASValidatingAdmissionPolicies(ctx context.Context, hcp *hyperv1.H
 	return nil
 }
 
+// adapted from from https://github.com/openshift/cluster-network-operator/blob/c21e13f626cfcb21fdbd42076b57d00cebb064ca/pkg/controller/infrastructureconfig/predicates.go#L23
+func isOnPremPlatform(platformType hyperv1.PlatformType) bool {
+	switch platformType {
+	case hyperv1.OpenStackPlatform:
+		return true
+	default:
+		return false
+	}
+}
+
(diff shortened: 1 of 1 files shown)


control-plane-operator/hostedclusterconfigoperator/controllers/resources/kas/admissionpolicies.go
```diff
@@ -79,6 +88,9 @@ func reconcileConfigValidatingAdmissionPolicy(ctx context.Context, hcp *hyperv1.
 	if hcp.Spec.OLMCatalogPlacement == hyperv1.ManagementOLMCatalogPlacement {
 		configResources = append(configResources, "operatorhubs")
 	}
+	if !isOnPremPlatform(hcp.Spec.Platform.Type) {
```
enxebre, line 91: could we instead allow the cno client https://github.com/openshift/hypershift/pull/4303/files#diff-b6b6ca31afcd6a987e7138f0d8d83874e551721abca008aad75f0d30cbb3eed2R36-R37?

sjenning: > which could be further restricted to the service account of the CNO. However, we just want to get Openstack back online asap for now.

Yeah, we could.  Just wanted to get it working again vs complicating the code too much.

EmilienM:
/test e2e-openstack
The job will probably fail for other reasons fixed in https://github.com/openshift/hypershift/pull/4586 but at least we should see if the CNO reports an error or not.

Thanks @sjenning for the responsive fix!

EmilienM:
I confirm that the cluster network operator is healthy with that patch.

sjenning:
/hold
exploring an alternative solution

PR #2647, opened 2023-06-05
OCPBUGS-11882: Annotate HCP pods with the safe-to-evict-local-volume CA annotation

jparrill:
**Which issue(s) this PR fixes**:
Fixes #[OCPBUGS-11882](https://issues.redhat.com/browse/OCPBUGS-11882)

Depend on:
- https://github.com/openshift/aws-ebs-csi-driver-operator/pull/231
- https://github.com/openshift/cluster-network-operator/pull/1822
- https://github.com/openshift/aws-ebs-csi-driver-operator/pull/232
- https://github.com/openshift/cluster-network-operator/pull/1830

This is a second approach from https://github.com/openshift/hypershift/pull/2597. This makes more sense.

**Checklist**
- [x] Subject and description added to both, commit and PR.
- [x] Relevant issues have been referenced.
- [x] This change includes e2e and unit tests.

PR #4612, opened 2024-08-26
HOSTEDCP-1883: Remove hardcoded catalog images in CPO

Patryk-Stefanski:
**What this PR does / why we need it**: removes hardcoded catalog images in CPO which have to be updated every release, we now retrieve the openshift version from the ocp payload

Fixes # [HOSTEDCP-1883](https://issues.redhat.com/browse/HOSTEDCP-1883)

**Checklist**

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"sort"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
@@ -122,6 +123,11 @@ const (
 
 	hcpReadyRequeueInterval    = 1 * time.Minute
 	hcpNotReadyRequeueInterval = 15 * time.Second
+)
+
+var (
+	olmCatalogImagesOnce sync.Once
+	catalogImages        map[string]string
 )
 
 type HostedControlPlaneReconciler struct {
@@ -3831,7 +3837,7 @@ func (r *HostedControlPlaneReconciler) reconcileOperatorLifecycleManager(ctx con
 
 			var getCatalogImagesErr error
 			olmCatalogImagesOnce.Do(func() {
-				catalogImages, err = olm.GetCatalogImages(ctx, *hcp, pullSecret.Data[corev1.DockerConfigJsonKey])
+				catalogImages, err = olm.GetCatalogImages(ctx, *hcp, pullSecret.Data[corev1.DockerConfigJsonKey], registryclient.GetListDigest)
 				if err != nil {
 					getCatalogImagesErr = err
 					return
@@ -3908,7 +3914,7 @@ func (r *HostedControlPlaneReconciler) reconcileOperatorLifecycleManager(ctx con
 
 			olmManagerImage := ""
 			exists := false
-			if olmManagerImage, exists = releaseImageProvider.ComponentImages()["operator-lifecycle-manager"]; !exists {
+			if olmManagerImage, exists = releaseImageProvider.ImageExist("operator-lifecycle-manager"); !exists {
 				return fmt.Errorf("failed to get olm image from release image provider")
 			}
 
--- a/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
@@ -3906,10 +3906,16 @@ func (r *HostedControlPlaneReconciler) reconcileOperatorLifecycleManager(ctx con
 				}
 			}
 
+			olmManagerImage := ""
+			exists := false
+			if olmManagerImage, exists = releaseImageProvider.ComponentImages()["operator-lifecycle-manager"]; !exists {
+				return fmt.Errorf("failed to get olm image from release image provider")
+			}
+
 			olmDeployments := olm.OLMDeployments(p, hcp.Namespace)
 			for _, dep := range olmDeployments {
 				if _, err := createOrUpdate(ctx, r, dep.Manifest, func() error {
-					return dep.Reconciler(dep.Manifest, p.OwnerRef, p.DeploymentConfig, dep.Image)
+					return dep.Reconciler(dep.Manifest, p.OwnerRef, p.DeploymentConfig, dep.Image, olmManagerImage)
 				}); err != nil {
 					return fmt.Errorf("failed to reconcile %s deployment with image %s: %w", dep.Name, dep.Image, err)
 				}
--- a/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
@@ -3824,11 +3824,28 @@ func (r *HostedControlPlaneReconciler) reconcileOperatorLifecycleManager(ctx con
 
 			isImageRegistryOverrides := util.ConvertImageRegistryOverrideStringToMap(p.OLMCatalogsISRegistryOverridesAnnotation)
 
+			pullSecret := common.PullSecret(hcp.Namespace)
+			if err := r.Get(ctx, client.ObjectKeyFromObject(pullSecret), pullSecret); err != nil {
+				return fmt.Errorf("failed to get pull secret for namespace %s: %w", hcp.Namespace, err)
+			}
+
+			var getCatalogImagesErr error
+			olmCatalogImagesOnce.Do(func() {
+				catalogImages, err = olm.GetCatalogImages(ctx, *hcp, pullSecret.Data[corev1.DockerConfigJsonKey])
+				if err != nil {
+					getCatalogImagesErr = err
+					return
+				}
+			})
+			if getCatalogImagesErr != nil {
+				return fmt.Errorf("failed to get catalog images: %w", getCatalogImagesErr)
+			}
+
 			if r.ManagementClusterCapabilities.Has(capabilities.CapabilityImageStream) {
 				catalogsImageStream := manifests.CatalogsImageStream(hcp.Namespace)
 				if !overrideImages {
 					if _, err := createOrUpdate(ctx, r, catalogsImageStream, func() error {
-						return olm.ReconcileCatalogsImageStream(catalogsImageStream, p.OwnerRef, isImageRegistryOverrides)
+						return olm.ReconcileCatalogsImageStream(catalogsImageStream, p.OwnerRef, isImageRegistryOverrides, catalogImages)
 					}); err != nil {
 						return fmt.Errorf("failed to reconcile catalogs image stream: %w", err)
 					}
@@ -3839,12 +3856,7 @@ func (r *HostedControlPlaneReconciler) reconcileOperatorLifecycleManager(ctx con
 				}
 			} else {
 				if !overrideImages {
-					pullSecret := common.PullSecret(hcp.Namespace)
-					if err := r.Get(ctx, client.ObjectKeyFromObject(pullSecret), pullSecret); err != nil {
-						return fmt.Errorf("failed to get pull secret for namespace %s: %w", hcp.Namespace, err)
-					}
-
-					for name, catalog := range olm.CatalogToImage {
+					for name, catalog := range catalogImages {
 						imageRef, err := reference.Parse(catalog)
 						if err != nil {
 							return fmt.Errorf("failed to parse catalog image %s: %w", catalog, err)
--- a/control-plane-operator/controllers/hostedcontrolplane/olm/catalogs.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/olm/catalogs.go
@@ -78,23 +78,23 @@ var (
 	redHatOperatorsCatalogDeployment   = assets.MustDeployment(content.ReadFile, "assets/catalog-redhat-operators.deployment.yaml")
 )
 
-func ReconcileCertifiedOperatorsDeployment(deployment *appsv1.Deployment, ownerRef config.OwnerRef, dc config.DeploymentConfig, imageOverride string) error {
-	return reconcileCatalogDeployment(deployment, ownerRef, dc, certifiedCatalogDeployment, imageOverride)
+func ReconcileCertifiedOperatorsDeployment(deployment *appsv1.Deployment, ownerRef config.OwnerRef, dc config.DeploymentConfig, imageOverride, olmManagerImage string) error {
+	return reconcileCatalogDeployment(deployment, ownerRef, dc, certifiedCatalogDeployment, imageOverride, olmManagerImage)
 }
 
-func ReconcileCommunityOperatorsDeployment(deployment *appsv1.Deployment, ownerRef config.OwnerRef, dc config.DeploymentConfig, imageOverride string) error {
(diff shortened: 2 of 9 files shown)


control-plane-operator/controllers/hostedcontrolplane/olm/catalogs.go
```diff
@@ -132,21 +133,24 @@ func GetCatalogImages(ctx context.Context, hcp hyperv1.HostedControlPlane, pullS
...
+		"community-operators": fmt.Sprintf("registry.redhat.io/redhat/community-operator-index:v%s.%s", versionParts[0], versionParts[1]),
+		"redhat-marketplace":  fmt.Sprintf("registry.redhat.io/redhat/redhat-marketplace-index:v%s.%s", versionParts[0], versionParts[1]),
+		"redhat-operators":    fmt.Sprintf("registry.redhat.io/redhat/redhat-operator-index:v%s.%s", versionParts[0], versionParts[1]),
 	}
 
-	urls := make(map[string]string)
-	for _, operator := range operators {
-		urls[operator] = fmt.Sprintf("registry.redhat.io/redhat/%s-index:v%s", operator, majorMinor)
+	_, err = registryclient.GetListDigest(ctx, operators["certified-operators"], pullSecret)
+	if err != nil {
```
bryan-cox, line 144: Should we be more exact here on the error message? 

Only get the minor version -1 when the error is "no tag or digest specified". If the error was "failed to get repo setup" decrementing the minor version isn't going to fix that issue and the error should be reported back to the user for example.

sjenning:
```
    eventually.go:220: observed *v1beta1.HostedCluster e2e-clusters-4wstz/example-zxsnt invalid at RV 69846 after 10m0s: incorrect condition: wanted Degraded=False, got Degraded=True: UnavailableReplicas([certified-operators-catalog deployment has 2 unavailable replicas, redhat-marketplace-catalog deployment has 2 unavailable replicas, redhat-operators-catalog deployment has 2 unavailable replicas])
    --- FAIL: TestCreateCluster/ValidateHostedCluster (1609.14s)
```

catalog containers are failing with
```
level=fatal msg="cache directory has unexpected contents"
```

jparrill:
If in the end this PR want to be backported to old releases, we will need to add this other change in the backport: https://github.com/openshift/hypershift/pull/4707

test/e2e/util/util.go
```diff
@@ -1130,6 +1130,10 @@ func EnsurePodsWithEmptyDirPVsHaveSafeToEvictAnnotations(t *testing.T, ctx conte
 			"virt-launcher":                    "kubevirt.io",
 			"azure-disk-csi-driver-controller": "app",
 			"azure-file-csi-driver-controller": "app",
+			"certified-operators-catalog":      "app",
```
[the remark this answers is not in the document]
Patryk-Stefanski: I dont see why it wouldnt

control-plane-operator/controllers/hostedcontrolplane/olm/catalogs.go
```diff
@@ -118,19 +128,48 @@ func findTagReference(tags []imagev1.TagReference, name string) *imagev1.TagRefe
...
+
+	operators := map[string]string{
+		"certified-operators": fmt.Sprintf("registry.redhat.io/redhat/certified-operator-index:v%s.%s", versionParts[0], versionParts[1]),
+		"community-operators": fmt.Sprintf("registry.redhat.io/redhat/community-operator-index:v%s.%s", versionParts[0], versionParts[1]),
+		"redhat-marketplace":  fmt.Sprintf("registry.redhat.io/redhat/redhat-marketplace-index:v%s.%s", versionParts[0], versionParts[1]),
+		"redhat-operators":    fmt.Sprintf("registry.redhat.io/redhat/redhat-operator-index:v%s.%s", versionParts[0], versionParts[1]),
+	}
+
+	_, err = registryclient.GetListDigest(ctx, operators["certified-operators"], pullSecret)
+	if err != nil && ctx.Value(TestKey) != "test-reconcile-olm" {
```
sjenning, line 151: It is weird to have test-aware code in the function.

Also this whole block seems fragile.  Maybe we could break it out and make it unit testable by itself so the tests document the expected cases.

sjenning (changes requested):

test/e2e/util/util.go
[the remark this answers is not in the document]
sjenning: Added in https://github.com/openshift/hypershift/pull/2647 which was included in 4.14 :+1:

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -3826,10 +3844,16 @@ func (r *HostedControlPlaneReconciler) reconcileOperatorLifecycleManager(ctx con
 				}
 			}
 
+			olmManagerImage := ""
+			exists := false
+			if olmManagerImage, exists = releaseImageProvider.ComponentImages()["operator-lifecycle-manager"]; !exists {
```
muraee, line 3849:
```suggestion
			if olmManagerImage, exists = releaseImageProvider.ImageExist("operator-lifecycle-manager"); !exists {
```

control-plane-operator/controllers/hostedcontrolplane/olm/catalogs.go
```diff
@@ -118,19 +128,53 @@ func findTagReference(tags []imagev1.TagReference, name string) *imagev1.TagRefe
...
+		"community-operators": fmt.Sprintf("registry.redhat.io/redhat/community-operator-index:v%s.%s", major, minor),
+		"redhat-marketplace":  fmt.Sprintf("registry.redhat.io/redhat/redhat-marketplace-index:v%s.%s", major, minor),
+		"redhat-operators":    fmt.Sprintf("registry.redhat.io/redhat/redhat-operator-index:v%s.%s", major, minor),
+	}
+
+	for i := 0; i < 4; i++ {
+		_, err = registryclient.GetListDigest(ctx, operators["certified-operators"], pullSecret)
+		if err == nil || ctx.Value(TestKey) == "test-reconcile-olm" {
+			break
+		}
```
muraee, line 160: `ctx.Value(TestKey) == "test-reconcile-olm"` this is an unusual way to unit test, why do we need this?

also, please let's move it out of the loop at least for now.

control-plane-operator/controllers/hostedcontrolplane/olm/catalogs.go
```diff
@@ -118,19 +128,53 @@ func findTagReference(tags []imagev1.TagReference, name string) *imagev1.TagRefe
...
+	}
+
+	operators := map[string]string{
+		"certified-operators": fmt.Sprintf("registry.redhat.io/redhat/certified-operator-index:v%s.%s", major, minor),
+		"community-operators": fmt.Sprintf("registry.redhat.io/redhat/community-operator-index:v%s.%s", major, minor),
+		"redhat-marketplace":  fmt.Sprintf("registry.redhat.io/redhat/redhat-marketplace-index:v%s.%s", major, minor),
+		"redhat-operators":    fmt.Sprintf("registry.redhat.io/redhat/redhat-operator-index:v%s.%s", major, minor),
+	}
+
+	for i := 0; i < 4; i++ {
```
sjenning, line 156:
```suggestion
	for i := 0; i < len(operators); i++ {
```

control-plane-operator/controllers/hostedcontrolplane/olm/catalogs.go
sjenning: yes, this is still unit test aware.  we need to refactor so this is not required.

control-plane-operator/controllers/hostedcontrolplane/olm/catalogs.go
```diff
@@ -118,19 +128,53 @@ func findTagReference(tags []imagev1.TagReference, name string) *imagev1.TagRefe
...
-	"redhat-marketplace":  "registry.redhat.io/redhat/redhat-marketplace-index:v4.16",
-	"redhat-operators":    "registry.redhat.io/redhat/redhat-operator-index:v4.16",
+func GetCatalogImages(ctx context.Context, hcp hyperv1.HostedControlPlane, pullSecret []byte) (map[string]string, error) {
+	imageRef := hcp.Spec.ReleaseImage
+	imageConfig, _, _, err := registryclient.GetMetadata(ctx, imageRef, pullSecret)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get image metadata: %w", err)
+	}
+
+	version := strings.Split(imageConfig.Config.Labels["io.openshift.release"], ".")
```
sjenning, line 138: we should just parse this into a `semver.Version` and manipulate that
```suggestion
	version := semver.Parse(imageConfig.Config.Labels["io.openshift.release"])	
```

control-plane-operator/controllers/hostedcontrolplane/olm/catalogs.go
Patryk-Stefanski: this is set to 4 and not the length of operators because we want it to check the last 4 supported versions.

sjenning: ah ok, might be best to make this a variable/const with a name to the effect.  `4` is not self-documenting and there are no comments.

control-plane-operator/controllers/hostedcontrolplane/olm/catalogs.go
```diff
@@ -118,19 +122,47 @@ func findTagReference(tags []imagev1.TagReference, name string) *imagev1.TagRefe
...
+		"community-operators": fmt.Sprintf("registry.redhat.io/redhat/community-operator-index:v%s", version.String()),
+		"redhat-marketplace":  fmt.Sprintf("registry.redhat.io/redhat/redhat-marketplace-index:v%s", version.String()),
+		"redhat-operators":    fmt.Sprintf("registry.redhat.io/redhat/redhat-operator-index:v%s", version.String()),
+	}
+
+	for i := 0; i < 4; i++ {
+		_, err = digestLister(ctx, operators["certified-operators"], pullSecret)
+		if err != nil {
+			if !strings.Contains(err.Error(), "manifest unknown") {
+				return operators, err
```
muraee, line 148: why are we returning a valid value with the error
```suggestion
				return nil, err
```

Patryk-Stefanski: yeah we dont need to now, left over from the previous approach.

control-plane-operator/controllers/hostedcontrolplane/olm/catalogs.go
```diff
@@ -118,19 +122,47 @@ func findTagReference(tags []imagev1.TagReference, name string) *imagev1.TagRefe
...
+	operators := map[string]string{
+		"certified-operators": fmt.Sprintf("registry.redhat.io/redhat/certified-operator-index:v%s", version.String()),
+		"community-operators": fmt.Sprintf("registry.redhat.io/redhat/community-operator-index:v%s", version.String()),
+		"redhat-marketplace":  fmt.Sprintf("registry.redhat.io/redhat/redhat-marketplace-index:v%s", version.String()),
+		"redhat-operators":    fmt.Sprintf("registry.redhat.io/redhat/redhat-operator-index:v%s", version.String()),
+	}
+
+	for i := 0; i < 4; i++ {
+		_, err = digestLister(ctx, operators["certified-operators"], pullSecret)
+		if err != nil {
```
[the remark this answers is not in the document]
Patryk-Stefanski: were lowering the version by 1 until we find one that's available from the last 4 supported versions this is so that right after branching this doesn't fail on us when the latest catalog images arent available yet. last 4 version might be overkill.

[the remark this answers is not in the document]
muraee: right, but why we do that when `digestLister()` returns an error?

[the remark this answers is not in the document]
muraee: I guess the way this for loop is written is not clear, how about this

```go
func GetCatalogImages(ctx context.Context, hcp hyperv1.HostedControlPlane, pullSecret []byte, digestLister registryclient.DigestListerFN) (map[string]string, error) {
	imageRef := hcp.Spec.ReleaseImage
	imageConfig, _, _, err := registryclient.GetMetadata(ctx, imageRef, pullSecret)
	if err != nil {
		return nil, fmt.Errorf("failed to get image metadata: %w", err)
	}

	version, err := semver.Parse(imageConfig.Config.Labels["io.openshift.release"])
	if err != nil {
		return nil, fmt.Errorf("invalid OpenShift release version format: %s", imageConfig.Config.Labels["io.openshift.release"])
	}

	imageFound := false
	for i := 0; i < 4; i++ {
		imageRef := fmt.Sprintf("registry.redhat.io/redhat/certified-operator-index:v%s", version.String())
		_, err = digestLister(ctx, imageRef, pullSecret)
		if err == nil {
			// found an available image.
			imageFound = true
			break
		}
		if !strings.Contains(err.Error(), "manifest unknown") {
			return nil, err
		}

		// keep lowering the minor version until we find an available image.
		version.Minor--
	}

	if !imageFound {
		return nil, fmt.Errorf("couldn't find an image")
	}

	operators := map[string]string{
		"certified-operators": fmt.Sprintf("registry.redhat.io/redhat/certified-operator-index:v%s", version.String()),
		"community-operators": fmt.Sprintf("registry.redhat.io/redhat/community-operator-index:v%s", version.String()),
		"redhat-marketplace":  fmt.Sprintf("registry.redhat.io/redhat/redhat-marketplace-index:v%s", version.String()),
		"redhat-operators":    fmt.Sprintf("registry.redhat.io/redhat/redhat-operator-index:v%s", version.String()),
	}
	return operators, nil
}
```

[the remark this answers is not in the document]
Patryk-Stefanski: manifestUnknow error means that the manifest doesnt exist anything else points to other errors that are unexpected.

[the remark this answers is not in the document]
muraee: you need to break on `err == nil`, otherwise you will always run the loop 4 times, even if you found a valid image immediately. Making the break/exist condition of a loop explicit is always better.


Also we need to return a readable error if we don't find an image withing the supported versions, rather thann returning whatever generic error `digestLister()` returns.

control-plane-operator/controllers/hostedcontrolplane/olm/catalogs.go
```diff
@@ -118,19 +122,53 @@ func findTagReference(tags []imagev1.TagReference, name string) *imagev1.TagRefe
...
+			if !strings.Contains(err.Error(), "manifest unknown") {
+				return nil, err
+			}
+			if i == supportedVersions-1 {
+				return nil, fmt.Errorf("failed to get image digest for 4 previous versions of certified-operator-index: %w", err)
+			}
+			version.Minor--
+		} else {
+			break
+		}
```
muraee, line 153:
```suggestion
		if err == nil {
		        break
		}
		//manifest unknown error is expected if the image is not available.
		//If the all supported versions are checked and the image is still not available, return the error
		if !strings.Contains(err.Error(), "manifest unknown") {
			return nil, err
		}
		if i == supportedVersions-1 {
			return nil, fmt.Errorf("failed to get image digest for 4 previous versions of certified-operator-index: %w", err)
		}
		version.Minor--
```

Patryk-Stefanski: done

PR #4621, opened 2024-08-27
OCPBUGS-38474: copy image-registry AdditionalTrustedCA configmap into HC openshift-config

sjenning:


devguyio:
Since this wasn't caught by a test, how making sure that there's a test that covers this?

PR #4623, opened 2024-08-27
OCPBUGS-39035: Ensure AzureCluster.Status.FailureDomains is populated when any NodePool specifies an Availability Zone

fxierh:
Ensure AzureCluster.Status.FailureDomains is populated when any NodePool specifies an Availability Zone.

Otherwise CAPZ will create Azure VMs with both Availability Zone and Availability Set configured, which results in an error:
```bash
{
        "error": {
          "code": "BadRequest",
          "message": "Virtual Machine cannot be created because both Availability Zone and Availability Set were specified. Deploying an Availability Set to an Availability Zone isn’t supported."
        }
}
```

ref: 
https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/711d86187262612c8f7f02f1848c22fbf3b040cd/azure/scope/cluster.go#L778-L780

### DIFF
--- a/hypershift-operator/controllers/hostedcluster/internal/platform/azure/azure.go
+++ b/hypershift-operator/controllers/hostedcluster/internal/platform/azure/azure.go
@@ -62,7 +62,7 @@ func (a Azure) ReconcileCAPIInfraCR(
 	}
 
 	if _, err := createOrUpdate(ctx, client, azureCluster, func() error {
-		return reconcileAzureCluster(azureCluster, hcluster, apiEndpoint, azureClusterIdentity, controlPlaneNamespace)
+		return reconcileAzureCluster(ctx, client, azureCluster, hcluster, apiEndpoint, azureClusterIdentity)
 	}); err != nil {
 		return nil, fmt.Errorf("failed to reconcile Azure CAPI cluster: %w", err)
 	}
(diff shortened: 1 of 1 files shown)


hypershift-operator/controllers/hostedcluster/internal/platform/azure/azure.go
```diff
@@ -242,6 +242,22 @@ func reconcileAzureCluster(azureCluster *capiazure.AzureCluster, hcluster *hyper
...
+	// (which results in an error), ensure AzureCluster.Status.FailureDomains is populated when any NodePool
+	// specifies an Availability Zone.
+	npList, err := listNodePools(ctx, c, hcluster.Namespace, hcluster.Name)
+	if err != nil {
+		return err
+	}
+	for _, np := range npList {
+		if len(np.Spec.Platform.Azure.AvailabilityZone) > 0 {
+			azureCluster.Status.FailureDomains = capiv1.FailureDomains{
+				"1": capiv1.FailureDomainSpec{},
```
Patryk-Stefanski, line 255: would we not want a failureDomain per AZ ?

fxierh:
Fixed upstream: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/5098

/close

PR #4632, opened 2024-08-28
OCPBUGS-39147: regroup KAS certs into public and private certs

sjenning:
This mostly reverts https://github.com/openshift/hypershift/pull/4566 and regroups private mgmt cluster service network dns names for the KAS into their own cert stored in a new `kube-server-private` secret.


EmilienM:
Manually tested and it worked for me on OpenStack :+1: 
Our CI doesn't test IP endpoints, so it's not worth calling the job here.

PR #4635, opened 2024-08-29
HOSTEDCP-1933: Expose AWS Tenancy through NodePool API

muraee:
This allows launching AWS EC2 instances into dedicated instances or dedicated hosts.

Fixes #

**Checklist**

### DIFF
--- a/api/hypershift/v1alpha1/nodepool_types.go
+++ b/api/hypershift/v1alpha1/nodepool_types.go
@@ -823,6 +823,17 @@ type AWSNodePoolPlatform struct {
 	// +listMapKey=key
 	// +optional
 	ResourceTags []AWSResourceTag `json:"resourceTags,omitempty"`
+
+	// Tenancy indicates if instance should run on shared or single-tenant hardware.
+	//
+	// Possible values:
+	// default: NodePool instances run on shared hardware.
+	// dedicated: Each NodePool instance runs on single-tenant hardware.
+	// host: NodePool instances run on user's pre-allocated dedicated hosts.
+	//
+	// +optional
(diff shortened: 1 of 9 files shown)


api/hypershift/v1alpha1/nodepool_types.go
```diff
@@ -823,6 +823,17 @@ type AWSNodePoolPlatform struct {
...
+	// Tenancy indicates if instance should run on shared or single-tenant hardware.
+	//
+	// Possible values:
+	// default: NodePool instances run on shared hardware.
+	// dedicated: Each NodePool instance runs on single-tenant hardware.
+	// host: NodePool instances run on user's pre-allocated dedicated hosts.
+	//
+	// +optional
+	// +kubebuilder:validation:Enum:=default;dedicated;host
+	Tenancy string `json:"tenancy,omitempty"`
```
EmilienM, line 836: You could create a new type (like AWSTenancy).
An example here:

https://github.com/openshift/hypershift/blob/d8812cdc1ea756b0857c13e0436ba8f4137a172f/api/hypershift/v1beta1/hostedcluster_types.go#L1952-L1974

And in the code you'll use the constants directly. Very handy and cleaner, imo.

muraee: thanks for the suggestion @EmilienM.
This is field is just going to be passed through to CAPA as is, so we don't really need to use it's values directly.

EmilienM: maybe something which can be done on CAPA side as well (just an idea).

PR #1787, opened 2022-10-05
Default Ingress Passthrough for KubeVirt provider

davidvossel:
# Overview

This PR adds the new `hostedCluster.Spec.Platform.KubeVirt.BaseDomainPassthrough`  option, which automatically sets the HostedCluster's BaseDomain to a unique subdomain of the mgmt/infra cluster's `*apps` dns. The HCCO operator in turn sets up the correct services and routes to pass through traffic on that subdomain to the KubeVirt VMs.

We've already been using this technique for months now both in our manual development of the KubeVirt Platform and CI [[1]](https://github.com/openshift/release/blob/1b66770683fa462c3cdc0e1dacbe6796215edf9b/ci-operator/step-registry/hypershift/kubevirt/create/hypershift-kubevirt-create-chain.yaml#L58). The technique is also documented in our hypershift upstream docs here [[2]](https://hypershift-docs.netlify.app/how-to/kubevirt/create-kubevirt-cluster/ ). We're even seeing third parties document manually setting up this same technique[ [3]](https://almogelfassy.medium.com/cluster-as-a-service-the-next-phase-of-development-for-kubernetes-by-hypershift-4f1b8d28ad78 ).

What we know now is that this ingress passthrough technique is the primary way users evaluate the Hypershift KubeVirt platform. It's the easiest way to get a functioning cluster. The `hostedCluster.Spec.Platform.KubeVirt.BaseDomainPassthrough` option improves the user experience for people who would otherwise be manually deploying the same logic.

# Design

The figure below outlines how this technique works. **The orange boxes are the route and service that the HCCO manages in the infra/mgmt cluster which passes through traffic to the kubevirt guest cluster.**

kubevirt hypershift default route ingress notes

# Usage

With this feature, users of the `hypershift` cli simply omit the basedomain when creating their kubevirt platform cluster and that automatically renders a HostedCluster with the passthrough feature enabled. Below is an example of what the cli looks like that uses the automatic passthrough default ingress.

```
bin/hypershift create cluster kubevirt \
	--name $CLUSTER_NAME \
	--pull-secret $PULL_SECRET_PATH \
	--ssh-key $SSH_KEY_PATH  \
```

# References

1. https://github.com/openshift/release/blob/1b66770683fa462c3cdc0e1dacbe6796215edf9b/ci-operator/step-registry/hypershift/kubevirt/create/hypershift-kubevirt-create-chain.yaml#L58
[...]

PR #4636, opened 2024-08-29
 	OSASINFRA-3582: openstack: base domain passthrough for ingress

EmilienM:

Implements base domain passthrough (like Kubevirt does) for Ingress.
Basically taking the work from https://github.com/openshift/hypershift/pull/1787 and adapting for OpenStack.


control-plane-operator/hostedclusterconfigoperator/controllers/resources/ingress/reconcile.go
```diff
@@ -141,7 +148,11 @@ func ReconcileDefaultIngressPassthroughService(service *corev1.Service, defaultN
 	// secondary networks.
 	service.Spec.Selector = map[string]string{}
 
-	service.Spec.Type = corev1.ServiceTypeClusterIP
+	if hcp.Spec.Platform.Type == hyperv1.KubevirtPlatform {
+		service.Spec.Type = corev1.ServiceTypeClusterIP
+	} else {
+		service.Spec.Type = corev1.ServiceTypeLoadBalancer
```
EmilienM, line 154: I need to find a way to not having a floating IP created for that LB.
The alternative is to create a Pod (e.g. Nginx) and its `ServiceTypeClusterIP` like Kubevirt did, but for VMs.

mdbooth (commented):
There's a major missing piece here, which is how to route traffic from the internal ingress passthrough service to OpenStack workers.

KubeVirt does it here: https://github.com/openshift/hypershift/blob/0bd568b4ffde84b119e7cf5dc43020178218f165/control-plane-operator/hostedclusterconfigoperator/controllers/machine/setup.go

This is a reconciliation loop over CAPI Machines which creates service endpoints for the internal passthrough service for each internal address of each KubeVirt machine. I don't know (but perhaps @davidvossel can confirm), but I assume KubeVirt can do this because:

* They can assume these internal addresses are routable from the management cluster, presumably because the KubeVirt VMs are running in the management cluster
* Each machine exposes an ingress listener on the port required by the passthrough service
* There's no firewalling preventing traffic directly to these machines from the management cluster

We can't make any of these assumptions in OpenStack. I think this is the problem we need to solve.

The problem this PR is trying to solve is really just DNS: it allows to re-use the wildcard address created for the ingress on the management cluster for the workload cluster's ingress. The disadvantage is that when doing this, all ingress cluster for the workload cluster must pass through the management cluster in both directions. We should consider how much of a problem this could be in our target deployments.

Either way, we still have the problem of how to get traffic in to the workload cluster. The obvious way to do this would be via the FIP which was created by the workload cluster's ingress operator.

We might be able to do this by making the passthrough service a [headless service](https://kubernetes.io/docs/concepts/services-networking/service/#headless-services), and reconcile the endpoint for it from the public IP address of the workload cluster's ingress service.

EmilienM:
I'm not actively working on that and we have no plans to pursue this effort for now so I'll just close it and eventually re-open when requested/desired.

PR #4639, opened 2024-08-29
tolerations docs

davidvossel:
Documents the Hosted Cluster tolerations feature


bryan-cox:
Thanks @davidvossel 
/lgtm

PR #4646, opened 2024-08-30
HOSTEDCP-1767: feat: install prerequisite CRDs for AROHCP

devguyio:

> [!NOTE]
> This PR is stacked on top of #4643 . It adds a single commit on top it and shouldn't be merged until #4643 is merged.

When installing the HO operator on an AKS cluster, there are a few CRDs the following CRDs need to be installed first:

- podmonitors.monitoring.coreos.com
- prometheusrules.monitoring.coreos.com
- servicemonitors.monitoring.coreos.com
- routes.route.openshift.io

This PR adds the capability to the install command of installing those CRDs when the `--managed-service` is set to `ARO-HCP`.

Fixes [HOSTEDCP-1767](https://issues.redhat.com//browse/HOSTEDCP-1767)

**Checklist**

### DIFF
--- a/cmd/install/assets/assets.go
+++ b/cmd/install/assets/assets.go
@@ -13,6 +13,7 @@ import (
 	crclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+//go:embed prereqs-azure/*
 //go:embed hypershift-operator/*
 //go:embed cluster-api/*
 //go:embed cluster-api-provider-aws/*
--- a/Makefile
+++ b/Makefile
@@ -41,6 +41,13 @@ ifeq ("/","${HOME}")
 HOME=/tmp
 endif
 
+# Prometheus CRDs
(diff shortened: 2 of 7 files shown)


cmd/install/assets/assets.go
```diff
@@ -13,6 +13,7 @@ import (
 	crclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+//go:embed prereqs-azure/*
```
jparrill, line 16: Where those CRDs come from? I mean, for self-maintenance would we need to add a command to auto-generation/update of those CRDs in the Makefile, like we do with CAPI ones?

devguyio: That's a good idea! I wasn't aware we do that. I'll check it out.

bryan-cox:
We make need to hold/rethink this PR a bit in light of [RFE-6258](https://issues.redhat.com/browse/RFE-6258). I don't know that we will need the Prometheus CRs now. We will still need the Routes CR though.

devguyio:
No longer needed
/close

bryan-cox:
Ah I thought we still needed this

PR #4650, opened 2024-08-30
HOSTEDCP-1939: Trigger SharedIngress reconciliation immediately on start

muraee:
This is needed so that the sharedingress loadbalacner is created at the start without waiting for a HostedCluster to exist.

Fixes #

**Checklist**

### DIFF
--- a/hypershift-operator/controllers/sharedingress/sharedingress_controller.go
+++ b/hypershift-operator/controllers/sharedingress/sharedingress_controller.go
@@ -3,9 +3,10 @@ package sharedingress
 import (
 	"context"
 	"fmt"
+	"os"
+
 	"github.com/openshift/hypershift/cmd/log"
 	"k8s.io/apimachinery/pkg/api/errors"
-	"os"
 
 	routev1 "github.com/openshift/api/route/v1"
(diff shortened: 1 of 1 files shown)


hypershift-operator/controllers/sharedingress/sharedingress_controller.go
```diff
@@ -34,9 +38,20 @@ func (r *SharedIngressReconciler) SetupWithManager(mgr ctrl.Manager, createOrUpd
...
+	initialSync := make(chan event.GenericEvent)
+
 	builder := ctrl.NewControllerManagedBy(mgr).
 		For(&hyperv1.HostedCluster{}).
+		WatchesRawSource(source.Channel(initialSync, &handler.EnqueueRequestForObject{})).
 		Named("SharedIngressController")
+
+	go func() {
+		initialSync <- event.GenericEvent{Object: &hyperv1.HostedCluster{}}
+	}()
```
jparrill, line 53: wouldn't you need a defer?

muraee: no its not needed, this is an unbuffered channel, so the send will automatically for a receiver.

enxebre:
thanks!
/lgtm

PR #4660, opened 2024-09-03
OCPBUGS-39396: Fix multi-arch validation by prioritizing ReleaseImage check

LiangquanLi930:
Fix validation for multi-arch image by checking the absence of ReleaseImage before ReleaseStream
Fixes OCPBUGS-39396

**Checklist**


bryan-cox:
/retitle OCPBUGS-39396: Set Default ReleaseStream Only When ReleaseImage is Empty

PR #4668, opened 2024-09-04
HOSTEDCP-1942: Add functional tests for Etcd Recovery

jparrill:
Included 2 functional tests to the TestHAEtcdChaos:

- An etcd member with data corruption.
- A test case when the ETCD Cluster lose one member.

Fixes #[HOSTEDCP-1942](https://issues.redhat.com/browse/HOSTEDCP-1942)

### DIFF
--- a/test/e2e/chaos_test.go
+++ b/test/e2e/chaos_test.go
@@ -17,12 +17,14 @@ import (
 	cpomanifests "github.com/openshift/hypershift/control-plane-operator/controllers/hostedcontrolplane/manifests"
 	"github.com/openshift/hypershift/hypershift-operator/controllers/manifests"
 	appsv1 "k8s.io/api/apps/v1"
+	batchv1 "k8s.io/api/batch/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/utils/pointer"
 	"k8s.io/utils/ptr"
 
 	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
+	etcdrecoverymanifests "github.com/openshift/hypershift/hypershift-operator/controllers/manifests/etcdrecovery"
 	e2eutil "github.com/openshift/hypershift/test/e2e/util"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


test/e2e/chaos_test.go
```diff
@@ -240,6 +257,136 @@ func testKillAllMembers(parentCtx context.Context, client crclient.Client, clust
...
+
+		// Get ETCD member ID
+		innerCommand := fmt.Sprintf("member list | grep %s | awk '{print $1}' | tr -d ,", pod.Name)
+		memberDiscoveryCommand := append(baseCommand, innerCommand)
+
+		// Final etcd commands
+		command := append(baseCommand, "member", "remove", fmt.Sprintf("$(%s)", memberDiscoveryCommand))
+
+		t.Logf("Deleting ETCD Member: %s", pod.Name)
+		_, err = e2eutil.RunCommandInPod(ctx, client, pod.Spec.Containers[0].Name, pod.Namespace, command, pod.Spec.Containers[0].Name, 5*time.Minute)
```
muraee, line 358:
```suggestion
		_, err = e2eutil.RunCommandInPod(ctx, client, "etcd", pod.Namespace, command, pod.Spec.Containers[0].Name, 5*time.Minute)
```
this arg is for the component name, not container name.

jparrill: In this case is the same but as you mentioned, makes sense to hardcoded it.

test/e2e/chaos_test.go
```diff
@@ -240,6 +257,138 @@ func testKillAllMembers(parentCtx context.Context, client crclient.Client, clust
...
+			Namespace:     manifests.HostedControlPlaneNamespace(cluster.Namespace, cluster.Name),
+			LabelSelector: labels.Set(etcdSts.Spec.Selector.MatchLabels).AsSelector(),
+		})
+		g.Expect(err).NotTo(HaveOccurred(), "failed to get etcd pods")
+
+		pod := randomPods(etcdPods.Items, 1)[0]
+		command := fmt.Sprintf("find /var/lib/data/member/wal -type f -name \"*.wal\" -print0 | shuf -z -n1 | xargs -0 rm")
+
+		t.Logf("Deleting wal file from etcd pod: %s", pod.Name)
+		cmdStdout, err := e2eutil.RunCommandInPod(ctx, client, "etcd", pod.Namespace, []string{"/bin/sh", "-c", command}, pod.Spec.Containers[0].Name, 5*time.Minute)
```
muraee, line 286: this will run the command in a random pod, not the one you have above `pod := randomPods(etcdPods.Items, 1)[0]`
if you need to run in an exact pod, use PodExecuter directly see: https://github.com/openshift/hypershift/blob/main/test/e2e/util/util.go#L962-L992

```suggestion
		cmdStdout, err := e2eutil.RunCommandInPod(ctx, client, "etcd", pod.Namespace, []string{"/bin/sh", "-c", command}, "etcd", 5*time.Minute)
```
also please hardcode the container name as well, the reader don't know what the first container is/it might change.

sjenning (approved):
Note to self: this is one of the tests we'll have to immediately version gate when switching to `main` e2e for all branches

PR #4669, opened 2024-09-04
OSASINFRA-3584: openstack: remove support for floating IP

EmilienM:

This feature was added with the goal of being able to
have Ingress working on day 1 but the way it's implemented is racy
because it depends on a resource that Hypershift controller don't own,
and is owned by the Ingress controller.

Let's remove it now and re-add it later when it's safer.

### DIFF
--- a/docs/content/how-to/openstack/create-openstack-cluster.md
+++ b/docs/content/how-to/openstack/create-openstack-cluster.md
@@ -6,6 +6,8 @@ Install an OCP cluster running on VMs within a management OCP cluster
 
 * The HyperShift Operator with OpenStack support is currently in development and is not intended for production use.
 * OpenStack CSI (Cinder and Manila) are not functional.
+* Operators running in the workload cluster (e.g. console) won't be operational on day 1 and a manual and documented
+  action is required to make them work on day 2.
 
 ## Prerequisites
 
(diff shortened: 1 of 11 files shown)


mdbooth (commented):
/lgtm

mdbooth:
@EmilienM Some thoughts on how to bring this back.

Firstly, here's how AWS supports adding annotations to the router service in the ingress controller: https://github.com/openshift/cluster-ingress-operator/blob/8252ac492c04d161fbcf60ef82af2989c99f4a9d/pkg/operator/controller/ingress/load_balancer_service.go#L403-L447

We could easily add our own here.

Here's the list of supported annotations: https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md

Couple of things to note there. Firstly there doesn't seem to be a way to specify a floating IP, so we'd probably have to set loadbalancerIP directly as you're doing here. The difference is that it's not a race if the ingress operator does it on creation. Looks like the code lends itself to easily adding this ability.

Secondly, I note that there's a `loadbalancer.openstack.org/keep-floatingip` annotation whose default is false. I'll bet that if we just specify a loadbalancerIP without adding this, CCM will 'helpfully' delete our floating IP for us on deletion. If we created it ourselves, this is almost certainly not what we want.

docs/content/how-to/openstack/create-openstack-cluster.md
```diff
@@ -141,8 +128,7 @@ hcp create cluster openstack \
 --openstack-credentials-file $CLOUDS_YAML \
 --openstack-external-network-id $EXTERNAL_ID \
 --openstack-node-image-name $IMAGE_NAME \
---openstack-node-flavor $FLAVOR \
```
jparrill, line 144: is this needed to be removed?

jparrill (commented):
just a question, otherwise lgtm

docs/content/how-to/openstack/create-openstack-cluster.md
mdbooth: It hasn't been removed. Only `--openstack-ingress-floating-ip` has been removed.

PR #4704, opened 2024-09-11
HOSTEDCP-1569: e2e: add version gating for 4.17

sjenning:
This PR adds a simple mechanism for gating/skipping test cases based on the version of the latest release image

### DIFF
--- a/test/e2e/util/util.go
+++ b/test/e2e/util/util.go
@@ -1218,6 +1218,7 @@ func EnsureAdmissionPolicies(t *testing.T, ctx context.Context, mgmtClient crcli
 	}
 	guestClient := WaitForGuestClient(t, ctx, mgmtClient, hc)
 	t.Run("EnsureValidatingAdmissionPoliciesExists", func(t *testing.T) {
+		AtLeast(t, Version418)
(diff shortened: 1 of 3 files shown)


test/e2e/util/util.go
```diff
@@ -1218,6 +1218,9 @@ func EnsureAdmissionPolicies(t *testing.T, ctx context.Context, mgmtClient crcli
 	}
 	guestClient := WaitForGuestClient(t, ctx, mgmtClient, hc)
 	t.Run("EnsureValidatingAdmissionPoliciesExists", func(t *testing.T) {
+		if !AtLeast(Version418) {
```
muraee, line 1221: why not skip the whole test here
https://github.com/openshift/hypershift/blob/bb6974658cee08dac4f39e3c0d324135725dd54f/test/e2e/util/hypershift_framework.go#L153

sjenning: I thought about that but I think we want to use `t.Skip()` if the assertion isn't supported in the version under test.  That way there is a slot for it in the junit/testgrid explicitly showing the assertion is in the code, but it is being skipped.

PR #4705, opened 2024-09-11
OCPBUGS-38425: Update the ImageStream tag according to the OCP release

jparrill:
Now the OLM catalog version is purely based in the OCP Release and not hardcoded.

Fixes #[OCPBUGS-38425](https://issues.redhat.com/browse/OCPBUGS-38425)


jparrill:
/close
Looks like this fix will affect the CI execution because the newer and non-released catalogs are not GPG signed so the CI will fail. 

AFAIK this will be investigated by the team and fixed eventually.

PR #4730, opened 2024-09-16
NO-JIRA: Add ddcache entry into gitignore

jparrill:
This PR add `.ddcache` file to gitignore.

`.ddcache` file is created into the root of the repository by the SNYK CLI once it's executed in local.


jparrill:
@bryan-cox Thanks for the reminder. Done

PR #4731, opened 2024-09-16
HOSTEDCP-1966: wrap nodePool tenancy API field in Placement struct

muraee:
This allows us to group and easily add new placement options to the API in the future.

Fixes #

**Checklist**


api/hypershift/v1beta1/nodepool_types.go
```diff
@@ -833,6 +833,14 @@ type AWSNodePoolPlatform struct {
 	// +optional
 	ResourceTags []AWSResourceTag `json:"resourceTags,omitempty"`
 
+	// Placement specifies the placement options for the EC2 instances.
```
bryan-cox, line 836: Joel has mentioned to me in the past the godoc should be lowercase for the field here in the comment.

cc: @JoelSpeed in case you want to check out this new API field.

bryan-cox:
@muraee - are we expecting self-managed to be able to use this as well? I don't see a way to set this through HCP CLI.

api/hypershift/v1beta1/nodepool_types.go
devguyio: can you elaborate more? What exactly should be lowercase? The `Placement` word?

bryan-cox: Yeah `placement`

api/hypershift/v1beta1/nodepool_types.go
```diff
@@ -833,6 +833,14 @@ type AWSNodePoolPlatform struct {
...
+	//
+	// +optional
+	Placement *PlacementOptions `json:"placement,omitempty"`
+}
+
+// PlacementOptions specifies the placement options for the EC2 instances.
+type PlacementOptions struct {
```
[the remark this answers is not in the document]
muraee: I wanted to follow the AWS SDK naming convention which uses `Placement`. 
> in fact capa already have PlacementGroupPartition and PlacementGroupName

right, those fields are also part of the same placement input in the SDK, I don't know why they are not grouped in CAPA, but If we ever expose them to our API, we can add them to the same placement struct, which is why I think tenacnyOptions might not be a good naming in the future.

happy to change it if you disagree.

api/hypershift/v1beta1/nodepool_types.go
muraee: I'd rather we stay consistent across all fields, which are now all using upperCase. 
Maybe, we could have a ticket to change all to lowerCase.

api/hypershift/v1beta1/nodepool_types.go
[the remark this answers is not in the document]
enxebre: ok, that makes sense to me

api/hypershift/v1beta1/nodepool_types.go
JoelSpeed: The godoc should start with the `json` tag version, ie, the serialised version, since that's what the end user would normally interact with, and these docs are for end users to understand the field.

> I'd rather we stay consistent across all fields, which are now all using upperCase.

This perpetuates the incorrect version and increases what there is to fix later. If you start fixing now, you have fewer to fix later.

api/hypershift/v1beta1/nodepool_types.go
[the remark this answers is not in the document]
JoelSpeed: A "placement group" is a concept within AWS, so I wonder if the placement group related options are separate because otherwise they would have to be `placement.placementGroup{Name|Partition}`?

JoelSpeed (commented):
This PR currently appears to represent a breaking change to an API, I don't see the generated CRDs anywhere here but changing a field to move it into a struct will break any existing use of this field.

What are clusters using presently, v1alpha1 as storage or v1beta1? You can change the v1beta1 API without making an identical change to the v1alpha1 API if you have conversion webhooks, do you have conversion webhooks?

/hold until we've understood the impact this might have

api/hypershift/v1beta1/nodepool_types.go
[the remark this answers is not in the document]
muraee: image

from: https://github.com/aws/aws-sdk-go/blob/main/service/ec2/api.go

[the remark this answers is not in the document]
muraee: "placement group" is also part of the same input in the SDK convention like along `tenancy`, so it would be `placement.group{Name|Partition|Id}`

muraee:
@JoelSpeed 
the original PR which introduced the `tenancy` field was added very recently(it hit stage a couple of days ago), there are currently no consumers of this  field.

> What are clusters using presently, v1alpha1 as storage or v1beta1? You can change the v1beta1 API without making an identical change to the v1alpha1 API if you have conversion webhooks, do you have conversion webhooks?

v1beta1 is used. Yes, we shouldn't have to add it to v1alpha1, but we have a fuzzing test today that would fail otherwise. We are going to drop that test soon so we can stop doing this. fwiw we don't have any v1alpha1 API users.

JoelSpeed:
> v1beta1 is used. Yes, we shouldn't have to add it to v1alpha1, but we have a fuzzing test today that would fail otherwise. We are going to drop that test soon so we can stop doing this. fwiw we don't have any v1alpha1 API users.

In the future when you evolve the API, fuzz tests are good and I'd suggest not getting rid of them, but rather building conversion webhooks that understand how to store state that doesn't exist on a downgrade, and then restore the state on upgrade. This is a fairly common pattern but if the v1alpha1 is going away soon, then I guess it doesn't matter now.

Poke me next time you're planning an API version change 😉 

> the original PR which introduced the tenancy field was added very recently(it hit stage a couple of days ago), there are currently no consumers of this field.

Ack

/hold cancel

api/hypershift/v1beta1/nodepool_types.go
muraee: fair enough, updated.

PR #4735, opened 2024-09-17
OCPBUGS-41992: Sanitize ignition payload

jparrill:
Added a function which validates the JSON generated as Ignition payload and it also verifies the content of the payload is resilient and does not contains any issues

**Which issue(s) this PR fixes** *:
Fixes #[OCPBUGS-41992](https://issues.redhat.com/browse/OCPBUGS-41992)


ignition-server/cmd/start.go
```diff
@@ -247,10 +246,9 @@ func run(ctx context.Context, opts Options) error {
 			return
 		}
 
-		var jsonPayload map[string]interface{}
-		if err := json.Unmarshal(value.Payload, &jsonPayload); err != nil {
-			log.Printf("Invalid JSON payload")
-			http.Error(w, "Invalid payload", http.StatusInternalServerError)
+		if err := util.SanitizeIgnitionPayload(value.Payload); err != nil {
+			log.Printf("Invalid ignition payload")
```
enxebre, line 250: can we please include the error in the log print here?

support/util/util.go
```diff
@@ -476,3 +477,20 @@ func getHostedClusterScopeAnnotation(obj client.Object, r client.Reader) string
...
+// SanitizeIgnitionPayload make sure the IgnitionPayload is valid
+// and does not contain inconsistencies.
+func SanitizeIgnitionPayload(payload []byte) error {
+	var jsonPayload map[string]interface{}
+
+	if err := json.Unmarshal(payload, &jsonPayload); err != nil {
+		return fmt.Errorf("error unmarshalling Ignition payload: %v", err)
+	}
+
+	_, _, err := ignitionconfig.Parse(payload)
```
enxebre, line 490: This is nice!
We could also just 
```
	var jsonPayload ignitionapi.Config

	if err := json.Unmarshal(payload, &jsonPayload); err != nil {
		return fmt.Errorf("error unmarshalling Ignition payload: %v", err)
	}
```
that way reused the existing vendored api and save the additional 14kloc of deps.
We would lose the the "duplicated entry" unit test but I think that's ok since we own this ignition creation and it's pretty much a "proxy" that redirects to the real payload via merge.source. wdyt?

jparrill: Sure, thanks for the suggestion

enxebre:
thanks!
/approve
/lgtm

PR #4737, opened 2024-09-17
OCPBUGS-32592: Add role and rolebinding for Agent's Cluster CRs

CrystalChun:
Creates an additional role/rolebinding for Agent platform's resources (ClusterDeployment, AgentClusterInstall) in the Hosted Cluster namespace.

The Agent platform's resources such as ClusterDeployment and AgentClusterInstall were previously created in the Hosted Control Plane namespace. However, this caused issues when simulating the deletion portion of the backup and restore scenario since the hosted control plane namespace is removed when the hosted cluster is removed.

The [PR](https://github.com/openshift/cluster-api-provider-agent/pull/141) in the cluster-api-provider-agent side adjusts the resources to be deployed in the hosted cluster namespace, but the permissions to CRUD these resources in the hosted cluster namespace needs to be added via this change.

Fixes #OCPBUGS-32592

**Checklist**

### DIFF
--- a/cmd/cluster/agent/create.go
+++ b/cmd/cluster/agent/create.go
@@ -20,9 +20,10 @@ func DefaultOptions() *RawCreateOptions {
 }
 
 type RawCreateOptions struct {
-	APIServerAddress   string
-	AgentNamespace     string
-	AgentLabelSelector string
+	APIServerAddress       string
+	AgentNamespace         string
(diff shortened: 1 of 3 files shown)


CrystalChun:
/close
Decided to take a different approach in PR https://github.com/openshift/hypershift/pull/4741

PR #4741, opened 2024-09-17
OCPBUGS-32592: Add Annotation to skip deleting hcp namespace

CrystalChun:
Adds an annotation to prevent deletion of the hosted control plane namespace when set.

Near-term backup and restore for Agent platform requires that the HCP namespace remain intact during the hosted cluster deletion due to the Agent CRs that exist in that namespace.

Fixes #OCPBUGS-32592

**Checklist**

### DIFF
--- a/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
+++ b/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
@@ -3900,6 +3900,10 @@ func (r *HostedClusterReconciler) delete(ctx context.Context, hc *hyperv1.Hosted
 
 	r.KubevirtInfraClients.Delete(hc.Spec.InfraID)
 
+	if skipNSDeletion := hc.Annotations[hyperv1.SkipControlPlaneNamespaceDeletionAnnotation]; skipNSDeletion == "true" {
+		return true, nil
+	}
+
 	// Block until the namespace is deleted, so that if a hostedcluster is deleted and then re-created with the same name
 	// we don't error initially because we can not create new content in a namespace that is being deleted.
 	exists, err = hyperutil.DeleteIfNeeded(ctx, r.Client, &corev1.Namespace{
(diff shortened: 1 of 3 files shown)


hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -139,7 +139,8 @@ const (
 	jobHostedClusterNameLabel      = "hypershift.openshift.io/cluster-name"
 	jobHostedClusterNamespaceLabel = "hypershift.openshift.io/cluster-namespace"
 
-	etcdCheckRequeueInterval = 10 * time.Second
+	etcdCheckRequeueInterval          = 10 * time.Second
+	skipControlPlaneNamespaceDeletion = "hypershift.openshift.io/skip-delete-hosted-controlplane-namespace"
```
csrwng, line 143: typically we put these with other [api constants](https://github.com/openshift/hypershift/blob/ff5710e71806fb6e85c4bffdcde5b5ceb484a2c6/api/hypershift/v1beta1/hostedcluster_types.go#L28)

csrwng:
/hold
Let's wait to see if this is what is needed. Before deleting the namespace, we delete the CAPI cluster. It'd be good to know if that should be skipped as well.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
CrystalChun: Thanks Cesar! Moved it to that file. Is the naming alright?

CrystalChun:
> Let's wait to see if this is what is needed. Before deleting the namespace, we delete the CAPI cluster. It'd be good to know if that should be skipped as well.

AFAIU deleting the CAPI cluster doesn't affect the two resources we need (ClusterDeployment and AgentClusterInstall). I can test out the changes in this PR in my environment to confirm.

csrwng:
/hold cancel
Followed up with @jparrill and it looks like this should be sufficient for what we need.
@CrystalChun is it possible to test out a backup/restore with this annotation before we merge it?

CrystalChun:
> @CrystalChun is it possible to test out a backup/restore with this annotation before we merge it?
Yes! I just tested it out on my environment locally and it works as intended!

jparrill:
Thanks @CrystalChun!
/lgtm

PR #4754, opened 2024-09-19
NO-JIRA: remove CPO/CPPKIO from the hypershift-operator image build

sjenning:
`hypershift-operator` image builds are taking a long time in prow ~20m.

Don't build more than we need.


sjenning:
/override ci/prow/e2e-aws
it already passed once

PR #4768, opened 2024-09-20
NO-JIRA: cmd/dump: add OpenStack CAPI resources

EmilienM:
They were missing and will be useful for debug.


csrwng:
/retitle NO-JIRA: cmd/dump: add OpenStack CAPI resources

PR #4782, opened 2024-09-23
NO-JIRA: e2e: add service and cluster CIDRs as flags to the e2e framework

orenc1:
In some complex test scenarios, we would like to allow specifying custom IP ranges for the service and the cluster CIDR, to avoid overlapping addresses that use the defaults.

Fixes #

**Checklist**


davidvossel (commented):
/lgtm
/approve

PR #4801, opened 2024-09-25
OCPBUGS-42004: Add capability to Add Microsoft Managed Identity Sidecar Containers to HCP Pod Deployments

bryan-cox:
Adds two functions to include the managed identity sidecar containers in a deployment. The adapter-init is added as an init container and the adapter-server is added as a sidecar container in the related pod deployment it is included in.

Follow on work to integrate these functions and implement managed identity in HCP will be done in [OCPBUGS-42434](https://issues.redhat.com/browse/OCPBUGS-42434).

Fixes [OCPBUGS-42004](https://issues.redhat.com/browse/OCPBUGS-42004)

**Checklist**

### DIFF
--- a/support/azureutil/azureutil.go
+++ b/support/azureutil/azureutil.go
@@ -9,12 +9,21 @@ import (
 	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
 
 	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
 	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork/v5"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armresources"
+)
+
+// We received this images directly from Microsoft; we should expect them to change as Microsoft continues development on both containers.
+// We are scheduled to receive new updates to these containers in October 2024 to support Managed Identities. They currently only support Service Principal.
+// TODO past October, will we receive new versions?
+const (
+	AdapterInitImage   = "aromiwi.azurecr.io/artifact/b8e9ef87-cd63-4085-ab14-1c637806568c/buddy/adapter-init:20240905.9"
+	AdapterServerImage = "aromiwi.azurecr.io/artifact/b8e9ef87-cd63-4085-ab14-1c637806568c/buddy/adapter-server:20240905.5"
 )
 
 // GetSubnetNameFromSubnetID extracts the subnet name from a subnet ID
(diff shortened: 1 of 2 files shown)


support/azureutil/azureutil.go
```diff
@@ -217,3 +226,64 @@ func VerifyResourceGroupLocationsMatch(ctx context.Context, hc *hyperv1.HostedCl
 
 	return nil
 }
+
+// GetAzureCredentialsFromSecret gets the Service Principal client ID, client secret, and tenant ID from the credentials
+// secret. This function will be modified a bit once the Microsoft sidecar containers support Managed Identity are
+// delivered (expected Oct 2024).
```
jparrill, line 232: Do we know when in Oct will happen this to change? Maybe makes sense to wait until the change happens to put the final code?

support/azureutil/azureutil.go
```diff
@@ -217,3 +226,64 @@ func VerifyResourceGroupLocationsMatch(ctx context.Context, hc *hyperv1.HostedCl
...
+			return nil, fmt.Errorf("credentials secret for cluster doesn't have required key %s", expectedKey)
+		}
+	}
+
+	return &azureCredentials, nil
+}
+
+// AdapterInitContainer returns the Microsoft adapter-init init container. This container needs the NET_ADMIN permission
+// so the adapter-server sidecar container can intercept the Managed Identity Azure API authentication calls.
+func AdapterInitContainer() corev1.Container {
```
jparrill, line 253: This func is not being used, is that for a future implementation?

support/azureutil/azureutil.go
```diff
@@ -217,3 +226,64 @@ func VerifyResourceGroupLocationsMatch(ctx context.Context, hc *hyperv1.HostedCl
...
+					"NET_ADMIN",
+				},
+			},
+		}}
+}
+
+// AdapterServerContainer returns the Microsoft adapter-server sidecar container. Currently, this container mimics Azure
+// Managed Identity approval and returns an authentication token. The container currently needs a Service Principal to
+// do this. Future versions of this container will be able to take a Managed Identity instead.
+func AdapterServerContainer(clientID, clientSecret, tenantID string) corev1.Container {
```
jparrill, line 270: This func is not being used, is that for a future implementation?

support/azureutil/azureutil.go
```diff
@@ -217,3 +226,64 @@ func VerifyResourceGroupLocationsMatch(ctx context.Context, hc *hyperv1.HostedCl
...
+func GetAzureCredentialsFromSecret(ctx context.Context, c client.Client, namespace, credsName string) (*corev1.Secret, error) {
+	var azureCredentials corev1.Secret
+
+	// Retrieve the Azure credentials secret to extract the needed fields for the managed identity containers
+	credentialsSecretName := client.ObjectKey{Namespace: namespace, Name: credsName}
+	if err := c.Get(ctx, credentialsSecretName, &azureCredentials); err != nil {
+		return nil, fmt.Errorf("failed to get secret %s: %w", credentialsSecretName, err)
+	}
+
+	for _, expectedKey := range []string{"AZURE_CLIENT_ID", "AZURE_CLIENT_SECRET", "AZURE_TENANT_ID"} {
```
Patryk-Stefanski, line 242: Should we expect the subscription_id as well or do we retrieve that from elsewhere ?

support/azureutil/azureutil.go
bryan-cox: We do not know when we will receive it. Most likely no earlier than the end of October.

I think we should continue on and not hold up the work.

support/azureutil/azureutil.go
bryan-cox: Yeah, this is listed in the PR description - Follow on work to integrate these functions and implement managed identity in HCP will be done in [OCPBUGS-42434](https://issues.redhat.com/browse/OCPBUGS-42434).

support/azureutil/azureutil.go
bryan-cox: Yeah, this is listed in the PR description - Follow on work to integrate these functions and implement managed identity in HCP will be done in [OCPBUGS-42434](https://issues.redhat.com/browse/OCPBUGS-42434).

support/azureutil/azureutil.go
bryan-cox: No that is directly exposed in the API - https://github.com/openshift/hypershift/blob/da39c6232f098798325bc41eecaad1fe0c2f0090/api/hypershift/v1beta1/hostedcluster_types.go#L1851

support/azureutil/azureutil.go
```diff
@@ -217,3 +226,64 @@ func VerifyResourceGroupLocationsMatch(ctx context.Context, hc *hyperv1.HostedCl
 
 	return nil
 }
+
+// GetAzureCredentialsFromSecret gets the Service Principal client ID, client secret, and tenant ID from the credentials
+// secret. This function will be modified a bit once the Microsoft sidecar containers support Managed Identity are
+// delivered (expected Oct 2024).
+func GetAzureCredentialsFromSecret(ctx context.Context, c client.Client, namespace, credsName string) (*corev1.Secret, error) {
```
csrwng, line 233: I could see this one split into 2. In general, any code that requires a client, we put in the main hcp controller. This could do the same with a *corev1.Secret as an argument.

stevekuznetsov:
/hold

Sorry about this - the design was not finalized. I don't have access to the OCPBUGS component so I can't see how this was communicated. In any case even if we do end up with an init/sidecar, the version of the image in there needs to be able to change without releasing a new version of HyperShift, and we need to be able to rev that without causing a thundering herd, which implies that it needs to be a field on the `HostedCluster`.

bryan-cox:
/close

Closing this PR since Microsoft wants to move in a different direction and not use the Managed Identity Sidecar Containers

PR #4802, opened 2024-09-25
OCPBUGS-42433: Retry reconciliation with invalid configuration

csrwng:
Before this commit, if configuration was determined to be invalid, we simply returned a nil from the reconciliation loop and waited for something to change in the HostedCluster. This fails to account for when the reason for the invalid configuration is transient, such as a missing ConfigMap. When the ConfigMap is finally made available, the HostedCluster does not resume reconciliation.

Now an error is thrown when configuration is not valid, resulting in a requeue of the HostedCluster, until the invalid configuration is resolved.

Fixes #[OCPBUGS-42433](https://issues.redhat.com/browse/OCPBUGS-42433)

**Checklist**

### DIFF
--- a/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
+++ b/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
@@ -1136,8 +1136,8 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
 	{
 		validConfig := meta.FindStatusCondition(hcluster.Status.Conditions, string(hyperv1.ValidHostedClusterConfiguration))
 		if validConfig != nil && validConfig.Status == metav1.ConditionFalse {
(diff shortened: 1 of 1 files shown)


hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -1137,7 +1137,8 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
 		validConfig := meta.FindStatusCondition(hcluster.Status.Conditions, string(hyperv1.ValidHostedClusterConfiguration))
 		if validConfig != nil && validConfig.Status == metav1.ConditionFalse {
 			log.Error(fmt.Errorf("configuration is invalid"), "reconciliation is blocked", "message", validConfig.Message)
```
muraee, line 1139: should we remove the log here, since we are returning the error now?

PR #4282, opened 2024-06-27
HOSTEDCP-1788: Create seperate RGs for NSG and VNet

Patryk-Stefanski:
**What this PR does / why we need it**:
Creates seperate RGs for NSG and VNet

**Which issue(s) this PR fixes** *(optional, use `fixes #<issue_number>(, fixes #<issue_number>, ...)` format, where issue_number might be a GitHub issue, or a Jira story*:
[HOSTEDCP-1788](https://issues.redhat.com/browse/HOSTEDCP-1788)

**Checklist**
- [X] Subject and description added to both, commit and PR.
- [X] Relevant issues have been referenced.
- [ ] This change includes docs. 
- [ ] This change includes unit tests.

PR #4809, opened 2024-09-26
HOSTEDCP-1788: use getResourceGroupName for infra deletion in azure so destroy infra can be ran without --resource-group-name

Patryk-Stefanski:

use getResourceGroupName for infra deletion in azure so destroy infra can be ran without --resource-group-name

Follow up PR for comments on https://github.com/openshift/hypershift/pull/4282

Fixes # [HOSTEDCP-1788](https://issues.redhat.com/browse/HOSTEDCP-1788)

**Checklist**


bryan-cox:
Can you update the commit message with the why we need this? Also remove the `-2` in the subject.

PR #4811, opened 2024-09-26
OCPBUGS-42434: Add Managed Identity Support in Azure HC API

bryan-cox:
Adds fields in the Azure HostedCluster API for the client IDs related to the managed service identities used for the following control plane components: azure cloud provider, cluster-storage-operator, KMS, CAPZ, the control plane operator, the image registry operator, cluster ingress operator, cluster network operator, azure-disk-controller and azure-file-controller.

Fixes [OCPBUGS-42434](https://issues.redhat.com/browse/OCPBUGS-42434)

**Checklist**


api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1859,6 +1859,97 @@ type AzurePlatformSpec struct {
 	// +immutable
 	// +required
 	SecurityGroupID string `json:"securityGroupID,omitempty"`
+
+	// ManagedIdentities contains the client IDs related to the managed identities needed for HCP control plane
```
JoelSpeed, line 1863: Should start with the serialised version of the field name
```suggestion
	// managedIdentities contains the client IDs related to the managed identities needed for HCP control plane
```

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1859,6 +1859,97 @@ type AzurePlatformSpec struct {
...
+	// and data plane components that authenticate with Azure's API.
+	//
+	// +kubebuilder:validation:Required
+	// +required
```
JoelSpeed, line 1867: You only need one of these, typically we've gone for the kubebuilder version but the very latest release now accepts both. You should try to be consistent across your codebase and choose one

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1859,6 +1859,97 @@ type AzurePlatformSpec struct {
...
+
+// AzureResourceManagedIdentities contains the client IDs related to the managed identities needed for HCP control plane
+// and data plane components that authenticate with Azure's API.
+type AzureResourceManagedIdentities struct {
+	// ControlPlaneMIs contains the client IDs of all the managed identities on the HCP control plane needing to
+	// authenticate with Azure's API.
+	//
+	// +kubebuilder:validation:Required
+	// +required
+	ControlPlaneMIs *ControlPlaneManagedIdentities `json:"controlPlaneMIs"`
```
JoelSpeed, line 1879: Don't use abbreviations in field names

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1859,6 +1859,97 @@ type AzurePlatformSpec struct {
...
+// ManagedIdentityClientID is a client ID of a managed identity
+// +kubebuilder:validation:XValidation:rule="self.matches('^[{]?[0-9a-fA-F]{8}-([0-9a-fA-F]{4}-){3}[0-9a-fA-F]{12}[}]?$')",message="the client ID of a managed identity must be a valid UUID. It should be 5 groups of hyphen separated hexadecimal characters in the form 8-4-4-4-12."
+type ManagedIdentityClientID string
+
+// ControlPlaneManagedIdentities contains the client IDs of all the managed identities on the HCP control plane needing
+// to authenticate with Azure's API.
+// Managed identity regex pattern is from Microsoft here - https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules#microsoftmanagedidentity.
+// The format a managed identity should be `/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.ManagedIdentity/userAssignedIdentities/{managedIdentityName}`.
+type ControlPlaneManagedIdentities struct {
+	// AzureCloudProviderMIClientID is the client ID of a pre-existing managed identity ID associated with the azure
```
JoelSpeed, line 1893: All of these fields should have something in the godoc that explains the format, and therefore, the validations that are applied to the strings.

You have it on the struct, but it should be at the field level too

api/hypershift/v1beta1/hostedcluster_types.go
JoelSpeed, line 1879: Required structs should not be pointers
```suggestion
	ControlPlaneMIs ControlPlaneManagedIdentities `json:"controlPlaneMIs"`
```

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1859,6 +1859,97 @@ type AzurePlatformSpec struct {
 	// +immutable
 	// +required
 	SecurityGroupID string `json:"securityGroupID,omitempty"`
+
+	// ManagedIdentities contains the client IDs related to the managed identities needed for HCP control plane
+	// and data plane components that authenticate with Azure's API.
+	//
+	// +kubebuilder:validation:Required
+	// +required
+	ManagedIdentities *AzureResourceManagedIdentities `json:"managedIdentities,omitempty"`
```
JoelSpeed, line 1868: Required structs should not be pointers
```suggestion
	ManagedIdentities AzureResourceManagedIdentities `json:"managedIdentities,omitempty"`
```

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1859,6 +1859,97 @@ type AzurePlatformSpec struct {
 	// +immutable
 	// +required
 	SecurityGroupID string `json:"securityGroupID,omitempty"`
+
+	// ManagedIdentities contains the client IDs related to the managed identities needed for HCP control plane
+	// and data plane components that authenticate with Azure's API.
+	//
+	// +kubebuilder:validation:Required
```
JoelSpeed, line 1866: The AzurePlatformSpec has recently been added as it not currently shipped right? Adding new required fields to something that already has shipped would cause issues

bryan-cox: It's not been shipped for customers. This is part of ARO HCP so these fields will only be used by them. This PR will need to be backported to 4.17.z though as that is what they will be using to private preview.

api/hypershift/v1beta1/hostedcluster_types.go
bryan-cox: This has been added.

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1857,8 +1857,98 @@ type AzurePlatformSpec struct {
...
+}
+
+// AzureResourceManagedIdentities contains the client IDs related to the managed identities needed for HCP control plane
+// and data plane components that authenticate with Azure's API.
+type AzureResourceManagedIdentities struct {
+	// ControlPlaneManagedIdentities contains the client IDs of all the managed identities on the HCP control plane needing to
+	// authenticate with Azure's API.
+	//
+	// +kubebuilder:validation:Required
+	ControlPlaneManagedIdentities ControlPlaneManagedIdentities `json:"controlPlaneManagedIdentities"`
```
enxebre, line 1876: This is withiin ManagedIdentities already, just controlPlane? also do we need the controlplane vs dataplane distinction?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1857,8 +1857,98 @@ type AzurePlatformSpec struct {
...
+// to authenticate with Azure's API.
+// Managed identity regex pattern is from Microsoft here - https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules#microsoftmanagedidentity.
+// The format a managed identity should be `/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.ManagedIdentity/userAssignedIdentities/{managedIdentityName}`.
+type ControlPlaneManagedIdentities struct {
+	// azureCloudProviderManagedIdentityClientID is the client ID of a pre-existing managed identity ID associated with the azure
+	// cloud provider, aka ccm. The client ID of a managed identity must be a valid UUID. It should be 5 groups of
+	// hyphen separated hexadecimal characters in the form 8-4-4-4-12.
+	//
+	// +kubebuilder:validation:Required
+	AzureCloudProviderManagedIdentityClientID ManagedIdentityClientID `json:"azureCloudProviderManagedIdentityClientID"`
```
enxebre, line 1895: this is within Azure, ManagedIdentities, should this just be cloudProvider?

api/hypershift/v1beta1/hostedcluster_types.go
bryan-cox: I'm okay with making this just ControlPlane.

Yeah, we will be adding the DataPlane managed identities in [HOSTEDCP-1542](https://issues.redhat.com/browse/HOSTEDCP-1542), which should quickly follow on to the managed identity work.

api/hypershift/v1beta1/hostedcluster_types.go
bryan-cox: Ack, will make it so

muraee: I think this still need to be changed. its currently `spec.azure.managedIdentities.cloudProviderManagedIdentityClientID`
it should be just  `spec.azure.managedIdentities.cloudProvider`   or `spec.azure.managedIdentities.cloudProviderClientID`

same for all fields.

bryan-cox: Updated

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1857,8 +1857,98 @@ type AzurePlatformSpec struct {
...
+	//
+	// +kubebuilder:validation:Required
+	DiskClientID ManagedIdentityClientID `json:"diskClientID"`
+
+	// fileClientID is the client ID of a pre-existing managed identity ID associated with the
+	// azure-disk-controller. The client ID of a managed identity must be a valid UUID. It should be 5 groups of hyphen
+	// separated hexadecimal characters in the form 8-4-4-4-12.
+	//
+	// +kubebuilder:validation:Required
+	FileClientID ManagedIdentityClientID `json:"fileClientID"`
```
JoelSpeed, line 1951: Both os these are for the disk controller?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1857,8 +1857,98 @@ type AzurePlatformSpec struct {
...
+// +kubebuilder:validation:XValidation:rule="self.matches('^[{]?[0-9a-fA-F]{8}-([0-9a-fA-F]{4}-){3}[0-9a-fA-F]{12}[}]?$')",message="the client ID of a managed identity must be a valid UUID. It should be 5 groups of hyphen separated hexadecimal characters in the form 8-4-4-4-12."
+type ManagedIdentityClientID string
+
+// ControlPlaneManagedIdentities contains the client IDs of all the managed identities on the HCP control plane needing
+// to authenticate with Azure's API.
+// Managed identity regex pattern is from Microsoft here - https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules#microsoftmanagedidentity.
+// The format a managed identity should be `/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.ManagedIdentity/userAssignedIdentities/{managedIdentityName}`.
+type ControlPlaneManagedIdentities struct {
+	// cloudProviderClientID is the client ID of a pre-existing managed identity ID associated with the azure
+	// cloud provider, aka ccm. The client ID of a managed identity must be a valid UUID. It should be 5 groups of
```
JoelSpeed, line 1891: I would avoid the abbrevation CCM and instead just say `Azure Cloud Controller Manager and Azure Cloud Node Manager` assuming this does also apply to the CNM.

Since this is part of the control plane group of client IDs, I guess not for the CNM, and that's separately authenticated?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1857,8 +1857,98 @@ type AzurePlatformSpec struct {
...
+	// and data plane components that authenticate with Azure's API.
+	//
+	// +kubebuilder:validation:Required
+	ManagedIdentities AzureResourceManagedIdentities `json:"managedIdentities,omitempty"`
+}
+
+// AzureResourceManagedIdentities contains the client IDs related to the managed identities needed for HCP control plane
+// and data plane components that authenticate with Azure's API.
+type AzureResourceManagedIdentities struct {
+	// ControlPlane contains the client IDs of all the managed identities on the HCP control plane needing to
```
JoelSpeed, line 1872:
```suggestion
	// controlPlane contains the client IDs of all the managed identities on the HCP control plane needing to
```

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1857,8 +1857,98 @@ type AzurePlatformSpec struct {
...
+	// hexadecimal characters in the form 8-4-4-4-12.
+	//
+	// +kubebuilder:validation:Required
+	ControlPlaneClientID ManagedIdentityClientID `json:"controlPlaneClientID"`
+
+	// kmsClientID is the client ID of a pre-existing managed identity ID associated with Azure KMS. The client
+	// ID of a managed identity must be a valid UUID. It should be 5 groups of hyphen separated hexadecimal characters
+	// in the form 8-4-4-4-12.
+	//
+	// +optional
```
miguelsorianod, line 1915: Why is this one marked as optional?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1857,8 +1857,98 @@ type AzurePlatformSpec struct {
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="SecurityGroupID is immutable"
 	// +kubebuilder:validation:Required
 	// +immutable
-	// +required
```
miguelsorianod, line 1860: Whys is the required annotation removed from the security group? Is it correct?

If it is correct, what is then the behavior when you don't provide one?

api/hypershift/v1beta1/hostedcluster_types.go
bryan-cox: Yes it runs two different controllers

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1857,8 +1857,98 @@ type AzurePlatformSpec struct {
...
+}
+
+// AzureResourceManagedIdentities contains the client IDs related to the managed identities needed for HCP control plane
+// and data plane components that authenticate with Azure's API.
+type AzureResourceManagedIdentities struct {
+	// ControlPlane contains the client IDs of all the managed identities on the HCP control plane needing to
+	// authenticate with Azure's API.
+	//
+	// +kubebuilder:validation:Required
+	ControlPlane ControlPlaneManagedIdentities `json:"controlPlane"`
```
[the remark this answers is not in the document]
bryan-cox: Yes we do. Those will be for implementing workload identity.

api/hypershift/v1beta1/hostedcluster_types.go
bryan-cox: KMS is normally optional on HCP and not required

miguelsorianod: Doesn't the KMS identity become *required* if KMS is leveraged?

Looking at the Hypershift API of HostedCluster, for the case of AWS STS I don't see the KMS identity being defined in `.platform.aws.rolesRef` AWSPlatformSpec. Instead, it seems to be defined at the the functionality usage point. This is, in `.secretEncryption.kms.aws.auth.awsKms`.

This seems to be inconsistent. Are we aware of that difference? Is there some reason for it?
Is hypershift validating that it is "optional", but "required" when kms is/will be used for the cluster being created?

bryan-cox: Yeah that does seem inconsistent. I was not aware of the difference. 

I'm inclined to remove the managed identity that is needed here out of the HC and into `AzureKMSSpec`. WDYT @muraee since you worked on the KMS implementation?

api/hypershift/v1beta1/hostedcluster_types.go
JoelSpeed: Ahh yes fair point, you don't have control plane nodes! Ok, all good

api/hypershift/v1beta1/hostedcluster_types.go
JoelSpeed: You could use a CEL validation if they are on the same object to make sure that they are either both supplied, or neither are supplied

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1857,8 +1857,98 @@ type AzurePlatformSpec struct {
...
+
+	// Future placeholder - DataPlaneMIs * DataPlaneManagedIdentities
+}
+
+// ManagedIdentityClientID is a client ID of a managed identity
+// +kubebuilder:validation:XValidation:rule="self.matches('^[{]?[0-9a-fA-F]{8}-([0-9a-fA-F]{4}-){3}[0-9a-fA-F]{12}[}]?$')",message="the client ID of a managed identity must be a valid UUID. It should be 5 groups of hyphen separated hexadecimal characters in the form 8-4-4-4-12."
+type ManagedIdentityClientID string
```
JoelSpeed, line 1883: This string has a fixed length, can you set the minimum and maximum length markers to the correct value? 36 if I counted correctly?

api/hypershift/v1beta1/hostedcluster_types.go
JoelSpeed: Yet you've called them both azure-disk-controller, you should call them out as different things else the descriptions would lead me to believe they are for the same thing?

api/hypershift/v1beta1/hostedcluster_types.go
muraee: yeah I think it makes more sense to move this field to `AzureKMSSpec` and stay consistent.

bryan-cox:
Moving this PR back to draft. Microsoft indicated yesterday they do not want to move forward with the sidecar container implementation and instead use a backing certificate with the managed identity. The API will need to change a bit to acomodate this new change.

bryan-cox:
/unhold

We still need this PR when authenticating with certs. The PR is in draft as well.

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -2008,8 +2008,116 @@ type AzurePlatformSpec struct {
...
+	//
+	// More information on how the Secrets Store CSI driver works to do this can be found here:
+	// https://learn.microsoft.com/en-us/azure/aks/csi-secrets-store-driver.
+	//
+	// +kubebuilder:validation:Required
+	ManagementKeyVault ManagedAzureKeyVault `json:"managementKeyVault"`
+}
+
+// ManagedAzureKeyVault is the management cluster's Azure Key Vault where the managed identity certificates are stored.
+type ManagedAzureKeyVault struct {
```
miguelsorianod, line 2031: Isn't this Key Vault called "Managemenet Key Vault" instead of "Managed Key Vault" ?

cc @geoberle 

Asking to try to be consistent with the naming in all components

bryan-cox: Is there some other component where its named "Management Key Vault"?

bryan-cox: I kept this the same, but renamed the one for the managed identities in the latest push

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -2008,8 +2008,118 @@ type AzurePlatformSpec struct {
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="SecurityGroupID is immutable"
 	// +kubebuilder:validation:Required
 	// +immutable
-	// +required
-	SecurityGroupID string `json:"securityGroupID,omitempty"`
+	SecurityGroupID string `json:"securityGroupID"`
+
+	// keyVaultManagedIdentityClientID is the client ID of the managed identity which can access the Azure Key Vaults,
+	// in the management cluster, to retrieve secrets and certificates, like certificates related to the
+	// ControlPlaneManagedIdentities for example. The HostedCluster controller populates this field for ARO HCP
```
enxebre, line 2015: > The HostedCluster controller populates this field for ARO HCP
	// deployments if it isn't already initialized.
	//

Can we drop this statement? I don't think we want the controller to populate its own API. We can just let it default transparently without updating the field if we want to.

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -2008,8 +2008,118 @@ type AzurePlatformSpec struct {
...
+	// cloudProvider is a pre-existing managed identity associated with the azure cloud provider, aka cloud controller
+	// manager.
+	//
+	// +kubebuilder:validation:Required
+	CloudProvider ManagedIdentity `json:"cloudProvider"`
+
+	// clusterAPIAzure is a pre-existing managed identity associated with cluster-api azure.
+	//
+	// +kubebuilder:validation:Required
+	ClusterAPIAzure ManagedIdentity `json:"clusterAPIAzure"`
```
enxebre, line 2092: Azure is redundant here. Also we might want to consolidate with the term we use in aws i.e NodePoolManagement, thoughts?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -2008,8 +2008,118 @@ type AzurePlatformSpec struct {
...
+
+	// controlPlaneClientID is a pre-existing managed identity associated with the control plane operator.
+	//
+	// +kubebuilder:validation:Required
+	ControlPlane ManagedIdentity `json:"controlPlane"`
```
enxebre, line 2097: I find it non intuitive to have ControlPlane within ControlPlaneManagedIdentities. In aws we refer to this as controlPlaneOperator, thoughts?

enxebre:
some minor comments, lgtm.

api/hypershift/v1beta1/hostedcluster_types.go
miguelsorianod: > Also we might want to consolidate with the term we use in aws i.e NodePoolManagement, thoughts?
Although it is true that changing it to `NodePoolManagement` would consolidate with how it is in AWS, that name is not really accurate. The operator is the cluster api operator and not "node pool management".

Regarding "azure" being redundant, is it? my understanding is that we have capi (https://github.com/kubernetes-sigs/cluster-api-operator) , but also capz (https://github.com/kubernetes-sigs/cluster-api-provider-azure), and capa (https://github.com/kubernetes-sigs/cluster-api-provider-aws). Which one of them is representing?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -2556,6 +2666,11 @@ type AzureKMSSpec struct {
 	// secrets can continue to be decrypted until they are all re-encrypted with the active key.
 	// +optional
 	BackupKey *AzureKMSKey `json:"backupKey,omitempty"`
+
+	// kms is a pre-existing managed identity used to authenticate with Azure KMS.
+	//
+	// +kubebuilder:validation:Required
+	KMS ManagedIdentity `json:"kms,omitempty"`
```
muraee, line 2673: required fields should not have `omitempty`

api/hypershift/v1beta1/hostedcluster_types.go
bryan-cox: I'm good with calling it ControlPlaneOperator

api/hypershift/v1beta1/hostedcluster_types.go
bryan-cox: I like being explicit here since that MI is specifically for CAPZ and to avoid any confusion with CAPI or CAPA like Miguel alluded to. It's not a hill I'm willing to die on though 😄

api/hypershift/v1beta1/hostedcluster_types.go
bryan-cox: Discussed this on Slack with Alberto:

CS is going to use the same MI here for all HCs on a particular mgmt cluster. They wanted to pass this value in through the HO installation ([I'm setting it as an env var on the HO pod here](https://github.com/openshift/hypershift/pull/4877/commits/43274102ccf8e3f906534e26481cef9ae563a6da#diff-4edc0a874ce10f9d4f77594c9ff489c3275cb2adc788826a9ea2f48a64c336efR486)) and not populate it on each HC CR. We need this value on the CPO side for creating the SecretProviderClass CRs.

This field should get set in the HCP, at its creation time.

bennerv: Should this be part of status then instead of spec?

bryan-cox: @bennerv @miguelsorianod - I've updated this now. The field here is populated when the HCP is created.

miguelsorianod: I think we didn't understand each other when discussing this.

We understood the attribute would be populated in the HCP and not in the HC itself. The original comment by Ben still applies then

bryan-cox: Update - we discussed this offline (Miguel, Ben, & myself). We are going to remove this field and the HO will pass the KY MI client ID through an env var to the CPO.

api/hypershift/v1beta1/hostedcluster_types.go
enxebre: CAPI/CAPZ is an implementation detail we could technically transparently swap if we choose to. Therefore I'd prefer if we avoin baking a reference to it in the API, hence my suggestion to align with the aws naming we chose.
In any case I'm also ok with your preference if you feel strongly about it.

bryan-cox: I've updated this.

enxebre:
/retitle OCPBUGS-42434: Add Managed Identity Support in Azure HC API behind AROHCPManagedIdentities Feature gate
/lgtm

PR #4814, opened 2024-09-26
HOSTEDCP-1930: Add AWS Shared VPC support

csrwng:
Adds fields to the AWS platform section of the HostedCluster spec to
support cluster creation on a shared VPC.

Makes use of the new fields in the AWSEndpointServiceReconciler and the
DNS global config reconciliation.

Adds a flag (--vpc-owner-aws-credentials) to the following commands:
- hypershift create cluster aws
- hypershift create infra aws
- hypershift create iam aws
- hypershift destroy cluster aws
- hypershift destroy infra aws
- hypershift destroy iam aws

The new flag allows creating a VPC in a different AWS account from the
one where the HostedCluster is created.

Fixes #[HOSTEDCP-1930](https://issues.redhat.com/browse/HOSTEDCP-1930)

**Checklist**


api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1350,6 +1350,33 @@ type AWSPlatformSpec struct {
...
+	// will be created.
+	//
+	// +optional
+	SharedVPC *AWSSharedVPC `json:"sharedVPC,omitempty"`
+}
+
+// AWSSharedVPC contains fields needed to create a HostedCluster using a VPC that has been
+// created and shared from a different AWS account than the AWS account where the cluster
+// is getting created.
+type AWSSharedVPC struct {
```
bryan-cox, line 1365: These fields should be listed as required

bryan-cox, line 1365: Also, is there a regex validation that could be added to these?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1350,6 +1350,33 @@ type AWSPlatformSpec struct {
 	// +kubebuilder:default=false
 	// +optional
 	MultiArch bool `json:"multiArch"`
+
+	// SharedVPC contains fields that must be specified if the HostedCluster is using a VPC that was
+	// created in a different AWS account and has been shared to the AWS account where the HostedCluster
+	// will be created.
```
devguyio, line 1356:
```suggestion
	// SharedVPC contains fields that must be specified if the HostedCluster must use a VPC that is
	// created in a different AWS account and is shared with the AWS account where the HostedCluster
	// will be created.
```

Is there a reason you used the present continuous tense in  `if the HostedCluster is using a VPC` , and later used a future tense `where the HostedCluster will be created` ? 

Fwiw, I made this suggestion to show how I would have written it.

cmd/infra/aws/destroy.go
```diff
@@ -223,17 +245,22 @@ func (o *DestroyInfraOptions) DestroyInfra(ctx context.Context) error {
 		s3Client = delegatingClent.S3API
 	}
 
+	_ = vpcOwnerEC2Client
+	_ = vpcOwnerRoute53Client
+	_ = ramClient
```
sjenning, line 250: these can be removed right?

csrwng: yes, ty!

devguyio (changes requested):
cool! I left a few comments but otherwise lgtm

cmd/infra/aws/create.go
```diff
@@ -249,15 +279,131 @@ func (o *CreateInfraOptions) CreateInfra(ctx context.Context, l logr.Logger) (*C
...
+				SubnetIds: allSubnetIDsToShare,
+			})
+			if err != nil || len(subnetResult.Subnets) != len(allSubnetIDsToShare) {
+				l.Info("Waiting for subnets to be available in cluster creator account")
+				return fmt.Errorf("not ready yet")
+			}
+			return nil
+		}); err != nil {
+			return nil, err
+		}
```
devguyio, line 369: Maybe this should be extracted to a separate function for sharing the subnets?

api/hypershift/v1beta1/hostedcluster_types.go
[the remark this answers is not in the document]
csrwng: thx!, no specific reason for the tenses :)

cmd/infra/aws/create.go
csrwng: Thx, addressed

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1350,6 +1350,45 @@ type AWSPlatformSpec struct {
...
+type AWSSharedVPC struct {
+
+	// IngressRoleARN is the ARN of a role in the VPC owner account that allows managing DNS records
+	// in the private hosted zone of the HostedCluster. The ingress operator will assume this role to
+	// manage DNS records.
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Pattern:=`^arn:(aws|aws-cn|aws-us-gov):iam::[0-9]{12}:role\/.*$`
+	// +required
+	IngressRoleARN string `json:"ingressRoleARN"`
```
enxebre, line 1374: have we considered having RolesRef AWSSharedVPCRolesRef `json:"rolesRef"`?

csrwng: I've now added a struct for AWSSharedVPCRolesRef. I also documented the referenced roles in the same way we do for AWSRolesRef. Thx!

control-plane-operator/main.go
```diff
@@ -458,8 +471,18 @@ func NewStartCommand() *cobra.Command {
 				os.Exit(1)
 			}
 
+			var assumeRole *string
+			var localZoneID *string
+
+			if hcp.Spec.Platform.AWS != nil && hcp.Spec.Platform.AWS.SharedVPC != nil {
```
enxebre, line 477: why are we introducing this pattern instead of propagating fields to HCP and let the controller reconcile it as in present implementation https://github.com/openshift/hypershift/blob/4ab267a7179127d196b147e40532fb18c4ce65a6/control-plane-operator/controllers/awsprivatelink/awsprivatelink_controller.go#L147

enxebre:
I see a bomb when the docs try to render mermaid https://deploy-preview-4814--hypershift-docs.netlify.app/how-to/aws/shared-vpc/

control-plane-operator/main.go
csrwng: The main reason is that we construct the aws clients for the AWSEndpointServiceReconciler on setup:
https://github.com/openshift/hypershift/blob/4ab267a7179127d196b147e40532fb18c4ce65a6/control-plane-operator/controllers/awsprivatelink/awsprivatelink_controller.go#L194-L218

For the shared vpc case, the client needs to be setup to assume the shared vpc control plane role if configured.
I'm fine changing that pattern in that reconciler so we don't have to look up the HostedCluster in main.go
tbh, not sure how much overhead is added by creating the client on every reconcile.

csrwng: Follow up tracked with https://issues.redhat.com/browse/HOSTEDCP-2012

PR #4825, opened 2024-09-27
HOSTEDCP-1706: feat: add a metric for cluster size override annotation

devguyio:

Add a metric to track the rosa hosted clusters which have the `hypershift.openshift.io/cluster-size-override` annotation set. 

Fixes [HOSTEDCP-1706](https://issues.redhat.com//browse/HOSTEDCP-1706)

**Checklist**

### DIFF
--- a/hypershift-operator/controllers/hostedcluster/metrics/metrics.go
+++ b/hypershift-operator/controllers/hostedcluster/metrics/metrics.go
@@ -89,6 +89,9 @@ const (
 
 	EtcdManualInterventionRequiredMetricName = "hypershift_etcd_manual_intervention_required"
 	etcdManualInterventionRequiredMetricHelp = "Indicates that manual intervention is required to recover the ETCD cluster"
+
+	ClusterSizeOverrideMetricName = "hypershift_cluster_size_override_instances"
+	clusterSizeOverrideMetricHelp = "Number of HostedClusters with a cluster size override annotation"
 )
 
 // semantically constant - not supposed to be changed at runtime
@@ -177,7 +180,11 @@ var (
 
 	etcdManualInterventionRequiredMetricDesc = prometheus.NewDesc(
 		EtcdManualInterventionRequiredMetricName, etcdManualInterventionRequiredMetricHelp,
-		append(hclusterLabels, "rosa_environment", "rosa_id"), nil)
+		append(hclusterLabels, "environment", "internal_id"), nil)
+
+	clusterSizeOverrideMetricDesc = prometheus.NewDesc(
+		ClusterSizeOverrideMetricName, clusterSizeOverrideMetricHelp,
+		append(hclusterLabels, "environment", "internal_id", "size"), nil)
 )
 
 type hostedClustersMetricsCollector struct {
@@ -460,6 +467,7 @@ func (c *hostedClustersMetricsCollector) Collect(ch chan<- prometheus.Metric) {
 			}
 
 			// etcdManualInterventionRequiredMetric
+			// clusterSizeOverrideMetric
 			{
 				metricLabels := make(map[string]string, 0)
 				if hcluster.Spec.Platform.Type == hyperv1.AWSPlatform && hcluster.Spec.Platform.AWS.ResourceTags != nil {
@@ -468,7 +476,7 @@ func (c *hostedClustersMetricsCollector) Collect(ch chan<- prometheus.Metric) {
 						case "api.openshift.com/environment":
 							metricLabels["environment"] = resourceTag.Value
 						case "api.openshift.com/id":
-							metricLabels["id"] = resourceTag.Value
+							metricLabels["internal_id"] = resourceTag.Value
 						case "red-hat-clustertype":
 							metricLabels["cluster_type"] = resourceTag.Value
 						}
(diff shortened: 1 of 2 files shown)


jparrill (commented):
Overall lgtm, I'm not a big fan of changing names of imports until we introduce a new import that maybe conflicts with the existent one, but I also don't have hard feelings to approve it.

devguyio:
> Overall lgtm, I'm not a big fan of changing names of imports until we introduce a new import that maybe conflicts with the existent one, but I also don't have hard feelings to approve it.

Thanks! I see your point about changing imports but for this particular case I renamed because:

1. I'm trying to follow a "leave the code you touch better than you found it" approach to improve our codebase.
2. The previous name `hyperv1` IHMO is misleading. This is a v1beta1 API and that import name gives the wrong assumption it's a v1, stable API. E.g. the import above  it is `configv1`, yet it's actually a v1 API.

hypershift-operator/controllers/hostedcluster/metrics/metrics.go
```diff
@@ -472,17 +480,26 @@ func (c *hostedClustersMetricsCollector) Collect(ch chan<- prometheus.Metric) {
...
 						)
 
 					}
+
+					if sizeOverride := hcluster.Annotations[hyperv1beta1.ClusterSizeOverrideAnnotation]; sizeOverride != "" {
+						ch <- prometheus.MustNewConstMetric(
+							clusterSizeOverrideMetricDesc,
+							prometheus.GaugeValue,
+							1.0,
+							append(hclusterLabelValues, metricLabels["environment"], metricLabels["id"], sizeOverride)...,
```
typeid, line 500: **Question:** we are already adding an `_id` label, how does that differ from the `id` here? It might be confusing to have an `_id` label as well as `id`.

Refs:
- https://github.com/openshift/hypershift/blob/37db51a9793e78bdf3a7621e958f605fe1f04c84/hypershift-operator/controllers/hostedcluster/metrics/metrics.go#L330
- https://github.com/openshift/hypershift/blob/37db51a9793e78bdf3a7621e958f605fe1f04c84/hypershift-operator/controllers/hostedcluster/metrics/metrics.go#L129

devguyio: as agreed offline, we're keeping `_id` and will name other one `internal_id`

hypershift-operator/controllers/hostedcluster/metrics/metrics.go
```diff
@@ -104,7 +107,7 @@ var (
 		configv1.IdentityProviderTypeRequestHeader,
 	}
 
-	knownPlatforms = hyperv1.PlatformTypes()
+	knownPlatforms = hyperv1beta1.PlatformTypes()
```
enxebre, line 110: fwiw we use hyperv1 across the code base, that alias would save a lot of replacements when there's an api bump.

hypershift-operator/controllers/hostedcluster/metrics/metrics.go
```diff
@@ -173,7 +176,11 @@ var (
 
 	etcdManualInterventionRequiredMetricDesc = prometheus.NewDesc(
 		EtcdManualInterventionRequiredMetricName, etcdManualInterventionRequiredMetricHelp,
-		append(hclusterLabels, "rosa_environment", "rosa_id"), nil)
+		append(hclusterLabels, "environment", "internal_id"), nil)
```
enxebre, line 179: how does this change relates to the PR?

enxebre: I see is for consistency below

enxebre: Would this break any consumption? cc @typeid

hypershift-operator/controllers/hostedcluster/metrics/metrics.go
```diff
@@ -456,30 +463,41 @@ func (c *hostedClustersMetricsCollector) Collect(ch chan<- prometheus.Metric) {
...
 							etcdManualInterventionRequiredMetricDesc,
 							prometheus.GaugeValue,
 							etcdManualInterventionRequiredValue,
-							append(hclusterLabelValues, metricLabels["rosa_environment"], metricLabels["rosa_id"])...,
+							append(hclusterLabelValues, metricLabels["environment"], metricLabels["internal_id"])...,
+						)
+
+					}
+
+					if sizeOverride := hcluster.Annotations[hyperv1beta1.ClusterSizeOverrideAnnotation]; sizeOverride != "" {
```
enxebre, line 495: what happens if the annotation is removed?

hypershift-operator/controllers/hostedcluster/metrics/metrics_test.go
```diff
@@ -774,10 +774,10 @@ func TestReportEtcdManualInterventionRequired(t *testing.T) {
 			Metric: []*dto.Metric{{
 				Label: []*dto.LabelPair{
 					{Name: ptr.To("_id"), Value: ptr.To("id")},
+					{Name: ptr.To("environment"), Value: ptr.To("")},
+					{Name: ptr.To("internal_id"), Value: ptr.To("")},
 					{Name: ptr.To("name"), Value: ptr.To("hc")},
```
enxebre, line 779: Please include a Test* for this

hypershift-operator/controllers/hostedcluster/metrics/metrics.go
typeid: This would not break anything on our side. Looks fine to me.

hypershift-operator/controllers/hostedcluster/metrics/metrics.go
devguyio: fixed

hypershift-operator/controllers/hostedcluster/metrics/metrics.go
devguyio: the metric won't be exposed anymore. So the time series for this HC will stop at the last time the metric showed a value.

hypershift-operator/controllers/hostedcluster/metrics/metrics_test.go
devguyio: Done

PR #4826, opened 2024-09-28
NO-JIRA: Fix recover-etcd cmd marking a missing flag required

devguyio:

The `recover-etcd` command was setting up a flag that doesn't exist as required. This PR cleans up that and fixes the require flags handling logic.

There were two main problems that were hidden by the swallowed errors:
1. The `etcd-endpoints` flag didn't exist
Here's a test when I changed the code to not swallow the error 

```shell
./bin/hypershift-operator recover-etcd --help
error occurred no such flag -etcd-endpoints
```
2. The `namespace` flag was set using the wrong method. It is a `PersistedFlag` so must be set using `MarkPersistentFlagRequired`  instead of `MarkFlagRequired`. 

```shell
./bin/hypershift-operator recover-etcd --help
error setting up namespace flag as required: no such flag -namespace
```

**Checklist**

### DIFF
--- a/etcd-recovery/etcdrecovery.go
+++ b/etcd-recovery/etcdrecovery.go
@@ -66,8 +66,12 @@ func NewRecoveryCommand() *cobra.Command {
 	cmd.PersistentFlags().StringVar(&opts.etcdClientKeyFile, "etcd-client-key", opts.etcdClientKeyFile, "etcd client cert key file.")
 	cmd.PersistentFlags().StringVar(&opts.etcdCAFile, "etcd-ca-cert", opts.etcdCAFile, "etcd trusted CA cert file.")
 	cmd.PersistentFlags().StringVar(&opts.namespace, "namespace", "", "namespace of etcd cluster")
-	_ = cmd.MarkFlagRequired("etcd-endpoints")
-	_ = cmd.MarkFlagRequired("namespace")
+
+	err := cmd.MarkPersistentFlagRequired("namespace")
+	if err != nil {
+		cmd.PrintErrf("error setting up namespace flag as required: %v\n", err)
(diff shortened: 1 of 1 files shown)


etcd-recovery/etcdrecovery.go
```diff
@@ -66,8 +66,12 @@ func NewRecoveryCommand() *cobra.Command {
 	cmd.PersistentFlags().StringVar(&opts.etcdClientKeyFile, "etcd-client-key", opts.etcdClientKeyFile, "etcd client cert key file.")
 	cmd.PersistentFlags().StringVar(&opts.etcdCAFile, "etcd-ca-cert", opts.etcdCAFile, "etcd trusted CA cert file.")
 	cmd.PersistentFlags().StringVar(&opts.namespace, "namespace", "", "namespace of etcd cluster")
-	_ = cmd.MarkFlagRequired("etcd-endpoints")
-	_ = cmd.MarkFlagRequired("namespace")
+
+	err := cmd.MarkPersistentFlagRequired("namespace")
```
jparrill, line 70: It's necessary to check the err and os.Exit? I mean, it will automatically do it exiting with `os.Exit(1)` and giving this error msg: `Error: required flag(s) "namespace" not set`

jparrill (commented):
small nit, otherwise lgtm

etcd-recovery/etcdrecovery.go
devguyio: Yea it's necessary as this err returned is more of a "misconfiguration" error that needs to be caught at runtime. Like marking a flag required that doesn't exist. So in our case this will happen if later one someone added a code that changed the `namespace` flag name without changing this line to the new name.

PR #4832, opened 2024-10-01
Enforce privileged PSA by default

Evan-Reilly:
OpenShiftPodSecurityAdmission feature gate is not enabled by default in 4.17 any longer: https://github.com/openshift/api/blob/release-4.17/features/features.go, https://github.com/openshift/api/pull/2018. Therefore, to ensure namespace security is not enforcing restricted by default any longer, this PR is required to properly enforce privileged PSA by default.

Fixes #

**Checklist**


rtheis (changes requested):
Please move this to release-4.17 only.

PR #4836, opened 2024-10-01
NO-JIRA: e2e: skip ValidKubeVirtInfraNetworkMTU condition check on 4.14 and ea…

sjenning:
4.14 kubevirt presubs are failing with
```
eventually.go:220: observed *v1beta1.HostedCluster e2e-clusters-qlpcv/example-5sdbz invalid at RV 90020 after 10m0s: missing condition: wanted ValidKubeVirtInfraNetworkMTU=True, did not find condition of this type
```
This condition was added in 4.15 and shouldn't be checked on earlier releases.


davidvossel (commented):
/lgtm
/approve

PR #4838, opened 2024-10-01
HOSTEDCP-1802: Refactor hc reconcile

devguyio:

Refactor the main `reconcile` method of the `HostedClusterReconciler`. This is the first of multiple PRs to refactor the HostedClusterReconciler. 

**Motive behind this PR**
- The `reconcile` function is ~1500 loc long, which means that it has a huge mental model that's very difficult to grasp.
- This makes it a brittle function and sensitive to modifications.

> [!NOTE]
> This PR is better reviewed by commits, as the changes are broken down into separate commits, each represents a step in the refactoring process.

> [!NOTE]
> There will be more PRs that break down the methods and dose more refactoring, but this is the first step.

**What's next?**
- Refactor the `reconcileWorld` method
- Introduce higher level reconciler abstraction.

**Checklist**


hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -1828,6 +1744,100 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
 	return result, nil
 }
 
+// reconcileMandatoryConditions sets all conditions that must be set even if the HC is being deleted.
+func (r *HostedClusterReconciler) reconcileMandatoryConditions(ctx context.Context, hcluster *hyperv1.HostedCluster,
```
enxebre: can we have a unit test for this?

devguyio:
Can be revisited later if necessary
/close

PR #4839, opened 2024-10-02
HOSTEDCP-1801: [CPO Refactor] Introduced ControlPlaneComponent CRD

muraee:

- An instance of this resource will be created for each ControlPlane component to report its status, similar to OC cluster operators.
- Add functionality for components to define their dependencies.

Fixes #

**Checklist**


api/hypershift/v1beta1/controlplanecomponent_types.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+// ControlPlaneComponentStatus defines the observed state of ControlPlaneComponent
+type ControlPlaneComponentStatus struct {
+	// Resources is a list of the resources reconciled by this component.
+	// +optional
+	Resources []ComponentResource `json:"resources,omitempty"`
+
+	// Version reports the current version of this component.
+	Version string `json:"version,omitempty"`
```
csrwng, line 47: I'm wondering if it makes sense to report multiple versions here. For example, for kube-apiserver clusteroperator on a standalone cluster, I get:

```
  versions:
  - name: raw-internal
    version: 4.18.0-0.nightly-2024-10-01-062804
  - name: kube-apiserver
    version: 1.30.3
  - name: operator
    version: 4.18.0-0.nightly-2024-10-01-062804
```

muraee: I think we are more interested in the version of the components we reconcile/deploy than the operands they deploy in our case, right? imo I think having multiple versions reported might be confusing.

api/hypershift/v1beta1/controlplanecomponent_types.go
```diff
@@ -0,0 +1,83 @@
...
+	WaitingForDependenciesReason string = "WaitingForDependencies"
+	// ReconciliationErrorReason indicates that there was an error during the reconcilation of the ControlPlaneComponent.
+	ReconciliationErrorReason string = "ReconciliationError"
+)
+
+// ControlPlaneComponentSpec defines the desired state of ControlPlaneComponent
+type ControlPlaneComponentSpec struct {
+}
+
+type ComponentResource struct {
```
csrwng, line 32: needs godoc

api/hypershift/v1beta1/controlplanecomponent_types.go
```diff
@@ -0,0 +1,83 @@
...
+	Kind string `json:"kind"`
```
csrwng, line 33: godoc

for consistency, we should indicate that these are required if they're not optional
and if they are immutable, then also // +immutable

api/hypershift/v1beta1/controlplanecomponent_types.go
csrwng: If this is optional, we should have +optional in the godoc

support/controlplane-component/status.go
```diff
@@ -0,0 +1,167 @@
...
+		return nil, err
+	}
+
+	unavailableDependencies := sets.New(c.dependencies...)
+	for _, component := range componentsList.Items {
+		if !unavailableDependencies.Has(component.Name) {
+			continue
+		}
+
+		availableCondition := meta.FindStatusCondition(component.Status.Conditions, string(hyperv1.ControlPlaneComponentAvailable))
```
csrwng, line 37: I don't think it's enough to check that the dependency is available. It should also match the desired version. That way reconcile during an upgrade can occur in a predetermined order. First etcd, then kas, etc. All of them will be available, but not necessarily at the desired version.

support/controlplane-component/status.go
```diff
@@ -0,0 +1,167 @@
...
+		})
+
+		return nil
+	}); err != nil {
+		return err
+	}
+
+	if len(unavailableDependencies) > 0 {
+		meta.SetStatusCondition(&component.Status.Conditions, metav1.Condition{
+			Type:    string(hyperv1.ControlPlaneComponentAvailable),
```
csrwng, line 72: maybe we should use a different condition here (unless we only want to use dependencies for initial deploy). During an upgrade, a component can be available but still be waiting for dependencies. It might be useful to mimic the ocp clusteroperator conditions (available, progressing, degraded). It is also possible for the component to be available and to have failed reconciliation.

support/controlplane-component/status.go
```diff
@@ -0,0 +1,167 @@
...
+			reason = "Error"
+			message = err.Error()
+		} else {
+			reason = hyperv1.NotFoundReason
+			message = fmt.Sprintf("%s Deployment not found", deployment.Name)
+		}
+		return
+	}
+
+	for _, cond := range deployment.Status.Conditions {
```
csrwng, line 124: we already have a function for this
https://github.com/openshift/hypershift/blob/a0191dbda4ac75bd8ee19869d9a952aa508b3f2b/support/util/deployment.go#L15

support/controlplane-component/status.go
```diff
@@ -0,0 +1,167 @@
...
+			reason = "Error"
+			message = err.Error()
+		} else {
+			reason = hyperv1.NotFoundReason
+			message = fmt.Sprintf("%s StatefulSet not found", statefulSet.Name)
+		}
+		return
+	}
+
+	if statefulSet.Status.ReadyReplicas >= ptr.Deref(statefulSet.Spec.Replicas, 0) {
```
csrwng, line 158: https://github.com/openshift/hypershift/blob/a0191dbda4ac75bd8ee19869d9a952aa508b3f2b/support/util/deployment.go#L28

api/hypershift/v1beta1/controlplanecomponent_types.go
csrwng: If not optional, then it should not have omitempty

csrwng (commented):
Some comments, thx

support/controlplane-component/status.go
muraee: makes sense.

support/controlplane-component/status.go
muraee: updated

support/controlplane-component/status.go
muraee: I'd rather use the Deployment available condition here, so we stay consistent with however k8s defines a deployment to be ready, and also to use the condition message when its false.

support/controlplane-component/status.go
muraee: the way I see it, is that a component's functionality requires its dependency to be available, so during an upgrade, a dependency upgrading means its not available and therefore the dependent component functionality is hindered so we mark it as unavailable (although technically its deployment is available).

e.g. if etcd is upgrading, then KAS can't be operational during that time and we mark it as unavailable until etcd finishes the upgrade and kas can start upgrading.

support/controlplane-component/status.go
csrwng: My understanding, the difference between ours and the upstream kube is that we take into consideration whether the latest spec is available, not just whether the deployment is available.

PR #4840, opened 2024-10-02
OCPBUGS-42579: Add network policies for konnectivity server and ignition server proxy

davidvossel:
This fixes two issues that impact HCP with the NodePort publishing strategy.

1. The konnectivity server component has moved in and out of the kas pod throughout several releases. The result is, some releases have network policies that allow connections to the konnectivity server, and others do not. To account for this, the hypershift operator needs to support both when the konnectivity server is in the kas pod and when it is outside of the pod.

This is resolved by adding a new network policy that allows access to the konnectivity server when it is hosted within the kas pod (which is how this is currently deployed in 4.17)

2. Nodes currently can't access the ignition proxy to retrieve their configs. This is resolved by adding a network policy for the ignition proxy.

### DIFF
--- a/hypershift-operator/controllers/hostedcluster/network_policies.go
+++ b/hypershift-operator/controllers/hostedcluster/network_policies.go
@@ -151,6 +151,13 @@ func (r *HostedClusterReconciler) reconcileNetworkPolicies(ctx context.Context, 
 				}); err != nil {
 					return fmt.Errorf("failed to reconcile ignition nodeport network policy: %w", err)
 				}
+				// Reconcile nodeport-ignition-proxy Network Policy
+				policy = networkpolicy.NodePortIgnitionProxyNetworkPolicy(controlPlaneNamespaceName)
+				if _, err := createOrUpdate(ctx, r.Client, policy, func() error {
+					return reconcileNodePortIgnitionProxyNetworkPolicy(policy, hcluster)
+				}); err != nil {
+					return fmt.Errorf("failed to reconcile ignition proxy nodeport network policy: %w", err)
+				}
 			}
 		case hyperv1.Konnectivity:
 			if svc.ServicePublishingStrategy.Type == hyperv1.NodePort {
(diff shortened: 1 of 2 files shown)


hypershift-operator/controllers/hostedcluster/network_policies.go
```diff
@@ -151,6 +151,13 @@ func (r *HostedClusterReconciler) reconcileNetworkPolicies(ctx context.Context,
 				}); err != nil {
 					return fmt.Errorf("failed to reconcile ignition nodeport network policy: %w", err)
 				}
+				// Reconcile nodeport-ignition-procy Network Policy
```
jparrill, line 154: nit: nodeport-ignition-procy

hypershift-operator/controllers/hostedcluster/network_policies.go
```diff
@@ -161,6 +168,15 @@ func (r *HostedClusterReconciler) reconcileNetworkPolicies(ctx context.Context,
 				}); err != nil {
 					return fmt.Errorf("failed to reconcile konnectivity nodeport network policy: %w", err)
 				}
+
+				// Reconcile nodeport-konnectivity Network Policy when konnectivity is hosted in the kas pod
+				policy = networkpolicy.NodePortKonnectivityKASNetworkPolicy(controlPlaneNamespaceName)
+				if _, err := createOrUpdate(ctx, r.Client, policy, func() error {
```
jparrill, line 174: I think this will create the policy always, we need to identify where is konnectivity and then place the netpol accordingly.

jparrill (commented):
Hey David, thanks for the PR. It's there any chance to add a unitTest around the netpol reconciliation to identify the policy is placed properly when is in or out of the KAS?

hypershift-operator/controllers/hostedcluster/network_policies.go
davidvossel: > we need to identify where is konnectivity and then place the netpol accordingly.

how? the hypershift operator isn't responsible for laying down the konnectivity agent.

davidvossel (commented):
> It's there any chance to add a unitTest around the netpol reconciliation to identify the policy is placed properly when is in or out of the KAS?

i don't believe the hypershift operator has knowledge of where the konnectivity server is placed.

davidvossel:
/cherry-pick release-4.17 release-4.16 release-4.15 release-4.14

PR #4823, opened 2024-09-27
HOSTEDCP-1976: Nodepool controller cleanup

enxebre:
**What this PR does / why we need it**:
- Move platform specific logic into their own files
- Rearrange validations

**Which issue(s) this PR fixes** *(optional, use `fixes #<issue_number>(, fixes #<issue_number>, ...)` format, where issue_number might be a GitHub issue, or a Jira story*:
Fixes #

**Checklist**
- [ ] Subject and description added to both, commit and PR.
- [ ] Relevant issues have been referenced.
- [ ] This change includes docs. 
- [ ] This change includes unit tests.

PR #4795, opened 2024-09-25
HOSTEDCP-1984: Refactor capi logic out from NodePool controller

enxebre:
This moves the CAPI related logic into their own file and add test coverage particularly for the reconcile function. Additional refactor of the business logic itself is left out intentionally for now to contain the scope of the refactor and avoid backward compatibility issues.
```
➜  hypershift git:(nodepool-refactor-capi) ✗ go tool cover -func=coverage.out | grep capi
github.com/openshift/hypershift/hypershift-operator/controllers/nodepool/capi.go:45:			newCAPI						100.0%
github.com/openshift/hypershift/hypershift-operator/controllers/nodepool/capi.go:60:			Reconcile					65.2%
github.com/openshift/hypershift/hypershift-operator/controllers/nodepool/capi.go:175:			cleanupMachineTemplates				78.1%
github.com/openshift/hypershift/hypershift-operator/controllers/nodepool/capi.go:235:			deleteMachineDeployment				30.8%
github.com/openshift/hypershift/hypershift-operator/controllers/nodepool/capi.go:256:			pauseMachineDeployment				70.0%
github.com/openshift/hypershift/hypershift-operator/controllers/nodepool/capi.go:274:			deleteMachineSet				30.8%
github.com/openshift/hypershift/hypershift-operator/controllers/nodepool/capi.go:295:			Pause						60.0%
github.com/openshift/hypershift/hypershift-operator/controllers/nodepool/capi.go:309:			pauseMachineSet					70.0%
github.com/openshift/hypershift/hypershift-operator/controllers/nodepool/capi.go:329:			deleteMachineHealthCheck			30.8%
github.com/openshift/hypershift/hypershift-operator/controllers/nodepool/capi.go:350:			reconcileMachineDeployment			72.2%
github.com/openshift/hypershift/hypershift-operator/controllers/nodepool/capi.go:571:			taintsToJSON					75.0%
github.com/openshift/hypershift/hypershift-operator/controllers/nodepool/capi.go:580:			reconcileMachineHealthCheck			96.4%
github.com/openshift/hypershift/hypershift-operator/controllers/nodepool/capi.go:662:			setMachineDeploymentReplicas			100.0%
github.com/openshift/hypershift/hypershift-operator/controllers/nodepool/capi.go:695:			machineTemplateBuilders				26.3%
github.com/openshift/hypershift/hypershift-operator/controllers/nodepool/capi.go:852:			generateMachineTemplateName			100.0%
[...]

PR #4847, opened 2024-10-03
NO-JIRA: Fixes kubevirt image cacher

davidvossel:
This fixes a regression to the kubevirt provider that was introduced sept 26th by some nodepool refactoring PRs, https://github.com/openshift/hypershift/pull/4795 and https://github.com/openshift/hypershift/pull/4823


davidvossel:
/hold

I need to validate this with the extended kubevirt tests to make sure the issue is resolved.

PR #4848, opened 2024-10-03
OSASINFRA-3623: openstack: Only upload required cloud credentials

stephenfin:

A `clouds.yaml` can contain multiple "clouds" or sets of credentials. If you provide a `clouds.yaml` with more than one entry, we should only use the relevant entry in our cloud credentials secret. Start doing this by building our own `clouds.yaml` with only a single entry. This also allows us to start doing other smart (TM) things, like using the `cacert` value in a cloud rather than insisting users provide this manually.

### DIFF
--- a/cmd/cluster/openstack/create.go
+++ b/cmd/cluster/openstack/create.go
@@ -20,6 +20,10 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
+const (
+	credentialCloudName = "openstack"
+)
+
 func DefaultOptions() *RawCreateOptions {
 	return &RawCreateOptions{NodePoolOpts: openstacknodepool.DefaultOptions()}
 }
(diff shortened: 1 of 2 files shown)


EmilienM:
/test e2e-openstack
to ensure no regression.

@stephenfin I guess you manually tested it so I'm fine with your change.

Thanks!

/lgtm

cmd/cluster/openstack/create_test.go
```diff
@@ -130,3 +131,178 @@ func TestCreateCluster(t *testing.T) {
 		})
 	}
 }
+
+func TestExtractCloud(t *testing.T) {
```
jparrill, line 135: It's possible here to use a test struct which could cover multiple cases? 
This is an example: https://github.com/openshift/hypershift/blob/main/api/util/configrefs/refs_test.go#L36

- This will make the readability better
- Reduce duplicated code
- Better organization.

cmd/cluster/openstack/create_test.go
```diff
@@ -130,3 +131,178 @@ func TestCreateCluster(t *testing.T) {
...
+	dumpYAML := func(t *testing.T, path string, contents map[string]any) {
+		data, err := yaml.Marshal(contents)
+		if err != nil {
+			t.Fatalf("failed to marshal YAML: %v", err)
```
jparrill, line 139: It's preferred to use `t.Errorf` instead of `t.Fatalf` in order to avoid the test termination before see all the results.

jparrill (commented):
Dropped some more comments

cmd/cluster/openstack/create.go
```diff
@@ -282,3 +269,56 @@ func findOpenStackCredentialsFile() (string, error) {
 
 	return "", nil
 }
+
+// extractCloud extracts the relevant cloud from a provided clouds.yaml and return a new clouds.yaml
+// with only that cloud in it and using a well-known cloud name
+func extractCloud(cloudsYAMLPath, caCertPath, cloudName string) ([]byte, []byte, error) {
```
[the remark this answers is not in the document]
stephenfin: Done

cmd/cluster/openstack/create_test.go
stephenfin: I explicitly avoided using a table-driven test here since there wasn't enough overlap between test cases. The resulting test case was full of "single-use" conditional branches that ultimately made the test case much more difficult to read (readability, organization) without a corresponding decrease in LOC (duplication).

cmd/cluster/openstack/create_test.go
stephenfin: I think `t.Fatalf` is expected here: an error here would indicate that we failed to marshall/write our test data to disk. If that happens, continuing would be misleading, since it could suggest an issue in the code-under-test rather than the test environment.

I used `t.Error` where appropriate in the test cases since in those cases I do want to see all results, as you note.

stephenfin:
@jparrill Anything else you'd like to see here? Are you okay with my responses?

cmd/cluster/openstack/create.go
```diff
@@ -30,6 +34,7 @@ func BindOptions(opts *RawCreateOptions, flags *pflag.FlagSet) {
 }
 
 func bindCoreOptions(opts *RawCreateOptions, flags *pflag.FlagSet) {
+	// TODO(stephenfin): This is unnecessary given the information should already be in clouds.yaml. We should deprecate and remove it.
```
EmilienM, line 37: :+1: for UX, much easier. Let's do that asap after this PR has landed.

EmilienM:
/cherry-pick release-4.17
to make it easy in our CI when we remove the deprecated CA flag.

PR #4852, opened 2024-10-04
OCPBUGS-38132: add ValidIDPConfiguration condition to report IDP config issues

sjenning:
Right now, IDP config/reachability issues are fatal to the CPO reconcile loop and lead to all manner of weird issues due to partial reconciliation of the HCP.

This makes the error non-fatal to the reconcile loop and reports any IDP issues on a new HC condition `ValidIDPConfiguration`

cc @csrwng

### DIFF
--- a/api/hypershift/v1beta1/hostedcluster_conditions.go
+++ b/api/hypershift/v1beta1/hostedcluster_conditions.go
@@ -101,6 +101,11 @@ const (
 	// A failure here may require external user intervention to resolve. E.g. oidc was deleted out of band.
 	ValidOIDCConfiguration ConditionType = "ValidOIDCConfiguration"
 
+	// ValidIDPConfiguration indicates if the Identity Provider configuration is valid.
+	// A failure here may require external user intervention to resolve
+	// e.g. the user-provided IDP configuration provided is invalid or the IDP is not reachable.
+	ValidIDPConfiguration ConditionType = "ValidIDPConfiguration"
+
 	// ValidReleaseImage indicates if the release image set in the spec is valid
 	// for the HostedCluster. For example, this can be set false if the
 	// HostedCluster itself attempts an unsupported version before 4.9 or an
--- a/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
@@ -3435,6 +3435,31 @@ func (r *HostedControlPlaneReconciler) reconcileOAuthServer(ctx context.Context,
 	}); err != nil {
 		return fmt.Errorf("failed to reconcile oauth deployment: %w", err)
 	}
+
+	// Report any IDP configuration errors as a condition on the HCP
+	new := metav1.Condition{
+		Type:    string(hyperv1.ValidIDPConfiguration),
+		Status:  metav1.ConditionTrue,
+		Reason:  "IDPConfigurationValid",
+		Message: "Identity provider configuration is valid",
+	}
+	if _, _, err := oauth.ConvertIdentityProviders(ctx, p.IdentityProviders(), p.OauthConfigOverrides, r, hcp.Namespace); err != nil {
+		// Report the error in a condition on the HCP
+		r.Log.Error(err, "failed to initialize identity providers")
+		new = metav1.Condition{
+			Type:    string(hyperv1.ValidIDPConfiguration),
+			Status:  metav1.ConditionFalse,
+			Reason:  "IDPConfigurationError",
+			Message: fmt.Sprintf("failed to initialize identity providers: %v", err),
+		}
+	}
+	// Update the condition on the HCP if it has changed
+	if meta.SetStatusCondition(&hcp.Status.Conditions, new) {
+		if err := r.Status().Update(ctx, hcp); err != nil {
+			return fmt.Errorf("failed to update valid IDP configuration condition: %w", err)
+		}
+	}
+
 	return nil
 }
 
--- a/control-plane-operator/controllers/hostedcontrolplane/oauth/config.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/oauth/config.go
@@ -52,9 +52,11 @@ func ReconcileOAuthServerConfig(ctx context.Context, cm *corev1.ConfigMap, owner
 
 func generateOAuthConfig(ctx context.Context, client crclient.Client, namespace string, params *OAuthConfigParams) (*osinv1.OsinServerConfig, error) {
 	var identityProviders []osinv1.IdentityProvider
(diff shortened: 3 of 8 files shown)


control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -3415,6 +3416,33 @@ func (r *HostedControlPlaneReconciler) reconcileOAuthServer(ctx context.Context,
...
+		// Report the error in a condition on the HCP
+		new = &metav1.Condition{
+			Type:    string(hyperv1.ValidIDPConfiguration),
+			Status:  metav1.ConditionFalse,
+			Reason:  "IDPConfigurationError",
+			Message: fmt.Sprintf("failed to initialize identity providers: %v", err),
+		}
+	}
+	// Update the condition on the HCP if it has changed
+	if !reflect.DeepEqual(orig, new) {
```
csrwng, line 3438: This would always be false because orig will have a LastTransitionTime, no?

csrwng: Also, could we get in a hot loop if the error returned is slightly different every time?

sjenning: > This would always be false because orig will have a LastTransitionTime, no?

yes, I do need to clear that before the comparison

> Also, could we get in a hot loop if the error returned is slightly different every time?

I thought about this.  We could just update when the Status changes.  However, the user moves from one invalid config to another and the error changes, we wouldn't be reporting the new error.  :thinking:

api/hypershift/v1beta1/hostedcluster_conditions.go
```diff
@@ -101,6 +101,11 @@ const (
 	// A failure here may require external user intervention to resolve. E.g. oidc was deleted out of band.
 	ValidOIDCConfiguration ConditionType = "ValidOIDCConfiguration"
 
+	// ValidIDPConfiguration indicates if the Identity Provider configuration is valid.
```
enxebre, line 104: Please include this in the list validated by validateHostedClusterConditions e2e

control-plane-operator/controllers/hostedcontrolplane/oauth/config.go
```diff
@@ -52,9 +52,11 @@ func ReconcileOAuthServerConfig(ctx context.Context, cm *corev1.ConfigMap, owner
 
 func generateOAuthConfig(ctx context.Context, client crclient.Client, namespace string, params *OAuthConfigParams) (*osinv1.OsinServerConfig, error) {
 	var identityProviders []osinv1.IdentityProvider
-	identityProviders, _, err := convertIdentityProviders(ctx, params.IdentityProviders, params.OauthConfigOverrides, client, namespace)
+	identityProviders, _, err := ConvertIdentityProviders(ctx, params.IdentityProviders, params.OauthConfigOverrides, client, namespace)
 	if err != nil {
-		return nil, err
+		// Eat the error here, since we don't want to fail the deployment if the identity providers are invalid
```
Patryk-Stefanski, line 57: It would be nice if we also had a trace of this error in the logs when debugging, could we log it as a warning ?

control-plane-operator/controllers/hostedcontrolplane/oauth/deployment.go
```diff
@@ -164,12 +164,13 @@ func ReconcileDeployment(ctx context.Context, client client.Client, deployment *
 
 	deploymentConfig.ApplyTo(deployment)
 	if len(identityProviders) > 0 {
-		_, volumeMountInfo, err := convertIdentityProviders(ctx, identityProviders, providerOverrides, client, deployment.Namespace)
-		if err != nil {
-			return err
+		_, volumeMountInfo, err := ConvertIdentityProviders(ctx, identityProviders, providerOverrides, client, deployment.Namespace)
+		// Eat the error here, since we don't want to fail the deployment if the identity providers are invalid
```
Patryk-Stefanski, line 168: It would be nice if we also had a trace of this error in the logs when debugging, could we log it as a warning ?

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -3414,6 +3415,35 @@ func (r *HostedControlPlaneReconciler) reconcileOAuthServer(ctx context.Context,
 	}); err != nil {
 		return fmt.Errorf("failed to reconcile oauth deployment: %w", err)
 	}
+
+	// Report any IDP configuration errors as a condition on the HCP
+	orig := meta.FindStatusCondition(hcp.Status.Conditions, string(hyperv1.ValidIDPConfiguration))
+	// zero out LastTransitionTime so that we can compare the condition without it
+	orig.LastTransitionTime = metav1.Time{}
```
Patryk-Stefanski, line 3422: Will this throw a nil pointer if the status isn't found ?

Patryk-Stefanski, line 3422: Will this cause a nil pointer if the status doesn't yet exist?

control-plane-operator/controllers/hostedcontrolplane/oauth/config.go
sjenning: This would be nice but we don't have access to the `logger` for the `HostedControlPlaneReconciler` here and piping in new arguments will likely created backporting pains.

sjenning: I logged it where we eval the condition so that it still appears in the CPO logs :+1:

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
muraee: `meta.SetStatusCondition()` retrieves the existing condition, compares it with the new condition and returns a bool if the condition changed, you shouldn't need to retrieve orig condition to compare manaully.
```golang
changed := meta.SetStatusCondition(&hcp.Status.Conditions, new)
if changed {
    if err := r.Status().Update(ctx, hcp); err != nil {
	return fmt.Errorf("failed to update valid IDP configuration condition: %w", err)
    }
}
```

api/hypershift/v1beta1/hostedcluster_conditions.go
sjenning: I will add this in a follow on.  If I add it now, the upgrade test fails as we check the conditions on the n-1 release before we upgrade to the release under test.

sjenning:
/test e2e-aks

after https://github.com/openshift/release/pull/57823 merged

PR #4856, opened 2024-10-07
NO-JIRA: switch to using ptr instead of pointer

Patryk-Stefanski:
**What this PR does / why we need it**: use ptr instead of pointer as it has been deprecated

Fixes #

**Checklist**

### DIFF
--- a/cmd/install/assets/hypershift_operator.go
+++ b/cmd/install/assets/hypershift_operator.go
@@ -23,7 +23,6 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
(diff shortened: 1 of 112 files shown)


cmd/install/assets/hypershift_operator.go
```diff
@@ -633,7 +632,7 @@ func (o HyperShiftOperatorDeployment) Build() *appsv1.Deployment {
 							Command:         []string{"/usr/bin/hypershift-operator"},
 							Args:            []string{"init"},
 							SecurityContext: &corev1.SecurityContext{
-								RunAsUser:              k8sutilspointer.Int64(1000),
+								RunAsUser:              ptr.To(int64(1000)),
```
muraee, line 635: please use generics instead of casting
```suggestion
								RunAsUser:              ptr.To[int64](1000),
```
same comment for the rest

PR #4857, opened 2024-10-07
OCPBUGS-35878: No PDBs creation on ControllerAvailability SingleReplica policy

jparrill:
Fixes #[OCPBUGS-35878](https://issues.redhat.com/browse/OCPBUGS-35878)

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
@@ -2775,13 +2775,15 @@ func (r *HostedControlPlaneReconciler) reconcileManagedEtcd(ctx context.Context,
 		r.Log.Info("reconciled etcd servicemonitor", "result", result)
 	}
 
-	pdb := manifests.EtcdPodDisruptionBudget(hcp.Namespace)
-	if result, err := createOrUpdate(ctx, r, pdb, func() error {
-		return etcd.ReconcilePodDisruptionBudget(pdb, p)
-	}); err != nil {
-		return fmt.Errorf("failed to reconcile etcd pdb: %w", err)
-	} else {
-		r.Log.Info("reconciled etcd pdb", "result", result)
+	if hcp.Spec.ControllerAvailabilityPolicy == hyperv1.HighlyAvailable {
+		pdb := manifests.EtcdPodDisruptionBudget(hcp.Namespace)
+		if result, err := createOrUpdate(ctx, r, pdb, func() error {
+			return etcd.ReconcilePodDisruptionBudget(pdb, p)
+		}); err != nil {
+			return fmt.Errorf("failed to reconcile etcd pdb: %w", err)
+		} else {
+			r.Log.Info("reconciled etcd pdb", "result", result)
+		}
 	}
 
 	// reconcile etcd-defrag-operator serviceAccount, role and roleBinding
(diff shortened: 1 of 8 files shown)


control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -4234,12 +4244,14 @@ func (r *HostedControlPlaneReconciler) reconcileRouter(ctx context.Context, hcp
 			return fmt.Errorf("failed to reconcile router deployment: %w", err)
 		}
 
-		pdb := manifests.RouterPodDisruptionBudget(hcp.Namespace)
-		if _, err := createOrUpdate(ctx, r.Client, pdb, func() error {
-			ingress.ReconcileRouterPodDisruptionBudget(pdb, hcp.Spec.ControllerAvailabilityPolicy, config.OwnerRefFrom(hcp))
-			return nil
-		}); err != nil {
-			return fmt.Errorf("failed to reconcile router pod disruption budget: %w", err)
+		if hcp.Spec.ControllerAvailabilityPolicy == hyperv1.HighlyAvailable {
```
enxebre, line 4247: All the Reconcile*PodDisruptionBudget funcs already receive `hcp.Spec.ControllerAvailabilityPolicy` as parameter and call `util.ReconcilePodDisruptionBudget` underneath, so it doesn't make sense to wrap them within `hcp.Spec.ControllerAvailabilityPolicy`. The implementation of util.ReconcilePodDisruptionBudget dictates the output and it can be unit tested

jparrill:
> /hold [#4857 (comment)](https://github.com/openshift/hypershift/pull/4857#discussion_r1791571080)
> 
> This pdbs prevent CI from going flake because of autoscaling relocating pods. If we get rid of the PDBs we need to prevent the pods from being drained by autoscaling in CI. To that end we could e.g. use the "cluster-autoscaler.kubernetes.io/safe-to-evict": "false" annotation can use PTAL @sjenning

Yep I already talked about this with Seth, this is still in WIP, we were thinking too in read the pod logs, check for leader election losses and also kas disruption issues. With that we can differentiate the real crashes from the not relevant ones.

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
jparrill: Ok, let me change this a little bit. I will remove the availability from the `util.ReconcilePodDisruptionBudget` function to return always the same configuration.

support/util/pdb.go
```diff
@@ -1,22 +1,12 @@
 package util
```
wking, line 1: Pulling [this comment][1] down into a thread, because I don't understand:

> This pdbs prevent CI from going flake because of autoscaling relocating pods. If we get rid of the PDBs we need to prevent the pods from being drained by autoscaling in CI. To that end we could e.g. use the "cluster-autoscaler.kubernetes.io/safe-to-evict": "false" annotation

and I want to unpack that, but I don't want to clutter up the main comment stream, and we can mark this thread resolved once discussion wraps up.

If these pods are `SingleReplica` instead of `HighlyAvailable`, isn't it expected that they go unavailable while being shifted around due to autoscaling or anything else that wants to shuffle workload Pods around for whatever reason?  `cluster-autoscaler.kubernetes.io/safe-to-evict: "false"` seems like it would achieve CI-stability with similar drain-blocking costs to the current `minAvailable: 1` PodDisruptionBudgets.  I'd expect us to:

* Drop the PDBs for `SingleReplica` (like this pull is currently doing in 4a82de8c4c),
* Remove any orphan PDBs that older versions of the HostedControlPlane controller had created for those `SingleReplica` clusters (e.g. so updating your HostedCluster from 4.17.0 to 4.18.hasThisFix wouldn't leave orphan PDBs lying around), and
* Improve any CI that assumes `SingleReplica` services will be highly available, so it can handle some discruption gracefully, without failing.  Maybe that's what [this comment][2] is imagining?

[1]: https://github.com/openshift/hypershift/pull/4857#issuecomment-2399402981
[2]: https://github.com/openshift/hypershift/pull/4857#issuecomment-2402202576

PR #4866, opened 2024-10-08
HOSTEDCP-2025: Allow setting the VPC CIDR when creating AWS infrastructure

csrwng:
This makes it possible to create VPC peering connections between different cluster infrastructures. It is a prerequisite for more extensive proxy testing.

Fixes #[HOSTEDCP-2025](https://issues.redhat.com/browse/HOSTEDCP-2025)

**Checklist**

### DIFF
--- a/cmd/cluster/aws/create.go
+++ b/cmd/cluster/aws/create.go
@@ -43,6 +43,7 @@ type RawCreateOptions struct {
 	ProxyVPCEndpointServiceName string
 	SingleNATGateway            bool
 	MultiArch                   bool
+	VPCCIDR                     string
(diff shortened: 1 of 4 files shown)


cmd/cluster/aws/create.go
```diff
@@ -181,6 +182,11 @@ func (o *ValidatedCreateOptions) Complete(ctx context.Context, opts *core.Create
 			*into = value
 		}
 	}
+
+	if err := awsutil.ValidateVPCCIDR(o.VPCCIDR); err != nil {
```
muraee, line 186: should this be called inside `validateAWSOptions()` instead?

csrwng: Makes sense, moved.

PR #4868, opened 2024-10-09
OCPBUGS-43087: openstack/e2e: re-work nodepool tests

EmilienM:
* When running the nodepool tests, the HostedCluster is created with a
  initial Nodepool and a replica == 0. That involves that the
  router-default won't exist and therefore we can't expect the Ingress to
  work. So let's skip the DNS record creation when there is no nodepool
  but show a warning because we can't expect the cluster to be healthy.

* Remove double get guestClient when creating Route54 records.

* nodepool: force replica to 1 for the default since we need one
  nodepool for Ingress to be working.

* Use an existing function to test the Nodepool replicas and the nodes
  numbers.

* skip NTO tests which can't work on the OpenStack platform at this
  time.

* nodepool/openstack: support AMD64 arch.

* nodepool/openstack: add missing watch in the reconciler.

### DIFF
--- a/hypershift-operator/controllers/nodepool/capi.go
+++ b/hypershift-operator/controllers/nodepool/capi.go
@@ -1281,6 +1281,11 @@ func (c *CAPI) listMachineTemplates() ([]client.Object, error) {
 		if err != nil {
 			return nil, err
 		}
+	case hyperv1.OpenStackPlatform:
+		gvk, err = apiutil.GVKForObject(&capiopenstack.OpenStackMachineTemplate{}, api.Scheme)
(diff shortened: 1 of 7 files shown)


EmilienM:
CI results: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_hypershift/4868/pull-ci-openshift-hypershift-main-e2e-openstack-nodepools/1844464816853356544

jparrill (commented):
lgtm

jparrill (commented):
/lgtm

PR #4870, opened 2024-10-09
HOSTEDCP-1773: Introduce feature gate mechanism.

enxebre:
Feature gates in OpenShift allows to ensure everything knew works together and optimizes for rapid evaluation in CI to promote without ever releasing a public preview.
There are no guarantees that your fleet continues to be operational long term after you enable a feature gate.

In the HCP context, there are multiple non exclusive scenarios where features might need to be gated:

1 - A feature that impacts HO install
E.g. New CRDs are required for CPOv2 (control plane operator version 2)

2 - A feature that impacts the whole cluster fleet / HO
E.g. Introduce fleet wise shared ingress to be validated in targeted environments

3 - A feature that impacts individual clusters
E.g. Introduce using CPOv2 for some HC to develop feedback

4 - A feature that impacts API
E.g. Introduce a new provider like Openstack
E.g. Introduce a new field/feature like AWS tenancy

5 - A feature specific for an OCP component
Components honour existing standalone in-cluster OCP feature gate mechanisim

The currently ongoing implementation of feature gates expose this in the CLI at install time
```
--feature-gates mapStringBool                    A set of key=value pairs that describe feature gates for alpha/experimental features. Options are:
                                                    AllAlpha=true|false (ALPHA - default=false)
                                                    AllBeta=true|false (BETA - default=false)
                                                    OpenStack=true
```

In a follow up we plan to introduce support to also signal a feature gate granularly at the HC level.
Eventually support for at least 1, 2, 3 and 4 aforementioned scenarios will most likely converge into a single API.

This PR uses 4 OpenStack platform as the driving use case.

Fixes #

**Checklist**


enxebre:
/hold
a bit

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -489,6 +489,13 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
 			}
 		}
 
+		// Use openshift markers https://github.com/openshift/api/blob/82e082220d910f489d35880174e2eb90e21f5589/example/v1/types_stable.go#L109-L111
+		// to generate different CRDs for different featureGates so validation would prevent the HC from being applied with a platfom that is not enabled.
+		if hcluster.Spec.Platform.Type == hyperv1.OpenStackPlatform && !featuregate.Gates.Enabled(featuregate.OpenStack) {
+			log.Info("OpenStack platform is not supported without enabling the featureg gate")
```
bryan-cox, line 495: typo
```suggestion
			log.Info("OpenStack platform is not supported without enabling the feature gate")
```

docs/content/how-to/feature-gates.md
bryan-cox, line 1: Should we add a reference to https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/ somewhere here?

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
enxebre: thanks!

docs/content/how-to/feature-gates.md
enxebre: Added a mention to k8s.io/component-base/featuregate

docs/content/how-to/feature-gates.md
```diff
@@ -0,0 +1,34 @@
...
+4 - A feature that impacts API
+E.g. Introduce a new provider like Openstack
+E.g. Introduce a new field/feature like AWS tenancy
+
+5 - A feature specific for an OCP component
+Components honour existing standalone in-cluster OCP feature gate mechanisim
+
+The currently ongoing implementation of feature gates expose this in the CLI at install time
+```
+--feature-gates mapStringBool                    A set of key=value pairs that describe feature gates for alpha/experimental features. Options are:
```
jparrill, line 26: Could you put a sample of how the Hosted Cluster should be created with CLI and a FeatureGate enabled?

EmilienM:
FYI I've done the CI part of it: https://github.com/openshift/release/pull/57849
So as soon as this one merges we can update the CI and limit OpenStack jobs downtime.

cc @mdbooth

enxebre:
/close
in favour https://github.com/openshift/hypershift/pull/4918

PR #4872, opened 2024-10-09
ARO-11084: Helm chart generator

geoberle:

Introduce a `install helm` command to generate a helm chart with some templatization options similar to the `install render --template` command.

the PR also introduces a way to template the registry overrides option of the operator

required by https://github.com/Azure/ARO-HCP/pull/698

Fixes #[ARO-11084](https://issues.redhat.com//browse/ARO-11084)

**Checklist**


cmd/install/install_helm.go
```diff
@@ -0,0 +1,153 @@
...
+}
+
+func NewHelmRenderCommand(opts *Options) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:          "helm",
+		Short:        "Generate a Helm chart for the HyperShift Operator",
+		SilenceUsage: true,
+	}
+
+	cmd.Flags().StringVar(&opts.OutputFile, "output-dir", "", "File to write the rendered manifests to. Writes to STDOUT if not specified.")
```
csrwng, line 43: nit: The description is not accurate

csrwng (commented):
Just one nit, otherwise, lgtm

cmd/install/install_helm.go
csrwng: also, since you are defaulting to `./chart`, it would make sense to set that as the default value of opts.OutputFile and pass it to StringVar

geoberle: both adressed. ty

csrwng:
/hold
@geoberle looks like you need to fix the unit tests

cmd/install/install_helm_test.go
```diff
@@ -0,0 +1,69 @@
...
+		if err != nil {
+			t.Fatal(err)
+		}
+		if len(files) == 0 {
+			t.Fatalf("%s directory is empty", dirPath)
+		}
+	}
+
+	// check if the Chart.yaml file exists
+	chartPath := tmpDir + "/Chart.yaml"
```
enxebre, line 58: Please let's add a test that validate the expected output for these files with fixtures, see https://github.com/openshift/hypershift/blob/main/cmd/cluster/aws/create_test.go#L227 for an example

cmd/install/install_render.go
```diff
@@ -25,23 +25,31 @@ const (
...
+	OIDCS3CredsSecretKey:        "OIDC_S3_CREDS_SECRET_KEY",
+	AWSPrivateRegion:            "AWS_PRIVATE_REGION",
+	AWSPrivateCredsSecret:       "AWS_PRIVATE_CREDS_SECRET",
+	AWSPrivateCredsSecretKey:    "AWS_PRIVATE_CREDS_SECRET_KEY",
+	ExternalDNSCredsSecret:      "EXTERNAL_DNS_CREDS_SECRET",
+	ExternalDNSDomainFilter:     "EXTERNAL_DNS_DOMAIN_FILTER",
+	ExternalDNSTxtOwnerID:       "EXTERNAL_DNS_TXT_OWNER_ID",
+	ExternalDNSImage:            "EXTERNAL_DNS_IMAGE",
+	RegistryOverrides:           "REGISTRY_OVERRIDES",
+	AROHCPKeyVaultUsersClientID: "AZURE_KEYVAULT_CLIENT_ID",
```
enxebre, line 46: are these changes related to helm or is this using the PR to update this command with newest fields? If so can we please split commits accordingly?

cmd/install/install_helm_test.go
geoberle: done

cmd/install/install_render.go
geoberle: ack. working on it

geoberle: i've split the work into two commits.
1) introduce helm generator while not changing anything on the output of the openshift template generator
2) introduce new fields

cmd/install/testdata/zz_fixture_TestHelmCommand_values.yaml
```diff
@@ -0,0 +1,17 @@
+aws:
```
enxebre, line 1: We'll want to have test cases to validate these fields are actually populated as you would expect. Otherwise the value of the test is quite limited. I'm ok if we want to do that in a follow up as we move forward.

PR #4878, opened 2024-10-09
HOSTEDCP-1994: Add Secrets Store CSI RBAC Policies for HO and CPO

bryan-cox:
This PR adds the secrets store CSI driver to the go.mod and adds the needed Secrets Store CSI RBAC policies to the HO and CPO for ARO HCP deployments.

Fixes #

**Checklist**


support/azureutil/azureutil.go
```diff
@@ -217,3 +218,8 @@ func VerifyResourceGroupLocationsMatch(ctx context.Context, hc *hyperv1.HostedCl
 
 	return nil
 }
+
+// IsAroHcp returns true if the managed service environment variable is set to ARO-HCP
+func IsAroHcp() bool {
```
csrwng, line 223: nit: s/Hcp/HCP/

cmd/install/assets/hypershift_operator.go
```diff
@@ -941,7 +941,7 @@ type HyperShiftOperatorClusterRole struct {
 	EnableCVOManagementClusterMetricsAccess bool
 }
 
-func (o HyperShiftOperatorClusterRole) Build() *rbacv1.ClusterRole {
+func (o HyperShiftOperatorClusterRole) Build(managedService string) *rbacv1.ClusterRole {
```
csrwng, line 944: This is not the pattern we've been using for the build functions. Generally any params are set as fields, the Build() has no args

csrwng (commented):
A couple of comments

cmd/install/assets/hypershift_operator.go
bryan-cox: Ahh I see now what you mean.

PR #4884, opened 2024-10-10
OCPBUGS-44964: Reconcile SecretProviderClass for Ingress on ARO HCP

bryan-cox:
Reconcile the SecretProviderClass for the ingress operator for ARO HCP deployments. The SecretProviderClass is used by the Secrets Store CSI driver to mount a certificate to a volume in the ingress pod deployment.

Fixes [OCPBUGS-44964](https://issues.redhat.com/browse/OCPBUGS-44964)

**Checklist**

### DIFF
--- a/cmd/infra/azure/create.go
+++ b/cmd/infra/azure/create.go
@@ -68,6 +68,7 @@ type CreateInfraOptions struct {
 	ManagedIdentityKeyVaultTenantID string
 	TechPreviewEnabled              bool
 	ManagedIdentitiesFile           string
+	DNSZoneRG                       string
 }
 
 type CreateInfraOutput struct {
@@ -229,7 +230,7 @@ func (o *CreateInfraOptions) Run(ctx context.Context, l logr.Logger) (*CreateInf
 		}
 
 		// Create ServicePrincipals with backing certificates
-		cmdStr := buildCreateServicePrincipalCommand(subscriptionID, resourceGroupName, nsgResourceGroupName, vnetResourceGroupName, ingress, o.InfraID, o.ManagedIdentityKeyVaultName)
+		cmdStr := buildCreateServicePrincipalCommand(subscriptionID, resourceGroupName, nsgResourceGroupName, vnetResourceGroupName, ingress, o.InfraID, o.ManagedIdentityKeyVaultName, o.DNSZoneRG)
 		clientID, err := createServicePrincipalWithCertificate(cmdStr)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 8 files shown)


bryan-cox:
Needs the following PRs before we can continue with this one:
- https://github.com/openshift/hypershift/pull/4884
- https://github.com/openshift/cluster-ingress-operator/pull/1151

support/config/constants.go
```diff
@@ -50,11 +50,38 @@ const (
...
-	DefaultMachineNetwork = "10.0.0.0/16"
+	ManagedAzureClientIdEnvVarKey        = "ARO_HCP_MI_CLIENT_ID"
+	ManagedAzureTenantIdEnvVarKey        = "ARO_HCP_TENANT_ID"
+	ManagedAzureCertificatePathEnvVarKey = "ARO_HCP_CLIENT_CERTIFICATE_PATH"
+	ManagedAzureCertificateMountPath     = "/mnt/certs"
+	ManagedAzureCertificatePath          = "/mnt/certs/"
+	ManagedAzureSecretsStoreCSIDriver    = "secrets-store.csi.k8s.io"
+	ManagedAzureSecretProviderClass      = "secretProviderClass"
+
+	ManagedAzureCPOSecretProviderClassName                = "managed-azure-cpo"
```
bryan-cox, line 72: Most of these are not used yet but will be. I wanted to include them here now to avoid rebase issues with the other related PRs.

cmd/infra/azure/create.go
```diff
@@ -387,12 +387,13 @@ func buildCreateServicePrincipalCommand(subscriptionID, managedResourceGroupName
 	managedRG := fmt.Sprintf("/subscriptions/%s/resourceGroups/%s", subscriptionID, managedResourceGroupName)
 	nsgRG := fmt.Sprintf("/subscriptions/%s/resourceGroups/%s", subscriptionID, nsgResourceGroupName)
 	vnetRG := fmt.Sprintf("/subscriptions/%s/resourceGroups/%s", subscriptionID, vnetResourceGroupName)
+	dnsZoneRG := fmt.Sprintf("/subscriptions/%s/resourceGroups/os4-common", subscriptionID)
```
enxebre, line 390: why is this resource group name hardcoded?

bryan-cox: Right, that should be a flag. This is just needed for CI/testing purposes. In a real world scenario, the permissions will already be setup prior to the HostedCluster creation.

bryan-cox:
/test e2e-aks

I tried out the command locally to add the ingress scope and that seems to be working fine.

PR #4886, opened 2024-10-10
NO-JIRA: cli: add feature-set flag to create cluster

sjenning:
In support of testing [SDN-5238](https://issues.redhat.com//browse/SDN-5238)

Avoid the need to do this https://github.com/openshift/release/pull/56065/files#diff-2dd93dce12c5420e483d70b5984b56fcf0d1fe4084d01901c6d5705e09cb619bR75-R92

### DIFF
--- a/cmd/cluster/core/create.go
+++ b/cmd/cluster/core/create.go
@@ -54,6 +54,7 @@ func DefaultOptions() *RawCreateOptions {
 		Arch:                           "amd64",
 		OLMCatalogPlacement:            hyperv1.ManagementOLMCatalogPlacement,
 		NetworkType:                    string(hyperv1.OVNKubernetes),
+		FeatureSet:                     string(configv1.Default),
 	}
 }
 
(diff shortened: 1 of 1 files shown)


cmd/cluster/core/create.go
```diff
@@ -622,6 +644,17 @@ func (opts *RawCreateOptions) Validate(ctx context.Context) (*ValidatedCreateOpt
 		return nil, fmt.Errorf("specified arch %q is not supported", opts.Arch)
 	}
 
+	// Validate feature set is "", TechPreviewNoUpgrade, or DevPreviewNoUpgrade
+	switch opts.FeatureSet {
+	case string(configv1.Default):
+	case string(configv1.TechPreviewNoUpgrade):
+	case string(configv1.DevPreviewNoUpgrade):
+	case string(configv1.CustomNoUpgrade):
+		return nil, fmt.Errorf("only predefined feature set are supported by the feature-set flag")
```
bryan-cox, line 653: nit
```suggestion
		return nil, fmt.Errorf("only a predefined feature set is supported by the feature-set flag")
```

cmd/cluster/core/create.go
bryan-cox, line 1: Can the "why" be added to the commit?

PR #4889, opened 2024-10-10
HOSTEDCP-2026: Add support for static control plane operator overrides

csrwng:
Adds a file where we can statically specify replacement CPO images for specific OCP releases. Enables e2e testing of this combination by setting previous and latest release images in e2e based on the presence of an environment variable.

Fixes #[HOSTEDCP-2026](https://issues.redhat.com/browse/HOSTEDCP-2026)

**Checklist**

### DIFF
--- a/hypershift-operator/controlplaneoperator-overrides/overrides.go
+++ b/hypershift-operator/controlplaneoperator-overrides/overrides.go
@@ -1,0 +1,75 @@
+package controlplaneoperatoroverrides
+
+import (
+	_ "embed"
+	"fmt"
+	"os"
+
+	"github.com/openshift/hypershift/support/config"
+	"gopkg.in/yaml.v2"
+)
+
+type CPOOverrides struct {
+	Overrides []CPOOverride           `yaml:"overrides,omitempty"`
+	Testing   CPOOverrideTestReleases `yaml:"testing"`
+}
+
+type CPOOverride struct {
+	Version  string `yaml:"version"`
+	CPOImage string `yaml:"cpoImage"`
+}
+
+type CPOOverrideTestReleases struct {
+	Latest   string `yaml:"latest"`
+	Previous string `yaml:"previous"`
+}
+
+//go:embed assets/overrides.yaml
+var overridesYAML []byte
+
+var (
+	overrides          = mustLoadOverrides()
+	overridesByVersion map[string]*CPOOverride
+)
+
+func init() {
+	initOverridesByVersion()
+}
+
+func IsOverridesEnabled() bool {
+	return os.Getenv(config.CPOOverridesEnvVar) == "1"
+}
+
+func CPOImage(version string) string {
+	if override, exists := overridesByVersion[version]; exists {
+		return override.CPOImage
+	}
+	return ""
+}
+
+func LatestOverrideTestReleases() (string, string) {
+	return overrides.Testing.Latest, overrides.Testing.Previous
+}
+
+func mustLoadOverrides() *CPOOverrides {
+	overrides, err := loadOverrides(overridesYAML)
+	if err != nil {
+		panic(fmt.Sprintf("Failed to load cpo overrides: %v", err))
+	}
+	return overrides
+}
+
+func loadOverrides(yamlContent []byte) (*CPOOverrides, error) {
+	result := &CPOOverrides{}
+	if err := yaml.Unmarshal(yamlContent, result); err != nil {
+		return nil, err
+	}
+	return result, nil
+}
+
+func initOverridesByVersion() {
+	overridesByVersion = map[string]*CPOOverride{}
+	for i, override := range overrides.Overrides {
+		overridesByVersion[override.Version] = &overrides.Overrides[i]
+	}
+}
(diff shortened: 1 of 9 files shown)


hypershift-operator/controlplaneoperator-overrides/overrides.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+func CPOImage(version string) string {
+	return lookupCPOImage(version)
+}
+
+func LatestOverrideTestReleases() (string, string) {
+	if len(overrides.Overrides) > 0 {
+		testingReleases := overrides.Overrides[len(overrides.Overrides)-1].Testing
+		return testingReleases.Latest, testingReleases.Previous
```
jparrill, line 50: This implies that the overrideReleases should follow a concrete order in the file?

hypershift-operator/controlplaneoperator-overrides/overrides.go
```diff
@@ -0,0 +1,84 @@
...
+)
+
+type CPOOverrides struct {
+	Overrides []CPOOverride `yaml:"overrides,omitempty"`
+}
+
+type CPOOverride struct {
+	Version  string                  `yaml:"version"`
+	CPOImage string                  `yaml:"cpoImage"`
+	Testing  CPOOverrideTestReleases `yaml:"testing"`
```
jparrill, line 19: What is this field for?

hypershift-operator/controlplaneoperator-overrides/overrides.go
```diff
@@ -0,0 +1,84 @@
...
+	if len(overrides.Overrides) > 0 {
+		testingReleases := overrides.Overrides[len(overrides.Overrides)-1].Testing
+		return testingReleases.Latest, testingReleases.Previous
+	}
+	return "", ""
+}
+
+func lookupCPOImage(version string) string {
+	fmt.Println("overridesByVersion", overridesByVersion)
+	if override, exists := overridesByVersion[version]; exists {
```
jparrill, line 57: IIRC this was for managed services, they will move to supporting disconnected env soon (?) how should we deal with that in here? Should we put the destination internal registry they will use for that? if that's not the case I think we will need to use a releaseProvider to handle the overrides of the overrides (😅)

hypershift-operator/controlplaneoperator-overrides/overrides.go
csrwng: This is for our e2e testing. It specifies which release images should be used for testing this particular override.

hypershift-operator/controlplaneoperator-overrides/overrides.go
csrwng: Only for testing. The thinking is that when we test overrides, we only test the latest mapping.

hypershift-operator/controlplaneoperator-overrides/overrides.go
csrwng: Like any other image we deliver to service delivery, they will need a ICSP in the mgmt cluster to replace with their local registry.

hypershift-operator/controlplaneoperator-overrides/overrides.go
```diff
@@ -0,0 +1,83 @@
...
+func init() {
+	initOverridesByVersion()
+}
+
+func IsOverridesEnabled() bool {
+	return os.Getenv(config.CPOOverridesEnvVar) == "1"
+}
+
+func CPOImage(version string) string {
+	return lookupCPOImage(version)
```
muraee, line 44: this can function can be inlined here.

hypershift-operator/controlplaneoperator-overrides/overrides.go
```diff
@@ -0,0 +1,83 @@
...
+//go:embed assets/overrides.yaml
+var overridesYAML []byte
+
+var (
+	overrides          = mustLoadOverrides()
+	overridesByVersion map[string]*CPOOverride
+)
+
+func init() {
+	initOverridesByVersion()
```
muraee, line 36: this can function can be inlined here.

hypershift-operator/controlplaneoperator-overrides/overrides.go
muraee: why do we need to specify `testing` field for each version then?

hypershift-operator/controlplaneoperator-overrides/overrides.go
csrwng: it's used in the unit tests

hypershift-operator/controlplaneoperator-overrides/overrides.go
csrwng: thx, done

hypershift-operator/controlplaneoperator-overrides/overrides.go
csrwng: because just the version is not enough to know which release image to use

muraee: hmm, I mean if all other testing entries are not used because always use the last entry, why not have a single testing field? I 
```go
type CPOOverrides struct {
	Overrides []CPOOverride `yaml:"overrides,omitempty"`
	Testing CPOOverrideTestReleases `yaml:"testing"`
}
```

csrwng: Makes sense, updated

PR #4902, opened 2024-10-11
HOSTEDCP-2028: e2e: use TechPreviewNoUpgrade feature set in TestCreateCluster

sjenning:
We currently have no test coverage for `TechPreviewNoUpgrade` feature set in hypershift.  We uncovered https://github.com/openshift/cluster-storage-operator/pull/519 and https://github.com/openshift/hypershift/pull/4887 recently.

As of `4.18.0-0.ci-2024-10-11-065556`, this was working again.

This makes `TestCreateCluster` deploy a `TechPreviewNoUpgrade` cluster so we can ensure it stays working.

### DIFF
--- a/test/e2e/create_cluster_test.go
+++ b/test/e2e/create_cluster_test.go
@@ -10,6 +10,7 @@ import (
 	"testing"
 
 	. "github.com/onsi/gomega"
+	configv1 "github.com/openshift/api/config/v1"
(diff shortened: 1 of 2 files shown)


sjenning:
/hold
until SRO is removed from the CSO

sjenning:
actually we can just filter them ourselves

/hold cancel

PR #4903, opened 2024-10-11
OCPBUGS-44967: Reconcile SecretProvider for CNCC on ARO HCP

bryan-cox:
Reconcile the SecretProviderClass for the cluster network operator for ARO HCP deployments. The SecretProviderClass is used by the Secrets Store CSI driver to mount a certificate to a volume in the ingress pod deployment.

Fixes [OCPBUGS-44967](https://issues.redhat.com/browse/OCPBUGS-44967)

**Checklist**


bryan-cox:
Needs the following PRs before we can continue with this one:

- https://github.com/openshift/hypershift/pull/4884
- https://github.com/openshift/cluster-network-operator/pull/2522
- https://github.com/openshift/cloud-network-config-controller/pull/156

PR #4906, opened 2024-10-14
OCPBUGS-43313: Delete CAPI Cluster CR on HC deletion early

enxebre:
Currently the CAPI Cluster is deleted by deleting the namespace and we block HC deletion on the HCP resource. We should signal CAPI Cluster resource for deletion early so child objects make decisions based on this. E.g. exlude drain and wait for volumes. This would also prevent edge chicken-egg scenarios

Fixes #

**Checklist**


csrwng:
We are deleting the CAPI cluster much earlier than that already
https://github.com/openshift/hypershift/blob/960357ff58ecf72c3c0c7f43a7540ac1c4acda7d/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go#L3835-L3840

PR #4914, opened 2024-10-15
OCPBUGS-43374 (manual backport) openstack/e2e: re-work nodepool tests

EmilienM:

This is a manual backport (almost clean) which will help to get our 4.17 CI able to test NodePools.

* When running the nodepool tests, the HostedCluster is created with a
  initial Nodepool and a replica == 0. That involves that the
  router-default won't exist and therefore we can't expect the Ingress to
  work. So let's skip the DNS record creation when there is no nodepool
  but show a warning because we can't expect the cluster to be healthy.

* Remove double get guestClient when creating Route54 records.

* nodepool: force replica to 1 for the default since we need one
  nodepool for Ingress to be working.

* Use an existing function to test the Nodepool replicas and the nodes
  numbers.

* skip NTO tests which can't work on the OpenStack platform at this
  time.

* nodepool/openstack: support AMD64 arch.

* nodepool/openstack: add missing watch in the reconciler


hypershift-operator/controllers/nodepool/nodepool_controller.go
```diff
@@ -2807,6 +2812,11 @@ func (r *NodePoolReconciler) listMachineTemplates(nodePool *hyperv1.NodePool) ([
 		if err != nil {
 			return nil, err
 		}
+	case hyperv1.OpenStackPlatform:
```
EmilienM, line 2815: the only cherry-pick conflict was here because now the CAPI stuff was moved into capi.go. But I ported the change here.

csrwng:
/retitle OCPBUGS-43374: [release-4.17] (manual backport) openstack/e2e: re-work nodepool tests

EmilienM:
/test e2e-openstack
for regression testing

hypershift-operator/controllers/nodepool/nodepool_controller.go
EmilienM: no, what I meant here is that the change here was "manual" and not from the cherry-pick, because in `main` this was done in `capi.go` and in this branch it's in `nodepool_controller.go`. Otherwise, same function, same goal.

jparrill:
Now is linked with Jira correctly :). We will need to wait until the parent issue get moved to verify by QE people. After that someone will need to execute another `/jira refresh` and got merged.

/lgtm

PR #4926, opened 2024-10-16
HOSTEDCP-1994: Pass Azure Key Vault's Managed Identity's Client ID to HO and CPO Deployments

bryan-cox:
In ARO HCP deployments, a managed identity will be assigned to each AKS management cluster. This managed identity will be responsible for pulling out the secrets from any Azure Key Vault on the management cluster, such as the HCP component's managed identity's backing certificate.

Since there is a single managed identity doing this for all HostedClusters on a management cluster, it makes sense for this value to be passed to the HyperShift Operator (HO) installed on the management cluster. However, the control plane operator (CPO) also needs this value when it creates the SecretProviderClass for the Secrets Store CSI driver. 

This PR:
- adds a flag to the HO installation to pass the client ID of the previously mentioned managed identity
- passes the client ID to the CPO deployment

This PR is a part of [HOSTEDCP-1994](https://issues.redhat.com/browse/HOSTEDCP-1994)

**Checklist**

### DIFF
--- a/cmd/install/assets/hypershift_operator.go
+++ b/cmd/install/assets/hypershift_operator.go
@@ -379,6 +379,7 @@ type HyperShiftOperatorDeployment struct {
 	EnableSizeTagging                       bool
 	EnableEtcdRecovery                      bool
 	EnableCPOOverrides                      bool
+	AROHCPKeyVaultUsersClientID             string
 	TechPreviewNoUpgrade                    bool
 }
 
@@ -493,6 +494,13 @@ func (o HyperShiftOperatorDeployment) Build() *appsv1.Deployment {
 		envVars = append(envVars, corev1.EnvVar{
 			Name:  "MANAGED_SERVICE",
 			Value: o.ManagedService,
+		})
+	}
+
+	if len(o.AROHCPKeyVaultUsersClientID) > 0 {
+		envVars = append(envVars, corev1.EnvVar{
+			Name:  config.AROHCPKeyVaultManagedIdentityClientID,
+			Value: o.AROHCPKeyVaultUsersClientID,
 		})
 	}
 
--- a/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
+++ b/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
@@ -2721,6 +2721,15 @@ func reconcileControlPlaneOperatorDeployment(
 		)
 	}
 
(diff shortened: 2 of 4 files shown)


cmd/install/assets/hypershift_operator.go
```diff
@@ -378,6 +378,7 @@ type HyperShiftOperatorDeployment struct {
 	EnableSizeTagging                       bool
 	EnableEtcdRecovery                      bool
 	EnableCPOOverrides                      bool
+	AroHCPKeyVaultUsersClientID             string
```
csrwng, line 381: s/Aro/ARO/

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -131,8 +131,9 @@ const (
 	controlPlanePKIOperatorSignsCSRsLabel                      = "io.openshift.hypershift.control-plane-pki-operator-signs-csrs"
 	useRestrictedPodSecurityLabel                              = "io.openshift.hypershift.restricted-psa"
 
-	etcdEncKeyPostfix    = "-etcd-encryption-key"
-	managedServiceEnvVar = "MANAGED_SERVICE"
+	etcdEncKeyPostfix                     = "-etcd-encryption-key"
+	managedServiceEnvVar                  = "MANAGED_SERVICE"
+	aroHCPKeyVaultManagedIdentityClientID = "ARO_HCP_KEY_VAULT_USER_CLIENT_ID"
```
csrwng, line 136: given that this will in turn be used by the CPO, it should go in https://github.com/openshift/hypershift/blob/main/support/config/constants.go

csrwng: and used by the CLI as well

csrwng (commented):
lgtm, a couple of nits

enxebre:
lgtm other than Cesar feedback

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
muraee, line 136:
```suggestion
	aroHCPKeyVaultManagedIdentityClientID = "ARO_HCP_KEYVAULT_USER_CLIENT_ID"
```

bryan-cox: I am planning to keep them at two words since that's what is in their documentation - https://azure.microsoft.com/en-us/products/key-vault/

PR #4941, opened 2024-10-18
HOSTEDCP-2054: [CPO Refactor] Refactor kube-apiserver

muraee:

Fixes #

**Checklist**


support/controlplane-component/controlplane-component.go
```diff
@@ -299,9 +299,18 @@ func (c *controlPlaneWorkload) defaultOptions(cpContext ControlPlaneContext, pod
...
-	deploymentConfig.Scheduling.PriorityClass = config.DefaultPriorityClass
-	if cpContext.HCP.Annotations[hyperv1.ControlPlanePriorityClass] != "" {
-		deploymentConfig.Scheduling.PriorityClass = cpContext.HCP.Annotations[hyperv1.ControlPlanePriorityClass]
+
+	// TODO: should all request serving omponents get APICriticalPriorityClass?
+	if c.IsRequestServing() {
+		deploymentConfig.Scheduling.PriorityClass = config.APICriticalPriorityClass
+		if priorityClass := cpContext.HCP.Annotations[hyperv1.APICriticalPriorityClass]; priorityClass != "" {
+			deploymentConfig.Scheduling.PriorityClass = priorityClass
+		}
```
muraee, line 308: @enxebre @csrwng question for you :)

csrwng: I think it's just for the components that are in the API request path, and are required for auth (as in the oauth server). Even non-request serving components like openshift-apiserver and openshift-oauth-apiserver I would consider to be APICriticalPriority. However, something like ignition server that is request serving, I would not classify as APICriticalPriority, given that workers can just keep retrying to hit it if it's unavailable.

muraee: I see, thanks!

control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth_types.go
```diff
@@ -0,0 +1,289 @@
+package kas
+
+import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+// TODO
+// We are currently copying this type from k8s.io/apiserver because it is
+// not yet in the kube base we use for 4.16.
+// In 4.17, we should switch back to using the type in k8s.io/apiserver
+// and remove this copy
+// https://github.com/openshift/kubernetes/pull/1881
```
csrwng, line 10: Seems like we can get rid of this code now

control-plane-operator/controllers/hostedcontrolplane/v2/kas/component.go
```diff
@@ -0,0 +1,108 @@
...
+		WithManifestAdapter(
+			"prometheus-recording-rules.yaml",
+			component.WithAdaptFunction(adaptRecordingRules),
+		).
+		WithManifestAdapter(
+			"aws-pod-identity-webhook-kubeconfig.yaml",
+			component.EnableForPlatform(hyperv1.AWSPlatform),
+			component.WithAdaptFunction(adaptAWSPodIdentityWebhookKubeconfigSecret),
+		).
+		WatchResource(&corev1.ConfigMap{}, "kas-config").
```
csrwng, line 103: There's nothing correlating this with 

```golang
		WithManifestAdapter(
			"config.yaml",
			component.WithAdaptFunction(adaptKubeAPIServerConfig),
		).
```

I think we should either name the files the same as the resource name or maybe make the Watch (true/false) a parameter of the WithManifestXXX functions.

control-plane-operator/controllers/hostedcontrolplane/v2/kas/deployment.go
```diff
@@ -0,0 +1,465 @@
...
+		Name: auditWebhookConfigFileVolumeName,
+	}
+	v.Secret = &corev1.SecretVolumeSource{}
+	v.Secret.SecretName = auditWebhookRef.Name
+	return v
+}
+
+func invokeBootstrapRenderScript(workDir, payloadVersion, featureGateYaml string) string {
+
+	var script = `#!/bin/sh
```
csrwng, line 344: It'd be nice if we could make this a static script in the deployment.yaml 
One possibility could be to replace params with env vars

control-plane-operator/controllers/hostedcontrolplane/v2/kas/kms/aws.go
```diff
@@ -0,0 +1,288 @@
...
+	if len(p.awsAuth.AWSKMSRoleARN) == 0 {
+		return fmt.Errorf("aws kms role arn not specified")
+	}
+	podSpec.Volumes = append(podSpec.Volumes,
+		util.BuildVolume(kasVolumeAWSKMSCredentials(), buildVolumeAWSKMSCredentials(aws.AWSKMSCredsSecret("").Name)),
+		util.BuildVolume(kasVolumeKMSSocket(), buildVolumeKMSSocket),
+		util.BuildVolume(kasVolumeAWSKMSCloudProviderToken(), buildKASVolumeAWSKMSCloudProviderToken),
+	)
+	var container *corev1.Container
+	for i, c := range podSpec.Containers {
```
csrwng, line 158: can we use util.UpdateContainer here?

control-plane-operator/controllers/hostedcontrolplane/v2/kas/kms/aws.go
```diff
@@ -0,0 +1,288 @@
...
+				Resources: config.KMSEncryptedObjects(),
+				Providers: providerConfiguration,
+			},
+		},
+	}
+	return encryptionConfig, nil
+}
+
+func (p *awsKMSProvider) ApplyKMSConfig(podSpec *corev1.PodSpec) error {
+	if len(p.activeKey.ARN) == 0 || len(p.kmsImage) == 0 {
```
csrwng, line 141: I wonder if it'd be more consistent if we have a static podspec for this functionality that we can simply merge with the main deployment template podspec (and adapt as necessary)

control-plane-operator/controllers/hostedcontrolplane/v2/kas/kms/ibmcloud.go
```diff
@@ -0,0 +1,309 @@
...
+				Key: hyperv1.IBMCloudIAMAPIKeySecretKey,
+			},
+		}
+		podSpec.Volumes = append(podSpec.Volumes, util.BuildVolume(kasVolumeIBMCloudKMSCustomerCredentials(), buildVolumeIBMCloudKMSCustomerCredentials(p.ibmCloud.Auth.Unmanaged.Credentials.Name)))
+	case hyperv1.IBMCloudKMSManagedAuth:
+	default:
+		return fmt.Errorf("unrecognized ibmcloud kms auth type %s", p.ibmCloud.Auth.Type)
+	}
+	podSpec.Containers = append(podSpec.Containers, util.BuildContainer(kasContainerIBMCloudKMS(), buildKASContainerIBMCloudKMS(p.kmsImage, p.ibmCloud.Region, kmsKPInfo, customerAPIKeyReference)))
+	var container *corev1.Container
```
csrwng, line 295: should use util.UpdateContainer

csrwng (commented):
Some comments. In general, I'd like for us to be more consistent in the use of static yaml templates. If that's what we're mainly going to use, then we should update the parts where we're still building containers with code.

control-plane-operator/controllers/hostedcontrolplane/v2/kas/deployment.go
```diff
@@ -0,0 +1,465 @@
...
+	}
+
+	if auditPolicyConfig(hcp).Profile != configv1.NoneAuditProfileType {
+		deployment.Spec.Template.Spec.Containers = append(deployment.Spec.Template.Spec.Containers, buildAuditLogsContainer())
+	}
+
+	// With managed etcd, we should wait for the known etcd client service name to
+	// at least resolve before starting up to avoid futile connection attempts and
+	// pod crashing. For unmanaged, make no assumptions.
+	if hcp.Spec.Etcd.ManagementType == hyperv1.Managed {
```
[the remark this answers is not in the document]
muraee: makes sense.

control-plane-operator/controllers/hostedcontrolplane/v2/assets/kube-apiserver/audit-config.yaml
```diff
@@ -0,0 +1,57 @@
+apiVersion: v1
+data:
+  policy.yaml: |
```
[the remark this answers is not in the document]
muraee: true, removed.

control-plane-operator/controllers/hostedcontrolplane/v2/kas/component.go
muraee: renamed the file.
> maybe make the Watch (true/false) a parameter of the WithManifestXXX functions.

The reason I have the watch as a separate option, is because components might want to watch resources that they don't necessarily reconcile.

control-plane-operator/controllers/hostedcontrolplane/v2/kas/deployment.go
muraee: done

control-plane-operator/controllers/hostedcontrolplane/v2/kas/kms/aws.go
muraee: updated to let the providers return a config struct instead of taking `podSpec` as an argument.
Let me know if that looks good to you.

control-plane-operator/controllers/hostedcontrolplane/v2/kas/component.go
```diff
@@ -0,0 +1,108 @@
...
+		WatchResource(&corev1.ConfigMap{}, "kas-audit-config").
+		WatchResource(&corev1.ConfigMap{}, "auth-config").
+		WatchResource(&corev1.Secret{}, "kas-secret-encryption-config").
```
enxebre, line 106: how about the other resource this component reconcile, don't we want to watch those? e.g. PDB, capi-kubeconfig...

enxebre: ah, I got confused by the term

PR #4942, opened 2024-10-18
NO-JIRA: Change globalConfigString to drop empty imageStreamImportMode

enxebre:
This PR https://github.com/openshift/api/pull/1928 introduced a string field which has no omitempty tag.
    This result in our mashalling transparently changing. This produces a different configuration Hash.
    This PR drops the field when empty from the mashalled string to keep backward compatibility.
    Implementing this at the marshal operation level might result in undesired impact as we might pontetially modify other fields and ordering is not deterministic.
    The PR also increases test coverage using raw strings to make sure they fail if any input for our Hash generation ever changes transparently

/hold

Fixes #

**Checklist**


hypershift-operator/controllers/nodepool/config.go
```diff
@@ -319,3 +321,35 @@ func encode(obj runtime.Object, ser *serializer.Serializer) ([]byte, error) {
...
+	if err := enc.Encode(proxy); err != nil {
+		return "", fmt.Errorf("failed to encode proxy global config: %w", err)
+	}
+
+	if err := enc.Encode(image); err != nil {
+		return "", fmt.Errorf("failed to encode image global config: %w", err)
+	}
+
+	// This PR https://github.com/openshift/api/pull/1928 introduced a string field which has no omitempty tag.
+	// This result in our mashalling transparently changing. This produces a different configuration Hash.
```
bryan-cox, line 349:
```suggestion
	// This results in our marshaling transparently changing. This produces a different configuration Hash.
```

hypershift-operator/controllers/nodepool/config.go
```diff
@@ -319,3 +321,35 @@ func encode(obj runtime.Object, ser *serializer.Serializer) ([]byte, error) {
...
+	// This PR drops the field when empty from the mashalled string to keep backward compatibility.
```
bryan-cox, line 350:
```suggestion
	// This PR drops the field when empty from the marshaled string to keep backward compatibility.
```

hypershift-operator/controllers/nodepool/config.go
```diff
@@ -319,3 +321,35 @@ func encode(obj runtime.Object, ser *serializer.Serializer) ([]byte, error) {
...
+	// Implementing this at the marshal operation level might result in undesired impact as we might pontetially modify other fields and ordering is not deterministic
```
bryan-cox, line 351:
```suggestion
	// Implementing this at the marshal operation level might result in undesired impact as we might potentially modify other fields and ordering is not deterministic
```

hypershift-operator/controllers/nodepool/config_test.go
```diff
@@ -108,6 +109,17 @@ spec:
   kernelType: ""
   osImageURL: ""
 `
+	globalConfig := hyperv1.ClusterConfiguration{
+		Authentication: &configv1.AuthenticationSpec{},
+		Image:          &configv1.ImageSpec{},
+		Proxy:          &configv1.ProxySpec{},
+	}
+	// Validating against this ensure that if marshalling of the config APIs ever produces a different output
```
bryan-cox, line 117:
```suggestion
	// Validating against this ensure that if marshaling of the config APIs ever produces a different output
```

go.sum
```diff
@@ -1270,8 +1270,8 @@ github.com/ruudk/golang-pdf417 v0.0.0-20181029194003-1af4ab5afa58/go.mod h1:6lfF
 github.com/ruudk/golang-pdf417 v0.0.0-20201230142125-a7e3863a1245/go.mod h1:pQAZKsJ8yyVxGRWYNEm9oFB8ieLgKFnamEyDmSA0BRk=
 github.com/sergi/go-diff v1.3.1 h1:xkr+Oxo4BOQKmkn/B9eMK0g5Kg/983T9DqqPHwYqD+8=
 github.com/sergi/go-diff v1.3.1/go.mod h1:aMJSSKb2lpPvRNec0+w3fl7LP9IOFzdc9Pa4NFbPK1I=
-github.com/shopspring/decimal v1.4.0 h1:bxl37RwXBklmTi0C79JfXCEBD1cqqHt0bbgBAGFp81k=
-github.com/shopspring/decimal v1.4.0/go.mod h1:gawqmDU56v4yIKSwfBSFip1HdCCXN8/+DMd9qYNcwME=
+github.com/shopspring/decimal v1.3.1 h1:2Usl1nmF/WZucqkFZhnfFYxxxu8LG21F6nPQBE5gKV8=
```
sjenning, line 1273: these deps are getting older.  doesn't seem right.  doesn't seem related to the change either.

hypershift-operator/controllers/nodepool/config.go
```diff
@@ -319,3 +321,35 @@ func encode(obj runtime.Object, ser *serializer.Serializer) ([]byte, error) {
 	}
 	return buff.Bytes(), nil
 }
+
+func globalConfigString(hcluster *hyperv1.HostedCluster) (string, error) {
```
sjenning, line 325: why are we moving this function? seems like more diff that needed.

enxebre: It's actually left over of previous refactor. The func belongs to config.go where is being consumed. I put a dedicated commit for this now.

bryan-cox:
/retitle NO-JIRA: Change globalConfigString to drop empty imageStreamImportMode

PR #4943, opened 2024-10-18
NO-JIRA: Added netlify config to unblock docs PRs

jparrill:
This PR Unblocks the Doc generation in netlify.


jparrill:
Looks like Netlify fixed the issue with the affected images and now python venv path it's correctly set, so this PR is not neccesary anymore
/close

PR #4945, opened 2024-10-18
HOSTEDCP-2000: Add E2E test validating the node runtime

jparrill:
Starting from OpenShift 4.18, the container runtime on worker nodes has been changed to crun. This new test validates that all those worker nodes in this release version contain the correct runtime handlers.

Fixes #[HOSTEDCP-2000](https://issues.redhat.com/browse/HOSTEDCP-2000)

### DIFF
--- a/test/e2e/nodepool_upgrade_test.go
+++ b/test/e2e/nodepool_upgrade_test.go
@@ -215,5 +215,6 @@ func (ru *NodePoolUpgradeTest) Run(t *testing.T, nodePool hyperv1.NodePool, node
 		},
 		e2eutil.WithTimeout(20*time.Minute),
 	)
-	e2eutil.WaitForReadyNodesByNodePool(t, ctx, ru.hostedClusterClient, &nodePool, ru.hostedCluster.Spec.Platform.Type)
+	newNodes := e2eutil.WaitForReadyNodesByNodePool(t, ctx, ru.hostedClusterClient, &nodePool, ru.hostedCluster.Spec.Platform.Type)
+	e2eutil.EnsureNodesRuntime(t, newNodes)
 }
--- a/test/e2e/util/util.go
+++ b/test/e2e/util/util.go
@@ -788,6 +788,31 @@ func EnsureNetworkPolicies(t *testing.T, ctx context.Context, c crclient.Client,
 	})
 }
 
+// EnsureNodesRuntime ensures that all nodes in the NodePool have the expected runtime handlers.
+// This is only supported on 4.18+ when the default runtime is changed to crun.
+func EnsureNodesRuntime(t *testing.T, nodes []corev1.Node) {
+	AtLeast(t, Version418)
+	g := NewWithT(t)
+
+	validHandlers := map[string]bool{
+		"runc": false,
+		"crun": false,
+	}
+
+	for _, node := range nodes {
+		g.Expect(node.Status.RuntimeHandlers).NotTo(BeNil(), "node %s is missing runtime handlers", node.Name)
+		for _, handler := range node.Status.RuntimeHandlers {
+			if _, ok := validHandlers[handler.Name]; ok {
+				validHandlers[handler.Name] = true
+			}
+		}
+
+		for handler, present := range validHandlers {
+			g.Expect(present).To(BeTrue(), "node %s is missing runtime handler %s", node.Name, handler)
+		}
+	}
+}
+
 func getComponentName(pod *corev1.Pod) string {
 	if pod.Labels["app"] != "" {
 		return pod.Labels["app"]


test/e2e/util/util.go
```diff
@@ -757,6 +757,35 @@ func EnsureNetworkPolicies(t *testing.T, ctx context.Context, c crclient.Client,
...
+	nodeList := &corev1.NodeList{}
+	err := c.List(ctx, nodeList)
+	g.Expect(err).NotTo(HaveOccurred(), "failed to list nodes")
+
+	validHandlers := map[string]bool{
+		"runc": false,
+		"crun": false,
+	}
+	for _, node := range nodeList.Items {
+		if node.Status.RuntimeHandlers != nil {
```
enxebre, line 775: this would pass is runtimeHandlers are nil.

test/e2e/util/util.go
```diff
@@ -757,6 +757,35 @@ func EnsureNetworkPolicies(t *testing.T, ctx context.Context, c crclient.Client,
...
+			for _, handler := range node.Status.RuntimeHandlers {
```
enxebre, line 776: fwiw you could something like

```
	for _, node := range nodeList.Items {
		if node.Status.RuntimeHandlers != nil {
			fail
		}
		Expect(node.Status.RuntimeHandlers).To(ConsistOf(
			NodeRuntimeHandler{Name: "runc"},
			NodeRuntimeHandler{Name: "crun"},
		))
	}
```

test/e2e/util/util.go
```diff
@@ -757,6 +757,35 @@ func EnsureNetworkPolicies(t *testing.T, ctx context.Context, c crclient.Client,
 	})
 }
 
+// EnsureNodesRuntime ensures that all nodes in the NodePool have the expected runtime handlers.
+// This is only supported on 4.18+ when the default runtime is changed to crun.
+func EnsureNodesRuntime(t *testing.T, ctx context.Context, c crclient.Client, nodepool *hyperv1.NodePool) {
```
enxebre, line 762: nodepool in the signature seems unused

csrwng: Seems like you should filter nodes by NodePool in the list below, given that you're doing this when the given nodepool is ready, not necessarily all nodes.

test/e2e/util/util.go
jparrill: I like it, but we need to upgrade the OAPI dependency in order to do this kind of comparison. I can do that if you agree. There is another field called UserNamespaces in the [NodeRuntimeHandlerFeatures](https://pkg.go.dev/k8s.io/api@v0.31.1/core/v1#NodeRuntimeHandlerFeatures) struct which needs to be set for that comparision. WDYT?

test/e2e/util/util.go
jparrill: But all the nodes, independently of the nodepool should has those runtimeHandlers. I really matter the nodepool assigned to that node? If it matter I can add it :). just let me know.

jparrill: Ahh ok, makes sense. I will modify the tests in that way. btw it's already gated behind `Version18` tests :).

test/e2e/nodepool_upgrade_test.go
```diff
@@ -203,4 +203,5 @@ func (ru *NodePoolUpgradeTest) Run(t *testing.T, nodePool hyperv1.NodePool, node
 		e2eutil.WithTimeout(20*time.Minute),
 	)
 	e2eutil.WaitForReadyNodesByNodePool(t, ctx, ru.hostedClusterClient, &nodePool, ru.hostedCluster.Spec.Platform.Type)
+	e2eutil.EnsureNodesRuntime(t, ctx, ru.hostedClusterClient, nodes)
```
enxebre, line 206: This would check against the original Nodes for the NodePool.

If you want to check against the ones produced by the rollout
You'll need to get them in the wait invocation e.g.
```
newNodes := e2eutil.EnsureNodesRuntime(t, ctx, ru.hostedClusterClient, nodes)
```
We can actually validate both before and after

test/e2e/util/util.go
```diff
@@ -757,6 +757,30 @@ func EnsureNetworkPolicies(t *testing.T, ctx context.Context, c crclient.Client,
 	})
 }
 
+// EnsureNodesRuntime ensures that all nodes in the NodePool have the expected runtime handlers.
+// This is only supported on 4.18+ when the default runtime is changed to crun.
+func EnsureNodesRuntime(t *testing.T, ctx context.Context, c crclient.Client, nodes []corev1.Node) {
```
enxebre, line 762: ctx and client seems unused so unnecessary in the signature

PR #4955, opened 2024-10-21
OCPBUGS-38132: cpo: produce valid oauth config even if idp config is invalid

sjenning:
follow up fix to https://github.com/openshift/hypershift/pull/4852

After https://github.com/openshift/hypershift/pull/4852, we returned `nil` instead of the valid oauth config (without the invalid IDPs).  We want is a valid oauth config with the invalid IDPs removed.


sjenning:
/hold
for manual testing

sjenning:
done
/hold cancel

PR #4956, opened 2024-10-21
NO-JIRA: allow e2e tests to skip TearDown

EmilienM:

In some cases, it can be useful to not tear down the HostedClusters
that are deployed by the e2e tests.
For example when debugging or also when we only want to run one single
test that will deploy a cluster and then later we want to use that
cluster for e.g. run the CSI test suite.


EmilienM:
/test e2e-aks e2e-aws-4-17 hypershift-operator-main-on-pull-request
flakes, unrelated to this PR

EmilienM:
https://redhat-external.slack.com/archives/C01C8502FMM/p1729685813913709

This was rejected by the Hypershift team, for good reasons. I was hoping otherwise but this is fine.

/close

PR #4960, opened 2024-10-21
HOSTEDCP-1994: Add generic create function for SecretProvider

bryan-cox:
Add generic create function for SecretProviderClass for ARO HCP. 

Example usage - https://github.com/openshift/hypershift/pull/4884.

Fixes #

**Checklist**

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/manifests/secretproviderclass.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/manifests/secretproviderclass.go
@@ -1,0 +1,50 @@
+package manifests
+
+import (
+	"fmt"
+	"github.com/openshift/hypershift/support/azureutil"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	secretsstorev1 "sigs.k8s.io/secrets-store-csi-driver/apis/v1"
+)
+
+const (
+	objectFormat = `
+array:
+  - |
+    objectName: %s
+    objectType: secret
+`
+)
+
+// ManagedAzureKeyVaultSecretProviderClass returns an instance of a SecretProviderClass completed with its name, Azure Key Vault set
+// up, and the certificate name it needs to pull from the Key Vault.
+//
+// https://learn.microsoft.com/en-us/azure/aks/csi-secrets-store-identity-access?tabs=azure-portal&pivots=access-with-a-user-assigned-managed-identity
+func ManagedAzureKeyVaultSecretProviderClass(secretProviderClassName, namespace, keyVaultName, KeyVaultTenantID, certificateName string) *secretsstorev1.SecretProviderClass {
+	return &secretsstorev1.SecretProviderClass{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      secretProviderClassName,
+			Namespace: namespace,
+		},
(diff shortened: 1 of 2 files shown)


control-plane-operator/controllers/hostedcontrolplane/manifests/secretproviderclass.go
```diff
@@ -0,0 +1,45 @@
...
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	secretsstorev1 "sigs.k8s.io/secrets-store-csi-driver/apis/v1"
+)
+
+// SecretProviderClassForAroHCP returns an instance of a SecretProviderClass completed with its name, Azure Key Vault set
+// up, and the certificate name it needs to pull from the Key Vault.
+//
+// https://learn.microsoft.com/en-us/azure/aks/csi-secrets-store-identity-access?tabs=azure-portal&pivots=access-with-a-user-assigned-managed-identity
+func SecretProviderClassForAroHCP(secretProviderClassName, namespace, keyVaultName, KeyVaultTenantID, certificateName string) *secretsstorev1.SecretProviderClass {
```
enxebre, line 14: can we not add specific ref for ForAROHCP? This is a building block we happen to instantiate when installing in that env but it ins't necessarily exclusive to it. Should this be ManagedAzureKeyVaultSecretProviderClass or similar?

control-plane-operator/controllers/hostedcontrolplane/manifests/secretproviderclass.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+}
+
+// formatSecretProviderClassObject places the certificate name in the appropriate string structure the
+// SecretProviderClass expects for an object. More details here:
+// - https://learn.microsoft.com/en-us/azure/aks/csi-secrets-store-identity-access?tabs=azure-portal&pivots=access-with-a-user-assigned-managed-identity#configure-managed-identity
+// - https://secrets-store-csi-driver.sigs.k8s.io/concepts.html?highlight=object#custom-resource-definitions-crds
+func formatSecretProviderClassObject(certName string) string {
+	return `
+array:
```
muraee, line 40: should this be a const?
Also, I think you should use `fmt.Sprintf` instead of strings concatenation with `+`

control-plane-operator/controllers/hostedcontrolplane/manifests/secretproviderclass.go
```diff
@@ -0,0 +1,52 @@
...
+		},
+	}
+}
+
+// formatSecretProviderClassObject places the certificate name in the appropriate string structure the
+// SecretProviderClass expects for an object. More details here:
+// - https://learn.microsoft.com/en-us/azure/aks/csi-secrets-store-identity-access?tabs=azure-portal&pivots=access-with-a-user-assigned-managed-identity#configure-managed-identity
+// - https://secrets-store-csi-driver.sigs.k8s.io/concepts.html?highlight=object#custom-resource-definitions-crds
+func formatSecretProviderClassObject(certName string) string {
+	return fmt.Sprint(objectFormatBeginning, certName, objectFormatEnd)
```
csrwng, line 51: My preference would be to have this as a single format string with params

csrwng (commented):

PR #4963, opened 2024-10-22
HOSTEDCP-1826: review Azure API

enxebre:

Fixes #

**Checklist**


bryan-cox:
/lgtm
/hold 
Hold for tests; others to take a look; sync with ARO folks

bryan-cox:
/test e2e-aks
flaky test

api/hypershift/v1beta1/azure.go
```diff
@@ -52,33 +46,20 @@ type AzureNodePoolPlatform struct {
...
+	// osDisk provides configuration for the OS disk for the nodepool.
+	// This can be used to configure the size, storage account type, encryption options and whether the disk is persistent or ephemeral.
+	// When not provided, the platform will choose reasonable defaults which are subject to change over time.
+	// Review the fields within the osDisk for more details.
+	OSDisk AzureNodePoolOSDisk `json:"osDisk"`
 
-	// AvailabilityZone is the failure domain identifier where the VM should be attached to. This must not be specified
-	// for clusters in a location that does not support AvailabilityZone.
+	// availabilityZone is the failure domain identifier where the VM should be attached to.
+	// This must not be specified for clusters in a location that does not support AvailabilityZone because... TODO: why?
```
bennerv, line 56: >because... TODO: why?

It will fail to provision?  We should create availability sets instead if not specified?

api/hypershift/v1beta1/azure.go
```diff
@@ -90,74 +71,116 @@ type AzureNodePoolPlatform struct {
...
 	// under HostedCluster.Spec.Platform.Azure.ResourceGroupName.
 	//
 	// If this field is not supplied, the Service Principal credentials will be written to a file on the disk of each VM
 	// in order to be accessible by the cloud provider; the aforementioned credentials provided are the same ones as
 	// HostedCluster.Spec.Platform.Azure.Credentials. However, this is less secure than using a managed identity.
 	//
+	// TODO: What is the valid character set for this field? What about minimum and maximum lengths?
+	//
 	// +optional
 	MachineIdentityID string `json:"machineIdentityID,omitempty"`
```
bennerv, line 116: I thought we were removing machine identity cc @bryan-cox

bryan-cox: From the API, yeah we still need to. I can put in a follow up PR if Alberto doesn't want to do that here. This PR is essentially blocking the cert work.

api/hypershift/v1beta1/azure.go
```diff
@@ -52,33 +46,20 @@ type AzureNodePoolPlatform struct {
...
-	// for clusters in a location that does not support AvailabilityZone.
+	// availabilityZone is the failure domain identifier where the VM should be attached to.
+	// This must not be specified for clusters in a location that does not support AvailabilityZone because... TODO: why?
+	// Availability zones are identified by numbers, either 1, 2 or 3.
 	//
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=3
 	// +optional
-	AvailabilityZone string `json:"availabilityZone,omitempty"`
+	AvailabilityZone int32 `json:"availabilityZone,omitempty"`
```
miguelsorianod, line 62: Can we change this to a string?

In CAPZ this is a string https://capz.sigs.k8s.io/reference/v1beta1-api#infrastructure.cluster.x-k8s.io/v1beta1.AzureMachineTemplateResource, as well as in Azure (https://learn.microsoft.com/en-us/rest/api/compute/resource-skus/list?view=rest-compute-2024-07-01&tabs=HTTP#resourceskulocationinfo) instead of an integer. I understand that it  currently maps to integers but if validation is desired that can be performed anyway with a string, but we would be consistent with what's in Azure we give space in case in the future changes.

enxebre: I had the same thought, I don't think we can have guarantees that azure will never introduce a different string that can't be mapped to a int. cc @JoelSpeed since you first introduce the change, any objections?

JoelSpeed: The potential to not be an integer in the future is a fair one

What about using the kube IntOrString type? That would mean we don't have to force users to quote their integers, but, does it make validation harder? Perhaps

What validation can we impose now if we make this a string type, to ensure the zones provided are only `"1"`, `"2"` and `"3"` as we currently expect them to be?

miguelsorianod: > What validation can we impose now if we make this a string type, to ensure the zones provided are only "1", "2" and "3" as we currently expect them to be?

Yes. In the future that validation then can then be relaxed if needed

miguelsorianod: > What about using the kube IntOrString type? That would mean we don't have to force users to quote their integers, but, does it make validation harder? Perhaps

For me a string would be just fine. I don't see the need of coupling it to kube IntOrString or add more complexity. In Azure/CAPZ they don't do that.

enxebre: ok thanks! I choose to go with a string for simplicity, future flexibility while still keeping validation UX

api/hypershift/v1beta1/azure.go
enxebre: yes, let's follow up in different PR for that.

enxebre:
/hold 
to ship https://github.com/openshift/hypershift/pull/4872 first so it can be consumed without this changes

PR #4965, opened 2024-10-22
OCPBUGS-43698: dump: make ControlPlaneComponent's GKV optional

EmilienM:

We now have a new list named `featureGatedResources` which will contain
the GVKs that live behind a feature gate (e.g. ControlPlaneComponent).
This list, combined with the previously tested `ocpResources`, will be
registering the GVKs that actually exist on the management cluster and
ignore the rest.


EmilienM:
I checked the logs and it fixed the bug.
/hold cancel

MaysaMacedo:
/test e2e-openstack

feel free to unhold if the error looks unrelated to you

PR #4977, opened 2024-10-24
HOSTEDCP-967: Drop alpha api and conversion

enxebre:
We originally develop our whole HC/NodePool API as alpha. Then we promote to it beta and enable conversion webhook to ease early consumers transitions e.g. ocm dev environments. While in practice some platforms that belong to the same API group remained not shipped/alpha support like Azure.

Today HC/HCP/NodePool alpha API consumption is not supported anymore, therefore this PR drops all the decommissioned code including the API types, the conversion webhook and the CRDs.

Further evolution of the API will be driven by feature gates https://github.com/openshift/hypershift/pull/4918
going from alpha to stable binary state (stable being "beta" or "v1" once hypershift.openshift.io is promoted).

Other APIs in this repo belonging to a different group might still choose to transition their group from v1alpha1 to stable.

In general we aim to adhere to https://github.com/openshift/enhancements/blob/master/dev-guide%2Fapi-conventions.md
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#adding-unstable-features-to-stable-versions

Fixes #

**Checklist**


rtheis (commented):
Thank you.  Per https://ibm-argonauts.slack.com/archives/C01C8502FMM/p1730233710697949, IBM is OK with this removal.

PR #4927, opened 2024-10-16
OCPBUGS-44960: (OSASINFRA-3643) bump k8s and dependencies

EmilienM:
Changes in `go.mod`:

* Use `github.com/distribution/reference`
* Bump `github.com/openshift/api` and `client-go` to latest
* Bump `github.com/openshift/library-go` to latest
* Bump `github.com/prometheus/common` to `v0.55.0`
* Bump k8s and deps to `v0.31.1`
* Bump CAPI to `v1.8.5`
* Bump `sigs.k8s.io/controller-runtime` to `v0.19.0`

Then code changes to adjust with the new versions.

Also there is a commit to handle the `CloudLoadBalancerConfig` in AWS PlatformStatus: the default value is handled server-side, not client-side. Also, if featuregate is not enabled, the field doesn't exist in the CRD schema, so the content should be stripped server-side. So now check if the PlatformStatus is nil and if yes, we set it to `&configv1.PlatformStatus{}`. We do the same for `PlatformStatus.AWS` where if nil we set it to `&configv1.AWSPlatformStatus{}`. That way, we leave the controllers to reconcile that object.

PR #4980, opened 2024-10-24
OCPNODE-2972: add support for minimumKubeletVersion

haircommander:
Add support for minimum kubelet version
carries https://github.com/openshift/api/pull/2059 and https://github.com/openshift/hypershift/pull/4927 

Fixes #

**Checklist**


enxebre:
/hold
Let's make sure we get https://github.com/openshift/api/pull/2059 in first

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -847,6 +867,39 @@ func (r *HostedControlPlaneReconciler) validateConfigAndClusterCapabilities(hc *
...
+		}
+		if err := nodelib.ValidateMinimumKubeletVersion(nodesClient, minVersion); err != nil {
+			return fmt.Errorf("validating failed for %s: %v", minVersion, err)
+		}
+	}
+	return nil
+}
+
+// A helper function to factor out the pieces that rely on a functioning guest cluster, added for unit testing purposes.
+func (r *HostedControlPlaneReconciler) getHostedClusterNodesClient(ctx context.Context, hc *hyperv1.HostedControlPlane) (corev1client.NodesGetter, error) {
```
enxebre, line 891: hypershift cohesively groups adhoc controllers that need to reconcile guest cluster resources within the hostedclusterconfigoperator.
See https://github.com/openshift/hypershift/tree/main/control-plane-operator/hostedclusterconfigoperator/controllers/nodecount

Please add the logic to set the current minimum version in the HCP there (similar to what we are doing for node count). Ideally this logic would its own controller that is instantiated here https://github.com/openshift/hypershift/blob/main/control-plane-operator/hostedclusterconfigoperator/controllers/nodecount/setup.go#L22 so then we can run it only when the MinimumKubeletVersion gate is enabled

haircommander: are you opposed to me running it in https://github.com/openshift/hypershift/blob/main/control-plane-operator/hostedclusterconfigoperator/controllers/node/node.go instead and gating that behavior on the feature gate, to reduce bolierplate?

haircommander: alrighty we chatted offline and the new controller was suggested, so I've adopted that and used that to pass the oldestKubeletVersion to the HCP controller

api/hypershift/v1beta1/hosted_controlplane.go
```diff
@@ -303,6 +303,9 @@ type HostedControlPlaneStatus struct {
 
 	// NodeCount tracks the number of nodes in the HostedControlPlane.
 	NodeCount *int `json:"nodeCount,omitempty"`
+
+	// OldestKubeletVersion tracks the oldest kubelet version in a hosted cluster
```
enxebre, line 307: This must be behind the featuregate marker and we will want the same feature gated field within the the HostedCluster

haircommander: I'm curious what you're imagining the data flow to be. `kubelet versions -> HostedCluster -> HostedControlPlane`, or `kubelet versions -> HostedControlPlane -> Hosted Cluster` or `kubelet versions -> HostedCluster ; kubelet versions - > HostedControlPlane`

haircommander: I think I'm confused why we need to involve the HostedCluster object at all here. HCP has the client connection and has the oldestKubeletVersion field, so it can validate the minimumKubeletversion and set a condition on the HCP object. It seems to me that when HCP condition is set, that will bubble up to the HC object which will then get exposed. Why do we need an additional field + additional controller for that field and potentially duplicate the checks

control-plane-operator/hostedclusterconfigoperator/controllers/featuregate/controller.go
```diff
@@ -0,0 +1,83 @@
...
+
+func getOldestKubeletVersion(nodes []corev1.Node) *semver.Version {
+	var oldestVersion *semver.Version
+	for _, node := range nodes {
+		vStr := node.Status.NodeInfo.KubeletVersion
+		v, err := nodelib.ParseKubeletVersion(vStr)
+		if err != nil {
+			continue
+		}
+		if oldestVersion == nil || v.LT(*oldestVersion) {
```
enxebre, line 78: why do we need oldestVersion == nil here?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1497,6 +1497,10 @@ type HostedClusterStatus struct {
 	// Platform contains platform-specific status of the HostedCluster
 	// +optional
 	Platform *PlatformStatus `json:"platform,omitempty"`
+
+	// OldestKubeletVersion tracks the oldest kubelet version in a hosted cluster
+	// +openshift:enable:FeatureGate=MinimumKubeletVersion
```
enxebre, line 1502: for this to take effect you'll want to set disable/enable the v1beta1/featuregates, then make update

control-plane-operator/hostedclusterconfigoperator/controllers/featuregate/controller.go
haircommander: if it's the first version found, wouldn't we segfault if we did v.LT(*oldestVersion)?

cmd/cluster/aws/testdata/zz_fixture_TestCreateCluster_minimal_flags_necessary_to_render.yaml
```diff
@@ -26,6 +26,7 @@ metadata:
 spec:
   autoscaling: {}
   configuration:
+    node: null
```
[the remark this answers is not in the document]
haircommander: I misspelled omitempty, this seems to be a vestige that's been replaced. let me see if I can fix it

control-plane-operator/hostedclusterconfigoperator/controllers/featuregate/setup.go
```diff
@@ -0,0 +1,43 @@
...
+
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+const ControllerName = "featuregate"
+
+func Setup(ctx context.Context, opts *operator.HostedClusterConfigOperatorConfig) error {
+	if !featuregate.Gates.Enabled(featuregate.MinimumKubeletVersion) {
```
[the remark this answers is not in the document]
haircommander: updated! I think :)

control-plane-operator/main.go
```diff
@@ -185,6 +186,7 @@ func NewStartCommand() *cobra.Command {
 			"Format is: name1=image1,name2=image2. \"nameX\" is name of an image in OpenShift release (e.g. \"cluster-network-operator\"). "+
 			"\"imageX\" is container image name (e.g. \"quay.io/foo/my-network-operator:latest\"). The container image name is still subject of registry name "+
 			"replacement when --registry-overrides is used.")
+	featuregate.MutableGates.AddFlag(cmd.Flags())
```
enxebre, line 189: The way it's right now, it seems these flags are actually never exercised?

hypershift-operator/main.go
```diff
@@ -343,6 +343,7 @@ func run(ctx context.Context, opts *StartOptions, log logr.Logger) error {
 		CertRotationScale:                       certRotationScale,
 		EnableCVOManagementClusterMetricsAccess: enableCVOManagementClusterMetricsAccess,
 		EnableEtcdRecovery:                      enableEtcdRecovery,
+		FeatureGates:                            featuregate.MutableGates.GetAll(),
```
enxebre, line 346: this seems unused?

enxebre:
Thanks! dropped some feedback on the feature flags plumbing, not blocking to me since I'd rather prevent this PR from keep growing

PR #4986, opened 2024-10-25
HOSTEDCP-2055: [CPO Refactor] Refactor kube-controller-manager

muraee:

Fixes #

**Checklist**

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/testdata/kube-controller-manager/zz_fixture_TestControlPlaneComponents.yaml
+++ b/control-plane-operator/controllers/hostedcontrolplane/testdata/kube-controller-manager/zz_fixture_TestControlPlaneComponents.yaml
@@ -1,0 +1,296 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  creationTimestamp: null
+  labels:
+    hypershift.openshift.io/managed-by: control-plane-operator
+  name: kube-controller-manager
+  namespace: hcp-namespace
+  ownerReferences:
+  - apiVersion: hypershift.openshift.io/v1beta1
+    blockOwnerDeletion: true
+    controller: true
+    kind: HostedControlPlane
+    name: hcp
+    uid: ""
+  resourceVersion: "1"
+spec:
+  replicas: 2
+  revisionHistoryLimit: 2
+  selector:
+    matchLabels:
+      app: kube-controller-manager
+  strategy:
+    rollingUpdate:
+      maxSurge: 1
+      maxUnavailable: 0
+    type: RollingUpdate
+  template:
+    metadata:
+      annotations:
+        cluster-autoscaler.kubernetes.io/safe-to-evict-local-volumes: logs,certs
+        component.hypershift.openshift.io/config-hash: "1252551421580954"
+        hypershift.openshift.io/release-image: ""
+      creationTimestamp: null
+      labels:
+        app: kube-controller-manager
+        hypershift.openshift.io/control-plane-component: kube-controller-manager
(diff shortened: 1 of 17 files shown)


control-plane-operator/controllers/hostedcontrolplane/testdata/kube-controller-manager/zz_fixture_TestControlPlaneComponents.yaml
```diff
@@ -0,0 +1,296 @@
...
+      - emptyDir: {}
+        name: logs
+      - name: kubeconfig
+        secret:
+          defaultMode: 416
+          secretName: kube-controller-manager-kubeconfig
+      - name: cluster-signer
+        secret:
+          defaultMode: 416
+          secretName: cluster-signer-ca
```
enxebre, line 281: tangential to the PR but why is testdata outside v2/?

enxebre: I see it's probably to satisfy CompareWithFixture from TestControlPlaneComponents

control-plane-operator/controllers/hostedcontrolplane/testdata/kube-controller-manager/zz_fixture_TestControlPlaneComponents_component.yaml
```diff
@@ -0,0 +1,37 @@
...
+metadata:
+  creationTimestamp: null
+  name: kube-controller-manager
+  namespace: hcp-namespace
+  resourceVersion: "1"
+spec: {}
+status:
+  conditions:
+  - lastTransitionTime: null
+    message: kube-controller-manager Deployment Available condition not found
```
enxebre, line 12: tangential: is there a place where we are unit testing reconcileComponentStatus for cases where conditions actually exists?

enxebre: Also is there anything preventing us from testing all assets against fixtures? not only Deployment, StatefulSet

control-plane-operator/controllers/hostedcontrolplane/testdata/kube-controller-manager/zz_fixture_TestControlPlaneComponents.yaml
muraee: right, CompareWithFixture generates testdata for `TestControlPlaneComponents` there, since the test is defined in hostedcontrolplane_controller_test.go

control-plane-operator/controllers/hostedcontrolplane/testdata/kube-controller-manager/zz_fixture_TestControlPlaneComponents_component.yaml
muraee: > is there a place where we are unit testing reconcileComponentStatus for cases where conditions actually exists?

we don't, but yeah that makes sense, I will create a ticket for that.

muraee: > Also is there anything preventing us from testing all assets against fixtures? not only Deployment, StatefulSet

not really, its just that the generic unit test, doesn't know what resources each component is generating. so we would probably need to list every possible resource and see if one exists and then create a fixture for it. I don't know if there is an easier way to do that.

PR #4990, opened 2024-10-28
OCPBUGS-43883: openstack: bump CAPO to v0.11.0

EmilienM:

### vendoring: bump CAPO to v0.11.0
* Edit `go.mod` file and bump CAPO to `v0.11.0`
* Run: `go mod tidy && go mod vendor`

### Add ORC new dependency
This currently live in CAPO.

### Run `make update`
... to update generated assets.

### v0.11.0 code adjustments
* Rename `capiopenstack` to `capiopenstackv1beta1` to make it clearer
  what API version we import.
* Add `OpenStackServer` to the dump, it'll be useful for debug.
* Flavor is now an optional string (pointer).
* Change how we convert the OpenStackIdentityReference.


cmd/cluster/core/dump.go
```diff
@@ -349,9 +350,10 @@ func DumpCluster(ctx context.Context, opts *DumpOptions) error {
 		&capiazure.AzureCluster{},
 		&capiazure.AzureMachine{},
 		&capiazure.AzureMachineTemplate{},
-		&capiopenstack.OpenStackCluster{},
-		&capiopenstack.OpenStackMachine{},
-		&capiopenstack.OpenStackMachineTemplate{},
+		&capiopenstackv1alpha1.OpenStackServer{},
```
MaysaMacedo, line 353: do we also need to add the openstackserver at https://github.com/openshift/hypershift/blob/main/cmd/install/assets/assets.go#L54-L58?

EmilienM: done, thx

PR #4997, opened 2024-10-29
OCPBUGS-42434: Reconcile SecretProvider for CPO on ARO HCP

bryan-cox:
Reconcile the SecretProviderClass for the control plane operator for ARO HCP deployments. The SecretProviderClass is used by the Secrets Store CSI driver to mount a certificate to a volume in the control plane operator pod deployment.

Fixes [OCPBUGS-42434](https://issues.redhat.com//browse/OCPBUGS-42434)

**Checklist**


bryan-cox:
/close

in favor of https://github.com/openshift/hypershift/pull/5160

PR #4999, opened 2024-10-30
HOSTEDCP-1971: HyperShift operator upgrade test for rollout validation

devguyio:

This PR adds an upgrade test to validate that a HO upgrade won't cause any unnecessary node rollout. 

1. The setup stage creates a HostedCluster and installs an initial HO image.
2. The test stage upgrades the HyprShift Operator to the version under test and validates that the upgrade did not cause a node rollout.

Fixes [HOSTEDCP-1971](https://issues.redhat.com//browse/HOSTEDCP-1971)

**How to test** :

1. Build the e2e binary
```shell
make e2e
```
2. With your kubeconfig pointing to a fresh cluster run the upgrade-test stage
```shell
./bin/test-e2e -test.run ^TestUpgradeHyperShift$ -test.v -test.timeout 0 \
       --upgrade.run-tests \
      --e2e.hypershift-operator-initial-image "quay.io/hypershift/hypershift-operator:initial"
       --e2e.aws-credentials-file "${AWS_CREDS}" \
       --e2e.pull-secret-file "${PULL_SECRET}" \
       --e2e.aws-region "${REGION}" \
       --e2e.availability-zones "${REGION}a,${REGION}b,${REGION}c" \
       --e2e.aws-oidc-s3-bucket-name "${BUCKET_NAME}" \
       --e2e.base-domain "${BASE_DOMAIN}" \
      --e2e.aws-oidc-s3-credentials ${AWS_CREDS}\
      --e2e.external-dns-credentials ${AWS_CREDS}\
      --e2e.aws-private-credentials-file ${AWS_CREDS}\
```

Here's a successful upgrade test run

[asciicast](https://asciinema.org/a/BlunfPFNihdWAW4Hufl0I8LFH)

**Checklist**


support/conditions/conditions.go
```diff
@@ -39,8 +39,10 @@ func ExpectedHCConditions(hostedCluster *hyperv1.HostedCluster) map[hyperv1.Cond
 
 		conditions[hyperv1.ValidAWSIdentityProvider] = metav1.ConditionTrue
 		conditions[hyperv1.AWSDefaultSecurityGroupCreated] = metav1.ConditionTrue
-		conditions[hyperv1.AWSEndpointAvailable] = metav1.ConditionTrue
-		conditions[hyperv1.AWSEndpointServiceAvailable] = metav1.ConditionTrue
+		if hostedCluster.Spec.Platform.AWS.EndpointAccess == hyperv1.Private {
```
enxebre, line 42: publicAndPrivate topology must also satisfy this.

test/e2e/util/hypershift_framework.go
```diff
@@ -69,15 +74,24 @@ func NewHypershiftTest(t *testing.T, ctx context.Context, test hypershiftTestFun
 	}
 }
 
+// NewHypershiftTestWithManualTeardown is used when the test needs to control when to trigger the teardown. Please note
+// that teardown will be triggered anyway in case of errors (e.g. if creating a HC fails)
+func NewHypershiftTestWithManualTeardown(t *testing.T, ctx context.Context, test hypershiftTestFunc) *hypershiftTest {
+	hst := NewHypershiftTest(t, ctx, test)
+	hst.manualTeardown = true
```
muraee, line 81: I don't think we want to add the option to skip teardown, it can be easily be misused and leave resources in the cloud.

For the purpose of this task, I don't really think you need to use the hypershift e2e framework to create the cluster. All the validations ran by the framework are not important for the upgrade test and will unnecessarily increase the execution time.

devguyio: - Who exactly would misuse this? I don't think this is a public API nor a customer used one, correct? 
- I prefer reusing the same logic that validates that a created test HC is healthy. That way if one HC health invariant changes in one place, it changes in another.
- This test is expected to run solo in parallel to the existing e2e tests, so it won't affect the overall pipeline execution time.

support/conditions/conditions.go
```diff
@@ -39,8 +39,10 @@ func ExpectedHCConditions(hostedCluster *hyperv1.HostedCluster) map[hyperv1.Cond
 
 		conditions[hyperv1.ValidAWSIdentityProvider] = metav1.ConditionTrue
 		conditions[hyperv1.AWSDefaultSecurityGroupCreated] = metav1.ConditionTrue
-		conditions[hyperv1.AWSEndpointAvailable] = metav1.ConditionTrue
-		conditions[hyperv1.AWSEndpointServiceAvailable] = metav1.ConditionTrue
+		if hostedCluster.Spec.Platform.AWS.EndpointAccess == hyperv1.Private || hostedCluster.Spec.Platform.AWS.EndpointAccess == hyperv1.PublicAndPrivate {
+			conditions[hyperv1.AWSEndpointAvailable] = metav1.ConditionTrue
+			conditions[hyperv1.AWSEndpointServiceAvailable] = metav1.ConditionTrue
+		}
```
sjenning, line 45: unrelated change

test/e2e/util/hypershift_framework.go
```diff
@@ -304,8 +342,8 @@ func (h *hypershiftTest) createHostedCluster(opts *PlatformAgnosticOptions, plat
 	return hc
 }
 
-// NOTE: teardownHostedCluster shouldn't start any subtests with t.Run() when cleanupPhase=True, this is not a supported operation and will fail immediately
-func teardownHostedCluster(t *testing.T, ctx context.Context, hc *hyperv1.HostedCluster, client crclient.Client, opts *PlatformAgnosticOptions, artifactDir string, cleanupPhase bool) {
+// NOTE: TeardownHC shouldn't start any subtests with t.Run() when cleanupPhase=True, this is not a supported operation and will fail immediately
+func TeardownHC(t *testing.T, ctx context.Context, hc *hyperv1.HostedCluster, client crclient.Client, opts *PlatformAgnosticOptions, artifactDir string, cleanupPhase bool) {
```
sjenning, line 346: don't need to rename, just make public

test/e2e/util/hypershift_framework.go
```diff
@@ -128,7 +159,8 @@ func (h *hypershiftTest) before(hostedCluster *hyperv1.HostedCluster, opts *Plat
 }
 
 // runs after each test.
-func (h *hypershiftTest) after(hostedCluster *hyperv1.HostedCluster, platform hyperv1.PlatformType) {
+func (h *hypershiftTest) after(platform hyperv1.PlatformType) {
```
sjenning, line 162: unrelated refactor (along with changes to `teardown` and `postTeardown` signatures)

test/e2e/upgrade_cluster_test.go
```diff
@@ -0,0 +1,186 @@
...
+
+	// Get the newly created default NodePool
+	nodepools := &hyperv1.NodePoolList{}
+	if err := client.List(ctx, nodepools, crclient.InNamespace(namespace)); err != nil {
+		t.Fatalf("failed to list nodepools in namespace %s: %v", namespace, err)
+	}
+	g.Expect(nodepools.Items).ToNot(BeEmpty())
+	g.Expect(len(nodepools.Items)).To(BeEquivalentTo(len(zones)))
+
+	nodePool := nodepools.Items[0]
```
sjenning, line 117: seems we assume `len(nodepools.Items)` (and therefore `len(zones)`) to be `1`

test/e2e/upgrade_cluster_test.go
```diff
@@ -0,0 +1,186 @@
...
+	machineDeployments := &v1beta1.MachineDeploymentList{}
+	hcpNamespace := fmt.Sprintf("%s-%s", namespace, preUpgradeInfo.Name)
+
+	if err := client.List(ctx, machineDeployments, crclient.InNamespace(hcpNamespace)); err != nil {
+		t.Fatalf("failed to list machinedeployments in namespace %s: %v", hcpNamespace, err)
+	}
+	g.Expect(machineDeployments.Items).ToNot(BeEmpty())
+	g.Expect(len(machineDeployments.Items)).To(BeEquivalentTo(len(zones)))
+
+	md := machineDeployments.Items[0]
```
sjenning, line 133: same here (assuming len `1`)

test/e2e/upgrade_cluster_test.go
```diff
@@ -0,0 +1,186 @@
...
+	g.Expect(nodepools.Items).ToNot(BeEmpty())
+	g.Expect(len(nodepools.Items)).To(BeEquivalentTo(len(zones)))
+
+	nodePool := nodepools.Items[0]
+
+	g.Expect(nodePool.Generation).To(Equal(preUpgradeInfo.NodePoolGeneration))
+	g.Expect(nodePool.Annotations[nodePoolAnnotationCurrentConfig]).To(Equal(preUpgradeInfo.NodePoolCurrentConfig))
+	g.Expect(nodePool.Annotations[nodePoolAnnotationCurrentConfigVersion]).To(Equal(preUpgradeInfo.NodePoolCurrentConfigVersion))
+
+	// Get the newly created default MachineDeployment
```
sjenning, line 123: Why are we doing asserts on MachineDeployement underneath NodePools? It seems like checking the `NodePool` is sufficient.

enxebre: I agree NodePools  should be sufficient in principle. Though I think additionally asserting on the MachineDeployment is a good extra safety check.

devguyio:
@sjenning @muraee @enxebre I've pushed the new version of the test.

Please review this by commit as I separated the refactoring across commits

test/e2e/upgrade_cluster_test.go
```diff
@@ -0,0 +1,189 @@
...
+						t.Logf("failed to cleanup HyperShift Operator: %v", err)
+					}
+				})
+			})
+		})
+
+		t.Run("Execute upgrade test", func(tt *testing.T) {
+			tt.Logf("Executing upgrade test")
+			e2eutil.NewHypershiftTest(tt, ctx, func(tt *testing.T, g gomega.Gomega, mc crclient.Client, hc *hyperv1.HostedCluster) {
+				tt.Logf("HostedCluster %s/%s created", hc.Namespace, hc.Name)
```
enxebre, line 97: nit: `client.ObjectKeyFromObject(hcluster).String()`

test/e2e/upgrade_cluster_test.go
```diff
@@ -0,0 +1,189 @@
...
+						g.Expect(ok).To(gomega.BeTrue())
+
+						tt.Logf("Generation: Want %d, Got %d", preUpgradeNodePool.Generation, nodePool.Generation)
+						tt.Logf("CurrentConfig: Want %s, Got %s", preUpgradeNodePool.Annotations[nodePoolAnnotationCurrentConfig], nodePool.Annotations[nodePoolAnnotationCurrentConfig])
+						tt.Logf("CurrentConfigVersion: Want %s, Got %s", preUpgradeNodePool.Annotations[nodePoolAnnotationCurrentConfigVersion], nodePool.Annotations[nodePoolAnnotationCurrentConfigVersion])
+
+						// Check if the node pool has been updated
+						g.Expect(nodePool.Generation).To(gomega.Equal(preUpgradeNodePool.Generation))
+						g.Expect(nodePool.Annotations[nodePoolAnnotationCurrentConfig]).To(gomega.Equal(preUpgradeNodePool.Annotations[nodePoolAnnotationCurrentConfig]))
+						g.Expect(nodePool.Annotations[nodePoolAnnotationCurrentConfigVersion]).To(gomega.Equal(preUpgradeNodePool.Annotations[nodePoolAnnotationCurrentConfigVersion]))
```
enxebre, line 174: can we please include a check for the underlying Deployment generation?

test/e2e/upgrade_cluster_test.go
```diff
@@ -0,0 +1,189 @@
...
+		var once sync.Once
+		var cleanup e2eutil.Cleanup
+		var err error
+
+		var mgmtClient crclient.Client
+		var hostedCluster *hyperv1.HostedCluster
+
+		var nodePoolsMap map[string]*hyperv1.NodePool
+
+		hyperShiftOperatorLatestImage := globalOpts.HyperShiftOperatorLatestImage
```
enxebre, line 38: Is this manipulating the HO on the management cluster for the test it self?
If so that would impact and invalidate all the other e2e test running on this management.
Unless you are planning to run this in isolation. If so can you please clarify in the PR desc?

test/e2e/util/install.go
```diff
@@ -0,0 +1,234 @@
...
+}
+
+type Builder func(ctx context.Context, logger logr.Logger, opts *Options) (Cleanup, error)
+
+// InstallHyperShiftCRDs generates and applies CRDs to the cluster for setup of the HyperShift Operator.
+//
+// A closure is returned that knows how to clean up the content on the cluster.
+func InstallHyperShiftCRDs(ctx context.Context, logger logr.Logger, opts *Options) (Cleanup, error) {
+	installLogPath := filepath.Join("install", "hypershift-install-crds.log")
+	renderCmd := exec.CommandContext(ctx, opts.HyperShiftCLIPath,
```
muraee, line 42: why are we running the CLI command locally instead of just calling the CLI code directly?

test/e2e/upgrade_cluster_test.go
```diff
@@ -0,0 +1,189 @@
...
+	nodePoolAnnotationCurrentConfigVersion = "hypershift.openshift.io/nodePoolCurrentConfigVersion"
+	hostedClusterUpgradeTestLabel          = "hypershift.openshift.io/upgrade-test"
+)
+
+// TestUpgradeHyperShift validates that a HyperShift Operator upgrade won't unnecessarily cause a node rollout.
+// The test must be triggered via the flag upgrade.run-tests and the HyperShift Operator initial image must be set
+// via the flag e2e.hypershift-operator-initial-image.
+func TestUpgradeHyperShift(t *testing.T) {
+	if globalOpts.RunUpgradeTest {
+		var once sync.Once
```
muraee, line 29: nit: flip if/else so its easier to read.
```suggestion
func TestUpgradeHyperShift(t *testing.T) {
	if !globalOpts.RunUpgradeTest {
		// This test should be triggered intentionally by setting the option above,
		// otherwise skip it.
		t.SkipNow()
           }
           // test code
           .....
}
```

test/e2e/upgrade_cluster_test.go
```diff
@@ -0,0 +1,189 @@
...
+		}
+
+		t.Run("Install initial HyperShift Operator", func(tt *testing.T) {
+			tt.Logf("Installing HyperShift Operator with image %s", hyperShiftOperatorInitialImage)
+			globalOpts.HyperShiftOperatorLatestImage = hyperShiftOperatorInitialImage
+			cleanup, err = e2eutil.InstallHyperShiftOperatorFull(tt, ctx, globalOpts)
+			if err != nil {
+				tt.Error("Failed to install initial HyperShift Operator")
+				tt.Fail()
+				t.FailNow()
```
muraee, line 80: you can't do that, `FailNow()` must be called from the same goroutine. `t.Run()` starts a new goroutine.

It's also not needed, if a subtest fails, the outer test also fail.

test/e2e/upgrade_cluster_test.go
devguyio: Yeah the plan is to run this test in isolation in its own ci-operator presubmit. 

Creating two nested HCs using the e2e test framework would result in a very complex code IMHO. 

Will clarify in the PR and test description.

test/e2e/upgrade_cluster_test.go
```diff
@@ -0,0 +1,189 @@
...
+			}
+			// Set up the cleanup function on the outer test
+			t.Cleanup(func() {
```
muraee, line 83: Registering cleanup for the outer test from a subtest is weird. Why are we doing that?

test/e2e/util/install.go
devguyio: this code is surfaced from the integration package. I'd say any improvement needed there can be done in a separate PR as long as there're no huge concerns.

The reason it's copied is after a discussion with @enxebre where we agreed to copy any code we'll use from there as a signal to what exactly is useful for the current e2e from that integration package and be able to track that.

test/e2e/upgrade_cluster_test.go
devguyio: Ah yeah I totally forgot about the same goroutine.

> It's also not needed, if a subtest fails, the outer test also fail.

I believe it is needed as when the subtest fails, the outer test doesn't fail immediately, instead Run returns false. It's why on other Run() executions I used the returned value to set the outer test as failed. Here's the doc for Run

```
// Run runs f as a subtest of t called name. It runs f in a separate goroutine
// and blocks until f returns or calls t.Parallel to become a parallel test.
// Run reports whether f succeeded (or at least did not fail before calling t.Parallel).
//
// Run may be called simultaneously from multiple goroutines, but all such calls
// must return before the outer test function for t returns.
```

test/e2e/upgrade_cluster_test.go
devguyio: Can you clarify why is it weird? I think weird is pretty subjective :) .

test/e2e/upgrade_cluster_test.go
muraee: right, if the point is to stop execution afterwards, they yeah checking the return value of `t.Run()` or checking `t.Failed()` is the correct way to do it.

test/e2e/util/install.go
muraee: my question is how is it better to exec a command locally, instead of calling our own code directly?

muraee: > Basically, using this code as-is allows us to get a running upgrade test in place faster, and we can later make it better.

It's not really `as is` if we are introducing new code (even if its copied from somewhere else)

test/e2e/upgrade_cluster_test.go
muraee: imo, I don't think 'install hypershit operator' should be a subtest, its should just be a part of the main test, which will eliminate this problem and the extra checks.

devguyio:
@muraee @enxebre changes pushed and comments addressed. PTAL

test/e2e/upgrade_hypershift_operator_test.go
```diff
@@ -0,0 +1,226 @@
...
+
+// TestUpgradeHyperShiftOperator validates that a HyperShift Operator upgrade won't unnecessarily cause a node rollout.
+// The test must be triggered via the flag upgrade.run-tests and the HyperShift Operator initial image must be set
+// via the flag e2e.hypershift-operator-initial-image.
+func TestUpgradeHyperShiftOperator(t *testing.T) {
+	if !globalOpts.RunUpgradeTest {
+		// This test should be triggered intentionally by setting the option above,
+		// otherwise skip it.
+		t.SkipNow()
+	} else {
```
muraee, line 34: nit: we can drop the `else {}` here, `t.skipNow()` will stop execution immediately.

test/e2e/upgrade_hypershift_operator_test.go
```diff
@@ -0,0 +1,226 @@
...
+			clusterOpts.NodePoolReplicas = 1
+		}
+
+		clusterOpts.BeforeApply = func(o crclient.Object) {
+			switch obj := o.(type) {
+			case *hyperv1.HostedCluster:
+				if obj.Annotations == nil {
+					obj.Annotations = make(map[string]string)
+				}
+				obj.Annotations[hyperv1.ControlPlaneOperatorV2Annotation] = "true"
```
muraee, line 77: I don't think we want to enable CPO V2 for the upgrade test for now.
```suggestion
```

test/e2e/util/options.go
```diff
@@ -0,0 +1,431 @@
...
+	// parent of the management cluster, name and namespace of the management cluster
+	// so we can create additional nodepools for it.
+	ManagementParentKubeconfig string
+	ManagementClusterNamespace string
+	ManagementClusterName      string
+	// If set, the UpgradeControlPlane test will upgrade control plane without
+	// reconciling PKI.
+	DisablePKIReconciliation bool
+
+	HyperShiftCLIPath              string
```
muraee, line 65: this and `OCPath` are not needed anymore, right?

test/e2e/upgrade_hypershift_operator_test.go
```diff
@@ -0,0 +1,226 @@
...
+				t.Logf("Upgrading HyperShift Operator to image %s", hyperShiftOperatorLatestImage)
+				globalOpts.HyperShiftOperatorLatestImage = hyperShiftOperatorLatestImage
+				// Note that we're replacing previous HO installation cleanup function with the new one
+				cleanup, err = e2eutil.InstallHyperShiftOperator(ctx, globalOpts)
+				if err != nil {
+					t.Fatal("Failed to upgrade HyperShift Operator")
+				}
+			})).To(gomega.BeTrue(), "Upgrade HyperShift Operator should succeed")
+
+			g.Expect(t.Run("Verify upgrade invariants", func(t *testing.T) {
```
muraee, line 161: imo, it would be cleaner if the steps in the test `Calculate HyperShift Operator upgrade invariants` -> `Verify upgrade invariants` are not subtests, as you had to wrap them with `g.Expect()` which would get us double error if the subtest fail, and the fact that they share state like nodePoolMap and machineDeploymentMap.

no strong feelings though.

test/e2e/upgrade_hypershift_operator_test.go
```diff
@@ -0,0 +1,226 @@
...
+					t.Logf("Verifying NodePool %s", nodePool.Name)
+					preUpgradeNodePool, ok := nodePoolsMap[nodePool.Name]
+					g.Expect(ok).To(gomega.BeTrue(), "Should have a pre-upgrade nodepool with the name %s",
+						nodePool.Name)
+
+					t.Logf("Generation: Want %d, Got %d", preUpgradeNodePool.Generation, nodePool.Generation)
+					t.Logf("CurrentConfig: Want %s, Got %s",
+						preUpgradeNodePool.Annotations[nodePoolAnnotationCurrentConfig],
+						nodePool.Annotations[nodePoolAnnotationCurrentConfig])
+					t.Logf("CurrentConfigVersion: Want %s, Got %s",
```
muraee, line 186: we are checking those values with `g.Expect` down, right? why do we need these logs? I think it will make it harder to find the real error if there is one.

test/e2e/upgrade_hypershift_operator_test.go
devguyio: I'm optimizing more for the test output readability, IMHO any test has three main steps, pre-condition, action, verification. 

So now the test output looks like this, which is IMHO is a cleaner and more readable output.

```
    --- PASS: TestUpgradeHyperShiftOperator/Main (16.71s)
        --- PASS: TestUpgradeHyperShiftOperator/Main/Calculate_HyperShift_Operator_upgrade_invariants (0.32s)
        --- PASS: TestUpgradeHyperShiftOperator/Main/Upgrade_HyperShift_Operator (15.77s)
        --- PASS: TestUpgradeHyperShiftOperator/Main/Verify_upgrade_invariants (0.22s)
```

test/e2e/upgrade_hypershift_operator_test.go
devguyio: I think it's crucial to have visibility over the test invariants in the case of success as well, to detect any error in the logic or any bug in the test. 

Here's an example output for the case of a failure, do you find it harder to find the real error?

```
=== RUN   TestUpgradeHyperShiftOperator/Main/Verify_upgrade_invariants
    upgrade_hypershift_operator_test.go:162: Verifying upgrade invariants
    upgrade_hypershift_operator_test.go:177: Verifying NodePool example-gddm2-us-east-1a
    upgrade_hypershift_operator_test.go:182: Generation: Want 1, Got 2
    upgrade_hypershift_operator_test.go:183: CurrentConfig: Want 00d3f987, Got 00d3f987
    upgrade_hypershift_operator_test.go:186: CurrentConfigVersion: Want f287df15, Got f287df15
    upgrade_hypershift_operator_test.go:191: 
        Pre-upgrade and post-upgrade NodePool generations should match
        Expected
            <int64>: 1
        not to equal
            <int64>: 2
=== NAME  TestUpgradeHyperShiftOperator/Main
```

test/e2e/util/options.go
devguyio: correct, good catch! fixed!

test/e2e/upgrade_hypershift_operator_test.go
devguyio: fixed

test/e2e/upgrade_hypershift_operator_test.go
devguyio: Fixed

support/conditions/conditions.go
devguyio: addressed

test/e2e/upgrade_cluster_test.go
```diff
@@ -0,0 +1,189 @@
+//go:build e2e
```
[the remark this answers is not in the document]
devguyio: addresed

test/e2e/upgrade_cluster_test.go
devguyio: done

test/e2e/upgrade_cluster_test.go
devguyio: I added the MD generation check, but I didn't use the `EnsureMachineDeploymentGeneration` as I wanted to surface more information in this test specifically like the MD name and log the generation.

test/e2e/upgrade_cluster_test.go
devguyio: Done

test/e2e/util/install.go
devguyio: Addressed

test/e2e/upgrade_cluster_test.go
devguyio: Addressed

test/e2e/upgrade_cluster_test.go
devguyio: Addressed

test/e2e/upgrade_cluster_test.go
devguyio: Addressed.

test/e2e/upgrade_hypershift_operator_test.go
muraee: > to detect any error in the logic or any bug in the test.

you might have a bug in the logs only, since you got duplicate code now. 

> Here's an example output for the case of a failure, do you find it harder to find the real error?

right, you can see the error at the bottom, but I got distracted reading the logs at beginning first. 
Specifically this `CurrentConfig: Want 00d3f987, Got 00d3f987` as it's not easy to see if they match or not immediately :D

again no strong feelings here.

devguyio: Got it, yeah I can see your point. 

Okay will push a version that's just a log of the current state without the want and got and lemme know your thoughts.

devguyio: Done. Now it'd look like this


```
=== RUN   TestUpgradeHyperShiftOperator/Main/Verify_upgrade_invariants
    upgrade_hypershift_operator_test.go:162: Verifying upgrade invariants
    upgrade_hypershift_operator_test.go:177: Verifying NodePool example-gddm2-us-east-1a
    upgrade_hypershift_operator_test.go:182: Generation:  2
    upgrade_hypershift_operator_test.go:183: CurrentConfig:  00d3f987
    upgrade_hypershift_operator_test.go:186: CurrentConfigVersion: f287df15
    upgrade_hypershift_operator_test.go:191: 
        Pre-upgrade and post-upgrade NodePool generations should match
        Expected
            <int64>: 1
        not to equal
            <int64>: 2
=== NAME  TestUpgradeHyperShiftOperator/Main
```

Thoughts?

muraee: thanks, looks better. Why does it say  "Expected <int64>: 1 `not to equal` <int64>: 2" though? shouldn't it be saying `to equal`

devguyio: hahaha, the only way for me to get the test to fail is to change the condition temporary to "NotEqual" instead of "Equal" 😄 , so the actual failure I got on my machine was  `Expected 1 not to equal 1` .

But when I copied it here, I changed the 2nd value to 2 to match the real error that would show in the test, but I forgot to change the `not to equal` to `equal` 😅 

So yeah, it'd actually be

```
        Pre-upgrade and post-upgrade NodePool generations should match
        Expected
            <int64>: 1
         to equal
            <int64>: 2
```

muraee: fair enough XD

test/e2e/e2e_test.go
```diff
@@ -78,72 +62,85 @@ func init() {
...
-	flag.StringVar(&globalOpts.configurableClusterOptions.KubeVirtRootVolumeVolumeMode, "e2e.kubevirt-root-volume-volume-mode", "Filesystem", "The root pvc volume mode")
-	flag.StringVar(&globalOpts.configurableClusterOptions.KubeVirtInfraKubeconfigFile, "e2e.kubevirt-infra-kubeconfig", "", "path to the kubeconfig file of the external infra cluster")
-	flag.StringVar(&globalOpts.configurableClusterOptions.KubeVirtInfraNamespace, "e2e.kubevirt-infra-namespace", "", "the namespace on the infra cluster the workers will be created on")
-	flag.IntVar(&globalOpts.configurableClusterOptions.NodePoolReplicas, "e2e.node-pool-replicas", 2, "the number of replicas for each node pool in the cluster")
+	flag.StringVar(&globalOpts.ConfigurableClusterOptions.AWSCredentialsFile, "e2e.aws-credentials-file", "", "path to AWS credentials")
+	flag.StringVar(&globalOpts.ConfigurableClusterOptions.Region, "e2e.aws-region", "us-east-1", "AWS region for clusters")
+	flag.StringVar(&globalOpts.ConfigurableClusterOptions.PrivatePlatform, "e2e.private-platform", "None", "Platform on which private clusters are supported by the HyperShift Operator (supports \"AWS\" or \"None\")")
```
enxebre, line 67: are this options only relevant for the installing the HO in the TestUpgradeHyperShiftOperator test? if so can we clarify this in the fields description?

test/e2e/upgrade_hypershift_operator_test.go
```diff
@@ -0,0 +1,217 @@
...
+			for i := range machineDeployments.Items {
+				t.Logf("Found MachineDeployment %s", machineDeployments.Items[i].Name)
+				machineDeploymentMap[machineDeployments.Items[i].Name] = &machineDeployments.Items[i]
+			}
+
+		})).To(gomega.BeTrue(), "Calculating HyperShift Operator upgrade invariants should succeed")
+
+		g.Expect(t.Run("Upgrade HyperShift Operator", func(t *testing.T) {
+			t.Logf("Upgrading HyperShift Operator to image %s", hyperShiftOperatorLatestImage)
+			globalOpts.HyperShiftOperatorLatestImage = hyperShiftOperatorLatestImage
```
enxebre, line 149: how do we know this is actually upgrading? can we print the current running image?

devguyio: good point 👍🏽

test/e2e/upgrade_hypershift_operator_test.go
```diff
@@ -0,0 +1,217 @@
...
+				"Number of pre-ugrade and post-upgrade NodePools should match")
+
+			for _, nodePool := range postUpgradeNodePools.Items {
+				t.Logf("Verifying NodePool %s", nodePool.Name)
+				preUpgradeNodePool, ok := nodePoolsMap[nodePool.Name]
+				g.Expect(ok).To(gomega.BeTrue(), "Should have a pre-upgrade nodepool with the name %s",
+					nodePool.Name)
+
+				t.Logf("Generation: %d", nodePool.Generation)
+				t.Logf("CurrentConfig: %s", nodePool.Annotations[nodePoolAnnotationCurrentConfig])
```
enxebre, line 179: can we also check the relevant conditions?
```
NodePoolUpdatingVersionConditionType = "UpdatingVersion"
	// NodePoolUpdatingConfigConditionType signals if a config update is currently happening in NodePool.
	NodePoolUpdatingConfigConditionType = "UpdatingConfig"
	// NodePoolUpdatingPlatformMachineTemplateConditionType signals if a platform machine template update is currently happening in NodePool.
	NodePoolUpdatingPlatformMachineTemplateConditionType = "UpdatingPlatformMachineTemplate"
```

test/e2e/e2e_test.go
enxebre: actually I think they deserve their own substruct like globalOpts.installHOConfig or similar

test/e2e/upgrade_hypershift_operator_test.go
devguyio: Done

test/e2e/e2e_test.go
devguyio: Done

test/e2e/upgrade_hypershift_operator_test.go
devguyio: Done

test/e2e/upgrade_hypershift_operator_test.go
```diff
@@ -0,0 +1,247 @@
...
+					"Pre-upgrade and post-upgrade NodePool current config version should match")
+
+				conditions, err := e2eutil.Conditions(&nodePool)
+				g.Expect(err).ToNot(gomega.HaveOccurred(), "Retrieving NodePool conditions shouldn't return errors")
+				targetConditions := sets.NewString(hyperv1.NodePoolUpdatingVersionConditionType,
+					hyperv1.NodePoolUpdatingConfigConditionType, hyperv1.NodePoolUpdatingPlatformMachineTemplateConditionType)
+				for _, c := range conditions {
+					if targetConditions.Has(c.Type) {
+						t.Logf("Found condition %s of NodePool %s", c.String(), nodePool.Name)
+						g.Expect(c.Status).NotTo(gomega.Equal(metav1.ConditionTrue), "Condition %s of nodepool %s shouldn't be True", nodePool.Name, c.Type)
```
enxebre, line 217: nit: we know the expected state should bit false.

test/e2e/upgrade_hypershift_operator_test.go
enxebre: sorry where is this done? I still see the same log sentence here
`t.Logf("Upgrading HyperShift Operator to image %s", hyperShiftOperatorLatestImage)`

enxebre:
/approve
lgtm other than https://github.com/openshift/hypershift/pull/4999#discussion_r1861713249
/hold

Should we merge the release PR first which would go green as it will noop because there's no TestUpgradeHyperShift.
Then we merge this one with CI already blocking on green and iterating in this same PR is needed?

test/e2e/upgrade_hypershift_operator_test.go
devguyio: After each installation / upgrade of the HO, I get and verify the image. The few lines after the initial installation and the upgrade is done 

https://github.com/openshift/hypershift/pull/4999/files#diff-421ca9e9fe9c0358db093bc1035943aa8b1e1870b99f30a7480d8a28a885e65cR169-R171

test/e2e/upgrade_hypershift_operator_test.go
devguyio: Fixed

devguyio:
> /approve lgtm other than [#4999 (comment)](https://github.com/openshift/hypershift/pull/4999#discussion_r1861713249) /hold
> 
> Should we merge the release PR first which would go green as it will noop because there's no TestUpgradeHyperShift. Then we merge this one with CI already blocking on green and iterating in this same PR is needed?

I'm working on updating the release PR. Sure we can do that indeed.

test/e2e/upgrade_hypershift_operator_test.go
```diff
@@ -0,0 +1,247 @@
...
+	g.Expect(err).ToNot(gomega.HaveOccurred(), "Getting HyperShiftOperator image shouldn't return errors")
+	g.Expect(operatorImage).To(gomega.Equal(hyperShiftOperatorInitialImage))
+
+	t.Logf("Observed pre-upgrade HyperShift Operator image %q", operatorImage)
+
+	t.Log("Executing upgrade test")
+	e2eutil.NewHypershiftTest(t, ctx, func(t *testing.T, g gomega.Gomega, mc crclient.Client, hc *hyperv1.HostedCluster) {
+		t.Logf("HostedCluster %s created", crclient.ObjectKeyFromObject(hc))
+		// Sanity check the cluster by waiting for the nodes to report ready
+		t.Log("Waiting for nodes to report ready")
```
enxebre, line 110: we comment and log "waiting for nodes" but that's not what WaitForGuestClient does?

devguyio:
for e2e-aws, looks like ` TestCreateClusterV2/ValidateHostedCluster` is flaky.

/retest-required

sjenning:
`verify` failing with
```
Makefile:79: *** untracked files detected: M test/e2e/e2e_test.go.  Stop.
```
Probably a `go fmt` change.  Did you run `make update`?

PR #5001, opened 2024-10-30
OCPBUGS-44001: handle errors returned by createServicePrincipalWithCe…

Patryk-Stefanski:

- handle errors returned by createServicePrincipalWithCertificate command
- use generated vnet and nsg resource group names

Fixes # [OCPBUGS-44001](https://issues.redhat.com/browse/OCPBUGS-44001)

**Checklist**

### DIFF
--- a/cmd/infra/azure/create.go
+++ b/cmd/infra/azure/create.go
@@ -165,14 +165,14 @@ func (o *CreateInfraOptions) Run(ctx context.Context, l logr.Logger) (*CreateInf
 	} else {
 		// Create a resource group for network security group
 		nsgResourceGroupName = o.Name + "-nsg"
-		nsgRG, msg, err := createResourceGroup(ctx, o, azureCreds, nsgResourceGroupName, subscriptionID)
+		nsgResourceGroupName, msg, err = createResourceGroup(ctx, o, azureCreds, nsgResourceGroupName, subscriptionID)
 		if err != nil {
 			return nil, fmt.Errorf("failed to create resource group for network security group: %w", err)
 		}
-		l.Info(msg, "name", nsgRG)
+		l.Info(msg, "name", nsgResourceGroupName)
 
 		// Create a network security group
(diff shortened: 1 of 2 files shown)


cmd/infra/azure/create.go
```diff
@@ -165,14 +165,14 @@ func (o *CreateInfraOptions) Run(ctx context.Context, l logr.Logger) (*CreateInf
 	} else {
 		// Create a resource group for network security group
 		nsgResourceGroupName = o.Name + "-nsg"
-		nsgRG, msg, err := createResourceGroup(ctx, o, azureCreds, nsgResourceGroupName, subscriptionID)
+		nsgResourceGroupName, msg, err = createResourceGroup(ctx, o, azureCreds, nsgResourceGroupName, subscriptionID)
```
bryan-cox, line 168: Is this not the same name as above?

cmd/infra/azure/create.go
```diff
@@ -201,14 +201,14 @@ func (o *CreateInfraOptions) Run(ctx context.Context, l logr.Logger) (*CreateInf
 	} else {
 		//create a resource group for virtual network
 		vnetResourceGroupName = o.Name + "-vnet"
-		vnetRG, msg, err := createResourceGroup(ctx, o, azureCreds, vnetResourceGroupName, subscriptionID)
+		vnetResourceGroupName, msg, err = createResourceGroup(ctx, o, azureCreds, vnetResourceGroupName, subscriptionID)
```
bryan-cox, line 204: Similar comment as above

cmd/infra/azure/create.go
Patryk-Stefanski: nope when we create a resource group name we also have a unique identifier as a suffix

cmd/infra/azure/create.go
Patryk-Stefanski: nope when we create a resource group name we also have a unique identifier as a suffix

PR #5013, opened 2024-10-31
OCPBUGS-44037: change Azure infra destroy CLI to use prefixes and inf…

Patryk-Stefanski:

- fixes OCPBUGS-44037 by changing Azure infra destroy CLI to use prefixes and infra-id for SP deletion.
- 
Fixes # [OCPBUGS-44037](https://issues.redhat.com/browse/OCPBUGS-44037)

**Checklist**

### DIFF
--- a/cmd/infra/azure/destroy.go
+++ b/cmd/infra/azure/destroy.go
@@ -2,12 +2,12 @@ package azure
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"os/exec"
 	"strings"
 
-	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
 	"github.com/openshift/hypershift/cmd/log"
 	"github.com/openshift/hypershift/cmd/util"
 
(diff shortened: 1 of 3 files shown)


cmd/infra/azure/destroy.go
```diff
@@ -126,50 +127,61 @@ func (o *DestroyInfraOptions) GetResourceGroupName() string {
...
+		if err != nil {
+			return fmt.Errorf("failed to delete application, %s: %w", app.AppID, err)
+		}
+		if strings.Contains(string(output), "ERROR") {
+			return errors.New(string(output))
+		}
+		logger.Info("Deleted application", "name", app.Name, "appId", app.AppID)
 	}
 
 	if strings.Contains(string(output), "ERROR") {
```
fxierh, line 187: This part appears to be unused

cmd/infra/azure/destroy.go
```diff
@@ -126,50 +127,61 @@ func (o *DestroyInfraOptions) GetResourceGroupName() string {
...
-	if err != nil {
-		return fmt.Errorf("failed to delete service principal, %s: %w", clientID, err)
+	for _, app := range filteredApplications {
+		cmdStr = `az ad app delete --id ` + app.AppID
+		cmd = exec.Command("sh", "-c", cmdStr)
+		output, err = cmd.CombinedOutput()
+		if err != nil {
+			return fmt.Errorf("failed to delete application, %s: %w", app.AppID, err)
+		}
+		if strings.Contains(string(output), "ERROR") {
```
fxierh, line 181: If we invoke az directly (without using sh -c), then it suffices to check if the returned error is nil imo

Patryk-Stefanski: I like being able to see what the error was rather than just seeing and `exit status 1` it easier when it comes to debugging.

PR #5029, opened 2024-11-01
[release-4.17] OSASINFRA-3623: openstack: Only upload required cloud credentials

EmilienM:

This backport will ease the user experience for our cloud credentials.
They are well tested by our CI and we also performed manual testing.
- **trivial: Inline clouds.yaml validation**
- **openstack: Support clouds.yaml discovery, different cloud names**
- **openstack: Only upload the cloud we want**
- **openstack: Support reading cacert from clouds.yaml**


EmilienM:
probably not useful for 4.17
/close

PR #5030, opened 2024-11-01
WIP - cmd/openstack: remove cred file & cacert options

EmilienM:

These options shouldn't be necessary with the recent changes in the CLI.
The user experience is being simplified here to avoid potential misuse
of clouds.yaml.

### DIFF
--- a/cmd/cluster/openstack/create.go
+++ b/cmd/cluster/openstack/create.go
@@ -34,17 +34,13 @@ func BindOptions(opts *RawCreateOptions, flags *pflag.FlagSet) {
 }
 
 func bindCoreOptions(opts *RawCreateOptions, flags *pflag.FlagSet) {
-	// TODO(stephenfin): This is unnecessary given the information should already be in clouds.yaml. We should deprecate and remove it.
-	flags.StringVar(&opts.OpenStackCredentialsFile, "openstack-credentials-file", opts.OpenStackCredentialsFile, "Path to the OpenStack credentials file (optional)")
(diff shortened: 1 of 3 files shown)


cmd/cluster/openstack/create.go
```diff
@@ -34,17 +34,13 @@ func BindOptions(opts *RawCreateOptions, flags *pflag.FlagSet) {
 }
 
 func bindCoreOptions(opts *RawCreateOptions, flags *pflag.FlagSet) {
-	// TODO(stephenfin): This is unnecessary given the information should already be in clouds.yaml. We should deprecate and remove it.
-	flags.StringVar(&opts.OpenStackCredentialsFile, "openstack-credentials-file", opts.OpenStackCredentialsFile, "Path to the OpenStack credentials file (optional)")
```
stephenfin, line 38: I think I put the TODO on the wrong line. We might want to keep this if we want to allow users to specify an exact `clouds.yaml` and not rely on auto-discovery? Thoughts?

EmilienM: I'm thinking of the use case where a customer provides a clouds.yaml without the CAcert because they have the CAcert of the cloud part of their trusted CA (e.g. on a company laptop, where CA trust is handled by the IT admins. They don't need the CAcert, but the HCP would need it).
In fact I'm thinking this whole PR is dangerous and we can still offer that flexibility for now. I also want to discuss with others in the team to know what they think.

cc @mandre @pierreprinetti @mdbooth @MaysaMacedo

EmilienM:
I'm going to close it for now. I think it's fine the way it is now and flexibility (at low enough cost) is offered to our customers.

EmilienM:
I don't think we'll do that for now.

PR #5034, opened 2024-11-04
HOSTEDCP-1924: Set the InfrastructureAvailabilityPolicy to HighlyAvailable by default

LiangquanLi930:
This is important in cases such as compact HCP. When the InfrastructureAvailablityPolicy is set to Single, it causes issues such as:
+ VMs not properly live migrating on node drains/upgrades
  + Per VM PDBs are never created
  + VMs are forcefully restarted on cluster upgrade
+ Virt component replicas are not scaled to total node count – this can lead to Virt control plane delays under load

**Checklist**

### DIFF
--- a/api/hypershift/v1beta1/hostedcluster_types.go
+++ b/api/hypershift/v1beta1/hostedcluster_types.go
@@ -421,7 +421,7 @@ type HostedClusterSpec struct {
 	// SingleReplica.
 	//
 	// +optional
-	// +kubebuilder:default:="SingleReplica"
+	// +kubebuilder:default:="HighlyAvailable"
 	// +immutable
 	InfrastructureAvailabilityPolicy AvailabilityPolicy `json:"infrastructureAvailabilityPolicy,omitempty"`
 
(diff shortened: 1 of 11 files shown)


EmilienM:
What? Why are we doing this?

LiangquanLi930:
> What? Why are we doing this?

Demand from https://issues.redhat.com/browse/HOSTEDCP-1923
cc @bryan-cox

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -421,7 +421,7 @@ type HostedClusterSpec struct {
 	// SingleReplica.
 	//
 	// +optional
-	// +kubebuilder:default:="SingleReplica"
+	// +kubebuilder:default:="HighlyAvailable"
```
EmilienM, line 424: You simply can't do that.
This API is released as v1beta1, every change has to be backward compatible and this is not the case here.
That's why I raised my concerns.

EmilienM:
nevermind my comments, HA is enabled by default when using the HCP cli. This makes sense.

PR #5044, opened 2024-11-05
NO-JIRA: Update the HCP migration script

jparrill:
This script allows the user to migrate a HCP among different management clusters. This is not supported as it is now using the script. This feature will be released presumably in 4.19+


contrib/migration/README.md
```diff
@@ -0,0 +1,52 @@
+# BM HCP Migration Script (Alpha)
+
+## Disclaimer
+**This is not supported** as it is now. Eventually this feature will be released using OADP and a custom plugin to cover this functionality.
```
zanetworker, line 4: Can we rephrase to something like: 

This script is a targeted solution meant as a stop-gap and designed to enable functionality for specific cases as an interim measure. 

Full support will ultimately be provided through OADP as part of the long-term solution.

contrib/migration/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Disclaimer
+
+This script is a targeted solution meant as a stop-gap and designed to enable functionality for specific cases as an interim measure.
+
+Full support will ultimately be provided through OADP as part of the long-term solution.
+
+## Prerequisites
+
+- >= MCE 2.8 (For now)
```
zanetworker, line 11: 4.18? I thought it works for 4.17 (to be released Nov 5th 2024?)

jparrill:
We need another label to add for `contrib` additions.

contrib/migration/README.md
jparrill: Not 100%, @CrystalChun needs to backport something to Assisted Installer to make possible the node adoption in the new management cluster.

CrystalChun: Once this PR https://github.com/openshift/assisted-service/pull/6968 merges, it'll be available in MCE 2.7

jparrill: Ok, let me change the versions then. Thanks folks :)

PR #5045, opened 2024-11-05
OCPBUGS-44221: remove hard dependcy on kubeconfig for hypershift CLI

Patryk-Stefanski:
**What this PR does / why we need it**: removes hard dependency on kubeconfig when running hypershift --help

Fixes # [OCPBUGS-44221](https://issues.redhat.com/browse/OCPBUGS-44221)

**Checklist**

### DIFF
--- a/cmd/cluster/azure/create.go
+++ b/cmd/cluster/azure/create.go
@@ -35,6 +35,10 @@ func DefaultOptions(client crclient.Client, log logr.Logger) (*RawCreateOptions,
 		NodePoolOpts:       azurenodepool.DefaultOptions(),
 	}
 
+	if client == nil {
+		return rawCreateOptions, nil
+	}
+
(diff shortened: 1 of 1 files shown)


enxebre:
/hold
The client is used in many other places in the cli. You need access to the cluster to install and create HCs. How is this change different?

Patryk-Stefanski:
> /hold The client is used in many other places in the cli. You need access to the cluster to install and create HCs. How is this change different?

This is used in the defaultOptions() function which prevents the hypershift --help command from running, I don't think that should require cluster access.

bryan-cox:
@Patryk-Stefanski - there is a typo in the commit subject and the commit description should be added per 5 in https://hypershift-docs.netlify.app/contribute/#prior-to-submitting-a-pull-request.

enxebre:
/hold cancel
/approve

lgtm pending addressing @bryan-cox comment

PR #5047, opened 2024-11-05
NO-JIRA: Put hostedCluster.spec.tolerations behind a feature gate

enxebre:
Since we introduced feature gates new field should start behind a feature gate and satisfy a promotion criteria (to be defined) to be promoted to stable. For now let's gate this field at minimum until:
- There's e2e test coverage for the feature that include day 2 changes of this field.
- Duplicating the corev1.toleration struct so we can add on admission validation which is missing atm.
    - e2e test this on creation UX for failures via https://github.com/openshift/hypershift/blob/main/test/e2e/create_cluster_test.go#L33-L48
- Document the field with the expected behaviour for day 1 and day 2, e.g what happens if I change this day2? Are deleted tolerations deleted or kept? Are changes additive?
    - e2e test day 2 on update UX for failures via https://github.com/openshift/hypershift/blob/d6f79f6cd0a638e07f82b6c57bff6c23a6c8d2c0/test/e2e/util/util.go#L977

Fixes #

**Checklist**

### DIFF
--- a/api/hypershift/v1beta1/featuregates/featureGate-Hypershift-Default.yaml
+++ b/api/hypershift/v1beta1/featuregates/featureGate-Hypershift-Default.yaml
@@ -16,6 +16,9 @@
             {
                 "disabled": [
                     {
+                      "name": "HostedClusterTolerations"
+                    },
+                    {
                       "name": "AROHCPManagedIdentities"
                     },
                     {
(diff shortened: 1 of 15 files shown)


enxebre:
cc @davidvossel

enxebre:
/hold
since the boat might have already shipped

PR #5048, opened 2024-11-05
WIP: test/e2e: add e2e coverage for setting HCP pod tolerations

sjenning:
This is expected to fail until we fix https://issues.redhat.com/browse/OCPBUGS-43357


sjenning:
Done in https://github.com/openshift/hypershift/pull/5543

PR #5074, opened 2024-11-06
OCPBUGS-42434: Authenticate Azure KMS with cert authentication

bryan-cox:
This commit authenticates Azure KMS with certificate authentication in order to communicate with Azure Cloud API. The certificate is stored in an Azure key vault and mounted into the KAS pod through a Secrets Store CSI driver SecretProviderClass.

Fixes [OCPBUGS-42434:](https://github.com/openshift/hypershift/pull/4997/files#top)

**Checklist**


bryan-cox:
/close

in favor of https://github.com/openshift/hypershift/pull/5160

PR #5079, opened 2024-11-07
HOSTEDCP-2003: Add sharedVPC support on product-cli

jparrill:
Add `--vpc-owner-aws-creds` flag in `hcp` product CLI to support sharedVPC cluster creation

Fixes #[HOSTEDCP-2003](https://issues.redhat.com/browse/HOSTEDCP-2003)


csrwng:
/hold
Before merging this, we should decide what the expected workflow is going to be from a product point of view
cc @zanetworker

PR #5090, opened 2024-11-08
HOSTEDCP-2075: Hostedcluster api validation

enxebre:
This PR massively increase docs and on admission validation via CEL for the HostedCluster API, it covers most of the not allowed scenarios in e2e tests.

Some of the CEL validations are restrictive with respect to current state which is technically a breaking change. However many would also fall under bugfix category since allowed input currently result in really bad UX producing non functional and sometimes unrecoverable HCs with no signal no the user. Let's inspect each change separately and analyze how agresive/flexible we might to go for each of them. The goal is to improve UX for well known broken scenarios that are causing user frustration.

Fixes #

**Checklist**


test/e2e/create_cluster_test.go
```diff
@@ -32,30 +34,605 @@ func TestOnCreateAPIUX(t *testing.T) {
 		client, err := e2eutil.GetClient()
 		g.Expect(err).NotTo(HaveOccurred(), "couldn't get client")
 
+		// testCases := []struct {
```
enxebre, line 37: will address this when the PR is /lgtmed

enxebre:
/hold
a bit

enxebre:
cc @davidvossel @rtheis PTAL particularly at the HostedCluster API CEL validations - the E2E validations might help to better visualize the use cases - Let's analyze carefully how any of this would impact existing clusters.

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -351,141 +351,161 @@ const (
...
-	//
+	// sshKey is a local reference to a Secret that must have a "id_rsa.pub" key whose content must be the public part of 1..N SSH keys.
+	// If the reference is set but none of the above requirements are met, the HostedCluster will enter a degraded state.
+	// TODO(alberto): Signal this in a condition.
+	// When sshKey is set, the controllers will generate a machineConfig with the sshAuthorizedKeys https://coreos.github.io/ignition/configuration-v3_2/ populated with this value.
+	// This MachineConfig will be part of every payload generated by the controllers for any NodePool of the HostedCluster.
+	// Changing this value will trigger a rollout for all existing NodePools in the cluster.
+	// +optional
 	// +immutable
 	SSHKey corev1.LocalObjectReference `json:"sshKey"`
```
muraee, line 485: shouldn't this be a pointer with `omitempty` if its optional?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -351,141 +351,161 @@ const (
...
-	// SSHKey references an SSH key to be injected into all cluster node sshd
-	// servers. The secret must have a single key "id_rsa.pub" whose value is the
-	// public part of an SSH key.
-	//
+	// sshKey is a local reference to a Secret that must have a "id_rsa.pub" key whose content must be the public part of 1..N SSH keys.
+	// If the reference is set but none of the above requirements are met, the HostedCluster will enter a degraded state.
+	// TODO(alberto): Signal this in a condition.
+	// When sshKey is set, the controllers will generate a machineConfig with the sshAuthorizedKeys https://coreos.github.io/ignition/configuration-v3_2/ populated with this value.
+	// This MachineConfig will be part of every payload generated by the controllers for any NodePool of the HostedCluster.
+	// Changing this value will trigger a rollout for all existing NodePools in the cluster.
```
muraee, line 482: This is marked as `+immutable` so it can't be changed, right?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -351,141 +351,161 @@ const (
...
+	// The default value is kubernetes.default.svc, which only works for in-cluster
 	// validation.
-	//
+	// If the platform is AWS and this value is set, the controller will update an s3 object with the appropriate OIDC documents (using the serviceAccountSigningKey info) to into that issuerURL.
+	// The expectation is for this s3 url to be backed by an OIDC provider in the AWS IAM.
 	// +kubebuilder:default:="https://kubernetes.default.svc"
 	// +immutable
 	// +optional
-	// +kubebuilder:validation:Format=uri
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf", message="omitempty is immutable"
```
muraee, line 498:
```suggestion
	// +kubebuilder:validation:XValidation:rule="self == oldSelf", message="issuerURL is immutable"
```

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -703,76 +743,124 @@ type NodePortPublishingStrategy struct {
...
 	BaseDomainPrefix *string `json:"baseDomainPrefix,omitempty"`
 
-	// PublicZoneID is the Hosted Zone ID where all the DNS records that are
-	// publicly accessible to the internet exist.
-	//
+	// publicZoneID is the Hosted Zone ID where all the DNS records that are publicly accessible to the internet exist.
+	// This field is optional and mainly leveraged in cloud environments where the DNS records for the .baseDomain are created by controllers in this zone.
+	// Once set, this value is immutable.
 	// +optional
+	// +kubebuilder:validation:XValidation:rule=`oldSelf == "" || self == oldSelf`, message="clusterID is immutable"
```
muraee, line 796:
```suggestion
	// +kubebuilder:validation:XValidation:rule=`oldSelf == "" || self == oldSelf`, message="publicZoneID is immutable"
```

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -703,76 +743,124 @@ type NodePortPublishingStrategy struct {
...
 	PublicZoneID string `json:"publicZoneID,omitempty"`
 
-	// PrivateZoneID is the Hosted Zone ID where all the DNS records that are only
-	// available internally to the cluster exist.
-	//
+	// privateZoneID is the Hosted Zone ID where all the DNS records that are only available internally to the cluster exist.
+	// This field is optional and mainly leveraged in cloud environments where the DNS records for the .baseDomain are created by controllers in this zone.
+	// Once set, this value is immutable.
 	// +optional
+	// +kubebuilder:validation:XValidation:rule=`oldSelf == "" || self == oldSelf`, message="clusterID is immutable"
```
muraee, line 806:
```suggestion
	// +kubebuilder:validation:XValidation:rule=`oldSelf == "" || self == oldSelf`, message="privateZoneID is immutable"
```

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -351,141 +351,161 @@ const (
...
+
 type HostedClusterSpec struct {
-	// Release specifies the desired OCP release payload for the hosted cluster.
-	//
-	// Updating this field will trigger a rollout of the control plane. The
-	// behavior of the rollout will be driven by the ControllerAvailabilityPolicy
-	// and InfrastructureAvailabilityPolicy.
+	// release specifies the desired OCP release payload for all the hosted cluster components.
+	// This includes those components running management side like the Kube API Server and the CVO but also the operands which land in the hosted cluster data plane like the ingress controller, ovn agents, etc.
+	// The maximum and minimum supported release versions are determined by running Hypersfhit Operator.
```
bryan-cox, line 359:
```suggestion
	// The maximum and minimum supported release versions are determined by the running Hypershift Operator.
```

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -351,141 +351,161 @@ const (
...
 	Release Release `json:"release"`
 
-	// ControlPlaneRelease specifies the desired OCP release payload for
-	// control plane components running on the management cluster.
-	// Updating this field will trigger a rollout of the control plane. The
-	// behavior of the rollout will be driven by the ControllerAvailabilityPolicy
-	// and InfrastructureAvailabilityPolicy.
-	// If not defined, Release is used
+	// controlPlaneRelease like spec.release but only for the components running on the management cluster.
+	// This excludes any operand which land in the hosted cluster data plane.
```
bryan-cox, line 367:
```suggestion
	// This excludes any operand which will land in the hosted cluster data plane.
```

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -351,141 +351,161 @@ const (
...
 
+	// infraID is a globally unique identifier for the cluster. This is expected to consist of lowercase alphanumeric characters or '-', start and end with an alphanumeric character
+	// This identifier will be used to associate various cloud resources with the HostedCluster and its associated NodePools.
+	// infraID is used to compute and tag created resources with "kubernetes.io/cluster/"+hcluster.Spec.InfraID which has contractual meaning for the cloud provider implementations.
+	// If a value is not specified, a random infraID will be generated and set by the controller.
+	// Once set, this value is immutable.
+	// +kubebuilder:validation:XValidation:rule="self.matches('^[a-z0-9]([-a-z0-9]*[a-z0-9])?$')",message="infraID must consist of lowercase alphanumeric characters or '-', start and end with an alphanumeric character, and be between 1 and 253 characters"
+	// +kubebuilder:validation:XValidation:rule=`oldSelf == "" || self == oldSelf`, message="infraID is immutable"
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:validation:MinLength=1
```
bryan-cox, line 396: nit: can this really only be a minimum of 1 character?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -351,141 +351,161 @@ const (
...
+	// +immutable
 	Services []ServicePublishingStrategyMapping `json:"services"`
 
-	// PullSecret references a pull secret to be injected into the container
-	// runtime of all cluster nodes. The secret must have a key named
-	// ".dockerconfigjson" whose value is the pull secret JSON.
+	// pullSecret is a local reference to a Secret that must have a ".dockerconfigjson" key whose content must be a valid Openshift pull secret JSON.
+	// If the reference is set but none of the above requirements are met, the HostedCluster will enter a degraded state.
+	// TODO(alberto): Signal this in a condition.
```
bryan-cox, line 470: nit: would be nice to link a Jira ticket here so we don't lose track of this TODO.

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -351,141 +351,161 @@ const (
...
+	// This pull secret will be part of every payload generated by the controllers for any NodePool of the HostedCluster
+	// and it will be injected into the container runtime of all NodePools.
+	// Changing this value will trigger a rollout for all existing NodePools in the cluster.
```
bryan-cox, line 473: Maybe we should add an explicit statement here that changing its contents does not trigger a rollout? The difference here has caused a lot of confusion in the past. Many folks think changing the pull secret contents changes their pull secret and should trigger a rollout when it does not.

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -351,141 +351,161 @@ const (
...
+	// +required
 	PullSecret corev1.LocalObjectReference `json:"pullSecret"`
 
-	// SSHKey references an SSH key to be injected into all cluster node sshd
-	// servers. The secret must have a single key "id_rsa.pub" whose value is the
-	// public part of an SSH key.
-	//
+	// sshKey is a local reference to a Secret that must have a "id_rsa.pub" key whose content must be the public part of 1..N SSH keys.
+	// If the reference is set but none of the above requirements are met, the HostedCluster will enter a degraded state.
+	// TODO(alberto): Signal this in a condition.
```
bryan-cox, line 479: nit: would be nice to link a Jira ticket here so we don't lose track of this TODO.

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -351,141 +351,161 @@ const (
...
-	// a single key named "key". If not specified, a service account signing key will
-	// be generated automatically for the cluster. When specifying a service account
-	// signing key, a IssuerURL must also be specified.
+	// ServiceAccountSigningKey is a local reference to a secret that must have a "key" key whose content must be the private key
+	// used by the service account token issuer.
+	// If not specified, a service account signing key will
+	// be generated automatically for the cluster.
+	// When specifying a service account signing key, an IssuerURL must also be specified.
+	// If the reference is set but none of the above requirements are met, the HostedCluster will enter a degraded state.
+	// TODO(alberto): Signal this in a condition.
```
bryan-cox, line 508: nit: would be nice to link a Jira ticket here so we don't lose track of this TODO.

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -351,141 +351,161 @@ const (
...
 	//
 	// +immutable
 	// +kubebuilder:validation:Optional
```
bryan-cox, line 511: Does this need to be removed? I see the one on `Configuration` below was removed.

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -513,14 +532,18 @@ type HostedClusterSpec struct {
 
 	// ImageContentSources specifies image mirrors that can be used by cluster
 	// nodes to pull content.
-	//
+	// When imageContentSources is set, the controllers will generate a machineConfig.
+	// This MachineConfig will be part of every payload generated by the controllers for any NodePool of the HostedCluster.
+	// Changing this value will trigger a rollout for all existing NodePools in the cluster.
```
bryan-cox, line 537: Just a thought, should we add some moniker for fields like this that trigger rollouts so anyone could easily search for such fields.

Like `// +rollout` or `// +causesRollout` or something else.

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -513,14 +532,18 @@ type HostedClusterSpec struct {
...
-	// +immutable
 	ImageContentSources []ImageContentSource `json:"imageContentSources,omitempty"`
 
-	// AdditionalTrustBundle is a reference to a ConfigMap containing a
-	// PEM-encoded X.509 certificate bundle that will be added to the hosted controlplane and nodes
-	//
+	// AdditionalTrustBundle is a local reference to a ConfigMap that must have a "ca-bundle.crt" key
+	// whose content must be a PEM-encoded X.509 certificate bundle that will be added to the hosted controlplane and nodes
+	// If the reference is set but none of the above requirements are met, the HostedCluster will enter a degraded state.
+	// TODO(alberto): Signal this in a condition.
```
bryan-cox, line 544: nit: would be nice to link a Jira ticket here so we don't lose track of this TODO.

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -533,15 +556,18 @@ type HostedClusterSpec struct {
...
 	FIPS bool `json:"fips"`
 
-	// PausedUntil is a field that can be used to pause reconciliation on a resource.
-	// Either a date can be provided in RFC3339 format or a boolean. If a date is
+	// pausedUntil is a field that can be used to pause reconciliation on the HostedCluster controller. Resulting in any change to the HostedCluster being ignored.
+	// Either a date can be provided in RFC3339 format or a boolean as in 'true', 'false', 'True', 'False'. If a date is
 	// provided: reconciliation is paused on the resource until that date. If the boolean true is
 	// provided: reconciliation is paused on the resource until the field is removed.
+	// +kubebuilder:validation:MaxLength=35
+	// +kubebuilder:validation:MinLength=1
```
bryan-cox, line 569: nit: minimum length is at least 4

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -556,8 +582,11 @@ type HostedClusterSpec struct {
 	// +immutable
 	OLMCatalogPlacement OLMCatalogPlacement `json:"olmCatalogPlacement,omitempty"`
 
-	// NodeSelector when specified, must be true for the pods managed by the HostedCluster to be scheduled.
-	//
+	// NodeSelector when specified, is propagated to all hostedCluster Deployments and Stateful sets running management side.
+	// It must be satisfied by the management Nodes for the pods to be scheduled. Otherwise the HostedCluster will enter a degraded state.
+	// Changes to this field will propagate to existing Deployments and StatefulSets.
+	// +kubebuilder:validation:XValidation:rule="size(self) <= 50",message="nodeSelector map can have at most 5 entries"
+	// TODO(alberto): add additional validation for the map key/values.
```
bryan-cox, line 589: nit: would be nice to link a Jira ticket here so we don't lose track of this TODO.

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -960,27 +1054,31 @@ type Release struct {
 // ClusterAutoscaling specifies auto-scaling behavior that applies to all
 // NodePools associated with a control plane.
 type ClusterAutoscaling struct {
-	// MaxNodesTotal is the maximum allowable number of nodes across all NodePools
-	// for a HostedCluster. The autoscaler will not grow the cluster beyond this
+	// maxNodesTotal is the maximum allowable number of nodes for the Autoscaler scale out to be operational.
+	// The autoscaler will not grow the cluster beyond this number.
+	// If ommited, the autoscaler will not have a maximum limit.
```
bryan-cox, line 1059:
```suggestion
	// If omitted, the autoscaler will not have a maximum limit.
```

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -680,20 +716,24 @@ var (
...
-	OIDC ServiceType = "OIDC"
-
 	// Ignition is the control plane ignition service for nodes.
 	Ignition ServiceType = "Ignition"
 
 	// OVNSbDb is the optional control plane ovn southbound database service used by OVNKubernetes CNI.
 	// Deprecated: This service is no longer used by OVNKubernetes CNI for >= 4.14.
 	OVNSbDb ServiceType = "OVNSbDb"
+
+	// OIDC is the control plane OIDC service.
```
enxebre, line 726: @sjenning please prove me wrong

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1353,6 +1456,8 @@ type ClusterConfiguration struct {
 	// registries, and policies to block or allow registry hostnames.
 	// When exposing OpenShift's image registry to the public, this also lets cluster
 	// admins specify the external hostname.
+	// Changing this value will trigger a rollout for all existing NodePools in the cluster.
+	// TODO(alberto): elaborate why.
```
bryan-cox, line 1460: nit: would be nice to link a Jira ticket here so we don't lose track of this TODO.

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -703,76 +743,124 @@ type NodePortPublishingStrategy struct {
...
+// +kubebuilder:validation:XValidation:rule="(!has(self.machineNetwork) && self.clusterNetwork.all(c, self.serviceNetwork.all(s, c.cidr != s.cidr)) || (has(self.machineNetwork) && (self.machineNetwork.all(m, self.clusterNetwork.all(c, m.cidr != c.cidr)) && self.machineNetwork.all(m, self.serviceNetwork.all(s, m.cidr != s.cidr)) && self.clusterNetwork.all(c, self.serviceNetwork.all(s, c.cidr != s.cidr)))))",message="CIDR ranges in machineNetwork, clusterNetwork, and serviceNetwork must be unique and non-overlapping"
 type ClusterNetworking struct {
-	// MachineNetwork is the list of IP address pools for machines.
-	//
+	// machineNetwork is the list of IP address pools for machines.
+	// This is might be used among other things to generate appropriate networking security groups in some clouds providers.
+	// Currently only one entry is supported.
+	// This field is immutable.
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="machineNetwork is immutable and cannot be modified once set."
+	// +kubebuilder:validation:MaxItems=1
```
enxebre, line 823: @csrwng @sjenning @jparrill is this min/max fair?

bryan-cox (changes requested):
Some notes:
- several fields are listed with a minimum length of 1 when that seems wrong, example: Address.
- Not all fields are marked as either optional or required. Should we go ahead and do that? (I think we have a ticket in the backlog for this though).

api/util/ipnet/ipnet.go
```diff
@@ -16,6 +16,8 @@ var nullBytes = []byte(nullString)
 // IPNet wraps net.IPNet to get CIDR serialization.
 //
 // +kubebuilder:validation:Type=string
+// +kubebuilder:validation:MaxLength=43
+// +kubebuilder:validation:XValidation:rule=`self.matches('^((\\d{1,3}\\.){3}\\d{1,3}/\\d{1,2})$') || self.matches('^([0-9a-fA-F]{0,4}:){2,7}([0-9a-fA-F]{0,4})?/[0-9]{1,3}$')`,message="cidr must be a valid IPv4 or IPv6 CIDR notation (e.g., 192.168.1.0/24 or 2001:db8::/64)"
```
enxebre, line 20: without this validation here, the only input protection is unmarshall first. For any already marshalled resource all strings here are swallowed.

api/hypershift/v1beta1/nodepool_types.go
```diff
@@ -142,17 +141,35 @@ type NodePoolSpec struct {
...
+	// https://github.com/openshift/cluster-node-tuning-operator/blob/2c76314fb3cc8f12aef4a0dcd67ddc3677d5b54f/pkg/apis/tuned/v1/tuned_types.go
+	//
+	// The PerformanceProfile API is defined here:
+	// https://github.com/openshift/cluster-node-tuning-operator/tree/b41042d42d4ba5bb2e99960248cf1d6ae4935018/pkg/apis/performanceprofile/v2
+	//
+	// Each ConfigMap must have a single key named "tuning" whose value is the
+	// JSON or YAML of a serialized Tuned or PerformanceProfile.
+	// Changing this field will trigger a NodePool rollout.
+	// +kubebuilder:validation:Optional
+	TuningConfig []corev1.LocalObjectReference `json:"tuningConfig,omitempty"`
```
sjenning, line 161: why was this moved within the struct def?

enxebre: sorry, you caught me again moving stuff around and making review confusing! I just thought it was better cohesively located next the .config. I'll put it back to avoid confusion.

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -351,141 +351,161 @@ const (
...
 	// +immutable
 	Etcd EtcdSpec `json:"etcd"`
 
-	// Services specifies how individual control plane services are published from
-	// the hosting cluster of the control plane.
-	//
-	// If a given service is not present in this list, it will be exposed publicly
-	// by default.
+	// services specifies how individual control plane services endpoints are published for consumption.
+	// This requires APIServer;OAuthServer;Konnectivity;Ignition.
```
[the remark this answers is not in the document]
sjenning: IIRC, when creating HCPs with kubevirt workers on MCE, they do not include an `APIServer` publishing strategy and rely on us to default it.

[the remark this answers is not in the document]
sjenning: yeah, confirmed
Screenshot from 2024-11-08 10-44-22

api/hypershift/v1beta1/hostedcluster_types.go
csrwng: There is other config that should also cause a rollout, such as proxy and featuregate

csrwng (commented):
Some comments. In general, it'd be great if we could find a way to unit test field validations.

enxebre:
>Some comments. In general, it'd be great if we could find a way to unit test field validations.

@csrwng they are all e2e tested here https://github.com/openshift/hypershift/pull/5090/commits/dfffc163346574a1342e06d6e8d435c9ef8582a1#diff-1ab8a204348b5b2e9d5d5995903341902c62d82f84061b1b7cd9e0858702e6c1

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -351,141 +351,161 @@ const (
...
+	// +kubebuilder:validation:XValidation:rule="self.all(s, !(s.service == 'APIServer' && s.servicePublishingStrategy.type == 'Route') || has(s.servicePublishingStrategy.route.hostname))",message="If serviceType is 'APIServer' and publishing strategy is 'Route', then hostname must be set"
+	// +kubebuilder:validation:XValidation:rule="['APIServer', 'OAuthServer', 'Konnectivity', 'Ignition'].all(requiredType, self.exists(s, s.service == requiredType))",message="Services list must contain at least 'APIServer', 'OAuthServer', 'Konnectivity', and 'Ignition' service types"
+	// +kubebuilder:validation:XValidation:rule="self.filter(s, s.servicePublishingStrategy.type == 'Route' && has(s.servicePublishingStrategy.route) && has(s.servicePublishingStrategy.route.hostname)).all(x, self.filter(y, y.servicePublishingStrategy.type == 'Route' && (has(y.servicePublishingStrategy.route) && has(y.servicePublishingStrategy.route.hostname) && y.servicePublishingStrategy.route.hostname == x.servicePublishingStrategy.route.hostname)).size() <= 1)",message="Each nodePort publishingStrategy 'nodePort' and 'hostname', and each route publishingStrategy 'hostname' must be unique within the Services list."
```
JoelSpeed, line 463: From 1.29, you can use optionals, so you could substitute 
```
has(s.servicePublishingStrategy.route) && has(s.servicePublishingStrategy.route.hostname)
```
with
```
s.servicePublishingStrategy.?route.hostname.hasValue()
```
Perhaps doesn't save much in this example but for longer trees that you're traversing looking for values in the future

You can also `orValue("...")` to provide a default if not set

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -351,141 +351,161 @@ const (
...
+	// +kubebuilder:validation:XValidation:rule="self.filter(s, s.servicePublishingStrategy.type == 'NodePort' && has(s.servicePublishingStrategy.nodePort) && has(s.servicePublishingStrategy.nodePort.address) && has(s.servicePublishingStrategy.nodePort.port)).all(x, self.filter(y, y.servicePublishingStrategy.type == 'NodePort' && (has(y.servicePublishingStrategy.nodePort) && has(y.servicePublishingStrategy.nodePort.address) && y.servicePublishingStrategy.nodePort.address == x.servicePublishingStrategy.nodePort.address && has(y.servicePublishingStrategy.nodePort.port) && y.servicePublishingStrategy.nodePort.port == x.servicePublishingStrategy.nodePort.port )).size() <= 1)",message="Each nodePort publishingStrategy 'nodePort' and 'hostname', and each route publishingStrategy 'hostname' must be unique within the Services list."
+	// +immutable
```
JoelSpeed, line 465: I don't see CEL validating this?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -351,141 +351,161 @@ const (
...
-	// runtime of all cluster nodes. The secret must have a key named
-	// ".dockerconfigjson" whose value is the pull secret JSON.
+	// pullSecret is a local reference to a Secret that must have a ".dockerconfigjson" key whose content must be a valid Openshift pull secret JSON.
+	// If the reference is set but none of the above requirements are met, the HostedCluster will enter a degraded state.
+	// TODO(alberto): Signal this in a condition.
+	// This pull secret will be part of every payload generated by the controllers for any NodePool of the HostedCluster
+	// and it will be injected into the container runtime of all NodePools.
+	// Changing this value will trigger a rollout for all existing NodePools in the cluster.
+	// +required
 	PullSecret corev1.LocalObjectReference `json:"pullSecret"`
```
JoelSpeed, line 475: Prefer to use our own types, instead of importing, it means we can tailor the output, and, it means that if the external type adds a new field, we don't magically gain that in our API, as a new unimplemented field

api/hypershift/v1beta1/hostedcluster_types.go
JoelSpeed: Doesn't have to be, if we consider this a configuration API and we want the field to be discoverable and there are no required fields within the struct

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -619,34 +648,41 @@ type ImageContentSource struct {
...
 	// +immutable
 	Service ServiceType `json:"service"`
 
-	// ServicePublishingStrategy specifies how to publish Service.
+	// servicePublishingStrategy specifies how to publish a service endpoint.
 	ServicePublishingStrategy `json:"servicePublishingStrategy"`
 }
 
 // ServicePublishingStrategy specfies how to publish a ServiceType.
+// +kubebuilder:validation:XValidation:rule="(self.type == 'NodePort' && has(self.nodePort) && !has(self.loadBalancer) && !has(self.route)) || (self.type == 'LoadBalancer' && !has(self.nodePort) && !has(self.route)) || (self.type == 'Route' && !has(self.nodePort) && !has(self.loadBalancer)) || ((self.type == 'None' || self.type == 'S3') && !has(self.nodePort) && !has(self.loadBalancer) && !has(self.route))",message="Only the field corresponding to the 'type' should be set, or none if Type is 'None' or 'S3'."
```
JoelSpeed, line 669: This is easier when you break it down into multiple ternarys
```
// +kubebuilder:validation:XValidation:rule="self.type == 'NodePort' ? has(self.nodePort) : !has(self.nodePort)",message="nodePort is required when type is NodePort, and forbidden otherwise"
```

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -680,20 +716,24 @@ var (
...
+	OIDC ServiceType = "OIDC"
 )
 
 // NodePortPublishingStrategy specifies a NodePort used to expose a service.
 type NodePortPublishingStrategy struct {
 	// Address is the host/ip that the NodePort service is exposed over.
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:validation:MinLength=1
+	// +required
 	Address string `json:"address"`
```
JoelSpeed, line 737: No format validation?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -703,76 +743,124 @@ type NodePortPublishingStrategy struct {
...
+	// +kubebuilder:validation:MinLength=1
 	// +immutable
 	PrivateZoneID string `json:"privateZoneID,omitempty"`
 }
 
-// ClusterNetworking specifies network configuration for a cluster.
+// clusterNetworking specifies network configuration for a cluster.
+// All CIDRs must be unique. Additional validation to check for CIDRs overlap and consistent network stack is perfomed by the controllers.
+// Failing that validation will result in the HostedCluster being degraded and the validConfiguration condition being false.
+// +kubebuilder:validation:XValidation:rule="(!has(self.machineNetwork) && self.clusterNetwork.all(c, self.serviceNetwork.all(s, c.cidr != s.cidr)) || (has(self.machineNetwork) && (self.machineNetwork.all(m, self.clusterNetwork.all(c, m.cidr != c.cidr)) && self.machineNetwork.all(m, self.serviceNetwork.all(s, m.cidr != s.cidr)) && self.clusterNetwork.all(c, self.serviceNetwork.all(s, c.cidr != s.cidr)))))",message="CIDR ranges in machineNetwork, clusterNetwork, and serviceNetwork must be unique and non-overlapping"
```
JoelSpeed, line 816: How can you tell if they are unique by comparing them as strings? CIDR ranges can actually be non-unique because of the bit length, `192.168.0.1/24` is not a correct manifestation of a CIDR network, since it has non-zero bits after the bit mask, however, it would equate to `192.168.0.0/24` if you parsed it as a CIDR

I also don't think you can tell if they are overlapping in this way.

Since OCP 4.16 we've had a CIDR library in the API server

You probably want something like:
```
isCIDR(self) && !cidr(self).containsCIDR(other)
```
I think you probably also want to try concatenating the lists to make this easier to deal with
```
(self.?machineNetwork.orValue([]) + self.clusterNetwork + self.serviceNetwork).all(x, 
    (self.?machineNetwork.orValue([]) + self.clusterNetwork + self.serviceNetwork).exists_one(y, cidr(x).containsCIDR(y))
)
```
Validating that they are in fact valid CIDRs in a separate rule is also probably preferable, so that one can fail first, and make this one easier to read, you can do that on the struct field itself though.

Al of the above works from 4.16 onwards

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -703,76 +743,124 @@ type NodePortPublishingStrategy struct {
...
 type ClusterNetworking struct {
-	// MachineNetwork is the list of IP address pools for machines.
-	//
+	// machineNetwork is the list of IP address pools for machines.
```
JoelSpeed, line 818: What happens when this field is omitted/has no entries?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -703,76 +743,124 @@ type NodePortPublishingStrategy struct {
...
 
-	// ClusterNetwork is the list of IP address pools for pods.
-	//
+	// clusterNetwork is the list of IP address pools for pods.
+	// Defaults to cidr: "10.132.0.0/14".
+	// Currently only one entry is supported.
+	// This field is immutable.
 	// +immutable
+	// +optional
 	// +kubebuilder:default:={{cidr: "10.132.0.0/14"}}
```
JoelSpeed, line 836: This defaulting doesn't work for structured clients, you have no omitempty

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -703,76 +743,124 @@ type NodePortPublishingStrategy struct {
...
+	// serviceNetwork is the list of IP address pools for services.
+	// Defaults to cidr: "172.31.0.0/16".
+	// Currently only one entry is supported.
+	// This field is immutable.
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="serviceNetwork is immutable and cannot be modified once set."
+	// +kubebuilder:validation:MaxItems=1
+	// +kubebuilder:validation:MinItems=1
 	// +optional
 	// +kubebuilder:default:={{cidr: "172.31.0.0/16"}}
 	ServiceNetwork []ServiceNetworkEntry `json:"serviceNetwork"`
```
JoelSpeed, line 851: Needs omitempty for defaulting to work

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -787,10 +875,10 @@ type MachineNetworkEntry struct {
 // ClusterNetworkEntry is a single IP address block for pod IP blocks. IP blocks
 // are allocated with size 2^HostSubnetLength.
 type ClusterNetworkEntry struct {
-	// CIDR is the IP block address pool.
+	// cidr is the IP block address pool.
 	CIDR ipnet.IPNet `json:"cidr"`
```
JoelSpeed, line 879: What schema is generated for this type? Can you validate the fact that this is a real CIDR?
```
isCIDR(self)
```

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1006,21 +1104,25 @@ const (
 )
 
 // EtcdSpec specifies configuration for a control plane etcd cluster.
+// +kubebuilder:validation:XValidation:rule="(self.managementType == 'Managed' && has(self.managed) && !has(self.unmanaged) || self.managementType == 'Unmanaged' && has(self.unmanaged) && !has(self.managed))",message="Only managed configuration must be set when managementType is Managed, and only unmanaged configuration must be set when managementType is Unmanaged"
```
JoelSpeed, line 1107: As above, break into ternaries specific to the type value

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1006,21 +1104,25 @@ const (
...
-	//
+	// managementType defines how the etcd cluster is managed.
+	// This can be either Managed or Unmanaged.
+	// This field is immutable.
 	// +unionDiscriminator
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf", message="managementType is immutable"
+	// +required
 	// +immutable
-	ManagementType EtcdManagementType `json:"managementType"`
+	ManagementType EtcdManagementType `json:"managementType,omitempty"`
```
JoelSpeed, line 1116: Required and omitempty are not friends

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1030,7 +1132,7 @@ type EtcdSpec struct {
 // ManagedEtcdSpec specifies the behavior of an etcd cluster managed by
 // HyperShift.
 type ManagedEtcdSpec struct {
-	// Storage specifies how etcd data is persisted.
+	// storage specifies how etcd data is persisted.
 	Storage ManagedEtcdStorageSpec `json:"storage"`
```
JoelSpeed, line 1136: Should have an explicit tag for required

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1079,16 +1182,16 @@ type ManagedEtcdStorageSpec struct {
...
 type PersistentVolumeEtcdStorageSpec struct {
-	// StorageClassName is the StorageClass of the data volume for each etcd member.
-	//
+	// storageClassName is the StorageClass of the data volume for each etcd member.
 	// See https://kubernetes.io/docs/concepts/storage/persistent-volumes#class-1.
-	//
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf", message="storageClassName is immutable"
 	// +optional
 	// +immutable
 	StorageClassName *string `json:"storageClassName,omitempty"`
```
JoelSpeed, line 1190: Why does this need to be a pointer? Is the empty string a valid opinion?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1079,16 +1182,16 @@ type ManagedEtcdStorageSpec struct {
...
 
-	// Size is the minimum size of the data volume for each etcd member.
-	//
+	// size is the minimum size of the data volume for each etcd member.
+	// Default is 8Gi.
```
JoelSpeed, line 1193: Is there a minimum size?

You can use the CEL quantity library to validate min/max/valid quantities on this field

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1108,7 +1211,7 @@ type UnmanagedEtcdSpec struct {
 	// +kubebuilder:validation:Pattern=`^https://`
 	Endpoint string `json:"endpoint"`
 
-	// TLS specifies TLS configuration for HTTPS etcd client endpoints.
+	// tls specifies TLS configuration for HTTPS etcd client endpoints.
 	TLS EtcdTLSConfig `json:"tls"`
```
JoelSpeed, line 1215: Explicit required tag

api/util/ipnet/ipnet.go
JoelSpeed, line 20: CEL has a CIDR parsing library, it's in OpenShift API servers from 4.16 onwards, you could use that instead of the pattern

api/hypershift/v1beta1/nodepool_types.go
JoelSpeed, line 161: All lists should have max lengths

api/hypershift/v1beta1/hostedcluster_types.go
enxebre: dropped the marker

api/hypershift/v1beta1/hosted_controlplane.go
```diff
@@ -184,7 +184,8 @@ type HostedControlPlaneSpec struct {
 	Tolerations []corev1.Toleration `json:"tolerations,omitempty"`
 }
 
-// AvailabilityPolicy specifies a high level availability policy for components.
+// availabilityPolicy specifies a high level availability policy for components.
```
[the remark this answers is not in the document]
enxebre: You are pointing to the hosted control plane API file which I'm leaving out of the scope of this PR. The hc API docs would render the .controllerAvailabilityPolicy comment

api/hypershift/v1beta1/hostedcluster_types.go
enxebre: This will be a separate activity from this PR.

api/hypershift/v1beta1/hostedcluster_types.go
enxebre: This will be a separate activity from this PR.

api/hypershift/v1beta1/hostedcluster_types.go
enxebre: This will be a separate activity

api/hypershift/v1beta1/hostedcluster_types.go
enxebre: This will be a different activity

api/hypershift/v1beta1/hostedcluster_types.go
enxebre: This will be a different activity

api/hypershift/v1beta1/hostedcluster_types.go
enxebre: consolidating on using 	// +optional everywhere as we go

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -351,141 +351,161 @@ const (
 // +kubebuilder:validation:XValidation:rule=`self.platform.type == "Azure" ? self.services.exists(s, s.service == "OAuthServer" && s.servicePublishingStrategy.type == "Route" && s.servicePublishingStrategy.route.hostname != "") : true`,message="Azure platform requires OAuthServer Route service with a hostname to be defined"
 // +kubebuilder:validation:XValidation:rule=`self.platform.type == "Azure" ? self.services.exists(s, s.service == "Konnectivity" && s.servicePublishingStrategy.type == "Route" && s.servicePublishingStrategy.route.hostname != "") : true`,message="Azure platform requires Konnectivity Route service with a hostname to be defined"
 // +kubebuilder:validation:XValidation:rule=`self.platform.type == "Azure" ? self.services.exists(s, s.service == "Ignition" && s.servicePublishingStrategy.type == "Route" && s.servicePublishingStrategy.route.hostname != "") : true`,message="Azure platform requires Ignition Route service with a hostname to be defined"
+// +kubebuilder:validation:XValidation:rule=`has(self.issuerURL) || !has(self.serviceAccountSigningKey)`,message="If serviceAccountSigningKey is set, issuerURL must be set"
+
```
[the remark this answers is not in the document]
enxebre: I guess the right UX would be to make .name required, as there's no semantic meaning or support for setting the struct with out a .name.

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -351,141 +351,161 @@ const (
...
+	// This is used by the control plane operator to determine and signal the appropriate available upgrades in the hostedCluster.status.
+	// +kubebuilder:validation:XValidation:rule="isURL(self)",message="updateService must be a valid absolute URL"
 	// +optional
 	UpdateService configv1.URL `json:"updateService,omitempty"`
 
-	// channel is an identifier for explicitly requesting that a non-default
-	// set of updates be applied to this cluster. The default channel will be
-	// contain stable updates that are appropriate for production clusters.
-	//
+	// channel is an identifier for explicitly requesting that a non-default set of updates be applied to this cluster.
```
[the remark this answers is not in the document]
enxebre: updated and added a todo consider the backend to use "Default" channel when this is omitted

api/hypershift/v1beta1/hosted_controlplane.go
```diff
@@ -117,6 +117,8 @@ type HostedControlPlaneSpec struct {
 
 	// Services defines metadata about how control plane services are published
 	// in the management cluster.
+	// +kubebuilder:validation:MaxItems=6
+	// +kubebuilder:validation:MinItems=4
```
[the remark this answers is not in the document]
enxebre: This is explained in hostedCluster.Services.
hosted control plane API is beyond the scope of this PR, this change was a side effect of some shared types to prevent hosted control plane API from breaking error cost budget

api/hypershift/v1beta1/hostedcluster_types.go
enxebre: nice, does isCIDR support ipv6 format?

enxebre: Seems it does based on https://github.com/openshift/api/blob/a2817b89f7e0989016967055a142d5e88bed18de/operator/v1/types_network.go#L487

enxebre: See https://github.com/openshift/hypershift/pull/5090#discussion_r1836890598

api/hypershift/v1beta1/hostedcluster_types.go
enxebre: Ideally this would be a pointer and the struct would have a required field .Name. That captures the valid semantics.
However, that would be now breaking for consumers not setting the field .name which has a valid code path.
By making it optional we at least allow the use case for serialize resources i.e. yaml to be applied without the field.

api/hypershift/v1beta1/hostedcluster_types.go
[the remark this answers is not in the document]
enxebre: That would be intrusive thought for other fields using the ref as a non pointer. https://github.com/openshift/hypershift/pull/5090#discussion_r1836992857

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -351,141 +351,161 @@ const (
...
+	// The default channel will contain stable updates that are appropriate for production clusters.
+	// +kubebuilder:validation:MaxLength=100
+	// +kubebuilder:validation:MinLength=1
 	// +optional
 	Channel string `json:"channel,omitempty"`
```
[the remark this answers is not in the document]
enxebre: cc @wking @LalatenduMohanty is there any input validation that can be done for channel?

api/hypershift/v1beta1/hostedcluster_types.go
enxebre: Thanks for pointing this out, yes this is tech debt. I'll revisit this in a different PR, this iteration is getting big enough

api/util/ipnet/ipnet.go
enxebre: Discussed offline this is available in vanilla kube from 1.31 API servers and in Openshift from 4.16.
There are use cases that consume hypershift using non Openshift versions before 1.31.
So I'll punt on this for now and add a todo
@JoelSpeed Do you have a link to the proposal where support for this was added in upstream?

api/hypershift/v1beta1/hostedcluster_types.go
enxebre: Not really, I don't think so. I'll added a TODO to revisit "breaking" changes that consolidate pointer vs non-pointer

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -703,76 +743,124 @@ type NodePortPublishingStrategy struct {
...
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:validation:MinLength=1
 	// +optional
 	Hostname string `json:"hostname,omitempty"`
 }
 
 // RoutePublishingStrategy specifies options for exposing a service as a Route.
 type RoutePublishingStrategy struct {
-	// Hostname is the name of the DNS record that will be created pointing to the Route.
+	// Hostname is the name of the DNS record that will be created pointing to the Route and passed through to consumers of the service.
```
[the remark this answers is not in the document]
bryan-cox: It looks like this still needs done.

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -351,152 +351,178 @@ const (
...
+	// Possible values are HighlyAvailable and SingleReplica. The default value is SingleReplica.
 	// +optional
 	// +kubebuilder:default:="SingleReplica"
-	// +immutable
 	InfrastructureAvailabilityPolicy AvailabilityPolicy `json:"infrastructureAvailabilityPolicy,omitempty"`
 
-	// DNS specifies DNS configuration for the cluster.
-	//
-	// +immutable
+	// dns specifies the DNS configuration for the hosted cluster ingress.
```
muraee, line 436: missing `+optional`

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -351,152 +351,178 @@ const (
...
 	DNS DNSSpec `json:"dns,omitempty"`
 
-	// Networking specifies network configuration for the cluster.
-	//
-	// +immutable
+	// networking specifies network configuration for the hosted cluster.
+	// Defaults to OVNKubernetes with a cluster network of cidr: "10.132.0.0/14" and a service network of cidr: "172.31.0.0/16".
 	// +kubebuilder:default={networkType: "OVNKubernetes", clusterNetwork: {{cidr: "10.132.0.0/14"}}, serviceNetwork: {{cidr: "172.31.0.0/16"}}}
 	Networking ClusterNetworking `json:"networking"`
```
muraee, line 442: missing `+optional` or `+required`

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -351,152 +351,178 @@ const (
...
-	// Etcd specifies configuration for the control plane etcd cluster. The
-	// default ManagementType is Managed. Once set, the ManagementType cannot be
+	// etcd specifies configuration for the control plane etcd cluster. The
+	// default managementType is Managed. Once set, the managementType cannot be
 	// changed.
 	//
-	// +kubebuilder:validation:Optional
 	// +kubebuilder:default={managementType: "Managed", managed: {storage: {type: "PersistentVolume", persistentVolume: {size: "8Gi"}}}}
 	// +immutable
 	Etcd EtcdSpec `json:"etcd"`
```
muraee, line 456: missing `+optional` or `+required`

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -619,34 +655,46 @@ type ImageContentSource struct {
...
+	// type is the publishing strategy used for the service.
+	// It can be LoadBalancer;NodePort;Route;None;S3
 	//
 	// +kubebuilder:validation:Enum=LoadBalancer;NodePort;Route;None;S3
-	// +immutable
+	// +required
 	Type PublishingStrategyType `json:"type"`
 
-	// NodePort configures exposing a service using a NodePort.
+	// nodePort configures exposing a service using a NodePort.
```
muraee, line 690: missing `+optional`  same for the other strategy fields

api/hypershift/v1beta1/nodepool_types.go
```diff
@@ -142,17 +141,20 @@ type NodePoolSpec struct {
 	// * ImageDigestMirrorSet
 	//
 	// This is validated in the backend and signaled back via validMachineConfig condition.
+	// Changing this field will trigger a NodePool rollout.
```
muraee, line 144: should we add `+rollout` here as in the HostedCluster API?

bryan-cox:
There are additional linters we can add in the future if we want - https://golangci-lint.run/usage/linters/

api/util/ipnet/ipnet.go
JoelSpeed: No design doc no, but added in https://github.com/kubernetes/kubernetes/pull/121912

enxebre:
/override Red Hat Konflux / hypershift-operator-main-on-pull-request

enxebre:
/override Red Hat Konflux / hypershift-operator-main-on-pull-request

enxebre:

rtheis (commented):
Thanks for the heads up on this PR.  I don't have any concerns at the moment.

PR #5091, opened 2024-11-08
OCPBUGS-44375: openstack: make external network ID really optional

EmilienM:

When we introduced the support for providing the external network ID, we
used pointers and this is causing pointer nil issues when the user
doesn't provide the external network ID (which is a valid use case).

For that, I refactored a few function's signatures (and I did it in the
same way as it was done for the v2 controllers to ease readability).

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/cloud/openstack/providerconfig.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/cloud/openstack/providerconfig.go
@@ -3,6 +3,7 @@ package openstack
 import (
 	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/utils/ptr"
 )
 
 const (
@@ -19,11 +20,11 @@ const (
 // ReconcileCloudConfigSecret reconciles the cloud config secret.
 // For some controllers (e.g. Manila CSI, CNCC, etc), the cloud config needs to be stored in a secret.
 // In the hosted cluster config operator, we create the secrets needed by these controllers.
-func ReconcileCloudConfigSecret(externalNetworkID *string, secret *corev1.Secret, cloudName string, credentialsSecret *corev1.Secret, caCertData []byte) error {
+func ReconcileCloudConfigSecret(platformSpec *hyperv1.OpenStackPlatformSpec, secret *corev1.Secret, credentialsSecret *corev1.Secret, caCertData []byte) error {
 	if secret.Data == nil {
 		secret.Data = map[string][]byte{}
 	}
-	config := getCloudConfig(externalNetworkID, cloudName, credentialsSecret, caCertData)
+	config := getCloudConfig(platformSpec, credentialsSecret, caCertData)
 	if caCertData != nil {
 		secret.Data[CABundleKey] = caCertData
 	}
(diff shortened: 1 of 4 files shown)


EmilienM:
/cherry-pick release-4.17
It'll probably fail in merge conflict but at least it's my reminder to do it.

EmilienM:
/test e2e-openstack
weird failures

control-plane-operator/controllers/hostedcontrolplane/cloud/openstack/providerconfig.go
```diff
@@ -48,20 +48,20 @@ func ReconcileCloudConfigConfigMap(externalNetworkID *string, cm *corev1.ConfigM
...
+	config += "cloud=" + platformSpec.IdentityRef.CloudName + "\n"
 	// This takes priority over the 'cacert' value in 'clouds.yaml' and we therefore
 	// unset then when creating the initial secret.
 	if caCertData != nil {
 		config += "ca-file=" + CaDir + "/" + CABundleKey + "\n"
 	}
 	config += "\n[LoadBalancer]\nmax-shared-lb = 1\nmanage-security-groups = true\n"
-	if externalNetworkID != nil && *externalNetworkID != "" {
-		config += "floating-network-id = " + *externalNetworkID + "\n"
+	if platformSpec.ExternalNetwork != nil && *platformSpec.ExternalNetwork.ID != "" {
```
muraee, line 63: you are dereferencing `ID` which is a pointer, without checking if its nil first. you can use ptr.Deref() as done in v2 controller here:
https://github.com/openshift/hypershift/blob/2b10a4454e3c6a0d8359873b10852e9337a17ffe/control-plane-operator/controllers/hostedcontrolplane/v2/cloud_controller_manager/openstack/config.go#L86-L87

EmilienM: great advise, thanks a lot

muraee: you still need `if platformSpec.ExternalNetwork != nil`

control-plane-operator/controllers/hostedcontrolplane/cloud/openstack/providerconfig.go
```diff
@@ -48,20 +49,23 @@ func ReconcileCloudConfigConfigMap(externalNetworkID *string, cm *corev1.ConfigM
...
 	// This takes priority over the 'cacert' value in 'clouds.yaml' and we therefore
 	// unset then when creating the initial secret.
 	if caCertData != nil {
 		config += "ca-file=" + CaDir + "/" + CABundleKey + "\n"
 	}
 	config += "\n[LoadBalancer]\nmax-shared-lb = 1\nmanage-security-groups = true\n"
-	if externalNetworkID != nil && *externalNetworkID != "" {
-		config += "floating-network-id = " + *externalNetworkID + "\n"
+	if platformSpec.ExternalNetwork != nil {
+		externalNetworkID := ptr.Deref(platformSpec.ExternalNetwork.ID, "")
```
stephenfin, line 65: This isn't for now, given it's already an issue, but this means we just ignore `ExternalNetwork` if `Filter` is set, right? Odd that we even supported that in the first place if so. I'm guessing it's inherited from CAPO.

stephenfin (commented):
/lgtm

One comment/question inline but not blocking (or even directly related to this)

control-plane-operator/controllers/hostedcontrolplane/cloud/openstack/providerconfig.go
EmilienM: Right now we don't support anything but the ID indeed. We supported the name for little time and realized this was unsafe in this case.
I think for now it's ok.

PR #5092, opened 2024-11-08
OCPBUGS-44041: add flags for passing in ManageIdentities on Azure HC …

Patryk-Stefanski:
**What this PR does / why we need it**: Adds a managedIdentitiesFile flag to azure create cluster so that we can pass in pre created certnames and client ids when creating hostedclusters. This allows for avoiding the az and jq dependency.

Fixes # [OCPBUGS-44041](https://issues.redhat.com/browse/OCPBUGS-44041)

**Checklist**


cmd/cluster/azure/create.go
```diff
@@ -210,11 +217,6 @@ func (o *ValidatedCreateOptions) Complete(ctx context.Context, opts *core.Create
 		if err != nil {
 			return nil, fmt.Errorf("failed to create infra: %w", err)
 		}
-
```
Patryk-Stefanski, line 213: deleted this becasue its now done as part of infra.run function

PR #5093, opened 2024-11-08
OCPBUGS-44438: Remove non-matching feature-gated CVO manifests from payload 

petr-muller:

CVO manifests contain some feature-gated ones:

- since at least 4.16, there are feature-gated `ClusterVersion` CRDs
- `UpdateStatus` feature is delivered through DevPreview (now) and TechPreview (later) feature set

We observed HyperShift CI jobs to fail when adding DevPreview-gated deployment manifests, which was unexpected. Investigating further, we discovered that HyperShift applies these manifests:

[cluster-version-operator-665c5789d5-8sr59-bootstrap.log](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-version-operator/1091/pull-ci-openshift-cluster-version-operator-master-e2e-hypershift-conformance/1853764751801192448/artifacts/e2e-hypershift-conformance/dump/artifacts/namespaces/clusters-c01d0e18fc19f1e0757b/core/pods/logs/cluster-version-operator-665c5789d5-8sr59-bootstrap.log):

```
error: error parsing /var/payload/manifests/0000_00_update-status-controller_03_deployment-DevPreviewNoUpgrade.yaml: error converting YAML to JSON: yaml: invalid map key: map[interface {}]interface {}{".ReleaseImage":interface {}(nil)}
```

But even without these added manifests, this happens for existing `ClusterVersion` CRD manifests present in the payload:

```console
$ ls -1 manifests/*clusterversions*crd.yaml
manifests/0000_00_cluster-version-operator_01_clusterversions-CustomNoUpgrade.crd.yaml
manifests/0000_00_cluster-version-operator_01_clusterversions-Default.crd.yaml
manifests/0000_00_cluster-version-operator_01_clusterversions-DevPreviewNoUpgrade.crd.yaml
manifests/0000_00_cluster-version-operator_01_clusterversions-TechPreviewNoUpgrade.crd.yaml
```

In a passing HyperShift CI job, the [same log](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-version-operator/1105/pull-ci-openshift-cluster-version-operator-master-e2e-hypershift-conformance/1854708481941049344/artifacts/e2e-hypershift-conformance/dump/artifacts/namespaces/clusters-a5d1b5c3fcb2445935f2/core/pods/logs/cluster-version-operator-96cdfbf7c-cxt9r-bootstrap.log) shows that all four manifests are applied instead of just one:

```
customresourcedefinition.apiextensions.k8s.io/clusterversions.config.openshift.io created
customresourcedefinition.apiextensions.k8s.io/clusterversions.config.openshift.io configured
customresourcedefinition.apiextensions.k8s.io/clusterversions.config.openshift.io configured
customresourcedefinition.apiextensions.k8s.io/clusterversions.config.openshift.io configured
```

This likely means that HyperShift hosted clusters end up using TechPreviewNoUpgrade `ClusterVersion` CRD?

The proper fix is probably to wire through the `FeatureGate` with desired featureset through `CVOParams` but we would need a bit of selection logic so for now we can just remove entropy by deleting all feature-gated manifests instead of stumbling at them.

OCPBUGS-44438

**Checklist**

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/cvo/params.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/cvo/params.go
@@ -5,6 +5,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/utils/ptr"
 
+	configv1 "github.com/openshift/api/config/v1"
 	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
 	"github.com/openshift/hypershift/control-plane-operator/controllers/hostedcontrolplane/imageprovider"
 
@@ -21,6 +22,7 @@ type CVOParams struct {
 	OwnerRef                config.OwnerRef
 	DeploymentConfig        config.DeploymentConfig
 	PlatformType            hyperv1.PlatformType
+	FeatureSet              configv1.FeatureSet
 }
 
 func NewCVOParams(hcp *hyperv1.HostedControlPlane, releaseImageProvider imageprovider.ReleaseImageProvider, setDefaultSecurityContext, enableCVOManagementClusterMetricsAccess bool) *CVOParams {
@@ -37,6 +39,9 @@ func NewCVOParams(hcp *hyperv1.HostedControlPlane, releaseImageProvider imagepro
 	// This could happen for example in local dev enviroments if the "OPERATE_ON_RELEASE_IMAGE" env variable is not set.
 	if p.ReleaseImage == "" {
 		p.ReleaseImage = hcp.Spec.ReleaseImage
+	}
+	if hcp.Spec.Configuration != nil && hcp.Spec.Configuration.FeatureGate != nil {
+		p.FeatureSet = hcp.Spec.Configuration.FeatureGate.FeatureSet
 	}
 
 	if enableCVOManagementClusterMetricsAccess {
--- a/control-plane-operator/controllers/hostedcontrolplane/cvo/reconcile.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/cvo/reconcile.go
@@ -115,7 +115,7 @@ func cvoLabels() map[string]string {
 
 var port int32 = 8443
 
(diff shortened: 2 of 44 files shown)


petr-muller:
/hold

Holding until branch cut

petr-muller:
I have adopted Seth's approach from https://github.com/openshift/hypershift/pull/5096. It is closer to desired behavior than my original proposal, I just tweaked it so it removes all non-matching manifests, instead of relying on a `Default` always present and the feature set one applied over it (which would not happen for `CustomNoUpgrade`).

muraee:
please also add those changes here:
https://github.com/openshift/hypershift/blob/da24e17fcccb7d6e35cfcda584938c8ba48a5905/control-plane-operator/controllers/hostedcontrolplane/v2/cvo/deployment.go#L124

petr-muller:
@muraee done, thank you!

I have also added an iteration to the `TestControlPlaneComponents` to test the control plane generated for `TechPreviewNoUpgrade` featureSet. I populated the fixtures with exact copies of the default ones, and then in a separate commit added the actual test and updated the fixture, so the commit clearly shows how the TP ones differs from the default ones.

petr-muller:
Refactored to include https://github.com/openshift/hypershift/pull/5188

petr-muller:
Had to rebase to pull fixture changes from https://github.com/openshift/hypershift/pull/5160/files#diff-33e09167c08ede07364a11e60abb341ce6e4141143fb1a7033c8af966f2c32aa to the TechPreview ones

petr-muller:
/test e2e-aws-upgrade-hypershift-operator

Failed just on destroy :crossed_fingers:

petr-muller:
LOL I was not expecting Konflux to piggyback on Prow commands

PR #5095, opened 2024-11-10
OCPBUGS-44413: add missing assets and RBAC for OpenStack with newer CAPO

EmilienM:
**Why we need this PR**:

We recently bumped CAPO (the CAPI provider for OpenStack) to v0.11 and we didn't update the assets correctly for a new dependency: [ORC](https://github.com/k-orc/openstack-resource-controller).

**What this PR does**:

* Add ORC image to support/api/scheme, it's a dependency of CAPO v0.11 which was recently bumped here.
* Run `go mod tidy && go mod vendor` so ORC is now vendored.
* Update the `Makefile` to generate the assets for ORC within the CAPO directory.
* Run `make update` to generate new assets.
* Add the new ORC resource to `dump` command to ease debug.
* Add RBAC rules to allow `List` and `Watch` the `Images` resource on `openstack.k-orc.cloud`, based on these CAPO [logs](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/logs/periodic-ci-shiftstack-ci-release-4.18-e2e-openstack-dualstack-v6primary/1855288533627965440/artifacts/e2e-openstack-dualstack-v6primary/ipi-install-install/artifacts/.openshift_install-1731174496.log) (note: the rules are added in the Hypershift Operator, in the CLI and in the Hypershift HostedCluster controller, this is expected so they are created early and reconciled by the controller to ensure upgrades if any):
```
failed to list *v1alpha1.Image: images.openstack.k-orc.cloud is forbidden: User "system:serviceaccount:e2e-clusters-5bc8z-example-fgd4z:capi-provider" cannot list resource "images" in API group "openstack.k-orc.cloud" in the namespace "e2e-clusters-5bc8z-example-fgd4z"
failed to watch *v1alpha1.Image: failed to list *v1alpha1.Image: images.openstack.k-orc.cloud is forbidden: User "system:serviceaccount:e2e-clusters-5bc8z-example-fgd4z:capi-provider" cannot list resource "images" in API group "openstack.k-orc.cloud" in the namespace "e2e-clusters-5bc8z-example-fgd4z"
```
The verbs will likely evolve once we use ORC more extensively, to handle the images for example.
* `testdata` fixtures were updated by running `UPDATE=true go test ./...`.

Note: no backport is required for this PR as the bump happened in 4.18 only.


go.mod
```diff
@@ -37,6 +37,7 @@ require (
 	github.com/google/gofuzz v1.2.0
 	github.com/google/uuid v1.6.0
 	github.com/grpc-ecosystem/go-grpc-prometheus v1.2.0
+	github.com/k-orc/openstack-resource-controller v0.0.0-00010101000000-000000000000
```
bryan-cox, line 40: Is there not a tag version we could point to here?

pierreprinetti: not yet. Also as you can see line 273 (formerly 272) of this same `go.mod`, the requirement is `replace`d because the code for openstack-resource-controller temporarily lives in the cluster-api-provider-openstack repository

bryan-cox: Ok, just wanted to double check

EmilienM:
@Patryk-Stefanski e2e-aws-4-17 is still unstable it seems :confused: - would it be possible to override so we can get this landed?

PR #5103, opened 2024-11-12
NO-JIRA: docs/openstack: clarify CLI

EmilienM:

Add missing hypershift CLI and clarify why we need it.


mandre (commented):
/lgtm

PR #5116, opened 2024-11-13
NO-JIRA: docs: minor fix for OpenStack

EmilienM:

Minor fix on how we get kubeadmin password.


bryan-cox:
/retitle NO-JIRA: docs: minor fix for OpenStack

PR #5126, opened 2024-11-14
OCPBUGS-44476: Use ingress role in private link controller for DNS operations

csrwng:
Before this commit, the private link controller was assuming the same role for all cloud operations in a shared VPC infrastructure. In order to support the topology in which private hosted zones are created in the cluster account and not in the VPC owner account, the private link controller will now use the ingress role for route53 operations and only use the shared control plane operator role for vpc endpoint operations. This allows customers to create the ingress role in the account where the private hosted zones live, and the ingress role can be used to manage dns records by both the ingress operator and the private link controller.

Fixes #[OCPBUGS-44476](https://issues.redhat.com/browse/OCPBUGS-44476)

**Checklist**


bryan-cox:
/lgtm
/hold

for others to take a look

csrwng:
aks flake looks like due to `TestNTOMachineConfigGetsRolledOut`
/retest-required
/hold cancel

PR #5138, opened 2024-11-15
NO-JIRA: openstack/e2e: allow to change AZ name

EmilienM:

In some environments, we'll want to change the name of the availability
zone where the nodepool AZ will be tested on.
Let's allow that with this change.

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -564,9 +564,8 @@ func (p *options) DefaultOpenStackOptions() hypershiftopenstack.RawCreateOptions
 		OpenStackExternalNetworkID: p.configurableClusterOptions.OpenStackExternalNetworkID,
 		NodePoolOpts: &openstacknodepool.RawOpenStackPlatformCreateOptions{
 			OpenStackPlatformOptions: &openstacknodepool.OpenStackPlatformOptions{
-				Flavor:         p.configurableClusterOptions.OpenStackNodeFlavor,
-				ImageName:      p.configurableClusterOptions.OpenStackNodeImageName,
-				AvailabityZone: p.configurableClusterOptions.OpenStackNodeAvailabilityZone,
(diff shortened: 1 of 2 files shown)


test/e2e/e2e_test.go
```diff
@@ -562,9 +562,8 @@ func (p *options) DefaultOpenStackOptions() hypershiftopenstack.RawCreateOptions
 		OpenStackExternalNetworkID: p.configurableClusterOptions.OpenStackExternalNetworkID,
 		NodePoolOpts: &openstacknodepool.RawOpenStackPlatformCreateOptions{
 			OpenStackPlatformOptions: &openstacknodepool.OpenStackPlatformOptions{
-				Flavor:         p.configurableClusterOptions.OpenStackNodeFlavor,
-				ImageName:      p.configurableClusterOptions.OpenStackNodeImageName,
-				AvailabityZone: p.configurableClusterOptions.OpenStackNodeAvailabilityZone,
```
EmilienM, line 567: we don't want all Nodepools to have AZ specified, so we remove it from here. This change is expected.

test/e2e/nodepool_osp_az_test.go
```diff
@@ -105,10 +105,17 @@ func (o OpenStackAZTest) BuildNodePoolManifest(defaultNodepool hyperv1.NodePool)
...
-	nodePool.Spec.Platform.OpenStack.AvailabilityZone = defaultAvailabilityZone
+	nodePool.Spec.Platform.OpenStack.AvailabilityZone = getAZName()
 	return nodePool, nil
 }
 
+func getAZName() string {
+	if globalOpts.configurableClusterOptions.OpenStackNodeAvailabilityZone != "" {
+		return globalOpts.configurableClusterOptions.OpenStackNodeAvailabilityZone
+	}
+	return defaultAvailabilityZone
```
EmilienM, line 116: the default AZ "nova" is safe, as it's present in most clouds. Though it's good we can easily override it.

PR #5139, opened 2024-11-15
OSASINFRA-3492: feat(openstack): leverage ORC to manage the release image

EmilienM:
This PR enhances OpenStack support in HyperShift by improving image management
and ORC (OpenStack Resource Controller) integration.

- Added a new `RetentionPolicy` type with `Orphan` and `Prune` options.
- Implemented `ImageRetentionPolicy` for OpenStack platform in `OpenStackPlatformSpec`.
- Updated CRDs to include `imageRetentionPolicy` with a default value of `Prune`.
- Enhanced the OpenStack CLI options to support `--openstack-image-retention-policy`.
- Updated OpenStack resource management to ensure proper handling of Glance images via ORC.
- Removed the mandatory requirement for `--openstack-node-image-name`, allowing ORC to manage RHCOS images dynamically.
- Modified RBAC policies for ORC to ensure proper permissions on `images` resources.
- Refactored related test cases and documentation to align with the new behavior.

This change improves flexibility in managing OpenStack resources, enabling users to retain or prune images based on policy.

### DIFF
--- a/api/hypershift/v1beta1/hostedcluster_types.go
+++ b/api/hypershift/v1beta1/hostedcluster_types.go
@@ -356,6 +356,21 @@ const (
 	AWSMachinePublicIPs = "hypershift.openshift.io/aws-machine-public-ips"
 )
 
+// RetentionPolicy defines the policy for handling resources associated with a cluster when the cluster is deleted.
+//
+// +kubebuilder:validation:Enum:=Orphan;Prune
+type RetentionPolicy string
+
+const (
+	// OrphanRetentionPolicy will keep the resources associated with the cluster
+	// when the cluster is deleted.
+	OrphanRetentionPolicy RetentionPolicy = "Orphan"
+
+	// PruneRetentionPolicy will delete the resources associated with the cluster
+	// when the cluster is deleted.
+	PruneRetentionPolicy RetentionPolicy = "Prune"
+)
+
 // +kubebuilder:validation:Enum=ImageRegistry
 type OptionalCapability string
 
--- a/api/hypershift/v1beta1/openstack.go
+++ b/api/hypershift/v1beta1/openstack.go
@@ -1,5 +1,7 @@
 package v1beta1
 
+import "fmt"
+
 // PortSecurityPolicy defines whether or not to enable port security on a port.
 type PortSecurityPolicy string
 
@@ -12,7 +14,31 @@ const (
 
 	// PortSecurityDefault uses the default port security policy.
 	PortSecurityDefault PortSecurityPolicy = ""
+
+	// PruneRetentionPolicy is the default policy for handling OpenStack Glance Images
+	// when the HostedCluster is deleted.
+	DefaultImageRetentionPolicy RetentionPolicy = PruneRetentionPolicy
 )
+
+func (p *RetentionPolicy) String() string {
+	return string(*p)
+}
+
+func (p *RetentionPolicy) Set(s string) error {
+	switch s {
+	case string(OrphanRetentionPolicy):
+		*p = OrphanRetentionPolicy
+	case string(PruneRetentionPolicy):
+		*p = PruneRetentionPolicy
+	default:
+		return fmt.Errorf("invalid RetentionPolicy: %s", s)
+	}
+	return nil
+}
+
+func (p *RetentionPolicy) Type() string {
+	return "RetentionPolicy"
+}
 
 type OpenStackNodePoolPlatform struct {
 	// Flavor is the OpenStack flavor to use for the node instances.
@@ -136,6 +162,20 @@ type OpenStackPlatformSpec struct {
 	// +kubebuilder:validation:XValidation:rule="isIP(self)",message="floatingIP must be a valid IPv4 or IPv6 address"
 	// +optional
 	IngressFloatingIP string `json:"ingressFloatingIP,omitempty"`
+
+	// imageRetentionPolicy defines the policy for handling resources associated with the image
+	// when the cluster is deleted.
+	// The default (set by the Nodepool controller) matches OpenStack Resource Controller (ORC)
+	// behavior where the OpenStack Glance Image is pruned when the Image object is deleted during the
+	// HostedCluster deletion.
+	// It is defined at the HostedCluster level and will be used for all nodepools images
+	// so there is no conflict between different ORC objects.
+	// On day 2 operations, if this field is changed, the corresponding ORC object will be updated
+	// to reflect the chosen retention policy (prune or orphan) which is translated into ORC format
+	// (either 'delete' or 'detach' in ORC terminology).
+	//
+	// +optional
+	ImageRetentionPolicy RetentionPolicy `json:"imageRetentionPolicy,omitempty"`
 }
 
 // OpenStackIdentityReference is a reference to an infrastructure
--- a/cmd/cluster/openstack/create.go
+++ b/cmd/cluster/openstack/create.go
@@ -14,7 +14,6 @@ import (
 	"github.com/openshift/hypershift/support/config"
 
 	corev1 "k8s.io/api/core/v1"
-	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -29,7 +28,10 @@ const (
 )
 
 func DefaultOptions() *RawCreateOptions {
-	return &RawCreateOptions{NodePoolOpts: openstacknodepool.DefaultOptions()}
+	return &RawCreateOptions{
+		OpenStackImageRetentionPolicy: hyperv1.PruneRetentionPolicy,
+		NodePoolOpts:                  openstacknodepool.DefaultOptions(),
+	}
 }
 
 func BindOptions(opts *RawCreateOptions, flags *pflag.FlagSet) {
@@ -45,15 +47,17 @@ func bindCoreOptions(opts *RawCreateOptions, flags *pflag.FlagSet) {
 	flags.StringVar(&opts.OpenStackExternalNetworkID, "openstack-external-network-id", opts.OpenStackExternalNetworkID, "ID of the OpenStack external network (optional)")
 	flags.StringVar(&opts.OpenStackIngressFloatingIP, "openstack-ingress-floating-ip", opts.OpenStackIngressFloatingIP, "An available floating IP in your OpenStack cluster that will be associated with the OpenShift ingress port (optional)")
 	flags.StringSliceVar(&opts.OpenStackDNSNameservers, "openstack-dns-nameservers", opts.OpenStackDNSNameservers, "List of DNS nameservers to use for the cluster (optional)")
+	flags.Var(&opts.OpenStackImageRetentionPolicy, "openstack-image-retention-policy", "OpenStack Glance Image retention policy. Valid values are 'Orphan' and 'Prune'. By default images are pruned. (optional)")
 }
 
 type RawCreateOptions struct {
-	OpenStackCredentialsFile   string
-	OpenStackCloud             string
-	OpenStackCACertFile        string
-	OpenStackExternalNetworkID string
-	OpenStackIngressFloatingIP string
-	OpenStackDNSNameservers    []string
+	OpenStackCredentialsFile      string
+	OpenStackCloud                string
+	OpenStackCACertFile           string
+	OpenStackExternalNetworkID    string
+	OpenStackIngressFloatingIP    string
+	OpenStackDNSNameservers       []string
+	OpenStackImageRetentionPolicy hyperv1.RetentionPolicy
 
 	NodePoolOpts *openstacknodepool.RawOpenStackPlatformCreateOptions
 }
@@ -134,6 +138,10 @@ func (o *RawCreateOptions) Validate(ctx context.Context, opts *core.CreateOption
 		return nil, err
 	}
 
+	if err := validateRetentionPolicy(o.OpenStackImageRetentionPolicy); err != nil {
+		return nil, err
+	}
+
 	if opts.ExternalDNSDomain != "" {
 		err := fmt.Errorf("--external-dns-domain is not supported on OpenStack")
 		opts.Log.Error(err, "Failed to create cluster")
@@ -194,6 +202,8 @@ func (o *RawCreateOptions) ApplyPlatformSpecifics(cluster *hyperv1.HostedCluster
 		cluster.Spec.Networking.MachineNetwork = []hyperv1.MachineNetworkEntry{{CIDR: *ipnet.MustParseCIDR(config.DefaultMachineNetwork)}}
 	}
 
+	cluster.Spec.Platform.OpenStack.ImageRetentionPolicy = hyperv1.RetentionPolicy(o.OpenStackImageRetentionPolicy)
+
 	return nil
 }
 
@@ -224,27 +234,6 @@ func (o *CreateOptions) GenerateResources() ([]client.Object, error) {
 
 	resources = append(resources, credentialsSecret)
 
-	resources = append(resources, &rbacv1.Role{
-		TypeMeta: metav1.TypeMeta{
-			Kind:       "Role",
-			APIVersion: rbacv1.SchemeGroupVersion.String(),
-		},
-		ObjectMeta: metav1.ObjectMeta{
-			Namespace: o.namespace,
-			Name:      "capi-provider-role",
-		},
-		// The following rule is required for CAPO to watch for the Images resources created by ORC,
-		// which is a dependency since CAPO v0.11.0.
-		// This rule is also defined in the Hypershift HostedCluster controller and the Hypershift Operator when creating
-		// the cluster.
-		Rules: []rbacv1.PolicyRule{
-			{
-				APIGroups: []string{"openstack.k-orc.cloud"},
-				Resources: []string{"images"},
-				Verbs:     []string{"list", "watch"},
-			},
-		},
-	})
 	return resources, nil
 }
 
@@ -377,4 +366,18 @@ func extractCloud(cloudsYAMLPath, caCertPath, cloudName string) ([]byte, []byte,
 	}
 
 	return cloudsYAML, caCert, nil
+}
+
+// validateRetentionPolicy validates the retention policy is valid
+func validateRetentionPolicy(policy hyperv1.RetentionPolicy) error {
+	if policy == "" {
+		return nil
+	}
+
+	switch policy {
+	case hyperv1.OrphanRetentionPolicy, hyperv1.PruneRetentionPolicy:
+		return nil
+	default:
+		return fmt.Errorf("invalid retention policy: %s", policy)
+	}
 }
--- a/cmd/install/assets/hypershift_operator.go
+++ b/cmd/install/assets/hypershift_operator.go
@@ -1172,14 +1172,11 @@ func (o HyperShiftOperatorClusterRole) Build() *rbacv1.ClusterRole {
 				Resources: []string{"ipaddresses", "ipaddresses/status"},
 				Verbs:     []string{"create", "delete", "get", "list", "update", "watch"},
 			},
-			// The following rule is required for CAPO to watch for the Images resources created by ORC,
-			// which is a dependency since CAPO v0.11.0.
-			// This rule is also defined in the Hypershift HostedCluster controller and the Hypershift CLI when creating
-			// the cluster.
+			// This allows hypershift operator to grant RBAC permissions for ORC to the capi-provider.
 			{
 				APIGroups: []string{"openstack.k-orc.cloud"},
-				Resources: []string{"images"},
-				Verbs:     []string{"list", "watch"},
+				Resources: []string{"images", "images/status"},
+				Verbs:     []string{rbacv1.VerbAll},
 			},
 			{ // This allows the kubevirt csi driver to hotplug volumes to KubeVirt VMs.
 				APIGroups: []string{"subresources.kubevirt.io"},
--- a/docs/content/how-to/openstack/hostedcluster.md
+++ b/docs/content/how-to/openstack/hostedcluster.md
@@ -4,18 +4,19 @@ Once all the [prerequisites](prerequisites.md) are met, it is now possible to cr
 
 Here are the available options specific to the OpenStack platform:
 
-| Option                           | Description                                                                                  | Required | Default       |
-|----------------------------------|----------------------------------------------------------------------------------------------|----------|---------------|
-| `--openstack-ca-cert-file`       | Path to the OpenStack CA certificate file                                                    | No       |               |
-| `--openstack-cloud`              | Name of the cloud in `clouds.yaml`                                                           | No       | `'openstack'` |
-| `--openstack-credentials-file`   | Path to the OpenStack credentials file                                                       | No       |               |
-| `--openstack-external-network-id`| ID of the OpenStack external network                                                         | No       |               |
-| `--openstack-ingress-floating-ip`| A floating IP for OpenShift ingress                                                          | No       |               |
-| `--openstack-node-additional-port`| Attach additional ports to nodes. Params: `network-id`, `vnic-type`, `disable-port-security`, `address-pairs`. | No       |               |
-| `--openstack-node-availability-zone`| Availability zone for the nodepool                                                        | No       |               |
-| `--openstack-node-flavor`        | Flavor for the nodepool                                                                      | Yes      |               |
-| `--openstack-node-image-name`    | Image name for the nodepool                                                                  | Yes      |               |
-| `--openstack-dns-nameservers`    | List of DNS server addresses that will be provided when creating the subnet                  | No       |               |
+| Option                                  | Description                                                                                  | Required | Default       |
+|-----------------------------------------|----------------------------------------------------------------------------------------------|----------|---------------|
+| `--openstack-ca-cert-file`              | Path to the OpenStack CA certificate file                                                    | No       |               |
+| `--openstack-cloud`                     | Name of the cloud in `clouds.yaml`                                                           | No       | `'openstack'` |
+| `--openstack-credentials-file`          | Path to the OpenStack credentials file                                                       | No       |               |
+| `--openstack-external-network-id`       | ID of the OpenStack external network                                                         | No       |               |
+| `--openstack-ingress-floating-ip`       | A floating IP for OpenShift ingress                                                          | No       |               |
+| `--openstack-node-additional-port`      | Attach additional ports to nodes. Params: `network-id`, `vnic-type`, `disable-port-security`, `address-pairs`. | No       |               |
+| `--openstack-node-availability-zone`    | Availability zone for the nodepool                                                           | No       |               |
+| `--openstack-node-flavor`               | Flavor for the nodepool                                                                      | Yes      |               |
+| `--openstack-image-retention-policy`    | OpenStack Glance Image retention policy. Valid values are 'Orphan' and 'Prune'.              | No       | `Prune`       |
+| `--openstack-node-image-name`           | Image name for the nodepool                                                                  | No       |               |
+| `--openstack-dns-nameservers`           | List of DNS server addresses that will be provided when creating the subnet                  | No       |               |
 
 Below is an example of how to create a cluster using environment variables and the `hcp` cli tool.
 
@@ -32,9 +33,6 @@ export WORKER_COUNT="2"
 # OpenStack resources for the HostedCluster will be created
 # in that project.
 export OS_CLOUD="openstack"
-
-# Image name is the name of the image in OpenStack that was pushed in the previous step.
-export IMAGE_NAME="rhcos"
 
 # Flavor for the nodepool
 export FLAVOR="m1.large"
@@ -68,7 +66,6 @@ hcp create cluster openstack \
 --openstack-credentials-file $CLOUDS_YAML \
 --openstack-ca-cert-file $CA_CERT_PATH \
 --openstack-external-network-id $EXTERNAL_NETWORK_ID \
---openstack-node-image-name $IMAGE_NAME \
 --openstack-node-flavor $FLAVOR \
 --openstack-ingress-floating-ip $INGRESS_FLOATING_IP \
 --openstack-dns-nameservers $DNS_NAMESERVERS
@@ -88,6 +85,13 @@ hcp create cluster openstack \
     `PreferredDuringSchedulingIgnoredDuringExecution` mode for `PodAntiAffinity`.
     If your management cluster doesn't have enough workers (less than 3), which is not recommended nor supported,
     you'll need to specify the `--control-plane-availability-policy` flag to `SingleReplica`.
+
+!!! note
+
+    When not providing the `--openstack-node-image-name` flag, the latest RHCOS image will be used.
+    ORC will handle the RHCOS image lifecycle by downloading the image from the OpenShift mirror and deleting it when it's no longer needed.
+    If you want ORC to not delete the image, you can create the HostedCluster with the following option: `--openstack-image-retention-policy=Orphan`.
+    This will prevent ORC from deleting the image resource.
 
 After a few moments we should see our hosted control plane pods up and running:
 
--- a/docs/content/how-to/openstack/prerequisites.md
+++ b/docs/content/how-to/openstack/prerequisites.md
@@ -15,7 +15,6 @@
 * OpenStack Octavia service must be running in the cloud hosting the guest cluster when ingress is configured with an Octavia load balancer.
   In the future, we'll explore other Ingress options like MetalLB.
 * The default external network (on which the kube-apiserver LoadBalancer type service is created) of the Management OCP cluster must be reachable from the guest cluster.
-* The RHCOS image must be uploaded to OpenStack (explained later).
 
 ## Install the HyperShift and HCP CLI
 
@@ -71,11 +70,19 @@ operator-755d587f44-lrtrq   1/1     Running   0          114s
 operator-755d587f44-qj6pz   1/1     Running   0          114s
 ```
 
-## Upload RHCOS image in OpenStack
+## Prepare the management cluster to store etcd locally
 
-For now, we need to manually push an RHCOS image that will be used when deploying the node pools
-on OpenStack. In the [future](https://issues.redhat.com/browse/OSASINFRA-3492), the CAPI provider (CAPO) will handle the RHCOS image
-lifecycle by using the image available in the chosen release payload.
+HostedClusters will have pod(s) for etcd and its performance is essential for the cluster health.
+In production environments, it's required to put etcd data on fast storage and in the case of OpenStack it'll be local storage.
+Follow this [procedure](etcd-local-storage.md) to leverage a well-known and tested solution.
+
+## Upload RHCOS image in OpenStack (optional)
+
+The user can specify which RHCOS image to use when deploying the node pools
+on OpenStack by uploading the image to the OpenStack cloud. If the image is not
+uploaded to OpenStack, [OpenStack Resource Controller (ORC)](https://github.com/k-orc/openstack-resource-controller) will
+manage the RHCOS image lifecycle by downloading the image from the OpenShift mirror and deleting it when it's no longer needed
+or left as an orphan image that can be re-used by other clusters and deleted when it's no longer needed.
 
 Here is an example of how to upload an RHCOS image to OpenStack:
 
@@ -87,12 +94,6 @@ openstack image create --disk-format qcow2 --file rhcos-openstack.x86_64.qcow2 r
 
     The `rhcos-openstack.x86_64.qcow2` file is the RHCOS image that was downloaded from the OpenShift mirror.
     You can download the latest RHCOS image from the [Red Hat OpenShift Container Platform mirror](https://mirror.openshift.com/pub/openshift-v4/dependencies/rhcos/).
-
-## Prepare the management cluster to store etcd locally
-
-HostedClusters will have pod(s) for etcd and its performance is essential for the cluster health.
-In production environments, it's required to put etcd data on fast storage and in the case of OpenStack it'll be local storage.
-Follow this [procedure](etcd-local-storage.md) to leverage a well-known and tested solution.
 
 ## Create a floating IP for the Ingress (optional)
 
--- a/hypershift-operator/controllers/hostedcluster/internal/platform/openstack/openstack.go
+++ b/hypershift-operator/controllers/hostedcluster/internal/platform/openstack/openstack.go
@@ -28,6 +28,12 @@ import (
 
 const (
 	sgRuleDescription = "Managed by the Hypershift Control Plane Operator"
+
+	// OpenStackResourceControllerLogLevel is the log level for OpenStack resource controller running
+	// alongside the Cluster API provider for OpenStack. The log level is set to 4 (info) by default.
+	// until we have a better way to control the log level.
+	// See https://github.com/k-orc/openstack-resource-controller/issues/156
+	OpenStackResourceControllerLogLevel = "4"
 )
 
 type OpenStack struct {
@@ -272,6 +278,9 @@ func (a OpenStack) CAPIProviderDeploymentSpec(hcluster *hyperv1.HostedCluster, _
 				"--namespace=$(MY_NAMESPACE)",
 				"--leader-elect",
 				"--health-probe-bind-address=:8081",
+				// Workaround until we address better logging in ORC:
+				// https://github.com/k-orc/openstack-resource-controller/issues/156
+				"--zap-log-level=" + OpenStackResourceControllerLogLevel,
 			},
 			SecurityContext: &corev1.SecurityContext{
 				AllowPrivilegeEscalation: &allowPrivilegeEscalation,
@@ -416,14 +425,12 @@ func (a OpenStack) CAPIProviderPolicyRules() []rbacv1.PolicyRule {
 			Resources: []string{"ipaddresses", "ipaddresses/status"},
 			Verbs:     []string{"create", "delete", "get", "list", "update", "watch"},
 		},
-		// The following rule is required for CAPO to watch for the Images resources created by ORC,
-		// which is a dependency since CAPO v0.11.0.
-		// This rule is also defined in the Hypershift Operator and the Hypershift CLI when creating
-		// the cluster.
+		// We deploy ORC in the same namespace as the CAPI provider, so we need to create the
+		// necessary RBAC policy rules for ORC to manage the OpenStack resources.
 		{
 			APIGroups: []string{"openstack.k-orc.cloud"},
-			Resources: []string{"images"},
-			Verbs:     []string{"list", "watch"},
+			Resources: []string{"images", "images/status"},
+			Verbs:     []string{rbacv1.VerbAll},
 		},
 	}
 }
--- a/hypershift-operator/controllers/nodepool/capi.go
+++ b/hypershift-operator/controllers/nodepool/capi.go
@@ -846,7 +846,7 @@ func (c *CAPI) machineTemplateBuilders() (client.Object, func(object client.Obje
 	case hyperv1.OpenStackPlatform:
 		template = &capiopenstackv1beta1.OpenStackMachineTemplate{}
 		var err error
-		machineTemplateSpec, err = openstack.MachineTemplateSpec(hcluster, nodePool)
+		machineTemplateSpec, err = openstack.MachineTemplateSpec(hcluster, nodePool, c.releaseImage)
 		if err != nil {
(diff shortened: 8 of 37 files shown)


support/releaseinfo/releaseinfo.go
```diff
@@ -69,14 +69,20 @@ type CoreOSImages struct {
...
 
 type CoreAzureDisk struct {
 	Release string `json:"release"`
 	URL     string `json:"url"`
 }
 
+type CoreOpenStackDisk struct {
+	Release string `json:"release"`
+	URL     string `json:"url"`
+}
```
EmilienM, line 84: I want to add the Hash as well.

EmilienM:
/cc bryan-cox 
since you worked on that topic.

hypershift-operator/controllers/hostedcluster/internal/platform/openstack/openstack.go
```diff
@@ -66,9 +69,66 @@ func (a OpenStack) ReconcileCAPIInfraCR(ctx context.Context, client client.Clien
...
+		return nil, err
+	}
+
 	return openStackCluster, nil
 }
 
+func reconcileOpenStackImageCR(ctx context.Context, client client.Client, createOrUpdate upsert.CreateOrUpdateFN, hcluster *hyperv1.HostedCluster, controlPlaneNamespace string) error {
+	openStackImage := orc.Image{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "rhcos" + hcluster.Name,
```
mdbooth, line 84:
```suggestion
			Name:      "rhcos-" + hcluster.Name,
```

mdbooth: We may also want to name it by the release image version so it can be shared between clusters 🤔

hypershift-operator/controllers/hostedcluster/internal/platform/openstack/openstack.go
```diff
@@ -66,9 +69,66 @@ func (a OpenStack) ReconcileCAPIInfraCR(ctx context.Context, client client.Clien
...
+			Namespace: controlPlaneNamespace,
+		},
+		Spec: orc.ImageSpec{},
+	}
+
+	if _, err := createOrUpdate(ctx, client, &openStackImage, func() error {
```
mdbooth, line 90: TODO: check what createOrUpdate() does. Given that we have applyconfigs it's trivial for us to SSA these, so there's really no need for a function that actually does create or update.

hypershift-operator/controllers/hostedcluster/internal/platform/openstack/openstack.go
```diff
@@ -445,3 +505,39 @@ func defaultWorkerSecurityGroupRules(machineCIDRs []string) []capo.SecurityGroup
...
+	if err != nil {
+		return rhcosImageURL, rhcosImageRelease, fmt.Errorf("failed to lookup release image: %w", err)
+	}
+
+	// We need to translate amd64 to x86_64 since that is what is in the release image stream
+	if _, ok := releaseImage.StreamMetadata.Architectures[hyperv1.ArchAliases[supportedArch]]; !ok {
+		return rhcosImageURL, rhcosImageRelease, fmt.Errorf("arch does not exist in release image, arch: %s", supportedArch)
+	}
+
+	rhcosImageURL = releaseImage.StreamMetadata.Architectures[hyperv1.ArchAliases[supportedArch]].RHCOS.OpenStackDisk.URL
```
mdbooth, line 532: I'm 99% sure we can grab the SHA256s from here, too. It's probably in OpenStackDisk.

hypershift-operator/controllers/nodepool/openstack/openstack.go
```diff
@@ -20,12 +18,9 @@ func MachineTemplateSpec(hcluster *hyperv1.HostedCluster, nodePool *hyperv1.Node
...
 		}
 	} else {
-		// TODO(emilien): Add support for using the image from the release payload.
-		// This will be possible when CAPO supports managing images in the OpenStack cluster:
-		// https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/2130
-		// For 4.17 we might leave this as is and let the user provide the image name as
-		// we plan to deliver the OpenStack provider as a dev preview.
-		return nil, fmt.Errorf("image name is required")
+		openStackMachineTemplate.Template.Spec.Image.ImageRef = &capiopenstackv1beta1.ResourceReference{
+			Name: "rhcos" + hcluster.Name,
```
mdbooth, line 22:
```suggestion
			Name: "rhcos-" + hcluster.Name,
```

(Or version number)

support/releaseinfo/releaseinfo.go
```diff
@@ -69,14 +69,20 @@ type CoreOSImages struct {
 }
 
 type CoreRHCOSImage struct {
-	AzureDisk CoreAzureDisk `json:"azure-disk"`
+	AzureDisk     CoreAzureDisk     `json:"azure-disk"`
+	OpenStackDisk CoreOpenStackDisk `json:"openstack-disk"`
```
mdbooth, line 73: Are you sure this is required? Doesn't seem to be used. What do you expect to go here?

mdbooth (commented):
Either needs wiring up to nodepool to use imageRef (preferrable), or we need to add a wait loop somewhere in here to ensure that the image is Available before creating any machines referencing it.

We don't want to do that second one, so... we should add an imageRef to Nodepool.

hypershift-operator/controllers/hostedcluster/internal/platform/openstack/openstack.go
```diff
@@ -498,3 +505,39 @@ func defaultWorkerSecurityGroupRules(machineCIDRs []string) []capo.SecurityGroup
 
 	return ingressRules
 }
+
+// lookupRHCOSImage looks up a release image and extracts the RHCOS image URL and release version
+func lookupRHCOSImage(ctx context.Context, client client.Client, hcluster *hyperv1.HostedCluster) (string, string, error) {
```
bryan-cox, line 510: We should just leverage the existing `lookupRHCOSImage` and make it an exported function. We could pass in a platform field to differentiate between OSP and Azure.

hypershift-operator/controllers/hostedcluster/internal/platform/openstack/openstack.go
EmilienM: I don't know if the resource could be shared across multiple clusters, that's an interesting idea. I'll investigate for sure.

hypershift-operator/controllers/hostedcluster/internal/platform/openstack/openstack.go
EmilienM: From the doc:

```
// CreateOrUpdate is a copy of controllerutil.CreateOrUpdate with
// an important difference: It copies a number of fields from the object
// on the server to the mutated object if unset in the latter. This
// avoids unnecessary updates when our code sets a whole struct that
// has fields that get defaulted by the server.
```

hypershift-operator/controllers/nodepool/capi.go
```diff
@@ -67,6 +67,13 @@ func (c *CAPI) Reconcile(ctx context.Context) error {
 		return err
 	}
 
+	// Reconcile ORC resources
+	if nodePool.Spec.Platform.Type == hyperv1.OpenStackPlatform {
```
EmilienM, line 71: I don't know if this pattern will be accepted. Suggestions are welcome!

hypershift-operator/controllers/nodepool/openstack/openstack.go
```diff
@@ -72,3 +77,54 @@ func MachineTemplateSpec(hcluster *hyperv1.HostedCluster, nodePool *hyperv1.Node
 	}
 	return openStackMachineTemplate, nil
 }
+
+func ReconcileOpenStackImageCR(ctx context.Context, client client.Client, createOrUpdate upsert.CreateOrUpdateFN, hcluster *hyperv1.HostedCluster, controlPlaneNamespace string, release *releaseinfo.ReleaseImage) error {
+	openStackImage := orc.Image{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "rhcos-" + hcluster.Name,
```
EmilienM, line 84: TODO: name

hypershift-operator/controllers/nodepool/openstack/openstack.go
```diff
@@ -72,3 +77,54 @@ func MachineTemplateSpec(hcluster *hyperv1.HostedCluster, nodePool *hyperv1.Node
...
+	}
+
+	openStackImageSpec.CloudCredentialsRef = orc.CloudCredentialsReference{
+		SecretName: hcluster.Spec.Platform.OpenStack.IdentityRef.Name,
+		CloudName:  hcluster.Spec.Platform.OpenStack.IdentityRef.CloudName,
+	}
+
+	openStackImageSpec.Resource = &orc.ImageResourceSpec{
+		// THIS IS NOT GOOD, NEEDS TO BE FIXED (something related to the image itself should be used, like version)
+		Name: "rhcos-" + hcluster.Name,
```
EmilienM, line 115: TODO: name

hypershift-operator/controllers/nodepool/kubevirt/kubevirt.go
```diff
@@ -54,7 +54,7 @@ func defaultImage(releaseImage *releaseinfo.ReleaseImage) (string, string, error
 	return containerImage, split[1], nil
 }
 
-func unsupportedOpenstackDefaultImage(releaseImage *releaseinfo.ReleaseImage) (string, string, error) {
+func UnsupportedOpenstackDefaultImage(releaseImage *releaseinfo.ReleaseImage) (string, string, error) {
```
EmilienM, line 57: We probably want to move that function to utils or something

hypershift-operator/controllers/nodepool/capi.go
EmilienM: @bryan-cox I realized that we probably need to handle the OpenStack RHCOS image by Nodepool, since the release image can be overridden per Nodepool?

EmilienM:
For myself: we need to update RBAC:

```
{"level":"error","ts":"2024-11-18T21:20:51Z","msg":"Failed to reconcile NodePool","controller":"nodepool","controllerGroup":"hypershift.openshift.io","controllerKind":"NodePool","NodePool":{"name":"example-n42fg","namespace":"e2e-clusters-mgz8n"},"namespace":"e2e-clusters-mgz8n","name":"example-n42fg","reconcileID":"dc9ea120-8120-400a-af0a-3b265526072a","error":"images.openstack.k-orc.cloud is forbidden: User \"system:serviceaccount:hypershift:operator\" cannot create resource \"images\" in API group \"openstack.k-orc.cloud\" in the namespace \"e2e-clusters-mgz8n-example-n42fg\"","stacktrace":"github.com/openshift/hypershift/hypershift-operator/controllers/nodepool.(*NodePoolReconciler).Reconcile\n\t/hypershift/hypershift-operator/controllers/nodepool/nodepool_controller.go:205\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile\n\t/hypershift/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:114\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n\t/hypershift/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:311\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/hypershift/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:261\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\t/hypershift/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:222"}
```

hypershift-operator/controllers/hostedcluster/internal/platform/openstack/openstack.go
EmilienM: I thought about it today and if a customer want a shared image, they can just pre-create it and use `--openstack-node-image-name`.

hypershift-operator/controllers/nodepool/openstack/openstack.go
```diff
@@ -72,3 +76,62 @@ func MachineTemplateSpec(hcluster *hyperv1.HostedCluster, nodePool *hyperv1.Node
 	}
 	return openStackMachineTemplate, nil
 }
+
+func ReconcileOpenStackImageCR(ctx context.Context, client client.Client, createOrUpdate upsert.CreateOrUpdateFN, hcluster *hyperv1.HostedCluster, controlPlaneNamespace string, release *releaseinfo.ReleaseImage) error {
+	openStackImage := orc.Image{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "rhcos-" + hcluster.Name,
+			Namespace: controlPlaneNamespace,
+			// TODO: add proper cleanup in CAPI resources cleanup
```
EmilienM, line 85: we need to add proper cleanup in the controller for removing the object.

hypershift-operator/controllers/nodepool/capi.go
```diff
@@ -1375,3 +1382,7 @@ func (r *NodePoolReconciler) getMachinesForNodePool(ctx context.Context, nodePoo
 
 	return sortedByCreationTimestamp(machinesForNodePool), nil
 }
+
+func (c *CAPI) reconcileORCResources(ctx context.Context) error {
+	return openstack.ReconcileOpenStackImageCR(ctx, c.Client, c.CreateOrUpdate, c.hostedCluster, c.controlplaneNamespace, c.releaseImage)
```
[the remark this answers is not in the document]
mdbooth: It's because ORC isn't logging anything steady state at V(2). I think we should change this in ORC, but for now it might be better just to bump CAPO's log level from 2 to 3 or 4.

hypershift-operator/controllers/nodepool/openstack/openstack.go
```diff
@@ -72,3 +76,62 @@ func MachineTemplateSpec(hcluster *hyperv1.HostedCluster, nodePool *hyperv1.Node
...
+			OwnerReferences: []metav1.OwnerReference{
+				{
+					APIVersion: hcluster.APIVersion,
+					Kind:       hcluster.Kind,
+					Name:       hcluster.Name,
+					UID:        hcluster.UID,
+				},
+			},
+		},
```
mdbooth, line 94: I'm pretty sure this is the problem: owner references can't be cross-namespace. Because this reference is invalid, the k8s gc will automatically delete it at some point. When that happens this controller will presumably re-create it and we get into the weird situation we see in the logs where we're referencing a deleted object.

The object needs to be owned by something in the same namespace. The HostedCP is the obvious candidate.

EmilienM: This was a mistake, I'm not sure why I used that namespace :/

hypershift-operator/controllers/nodepool/openstack/openstack.go
```diff
@@ -72,3 +76,62 @@ func MachineTemplateSpec(hcluster *hyperv1.HostedCluster, nodePool *hyperv1.Node
 	}
 	return openStackMachineTemplate, nil
 }
+
+func ReconcileOpenStackImageCR(ctx context.Context, client client.Client, createOrUpdate upsert.CreateOrUpdateFN, hcluster *hyperv1.HostedCluster, release *releaseinfo.ReleaseImage) error {
+	openStackImage := orc.Image{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "rhcos-" + hcluster.Name,
```
mdbooth, line 83: I think you're going to want to include the RHCOS image version as a component of the name. The problem is, when we inevitably need to bump the image version in the hosted cluster, what do we call the new one? It will have to have a different name, and we will need to update the NodePool to refer to it. If we standardise on always including the RHCOS version (and arch, I guess?) in the name then this problem goes away.

support/releaseinfo/releaseinfo.go
```diff
@@ -355,3 +355,24 @@ func (v ComponentVersions) DisplayNameLabel() string {
 	}
 	return buf.String()
 }
+
+func UnsupportedOpenstackDefaultImage(releaseImage *ReleaseImage) (string, string, error) {
```
mdbooth, line 359: Why is it unsupported?

hypershift-operator/controllers/nodepool/openstack/openstack.go
```diff
@@ -72,3 +76,62 @@ func MachineTemplateSpec(hcluster *hyperv1.HostedCluster, nodePool *hyperv1.Node
...
+		return fmt.Errorf("failed to lookup RHCOS image: %w", err)
+	}
+
+	openStackImageSpec.CloudCredentialsRef = orc.CloudCredentialsReference{
+		SecretName: hcluster.Spec.Platform.OpenStack.IdentityRef.Name,
+		CloudName:  hcluster.Spec.Platform.OpenStack.IdentityRef.CloudName,
+	}
+
+	openStackImageSpec.Resource = &orc.ImageResourceSpec{
+		Name: "rhcos-" + hcluster.Name,
```
mdbooth, line 122: **FOR CI ONLY**

We'll want an option to override this name so we can re-use the image between CI runs for different clusters. As written this will be expensive to run if the HostedCP is in AWS and the workers are in Vexxhost because we're going to download the bits to AWS before uploading them again to Vexxhost... for each HostedCluster in each CI run. Our shareholders might not be happy with us.

hypershift-operator/controllers/nodepool/openstack/openstack.go
```diff
@@ -72,3 +76,62 @@ func MachineTemplateSpec(hcluster *hyperv1.HostedCluster, nodePool *hyperv1.Node
 	}
 	return openStackMachineTemplate, nil
 }
+
+func ReconcileOpenStackImageCR(ctx context.Context, client client.Client, createOrUpdate upsert.CreateOrUpdateFN, hcluster *hyperv1.HostedCluster, release *releaseinfo.ReleaseImage) error {
+	openStackImage := orc.Image{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "rhcos-" + hcluster.Name,
+			Namespace: hcluster.Namespace,
```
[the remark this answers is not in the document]
MaysaMacedo: CAPI resources are in a different namespace. But I change the namespace of the ORC Image to be in the same of the CAPI deployment, otherwise no image would be created.

hypershift-operator/controllers/nodepool/openstack/openstack.go
MaysaMacedo: I changed. The image now has the following format, for example:
`rhcos-418.94.202411221729-0-may`

support/releaseinfo/releaseinfo.go
MaysaMacedo: I'm not sure. @EmilienM any ideas?

hypershift-operator/controllers/nodepool/openstack/openstack.go
MaysaMacedo: Can't we just remain using `openstack-node-image-name` for most of the CI jobs? and some running less frequently without it?

hypershift-operator/controllers/nodepool/openstack/openstack.go
EmilienM: I like the result Maysa.

support/releaseinfo/releaseinfo.go
EmilienM: That function was taken from kubevirt and we can / should certainly rename it.

hypershift-operator/controllers/nodepool/openstack/openstack.go
MaysaMacedo: `openstack-node-image-name` is a flag that can be passed in the `hcp create cluster` command and it defines the name of a pre-existent image to be used in the OSP VMs.

hypershift-operator/controllers/nodepool/openstack/openstack.go
```diff
@@ -72,3 +82,85 @@ func MachineTemplateSpec(hcluster *hyperv1.HostedCluster, nodePool *hyperv1.Node
...
+			Download: &orc.ImageContentSourceDownload{
+				URL:        imageURL,
+				Decompress: ptr.To(orc.ImageCompressionGZ),
+				Hash: &orc.ImageHash{
+					Algorithm: "sha256",
+					Value:     imageHash,
+				},
+			},
+		},
+	}
```
EmilienM, line 163: In order to be efficient, in CI we will need to specify `managedOptions.onDelete` to `detach` so the RHCOS image will be created once in Glance and re-used across multiple Hosted Clusters. The CI cleanup jobs will eventually take care of deleting the image but the Hypershift CI wouldn't bother doing it.

This is a CI use case. As a customer, the image would always be deleted when a HostedCluster is being deleted.

cc @mdbooth

hypershift-operator/controllers/nodepool/openstack/openstack.go
MaysaMacedo: I had to remove the cluster name, otherwise ORC wouldn't reuse the existent image from the same ocp version but had a different cluster name

hypershift-operator/controllers/nodepool/openstack/openstack.go
MaysaMacedo: done

support/releaseinfo/releaseinfo.go
MaysaMacedo: done

test/e2e/nodepool_test.go
```diff
@@ -125,6 +125,10 @@ func TestNodePool(t *testing.T) {
 						name: "OpenStackAZTest",
 						test: NewOpenStackAZTest(ctx, mgtClient, hostedCluster),
 					},
+					{
+						name: "OpenStackImageTest",
```
EmilienM, line 129: The e2e job timeouted, that makes me think we have too many tests maybe. We could combine this scenario with another one (AZ or multi net for example).

EmilienM:
/test e2e-openstack-aws
In the future runs, we tag the HostedClusters so the Glance image is not cleaned up after HC is deleted. That's made to save egress fees. Let's see if that works.

EmilienM:
@MaysaMacedo 
The CI job failure might be valid, especially `TestNodePoolReplaceUpgrade` ? Might be worse investigate.

EmilienM:
/hold
I've included the commit from https://github.com/openshift/hypershift/pull/5584 for early testing.

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -167,6 +167,10 @@ const (
 	// of the cluster will be removed, including image registry storage, ingress dns records, load balancers, and persistent storage.
 	CleanupCloudResourcesAnnotation = "hypershift.openshift.io/cleanup-cloud-resources"
 
+	// CleanupOrcImageResourcesAnnotation is an annotation that indicates whether an OpenStack Image should be
+	// removed when deleting the corresponding ORC Image. If set to "false", the OpenStack Image will not be removed.
+	CleanupOrcImageResourcesAnnotation = "hypershift.openshift.io/cleanup-orc-image-resource"
```
EmilienM, line 172: I would rename it to `CleanupORCImageResourcesAnnotation`.

Constants in Go typically follow [CamelCase](https://github.com/zecke/Kubernetes/blob/master/docs/devel/api-conventions.md#constants) and avoid underscores. Also, acronyms (like ORC) are usually all uppercase to maintain readability.

hypershift-operator/controllers/hostedcluster/internal/platform/openstack/openstack.go
```diff
@@ -204,7 +204,9 @@ func (a OpenStack) CAPIProviderDeploymentSpec(hcluster *hyperv1.HostedCluster, _
 					Args: []string{
 						"--namespace=$(MY_NAMESPACE)",
 						"--leader-elect",
-						"--v=2",
+						// We need to set the log level to 3 to get the logs from ORC.
+						// Once ORC follows logging guidelines, we should use V(2) again.
```
EmilienM, line 208: @mandre @mdbooth any updates on this?

hypershift-operator/controllers/nodepool/openstack/openstack.go
```diff
@@ -73,3 +83,91 @@ func MachineTemplateSpec(hcluster *hyperv1.HostedCluster, nodePool *hyperv1.Node
...
+	controlPlaneNamespace := fmt.Sprintf("%s-%s", nodePool.Namespace, strings.ReplaceAll(nodePool.Spec.ClusterName, ".", "-"))
+	openstackCluster, err := getOpenStackClusterForHostedCluster(ctx, client, hcluster, controlPlaneNamespace)
+	if err != nil {
+		return err
+	}
+	openStackImage := orc.Image{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "rhcos-" + releaseVersion,
+			Namespace: controlPlaneNamespace,
+			// TODO: add proper cleanup in CAPI resources cleanup
```
EmilienM, line 111: we can remove that comment.

EmilienM:
/hold cancel
passing CI, ready for review

EmilienM:
/hold cancel
This is ready (again) to be reviewed.

cmd/cluster/openstack/create.go
```diff
@@ -233,15 +233,15 @@ func (o *CreateOptions) GenerateResources() ([]client.Object, error) {
...
+		// The following rule is required for CAPO to reconcile for the Images resources created by ORC,
 		// which is a dependency since CAPO v0.11.0.
 		// This rule is also defined in the Hypershift HostedCluster controller and the Hypershift Operator when creating
 		// the cluster.
 		Rules: []rbacv1.PolicyRule{
 			{
 				APIGroups: []string{"openstack.k-orc.cloud"},
-				Resources: []string{"images"},
-				Verbs:     []string{"list", "watch"},
+				Resources: []string{"images", "images/status"},
```
mdbooth, line 243: Hypershift doesn't need any permissions on status. You should remove status here.
```suggestion
				Resources: []string{"images"},
```

cmd/install/assets/hypershift_operator.go
```diff
@@ -1172,14 +1172,14 @@ func (o HyperShiftOperatorClusterRole) Build() *rbacv1.ClusterRole {
...
-			// The following rule is required for CAPO to watch for the Images resources created by ORC,
+			// The following rule is required for CAPO to reconcile for the Images resources created by ORC,
 			// which is a dependency since CAPO v0.11.0.
 			// This rule is also defined in the Hypershift HostedCluster controller and the Hypershift CLI when creating
 			// the cluster.
 			{
 				APIGroups: []string{"openstack.k-orc.cloud"},
-				Resources: []string{"images"},
-				Verbs:     []string{"list", "watch"},
+				Resources: []string{"images", "images/status"},
```
mdbooth, line 1181:
```suggestion
				Resources: []string{"images"},
```

docs/content/how-to/openstack/hostedcluster.md
```diff
@@ -89,6 +85,13 @@ hcp create cluster openstack \
     If your management cluster doesn't have enough workers (less than 3), which is not recommended nor supported,
     you'll need to specify the `--control-plane-availability-policy` flag to `SingleReplica`.
 
+!!! note
+
+    When not providing the `--openstack-node-image-name` flag, the latest RHCOS image will be used.
+    ORC will handle the RHCOS image lifecycle by downloading the image from the OpenShift mirror and deleting it when it's no longer needed.
+    If you want ORC to not delete the image, you can created the HostedCluster with the following flag: `--annotations "hypershift.openshift.io/cleanup-orc-image-resource=false"`.
```
mdbooth, line 92:
```suggestion
    If you want ORC to not delete the image, you can create the HostedCluster with the following flag: `--annotations "hypershift.openshift.io/cleanup-orc-image-resource=false"`.
```

hypershift-operator/controllers/hostedcluster/internal/platform/openstack/openstack.go
```diff
@@ -416,14 +417,14 @@ func (a OpenStack) CAPIProviderPolicyRules() []rbacv1.PolicyRule {
...
-		// The following rule is required for CAPO to watch for the Images resources created by ORC,
+		// The following rule is required for CAPO to reconcile the Images resources created by ORC,
 		// which is a dependency since CAPO v0.11.0.
 		// This rule is also defined in the Hypershift Operator and the Hypershift CLI when creating
 		// the cluster.
 		{
 			APIGroups: []string{"openstack.k-orc.cloud"},
-			Resources: []string{"images"},
-			Verbs:     []string{"list", "watch"},
+			Resources: []string{"images", "images/status"},
```
mdbooth, line 426:
```suggestion
			Resources: []string{"images"},
```

mdbooth (commented):
I'd prefer that you remove the RBAC on image/status as it's not required, but apart from that it looks fine. We don't need to worry about whether the image we created is ready as CAPO will do that for us. Looks fairly straightforward 🤞

cmd/install/assets/hypershift_operator.go
```diff
@@ -1172,14 +1172,11 @@ func (o HyperShiftOperatorClusterRole) Build() *rbacv1.ClusterRole {
 				Resources: []string{"ipaddresses", "ipaddresses/status"},
 				Verbs:     []string{"create", "delete", "get", "list", "update", "watch"},
 			},
-			// The following rule is required for CAPO to watch for the Images resources created by ORC,
-			// which is a dependency since CAPO v0.11.0.
-			// This rule is also defined in the Hypershift HostedCluster controller and the Hypershift CLI when creating
-			// the cluster.
+			// This allows hypershift operator to grant RBAC permissions for ORC to the capi-provider.
```
mdbooth, line 1175: Thanks for the comment 🙏

mdbooth (approved):
/lgtm

docs/content/how-to/openstack/prerequisites.md
```diff
@@ -71,11 +70,18 @@ operator-755d587f44-lrtrq   1/1     Running   0          114s
...
+HostedClusters will have pod(s) for etcd and its performance is essential for the cluster health.
+In production environments, it's required to put etcd data on fast storage and in the case of OpenStack it'll be local storage.
+Follow this [procedure](etcd-local-storage.md) to leverage a well-known and tested solution.
+
+## Upload RHCOS image in OpenStack (optional)
 
-For now, we need to manually push an RHCOS image that will be used when deploying the node pools
-on OpenStack. In the [future](https://issues.redhat.com/browse/OSASINFRA-3492), the CAPI provider (CAPO) will handle the RHCOS image
-lifecycle by using the image available in the chosen release payload.
+The user can specific which RHCOS image to use when deploying the node pools
```
bryan-cox, line 81:
```suggestion
The user can specify which RHCOS image to use when deploying the node pools
```

hypershift-operator/controllers/hostedcluster/internal/platform/openstack/openstack.go
```diff
@@ -272,6 +272,7 @@ func (a OpenStack) CAPIProviderDeploymentSpec(hcluster *hyperv1.HostedCluster, _
 				"--namespace=$(MY_NAMESPACE)",
 				"--leader-elect",
 				"--health-probe-bind-address=:8081",
+				"--zap-log-level=4",
```
bryan-cox, line 275: Should this always be set to 4 or to some configurable value?

support/releaseinfo/releaseinfo.go
```diff
@@ -357,3 +357,36 @@ func (v ComponentVersions) DisplayNameLabel() string {
 	}
 	return buf.String()
 }
+
+func OpenstackDefaultImage(releaseImage *ReleaseImage) (string, string, error) {
+	arch, foundArch := releaseImage.StreamMetadata.Architectures["x86_64"]
+	if !foundArch {
+		return "", "", fmt.Errorf("couldn't find OS metadata for architecture %q", "x64_64")
```
bryan-cox, line 364:
```suggestion
		return "", "", fmt.Errorf("couldn't find OS metadata for architecture %q", "x86_64")
```

support/releaseinfo/releaseinfo.go
```diff
@@ -357,3 +357,36 @@ func (v ComponentVersions) DisplayNameLabel() string {
...
+		return "", "", fmt.Errorf("couldn't find OS metadata for the openstack qcow2.gz disk")
+	}
+
+	return disk.Location, disk.SHA256, nil
+}
+
+func OpenStackReleaseImage(releaseImage *ReleaseImage) (string, error) {
+	arch, foundArch := releaseImage.StreamMetadata.Architectures["x86_64"]
+	if !foundArch {
+		return "", fmt.Errorf("couldn't find OS metadata for architecture %q", "x64_64")
```
bryan-cox, line 385:
```suggestion
		return "", fmt.Errorf("couldn't find OS metadata for architecture %q", "x86_64")
```

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -167,6 +167,10 @@ const (
 	// of the cluster will be removed, including image registry storage, ingress dns records, load balancers, and persistent storage.
 	CleanupCloudResourcesAnnotation = "hypershift.openshift.io/cleanup-cloud-resources"
 
+	// CleanupORCImageResourcesAnnotation is an annotation that indicates whether an OpenStack Image should be
```
bryan-cox, line 170: Could you add the why here?

EmilienM: Done. Next patchset will have a longer description explaining why.

hypershift-operator/controllers/hostedcluster/internal/platform/openstack/openstack.go
mdbooth: Ideally this would use the same log level as everything else. It's a bug in ORC that we can't do that. Possibly worth adding a comment (with a bug link?) explaining why we're hard coding 4 here, but this is an appropriate workaround for now IMHO.

EmilienM: The logs in OpenStack-Resource-Controller (ORC) aren't yet well implemented and from my experience as a user, log level 4 (INFO) is what provides useful information for now for all resources that ORC handles.
Would you have a preference on how we can configure it?

EmilienM: I'll add a link so we can keep track of it.

bryan-cox: Maybe you could add it as a constant here in an OpenStack section - hypershift/support/config/constants.go? That way a comment could be added why its 4 as Matthew suggested. WDYT?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -167,6 +167,10 @@ const (
...
+	// removed when deleting the corresponding ORC Image. If set to "false", the OpenStack Image will not be removed.
+	CleanupORCImageResourcesAnnotation = "hypershift.openshift.io/cleanup-orc-image-resource"
```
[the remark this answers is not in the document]
EmilienM: That's a very valid question. When we originally looked at it, we were inspired by the field above `CleanupCloudResourcesAnnotation` and thought we could follow that pattern.
Yes we could move that field to `OpenStackPlatformSpec` but since this is TechPreview, I wasn't sure about changing the API for now and I'll rather expose that through a label rather than a stable API that we're supposed to maintain :shrug: 

Let us know if you prefer it to be part of the spec.

support/releaseinfo/releaseinfo.go
```diff
@@ -357,3 +357,36 @@ func (v ComponentVersions) DisplayNameLabel() string {
 	}
 	return buf.String()
 }
+
+func OpenstackDefaultImage(releaseImage *ReleaseImage) (string, string, error) {
```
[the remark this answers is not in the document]
EmilienM: done and I added unit tests to these functions.

api/hypershift/v1beta1/hostedcluster_types.go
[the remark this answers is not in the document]
enxebre: >That's a very valid question. When we originally looked at it, we were inspired by the field above CleanupCloudResourcesAnnotation and thought we could follow that pattern.

fwiw some of those decisions are influenced by not wanting to surface something to multiple platforms or products, or other historic reasons, or they might be just bad decisions. So question them :).

>Yes we could move that field to OpenStackPlatformSpec but since this is TechPreview, I wasn't sure about changing the API for now and I'll rather expose that through a label rather than a stable API that we're supposed to maintain 🤷

We don't want to add any new annotation unless there's a good technical reason for it. Please put in the API, that's the main point of having it TechPreview, you can try things and break it if you are not sure. The annotation would need the same level of maintenance once you expose it ga and is less explicit.

api/hypershift/v1beta1/openstack.go
```diff
@@ -136,6 +136,21 @@ type OpenStackPlatformSpec struct {
...
+	// which happens when the HostedCluster is deleted.
+	//
+	// If set to `true`, the OpenStack Image will not be removed, allowing it to be
+	// reused by other HostedClusters. This is useful in cases where images are shared
+	// across multiple clusters or in environments where we want to avoid unnecessary downloads
+	// for each test cluster, saving time and cost.
+	//
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf", message="SkipCleanupORCImageResources is immutable"
+	// +optional
+	SkipCleanupORCImageResources *bool `json:"SkipCleanupORCImageResources,omitempty"`
```
mdbooth, line 153: @JoelSpeed would tell us off for using a boolean here, btw. The wonks are currently telling us that booleans should generally be enums instead.

This is what we do in ORC, in fact. `onDelete` is one of `delete` or `detach`: https://k-orc.cloud/crd-reference/#ondelete. This leaves open the possibility of other behaviours in the future without requiring a backwards-incompatible change.

EmilienM: I thought about it too but I didn't find the user experience so great, in that particular case; but maybe it's fine as long as it's well documented?

bryan-cox: ^ Just FYI on OpenShift API conventions in case you didn't have it. Joel shared that with me in the recent past.

JoelSpeed: An `ImageRetentionPolicy: Prune | Orphan` might work here

mdbooth: I like that better than what we have in ORC, btw (which iirc I stole from ASO). Meh, it's baked now. Maybe we'll update it if we bump the api version.

EmilienM: Good idea Joel. I was hesitant to that because it's an abstraction to what ORC does, but I talked to Matt and he seems fine with it, so am I.

test/e2e/util/options.go
```diff
@@ -219,18 +220,26 @@ func (o *Options) DefaultNoneOptions() none.RawCreateOptions {
 }
 
 func (p *Options) DefaultOpenStackOptions() hypershiftopenstack.RawCreateOptions {
+	var e2eOpenStackImageRetentionPolicy hyperv1.RetentionPolicy
+	if p.ConfigurableClusterOptions.OpenStackImageRetentionPolicy == "" {
+		// Default to orphaning images for e2e tests so they can be reused
```
EmilienM, line 225: @mdbooth FYI I've put "safe" defaults here. Feedback is welcome.

cmd/cluster/openstack/create.go
```diff
@@ -45,15 +47,17 @@ func bindCoreOptions(opts *RawCreateOptions, flags *pflag.FlagSet) {
 	flags.StringVar(&opts.OpenStackExternalNetworkID, "openstack-external-network-id", opts.OpenStackExternalNetworkID, "ID of the OpenStack external network (optional)")
 	flags.StringVar(&opts.OpenStackIngressFloatingIP, "openstack-ingress-floating-ip", opts.OpenStackIngressFloatingIP, "An available floating IP in your OpenStack cluster that will be associated with the OpenShift ingress port (optional)")
 	flags.StringSliceVar(&opts.OpenStackDNSNameservers, "openstack-dns-nameservers", opts.OpenStackDNSNameservers, "List of DNS nameservers to use for the cluster (optional)")
+	flags.Var(&opts.OpenStackImageRetentionPolicy, "openstack-image-retention-policy", "OpenStack Glance Image retention policy. Valid values are 'Orphan' and 'Prune'. By default images are pruned. (optional)")
```
MaysaMacedo, line 50: if the option is required, should this not mention the optional in the message?

EmilienM: yeah I got confused in the discussion elsewhere where we said we could make it required with a default but after re-reading https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md#defaulting : 

> With configuration APIs, we typically default fields within the controller and not within the API.

So better to make it optional.

MaysaMacedo (commented):
/lgtm

EmilienM:
/test e2e-openstack-aws
Fixed a typo in API doc

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -356,6 +356,41 @@ const (
 	AWSMachinePublicIPs = "hypershift.openshift.io/aws-machine-public-ips"
 )
 
+// RetentionPolicy defines the policy for handling resources associated with a cluster when the cluster is deleted.
```
enxebre, line 359: just for organizational purposes I'd prefer to keep this self-contained within the Openstack file. Then me move it here if it's consumed agnostically or by more platforms

hypershift-operator/controllers/nodepool/openstack/openstack.go
```diff
@@ -73,3 +83,135 @@ func MachineTemplateSpec(hcluster *hyperv1.HostedCluster, nodePool *hyperv1.Node
...
+	cluster := capiopenstackv1beta1.OpenStackCluster{}
+
+	if err := c.Get(ctx, types.NamespacedName{Namespace: controlPlaneNamespace, Name: hcluster.Name}, &cluster); err != nil {
+		return cluster, fmt.Errorf("failed to get Cluster: %w", err)
+	}
+
+	return cluster, nil
+}
+
+func ReconcileOpenStackImageCR(ctx context.Context, client client.Client, createOrUpdate upsert.CreateOrUpdateFN, hcluster *hyperv1.HostedCluster, release *releaseinfo.ReleaseImage, nodePool *hyperv1.NodePool) error {
```
enxebre, line 97: please document this exported function

api/hypershift/v1beta1/hostedcluster_types.go
enxebre: fwiw I'd definitely like to adopt this for other platforms in future

api/hypershift/v1beta1/openstack.go
```diff
@@ -136,6 +140,17 @@ type OpenStackPlatformSpec struct {
 	// +kubebuilder:validation:XValidation:rule="isIP(self)",message="floatingIP must be a valid IPv4 or IPv6 address"
 	// +optional
 	IngressFloatingIP string `json:"ingressFloatingIP,omitempty"`
+
+	// ImageRetentionPolicy defines the policy for handling resources associated with the image
```
enxebre, line 144: nit: start fields doc with lowercase i.e imageRetentionPolicy, so gdocs match the serialized version of the field.

api/hypershift/v1beta1/openstack.go
```diff
@@ -136,6 +140,17 @@ type OpenStackPlatformSpec struct {
...
+	// when the cluster is deleted.
+	// The default (set by the Nodepool controller) matches OpenStack Resource Controller (ORC)
+	// behavior where the OpenStack Glance Image is pruned when the Image object is deleted during the
+	// HostedCluster deletion.
+	// It is defined at the HostedCluster level and will be used for all nodepools images
```
enxebre, line 149: can we explicitly document what happens if this is changed day 2?

EmilienM: Yes, and I also confirmed that the retention policy is mutable in ORC:
https://github.com/k-orc/openstack-resource-controller/blob/854b83db2bf6622d58f8583a643bdc1cab57b3fc/cmd/resource-generator/data/api.template#L78-L80

MaysaMacedo (approved):
/lgtm

EmilienM:
/hold
tests aren't looking good anymore.

api/hypershift/v1beta1/hostedcluster_types.go
enxebre: this is resolved but the declaration is still here, can we move it to openstack.go?

EmilienM:
I had to update unit tests; a minor change. Thanks for the approval @enxebre

PR #5156, opened 2024-11-19
HOSTEDCP-2182: Sync main and api go module vendoring to fix go workspaces 

csrwng:
Adds a target to the Makefile to do a 'go work sync' and ensure that the
main and api modules are synced whenever we update dependencies. This
makes it possible to work with go workspaces locally on developer
machines. The target will be executed whenever a 'make update' is run.

Fixes #[HOSTEDCP-2182](https://issues.redhat.com/browse/HOSTEDCP-2182)

**Checklist**


hack/workspace/go.work
```diff
@@ -0,0 +1,6 @@
+go 1.22.0
```
sjenning, line 1: is this going to force us all to use this version of go in order to run `make verify`?

csrwng: we already require that version of go because of https://github.com/openshift/hypershift/blob/main/go.mod

PR #5168, opened 2024-11-21
OCPBUGS-44655,OCPBUGS-43083: Fix IPv6 Disconnected HCP deployments

jparrill:
The issue with the MultiArch check is that it's recovering the manifests without passing through ImageMetadaProvider. I've added a new method to that which recovers the requested manifest and adds it to a new LRU Cache.

To pass the ImageMetadataProvider to the RegistryClient, implies to fall in a import cycle issue, so a new interface was created called ManifestProvider, which implement the same GetManifest methods as the original does.

Fixes:
- [OCPBUGS-44655](https://issues.redhat.com/browse/OCPBUGS-44655) - HO issue determining the cluster payload arch does not checks the ICSP/IDMS
- [OCPBUGS-43083](https://issues.redhat.com/browse/OCPBUGS-43083) - Pods cannot connect to apiserver in IPv6 disconnected hosted cluster

**Issues Solved**
- [x] Detecting MultiArch cannot get the right manifest (original GetManifests is not using the ImageMetadataProvider) 
  - Implemented GetManifests method inside of ImageMetadataProvider
- [x] ImageMetadataProvider does not understand the IDMS with only the domain (exact match works fine
  - Patched GetOverrideImages to understand IDMS namespace wildcards 
- [x] CVO is being override by a EnvVar and not getting the digest from OCP release
  - Implemented GetDigest method inside of ImageMetadataProvider 
- [x] CVO deployment’s payload-script does not contain EOF
- [x] IPv6 does not work, nodes cannot join the KAS
  - The `IsIPv4` function now accepts IPAddresses instead of only CIDRs.

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
@@ -988,6 +988,7 @@ func (r *HostedControlPlaneReconciler) reconcileCPOV2(ctx context.Context, hcp *
 		SetDefaultSecurityContext: r.SetDefaultSecurityContext,
 		MetricsSet:                r.MetricsSet,
 		EnableCIDebugOutput:       r.EnableCIDebugOutput,
+		ImageMetadataProvider:     r.ImageMetadataProvider,
 	}
 
 	var errs []error
--- a/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
@@ -175,6 +175,7 @@ type HostedControlPlaneReconciler struct {
 	awsSession                              *session.Session
 	reconcileInfrastructureStatus           func(ctx context.Context, hcp *hyperv1.HostedControlPlane) (infra.InfrastructureStatus, error)
 	EnableCVOManagementClusterMetricsAccess bool
+	ImageMetadataProvider                   util.ImageMetadataProvider
 
 	IsCPOV2 bool
 }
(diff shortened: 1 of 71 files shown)


support/util/imagemetadata.go
```diff
@@ -131,6 +128,310 @@ func (r *RegistryClientImageMetadataProvider) ImageMetadata(ctx context.Context,
...
+		if imageDigest, exists := digestCache.Get(imageRef); exists {
+			parsedImageRef.ID = string(imageDigest.(digest.Digest))
+			return imageDigest.(digest.Digest), &parsedImageRef, nil
+		}
+
+		ref = &parsedImageRef
+	}
+
+	// Get the image repo info based the source/mirrors in the ICSPs/IDMSs
+	for source, mirrors := range r.OpenShiftImageRegistryOverrides {
```
csrwng, line 202: It looks like you run this loop in a few places, should it be in a separate function?

csrwng (commented):
Some comments

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -3609,9 +3610,27 @@ func (r *HostedControlPlaneReconciler) reconcileClusterVersionOperator(ctx conte
 		}
 	}
 
+	// The CVO prepare-payload script needs the ReleaseImage digest for disconnected environments
```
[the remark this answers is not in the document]
jparrill: As we talked on slack, CVO image does not contain the `release-manifests` folder but OCP Release image does, so we need this.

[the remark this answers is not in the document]
csrwng: Please ignore this, as discussed, the CVO image cannot be used as a replacement for the release image.

jparrill:
Still hitting quota :( - `NatGatewayLimitExceeded: Performing this operation would exceed the limit of 100 NAT gateways`

jparrill:
Rebase from main to fix Public only clusters conditions in `TestClusterCreateV2`

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller_test.go
```diff
@@ -1659,13 +1662,24 @@ func TestControlPlaneComponents(t *testing.T) {
 		},
 	}
 
+	pullSecret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{Name: "pull-secret", Namespace: "hcp-namespace"},
+		Data: map[string][]byte{
+			corev1.DockerConfigJsonKey: []byte(`{}`),
+		},
+	}
```
muraee, line 1670: please move this to `componentsFakeDependencies()` function.

control-plane-operator/controllers/hostedcontrolplane/v2/cvo/deployment.go
```diff
@@ -37,11 +39,51 @@ func (cvo *clusterVersionOperator) adaptDeployment(cpContext component.WorkloadC
...
+	if err := cpContext.Client.Get(cpContext.Context, client.ObjectKeyFromObject(pullSecret), pullSecret); err != nil {
+		return fmt.Errorf("failed to get pull secret for namespace %s: %w", cpContext.HCP.Namespace, err)
+	}
+
+	cpRef, err := registryclient.GetCorrectArchImage(cpContext.Context, "cluster-version-operator", cpContext.HCP.Spec.ReleaseImage, pullSecret.Data[corev1.DockerConfigJsonKey], cpContext.ImageMetadataProvider)
+	if err != nil {
+		return fmt.Errorf("failed to parse control plane release image %s: %w", cpRef, err)
+	}
+
+	_, cpReleaseImageRef, err := cpContext.ImageMetadataProvider.GetDigest(cpContext.Context, cpRef, pullSecret.Data[corev1.DockerConfigJsonKey])
```
muraee, line 58: can we have a reusable function that we can just call here?

jparrill:
I have present that I need to add the unit tests for `discoverCVOReleaseImages` new function included in the PR. I will do it in a follow up PR because I need this PR merged for code freeze.

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller_test.go
muraee: sorry meant `componentsFakeObjects()` :D. You can also remove this line here now.

ignition-server/cmd/start.go
```diff
@@ -149,6 +149,10 @@ func setUpPayloadStoreReconciler(ctx context.Context, registryOverrides map[stri
 		return nil, fmt.Errorf("unable to create image file cache: %w", err)
 	}
 
+	imageMetaDataProvider := &util.RegistryClientImageMetadataProvider{
```
enxebre, line 152: why does the ignition-server needs the new metadata provider? Isn't this which is already available all we need https://github.com/openshift/hypershift/blob/main/ignition-server/cmd/start.go#L155-L165? so the you can invoke GetOpenShiftImageRegistryOverrides?

jparrill: The releaseProvider does not contain the ImageMetadata of the Release image itself, but the content. And to verify the Arch functions will need all the metadata including the layers.

PR #5176, opened 2024-11-22
NO-JIRA: Remove OpenStack workaround for Ingress Floating IP

EmilienM:

Remove some technical debt around the Ingress Floating IP reconcile for the OpenStack platform.


EmilienM:
/assign csrwng
as we worked together on this commit that we now revert.

EmilienM:
The fact that `e2e-openstack-aws-conformance` passed show there is no regression with that PR. That job uses a pre-created Floating IP used by the HostedCluster later on. So I think it's safe to assess that this PR is functionally OK. BTW it was a clean revert (except go imports due to recent changes).

EmilienM:
/test e2e-openstack-aws
I'll live debug the job in prow

EmilienM:
With e2e-openstack-aws-conformance passing, this is safe to merge.

PR #5178, opened 2024-11-22
OCPBUGS-44910: openstack: don't reconcile image registry config during bootstrap

EmilienM:

For platforms where cluster-image-registry-operator (CIRO) needs a PVC to be created, bootstrap needs to happen
in CIRO before the registry config is created. For now, this is the case for the OpenStack platform.
If the object exist, we reconcile the registry config for other other fields as it should be fine since the PVC would
exist at this point.

This patch will maintain a list of platforms that require a PVC for CIRO
(OpenStack only for now), check whether an image registry config already
exist and if it does, skip reconcile to let CIRO bootstrap doing it and
creating the needed PVC for CIRO.

### DIFF
--- a/control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
+++ b/control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
@@ -356,39 +356,57 @@ func (r *reconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Result
 		errs = append(errs, fmt.Errorf("failed to reconcile rbac: %w", err))
 	}
 
-	log.Info("reconciling registry config")
 	registryConfig := manifests.Registry()
-	if _, err := r.CreateOrUpdate(ctx, r.client, registryConfig, func() error {
-		registry.ReconcileRegistryConfig(registryConfig, r.platformType, hcp.Spec.InfrastructureAvailabilityPolicy)
-		return nil
-	}); err != nil {
-		errs = append(errs, fmt.Errorf("failed to reconcile imageregistry config: %w", err))
+	var registryConfigExists bool
+	// Check if the registry config exists
+	if err := r.client.Get(ctx, client.ObjectKeyFromObject(registryConfig), registryConfig); err != nil {
+		if !apierrors.IsNotFound(err) {
+			return ctrl.Result{}, fmt.Errorf("failed to get registry config: %w", err)
+		}
+	} else {
+		registryConfigExists = true
 	}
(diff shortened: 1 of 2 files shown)


control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -356,39 +356,57 @@ func (r *reconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Result
...
+		if !apierrors.IsNotFound(err) {
+			return ctrl.Result{}, fmt.Errorf("failed to get registry config: %w", err)
+		}
+	} else {
+		registryConfigExists = true
+	}
+
+	// For platforms where cluster-image-registry-operator (CIRO) needs a PVC to be created, bootstrap needs to happen
+	// in CIRO before the registry config is created. For now, this is the case for the OpenStack platform.
+	// If the object exist, we reconcile the registry config for other other fields as it should be fine since the PVC would
```
MaysaMacedo, line 372:
```suggestion
	// If the object exist, we reconcile the registry config for other fields as it should be fine since the PVC would
```

MaysaMacedo (commented):
/lgtm

mandre (commented):
I'm not clear on the relationship between CIRO and hypershift. If we need in all case for CIRO to write its config once, why don't we make it a needed condition in hypershift (for all platforms)? Why do we need to maintain a list of platforms that might use a PV as the storage for the image registry?

EmilienM:
/test unit
flake

MaysaMacedo:
/lgtm
/hold

waiting for the reply to Martin's question

EmilienM:
> I'm not clear on the relationship between CIRO and hypershift. If we need in all case for CIRO to write its config once, why don't we make it a needed condition in hypershift (for all platforms)? Why do we need to maintain a list of platforms that might use a PV as the storage for the image registry?

Creating configs before an operator starts seems to be a common design pattern in Hypershift. I'm not sure we want to change that for all the platforms even if it might work for CIRO in this particular case.
I think Cesar's proposal was to only make it for OpenStack as we're the only platform now that can leverage PVC for CIRO's backend.

csrwng:
> If we need in all case for CIRO to write its config once, why don't we make it a needed condition in hypershift (for all platforms)?

As Emilien mentioned, the reason is to limit the scope of this change to OpenStack. We may eventually want to do it for other platforms as well. But since that does have a broader impact, we want to be more intentional about it and maybe not just do it for CIRO but for other operators as well.

MaysaMacedo:
/lgtm
/hold

depends on the backport of  https://github.com/openshift/hypershift/pull/5178

PR #5179, opened 2024-11-23
Create AWS clients on every reconcile instead of at initialization

csrwng:
Moves client creation for the private link endpoint controller into the reconcile loop instead of when the reconciler is initially registered. This allows the controller to recover from initial issues assuming a shared vpc role.

Adds additional error logging when AWS cloud API calls fail.

Fixes #[OCPBUGS-44698](https://issues.redhat.com/browse/OCPBUGS-44698)

**Checklist**


control-plane-operator/controllers/awsprivatelink/awsprivatelink_controller.go
```diff
@@ -686,10 +789,11 @@ func (r *AWSEndpointServiceReconciler) reconcileAWSEndpointSecurityGroup(ctx con
 	ingressPermissions := supportawsutil.VPCEndpointSecurityGroupRules(machineCIDRs, vpcEndpointPort(awsEndpointService))
 	missingPermissions := diffPermissions(sg.IpPermissions, ingressPermissions)
 	if len(missingPermissions) > 0 {
-		if _, err = r.ec2Client.AuthorizeSecurityGroupIngressWithContext(ctx, &ec2.AuthorizeSecurityGroupIngressInput{
+		if _, err = ec2Client.AuthorizeSecurityGroupIngressWithContext(ctx, &ec2.AuthorizeSecurityGroupIngressInput{
 			GroupId:       aws.String(sgID),
 			IpPermissions: ingressPermissions,
 		}); err != nil {
+			log.Error(err, "failed to sect security group ingress rules", "id", sgID)
```
enxebre, line 796: typo sect?
Why are we unconditionally logging here then right below we are logging again via
```
			if supportawsutil.AWSErrorCode(err) != "InvalidPermission.Duplicate" {
				return fmt.Errorf("failed to set security group ingress rules, code: %s", supportawsutil.AWSErrorCode(err))
			}
			log.Info("WARNING: got duplicate permissions error when setting security group ingress permissions", "sgID", sgID)
```

control-plane-operator/controllers/awsprivatelink/awsprivatelink_controller.go
```diff
@@ -756,14 +861,17 @@ func (r *AWSEndpointServiceReconciler) createSecurityGroup(ctx context.Context,
 	describeSGInput := &ec2.DescribeSecurityGroupsInput{
 		GroupIds: []*string{aws.String(sgID)},
 	}
-	if err = r.ec2Client.WaitUntilSecurityGroupExistsWithContext(ctx, describeSGInput); err != nil {
+	if err = ec2Client.WaitUntilSecurityGroupExistsWithContext(ctx, describeSGInput); err != nil {
+		log.Error(err, "failed to wait for security group to exist", "id", sgID)
```
enxebre, line 865: why are we logging in addition to the returned error and with different message?

control-plane-operator/main.go
```diff
@@ -513,3 +488,7 @@ func NewStartCommand() *cobra.Command {
 
 	return cmd
 }
+
+func isAWS() bool {
+	return os.Getenv("AWS_REGION") != ""
```
enxebre, line 493: Should this be:
```
func isAWSPrivate() bool {
	return os.Getenv("AWS_REGION") != "" && hcp.Spec.Platform.Type == hyperv1.AWSPlatform
}
```

control-plane-operator/controllers/awsprivatelink/awsprivatelink_controller.go
```diff
@@ -187,12 +189,113 @@ const (
...
+	if !b.initialized {
+		return nil, nil, errors.New("clients not initialized")
+	}
+
+	// AWS_SHARED_CREDENTIALS_FILE and AWS_REGION envvar should be set in operator deployment
+	awsEndpointSession, err := b.awsSession()
+	if err != nil {
+		return nil, nil, err
+	}
+	if b.assumeEndpointRoleARN != "" {
```
enxebre, line 229: nit:
```
// When sharedVPC we need assume these additional roles
```
?

control-plane-operator/controllers/awsprivatelink/awsprivatelink_controller.go
```diff
@@ -187,12 +189,113 @@ const (
...
-	AssumeEndpointRoleARN *string
-	AssumeRoute53RoleARN  *string
-	LocalZoneID           *string
+	awsClientBuilder clientBuilder
+}
+
+type clientBuilder struct {
+	mu                    sync.Mutex
+	initialized           bool
+	assumeEndpointRoleARN string
```
enxebre, line 199: assumeSharedVPCEndpointRoleARN
assumeSharedVPCRoute53RoleARN
?

control-plane-operator/controllers/awsprivatelink/awsprivatelink_controller.go
```diff
@@ -187,12 +189,113 @@ const (
...
+	defer b.mu.Unlock()
+
+	return b.localZoneID
+}
+
+func (b *clientBuilder) initializeWithHCP(log logr.Logger, hcp *hyperv1.HostedControlPlane) {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+
+	if !b.initialized {
```
enxebre, line 260: would it make any difference if this func is just
```
b.warnOnDifferentValues(log, hcp)
b.setFromHCP(hcp)
b.initialized = true
```
without the conditional?

csrwng: we'd warn unnecessarily when initially setting the values

control-plane-operator/main.go
csrwng: We are no longer retrieving the hcp on startup, but we're always setting AWS_REGION on the CPO when platform.Type == AWS:
https://github.com/openshift/hypershift/blob/a3791d6fe5949fa50a3a7109893e54097ad7d496/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go#L2739-L2770 
So just retrieving the hcp on startup for the same check seemed unnecessary

control-plane-operator/controllers/awsprivatelink/awsprivatelink_controller.go
csrwng: > typo sect?
yes

Will also fix the unconditional logging

control-plane-operator/main.go
enxebre: ah I see. Where I was actually trying to get is that I guess we should only run this if we are in aws private, i.e. should we check hcp.spec.EndpointAccess?

control-plane-operator/controllers/awsprivatelink/awsprivatelink_controller.go
csrwng: The difference is that the returned error only contains the error code (it goes in the status of the resource) while the log line includes the full error returned from AWS.

I fixed the different error messages

control-plane-operator/controllers/awsprivatelink/awsprivatelink_controller.go
```diff
@@ -287,12 +377,18 @@ func (r *AWSEndpointServiceReconciler) Reconcile(ctx context.Context, req ctrl.R
...
 		}
-		completed, err := r.delete(ctx, awsEndpointService, r.ec2Client, r.route53Client)
+
+		ec2Client, route53Client, err := r.awsClientBuilder.getClients()
 		if err != nil {
-			return ctrl.Result{}, fmt.Errorf("failed to delete resource: %w", err)
-		}
-		if !completed {
-			return ctrl.Result{RequeueAfter: endpointServiceDeletionRequeueDuration}, nil
+			log.Error(err, "failed to get AWS client, skipping aws endpoint service cleanup")
```
enxebre, line 383: does this introduce a new scenario that wasn't possible before? we remove the finalizer if say a transient issue fails to get the client?

control-plane-operator/controllers/awsprivatelink/awsprivatelink_controller.go
enxebre: So for example here we'll log 
"failed to wait for security group to exist", "id", sgID"
and the controller runtime will log
"failed to wait for security group to exist (id: %s), code: %s", sgID, supportawsutil.AWSErrorCode(err)"

Seems unncecesary?

enxebre:
https://github.com/openshift/hypershift/pull/5179/files#r1856712444
https://github.com/openshift/hypershift/pull/5179#discussion_r1858009839
https://github.com/openshift/hypershift/pull/5179#discussion_r1858016237
/lgtm
/hold
feel free to cancel

control-plane-operator/controllers/awsprivatelink/awsprivatelink_controller.go
csrwng: The difference is that the latter will also go into the status of the AWSEndpointService

control-plane-operator/main.go
csrwng: Yup, we can. Since we need the hcp for that, I'll restore the code that fetched it on startup.

control-plane-operator/controllers/awsprivatelink/awsprivatelink_controller.go
csrwng: The only 2 possible errors are that the session cannot be created because env/cfg is incomplete or that the client builder was not initialized from a hcp. Either case is not recoverable with a retry.

PR #5191, opened 2024-11-26
OCPBUGS-44957: Use fleet manager configmaps to determine placeholder pod node antiaffinity

csrwng:
Before this commit, the placeholder controller looked at nodes with hosted cluster labels to determine which fleet manager labels it should exclude for placeholders. However, in the case when a hostedcluster control plane is cleaned up, but the HostedCluster CR itself has not been deleted due to the controlplane namespace not getting cleaned up, there are no longer any nodes with the hostedcluster label, but the hostedcluster is still associated with a particular pair label. Placeholder pods are not prevented from landing on nodes with this same pair label. These nodes cannot be used to allocate to a new HostedCluster because they are still associated with previous HostedCluster. This change modifies the placeholder controller to use the pair association configmaps in the placeholder namespace to determine which pair labels to avoid. This makes it consistent with the scheduler which also looks at the same configmaps to determine whether a fleetmanager pair label is available to use for a new cluster.

Fixes #[OCPBUGS-44957](https://issues.redhat.com/browse/OCPBUGS-44957)

**Checklist**

### DIFF
--- a/hypershift-operator/controllers/scheduler/placeholders.go
+++ b/hypershift-operator/controllers/scheduler/placeholders.go
@@ -66,12 +66,12 @@ func (r *PlaceholderScheduler) SetupWithManager(ctx context.Context, mgr ctrl.Ma
 			}
 			return &deployments, nil
 		},
-		listNodes: func(ctx context.Context, opts ...client.ListOption) (*corev1.NodeList, error) {
-			nodes := corev1.NodeList{}
-			if err := mgr.GetClient().List(ctx, &nodes, opts...); err != nil {
-				return nil, fmt.Errorf("could not list nodes: %w", err)
+		listConfigMaps: func(ctx context.Context, opts ...client.ListOption) (*corev1.ConfigMapList, error) {
+			configMaps := corev1.ConfigMapList{}
+			if err := mgr.GetClient().List(ctx, &configMaps, opts...); err != nil {
+				return nil, fmt.Errorf("could not list configmaps: %w", err)
 			}
-			return &nodes, nil
+			return &configMaps, nil
 		},
 	}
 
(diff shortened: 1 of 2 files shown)


hypershift-operator/controllers/scheduler/placeholders.go
```diff
@@ -66,12 +66,12 @@ func (r *PlaceholderScheduler) SetupWithManager(ctx context.Context, mgr ctrl.Ma
 			}
 			return &deployments, nil
 		},
-		listNodes: func(ctx context.Context, opts ...client.ListOption) (*corev1.NodeList, error) {
-			nodes := corev1.NodeList{}
-			if err := mgr.GetClient().List(ctx, &nodes, opts...); err != nil {
-				return nil, fmt.Errorf("could not list nodes: %w", err)
+		listConfigMaps: func(ctx context.Context, opts ...client.ListOption) (*corev1.ConfigMapList, error) {
+			configMaps := corev1.ConfigMapList{}
+			if err := mgr.GetClient().List(ctx, &configMaps, opts...); err != nil {
```
enxebre, line 71: should listConfigMaps include the namespace filter instead of passing it through in all invokations?

hypershift-operator/controllers/scheduler/placeholders.go
```diff
@@ -319,35 +323,37 @@ func (r *placeholderUpdater) reconcile(
...
 			}
 		}
 	}
 
-	if !pairedNodesAssignedToHostedClusters.Equal(existingPairedNodes) {
+	if !pairLabelsAssignedToHostedClusters.Equal(existingPairLabels) {
 		// the set of paired nodes on which hosted clusters have been scheduled has changed, we need to update the
 		// node affinity configuration on the deployment
-		return false, newDeployment(placeholderNamespace, placeholderSize, parsedIndex, pairedNodesAssignedToHostedClusters.UnsortedList()), nil
+		return false, newDeployment(placeholderNamespace, placeholderSize, parsedIndex, pairLabelsAssignedToHostedClusters.UnsortedList()), nil
```
enxebre, line 356: nit: using `pairLabelsAssignedToHostedClusters` and `pairedNodes` in `newDeployment` might be a bit confusing.

csrwng:
Previous failure on e2e-aws-417:
```
  - lastTransitionTime: "2024-11-26T16:04:01Z"
    message: "Throttling: Rate exceeded\n\tstatus code: 400, request id: 693a651c-2a52-4ed6-b15d-e58f398fe8e3"
    reason: AWSError
    status: "False"
    type: AWSEndpointAvailable
```
On awsendpointservice resource

PR #5197, opened 2024-11-28
OCPBUGS-44380: Reconcile proxy CA bundle into hosted cluster

csrwng:
The openshift controller manager operator syncs the CA bundle that was configured in the global proxy config into a configmap in the openshift-controller-manager namespace. This configmap is then used by the build controller to generate the CA that is mounted into builds for registry access. Because we don't run the controller manager operator in hypershift, this needs to be done by the HCCO. This code adds the reconciliation to the HCCO so that it puts the configmap in the right location.

Ref: https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/08c64512c055ae246b3a14f6d1088d39988c44fe/pkg/operator/usercaobservation/user_ca_observation_controller.go#L80-L90

Fixes #[OCPBUGS-44380](https://issues.redhat.com/browse/OCPBUGS-44380)

**Checklist**

### DIFF
--- a/control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
+++ b/control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
@@ -542,6 +542,11 @@ func (r *reconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Result
 		errs = append(errs, fmt.Errorf("failed to reconcile user cert CA bundle: %w", err))
 	}
 
+	log.Info("reconciling proxy CA bundle")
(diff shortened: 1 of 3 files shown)


control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -1397,6 +1402,29 @@ func (r *reconciler) reconcileUserCertCABundle(ctx context.Context, hcp *hyperv1
 	return nil
 }
```
enxebre, line 1404: let's follow up to document the func with the same PR rationale. So is easier to reason about future relocation of the logic across components

PR #5200, opened 2024-11-28
OCPBUGS-45182: Disable openstack-manila-csi-controllerplugin crash detection in e2e

EmilienM:

Whether it's on a Standalone OCP cluster or in a HostedCluster, we have a race condition
for Manila CSI with the NFS driver.
The result is openstack-manila-csi-controllerplugin crashes but eventually stabilizes after a restart.
We will solve the race condition.
For now we need to disable crash detection of openstack-manila-csi-controllerplugin to unblock the presubmit jobs.

### DIFF
--- a/test/e2e/util/util.go
+++ b/test/e2e/util/util.go
@@ -518,6 +518,11 @@ func EnsureNoCrashingPods(t *testing.T, ctx context.Context, client crclient.Cli
 				continue
 			}
 
+			// Temporary workaround for https://issues.redhat.com/browse/OCPBUGS-45182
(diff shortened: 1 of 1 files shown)


EmilienM:
/hold
we might fix it here: https://github.com/openshift/csi-operator/pull/334

EmilienM:
/hold
still testing a fix in csi-operator...

EmilienM:
We didn't come up with a proper fix in CSI operator but we realise that we might have more work to do on this one, so for now let's just tolerate the restart. We'll provide a cleaner fix in 4.19.
/hold cancel

EmilienM:
/close
I'll combine this fix into https://github.com/openshift/hypershift/pull/5202

PR #5211, opened 2024-12-02
OCPBUGS-45265: Use /livez/ping endpoint rather than /version

jonesbr17:
**What this PR does / why we need it**: /livez/ping is faster and more reliable than /version. We should prefer /livez/ping when possible.

Fixes # [OCPBUGS-45265](https://issues.redhat.com/browse/OCPBUGS-45265)

**Checklist**


hypershift-operator/controllers/nodepool/apiserver-haproxy/haproxy.cfg
```diff
@@ -23,7 +23,7 @@ frontend local_apiserver
 backend remote_apiserver
   mode tcp
   log global
-  option httpchk GET /version
+  option httpchk GET /livez/ping
```
rtheis, line 26: We cannot simply switch to `/livez/ping` because it isn't enabled for unauthenticated users by default.

rtheis (changes requested):

hypershift-operator/controllers/nodepool/haproxy.go
```diff
@@ -403,6 +414,63 @@ func generateHAProxyStaticPod(name, image, internalAPIAddress, configPath string
 	}
 }
 
+func generateNodePublicInfoViewerCR() func() ([]byte, error) {
```
rtheis, line 417: I don't know if this is the correct location to do this.  This RBAC is cluster scope and not nodepool scoped.  I suspect it belongs in the HCCO along side reconciling the in cluster Kubernetes API server endpoint.  Please ask Red Hat for their opinion.

rtheis:
@jonesbr17 please also fix the Jira.

hypershift-operator/controllers/nodepool/haproxy.go
rtheis: @jonesbr17 any updates on this?

sjenning: Yes, this is true.  HCCO should reconcile this.  Good catch @rtheis :+1:

sjenning:
/hold
Until the CR/CRB reconcile is moved to the HCCO.  I doubt that this works at all in its current state.

rtheis:
/remove-hold 

CR/CRB reconcile has been moved to the HCCO.

rtheis:
@jonesbr17 https://github.com/openshift/hypershift/pull/5211#issuecomment-2584791238 ptal

xueqzhan:
/payload-job periodic-ci-openshift-hypershift-release-4.19-periodics-e2e-aws-ovn-conformance

PR #5202, opened 2024-11-29
OCPBUGS-45189, OCPBUGS-45182: e2e/openstack: adjustments for missing manila apps & pods

EmilienM:
**What this PR does / why we need it**:

* In `EnsurePodsWithEmptyDirPVsHaveSafeToEvictAnnotations`, we need both
  `openstack-manila-csi` and `manila-csi-driver-controller`.
  See: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_hypershift/5200/pull-ci-openshift-hypershift-main-e2e-openstack/1862228917390151680/artifacts/e2e-openstack/hypershift-openstack-e2e-execute/artifacts/TestCreateCluster/namespaces/e2e-clusters-whcqh-example-nqcr9/apps/deployments/openstack-manila-csi-controllerplugin.yaml
  for an example. We were missing `openstack-manila-csi`.
* In `EnsureSATokenNotMountedUnlessNecessary`, we need to add
  `csi-driver` and `csi-driver-nfs`. See:
  https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_hypershift/5200/pull-ci-openshift-hypershift-main-e2e-openstack/1862228917390151680/artifacts/e2e-openstack/hypershift-openstack-e2e-execute/artifacts/TestCreateCluster/namespaces/e2e-clusters-whcqh-example-nqcr9/core/pods/openstack-manila-csi-controllerplugin-74567d9c96-m2m95.yaml
  These 2 containers have a mount for `kube-api-access-`.
* Whether it's on a Standalone OCP cluster or in a HostedCluster, we have a race condition for Manila CSI with the NFS driver. The result is openstack-manila-csi-controllerplugin crashes but eventually stabilizes after a restart. We will solve the race condition.
  For now we need to disable crash detection of openstack-manila-csi-controllerplugin to unblock the presubmit jobs.

PR #5215, opened 2024-12-02
NO-JIRA: e2e/utils: improve logs for EnsureSATokenNotMountedUnlessNecessary

EmilienM:

When `EnsureSATokenNotMountedUnlessNecessary` fails, it's hard to know what pod is problematic.
This PR adds that kind of information in the logs.


EmilienM:
/close
I moved the commit to #5202

PR #5216, opened 2024-12-02
OCPBUGS-44967: Pass only the certificate name for CNO deployment

bryan-cox:
Pass only the certificate name for the CNO deployment for managed Azure.
CNO uses its own certificate path here, https://github.com/openshift/cluster-network-operator/blob/7736bfe37f1276f771fcef03077f6d840eb6b862/pkg/network/cloud_network.go#L23.
This is combined with the certificate name here, https://github.com/openshift/cluster-network-operator/blob/7736bfe37f1276f771fcef03077f6d840eb6b862/pkg/network/cloud_network.go#L111.

Fixes OCPBUGS-44967

**Checklist**

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/cno/clusternetworkoperator.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/cno/clusternetworkoperator.go
@@ -622,11 +622,20 @@ if [[ -n $sc ]]; then kubectl --kubeconfig $kc delete --ignore-not-found validat
 	// to use on the CNCC deployment.
 	if azureutil.IsAroHCP() {
 		dep.Spec.Template.Spec.Containers[0].Env = append(dep.Spec.Template.Spec.Containers[0].Env,
-			azureutil.CreateEnvVarsForAzureManagedIdentity(params.AzureClientID, params.AzureTenantID, params.AzureCertificateName)...)
-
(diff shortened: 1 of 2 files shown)


control-plane-operator/controllers/hostedcontrolplane/cno/clusternetworkoperator.go
```diff
@@ -622,9 +622,18 @@ if [[ -n $sc ]]; then kubectl --kubeconfig $kc delete --ignore-not-found validat
...
+			corev1.EnvVar{
+				Name:  config.ManagedAzureClientIdEnvVarKey,
+				Value: params.AzureClientID,
+			},
+			corev1.EnvVar{
+				Name:  config.ManagedAzureTenantIdEnvVarKey,
+				Value: params.AzureTenantID,
+			},
+			corev1.EnvVar{
+				Name:  "ARO_HCP_CLIENT_CERTIFICATE_NAME",
```
enxebre, line 634: can we have this in constants as well?

enxebre: ARO_HCP_SECRET_PROVIDER_CLASS / ARO_HCP_CLIENT_CERTIFICATE_NAME

PR #5217, opened 2024-12-02
NO-JIRA: cpov2: revert oauth reconciliation

sjenning:
This is causing a hot rollout of the `oauth-openshift` deployment leading to a very high flake rate in CI.

Hot change is the addition/removal of the `hypershift.openshift.io/kubeadmin-secret-hash` annotation that is added by the HCCO and apparently stomped by the new reconciliation logic.


sjenning:
In the artifacts for `TestCreateClusterV2`
```
apiVersion: apps/v1
kind: Deployment
metadata:
  annotations:
    deployment.kubernetes.io/revision: "2"
  creationTimestamp: "2024-12-02T22:43:27Z"
  generation: 2 <---
...
  name: oauth-openshift
  namespace: e2e-clusters-mltlr-example-69ccb
```
/hold cancel

PR #5220, opened 2024-12-03
OCPBUGS-42320: Prevent IgnitionServer from flooding the API server with patch requests

jparrill:
This patch addresses the issue of excessive patch attempts in a short period, which could overwhelm the KAS in the control plane.  A new safeguards have been implemented to prevent this scenario which involves a conditional check before executing patch commands that could potentially flood the API.  

Fixes #[OCPBUGS-42320](https://issues.redhat.com/browse/OCPBUGS-42320)

### DIFF
--- a/ignition-server/controllers/tokensecret_controller.go
+++ b/ignition-server/controllers/tokensecret_controller.go
@@ -249,12 +249,18 @@ func (r *TokenSecretReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	compressedConfig := tokenSecret.Data[TokenSecretConfigKey]
 	config, err := util.DecodeAndDecompress(compressedConfig)
 	if err != nil {
+		errWithFullMsg := fmt.Errorf("failed to decode and decompress config: %w", err)
+		if hasSameReasonAndMessage(tokenSecret, InvalidConfigReason, errWithFullMsg) {
+			return ctrl.Result{}, errWithFullMsg
+		}
+
 		patch := tokenSecret.DeepCopy()
 		patch.Data[TokenSecretReasonKey] = []byte(InvalidConfigReason)
-		patch.Data[TokenSecretMessageKey] = []byte(fmt.Sprintf("Failed to decode and decompress config: %s", err))
+		patch.Data[TokenSecretMessageKey] = []byte(errWithFullMsg.Error())
 		if err := r.Client.Patch(ctx, patch, client.MergeFrom(tokenSecret)); err != nil {
 			return ctrl.Result{}, fmt.Errorf("failed to patch tokenSecret with payload content: %w", err)
 		}
+
 		return ctrl.Result{}, err
 	}
 
@@ -274,12 +280,20 @@ func (r *TokenSecretReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		return payload, err
 	}()
 	if err != nil {
+		// This patch could flood the API server, so we should only do it when the reason or message is different from the current one.
+		// More info here: https://issues.redhat.com/browse/OCPBUGS-42320.
+		errWithFullMsg := fmt.Errorf("failed to generate payload: %w", err)
+		if hasSameReasonAndMessage(tokenSecret, InvalidConfigReason, errWithFullMsg) {
+			return ctrl.Result{}, errWithFullMsg
+		}
+
 		patch := tokenSecret.DeepCopy()
 		patch.Data[TokenSecretReasonKey] = []byte(InvalidConfigReason)
-		patch.Data[TokenSecretMessageKey] = []byte(fmt.Sprintf("Failed to generate payload: %s", err))
+		patch.Data[TokenSecretMessageKey] = []byte(errWithFullMsg.Error())
 		if err := r.Client.Patch(ctx, patch, client.MergeFrom(tokenSecret)); err != nil {
 			return ctrl.Result{}, fmt.Errorf("failed to patch tokenSecret with payload content: %w", err)
 		}
+
 		return ctrl.Result{}, err
 	}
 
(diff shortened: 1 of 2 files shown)


enxebre:
/hold
This would delay reconciliation for any error (including transient errors) which we don't want to.
Controller runtime already has a builtin mechanism to backoff on failures [1] that uses client-go default rate limiter [2] which we are already exercising by default. It rate limites both per item and it has a BucketRateLimiter for worst case failure of multiple items like this one. Besides, we are using a cached client to get resources and it seems the flooding request are .patch. Based on this I suggest the following:
- Try to capture and type this specific disconnected failure and make it terminal so controller runtime doesn't requeue.
- Make the patch logic a bit smarter so only make the write request if really needed.
- When this happens "1. Start a disconnected HCP cluster with incorrect mirror-registry information
" we must signal that in the HC status somehow.

[1] https://github.com/kubernetes-sigs/controller-runtime/blob/9fd890c90e9f9ce92c61400377697e46d57c5166/pkg/internal/controller/controller.go#L321-L326
[2] https://github.com/kubernetes/client-go/blob/37045084c2aa82927b0e5ffc752861430fd7e4ab/util/workqueue/default_rate_limiters.go#L50-L56

jparrill:
Hey @enxebre:

> - Try to capture and type this specific disconnected failure and make it terminal so controller runtime doesn't requeue or increase requeue retry.

The implementation capturing the current error is a workaround until https://github.com/openshift/hypershift/pull/5168 got merged. In that PR I've get rid of all these functions and now they are part of the imageMetadataPovider. So the plan is to put a follow up PR with custom errors to be easy to be identify.

> - Make the patch logic a bit smarter so only make the write request if really needed.

Done

> - When this happens "1. Start a disconnected HCP cluster with incorrect mirror-registry information" we must signal that in the HC status somehow.

This it's not necessary because a new instance of the metadataProvider will be created in for each component that pull images from a registry.

WDYT?

ignition-server/controllers/tokensecret_controller.go
```diff
@@ -249,11 +254,13 @@ func (r *TokenSecretReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	compressedConfig := tokenSecret.Data[TokenSecretConfigKey]
 	config, err := util.DecodeAndDecompress(compressedConfig)
 	if err != nil {
-		patch := tokenSecret.DeepCopy()
-		patch.Data[TokenSecretReasonKey] = []byte(InvalidConfigReason)
-		patch.Data[TokenSecretMessageKey] = []byte(fmt.Sprintf("Failed to decode and decompress config: %s", err))
-		if err := r.Client.Patch(ctx, patch, client.MergeFrom(tokenSecret)); err != nil {
-			return ctrl.Result{}, fmt.Errorf("failed to patch tokenSecret with payload content: %w", err)
+		if !hasSameReasonAndMessage(tokenSecret, InvalidConfigReason, fmt.Sprintf("Failed to decode and decompress config: %s", err)) {
```
enxebre, line 257: nit: can we instead return early if hasSameReasonAndMessage, and otherwise run the bulk logic

ignition-server/controllers/tokensecret_controller.go
```diff
@@ -274,12 +281,24 @@ func (r *TokenSecretReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
 	}()
 	if err != nil {
-		patch := tokenSecret.DeepCopy()
-		patch.Data[TokenSecretReasonKey] = []byte(InvalidConfigReason)
-		patch.Data[TokenSecretMessageKey] = []byte(fmt.Sprintf("Failed to generate payload: %s", err))
-		if err := r.Client.Patch(ctx, patch, client.MergeFrom(tokenSecret)); err != nil {
-			return ctrl.Result{}, fmt.Errorf("failed to patch tokenSecret with payload content: %w", err)
+		// This patch could flood the API server, so we should only do it when the reason or message is different from the current one.
+		// More info here: https://issues.redhat.com/browse/OCPBUGS-42320.
+		if !hasSameReasonAndMessage(tokenSecret, InvalidConfigReason, fmt.Sprintf("Failed to generate payload: %s", err)) {
```
enxebre, line 286: nit: can we instead return early if hasSameReasonAndMessage, and otherwise run the bulk logic

enxebre:
>The current implementation capturing the current error in this PR is a workaround until https://github.com/openshift/hypershift/pull/5168 got merged. In that PR I've get rid of all these functions and now they are part of the imageMetadataPovider. So the plan is to put a follow up PR with custom errors to be easy to be identify.

Thanks! Let's scope this PR to only the patch logic then, that solves the problem of overloading kas. Then add proper error handling once the above is true.

>This it's not necessary because a new instance of the metadataProvider will be created in for each component that pull images from a registry.

If there's an issue pulling images that requires manual intervention like the one in the jira, this needs to be signaled to the user via .Status

jparrill:
Thanks for the review @enxebre. I will:

- Signal the error pulling image to the user via .status
- Better error handling 

in this https://github.com/openshift/hypershift/pull/5168 PR.

ignition-server/controllers/tokensecret_controller.go
```diff
@@ -249,12 +249,17 @@ func (r *TokenSecretReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	compressedConfig := tokenSecret.Data[TokenSecretConfigKey]
 	config, err := util.DecodeAndDecompress(compressedConfig)
 	if err != nil {
+		if hasSameReasonAndMessage(tokenSecret, InvalidConfigReason, fmt.Errorf("Failed to decode and decompress config: %w", err)) {
```
enxebre, line 252: This changing from upper to lower. Error messages should start with lower case, "failed..."
Also can we put this into a var and reuse it?

ignition-server/controllers/tokensecret_controller.go
```diff
@@ -274,12 +279,19 @@ func (r *TokenSecretReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		return payload, err
 	}()
 	if err != nil {
+		// This patch could flood the API server, so we should only do it when the reason or message is different from the current one.
+		// More info here: https://issues.redhat.com/browse/OCPBUGS-42320.
+		if hasSameReasonAndMessage(tokenSecret, InvalidConfigReason, fmt.Errorf("Failed to generate payload: %w", err)) {
+			return ctrl.Result{}, fmt.Errorf("Return early, failed to generate payload: %w", err)
```
enxebre, line 285: same than above. also can we keep the original message and drop "return early"?

enxebre:
thanks!
/lgtm

PR #5223, opened 2024-12-03
NO-JIRA: openstack/doc: add a requirement for Octavia

EmilienM:

LB has to be functional on the mgt cluster for kube-api.


mandre (commented):
/lgtm

PR #2867, opened 2023-08-01
MULTIARCH-3708: PowerVS - Fix cluster deletion when existing resources passed

dharaneeshvrd:
**What this PR does / why we need it**:
[MULTIARCH-3708](https://issues.redhat.com//browse/MULTIARCH-3708): PowerVS - Fix cluster deletion when existing resources passed

**Which issue(s) this PR fixes** *(optional, use `fixes #<issue_number>(, fixes #<issue_number>, ...)` format, where issue_number might be a GitHub issue, or a Jira story*:
Fixes #

**Checklist**
- [ ] Subject and description added to both, commit and PR.
- [ ] Relevant issues have been referenced.
- [ ] This change includes docs. 
- [ ] This change includes unit tests.

PR #5227, opened 2024-12-03
HOSTEDCP-2232: Destroy DHCP Server before PowerVS Cloud Instance deletion

Neha-dot-Yadav:
This PR adds functionality to delete DHCP server created in PowerVS Cloud Instance before the actual instance deletion.
We need it, so that we do not rely on PowerVS Cloud Instance's recursive delete functionality to delete all other dependent resources.

**Checklist**
- Subject and description added to both, commit and PR.
- This PR was tested locally.

### DIFF
--- a/cmd/infra/powervs/destroy.go
+++ b/cmd/infra/powervs/destroy.go
@@ -47,7 +47,6 @@ const (
 	powerVSCloudInstanceRemovedState = "removed"
 	powerVSJobCompletedState         = "completed"
 	powerVSJobFailedState            = "failed"
-	dhcpInstanceShutOffState         = "SHUTOFF"
 
 	// Resource name prefix
 	vpcLbNamePrefix = "kube"
@@ -268,7 +267,12 @@ func (options *DestroyInfraOptions) DestroyInfra(ctx context.Context, logger log
 	}
 
 	if !skipPowerVs {
-		if err = destroyPowerVsCloudInstance(ctx, logger, options, infra, powerVsCloudInstanceID, session); err != nil {
+		// destroy DHCP server first, to not rely on PowerVs Cloud Instance recursive delete
+		// NOTE: not calling dhcp destroy inside the destroyPowerVsCloudInstance, to avoid Private Network deletion getting stuck on modified
+		if err = destroyPowerVsDhcpServer(ctx, logger, powerVsCloudInstanceID, options, session); err != nil {
+			logger.Error(err, "error destroying DhcpServer")
+		}
+		if err = destroyPowerVsCloudInstance(ctx, logger, options, powerVsCloudInstanceID); err != nil {
 			errL = append(errL, fmt.Errorf("error destroying powervs cloud instance: %w", err))
 			logger.Error(err, "error destroying powervs cloud instance")
 		}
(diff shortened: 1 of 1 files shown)


cmd/infra/powervs/destroy.go
```diff
@@ -371,8 +376,64 @@ func deleteSecrets(name, namespace, cloudInstanceID string, accountID string, re
...
+	if err != nil {
+		return err
+	}
+
+	if dhcpServers == nil || len(dhcpServers) < 1 {
+		logger.Info("No DHCP servers available to delete in PowerVS")
+		return nil
+	}
+
+	dhcpID := *dhcpServers[0].ID
```
srivastav-abhishek, line 393: Is there a possibility that there are multiple dhcp servers in a workspace ? We may have to iterate through all of it and delete.

srivastav-abhishek (changes requested):

cmd/infra/powervs/destroy.go
```diff
@@ -371,8 +376,64 @@ func deleteSecrets(name, namespace, cloudInstanceID string, accountID string, re
...
+// destroyPowerVsDhcpServer destroying powervs dhcp server
+func destroyPowerVsDhcpServer(ctx context.Context, logger logr.Logger, cloudInstanceID string, session *ibmpisession.IBMPISession) error {
+	client := instance.NewIBMPIDhcpClient(ctx, session, cloudInstanceID)
+
+	dhcpServers, err := client.GetAll()
+	if err != nil {
+		return err
+	}
+
+	if dhcpServers == nil || len(dhcpServers) < 1 {
```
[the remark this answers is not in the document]
Neha-dot-Yadav: Changed it, thanks!

cmd/infra/powervs/destroy.go
Neha-dot-Yadav: No, there will be only one dhcp server.
https://github.com/openshift/hypershift/blob/main/cmd/infra/powervs/create.go#L130

cmd/infra/powervs/destroy.go
```diff
@@ -371,8 +376,64 @@ func deleteSecrets(name, namespace, cloudInstanceID string, accountID string, re
 	return nil
 }
 
+// destroyPowerVsDhcpServer destroying powervs dhcp server
+func destroyPowerVsDhcpServer(ctx context.Context, logger logr.Logger, cloudInstanceID string, session *ibmpisession.IBMPISession) error {
```
dharaneeshvrd, line 380: You need to consider the existing resource use case here. So you can't delete the resource always.

cmd/infra/powervs/destroy.go
```diff
@@ -371,8 +376,64 @@ func deleteSecrets(name, namespace, cloudInstanceID string, accountID string, re
...
+	f := func() (bool, error) {
+		dhcpInstance, err := instanceClient.Get(dhcpID)
+		if err != nil {
+			if err = isNotRetryableError(err, timeoutErrorKeywords); err == nil {
+				return false, nil
+			}
+			errMsg := err.Error()
+			// when instance becomes does not exist, infra destroy can proceed
+			// when there is flaky, internal server error in pvm-instance get call, retry destroy
+			if strings.Contains(errMsg, "pvm-instance does not exist") || strings.Contains(errMsg, "unable to get the server volume attachment list: Internal Server Error") {
```
dharaneeshvrd, line 413:
```suggestion
			if strings.Contains(errMsg, "pvm-instance does not exist") || strings.Contains(errMsg, "unable to get the server volume attachment list") {
```

dharaneeshvrd: You are returning `true, nil`, how can it retry when it gets this flaky error?

cmd/infra/powervs/destroy.go
```diff
@@ -371,8 +376,64 @@ func deleteSecrets(name, namespace, cloudInstanceID string, accountID string, re
...
+			}
+			return false, err
+		}
+
+		if dhcpInstance == nil {
+			return false, fmt.Errorf("dhcpInstance is nil")
+		}
+
+		logger.Info("Waiting for DhcpServer to destroy", "id", *dhcpInstance.PvmInstanceID, "status", *dhcpInstance.Status)
+		if *dhcpInstance.Status == dhcpInstanceDeletedState || *dhcpInstance.Status == dhcpServiceErrorState {
```
dharaneeshvrd, line 425: Why are you considering `dhcpServiceErrorState` here?

dharaneeshvrd (changes requested):

cmd/infra/powervs/destroy.go
Neha-dot-Yadav: Added the check!
Btw, the thought process behind not considering it was, shouldn't `skipPowerVs ` be set to `true` in case when the `powerVsCloudInstanceID` is user provided. So, if it is true, these lines should not get called and so, I left it.

cmd/infra/powervs/destroy.go
Neha-dot-Yadav: Rectified it, thanks!

cmd/infra/powervs/destroy.go
Neha-dot-Yadav: Removed it!
I actually took reference from your PR - https://github.com/openshift/hypershift/pull/2867/files#diff-f4fe34c4d57d805e9983ff19def305eeb4bd6ab244f96b4cc66076361ed478d4L388
Thought this check is added, so that the PowerVS instance deletion can proceed even if the dhcp server deletion is in error state, as anyways, the DHCP server will get deleted as  PowerVS instance recursive delete operation.

cmd/infra/powervs/destroy.go
```diff
@@ -371,8 +376,78 @@ func deleteSecrets(name, namespace, cloudInstanceID string, accountID string, re
 	return nil
 }
 
+// destroyPowerVsDhcpServer destroying powervs dhcp server
+func destroyPowerVsDhcpServer(ctx context.Context, logger logr.Logger, cloudInstanceID string, options *DestroyInfraOptions, session *ibmpisession.IBMPISession) error {
+	// Attempt dhcp server deletion only when cloud instance is created by hypershift
+	// Do not clean up, if PowerVS instance is user provided
+	if options.CloudInstanceID == "" {
```
dharaneeshvrd, line 383:
```suggestion
	if options.CloudInstanceID != "" { 
	    return nil
	}
	client := instance.NewIBMPIDhcpClient(ctx, session, cloudInstanceID)
```
We can do an early return to keep the code more readable IMO.

cmd/infra/powervs/destroy.go
```diff
@@ -371,8 +376,78 @@ func deleteSecrets(name, namespace, cloudInstanceID string, accountID string, re
...
+
+		if len(dhcpServers) == 0 {
+			logger.Info("No DHCP servers available to delete in PowerVS")
+			return nil
+		}
+
+		dhcpID := *dhcpServers[0].ID
+		logger.Info("Deleting DHCP server", "id", dhcpID)
+		err = client.Delete(dhcpID)
+		if err != nil {
```
dharaneeshvrd, line 399:
```suggestion
		if err = client.Delete(dhcpID); err != nil {
```

cmd/infra/powervs/destroy.go
Neha-dot-Yadav: added it, thanks!

cmd/infra/powervs/destroy.go
```diff
@@ -370,8 +374,74 @@ func deleteSecrets(name, namespace, cloudInstanceID string, accountID string, re
...
+			case strings.Contains(errMsg, "unable to get the server volume attachment list"):
+				// There is flaky internal server error in pvm-instance get call; retry destroy
+				return false, nil
+
+			default:
+				return false, err
+			}
+		}
+
+		if dhcpInstance == nil {
```
dharaneeshvrd, line 431: Can you pls explain on which it would enter this block?
Looks not needed to me.

Neha-dot-Yadav: It will enter this block if the dhcpid is wrong or empty.
Tested it using this block -
```
f := func() (bool, error) {
        logger.Info("Testing if it returns nil in case of dhcpID not being present or it returns error")
        dhcpInstanceTest, err := instanceClient.Get("c8a265a9-57bd-4156-baee-7ea12544d280f")
        logger.Info("Value of dhcpInstanceTest", "dhcpInstanceTest", dhcpInstanceTest)
        logger.Info("Ending this testing")

```

And it returned null when I gave dhcp id  which does not exist instead of giving error -
`2025-01-09T17:36:15+05:30	INFO	Testing if it returns nil in case of dhcpID not being present or it returns error
2025-01-09T17:36:19+05:30	INFO	Value of dhcpInstanceTest	{"dhcpInstanceTest": null}
2025-01-09T17:36:19+05:30	INFO	Ending this `testing``

dharaneeshvrd: Yes, but I think earlier delete would throw this error and it won't come to the get block itself. Can you please check?

dharaneeshvrd (approved):
/lgtm

PR #5229, opened 2024-12-04
NO-JIRA: test/e2e: detect rapidly updating deployments

sjenning:


test/e2e/util/util.go
```diff
@@ -548,6 +548,23 @@ func NoticePreemptionOrFailedScheduling(t *testing.T, ctx context.Context, clien
 	})
 }
 
+func EnsureNoRapidDevelopmentRollouts(t *testing.T, ctx context.Context, client crclient.Client, hostedCluster *hyperv1.HostedCluster) {
+	const maxAllowedGeneration = 30
```
enxebre, line 552: isn't this too permissive?

sjenning: Probably, I just didn't want false positives. Looking at deployments from `TestUpgradeControlPlane`, seems the highest generation is `4`.  Could tighten it to 10 to be sensitive.

sjenning:
Updated PR:
- `s/EnsureNoRapidDevelopmentRollouts/EnsureNoRapidDeploymentRollouts`
- tightened `maxAllowedGeneration` to `10`

PR #5257, opened 2024-12-08
HOSTEDCP-2206: Improve Unit Test Runtime

bryan-cox:
This PR:
- reduces the unit test runtime by ~90% from my tests locally
- reduces the CI runtime by ~72%*
- fixes some minor codespelling mistakes picked up by `pre-commit` hooks

_\*calculated from 8 runs of this PR vs 8 successful runs of the unit tests without this PR_

---

The majority of the time reduction is due from taking count from 25 to 1. Setting this value to 1 ensures all the unit tests are always run.
```
% go help testflag
...
	-count n
	    Run each test, benchmark, and fuzz seed n times (default 1).
	    If -cpu is set, run n times for each GOMAXPROCS value.
	    Examples are always run once. -count does not apply to
	    fuzz tests matched by -fuzz.
```

This PR also sets the parallel flag on unit tests according to the number of cores on the machine running the test. More info on the parallel flag can be found here:
```
% go help testflag
...
	-parallel n
	    Allow parallel execution of test functions that call t.Parallel, and
	    fuzz targets that call t.Parallel when running the seed corpus.
	    The value of this flag is the maximum number of tests to run
	    simultaneously.
	    While fuzzing, the value of this flag is the maximum number of
	    subprocesses that may call the fuzz function simultaneously, regardless of
	    whether T.Parallel is called.
	    By default, -parallel is set to the value of GOMAXPROCS.
	    Setting -parallel to values higher than GOMAXPROCS may cause degraded
	    performance due to CPU contention, especially when fuzzing.
	    Note that -parallel only applies within a single test binary.
	    The 'go test' command may run tests for different packages
	    in parallel as well, according to the setting of the -p flag
	    (see 'go help build').
```

---

Fixes [HOSTEDCP-2206](https://issues.redhat.com/browse/HOSTEDCP-2206)

**Checklist**

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -7,6 +7,7 @@ repos:
       - id: trailing-whitespace
         args:
           - --markdown-linebreak-ext=md
+        exclude:  docs/content/reference/api.md
   - repo: local
     hooks:
       - id: make-codespell
--- a/Makefile
+++ b/Makefile
@@ -118,19 +118,19 @@ hypershift-api: $(CONTROLLER_GEN) $(CODE_GEN)
 	rm -rf cmd/install/assets/hypershift-operator/*
 
 	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./api/..."
-	
+
 	# These consolidate with the 3 steps used to generate CRDs by openshift/api.
 	$(CODE_GEN) empty-partial-schemas --base-dir ./api/hypershift/v1beta1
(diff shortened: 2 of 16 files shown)


bryan-cox:
Documenting this for future us 😄 

We discussed leaving count at 1 for now during a grooming call today. If there is an issue with a unit test, that would have been caught with count being higher than one, we will parameterize the count value to 1 locally and change it to something like 10 in the unit tests.

PR #5262, opened 2024-12-09
OSASINFRA-3685: openstack: tag CAPI resources with InfraID

EmilienM:
By default, OpenStack resources created by CAPI will be tagged (in the
OpenStack way) with `openshiftClusterID=<infraID>`.

`openshiftClusterID` is a well-known tag prefix in ShiftOnStack IPI.

Additional tags can be added via the `HostedCluster.Spec.Platform.OpenStack.Tags` field
when creating the HostedCluster with a given YAML file.

### DIFF
--- a/hypershift-operator/controllers/hostedcluster/internal/platform/openstack/openstack.go
+++ b/hypershift-operator/controllers/hostedcluster/internal/platform/openstack/openstack.go
@@ -155,7 +155,11 @@ func reconcileOpenStackClusterSpec(hcluster *hyperv1.HostedCluster, openStackClu
 	openStackClusterSpec.ManagedSecurityGroups = &capo.ManagedSecurityGroups{
 		AllNodesSecurityGroupRules: defaultWorkerSecurityGroupRules(machineNetworksToStrings(machineNetworks)),
 	}
-	openStackClusterSpec.Tags = openStackPlatform.Tags
+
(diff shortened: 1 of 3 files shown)


hypershift-operator/controllers/hostedcluster/internal/platform/openstack/openstack.go
```diff
@@ -155,7 +155,13 @@ func reconcileOpenStackClusterSpec(hcluster *hyperv1.HostedCluster, openStackClu
 	openStackClusterSpec.ManagedSecurityGroups = &capo.ManagedSecurityGroups{
 		AllNodesSecurityGroupRules: defaultWorkerSecurityGroupRules(machineNetworksToStrings(machineNetworks)),
 	}
-	openStackClusterSpec.Tags = openStackPlatform.Tags
+
+	// Users are permitted to specify additional tags to be applied to the OpenStack resources
+	// but the default tag will be compliant with the OpenShift Cluster ID.
+	openStackClusterSpec.Tags = []string{"openshiftClusterID=" + hcluster.Spec.InfraID}
+	if len(openStackPlatform.Tags) > 0 {
```
csrwng, line 162: nit: no need to check length of slice

csrwng (commented):
Just one nit, otherwise lgtm

mdbooth:
/lgtm

Users will appreciate this.

PR #5265, opened 2024-12-10
NO-JIRA: contrib: update route53 zone cleaning utility

sjenning:
This utility is now installed in the `hypershift` namespace of the root CI cluster and a CronJob (daily) to clean up CI related hosted zones and records from the `ci` and `service.ci` zones.  The intent is to be non-disruptive to running tests by extracting the list of infraIDs in-use from the VPC list, then use that to find CI related zones/records that don't match an in-use infraID and remove them.

### DIFF
--- a/contrib/cleanzones/manifests/02-cronjob.yaml
+++ b/contrib/cleanzones/manifests/02-cronjob.yaml
@@ -1,0 +1,34 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: cleanzones
+  namespace: hypershift
+spec:
+  schedule: "0 0 * * *"
+  jobTemplate:
+    spec:
+      template:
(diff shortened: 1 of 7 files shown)


contrib/cleanzones/manifests/02-cronjob.yaml
```diff
@@ -0,0 +1,34 @@
...
+  namespace: hypershift
+spec:
+  schedule: "0 0 * * *"
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          containers:
+          - name: cleanzones
+            image: quay.io/sjenning/cleanzones:latest
```
csrwng, line 14: should we use quay.io/hypershift ?

csrwng (commented):

PR #5268, opened 2024-12-11
NO-JIRA: Update tekton task to latest working ref

devguyio:
Update tekton task to latest working ref. This is needed because currently a tag was deleted from `redhat-appstudio-tekton-catalog` and we need to do update this manually. 

**Checklist**


jparrill:
I think this PR is not necessary anymore. Looks like It's working now.

/hold

devguyio:
tags got fixed on the registry, no longer needed 

/close

PR #5286, opened 2024-12-13
NO-JIRA: Move oauth kubeadminsecret hash annotation logic from HCCO to CPO

muraee:
This avoids the conflict between HCCO and CPO reconciling the same deployment. All HCP components reconciliation should be done in the CPO

Fixes #

**Checklist**


bryan-cox:
Can you add the why/description to the commit? Otherwise lgtm.

PR #5287, opened 2024-12-13
CNTRLPLANE-10: Konflux build CPO only when needed

celebdor:
Changes that do not affect the Control Plane Operator should not trigger Konflux builds of the Control Plane Operator

Fixes #CNTRLPLANE-10

**Checklist**


bryan-cox:

PR #5307, opened 2024-12-16
ARO-13398, HOSTEDCP-1972: hypershift-operator: allow opting into pull secret rotation

stevekuznetsov:
There is no technical reason that we cannot react to changes to referenced secrets for pull credentials; allow users to opt into this behavior by labelling their secret.

A follow-up will make this change not cause node rollout.


enxebre:
>A follow-up will make this change not cause node rollout.

Please note that this change won't trigger a rollout either as it is now. Only changing the [ref will](https://github.com/openshift/hypershift/blob/ef8596d4d69a53eb60838ae45ffce2bca0bfa3b2/hypershift-operator/controllers/nodepool/config.go#L116).

So this PR will be no-op from ux pov atm. Can we include at commit in the PR to implement the inplace rollout?

enxebre:
>@enxebre ah right - I can update the hash in this PR as well - WDYT?

Unfortunately that's a no go because a non backward compatible change in how we build the hash would transparently cause a global rollout for the whole cluster fleet as soon as this change gets shipped to a management cluster, which we don't want to.
That's why this is sensitive and we need to think carefully about the implementation.

/ok-to-test

api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/hostedclusters.hypershift.openshift.io/AAA_ungated.yaml
```diff
@@ -3513,17 +3513,19 @@ spec:
...
-                  Changing the content of the secret inplace will not trigger a rollout and might result in unpredictable behaviour.
+                  Add the hypershift.openshift.io/reload=true label to the referenced pull secret to enable reloading of the content
+                  of the secret in-place.
                 properties:
                   name:
-                    default: ""
-                    description: |-
-                      Name of the referent.
-                      This field is effectively required, but due to backwards compatibility is
-                      allowed to be empty. Instances of this type with an empty value here are
```
stevekuznetsov, line 3523: This is technically backwards incompat as well, unclear - would have been invalid anyway?

stevekuznetsov: (I couldn't embed `corev1.LocalObjectReference` as it somehow broke the applyconfiggenerator)

stevekuznetsov:
I think I need another `/ok-to-test`? weird.

stevekuznetsov:
Nvm, was just impatient for Prow

stevekuznetsov:
e2e failing with

```
    util.go:145: ImagePullCredentialsValid=False: InvalidConfiguration(Referenced image pull credential e2e-clusters-xvwrz/example-49p4t-pull-secret is invalid - ensure the secret is of dockerconfigjson type and has the appropriate content)
```

 That's my new condition ... how was this working in the past with invalid pull credentials? huh?

stevekuznetsov:
@enxebre ready for review, the previous e2e was a known flake

PR #5309, opened 2024-12-17
OCPBUGS-46483: Correct ARO csi drivers cloud config

Phaow:
- Correct ARO csi drivers cloud config

Fixes # [OCPBUGS-46483](https://issues.redhat.com/browse/OCPBUGS-46483)

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/cloud/azure/providerconfig.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/cloud/azure/providerconfig.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
+	"github.com/openshift/hypershift/control-plane-operator/controllers/hostedcontrolplane/manifests"
 	"github.com/openshift/hypershift/support/azureutil"
 	"github.com/openshift/hypershift/support/config"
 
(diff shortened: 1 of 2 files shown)


Phaow:
@bryan-cox @gnufied The patch fixes the azure file csi driver could not provision volume succeed issue, for the disk csi issue should be an issue from driver itself,  I used upstream `azuredisk-csi:v1.31.1` and the patch verify disk csi driver works, so I suspect there's some known issue fixed between `v1.29.11` and `v1.31.1` caused `invalid memory address or nil pointer dereference` in disk csi provision. Could you help take a look when you get a chance? Thank you!

control-plane-operator/controllers/hostedcontrolplane/cloud/azure/providerconfig.go
```diff
@@ -43,8 +43,19 @@ func ReconcileCloudConfigWithCredentials(secret *corev1.Secret, hcp *hyperv1.Hos
 		return err
 	}
 
-	cfg.AADClientID = hcp.Spec.Platform.Azure.ManagedIdentities.ControlPlane.CloudProvider.ClientID
-	cfg.AADClientCertPath = config.ManagedAzureCertificatePath + hcp.Spec.Platform.Azure.ManagedIdentities.ControlPlane.CloudProvider.CertificateName
+	// Get the ClientID and CertificateName config for csi drivers in ARO
+	switch secret.Name {
+	case "azure-disk-csi-config":
```
enxebre, line 48: can we please get the name cases from the library funcs, e.g. manifests.AzureFileConfigWithCredentials("")?

control-plane-operator/controllers/hostedcontrolplane/cloud/azure/providerconfig.go
```diff
@@ -43,8 +43,19 @@ func ReconcileCloudConfigWithCredentials(secret *corev1.Secret, hcp *hyperv1.Hos
 		return err
 	}
 
-	cfg.AADClientID = hcp.Spec.Platform.Azure.ManagedIdentities.ControlPlane.CloudProvider.ClientID
-	cfg.AADClientCertPath = config.ManagedAzureCertificatePath + hcp.Spec.Platform.Azure.ManagedIdentities.ControlPlane.CloudProvider.CertificateName
+	// Get the ClientID and CertificateName config for csi drivers in ARO
```
enxebre, line 46: can we have a unit test for this func?

enxebre: Can we please make sure this change is also applied to cpo v2/ cc @muraee @bryan-cox

muraee: https://github.com/openshift/hypershift/blob/main/control-plane-operator/controllers/hostedcontrolplane/v2/cloud_controller_manager/azure/config.go#L60

control-plane-operator/controllers/hostedcontrolplane/cloud/azure/providerconfig.go
Phaow: fixed, thanks!

control-plane-operator/controllers/hostedcontrolplane/cloud/azure/providerconfig.go
Phaow: TBH, I could not understand the cpo v2 design how to handle the cloud secrets, I saw it introduces the `azure-credential-information` secret while seems no logic handle the csi drivers secrets. Not sure how to apply the fix also to cpo v2. @bryan-cox Could you please help give some info? Thank you!

muraee: true, this will be part for the storage v2 which is not yet refactored. You don't need to do anything extra for now in this PR.

control-plane-operator/controllers/hostedcontrolplane/cloud/azure/providerconfig.go
```diff
@@ -43,8 +44,19 @@ func ReconcileCloudConfigWithCredentials(secret *corev1.Secret, hcp *hyperv1.Hos
...
+	switch secret.Name {
+	case manifests.AzureDiskConfigWithCredentials(hcp.Namespace).Name:
```
bryan-cox, line 49: I don't think this is the right fix for this issue. We are already reconciling the right info for the secrets here- https://github.com/openshift/hypershift/blob/74287710a908357c17902185ca2cf5f3f58886c9/control-plane-operator/controllers/hostedcontrolplane/storage/azure.go#L28-L27

bryan-cox: Looks like we just need to add the certificate path to the above^

muraee: Those secrets are reconciled again and overriden here https://github.com/openshift/hypershift/blob/main/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go#L5094-L5106

control-plane-operator/controllers/hostedcontrolplane/cloud/azure/providerconfig.go
Phaow: @muraee Thanks for the checking and clarifying!

PR #5314, opened 2024-12-18
CNTRLPLANE-26: Disable building HO for s390x platform on PR

devguyio:
Due to capacity issues in s390x environment, we're instructed to stop building images on PR for s390x

**Checklist**

### DIFF
--- a/.tekton/hypershift-operator-main-pull-request.yaml
+++ b/.tekton/hypershift-operator-main-pull-request.yaml
@@ -206,34 +206,6 @@ spec:
             workspace: workspace-arm64
           - name: basic-auth
             workspace: git-auth
-      - name: clone-repository-s390x
-        params:
-          - name: url
-            value: $(params.git-url)
-          - name: revision
-            value: $(params.revision)
-        runAfter:
-          - init
-        taskRef:
-          kind: Task
-          params:
-            - name: name
-              value: git-clone
-            - name: bundle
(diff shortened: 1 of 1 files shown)


.tekton/hypershift-operator-main-pull-request.yaml
```diff
@@ -206,34 +206,6 @@ spec:
             workspace: workspace-arm64
           - name: basic-auth
             workspace: git-auth
-      - name: clone-repository-s390x
```
bryan-cox, line 209: Can we just comment everything out rather than removing it? That way we can just uncomment it back in once s390x is supported again.

bryan-cox (commented):
Would you please add the why/description to the commit message?

.tekton/hypershift-operator-main-pull-request.yaml
jparrill: +1

devguyio: Personally I'm not a big fan of commenting out code. Git history is always there and one can see whatever was removed. But sure, if that's what folks want.

devguyio:
/retitle "CNTRLPLANE-26: Disable building HO for s390x platform on PR"

.tekton/hypershift-operator-main-pull-request.yaml
devguyio: As agreed, a JIRA issue is now referenced

devguyio:
> Would you please add the why/description to the commit message?

Done

devguyio:
/retitle [CNTRLPLANE-26: Disable building HO for s390x platform on PR

devguyio:
/retitle CNTRLPLANE-26: Disable building HO for s390x platform on PR

PR #5321, opened 2024-12-19
OCPBUGS-38425: Return the right tagReference on Catalogs ImageStream

jparrill:
This is a manual cherry-pick of https://github.com/openshift/hypershift/pull/5187


jparrill:
Not having a way to validate this (because we need a 4.13 deployment to be upgraded to 4.14) I will close this PR with the Jira.

/close

PR #5341, opened 2024-12-30
OCPBUGS-47715: fix disconnected via CLI

LiangquanLi930:
In many disconnected environments, the execution environment may not have access to the payload image. Adjust the multi-arch check logic to skip the check when the payload is inaccessible.

Fixes [OCPBUGS-47715](https://issues.redhat.com/browse/OCPBUGS-47715)

**Checklist**

### DIFF
--- a/cmd/cluster/core/create.go
+++ b/cmd/cluster/core/create.go
@@ -649,7 +649,11 @@ func (opts *RawCreateOptions) Validate(ctx context.Context) (*ValidatedCreateOpt
 			return nil, fmt.Errorf("could not retrieve kube clientset: %w", err)
 		}
 		if err := validateMgmtClusterAndNodePoolCPUArchitectures(ctx, opts, kc, &hyperutil.RegistryClientImageMetadataProvider{}); err != nil {
-			return nil, err
+			if strings.Contains(err.Error(), "failed to retrieve manifest") {
+				opts.Log.Info("WARNING: Unable to access the payload, skipping the Architectures check.", "error", err.Error())
+			} else {
+				return nil, err
+			}
 		}
 	}
 


cmd/cluster/core/create.go
```diff
@@ -1008,43 +998,3 @@ func parseTolerationString(str string) (*corev1.Toleration, error) {
...
-
-	// Check if the release image is multi-arch
-	if len(opts.ReleaseImage) > 0 && len(opts.PullSecretFile) > 0 {
-		pullSecret, err := os.ReadFile(opts.PullSecretFile)
-		if err != nil {
-			return fmt.Errorf("failed to read pull secret file: %w", err)
-		}
-		validMultiArchImage, err = registryclient.IsMultiArchManifestList(ctx, opts.ReleaseImage, pullSecret, imageMetadataProvider)
-		if err != nil {
-			return err
```
bryan-cox, line 1026: Rather than outright removing the multi-arch validation function, we could just log an error message here.

cmd/cluster/core/create.go
```diff
@@ -1008,43 +998,3 @@ func parseTolerationString(str string) (*corev1.Toleration, error) {
...
-	// If not release image was provided, check if a release stream was provided instead and its multi-arch
-	if opts.ReleaseImage == "" && len(opts.ReleaseStream) > 0 && strings.Contains(opts.ReleaseStream, hyperv1.ArchitectureMulti) {
-		validMultiArchImage = true
-	}
-
-	// If a release image/stream is not multi-arch, check the mgmt & NodePool CPU architectures match
-	if !validMultiArchImage {
-		mgmtClusterCPUArch, err := hyperutil.GetMgmtClusterCPUArch(kc)
-		if err != nil {
-			return fmt.Errorf("failed to check mgmt cluster CPU arch: %v", err)
```
bryan-cox, line 1039: Rather than outright removing the multi-arch validation function, we could just log an error message here.

cmd/cluster/core/create.go
```diff
@@ -1008,43 +998,3 @@ func parseTolerationString(str string) (*corev1.Toleration, error) {
...
-		}
-
-		if !strings.EqualFold(mgmtClusterCPUArch, opts.Arch) {
-			return fmt.Errorf("multi-arch hosted cluster is not enabled and "+
```
bryan-cox, line 1043: Rather than outright removing the multi-arch validation function, we could just log an error message here.

bryan-cox (commented):
I think we should leave this function in place and just log an error message.

cmd/cluster/core/create.go
```diff
@@ -649,6 +649,9 @@ func (opts *RawCreateOptions) Validate(ctx context.Context) (*ValidatedCreateOpt
 			return nil, fmt.Errorf("could not retrieve kube clientset: %w", err)
 		}
 		if err := validateMgmtClusterAndNodePoolCPUArchitectures(ctx, opts, kc, &hyperutil.RegistryClientImageMetadataProvider{}); err != nil {
+			if strings.Contains(err.Error(), "failed to retrieve manifest") {
+				opts.Log.Info("WARNING: Unable to access the payload, skipping the Architectures check.", "error", err.Error())
```
sjenning, line 653: error says we are skipping, but we are still returning the error and `Validate()` fails right?

bryan-cox: Yeah, I must have misread something when I looked at this earlier. 

I think we need a `} else {` on L654

PR #5347, opened 2025-01-07
NO-JIRA:reconcile azure creds to a fixed secret in hcp namespace

Patryk-Stefanski:
This is blocking 4.17 aks e2e as it cannot create an HC because the reconcile of cloud credentials is blocking progression as it is unable to find the -cloud-credentials secret in HCP NS as made in the HC NS. In newer releases, we use a fixed secret in the HCP namespace instead so this is copying that behavior.

Fixes #

**Checklist**

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
@@ -2617,7 +2617,7 @@ func (r *HostedControlPlaneReconciler) reconcileCloudProviderConfig(ctx context.
 			return fmt.Errorf("failed to reconcile aws provider config: %w", err)
 		}
 	case hyperv1.AzurePlatform:
-		credentialsSecret := &corev1.Secret{ObjectMeta: metav1.ObjectMeta{Namespace: hcp.Namespace, Name: hcp.Spec.Platform.Azure.Credentials.Name}}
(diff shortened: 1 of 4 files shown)


jparrill:
Hey @Patryk-Stefanski It's not usual to put a PR against 4.17 directly, is this blocking you? It's there any way to use the typical backport flow to follow in this case?

Patryk-Stefanski:
> It's not usual to put a PR against 4.17 directly, is this blocking you? It's there any way to use the typical backport flow to follow in this case?

This is one of the things that is causing the 4.17 aks e2e to fail, there is a change in 4.18 that fixes this but it requires a manual backport of a few different PRs, I was hoping this would be a quick fix.

We might not need a 4.17 aks e2e, @bryan-cox any update on that ? We can close this PR if we don't need 4.17.

bryan-cox:
> > It's not usual to put a PR against 4.17 directly, is this blocking you? It's there any way to use the typical backport flow to follow in this case?
> 
> This is one of the things that is causing the 4.17 aks e2e to fail, there is a change in 4.18 that fixes this but it requires a manual backport of a few different PRs, I was hoping this would be a quick fix.
> 
> We might not need a 4.17 aks e2e, @bryan-cox any update on that ? We can close this PR if we don't need 4.17.

@Patryk-Stefanski no update right now

Patryk-Stefanski:
Closed as we currently wont be offering support for 4.17

PR #5353, opened 2025-01-08
OCPBUGS-48152: Fix mirror reference image name when ID is set

jparrill:
**What this PR does / why we need it**: When an imageName is set with the SHA256 and the ICSP/IDMS does not contain a Tag but an ID, the reference image name is using the original name instead of the mirror one. This PR fixes that situation.

Fixes #[OCPBUGS-48152](https://issues.redhat.com/browse/OCPBUGS-48152)


jparrill:
Looks like there is a Konflux outage we will need to try retest after that got fixed. More info here: https://redhat-internal.slack.com/archives/C04PZ7H0VA8/p1736438759559479

PR #5375, opened 2025-01-10
NO-JIRA: test/e2e: improve VAP test assertions

sjenning:
https://github.com/openshift/cluster-network-operator/pull/2612 broke our e2e because our assertions in `EnsureValidatingAdmissionPoliciesExists` make the assumption that the only VAPs in the guest cluster are ones that we create.

The PR refactors this check to:
* Not assume no other VAPs exist
* Not wait for VAPs to exist as this is called from `after()`, as in after the HostedCluster is fully deployed and the VAPs should already be created.  It is a bug if they are not created by the time the check runs.


sjenning:
self lgtm'ing and overriding to fix e2e
/override ci/prow/e2e-aws

PR #5403, opened 2025-01-16
OCPBUGS-48510: Allow ARM64 arch deployment on None platform type

eranco74:
This PR allows the NodePool API to set ARM64 arch for deployments with None platform type.

Fixes #[OCPBUGS-48510](https://issues.redhat.com/browse/OCPBUGS-48510)

### DIFF
--- a/api/hypershift/v1beta1/nodepool_types.go
+++ b/api/hypershift/v1beta1/nodepool_types.go
@@ -79,7 +79,7 @@ type NodePool struct {
 
 // NodePoolSpec is the desired behavior of a NodePool.
 // +kubebuilder:validation:XValidation:rule="!has(oldSelf.arch) || has(self.arch)", message="Arch is required once set"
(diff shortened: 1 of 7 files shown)


enxebre:
The issue description talks about agent but this is for None, what's the intersection?

eranco74:
> The issue description talks about agent but this is for None, what's the intersection?

The jira was cloned from similar issue about the agent platform,  I thought the description was generic (non aws/azure), the intersection is baremetal hosts.

bryan-cox:
Failures seem legitimate 
```
Error: CustomResourceDefinition.apiextensions.k8s.io "nodepools.hypershift.openshift.io" is invalid: spec.validation.openAPIV3Schema.properties[spec].x-kubernetes-validations[1].rule: Invalid value: apiextensions.ValidationRule{Rule:"self.arch != 'arm64' || has(self.platform.aws) || has(self.platform.azure) || has(self.platform.agent) || has(self.platform.type.none)", Message:"Setting Arch to arm64 is only supported for AWS, Azure, Agent and None", MessageExpression:"", Reason:(*apiextensions.FieldValueErrorReason)(nil), FieldPath:"", OptionalOldSelf:(*bool)(nil)}: compilation failed: ERROR: <input>:1:110: type 'string' does not support field selection
```

PR #5411, opened 2025-01-16
NO-JIRA: Fix dev docs for managed azure

gnufied:
Fix managed azure docs for development

### DIFF
--- a/contrib/managed-azure/setup_dev_environment.md
+++ b/contrib/managed-azure/setup_dev_environment.md
@@ -42,6 +42,12 @@ PARENT_DNS_RG="os4-common"
 HC_NAME=<HC_NAME>
 ```
 
+You can look up `USER_ACCOUNT_ID` by searching for your account in azure portal's search box and selecting Microsoft Entra ID.You can
+use value of `Object ID` as shown in azure portal.
+
+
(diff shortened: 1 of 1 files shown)


contrib/managed-azure/setup_dev_environment.md
```diff
@@ -42,6 +42,12 @@ PARENT_DNS_RG="os4-common"
 HC_NAME=<HC_NAME>
 ```
 
+You can look up `USER_ACCOUNT_ID` by searching for your account in azure portal's search box and selecting Microsoft Entra ID.You can
+use value of `Object ID` as shown in azure portal.
+
+
+Value of `$PARENT_DNS_ZONE` may be different for different teams. Check `os4-common` resourcegroup associated with your subscription-id.
```
Patryk-Stefanski, line 49:
```suggestion
Value of `$PARENT_DNS_ZONE` may be different for different teams. Check `os4-common` resourcegroup associated with your subscription-id for pre existing DNS zones.  If there are multiple DNS zones more than likely the parent DNS zone is the common suffix among them.
```

gnufied:
@Patryk-Stefanski ptal. thanks.

PR #5412, opened 2025-01-16
NO-JIRA: doc/openstack: improve index layout

EmilienM:

Double-level bullet points don't seem the work fine with the current
theme so we just adjust the layout a bit so it's prettier.

### DIFF
--- a/docs/content/how-to/openstack/index.md
+++ b/docs/content/how-to/openstack/index.md
@@ -21,8 +21,10 @@ Table of content:
 
 - [Prerequisites](prerequisites.md)
 - [Create a Hosted Cluster and a default Nodepool](hostedcluster.md)
-- Advanced scenarios: additional Nodepools can be created with specific configurations:
-  - [Nodepools distributed over Nova availability zones](az.md)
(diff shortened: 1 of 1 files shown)


docs/content/how-to/openstack/index.md
```diff
@@ -21,8 +21,10 @@ Table of content:
 
 - [Prerequisites](prerequisites.md)
 - [Create a Hosted Cluster and a default Nodepool](hostedcluster.md)
-- Advanced scenarios: additional Nodepools can be created with specific configurations:
-  - [Nodepools distributed over Nova availability zones](az.md)
-  - [Additional ports for Nodepools](additional-ports.md)
-  - [Node tuning, SR-IOV Network Operator](performance-tuning.md)
 - [Destroy a Hosted Cluster](destroy.md)
+
+Advanced scenarios: additional Nodepools can be created with specific configurations:
```
bryan-cox, line 26: Is the `-` missing here?

bryan-cox: I think it's fine. I wish I could see it deployed like we used to be able to from the PR.

I checked it out locally and it looks fine.

PR #5417, opened 2025-01-17
HOSTEDCP-2169: Add aro scheduler

Patryk-Stefanski:

The reason for implementing this scheduler is to help Azure Red Hat OpenShift (ARO) scale to 500 nodes. With this scheduler, ARO can dynamically control the CPU and memory requests for all pods and the Kube API Server (KAS) GoMem limits based on the number of nodes a `HostedCluster` has. This capability is essential for efficiently scaling a `HostedCluster` to 500 nodes.

Fixes # [HOSTEDCP-2169](https://issues.redhat.com/browse/HOSTEDCP-2169)

**Checklist**


hypershift-operator/controllers/scheduler/aro/aro_scheduler.go
```diff
@@ -0,0 +1,157 @@
...
+	if err := r.Get(ctx, types.NamespacedName{Name: "cluster"}, &config); err != nil {
+		return ctrl.Result{}, fmt.Errorf("could not get cluster sizing configuration: %w", err)
+	}
+
+	if condition := meta.FindStatusCondition(config.Status.Conditions, schedulingv1alpha1.ClusterSizingConfigurationValidType); condition == nil || condition.Status != metav1.ConditionTrue {
+		log.Info("Cluster sizing configuration is not valid, skipping for now")
+		return ctrl.Result{}, nil
+	}
+
+	// Find existing dedicated serving content Nodes for this HC.
```
csrwng, line 123: Please remove, this is not needed

hypershift-operator/controllers/scheduler/aro/aro_scheduler.go
```diff
@@ -0,0 +1,157 @@
+package aro
```
csrwng, line 1: nit: "aro" and "scheduler" are already part of the package name. I would name the file something like "controller.go" or "reconciler.go"

support/util/scheduler.go
```diff
@@ -0,0 +1,148 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+const (
+	GoMemLimitLabel = "hypershift.openshift.io/kas-go-mem-limit"
+	LBSubnetsLabel  = "hypershift.openshift.io/request-serving-subnets"
+)
+
+func UpdateHostedCluster(hc *hyperv1.HostedCluster, size string, config *schedulingv1alpha1.ClusterSizingConfiguration, nodes []corev1.Node) (*hyperv1.HostedCluster, error) {
```
csrwng, line 18: This could use unit tests

support/util/scheduler.go
```diff
@@ -0,0 +1,148 @@
...
+func configHasMHCTimeout(config *schedulingv1alpha1.ClusterSizingConfiguration) bool {
+	for _, size := range config.Spec.Sizes {
+		if size.Effects != nil && size.Effects.MachineHealthCheckTimeout != nil {
+			return true
+		}
+	}
+	return false
+}
+
+func ResourceRequestsToOverrideAnnotations(requests []schedulingv1alpha1.ResourceRequest) map[string]string {
```
csrwng, line 131: doesn't look like this needs to be public

hypershift-operator/controllers/scheduler/aro/aro_scheduler.go
bryan-cox: tagging on this... Alberto had me removing some of the ARO verbiage in favor of something like "managed Azure" in previous ARO HCP PRs. If you search around our codebase, you'll see where we already transitioned a lot to that type of language.

support/util/scheduler.go
Patryk-Stefanski: It needed here https://github.com/openshift/hypershift/pull/5417/files#diff-be4da6a931f52746ccd6902dca5fd9936ac0ffb1179417b9577d970aa3ce7bb4R756

hypershift-operator/controllers/scheduler/azure/controller.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	if err := r.updateAroHostedCluster(ctx, hc, desiredSize, &config); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	return ctrl.Result{}, nil
+}
+
+func (r *Scheduler) updateAroHostedCluster(ctx context.Context, hc *hyperv1.HostedCluster, size string, config *schedulingv1alpha1.ClusterSizingConfiguration) error {
```
enxebre, line 134: Please let's keep this project as product agnostic building blocks as much as possible while let it be platform aware. E.g. replace UpdateARO... with UpdateAzure?
What's the distinction between updateAROHostedCluster and updateROSAHostedCluster? Please name the func according to what the func actually does and let the diff be a param if needed

hypershift-operator/controllers/scheduler/azure/controller.go
```diff
@@ -0,0 +1,148 @@
...
+		return ctrl.Result{}, fmt.Errorf("failed to get cluster %q: %w", req.NamespacedName, err)
+	}
+	if !hc.DeletionTimestamp.IsZero() {
+		log.Info("hostedcluster is deleted, cleaning up")
+		if controllerutil.ContainsFinalizer(hc, schedulerFinalizer) {
+			if controllerutil.ContainsFinalizer(hc, hostedcluster.HostedClusterFinalizer) {
+				// Wait until the hosted cluster finalizer is removed
+				return ctrl.Result{}, nil
+			}
+			// Ensure that any placeholder deployment is deleted
```
enxebre, line 78: >// Ensure that any placeholder deployment is deleted

I don't see where we are honouring this. It doesn't seem this finalizer is actually doing anything.
Is this a TODO?

hypershift-operator/controllers/scheduler/azure/controller.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+	if !controllerutil.ContainsFinalizer(hc, schedulerFinalizer) {
+		controllerutil.AddFinalizer(hc, schedulerFinalizer)
+		if err := r.Update(ctx, hc); err != nil {
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{}, nil
+	}
+
+	isPaused, duration, err := util.ProcessPausedUntilField(hc.Spec.PausedUntil, time.Now())
```
enxebre, line 95: this would pause if there's bad input, is that what we want? Otherwise you can use IsReconciliationPaused

hypershift-operator/controllers/scheduler/azure/controller.go
enxebre: Also let's name the parent folders aws and azure consistently

hypershift-operator/controllers/scheduler/azure/controller.go
Patryk-Stefanski: This comment should be removed it was from previous code , we don't need placeholders for the aro scheduler

hypershift-operator/controllers/scheduler/azure/controller.go
bryan-cox: I had been using managed Azure previously - https://github.com/openshift/hypershift/pull/5417#discussion_r1920637513

hypershift-operator/controllers/scheduler/azure/controller.go
Patryk-Stefanski: The only time the pausedUntil field has a bad input is when a user intends to pause the reconciliation, so I think its better to pause it even if the input isn't valid.

support/util/scheduler.go
```diff
@@ -0,0 +1,174 @@
+package util
```
muraee, line 1: I think its better to put this file inside controller/scheduler/util package instead of the global util, since its only used there.

Patryk-Stefanski: done

hypershift-operator/controllers/scheduler/azure/controller.go
enxebre: sgtm, fwiw any bad input should be also prevented by CEL already

hypershift-operator/controllers/scheduler/util/scheduler.go
```diff
@@ -0,0 +1,174 @@
...
+		}
+
+		hc.Annotations[hyperv1.RequestServingNodeAdditionalSelectorAnnotation] = fmt.Sprintf("%s=%s", hyperv1.NodeSizeLabel, size)
+
+	}
+
+	return hc, nil
+}
+
+func UpdateHostedCluster(ctx context.Context, c client.Client, hc *hyperv1.HostedCluster, size string, config *schedulingv1alpha1.ClusterSizingConfiguration, nodes []corev1.Node) error {
```
enxebre, line 98: please document this public func // UpdateHostedCluster sets...

hypershift-operator/controllers/scheduler/util/scheduler.go
```diff
@@ -0,0 +1,174 @@
...
+func getGoMemLimitLabelFromNodes(nodes []corev1.Node) string {
+	for _, node := range nodes {
+		if node.Labels[GoMemLimitLabel] != "" {
+			return node.Labels[GoMemLimitLabel]
+		}
+	}
+	return ""
+}
+
+func SizeConfiguration(config *schedulingv1alpha1.ClusterSizingConfiguration, size string) *schedulingv1alpha1.SizeConfiguration {
```
enxebre, line 139: please document this public func

enxebre:
can you please include in the PR desc or godoc the purpose and outcome of this new controller? e.g.
"The controller/scheduler reconciles HostedClusters applyng the corresponding annotations based on the size label assigned to the given HostedCluster. Based on those annotations the hostedCluster controller adjust compute consumption choices." (if that's accurate indeed)

hypershift-operator/controllers/scheduler/aro/aro_scheduler.go
enxebre: This Repo allocates a project which supports multiple products. There's anything about this code that makes it aro specific. I see it as an agnostic building block that we might choose to leverage in e.g. self-hosted azure. It's a runtime choice to tie it with any particular product. My suggestion is we capture this as semantics when naming things.
With that in mind It would makes sense to me that we name everything here after "Azure". Then, at runtime, we choose to run this controller if IsAroHCP

hypershift-operator/controllers/scheduler/azure/controller.go
```diff
@@ -0,0 +1,107 @@
...
+			for _, hc := range hostedClusters.Items {
+				out = append(out, reconcile.Request{NamespacedName: types.NamespacedName{Namespace: hc.Namespace, Name: hc.Name}})
+			}
+			return out
+		})).
+		Named(aroScheduler)
+	return builder.Complete(r)
+}
+
+func (r *Scheduler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
enxebre, line 58: can we please have a unit for TestReconcile()

hypershift-operator/controllers/scheduler/azure/controller.go
```diff
@@ -0,0 +1,106 @@
...
+	log := ctrl.LoggerFrom(ctx)
+	err := r.Get(ctx, req.NamespacedName, hc)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			log.Info("hostedcluster not found, aborting reconcile")
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, fmt.Errorf("failed to get cluster %q: %w", req.NamespacedName, err)
+	}
+
```
enxebre, line 68: even though we need no finalizers, we should return early here if the hc has a deletion timestamp

hypershift-operator/controllers/scheduler/azure/controllers_test.go
```diff
@@ -0,0 +1,212 @@
...
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+func TestAzureSchedulerReconcile(t *testing.T) {
```
enxebre, line 23: nit: this is within the scheduler/azure, so TestReconcile seems should be sufficient

hypershift-operator/controllers/scheduler/azure/controllers_test.go
```diff
@@ -0,0 +1,212 @@
...
+			expectErrText: "could not find size configuration for size",
+			expectRequeue: false,
+		},
+		{
+			name:          "valid hosted cluster",
+			hc:            hostedcluster(),
+			sizingConfig:  sizingConfig,
+			expectError:   false,
+			expectRequeue: false,
+			expectAnnotations: map[string]string{
```
enxebre, line 165: any reason to not validate the rest of the [effects](https://github.com/openshift/hypershift/pull/5417/files#diff-7cae82863c80900f5456f966ca5bdfc20378a56d997429c284fcf10f82d8425cR9)?

enxebre:
can we can an e2e similar to TestCreateClusterRequestServingIsolation?

enxebre:
lgtm other than https://github.com/openshift/hypershift/pull/5417#discussion_r1930171344
I'd defer to @csrwng for his review and tag

hypershift-operator/controllers/scheduler/azure/controllers_test.go
Patryk-Stefanski: ill add test cases for the rest of the effects to TestSetHostedClusterSchedulingAnnotations

Patryk-Stefanski:
> can we can an e2e similar to TestCreateClusterRequestServingIsolation?

Yep I was planning on creating a follow up task to e2e this.

docs/content/how-to/azure/scheduler.md
```diff
@@ -0,0 +1,169 @@
...
+
+The Azure scheduler controller is straightforward. It checks the label set by the `HostedClusterSizing` controller and retrieves the cluster sizing configuration associated with the tshirt size. Based on the configuration, it can modify the `HostedCluster` with annotations for the specified fields. These annotations are then used by different controllers to propagate the required changes to the appropriate pods and containers.
+
+## How to Use
+
+### Prerequisites
+
+- AKS cluster with cluster-autoscaler enabled and using Standard\_D4s\_v4 VMs for this example. (--enable-cluster-autoscaler flag when installing AKS cluster, with --min-count 2 --max-count 6)
+- Hypershift operator with size tagging enabled. (--enable-size-tagging flag when installing hypershift operator)
+- HostedCluster with the `ClusterSizingConfiguration` resource created. (Default resource is created by the hypershift operator)
```
csrwng, line 50: This is a little confusing. What is a HostedCluster with ClusterSizingConfiguration? 
I would call them out as separate prereqs.
It would also be good to specify the name of the HostedCluster and NodePool that you're using throughout your example.

docs/content/how-to/azure/scheduler.md
```diff
@@ -0,0 +1,169 @@
...
+        name: medium
+      - criteria:
+          from: 5
+        name: large
+      transitionDelay:
+        decrease: 0s
+        increase: 0s
+    ```
+
+3. Scale nodepool up to 3 nodes:
```
csrwng, line 99: Before this there is no mention of nodepool pstefans-3

csrwng (commented):
Just a nit on the doc, otherwise lgtm

PR #5420, opened 2025-01-17
OCPBUGS-45511: Compare contents of user-ca-bundle-config.yaml to the correct configmap

csrwng:
PR #4331 added a check in the local ignition provider to validate that the bundle in the machine-config-server configmap under user-ca-bundle-config.yaml matched the bundle in the control plane namespace. However, the comparison was using the wrong configmap. The source of user-ca-bundle-config.yaml is trusted-ca-bundle-managed and not user-ca-bundle. This fix keeps the verification of the hash against user-ca-bundle since that hash is generated by the nodepool controller. However, when comparing the contents of the mcs configmap, it switches to the trusted-ca-bundle-managed configmap, since both should contain the aggregate of .spec.additionalTrustBundle and

Fixes #[OCPBUGS-45511](https://issues.redhat.com/browse/OCPBUGS-45511)

**Checklist**

### DIFF
--- a/ignition-server/controllers/local_ignitionprovider.go
+++ b/ignition-server/controllers/local_ignitionprovider.go
@@ -85,8 +85,11 @@ type LocalIgnitionProvider struct {
 
 var _ IgnitionProvider = (*LocalIgnitionProvider)(nil)
 
-const pullSecretName = "pull-secret"
-const additionalTrustBundleName = "user-ca-bundle"
+const (
+	pullSecretName            = "pull-secret"
+	additionalTrustBundleName = "user-ca-bundle"
+	managedTrustBundleName    = "trusted-ca-bundle-managed"
+)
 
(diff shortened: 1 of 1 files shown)


ignition-server/controllers/local_ignitionprovider.go
```diff
@@ -87,6 +87,7 @@ var _ IgnitionProvider = (*LocalIgnitionProvider)(nil)
 
 const pullSecretName = "pull-secret"
 const additionalTrustBundleName = "user-ca-bundle"
+const managedTrustBundleName = "trusted-ca-bundle-managed"
```
bryan-cox, line 90: nit: probably should have a const () block now

csrwng: fixed

bryan-cox:

PR #5404, opened 2025-01-16
PODAUTO-303: Add the most basic e2e for autoNode via Karpenter

enxebre:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, use `fixes #<issue_number>(, fixes #<issue_number>, ...)` format, where issue_number might be a GitHub issue, or a Jira story*:
Fixes #

/hold 
needs https://github.com/openshift/release/pull/60715

**Checklist**
- [ ] Subject and description added to both, commit and PR.
- [ ] Relevant issues have been referenced.
- [ ] This change includes docs. 
- [ ] This change includes unit tests.

PR #5439, opened 2025-01-20
HOSTEDCP-2249: Reconcile karpenter user-data secret programmatically instead of creating a mock nodePool

muraee:

Fixes #

**Checklist**


enxebre:
/hold
on https://github.com/openshift/hypershift/pull/5404

hypershift-operator/controllers/hostedcluster/karpenter.go
```diff
@@ -92,21 +90,52 @@ spec:
...
+	if err != nil {
+		return err
+	}
+
+	releaseImage, err := releaseProvider.Lookup(ctx, nodePool.Spec.Release.Image, pullSecretBytes)
+	if err != nil {
+		return err
+	}
+
+	haproxy := haproxy.HAProxy{
```
enxebre, line 112: Not for this PR but we need to follow up to hide the haproxy generation behind the Config logic.

Then the rest of this logic could be encapsulated within a func token() as in the nodepool_controller if we have a common reconciler interface which implements the methods to get the image

enxebre: can we have this logic to be a unit tested function which validates the expected userData is generated with their current contractual expectations, i.e. hyperv1.NodePoolLabel and hypershift.openshift.io/ami?

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -1832,6 +1826,10 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
 		}
 	}
 
+	if err := r.reconcileKarpenterOperator(ctx, createOrUpdate, hcluster, hcp, r.HypershiftOperatorImage, controlPlaneOperatorImage); err != nil {
```
enxebre, line 1829: why are we moving this?

muraee: I just moved it to the end, just to make sure an error here doesn't block the HC from progressing.

hypershift-operator/controllers/hostedcluster/karpenter.go
muraee: added a unit test

enxebre:
unit is failing, lgtm otherwise

hypershift-operator/controllers/nodepool/apiserver-haproxy/haproxy.go
```diff
@@ -71,11 +73,11 @@ func (r *HAProxy) isHAProxyIgnitionConfigManaged(ctx context.Context, hcluster *
 		return false, "", fmt.Errorf("failed to look up image metadata for %s: %w", controlPlaneOperatorImage, err)
 	}
 
-	_, cpoSkips := util.ImageLabels(controlPlaneOperatorImageMetadata)[controlPlaneOperatorSkipsHAProxyConfigGenerationLabel]
+	_, cpoSkips := util.ImageLabels(controlPlaneOperatorImageMetadata)[ControlPlaneOperatorSkipsHAProxyConfigGenerationLabel]
 	return cpoSkips, controlPlaneOperatorImage, nil
 }
 
-func (r *HAProxy) reconcileHAProxyIgnitionConfig(ctx context.Context, componentImages map[string]string, hcluster *hyperv1.HostedCluster, controlPlaneOperatorImage string) (cfg string, missing bool, err error) {
+func (r *HAProxy) reconcileHAProxyIgnitionConfig(ctx context.Context, hcluster *hyperv1.HostedCluster, controlPlaneOperatorImage string) (cfg string, err error) {
```
enxebre, line 80: why are we changing this returning signature?

hypershift-operator/controllers/hostedcluster/karpenter.go
```diff
@@ -65,71 +66,54 @@ spec:
...
-				Subnet: hyperv1.AWSResourceReference{
-					// TODO(alberto): this is just to pass cel.
-					// Setting an ID instead of filter would break publicAndPrivate topology because the AWSEndpointService won't find the subnet.
-					// We'll move to generate the userdata for karpenter programatically.
-					Filters: []hyperv1.Filter{
-						{
-							Name:   "subnet-none",
-							Values: []string{"none"},
-						},
+			Management: hyperv1.NodePoolManagement{
```
enxebre, line 83: we don't need to define .management, do we?

hypershift-operator/controllers/nodepool/apiserver-haproxy/haproxy.go
muraee: the returned bool serves no purpose. returning true will only results in the calling function returning an ambiguous error message "waiting for missing components" instead of an actual error. 
I changed it so we always return a proper error message.

hypershift-operator/controllers/hostedcluster/karpenter.go
muraee: removed

PR #5446, opened 2025-01-20
OCPBUGS-48626: openstack: add DNS option to CLI

EmilienM:

Allow the users to provide the DNS nameservers that will be used when
creating the OpenStack subnets.
In some environments where DNS forwarding is not enabled, the subnet
needs to have DNS nameservers for external resolution, otherwise the
Nodepools won't be able to e.g. reach the ignition endpoint.

This was already possible before by rendering the HostedCluster spec
first but now it's possible to do it via the CLI for a better UX, since
this option might be used by a majority of our customers.


bshephar (approved):
Nice UX improvement.

test/e2e/e2e_test.go
```diff
@@ -407,3 +408,831 @@ func dumpTestMetrics(log logr.Logger, artifactDir string) {
 	}
 	log.Info("Successfully wrote metrics to artifacts")
 }
+<<<<<<< HEAD
+||||||| parent of 8c5d99bc2 (OCPBUGS-48626: openstack: add DNS option to CLI)
```
MaysaMacedo, line 412: leftover from merge conflict

cmd/cluster/openstack/create.go
```diff
@@ -170,6 +172,19 @@ func (o *RawCreateOptions) ApplyPlatformSpecifics(cluster *hyperv1.HostedCluster
 		cluster.Spec.Platform.OpenStack.IngressFloatingIP = o.OpenStackIngressFloatingIP
 	}
 
+	// If the user has specified DNS nameservers, it'll be used when creating the managed subnet(s).
+	// In the case where the user wants to override the other fields of the managed subnets, they can
+	// first render the cluster spec and then provide the needed configuration as API allows (for
+	// example the allocation pools and DNS nameservers).
```
MaysaMacedo, line 178: no need to mention DNS nameservers again as it will be possible to override with the CLI. No no need to render the manifest.

PR #5453, opened 2025-01-22
OCPBUGS-48708: correct aro azureFile config and permissions

Phaow:
- Correct ARO file csi driver cloud config and permissions.

Fixes # [OCPBUGS-48708](https://issues.redhat.com/browse/OCPBUGS-48708)

**Test record(Manually build the control plane operator image and replcaed in a live cluster verify it works)** 
https://jenkins-csb-openshift-qe-mastern.dno.corp.redhat.com/job/ocp-common/job/ginkgo-test/285274/console
```console
79 pass, 13 skip (46m1s)
```

### DIFF
--- a/cmd/infra/azure/create.go
+++ b/cmd/infra/azure/create.go
@@ -806,6 +806,8 @@ func assignServicePrincipalRoles(subscriptionID, managedResourceGroupName, nsgRe
 		scopes = append(scopes, vnetRG, dnsZoneRG)
 	case cpo:
 		scopes = append(scopes, nsgRG, vnetRG)
+	case azureFile:
+		scopes = append(scopes, nsgRG, vnetRG)
 	case nodePoolMgmt:
 		scopes = append(scopes, vnetRG)
 	}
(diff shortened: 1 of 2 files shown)


cmd/infra/azure/create.go
```diff
@@ -806,6 +806,8 @@ func assignServicePrincipalRoles(subscriptionID, managedResourceGroupName, nsgRe
 		scopes = append(scopes, vnetRG, dnsZoneRG)
 	case cpo:
 		scopes = append(scopes, nsgRG, vnetRG)
+	case azureFile:
```
bryan-cox, line 809: I don't believe this is needed. Could we remove this and retest? We are going to be stripping down these permissions soon.

Phaow: I tried with remove the permission previously, the file nfs volume provision got failed of permission issue.

bryan-cox: I'm mistaken. The operator on the HCP side will need permissions.

PR #5315, opened 2024-12-18
NO-JIRA: disable image registry for testing

tjungblu:
testing to see the behavior on CI

PR #5456, opened 2025-01-22
ARO-14354,ARO-14356,ARO-14355: introduce image registry capability behind feature gate

flavianmissi:
We're introducing the concept of capabilities to Hypershift on a limited basis, as described in https://github.com/openshift/enhancements/pull/1729.

This PR implements the changes necessary to disable the image registry capability at hosted cluster install time.

** What changed **

The changes listed below are what to expect when creating a HostedCluster with the Image Registry capability disabled (TODO: explain how to do this).

* `ClusterVersion` capabilities are set to the current set of capabilities (taken from openshift/api)
* The image-registry operator and its associated resources are not created or reconciled
  * Resources include but are not limited to: operator pods, certificates, cloud credentials, pvc, `configs.imageregistry.operator.openshift.io`, image registry pods, relevant service accounts, cluster roles and bindings, etc.
* openshift-controller-manager's pull-secrets controller is disabled
* WIP

Fixes #[ARO-14354](https://issues.redhat.com//browse/ARO-14354)

**Checklist**


flavianmissi:
/retitle ARO-14442,ARO-14354,ARO-14356,ARO-14355: introduce image registry capability behind feature gate

api/hypershift/v1beta1/hosted_controlplane.go
```diff
@@ -200,6 +200,18 @@ type HostedControlPlaneSpec struct {
...
+	// DisabledCapabilities when specified, sets the cluster version BaselineCapabilitySet to None
+	// and sets all AdditionalEnabledCapabilities BUT the ones supplied in DisabledCapabilities.
+	// This effectively disables that capability on the hosted cluster.
+	//
+	// When this is not supplied, the cluster will use the DefaultCapabilitySet defined for the respective
+	// OpenShift version.
+	//
+	// +listType=atomic
+	// +optional
+	// +openshift:enable:FeatureGate=DisableClusterCapabilities
```
enxebre, line 213: please use lower case for godocs so it generates the serialized version of the field name.
Please document what happens if I change the field day 2.
What's the rationale to deviate from https://github.com/openshift/installer/blob/main/pkg/types/installconfig.go#L537-L548

do we foresee this semantic growing so that it deserves its own struct?

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -1318,9 +1316,11 @@ func (r *HostedControlPlaneReconciler) reconcile(ctx context.Context, hostedCont
 	}
 
 	// Reconcile image registry operator
-	r.Log.Info("Reconciling Image Registry Operator")
-	if err := r.reconcileImageRegistryOperator(ctx, hostedControlPlane, releaseImageProvider, userReleaseImageProvider, createOrUpdate); err != nil {
-		return fmt.Errorf("failed to reconcile image registry operator: %w", err)
+	if capabilities.IsImageRegistryCapabilityEnabled(hostedControlPlane.Spec.DisabledCapabilities) {
```
enxebre, line 1319: We are not plumbing the featuregate flags into the cpo/hcco binaries. You'd need to do it similar to what we do for the HO
https://github.com/openshift/hypershift/blob/main/cmd/install/assets/hypershift_operator.go#L428-L430
https://github.com/openshift/hypershift/blob/main/hypershift-operator/main.go#L156

Or alternative you can just gate on the field being set since it will only exist in the CRD if the HO is installed with --tech-preview-no-upgrade. See autoNode for an example of this.

api/hypershift/v1beta1/hosted_controlplane.go
```diff
@@ -200,6 +200,18 @@ type HostedControlPlaneSpec struct {
...
+	DisabledCapabilities []OptionalCapability `json:"disabledCapabilities,omitempty"`
```
enxebre, line 214: you'll need to add the featuregate to the files within v1beta1/featuregates/* then run make update to generate the CRDs

api/hypershift/v1beta1/hosted_controlplane.go
flavianmissi: > What's the rationale to deviate from https://github.com/openshift/installer/blob/main/pkg/types/installconfig.go#L537-L548

In Hypershift's case, we'll start off by only supporting the disabling of the Image Registry. Because of this, we think it'd be easier for users to simply specify what they want to disable. If we were to follow the OCP capabilities API then users would have to list all capabilities minus the ImageRegistry, and we don't even support the rest of the capabilities available in OCP. 

cc @sjenning 

> do we foresee this semantic growing so that it deserves its own struct?

Not sure I fully understand the question, but I'll try to answer anyway. 
What I expect to happen is that it will be possible in the future to specify `EnabledCapabilities`. It will likely be used to enable capabilities which aren't enabled by default (although we have no examples of this at the moment of writing). It might be used in conjunction with `DisabledCapabilities` or on its own.

> Please document what happens if I change the field day 2.
I'll work on this.

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
flavianmissi: If going on the route of adding a `--feature-gates` flag to cpo and hcco binaries, where/what would set this flag?

While the autoNode approach seems easier, I personally prefer the more explicit one with the feature gate check. Do you have a preference?

api/hypershift/v1beta1/hosted_controlplane.go
flavianmissi: done.

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
enxebre: You'd need to set them similarly to the links above but where we run those binaries, see links below.
https://github.com/openshift/hypershift/blob/main/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go#L2532
https://github.com/openshift/hypershift/blob/main/control-plane-operator/controllers/hostedcontrolplane/configoperator/reconcile.go#L391

My preference would be to use the api field which is the authoritative source. That way, once you drop the feature gate, the field will be part of the CRD and the business logic will just work with no more changes required. I see value for the flags only for features that are not api driven.

api/hypershift/v1beta1/hosted_controlplane.go
enxebre: let's also include a min/max validation for the slice

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -1235,11 +1240,37 @@ func (r *reconciler) reconcileKonnectivityAgent(ctx context.Context, hcp *hyperv
 	return errors.NewAggregate(errs)
 }
 
+// calculateEnabledCapabilities returns the difference between the default set
+// of enabled capabilities (vCurrent) and the given set of capabilities to
+// disable, alphabetically sorted.
```
enxebre, line 1245: please unit test this

enxebre:
Let's include unit tests for all the reconciliations that are behind the new check. If code coupling makes that impractical let's validate the same within an e2e.
Also this change needs to be incorporated within the cpov2 as well cc @muraee

muraee:
thanks @enxebre, seems this change is only affecting image-registry operator, which is not refactored yet, so no further changes in cpov2 is required right now.

sjenning:
FYI need for rebase caused by https://github.com/openshift/hypershift/pull/5549 removing the `HCPPodsLabels` feature gate

flavianmissi:
Thanks, Seth - I've removed the api and feature gate related changes for now, those are in a separate PR now: https://github.com/openshift/hypershift/pull/5554

This will wreak havoc in this PRs tests, as soon as the other PR merges I'll rebase this one and things should stabilize.

api/hypershift/v1beta1/hosted_controlplane.go
```diff
@@ -34,6 +34,7 @@ type HostedControlPlane struct {
 }
 
 // HostedControlPlaneSpec defines the desired state of HostedControlPlane
+// +kubebuilder:validation:XValidation:rule=`self.capabilities == oldSelf.capabilities`, message="Capabilities is immutable. Changes might result in unpredictable and disruptive behavior."
```
enxebre, line 37: can we instead have this validation at the field level with self == oldself
e.g. like with IssuerURL

support/capabilities/hosted_control_plane_capabilities_test.go
```diff
@@ -0,0 +1,118 @@
...
+	}{
+		{
+			name:                       "returns false when image registry capability is disabled",
+			disabledCapabilities:       []hyperv1.OptionalCapability{hyperv1.ImageRegistryCapability},
+			expectImageRegistryEnabled: false,
+		},
+		{
+			name:                       "returns true when image registry capability is enabled",
+			disabledCapabilities:       nil,
+			expectImageRegistryEnabled: true,
```
enxebre, line 26: are these two cases identical?

api/hypershift/v1beta1/hosted_controlplane.go
flavianmissi: I'll try that, thanks!

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources_test.go
```diff
@@ -955,6 +955,84 @@ func TestReconcileClusterVersion(t *testing.T) {
...
+	}
+	err := r.reconcileClusterVersion(context.Background(), hcp)
+	g.Expect(err).ToNot(HaveOccurred())
+	err = fakeClient.Get(context.Background(), client.ObjectKeyFromObject(clusterVersion), clusterVersion)
+	g.Expect(err).ToNot(HaveOccurred())
+
+	expectedCapabilities := &configv1.ClusterVersionCapabilitiesSpec{
+		BaselineCapabilitySet: configv1.ClusterVersionCapabilitySetNone,
+		AdditionalEnabledCapabilities: []configv1.ClusterVersionCapability{
+			configv1.ClusterVersionCapabilityBuild,
```
enxebre, line 1014: given we use this in many tests, we might want to have a common variable with the full list, then for comparison we subtract ClusterVersionCapabilityImageRegistry from that common source?

support/capabilities/hosted_control_plane_capabilities_test.go
flavianmissi: Ops! Nice catch.

control-plane-operator/controllers/hostedcontrolplane/cvo/reconcile.go
```diff
@@ -377,22 +388,26 @@ func cvoBootstrapScript(clusterVersionJSON []byte) string {
...
 	scriptTemplate := `#!/bin/bash
 set -euo pipefail
+MANIFEST_DIR=%s/manifests
+ls -la ${MANIFEST_DIR}
 cat > /tmp/clusterversion.json <<EOF
 %s
 EOF
 oc get ns openshift-config &> /dev/null || oc create ns openshift-config
 oc get ns openshift-config-managed &> /dev/null || oc create ns openshift-config-managed
+oc apply -f ${MANIFEST_DIR}/0000_00_cluster-version-operator_01_clusterversions*
```
enxebre, line 398: why do we need this additional commands now?

flavianmissi: Some of it was added to help troubleshooting - I'll remove those lines.

This line (398) specifically was added here because we needed to create the ClusterVersion resource earlier than the rest of the stuff in the manifests dir, so that the ClusterVersion was created before the CVO itself, since it's not possible to edit the field after the CVO is running (i.e on a running cluster).
It was @tjungblu who unearthed this requirement, but if I remember correctly the reasoning was that the [resource definition itself did not yet exist](https://github.com/openshift/hypershift/pull/5315#discussion_r1895786134) at this point, and since we wanted to create the ClusterVersion before the CVO, we had to move the CRD creation up here.

Maybe I should add a code comment here to clarify this?

test/e2e/create_cluster_test.go
```diff
@@ -1396,6 +1421,98 @@ func TestCreateClusterPrivateWithRouteKAS(t *testing.T) {
 	testCreateClusterPrivate(t, true)
 }
 
+// TestCreateClusterWithDisabledCapabilities implements a test that creates a cluster
+// with the ImageRegistry capability disabled, then attempts to enable it.
+func TestCreateClusterWithDisabledCapabilities(t *testing.T) {
```
enxebre, line 1426: we'll want to avoid increasing the cloud load by creating yet another hc for this. Instead we can reuse the existing TestCreateClusterCustomConfig to include these validations

flavianmissi: We talked about this on Slack: I'll drop the e2e from this PR and implement this comment on a separate PR, where we'll also drop the tech preview feature gate.

control-plane-operator/controllers/hostedcontrolplane/cvo/reconcile.go
tjungblu: feel free to revert them, but it took me three weeks to debug a yaml identation issue with this silly EOF bash imputation. And then it took me another two weeks to figure out why the apiserver was rejecting the apply and didn't complain about it. 

So I think it's useful, but I don't know how often you guys want to touch that code anyway.

flavianmissi: Thanks for all that work, Thomas, I'm planning on leaving the parsing separated from the apply, just like you ended up doing. The troubleshooting I was referring to I will remove is the `ls` and `oc get ... -oyaml` lines

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources_test.go
flavianmissi: Now that I removed the e2e tests from this PR, this is the only place that builds this list, so I'll defer this to the follow-up PR.

control-plane-operator/controllers/hostedcontrolplane/v2/ocm/config.go
```diff
@@ -38,7 +39,7 @@ func adaptConfigMap(cpContext component.WorkloadContext, cm *corev1.ConfigMap) e
 		return fmt.Errorf("failed to read observed global config: %w", err)
 	}
 
-	adaptConfig(config, cpContext.HCP.Spec.Configuration, cpContext.ReleaseImageProvider, observedConfig.Build)
+	adaptConfig(config, cpContext.HCP.Spec.Configuration, cpContext.ReleaseImageProvider, observedConfig.Build, cpContext.HCP.Spec.Capabilities)
```
muraee, line 42: thanks! I see you also change other components like oapi and cvo, can you please also make the change for their v2 counter parts as you did here.

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -1339,11 +1335,12 @@ func (r *HostedControlPlaneReconciler) reconcile(ctx context.Context, hostedCont
 
 	if !r.IsCPOV2 {
 		// Reconcile image registry operator
-		r.Log.Info("Reconciling Image Registry Operator")
-		if err := r.reconcileImageRegistryOperator(ctx, hostedControlPlane, releaseImageProvider, userReleaseImageProvider, createOrUpdate); err != nil {
-			return fmt.Errorf("failed to reconcile image registry operator: %w", err)
+		if capabilities.IsImageRegistryCapabilityEnabled(hostedControlPlane.Spec.Capabilities) {
```
muraee, line 1338: you would need to create a predicate to stop reconciliation in v2 here: https://github.com/openshift/hypershift/blob/main/control-plane-operator/controllers/hostedcontrolplane/v2/registryoperator/component.go#L34

see https://github.com/openshift/hypershift/blob/main/control-plane-operator/controllers/hostedcontrolplane/v2/csi/kubevirt/component.go#L48 as example.

flavianmissi:
Thanks for the comments @muraee. Do you think I can address them on a follow-up PR or is it a deal breaker to merge this one? I'll follow your lead, so just let me know.

muraee:
@flavianmissi I prefer to have the changes in the same PR, so as a reviewer I can compare the changes and make sure all of them are done in v2 as well. If you think this is not feasible, I am fine with a follow-up PR.

flavianmissi:
Makes sense @muraee , I'll put it on hold until I can get to that, which will probably happen sometime tomorrow.

/hold

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
flavianmissi: Done.

control-plane-operator/controllers/hostedcontrolplane/v2/ocm/config.go
flavianmissi: Done.

flavianmissi:
I believe I addressed all the V2 changes @muraee, could you PTAL? I have grouped the v2 changes in the same commit as the v1 changes, so you should be able to see equivalent v2 changes side by side with v1 for each commit.

control-plane-operator/controllers/hostedcontrolplane/v2/kas/params.go
```diff
@@ -118,6 +118,10 @@ func NewConfigParams(hcp *hyperv1.HostedControlPlane) KubeAPIServerConfigParams
 		kasConfig.MaxMutatingRequestsInflight = mutatingReqInflight
 	}
 
+	if capabilities.IsImageRegistryCapabilityEnabled(hcp.Spec.Capabilities) {
```
muraee, line 121: v1 kas doesn't seem to have this change. Also you made a change to v1 oapi but not v2. I assume you meant to do that here instead of the kas change?

muraee: for oapi v2, this is defined in the config manifest directly https://github.com/openshift/hypershift/blob/main/control-plane-operator/controllers/hostedcontrolplane/v2/assets/openshift-apiserver/config.yaml#L44

so in the code you need to set this to empty string when the capability is disabled.

flavianmissi: Thanks for the pointer, hopefully I covered everything this time around, PTAL

muraee: thanks! looks good

flavianmissi:
all review comments have been addressed, removing the hold.
/hold cancel

sjenning:
Tested this manually and everything looks good :+1: 

* No `openshift-image-registry` namespace in hosted cluster
* No `image-registry` ClusterOperator in hosted cluster
* `configs.imageregistry.operator.openshift.io` CRD exist but `cluster` config does not
* No guest cluster SAs have `imagePullSecret` set
* Guest cluster fully rolls out at the CVO and HC levels
* `image-registry-operator` is not running in the HCP
* OCM config in the HCP has `serviceaccount-pull-secrets` controller disabled
```
    controllers:
    - '*'
    - -openshift.io/serviceaccount-pull-secrets
...
    dockerPullSecret:
      internalRegistryHostname: ""
```
* API validation/CEL works as expected too.  I'm not able to change the capabilities after initial create.

flavianmissi:
Konflux checks seem to be holding off the merge...

/retest

PR #5458, opened 2025-01-22
CNTRLPLANE-216: Add KubeAPIExteralName api

jparrill:
This new API autogenerate a new kubeconfig based on a desired url and managed by HCP.

The current implementation of HCP to define the KAS address is using the ServicePublishingType, where you can define a  hostname and that, will customize the API and API-INT interfaces where all the kubeconfigs (excepts the ones pointing to local) are pointing to. This makes the customization rigid and customers are asking for more flexibility.

This PR adds a new API which creates a new Kubeconfig (certificate based) pointing to the desired address. With this new API you can:

- Modify the address day-1 and day-2 without affectation of the DataPlane (No rollout expected)
- Generation of a new Kubeconfig based on certificates (managed by HCP)
- The HC and HCP are in sync about this new Kubeconfig and reported in the status as `customKubeconfig`
- The target of this new API are the scenarios of Public and PublicAndPrivate

> ⚠️ **IMPORTANT!** The external DNSs URL where you plan to point to and you will specify it in this new API, should be properly configured in order to makes this work.

- Fixes #[HOSTEDCP-1960](https://issues.redhat.com/browse/HOSTEDCP-1960)

**Checklist**

### DIFF
--- a/api/hypershift/v1beta1/hosted_controlplane.go
+++ b/api/hypershift/v1beta1/hosted_controlplane.go
@@ -115,6 +115,20 @@ type HostedControlPlaneSpec struct {
 	// +optional
 	KubeConfig *KubeconfigSecretRef `json:"kubeconfig,omitempty"`
 
+	// kubeAPIServerDNSName specifies a desired DNS name to resolve to the KAS.
+	// When set, the controller will automatically generate a secret with kubeconfig and expose it in the hostedCluster Status.customKubeconfig field.
+	// If it's set or removed day 2, the kubeconfig generated secret will be created, recreated or deleted.
+	// The DNS entries should be resolvable from the cluster, so this should be manually configured in the DNS provider.
+	// This field works in conjunction with configuration.APIServer.ServingCerts.NamedCertificates to enable
+	// access to the API server via a custom domain name. The NamedCertificates provide the TLS certificates
+	// for the custom domain, while this field triggers the generation of a kubeconfig that uses those certificates.
+	//
+	// +kubebuilder:validation:XValidation:rule=`self == "" || self.matches('^(?:(?:[a-zA-Z0-9-]+\\.)+[a-zA-Z]{2,}|[a-zA-Z0-9-]+)$')`,message="kubeAPIServerDNSName must be a valid URL name (e.g., api.example.com)"
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:example: "api.example.com"
+	// +optional
+	KubeAPIServerDNSName string `json:"kubeAPIServerDNSName,omitempty"`
+
 	// Services defines metadata about how control plane services are published
 	// in the management cluster.
 	// +kubebuilder:validation:MaxItems=6
@@ -313,6 +327,15 @@ type HostedControlPlaneStatus struct {
 	// KubeConfig is a reference to the secret containing the default kubeconfig
 	// for this control plane.
 	KubeConfig *KubeconfigSecretRef `json:"kubeConfig,omitempty"`
+
+	// customKubeconfig references an external custom kubeconfig secret.
+	// This field is populated in the status when a custom kubeconfig secret has been generated
+	// for the hosted cluster. It contains the name and key of the secret located in the
+	// hostedCluster namespace. This field is only populated when kubeApiExternalName is set.
+	// If this field is removed during a day 2 operation, the referenced secret will be deleted
+	// and this field will be removed from the hostedCluster status.
+	// +optional
+	CustomKubeconfig *KubeconfigSecretRef `json:"customKubeconfig,omitempty"`
 
 	// KubeadminPassword is a reference to the secret containing the initial kubeadmin password
 	// for the guest cluster.
--- a/api/hypershift/v1beta1/hostedcluster_types.go
+++ b/api/hypershift/v1beta1/hostedcluster_types.go
@@ -471,6 +471,20 @@ type HostedClusterSpec struct {
 	// +required
 	Platform PlatformSpec `json:"platform"`
 
+	// kubeAPIServerDNSName specifies a desired DNS name to resolve to the KAS.
+	// When set, the controller will automatically generate a secret with kubeconfig and expose it in the hostedCluster Status.customKubeconfig field.
+	// If it's set or removed day 2, the kubeconfig generated secret will be created, recreated or deleted.
+	// The DNS entries should be resolvable from the cluster, so this should be manually configured in the DNS provider.
+	// This field works in conjunction with configuration.APIServer.ServingCerts.NamedCertificates to enable
+	// access to the API server via a custom domain name. The NamedCertificates provide the TLS certificates
+	// for the custom domain, while this field triggers the generation of a kubeconfig that uses those certificates.
+	// This API endpoint only works in OCP version 4.19 or later. Older versions will result in a no-op.
+	// +kubebuilder:validation:XValidation:rule=`self == "" || self.matches('^(?:(?:[a-zA-Z0-9-]+\\.)+[a-zA-Z]{2,}|[a-zA-Z0-9-]+)$')`,message="kubeAPIServerDNSName must be a valid URL name (e.g., api.example.com)"
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:example: "api.example.com"
+	// +optional
+	KubeAPIServerDNSName string `json:"kubeAPIServerDNSName,omitempty"`
+
 	// controllerAvailabilityPolicy specifies the availability policy applied to critical control plane components like the Kube API Server.
 	// Possible values are HighlyAvailable and SingleReplica. The default value is HighlyAvailable.
 	// This field is immutable.
@@ -1477,6 +1491,11 @@ type HostedClusterStatus struct {
 	// for the cluster.
 	// +optional
 	KubeConfig *corev1.LocalObjectReference `json:"kubeconfig,omitempty"`
+
+	// CustomKubeconfig is a local secret reference to the external custom kubeconfig.
+	// Once the hypershift operator sets this status field, it will generate a secret with the specified name containing a kubeconfig within the `HostedCluster` namespace.
+	// +optional
+	CustomKubeconfig *corev1.LocalObjectReference `json:"customKubeconfig,omitempty"`
 
 	// KubeadminPassword is a reference to the secret that contains the initial
 	// kubeadmin user password for the guest cluster.
--- a/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
@@ -739,7 +739,8 @@ func (r *HostedControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.R
 		meta.SetStatusCondition(&hostedControlPlane.Status.Conditions, condition)
 	}
 
-	kubeconfig := manifests.KASExternalKubeconfigSecret(hostedControlPlane.Namespace, hostedControlPlane.Spec.KubeConfig)
+	// Admin Kubeconfig
+	kubeconfig := manifests.KASAdminKubeconfigSecret(hostedControlPlane.Namespace, hostedControlPlane.Spec.KubeConfig)
 	if err := r.Get(ctx, client.ObjectKeyFromObject(kubeconfig), kubeconfig); err != nil {
 		if !apierrors.IsNotFound(err) {
 			return reconcile.Result{}, err
@@ -749,9 +750,14 @@ func (r *HostedControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.R
 			Name: kubeconfig.Name,
 			Key:  DefaultAdminKubeconfigKey,
 		}
+
 		if hostedControlPlane.Spec.KubeConfig != nil {
 			hostedControlPlane.Status.KubeConfig.Key = hostedControlPlane.Spec.KubeConfig.Key
 		}
+	}
+
+	if err := setKASCustomKubeconfigStatus(ctx, hostedControlPlane, r.Client); err != nil {
+		return reconcile.Result{}, err
 	}
 
 	explicitOauthConfig := hostedControlPlane.Spec.Configuration != nil && hostedControlPlane.Spec.Configuration.OAuth != nil
@@ -3008,12 +3014,33 @@ func (r *HostedControlPlaneReconciler) reconcileKubeAPIServer(ctx context.Contex
 		return fmt.Errorf("failed to reconcile localhost kubeconfig secret: %w", err)
 	}
 
-	externalKubeconfigSecret := manifests.KASExternalKubeconfigSecret(hcp.Namespace, hcp.Spec.KubeConfig)
-	if _, err := createOrUpdate(ctx, r, externalKubeconfigSecret, func() error {
-		if !util.IsPublicHCP(hcp) && !util.IsRouteKAS(hcp) {
-			return kas.ReconcileExternalKubeconfigSecret(externalKubeconfigSecret, clientCertSecret, rootCA, p.OwnerRef, p.InternalURL(), p.ExternalKubeconfigKey())
+	// Generate the new KASCustomKubeconfig secret if the KubeAPIServerDNSName is set
+	kasCustomKubeconfigSecret := manifests.KASCustomKubeconfigSecret(hcp.Namespace, nil)
+	if len(hcp.Spec.KubeAPIServerDNSName) > 0 {
+		newRootCA, err := includeServingCertificates(ctx, r.Client, hcp, rootCA)
+		if err != nil {
+			return fmt.Errorf("failed to include serving certificates: %w", err)
 		}
-		return kas.ReconcileExternalKubeconfigSecret(externalKubeconfigSecret, clientCertSecret, rootCA, p.OwnerRef, p.ExternalURL(), p.ExternalKubeconfigKey())
+
+		if _, err := createOrUpdate(ctx, r, kasCustomKubeconfigSecret, func() error {
+			return kas.ReconcileKASCustomKubeconfigSecret(kasCustomKubeconfigSecret, clientCertSecret, newRootCA, p.OwnerRef, p.CustomExternalURL(), p.KASCustomKubeconfigKey())
+		}); err != nil {
+			return fmt.Errorf("failed to reconcile custom external kubeconfig secret: %w", err)
+		}
+	} else {
+		// Cleanup the new kasCustomKubeconfigSecret secret if the KubeAPIServerDNSName is removed
+		if _, err := util.DeleteIfNeeded(ctx, r.Client, kasCustomKubeconfigSecret); err != nil {
+			return fmt.Errorf("failed to delete customKubeconfig status from HCP object: %w", err)
+		}
+	}
+
+	// Renamed the old externalKubeconfigSecret to adminKubeconfigSecret
+	adminKubeconfigSecret := manifests.KASAdminKubeconfigSecret(hcp.Namespace, hcp.Spec.KubeConfig)
+	if _, err := createOrUpdate(ctx, r, adminKubeconfigSecret, func() error {
+		if !util.IsPublicHCP(hcp) && !util.IsRouteKAS(hcp) {
+			return kas.ReconcileKASCustomKubeconfigSecret(adminKubeconfigSecret, clientCertSecret, rootCA, p.OwnerRef, p.InternalURL(), p.ExternalKubeconfigKey())
+		}
+		return kas.ReconcileKASCustomKubeconfigSecret(adminKubeconfigSecret, clientCertSecret, rootCA, p.OwnerRef, p.ExternalURL(), p.ExternalKubeconfigKey())
 	}); err != nil {
 		return fmt.Errorf("failed to reconcile external kubeconfig secret: %w", err)
 	}
@@ -5571,7 +5598,7 @@ func (r *HostedControlPlaneReconciler) validateAzureKMSConfig(ctx context.Contex
 }
 
 func (r *HostedControlPlaneReconciler) GetGuestClusterClient(ctx context.Context, hcp *hyperv1.HostedControlPlane) (*kubernetes.Clientset, error) {
-	kubeconfigSecret := manifests.KASExternalKubeconfigSecret(hcp.Namespace, hcp.Spec.KubeConfig)
+	kubeconfigSecret := manifests.KASAdminKubeconfigSecret(hcp.Namespace, hcp.Spec.KubeConfig)
 	if err := r.Get(ctx, client.ObjectKeyFromObject(kubeconfigSecret), kubeconfigSecret); err != nil {
 		return nil, err
 	}
@@ -5675,4 +5702,55 @@ func (r *HostedControlPlaneReconciler) verifyResourceGroupLocationsMatch(ctx con
 		return fmt.Errorf("the locations of the resource groups do not match - vnet location: %v; network security group location: %v; managed resource group location: %v", ptr.Deref(vnet.Location, ""), ptr.Deref(nsg.Location, ""), ptr.Deref(rg.Location, ""))
 	}
 	return nil
+}
+
+func setKASCustomKubeconfigStatus(ctx context.Context, hcp *hyperv1.HostedControlPlane, c client.Client) error {
+	customKubeconfig := manifests.KASCustomKubeconfigSecret(hcp.Namespace, nil)
+	if err := c.Get(ctx, client.ObjectKeyFromObject(customKubeconfig), customKubeconfig); err != nil {
+		if !apierrors.IsNotFound(err) {
+			return fmt.Errorf("failed to get custom kubeconfig secret: %w", err)
+		}
+	}
+
+	if len(hcp.Spec.KubeAPIServerDNSName) > 0 {
+		// Reconcile custom kubeconfig status
+		hcp.Status.CustomKubeconfig = &hyperv1.KubeconfigSecretRef{
+			Name: customKubeconfig.Name,
+			Key:  DefaultAdminKubeconfigKey,
+		}
+	} else {
+		// Cleanning up custom kubeconfig status
+		hcp.Status.CustomKubeconfig = nil
+	}
+
+	return nil
+}
+
+// includeServingCertificates includes additional serving certificates into the provided root CA ConfigMap.
+// It retrieves the named certificates specified in the HostedControlPlane's APIServer configuration and appends
+// their contents to the "ca.crt" entry in the root CA ConfigMap.
+func includeServingCertificates(ctx context.Context, c client.Client, hcp *hyperv1.HostedControlPlane, rootCA *corev1.ConfigMap) (*corev1.ConfigMap, error) {
+	var tlsCRT string
+	newRootCA := rootCA.DeepCopy()
+
+	if hcp.Spec.Configuration != nil && hcp.Spec.Configuration.APIServer != nil && len(hcp.Spec.Configuration.APIServer.ServingCerts.NamedCertificates) > 0 {
+		for _, servingCert := range hcp.Spec.Configuration.APIServer.ServingCerts.NamedCertificates {
+			newCRT := &corev1.Secret{}
+			if err := c.Get(ctx, client.ObjectKey{Namespace: hcp.Namespace, Name: servingCert.ServingCertificate.Name}, newCRT); err != nil {
+				return nil, fmt.Errorf("failed to get serving certificate secret: %w", err)
+			}
+
+			if len(tlsCRT) <= 0 {
+				tlsCRT = newRootCA.Data["tls.crt"]
+			}
+
+			tlsCRT = fmt.Sprintf("%s\n%s", tlsCRT, string(newCRT.Data["tls.crt"]))
+		}
+
+		if len(tlsCRT) > 0 {
+			newRootCA.Data["tls.crt"] = tlsCRT
+		}
+	}
+
+	return newRootCA, nil
 }
--- a/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
@@ -5741,14 +5741,14 @@ func includeServingCertificates(ctx context.Context, c client.Client, hcp *hyper
 			}
 
 			if len(tlsCRT) <= 0 {
-				tlsCRT = newRootCA.Data["tls.crt"]
+				tlsCRT = newRootCA.Data["ca.crt"]
 			}
 
 			tlsCRT = fmt.Sprintf("%s\n%s", tlsCRT, string(newCRT.Data["tls.crt"]))
 		}
 
 		if len(tlsCRT) > 0 {
-			newRootCA.Data["tls.crt"] = tlsCRT
+			newRootCA.Data["ca.crt"] = tlsCRT
 		}
 	}
 
--- a/control-plane-operator/controllers/hostedcontrolplane/v2/kas/kubeconfig.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/v2/kas/kubeconfig.go
@@ -101,6 +101,7 @@ func adapExternalAdminKubeconfigSecret(cpContext component.WorkloadContext, secr
 	}
 
 	url := externalURL(cpContext.InfraStatus)
+
 	if !util.IsPublicHCP(cpContext.HCP) && !util.IsRouteKAS(cpContext.HCP) {
 		url = internalURL(cpContext.InfraStatus, cpContext.HCP.Name)
 	}
@@ -116,7 +117,26 @@ func adapExternalAdminKubeconfigSecret(cpContext component.WorkloadContext, secr
 	return nil
 }
 
+func adaptCustomAdminKubeconfigSecret(cpContext component.WorkloadContext, secret *corev1.Secret) error {
+	hcp := cpContext.HCP
+	apiServerPort := util.KASPodPort(hcp)
+	url := customExternalURL(hcp.Spec.KubeAPIServerDNSName, apiServerPort)
+	kubeconfig, err := GenerateKubeConfig(cpContext, manifests.SystemAdminClientCertSecret(hcp.Namespace), url)
+	if err != nil {
+		return fmt.Errorf("failed to generate kubeconfig: %w", err)
+	}
+
+	if secret.Data == nil {
+		secret.Data = map[string][]byte{}
+	}
+	secret.Data[externalKubeconfigKey(hcp)] = kubeconfig
+
+	return nil
+
+}
+
 func adaptBootstrapKubeconfigSecret(cpContext component.WorkloadContext, secret *corev1.Secret) error {
+
 	url := externalURL(cpContext.InfraStatus)
 	if util.IsPrivateHCP(cpContext.HCP) {
 		url = internalURL(cpContext.InfraStatus, cpContext.HCP.Name)
@@ -201,13 +221,17 @@ func InClusterKASURL(platformType hyperv1.PlatformType) string {
 	return fmt.Sprintf("https://%s:%d", manifests.KubeAPIServerServiceName, config.KASSVCPort)
 }
 
+func customExternalURL(address string, port int32) string {
+	return fmt.Sprintf("https://%s:%d", pki.AddBracketsIfIPv6(address), port)
+}
+
 func externalURL(infraStatus infra.InfrastructureStatus) string {
 	return fmt.Sprintf("https://%s:%d", pki.AddBracketsIfIPv6(infraStatus.APIHost), infraStatus.APIPort)
 }
 
 func internalURL(infraStatus infra.InfrastructureStatus, hcpName string) string {
 	internalAddress := fmt.Sprintf("api.%s.hypershift.local", hcpName)
-	return fmt.Sprintf("https://%s:%d", pki.AddBracketsIfIPv6(internalAddress), infraStatus.APIPort)
+	return fmt.Sprintf("https://%s:%d", internalAddress, infraStatus.APIPort)
 }
 
 func externalKubeconfigKey(hcp *hyperv1.HostedControlPlane) string {
--- a/docs/content/how-to/aws/define-custom-kube-api-name.md
+++ b/docs/content/how-to/aws/define-custom-kube-api-name.md
@@ -1,0 +1,35 @@
+---
+title: Define Custom KubeAPI Name
+---
+
+## What is this for?
+
+`KubeAPICustomName` is a spec field used to declare a custom Kubernetes API URI. To make this work, you simply need to define the URI (e.g., `api.example.com`) in the `HostedCluster` object.
+
+## How does this work?
+
+- This can be defined both during day-1 (initial setup) and day-2 (post-deployment updates).
+- The CPO (ControlPlaneOperator) controllers will create a new kubeconfig stored in the HCP namespace. This kubeconfig will be based on certificates and named `custom-admin-kubeconfig`.
+- The certificates are generated from the root CA, with their expiration and renewal managed by the `HostedControlPlane`.
+- The CPO will report a new kubeconfig, called `CustomKubeconfig`, in the `HostedControlPlane`. This kubeconfig will use the new server defined in the `KubeAPICustomName` field.
+- This custom kubeconfig will also be referenced in the `HostedCluster` object under the status field as `CustomKubeconfig`.
+- A new secret, named `{HOSTEDCLUSTER_NAME}-custom-admin-kubeconfig`, will be created in the `HostedCluster` namespace. This secret can be used to easily access the HostedCluster API server.
+
+!!! NOTE
+    This does not directly affect the dataplane, so no rollouts are expected to occur.
+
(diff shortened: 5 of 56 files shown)


api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -432,6 +432,13 @@ type HostedClusterSpec struct {
 	// +required
 	Platform PlatformSpec `json:"platform"`
 
+	// kubeApiExternalName specifies the external name for the Kube API Server service.
+	// +kubebuilder:validation:MaxLength=100
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:example: "api.example.com"
+	// +optional
+	KubeAPIExternalName string `json:"kubeApiExternalName,omitempty"`
```
enxebre, line 440: kubeApiExternalName -> kubeAPIExternalName

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1423,6 +1430,10 @@ type HostedClusterStatus struct {
 	// +optional
 	KubeConfig *corev1.LocalObjectReference `json:"kubeconfig,omitempty"`
 
+	// CustomKubeConfig specifies the name and key for the External Custom kubeconfig secret
```
enxebre, line 1433: Let's use lower case by convention for api field docs so gdocs generation result in the serialized version of the field, i.e. "customKubeConfig"
Let's document what happens if I set and don't set this field day 1 and day 2.
E.g. "This field is optional and only allowed if kubeApiExternalName is not empty. When set, they result in the generation of a secret with the given name containing a Kubeconfig within the hostedCluster namespace and a referenced by the hostedCluster.status.... When removed day 2...."

Is the underlying business logic tied with the .kubeAPIExternalName? If so let's have cel validation here to capture that. Also we can probably put both fields behind the same struct.

we use "custom" term here but "external..." above can we consolidate on using a single one all the time.

api/hypershift/v1beta1/hostedcluster_types.go
enxebre: See https://github.com/openshift/hypershift/pull/5458/files#r1930322503

enxebre: Let's use same cel regex validation we do for BaseDomain and others

enxebre:
>HostedControlPlanes does not follows the separation of API and API-INT

Can we elaborate on this? Which topologies is this use case targeting: privateAndPublic, private, public?
For privateAndPublic we do have decouple internal and external urls routing to kas, so what's this solving?
Seems the PR is effectively implementing a kubeconfig autogeneration using a passed through URL, let's make sure that's clearly stated in the PR desc and API docs, also let's clarify how this is planned to be consumed and that consumption solves the driving rfe.

api/hypershift/v1beta1/hostedcluster_types.go
jparrill: Makes sense, I've set all to custom which I think is the most accurate term, I also solved the comments you mentioned in the review

api/hypershift/v1beta1/hosted_controlplane.go
```diff
@@ -202,6 +206,26 @@ type HostedControlPlaneSpec struct {
...
+// +optional
+type KubeAPICustomKubeconfig struct {
+	// CustomKubeConfig specifies the name and key for the External Custom kubeconfig secret.
+	// When set, it triggers the generation of a secret with the specified name containing a kubeconfig within the `HostedCluster` namespace.
+	// This kubeconfig will also be referenced in the `HostedCluster.status` as `customKubeconfig`.
+	// If removed during day-2 operations, all related secrets and status references will also be deleted.
+	// +optional
+	CustomKubeConfig *KubeconfigSecretRef `json:"customkubeconfig,omitempty"`
+
+	// kubeApiCustomName specifies the external name for the Kube API Server service.
```
csrwng, line 219: s/kubeApiCustomName/kubeAPICustomName

api/hypershift/v1beta1/hosted_controlplane.go
```diff
@@ -115,6 +115,10 @@ type HostedControlPlaneSpec struct {
 	// +optional
 	KubeConfig *KubeconfigSecretRef `json:"kubeconfig,omitempty"`
 
+	// KubeAPICustomKubeconfig is a group that encompasses an additional KAS server address and the customKubeconfig generated for the kube-apiserver.
```
csrwng, line 118: fix

api/hypershift/v1beta1/hosted_controlplane.go
```diff
@@ -300,6 +324,12 @@ type HostedControlPlaneStatus struct {
 	// for this control plane.
 	KubeConfig *KubeconfigSecretRef `json:"kubeConfig,omitempty"`
 
+	// CustomKubeConfig specifies the name and key for the External Custom kubeconfig secret.
```
csrwng, line 327: fix

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -432,6 +432,15 @@ type HostedClusterSpec struct {
 	// +required
 	Platform PlatformSpec `json:"platform"`
 
+	// kubeApiCustomName specifies the external name for the Kube API Server service.
```
csrwng, line 435: s/kubeApiCustomName/kubeAPICustomName/

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1423,6 +1432,13 @@ type HostedClusterStatus struct {
 	// +optional
 	KubeConfig *corev1.LocalObjectReference `json:"kubeconfig,omitempty"`
 
+	// CustomKubeConfig specifies the name and key for the External Custom kubeconfig secret.
```
csrwng, line 1435: fix

csrwng (commented):
Some nits in the API. In general, please use consistent case for "kubeconfig" wherever it appears in the API.

control-plane-operator/controllers/hostedcontrolplane/v2/kas/deployment.go
```diff
@@ -90,6 +90,29 @@ func adaptDeployment(cpContext component.WorkloadContext, deployment *appsv1.Dep
 	return nil
 }
 
+func adaptCustomAdminKubeconfigSecret(cpContext component.WorkloadContext, secret *corev1.Secret) error {
+	hcp := cpContext.HCP
+
+	if hcp.Spec.KubeConfig != nil {
```
muraee, line 96: I assume you want `hcp.Spec.KubeAPICustomKubeconfig != nil` here, right?

control-plane-operator/controllers/hostedcontrolplane/v2/kas/deployment.go
```diff
@@ -90,6 +90,29 @@ func adaptDeployment(cpContext component.WorkloadContext, deployment *appsv1.Dep
 	return nil
 }
 
+func adaptCustomAdminKubeconfigSecret(cpContext component.WorkloadContext, secret *corev1.Secret) error {
```
muraee, line 93: please move this function to `kubeconfig.go` file

support/controlplane-component/common.go
```diff
@@ -41,3 +41,13 @@ func EnableForPlatform(platform hyperv1.PlatformType) option {
 		return cpContext.HCP.Spec.Platform.Type == platform
 	})
 }
+
+// EnableIfCustomKubeconfig is a helper predicate for the common use case of enabling a resource when a KubeAPICustomKubeconfig is specified.
+func EnableIfCustomKubeconfig() option {
+	return WithPredicate(func(cpContext WorkloadContext) bool {
```
muraee, line 47: I don't think any other component will use this predicate other than kas.
let's move it to the kas package where its used. 
See this as an example https://github.com/muraee/hypershift/blob/karpenter-userdata/control-plane-operator/controllers/hostedcontrolplane/v2/cno/component.go#L47

enxebre:
can we please split the commits having a dedicated one for the autogenerated code so it's easier to exercise reviews

api/hypershift/v1beta1/hosted_controlplane.go
```diff
@@ -202,6 +206,26 @@ type HostedControlPlaneSpec struct {
...
 }
 
+// kubeAPICustomKubeconfig is a group that englobes an additional KAS server address and the customkubeconfig generated for the kube-apiserver.
+// +optional
+type KubeAPICustomKubeconfig struct {
+	// customKubeConfig specifies the name and key for the External Custom kubeconfig secret.
+	// When set, it triggers the generation of a secret with the specified name containing a kubeconfig within the `HostedCluster` namespace.
+	// This kubeconfig will also be referenced in the `HostedCluster.status` as `customkubeconfig` (remarking that in the HC status is lowercase meanwhile this one is camelCase to keep the consistency with the rest of the spec).
+	// If removed during day-2 operations, all related secrets and status references will also be deleted.
+	// +optional
```
enxebre, line 216: please add docs and cel validation that only pass if these fields are both set, unless other cases are supported.

api/hypershift/v1beta1/hosted_controlplane.go
```diff
@@ -115,6 +115,10 @@ type HostedControlPlaneSpec struct {
 	// +optional
 	KubeConfig *KubeconfigSecretRef `json:"kubeconfig,omitempty"`
 
+	// kubeAPICustomKubeconfig is a group that encompasses an additional KAS server address and the customkubeconfig generated for the kube-apiserver.
```
enxebre, line 118: Can we capture all the following:
customKubeconfig specifies the configuration to generate a kubeconfig resolving through a kasDNSName. For this kubeconfig to successfully resolve, this needs to be used in conjunction with hostedCluster.apiServer.ServingCerts and hostedCluster.apiServer.ClientCA. DNS for this name resolution needs to be setup out of band to resolve to the kas publishing strategy

api/hypershift/v1beta1/hosted_controlplane.go
```diff
@@ -202,6 +206,26 @@ type HostedControlPlaneSpec struct {
...
+	CustomKubeConfig *KubeconfigSecretRef `json:"customKubeConfig,omitempty"`
+
+	// kubeAPICustomName specifies the external name for the Kube API Server service.
+	// This is used to configure the Kube API Server service to use a different name, so
+	// the DNS should be configured to resolve the external name to the Kube API Server service.
+	// +kubebuilder:validation:XValidation:rule=`self == "" || self.matches('^(?:(?:[a-zA-Z0-9-]+\\.)+[a-zA-Z]{2,}|[a-zA-Z0-9-]+)$')`,message="kubeAPICustomName must be a valid URL name (e.g., api.example.com)"
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:example: "api.example.com"
+	// +optional
+	KubeAPICustomName string `json:"kubeAPICustomName,omitempty"`
```
enxebre, line 226: this is within customKubeconfig shoud no need to name it "custom" again, also this is a dns name, right?
I would name similar to
customKubeConfig {
 kasDNSName
 secretRef
}

api/hypershift/v1beta1/hosted_controlplane.go
jparrill: I don't think the validation is necessary at all:
- The user defines the `KasDNSName` in the HostedCluster
- The Controller grab that name and put it in the HostedControlPlane object inside of the `CustomKubeconfig` struct
- The Controller will also generate the Certificate as an Opaque `SecretRef`.

So this will be managed by us and not by the user and does not needs to be controlled by CEL.

enxebre: ok, so If this is only an implementation detail with well known values then, there's no need for it in the API at all.

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -736,11 +737,51 @@ func (r *HostedControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.R
 			Name: kubeconfig.Name,
 			Key:  DefaultAdminKubeconfigKey,
 		}
+
 		if hostedControlPlane.Spec.KubeConfig != nil {
 			hostedControlPlane.Status.KubeConfig.Key = hostedControlPlane.Spec.KubeConfig.Key
 		}
 	}
 
+	if hostedControlPlane.Spec.KasCustomKubeconfig != nil && len(hostedControlPlane.Spec.KasCustomKubeconfig.KasDNSName) > 0 {
```
enxebre, line 746: please unit test this

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -2975,12 +3016,29 @@ func (r *HostedControlPlaneReconciler) reconcileKubeAPIServer(ctx context.Contex
 		return fmt.Errorf("failed to reconcile localhost kubeconfig secret: %w", err)
 	}
 
-	externalKubeconfigSecret := manifests.KASExternalKubeconfigSecret(hcp.Namespace, hcp.Spec.KubeConfig)
-	if _, err := createOrUpdate(ctx, r, externalKubeconfigSecret, func() error {
+	// Generate the new externalKubeconfigSecret if the external name is set
```
enxebre, line 3019: please unit test this

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -1967,6 +2035,19 @@ func reconcileHostedControlPlane(hcp *hyperv1.HostedControlPlane, hcluster *hype
 		hcp.Spec.SecretEncryption = hcluster.Spec.SecretEncryption.DeepCopy()
 	}
 
+	// If ROSA passes KubeAPICustomKubeconfig.CustomKubeConfig, we should pass it through to the HCP, otherwise
```
enxebre, line 2038: what is KubeAPICustomKubeconfig.CustomKubeConfig?

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -1967,6 +2035,19 @@ func reconcileHostedControlPlane(hcp *hyperv1.HostedControlPlane, hcluster *hype
...
+	// we will keep it in blank for name auto-generation.
+	// If ROSA passes the KasDNSName but KasDNSName it's not set in the HC, we will not set anything.
+	if len(hcluster.Spec.KasDNSName) >= 0 {
+		if hcp.Spec.KasCustomKubeconfig != nil && hcp.Spec.KasCustomKubeconfig.SecretRef != nil {
```
enxebre, line 2042: why do we need an if here or the KasCustomKubeconfig struct? seems all the info the hcp controller needs is the KasDNSName

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -2975,12 +3016,29 @@ func (r *HostedControlPlaneReconciler) reconcileKubeAPIServer(ctx context.Contex
...
+	if len(hcp.Spec.KasCustomKubeconfig.KasDNSName) > 0 {
+		if hcp.Spec.Configuration != nil && hcp.Spec.Configuration.APIServer != nil && len(hcp.Spec.Configuration.APIServer.ClientCA.Name) > 0 {
+			rootCA = manifests.KASExternalCAConfigMap(hcp.Spec.Configuration.APIServer.ClientCA.Name)
```
jparrill, line 3022: ClientCA is not the required field, but https://docs.openshift.com/container-platform/4.17/rest_api/config_apis/apiserver-config-openshift-io-v1.html#spec-servingcerts-namedcertificates-2

api/hypershift/v1beta1/hosted_controlplane.go
jparrill: I'm followed the same approach than we do with `kubeconfig`, I was thinking this is necessary for ROSA or MCE in any way.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
jparrill: This is what I've mentioned above https://github.com/openshift/hypershift/pull/5458#discussion_r1949425238 so maybe I'm wrong. If this is not used by MCE or ROSA, I can remove it.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
jparrill: same as above

jparrill: To clarify, I've removed the `hcp.spec.KasCustomKubeconfig` because it's an old implementation detail that is not relevant anymore and I've changed the PR to simplify the implementation in that way.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
jparrill: To clarify, I've removed the hcp.spec.KasCustomKubeconfig because it's an old implementation detail that is not relevant anymore and I've changed the PR to simplify the implementation in that way.

api/hypershift/v1beta1/hosted_controlplane.go
jparrill: To clarify, I've removed the hcp.spec.KasCustomKubeconfig because it's an old implementation detail that is not relevant anymore and I've changed the PR to simplify the implementation in that way.

api/hypershift/v1beta1/hosted_controlplane.go
```diff
@@ -115,6 +115,16 @@ type HostedControlPlaneSpec struct {
...
 
+	// kasDNSName specifies a desired DNS name to resolve to the KAS.
+	// This can only be set if secretRef is also set. When set, the controller will automatically generate a secret with kubeconfig and expose it in the hostedCluster Status.
+	// If it's set or removed day 2, all generated objects will be created, recreated or deleted.
+	// The DNS entries should be resolvable from the cluster, so this should be manually configured in the DNS provider.
+	// +kubebuilder:validation:XValidation:rule=`self == "" || self.matches('^(?:(?:[a-zA-Z0-9-]+\\.)+[a-zA-Z]{2,}|[a-zA-Z0-9-]+)$')`,message="kasDNSName must be a valid URL name (e.g., api.example.com)"
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:example: "api.example.com"
+	// +optional
+	KasDNSName string `json:"kasDNSName,omitempty"`
```
csrwng, line 126: s/KasDNSName/KubeAPIServerDNSName/

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -456,6 +456,16 @@ type HostedClusterSpec struct {
...
 
+	// kasDNSName specifies a desired DNS name to resolve to the KAS.
+	// This can only be set if secretRef is also set. When set, the controller will automatically generate a secret with kubeconfig and expose it in the hostedCluster Status.
+	// If it's set or removed day 2, all generated objects will be created, recreated or deleted.
+	// The DNS entries should be resolvable from the cluster, so this should be manually configured in the DNS provider.
+	// +kubebuilder:validation:XValidation:rule=`self == "" || self.matches('^(?:(?:[a-zA-Z0-9-]+\\.)+[a-zA-Z]{2,}|[a-zA-Z0-9-]+)$')`,message="kubeApiCustomName must be a valid URL name (e.g., api.example.com)"
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:example: "api.example.com"
+	// +optional
+	KasDNSName string `json:"kasDNSName,omitempty"`
```
csrwng, line 467: s/KasDNSName/KubeAPIServerDNSName/

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -5764,3 +5786,57 @@ func verifyResourceGroupLocationsMatch(ctx context.Context, hcp *hyperv1.HostedC
...
+	newRootCA := rootCA.DeepCopy()
+
+	if hcp.Spec.Configuration != nil && hcp.Spec.Configuration.APIServer != nil && len(hcp.Spec.Configuration.APIServer.ServingCerts.NamedCertificates) > 0 {
+		for _, servingCert := range hcp.Spec.Configuration.APIServer.ServingCerts.NamedCertificates {
+			newCRT := &corev1.Secret{}
+			if err := c.Get(ctx, client.ObjectKey{Namespace: hcp.Namespace, Name: servingCert.ServingCertificate.Name}, newCRT); err != nil {
+				return nil, fmt.Errorf("failed to get serving certificate secret: %w", err)
+			}
+
+			if len(tlsCRT) <= 0 {
```
csrwng, line 5829: you can also just write all certs to a byte buffer.
see https://github.com/openshift/hypershift/blob/8f40e802d2e84a4717a718673cdc7db4453a3c86/control-plane-operator/controllers/hostedcontrolplane/pki/ca.go#L21-L28

csrwng (commented):
A couple of comments. I also don't see code to set the apiServerURL in the infrastructure status of the hosted cluster.

api/hypershift/v1beta1/hostedcluster_types.go
jparrill: done

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
jparrill: done

jparrill:
> A couple of comments. I also don't see code to set the apiServerURL in the infrastructure status of the hosted cluster.

Thanks @csrwng. Included!

docs/content/how-to/aws/define-custom-kube-api-name.md
```diff
@@ -0,0 +1,35 @@
...
+- A new secret, named `{HOSTEDCLUSTER_NAME}-custom-admin-kubeconfig`, will be created in the `HostedCluster` namespace. This secret can be used to easily access the HostedCluster API server.
+
+!!! NOTE
+    This does not directly affect the dataplane, so no rollouts are expected to occur.
+
+- If you remove this field from the spec, all newly generated secrets and the `CustomKubeconfig` reference will be removed from the cluster and from the status field.
+
+## Additional Notes
+
+This other field called `CustomKubeConfig` is optional and can only be used if `KubeAPICustomName` is not empty. When set, it triggers the generation of a secret with the specified name containing a kubeconfig within the `HostedCluster` namespace. This kubeconfig will also be referenced in the `HostedCluster.status` as `customkubeconfig`. If removed during day-2 operations, all related secrets and status references will also be deleted:
```
enxebre, line 25: This doesn't seem accurate anymore?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -456,6 +456,16 @@ type HostedClusterSpec struct {
 	// +required
 	Platform PlatformSpec `json:"platform"`
 
+	// KubeAPIServerDNSName specifies a desired DNS name to resolve to the KAS.
+	// This can only be set if secretRef is also set. When set, the controller will automatically generate a secret with kubeconfig and expose it in the hostedCluster Status.
```
enxebre, line 460: This comment doesn't seem accurate anymore?

Can we please also document in this API field how the kubeconfig generation it triggers will consume configuration.apiserver.ca/certs?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -456,6 +456,16 @@ type HostedClusterSpec struct {
...
+	// If it's set or removed day 2, all generated objects will be created, recreated or deleted.
```
enxebre, line 461: can we be specific i.e. r/"all generated objects"/ "the kubeconfig generated secret"?

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -5764,3 +5786,57 @@ func verifyResourceGroupLocationsMatch(ctx context.Context, hcp *hyperv1.HostedC
...
+
+	if len(hcp.Spec.KubeAPIServerDNSName) > 0 {
+		// Reconcile custom kubeconfig status and secret
+		hcp.Status.CustomKubeConfig = &hyperv1.KubeconfigSecretRef{
+			Name: customKubeconfig.Name,
+			Key:  DefaultAdminKubeconfigKey,
+		}
+	} else {
+		// Cleanning up custom kubeconfig status and secret if the external name is removed
+		if _, err := util.DeleteIfNeeded(ctx, c, customKubeconfig); err != nil {
```
enxebre, line 5806: issuing a delete request in setCustomKubeconfigStatus seems out of the scope of this func. I'd expect deletion to haven at the same place we reconcile for creation

control-plane-operator/controllers/hostedcontrolplane/v2/kas/kubeconfig.go
```diff
@@ -116,7 +117,31 @@ func adapExternalAdminKubeconfigSecret(cpContext component.WorkloadContext, secr
 	return nil
 }
 
+func adaptCustomAdminKubeconfigSecret(cpContext component.WorkloadContext, secret *corev1.Secret) error {
+	hcp := cpContext.HCP
+
+	if len(hcp.Spec.KubeAPIServerDNSName) > 0 {
+		secret.Name = hcp.Spec.KubeAPIServerDNSName
```
muraee, line 124: why are we changing the secret name here?

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -5764,3 +5786,57 @@ func verifyResourceGroupLocationsMatch(ctx context.Context, hcp *hyperv1.HostedC
...
+			newCRT := &corev1.Secret{}
+			if err := c.Get(ctx, client.ObjectKey{Namespace: hcp.Namespace, Name: servingCert.ServingCertificate.Name}, newCRT); err != nil {
+				return nil, fmt.Errorf("failed to get serving certificate secret: %w", err)
+			}
+
+			if len(tlsCRT) <= 0 {
+				tlsCRT = newRootCA.Data["tls.crt"]
+			}
+
+			tlsCRT = fmt.Sprintf("%s\n%s", tlsCRT, string(newCRT.Data["tls.crt"]))
```
enxebre, line 5833: can we please create a ticket to make sure we validate earlier the secrets referenced by Configuration.APIServer.ServingCerts.NamedCertificates has the expected keys or fail in a condition otherwise?

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -5764,3 +5786,57 @@ func verifyResourceGroupLocationsMatch(ctx context.Context, hcp *hyperv1.HostedC
...
+		}
+
+		if len(tlsCRT) > 0 {
```
enxebre, line 5836: how are the CAs for these certs plumbed in?

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
jparrill: https://issues.redhat.com/browse/CNTRLPLANE-237

control-plane-operator/controllers/hostedcontrolplane/v2/kas/kubeconfig.go
jparrill: nice catch 🙏

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -5764,3 +5791,54 @@ func verifyResourceGroupLocationsMatch(ctx context.Context, hcp *hyperv1.HostedC
...
+			newRootCA.Data["tls.crt"] = tlsCRT
```
enxebre, line 5839: you are storing this in tls.crt, but I can't see where that key is being consumed?
Once that's clarified let's have a unit that takes hcp.Spec.Configuration.APIServer and validate the kubeconfig generated content.

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
enxebre: those serving certs are already mounted on the kas pod. I don't see how they relate to the generated Kubeconfig?
What I would expect to meet this rfe is that the generated Kubeconfig includes the ca that let's the client to verify the server authenticity for the custom dns name used in the Kubeconfig, so the generated kubeconfig just works.
Also see https://github.com/openshift/hypershift/pull/5458/commits/fec7f62de45095649e3056b98c22cfb179ea5be8#r1954714596 you are storing the certs but they're not being consumed.

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -454,6 +454,16 @@ type HostedClusterSpec struct {
 	// +required
 	Platform PlatformSpec `json:"platform"`
 
+	// KubeAPIServerDNSName specifies a desired DNS name to resolve to the KAS.
+	// When set, the controller will automatically generate a secret with kubeconfig and expose it in the hostedCluster Status.
+	// If it's set or removed day 2, the kubeconfig generated secret will be created, recreated or deleted.
```
enxebre, line 459: Can we document how this field relates to configuration.APIServer.ServingCerts.NamedCertificates?

api/hypershift/v1beta1/hosted_controlplane.go
```diff
@@ -313,6 +323,12 @@ type HostedControlPlaneStatus struct {
 	// for this control plane.
 	KubeConfig *KubeconfigSecretRef `json:"kubeConfig,omitempty"`
 
+	// customKubeConfig specifies the name and key for the External Custom kubeconfig secret.
```
enxebre, line 326: This doc sounds like this is an spec field but it's a status one.
>When set, they result in the generation of a secret with the given name containing a Kubeconfig within the hostedCluster namespace and a referenced by the hostedCluster.status object.

This field doesn't cause a generation of a secret, this field is the result of generating that secret, right?

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -2987,12 +2993,33 @@ func (r *HostedControlPlaneReconciler) reconcileKubeAPIServer(ctx context.Contex
 		return fmt.Errorf("failed to reconcile localhost kubeconfig secret: %w", err)
 	}
 
-	externalKubeconfigSecret := manifests.KASExternalKubeconfigSecret(hcp.Namespace, hcp.Spec.KubeConfig)
-	if _, err := createOrUpdate(ctx, r, externalKubeconfigSecret, func() error {
+	// Generate the new customExternalKubeconfig secret if the KubeAPIServerDNSName is set
```
enxebre, line 2996: This seems to be using 3 different terms for the same thing: the term set(CustomKubeconfig)Status in l752, the term customExternalKubeconfig, and (KASCustomKubeconfig)Secret in l2997. Can we use one consistently everywhere?

enxebre: Another one Reconcile(ExternalKubeconfigSecret) in l3005

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -5634,3 +5661,54 @@ func verifyResourceGroupLocationsMatch(ctx context.Context, hcp *hyperv1.HostedC
...
+			}
+
+			tlsCRT = fmt.Sprintf("%s\n%s", tlsCRT, string(newCRT.Data["tls.crt"]))
+		}
+
+		if len(tlsCRT) > 0 {
+			newRootCA.Data["tls.crt"] = tlsCRT
+		}
+	}
+
```
enxebre, line 5712: can we drop this changes from the commit which are reverted by the next one?

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -5634,3 +5661,54 @@ func verifyResourceGroupLocationsMatch(ctx context.Context, hcp *hyperv1.HostedC
...
+func setCustomKubeconfigStatus(ctx context.Context, hcp *hyperv1.HostedControlPlane, c client.Client) error {
+	customKubeconfig := manifests.KASCustomKubeconfigSecret(hcp.Namespace, nil)
+	if err := c.Get(ctx, client.ObjectKeyFromObject(customKubeconfig), customKubeconfig); err != nil {
+		if !apierrors.IsNotFound(err) {
+			return fmt.Errorf("failed to get custom kubeconfig secret: %w", err)
+		}
+	}
+
+	if len(hcp.Spec.KubeAPIServerDNSName) > 0 {
```
enxebre, line 5673: shouldn't this only be set if the secret was found?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -454,6 +454,16 @@ type HostedClusterSpec struct {
...
+	// The DNS entries should be resolvable from the cluster, so this should be manually configured in the DNS provider.
+	// +kubebuilder:validation:XValidation:rule=`self == "" || self.matches('^(?:(?:[a-zA-Z0-9-]+\\.)+[a-zA-Z]{2,}|[a-zA-Z0-9-]+)$')`,message="kubeAPIServerDNSName must be a valid URL name (e.g., api.example.com)"
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:example: "api.example.com"
+	// +optional
+	KubeAPIServerDNSName string `json:"kubeAPIServerDNSName,omitempty"`
```
enxebre, line 465: do we foresee this config growing in future in a way that It would deserve having its own struct now so it can allocate more properties in future?

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
jparrill: 👍 Thanks! I will put all of them as *KASCustomKubeconfig*. I think is the most accurate name.

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -454,6 +454,20 @@ type HostedClusterSpec struct {
 	// +required
 	Platform PlatformSpec `json:"platform"`
 
+	// kubeAPIServerDNSName specifies a desired DNS name to resolve to the KAS.
+	// When set, the controller will automatically generate a secret with kubeconfig and expose it in the hostedCluster Status.customkubeconfig field.
+	// If it's set or removed day 2, the kubeconfig generated secret will be created, recreated or deleted.
+	// The DNS entries should be resolvable from the cluster, so this should be manually configured in the DNS provider.
+	// This field works in conjunction with configuration.APIServer.ServingCerts.NamedCertificates to enable
+	// access to the API server via a custom domain name. The NamedCertificates provide the TLS certificates
```
enxebre, line 462: nice, thanks!

api/hypershift/v1beta1/hosted_controlplane.go
```diff
@@ -115,6 +115,16 @@ type HostedControlPlaneSpec struct {
 	// +optional
 	KubeConfig *KubeconfigSecretRef `json:"kubeconfig,omitempty"`
 
+	// KubeAPIServerDNSName specifies a desired DNS name to resolve to the KAS.
+	// When set, the controller will automatically generate a secret with kubeconfig and expose it in the hostedCluster Status.
+	// If it's set or removed day 2, the kubeconfig generated secret will be created, recreated or deleted.
+	// The DNS entries should be resolvable from the cluster, so this should be manually configured in the DNS provider.
```
enxebre, line 121: can we make this doc match the one in hostedcluster_types?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1460,6 +1474,11 @@ type HostedClusterStatus struct {
 	// +optional
 	KubeConfig *corev1.LocalObjectReference `json:"kubeconfig,omitempty"`
 
+	// customkubeconfig is a local secret reference to the external custom kubeconfig.
+	// Once the hypershift operator sets this status field, it will generate a secret with the specified name containing a kubeconfig within the `HostedCluster` namespace.
+	// +optional
+	CustomKubeConfig *corev1.LocalObjectReference `json:"customkubeconfig,omitempty"`
```
enxebre, line 1480: the marker has a typo, should be customKubeconfig

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -454,6 +454,20 @@ type HostedClusterSpec struct {
...
 	Platform PlatformSpec `json:"platform"`
 
+	// kubeAPIServerDNSName specifies a desired DNS name to resolve to the KAS.
+	// When set, the controller will automatically generate a secret with kubeconfig and expose it in the hostedCluster Status.customkubeconfig field.
+	// If it's set or removed day 2, the kubeconfig generated secret will be created, recreated or deleted.
+	// The DNS entries should be resolvable from the cluster, so this should be manually configured in the DNS provider.
+	// This field works in conjunction with configuration.APIServer.ServingCerts.NamedCertificates to enable
+	// access to the API server via a custom domain name. The NamedCertificates provide the TLS certificates
+	// for the custom domain, while this field triggers the generation of a kubeconfig that uses those certificates.
+	//
```
enxebre, line 464: can we document that this will only work in >= 4.19 and will cause no-op otherwise?

test/e2e/util/util.go
```diff
@@ -1320,6 +1390,173 @@ func EnsureGuestWebhooksValidated(t *testing.T, ctx context.Context, guestClient
 	})
 }
 
+func EnsureKubeAPIDNSName(t *testing.T, ctx context.Context, mgmtClient crclient.Client, hc *hyperv1.HostedCluster) {
+	//AtLeast(t, Version419)
+	var (
+		hcKASCustomKubeconfigSecretName string
+		customApiServerHost             = "localhost"
```
enxebre, line 1397: can we validate this with a real domain (that might just resolve to 127.0.0.1) and passing configuration.apisever.certs?

api/hypershift/v1beta1/hostedcluster_types.go
jparrill: Humm following the above pattern of `KubeConfig` I've set CustomKubeconfig. Is that alright?

test/e2e/util/util.go
jparrill: yep, the plan I had is to add new E2E covering this onces this got merged. That works for you? (We need to got this merged before the feature freeze)

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -5634,3 +5661,54 @@ func verifyResourceGroupLocationsMatch(ctx context.Context, hcp *hyperv1.HostedC
...
+		// Cleanning up custom kubeconfig status
+		hcp.Status.CustomKubeconfig = nil
+	}
+
+	return nil
+}
+
+// includeServingCertificates includes additional serving certificates into the provided root CA ConfigMap.
+// It retrieves the named certificates specified in the HostedControlPlane's APIServer configuration and appends
+// their contents to the "tls.crt" entry in the root CA ConfigMap.
```
csrwng, line 5689: nit: there is no "tls.crt" entry in the root CA ConfigMap, it's "ca.crt"

support/util/util.go
```diff
@@ -655,3 +655,8 @@ func GetControlPlaneOperatorImageLabels(ctx context.Context, hc *hyperv1.HostedC
 
 	return ImageLabels(controlPlaneOperatorImageMetadata), nil
 }
+
+// EnableIfCustomKubeconfig returns true if the hosted control plane has a custom kubeconfig defined
+func EnableIfCustomKubeconfig(hcp *hyperv1.HostedControlPlane) bool {
```
csrwng, line 660: nit: the name of the function should reflect what it does. In this case this function is not enabling anything, all it's doing is determining if a KubeAPIServerDNSName is specified.

csrwng (commented):
lgtm overall
a couple of really minor nits that you can feel free to ignore
I tested this with a public AWS server with a custom certificate. Things worked as expected.

test/e2e/util/util.go
enxebre: without that test, there's zero evidence this works as it will be consumed and there's zero guarantees it won't be broken. We need that in place before this change ships into any publicly consumable artifact.

jparrill:
For xref, sent a PR to ACM UI to fix the marker name of CustomKubeconfig https://github.com/openshift-assisted/assisted-installer-ui/pull/2862

jparrill:
Because this PR is getting huge, I will separate [the another required E2E](https://issues.redhat.com/browse/CNTRLPLANE-217) into a new PR.

bryan-cox:

bryan-cox:

PR #5469, opened 2025-01-23
CNTRLPLANE-173: Assign known ARO roles for CP IDs for ARO HCP

bryan-cox:
This PR assigns the appropriate roles for control plane identities where existing ARO HCP roles already exist for ARO HCP.

Fixes [CNTRLPLANE-173](https://issues.redhat.com/browse/CNTRLPLANE-173)

**Checklist**

### DIFF
--- a/cmd/infra/azure/create.go
+++ b/cmd/infra/azure/create.go
@@ -825,28 +825,36 @@ func assignServicePrincipalRoles(subscriptionID, managedResourceGroupName, nsgRe
 	vnetRG := fmt.Sprintf("/subscriptions/%s/resourceGroups/%s", subscriptionID, vnetResourceGroupName)
 	dnsZoneRG := fmt.Sprintf("/subscriptions/%s/resourceGroups/%s", subscriptionID, dnsZoneResourceGroupName)
 
-	contributorRole := "Contributor"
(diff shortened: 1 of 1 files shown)


Patryk-Stefanski:
lgtm

cmd/infra/azure/create.go
```diff
@@ -825,26 +825,36 @@ func assignServicePrincipalRoles(subscriptionID, managedResourceGroupName, nsgRe
...
+	case ciro:
+		role = "Azure Red Hat OpenShift Image Registry Operator Role"
 	case nodePoolMgmt:
 		scopes = append(scopes, vnetRG)
 	}
 
 	// Assign contributor role to service principal
 	for _, scope := range scopes {
-		if err := assignRole(assigneeID, contributorRole, scope); err != nil {
+		if err := assignRole(assigneeID, role, scope); err != nil {
```
heliubj18, line 857: is the role value correct here? I see its value has been changed multiple times

bryan-cox: Discussed on a Slack huddle, this role is correct. Eventually all the CPO components will have their own role and none will have contributor role.

PR #5475, opened 2025-01-25
OCPBUGS-50518: Kubernetes API Server apply-bootstrap container does not respect SIGTERM

jared-hayes-dev:
Modify apply-bootstrap container to trap SIGTERM, and handle it correctly before exiting to ensure the container terminates within the specified grace period.

Fixes #

**Checklist**

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/kas/deployment.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/kas/deployment.go
@@ -799,6 +799,12 @@ cp /tmp/manifests/* %[1]s
 
 func applyBootstrapManifestsScript(workDir string) string {
 	var script = `#!/bin/sh
+function cleanup() {
+	pkill -P $$$
+	wait
+	exit
+}
+trap cleanup SIGTERM
 while true; do
   if oc apply -f %[1]s; then
     echo "Bootstrap manifests applied successfully."
@@ -814,7 +820,8 @@ while true; do
   sleep 1
 done
 while true; do
-  sleep 1000
+  sleep 1000 &
+  wait $!
 done
 `
 	return fmt.Sprintf(script, workDir)
--- a/control-plane-operator/controllers/hostedcontrolplane/testdata/kube-apiserver/zz_fixture_TestControlPlaneComponents.yaml
+++ b/control-plane-operator/controllers/hostedcontrolplane/testdata/kube-apiserver/zz_fixture_TestControlPlaneComponents.yaml
@@ -85,6 +85,12 @@ spec:
         - -c
         - |
           #!/bin/sh
+          function cleanup() {
+            pkill -P $$$
+            wait
+            exit
+          }
+          trap cleanup SIGTERM
           while true; do
             if oc apply -f /work; then
               echo "Bootstrap manifests applied successfully."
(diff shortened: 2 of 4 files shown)


rtheis:
@jared-hayes-dev similar to https://github.com/openshift/hypershift/pull/5491, please update the test data files too.

control-plane-operator/controllers/hostedcontrolplane/kas/deployment.go
```diff
@@ -799,6 +799,12 @@ cp /tmp/manifests/* %[1]s
 
 func applyBootstrapManifestsScript(workDir string) string {
 	var script = `#!/bin/sh
+function cleanup() {
```
rtheis, line 802: Please also change https://github.com/openshift/hypershift/blob/607f618b03351a2313a24261b76e2f580f371ca3/control-plane-operator/controllers/hostedcontrolplane/v2/assets/kube-apiserver/deployment.yaml#L26-L43

rtheis (changes requested):

control-plane-operator/controllers/hostedcontrolplane/kas/deployment.go
```diff
@@ -799,6 +799,12 @@ cp /tmp/manifests/* %[1]s
...
+	pkill -P $$$
+	wait
+	exit
+}
+trap cleanup SIGTERM
```
rtheis, line 807: As far as I can tell, my testing shows that this won't work if the script is running `sleep 1000` because you won't get control in the trap.  I think you need to send the sleep to the background and wait for it or shorten the sleep.

rtheis (changes requested):

control-plane-operator/controllers/hostedcontrolplane/kas/deployment.go
jared-hayes-dev: Agreed, the sleep should be run in bg and be waited on. I accidentally dropped that change on when I pushed up, will have it back.

rtheis: @jared-hayes-dev reminder to push the changes.

rtheis:
@jared-hayes-dev please provide a valid Jira reference.

jared-hayes-dev:
/retitle OCPBUGS-50518: Kubernetes API Server apply-bootstrap container does not respect SIGTERM

rtheis:
@Joseph-Goergen ptal

devguyio (commented):
Thanks! left a question

control-plane-operator/controllers/hostedcontrolplane/kas/deployment.go
```diff
@@ -799,6 +799,12 @@ cp /tmp/manifests/* %[1]s
 
 func applyBootstrapManifestsScript(workDir string) string {
 	var script = `#!/bin/sh
+function cleanup() {
+	pkill -P $$$
```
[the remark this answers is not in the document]
jared-hayes-dev: $$$ is reduced to $$, this behavior is described below
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#container-v1-core
and here is a discussion in an issue about this exact behavior https://github.com/kubernetes/kubernetes/issues/101137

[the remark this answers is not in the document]
devguyio: thanks! that was super useful !

rtheis:
@jared-hayes-dev please manually cherry-pick this fix to the `release-4.18` branch.

PR #5296, opened 2024-12-16
[release-4.17] OCPBUGS-46440: Allow ARM64 arch deployment on Agent platform

jparrill:
This a manual backport of https://github.com/openshift/hypershift/pull/5276

PR #5482, opened 2025-01-27
OCPBUGS-49340: Allow ARM64 arch deployment on Agent platform

jparrill:
Manual backport from https://github.com/openshift/hypershift/pull/5296


jparrill:
/close

Not needed in 4.16 because the issue was reported against MCE 2.7. Closing it

PR #5483, opened 2025-01-27
OCPBUGS-49370: openstack: update CAPI provider deployment spec

EmilienM:

`metrics-bind-addr` was deprecated and is currently removed, we should
stop using it so we can bump the version of CAPI and CAPO.


EmilienM:
CAPO started with no issue: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_hypershift/5483/pull-ci-openshift-hypershift-main-e2e-openstack-aws-conformance/1883900383454040064/artifacts/e2e-openstack-aws-conformance/dump/artifacts/namespaces/clusters-365e966cf37a03044f3f/core/pods/logs/capi-provider-699b99db9c-l7svz-manager.log

PR #5487, opened 2025-01-27
OCPBUGS-48532: Fix IsProgressing condition in HostedClusters

csrwng:
After #5168, the release image passed to the CVO is an image with a digest (which is needed for offline). Because this doesn't always match the release image specified in the HostedCluster.spec.release.image field, code that relies on comparing the image in CVO status with what's in the spec is now not matching. One effect is that the Progressing condition remains set to "true" even if a HostedCluster has completed provisioning.

This PR makes the following changes in order to fix it:
- Adds code to the HyperShift operator to compare either the image in spec as is or converted to a digest with the image reported by the CVO in status.
- Removes the call to obtain the release image for a specific architecture when obtaining an image with digest for the release.
- Removes code that uses fmt.Sprintf to obtain an image ref string from a docker reference in favor of ref.String() except in places where we can't rely on ref.String() to give us the expected reference.

Fixes #[OCPBUGS-48532](https://issues.redhat.com/browse/OCPBUGS-48532)

**Checklist**

### DIFF
--- a/support/util/imagemetadata_test.go
+++ b/support/util/imagemetadata_test.go
@@ -152,6 +152,7 @@ func TestGetManifest(t *testing.T) {
 		})
 	}
 }
+
 func TestGetDigest(t *testing.T) {
 	ctx := context.TODO()
 	pullSecret := []byte("{}")
@@ -166,7 +167,7 @@ func TestGetDigest(t *testing.T) {
 	}{
 		{
 			name:        "if failed to parse image reference",
-			imageRef:    "invalid-image-ref",
+			imageRef:    "::invalid-image-ref",
 			pullSecret:  pullSecret,
 			expectedErr: true,
 		},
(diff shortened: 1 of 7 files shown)


support/releaseinfo/registryclient/client.go
```diff
@@ -424,8 +423,12 @@ func GetCorrectArchImage(ctx context.Context, component string, imageRef string,
 	}
 
 	if isMultiArchImage {
-		operatingSystem := runtime.GOOS
-		arch := runtime.GOARCH
+		// operatingSystem := runtime.GOOS
+		// arch := runtime.GOARCH
+
+		operatingSystem := "linux"
+		arch := "amd64"
```
jparrill, line 430: Is that a reason why is this hardcoded?

support/util/imagemetadata_test.go
```diff
@@ -166,7 +167,7 @@ func TestGetDigest(t *testing.T) {
 	}{
 		{
 			name:        "if failed to parse image reference",
-			imageRef:    "invalid-image-ref",
+			imageRef:    "::invalid-image-ref",
```
jparrill, line 170: Is this in purpose?

jparrill (commented):
dropped some comments.

It would be great if we can validate this deploying a HC disconnected in our BM Lab.

Thanks!

support/util/imagemetadata_test.go
csrwng: Yes, because I changed how we obtain the image reference to use the ref.String() function, a name by itself is considered valid (before with the fmt.Sprintf() it was getting converted to something like: "//invalid-image-ref"). To still have an invalid ref, I added the extra "::"

jparrill: Gotcha, thanks

jparrill:
Tested on the BM laboratory with an IPv6/Disconnected scenario using a 4.16.12 OCP payload is working fine.

PR #5488, opened 2025-01-27
CNTRLPLANE-187: Remove MachineIdentityID from Azure HyperShift API

bryan-cox:
This PR removes MachineIdentityID from the Azure HyperShift API, CLI, and NodePool controller. This field is not needed for ARO HCP.

Fixes [CNTRLPLANE-187](https://issues.redhat.com/browse/CNTRLPLANE-187)

**Checklist**


enxebre:
we'll revaluate the best way to introduce support for this when needed.
/lgtm

PR #5495, opened 2025-01-28
OCPBUGS-48688: Add Tenant ID to Azure API & Remove Credentials from CPO

bryan-cox:
This PR:
- exposes a field for the tenant ID in the Azure API. This value was originally captured in the Credentials field of the Azure API, but that field will be removed in a future commit
- updates the CLI and the HyperShift Operator (HO) to use Tenant ID directly from the HC Azure API spec rather than the previous Credentials secret. References to the Credentials secret, in the CLI and HO, were also removed in this commit
- updates control plane operator (CPO) to use Tenant ID directly from the HC Azure API spec rather than the previous Credentials secret. References to the Credentials secret, CPO, were also removed in this commit

Fixes #

**Checklist**


bryan-cox:
I will have a follow up PR to remove Credentials from the API. I need this PR to go through and be backported to 4.18 before I can do that so that `ci/prow/e2e-aks-4-18` can pass

PR #5496, opened 2025-01-29
NO-JIRA: switch to resource.k8s.io/v1beta1 after kube 1.32 rebase

sjenning:
This fix is required to unblock the payload jobs

Replicates https://github.com/openshift/cluster-kube-apiserver-operator/pull/1777/commits/dea2e4fb8faafda391789ed662916854f15b8dff


bryan-cox:
/retitle NO-JIRA: switch to resource.k8s.io/v1beta1 after kube 1.32 rebase

PR #5505, opened 2025-01-29
OCPBUGS-49686: Add ObjectEncoding to the Azure API and SecretProviderClass Reconciliation in the CPO

bryan-cox:
This PR:
- adds ObjectEncoding to the Azure API so one can specify how a secret is encoded
- adds the objectEncoding to SecretProviderClass params so the right decoding option can be used on the Azure Key Vault secret

Fixes #

**Checklist**

### DIFF
--- a/api/hypershift/v1beta1/azure.go
+++ b/api/hypershift/v1beta1/azure.go
@@ -435,6 +435,10 @@ type AzurePlatformSpec struct {
 	TenantID string `json:"tenantID"`
 }
 
+// ObjectEncodingFormat is the type of encoding for an Azure Key Vault secret
+// +kubebuilder:validation:Enum:=utf-8;hex;base64
+type ObjectEncodingFormat string
+
 // ManagedAzureKeyVault is an Azure Key Vault on the management cluster.
 type ManagedAzureKeyVault struct {
 	// name is the name of the Azure Key Vault on the management cluster.
@@ -478,6 +482,20 @@ type ManagedIdentity struct {
 	//
 	// +kubebuilder:validation:Required
 	CertificateName string `json:"certificateName"`
+
+	// objectEncoding represents the encoding for the Azure Key Vault secret containing the certificate related to
+	// CertificateName. objectEncoding needs to match the encoding format used when the certificate was stored in the
+	// Azure Key Vault. If objectEncoding doesn't match the encoding format of the certificate, the certificate will
+	// unsuccessfully be read by the Secrets CSI driver and an error will occur. This error will only be visible on the
+	// SecretProviderClass custom resource related to the managed identity.
+	//
+	// The default value is utf-8.
+	//
+	// See this for more info - https://github.com/Azure/secrets-store-csi-driver-provider-azure/blob/master/website/content/en/getting-started/usage/_index.md
+	//
+	// +kubebuilder:validation:Enum:=utf-8;hex;base64
+	// +kubebuilder:default:="utf-8"
+	ObjectEncoding ObjectEncodingFormat `json:"objectEncoding"`
 }
 
 // ControlPlaneManagedIdentities contains the managed identities on the HCP control plane needing to authenticate with
--- a/control-plane-operator/controllers/hostedcontrolplane/secretproviderclass/secretproviderclass.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/secretproviderclass/secretproviderclass.go
@@ -14,6 +14,7 @@ const (
 array:
   - |
     objectName: %s
+    objectEncoding: %s
     objectType: secret
 `
 )
@@ -22,7 +23,7 @@ array:
 // Key Vault setup, and the certificate name it needs to pull from the Key Vault.
 //
 // https://learn.microsoft.com/en-us/azure/aks/csi-secrets-store-identity-access?tabs=azure-portal&pivots=access-with-a-user-assigned-managed-identity
-func ReconcileManagedAzureSecretProviderClass(secretProviderClass *secretsstorev1.SecretProviderClass, hcp *hyperv1.HostedControlPlane, certName string) {
+func ReconcileManagedAzureSecretProviderClass(secretProviderClass *secretsstorev1.SecretProviderClass, hcp *hyperv1.HostedControlPlane, certName, objectEncoding string) {
 	secretProviderClass.Spec = secretsstorev1.SecretProviderClassSpec{
 		Provider: "azure",
 		Parameters: map[string]string{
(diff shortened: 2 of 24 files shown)


control-plane-operator/controllers/hostedcontrolplane/secretproviderclass/secretproviderclass.go
```diff
@@ -34,6 +34,10 @@ func ReconcileManagedAzureSecretProviderClass(secretProviderClass *secretsstorev
 			"objects":                formatSecretProviderClassObject(certName),
 		},
 	}
+
+	if hcp.Spec.Platform.Azure.ObjectEncoding != "" {
+		secretProviderClass.Spec.Parameters["objectEncoding"] = string(hcp.Spec.Platform.Azure.ObjectEncoding)
```
machi1990, line 39: I think this has to go in https://github.com/openshift/hypershift/blob/b378274863c22f010a16b6cd7e0d31e46b058557/control-plane-operator/controllers/hostedcontrolplane/secretproviderclass/secretproviderclass.go#L16 

```
array:
  - |
    objectName: <the-name>
    objectType: secret
    objectEncoding: <the-encoding>
```
As par the types in https://github.com/Azure/secrets-store-csi-driver-provider-azure/blob/e40b93b7f34e6f8e5f0d278fe7ffa7ad52a69906/pkg/provider/types/types.go#L72

miguelsorianod: I think so too :thinking: , and it is applied per secret :thinking:

miguelsorianod: This also means that if it is added there everyone that calls this method will use that encodingformat, no matter where the code comes from. Is that the intention?

api/hypershift/v1beta1/azure.go
```diff
@@ -433,8 +433,20 @@ type AzurePlatformSpec struct {
...
 	// +required
 	TenantID string `json:"tenantID"`
+
+	// Represents the encoding for the Azure Key Vault secret
+	// Plumbs the objectEncoding for secrets store provider - https://github.com/Azure/secrets-store-csi-driver-provider-azure/blob/master/website/content/en/getting-started/usage/_index.md
+	//
+	// +kubebuilder:validation:Enum:=utf-8;hex;base64
+	// +kubebuilder:default:=utf-8
+	// +kubebuilder:validation:Optional
+	ObjectEncoding ObjectEncodingFormat `json:"objectEncoding"`
```
miguelsorianod, line 443: This is adding the attribute in `.platform.azure.objectEncoding`. If I understand this correctly, this means that conceptually this will apply to all keyvault secrets/certificates that are mounted using CSI driver, no matter where they come from.
Is this intentional? An alternative is in `.platform.azure.managedIdentities.controlPlane.managedIdentitiesKeyVault` ?

Take into account that other keyvault references are specified too in the spec (although I don't know if those end up being leveraged by CSI driver), like `.secretEncryption.kms.azure.activeKey`

api/hypershift/v1beta1/azure.go
```diff
@@ -433,8 +433,20 @@ type AzurePlatformSpec struct {
 	//
 	// +required
 	TenantID string `json:"tenantID"`
+
+	// Represents the encoding for the Azure Key Vault secret
```
enxebre, line 437: I anticipate the current docs will be difficult to grasp for an API consumer with little context:
- Please document the field starting with the serialized version of the field name i.e. objectEncoding so it's visible in that consumable way by godoc generated content.

- Please elaborate the intent of the field day 1 and also what happens if it's changed day 2. E.g "objectEncoding needs to match the format of the certificates that an API consumer has previously stored in Azure Key Vault. The value is passed down to the secretProviderClass managed by the controller to inject the certificates and let multiple components to authenticate against Azure. This includes, machine management, storage, Enumerate all components here..... If objectEncoding doesn't match the format used to store the content in key vault then (what happens)..... If you change this value day then (what happens)....."

- Feels pretty random to have this in the top level spec. Wouldn't .managedAzureKeyVault be a better place for it ti be?
This will make sense if we think this value will always apply to anything consumed from keyvault so we have a single input, otherwise we'd need to structure it in way that enables more flexibility (e.g. have it at azureResourceManagedIdentities.objectEncoding).

control-plane-operator/controllers/hostedcontrolplane/secretproviderclass/secretproviderclass.go
```diff
@@ -15,14 +15,15 @@ array:
   - |
     objectName: %s
     objectType: secret
+	objectEncoding: %s
```
machi1990, line 18: is the indentation okay here? Just want to be sure that this won't cause issue

bryan-cox: In my IDE, it's not showing up like that. I am having issues with testing this manually though.

control-plane-operator/controllers/hostedcontrolplane/secretproviderclass/secretproviderclass.go
```diff
@@ -31,15 +33,16 @@ func ReconcileManagedAzureSecretProviderClass(secretProviderClass *secretsstorev
...
 // formatSecretProviderClassObject places the certificate name in the appropriate string structure the
-// SecretProviderClass expects for an object. More details here:
+// SecretProviderClass expects for an object and specified the objectEncoding. More details here:
 // - https://learn.microsoft.com/en-us/azure/aks/csi-secrets-store-identity-access?tabs=azure-portal&pivots=access-with-a-user-assigned-managed-identity#configure-managed-identity
 // - https://secrets-store-csi-driver.sigs.k8s.io/concepts.html?highlight=object#custom-resource-definitions-crds
-func formatSecretProviderClassObject(certName string) string {
-	return fmt.Sprintf(objectFormat, certName)
+func formatSecretProviderClassObject(certName, objectEncoding string) string {
+	formattedString := fmt.Sprintf(objectFormat, certName, objectEncoding)
+	return strings.ReplaceAll(formattedString, "\t", "    ")
```
machi1990, line 47: this is no longer needed

control-plane-operator/controllers/hostedcontrolplane/secretproviderclass/secretproviderclass.go
```diff
@@ -22,7 +23,7 @@ array:
 // Key Vault setup, and the certificate name it needs to pull from the Key Vault.
 //
 // https://learn.microsoft.com/en-us/azure/aks/csi-secrets-store-identity-access?tabs=azure-portal&pivots=access-with-a-user-assigned-managed-identity
-func ReconcileManagedAzureSecretProviderClass(secretProviderClass *secretsstorev1.SecretProviderClass, hcp *hyperv1.HostedControlPlane, certName string) {
+func ReconcileManagedAzureSecretProviderClass(secretProviderClass *secretsstorev1.SecretProviderClass, hcp *hyperv1.HostedControlPlane, certName, objectEncoding string) {
```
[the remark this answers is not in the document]
bryan-cox: I would do that, but we are changing function again in a couple of weeks for the MI phase 3 work.

api/hypershift/v1beta1/azure.go
```diff
@@ -478,6 +482,17 @@ type ManagedIdentity struct {
 	//
 	// +kubebuilder:validation:Required
 	CertificateName string `json:"certificateName"`
+
+	// objectEncoding represents the encoding for the Azure Key Vault secret containing the certificate related to
+	// CertificateName. objectEncoding needs to match the encoding format used when the certificate was stored in the
+	// Azure Key Vault. If objectEncoding doesn't match the encoding format of the certificate, the certificate will
+	// unsuccessfully be read by the Secrets CSI driver and an error will occur.
```
enxebre, line 489: can we document what's the default?
will this error be signaled back in any way to the api consumer?

bryan-cox: Updated

bryan-cox: Also this API change is only for M2. This will not be needed for M3 and will be removed when we start the MI phase 3 work.

PR #5510, opened 2025-01-30
OCPBUGS-48688: Remove Credentials from the Azure API

bryan-cox:
This PR:
- removes the Credentials field from the Azure API since it is no longer needed. Originally, the client secret info, subscription id, and tenant id were stored in this secret reference. However, client secret implementation has been removed as a result of workload identity work, and subscription id and tenant id are directly exposed in the Azure API now.
- updates the CLI and the HyperShift Operator (HO) to use Tenant ID directly from the HC Azure API spec rather than the previous Credentials secret. References to the Credentials secret, in the CLI and HO, were also removed in this commit.

Fixes [OCPBUGS-48688](https://issues.redhat.com/browse/OCPBUGS-48688)

**Checklist**


miguelsorianod:
lgtm but I'll defer approval/reviews to other Hypershift team members

PR #5515, opened 2025-01-30
NO-ISSUE: docs/content/how-to/disaster-recovery/etcd-recovery: Explicitly delegate to admin

wking:

The outgoing line left the responsibility for the deletion ambiguous. Maybe the HyperShift operator had automated that deletion?  Maybe the admin had to manually delete, and the automation was just scoped to post-deletion recovery?  My incoming line is trying to make it clear that it's the latter, and that HyperShift is not involved in the Pod/PVC deletion that initiates recovery, even when `--enable-etcd-recovery` is true.

**Checklist**


bryan-cox:
@jparrill - you should take a look

PR #5522, opened 2025-01-31
HOSTEDCP-2257: Vendor karpenter CRDs

muraee:

Fixes #

**Checklist**


elmiko:
we now have the karpenter repos mirrored into openshift org, do we want to use those for importing?

or perhaps add as a followup to this as we are still configuring the local repos.

muraee:
@elmiko let's do that in a followup, will also need to change the CRDs we install in the guest cluster.

PR #5525, opened 2025-01-31
OCPBUGS-48619: Add HostedCluster additional trustbundles to konnectivity-https-proxy

csrwng:
When using the konnectivity-https-proxy sidecar to communicate through workers, the proxy dials any configured proxy directly and fails to connect if it cannot trust the proxy's serving certificate. This commit adds mounts for the trusted-ca-bundle-managed configmap to sidecar containers that run the konnectivity-https-proxy. The configmap contains the aggregate of .spec.additionalTrustBundle and

This fixes the issue of the ingress operator not becoming available when using a proxy with a custom certificate.

In addition it propagates any proxy environment variables that are set on the control plane operator to the konnectivity-https-proxy containers. This is needed when the management cluster itself requires a proxy and the proxy container dials an address directly such as when it needs to communicate with a cloud provider endpoint.

Fixes #[OCPBUGS-48619](https://issues.redhat.com/browse/OCPBUGS-48619)

**Checklist**

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/ingressoperator/ingressoperator.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/ingressoperator/ingressoperator.go
@@ -2,12 +2,14 @@ package ingressoperator
 
 import (
 	"fmt"
+	"path"
 
 	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
 	"github.com/openshift/hypershift/control-plane-operator/controllers/hostedcontrolplane/imageprovider"
 	"github.com/openshift/hypershift/control-plane-operator/controllers/hostedcontrolplane/kas"
 	"github.com/openshift/hypershift/control-plane-operator/controllers/hostedcontrolplane/manifests"
 	"github.com/openshift/hypershift/support/azureutil"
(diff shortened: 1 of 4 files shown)


muraee:
can you please also update v2 code here: https://github.com/openshift/hypershift/blob/main/support/controlplane-component/konnectivity-container.go

This part https://github.com/openshift/hypershift/blob/main/support/controlplane-component/konnectivity-container.go#L87-L105  can be removed then if we are mounting `trusted-ca-bundle-managed` configMap, right?

devguyio:
@csrwng can you please describe how can this PR be tested?

csrwng:
> can you please describe how can this PR be tested?

Basically by creating a HostedCluster that uses a secure proxy. Without this fix, the cluster fails to complete because the ingress operator never reports as available.

csrwng:
@muraee I believe we're already doing the right thing in the v2 code. For the existing non-v2 code I opted to mount the trusted-ca-bundle-managed configmap for convenience (just one thing to mount). But in v2, if we're already mounting both the .spec.additionalTrustBundle and the .spec.configuration.proxy.trustCA, then I think we're covered.

PR #5526, opened 2025-01-31
HOSTEDCP-2259: CLI: enable secure proxy creation

csrwng:
Adds '--enable-secure-proxy' flag to 'hypershift create infra aws' and 'hypershift create cluster aws' commands.

The flag is similar to the existing '--enable-proxy' flag, except that it creates a proxy that exposes a ssl endpoint with custom serving certificate.

This is needed to automate testing of secure proxy use cases.

Fixes #[HOSTEDCP-2259](https://issues.redhat.com/browse/HOSTEDCP-2259)

**Checklist**

### DIFF
--- a/cmd/cluster/aws/create.go
+++ b/cmd/cluster/aws/create.go
@@ -2,7 +2,9 @@ package aws
 
 import (
 	"context"
+	"encoding/base64"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 
@@ -42,6 +44,7 @@ type RawCreateOptions struct {
 	Zones                        []string
 	EtcdKMSKeyARN                string
 	EnableProxy                  bool
+	EnableSecureProxy            bool
 	ProxyVPCEndpointServiceName  string
 	SingleNATGateway             bool
 	MultiArch                    bool
@@ -67,6 +70,10 @@ func (o *RawCreateOptions) Validate(ctx context.Context, opts *core.CreateOption
 		return nil, err
 	}
 
+	if o.EnableProxy && o.EnableSecureProxy {
+		return nil, errors.New("use --enable-proxy or --enable-secure-proxy but not both")
+	}
+
 	return &ValidatedCreateOptions{
 		validatedCreateOptions: &validatedCreateOptions{
 			RawCreateOptions: o,
@@ -82,6 +89,7 @@ type completedCreateOptions struct {
 	iamInfo           *awsinfra.CreateIAMOutput
 	arch              string
 	externalDNSDomain string
+	namespace         string
 }
 
 type CreateOptions struct {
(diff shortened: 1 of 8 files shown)


cmd/infra/aws/create.go
```diff
@@ -596,13 +672,60 @@ yum install -y squid
...
 systemctl enable --now squid
+
 mkdir -p /home/ec2-user/.ssh
 chmod 0700 /home/ec2-user/.ssh
 echo -e '%s' >/home/ec2-user/.ssh/authorized_keys
 chmod 0600 /home/ec2-user/.ssh/authorized_keys
 chown -R ec2-user:ec2-user /home/ec2-user/.ssh
 `
 
+const secureProxyConfigurationScript = `#!/bin/bash
```
devguyio, line 683: maybe we can extract these two script constants and the generation function in a separate file for better visibility given how significant they are?

cmd/infra/aws/proxyca.go
```diff
@@ -0,0 +1,44 @@
...
+	"golang.org/x/crypto/ssh"
+)
+
+func fetchProxyCA(publicAddr, privateAddr string, privateKey []byte) (string, error) {
+	signer, err := ssh.ParsePrivateKey(privateKey)
+	if err != nil {
+		log.Fatalf("unable to parse private key: %v", err)
+	}
+
+	config := &ssh.ClientConfig{
```
devguyio, line 17: same comment about separating the naming and creation logic of CRs in util function

devguyio (changes requested):
Awesome work! I left a few commens and I think it'd be great if we have tests and docs update. Were you planning on having that in separate PRs? or do you wanna add them to this one?

cmd/cluster/aws/create.go
```diff
@@ -42,6 +45,7 @@ type RawCreateOptions struct {
 	Zones                        []string
 	EtcdKMSKeyARN                string
 	EnableProxy                  bool
+	EnableSecureProxy            bool
```
devguyio, line 48: something I just noticed, this option is missing a matching flag as part of `bindCoreOptions` function 
https://github.com/openshift/hypershift/blob/168c2a57e7fc89dad00e8127b279bcc162b2c03e/cmd/cluster/aws/create.go#L446

similar to `EnableProxy` option

https://github.com/openshift/hypershift/blob/168c2a57e7fc89dad00e8127b279bcc162b2c03e/cmd/cluster/aws/create.go#L457

csrwng:
> do we need to keep --enable-proxy? Can we drop it in favour of this?

We could drop it. Eventually. 
Right now we run a proxy test that uses the old flag. Switching to a secure proxy would only pass if we get https://github.com/openshift/hypershift/pull/5525 merged. And because we use the main branch for all e2e's, it would mean that we can't switch until we get that PR backported to all supported branches.

csrwng:
Thank you for the review @devguyio. I have addressed your comments.

devguyio:
Thanks! 

/lgtm

PR #5540, opened 2025-02-03
PODAUTO-318: Implement HCP karpenter deletion

maxcao13:
This commit implements supporting cascading deletion of karpenter provisioned nodes upon deletion of a hosted control plane. Also modifies the current Karpenter e2e test to test deletion of an HCP is not blocked by karpenter nodes, and adds a unit test.

Fixes https://issues.redhat.com/browse/PODAUTO-318

**Checklist**

### DIFF
--- a/karpenter-operator/controllers/karpenter/karpenter_controller.go
+++ b/karpenter-operator/controllers/karpenter/karpenter_controller.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"sort"
+	"time"
 
 	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
 	"github.com/openshift/hypershift/karpenter-operator/controllers/karpenter/assets"
@@ -30,7 +31,9 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/source"
+	karpenterv1 "sigs.k8s.io/karpenter/pkg/apis/v1"
 )
 
 const (
@@ -102,6 +105,21 @@ func (r *Reconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager, man
 		return fmt.Errorf("failed to watch Deployment: %w", err)
 	}
 
+	namespacedPredicates := predicate.NewPredicateFuncs(func(object client.Object) bool {
+		return object.GetNamespace() == r.Namespace
+	})
+	// Watch the HCP management side.
+	if err := c.Watch(source.Kind[client.Object](managementCluster.GetCache(), &hyperv1.HostedControlPlane{}, handler.EnqueueRequestsFromMapFunc(
+		func(ctx context.Context, o client.Object) []ctrl.Request {
+			if o.GetNamespace() != r.Namespace {
+				return nil
+			}
+			return []ctrl.Request{{NamespacedName: client.ObjectKeyFromObject(o)}}
+		},
+	), namespacedPredicates)); err != nil {
+		return fmt.Errorf("failed to watch HostedControlPlane: %w", err)
+	}
+
 	// Trigger initial sync.
 	initialSync := make(chan event.GenericEvent)
 	if err := c.Watch(source.Channel(initialSync, &handler.EnqueueRequestForObject{})); err != nil {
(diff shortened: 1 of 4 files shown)


karpenter-operator/controllers/karpenter/karpenter_controller.go
```diff
@@ -128,13 +128,35 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+				// If we still get the NodePool, but it's already marked as terminating, we don't need to call Delete again
+				if !nodePool.GetDeletionTimestamp().IsZero() {
+					continue
+				}
+				if err := r.GuestClient.Delete(ctx, &nodePool); err != nil {
+					return ctrl.Result{}, fmt.Errorf("failed to delete NodePool: %w", err)
+				}
+			}
+			// Wait until all NodePools are deleted before removing the finalizer
+			if len(nodePoolList.Items) > 0 {
```
enxebre, line 150: This will cause the nodePools to immediately be removed from etcd and signal background deletion for the nodeClaims, so we also need to wait for all nodeClaims to go away, agree?

maxcao13: Ah right, makes sense, I was assuming the default behaviour was foreground deletion. Would it make sense to just set foreground deletion specifically here?

maxcao13:
/cc joelsmith elmiko mtulio
If you guys don't mind taking a look as well ^^

karpenter-operator/controllers/karpenter/karpenter_controller.go
enxebre: Lets codify nodeClaim wait for deletion explicitly so we ensure our criteria for infra removal can't be broken by a change/bug in the karpenter ownership chain. If we want to signal the request as foreground additional that seems fine as well.

karpenter-operator/controllers/karpenter/karpenter_controller.go
```diff
@@ -128,13 +129,48 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+				}
+				if err := r.GuestClient.Delete(ctx, &nodePool, &client.DeleteOptions{
+					PropagationPolicy: ptr.To(metav1.DeletePropagationForeground),
+				}); err != nil {
+					return ctrl.Result{}, fmt.Errorf("failed to delete NodePool: %w", err)
+				}
+			}
+			// Wait until all NodePools are deleted before removing the finalizer
+			if len(nodePoolList.Items) > 0 {
+				return ctrl.Result{}, nil
```
enxebre, line 154: we should requeue here and also add a watcher for the HCP resource in SetupWithManager

karpenter-operator/controllers/karpenter/karpenter_controller_test.go
```diff
@@ -227,3 +231,202 @@ func TestGetUserDataSecret(t *testing.T) {
...
+			}
+
+			ctx := log.IntoContext(context.Background(), testr.New(t))
+
+			_, err := r.Reconcile(ctx, ctrl.Request{NamespacedName: client.ObjectKeyFromObject(tc.hcp)})
+
+			g.Expect(err).NotTo(HaveOccurred())
+
+			// reconcile "enough" times and check if karpenter finalizer on the hcp still exists
+			numReconciles := 3
```
enxebre, line 410: Nice test! For the unit test scenario, shouldn't this just need one single reconcile?

muraee:
should we get this https://github.com/openshift/hypershift/pull/5522 merged first?

elmiko (commented):
this is looking nice, thanks @maxcao13 !

i have a question, but it's not a blocker.

maxcao13:
| should we get this https://github.com/openshift/hypershift/pull/5522 merged first?
Probably makes sense for this to wait for that to be merged (whether we import the upstream or mirrored deps) since it's easier for me to resolve the TODOs this way.

karpenter-operator/controllers/karpenter/karpenter_controller_test.go
maxcao13: Yes you are right, fixed, thanks!

enxebre:
ok let's ship https://github.com/openshift/hypershift/pull/5522 then rebased and proceed here

muraee:
you can register it in `main.go` just like the NodeClaim:
https://github.com/openshift/hypershift/blob/main/karpenter-operator/main.go#L82-L85

enxebre:
/approve
ptal @muraee

karpenter-operator/controllers/karpenter/karpenter_controller.go
```diff
@@ -126,13 +144,43 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		return ctrl.Result{}, err
 	}
 	if hcp.DeletionTimestamp != nil {
-		// TODO(alberto): implement deletion. E.g. loop over nodeClaims delete them, wait and delete karpeneter deployment.
 		if controllerutil.ContainsFinalizer(hcp, karpenterFinalizer) {
+			nodePoolList := &karpenterv1.NodePoolList{}
+			if err := r.GuestClient.List(ctx, nodePoolList, client.InNamespace(r.Namespace)); err != nil {
```
muraee, line 149: why are we using `client.InNamespace(r.Namespace)` here? this will list nodePools in the HCP namespace which doesn't exist in the guest cluster.

karpenter-operator/controllers/karpenter/karpenter_controller.go
```diff
@@ -126,13 +144,43 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+				}
+			}
+			// Wait until all NodePools are deleted before removing the finalizer
+			if len(nodePoolList.Items) > 0 {
+				return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
+			}
+
+			nodeClaimList := &karpenterv1.NodeClaimList{}
+			// Make sure all NodeClaims are actually gone
+			if err := r.GuestClient.List(ctx, nodeClaimList, client.InNamespace(r.Namespace)); err != nil {
```
muraee, line 171: same comment as above

karpenter-operator/controllers/karpenter/karpenter_controller.go
maxcao13: My bad, thanks for catching that.

elmiko (commented):
thanks @maxcao13 , lgtm

PR #5541, opened 2025-02-03
CNTRLPLANE-172: cli: azure: allow assigning custom HCP roles

sjenning:
A follow on to https://github.com/openshift/hypershift/pull/5469

We don't yet have built-in roles for the HCP components and that process might take a while.

In the meantime, we can exercise the limited permission set in custom roles that are manually created by the user before cluster creation.

The role names are
* `Azure Red Hat OpenShift Control Plane Operator Role` for the CPO
* `Azure Red Hat OpenShift NodePool Management Role` for CAPZ i.e. nodePoolManagement

Once these roles are built-in, we can safely assume the roles always exist and remove this flag.

### DIFF
--- a/cmd/infra/azure/create.go
+++ b/cmd/infra/azure/create.go
@@ -70,6 +70,7 @@ type CreateInfraOptions struct {
 	DataPlaneIdentitiesFile     string
 	AssignServicePrincipalRoles bool
 	DNSZoneRG                   string
+	AssignCustomHCPRoles        bool
 }
 
 type CreateInfraOutput struct {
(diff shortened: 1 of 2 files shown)


cmd/infra/azure/create.go
```diff
@@ -849,6 +853,9 @@ func assignServicePrincipalRoles(subscriptionID, managedResourceGroupName, nsgRe
 		role = "Azure Red Hat OpenShift Image Registry Operator Role"
 	case nodePoolMgmt:
 		scopes = append(scopes, vnetRG)
+		if assignCustomHCPRoles {
+			role = "Azure Red Hat OpenShift Cluster API Role"
```
bryan-cox, line 857: Wondering if this should be Azure Red Hat OpenShift NodePool Management Role to keep it agnostic from CAPI/CAPZ? That would be in line with what we do for AWS as well - https://github.com/openshift/hypershift/blob/103d1486a1326152865fc7dfb1e91d2cc147ef18/api/hypershift/v1beta1/aws.go#L508.

ref - https://github.com/openshift/hypershift/pull/4811#discussion_r1800604551

PR #5551, opened 2025-02-04
OCPBUGS-49825: use registryOverrides when automaitcally retrieving catalog images for hosted control plane

Patryk-Stefanski:

- We should use the registry overrides if provided when automatically retrieving the latest available catalog images, this can cause trouble when the pullsecret doesn't give access to the default redhat registry expected ie 'registry.redhat.io/redhat'

Fixes #

**Checklist**

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
@@ -10,7 +10,6 @@ import (
 	"os"
 	"sort"
 	"strings"
-	"sync"
 	"time"
 
 	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
@@ -99,7 +98,6 @@ import (
 	"github.com/openshift/hypershift/support/metrics"
 	"github.com/openshift/hypershift/support/proxy"
 	"github.com/openshift/hypershift/support/releaseinfo"
-	"github.com/openshift/hypershift/support/releaseinfo/registryclient"
 	"github.com/openshift/hypershift/support/secretproviderclass"
 	"github.com/openshift/hypershift/support/thirdparty/library-go/pkg/image/reference"
 	"github.com/openshift/hypershift/support/upsert"
@@ -165,8 +163,7 @@ const (
 )
 
 var (
-	olmCatalogImagesOnce sync.Once
-	catalogImages        map[string]string
+	catalogImages map[string]string
 )
 
 type HostedControlPlaneReconciler struct {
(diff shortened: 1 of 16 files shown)


control-plane-operator/controllers/hostedcontrolplane/olm/catalogs.go
```diff
@@ -135,17 +135,38 @@ func GetCatalogImages(ctx context.Context, hcp hyperv1.HostedControlPlane, pullS
...
 	//check catalogs of last 4 supported version in case new version is not available
 	supportedVersions := 4
+	imageRegistry := ""
 	for i := 0; i < supportedVersions; i++ {
-		_, err = digestLister(ctx, fmt.Sprintf("registry.redhat.io/redhat/certified-operator-index:v%d.%d", version.Major, version.Minor), pullSecret)
-		if err == nil {
-			break
+		for _, registry := range registries {
+			testImage := fmt.Sprintf("%s/certified-operator-index:v%d.%d", registry, version.Major, version.Minor)
+
```
enxebre, line 156: should this use imageMetaDataProvider to pick the right mirrors in the case no annotations are pased through? @jparrill

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -3980,7 +3979,7 @@ func (r *HostedControlPlaneReconciler) reconcileOperatorLifecycleManager(ctx con
 
 			var getCatalogImagesErr error
 			olmCatalogImagesOnce.Do(func() {
-				catalogImages, err = olm.GetCatalogImages(ctx, *hcp, pullSecret.Data[corev1.DockerConfigJsonKey], registryclient.GetListDigest, r.ImageMetadataProvider)
+				catalogImages, err = olm.GetCatalogImages(ctx, *hcp, pullSecret.Data[corev1.DockerConfigJsonKey], r.ImageMetadataProvider.GetDigest, r.ImageMetadataProvider, isImageRegistryOverrides)
```
muraee, line 3982: we are already passing `ImageMetadataProvider`, why do we need to pass `ImageMetadataProvider.GetDigest` explicitly? I think we can drop it.

Patryk-Stefanski: It so that we can mock it for the test, https://github.com/openshift/hypershift/pull/5551/files#diff-5846297ea35ea4e63b978b59b6773f58ff8d63668d14648f764ec523b92a5554R149

muraee: that test is in the HCCO and doesn't seem to be calling `olm.GetCatalogImages` func. Even if it does you also pass a mock imageMetaDataProvider there, which will container the mock `GetDigest` func

Patryk-Stefanski: yeah it tests the reconcileOLM which uses NewOperatorLifecycleManagerParams which uses the olm.getCatalogImages.

And the way the fakeImageMetaDataProvider is designed still relies on a image being available in a repo so I couldn't mock the image in a registry for the getRepoSetup() func which causes this unit test to fail when using the fakeImageMetaDataProvider.

muraee: you could just create a new fakeImageMetaDataProvider and use it for that test, no need to reuse the existing one.

muraee: this code is already confusing with too many things passed around from different places.

Patryk-Stefanski: I created another fakeRegistryClientProvider to mock the functions for the HCCO where its needed in the olm reconcile.

muraee: great, thanks. This is not needed now right? https://github.com/openshift/hypershift/pull/5551/files#diff-2cc6d15e7cd4dc8dbc5d56444fc81df070bc6749714b11e8b6646a50d79003a0R389
```
type GetDigestFN = func(ctx context.Context, imageRef string, pullSecret []byte) (digest.Digest, *reference.DockerImageReference, error)
```

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -261,7 +261,7 @@ func (r *HostedControlPlaneReconciler) registerComponents() {
 		snapshotcontrollerv2.NewComponent(),
 	)
 	r.components = append(r.components,
-		olmv2.NewComponents(r.ManagementClusterCapabilities.Has(capabilities.CapabilityImageStream))...,
+		olmv2.NewComponents(r.ImageMetadataProvider, r.ManagementClusterCapabilities.Has(capabilities.CapabilityImageStream))...,
```
muraee, line 264: ImageMetadataProvider is already passed down in the cpContext, no need to pass it here.

control-plane-operator/controllers/hostedcontrolplane/olm/catalogs.go
```diff
@@ -154,21 +174,21 @@ func GetCatalogImages(ctx context.Context, hcp hyperv1.HostedControlPlane, pullS
...
 }
 
-func ReconcileCatalogsImageStream(imageStream *imagev1.ImageStream, ownerRef config.OwnerRef, isImageRegistryOverrides map[string][]string, catalogImages map[string]string) error {
+func ReconcileCatalogsImageStream(imageStream *imagev1.ImageStream, ownerRef config.OwnerRef, catalogImages map[string]string) error {
 	imageStream.Spec.LookupPolicy.Local = true
 	if imageStream.Spec.Tags == nil {
 		imageStream.Spec.Tags = []imagev1.TagReference{}
 	}
-	for name, image := range getCatalogToImageWithISImageRegistryOverrides(catalogImages, isImageRegistryOverrides) {
+	for name, image := range catalogImages {
```
muraee, line 191: I see you didn't made add this change in v2, any reason why not?
https://github.com/openshift/hypershift/pull/5551/files#diff-9af8dd67b1580cb012d9148e01dc409ca6d4a97aa600c4fe7d8cf06560dcbd2bL44-L45

PR #5554, opened 2025-02-04
ARO-14442: introduce capabilities api

flavianmissi:
We're introducing the concept of capabilities to Hypershift on a limited basis, as described in https://github.com/openshift/enhancements/pull/1729.

**Which issue(s) this PR fixes** *:
Fixes [ARO-14442](https://issues.redhat.com//browse/ARO-14442)

**Checklist**


api/hypershift/v1beta1/hosted_controlplane.go
```diff
@@ -200,6 +200,11 @@ type HostedControlPlaneSpec struct {
 	// +kubebuilder:validation:MaxProperties=20
 	// +optional
 	Labels map[string]string `json:"labels,omitempty"`
+
+	// capabilities allows for disabling optional components at cluster install time.
+	// This field is optional and once set cannot be changed.
+	// +optional
+	Capabilities Capabilities `json:"capabilities,omitempty"`
```
enxebre, line 207: should we featuregate this instead of the inner field?
Let's also add soime validation tests https://github.com/openshift/hypershift/blob/607f618b03351a2313a24261b76e2f580f371ca3/test/e2e/create_cluster_test.go#L27

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -353,6 +353,36 @@ const (
...
+	// will effectively enable the capability on the cluster, causing the
+	// referred component to be installed.
+	//
+	// It is not possible to add capabilities to disabledCapabilities. In
+	// other words, it's not possible to disable a capability at runtime.
+	//
+	// +listType=atomic
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="disabledCapabilities is immutable"
+	// +optional
+	// +openshift:enable:FeatureGate=DisableClusterCapabilities
```
enxebre, line 382: should we add min/max?

flavianmissi: Right, thanks for reminding me about this. For now we only have one valid value: `ImageRegistry`, so should the min/max both be 13 (len(ImageRegistry))? Or are there some sensible defaults we can use instead?

enxebre:
ptal @sjenning

api/hypershift/v1beta1/hosted_controlplane.go
flavianmissi: We decided on slack that this can't be done at this point because the field is behind a tech preview feature gate, so we'll defer the tests for now.

flavianmissi:
I don't see how the test failures are connected to the changes 🤔 

/retest

flavianmissi:
from latest e2e-aws failure:

```
eventually.go:223: observed *v1beta1.HostedCluster e2e-clusters-rjbch/example-9xpt7 invalid at RV 46558 after 10m0s: incorrect condition: wanted AWSEndpointAvailable=True, got AWSEndpointAvailable=False: AWSError(failed to create vpc endpoint: RequestLimitExceeded)
```
unsure if any action is needed. It'll probably clear up in the next few hours.

sjenning:
This has flaked a few times in a row.   I'm confident is passes.

/override ci/prow/e2e-aws

PR #5555, opened 2025-02-04
WIP - KubeAPIServerVirtualIP

EmilienM:
/hold

### DIFF
--- a/api/hypershift/v1beta1/openstack.go
+++ b/api/hypershift/v1beta1/openstack.go
@@ -136,6 +136,13 @@ type OpenStackPlatformSpec struct {
 	// +kubebuilder:validation:XValidation:rule="isIP(self)",message="floatingIP must be a valid IPv4 or IPv6 address"
 	// +optional
 	IngressFloatingIP string `json:"ingressFloatingIP,omitempty"`
+
+	// KubeAPIServerVirtualIP is an available floating IP in your OpenStack cluster that will
+	// be associated with the kubeapi-server load balancer port.
+	//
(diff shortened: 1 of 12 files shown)


api/hypershift/v1beta1/openstack.go
```diff
@@ -136,6 +136,13 @@ type OpenStackPlatformSpec struct {
 	// +kubebuilder:validation:XValidation:rule="isIP(self)",message="floatingIP must be a valid IPv4 or IPv6 address"
 	// +optional
 	IngressFloatingIP string `json:"ingressFloatingIP,omitempty"`
+
+	// KubeAPIServerVirtualIP is an available floating IP in your OpenStack cluster that will
+	// be associated with the kubeapi-server load balancer port.
+	//
+	// +kubebuilder:validation:XValidation:rule="isIP(self)",message="kubeAPIServerVirtualIP must be a valid IPv4 or IPv6 address"
+	// +optional
+	KubeAPIServerVirtualIP string `json:"kubeAPIServerVirtualIP,omitempty"`
```
EmilienM, line 145: Brain dump idea: create a type that has an IP and a hostname as well.
The hostname would be used to update `HostedCluster.Status.ControlPlaneEndpoint` (if it's possible) or at least generate a kubeconfig using that hostname (a hostname is prettier than an IP).
And I think we would need to create the Service annotation to give the hostname, so the certificate would be valid.

EmilienM:
Alternative prototype here: https://github.com/openshift/hypershift/pull/5718

PR #5560, opened 2025-02-05
NO-JIRA: Update cpov2 docs

muraee:

Fixes #

**Checklist**

### DIFF
--- a/support/controlplane-component/README.md
+++ b/support/controlplane-component/README.md
@@ -4,7 +4,7 @@
 
 Every component needs to create a directory under `/control-plane-operator/controllers/hostedcontrolplane/v2/assets` to host its manifests, the name of the directory is your component's name
 ```shell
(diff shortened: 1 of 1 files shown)


sjenning:
should we not move this to `docs` so that e2e's don't run when we update these docs?

muraee:
@sjenning I didn't put it in docs, so it won't be visible on the site just yet, since we haven't fully finished the implementation.
shouldn't the e2e be skipped for any `.md` file change instead of just looking at `docs` dir?

sjenning:
Our regex in the ci-operator config only skip `.md` files in the root of the repo
```
skip_if_only_changed: (^(\.tekton|docs|examples|enhancements|contrib)/)|(^[A-Z]+\.md$)|((^|/)OWNERS$)|(/overrides\.yaml$)
```

PR #5359, opened 2025-01-08
OCPBUGS-45600: Enforce privileged PSA by default

Evan-Reilly:
Align HyperShift with OpenShift's expectation that PSA is enforcing privileged by default instead of restricted.

**Which issue(s) this PR fixes** *(optional, use `fixes #<issue_number>(, fixes #<issue_number>, ...)` format, where issue_number might be a GitHub issue, or a Jira story*:
Fixes #

https://issues.redhat.com/browse/OCPBUGS-45600

**Checklist**
- [x] Subject and description added to both, commit and PR.
- [ ] Relevant issues have been referenced.
- [ ] This change includes docs. 
- [ ] This change includes unit tests.

PR #5566, opened 2025-02-05
OCPBUGS-45600: enforce privileged PSA by default

sjenning:
supersedes https://github.com/openshift/hypershift/pull/5359


rtheis (approved):
/lgtm
/approved

PR #5569, opened 2025-02-05
OTA-1405: Add the ClusterVersionOperator API

DavidHurta:
**Add the ClusterVersionOperator API**:

Add the feature-gated ClusterVersionOperator API for HyperShift, which is described in more detail in the [CVO Log Level API enhancement](https://github.com/openshift/enhancements/blob/master/enhancements/update/cvo-log-level-api.md#hypershift).

This PR only introduces the API changes. The API will be implemented as part of the [OTA-1406](https://issues.redhat.com/browse/OTA-1406) issue. This is to break down the problem into smaller PRs. In case it is more desired to couple the API changes together with the implementation, please let me know.

Fixes #[OTA-1405](https://issues.redhat.com/browse/OTA-1405)

**Checklist**


api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -546,6 +546,12 @@ type HostedClusterSpec struct {
 	// +optional
 	Configuration *ClusterConfiguration `json:"configuration,omitempty"`
 
+	// OperatorConfiguration specifies configuration for individual OCP operators in the cluster.
```
enxebre, line 549: please start doc with lower case so generate gdoc uses the serialized version of the field i.e. operatorConfiguration.
Same for the inner field.
FWIW we'll add linter for this.

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1595,6 +1601,14 @@ type ClusterConfiguration struct {
 	Proxy *configv1.ProxySpec `json:"proxy,omitempty"`
 }
 
+// OperatorConfiguration specifies configuration for individual OCP operators in the cluster.
+type OperatorConfiguration struct {
+	// ClusterVersionOperator specifies the configuration for the Cluster Version Operator in the hosted cluster.
+	//
+	// +optional
+	ClusterVersionOperator ClusterVersionOperatorSpec `json:"clusterVersionOperator,omitempty"`
```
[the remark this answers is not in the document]
DavidHurta: I have trouble remembering. I went through the comments on the enhancement, I thought about the decision, and I have refreshed my memory on the [practices regarding optional and required tags](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required). However, it seems that the non-pointer reference is unintentional. I have changed the field to be a pointer type since the field is optional, and there is no specific reason to make it a non-pointer type. Thank you for noticing.

api/hypershift/v1beta1/hostedcluster_types.go
DavidHurta: Thanks, the doc comments for the `operatorConfiguration` and `clusterVersionOperator` fields have been updated!

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -570,6 +570,12 @@ type HostedClusterSpec struct {
 	// +optional
 	Configuration *ClusterConfiguration `json:"configuration,omitempty"`
 
+	// operatorConfiguration specifies configuration for individual OCP operators in the cluster.
+	//
+	// +optional
+	// +openshift:enable:FeatureGate=ClusterVersionOperatorConfiguration
```
enxebre, line 576: please include this gate within api/hypershift/v1beta1/featuregates/ yaml files before running make update

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -570,6 +570,12 @@ type HostedClusterSpec struct {
...
+	OperatorConfiguration *OperatorConfiguration `json:"operatorConfiguration,omitempty"`
```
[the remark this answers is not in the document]
DavidHurta: Yes! The relevant https://github.com/openshift/enhancements/blob/master/enhancements/update/cvo-log-level-api.md enhancement proposes that as well. I have forgotten to add that, sorry about that.

Now included.

Do you prefer to merge the propagation in this PR? My initial goal was only to introduce the API changes, and support the API in https://issues.redhat.com/browse/OTA-1406. As part of an iterative approach and because the support itself may take some additional effort for a developer, which is not that active in the HyperShift repository.

api/hypershift/v1beta1/operator.go
```diff
@@ -0,0 +1,35 @@
...
+
+	// TraceAll is used when something is broken at the level of API content/decoding. It will dump complete body
+	// content. If you turn this on in a production cluster prepare from serious performance issues and massive amounts
+	// of logs. In kube, this is probably glog=8.
+	TraceAll LogLevel = "TraceAll"
+)
+
+// ClusterVersionOperatorSpec is the specification of the desired behavior of the Cluster Version Operator.
+type ClusterVersionOperatorSpec struct {
+	// operatorLogLevel is an intent based logging for the operator itself. It does not give fine-grained control,
```
enxebre, line 27: it'd be nice to explicitly specify what happens if this  is changed day 2

PR #5578, opened 2025-02-06
NO-JIRA: add optional step for creating release image using cluster b…

Patryk-Stefanski:
…ot and multiple prs

Fixes #

**Checklist**

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -35,17 +35,42 @@ What do I need to do to test...
 
         $ bin/hypershift-operator run
 
+### Building Custom Images
+To build images that can be both used for HyperShift installation and Hosted Cluster creation
+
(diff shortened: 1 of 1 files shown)


HACKING.md
```diff
@@ -57,6 +57,20 @@ What do I need to do to test...
 
 1. Build and push a custom release image to your own repository.
 
+         docker build --platform linux/amd64 -t quay.io/blah/controlplaneoperator:<tag> -f Dockerfile.control-plane .
```
bryan-cox, line 60: I would suggest to use the dev Dockerfile here to build both the HO and CPO image at the same time - https://github.com/openshift/hypershift/blob/fdd76ee5a56a9384216e1fa87ab5919f4a59fa19/Dockerfile.dev#L1

PR #5579, opened 2025-02-06
NO-JIRA: e2e: combine OpenStack advanced tests within one Nodepool

EmilienM:

This will make our tests to run faster, especially if we keep adding
tests.


EmilienM:
/hold
I wonder if it would make things even faster to run the `OpenStackAdvancedTest` in another HostedCluster? This one seems to have a lot of tests already. Does it have an impact on the tests duration?

EmilienM:
I put hold because I want to see better results, which might come with https://github.com/openshift/release/pull/61363

EmilienM:
this helps save 30 min :partying_face: 
/hold cancel

PR #5585, opened 2025-02-07
OCPBUGS-45600: override OpenShiftPodSecurityAdmission featue gate to false for HCP components

sjenning:
Follow on to #5566

Discovered that, while the FeatureGate cluster config within the guest cluster shows that `OpenshiftPodSecurityAdmission` feature gate is disabled, we do not use that for the feature gate configuration of components on the HCP.

In the guest cluster, we use the cluster-config-operator to set the FeatureGate status based on the feature set (the right thing to do).

For the HCP, we decode the feature gates ourselves from the vendored openshift/api (the wrong thing to do).  Our vendored o/api currently has `OpenshiftPodSecurityAdmission` in the Default FeatureSet so it is being set to true for the cluster-policy-controller running in the HCP.

This results in the `pod-security.kubernetes.io/enforce` being authoritatively set all the time, rendering the KAS config defaults inert.

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/testdata/kube-apiserver/zz_fixture_TestControlPlaneComponents.yaml
+++ b/control-plane-operator/controllers/hostedcontrolplane/testdata/kube-apiserver/zz_fixture_TestControlPlaneComponents.yaml
@@ -29,7 +29,7 @@ spec:
     metadata:
       annotations:
         cluster-autoscaler.kubernetes.io/safe-to-evict-local-volumes: bootstrap-manifests,logs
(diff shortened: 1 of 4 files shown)


sjenning:
/hold

until test artifacts verify that the config for cluster-policy-controller now has `OpenshiftPodSecurityAdmission` gate explicitly set `false`

sjenning:
confirmed that `OpenshiftPodSecurityAdmission=false` in the cluster-policy-controller config

/hold cancel

rtheis (commented):
Thank you.

PR #5600, opened 2025-02-11
HOSTEDCP-2262: Karpenter openshift ec2class

muraee:
While VAP is ok for implementing shared ownership it has some drawbacks. e.g. it forces us to change the builtin CEL rules of the API for required fields which is a maintenance burden and error prone. Besides it doesn't gives control to future api pivots we might need to execute to satisfy business needs. E.g. expose a field for dual stream support which requires picking a different ami, e.g expose a field for kubeletconfig that let us include that in the payload generation.

This PR introduces a new CRD `OpenshiftEC2NodeClass` to be the main interaction point for users with Karpenter instead of directly creating/modifying `EC2NodeClass` objects.

Fixes #

**Checklist**


api/karpenter/v1alpha1/karpenter_types.go
```diff
@@ -0,0 +1,423 @@
+package v1alpha1
```
enxebre, line 1: Let's make this beta, this won't ship as alpha

api/karpenter/v1alpha1/karpenter_types.go
```diff
@@ -0,0 +1,423 @@
...
+	Tags map[string]string `json:"tags,omitempty"`
+
+	// Kubelet defines args to be used when configuring kubelet on provisioned nodes.
+	// They are a subset of the upstream types, recognizing not all options may be supported.
+	// Wherever possible, the types and names should reflect the upstream kubelet types.
+	// +kubebuilder:validation:XValidation:message="imageGCHighThresholdPercent must be greater than imageGCLowThresholdPercent",rule="has(self.imageGCHighThresholdPercent) && has(self.imageGCLowThresholdPercent) ?  self.imageGCHighThresholdPercent > self.imageGCLowThresholdPercent  : true"
+	// +kubebuilder:validation:XValidation:message="evictionSoft OwnerKey does not have a matching evictionSoftGracePeriod",rule="has(self.evictionSoft) ? self.evictionSoft.all(e, (e in self.evictionSoftGracePeriod)):true"
+	// +kubebuilder:validation:XValidation:message="evictionSoftGracePeriod OwnerKey does not have a matching evictionSoft",rule="has(self.evictionSoftGracePeriod) ? self.evictionSoftGracePeriod.all(e, (e in self.evictionSoft)):true"
+	// +optional
+	Kubelet *KubeletConfiguration `json:"kubelet,omitempty" hash:"ignore"`
```
enxebre, line 84: let's drop kubelet config for now

karpenter-operator/controllers/aws/ec2_nodeclass.go
```diff
@@ -0,0 +1,376 @@
+package aws
```
enxebre: let's move this within the existing karpenter/ folder and let it be nodeclass_controller. Or put it under nodeclass/ folder

karpenter-operator/controllers/aws/ec2_nodeclass.go
```diff
@@ -0,0 +1,376 @@
...
+		vap.Spec.MatchConditions = []admissionv1.MatchCondition{
+			{
+				Name:       "exclude-hcco-user",
+				Expression: "'system:hosted-cluster-config' != request.userInfo.username",
+			},
+		}
+
+		vap.Spec.Validations = []admissionv1.Validation{
+			{
+				Expression: "has(oldObject.spec) && has(object.spec) && object.spec == oldObject.spec",
```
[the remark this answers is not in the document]
muraee: name and namespace are already immutable. Do we care about preventing labels and annotations from being updated?

karpenter-operator/controllers/aws/ec2_nodeclass.go
```diff
@@ -0,0 +1,376 @@
...
+	upsert.CreateOrUpdateProvider
+}
+
+func (r *EC2NodeClassReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager, managementCluster cluster.Cluster) error {
+	r.managementClient = managementCluster.GetClient()
+	r.guestClient = mgr.GetClient()
+	r.CreateOrUpdateProvider = upsert.New(false)
+
+	// First install the CRDs so we can create a watch below.
+	if err := r.reconcileCRDs(ctx, true); err != nil {
```
enxebre: I'd expect we let the main controller handle crd installation, then this controller is just implementing mirroring

karpenter-operator/controllers/aws/ec2_nodeclass.go
```diff
@@ -0,0 +1,376 @@
...
+			func(ctx context.Context, o client.Object) []ctrl.Request {
+				// Only watch our Karpenter CRDs
+				switch o.GetName() {
+				case "ec2nodeclasses.karpenter.k8s.aws",
+					"openshiftec2nodeclasses.karpenter.hypershift.openshift.io":
+					return []ctrl.Request{{NamespacedName: client.ObjectKey{Namespace: r.Namespace}}}
+				}
+				return nil
+			},
+		)).
```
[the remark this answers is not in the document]
muraee: `For(&hyperkarpenterv1.OpenshiftEC2NodeClass{}).` adds a watch for openshiftnodeclass.

karpenter-operator/controllers/aws/ec2_nodeclass.go
```diff
@@ -0,0 +1,376 @@
...
+	}
+	if err := utilerrors.NewAggregate(errs); err != nil {
+		return fmt.Errorf("failed to reconcile CRDs: %w", err)
+	}
+	log.Info("Reconciled CRDs", "op", op)
+
+	return nil
+}
+
+func reconcileEC2NodeClass(ec2NodeClass *awskarpenterv1.EC2NodeClass, openshiftEC2NodeClass *hyperkarpenterv1.OpenshiftEC2NodeClass, hcp *hyperv1.HostedControlPlane, userDataSecret *corev1.Secret) error {
```
enxebre: we need units for the mirroring reconciliation. Might be also nice to have fuzz testing

karpenter-operator/controllers/aws/ec2_nodeclass.go
muraee: my thoughts are to let this controller which is aws specific to install its CRDs (EC2NodeClass, etc), and then when we add support for azure, we would need a new controller which will install the azure specific CRDs (AKSNodeClass, etc).

karpenter-operator/controllers/aws/ec2_nodeclass.go
muraee: moved under nodeclass/ and renamed to `ec2_nodeclass_controller`, so we can differentiate it from azure `aks_nodeclass_controller` when we add that.

karpenter-operator/controllers/aws/ec2_nodeclass.go
muraee: updated the unit test

karpenter-operator/controllers/nodeclass/ec2_nodeclass_controller.go
```diff
@@ -0,0 +1,394 @@
...
+		return ctrl.Result{}, err
+	}
+
+	if _, err := r.CreateOrUpdate(ctx, r.guestClient, ec2NodeClass, func() error {
+		return reconcileEC2NodeClass(ec2NodeClass, openshiftEC2NodeClass, hcp, userDataSecret)
+	}); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	if err := r.reconcileStatus(ctx, ec2NodeClass, openshiftEC2NodeClass); err != nil {
```
enxebre: why are we not following the same CreateOrUpdate pattern here?

muraee: CreateOrUpdate  doesn't update status right? we would need to add the type explicitly in upsert.go to for it to work. I just followed what we do for HCP with the client patch.

api/karpenter/v1beta1/karpenter_types.go
```diff
@@ -0,0 +1,414 @@
...
+	// disabled, with httpPutResponseLimit of 1, and with httpTokens
+	// required.
+	// +kubebuilder:default={"httpEndpoint":"enabled","httpProtocolIPv6":"disabled","httpPutResponseHopLimit":1,"httpTokens":"required"}
+	// +optional
+	MetadataOptions *MetadataOptions `json:"metadataOptions,omitempty"`
+
+	// Context is a Reserved field in EC2 APIs
+	// https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateFleet.html
+	// +optional
+	Context *string `json:"context,omitempty"`
```
enxebre, line 112: any reason to expose this?

api/karpenter/v1beta1/karpenter_types.go
```diff
@@ -0,0 +1,414 @@
...
+	//
+	// Refer to recommended, security best practices
+	// (https://aws.github.io/aws-eks-best-practices/security/docs/iam/#restrict-access-to-the-instance-profile-assigned-to-the-worker-node)
+	// for limiting exposure of Instance Metadata and User Data to pods.
+	// If omitted, defaults to httpEndpoint enabled, with httpProtocolIPv6
+	// disabled, with httpPutResponseLimit of 1, and with httpTokens
+	// required.
+	// +kubebuilder:default={"httpEndpoint":"enabled","httpProtocolIPv6":"disabled","httpPutResponseHopLimit":1,"httpTokens":"required"}
+	// +optional
+	MetadataOptions *MetadataOptions `json:"metadataOptions,omitempty"`
```
enxebre, line 107: I don't we want to expose this, instead we can just mirror to the default required

api/karpenter/v1beta1/karpenter_types.go
```diff
@@ -0,0 +1,414 @@
...
+	// +kubebuilder:validation:Pattern:="sg-[0-9a-z]+"
+	// +optional
+	ID string `json:"id,omitempty"`
+
+	// Name is the security group name in EC2.
+	// This value is the name field, which is different from the name tag.
+	Name string `json:"name,omitempty"`
+}
+
+// KubeletConfiguration defines args to be used when configuring kubelet on provisioned nodes.
```
enxebre, line 151: this seems unused

karpenter-operator/controllers/nodeclass/ec2_nodeclass_controller.go
```diff
@@ -0,0 +1,394 @@
...
+		if err := r.guestClient.Status().Patch(ctx, openshiftNodeClass, client.MergeFrom(originalObj)); err != nil {
+			return fmt.Errorf("failed to update status: %v", err)
+		}
+	}
+
+	log.Info("Reconciled OpenshiftEC2NodeClass status")
+	return nil
+}
+
+func (r *EC2NodeClassReconciler) reconcileVAP(ctx context.Context) error {
```
enxebre: what happens if the ec2class is deleted out of band?

muraee: we will just create it back. should we prevent deletion as well?

api/karpenter/v1beta1/karpenter_types.go
muraee: will drop.

api/karpenter/v1beta1/karpenter_types.go
muraee: why not, we expose the option in the NodePool API (via annotation)

devguyio:
Why is this API added first as `v1beta1` and not `v1alpha1` ?

muraee:
@devguyio This API is just a mirror (with some fields removed) of the upsteream `EC2NodeClass` type. And we can't expose it to users in v1alpha.

> Do you mind adding some context behind this PR in the description?

It's in the jira ticket, I added to the PR description as well.

enxebre:
/lgtm
cc @maxcao13

PR #5601, opened 2025-02-11
OCPBUGS-50014: Restart HCP on stale serving certs

csrwng:
Adds code to the HyperShift operator to initiate a restart of control plane workloads when stale certificates are detected. This is limited for hosted clusters with OCP version < 4.19. In 4.19, we will handle automatic restarts within the OCP v2 code.

Fixes #[OCPBUGS-50014](https://issues.redhat.com/browse/OCPBUGS-50014)

**Checklist**


hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -1843,20 +1849,151 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
...
+		return hyperutil.HashSimple(pemBytes), nil
+	}
+}
+
+// shouldCheckForStaleCerts returns true if a HostedCluster should be checked for stale certs
+// The following pre-conditions must be met:
+// - The HostedCluster must be managing its own PKI (the hyperv1.DisablePKIReconciliationAnnotation is not present)
+// - The HostedCluster has been available once
+// - The cluster version of the HostedCluster is < 4.19.0
+var version419 = semver.MustParse("4.19.0")
```
bryan-cox, line 1965: There's an exportable var of this already in version.go. I wonder if we should add this to config.go and update both 🤔

csrwng: The one in version.go should change as we move the main branch to a new version, so I can't rely on that. I will add it to config.go

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -1843,20 +1852,144 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
...
+		}
+		if value := servingCertSecret.Data[corev1.TLSCertKey]; len(value) == 0 {
+			return "", fmt.Errorf("no value for KAS serving certificate in %s", client.ObjectKeyFromObject(servingCertSecret).String())
+		} else {
+			return hyperutil.HashSimple(value), nil
+		}
+	}
+}
+
+func (r *HostedClusterReconciler) kasServingCertHashFromEndpoint(kasHostAndPort string) func() (string, error) {
```
enxebre, line 1939: neat!

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -1843,20 +1852,144 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
...
+	kasServingCertHashFromEndpoint func() (string, error)) func() (map[string]string, error) {
+	return func() (map[string]string, error) {
+		if !shouldCheckForStaleCerts() {
+			return nil, nil
+		}
+
+		// The hash from the KAS serving secret is the source of truth for what
+		// certificate we should be using for KAS
+		// We can check whether we need to restart pods in one of 2 ways:
+		// 1. We have previously saved an annotation on the HCP that says which was the last
```
enxebre, line 1873: nit: 1 also uses the annotation to restart but reading this comment is not clear.

enxebre:
any chance to e2e test this?
lgtm

csrwng:
> any chance to e2e test this?
@enxebre my fear with a e2e is that in order to not make it flaky, we need to allow for a long enough cert expiration period (say ~30 min), and waiting for certs to expire might make the presub test take longer than what we'd like. 

I will add an e2e test that only runs when a certain env var is set, and then pass that for periodics only.

csrwng:
Pushed fix/ut for setting/removing autoscaling annotation.

PR #5602, opened 2025-02-11
HOSTEDCP-2241: [CPOv2 Refactor] Refactor ingress operator

muraee:

Fixes #

**Checklist**

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/v2/assets/ingress-operator/deployment.yaml
+++ b/control-plane-operator/controllers/hostedcontrolplane/v2/assets/ingress-operator/deployment.yaml
@@ -1,0 +1,91 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: ingress-operator
+spec:
+  replicas: 1
+  revisionHistoryLimit: 2
+  selector:
+    matchLabels:
+      name: ingress-operator
+  strategy:
+    type: Recreate
+  template:
+    metadata:
+      annotations:
+        target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
+      labels:
+        app: ingress-operator
+        name: ingress-operator
+    spec:
+      containers:
+      - command:
+        - ingress-operator
+        - start
+        - --namespace
+        - openshift-ingress-operator
+        - --image
+        - $(IMAGE)
+        - --canary-image
+        - $(CANARY_IMAGE)
+        - --release-version
+        - $(RELEASE_VERSION)
+        - --metrics-listen-addr
+        - 0.0.0.0:60000
+        env:
+        - name: RELEASE_VERSION
+          value: ""
+        - name: IMAGE
+          value: ""
+        - name: CANARY_IMAGE
+          value: ""
+        - name: KUBECONFIG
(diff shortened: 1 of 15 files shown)


control-plane-operator/controllers/hostedcontrolplane/v2/assets/ingress-operator/deployment.yaml
```diff
@@ -0,0 +1,97 @@
...
+metadata:
+  name: ingress-operator
+spec:
+  replicas: 1
+  revisionHistoryLimit: 2
+  selector:
+    matchLabels:
+      name: ingress-operator
+  strategy:
+    type: Recreate
```
enxebre, line 12: why is this set here? should we enforce this value?

control-plane-operator/controllers/hostedcontrolplane/v2/assets/ingress-operator/deployment.yaml
```diff
@@ -0,0 +1,97 @@
...
+          requests:
+            cpu: 10m
+            memory: 10Mi
+        volumeMounts:
+        - mountPath: /var/run/secrets/openshift/serviceaccount
+          name: serviceaccount-token
+        - mountPath: /etc/kubernetes
+          name: admin-kubeconfig
+      volumes:
+      - name: ingress-operator-kubeconfig
```
enxebre, line 88: ingress kubeconfig can be self service after https://github.com/openshift/hypershift/pull/5607

enxebre:
Thanks! it needs rebase, lgtm otherwise

control-plane-operator/controllers/hostedcontrolplane/v2/assets/ingress-operator/deployment.yaml
muraee: not all components use `Recreate`, a lot of them use `RollingUpdate`

support/controlplane-component/defaults.go
```diff
@@ -51,6 +51,10 @@ func (c *controlPlaneWorkload) defaultOptions(cpContext ControlPlaneContext, pod
 		return nil, err
 	}
 
+	if c.serviceAccountKubeConfigOpts != nil {
+		c.addServiceAccountKubeconfigVolumes(podTemplateSpec)
```
muraee, line 55: moved this before konnectivity-container injection to avoid adding a duplicate volumeMount, since konnectivity adds its own kubeconfig volumeMount.

control-plane-operator/controllers/hostedcontrolplane/v2/assets/ingress-operator/deployment.yaml
enxebre: I think we should enable this via library .withStrategy

muraee: I think it makes sense to leave it in the manifest to reduce the code part and keep the manifest as the authoritative source.
We also support statefulsets and cronJobs (soon via https://github.com/openshift/hypershift/pull/5663) which don't use the same field type,

PR #5607, opened 2025-02-11
NO-JIRA: CPOv2 refactor self service sa kubeconfig

enxebre:
Add cpov2 capability to self service sa kubeconfig  for components
includes https://github.com/openshift/hypershift/pull/5603

Fixes #

**Checklist**


support/controlplane-component/kubeconfig.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+func (c *controlPlaneWorkload) serviceAccountKubeconfigSecret() *corev1.Secret {
+	return &corev1.Secret{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "v1",
+			Kind:       "Secret",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      c.serviceAccountKubeConfigOpts.Name + "-kubeconfig",
```
muraee, line 58: should we use the component name here instead?

muraee:
since there is no kubeconfig manifest anymore, the secret will not be in the controlPlaneComponent status like the other manifests here https://github.com/openshift/hypershift/blob/main/support/controlplane-component/status.go#L81-L85

can we add it please

support/controlplane-component/builder.go
```diff
@@ -82,6 +82,17 @@ func (b *controlPlaneWorkloadBuilder[T]) InjectAvailabilityProberContainer(opts
...
 
+// InjectServiceAccountKubeConfig will cause the generation of a secret with a kubeconfig using certificates for the given named service account
+// and the volume mounts for that secret withing the given mountPath.
+func (b *controlPlaneWorkloadBuilder[T]) InjectServiceAccountKubeConfig(opts ServiceAccountKubeConfigOpts) *controlPlaneWorkloadBuilder[T] {
+	b.workload.serviceAccountKubeConfigOpts = &opts
+	return b
+}
+
+type ServiceAccountKubeConfigOpts struct {
+	Name, Namespace, MountPath string
```
muraee, line 93: I think we also need a predicate option here. for example, the DNS operator will need to disable this if `hyperv1.DisablePKIReconciliationAnnotation` is present.

support/controlplane-component/kubeconfig.go
enxebre: updated

enxebre:
updated according to feedback and covered with units

support/controlplane-component/status.go
```diff
@@ -89,6 +89,14 @@ func (c *controlPlaneWorkload) reconcileComponentStatus(cpContext ControlPlaneCo
 		return err
 	}
 
+	if c.serviceAccountKubeConfigOpts != nil {
```
muraee, line 92: can we also check `hyperv1.DisablePKIReconciliationAnnotation` annotation here.

support/controlplane-component/controlplane-component.go
```diff
@@ -253,6 +256,17 @@ func (c *controlPlaneWorkload) update(cpContext ControlPlaneContext) error {
 		return err
 	}
 
+	if c.serviceAccountKubeConfigOpts != nil {
+		_, disablePKIReconciliationAnnotation := cpContext.HCP.Annotations[hyperv1.DisablePKIReconciliationAnnotation]
+		if !disablePKIReconciliationAnnotation {
```
muraee, line 261: I think we need an else here to handle the secret deletion when `hyperv1.DisablePKIReconciliationAnnotation` annotation is added day2

support/controlplane-component/status.go
enxebre: done

support/controlplane-component/controlplane-component.go
enxebre: I don't think that replicates current behaviour, does it?

muraee: yeah I don't think we do that in v1, but I think it makes sense to cleanup, no strong feelings though,

support/controlplane-component/kubeconfig.go
```diff
@@ -0,0 +1,77 @@
...
+
+func (c *controlPlaneWorkload) serviceAccountKubeconfigSecret() *corev1.Secret {
+	return &corev1.Secret{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "v1",
+			Kind:       "Secret",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      c.serviceAccountKubeconfigSecretName(),
+			Namespace: c.serviceAccountKubeConfigOpts.Namespace,
```
muraee, line 67: this should be HCP namespace

support/controlplane-component/controlplane-component.go
enxebre: Let's dont' deviate anything from current behaviour before this is fully migrated and delivered

muraee:
/retitle NO-JIRA: CPOv2 refactor self service sa kubeconfig

PR #5618, opened 2025-02-13
OCPBUGS-37323: MarkPersistentFlagRequired does not work since CLI refactoring

georgelipceanu:
Enforces --name and --pull-secret flag requirements to Hypershift create cluster command since omitting them can lead to misconfigured deployments.
Fixes #[OCPBUGS-37323](https://issues.redhat.com/browse/OCPBUGS-37323)

**Checklist**

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/cno/clusternetworkoperator.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/cno/clusternetworkoperator.go
@@ -104,22 +104,23 @@ type Images struct {
 }
 
 type Params struct {
-	ReleaseVersion          string
-	AvailabilityProberImage string
-	HostedClusterName       string
-	CAConfigMap             string
-	CAConfigMapKey          string
-	APIServerAddress        string
-	APIServerPort           int32
-	TokenAudience           string
-	Images                  Images
-	OwnerRef                config.OwnerRef
-	DeploymentConfig        config.DeploymentConfig
-	IsPrivate               bool
-	DefaultIngressDomain    string
-	AzureClientID           string
-	AzureTenantID           string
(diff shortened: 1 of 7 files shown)


cmd/cluster/core/create.go
```diff
@@ -43,7 +43,7 @@ import (
 func DefaultOptions() *RawCreateOptions {
 	return &RawCreateOptions{
 		Namespace:                      "clusters",
-		Name:                           "example",
+		Name:                           "",
```
bryan-cox, line 46: Why was example removed here?

bryan-cox (commented):
Did you determine the root cause of the issue with MarkPersistentFlagRequired?

cmd/cluster/core/create.go
georgelipceanu: The DefaultOptions() function sets the name value to "example" which is passed into the BindOptions() function found at https://github.com/openshift/hypershift/blob/8be1d9c6f8f79106444e48f2b7d0069b942ba0d7/product-cli/cmd/cluster/cluster.go#L27 and since the name is set, the flag was considered set. I couldn't find where the MarkPersistentFlagRequired was being used in the create commands and I only see it set on the destroy commands, where it works as expected.

georgelipceanu: Just tried that there and it seems to be working fine as well, I remember trying to do the same thing before and it wouldn't work but I'm assuming it was because I forgot to rebuild the binaries or something like that.
Either way I'll just make that change now.

bryan-cox:
/retitle OCPBUGS-37323: MarkPersistentFlagRequired does not work since CLI refactoring

PR #5622, opened 2025-02-13
NO-JIRA: bump Cluster API Provider OpenStack to v0.12 (for 4.19 only)

MaysaMacedo:

- Bumps CAPO dependency to v0.12.0
- Updates the generated assets with make update

Fixes #

**Checklist**


go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/openshift/hypershift
 
-go 1.23.0
+go 1.23.4
```
EmilienM, line 3: One way to drop that dependency is to release 0.12.1 in CAPI which relies has 1.23.0 instead of 1.23.4.
I think it's better practice to not hard-code the patch version of go in go.mod AFIK.

MaysaMacedo: okay, I just fixed it.

EmilienM:
/retitle NO-JIRA: bump Cluster API Provider OpenStack to v0.12 (for 4.19 only)

PR #5657, opened 2025-02-17
NO-JIRA: update hypershift azure roles to use ids

Patryk-Stefanski:

Fixes #

**Checklist**

### DIFF
--- a/cmd/infra/azure/create.go
+++ b/cmd/infra/azure/create.go
@@ -271,7 +271,7 @@ func (o *CreateInfraOptions) Run(ctx context.Context, l logr.Logger) (*CreateInf
 		if err != nil {
 			return nil, err
 		}
-		err = assignRole(objectID, "Azure Red Hat OpenShift Image Registry Operator Role", managedRG)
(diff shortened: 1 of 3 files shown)


cmd/infra/azure/create.go
```diff
@@ -824,32 +824,32 @@ func assignServicePrincipalRoles(subscriptionID, managedResourceGroupName, nsgRe
...
-	role := "Contributor"
+	role := "b24988ac-6180-42a0-ab88-20f7382dd24c"
 
 	scopes := []string{managedRG}
 
 	// TODO CNTRLPLANE-171: CPO, KMS, and NodePoolManagement will need new roles that do not exist today
 	switch component {
 	case cloudProvider:
-		role = "Azure Red Hat OpenShift Cloud Controller Manager Role"
+		role = "5b7237c5-45e1-49d6-bc18-a1f62f400748"
```
bryan-cox, line 834: I don't think this is right. It's the same as the azure disk and they were not the same as before.

bryan-cox: I fixed this. This should fix the failing e2e-aks test as well 🤞🏻 .

PR #5663, opened 2025-02-18
HOSTEDCP-2246: [CPOv2 Refactor] Refactor OLM

muraee:

Fixes #

**Checklist**

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/v2/olm/collect_profiles/component.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/v2/olm/collect_profiles/component.go
@@ -1,0 +1,40 @@
+package collectprofiles
+
+import (
+	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
+	component "github.com/openshift/hypershift/support/controlplane-component"
+)
+
+const (
+	ComponentName = "olm-collect-profiles"
+)
+
+var _ component.ComponentOptions = &olmCollectProfiles{}
+
+type olmCollectProfiles struct{}
+
+// IsRequestServing implements controlplanecomponent.ComponentOptions.
+func (r *olmCollectProfiles) IsRequestServing() bool {
+	return false
+}
+
+// MultiZoneSpread implements controlplanecomponent.ComponentOptions.
+func (r *olmCollectProfiles) MultiZoneSpread() bool {
+	return false
+}
+
+// NeedsManagementKASAccess implements controlplanecomponent.ComponentOptions.
+func (r *olmCollectProfiles) NeedsManagementKASAccess() bool {
+	return true
+}
+
+func NewComponent() component.ControlPlaneComponent {
+	return component.NewCronJobComponent(ComponentName, &olmCollectProfiles{}).
+		WithAdaptFunction(adaptCronJob).
+		WithPredicate(predicate).
+		Build()
+}
+
+func predicate(cpContext component.WorkloadContext) (bool, error) {
+	return cpContext.HCP.Spec.Platform.Type != hyperv1.IBMCloudPlatform, nil
+}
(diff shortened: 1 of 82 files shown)


support/controlplane-component/controlplane-component.go
```diff
@@ -40,6 +41,7 @@ type ControlPlaneContext struct {
 	ReleaseImageProvider     imageprovider.ReleaseImageProvider
 	UserReleaseImageProvider imageprovider.ReleaseImageProvider
 	ImageMetadataProvider    util.ImageMetadataProvider
+	DigestLister             registryclient.DigestListerFN
```
enxebre, line 44: can we document DigestLister and the other Public variables here

control-plane-operator/controllers/hostedcontrolplane/v2/olm/collect_profiles/component.go
```diff
@@ -0,0 +1,40 @@
...
+	return false
+}
+
+// NeedsManagementKASAccess implements controlplanecomponent.ComponentOptions.
+func (r *olmCollectProfiles) NeedsManagementKASAccess() bool {
+	return true
+}
+
+func NewComponent() component.ControlPlaneComponent {
+	return component.NewCronJobComponent(ComponentName, &olmCollectProfiles{}).
```
enxebre, line 32: why couldn't this just be a manifest with an adapter?

support/controlplane-component/controlplane-component.go
```diff
@@ -311,57 +278,14 @@ func (c *controlPlaneWorkload) reconcileWorkload(cpContext ControlPlaneContext)
...
-
-	deploymentConfig, err := c.defaultOptions(cpContext, &deployment.Spec.Template, deployment.Spec.Replicas, existingResources)
+	deploymentConfig, err := c.defaultOptions(cpContext, c.workloadProvider.PodTemplateSpec(workloadObj), c.workloadProvider.Replicas(workloadObj))
 	if err != nil {
 		return err
 	}
-	deploymentConfig.ApplyTo(deployment)
-	return nil
-}
+	c.workloadProvider.ApplyOptionsTo(cpContext, workloadObj, oldWorkloadObj, deploymentConfig)
```
enxebre, line 285: I wouldn't expect we need to pass deploymentConfig through any signaturem specially public

support/controlplane-component/cronjob.go
```diff
@@ -0,0 +1,62 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var _ WorkloadProvider[*batchv1.CronJob] = &cronJobProvider{}
+
+type cronJobProvider struct {
+}
+
+// ApplyOptionsTo implements workload.
```
enxebre, line 18: can we document the interface funcs?

support/controlplane-component/controlplane-component.go
muraee: the problem is there is not generic function for `deploymentConfig.ApplyTo` si each workloadProvider needs to call the correct function. I'd like to get rid of deploymentConfig in the cpov2 completely once finish the refactors to make the code more readable and explicit.

> through any signaturem specially public

technically this should be only used within this package, I can make it the func private

enxebre: we could hide this within the deploymentConfig package rather than surface it in this library
```
	switch any(workloadObj).(type) {
	case appsv1.Deployment:
		// call right function
	case stateful:
```

not a blocker for me

support/controlplane-component/controlplane-component.go
muraee: done

support/controlplane-component/cronjob.go
muraee: done

support/controlplane-component/controlplane-component.go
muraee: thanks, I will follow-up with a PR to remove deploymentConfig from v2

PR #5664, opened 2025-02-18
OCPBUGS-51007: test/e2e: fix WaitForImageRollout to actually wait on upgrade case

sjenning:
The general idea of the fix is to detect the case where the HC image has been changed but that change has not yet propagated to HostedCluster status via ClusterVersion status.

The PR detects this by recording the `CompletionTime` of the latest rollout in the version history when  `WaitForImageRollout` is first called, immediately after the image is changed on the HostedCluster.

As long as the latest rollout in the version history continues to match this recorded time, the update has not yet propagated through ClusterVersion and to the HostedCluster status.  Once the propagation occurs, the latest entry in the version history will be `Partial` until the rollout completes, for which there is an existing block.

### DIFF
--- a/test/e2e/create_cluster_test.go
+++ b/test/e2e/create_cluster_test.go
@@ -1411,6 +1411,8 @@ func testCreateClusterPrivate(t *testing.T, enableExternalDNS bool) {
 	clusterOpts := globalOpts.DefaultClusterOptions(t)
 	clusterOpts.ControlPlaneAvailabilityPolicy = string(hyperv1.SingleReplica)
 	clusterOpts.AWSPlatform.EndpointAccess = string(hyperv1.Private)
+	// DO NOT COMMIT
+	clusterOpts.NodePoolReplicas = 0
 	expectGuestKubeconfHostChange := false
 	if !enableExternalDNS {
 		clusterOpts.ExternalDNSDomain = ""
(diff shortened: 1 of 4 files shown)


sjenning:
/hold
until fix verification by test artifacts

test/e2e/create_cluster_test.go
```diff
@@ -1448,7 +1447,6 @@ func testSwitchFromPublicToPrivate(ctx context.Context, client crclient.Client,
 			obj.Spec.Platform.AWS.EndpointAccess = hyperv1.Private
 		})
 		g.Expect(err).ToNot(HaveOccurred(), "failed to update hostedcluster EndpointAccess")
-
-		e2eutil.ValidatePrivateCluster(t, ctx, client, hostedCluster, clusterOpts)
```
enxebre, line 1452: why are we losing this signal?

enxebre:
Why do we need to lose test coverage to fix WaitForImageRollout?
We know the release image we are waiting to rollout. Shouldn't we just loop over the .history, find the match with .history.[n].image and .history.[n].state is Completed?

test/e2e/create_cluster_test.go
sjenning: This was not actually testing anything as the cluster is validated before the switch (all COs ready, CV rolled out, nodes ready).  Then we switch and immediately revalidate, all assertions pass in 0s.

What we should really do is check the KAS and oauth routes and make sure that they transition properly.

sjenning:
> We know the release image we are waiting to rollout.
We do not.

The CVO reports the image after sha tag replacement and registry overrides.  So we can't simply compare the image in the HC spec to the image in the CV status.

PR #5672, opened 2025-02-19
CNTRLPLANE-35: skip EnsureValidatingAdmissionPoliciesExists/EnsureVal…

Patryk-Stefanski:
…idatingAdmissionPoliciesCheckDeniedRequests based on HC CPO version

We need to skip based on the HC CPO version because when running minor upgrade tests we dont want to use the latest release image version as this would give innacurate test results

Fixes # [CNTRLPLANE-35](https://issues.redhat.com/browse/CNTRLPLANE-35)

**Checklist**

### DIFF
--- a/test/e2e/util/version.go
+++ b/test/e2e/util/version.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"testing"
 
+	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
 	"github.com/openshift/hypershift/support/releaseinfo"
 
 	"github.com/blang/semver"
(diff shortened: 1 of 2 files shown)


test/e2e/util/version.go
```diff
@@ -61,6 +62,13 @@ func AtLeast(t *testing.T, version semver.Version) {
 	}
 }
 
+func CPOAtLeast(t *testing.T, version semver.Version, hc *hyperv1.HostedCluster) {
+	cpoVersion := semver.MustParse(hc.Status.Version.Desired.Version)
```
sjenning, line 66: I feel like we should panic here if `hc.Status.Version == nil` or `hc.Status.Version.Desired.Version == ""`

That means that `CPOAtLeast()` was called from a context where `WaitForImageRollout()` was not called first.

Patryk-Stefanski: This i ran as part of the t.after() function which runs only after the test and after the t.before() is called which calls the WaitForImageRoolout(). 

I will add nil checks just to make sure we don't panic and return some nice logs instead.

PR #5706, opened 2025-02-25
CNTRLPLANE-244: Enable initial GitHub Action jobs

bryan-cox:
This PR enables an initial set of GitHub Action jobs including: codespell, golangci-lint, verify, and unit testing

A part of [CNTRLPLANE-244](https://issues.redhat.com/browse/CNTRLPLANE-244)

**Checklist**

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -1,0 +1,19 @@
+name: golangci-lint
+on:
+  pull_request:
+    branches:
+      - main
+jobs:
+  golangci-lint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+      - name: Setup Go environment
+        uses: actions/setup-go@v5.3.0
+        with:
+          go-version-file: go.mod
+      - name: Run golangci-lint
+        uses: golangci/golangci-lint-action@v6.5.0
+        with:
+          args: --timeout=15m --config=.golangci.yml --issues-exit-code=0
--- a/.github/workflows/make-verify.yml
+++ b/.github/workflows/make-verify.yml
@@ -1,0 +1,17 @@
+name: make-verify
+on:
+  pull_request:
+    branches:
+      - main
+jobs:
+  make-verify:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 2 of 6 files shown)


.github/workflows/make-verify.yml
```diff
@@ -0,0 +1,17 @@
...
+    branches:
+      - main
+jobs:
+  make-verify:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+      - name: Setup Go environment
+        uses: actions/setup-go@v5.3.0
```
sjenning, line 13: does this always deploy go with the version that matches go.mod?

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,19 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@v4
+      - name: Setup Go environment
+        uses: actions/setup-go@v5.3.0
+        with:
+          go-version-file: go.mod
+      - name: Run golangci-lint
+        uses: golangci/golangci-lint-action@v6.5.0
+        with:
+          args: --timeout=30m --config=.golangci.yml --issues-exit-code=0
```
sjenning, line 19: newline

.github/workflows/unit-test.yml
```diff
@@ -0,0 +1,29 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+      - name: Setup Go environment
+        uses: actions/setup-go@v5.3.0
+        with:
+          go-version-file: go.mod
+          cache: false
+      - name: Run unit tests
+        run: go test -race -parallel=$(nproc) -count=1 -timeout=30m -coverprofile cover.out -covermode=atomic -coverpkg=./... ./...
```
sjenning, line 23: wondering if we want to tighten this if we pay by the minute.  maybe 15m?  seems like it completes in 8m atm.

.github/workflows/make-verify.yml
bryan-cox: It looks like it will always match our go.mod per https://github.com/actions/setup-go?tab=readme-ov-file#getting-go-version-from-the-gomod-file

PR #5710, opened 2025-02-26
OSASINFRA-3733: deploy ORC on release payload >= 4.19

EmilienM:

ORC was added to the release payload in 4.19 so we don't want to
install it in 4.18.

### DIFF
--- a/hypershift-operator/controllers/hostedcluster/internal/platform/openstack/openstack.go
+++ b/hypershift-operator/controllers/hostedcluster/internal/platform/openstack/openstack.go
@@ -22,6 +22,8 @@ import (
 	capo "sigs.k8s.io/cluster-api-provider-openstack/api/v1beta1"
 	capiv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/blang/semver"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


hypershift-operator/controllers/hostedcluster/internal/platform/openstack/openstack.go
```diff
@@ -252,66 +256,76 @@ func (a OpenStack) CAPIProviderDeploymentSpec(hcluster *hyperv1.HostedCluster, _
...
-						},
-						InitialDelaySeconds: 15,
-						PeriodSeconds:       20,
+				}},
+			}},
+	}
+
+	// Add the ORC manager container if the payload version is 4.19 or later
+	// ORC was decoupled from CAPO in 4.19 but was part of CAPO in 4.18.
+	if a.payloadVersion != nil && a.payloadVersion.Major == 4 && a.payloadVersion.Minor >= 19 {
```
enxebre, line 265: This would benefit from unit tests that validate the spec based on the passed payloadVersion

EmilienM:
/hold
this didn't work.
@enxebre you're right. I'll take a look.

EmilienM:
I don't understand why ORC isn't deployed with that PR: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_hypershift/5710/pull-ci-openshift-hypershift-main-e2e-openstack-aws/1894836315447365632/artifacts/e2e-openstack-aws/hypershift-openstack-e2e-execute/artifacts/TestCreateCluster/namespaces/e2e-clusters-5qbwh-create-cluster-x8m4g/core/pods/capi-provider-57b8b46b79-m6wjk.yaml

The release payload that is used should be 4.19 :thinking:

EmilienM:
/hold cancel
I think this is good to go, ORC was deployed.

PR #5718, opened 2025-02-27
openstack: better support for BYON & custom KAS VIP

EmilienM:
WIP

### DIFF
--- a/api/hypershift/v1beta1/openstack.go
+++ b/api/hypershift/v1beta1/openstack.go
@@ -136,6 +136,12 @@ type OpenStackPlatformSpec struct {
 	// +kubebuilder:validation:XValidation:rule="isIP(self)",message="floatingIP must be a valid IPv4 or IPv6 address"
 	// +optional
 	IngressFloatingIP string `json:"ingressFloatingIP,omitempty"`
+
+	// KASPortID is the ID of a pre-existing port to use for the API server VIP.
+	//
+	// +kubebuilder:validation:Format:=uuid
+	// +optional
+	KASPortID string `json:"kasPortID,omitempty"`
 }
 
 // OpenStackIdentityReference is a reference to an infrastructure
--- a/control-plane-operator/controllers/hostedcontrolplane/kas/service.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/kas/service.go
@@ -74,6 +74,13 @@ func ReconcileService(svc *corev1.Service, strategy *hyperv1.ServicePublishingSt
 		} else {
 			svc.Spec.Type = corev1.ServiceTypeClusterIP
 		}
+		if hcp.Spec.Platform.Type == hyperv1.OpenStackPlatform && hcp.Spec.Platform.OpenStack != nil && hcp.Spec.Platform.OpenStack.KASPortID != "" {
+			// https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md#service-annotations
+			svc.Annotations["loadbalancer.openstack.org/port-id"] = hcp.Spec.Platform.OpenStack.KASPortID
+			// the loadbalancer VIP won't be associated with a floating IP.
+			// This is a hack now for the PoC; we need to design it properly, because this is very opinionated.
+			svc.Annotations["service.beta.kubernetes.io/openstack-internal-load-balancer"] = "true"
+		}
 	case hyperv1.NodePort:
 		svc.Spec.Type = corev1.ServiceTypeNodePort
 		if portSpec.NodePort == 0 && strategy.NodePort != nil {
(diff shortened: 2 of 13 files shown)


mdbooth:
I really like this, I think it's really promising.

I'm assuming we're either allowing CAPO to create the cluster 'default' network, or allowing the user to pass it in (BYON). Do we need to support creating the api VIP on anything other than that subnet? I feel like we've already got that subnet (created ourselves or BYON), and it's presumably already guaranteed to be routable from everywhere inside the cluster. Could we just use that and potentially save the user from a footgun?

I also think we're should do the RBAC thing ourselves if we're creating the network, or would we only support this feature for BYON?

EmilienM:
> I really like this, I think it's really promising.
> 
> I'm assuming we're either allowing CAPO to create the cluster 'default' network, or allowing the user to pass it in (BYON). Do we need to support creating the api VIP on anything other than that subnet? I feel like we've already got that subnet (created ourselves or BYON), and it's presumably already guaranteed to be routable from everywhere inside the cluster. Could we just use that and potentially save the user from a footgun?
> 
> I also think we're should do the RBAC thing ourselves if we're creating the network, or would we only support this feature for BYON?

I think we should keep the "BYON" model where there are so many use cases out there that we can't support them all, so I'm not in favor of controlling the RBAC in e.g. HCP or CAPO.

control-plane-operator/controllers/hostedcontrolplane/kas/service.go
```diff
@@ -74,6 +74,13 @@ func ReconcileService(svc *corev1.Service, strategy *hyperv1.ServicePublishingSt
 		} else {
 			svc.Spec.Type = corev1.ServiceTypeClusterIP
 		}
+		if hcp.Spec.Platform.Type == hyperv1.OpenStackPlatform && hcp.Spec.Platform.OpenStack != nil && hcp.Spec.Platform.OpenStack.KASPortID != "" {
+			// https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md#service-annotations
+			svc.Annotations["loadbalancer.openstack.org/port-id"] = hcp.Spec.Platform.OpenStack.KASPortID
+			// the loadbalancer VIP won't be associated with a floating IP.
+			// This is a hack now for the PoC; we need to design it properly, because this is very opinionated.
+			svc.Annotations["service.beta.kubernetes.io/openstack-internal-load-balancer"] = "true"
```
EmilienM, line 82: we could also set `loadbalancer.openstack.org/hostname` so we get a control plane endpoint hostname instead of IP. Customers would preprovision the KAS VIP port and its DNS record before deploying the HCP.

control-plane-operator/controllers/hostedcontrolplane/kas/service.go
```diff
@@ -74,6 +74,13 @@ func ReconcileService(svc *corev1.Service, strategy *hyperv1.ServicePublishingSt
 		} else {
 			svc.Spec.Type = corev1.ServiceTypeClusterIP
 		}
+		if hcp.Spec.Platform.Type == hyperv1.OpenStackPlatform && hcp.Spec.Platform.OpenStack != nil && hcp.Spec.Platform.OpenStack.KASPortID != "" {
+			// https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md#service-annotations
+			svc.Annotations["loadbalancer.openstack.org/port-id"] = hcp.Spec.Platform.OpenStack.KASPortID
+			// the loadbalancer VIP won't be associated with a floating IP.
+			// This is a hack now for the PoC; we need to design it properly, because this is very opinionated.
```
mdbooth, line 81: FYI I don't see this as a hack. We don't want a FIP in this case.

control-plane-operator/controllers/hostedcontrolplane/kas/service.go
mdbooth: What would be the advantage of this?

api/hypershift/v1beta1/openstack.go
```diff
@@ -136,6 +136,12 @@ type OpenStackPlatformSpec struct {
 	// +kubebuilder:validation:XValidation:rule="isIP(self)",message="floatingIP must be a valid IPv4 or IPv6 address"
 	// +optional
 	IngressFloatingIP string `json:"ingressFloatingIP,omitempty"`
+
+	// KASPortID is the ID of a pre-existing port to use for the API server VIP.
+	//
+	// +kubebuilder:validation:Format:=uuid
+	// +optional
+	KASPortID string `json:"kasPortID,omitempty"`
```
mdbooth, line 144: Why a port rather than a subnet?

control-plane-operator/controllers/hostedcontrolplane/kas/service.go
EmilienM: I can name many benefits of using a hostname vs an IP but here are a few from the top of my head:
* allow to change the underlying IP address of the Kubernetes API server without requiring reconfiguration of clients
* security policies and firewall rules are easier to manage when using DNS-based allowlists instead of IP-based rules
* (to be verified) easier to handle TLS certificates

PR #5722, opened 2025-02-28
OCPBUGS-52172: [release-4.15] Add HostedCluster additional trustbundles to konnectivity-https-proxy

enxebre:
Manual cherry-pick for https://github.com/openshift/hypershift/pull/5707

When using the konnectivity-https-proxy sidecar to communicate through workers, the proxy dials any configured proxy directly and fails to connect if it cannot trust the proxy's serving certificate. This commit adds mounts for the trusted-ca-bundle-managed configmap to sidecar containers that run the konnectivity-https-proxy. The configmap contains the aggregate of .spec.additionalTrustBundle and

This fixes the issue of the ingress operator not becoming available when using a proxy with a custom certificate.

In addition it propagates any proxy environment variables that are set on the control plane operator to the konnectivity-https-proxy containers. This is needed when the management cluster itself requires a proxy and the proxy container dials an address directly such as when it needs to communicate with a cloud provider endpoint.

Fixes #

**Checklist**

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/ingressoperator/ingressoperator.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/ingressoperator/ingressoperator.go
@@ -2,12 +2,15 @@ package ingressoperator
 
 import (
 	"fmt"
+	"path"
 
 	configv1 "github.com/openshift/api/config/v1"
 	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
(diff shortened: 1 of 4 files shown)


rtheis:
Please hold any 4.14 cherry-pick due to this breaking IBM Cloud.

rtheis:
Feel free to continue with a cherry-pick so long as it includes https://github.com/openshift/hypershift/pull/5813 too.  Thank you.

PR #5724, opened 2025-02-28
CNTRLPLANE-269: Remove DeploymentConfig usage from cpov2

muraee:
This removes the usage of the legacy DeploymentConfig from cpov2 and moves all the PodTemplateSpec defaulting logic to the ControlPlaneComponent. This includes setting annotations, labels, isolation, colocation, spread, etc.

Fixes #

**Checklist**

### DIFF
--- a/support/controlplane-component/defaults.go
+++ b/support/controlplane-component/defaults.go
@@ -3,6 +3,7 @@ package controlplanecomponent
 import (
 	"context"
 	"fmt"
+	"maps"
 	"slices"
 	"strings"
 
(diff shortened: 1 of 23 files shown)


support/controlplane-component/defaults.go
```diff
@@ -31,26 +52,42 @@ var (
...
-		c.Name() == "packageserver" { // TODO: investigate why packageserver needs AutomountServiceAccountToken or set NeedsManagementKASAccess to true.
-		podTemplateSpec.Spec.AutomountServiceAccountToken = ptr.To(true)
+func (c *controlPlaneWorkload[T]) setDefaultOptions(cpContext ControlPlaneContext, workloadObj T, existingResources map[string]corev1.ResourceRequirements) error {
+	hcp := cpContext.HCP
+
+	labels := workloadObj.GetLabels()
+	if labels == nil {
+		labels = map[string]string{}
+	}
+	labels[managedByLabel] = "control-plane-operator-v2"
```
enxebre, line 62: I'd prefer if we don't change this since upper level tooling might be doing assumptions

muraee: haven't thought about that, will change it back.

enxebre:
/hold
let's hold until we switch

bryan-cox:
We should not be making large changes without a Jira ticket.

Why is this change needed? The PR description does not reference the why. The commit description also does not reference the why. These should be added. It's hard for someone to quickly understand why we need this change.

bryan-cox:
> @bryan-cox Updated the PR with the jira ticket. This is continuation of the effort to switch to CPOv2, It's moving the old DeploymentConfig code to cpov2.

Right but we still should be adding the why to the PR and to the commit if the commit is not extremely clear. We ask others to do the same - https://hypershift.pages.dev/contribute/#prior-to-submitting-a-pull-request.

muraee:
created https://issues.redhat.com/browse/CNTRLPLANE-844 to track the removal of the remaining refs to DeploymentConfig

PR #5746, opened 2025-03-03
OLS-1370: Add full aws permission list and point to it from "getting started"

JoaoFula:
**What this PR does / why we need it**: Adding the full list of aws permissions for creation of hypershift hosted clusters. Also adding it as a separate for user friendliness. Like this users can just copy one set of permissions to a policy and be able to create any IAM user capable of spinning hypershift hosted clusters.

**Which issue(s) this PR fixes** *(optional, use `fixes #<issue_number>(, fixes #<issue_number>, ...)` format, where issue_number might be a GitHub issue, or a Jira story*: The issue where it was hard to find what set of permissions we needed and it required us to try many times and note down every permission that was missing from the lists. Furthermore, we compressed it to one list to make it easy to copy for any user.
Fixes #

**Checklist**


bryan-cox:
/retitle OLS-1370: Add full aws permission list and point to it from "getting started"

PR #2408, opened 2023-04-12
HOSTEDCP-807: Requeue HCP always

muraee:
**What this PR does / why we need it**:
Requeue HCP controller with different intervals depending on the Ready status. This will ensure external resources are reconciled in a timely fashion.

**Which issue(s) this PR fixes** *(optional, use `fixes #<issue_number>(, fixes #<issue_number>, ...)` format, where issue_number might be a GitHub issue, or a Jira story*:
Fixes #[HOSTEDCP-807](https://issues.redhat.com//browse/HOSTEDCP-807) (LoadBalancer taking a long time to report ready, because the reconcile loop can't watch its status).

**Checklist**
- [x] Subject and description added to both, commit and PR.
- [x] Relevant issues have been referenced.
- [ ] This change includes docs. 
- [ ] This change includes unit tests.

PR #1913, opened 2022-12-01
Add ValidIdentityProvider condition

enxebre:
There's scenarios where the identity provider might have been conflicted, e.g. delete by accident. In that case the HC is not able to operate successfully e.g. aws endpoints, ec2 machines... and there's no clear signal in the HC. This is also true if the HC is in deleting state, which will hang forever with no clear signal.

This PR tries to allivites those scenarios by capturing them in a condition.

**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, use `fixes #<issue_number>(, fixes #<issue_number>, ...)` format, where issue_number might be a GitHub issue, or a Jira story*:
Ref [HOSTEDCP-660](https://issues.redhat.com/browse/HOSTEDCP-660)

**Checklist**
- [ ] Subject and description added to both, commit and PR.
- [ ] Relevant issues have been referenced.
- [ ] This change includes docs. 
- [ ] This change includes unit tests.

PR #5751, opened 2025-03-03
OCPBUGS-52223: refactor aws identity health check into new controller

sjenning:
The cloud provider health check runs every time through the HCP reconcile loop and result in ~800k calls to `DescribeVpcEndpoints` per day in our CI.  This is 25-30% of our total AWS API call volume in our CI account and is contributing to API throttling.

xref
Add identity provider health check https://github.com/openshift/hypershift/pull/1913
Requeue HCP always https://github.com/openshift/hypershift/pull/2408

This PR refactors the health check into its own controller to achieve two goals:
1) Do health checks periodically, even HCP is not changing.  This is already done because with requeue the HCP as a hack to deal with unwatchable changes i.e. AWS LB stuff, however, I think we want to refactor this as well in a follow-on PR so we can stop requeuing the HCP on reconcile success.
2) Aggressively limit how often the health check is done in both the failure and success cases (30s for failure, 5m for success)

We could refactor this in a different way where the code remains in the HCP reconciler, but we just record a `lastAWSIdentityCheckTime` and skip if the reconcile was triggered in less time than some `delayAWSIdentityCheckDuration`.  This only works if we always requeue the HCP, even on reconcile success, which we currently do.

### DIFF
--- a/control-plane-operator/controllers/healthcheck/healthcheck_controller.go
+++ b/control-plane-operator/controllers/healthcheck/healthcheck_controller.go
@@ -1,0 +1,100 @@
+package healthcheck
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/go-logr/logr"
+)
+
+const (
+	successRequeueInterval = 5 * time.Minute
+	failureRequeueInterval = 30 * time.Second
+)
+
+type HealthCheckUpdater struct {
+	client.Client
+	HostedControlPlane client.ObjectKey
+
+	log logr.Logger
+}
+
+func (hcu *HealthCheckUpdater) SetupWithManager(mgr ctrl.Manager) error {
+	return mgr.Add(hcu)
+}
+
+func (hcu *HealthCheckUpdater) Start(ctx context.Context) error {
+	hcu.log = ctrl.LoggerFrom(ctx).WithName("health-check-updater")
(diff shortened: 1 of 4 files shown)


control-plane-operator/controllers/healthcheck/healthcheck_controller.go
```diff
@@ -0,0 +1,94 @@
...
+
+	hostedControlPlane := &hyperv1.HostedControlPlane{}
+	err := hcu.Client.Get(ctx, hcu.HostedControlPlane, hostedControlPlane)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil
+		}
+		return fmt.Errorf("failed to get HostedControlPlane: %w", err)
+	}
+	originalHostedControlPlane := hostedControlPlane.DeepCopy()
```
muraee, line 68: should we abort here if the HCP is deleting?

sjenning: yes we should.  it wouldn't really cause an issue but would be messy on teardown (unable to update resource that is being deleted)

control-plane-operator/controllers/healthcheck/healthcheck_controller.go
```diff
@@ -0,0 +1,100 @@
...
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil
+		}
+		return fmt.Errorf("failed to get HostedControlPlane: %w", err)
+	}
+
+	// Return early if deleting
+	if !hostedControlPlane.DeletionTimestamp.IsZero() {
+		return nil
```
muraee, line 71: should we also stop the ticker here?

sjenning: I don't think so.  It is true that there is no point in running anymore if the HCP is being deleted, however, if we return nil, we wait `successRequeueInterval` before retrying.  If the HCP is being deleted, we likely won't survive that long.  `Start()` `ctx` will be cancelled as the CPO is terminated and `ticker.Stop()` will be called in the `defer`.

sjenning:

sjenning:

PR #5765, opened 2025-03-04
tekton: fix coverity migration

sjenning:
Required migration was not done in https://github.com/openshift/hypershift/pull/5747

https://github.com/konflux-ci/build-definitions/blob/main/task/sast-coverity-check-oci-ta/0.2/MIGRATION.md#action-from-users


sjenning:
Done in https://github.com/openshift/hypershift/pull/5755

PR #5767, opened 2025-03-05
OCPBUGS-52331: CPO v2: rollout workloads on any changes to mounted configmaps/secrets

csrwng:
Removes the RolloutOnSecretChange and RolloutOnConfigMapChange functions in favor of taking all configmaps and secrets into consideration when computing a config hash for a given component.

Fixes reconciliation of service account kubeconfig secrets by reading existing values before the adapt function so we don't re-generate a certificate with every reconcile.

Fixes #[OCPBUGS-52331](https://issues.redhat.com/browse/OCPBUGS-52331)

**Checklist**

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/v2/kas/component.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/v2/kas/component.go
@@ -101,14 +101,13 @@ func NewComponent() component.ControlPlaneComponent {
 			"aws-pod-identity-webhook-kubeconfig.yaml",
 			component.EnableForPlatform(hyperv1.AWSPlatform),
 			component.WithAdaptFunction(adaptAWSPodIdentityWebhookKubeconfigSecret),
+			component.ReconcileExisting(),
 		).
 		WithManifestAdapter(
 			"azure-kms-secretprovider.yaml",
 			component.WithAdaptFunction(kms.AdaptAzureSecretProvider),
 			component.WithPredicate(enableAzureKMSSecretProvider),
 		).
-		RolloutOnConfigMapChange("kas-config", "kas-audit-config", "auth-config").
-		RolloutOnSecretChange("kas-secret-encryption-config").
 		Build()
 }
 
--- a/support/controlplane-component/controlplane-component.go
+++ b/support/controlplane-component/controlplane-component.go
@@ -58,6 +58,9 @@ type ControlPlaneContext struct {
 
 	// This is needed for the generic unit test, so we can always generate a fixture for the components deployment/statefulset.
 	SkipPredicate bool
+
+	// SkipCertificateSigning is used for the generic unit test to skip the signing of certificates and maintain a stable output.
+	SkipCertificateSigning bool
 }
 
 type WorkloadContext struct {
@@ -74,6 +77,9 @@ type WorkloadContext struct {
 	SetDefaultSecurityContext bool
 	EnableCIDebugOutput       bool
 	MetricsSet                metrics.MetricsSet
+
+	// skip generation of certificates for unit tests
+	SkipCertificateSigning bool
 }
 
 func (cp *ControlPlaneContext) workloadContext() WorkloadContext {
@@ -88,6 +94,7 @@ func (cp *ControlPlaneContext) workloadContext() WorkloadContext {
 		EnableCIDebugOutput:       cp.EnableCIDebugOutput,
 		MetricsSet:                cp.MetricsSet,
 		ImageMetadataProvider:     cp.ImageMetadataProvider,
+		SkipCertificateSigning:    cp.SkipCertificateSigning,
 	}
 }
 
@@ -116,9 +123,6 @@ type controlPlaneWorkload[T client.Object] struct {
 	manifestsAdapters map[string]genericAdapter
 	// predicate is called at the beginning, the component is disabled if it returns false.
 	predicate func(cpContext WorkloadContext) (bool, error)
-	// These secrets/configMaps will cause the Deployment/statefulset to rollout when changed.
-	rolloutSecretsNames    []string
-	rolloutConfigMapsNames []string
 
 	// if provided, konnectivity proxy container and required volumes will be injected into the deployment/statefulset.
 	konnectivityContainerOpts *KonnectivityContainerOptions
@@ -136,7 +140,7 @@ func (c *controlPlaneWorkload[T]) Name() string {
 	return c.name
 }
 
-// reconcile implements ControlPlaneComponent.
+// Reconcile implements ControlPlaneComponent.
 func (c *controlPlaneWorkload[T]) Reconcile(cpContext ControlPlaneContext) error {
 	workloadContext := cpContext.workloadContext()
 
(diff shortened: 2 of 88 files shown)


support/controlplane-component/generic-adapter.go
```diff
@@ -31,6 +32,12 @@ func WithPredicate(predicate Predicate) option {
 	}
 }
 
+func ReconcileExisting() option {
```
enxebre, line 35: please let's document this exported function and/or the field

support/controlplane-component/defaults.go
```diff
@@ -101,62 +101,101 @@ func (c *controlPlaneWorkload[T]) defaultOptions(cpContext ControlPlaneContext,
...
-		configMap := &corev1.ConfigMap{
-			ObjectMeta: metav1.ObjectMeta{
-				Name:      configMapName,
-				Namespace: cpContext.HCP.Namespace,
-			},
-		}
-		if err := cpContext.Client.Get(cpContext, client.ObjectKeyFromObject(configMap), configMap); err != nil {
-			if apierrors.IsNotFound(err) {
-				continue
+func podSecretNames(spec *corev1.PodSpec) []string {
```
muraee, line 121: should we pass a client to this function and make it just return a list of secrets directly, instead of having to create a generic fetchResource.

support/controlplane-component/defaults.go
```diff
@@ -101,62 +101,101 @@ func (c *controlPlaneWorkload[T]) defaultOptions(cpContext ControlPlaneContext,
...
+	podTemplate.Annotations["component.hypershift.openshift.io/config-hash"] = hashString
 	return nil
 }
 
+func computeResourceHash(secretNames, configMapNames []string,
+	fetchSecret func(string) (*corev1.Secret, error),
+	fetchConfigMap func(string) (*corev1.ConfigMap, error),
+) (string, error) {
+	slices.Sort(secretNames)
+	slices.Sort(configMapNames)
```
muraee, line 174: We don't really need to sort here, sorting the hashes at the end `slices.Sort(hashes)` should be sufficient, right?

support/controlplane-component/defaults_test.go
```diff
@@ -0,0 +1,60 @@
...
+				},
+				func(name string) (*corev1.ConfigMap, error) {
+					return configMapData[name], nil
+				})
+			g.Expect(err).ToNot(HaveOccurred())
+			if hashValue == "" {
+				hashValue = result
+				break
+			}
+			g.Expect(result).To(Equal(hashValue), "Hash value must remain the same for the same data")
```
enxebre, line 31: can you help me grok how this expectation is true?

support/controlplane-component/controlplane-component.go
```diff
@@ -16,6 +16,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
```
muraee, line 19: duplicate import, already imported as `apierrors`

support/controlplane-component/defaults_test.go
csrwng: The test is to ensure that the computed hash remains consistent through several calls to computeResourceHash, so we fuzz a set of inputs once, and then compute the hash several times. Every time we compute it, it should remain the same.

support/controlplane-component/defaults.go
csrwng: The reason I split them is that it should be easier to test whether we're returning the right resources given a podSpec without involving a client. The intent was more/easier ut, which reminds me that I didn't include UTs for this, will add.

support/controlplane-component/defaults.go
csrwng: removed

support/controlplane-component/defaults_test.go
```diff
@@ -0,0 +1,204 @@
...
+			},
+			expect: []string{"pcm1", "pcm2"},
+		},
+		{
+			name: "volumes and projected",
+			podSpec: corev1.PodSpec{
+				Volumes: []corev1.Volume{cmVolume("cm1"), cmVolume("cm2"), projectedCMVolume("pcms", "pcm3", "pcm4"), secretVolume("s1")},
+			},
+			expect: []string{"cm1", "cm2", "pcm3", "pcm4"},
+		},
```
devguyio, line 104: can we also add the case of empty volumes and a projectedVolume that's not neither a ConfigMap nor a Secret to verify that we skip it?

devguyio (changes requested):
nice fix! I left a few comments and questions. 

I think this PR needs an e2e fix that would fail without this fix. e.g. a test that verifies that an existing SA kubeconfig doesn't get regenrated  after reconciliation

enxebre:
>Fixes reconciliation of service account kubeconfig secrets by reading existing values before the adapt function so we don't re-generate a certificate with every reconcile.

For this, can we add a immutability case for testReconcile?

>Removes the RolloutOnSecretChange and RolloutOnConfigMapChange functions in favor of taking all configmaps and secrets into consideration when computing a config hash for a given component.

This changes behaviour from current cpo, right? If so I'd rather postpone this change until the switch is complete

lgtm overall pending others comments

support/controlplane-component/defaults.go
muraee: ack

support/controlplane-component/defaults.go
```diff
@@ -61,6 +57,10 @@ func (c *controlPlaneWorkload[T]) defaultOptions(cpContext ControlPlaneContext,
 		c.konnectivityContainerOpts.injectKonnectivityContainer(cpContext, &podTemplateSpec.Spec)
 	}
 
+	if err := c.applyWatchedResourcesAnnotation(cpContext, podTemplateSpec); err != nil {
```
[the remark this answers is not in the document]
csrwng: because I want the hash to take into account all mounted configmaps and secrets, and injecting the konnectivity container potentially adds more volumes to the pod

csrwng:
> This changes behaviour from current cpo, right? If so I'd rather postpone this change until the switch is complete

It doesn't. It only impacts the CPO v2.

enxebre:
>It doesn't. It only impacts the CPO v2.

Yeh that's what I mean, It deviates from current cpo behaviour. It'd be good to minimize cpov2 new improvements over existing cpo behaviour until the migration is done and cpov2 is fully e2e tested. Otherwise, we increase the risk of going down a rabbit whole by conflating both migration of current state with new improvements. Not a blocker to me anyways.

devguyio (approved):

csrwng:

control-plane-operator/controllers/hostedcontrolplane/v2/kas/component.go
```diff
@@ -101,14 +101,16 @@ func NewComponent() component.ControlPlaneComponent {
...
 		).
 		WithManifestAdapter(
 			"azure-kms-secretprovider.yaml",
 			component.WithAdaptFunction(kms.AdaptAzureSecretProvider),
 			component.WithPredicate(enableAzureKMSSecretProvider),
 		).
-		RolloutOnConfigMapChange("kas-config", "kas-audit-config", "auth-config").
-		RolloutOnSecretChange("kas-secret-encryption-config").
+		ExcludeConfigMapsFromConfigHash(
+			"client-ca",
```
muraee, line 112: I don't think the individual components should decide whether to exclude this configMap or not, I think we should just always exclude it as it will cause any component to keep restarting.  wdyt?

csrwng: sgtm, will update

csrwng:
Will retest when https://github.com/openshift/hypershift/pull/5860 merges

PR #5784, opened 2025-03-07
CNTRLPLANE-267: Restructure the OWNERS usage in the whole repo

devguyio:

Restructre the usage of OWNERS in order to leverage Prow's auto assigment of approvers and reviewers as well as to streamline the management of OWNERS.

Restructure the usage of OWNERS to leverage Prow's auto-assignment of approvers and reviewers, as well as streamline the management of OWNERS.

- Move to a single OWNERS file with multiple filters instead of multiple OWNERS files.
- Introduce OWNERS_ALIASES
- Introduce emeritus_approvers

## Why?

Currently, prow's auto-assignment of approvers and reviewers is totally ignored. It can include non-core members and so there's no confidence in the assignment process.

However, by having a well maintained OWNERS lists that reflects the reality, we can improve the reviewing experience significantly by leveraging prow's auto-assignment.

This way we can achieve:

1. Fair distribution of code reviews burden.
1. Faster code reviews by leveraging GitHub notifications for being assigned.
1. Spreading the knowledge by spreading reviews across the team.
1. Reducing the noise in the team's slack channel by reducing the necessity to ping folks for reviews over and over.

If adopted, the expectation would be whoever is assigned can either review the PR or change the assignment to the correct person who should review it.

In short, PR reviewers assignment will be the signaling mechanism between team members.

Fixes #CNTRLPLANE-267

## Checklist


OWNERS
```diff
@@ -1,16 +1,72 @@
...
+  "cmd":
+    labels:
+    - area/cli
+  "control-plane-operator":
+    labels:
+    - area/control-plane-operator
+  "control-plane-pki-operator":
+    labels:
+    - area/control-plane-pki-operator
+    approvers:
```
bryan-cox, line 37: Why does this section have tech leads specifically called out?

OWNERS_ALIASES
```diff
@@ -0,0 +1,30 @@
...
+    - muraee
+    - sdminonne
+    - sjenning
+  kubevirt-approvers:
+    - davidvossel
+    - nirarg
+  kubevirt-reviewers:
+    - davidvossel
+    - nirarg
+  ibm-reviewers:
```
bryan-cox, line 28: We might want to check to see if they want anyone else added.

OWNERS_ALIASES
```diff
@@ -0,0 +1,30 @@
...
+    - bryan-cox
+    - csrwng
+    - devguyio
+    - enxebre
+    - jparrill
+    - muraee
+    - sdminonne
+    - sjenning
+  kubevirt-approvers:
+    - davidvossel
```
bryan-cox, line 23: I don't think David should be included anymore. I thought he left kubevirt. I think Oren should be added perhaps?

Same comment for reviewers below.

OWNERS_ALIASES
rtheis: We do but can add them in a follow-up PR.

OWNERS
```diff
@@ -1,16 +1,72 @@
...
-- hasueki
-- rtheis
+filters:
+  ".*":
+    approvers:
+    - core-approvers
+    reviewers:
+    - core-reviewers
+    emeritus-approvers:
+    - davidvossel
```
jparrill, line 8: David is not in Kubevirt anymore, so you will need to add Oren Cohen for that matters.

OWNERS
```diff
@@ -1,16 +1,72 @@
...
+  # KubeVirt specific OWNERS
```
jparrill, line 9: I think we will need another 2 sections:
- For Openstack provider where Emilien Macchi.
- For Agent provider Nick Carboni and Crystal.
- For IBM Hide could be the choice.

WDYT?

bryan-cox: Might check with Emilien about who else can be added to OpenStack. Matt Booth may be added for instance.

EmilienM: @stephenfin @mdbooth at least.

Thanks

devguyio: > I think we will need another 2 sections for reviewers:
> 
>     * For Openstack provider where Emilien Macchi.
> 
>     * For Agent provider Nick Carboni and Crystal.
> 
>     * For IBM Hide or/and Richard could be the choices.
> 
> 
> WDYT?

sure! however let's do that in a separate PR for two reasons:
1- this PR was mainly trying to reflect the current state.
2- (and mostly this) because I'm still testing the best way to tackle the OWNERS for the non-core teams, and there's a possibility we'll need to change it to nested OWNERS files again. 

So closely after this PR gets merged, I'll test and see if this setup works or we need nested OWNERS files, then I'll create the relevant PRs introducing the needed changes and those three suggestions you mentioned.

bryan-cox: @EmilienM moved to a different team so he's no longer on OpenStack.

OWNERS
devguyio: it's how it is now, the OWNERS file under the `control-plane-pki-operator` has @sjenning , @csrwng and @enxebre as approvers. So they'd be preferred when suggested as approvers.

devguyio:
This is now updated and ready for review

enxebre:
verify complains on commit conventions lgtm otherwise
/approve

rtheis (approved):
/lgtm

PR #5793, opened 2025-03-07
CNTRLPLANE-111: Fix KMS MIv3 Issues

bryan-cox:
This PR fixes a few issues when KMS was converted over to MIv3:
- adds omitempty to ClientID and CertificatName in Azure API since these fields are no longer required in managed identity version 3
- renames the KMSUserAssignedCredsFile flag to KMSUserAssignedCredsSecretName. This more accurately represents the   flag's purpose, which is to supply the HC with the name of the secret in Azure Key Vault containing the managed identity needed when using KMS.

A part of [CNTRLPLANE-111](https://issues.redhat.com//browse/CNTRLPLANE-111)

**Checklist**

### DIFF
--- a/api/hypershift/v1beta1/azure.go
+++ b/api/hypershift/v1beta1/azure.go
@@ -465,15 +465,17 @@ type ManagedIdentity struct {
 	// Deprecated: This field was previously required as part of the MIWI phase 2 work; however, this field will be
 	// removed as part of the MIWI phase 3 work, https://issues.redhat.com/browse/OCPSTRAT-1856.
 	//
+	// +optional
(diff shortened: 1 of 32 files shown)


api/hypershift/v1beta1/azure.go
```diff
@@ -466,14 +466,14 @@ type ManagedIdentity struct {
 	// removed as part of the MIWI phase 3 work, https://issues.redhat.com/browse/OCPSTRAT-1856.
 	//
 	// +kubebuilder:validation:XValidation:rule="self.matches('^[{]?[0-9a-fA-F]{8}-([0-9a-fA-F]{4}-){3}[0-9a-fA-F]{12}[}]?$')",message="the client ID of a managed identity must be a valid UUID. It should be 5 groups of hyphen separated hexadecimal characters in the form 8-4-4-4-12."
-	ClientID string `json:"clientID"`
+	ClientID string `json:"clientID,omitempty"`
```
muraee, line 469: should we mark as `+optional`?

PR #5795, opened 2025-03-07
OCPBUGS-52655: Add flag to set etcd storage size

bryan-cox:
This PR adds a flag to the HyperShift CLI allowing the etcd storage size to be set during cluster creation.

Fixes OCPBUGS-52655

**Checklist**

### DIFF
--- a/cmd/cluster/core/create.go
+++ b/cmd/cluster/core/create.go
@@ -27,6 +27,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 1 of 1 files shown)


cmd/cluster/core/create.go
```diff
@@ -330,6 +333,11 @@ func prototypeResources(opts *CreateOptions) (*resources, error) {
 		prototype.Cluster.Spec.Etcd.Managed.Storage.PersistentVolume.StorageClassName = ptr.To(opts.EtcdStorageClass)
 	}
 
+	if opts.EtcdStorageSize != "" {
+		etcdStorageSize := resource.MustParse(opts.EtcdStorageSize)
```
muraee, line 337: MustParse will panic if the value is invalid, can we please return an error here instead.

PR #5827, opened 2025-03-13
OCPBUGS-53050: Handle multiple mirror entries for source

LiangquanLi930:
According to OCP IDMS, when the current mirror image is unavailable, the system should continue searching for an available mirror image.

Fixes # https://issues.redhat.com/browse/OCPBUGS-53050

**Checklist**

### DIFF
--- a/support/releaseinfo/registry_image_content_policies.go
+++ b/support/releaseinfo/registry_image_content_policies.go
@@ -5,6 +5,8 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/openshift/hypershift/support/releaseinfo/registryclient"
+
 	ctrl "sigs.k8s.io/controller-runtime"
 )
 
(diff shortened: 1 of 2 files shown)


xiuwang:
Pre-merge test this PR pass, hosted cluster could hold multiple mirrors, when the first place is failed to pull, will fall down to second place in order....  and so on.
Test details in https://issues.redhat.com/browse/OCPBUGS-53050?focusedId=26781151&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-26781151

support/releaseinfo/registry_image_content_policies.go
```diff
@@ -32,8 +34,12 @@ func (p *ProviderWithOpenShiftImageRegistryOverridesDecorator) Lookup(ctx contex
...
-					return releaseImage, nil
+					// Verify mirror image availability.
+					if _, _, err = registryclient.GetRepoSetup(ctx, image, pullSecret); err == nil {
+						p.mirroredReleaseImage = image
+						return releaseImage, nil
+					}
+					continue
 				}
 
 				logger.Error(err, "Failed to look up release image using registry mirror", "registry mirror", registryReplacement)
```
bryan-cox, line 45: Should this be a warning instead since we are potentially looking through lots of mirrors?

jparrill (commented):
LGTM overall, I see 2 issues in the approach we are following (maybe we can address them in a separated PR):

- The unit tests should not try to pull a real image (I did that mistake adding those to the unit tests initially), I think this should be part of the integration tests. Maybe we could have a response json in some place and using it as a mocked response from quay.
- We need to cover this scenario in Integration/E2E to avoid gaps into the normal disconnected workflow

LiangquanLi930:
/jira backport release-4.18,release-4.17,release-4.16,release-4.15,release-4.14

PR #5831, opened 2025-03-13
OCPBUGS-50562: Sync RBAC for attaching volumes on VM level

akalenyu:
This is basically about adapting RBAC for this CSI driver fork PR:
https://github.com/openshift/kubevirt-csi-driver/pull/53
The RBAC is hardcoded here and thus needs to be changed as well.

Fixes #

**Checklist**


akalenyu:
/hold
to make sure ordering is right (and that this makes sense)

akalenyu:
> Only question I have at this point since we switched from VMIs to VMs, do we still need the RBAC to the VMI endpoints for addvolume and removevolume?

Since this goes in before the actual changes (which may or may not be inverted dependency) I opted for the safe side of keeping both

awels:
Okay but at some point we need to remove that RBAC.
/lgtm

cmd/install/assets/hypershift_operator.go
```diff
@@ -1183,6 +1183,11 @@ func (o HyperShiftOperatorClusterRole) Build() *rbacv1.ClusterRole {
 				Resources: []string{"virtualmachineinstances/addvolume", "virtualmachineinstances/removevolume"},
 				Verbs:     []string{rbacv1.VerbAll},
 			},
+			{ // This allows the kubevirt csi driver to hotplug volumes to KubeVirt VMs.
+				APIGroups: []string{"subresources.kubevirt.io"},
+				Resources: []string{"virtualmachines/addvolume", "virtualmachines/removevolume"},
+				Verbs:     []string{rbacv1.VerbAll},
```
bryan-cox, line 1189: Does this really need all or is get just fine?

akalenyu: I am pretty sure it only needs "update", good call. Not sure why it uses * prior to this

awels (approved):

bryan-cox:
/overrride "Red Hat Konflux / hypershift-operator-main-enterprise-contract / hypershift-operator-main"

bryan-cox:
/overrride "Red Hat Konflux / hypershift-operator-main-enterprise-contract / hypershift-operator-main"

bryan-cox:

bryan-cox:

bryan-cox:
/override "Red Hat Konflux / control-plane-operator-enterprise-contract / pr group"

PR #5840, opened 2025-03-14
WIP: CNTRLPLANE-334: Update KAS structured authentication configuration generation logic to use new `uid` and `extra` field

everettraven:
- Updates the KAS structured authentication generation logic to use the new `uid` and `extra` claim mapping fields introduced to the `authentications.config.openshift.io` CRD in https://github.com/openshift/api/pull/2234

Fixes #

**Checklist**


everettraven:
It is expected that all tests will fail as this is consuming unmerged changes to the openshift/api dependency.

rtheis:
@jonesbr17 ptal

everettraven:
/testwith openshift/hypershift/main/e2e-aws openshift/api#2234

everettraven:
/testwith openshift/hypershift/main/e2e-aws openshift/api#2234

everettraven:
/testwith openshift/hypershift/main/e2e-aws https://github.com/openshift/api/pull/2234

everettraven:
/payload-with-prs 4.19 ci blocking https://github.com/openshift/api/pull/2234

cmd/install/assets/hypershift-operator/zz_generated.crd-manifests/hostedclusters-Default.crd.yaml
```diff
@@ -618,243 +618,6 @@ spec:
                         required:
                         - name
                         type: object
-                      oidcProviders:
```
everettraven, line 621: This seems wrong. Digging.

control-plane-operator/controllers/hostedcontrolplane/kas/auth.go
```diff
@@ -48,50 +53,255 @@ func generateAuthConfig(spec *configv1.AuthenticationSpec, ctx context.Context,
...
+	if err != nil {
+		return out, fmt.Errorf("generating username claim mapping: %v", err)
+	}
+
+	groups := generateGroupsClaimMapping(claimMappings.Groups)
+
+	out.Username = username
+	out.Groups = groups
+
+	if featuregate.Gates.Enabled(featuregate.ExternalOIDCWithUIDAndExtraClaimMappings) {
```
everettraven, line 138: TODO: pending https://github.com/openshift/hypershift/pull/5976 merging, will need to update how the feature gates are being used here

control-plane-operator/controllers/hostedcontrolplane/kas/auth.go
```diff
@@ -3,95 +3,318 @@ package kas
...
-	authConfig, err := generateAuthConfig(p.Authentication, ctx, c, config.Namespace)
+
+	authConfig, err := GenerateAuthConfig(p.Authentication, ctx, c, config.Namespace)
 	if err != nil {
 		return fmt.Errorf("failed to generate authentication config: %w", err)
 	}
+
+	// TODO: using the default compiler means that we are allowing CEL library usage for the Kube version that maps to our
+	// dependency import. This should align with the version of Kubernetes that will be running for a guest cluster instead
+	// since that is what will actually load and validate the configuration.
```
everettraven, line 42: 2 lines of thought here:
1. We leave this as-is. This means we validate against progressively newer validations every time we bump our kube dependencies. Should _generally_ be safe as kube tries to avoid breaking changes and is heavily concerned with backwards compatibility. However, we may end up with false-positives where a CEL library is available in the version we are using for our dependency, but not available on the KAS being deployed. Seems like a non-starter, but sharing my train of thought here for others.
2. We implement a system to map an OCP version --> Kubernetes version. This seems like the only reasonable way to really approach this. In practice, I'm not sure how one would extract this information from the `HostedControlPlane` resource as it doesn't seem to include the OCP release version it is using in the spec anywhere. Doing some digging on this.

everettraven:
Closing in favor of #6073

PR #5849, opened 2025-03-16
NOJIRA: openstack: ORC won't delete the image after HostedCluster deletion

EmilienM:

We have introduced a new feature where ORC handles the lifecycle of
OpenStack Glance images for the release image from OpenShift mirrors.
By default, we were deleting the image when the HostedCluster was
deleted.
But this will be problematic when a customer deploys more than one
HostedCluster, and then after the deletion of one of them the image
would have been destroyed, which would be problematic for the other
HostedCluster where a Nodepools couldn't be scaled up for example.

It is safer to orphan the images (with proper documentation that they
are orphaned) and provide an option to let them follow the
HostedCluster's lifecycle when desired (also documented).

The default in ORC is to delete the image once the ORC object is deleted
but this would cause problems for us in Hypershift, so we don't follow
that default.

This patch updates API, code, tests and docs to address that change.

### DIFF
--- a/api/hypershift/v1beta1/openstack.go
+++ b/api/hypershift/v1beta1/openstack.go
@@ -17,7 +17,7 @@ const (
 
 	// PruneRetentionPolicy is the default policy for handling OpenStack Glance Images
 	// when the HostedCluster is deleted.
-	DefaultImageRetentionPolicy RetentionPolicy = PruneRetentionPolicy
+	DefaultImageRetentionPolicy RetentionPolicy = OrphanRetentionPolicy
 )
 
 func (p *RetentionPolicy) String() string {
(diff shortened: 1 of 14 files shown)


mdbooth:
The implication of this PR is that you're expecting multiple HostedClusters to use the same glance images. They shouldn't be doing that. It is not safe for a resource to have multiple owners. A glance image should be owned by the HostedCluster which created it, and not used by any other HostedCluster.

The 'adopt with the same name' hack is... a hack. It works for CI, but it's not something we want to recommend to customers, and honestly I expect ORC to break it.

I don't believe we should do this. If this is a problem we should fix the root cause, which is that production HostedClusters are sharing glance images.

EmilienM:
/close
We've discussed this with Matt and we're going to solve it step by step.

PR #5853, opened 2025-03-17
NO-JIRA: openstack: revert `OpenStackImageRetentionPolicy` + use a unique image name per Hosted Cluster

EmilienM:

- Revert of `OpenStackImageRetentionPolicy` for now. We have identified that the current code that was implemented isn't suitable for production.
  We want to re-work it but in the meantime we just revert it.
- Added `ClusterImageName` function to encapsulate image name generation for a HostedCluster.
- Updated `reconcileOpenStackImageCR` and `ReconcileOpenStackImageSpec`
  to use `ClusterImageName` instead of manually constructing image names.
- Use an image name that now contains the Hosted Cluster name in it so
  it's unique and per Hosted Cluster. The OpenStack Glance image won't
  be shared across Hosted Clusters and its lifecycle will be the same as
  the HostedCluster using it (by default).

### DIFF
--- a/hypershift-operator/controllers/nodepool/openstack.go
+++ b/hypershift-operator/controllers/nodepool/openstack.go
@@ -74,7 +74,7 @@ func (r *NodePoolReconciler) reconcileOpenStackImageCR(ctx context.Context, clie
 			OwnerReferences: []metav1.OwnerReference{
 				{
 					// Since there is no code that deletes the ORC image object, the only way the OpenStack Glance image
-					// can be deleted is when the OpenStackCluster CR is deleted and that the imageRetentionPolicy is set to "Prune".
+					// can be deleted is when the OpenStackCluster CR is deleted.
 					// That means Nodepools can share the same image and changing the image of a Nodepool will not affect the other Nodepools.
 					APIVersion: openstackCluster.APIVersion,
 					Kind:       openstackCluster.Kind,
--- a/api/hypershift/v1beta1/openstack.go
+++ b/api/hypershift/v1beta1/openstack.go
@@ -1,7 +1,5 @@
 package v1beta1
 
-import "fmt"
-
 // PortSecurityPolicy defines whether or not to enable port security on a port.
 type PortSecurityPolicy string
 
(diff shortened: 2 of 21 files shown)


hypershift-operator/controllers/nodepool/openstack.go
```diff
@@ -57,16 +57,15 @@ func (r *NodePoolReconciler) setOpenStackConditions(ctx context.Context, nodePoo
...
 func (r *NodePoolReconciler) reconcileOpenStackImageCR(ctx context.Context, client client.Client, hcluster *hyperv1.HostedCluster, release *releaseinfo.ReleaseImage, nodePool *hyperv1.NodePool) (string, error) {
-	releaseVersion, err := openstack.OpenStackReleaseImage(release)
+	controlPlaneNamespace := fmt.Sprintf("%s-%s", nodePool.Namespace, strings.ReplaceAll(nodePool.Spec.ClusterName, ".", "-"))
+	openstackCluster, err := openstack.GetOpenStackClusterForHostedCluster(ctx, client, hcluster, controlPlaneNamespace)
 	if err != nil {
 		return "", err
 	}
-	controlPlaneNamespace := fmt.Sprintf("%s-%s", nodePool.Namespace, strings.ReplaceAll(nodePool.Spec.ClusterName, ".", "-"))
-	openstackCluster, err := openstack.GetOpenStackClusterForHostedCluster(ctx, client, hcluster, controlPlaneNamespace)
+	imageName, err := openstack.ClusterImageName(hcluster, release)
```
mdbooth, line 65: You don't actually need to change `imageName` here. You can leave it as before. This is the name of the ORC image, which is already namespaced (because it's literally in a kubernetes namespace).

ReconcileOpenStackImageSpec is setting the name of the glance image, which is what needs to be prefixed with the hosted cluster.

This is ok, I guess, if you think it's simpler. It just struck me as weird that we were calling ClusterImageName() twice.

EmilienM: But we could have multiple Hosted Clusters within the same namespace, which is why I added it here.

EmilienM: Since the ORC resource is in the HC namespace, we can use the previous name, gotcha.

mdbooth (commented):
/lgtm

PR #5854, opened 2025-03-17
CNTRLPLANE-331: Add string fields for CVO status AvailableUpdates and…

tjungblu:
… ConditionalUpdates

To work around the following two shortcomings of ACM:
* ACM-18883 Enable RawFeedbackJsonString by default
* ACM-18884 Increase/Configure max limit on ManifestWork rawjson type

... we agreed to create another field in the HCP status to write a single string of all available and conditional updates.

This is due to the fact that the the ManifestWorks CRD has a limit of 1024 characters when using rawJSON as a feedback type, which we would need when we would like to get all available updates for a given HostedCluster. In addition to that limit, a feature flag needs to be enabled by ACM to allow the rawJSON as a feedback type in the first place.

It is likely that changing the defaults will take an undefined and long enough time and thus we are unblocking ourselves by adding two dedicated new status fields we can directly query.


api/hypershift/v1beta1/hosted_controlplane.go
```diff
@@ -302,6 +302,18 @@ type HostedControlPlaneStatus struct {
 	// +optional
 	ReleaseImage string `json:"releaseImage,omitempty"`
 
+	// AvailableUpdateVersions is a comma separated list of all available update versions.
+	//
+	// Deprecated: Use versionStatus.availableUpdates instead.
+	// +optional
+	AvailableUpdateVersions string `json:"availableUpdateVersions,omitempty"`
```
wking, line 309: New structured properties seems heavy for a throw-away property we want to drop later?  Could we use a custom `type` for `status.conditions[]` to set a string without getting tied up in CRD structure?

api/hypershift/v1beta1/hosted_controlplane.go
```diff
@@ -302,6 +302,18 @@ type HostedControlPlaneStatus struct {
...
+
+	// ConditionalUpdateVersions is a comma separated list of all conditional update versions.
+	//
+	// Deprecated: Use versionStatus.conditionalUpdates instead.
+	// +optional
+	ConditionalUpdateVersions string `json:"conditionalUpdateVersions,omitempty"`
```
wking, line 315: Conditional update version strings are probably not all that useful without knowing which of the risks associated with each target the cluster-version operator was concerned about.  Maybe just focus on `availableUpdates` for this hopefully-short-term workaround?

control-plane-operator/hostedclusterconfigoperator/controllers/hcpstatus/hcpstatus.go
```diff
@@ -166,3 +171,19 @@ func (h *hcpStatusReconciler) reconcile(ctx context.Context, hcp *hyperv1.Hosted
 
 	return nil
 }
+
+func compileUpdateVersions(v []configv1.Release) string {
+	var versions []string
+	for _, release := range v {
+		versions = append(versions, release.Version)
+	}
+	return strings.Join(versions, ",")
```
wking, line 180: Checking to see how this works for [an old 4.12.16 cluster][1] with lots of newer 4.12 and 4.13 patch updates:

```console
$ curl -sL https://raw.githubusercontent.com/openshift/oc/fe18ae25a123e896fd218210073141f7f5811465/pkg/cli/admin/upgrade/recommend/examples/4.12.16-longest-recommended-cv.yaml | yaml2json | jq -r '[.status.availableUpdates[].version] | join(",")' | wc -c
576
```

That ClusterVersion snapshot is months old now, so also testing vs. current Cincinnati:

```console
cincinnati-graph-data $ hack/show-edges.py --cincinnati https://api.openshift.com/api/upgrades_info/graph stable-4.13 | grep '^4[.]12[.]16 ' | sed 's/.*> //' | tr '\n' ',' | sed 's/,$//' | wc -c
822
```

So we're getting into the ballpark of 1024 characters, but still have enough headroom that maybe logic like "prefer the latest *n* patches in each `4.y`" won't be necessary.

[1]: https://github.com/openshift/oc/blob/fe18ae25a123e896fd218210073141f7f5811465/pkg/cli/admin/upgrade/recommend/examples/4.12.16-longest-recommended-cv.yaml#L14

api/hypershift/v1beta1/hosted_controlplane.go
tjungblu: adding a new field seems a bit more consistent, if you look at the ones above, which are also created for the purposes of the feedback rules

@sjenning any opinions on this?

control-plane-operator/hostedclusterconfigoperator/controllers/hcpstatus/hcpstatus.go
tjungblu: the funny thing is, there is no kubevalidation on the strings. So you could put gigabytes worth of string data in there ;-) Well, up until the etcd limit per key.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -654,6 +654,17 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
 		}
 	}
 
+	{
+		updatesCondition := &metav1.Condition{
+			Type:               "ClusterVersionAvailableUpdateVersions",
+			Status:             metav1.ConditionTrue,
+			Reason:             hyperv1.AsExpectedReason,
+			Message:            compileUpdateVersions(hcluster.Status.Version.AvailableUpdates),
```
wking, line 662: [`HostedClusterStatus.Version` is a `*ClusterVersionStatus`][1], so you're leaning on `computeClusterVersionStatus` earlier in this function to never return `nil`.  Which [looks safe today][2], but it might be worth being more defensive here and wrapping this block in an `if  hcluster.Status.Version != nil {` just in case?  Because "HostedCluster-controller explodes on a segfault" would be bad, and "HostedCluster-controller doesn't populate `ClusterVersionAvailableUpdateVersions`" would be a much smaller deal.  I guess you could have an `else` block that set this condition `False` with "status.version is nil, probably take a look at the underlying HostedControlPlane's `status` next?" or some similar message?

[1]: https://github.com/openshift/hypershift/blob/ece810c2c0886cb4f708f41b9f89de0522bd8566/api/hypershift/v1beta1/hostedcluster_types.go#L1466-L1471
[2]: https://github.com/openshift/hypershift/blob/ece810c2c0886cb4f708f41b9f89de0522bd8566/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go#L3662-L3741

tjungblu: put the `compileUpdateVersions` parameter one level higher and added nil and empty checks more explicitly in there

wking (approved):
I'm not an approver, but looks good to me :)

/lgtm

sjenning:
/hold

This isn't a great, using the `message` field of a `condition` to pass data and the condition always being `True`.

I would rather add a new field to the `HostedClusterStatus` instead.

sjenning:
Before you add the status field, let's meet about this.  I would prefer no change to hypershift since we already report this information but we are working around limitations in other layers.

Other ideas:
- Annotation on the HC (minimize API impact and can remove in the future when ACM supports this)
- Transform the HC when processing for ManifestWork

tjungblu:
/close

we decided in a meeting today that we're waiting for the ACM tickets to get resolved, no need to duplicate the information in the meantime.

PR #5861, opened 2025-03-18
OCPBUGS-53263: Fixed cpov2 reconciliation making no-op calls to api-server

muraee:

Fixes #

**Checklist**

### DIFF
--- a/support/upsert/apply.go
+++ b/support/upsert/apply.go
@@ -1,0 +1,173 @@
+package upsert
+
+import (
+	"context"
+	"fmt"
+	"maps"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/sets"
+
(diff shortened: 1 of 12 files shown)


support/upsert/apply.go
```diff
@@ -0,0 +1,143 @@
...
+	if annotations == nil {
+		annotations = make(map[string]string)
+	}
+	maps.Copy(annotations, mutated.GetAnnotations())
+	mutated.SetAnnotations(annotations)
+
+	finalizers := sets.New(original.GetFinalizers()...).Insert(mutated.GetFinalizers()...)
+	mutated.SetFinalizers(sets.List(finalizers))
+
+	mutated.SetResourceVersion(original.GetResourceVersion())
```
enxebre, line 116: why is the ResourceVersion not ignored?

muraee: for future reference: This needs to be set and passed back by clients for concurrency checks
ref: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#concurrency-control-and-consistency

enxebre: fwiw how you implement passing that back might be tied to your patching strategy, e.g. I don't think we'd set it if using SSA

muraee: true, I have been looking at the patch code from controller-util and they set this field only when optimistLock is true
see:
https://github.com/kubernetes-sigs/controller-runtime/blob/5dfe3216fb7fd7f5afb61d6d0f8956c7bec8df62/pkg/client/patch.go#L104-L115

sjenning:
failed due to need for forced promotion in ci payload stream

/override ci/prow/e2e-aws

PR #5870, opened 2025-03-19
AUTOSCALE-44: add autonode karpenter drift hcp upgrade e2e test

maxcao13:
This PR adds a new test to the karpenter e2e suite which upgrades the hcp and validates that karpenter nodes are drift and upgraded to the control-plane's new release image version. Also validates that any pods that were already scheduled to the old karpenter node, will be rescheduled to the new karpenter node.

Also this PR lowers the CPU of the tested workload to avoid them being unschedulable because of the small CPU allocation of `t3.large` instances.

Fixes: https://issues.redhat.com/browse/AUTOSCALE-44

**Checklist**

### DIFF
--- a/support/api/scheme.go
+++ b/support/api/scheme.go
@@ -23,7 +23,9 @@ import (
 	agentv1 "github.com/openshift/cluster-api-provider-agent/api/v1beta1"
 
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/runtime/serializer/json"
 	apiserverconfigv1 "k8s.io/apiserver/pkg/apis/apiserver/v1"
 	kasv1beta1 "k8s.io/apiserver/pkg/apis/apiserver/v1beta1"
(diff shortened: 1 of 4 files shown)


maxcao13:
/cc @elmiko @muraee @enxebre

ready for review.

enxebre:
Thanks, this is great! is there any nodepool condition we should be checking as part of the process?

Each new Test* creates a new HC and increase the aws cloud account load which we are trying to avoid. With that in mind, can we instead validate this logic within the existing TestKarpenter?

enxebre:
>AFAIK, karpenter doesn't actually expose any drifted conditions/statuses atm, but only emit drifted events. There exists this PR https://github.com/kubernetes-sigs/karpenter/pull/2090 which would be helpful, but we would need to wait for it to merge and then pull the latest upstream image.
Unless I'm misinterpreting and you are talking about some other conditions we would need to check. WDYT?

nodeClaim resources signal a "drifted" condition with reason nodeclassDrifted you can use as a criteria. We should actually also validate that condition is false if not performing an ugprade.

>Sure, makes sense to me. Is the idea moving forward to move as much functionality validations into one Test hostedcluster as much as possible?

Yes, this can be different t.run( within the same Test though. Also to go out of the feature gate we can possibly have 2 Test* so we smoke test n karpenter instances running simultaneously.

/approve
/test e2e-aws-techpreview

to check for flakes

enxebre:
/override Red Hat Konflux / hypershift-operator-main-on-pull-request

enxebre:
/override Red Hat Konflux / hypershift-operator-main-enterprise-contract / hypershift-operator-main

PR #5871, opened 2025-03-19
OCPBUGS-46379: Kas bootstrap bin

enxebre:
kas-bootstrap is a tool to run the pre-required actions for bootstraping the kas during cluster creation (or upgrade).
It will apply some CRDs rendered by the cluster-config-operator and update the featureGate CR status by appending the git FeatureGate status.

It aims to alleviate the current bash scripts fragility, fix the fact we always replace instead of append the featureGate.status and include current and any upcoming changes to this logic with appropriate test coverage

Follow up:
Move the logic from kasContainerApplyBootstrap to kasContainerBootstrap and drop the former.
Add it to cpov2

Fixes #OCPBUGS-46379

**Checklist**

### DIFF
--- a/kas-bootstrap/kas_boostrap.go
+++ b/kas-bootstrap/kas_boostrap.go
@@ -1,0 +1,142 @@
+package kasbootstrap
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	configv1 "github.com/openshift/api/config/v1"
+
+	equality "k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/sets"
+
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+
+	"go.uber.org/zap/zapcore"
+)
+
+func init() {
+	utilruntime.Must(configv1.Install(configScheme))
+}
+
+var (
+	configScheme = runtime.NewScheme()
+	configCodecs = serializer.NewCodecFactory(configScheme)
+)
+
+func run(ctx context.Context, opts Options) error {
+	logger := zap.New(zap.JSONEncoder(func(o *zapcore.EncoderConfig) {
+		o.EncodeTime = zapcore.RFC3339TimeEncoder
+	}))
+	ctrl.SetLogger(logger)
+
+	cfg, err := ctrl.GetConfig()
+	if err != nil {
+		return fmt.Errorf("failed to get config: %w", err)
+	}
+	c, err := client.New(cfg, client.Options{Scheme: configScheme})
+	if err != nil {
+		return fmt.Errorf("failed to create client: %w", err)
+	}
+
+	content, err := os.ReadFile(filepath.Join(opts.RenderedFeatureGatePath, "99_feature-gate.yaml"))
+	if err != nil {
+		return fmt.Errorf("failed to read featureGate file: %w", err)
+	}
+
+	renderedFeatureGate, err := parseFeatureGateV1(content)
+	if err != nil {
+		return fmt.Errorf("failed to parse featureGate file: %w", err)
+	}
+
+	if err := reconcileFeatureGate(ctx, c, renderedFeatureGate); err != nil {
+		return fmt.Errorf("failed to reconcile featureGate: %w", err)
+	}
+
+	// we want to keep the process running during the lifecycle of the Pod because the Pod runs with restartPolicy=Always
+	// and it's not possible for individual containers to have a dedicated restartPolicy like onFailure.
+
+	// start a goroutine that will close the done channel when the context is done.
+	done := make(chan struct{})
+	go func() {
+		<-ctx.Done()
+		close(done)
+	}()
+
+	logger.Info("kas-bootstrap process completed successfully, waiting for termination signal")
+	<-done
+
+	return nil
+}
+
+// reconcileFeatureGate reconciles the featureGate CR status appending the renderedFeatureGate status.featureGates to the existing featureGates.
+// It will not fail if the clusterVersion is not found as this is expected for a brand new cluster.
+// But it will remove any featureGates that are not in the clusterVersion.Status.History if it exists.
+func reconcileFeatureGate(ctx context.Context, c client.Client, renderedFeatureGate *configv1.FeatureGate) error {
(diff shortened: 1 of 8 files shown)


enxebre:
cc @wking

kas-bootstrap/kas_boostrap.go
```diff
@@ -0,0 +1,133 @@
...
+		return fmt.Errorf("failed to parse featureGate file: %w", err)
+	}
+
+	if err := reconcileFeatureGate(ctx, c, renderedFeatureGate); err != nil {
+		return fmt.Errorf("failed to reconcile featureGate: %w", err)
+	}
+
+	// we want to keep the process running during the lifecycle of the Pod.
+	// start a goroutine that will close the done channel when the context is done
+	done := make(chan struct{})
```
muraee, line 64: can you explain why we want to keep the process running?

muraee:
don't we want to add this in cpov2 as well?

enxebre:
>don't we want to add this in cpov2 as well?

Yes, It's stated as a follow up in the PR desc. This is just the first deliverable to keep PR sizing small

kas-bootstrap/kas_boostrap.go
enxebre: because the pod RestartPolicy is Always, this mimics current behaviour and I would defer deviating from it to a different change

muraee: ack, any reason we are not adding this an an init container instead?

wking: I'd guess it can't be an init container because it needs the actual API-server in another long-lived container to talk to.  But couldn't we set [a container-scoped `restartPolicy: OnFailure`][1] for this container to get both:

* The ability to exit 0 when we were successfully reconciled and recover the resources the container process had been consuming.  Until some future when management of these resources moves from "successfully reconciled once per 4.y.z release" to "actively watched and managed with some regularity", which would be nice, but is likely more than we want to bite off in a single pull request.
* Reporting via `KubePodCrashLooping` if the container has trouble, while the container continues to relaunch and retry.  Not as direct as having the controlling CPO know _why_ the container was having trouble, but at least there would be a sign of trouble visible in Kube at a higher level than "dip into the container's logs".

[1]: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#restart-policy

kas-bootstrap/kas_boostrap.go
```diff
@@ -0,0 +1,133 @@
...
+	var clusterVersion configv1.ClusterVersion
+	err := c.Get(ctx, client.ObjectKey{Name: "version"}, &clusterVersion)
+	if err != nil {
+		// we don't fail if we can't get the clusterVersion, we will just not update the featureGate.
+		// This is always the case for a brand new cluster as the clusterVersion is not created yet.
+		logger.Info("WARNING: failed to get clusterVersion. This is expected for a brand new cluster", "error", err)
+	} else {
+		for _, cvoVersion := range clusterVersion.Status.History {
+			knownVersions = sets.NewString(clusterVersion.Status.Desired.Version)
+			knownVersions.Insert(cvoVersion.Version)
```
wking, line 92: standalone OCP currently doesn't do any garbage collection, so what you have now is fine as it stands.  But once you hit your first `Completed` entry and insert that into `knownVersions`, you can `break`, because there shouldn't be anything left on the cluster that cares about those ancient releases anymore:

```go
for _, cvoVersion := range clusterVersion.Status.History {
	knownVersions = sets.NewString(clusterVersion.Status.Desired.Version)
	knownVersions.Insert(cvoVersion.Version)
	if cvoVersion.State == configv1.CompletedUpdate {
		break
	}
}
```

kas-bootstrap/kas_boostrap.go
csrwng: @wking this last comment led me to do a little bit of experimentation on a 4.19 ci cluster :) 
- Tried changing the restart policy of a side container under `.spec.containers`, and failed admission with:
`* spec.template.spec.containers[1].restartPolicy: Forbidden: may not be set for non-init containers`

- Then tried moving the container under `.spec.initContainers` with a restartPolicy of `OnFailure`, and also failed admission with:
`* spec.template.spec.initContainers[0].restartPolicy: Unsupported value: "OnFailure": supported values: "Always"`

- Then tried changing the initContainer restartPolicy to `Always` and the deployment was accepted. The init container ended up running as a side container (which was new to me :)). I could not see a difference though between the additional container under `.spec.containers` or the init container with restartPolicy as Always.

Bottom line, I think what we have here is fine.

kas-bootstrap/kas_boostrap.go
```diff
@@ -0,0 +1,133 @@
...
+			continue
+		}
+		if len(knownVersions) > 0 && !knownVersions.Has(featureGateValues.Version) {
+			continue
+		}
+		desiredFeatureGates = append(desiredFeatureGates, featureGateValues)
+	}
+
+	if equality.Semantic.DeepEqual(desiredFeatureGates, featureGate.Status.FeatureGates) {
+		logger.Info("There is no updated for featureGate.Status.FeatureGates")
```
csrwng, line 115: really minor nit: s/updated/update/

csrwng (commented):
In general lgtm, definitely a step in the right direction

kas-bootstrap/kas_boostrap.go
enxebre: Thanks, updated the logic and unit coverage to reflect this

kas-bootstrap/kas_boostrap.go
```diff
@@ -0,0 +1,133 @@
...
+		return nil
+	}
+
+	featureGate.Status.FeatureGates = desiredFeatureGates
+	if err := c.Status().Update(ctx, &featureGate); err != nil {
```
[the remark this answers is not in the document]
enxebre: Thanks, updated the logic to patch instead

kas-bootstrap/kas_boostrap.go
enxebre: yes, the reason I didn't just set restart on failure for this container is that afaik individual containers can't supersede the Pod restartPolicy. Moving it to init as a side container will technically differ operationally from what we do at the moment and possibly causing more restarts while racing rendering for no value? so I'd rather keep it as it is to keep changes scoped and defer any further change to different PRs. After this one we'll still need to move the apply logic to this binary.
I added a comment in code to clarify about the restart policies.

enxebre:

enxebre:

rtheis:
Thank you @enxebre.  Are you all okay with or considering a cherry-pick of this fix to 4.18?

enxebre:
>Thank you @enxebre. Are you all okay with or considering a cherry-pick of this fix to 4.18?

I think we can backport this and https://github.com/openshift/hypershift/pull/5937 together

rtheis:
@enxebre do you still think that a backport is doable for these changes?

PR #5873, opened 2025-03-20
CNTRLPLANE-350: add NodePool minor version compatibility check

LiangquanLi930:
This commit adds validation to ensure NodePool versions are compatible with the HostedCluster version:
- For 4.even versions, allows y-2 difference
- For 4.odd versions, allows y-1 difference
- NodePool version cannot be higher than control plane version

The validation is skipped in disconnected environments where registry access is not available.

**Jira story**
https://issues.redhat.com/browse/CNTRLPLANE-330
https://issues.redhat.com/browse/CNTRLPLANE-350

**Log**
```
➜  oc get hostedcluster -A              
NAMESPACE   NAME          VERSION                              KUBECONFIG                     PROGRESS    AVAILABLE   PROGRESSING   MESSAGE
clusters    liangli0320   4.19.0-0.nightly-2025-03-19-080024   liangli0320-admin-kubeconfig   Completed   True        False         The hosted control plane is available
➜  bin/hcp create nodepool aws --name test07 --cluster-name liangli0320  --node-count  1 --release-image quay.io/openshift-release-dev/ocp-release:4.18.5-x86_64
NodePool test07 created
➜  bin/hcp create nodepool aws --name test07 --cluster-name liangli0320  --node-count  1 --release-image quay.io/openshift-release-dev/ocp-release:4.17.21-x86_64
{"level":"error","ts":"2025-03-20T17:24:49+08:00","msg":"Failed to create nodepool","error":"NodePool minor version 4.17 is not compatible with the HostedCluster minor version 4.19 (max allowed difference: 1)","stacktrace":"github.com/openshift/hypershift/cmd/nodepool/core.(*CreateNodePoolOptions).CreateRunFunc.func1\n\t/Users/daquan/hypershift/hypershift/cmd/nodepool/core/create.go:47\ngithub.com/spf13/cobra.(*Command).execute\n\t/Users/daquan/hypershift/hypershift/vendor/github.com/spf13/cobra/command.go:985\ngithub.com/spf13/cobra.(*Command).ExecuteC\n\t/Users/daquan/hypershift/hypershift/vendor/github.com/spf13/cobra/command.go:1117\ngithub.com/spf13/cobra.(*Command).Execute\n\t/Users/daquan/hypershift/hypershift/vendor/github.com/spf13/cobra/command.go:1041\ngithub.com/spf13/cobra.(*Command).ExecuteContext\n\t/Users/daquan/hypershift/hypershift/vendor/github.com/spf13/cobra/command.go:1034\nmain.main\n\t/Users/daquan/hypershift/hypershift/product-cli/main.go:59\nruntime.main\n\t/Users/daquan/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.23.0.darwin-arm64/src/runtime/proc.go:272"}
Error: NodePool minor version 4.17 is not compatible with the HostedCluster minor version 4.19 (max allowed difference: 1)
NodePool minor version 4.17 is not compatible with the HostedCluster minor version 4.19 (max allowed difference: 1)
```

**Checklist**

### DIFF
--- a/cmd/nodepool/core/create.go
+++ b/cmd/nodepool/core/create.go
@@ -9,13 +9,17 @@ import (
 	"github.com/openshift/hypershift/cmd/log"
 	"github.com/openshift/hypershift/cmd/util"
 	hyperapi "github.com/openshift/hypershift/support/api"
+	"github.com/openshift/hypershift/support/releaseinfo"
 
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 
 	ctrl "sigs.k8s.io/controller-runtime"
 	crclient "sigs.k8s.io/controller-runtime/pkg/client"
 
+	"github.com/blang/semver"
 	"github.com/spf13/cobra"
 )
 
@@ -51,6 +55,10 @@ func (o *CreateNodePoolOptions) CreateRunFunc(platformOpts PlatformOptions) func
 func (o *CreateNodePoolOptions) Validate(ctx context.Context, c crclient.Client) error {
 	// Validate HostedCluster payload can support the NodePool CPU type
 	if err := validateHostedClusterPayloadSupportsNodePoolCPUArch(ctx, c, o.ClusterName, o.Namespace, o.Arch); err != nil {
+		return err
+	}
+
+	if err := validMinorVersionCompatibility(ctx, c, o.ClusterName, o.Namespace, o.ReleaseImage, &releaseinfo.RegistryClientProvider{}); err != nil {
 		return err
 	}
 
@@ -180,6 +188,98 @@ func validateHostedClusterPayloadSupportsNodePoolCPUArch(ctx context.Context, cl
 
 	if hc.Status.PayloadArch != "" && hc.Status.PayloadArch != hyperv1.Multi && hc.Status.PayloadArch != hyperv1.ToPayloadArch(arch) {
 		return fmt.Errorf("NodePool CPU arch, %s, is not supported by the HostedCluster payload type, %s", arch, hc.Status.PayloadArch)
+	}
+
+	return nil
+}
+
+// validMinorVersionCompatibility validates that the NodePool version is compatible with the HostedCluster version.
+// For 4.even versions, it allows y-2 difference.
+// For 4.odd versions, it allows y-1 difference.
+// NodePool version cannot be higher than control plane version.
+func validMinorVersionCompatibility(ctx context.Context, client crclient.Client, name, namespace, nodePoolReleaseImage string, releaseProvider releaseinfo.Provider) error {
+	if nodePoolReleaseImage == "" {
+		return nil
+	}
+	logger := ctrl.LoggerFrom(ctx)
+
+	hcluster := &hyperv1.HostedCluster{}
+	if err := client.Get(ctx, types.NamespacedName{Namespace: namespace, Name: name}, hcluster); err != nil {
+		if !apierrors.IsNotFound(err) {
+			// For other errors (e.g. API server issues, RBAC problems), we should return the error
+			return fmt.Errorf("failed to get HostedCluster to check version compatibility: %w", err)
+		}
+
+		// This is expected to happen when we create a cluster since there is no created HostedCluster CR to check the
+		// payload from.
+		logger.Info("WARNING: failed to get HostedCluster to check version compatibility")
+		return nil
+	}
+
+	// Get the control plane version string
+	var controlPlaneVersionStr string
+	if len(hcluster.Status.Version.History) == 0 {
+		// If the cluster is in the process of installation, there is no history
+		// Use the desired version as the control plane version
+		controlPlaneVersionStr = hcluster.Status.Version.Desired.Version
+	} else {
+		// If the cluster is installed or upgrading
+		// Start with the most recent version from history as the default
+		controlPlaneVersionStr = hcluster.Status.Version.History[len(hcluster.Status.Version.History)-1].Version
+		// Update with any more recent Completed version if found
+		for _, history := range hcluster.Status.Version.History {
+			if history.State == "Completed" {
+				controlPlaneVersionStr = history.Version
+				break
+			}
+		}
+	}
+
+	// Parse control plane version
+	controlPlaneVersion, err := semver.Parse(controlPlaneVersionStr)
+	if err != nil {
+		return fmt.Errorf("parsing control plane version (%s): %w", controlPlaneVersionStr, err)
+	}
+
+	pullSecret := &corev1.Secret{}
+	if err = client.Get(ctx, types.NamespacedName{Namespace: namespace, Name: hcluster.Spec.PullSecret.Name}, pullSecret); err != nil {
+		return fmt.Errorf("failed to get pull secret: %w", err)
+	}
+
+	releaseImage, err := releaseProvider.Lookup(ctx, nodePoolReleaseImage, pullSecret.Data[corev1.DockerConfigJsonKey])
+	if err != nil {
+		// Skip version check in disconnected environment where registry access is not available
+		logger.Info("WARNING: Unable to access the payload, skipping the Minor Version check.", "error", err.Error())
+		return nil
+	}
+
+	// Parse NodePool version
+	nodePoolVersion, err := semver.Parse(releaseImage.Version())
+	if err != nil {
+		return fmt.Errorf("parsing NodePool version (%s): %w", releaseImage.Version(), err)
+	}
+
+	// NodePool version cannot be higher than control plane version
+	if nodePoolVersion.GT(controlPlaneVersion) {
+		return fmt.Errorf("NodePool version %s cannot be higher than the HostedCluster version %s",
+			nodePoolVersion, controlPlaneVersion)
+	}
+
+	// Calculate minor version difference
+	versionDiff := int64(controlPlaneVersion.Minor - nodePoolVersion.Minor)
+
+	// For 4.even versions, allow y-2 difference
+	// For 4.odd versions, allow y-1 difference
+	maxAllowedDiff := int64(2)
+	if controlPlaneVersion.Minor%2 == 1 {
+		maxAllowedDiff = 1
+	}
+
+	if versionDiff > maxAllowedDiff {
+		return fmt.Errorf("NodePool minor version %d.%d is not compatible with the HostedCluster minor version %d.%d (max allowed difference: %d)",
+			nodePoolVersion.Major, nodePoolVersion.Minor,
+			controlPlaneVersion.Major, controlPlaneVersion.Minor,
+			maxAllowedDiff)
 	}
 
 	return nil
--- a/cmd/nodepool/core/create_test.go
+++ b/cmd/nodepool/core/create_test.go
@@ -8,7 +8,11 @@ import (
 
 	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
 	"github.com/openshift/hypershift/support/api"
+	fakereleaseprovider "github.com/openshift/hypershift/support/releaseinfo/fake"
 
+	configv1 "github.com/openshift/api/config/v1"
+
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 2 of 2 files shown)


cmd/nodepool/core/create.go
```diff
@@ -184,3 +191,72 @@ func validateHostedClusterPayloadSupportsNodePoolCPUArch(ctx context.Context, cl
...
+	hcluster := &hyperv1.HostedCluster{}
+	if err := client.Get(ctx, types.NamespacedName{Namespace: namespace, Name: name}, hcluster); err != nil {
+		// This is expected to happen when we create a cluster since there is no created HostedCluster CR to check the
+		// payload from.
+		logger.Info("WARNING: failed to get HostedCluster to check version compatibility")
+		return nil
+	}
+
+	// Parse control plane version
+	controlPlaneVersion, err := semver.Parse(hcluster.Status.Version.Desired.Version)
```
wking, line 214: This is going to be a bit sticky, but `Desired` here is because this is what the cluster is aiming to reach, and doesn't talk about what the cluster has actually delivered.  For example:

1. HostedCluster running 4.18
2. You ask the HostedCluster to update to a 4.19 release.
3. HyperShift operator passes the request down to the HostedControlPlane and rolls out a new HostedControlPlane operator.
4. HCP-operator passes the request down to ClusterVersion and rolls out a new cluster-version operator.
5. CVO reports the 4.19 target in ClusterVersion's `status.desired`, a `Partial` entry in ClusterVersion's `status.history[0]`, and starts working on rolling out the portion of the hosted control plane that it manages.
6. All the 4.19 stuff has rolled out!  CVO records a `Completed` entry in ClusterVersion's `status.history[0]`.

By looking at `hcluster.Status.Version.Desired.Version`, you'll start allowing 4.19 NodePools at (5), but I'd guess you probably actually want to wait until (6).

cmd/nodepool/core/create.go
```diff
@@ -184,3 +191,72 @@ func validateHostedClusterPayloadSupportsNodePoolCPUArch(ctx context.Context, cl
...
+
+	// Parse NodePool version
+	nodePoolVersion, err := semver.Parse(releaseImage.Version())
+	if err != nil {
+		return fmt.Errorf("parsing NodePool version (%s): %w", releaseImage.Version(), err)
+	}
+
+	// NodePool version cannot be higher than control plane version
+	if nodePoolVersion.Minor > controlPlaneVersion.Minor {
+		return fmt.Errorf("NodePool minor version %d.%d cannot be higher than the HostedCluster minor version %d.%d",
```
wking, line 239: HyperShift's position on patch-version skew is [unclear][1], but in the absence of explicit docs allowing 4.y.(z > z') NodePools on 4.y.z' HostedClusters, you may want to also have a guard for that:

```go
else if nodePoolVersion.Minor == controlPlaneVersion.Minor && nodePoolVersion.Patch > controlPlaneVersion.Patch {
```

Or if HyperShift's patch-skew policy is "go for it; we test and support that!" then hooray, you're fine here without trying to check that situation :)

[1]: https://issues.redhat.com/browse/RFE-6410

cmd/nodepool/core/create.go
LiangquanLi930: For an upgrade from 4.18 to 4.19, given the following update status:
```
{
  "availableUpdates": null,
  "desired": {
    "image": "quay.io/openshift-release-dev/ocp-release@sha256:3c7decd8e09329d5206a96bcc19838d25bdc3af9c9565f249aa91494c7beb7db",
    "version": "4.19.0-ec.3"
  },
  "history": [
    {
      "completionTime": null,
      "image": "quay.io/openshift-release-dev/ocp-release@sha256:3c7decd8e09329d5206a96bcc19838d25bdc3af9c9565f249aa91494c7beb7db",
      "startedTime": "2025-03-21T10:42:39Z",
      "state": "Partial",
      "verified": false,
      "version": "4.19.0-ec.3"
    },
    {
      "completionTime": "2025-03-21T10:33:24Z",
      "image": "quay.io/openshift-release-dev/ocp-release@sha256:61fdad894f035a8b192647c224faf565279518255bdbf60a91db4ee0479adaa6",
      "startedTime": "2025-03-21T10:24:09Z",
      "state": "Completed",
      "verified": false,
      "version": "4.18.6"
    }
  ],
  "observedGeneration": 1
}
```
IIUC, we should allow the NodePool to remain on 4.18 rather than 4.19. So, we should extract the first version in the history list where state is Completed, which in this case is 4.18.6. Is that correct?

cmd/nodepool/core/create.go
```diff
@@ -184,3 +191,95 @@ func validateHostedClusterPayloadSupportsNodePoolCPUArch(ctx context.Context, cl
...
+	if nodePoolReleaseImage == "" {
+		return nil
+	}
+	logger := ctrl.LoggerFrom(ctx)
+
+	hcluster := &hyperv1.HostedCluster{}
+	if err := client.Get(ctx, types.NamespacedName{Namespace: namespace, Name: name}, hcluster); err != nil {
+		// This is expected to happen when we create a cluster since there is no created HostedCluster CR to check the
+		// payload from.
+		logger.Info("WARNING: failed to get HostedCluster to check version compatibility")
```
wking, line 209: but do you have to allow NodePool creation to complete before there's a HostedCluster?  You could decide to block until the HostedCluster exists and have the `create` caller retry after that point.  Or you could loop some number of times or for some duration here on the does-not-exist failure mode, waiting for the HostedCluster to start existing in parallel.

cmd/nodepool/core/create.go
```diff
@@ -184,3 +191,95 @@ func validateHostedClusterPayloadSupportsNodePoolCPUArch(ctx context.Context, cl
...
+// For 4.odd versions, it allows y-1 difference.
+// NodePool version cannot be higher than control plane version.
+func validMinorVersionCompatibility(ctx context.Context, client crclient.Client, name, namespace, nodePoolReleaseImage string) error {
+	if nodePoolReleaseImage == "" {
+		return nil
+	}
+	logger := ctrl.LoggerFrom(ctx)
+
+	hcluster := &hyperv1.HostedCluster{}
+	if err := client.Get(ctx, types.NamespacedName{Namespace: namespace, Name: name}, hcluster); err != nil {
```
wking, line 206: You probably want to distinguish the [`IsNotFound`][1] case from other errors, so you don't treat situations like Kube-API hiccups or RBAC wrinkles as "do whatever you want!" unguarded windows.

[1]: https://pkg.go.dev/k8s.io/apimachinery/pkg/api/errors#IsNotFound

cmd/nodepool/core/create.go
```diff
@@ -184,3 +191,95 @@ func validateHostedClusterPayloadSupportsNodePoolCPUArch(ctx context.Context, cl
...
+	var controlPlaneVersionStr string
+	if len(hcluster.Status.Version.History) == 0 {
+		// If the cluster is in the process of installation, there is no history
+		// Use the desired version as the control plane version
+		controlPlaneVersionStr = hcluster.Status.Version.Desired.Version
+	} else {
+		// If the cluster is installed or upgrading
+		// Find the last completed version from history as the control plane version
+		for _, history := range hcluster.Status.Version.History {
+			if history.State == "Completed" {
```
wking, line 223: There's no guarantee that `History` will have a `Completed` entry; it could be full of `Partial`.  A safer approach here is to assume that the Kube API is the oldest version, and then update with any more-recent `Completed` version.  Something like:

```go
controlPlaneVersionStr = hcluster.Status.Version.History[len(hcluster.Status.Version.History)-1].Version
for _, history := range hcluster.Status.Version.History {
	if history.State == "Completed" {
	....
```

cmd/nodepool/core/create.go
LiangquanLi930: When creating a HostedCluster using the HyperShift CLI, a default NodePool is also created. This is just CLI test code—we also add validation on the controller side.

cmd/nodepool/core/create.go
LiangquanLi930: done

cmd/nodepool/core/create.go
LiangquanLi930: done

cmd/nodepool/core/create.go
LiangquanLi930: done

cmd/nodepool/core/create.go
```diff
@@ -184,3 +191,95 @@ func validateHostedClusterPayloadSupportsNodePoolCPUArch(ctx context.Context, cl
...
+
+	// NodePool version cannot be higher than control plane version
+	if nodePoolVersion.Minor > controlPlaneVersion.Minor {
+		return fmt.Errorf("NodePool minor version %d.%d cannot be higher than the HostedCluster minor version %d.%d",
+			nodePoolVersion.Major, nodePoolVersion.Minor,
+			controlPlaneVersion.Major, controlPlaneVersion.Minor)
+	} else if nodePoolVersion.Minor == controlPlaneVersion.Minor && nodePoolVersion.Patch > controlPlaneVersion.Patch {
+		// If the NodePool minor version is the same as the control plane minor version,
+		// And the NodePool patch version is greater than the control plane patch version, return an error
+		return fmt.Errorf("NodePool patch version %d.%d.%d cannot be higher than the HostedCluster patch version %d.%d.%d",
```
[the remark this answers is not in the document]
LiangquanLi930: done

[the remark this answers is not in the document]
wking: Yup, 5e1b1a6fcb handles this case now; feel free to mark this thread resolved.

cmd/nodepool/core/create.go
wking: Sounds like that controller-side validation is future-plans, not current-implementation.  But I'm fine assuming that there will be some controller-side backstops, and that this client code is just best-effort attempts to warn command-line users without bothering to go as far and slow as an API NodePool creation attempt, and letting these through here with the warning :+1:

wking (approved):
Looks good!  My only outstanding concern is [the race when HostedCluster hasn't been created yet][1], but as you've pointed out in that thread, anything crazy should be caught by coming controller-side validation.  So nothing else I can see that needs changing in this pull request, and I'm approving from my end.  I'll leave the LGTM to someone HyperShiftier.

[1]: https://github.com/openshift/hypershift/pull/5873#discussion_r2010469993

wking:
Ah I didn't realize [my approval][1] would cause [Prow to LGTM][2].

/hold so HyperShiftier folks have time to review

[1]: https://github.com/openshift/hypershift/pull/5873#pullrequestreview-2712025702
[2]: https://github.com/openshift/hypershift/pull/5873#event-16965412458

cmd/nodepool/core/create_test.go
```diff
@@ -102,3 +107,106 @@ func TestValidateHostedClusterPayloadSupportsNodePoolCPUArch(t *testing.T) {
...
+	log.SetLogger(zap.New(zap.UseDevMode(true)))
+
+	tests := []struct {
+		name                 string
+		controlPlaneVersion  string
+		nodePoolReleaseImage string
+		expectedError        string
+	}{
+		{
+			name:                 "same version",
```
enxebre, line 121: nit: usually prefer "when ... it should ..." format

cmd/nodepool/core/create.go
```diff
@@ -184,3 +192,99 @@ func validateHostedClusterPayloadSupportsNodePoolCPUArch(ctx context.Context, cl
...
+	} else if nodePoolVersion.Minor == controlPlaneVersion.Minor && nodePoolVersion.GT(controlPlaneVersion) {
+		// If the NodePool minor version is the same as the control plane minor version,
+		// And the NodePool version is greater than the control plane version (including pre-release and build versions)
+		return fmt.Errorf("NodePool version %s cannot be higher than the HostedCluster version %s", nodePoolVersion, controlPlaneVersion)
+	}
+
+	// Calculate minor version difference
+	versionDiff := int64(controlPlaneVersion.Minor - nodePoolVersion.Minor)
+
+	// For 4.even versions, allow y-2 difference
```
enxebre, line 275: where is the even/odd distinction coming from?

jparrill (changes requested):
Thank you!

cmd/nodepool/core/create.go
LiangquanLi930: https://issues.redhat.com/browse/OCPSTRAT-1707
> Requirements (aka. Acceptance Criteria):
> + The HCP documentation accurately reflects the supported NodePool minor version skew of y-2 and removes the floating reference to [upstream kubernetes](https://kubernetes.io/releases/version-skew-policy/).
> + HCP prevents the creation of NodePools with a skew greater than y-2 (for 4.even control planes) or y-1 (for 4.odd control planes) through code guardrails. Example of how[ its done in standalone OpenShift|https://github.com/openshift/cluster-kube-apiserver-operator/pull/1199#issue-958481467].
> + Existing NodePools with a skew greater than y-2 are not impacted (but upgrading the control plane may be blocked).
> + Appropriate error messages are provided to users when attempting to create NodePools with an unsupported skew.

cmd/nodepool/core/create.go
```diff
@@ -184,3 +192,99 @@ func validateHostedClusterPayloadSupportsNodePoolCPUArch(ctx context.Context, cl
...
+	if err != nil {
+		return fmt.Errorf("parsing NodePool version (%s): %w", releaseImage.Version(), err)
+	}
+
+	// NodePool version cannot be higher than control plane version
+	if nodePoolVersion.Minor > controlPlaneVersion.Minor {
+		return fmt.Errorf("NodePool minor version %d.%d cannot be higher than the HostedCluster minor version %d.%d",
+			nodePoolVersion.Major, nodePoolVersion.Minor,
+			controlPlaneVersion.Major, controlPlaneVersion.Minor)
+	} else if nodePoolVersion.Minor == controlPlaneVersion.Minor && nodePoolVersion.GT(controlPlaneVersion) {
```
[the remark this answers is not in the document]
LiangquanLi930: This is due to [RFE-6410](https://issues.redhat.com/browse/RFE-6410). If the NodePool Z version is allowed to be greater than the ControlPlane version, we can simply remove this else if statement.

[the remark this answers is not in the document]
enxebre: I see, my point is that you are not looking at the major version at all, the unit test above will pass right now and it should not.
Also let's don't conflate this PR with RFE-6410 which would require involved changes for the backend, the CLI automation does not dictate support, the API does.

cmd/nodepool/core/create.go
```diff
@@ -184,3 +192,99 @@ func validateHostedClusterPayloadSupportsNodePoolCPUArch(ctx context.Context, cl
...
+	if err := client.Get(ctx, types.NamespacedName{Namespace: namespace, Name: name}, hcluster); err != nil {
+		if apierrors.IsNotFound(err) {
+			// This is expected to happen when we create a cluster since there is no created HostedCluster CR to check the
+			// payload from.
+			logger.Info("WARNING: failed to get HostedCluster to check version compatibility")
+			return nil
+		}
+		// For other errors (e.g. API server issues, RBAC problems), we should return the error
+		return fmt.Errorf("failed to get HostedCluster to check version compatibility: %w", err)
+	}
```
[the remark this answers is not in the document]
LiangquanLi930: Yes, IMO, further verification should be handled by the controller.

cmd/nodepool/core/create.go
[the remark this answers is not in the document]
LiangquanLi930: Ah, sorry, I didn't get your point. In my mind, all OCP versions are 4.x by default, so I didn't add a check. updated

cmd/nodepool/core/create_test.go
```diff
@@ -102,3 +107,106 @@ func TestValidateHostedClusterPayloadSupportsNodePoolCPUArch(t *testing.T) {
 		})
 	}
 }
+
+func TestValidMinorVersionCompatibility(t *testing.T) {
+	log.SetLogger(zap.New(zap.UseDevMode(true)))
```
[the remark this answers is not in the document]
LiangquanLi930: Thanks, updated

cmd/nodepool/core/create.go
```diff
@@ -184,3 +192,99 @@ func validateHostedClusterPayloadSupportsNodePoolCPUArch(ctx context.Context, cl
...
+	if err != nil {
+		return fmt.Errorf("parsing control plane version (%s): %w", controlPlaneVersionStr, err)
+	}
+
+	pullSecret := &corev1.Secret{}
+	if err = client.Get(ctx, types.NamespacedName{Namespace: namespace, Name: hcluster.Spec.PullSecret.Name}, pullSecret); err != nil {
+		return fmt.Errorf("failed to get pull secret: %w", err)
+	}
+	releaseProvider := &releaseinfo.RegistryClientProvider{}
+	releaseImage, err := releaseProvider.Lookup(ctx, nodePoolReleaseImage, pullSecret.Data[corev1.DockerConfigJsonKey])
```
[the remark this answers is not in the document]
LiangquanLi930: Thanks! done

cmd/nodepool/core/create_test.go
LiangquanLi930: done

cmd/nodepool/core/create.go
```diff
@@ -184,3 +192,100 @@ func validateHostedClusterPayloadSupportsNodePoolCPUArch(ctx context.Context, cl
...
+
+	// Parse NodePool version
+	nodePoolVersion, err := semver.Parse(releaseImage.Version())
+	if err != nil {
+		return fmt.Errorf("parsing NodePool version (%s): %w", releaseImage.Version(), err)
+	}
+
+	// NodePool version cannot be higher than control plane version
+	if nodePoolVersion.Minor > controlPlaneVersion.Minor || nodePoolVersion.Major > controlPlaneVersion.Major {
+		return fmt.Errorf("NodePool version %d.%d cannot be higher than the HostedCluster version %d.%d",
```
enxebre, line 264: this still seems semantically incorrect, a hc with greater major but less minor will fail here with this message, and nested if/else become difficult to read.

Isn't this all we need?
```
// nodePool version must be the same or lower.
if nodePoolVersion.GT(controlPlaneVersion) {
   fail
}
```

enxebre: And when we want to allow greater z
```
	// NodePool version cannot be higher than control plane version
	// major version must be the same.
	if nodePoolVersion.Major != controlPlaneVersion.Major {
		fail
	}

	// minor version must be the same or lower.
	if nodePoolVersion.Minor > controlPlaneVersion.Minor {
		fail
	}
```

enxebre:
lgtm
I'll defer to @jparrill to tag since he shared some feedback

bryan-cox:

bryan-cox:

cmd/nodepool/core/create.go
wking: #5931 is following up with the HyperShift-operator side validation discussed in this thread.

PR #5877, opened 2025-03-20
NO-JIRA: Improve error output msg for Azure CLI role cmds

bryan-cox:
This commit improves the error output messaging for Azure CLI role assignment commands. This should help with diagnosing issues with Azure.

N/A

**Checklist**


cmd/infra/azure/create.go
```diff
@@ -897,7 +897,7 @@ func execAzCommand(cmdStr string) ([]byte, error) {
 	cmd := exec.Command("sh", "-c", cmdStr)
 	output, err := cmd.CombinedOutput()
 	if err != nil {
-		return nil, fmt.Errorf("failed to execute az command: %w", err)
+		return nil, fmt.Errorf("failed to execute az command, output: %s err: %w", string(output), err)
```
sjenning, line 900: seems like we already return this output to the caller in an error below on 904?

bryan-cox: I removed that line now

PR #5879, opened 2025-03-21
OCPBUGS-53403: Remove synced annotations from HCP when they are removed from HC

csrwng:
This commit ensures that annotations that are synced by the HostedCluster are removed from the HostedControlPlane when they are removed from the HostedCluster. Annotations that the HostedCluster reconciler does not know about are not removed.

Fixes #[OCPBUGS-53403](https://issues.redhat.com/browse/OCPBUGS-53403)

**Checklist**


csrwng:

PR #5881, opened 2025-03-21
OCPBUGS-53462: Check for availability of individual catalog images

csrwng:
This commit refactors the code that fetches the images to use for OLM catalogs into a common support package. It checks individual catalog images instead of assuming that all of them have the same tag available. It also introduces a cache for images so that every reconcile loop does not result in registry lookups.

Fixes #[OCPBUGS-53462](https://issues.redhat.com/browse/OCPBUGS-53462)

**Checklist**


bryan-cox (approved):
/lgtm

csrwng:
The only failure in e2e-aws is because the previous version has `Degraded=True`:
```
  - lastTransitionTime: "2025-03-23T06:07:02Z"
    message: community-operators-catalog deployment has 1 unavailable replicas
    observedGeneration: 3
    reason: UnavailableReplicas
    status: "True"
    type: Degraded
 ```
 
 This is not going to succeed. Overriding that test so we can get this fix in.
/override ci/prow/e2e-aws

PR #5917, opened 2025-03-28
NO-JIRA: chore: Remove Azure autorest dependency

bryan-cox:
This PR removes the Azure KMS use of the autorest dependency. The autorest was only used to determine the Azure Key Vault DNS suffix. A custom function mimicking the same functionality as autorest was implemented instead.

N/A

**Checklist**


bryan-cox:

PR #5931, opened 2025-03-30
CNTRLPLANE-377: Add NodePool minor version compatibility check to HO

LiangquanLi930:
Add NodePool minor version compatibility check

This commit adds version compatibility checks between HostedCluster and NodePool:
1. Add new conditions:
   - ValidVersionConditionType for HostedCluster
   - NodePoolValidVersionConditionType for NodePool

2. Implement version validation logic:
   - NodePool version cannot be higher than HostedCluster version
   - For 4.even versions, allow y-2 difference
   - For 4.odd versions, allow y-1 difference

Fixes #

**Checklist**


docs/content/how-to/common/nodepool-version-compatibility.md
```diff
@@ -0,0 +1,44 @@
...
+    - Allowed NodePool versions: 4.18.5, 4.18.4, 4.17.z, 4.16.z
+    - Disallowed NodePool versions: 4.15.z and below, 4.19.z and above
+
+- If HostedCluster version is 4.17.5:
+    - Allowed NodePool versions: 4.17.5, 4.17.4, 4.16.z
+    - Disallowed NodePool versions: 4.15.z and below, 4.18.z and above
+
+## Status Checks
+
+### HostedCluster Status
```
enxebre, line 25: I would find it counter intuitive if a ValidVersion is false in HC when the version of the HC is actually ok. Let's focus on getting the NodePools signaling right first. Then we can build upon that

LiangquanLi930: After discussion, this PR focuses on nodepool controller

docs/content/how-to/common/nodepool-version-compatibility.md
```diff
@@ -0,0 +1,44 @@
...
+- Status is `True` when all NodePool versions are compatible
+- Status is `False` when there are incompatible NodePool versions, with detailed error messages
+
+### NodePool Status
+The NodePool controller sets the `ValidVersion` condition to indicate version compatibility status:
+- Status is `True` when the NodePool version is compatible with the HostedCluster version
+- Status is `False` when versions are incompatible, with detailed error messages
+
+## Error Handling
+When version incompatibility is detected, the system will:
```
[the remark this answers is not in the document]
LiangquanLi930: Okay, modify the NodePool controller to not block the reconcile.

api/hypershift/v1beta1/nodepool_conditions.go
```diff
@@ -18,6 +18,9 @@ const (
 	// NodePoolValidTuningConfigConditionType signals if the content within nodePool.spec.tuningConfig is valid.
 	// A failure here is unlikely to resolve without the changing user input.
 	NodePoolValidTuningConfigConditionType = "ValidTuningConfig"
+	// NodePoolValidVersionConditionType signals if the version of the NodePool is valid and compatible.
```
enxebre, line 21: this naming results confusing for me given we also have NodePoolValidReleaseImageConditionType.
Happy to hear other thoughts, one idea "supportedVersionSkew".
Also let's be specific in the condition docs, e.g. "supportedVersionSkew signals if the NodePool points to a version that falls within the supported skew policy with the control plane.
When false, the NodePool will keep trying to operate as usual even though there are no guarantees"

LiangquanLi930: `supportedVersionSkew` look good to me

api/hypershift/v1beta1/nodepool_conditions.go
```diff
@@ -18,6 +18,10 @@ const (
 	// NodePoolValidTuningConfigConditionType signals if the content within nodePool.spec.tuningConfig is valid.
 	// A failure here is unlikely to resolve without the changing user input.
 	NodePoolValidTuningConfigConditionType = "ValidTuningConfig"
+	// NodePoolSupportedVersionSkewConditionType signals if the NodePool points to a version that falls within the supported skew policy with the HostedCluster.
+	// When false, the NodePool will keep trying to operate as usual even though there are no guarantees
+	// A failure here is unlikely to resolve without changing the version input.
```
enxebre, line 23: changing the version input.
Can we be specific, e.g. without changing spec.release.
Can we also specify what is the supported skew policy.

api/hypershift/v1beta1/nodepool_conditions.go
```diff
@@ -116,4 +120,5 @@ const (
 	InvalidOpenStackMachineTemplate       = "InvalidOpenStackMachineTemplate"
 	CIDRConflictReason                    = "CIDRConflict"
 	NodePoolKubeVirtLiveMigratableReason  = "KubeVirtNodesNotLiveMigratable"
+	NodePoolInvalidVersionReason          = "InvalidVersion"
```
enxebre, line 123: should this be "unsupportedSkew"?

hypershift-operator/controllers/nodepool/conditions.go
```diff
@@ -824,3 +826,107 @@ func (r NodePoolReconciler) createValidGeneratedPayloadCondition(ctx context.Con
 
 	return condition, nil
 }
+
+func (r *NodePoolReconciler) supportedVersionSkewCondition(ctx context.Context, nodePool *hyperv1.NodePool, hcluster *hyperv1.HostedCluster) (*ctrl.Result, error) {
+	if hcluster.Status.Version == nil {
+		SetStatusCondition(&nodePool.Status.Conditions, hyperv1.NodePoolCondition{
+			Type:               hyperv1.NodePoolSupportedVersionSkewConditionType,
+			Status:             corev1.ConditionFalse,
```
enxebre, line 834: should this be "unknown"?

hypershift-operator/controllers/nodepool/conditions.go
```diff
@@ -824,3 +826,107 @@ func (r NodePoolReconciler) createValidGeneratedPayloadCondition(ctx context.Con
...
+			Reason:             hyperv1.NodePoolValidationFailedReason,
+			Message:            "HostedCluster version is not available yet, waiting for version information",
+			ObservedGeneration: nodePool.Generation,
+		})
+		return &ctrl.Result{}, fmt.Errorf("HostedCluster version is not available yet")
+	}
+	removeStatusCondition(&nodePool.Status.Conditions, hyperv1.NodePoolSupportedVersionSkewConditionType)
```
enxebre, line 841: why is this removing needed?

hypershift-operator/controllers/nodepool/conditions.go
```diff
@@ -824,3 +826,107 @@ func (r NodePoolReconciler) createValidGeneratedPayloadCondition(ctx context.Con
...
+		return &ctrl.Result{}, fmt.Errorf("parsing control plane version (%s): %w", controlPlaneVersionStr, err)
+	}
+
+	releaseImage, err := r.getReleaseImage(ctx, hcluster, nodePool.Status.Version, nodePool.Spec.Release.Image)
+	if err != nil {
+		SetStatusCondition(&nodePool.Status.Conditions, hyperv1.NodePoolCondition{
+			Type:               hyperv1.NodePoolSupportedVersionSkewConditionType,
+			Status:             corev1.ConditionFalse,
+			Reason:             hyperv1.NodePoolValidationFailedReason,
+			Message:            fmt.Sprintf("Failed to get release image: %v", err.Error()),
```
enxebre, line 873: should this be unknown?

hypershift-operator/controllers/nodepool/conditions.go
```diff
@@ -824,3 +826,107 @@ func (r NodePoolReconciler) createValidGeneratedPayloadCondition(ctx context.Con
...
+	}
+	// Parse control plane version
+	controlPlaneVersion, err := semver.Parse(controlPlaneVersionStr)
+	if err != nil {
+		return &ctrl.Result{}, fmt.Errorf("parsing control plane version (%s): %w", controlPlaneVersionStr, err)
+	}
+
+	releaseImage, err := r.getReleaseImage(ctx, hcluster, nodePool.Status.Version, nodePool.Spec.Release.Image)
+	if err != nil {
+		SetStatusCondition(&nodePool.Status.Conditions, hyperv1.NodePoolCondition{
```
enxebre, line 869: why are we setting condition in some errors but not in others?

hypershift-operator/controllers/nodepool/conditions.go
```diff
@@ -824,3 +826,107 @@ func (r NodePoolReconciler) createValidGeneratedPayloadCondition(ctx context.Con
...
+			Reason: hyperv1.NodePoolInvalidVersionReason,
+			Message: fmt.Sprintf("NodePool minor version %d.%d is not compatible with the HostedCluster minor version %d.%d (max allowed difference: %d)",
+				nodePoolVersion.Major, nodePoolVersion.Minor,
+				controlPlaneVersion.Major, controlPlaneVersion.Minor,
+				maxAllowedDiff),
+			ObservedGeneration: nodePool.Generation,
+		})
+		return nil, nil
+	}
+	removeStatusCondition(&nodePool.Status.Conditions, hyperv1.NodePoolSupportedVersionSkewConditionType)
```
enxebre, line 923: why removeStatusCondition here?

hypershift-operator/controllers/nodepool/nodepool_controller_test.go
```diff
@@ -1476,3 +1477,187 @@ func Test_validateHCPayloadSupportsNodePoolCPUArch(t *testing.T) {
...
+
+	testCases := []struct {
+		name              string
+		nodePool          *hyperv1.NodePool
+		hostedCluster     *hyperv1.HostedCluster
+		releaseProvider   *fakereleaseprovider.FakeReleaseProvider
+		expectedCondition *hyperv1.NodePoolCondition
+		expectedError     string
+	}{
+		{
```
enxebre, line 1541: can we also cover the case where if len(hcluster.Status.Version.History) == 0 {

hypershift-operator/controllers/nodepool/nodepool_controller_test.go
```diff
@@ -1476,3 +1477,187 @@ func Test_validateHCPayloadSupportsNodePoolCPUArch(t *testing.T) {
 		})
 	}
 }
+
+func TestSupportedVersionSkewCondition(t *testing.T) {
+	pullSecret, err := os.ReadFile("../../../hack/dev/fakePullSecret.json")
```
enxebre, line 1482: why do we need this? can't it just be []byte("pullSecret")?

hypershift-operator/controllers/nodepool/nodepool_controller_test.go
LiangquanLi930: done

hypershift-operator/controllers/nodepool/conditions.go
LiangquanLi930: From my perspective, the error from semver.Parse() seems more like a rare or accidental issue that would be unlikely to occur during actual execution. So I didn’t add a Condition for it. I'm not sure if my understanding is correct. I noticed that in some cases in this section of the code: https://github.com/openshift/hypershift/blob/1c615c778e4e321819827a78accd9ecd2199abcc/hypershift-operator/controllers/nodepool/conditions.go#L315-L326, Conditions aren’t set either.

test/e2e/nodepool_prev_release_test.go
```diff
@@ -36,6 +37,11 @@ func (npPrevTest *NodePoolPrevReleaseCreateTest) Setup(t *testing.T) {
 }
 
 func (npPrevTest *NodePoolPrevReleaseCreateTest) BuildNodePoolManifest(defaultNodepool hyperv1.NodePool) (*hyperv1.NodePool, error) {
+	isCompatible, err := e2eutil.ValidateNodePoolVersionCompatibility(npPrevTest.release, npPrevTest.hostedCluster.Spec.Release.Image)
```
enxebre, line 40: why do we need this check here?

test/e2e/nodepool_upgrade_test.go
```diff
@@ -53,6 +53,10 @@ func NewNodePoolInPlaceUpgradeTestManifest(hostedCluster *hyperv1.HostedCluster,
 }
 
 func (ipu *NodePoolInPlaceUpgradeTestManifest) BuildNodePoolManifest(defaultNodepool hyperv1.NodePool) (*hyperv1.NodePool, error) {
+	isCompatible, err := e2eutil.ValidateNodePoolVersionCompatibility(ipu.previousReleaseImage, ipu.hostedCluster.Spec.Release.Image)
```
enxebre, line 56: why do we need this check here? we own this tests, we want to deterministically make them compatible or error otherwise

LiangquanLi930:
code conflicts

test/e2e/nodepool_prev_release_test.go
LiangquanLi930: I’ve refactored this part of the code and moved it to Setup().

enxebre: This might produce false passes freely since it would transparently skip any test where this is not met?
Why would we want to do this at all? If a test creates a nodepool that breaks the policy, the test should legitimately fail the conditions check and fail

test/e2e/nodepool_prev_release_test.go
```diff
@@ -4,26 +4,29 @@
...
 
 type NodePoolPrevReleaseCreateTest struct {
 	DummyInfraSetup
-	hostedCluster *hyperv1.HostedCluster
-	release       string
-	clusterOpts   e2eutil.PlatformAgnosticOptions
+	hostedCluster     *hyperv1.HostedCluster
+	release           string
+	clusterOpts       e2eutil.PlatformAgnosticOptions
+	expectVersionSkew bool
```
enxebre, line 21: should this be named supportedVersionSkew=true/false
or expectedSupportedVersionSkew=true/false
?

test/e2e/nodepool_test.go
```diff
@@ -115,11 +115,11 @@ func TestNodePool(t *testing.T) {
 					},
 					{
 						name: "TestNodePoolPrevReleaseN1",
-						test: NewNodePoolPrevReleaseCreateTest(hostedCluster, globalOpts.N1MinorReleaseImage, clusterOpts),
+						test: NewNodePoolPrevReleaseCreateTest(hostedCluster, globalOpts.N1MinorReleaseImage, clusterOpts, true),
 					},
 					{
 						name: "TestNodePoolPrevReleaseN2",
```
enxebre, line 121: with the policy being proposed in this PR, wouldn't this depend if the hc version is even or odd?

LiangquanLi930: Even versions support N-2, while odd versions do not.
So, if hc is an even version:
```
{
	name: "TestNodePoolPrevReleaseN1",
	test: NewNodePoolPrevReleaseCreateTest(hostedCluster, globalOpts.N1MinorReleaseImage, clusterOpts, true),
},
{
	name: "TestNodePoolPrevReleaseN2",
	test: NewNodePoolPrevReleaseCreateTest(hostedCluster, globalOpts.N2MinorReleaseImage, clusterOpts, true),
}
```
But if hc is an odd version, it should be:
```
{
	name: "TestNodePoolPrevReleaseN1",
	test: NewNodePoolPrevReleaseCreateTest(hostedCluster, globalOpts.N1MinorReleaseImage, clusterOpts, true),
},
{
	name: "TestNodePoolPrevReleaseN2",
	test: NewNodePoolPrevReleaseCreateTest(hostedCluster, globalOpts.N2MinorReleaseImage, clusterOpts, false),
}
```

api/hypershift/v1beta1/nodepool_conditions.go
```diff
@@ -18,6 +18,14 @@ const (
 	// NodePoolValidTuningConfigConditionType signals if the content within nodePool.spec.tuningConfig is valid.
 	// A failure here is unlikely to resolve without changing user input.
 	NodePoolValidTuningConfigConditionType = "ValidTuningConfig"
+	// NodePoolSupportedVersionSkewConditionType signals if the NodePool points to a version that falls within the supported skew policy with the HostedCluster.
+	// NodePool version cannot be higher than the HostedCluster version.
+	// For 4.y versions:
+	// - For 4.even versions (e.g. 4.18), allows up to 2 minor version differences (4.17, 4.16)
+	// - For 4.odd versions (e.g. 4.17), allows up to 1 minor version difference (4.16)
+	// When false, the NodePool will keep trying to operate as usual even though there are no guarantees
```
wking, line 26: nit: this line should grow a trailing period, in case consumers like https://pkg.go.dev/github.com/openshift/hypershift/api/hypershift/v1beta1#pkg-constants decide to start unwrapping Godocs for constants the way they already do for functions.

hypershift-operator/controllers/nodepool/conditions.go
```diff
@@ -852,3 +854,103 @@ func (r NodePoolReconciler) validPlatformConfigCondition(ctx context.Context, no
...
+	if controlPlaneVersion.Minor%2 == 1 {
+		maxAllowedDiff = 1
+	}
+
+	if versionDiff > maxAllowedDiff {
+		SetStatusCondition(&nodePool.Status.Conditions, hyperv1.NodePoolCondition{
+			Type:   hyperv1.NodePoolSupportedVersionSkewConditionType,
+			Status: corev1.ConditionFalse,
+			Reason: hyperv1.NodePoolUnsupportedSkewReason,
+			Message: fmt.Sprintf("NodePool minor version %d.%d is not compatible with the HostedCluster minor version %d.%d (max allowed difference: %d)",
```
wking, line 940: nit: this generates messages like:

> NodePool minor version 4.15 is not compatible with the HostedCluster minor version 4.17 (max allowed difference: 1)

But that doesn't explain exactly how the difference integer fits in.  It might be easier to read if we construct the minimum allowed 4.y, like:

```golang
fmt.Sprintf("NodePool minor version %d.%d is less than %.%, which is the minimum NodePool version compatible with the %d.%d HostedCluster",
	nodePoolVersion.Major, nodePoolVersion.Minor,
	controlPlaneVersion.Major, controlPlaneVersion.Minor - maxAllowedDiff,
	controlPlaneVersion.Major, controlPlaneVersion.Minor)
```

to generate messages like:

> NodePool minor version 4.15 is less than 4.16, which is is the minimum NodePool version compatible with the 4.17 HostedCluster

enxebre:
/hold
until we have agreement on the policy, then lets consolidate back the PR and make sure it only impact versions >= N

docs/content/how-to/common/nodepool-version-compatibility.md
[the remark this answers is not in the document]
jparrill: IMHO it's part of the product references, not a how-to guide.

docs/content/how-to/common/nodepool-version-compatibility.md
```diff
@@ -0,0 +1,37 @@
...
+1. Set appropriate error conditions in both NodePool status
+2. Provide detailed error messages explaining the specific version incompatibility reason
+
+## End user Attention
+1. Ensure version compatibility when creating new NodePools
+2. Check version compatibility of all NodePools when upgrading HostedCluster
```
jparrill: In this case: 
- how will be the behavior of the HO?
- how we alert the user that there is an issue with the upgrade?
- Are we blocking the upgrade with a failure (I would expect that)? or we proceed and let the NodePool error to pop up in the status conditions

hypershift-operator/controllers/nodepool/conditions.go
```diff
@@ -852,3 +854,102 @@ func (r NodePoolReconciler) validPlatformConfigCondition(ctx context.Context, no
...
+func (r *NodePoolReconciler) supportedVersionSkewCondition(ctx context.Context, nodePool *hyperv1.NodePool, hcluster *hyperv1.HostedCluster) (*ctrl.Result, error) {
+	if hcluster.Status.Version == nil {
+		SetStatusCondition(&nodePool.Status.Conditions, hyperv1.NodePoolCondition{
+			Type:               hyperv1.NodePoolSupportedVersionSkewConditionType,
+			Status:             corev1.ConditionUnknown,
+			Reason:             hyperv1.NodePoolValidationFailedReason,
+			Message:            "HostedCluster version is not available yet, waiting for version information",
+			ObservedGeneration: nodePool.Generation,
+		})
+		return &ctrl.Result{}, fmt.Errorf("HostedCluster version is not available yet")
```
jparrill: Are we causing a short circuit returning this error, until the NP has a version set?

hypershift-operator/controllers/nodepool/conditions.go
jparrill: Agree that is unlikely to happen but also does not harm to be added as a msg of the condition in case of that parsing fails, it's a more deterministic approach.

hypershift-operator/controllers/nodepool/conditions.go
```diff
@@ -852,3 +854,102 @@ func (r NodePoolReconciler) validPlatformConfigCondition(ctx context.Context, no
...
+		return nil, nil
+	}
+
+	// Calculate minor version difference
+	versionDiff := int(controlPlaneVersion.Minor) - int(nodePoolVersion.Minor)
+
+	// All 4.y versions now support n-3 version skew
+	maxAllowedDiff := 3
+
+	if versionDiff > maxAllowedDiff {
```
jparrill: I would isolate this logic in a separated place to be reused in other places. Also could return a boolean that tells directly if the condition is true or false.

hypershift-operator/controllers/nodepool/nodepool_controller_test.go
```diff
@@ -1487,3 +1488,258 @@ func Test_validateHCPayloadSupportsNodePoolCPUArch(t *testing.T) {
...
+				Type:               hyperv1.NodePoolSupportedVersionSkewConditionType,
+				Status:             corev1.ConditionFalse,
+				Reason:             hyperv1.NodePoolUnsupportedSkewReason,
+				Message:            "NodePool minor version 4.14 is less than 4.15, which is the minimum NodePool version compatible with the 4.18 HostedCluster",
+				ObservedGeneration: 0,
+			},
+			expectedError: "",
+		},
+	}
+
```
jparrill, line 1760: I miss a case with Z versions

hypershift-operator/controllers/nodepool/conditions.go
LiangquanLi930: Yes, it short-circuits when `hcluster.Status.Version == nil`, because this check depends on the HC’s version.  So I added the short-circuit.  IIUC, the NodePool reconciliation should also wait until the control plane is ready.  I also noticed the `IgnitionEndpointAvailableCondition`, which similarly waits for the control plane to be ready.

docs/content/how-to/common/nodepool-version-compatibility.md
LiangquanLi930: Currently, we only notify users about version incompatibility through the NodePool’s status conditions; we don’t short-circuit the process.

hypershift-operator/controllers/nodepool/nodepool_controller_test.go
LiangquanLi930: updated

hypershift-operator/controllers/nodepool/conditions.go
```diff
@@ -852,3 +855,92 @@ func (r NodePoolReconciler) validPlatformConfigCondition(ctx context.Context, no
...
+			Status:             corev1.ConditionUnknown,
+			Reason:             hyperv1.NodePoolValidationFailedReason,
+			Message:            fmt.Sprintf("Failed to get release image: %v", err.Error()),
+			ObservedGeneration: nodePool.Generation,
+		})
+		return &ctrl.Result{}, fmt.Errorf("failed to look up release image metadata: %w", err)
+	}
+
+	// Parse NodePool version
+	nodePoolVersion, err := semver.Parse(releaseImage.Version())
```
jparrill: maybe it's more readable to instantiate the NodePoolCondition before all the conditionals and only modify the relevant fields on each scenario

hypershift-operator/controllers/nodepool/conditions.go
```diff
@@ -852,3 +855,92 @@ func (r NodePoolReconciler) validPlatformConfigCondition(ctx context.Context, no
...
+	if err != nil {
+		SetStatusCondition(&nodePool.Status.Conditions, hyperv1.NodePoolCondition{
```
jparrill, line 917: non-blocking nit, maybe it's more readable to instantiate the NodePoolCondition before all the conditionals and only modify the relevant fields on each scenario, something like:

```suggestion
		condition := &hyperv1.NodePoolCondition{
		    Type:               hyperv1.NodePoolSupportedVersionSkewConditionType,
			ObservedGeneration: nodePool.Generation,
		}
		
		// Parse NodePool version
	    nodePoolVersion, err := semver.Parse(releaseImage.Version())
	    if err != nil {
	        condition.Status := xxxx
	        condition.Reason := xxxx
	        condition.Message := xxxx
	        SetStatusCondition(&nodePool.Status.Conditions, condition)
	    }
	    ...
	    ...
```

jparrill (commented):
Dropped a small nit, otherwise lgtm

hypershift-operator/controllers/nodepool/conditions.go
LiangquanLi930: Thanks for the suggestion! In the current implementation, each condition is only created once within its own branch, so there's no runtime overhead. I find this style more readable as each error path is self-contained and explicit. I'd prefer to keep it as is. Thanks!

hypershift-operator/controllers/nodepool/conditions.go
```diff
@@ -852,3 +855,92 @@ func (r NodePoolReconciler) validPlatformConfigCondition(ctx context.Context, no
...
+
+	// Get the control plane version string
+	var controlPlaneVersionStr string
+	if len(hcluster.Status.Version.History) == 0 {
+		// If the cluster is in the process of installation, there is no history
+		// Use the desired version as the control plane version
+		controlPlaneVersionStr = hcluster.Status.Version.Desired.Version
+	} else {
+		// If the cluster is installed or upgrading
+		// Start with the most recent version from history as the default
```
csrwng, line 879: nit: update the comment if the intent is to start with the oldest version in history

hypershift-operator/controllers/nodepool/conditions.go
```diff
@@ -852,3 +855,92 @@ func (r NodePoolReconciler) validPlatformConfigCondition(ctx context.Context, no
 	SetStatusCondition(&nodePool.Status.Conditions, *condition)
 	return nil, nil
 }
+
+func (r *NodePoolReconciler) supportedVersionSkewCondition(ctx context.Context, nodePool *hyperv1.NodePool, hcluster *hyperv1.HostedCluster) (*ctrl.Result, error) {
+	if hcluster.Status.Version == nil {
+		SetStatusCondition(&nodePool.Status.Conditions, hyperv1.NodePoolCondition{
```
csrwng, line 861: nit: The code might be clearer if there was an inner function that returns (bool, error) and the condition is constructed once based on that, instead of constructing the condition on every error condition.

hypershift-operator/controllers/nodepool/conditions.go
```diff
@@ -852,3 +855,92 @@ func (r NodePoolReconciler) validPlatformConfigCondition(ctx context.Context, no
...
+			Type:               hyperv1.NodePoolSupportedVersionSkewConditionType,
+			Status:             corev1.ConditionUnknown,
+			Reason:             hyperv1.NodePoolValidationFailedReason,
+			Message:            "HostedCluster version is not available yet, waiting for version information",
+			ObservedGeneration: nodePool.Generation,
+		})
+		return &ctrl.Result{}, fmt.Errorf("HostedCluster version is not available yet")
```
csrwng: An error returned from this function short circuits other condition checking. Do we always want to short circuit other conditions, or should the error be reported in the condition itself and continue to check the others?

support/supportedversion/version.go
```diff
@@ -292,6 +292,41 @@ type ocpTags struct {
...
+// Returns nil if the version skew is supported, otherwise returns a descriptive error.
+// All 4.y versions support n-3 version skew (e.g., 4.18 HostedCluster supports NodePools running 4.17, 4.16, and 4.15).
+func ValidateVersionSkew(hostedClusterVersion, nodePoolVersion *semver.Version) error {
+	// Reject mismatched major versions
+	if nodePoolVersion.Major != hostedClusterVersion.Major {
+		return fmt.Errorf("NodePool major version %d must match HostedCluster major version %d",
+			nodePoolVersion.Major, hostedClusterVersion.Major)
+	}
+
+	if nodePoolVersion.GT(*hostedClusterVersion) {
```
csrwng, line 305: This means that if the nodepool and hc have the same minor but the np has a newer patch release, we consider it invalid, is that the intent?

test/e2e/nodepool_test.go
```diff
@@ -119,11 +119,19 @@ func TestNodePool(t *testing.T) {
...
-						test: NewNodePoolPrevReleaseCreateTest(hostedCluster, globalOpts.N2MinorReleaseImage, clusterOpts),
+						test: NewNodePoolPrevReleaseCreateTest(hostedCluster, globalOpts.N2MinorReleaseImage, clusterOpts, true),
+					},
+					{
+						name: "TestNodePoolPrevReleaseN3",
+						test: NewNodePoolPrevReleaseCreateTest(hostedCluster, globalOpts.N3MinorReleaseImage, clusterOpts, true),
+					},
+					{
+						name: "TestNodePoolPrevReleaseN4",
+						test: NewNodePoolPrevReleaseCreateTest(hostedCluster, globalOpts.N4MinorReleaseImage, clusterOpts, false),
```
csrwng, line 134: Would we always expect this test to pass? If the release is outside of the supported version skew, it would be possible that nodes do not become available, and that would be ok, no?

hypershift-operator/controllers/nodepool/conditions.go
LiangquanLi930: According to the API docs, History is ordered by recency with the newest update first. For example:
```yaml
  history:
    - completionTime: "2025-11-17T12:53:39Z"  # History[0] - newest
      state: Completed
      version: 4.21.0-0.ci-2025-11-17-064207
    - completionTime: "2025-11-17T12:31:36Z"  # History[1] - older
      state: Completed
      version: 4.20.0-0.nightly-2025-11-15-175712
```
updated it.

hypershift-operator/controllers/nodepool/conditions.go
LiangquanLi930: Changed all error returns to `nil, nil` to avoid short-circuiting other condition checks.

support/supportedversion/version.go
LiangquanLi930: @wking Do you know which strategy OCP is currently using? Should we be comparing the patch version instead?

wking: OCP doesn't care about the patch version, but there's also no supported way to get the compute Nodes out to a patch version that's greater than the control-plane's version (e.g. a 4.y.z control plane with 4.y.(z' > z) compute).  HyperShift [currently claims][1] to not support that patch skew, although personally:

* I don't see why we wouldn't support it, because the point of 4.y z streams is that we aren't changing cross-component APIs within the z stream.  However, there are also other components like OVN that [currently claim the ability to bump APIs or make other incompatible changes within a z stream][2], so... that makes it hard to know how much of the "4.y z streams don't bump APIs" is delivered policy vs. aspirational dreaming.
* If you _do_ care about patch skew for control-plane/compute interaction, minor-version skew becomes very sticky.  A 4.20.0 control plane needs to support 4.19.15 compute, to allow 4.19.15 -> 4.20.0 updates.  But does a 4.20.0 control plane support 4.19.0 compute (4.19.0 -> 4.20.0 control-plane updates are not supported in a single hop, you need to pass through 4.19.15 first, openshift/cincinnati-graph-data#8106)?  Does it support 4.19.19, even though 4.19.19 is about a month younger:

    ```console
    $ for V in 4.20.0 4.19.17 4.19.19; do echo -n "$V "; oc adm release info -o json "quay.io/openshift-release-dev/ocp-release:${V}-x86_64" | jq -r .config.created; done
    4.20.0 2025-10-16T13:35:24Z
    4.19.17 2025-10-16T15:34:43Z
    4.19.19 2025-11-14T00:12:22Z
    ```

    If we support 4.19.19 compute against a 4.20.0 control plane, why wouldn't we support 4.19.19 compute against a 4.19.17 control-plane, since 4.19.17 is a more recent build than 4.20.0?

I think the reasoning becomes much clearer if we commit to not caring about patch skew, and only focus on 4.y vs. 4.y' minor-version skew.  But that means committing to not changing APIs within a 4.y z-stream in a way that would cause compatibility to care about patch versions.  And it means we'd want to test new 4.y.z candidate releases with both "new 4.y.z HostedCluster and old 4.y.0 NodePools" and "old 4.y.0 HostedCluster and new 4.y.z NodePools" and similar for 4.y' NodePools within the minor-version skew constraints.  Likely with a more nuanced test suite, because 4.y.0 HostedClusters and NodePools will be exposed to bugs fixed by 4.y.z.

[1]: https://issues.redhat.com/browse/RFE-6410
[2]: https://issues.redhat.com/browse/RFE-5522

wking: As for what standalone does, the Kube API server guard is a bit different, because it [only uses the OCP version to calculate the allowed minor-version skew][1].  The actual skew calculation is just [looking at Kube API server and kubelet versions][2], so it will still set `Upgradeable=False` in cases like "4.odd ClusterVersion, 4.odd MachineConfigPool _targets_, some Nodes still yet to update and still running 4.(odd-1)'s kubelet version".  There's no MachineConfigPool knob to aim at a given version today though, so they don't need constraints on their MCP controller, like you'll want constraints on your NodePool controller.  But you may also want constraints on the HostedCluster controller (maybe the HostedControlPlane operator chipping into `ClusterVersionUpgradeable`?) to point out to cluster owners when they have kubelet Nodes on versions that would outstretch the minor-version skew limits if the HostedCluster was updated?  With a way to push past that guard in cases where the cluster owner felt that violating the skew was worth the risk (e.g. to update the bits running on the management cluster before the control-plane's 4.y went end-of-life).

[1]: https://github.com/openshift/cluster-kube-apiserver-operator/blob/aa82024dd2374f1561508f19a656366840619ed2/pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller.go#L70-L79
[2]: https://github.com/openshift/cluster-kube-apiserver-operator/blob/aa82024dd2374f1561508f19a656366840619ed2/pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller.go#L112-L123

csrwng: My preference here would be to default to being cautious, given that upstream kubernetes doesn't mention anything about supporting patch version skew. It simply states that the [kubelet cannot be newer than the kube apiserver](https://kubernetes.io/releases/version-skew-policy/#kubelet). If we're going to support newer patch versions, we will need to do some kind of validation to make sure that's actually going to work.

LiangquanLi930: Thanks for the discussion! Based on consensus, keep the current strict validation (full version comparison).

Once we’ve validated the NP.patch > HC.patch scenario in the future, we can come back and adjust this part of the code.

test/e2e/nodepool_test.go
LiangquanLi930: Thanks! updated. For N-4, now only validates the condition is set to False and skips node readiness checks.

jparrill:
Please @LiangquanLi930 the CLI is evaluating different the skew between the HC and NP versions:

```
λ hypershift git:(OCPBUGS-65713) /Users/jparrill/RedHat/RedHat_Engineering/hypershift/repos/hypershift/bin/hypershift create nodepool aws --auto-repair --cluster-name jparrill-hosted --name jparrill-hosted-4-18 --namespace clusters --node-upgrade-type replace --replicas 1 --release-image quay.io/openshift-release-dev/ocp-release:4.18.28-x86_64
{"level":"error","ts":"2025-11-19T18:42:58+01:00","msg":"Failed to create nodepool","error":"NodePool minor version 4.18 is not compatible with the HostedCluster minor version 4.21 (max allowed difference: 1)","stacktrace":"github.com/openshift/hypershift/cmd/nodepool/core.(*CreateNodePoolOptions).CreateRunFunc.func1\n\t/Users/jparrill/RedHat/RedHat_Engineering/hypershift/repos/hypershift/cmd/nodepool/core/create.go:48\ngithub.com/spf13/cobra.(*Command).execute\n\t/Users/jparrill/RedHat/RedHat_Engineering/hypershift/repos/hypershift/vendor/github.com/spf13/cobra/command.go:1015\ngithub.com/spf13/cobra.(*Command).ExecuteC\n\t/Users/jparrill/RedHat/RedHat_Engineering/hypershift/repos/hypershift/vendor/github.com/spf13/cobra/command.go:1148\ngithub.com/spf13/cobra.(*Command).Execute\n\t/Users/jparrill/RedHat/RedHat_Engineering/hypershift/repos/hypershift/vendor/github.com/spf13/cobra/command.go:1071\ngithub.com/spf13/cobra.(*Command).ExecuteContext\n\t/Users/jparrill/RedHat/RedHat_Engineering/hypershift/repos/hypershift/vendor/github.com/spf13/cobra/command.go:1064\nmain.main\n\t/Users/jparrill/RedHat/RedHat_Engineering/hypershift/repos/hypershift/main.go:78\nruntime.main\n\t/opt/homebrew/opt/go/libexec/src/runtime/proc.go:285"}
Error: NodePool minor version 4.18 is not compatible with the HostedCluster minor version 4.21 (max allowed difference: 1)
NodePool minor version 4.18 is not compatible with the HostedCluster minor version 4.21 (max allowed difference: 1)
```

Also this Bug (which is a release blocker) was created yesterday [OCPBUGS-65713](https://issues.redhat.com/browse/OCPBUGS-65713) and you PR mostly address the situation. 

I'm trying to reproduce the issue.

LiangquanLi930:
Based on discussion, we’ll fix the CLI bug in a separate PR.

PR #5943, opened 2025-03-31
OCPBUGS-53206: Use a CPO image label to determine whether to run expired cert remediation

csrwng:
Adds a commented label to the CPO image that should be uncommented whenever we switch to using CPO v2 as the default. Uses that label (and annotation) to determine whether we should remediate expired certs from the hypershift operator.

We were previously using the HC version, assuming that 4.19 would default to CPOv2, but that won't be the case.

Fixes #[OCPBUGS-53206](https://issues.redhat.com/browse/OCPBUGS-53206)

**Checklist**


csrwng:

PR #5944, opened 2025-03-31
WIP destroy

bryan-cox:

Fixes #

**Checklist**


bryan-cox:
/close 

I looked at the os4-common resource group today and we are not leaking any role assignments now. It doesn't seem necessary to delete the role assignments for the service principals serving as the control plane and data plane managed identities. I'm going to close this for now. This was almost all the way working except for the DNS RG not being passed to the infra destroy options for some reason.

PR #5949, opened 2025-04-01
OCPBUGS-13292: Add functionality to limit CAPI CRD installation on HO installation

bryan-cox:
This PR adds a new flag to the HO installation, limit-crd-install. This flag allows a user to limit what CAPI CRDs are installed per platform basis.

Fixes #

**Checklist**

### DIFF
--- a/cmd/install/assets/assets.go
+++ b/cmd/install/assets/assets.go
@@ -23,7 +23,7 @@ import (
 //go:embed cluster-api-provider-agent/*
 //go:embed cluster-api-provider-azure/*
 //go:embed cluster-api-provider-openstack/*
-var crds embed.FS
+var CRDS embed.FS
 
 //go:embed recordingrules/*
 var recordingRules embed.FS
@@ -82,14 +82,14 @@ func getContents(fs embed.FS, file string) []byte {
 // CustomResourceDefinitions returns all existing CRDs as controller-runtime objects
 func CustomResourceDefinitions(include func(path string, crd *apiextensionsv1.CustomResourceDefinition) bool, transform func(*apiextensionsv1.CustomResourceDefinition)) []crclient.Object {
 	var allCrds []crclient.Object
-	err := fs.WalkDir(crds, ".", func(path string, d fs.DirEntry, err error) error {
+	err := fs.WalkDir(CRDS, ".", func(path string, d fs.DirEntry, err error) error {
 		if err != nil {
 			panic(err)
 		}
 		if filepath.Ext(path) != ".yaml" {
 			return nil
 		}
-		crd := getCustomResourceDefinition(crds, path)
+		crd := getCustomResourceDefinition(CRDS, path)
 		if include(path, crd) {
 
 			if transform != nil {
--- a/cmd/install/install.go
+++ b/cmd/install/install.go
@@ -46,6 +46,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/apimachinery/pkg/util/yaml"
 	"k8s.io/utils/ptr"
+	"k8s.io/utils/set"
 
 	crclient "sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
@@ -58,6 +59,17 @@ const (
 	// ExternalDNSImage - This is specifically tag 1.1.0-3 from https://catalog.redhat.com/software/containers/edo/external-dns-rhel8/61d4c35023156829b87a434a?container-tabs=overview&tag=1.1.0-3&push_date=1671131187000
 	// TODO this needs to be updated to a multi-arch image including Arm - https://issues.redhat.com/browse/NE-1298
 	ExternalDNSImage = "registry.redhat.io/edo/external-dns-rhel8@sha256:638fb6b5fc348f5cf52b9800d3d8e9f5315078fc9b1e57e800cb0a4a50f1b4b9"
+)
+
+// ValidPlatforms should contain all the CAPI provider types we support; see also crds in assests.go
+// https://github.com/openshift/hypershift/blob/3ea313694d386763578646b157a8d4d3d187e98e/cmd/install/assets/assets.go#L26
+var ValidPlatforms = set.New[string](
+	"aws",
+	"azure",
+	"ibmcloud",
+	"kubevirt",
+	"agent",
+	"openstack",
 )
 
 type Options struct {
@@ -113,6 +125,7 @@ type Options struct {
 	TechPreviewNoUpgrade                      bool
 	RegistryOverrides                         string
 	RenderNamespace                           bool
+	PlatformsToInstall                        []string
 }
 
 func (o *Options) Validate() error {
@@ -167,6 +180,14 @@ func (o *Options) Validate() error {
 
 	if len(o.ManagedService) > 0 && o.ManagedService != hyperv1.AroHCP {
 		errs = append(errs, fmt.Errorf("not a valid managed service type: %s", o.ManagedService))
+	}
+
+	// Validate all the platforms in the list are valid
+	for _, platform := range o.PlatformsToInstall {
+		platformToCheck := strings.ToLower(platform)
+		if !ValidPlatforms.Has(platformToCheck) {
+			errs = append(errs, fmt.Errorf("not a valid platform type: %s", platform))
+		}
 	}
 	return errors.NewAggregate(errs)
 }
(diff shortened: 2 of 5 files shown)


cmd/install/install.go
```diff
@@ -242,6 +259,7 @@ func NewCommand() *cobra.Command {
...
 	cmd.PersistentFlags().BoolVar(&opts.TechPreviewNoUpgrade, "tech-preview-no-upgrade", opts.TechPreviewNoUpgrade, "If true, the HyperShift operator runs with TechPreviewNoUpgrade features enabled")
 	cmd.PersistentFlags().StringVar(&opts.RegistryOverrides, "registry-overrides", "", "registry-overrides contains the source registry string as a key and the destination registry string as value. Images before being applied are scanned for the source registry string and if found the string is replaced with the destination registry string. Format is: sr1=dr1,sr2=dr2")
+	cmd.PersistentFlags().StringSliceVar(&opts.PlatformsToInstall, "limit-crd-install", opts.PlatformsToInstall, "Used to limit the CAPI CRDs that are installed to a per platform basis (example: --limit-crd-install=AWS,Azure). If this flag is not specified, all CRDs for all platforms will be installed. Valid, case-insensitive values are: AWS, Azure, IBMCloud, KubeVirt, Agent, OpenStack.")
```
enxebre, line 262: this is an impl detail, let's don't surface that detail with a flag

cmd/install/assets/assets.go
```diff
@@ -15,18 +15,20 @@ import (
...
-//go:embed hypershift-operator/*
-//go:embed cluster-api/*
-//go:embed cluster-api-provider-aws/*
-//go:embed cluster-api-provider-ibmcloud/*
-//go:embed cluster-api-provider-kubevirt/*
-//go:embed cluster-api-provider-agent/*
-//go:embed cluster-api-provider-azure/*
-//go:embed cluster-api-provider-openstack/*
-var crds embed.FS
+var crds = []string{
```
enxebre: this would force us include this files within the image file system and anywhere you would want to run the cli install binary.
Let's get back to embed and we can still discriminate at runtime, see func setupCRDs( for some precedents

cmd/install/install.go
```diff
@@ -168,6 +169,22 @@ func (o *Options) Validate() error {
 	if len(o.ManagedService) > 0 && o.ManagedService != hyperv1.AroHCP {
 		errs = append(errs, fmt.Errorf("not a valid managed service type: %s", o.ManagedService))
 	}
+
+	validPlatforms := map[string]struct{}{
```
muraee, line 173: can we use a `sets.New` here instead of a map.

cmd/install/install.go
bryan-cox: Following up here - we discussed offline that it might be good to keep this as a flag for now even.

cmd/install/install.go
```diff
@@ -242,6 +263,7 @@ func NewCommand() *cobra.Command {
...
 	cmd.PersistentFlags().BoolVar(&opts.TechPreviewNoUpgrade, "tech-preview-no-upgrade", opts.TechPreviewNoUpgrade, "If true, the HyperShift operator runs with TechPreviewNoUpgrade features enabled")
 	cmd.PersistentFlags().StringVar(&opts.RegistryOverrides, "registry-overrides", "", "registry-overrides contains the source registry string as a key and the destination registry string as value. Images before being applied are scanned for the source registry string and if found the string is replaced with the destination registry string. Format is: sr1=dr1,sr2=dr2")
+	cmd.PersistentFlags().StringSliceVar(&opts.PlatformsToInstall, "limit-crd-install", opts.PlatformsToInstall, "Used to limit the CAPI CRDs that are installed to a per platform basis (example: --limit-crd-install=AWS,Azure). If this flag is not specified, all CRDs for all platforms will be installed. Valid, case-insensitive values are: AWS, Azure, IBMCloud, KubeVirt, Agent, OpenStack.")
```
enxebre, line 266: the flag docs mentions capi, do we want to let it be generic? i.e Used to limit the CRDs that are installed to a per platform

cmd/install/install_test.go
```diff
@@ -141,15 +161,38 @@ func TestSetupCRDs(t *testing.T) {
...
-			g.Expect(awsEndpointServicesCRD).ToNot(BeNil())
 
 			// Validate the feature set specific CRDs are applied.
 			g.Expect(nodePoolCRDS).To(HaveLen(1))
 			if tc.inputOptions.TechPreviewNoUpgrade {
 				g.Expect(nodePoolCRDS[0].GetAnnotations()["release.openshift.io/feature-set"]).To(Equal("TechPreviewNoUpgrade"))
 				return
 			}
+			if tc.inputOptions.PlatformsToInstall != nil {
+				if len(tc.inputOptions.PlatformsToInstall) > 1 {
```
enxebre, line 172: please avoid nesting for easier readability.

You can actually achieve the same coverage with

```
			// Compute wanted and unwanted platforms based on the test input.
			allPlatforms := set.New[string]("AWS", "Azure", "OpenStack", "Agent", "KubeVirt", "IBMCloud")
			wantedPlatforms := allPlatforms
			unwantedPlatforms := set.New[string]()
			if tc.inputOptions.PlatformsToInstall != nil {
				wantedPlatforms = set.New[string](tc.inputOptions.PlatformsToInstall...)
				unwantedPlatforms = allPlatforms.Difference(wantedPlatforms)
			}

			// Validate that no unwanted platform CRDs are present.
			for _, crd := range gotCRDs {
				for unwantedPlatform := range unwantedPlatforms {
					g.Expect(strings.ToLower(crd.GetName())).NotTo(ContainSubstring(strings.ToLower(unwantedPlatform)), "Found unwanted platform CRD")
				}
			}
```

And you can also add coverage for the wanted CRDs
```
			// Validate that all wanted platform CRDs are present.
			for platform := range wantedPlatforms {
				wantedCAPICRDsPerPlatform, err := fs.ReadDir(assets.CRDS, "cluster-api-provider-"+strings.ToLower(platform))
				if err == nil {
					var yamlFiles []fs.DirEntry
					for _, file := range wantedCAPICRDsPerPlatform {
						if filepath.Ext(file.Name()) == ".yaml" {
							yamlFiles = append(yamlFiles, file)
						}
					}
					wantedCAPICRDsPerPlatform = yamlFiles
				}

				gotCRDsPerPlatform := make([]string, 0)
				if platform == "IBMCloud" {
					platform = "ibm"
				}
				for _, crd := range gotCRDs {
					if strings.Contains(strings.ToLower(crd.GetName()), strings.ToLower(platform)) {
						gotCRDsPerPlatform = append(gotCRDsPerPlatform, crd.GetName())
					}
				}

				g.Expect(len(wantedCAPICRDsPerPlatform)).To(BeNumerically("<=", len(gotCRDsPerPlatform)), "Missing CRDs for platform %s", platform)
			}

			if wantedPlatforms.Has("AWS") {
				g.Expect(awsEndpointServicesCRD).ToNot(BeNil())
			}
```

cmd/install/install.go
```diff
@@ -630,8 +654,21 @@ func setupCRDs(opts Options, operatorNamespace *corev1.Namespace, operatorServic
 						}
 					}
 				}
+				if strings.Contains(path, "hypershift-operator/") {
```
enxebre, line 657: while this is ok I would expect we keep the pattern of default to true and cascade check to return false, e.g.
```
				if strings.Contains(path, "etcd") && opts.ExcludeEtcdManifests {
					return false
				}
				// Feature Generated CRDs are in
				// If the feature generated CRD has any featureSet version then it has the format nodepool-<featureSet>.
				if strings.Contains(path, "zz_generated.crd-manifests") {
					// Filter by featureSet.
					if !filterByFeatureSet(crd) {
						return false
					}

					if !filterByPlatform(crd) {
						return false
					}
				}

				if strings.Contains(path, "cluster-api-provider") {
					if len(opts.PlatformsToInstall) > 0 {
						for _, platform := range opts.PlatformsToInstall {
							if !strings.Contains(path, strings.ToLower(platform)) {
								return false
							}
						}
					}
				}
				return true
```

This is not a blocker to me though feel free to keep it as it is

cmd/install/install.go
```diff
@@ -168,6 +170,25 @@ func (o *Options) Validate() error {
...
+		"ibmcloud",
+		"kubevirt",
+		"agent",
+		"openstack",
+	)
+
+	// Validate all the platforms in the list are valid
+	for _, platform := range o.PlatformsToInstall {
+		platformToCheck := strings.ToLower(platform)
+		if _, exists := validPlatforms[platformToCheck]; !exists {
```
muraee, line 188:
```suggestion
		if !validPlatforms.Has(platformToCheck) {
```

enxebre:
thanks!
/approve

bryan-cox:

PR #5951, opened 2025-04-01
NO-JIRA: set default expriationDate tag in AWS if not set

sjenning:
Had a recent issue where new dev ran e2e locally against AWS infra without including an expirationDate tag causing orphaned resources


sjenning:

PR #5674, opened 2025-02-20
NO-JIRA: ho: watch AWSEndpointServices

sjenning:
https://github.com/openshift/hypershift/pull/2278 added code for propagating the conditions on the `AWSEndpointService` up to the `HostedCluster`.  However, the `HostedClusterReconciler` does not watch the `AWSEndpointService` type and thus can miss updates to the conditions.

This PR adds `AWSEndpointService` to the `managedResources` that are watched by the `HostedClusterReconciler`.

cc @muraee

PR #5960, opened 2025-04-01
OCPBUGS-52984: annotate AWSEndpointServices with HostedClusterAnnotation

sjenning:
Follow on to https://github.com/openshift/hypershift/pull/5674

Previous PR added `AWSEndpointService` objects to the `HostedCluster` controller Watch, however, `enqueueHostedClustersFunc()` is unable to map to a `HostedCluster` to reconcile because the `AWSEndpointService` type does not have an explicit mapping method or the `HostedClusterAnnotation`.

This PR adds the `HostedClusterAnnotation` to the `AWSEndpointService` so that `handleDefault` can map it to a `HostedCluster`

https://github.com/openshift/hypershift/blob/dd7cb9264226444e5026d67575562b11a932aa38/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go#L4194-L4205


enxebre:
how can we prevent regress on this and prevent the same from happening from new resources?
should we have an e2e that loops over r.managedResources and checks the annotation?

sjenning:

PR #5965, opened 2025-04-02
OCPBUGS-54439: Change KASGoMemLimit to a string pointer type

bryan-cox:
This PR changes the KASGoMemLimit to a string pointer type instead of a k8s api/resource Quantity pointer type and updates related unit tests.

Fixes #

**Checklist**

### DIFF
--- a/api/scheduling/v1alpha1/clustersizingconfiguration_types.go
+++ b/api/scheduling/v1alpha1/clustersizingconfiguration_types.go
@@ -109,9 +109,13 @@ type NodeCountCriteria struct {
 type Effects struct {
 
 	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Pattern=`^\d+(B|KiB|MiB|GiB|TiB)?$`
+	// +kubebuilder:validation:MaxLength=20
 
 	// KASGoMemLimit is the value to set for the $GOMEMLIMIT of the Kube APIServer container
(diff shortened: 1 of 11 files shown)


api/scheduling/v1alpha1/clustersizingconfiguration_types.go
```diff
@@ -111,7 +111,7 @@ type Effects struct {
 	// +kubebuilder:validation:Optional
 
 	// KASGoMemLimit is the value to set for the $GOMEMLIMIT of the Kube APIServer container
-	KASGoMemLimit *resource.Quantity `json:"kasGoMemLimit,omitempty"`
+	KASGoMemLimit *string `json:"kasGoMemLimit,omitempty"`
```
enxebre, line 114: why does this need to be a pointer?
Can you please document in the field the expected format input?
Can you add cel validation for the above that doesn't break existing input in managed environments?

bryan-cox: Discussed offline a bit here - https://redhat-internal.slack.com/archives/G01QS0P2F6W/p1743690038033419?thread_ts=1743515322.678699&cid=G01QS0P2F6W

Just made it a pointer to be consistent with the other strings in the same struct.

Added expected format & CEL.

csrwng:

PR #5775, opened 2025-03-06
ARO-15578: add disable cluster capabilities flag to cluster create cmd

flavianmissi:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes**
Fixes #[ARO-15578](https://issues.redhat.com//browse/ARO-15578)

**Checklist**
- [X] Subject and description added to both, commit and PR.
- [X] Relevant issues have been referenced.
- [ ] This change includes docs. 
- [X] This change includes unit tests.

PR #5969, opened 2025-04-03
CNTRLPLANE-233: Add --kas-dns-name CLI flag to consume KubeAPIServerDNSName API

jparrill:
This PR includes the new flag --kas-dns-name, which enables the customer to create a HostedCluster directly to consume KubeAPIServerDNSName API

Fixes #[CNTRLPLANE-233](https://issues.redhat.com/browse/CNTRLPLANE-233)

### DIFF
--- a/cmd/cluster/core/create.go
+++ b/cmd/cluster/core/create.go
@@ -108,6 +108,7 @@ func bindCoreOptions(opts *RawCreateOptions, flags *pflag.FlagSet) {
 	flags.StringVar(&opts.ReleaseStream, "release-stream", opts.ReleaseStream, "The OCP release stream for the cluster (e.g. 4-stable-multi), this flag is ignored if release-image is set")
 	flags.StringVar(&opts.FeatureSet, "feature-set", opts.FeatureSet, "The predefined feature set to use for the cluster (TechPreviewNoUpgrade or DevPreviewNoUpgrade)")
 	flags.StringSliceVar(&opts.DisableClusterCapabilities, "disable-cluster-capabilities", nil, "Optional cluster capabilities to disabled. The only currently supported value is ImageRegistry.")
+	flags.StringVar(&opts.KubeAPIServerDNSName, "kas-dns-name", opts.KubeAPIServerDNSName, "The custom DNS name for the kube-apiserver service. Make sure the DNS name is valid and addressable.")
 }
 
 // BindDeveloperOptions binds options that should only be exposed to developers in the `hypershift` CLI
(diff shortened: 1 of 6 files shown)


cmd/cluster/core/create.go
```diff
@@ -108,6 +108,7 @@ func bindCoreOptions(opts *RawCreateOptions, flags *pflag.FlagSet) {
 	flags.StringVar(&opts.ReleaseStream, "release-stream", opts.ReleaseStream, "The OCP release stream for the cluster (e.g. 4-stable-multi), this flag is ignored if release-image is set")
 	flags.StringVar(&opts.FeatureSet, "feature-set", opts.FeatureSet, "The predefined feature set to use for the cluster (TechPreviewNoUpgrade or DevPreviewNoUpgrade)")
 	flags.StringSliceVar(&opts.DisableClusterCapabilities, "disable-cluster-capabilities", nil, "Optional cluster capabilities to disabled. The only currently supported value is ImageRegistry.")
+	flags.StringVar(&opts.KubeAPIServerDNSName, "kas-dns-name", opts.KubeAPIServerDNSName, "The custom DNS name for the kube-apiserver service. Make sure the DNS name is valid and addressable.")
```
enxebre, line 111: please include unit tests that validate the produced resource has what's expected when this flag is set

jparrill: done

cmd/cluster/core/create_test.go
```diff
@@ -218,3 +249,120 @@ func TestValidate(t *testing.T) {
 		})
 	}
 }
+
```
enxebre, line 252: Insead of adding this new test can we consolidate with https://github.com/openshift/hypershift/pull/5775/files#diff-23216e73706c06453a939915d3547af12db4de806b4a1a7d81552269a539d5a6
i.e.
- can we please include validation for this field input and cover it within TestValidate

- Can we also add it to TestPrototypeResources

- Can we also include a case in aws/azure to validate the rendered resource against a fixture

jparrill:
override "Red Hat Konflux / hypershift-operator-main-enterprise-contract / hypershift-operator-main"

jparrill:

cmd/cluster/core/create_test.go
```diff
@@ -203,6 +205,16 @@ func TestValidate(t *testing.T) {
 			},
 			expectedErr: "",
 		},
+		{
+			name: "passes with KubeAPIServerDNSName",
```
enxebre, line 209: can we have a negative case, flag input for an invalid dns name?

jparrill: Humm we're not validating that opt in the validate function because we rely in the apimachinery to do that validation at API level.

enxebre: the field is validated with CEL at the API level. What I'm asking in this PR now is to validate the CLI input in func (opts *RawCreateOptions) Validate(ctx context.Context) to fail earlier and cover that in the unit.
just like we do for other fields see https://github.com/openshift/hypershift/blob/627d5bc233260e418706c139ef8b5c7cd781bdde/cmd/cluster/core/create.go#L620-L626

jparrill:

PR #5971, opened 2025-04-03
CNTRLPLANE-334: openshift/api dependency bumps

everettraven:
This PR bumps the openshift/api, openshift/api/tools, and openshift/controller-tools dependencies throughout hypershift to pull in the latest changes.

This is spun out from https://github.com/openshift/hypershift/pull/5840 to have a separation between general dependency bumps and the bumps necessary to add the new external OIDC fields.

Fixes #

**Checklist**


api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/hostedclusters.hypershift.openshift.io/AutoNodeKarpenter.yaml
```diff
@@ -385,11 +386,6 @@ spec:
                                 3. routes.route.openshift.io
                                 4. oauthaccesstokens.oauth.openshift.io
                                 5. oauthauthorizetokens.oauth.openshift.io
-                            enum:
-                            - ""
-                            - identity
-                            - aescbc
-                            - aesgcm
```
everettraven, line 392: Not sure why this ended up getting removed. These values look like they should be here no matter what based on the openshift/api featuregate enum validation https://github.com/openshift/api/blob/50cb52d04f83975bf054f400d27223e8893bb749/config/v1/types_apiserver.go#L212-L214

everettraven: Looking at the resulting merged CRDs, it looks like this set is still included by default. So this shouldn't cause any negative impacts.

PR #5976, opened 2025-04-03
CNTRLPLANE-334: add feature set aware feature gates to HO and CPO

everettraven:
- Adds feature set aware feature gates to the Hypershift Operator and Control Plane Operator. 
- Updates the `hypershift install --tech-preview-no-upgrade ...` command to set Hypershift Operator's feature set as `TechPreviewNoUpgrade`, automatically enabling all TechPreviewNoUpgrade feature gates.
- Updates the Hypershift Operator to pipe through to the Control Plane Operator the feature set it is being run with, configuring the Control Plane Operator with the same feature set.
- Adds feature gates to the Control Plane Operator, a pre-requisite to adding the logic in https://github.com/openshift/hypershift/pull/5840

This change is necessary, because it allows shared features between Hypershift and OpenShift to be updated following the standard feature promotion cycle that exists for OpenShift. This change lays the foundation of this process, and there is likely to be iterative improvement in the approach in the future.

Fixes #

**Checklist**


control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -356,6 +355,19 @@ func (r *HostedControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+	if featuregates.Gate().Enabled(featuregates.Foo) {
+		// Foo is a gate that should be enabled in both TP and default featuresets
+		r.Log.Info("FEATURE_GATE_TEST -- FOO")
+	}
+
+	if featuregates.Gate().Enabled(featuregates.Bar) {
+		// Bar is a gate that should only be enables in TP featureset
+		r.Log.Info("FEATURE_GATE_TEST -- BAR")
+	}
+	// ----
```
everettraven, line 369: Leaving this in for now as an example. Will be removed prior to merging.

cmd/install/assets/hypershift_operator.go
```diff
@@ -460,6 +464,14 @@ func (o HyperShiftOperatorDeployment) Build() *appsv1.Deployment {
 		},
 	}
 
+	// Add the new HYPERSHIFT_FEATURESET env var if TPNU is set.
```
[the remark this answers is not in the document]
enxebre: Let's focus this PR on the impl details and avoid any consumer impacting changes i.e. let's don't introduce/change install flags.

Seth will tell but I think he was expressing desire to not change an existing working plumbing implementation, which you are not doing here since the env variable solves a new need to propagate this semantic down to the CPO.

hypershift-operator/featuregate/feature.go
```diff
@@ -19,16 +23,43 @@ const (
...
+
+func FeatureSet() configv1.FeatureSet {
+	return globalFeatureSet
 }
 
-var defaultHypershiftFeatureGates = map[featuregate.Feature]featuregate.FeatureSpec{
-	// Every feature should be initiated here:
-	AROHCPManagedIdentities: {Default: false, PreRelease: featuregate.Alpha},
-	OpenStack:               {Default: false, PreRelease: featuregate.Alpha},
+func ConfigureFeatureSet(featureSet string) {
```
enxebre, line 56: can we please document this exported func?

control-plane-operator/featuregates/featuregates.go
```diff
@@ -0,0 +1,47 @@
...
+	// Default to configuring the Default featureset
+	ConfigureFeatureSet(string(configv1.Default))
+}
+
+var globalGate featuregate.FeatureGate
+
+func Gate() featuregate.FeatureGate {
+	return globalGate
+}
+
```
enxebre, line 38: let's document this func/vars

hypershift-operator/featuregate/feature.go
```diff
@@ -19,16 +23,43 @@ const (
 	OpenStack featuregate.Feature = "OpenStack"
 )
 
+var (
+	allFeatures = featuregates.NewFeatureSetAwareFeatures()
+
+	aroHCPManagedIdentitiesFeature = featuregates.NewFeature(AROHCPManagedIdentities, featuregates.WithEnableForFeatureSets(configv1.TechPreviewNoUpgrade))
```
enxebre, line 29: fwiw just a note orthogonal to the PR. While the plumbing make sense, in practice we don't even need these two in particular atm, aroHCPManagedIdentities doesn't exist anymore and openStack is basically gated implicitily by its existence in the API.

control-plane-operator/featuregates/featuregates.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+// Define new featuregates here
+const (
+	Foo featuregate.Feature = "Foo"
+	Bar featuregate.Feature = "Bar"
+)
+
+var (
+	allFeatures = featuregates.NewFeatureSetAwareFeatures()
```
enxebre, line 18: I'm fine keeping this source for set<-> features separate from the HO as you proposed so feel free to ignore my comment on the sync discussion

hypershift-operator/featuregate/feature.go
everettraven: I leave these in for now since this is pretty much just a conversion of the existing feature gates to the new format. Maybe we can create a separate ticket or something to track the need to remove stale feature gates?

enxebre: yes, let's leave it aside from that PR. I was just trying to make visible all the context since we are here

cmd/install/assets/hypershift_operator.go
```diff
@@ -404,17 +401,6 @@ type HyperShiftOperatorDeployment struct {
...
-// String returns a string containing all enabled feature gates, formatted as "key1=value1,key2=value2,...".
-func featureGateString() string {
-	featureGates := make([]string, 0)
-	for feature := range featuregate.MutableGates.GetAll() {
-		featureGates = append(featureGates, fmt.Sprintf("%s=true", feature))
-	}
-
-	sort.Strings(featureGates)
-	return strings.Join(featureGates, ",")
-}
```
everettraven, line 416: Removed because there are no longer any references to this function with this change.

pkg/featuregates/featuregates.go
```diff
@@ -0,0 +1,114 @@
+package featuregates
```
enxebre, line 1: Not a blocker, but the generic stuff in this repo which crosses components usually goes within support/. We can follow up for that if we want to.

control-plane-operator/main.go
```diff
@@ -266,12 +264,16 @@ func NewStartCommand() *cobra.Command {
 			"SOCKS5_PROXY_IMAGE":                   &socks5ProxyImage,
 			"AVAILABILITY_PROBER_IMAGE":            &availabilityProberImage,
 			"TOKEN_MINTER_IMAGE":                   &tokenMinterImage,
+			"HYPERSHIFT_FEATURESET":                &featureSet,
```
enxebre, line 267: given the number of consumptions for HYPERSHIFT_FEATURESET, we might want to put that into an exported var within support/

enxebre:

PR #5995, opened 2025-04-07
OCPBUGS-51800: Fix golang crypto dependency go.mod replacement

celebdor:

Fix replace statement introduced in #5902  so that golang crypto finally uses the openshift fixed fork that does not require golang 1.23.

Fixes #OCPBUGS-51800

**Checklist**


celebdor:
/override "Red Hat Konflux / hypershift-release-mce-25-on-pull-request"

PR #6015, opened 2025-04-09
OCPBUGS-54763: openstack: set port security only if explicitly specified

EmilienM:

Avoids setting DisablePortSecurity by default to prevent failures when networks
are project-owned and the default Neutron policy won't allow explicit port security
options.


stephenfin:
/lgtm

Using the default makes sense here.

EmilienM:
/retest
konflux not happy

PR #6035, opened 2025-04-14
NO-JIRA: Bump o/api version to 4.19 tag

enxebre:
Without this bump the ocp GatewayAPI tests fail since the feature is not enable in default featureset and so the ingress operator doesn't install the CRDs

Run
go get github.com/openshift/api@release-4.19
go mod tidy
go mod vendor
make update

Fixes #

**Checklist**


enxebre:

PR #6039, opened 2025-04-15
CNTRLPLANE-709: Add ability to limit platform specific resources managed by HostedCluster

bryan-cox:
This PR allows the HostedCluster to only watch platform specific resources based on what platforms installed in the HO. If the HO does not limit the platform CRDs installed, the HostedCluster watches all platform CRDs as it did before this PR.

[CNTRLPLANE-709](https://issues.redhat.com/browse/CNTRLPLANE-709)

**Checklist**

### DIFF
--- a/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
+++ b/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
@@ -4260,13 +4260,15 @@ func (r *HostedClusterReconciler) delete(ctx context.Context, hc *hyperv1.Hosted
 		return false, err
 	}
 
-	exists, err := deleteAWSEndpointServices(ctx, r.Client, hc, controlPlaneNamespace)
-	if err != nil {
-		return false, err
-	}
-	if exists {
-		log.Info("Waiting for awsendpointservice deletion", "controlPlaneNamespace", controlPlaneNamespace)
-		return false, nil
+	if hc.Spec.Platform.Type == hyperv1.AWSPlatform {
+		exists, err := deleteAWSEndpointServices(ctx, r.Client, hc, controlPlaneNamespace)
+		if err != nil {
+			return false, err
+		}
+		if exists {
+			log.Info("Waiting for awsendpointservice deletion", "controlPlaneNamespace", controlPlaneNamespace)
+			return false, nil
+		}
 	}
 
 	if r.ManagementClusterCapabilities.Has(capabilities.CapabilityRoute) {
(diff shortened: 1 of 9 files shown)


hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
bryan-cox, line 1: This is failing at the moment and I don't quite understand why - 
```
the set of resources that are being created differs from the one that is being watched:   []string{
          	... // 9 identical elements
          	"*v1.Route",
          	"*v1.Secret",
        - 	"*v1.SecretProviderClass",
          	"*v1.Service",
          	"*v1.ServiceAccount",
          	"*v1alpha1.KubevirtCluster",
          	"*v1beta1.AgentCluster",
          	"*v1beta1.AzureCluster",
        - 	"*v1beta1.AzureClusterIdentity",
          	"*v1beta1.Cluster",
          	"*v1beta1.HostedControlPlane",
        + 	"*v1beta1.OpenStackCluster",
          	"*v1beta2.AWSCluster",
          	"*v1beta2.IBMVPCCluster",
          }
```

bryan-cox: I don't understand how `client.createdTypes.List()` is being populated.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -158,6 +160,16 @@ const (
 // NoopReconcile is just a default mutation function that does nothing.
 var NoopReconcile controllerutil.MutateFn = func() error { return nil }
 
+var capiRelatedManagedResourcesToWatch = []client.Object{
```
enxebre, line 163: AWSEndpointService is not a capi resource and there might be others. Let's rather use platform as the driver, e.g. platformSpecificResources

support/util/resources.go
```diff
@@ -0,0 +1,63 @@
...
+)
+
+func GetHostedClusterManagedResources(platformsInstalled string) []client.Object {
+	var managedResources []client.Object
+
+	platformsInstalledList := strings.Split(platformsInstalled, ",")
+	for _, platform := range platformsInstalledList {
+		platform = strings.ToLower(strings.TrimSpace(platform))
+		switch platform {
+		case "aws":
```
enxebre, line 25: can we somehow use the PlatformType's for this and any reference to platform segregation?

support/util/resources.go
```diff
@@ -0,0 +1,106 @@
...
+
+func GetHostedClusterManagedResources(platformsInstalled string) []client.Object {
+	var managedResources []client.Object
+
+	platformsInstalledList := strings.Split(platformsInstalled, ",")
+	for _, platform := range platformsInstalledList {
+		platform = strings.ToLower(strings.TrimSpace(platform))
+		switch {
+		case strings.EqualFold(platform, string(hyperv1.AWSPlatform)):
+			managedResources = append(managedResources, BaseResources...)
```
csrwng, line 64: Why not append BaseResources once before the platform-specific ones?

bryan-cox: The None platform doesn't have that resource and including it by default was causing the unit test to fail.

PR #6041, opened 2025-04-16
Testing techpreview please ignore

jkyros:
Trying to rule out whether the karpenter/deps bump broke the e2e-aws-techpreview or not. It looks like no, I expect it to fail the same way here as it does there, but I want to make sure the deps didn't weirdly break it. 

Fixes #

**Checklist**


jkyros:
Looks like it's not me :smile: :frowning:

PR #6043, opened 2025-04-16
NO-JIRA: version check csi-snapshot-webhook for EnsureSATokenNotMountedUnlessNecessary

sjenning:
https://github.com/openshift/hypershift/pull/6004 removed `csi-snapshot-webhook` from the excluded prefixes, however, the webhook is only removed back to 4.18.  4.16 and 4.17 still have it and it still needs to be excluded.

This is causing payload test failures.


bryan-cox (approved):
/lgtm

sjenning:
/override ci/prow/e2e-aws

PR #6048, opened 2025-04-17
CNTRLPLANE-111: Support KMSv2 encryption encryption for ARO HCP, with MIv3

bryan-cox:
This PR:
- updates Azure KMS implementation by updating the encryption configuration to only support v2 parameters
- have Azure KMS use it's own Azure authentication configuration; previously, it was reusing CP/CCM configuration
- reorders some of the Azure how-to pages in our docs
- adds a how-to section on setting up and verifying etcd encryption

Fixes [CNTRLPLANE-111](https://issues.redhat.com//browse/CNTRLPLANE-111)

**Notes to Reviewers**:
- The first commit contains the actual code changes. The other commits are docs related.
- CacheSize: ptr.To\[int32\]\(100\), was removed because it is not valid for v2

**Checklist**

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
@@ -3261,27 +3261,11 @@ func (r *HostedControlPlaneReconciler) reconcileKubeAPIServer(ctx context.Contex
 			if hcp.Spec.SecretEncryption.KMS == nil {
 				return fmt.Errorf("kms metadata not specified")
 			}
-			if _, err := createOrUpdate(ctx, r, encryptionConfigFile, func() error {
-				return kas.ReconcileKMSEncryptionConfig(encryptionConfigFile, p.OwnerRef, hcp.Spec.SecretEncryption.KMS)
-			}); err != nil {
-				return fmt.Errorf("failed to reconcile kms encryption config secret: %w", err)
-			}
 
 			if _, err := createOrUpdate(ctx, r, encryptionConfigFile, func() error {
 				return kas.ReconcileKMSEncryptionConfig(encryptionConfigFile, p.OwnerRef, hcp.Spec.SecretEncryption.KMS)
 			}); err != nil {
 				return fmt.Errorf("failed to reconcile kms encryption config secret: %w", err)
-			}
-
-			if hyperazureutil.IsAroHCP() {
-				// Reconcile the SecretProviderClass
-				kmsSecretProviderClass := manifests.ManagedAzureSecretProviderClass(config.ManagedAzureKMSSecretProviderClassName, hcp.Namespace)
-				if _, err := createOrUpdate(ctx, r, kmsSecretProviderClass, func() error {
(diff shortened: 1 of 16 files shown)


control-plane-operator/controllers/hostedcontrolplane/kas/kms/azure.go
```diff
@@ -79,7 +78,6 @@ func (p *azureKMSProvider) GenerateKMSEncryptionConfig(apiVersion string) (*v1.E
 			Name:       fmt.Sprintf("%s-%s", azureProviderConfigNamePrefix, activeKeyHash),
 			APIVersion: apiVersion,
 			Endpoint:   azureActiveKMSUnixSocket,
-			CacheSize:  ptr.To[int32](100),
```
bryan-cox, line 82: CacheSize: ptr.To[int32], was removed because it is not valid for v2

control-plane-operator/controllers/hostedcontrolplane/kas/params.go
```diff
@@ -289,12 +288,8 @@ func NewKubeAPIServerParams(ctx context.Context, hcp *hyperv1.HostedControlPlane
 		},
 	}
 
-	switch hcp.Spec.Platform.Type {
-	case hyperv1.AWSPlatform:
+	if hcp.Spec.Platform.Type == hyperv1.AWSPlatform {
 		params.CloudProvider = aws.Provider
-	case hyperv1.AzurePlatform:
-		params.CloudProvider = azure.Provider
```
bryan-cox, line 296: This doesn't need set for Azure - https://cloud-provider-azure.sigs.k8s.io/install/azure-ccm/#kube-apiserver

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -3178,13 +3173,12 @@ func (r *HostedControlPlaneReconciler) reconcileKubeAPIServer(ctx context.Contex
 			}
 
 			if hyperazureutil.IsAroHCP() {
-				// Reconcile the SecretProviderClass
-				kmsSecretProviderClass := manifests.ManagedAzureSecretProviderClass(config.ManagedAzureKMSSecretProviderClassName, hcp.Namespace)
-				if _, err := createOrUpdate(ctx, r, kmsSecretProviderClass, func() error {
-					secretproviderclass.ReconcileManagedAzureSecretProviderClass(kmsSecretProviderClass, hcp, hcp.Spec.SecretEncryption.KMS.Azure.KMS, true)
-					return nil
+				// Reconcile the Azure KMS Config Secret
```
enxebre, line 3176: should this be part of hc/internal/platform ReconcileCredentials for consistency with aws? is the cloudNetworkConfigCreds there still needed?
alternative should we make this part of the kms.IKMSProvider interface?

bryan-cox: cloudNetworkConfigCreds is still needed for CCM/CP so no need to mess with it here.

PR #6049, opened 2025-04-17
OCPBUGS-54861: fix(hostedCluster): Skip ignition server setup if disabled

a-dsouza:
**What this PR does / why we need it**: missing pieces for ignition server disable feature

Fixes # https://issues.redhat.com/browse/OCPBUGS-54861

**Checklist**

### DIFF
--- a/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
+++ b/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
@@ -940,57 +940,59 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
 
 	// Set Ignition Server endpoint
 	{
-		serviceStrategy := servicePublishingStrategyByType(hcluster, hyperv1.Ignition)
-		if serviceStrategy == nil {
-			// We don't return the error here as reconciling won't solve the input problem.
-			// An update event will trigger reconciliation.
-			log.Error(fmt.Errorf("ignition server service strategy not specified"), "")
-			return ctrl.Result{}, nil
-		}
-		switch serviceStrategy.Type {
-		case hyperv1.Route:
-			if serviceStrategy.Route != nil && serviceStrategy.Route.Hostname != "" {
-				hcluster.Status.IgnitionEndpoint = serviceStrategy.Route.Hostname
-			} else {
-				ignitionServerRoute := ignitionserver.Route(controlPlaneNamespace.GetName())
-				if err := r.Client.Get(ctx, client.ObjectKeyFromObject(ignitionServerRoute), ignitionServerRoute); err != nil {
-					if !apierrors.IsNotFound(err) {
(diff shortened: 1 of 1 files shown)


hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -1010,47 +1012,49 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
...
+				Type:   string(hyperv1.IgnitionEndpointAvailable),
+				Status: metav1.ConditionUnknown,
+				Reason: hyperv1.StatusUnknownReason,
+			}
+			// Check to ensure the deployment exists and is available.
+			deployment := ignitionserver.Deployment(controlPlaneNamespace.Name)
+			if err := r.Client.Get(ctx, client.ObjectKeyFromObject(deployment), deployment); err != nil {
+				if apierrors.IsNotFound(err) {
+					newCondition = metav1.Condition{
+						Type:    string(hyperv1.IgnitionEndpointAvailable),
```
enxebre, line 1027: without the check this condition would legitimately signal this state here, are you intentionally wanting to avoid that?

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -940,57 +940,59 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
...
-						return ctrl.Result{}, fmt.Errorf("failed to get ignitionServerRoute: %w", err)
-					}
-				}
-				if ignitionServerRoute.Spec.Host != "" {
-					hcluster.Status.IgnitionEndpoint = ignitionServerRoute.Spec.Host
-				}
-			}
-		case hyperv1.NodePort:
-			if serviceStrategy.NodePort == nil {
+		if _, exists := hcp.Annotations[hyperv1.DisableIgnitionServerAnnotation]; !exists {
```
enxebre, line 943: not to block this particular PR on this, but we should looking into integrate all these annotations within the API.
Instead of wrapping the whole thing with an if, we could drop all the checks that return `ctrl.Result{}, nil in this block` (This is validated by CEL now), then make this a func SetIgnitionServerEndpoint that just return earlier DisableIgnitionServerAnnotation exists and keep running current logic otherwise.

I'm ok with the change as it's though, since the above should possibly be addressed as part of a targeted bigger effort to rework this controller as a whole.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
a-dsouza: Good point--if the state is purely informational, I think it's fine to avoid it being set, but now I'm thinking about it, I'm worried if that state is referenced elsewhere or if it's used to determine the state of the HostedCluster resource. I'm not certain of that yet--do you happen to know?

PR #6060, opened 2025-04-18
CNTRLPLANE-713: Use feature flags from release payload for control plane workloads

csrwng:
We currently render feature flags using cluster-config-api container inside the Kube APIServer and Ignition Server for rendering into the hosted cluster. However, for components that we run in the control plane that also use feature flags (Kube APIServer, Kube Controller Manager, Kube Scheduler, Cluster Policy Controller), we are simply generating feature flags from vendored code in the HyperShift repo. This may not reflect the reality of the particular release payload that we are running.

This commit adds a job to render feature gates into a configmap that can be used to configure the control plane workloads. For v2, it introduces a new type of component (Job).

Fixes #[CNTRLPLANE-713](https://issues.redhat.com/browse/CNTRLPLANE-713)

**Checklist**


control-plane-operator/controllers/hostedcontrolplane/fg/job.go
```diff
@@ -0,0 +1,104 @@
...
+func getPayloadVersionAndFeatureGateYAML(job *batchv1.Job) (string, string) {
+	if len(job.Spec.Template.Spec.InitContainers) == 0 {
+		return "", ""
+	}
+	var payloadVersion, fgYAML string
+	for _, env := range job.Spec.Template.Spec.InitContainers[0].Env {
+		if env.Name == "PAYLOAD_VERSION" {
+			payloadVersion = env.Value
+			continue
+		}
```
enxebre, line 97: fwiw it shouldn't hurt but just so we are aware this will trigger update if e.g. fg crd includes a new defaulted or not omitempty field

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -2769,6 +2791,62 @@ func (r *HostedControlPlaneReconciler) reconcileCSIDriver(ctx context.Context, h
...
+					return false, nil
+				}
+				// If job has completed, ensure that the resulting configmap exists and has valid data.
+				// If not, delete the job so it can be recreated
+				if cm, err := config.FeatureGateConfigMap(ctx, r.Client, hcp.Namespace); err != nil {
+					if !apierrors.IsNotFound(err) {
+						return false, err
+					}
+				} else {
+					if _, err := config.ParseFeatureGates(cm); err == nil {
```
enxebre, line 2828: why are we not surfacing this err?

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -2769,6 +2791,62 @@ func (r *HostedControlPlaneReconciler) reconcileCSIDriver(ctx context.Context, h
 	return nil
 }
 
+func (r *HostedControlPlaneReconciler) reconcileFeatureGateGenerationJob(ctx context.Context, hcp *hyperv1.HostedControlPlane, createOrUpdate upsert.CreateOrUpdateFN, releaseImageProvider, userReleaseImageProvider *imageprovider.SimpleReleaseImageProvider) (bool, error) {
```
enxebre, line 2794: can we have a unit with cases for the bool?

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
enxebre: fwiw I find it easier to read by avoiding nested ifs an failing earlier explicitily, e.g.
```
		// There is an existing job, check its completion status and whether it matches the spec we want
		failedCond := util.FindJobCondition(existing, batchv1.JobFailed)
		if failedCond != nil && failedCond.Status == corev1.ConditionTrue {
			// If the job has failed, delete it
			_, err := util.DeleteIfNeededWithOptions(ctx, r.Client, existing, client.PropagationPolicy(metav1.DeletePropagationForeground))
			return false, err
		}
		if needsUpdate := fg.NeedsUpdate(existing, updated); needsUpdate {
			// If the job needs to be updated, delete it
			_, err := util.DeleteIfNeededWithOptions(ctx, r.Client, existing, client.PropagationPolicy(metav1.DeletePropagationForeground))
			return false, err
		}

		completedCond := util.FindJobCondition(existing, batchv1.JobComplete)
		if completedCond == nil || completedCond.Status != corev1.ConditionTrue {
			// Not done yet
			return false, nil
		}
		
		// If the job has completed, ensure that the resulting configmap exists and has valid data.
		cm, err := config.FeatureGateConfigMap(ctx, r.Client, hcp.Namespace);
		if err != nil {
				return false, err
			}
		}

		if _, err := config.ParseFeatureGates(cm); err != nil {
			return false, err
		}
		// Only if feature gates exist and are valid, then we are ready
		return true, nil
```

feel free to adopt/ignore

control-plane-operator/controllers/hostedcontrolplane/fg/assets/job.yaml
```diff
@@ -0,0 +1,71 @@
...
+            valueFrom:
+              fieldRef:
+                fieldPath: metadata.namespace
+          command:
+          - /bin/bash
+          args:
+          - -c
+          - |-
+            #!/bin/bash
+            if oc get cm feature-gate; then
```
enxebre, line 59: having the CM annotated with the version applying it might help in future debugging

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
csrwng: I updated the logic, thx!

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
csrwng: done

csrwng:
@enxebre I addressed your comments, thanks for the review

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller_test.go
```diff
@@ -2057,3 +2077,202 @@ func componentsFakeDependencies(componentName string, namespace string) []client
...
+				ReleaseProvider:     &fakereleaseprovider.FakeReleaseProvider{},
+				UserReleaseProvider: &fakereleaseprovider.FakeReleaseProvider{},
+			}
+			proceed, err := r.reconcileFeatureGateGenerationJob(context.Background(), hcp, controllerutil.CreateOrUpdate, &fakeReleaseProvider{}, &fakeReleaseProvider{})
+			if tc.expectedError {
+				g.Expect(err).To(HaveOccurred())
+			} else {
+				g.Expect(err).NotTo(HaveOccurred())
+				if tc.expectedProceed {
+					g.Expect(proceed).To(BeTrue())
```
enxebre, line 2261: shouldn't this just be 
g.Expect(proceed).To(tc.expectedProceed)

?

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller_test.go
```diff
@@ -2057,3 +2077,202 @@ func componentsFakeDependencies(componentName string, namespace string) []client
...
+		{
+			name:            "completed job, invalid configmap",
+			job:             completedJob(expectedPayloadVersion, expectedFGYaml),
+			cm:              invalidConfigMap(),
+			expectJobExists: false,
+			expectedProceed: false,
+		},
+	}
+
+	for _, tc := range tests[3:] {
```
enxebre, line 2232: why does this loop start on 3?

PR #6062, opened 2025-04-19
AUTOSCALE-127: fix e2e autonode drift test

maxcao13:
Fixes the test by grabbing the correct part of the string from node.Status.NodeInfo.OSImage that now points to the rhcos version. See https://github.com/openshift/release/pull/63681#issuecomment-2816838538 as to why we need this.

Fixes #

**Checklist**


jkyros:
Techpreview passed, the parsing is still kinda brittle and may break if the string format changes again, but this makes it work again for now and we can think on how to fix that more elegantly for later -- hopefully this is okay  :) 
/lgtm

PR #6063, opened 2025-04-21
OCPBUGS-55226: switch termination message policy to TerminationMessageFallbackToLogs…

deads2k:
…OnError for debugability

We do this for standalone and it makes a huge difference.  I think I found the specifically set values and the default setting for mutation.

I'd also like to write a test.  Where can I add a test that watches all pods in guest cluster namespaces and openshift-* namespaces in the management cluster for this setting?


enxebre:
thanks!
>I'd also like to write a test. Where can I add a test that watches all pods in guest cluster namespaces and openshift-* namespaces in the management cluster for this setting?

This change will mostly impact those deployments running pods within the control plane namespace. For a similar validation example see e.g. EnsureAllContainersHavePullPolicyIfNotPresent

sjenning:
superseded by https://github.com/openshift/hypershift/pull/6122

PR #6074, opened 2025-04-24
OCPBUGS-55504: Add postStart hook to kube-apiserver container

muraee:
- This blocks the creation of other containers specifically the bootstrap container until kas is ready.

Fixes #

**Checklist**


bryan-cox:
This doesn't seem like something that should be NO-JIRA. We should have a Jira ticket for this imo.

bryan-cox:
/hold cancel

#5792 has merged

PR #6094, opened 2025-04-29
CNTRLPLANE-645: Rename NodeCount flag to Replicas

georgelipceanu:
This PR changes the `--node-count` flag to be renamed to `--replicas` so that it is consistent with what the term is called in the API.

Fixes #[CNTRLPLANE-645](https://issues.redhat.com/browse/CNTRLPLANE-645)

**Checklist**

### DIFF
--- a/cmd/nodepool/create.go
+++ b/cmd/nodepool/create.go
@@ -30,7 +30,7 @@ func NewCreateCommand() *cobra.Command {
 		Name:            "example",
 		Namespace:       "clusters",
 		ClusterName:     "example",
-		NodeCount:       2,
+		Replicas:        2,
 		ReleaseImage:    "",
 		NodeUpgradeType: hyperv1.UpgradeTypeReplace,
 		Arch:            "amd64",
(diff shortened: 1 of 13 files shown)


product-cli/cmd/nodepool/create.go
bryan-cox, line 1: I don't think we can just rename it in the `hcp cli`. We probably should follow the k8s deprecation policy - https://kubernetes.io/docs/reference/using-api/deprecation-policy/#deprecating-a-flag-or-cli.

georgelipceanu: The flag could be kept as `--node-count` in the HCP CLI, I just changed it since the Jira ticket asks for this to be changed in HCP also.
Reading the policy provided, would we keep both `--node-count` (with a warning of deprecation) and `--replicas` as flags, then deprecate after 12 months or 2 releases?

bryan-cox: @enxebre what would you recommend as the deprecation path here? Time based or release based?

georgelipceanu: I have `--node-count` marked as deprecated either way now 👍

devguyio: I agree that we should deprecate the `node-count` flag, that way we wouldn't break any existing tooling or CI using the current flag and would do a soft transitioning.

devguyio (approved):
Nice work. 
/lgtm

cmd/nodepool/create.go
```diff
@@ -30,15 +30,17 @@ func NewCreateCommand() *cobra.Command {
...
 		NodeUpgradeType: hyperv1.UpgradeTypeReplace,
 		Arch:            "amd64",
 	}
 
 	cmd.PersistentFlags().StringVar(&opts.Name, "name", opts.Name, "The name of the NodePool")
 	cmd.PersistentFlags().StringVar(&opts.Namespace, "namespace", opts.Namespace, "The namespace in which to create the NodePool")
-	cmd.PersistentFlags().Int32Var(&opts.NodeCount, "node-count", opts.NodeCount, "The number of nodes to create in the NodePool")
+	cmd.PersistentFlags().Int32Var(&opts.Replicas, "replicas", opts.Replicas, "The number of nodes to create in the NodePool")
+	cmd.PersistentFlags().Int32Var(&opts.Replicas, "node-count", opts.Replicas, "The number of nodes to create in the NodePool (deprecated, use '--replicas' instead)")
+	_ = cmd.PersistentFlags().MarkDeprecated("node-count", "please use '--replicas' instead")
```
bryan-cox, line 43: nit: could this be moved outside of all the flags set to persistent

bryan-cox: Yeah

PR #6100, opened 2025-04-30
OCPBUGS-54483: Add support for registry root entry only in the IDMS/ICSP

jparrill:
The reference.Parse function tries to retrieve the whole entry string, but if the entry reference does not contain Namespace and Name but registry, it will return the Registry field as the Name which is not a good behavior. We added the proper changes to have this in consideration and return the right override. Also we covered it with the unit test to avoid regressions.

- Fixes [OCPBUGS-54483](https://issues.redhat.com/browse/OCPBUGS-54483)

### DIFF
--- a/support/util/imagemetadata.go
+++ b/support/util/imagemetadata.go
@@ -374,7 +374,52 @@ func ImageLabels(metadata *dockerv1client.DockerImageConfig) map[string]string {
 	}
 }
 
-func GetRegistryOverrides(ctx context.Context, ref reference.DockerImageReference, source string, mirror string) (*reference.DockerImageReference, bool, error) {
+// Attempts only a root registry override match.
+func tryOnlyRootRegistryOverride(ref, sourceRef, mirrorRef reference.DockerImageReference) (*reference.DockerImageReference, bool, error) {
+	if sourceRef.Namespace == "" && sourceRef.Registry == "" && sourceRef.Name != "" {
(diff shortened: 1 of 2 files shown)


support/util/imagemetadata.go
```diff
@@ -387,6 +387,19 @@ func GetRegistryOverrides(ctx context.Context, ref reference.DockerImageReferenc
 		return nil, false, fmt.Errorf("failed to parse mirror image reference %q: %w", source, err)
 	}
 
+	// If the root registry is provided, once the image is parsed, it returns the registry in the Name field
```
bryan-cox, line 390: Maybe this section should be its own function to simplify `GetRegistryOverrides` further. IMO, we should refactor `GetRegistryOverrides` because of how complex it has gotten.

bryan-cox: It seems like we should at least make this section of new code less complex given it has compounded and nested if statements.

jparrill: Refactored, let me know if you have any concern

sdminonne:
I'm OK with this. @bryan-cox already had a look and I let him put the final stamp. Thanks!

jparrill:
> @jparrill - Target Version and Target Backport Versions need updated on the OCPBUGS

I've set the target version to 4.20.0 and backport to 4.17.z, 4.18.z and 4.19.0. I hope this PR got merged before the code freeze 😄

PR #6103, opened 2025-05-01
WIP: ho: changes to ref'ed resources triggers HC reconciliation

sjenning:


hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -154,6 +154,7 @@ const (
 
 	previouslySyncedRestartDateAnnotation = "hypershift.openshift.io/previous-restart-date"
 	kasServingCertHashAnnotation          = "hypershift.openshift.io/kas-serving-cert-hash"
+	referencedResourceAnnotationPrefix    = "hypershift.openshift.io/referenced-resource-"
```
csrwng, line 157: My pref would be something like:
`referenced-resource.hypershift.openshift.io/[name]`
A little less confusing when understanding what part is prefix and what part is name.

csrwng (commented):
lgtm, just one nit

sjenning:
HO only, no payload impact

PR #6122, opened 2025-05-06
OCPBUGS-55226: switch termination message policy to TerminationMessageFallbackToLogsOnError for debugability

sjenning:
supersedes #6063 

We do this for standalone and it makes a huge difference when full pod logs are not captured/available for any reason.


muraee:
This is pretty straight forward to do in CPOv2 similar to image pull policy enforcement here: https://github.com/openshift/hypershift/blob/main/support/controlplane-component/defaults.go#L224

muraee:
test fixtures need update for unit to pass
```
UPDATE=true go test ./...
```

support/controlplane-component/defaults.go
```diff
@@ -86,6 +86,8 @@ func (c *controlPlaneWorkload[T]) setDefaultOptions(cpContext ControlPlaneContex
 	if err != nil {
 		return err
 	}
+	enforceTerminationMessagePolicy(podTemplateSpec.Spec.InitContainers)
+	enforceTerminationMessagePolicy(podTemplateSpec.Spec.Containers)
```
muraee, line 90: To fix the e2e, we would need to move this call after we inject konnectivity/tokenMinter/avaialbiltiyProber containers so that those containers also get the new TerminationMessagePolicy. i.e. after 
https://github.com/openshift/hypershift/blob/e74a1bb4154e7b4e04e73d0e657711d25e1e4870/support/controlplane-component/defaults.go#L113-L119

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -1713,6 +1713,12 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
 	}
 	if capiProviderDeploymentSpec != nil {
 		proxy.SetEnvVars(&capiProviderDeploymentSpec.Template.Spec.Containers[0].Env)
+		for i := range capiProviderDeploymentSpec.Template.Spec.Containers {
```
enxebre, line 1716: all the changes in this file will trigger a rollout for those components across the fleet which might result in a resource consumption spike. We probably want to call this out for managed services consumers

enxebre:
/label tide/merge-method-squash
/lgtm

However e2e-aws-upgrade-hypershift-operator is failing because that uses a previews version of the HO as a first step.

sjenning:

sjenning:

sjenning:

PR #6127, opened 2025-05-07
WIP OCPBUGS-55892: make image registry managed identity optional

flavianmissi:

Fixes #

**Checklist**


cmd/infra/azure/create.go
```diff
@@ -246,6 +247,10 @@ func (o *CreateInfraOptions) Run(ctx context.Context, l logr.Logger) (*CreateInf
 			config.CNCC:          result.ControlPlaneMIs.ControlPlane.Network.ClientID,
 		}
 
+		if !slices.Contains(o.DisableClusterCapabilities, string(hyperv1.ImageRegistryCapability)) {
+			components[config.CIRO] = result.ControlPlaneMIs.ControlPlane.ImageRegistry.ClientID
```
flavianmissi, line 251: Is `result.ControlPlaneMIs.ControlPlane.ImageRegistry.ClientID` provided by the user running this command? If so, is it worth printing some some of message here in case they have both provided an image registry managed identity _and_ disabled the image registry capability? Am I overthinking this? 🤔

flavianmissi:
/retitle OCPBUGS-55892: make image registry managed identity optional

cmd/infra/azure/create.go
bryan-cox: Really client ID is not used in ARO HCP production. It's only here right now because we need to keep it around for CI purposes in the CLI.

bryan-cox:
@flavianmissi - The release note type and text on the bug ticket still needs filled out.

flavianmissi:
Now we're just missing a lgtm @bryan-cox, should I ping someone else for it?

PR #6134, opened 2025-05-08
OCPBUGS-55913: Fix Job status returning empty reason and message

muraee:
This fixes the issue of the controller failing to update ControlPlaneComponent status condition on older versions of openshift/k8s management clusters where the job controller doesn't set the reason and message fields in the Job conditions.

Fixes #

**Checklist**

### DIFF
--- a/support/controlplane-component/job.go
+++ b/support/controlplane-component/job.go
@@ -48,32 +48,42 @@ func (c *jobProvider) Replicas(object *batchv1.Job) *int32 {
 
 // IsAvailable implements WorkloadProvider.
 func (c *jobProvider) IsAvailable(job *batchv1.Job) (status metav1.ConditionStatus, reason string, message string) {
-	complete := util.FindJobCondition(job, batchv1.JobComplete)
-	if complete != nil {
-		return metav1.ConditionStatus(complete.Status), complete.Reason, complete.Message
-	}
-	failed := util.FindJobCondition(job, batchv1.JobFailed)
-	if failed != nil {
-		return metav1.ConditionFalse, failed.Reason, failed.Message
-	}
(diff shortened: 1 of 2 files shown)


support/controlplane-component/job.go
```diff
@@ -48,32 +48,42 @@ func (c *jobProvider) Replicas(object *batchv1.Job) *int32 {
...
+
+// JobCompletionStatus checks the status of a job and returns the appropriate condition status, reason, and message.
+// It checks if the job is complete or failed and returns the corresponding status.
+// If the job is neither complete nor failed, it returns unknown status.
+func JobCompletionStatus(job *batchv1.Job) (status metav1.ConditionStatus, reason string, message string) {
 	complete := util.FindJobCondition(job, batchv1.JobComplete)
-	if complete != nil {
-		return metav1.ConditionStatus(complete.Status), complete.Reason, complete.Message
+	if complete != nil && complete.Status == corev1.ConditionTrue {
+		return metav1.ConditionTrue, "JobComplete", "Job completed successfully"
```
sjenning, line 71: why not just do the same logic as the `failed` case i.e. use the Reason/Message if set, else default to hardcoded reason/message?

bryan-cox:
The commit message description has a grammatical typo in it.

The commit subject should also use the imperative.

support/controlplane-component/job.go
muraee: There wouldn't be anything interesting in the message if job completed successfully, figured we could save a few lines of code XD. 
can change it if you think that's better.

sjenning:
confirmed the fix on 4.17 mgmt cluster

/lgtm

PR #6137, opened 2025-05-09
CNTRLPLANE-942: update KAS bootstrapping to get RBR from CAO

everettraven:
- Updates the Kubernetes API server bootstrapping logic to get the `RoleBindingRestriction` CRD from the cluster-authentication-operator's manifest renderer if OAuth is enabled. This is needed so that the Kubernetes API server can start up successfully when the `RoleBindingRestriction` CRD is no longer included in the openshift/api generated payload manifests. See https://github.com/openshift/api/pull/2138
- Disables the `authorization.openshift.io/RestrictSubjectBindings` and `authorization.openshift.io/ValidateRoleBindingRestrictions` admission plugins on the Kubernetes API server if the OAuth is not enabled. This was proposed in https://github.com/openshift/enhancements/pull/1726

Fixes #

**Checklist**


everettraven:
Validated in https://prow.ci.openshift.org/view/gs/test-platform-results/logs/multi-pr-openshift-api-2138-openshift-hypershift-6137-openshift-cluster-authentication-operator-770-e2e-aws-ovn-hypershift/1930689956246720512 that the current state does indeed result in KAS bootstrapping successfully using the new cluster-authentication-operator manifest renderer.

More specifically, it can be seen in the bootstrap logs here: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/logs/multi-pr-openshift-api-2138-openshift-hypershift-6137-openshift-cluster-authentication-operator-770-e2e-aws-ovn-hypershift/1930689956246720512/artifacts/e2e-aws-ovn-hypershift/hypershift-aws-run-e2e-external/artifacts/TestNodePool_HostedCluster0/namespaces/e2e-clusters-ssvrc-node-pool-mvwgw/core/pods/logs/kube-apiserver-564cd7f94b-gsgq7-bootstrap.log

ignition-server/controllers/local_ignitionprovider.go
```diff
@@ -403,6 +403,63 @@ func (p *LocalIgnitionProvider) GetPayload(ctx context.Context, releaseImage, cu
...
+			if caoComposedImage != clusterAuthenticationOperatorImage {
+				clusterAuthenticationOperatorImage = caoComposedImage
+				log.Info(fmt.Sprintf("using mirrored %s image %v", clusterAuthenticationOperatorComponent, caoComposedImage))
+			}
+
+			file, err := os.Create(filepath.Join(binDir, clusterAuthenticationOperatorComponent))
+			if err != nil {
+				return fmt.Errorf("failed to create file: %w", err)
+			}
+			if err := file.Chmod(0777); err != nil {
```
bryan-cox: security nit: should we really use 777 here?

everettraven: This logic was mostly copied from directly above for the standard o/api manifest renderer which is `0777` as well.

The only tweaks I had made was image and binary location and variable names (i didn't want to override something in case it was used later).

I contemplated making this a common function to use as well. If this is something folks think would be useful I'm happy to do this.

I'm open to making changes here, and if we do I'll make it across the board.

I'll leave it up to those more familiar with the historical context of why `0777` was originally chosen for the existing logic to decide if this needs to change.

control-plane-operator/controllers/hostedcontrolplane/v2/assets/kube-apiserver/deployment.yaml
```diff
@@ -276,6 +276,39 @@ spec:
...
+          EOF
+          fi
+
+          /usr/bin/authentication-operator render \
+            --asset-output-dir /tmp/output \
+            --rendered-manifest-dir=/tmp/manifests \
+            --cluster-profile=ibm-cloud-managed \
+            --payload-version=$(PAYLOAD_VERSION)
+
+          cp /tmp/output/* /work
```
enxebre, line 298: why don't we need to copy /tmp/manifests

ignition-server/controllers/local_ignitionprovider.go
```diff
@@ -403,6 +403,63 @@ func (p *LocalIgnitionProvider) GetPayload(ctx context.Context, releaseImage, cu
 		return nil, fmt.Errorf("failed to download binaries: %w", err)
 	}
 
+	// As of OpenShift 4.20, the cluster-authentication-operator is responsible for managing
+	// the RoleBindingRestriction CRD and has been removed from the cluster-config-api
+	// payload image.
+	// Because this CRD is required for bootstrapping, ensure we use the cluster-authentication-operator
+	// payload image to get the RoleBindingRestriction CRD in the set of bootstrap manifests.
+	if payloadVersion.Major == 4 && payloadVersion.Minor >= 20 {
+		// set the component to the correct binary name and file path based on the payload version
```
enxebre: All this block seems analogous to what we do for clusterConfigComponent. But that binary is then executed via invokeFeatureGateRenderScript to render the manifests that are then consumed by the MCO binaries to generate an ign payload for nodes to join.
I don't think we need anything of that for the authenticator operator and I think the logic added for the kas config is all we need. Agree?

control-plane-operator/controllers/hostedcontrolplane/v2/assets/kube-apiserver/deployment.yaml
everettraven: I didn't have it copy `/tmp/manifests` because, for now, the cluster-authentication-operator manifest renderer does not actually read in the feature gates file or write to it. There is no guarantee this doesn't change in the future, but I suspect it is not very likely to change.

Because of this, the feature gates file that the original manifest rendering step would produce would sometimes get overwritten, leading to some failures to create a HostedCluster.

Instead of continuing to copy the manifests over and trying to handle this edge case now, it seemed reasonable to defer that handling until the cluster-authentication-operator manifest renderer actually needs to read and write feature gates to a file.

ignition-server/controllers/local_ignitionprovider.go
everettraven: I'm not sure I'm following.

Why would we need this for the `cluster-config-api` manifest renderer but not for the `cluster-authentication-operator`?

I'm not all that familiar with how all this code works together to create the final output and was piecing things together as I went the best I could. Anywhere we are pulling in the manifest renderer for the `cluster-config-api` seemed reasonable to also include the manifest renderer for the `cluster-authentication-operator`. 

With that in mind, I'll defer to you on whether or not it actually makes sense to add this logic - I'm happy to remove it if we don't need it.

control-plane-operator/controllers/hostedcontrolplane/v2/assets/kube-apiserver/deployment.yaml
enxebre: Let's please add a code comment articulating this.

enxebre: Also if that's the case, why handcrafting the featuregate yaml in the script at all?

ignition-server/controllers/local_ignitionprovider.go
enxebre: cluster-config-api binary is invoked via the invokeFeatureGateRenderScript func call. That results in some rendered manifests. That output is then consumed by the MCO binaries render commands. Ultimately the output of those commands is consumed by an ephemeral MCD to serve the ign payload.
I don't think anything of this applies to cao afaict. The code as written now in the PR is actually transparent since the cao binary is never executed. Besides the payload served to nodes won't have any impact on RoleBindingRestriction creation. Let's remove it.

control-plane-operator/controllers/hostedcontrolplane/v2/assets/kube-apiserver/deployment.yaml
everettraven: >Also if that's the case, why handcrafting the featuregate yaml in the script at all?

Fair point - I forgot to remove it :)

ignition-server/controllers/local_ignitionprovider.go
everettraven: Thanks for elaborating - that cleared up my confusion. Will remove it.

sjenning:
`verify` failure is an OOM flake btw

ignition-server/controllers/local_ignitionprovider.go
```diff
@@ -759,7 +759,7 @@ func copyFile(src, dst string) error {
 }
 
 func invokeFeatureGateRenderScript(binary, workDir, outputDir string, payloadVersion semver.Version, featureGateYAML string) string {
-	var script = `#!/bin/bash
+	script := `#!/bin/bash
```
enxebre: why this change?

control-plane-operator/controllers/hostedcontrolplane/v2/kas/config.go
```diff
@@ -354,6 +353,22 @@ func admissionPlugins() []string {
...
+		enabled = append(enabled, "authorization.openshift.io/RestrictSubjectBindings", "authorization.openshift.io/ValidateRoleBindingRestriction")
+	}
+
+	return enabled
+}
+
+func disabledAdmissionPlugins(cfg KubeAPIServerConfigParams) []string {
+	disabled := []string{}
+
+	if !util.ConfigOAuthEnabled(cfg.Authentication) {
```
enxebre, line 366: The automated fixtures tests are not covering config generation for this path.
Let's make sure we cover it. We could add coverage with the fixtures (see TestControlPlaneComponents) but that would result in many duplicated fixtures.
I think we can probably just add a unit for generateConfig()

cc @muraee in case you had other plans for adding new test cases such as this one.

enxebre: Also is there any e2e in place validating this change?

muraee: yeah, adding coverage in TestControlPlaneComponents will introduce too many duplicates . I think this should be covered in new local unit test for generateConfig() indeed.

ignition-server/controllers/local_ignitionprovider.go
everettraven: This was from an auto formatting change that I missed after removing my other changes. I can revert this.

control-plane-operator/controllers/hostedcontrolplane/v2/kas/config.go
everettraven: >Also is there any e2e in place validating this change?

No e2e tests in place. IIUC the only case where the OpenShift OAuth stack would be disabled is when the external OIDC configuration is specified and there is no pre-existing e2e tests for the external OIDC behavior.

I have plans to come back around to implementing e2e tests for the external OIDC feature, but if they _must_ be implemented before this can merge then I'll jump to that and return to this PR later.

I can add a new unit test for this change.

control-plane-operator/controllers/hostedcontrolplane/v2/kas/config.go
```diff
@@ -412,3 +409,58 @@ func auditWebhookConfigFile() string {
...
 }
+
+// enforceFeatureGates is a helper function that ensures the feature gates
+// are set to a particular state based on the provided feature gate strings.
+// If the existing set of feature gate strings does not contain a
+// desired feature gate string it is added.
+// If the existing set of feature gate strings does contain a desired feature gate,
+// but sets the state to a different value, it will be overrided to match
+// the desired state.
+func enforceFeatureGates(featureGates []string, addIfMissing ...string) []string {
```
enxebre: nit: addIfMissing seems a bit misleading to me, since if it's not missing is also clobbered

everettraven: Ah, good catch - I missed changing the name of this from the iteration where it was only added if missing :)

bryan-cox:
`verify` is legit failing. Run `make lint-fix` locally to fix the error and commit those changes. It should pass on CI after that.

everettraven:
@enxebre Does the failing Konflux check still need to be overridden?

enxebre:

PR #6147, opened 2025-05-12
NO-JIRA: Increase the content width in the upstream docs

bryan-cox:
This commit increases the content width in the upstream docs. The width is increased percentage wise so that a variety of monitor sizes should be accommodated.

N/A

**Checklist**


enxebre:
@bryan-cox  Did we lose the ability to see a preview on the PR?

/lgtm

PR #6155, opened 2025-05-14
OCPBUGS-53265: Use the image reference for the cache in the image metadata provider

ryan-cradick:
Allows release images to utilize tags (instead of a shah) and still utilize the image and image metadata caches.

Fixes #OCPBUGS-53265

**Checklist**

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -34,7 +34,6 @@ require (
 	github.com/go-jose/go-jose/v3 v3.0.4
 	github.com/go-logr/logr v1.4.2
 	github.com/go-logr/zapr v1.3.0
-	github.com/golang/groupcache v0.0.0-20241129210726-2c02b8208cf8
 	github.com/google/cel-go v0.22.0
 	github.com/google/go-cmp v0.7.0
 	github.com/google/gofuzz v1.2.0
@@ -159,6 +158,7 @@ require (
 	github.com/gogo/protobuf v1.3.2 // indirect
 	github.com/golang-jwt/jwt/v4 v4.5.2 // indirect
 	github.com/golang-jwt/jwt/v5 v5.2.2 // indirect
+	github.com/golang/groupcache v0.0.0-20241129210726-2c02b8208cf8 // indirect
 	github.com/golang/protobuf v1.5.4 // indirect
 	github.com/google/btree v1.1.3 // indirect
 	github.com/google/gnostic-models v0.6.9-0.20230804172637-c7be7c783f49 // indirect
--- a/support/util/imagemetadata.go
+++ b/support/util/imagemetadata.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"time"
 
 	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
 	"github.com/openshift/hypershift/support/releaseinfo"
@@ -13,6 +14,7 @@ import (
 	"github.com/openshift/hypershift/support/thirdparty/oc/pkg/cli/image/manifest"
 	"github.com/openshift/hypershift/support/thirdparty/oc/pkg/cli/image/manifest/dockercredentials"
 
+	"k8s.io/apimachinery/pkg/util/cache"
 	"k8s.io/client-go/rest"
 
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -20,14 +22,15 @@ import (
 	"github.com/blang/semver"
 	"github.com/docker/distribution"
 	"github.com/docker/distribution/registry/client/transport"
-	"github.com/golang/groupcache/lru"
 	"github.com/opencontainers/go-digest"
 )
 
+const cacheTTL = time.Hour * 12
+
 var (
-	imageMetadataCache = lru.New(1000)
-	manifestsCache     = lru.New(1000)
-	digestCache        = lru.New(1000)
+	imageMetadataCache = cache.NewLRUExpireCache(1000)
+	manifestsCache     = cache.NewLRUExpireCache(1000)
+	digestCache        = cache.NewLRUExpireCache(1000)
 )
 
 type ImageMetadataProvider interface {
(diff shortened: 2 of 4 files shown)


rtheis (changes requested):

rtheis (approved):
/lgtm

rtheis (approved):
/lgtm

support/util/imagemetadata.go
```diff
@@ -42,52 +42,26 @@ type RegistryClientImageMetadataProvider struct {
...
-				return imageConfigObject.(*dockerv1client.DockerImageConfig), nil
-			}
-		}
-		ref = &parsedImageRef
-	}
-
 	// Get the image repo info based the source/mirrors in the ICSPs/IDMSs
-	ref = seekOverride(ctx, r.OpenShiftImageRegistryOverrides, parsedImageRef)
+	ref := seekOverride(ctx, r.OpenShiftImageRegistryOverrides, parsedImageRef)
+	refPullSpec := ref.String()
```
devguyio, line 60: should we check for a nil ref here?

support/util/imagemetadata_test.go
```diff
@@ -146,7 +150,9 @@ func TestGetManifest(t *testing.T) {
 			}
 
 			if tc.validateCache {
-				_, exists := manifestsCache.Get(tc.expectedDigest)
+				parsedImageRef, err := reference.Parse(tc.imageRef)
+				g.Expect(err).NotTo(HaveOccurred())
+				_, exists := manifestsCache.Get(parsedImageRef.String())
 				g.Expect(exists).To(BeTrue())
 			}
```
devguyio: I'm thinking since we're always checking the cache now, can we also check that the image is not cached when `validateCache` is false

devguyio (commented):
Thanks for the fix! Left a few commens!

support/util/imagemetadata.go
ryan-cradick: The code didn't validate this before and in other places where `seekOverride` is called. I would speculate that this is done  because we already throw an error if `reference.Parse(imageRef)` returns an error and this same image is returned from `seekOverride` if there is an error. If you feel strongly about adding the check here, I would be glad to add it.

support/util/imagemetadata_test.go
ryan-cradick: Good idea!

rtheis (approved):
/lgtm

jeffnowicki (commented):
/lgtm

rtheis:
@ryan-cradick I think you need to fix the following:

```
./support/util/imagemetadata.go:199: mainfest ==> manifest
make: *** [Makefile:351: verify-codespell] Error 65
```

ryan-cradick:
> @ryan-cradick I think you need to fix the following:
> 
> ```
> ./support/util/imagemetadata.go:199: mainfest ==> manifest
> make: *** [Makefile:351: verify-codespell] Error 65
> ```

Not sure how it passed before. Thanks!

rtheis:
@csrwng ptal

go.mod
```diff
@@ -159,6 +158,7 @@ require (
 	github.com/gogo/protobuf v1.3.2 // indirect
 	github.com/golang-jwt/jwt/v4 v4.5.2 // indirect
 	github.com/golang-jwt/jwt/v5 v5.2.2 // indirect
```
enxebre, line 160: where's this change coming from?

enxebre:
lgtm overall ptal @jparrill

support/util/imagemetadata.go
```diff
@@ -42,52 +45,26 @@ type RegistryClientImageMetadataProvider struct {
...
-		ref = &parsedImageRef
-	}
-
 	// Get the image repo info based the source/mirrors in the ICSPs/IDMSs
-	ref = seekOverride(ctx, r.OpenShiftImageRegistryOverrides, parsedImageRef, pullSecret)
+	ref := seekOverride(ctx, r.OpenShiftImageRegistryOverrides, parsedImageRef, pullSecret)
+	refPullSpec := ref.String()
 
-	// If the image reference contains a digest, immediately look it up in the cache
-	if ref.ID != "" {
```
jparrill, line 84: What about if the seekOverrides recovers an ID? that would be overriden in every reconciliation

support/util/imagemetadata.go
```diff
@@ -42,52 +45,26 @@ type RegistryClientImageMetadataProvider struct {
...
-
-	parsedImageRef, err = reference.Parse(imageRef)
+	parsedImageRef, err := reference.Parse(imageRef)
 	if err != nil {
 		return nil, fmt.Errorf("failed to parse image reference %q: %w", imageRef, err)
 	}
 
-	// There are no ICSPs/IDMSs to process.
-	// That means the image reference should be pulled from the external registry
-	if len(r.OpenShiftImageRegistryOverrides) == 0 {
```
jparrill, line 70: Does it makes sense to look for an override if there is not any OpenShiftImageRegistryOverrides?

support/util/imagemetadata.go
```diff
@@ -98,18 +75,13 @@ func (r *RegistryClientImageMetadataProvider) ImageMetadata(ctx context.Context,
 		return nil, fmt.Errorf("failed to obtain root manifest for %s: %w", imageRef, err)
 	}
 
-	// If the image ref did not contain a digest, attempt looking it up by digest after we've fetched the digest
-	if ref.ID == "" {
```
jparrill, line 102: why don't check the cache if the ref.ID is not recovered?

support/util/imagemetadata.go
```diff
@@ -192,62 +164,40 @@ func (r *RegistryClientImageMetadataProvider) GetDigest(ctx context.Context, ima
...
-
-	parsedImageRef, err = reference.Parse(imageRef)
+	parsedImageRef, err := reference.Parse(imageRef)
 	if err != nil {
 		return nil, fmt.Errorf("failed to parse image reference %q: %w", imageRef, err)
 	}
 
-	// There are no ICSPs/IDMSs to process.
-	// That means the image reference should be pulled from the external registry
-	if len(r.OpenShiftImageRegistryOverrides) == 0 {
```
jparrill, line 222: Same comment as the first one

jparrill (commented):
Dropped some questions

go.mod
ryan-cradick: I assume this is referencing the `groupcache` dependency? It just shifted from a direct to indirect dependency once we switched from using the `lru` cache to the `cache.NewLRUExpireCache`.

support/util/imagemetadata.go
ryan-cradick: The name of the method `seekOverride` indicates that we are looking for an override and may not find one. Looking at the function, it does return the same image reference that was passed into the function if the length of `OpenShiftImageRegistryOverrides` is zero. Therefore, I think checking for the length of the override list at this point only complicates the code.

support/util/imagemetadata.go
ryan-cradick: The name of the method `seekOverride` indicates that we are looking for an override and may not find one. Looking at the function, it does return the same image reference that was passed into the function if the length of`OpenShiftImageRegistryOverrides` is zero. Therefore, I think checking for the length of the override list at this point only complicates the code.

support/util/imagemetadata.go
ryan-cradick: Prior to the change, the cache key was the `location.Manifest` and after the change, the cache key is the `refPullSpec`. The first time through the code, the config for the location will be cached using the `refPullSpec` key which will be a hit on the cache the next time this method is called with the same `refPullSpec`. Therefore, there is no need to check the cache here as it was already checked for the `refPullSpec` key.

support/util/imagemetadata.go
ryan-cradick: The `refPullSpec` string is now used as the cache key and is set right after `seekOverride` is called. The `refPullSpec` will include the ID when it is turned into a string.

support/util/imagemetadata.go
jparrill: Humm, but it's a return early measure. Did you test it in a real env?

ryan-cradick: Yes, I tested it with in our performance environment with 500+ clusters with `OpenShiftImageRegistryOverrides == 0`.

ryan-cradick: Testing was done in conjunction with https://issues.redhat.com/browse/OCPBUGS-53259, see comment there.

csrwng:

csrwng:

csrwng:

PR #6158, opened 2025-05-15
METAL-1404: Make baremetal capability disabled by default on hosted clusters

MahnoorAsghar:
Make baremetal capability disabled by default on hosted clusters

Fixes #METAL-1404: https://issues.redhat.com/browse/METAL-1404


enxebre:
Thanks @MahnoorAsghar! To reduce impact over already persisted resources and preserve control over defaults, baremetal will be disabled by default unless explicitly included in "enabled", but it doesn't need to be explicitly included in the "disabled". Let's close this PR and put a new one for the explicit opt-in implementation

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -369,14 +369,24 @@ const (
...
-type OptionalCapability string
-
-const ImageRegistryCapability OptionalCapability = OptionalCapability(configv1.ClusterVersionCapabilityImageRegistry)
-
-// capabilities allows disabling optional components at install time.
+// capabilities allows enabing or disabling optional components at install time.
 // Once set, it cannot be changed.
+// self.set1.all(e, !(e in self.set2)),
+//
+// +kubebuilder:validation:XValidation:rule="self.enabled.all(e, !(e in self.disabled))", message="Capabilities can not be both enabled and disabled at once."
```
MahnoorAsghar, line 376: CRD is invalid...error

dtantsur: I'm not overly experienced in advanced rules. In fact, for a rule so complex, I would go for the validation webhook (if the Hypershift API has any). I see that they do use quite complex rules here: https://github.com/openshift/hypershift/blob/main/api/hypershift/v1beta1/nodepool_types.go#L86-L89

The `enabled` and `disabled` fields are optional, I wonder if you need to use `has` to check for their presence. Your failure happens at the path `spec.validation.openAPIV3Schema.properties[spec].properties[capabilities].default`, and the `default` bit makes me suspect that an empty object no longer passes your validation. So what about

```
has(self.enabled) && has(self.disabled) && self.enabled.all(e, !(e in self.disabled))
```

?

MahnoorAsghar:
I'm not sure how to retry the Konflux failures, but it looks like unrelated failures

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -379,31 +379,32 @@ const (
...
+	// disabled when specified, explicitly disables the specified capabilitíes on the hosted cluster.
 	// Once set, this field cannot be changed.
 	//
 	// +listType=atomic
 	// +immutable
 	// +optional
 	// +kubebuilder:validation:MaxItems=25
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf", message="Disabled is immutable. Changes might result in unpredictable and disruptive behavior."
-	Disabled []OptionalCapability `json:"disabled,omitempty"`
+	Disabled []configv1.ClusterVersionCapability `json:"disabled,omitempty"`
```
enxebre: we don't want to consume configv1 here since that drops our enum validation. We want to be in control and gradually expose what can be enable or disable with the enum validation

enxebre: Same for Enabled

MahnoorAsghar: Done

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -379,25 +379,32 @@ const (
...
-	// OpenShift version.
+	// enabled when specified, explicitly enables the specified capabilitíes on the hosted cluster.
+	// Once set, this field cannot be changed.
 	//
+	// +listType=atomic
+	// +immutable
+	// +optional
+	// +kubebuilder:validation:MaxItems=25
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf", message="Enabled is immutable. Changes might result in unpredictable and disruptive behavior."
+	Enabled []OptionalCapability `json:"enabled,omitempty"`
```
enxebre, line 405: the list of valid input for Enabled is different than the one for Disabled. The API should reflect and enforce that so may be we need to different types OptionalCapabilityDisabled / OptionalCapabilityEnabled cc @JoelSpeed

JoelSpeed: Would you consider it an error for a user to force Disable a "Disabled by default" capability? Being explicit that they want it disabled allows them to distinguish between no opinion (the platform can choose) and having an explicit opinion (I definitely don't want this).

By allowing this choice, it allows us in the future to change whether capabilities are enabled/disabled by default without affecting customers who have an opinion.

It may not fit exactly this case, but that's the general rule for why we allow the enable/disable lists to contain the same elements (though not at the same time of course)

MahnoorAsghar: I'm alright with keeping it the same type for both enabled and disabled. If a user disables an already disabled capability, it can be a no-op. (But of course, whatever is better from the API perspective, I'm happy to do.)

enxebre: >By allowing this choice, it allows us in the future to change whether capabilities are enabled/disabled by default without affecting customers who have an opinion.

ok, I can get behind this. Thanks!

In any case let's please add validation for negative cases here TestOnCreateAPIUX

MahnoorAsghar: I'm trying to understand what you mean: do you mean more test cases in test/e2e/create_cluster_test.go:TestOnCreateAPIUX function?

MahnoorAsghar:
@bryan-cox 
Hello, could I get your review on this? (I see your comment, so there's a chance you've had a look already) :)

bryan-cox (approved):
/lgtm

bryan-cox:
/retest

For Konflux of course

PR #6131, opened 2025-05-08
[release-4.19] OCPBUGS-55693: Add support for registry root entry only in the IDMS/ICSP

jparrill:
Manual cherry pick of https://github.com/openshift/hypershift/pull/6100

PR #6163, opened 2025-05-16
OCPBUGS-54483: Add support for registry root entry only in the IDMS/ICSP

jparrill:
The reference.Parse function tries to retrieve the whole entry string, but if the entry reference does not contain Namespace and Name but registry, it will return the Registry field as the Name which is not a good behavior. We added the proper changes to have this in consideration and return the right override. Also we covered it with the unit test to avoid regressions.

Manual cherry pick from:
- https://github.com/openshift/hypershift/pull/6131


jparrill:
Need to add https://github.com/openshift/hypershift/pull/6198 too to this PR

PR #6180, opened 2025-05-22
CNTRLPLANE-902:  New metrics for HCP Azure

sdminonne:
It adds 2 metrics for Hosted Control Plane (HCP) on Azure.

For Managed HCP it adds metric: `hosted_cluster_managed_azure_info` with the following labels:
- _id
- name
- namespace
- location
- microsoft_subscription_id
- microsoft_resource_group_name
- microsoft_resource_type
- microsoft_resource_id

For not managed HCP it adds metric `hosted_cluster_azure_info` with the following labels:
- _id
- name
- namespace
- location
- microsoft_subscription_id
- microsoft_resource_group_name

The two metrics are mutually exsclusive.
 
Fixes #[CNTRLPLANE-902](https://issues.redhat.com//browse/CNTRLPLANE-902)

**Checklist**
- [x ] This change includes unit tests.

### DIFF
--- a/hypershift-operator/controllers/hostedcluster/metrics/metrics.go
+++ b/hypershift-operator/controllers/hostedcluster/metrics/metrics.go
@@ -95,8 +95,11 @@ const (
 	clusterSizeOverrideMetricHelp = "Number of HostedClusters with a cluster size override annotation"
 
 	HostedClusterManagedAzureInfoMetricName = "hosted_cluster_managed_azure_info"
-	HostedClusterManagedAzureInfoMetricHelp = "Reports ARO specific information about the given HostedCluster"
+	HostedClusterManagedAzureInfoMetricHelp = "Reports Azure managed (ARO) specific information about the given HostedCluster"
 	HostedClusterManagedAzureResourceType   = "hcpOpenShiftClusters"
+
+	HostedClusterAzureInfoMetricName = "hosted_cluster_azure_info"
+	HostedClusterAzureInfoMetricHelp = "Reports Azure information about the given HostedCluster"
 )
 
 // semantically constant - not supposed to be changed at runtime
@@ -191,7 +194,7 @@ var (
 		ClusterSizeOverrideMetricName, clusterSizeOverrideMetricHelp,
 		append(hclusterLabels, "environment", "internal_id", "size"), nil)
 
-	hostedClusterInfoDesc = prometheus.NewDesc(
+	managedAzureHostedClusterInfoDesc = prometheus.NewDesc(
 		HostedClusterManagedAzureInfoMetricName, HostedClusterManagedAzureInfoMetricHelp,
 		append(hclusterLabels,
 			"location",
@@ -199,6 +202,13 @@ var (
 			"microsoft_resource_group_name",
 			"microsoft_resource_type",
 			"microsoft_resource_id"), nil)
+
+	azureHostedClusterInfoDesc = prometheus.NewDesc(
+		HostedClusterAzureInfoMetricName, HostedClusterAzureInfoMetricHelp,
+		append(hclusterLabels,
+			"location",
+			"microsoft_subscription_id",
+			"microsoft_resource_group_name"), nil)
 )
 
 type hostedClustersMetricsCollector struct {
@@ -521,23 +531,34 @@ func (c *hostedClustersMetricsCollector) Collect(ch chan<- prometheus.Metric) {
 				}
 			}
 
-			//  ARO specific hostedClusterInfo
-			if azureutil.IsAroHCP() {
+			if hcluster.Spec.Platform.Azure != nil {
 				azInfo := hcluster.Spec.Platform.Azure
 				subID := azInfo.SubscriptionID
 				resGroup := azInfo.ResourceGroupName
-				resourceID := fmt.Sprintf("/subscriptions/%s/resourceGroups/%s/providers/Microsoft.RedHatOpenshift/hcpOpenShiftClusters/%s",
-					subID, resGroup, hcluster.Name)
-				ch <- prometheus.MustNewConstMetric(
-					hostedClusterInfoDesc,
-					prometheus.GaugeValue,
-					1.0,
-					append(hclusterLabelValues,
-						azInfo.Location,
-						subID,
-						resGroup,
-						HostedClusterManagedAzureResourceType,
-						resourceID)...)
+				if azureutil.IsAroHCP() {
+					// see https://github.com/Azure/ARO-HCP/blob/4134b5bb53782858047a0493f31b250c811eb84c/api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpclusters/preview/2024-06-10-preview/openapi.json#L131
+					resourceID := fmt.Sprintf("/subscriptions/%s/resourceGroups/%s/providers/Microsoft.RedHatOpenshift/hcpOpenShiftClusters/%s",
+						subID, resGroup, hcluster.Name)
+					ch <- prometheus.MustNewConstMetric(
+						managedAzureHostedClusterInfoDesc,
+						prometheus.GaugeValue,
+						1.0,
+						append(hclusterLabelValues,
+							azInfo.Location,
+							subID,
+							resGroup,
+							HostedClusterManagedAzureResourceType,
+							resourceID)...)
+				} else {
+					ch <- prometheus.MustNewConstMetric(
+						azureHostedClusterInfoDesc,
+						prometheus.GaugeValue,
+						1.0,
+						append(hclusterLabelValues,
+							azInfo.Location,
+							subID,
+							resGroup)...)
+				}
 			}
 
 			// invalidAwsCredsMetric
--- a/hypershift-operator/controllers/hostedcluster/metrics/metrics.go
+++ b/hypershift-operator/controllers/hostedcluster/metrics/metrics.go
@@ -8,6 +8,7 @@ import (
 
 	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
 	platformaws "github.com/openshift/hypershift/hypershift-operator/controllers/hostedcluster/internal/platform/aws"
+	"github.com/openshift/hypershift/support/azureutil"
 	"github.com/openshift/hypershift/support/conditions"
 
 	configv1 "github.com/openshift/api/config/v1"
@@ -92,6 +93,10 @@ const (
 
 	ClusterSizeOverrideMetricName = "hypershift_cluster_size_override_instances"
 	clusterSizeOverrideMetricHelp = "Number of HostedClusters with a cluster size override annotation"
+
+	HostedClusterManagedAzureInfoMetricName = "hosted_cluster_managed_azure_info"
+	HostedClusterManagedAzureInfoMetricHelp = "Reports ARO specific information about the given HostedCluster"
+	HostedClusterManagedAzureResourceType   = "hcpOpenShiftClusters"
 )
 
 // semantically constant - not supposed to be changed at runtime
@@ -185,6 +190,15 @@ var (
 	clusterSizeOverrideMetricDesc = prometheus.NewDesc(
 		ClusterSizeOverrideMetricName, clusterSizeOverrideMetricHelp,
 		append(hclusterLabels, "environment", "internal_id", "size"), nil)
+
+	hostedClusterInfoDesc = prometheus.NewDesc(
+		HostedClusterManagedAzureInfoMetricName, HostedClusterManagedAzureInfoMetricHelp,
+		append(hclusterLabels,
+			"location",
+			"microsoft_subscription_id",
+			"microsoft_resource_group_name",
+			"microsoft_resource_type",
+			"microsoft_resource_id"), nil)
 )
 
 type hostedClustersMetricsCollector struct {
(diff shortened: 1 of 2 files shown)


hypershift-operator/controllers/hostedcluster/metrics/metrics.go
```diff
@@ -92,6 +92,9 @@ const (
 
 	ClusterSizeOverrideMetricName = "hypershift_cluster_size_override_instances"
 	clusterSizeOverrideMetricHelp = "Number of HostedClusters with a cluster size override annotation"
+
+	HostedClusterInfoMetricName = "hosted_cluster_info"
```
simonpasquier, line 96: we might want 2 info metrics
* one `hosted_cluster_info` metric for all generic labels (at least hosted cluster name, namespace, _id and maybe platform type).
* one specific for Azure platform (`hosted_cluster_azure_info`) when platform type = "Azure".

One suggestion I made somewhere else was that we could put everything into "hosted_cluster_info" but it's recommended to have consistent labels per metric. And we might get similar requests in the future for AWS, GCP, ....

sdminonne: thanks @simonpasquier I'll discuss this with my team and let you know.

hypershift-operator/controllers/hostedcluster/metrics/metrics.go
```diff
@@ -185,6 +188,14 @@ var (
...
 		ClusterSizeOverrideMetricName, clusterSizeOverrideMetricHelp,
 		append(hclusterLabels, "environment", "internal_id", "size"), nil)
+
+	hostedClusterInfoDesc = prometheus.NewDesc(
+		HostedClusterInfoMetricName, HostedClusterInfoMetricHelp,
+		append(hclusterLabels, "Microsoft_subscriptionId",
+			"Microsoft_resourceGroupName",
+			"Microsoft_resourceType",
+			"Microsoft_resourceId",
+			"Location"), nil)
```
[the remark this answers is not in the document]
sdminonne: Note taken, if we'll split the metric we may want to get rid of the prefix

hypershift-operator/controllers/hostedcluster/metrics/metrics.go
sdminonne: @devguyio @simonpasquier at the moment we don't have any plan to revamp all metrics and I think I'd say I'm in favor of leaving this metrics as defined in the https://issues.redhat.com/browse/OCPSTRAT-2175

simonpasquier: I'm not sure to understand the revamp here. The question here is about adding 1 or 2 metrics?

````
hosted_cluster_info{name="foo",namespace="bar",_id="1",microsoft_subscription_id="xxx","microsoft_resource_group_name="xxx",...} 1
````

vs.

````
hosted_cluster_info{name="foo",namespace="bar",_id="1",} 1
azure_hosted_cluster_info{_id="1"subscription_id="xxx","microsoft_resource_group_nameame="xxx",...} 1
````

From experience, an Azure specific metric makes things easier to understand and query for users. Otherwise you'll have to prefix all `hosted_cluster_info` labels with the cloud provider. Also working with metrics which have labels that don't always exist is painful (for Prometheus, empty label value == the label doesn't exist).

Finally it's also less error-prone from an HyperShift developer perspective because you don't have to remember the whole list of labels when emitting the metric.

File: hypershift-operator/controllers/hostedcluster/metrics/metrics.go [hunk 3]
```diff
@@ -507,6 +518,25 @@ func (c *hostedClustersMetricsCollector) Collect(ch chan<- prometheus.Metric) {
 				}
 			}
 
+			//  ARO specific hostedClusterInfo
+			if hcluster.Spec.Platform.Type == hyperv1.AzurePlatform && hcluster.Spec.Platform.Azure != nil {
+				azInfo := hcluster.Spec.Platform.Azure
+				subID := azInfo.SubscriptionID
+				resGroup := azInfo.ResourceGroupName
+				resourceID := fmt.Sprintf("/subscriptions/%s/resourceGroups/%s/providers/Microsoft.RedHatOpenshift/hcpOpenShiftClusters/%s",
```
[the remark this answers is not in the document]
simonpasquier: Reading again https://issues.redhat.com/browse/OCPSTRAT-2175, it clearly explains why we need the full resource ID.

> ARO SRE uses the resourceId to perform remediation actions, send communication to customers, etc, when responding to incident tickets.

hypershift-operator/controllers/hostedcluster/metrics/metrics.go
```diff
@@ -185,6 +188,14 @@ var (
 	clusterSizeOverrideMetricDesc = prometheus.NewDesc(
 		ClusterSizeOverrideMetricName, clusterSizeOverrideMetricHelp,
 		append(hclusterLabels, "environment", "internal_id", "size"), nil)
+
+	hostedClusterInfoDesc = prometheus.NewDesc(
+		HostedClusterInfoMetricName, HostedClusterInfoMetricHelp,
+		append(hclusterLabels, "Microsoft_subscriptionId",
```
simonpasquier: @tony-schndr does it make sense to include other fields of the Azure's HostedCluster spec?
* Cloud
* Location
* VnetID
* SubnetID
* SecurityGroupID
* TenantID

tony-schndr: We need Location, I thought I saw that is already included?  TenantId might also be good to have though I don't think it is required by SRE and Geneva Shoebox doesn't list.
At the end of the day, everything else is nice to have but SRE has other means to get that information.  The resource ID / region allows us to look everything else up from the RP.

hypershift-operator/controllers/hostedcluster/metrics/metrics.go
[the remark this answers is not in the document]
sdminonne: I'm lost here, I thought all this was for ARO only. Maybe because the title of https://issues.redhat.com/browse/OCPSTRAT-2175 ?
@simonpasquier may you help me to find what would be `resourceID` for HCP in Azure *non*-ARO?

hypershift-operator/controllers/hostedcluster/metrics/metrics.go
sdminonne: @simonpasquier @tony-schndr thanks for this! It really help? May you this to the https://issues.redhat.com/browse/OCPSTRAT-2175 ? 
Eventually what about changing to `snake_case` as suggested by @devguyio https://github.com/openshift/hypershift/pull/6180#discussion_r2108610241
Ty!

hypershift-operator/controllers/hostedcluster/metrics/metrics.go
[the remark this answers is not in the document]
simonpasquier: Sorry for being dense! My comment was meant to clarify with myself that a "resource_id" label is indeed required for ARO SREs.

My remaining  question would be: is there a use case for HyperShift on Azure without ARO? Reading https://hypershift-docs.netlify.app/how-to/azure/create-azure-cluster_on_aks/ I would assume that the answer is "no", at least for now. If it changes in the future then it would be confusing have a resource_id label with `.../providers/Microsoft.RedHatOpenshift/hcpOpenShiftClusters/...`.

hypershift-operator/controllers/hostedcluster/metrics/metrics.go
```diff
@@ -507,6 +518,25 @@ func (c *hostedClustersMetricsCollector) Collect(ch chan<- prometheus.Metric) {
 				}
 			}
 
+			//  ARO specific hostedClusterInfo
+			if hcluster.Spec.Platform.Type == hyperv1.AzurePlatform && hcluster.Spec.Platform.Azure != nil {
```
bryan-cox: This should not use the platform if this is ARO HCP specific. We should use the `IsAROHCP` function that is used throughout the HyperShift codebase.

Self-managed will also have the platform type as Azure.

YuLi517:
@sdminonne As we already discussed and plan to implement 2 metrics, one for ARO-HCP (HyperShift on managed Azure), one for HyperShift on Azure not managed.  Would you mind to update our PR description with these 2 metrics?

sdminonne:
@YuLi517 I'm working on it

sdminonne:
@YuLi517 PR (and description) updated. May you double check and eventually let me know?
TY!

YuLi517:
> @YuLi517 PR (and description) updated. May you double check and eventually let me know? TY!

lgtm, thanks @sdminonne

simonpasquier (commented):
/lgtm

hypershift-operator/controllers/hostedcluster/metrics/metrics.go
```diff
@@ -507,6 +531,35 @@ func (c *hostedClustersMetricsCollector) Collect(ch chan<- prometheus.Metric) {
 				}
 			}
 
+			if hcluster.Spec.Platform.Azure != nil {
+				azInfo := hcluster.Spec.Platform.Azure
+				subID := azInfo.SubscriptionID
+				resGroup := azInfo.ResourceGroupName
+				if azureutil.IsAroHCP() {
```
sdminonne, line 538: @bryan-cox may you double check logic between line 534 and line 538? 
In my mind this should handle both unmanaged (line 534) and managed (line 538).
Ty!

hypershift-operator/controllers/hostedcluster/metrics/metrics.go
```diff
@@ -507,6 +531,35 @@ func (c *hostedClustersMetricsCollector) Collect(ch chan<- prometheus.Metric) {
...
+					resourceID := fmt.Sprintf("/subscriptions/%s/resourceGroups/%s/providers/Microsoft.RedHatOpenshift/hcpOpenShiftClusters/%s",
```
bryan-cox, line 540: Could we add a comment stating where this format is defined?

hypershift-operator/controllers/hostedcluster/metrics/metrics_test.go
```diff
@@ -1133,6 +1134,172 @@ func TestReportClusterSizeOverride(t *testing.T) {
...
+	value, found := os.LookupEnv("MANAGED_SERVICE")
+	if found {
+		managedServiceEnvVarPreSetValue = &value
+	}
+
+	os.Setenv("MANAGED_SERVICE", hyperv1.AroHCP)
+
+	if !found {
+		defer func() {
+			os.Unsetenv("MANAGED_SERVICE")
```
bryan-cox: I would use the test `Setenv` method rather than `os`. See here as example - https://github.com/openshift/hypershift/blob/8bc0d0d6059cd78a6100f0c503e42bcd70a783bc/cmd/cluster/aws/create_test.go#L94

hypershift-operator/controllers/hostedcluster/metrics/metrics_test.go
```diff
@@ -1133,6 +1134,172 @@ func TestReportClusterSizeOverride(t *testing.T) {
 	}
 }
 
+func TestHostedClusterManagedAzureInfo(t *testing.T) {
```
bryan-cox: Any reason not to combine the two unit tests together?

I was thinking you could splice them together and if the test case name is for managed Azure, you could set the env var in the test run and add the additional metric labels you are looking for.

hypershift-operator/controllers/hostedcluster/metrics/metrics_test.go
sdminonne: was not aware of this. :TY!

hypershift-operator/controllers/hostedcluster/metrics/metrics_test.go
sdminonne: The two metrics are mutually exclusive but I can combine the two tests anyway

hypershift-operator/controllers/hostedcluster/metrics/metrics.go
mociarain: When we have a public API there will be more official documentation we can link to but for now this should help: https://github.com/Azure/ARO-HCP/blob/4134b5bb53782858047a0493f31b250c811eb84c/api/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/hcpclusters/preview/2024-06-10-preview/openapi.json#L131

bryan-cox: I think adding a comment to the permalink would work for this comment 👍🏻

hypershift-operator/controllers/hostedcluster/metrics/metrics_test.go
sdminonne: Done

hypershift-operator/controllers/hostedcluster/metrics/metrics_test.go
sdminonne: Done

sdminonne:
@bryan-cox 

- modified the commtis comment (PTAL), 
- added permalink to  ARO `resource_id` resource
- get rid of `os.Setenv` replaced with `t.Setenv`
- merged the 2 tests

PTAL

commited on top for clarity, but I'll squash before merging

PR #6187, opened 2025-05-22
WIP: AUTOSCALE-266: Refactor karpenter-operator and karpenter into v2 components

maxcao13:
This PR refactors the karpenter-operator and karpenter into CPOv2 components.

As part of this PR, one of the commits allows the hcp_controller to now manage and reconcile the karpenter deployment instead of the karpenter-operator. This is needed because in order to use CPOv2, we need to pass in a cpoContext, and it is non-trivial to create a new cpoContext specially for the karpenter-operator container.

**Checklist**

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/v2/assets/karpenter-operator/deployment.yaml
+++ b/control-plane-operator/controllers/hostedcontrolplane/v2/assets/karpenter-operator/deployment.yaml
@@ -1,0 +1,50 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: karpenter-operator
+  namespace: HCP_NAMESPACE
+  labels:
+    app: karpenter-operator
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: karpenter-operator
+  template:
+    metadata:
+      labels:
+        app: karpenter-operator
+    spec:
+      serviceAccountName: karpenter-operator
+      terminationGracePeriodSeconds: 10
+      tolerations:
+        - key: "node-role.kubernetes.io/master"
+          effect: "NoSchedule"
+      volumes:
+        - name: target-kubeconfig
+          secret:
+            secretName: hcco-kubeconfig
+            defaultMode: 0640
+            items:
+              - key: kubeconfig
+                path: target-kubeconfig
+      containers:
+        - name: karpenter-operator
+          image: karpenter-operator # replaced by adaptDeployment
+          command:
+            - "/usr/bin/karpenter-operator"
+          args:
+            - "--target-kubeconfig=/mnt/kubeconfig/target-kubeconfig"
+            - "--namespace=$(MY_NAMESPACE)"
+          env:
+            - name: MY_NAMESPACE
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
+          resources:
+            requests:
+              memory: "60Mi"
+              cpu: "10m"
+          volumeMounts:
+            - name: target-kubeconfig
+              mountPath: "/mnt/kubeconfig"
--- a/control-plane-operator/controllers/hostedcontrolplane/v2/assets/karpenter/deployment.yaml
+++ b/control-plane-operator/controllers/hostedcontrolplane/v2/assets/karpenter/deployment.yaml
@@ -1,0 +1,97 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: karpenter
+  namespace: HCP_NAMESPACE
+  labels:
+    app: karpenter
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: karpenter
+  template:
+    metadata:
+      labels:
+        app: karpenter
+    spec:
+      serviceAccountName: karpenter
+      terminationGracePeriodSeconds: 10
+      tolerations:
+        - key: "node-role.kubernetes.io/master"
+          effect: "NoSchedule"
+      volumes:
+        - name: target-kubeconfig
+          secret:
+            # secretName: to be replaced by adaptDeployment
+            defaultMode: 0640
+            items:
+              - key: value
+                path: target-kubeconfig
+        - name: serviceaccount-token
+          emptyDir: {}
+        - name: provider-creds
+          secret:
+            secretName: "karpenter-credentials"
+      containers:
+        - name: karpenter
+          image: karpenter # replaced by payload
+          args:
+            - "--log-level=debug"
+          env:
+            - name: KUBECONFIG
+              value: "/mnt/kubeconfig/target-kubeconfig"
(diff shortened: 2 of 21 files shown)


control-plane-operator/controllers/hostedcontrolplane/v2/karpenteroperator/secret.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+func adaptCredentialsSecret(cpContext component.WorkloadContext, secret *corev1.Secret) error {
+	hcp := cpContext.HCP
+
+	awsCredentialsTemplate := `[default]
+	role_arn = %s
+	web_identity_token_file = /var/run/secrets/openshift/serviceaccount/token
+	sts_regional_endpoints = regional
+`
```
muraee, line 18: this template could be moved to the secret yaml file. The adapt function can then read the template from the secret object, replace the role_arn field and write it back.

control-plane-operator/controllers/hostedcontrolplane/v2/karpenteroperator/secret.go
```diff
@@ -0,0 +1,25 @@
...
+	arn := hcp.Spec.AutoNode.Provisioner.Karpenter.AWS.RoleARN
+
+	credentials := fmt.Sprintf(awsCredentialsTemplate, arn)
+	secret.Data = map[string][]byte{"credentials": []byte(credentials)}
+	secret.Type = corev1.SecretTypeOpaque
```
muraee, line 23: type is already set in the yaml file, no need to set it here.

control-plane-operator/controllers/hostedcontrolplane/v2/karpenteroperator/component.go
```diff
@@ -0,0 +1,66 @@
...
+		WithManifestAdapter("karpenter-credentials.yaml",
+			component.WithAdaptFunction(adaptCredentialsSecret),
+			component.ReconcileExisting(),
+		).
+		WithPredicate(predicate).
+		InjectAvailabilityProberContainer(util.AvailabilityProberOpts{}).
+		Build()
+}
+
+func predicate(cpContext component.WorkloadContext) (bool, error) {
```
muraee, line 51: shouldn't the predicate be checking if `.spec.autoNode` is enabled as well?

control-plane-operator/controllers/hostedcontrolplane/v2/assets/karpenter-operator/deployment.yaml
```diff
@@ -0,0 +1,50 @@
...
+  template:
+    metadata:
+      labels:
+        app: karpenter-operator
+    spec:
+      serviceAccountName: karpenter-operator
+      terminationGracePeriodSeconds: 10
+      tolerations:
+        - key: "node-role.kubernetes.io/master"
+          effect: "NoSchedule"
```
muraee, line 22:
```suggestion
```
tolerations are set automatically. They shouldn't be part of the manifest.

control-plane-operator/controllers/hostedcontrolplane/v2/karpenteroperator/component.go
maxcao13: That's a good point but `reconcileKarpenterOperator` already checks the AutoNode field anyways, so would this check just be good defensive practice then?
https://github.com/openshift/hypershift/blob/aedf7ae972a5fdc893f848ef26f413fadacf2199/hypershift-operator/controllers/hostedcluster/karpenter.go#L34

muraee: the predicate will also make sure that the deployment and resources are deleted when it's false, i.e. when autNode is disabled. So we definitely want to check here. 
I would personally remove the other check you linked in karpenter.go, but keeping it as good defensive practice also makes sense.

control-plane-operator/controllers/hostedcontrolplane/v2/karpenteroperator/secret.go
muraee: you are right, I was thinking of configMaps where we have the template in the yaml file. True, this doesn't work for secrets.

control-plane-operator/controllers/hostedcontrolplane/v2/assets/karpenter/deployment.yaml
```diff
@@ -0,0 +1,97 @@
...
+          effect: "NoSchedule"
+      volumes:
+        - name: target-kubeconfig
+          secret:
+            # secretName: to be replaced by adaptDeployment
+            defaultMode: 0640
+            items:
+              - key: value
+                path: target-kubeconfig
+        - name: serviceaccount-token
```
muraee, line 31: this volume and volume-mount are injected in the code, can be removed

control-plane-operator/controllers/hostedcontrolplane/v2/karpenter/role.go
```diff
@@ -0,0 +1,15 @@
...
+import (
+	"github.com/openshift/hypershift/support/config"
+	component "github.com/openshift/hypershift/support/controlplane-component"
+
+	rbacv1 "k8s.io/api/rbac/v1"
+)
+
+func adaptRole(cpContext component.WorkloadContext, role *rbacv1.Role) error {
+	ownerRef := config.OwnerRefFrom(cpContext.HCP)
+	ownerRef.ApplyTo(role)
```
muraee, line 12: why are we adding the HCP ownerRef only for the Role?

karpenter-operator/controllers/karpenter/karpenter_controller.go
```diff
@@ -138,6 +147,36 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+	imageProvider.ComponentImages()["token-minter"] = r.ControlPlaneOperatorImage
+	imageProvider.ComponentImages()[util.AvailabilityProberImageName] = r.ControlPlaneOperatorImage
+
+	cpContext := controlplanecomponent.ControlPlaneContext{
+		Context:              ctx,
+		Client:               r.ManagementClient,
+		ApplyProvider:        upsert.NewApplyProvider(false),
+		HCP:                  hcp,
+		ReleaseImageProvider: imageProvider,
+		OmitOwnerReference:   true,
```
muraee, line 172:
```suggestion
```
why is this true? the current code seems to be adding the HCP ownerRef to karpenter resources. 
related to my other comment https://github.com/openshift/hypershift/pull/6187#discussion_r2109013495

muraee:
can you please add the karpenter-operator component to `TestReconcileComponents()` here: https://github.com/openshift/hypershift/blob/main/hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go#L4189-L4200

enxebre:

PR #6195, opened 2025-05-27
OCPBUGS-56430: Add DR operations after HC restoration

jparrill:
This PR includes a new annotation and a new condition for the hosted clusters that has been restored using the HCP OADP plugin.

- Annotation: `hypershift.openshift.io/restored-from-backup`
- Condition: `HostedClusterRestoredFromBackup`

The annotation will be included by the plugin on restored clusters. The HC Controller bubbles down the annotation to the HCP and the HCCO triggers a procedure that makes sure the recovery is done properly through some concrete checks in the DataPlane side.

Once the HCCO detects `HostedClusterRestoredFromBackupAnnotation` annotation it begins the recovery actions and sets the new condition called HostedClusterRestoredFromBackup. 

It will stays on False until all the procedures and checks ends properly. After that the HCCO sets the new condition signaling the HostedCluster controller to remove the annotation of `HostedClusterRestoredFromBackupAnnotation` from the HC

- Fixes [OCPBUGS-56430](https://issues.redhat.com/browse/OCPBUGS-56430)

## Related PRs
- https://github.com/openshift/hypershift-oadp-plugin/pull/60
- https://github.com/openshift/hypershift-oadp-plugin/pull/62

### DIFF
--- a/api/hypershift/v1beta1/hostedcluster_types.go
+++ b/api/hypershift/v1beta1/hostedcluster_types.go
@@ -352,6 +352,12 @@ const (
 	// AWSMachinePublicIPs, if set to "true", results in an AWS machine template that creates machines with public IPs
 	// WARNING: This option is for development and testing purposes only
 	AWSMachinePublicIPs = "hypershift.openshift.io/aws-machine-public-ips"
+
+	// HostedClusterRestoredFromBackupAnnotation is set to true when the HostedCluster is restored from a backup using Hypershift
+	// OADP plugin. This annotation is set by the Hypershift OADP plugin during the Backup/Restore process. The annotation will trigger
+	// a process to check if the differents components in the DataPlane are working as expected. Checks:
+	// - Validates the monitoring stack is properly working after restoration, if not HCCO will restart the prometheus-k8s pods.
+	HostedClusterRestoredFromBackupAnnotation = "hypershift.openshift.io/restored-from-backup"
 )
 
 // RetentionPolicy defines the policy for handling resources associated with a cluster when the cluster is deleted.
--- a/control-plane-operator/hostedclusterconfigoperator/controllers/resources/recovery/recovery.go
+++ b/control-plane-operator/hostedclusterconfigoperator/controllers/resources/recovery/recovery.go
@@ -1,0 +1,52 @@
+package recovery
+
+import (
+	"context"
+	"fmt"
+
+	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func RecoverMonitoringStack(ctx context.Context, hcp *hyperv1.HostedControlPlane, c client.Client) error {
+	log := ctrl.LoggerFrom(ctx)
+	monitoringStackNS := "openshift-monitoring"
+	stsName := "prometheus-k8s"
+
+	// Check deployment of monitoring stack
+	prometheusSts := &appsv1.StatefulSet{}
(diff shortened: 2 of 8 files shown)


api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -352,6 +352,10 @@ const (
 	// AWSMachinePublicIPs, if set to "true", results in an AWS machine template that creates machines with public IPs
 	// WARNING: This option is for development and testing purposes only
 	AWSMachinePublicIPs = "hypershift.openshift.io/aws-machine-public-ips"
+
+	// HostedClusterRestoredFromBackupAnnotation is set to true when the HostedCluster is restored from a backup using Hypershift
+	// OADP plugin.
+	HostedClusterRestoredFromBackupAnnotation = "hypershift.openshift.io/restored-from-backup"
```
bryan-cox, line 360: Where does this annotation actually get applied to a HostedCluster?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -352,6 +352,10 @@ const (
 	// AWSMachinePublicIPs, if set to "true", results in an AWS machine template that creates machines with public IPs
 	// WARNING: This option is for development and testing purposes only
 	AWSMachinePublicIPs = "hypershift.openshift.io/aws-machine-public-ips"
+
+	// HostedClusterRestoredFromBackupAnnotation is set to true when the HostedCluster is restored from a backup using Hypershift
```
bryan-cox, line 356: Could we add more to the comment about what happens with this annotation? Something like when this annotation is set, blah blah blah happens.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/recovery/recovery.go
```diff
@@ -0,0 +1,52 @@
...
+	err := c.Get(ctx, types.NamespacedName{Namespace: monitoringStackNS, Name: stsName}, prometheusSts)
+	if err != nil {
+		if !apierrors.IsNotFound(err) {
+			return fmt.Errorf("failed to get prometheus statefulSet %s in namespace %s: %w", stsName, monitoringStackNS, err)
+		}
+
+		return fmt.Errorf("prometheus statefulSet is still starting, rescheduling reconciliation: %w", err)
+	}
+
+	if prometheusSts.Status.ReadyReplicas < prometheusSts.Status.Replicas {
```
bryan-cox, line 34: I asked Juanma offline about the necessity of force deleting the pods when the ready replicas are still less than the replicas. He stated the pods got stuck after restoration because of the volumes are still associated with the AWS volumes which potentially will not exists. This blocks the monitoring CO stuck on deployment and not reporting status. Once we delete the pods, that association is removed and the CO could continue the deployment.

api/hypershift/v1beta1/hostedcluster_types.go
jparrill: Described in the description :):

```
The annotation will be included by the plugin on restored clusters. The HC Controller bubbles down the annotation to the HCP and the HCCO triggers a procedure that makes sure the recovery is done properly through some concrete checks in the DataPlane side.
```

bryan-cox: 🤦🏻‍♂️

api/hypershift/v1beta1/hostedcluster_types.go
jparrill: Makes sense. Thank you

control-plane-operator/hostedclusterconfigoperator/controllers/resources/recovery/recovery.go
jparrill: Thank you for xref 🙏

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -352,6 +352,12 @@ const (
...
+	// OADP plugin. This annotation is set by the Hypershift OADP plugin during the Backup/Restore process. The annotation will trigger
+	// a process to check if the different in the DataPlane are working as expected. Checks:
```
bryan-cox: I think there is a typo here. What comes after different?

PR #6197, opened 2025-05-28
MGMT-20682: Enable disabling of openshift-samples capability

linoyaslan:
This PR is the first step in expanding support for disabling openshift capabilities in hypershift. Currently, only the ImageRegistry capability can be disabled. This change adds support for disabling the openshift-samples capability, who managed by the Cluster Version Operator. Additional capabilities will be supported in future updates.

**Checklist**

### DIFF
--- a/api/hypershift/v1beta1/hostedcluster_types.go
+++ b/api/hypershift/v1beta1/hostedcluster_types.go
@@ -369,10 +369,12 @@ const (
 	PruneRetentionPolicy RetentionPolicy = "Prune"
 )
 
-// +kubebuilder:validation:Enum=ImageRegistry
+// +kubebuilder:validation:Enum=ImageRegistry;openshift-samples
 type OptionalCapability string
 
 const ImageRegistryCapability OptionalCapability = OptionalCapability(configv1.ClusterVersionCapabilityImageRegistry)
+
+const OpenShiftSamplesCapability OptionalCapability = OptionalCapability(configv1.ClusterVersionCapabilityOpenShiftSamples)
 
 // capabilities allows disabling optional components at install time.
 // Once set, it cannot be changed.
--- a/test/e2e/util/util.go
+++ b/test/e2e/util/util.go
@@ -90,6 +90,32 @@ var expectedKasManagementComponents = []string{
 	"featuregate-generator",
 }
 
+type GuestClients struct {
+	CfgClient  *configv1client.Clientset
+	KubeClient *kubernetes.Clientset
+}
+
+// InitGuestClients initializes the Kubernetes and OpenShift config clients for the guest cluster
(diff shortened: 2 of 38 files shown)


test/e2e/util/util.go
```diff
@@ -2611,3 +2625,19 @@ func GenerateCustomCertificate(dnsNames []string, validity time.Duration) ([]byt
...
+func EnsureOpenshiftSamplesCapabilityDisabled(ctx context.Context, t *testing.T, g Gomega, clients *GuestClients) {
+	t.Run("EnsureOpenshiftSamplesCapabilityDisabled", func(t *testing.T) {
+		AtLeast(t, Version420)
+
+		_, err := clients.CfgClient.ConfigV1().ClusterOperators().Get(ctx, "openshift-samples", metav1.GetOptions{})
+		g.Expect(err).To(HaveOccurred())
+		g.Expect(err.Error()).To(ContainSubstring("clusteroperators.config.openshift.io \"openshift-samples\" not found"))
+
+		// ensure console resources are not present
+		_, err = clients.KubeClient.CoreV1().Namespaces().Get(ctx, "openshift-samples", metav1.GetOptions{})
```
enxebre, line 2639: This should be checking for the openshift-cluster-samples-operator
https://github.com/openshift/cluster-samples-operator/blob/main/manifests/01-namespace.yaml#L12C9-L12C43

```
teamspace-racedo-test        dev                   dev-admin-kubeconfig     Partial     True        False         The hosted control plane is available
➜  hypershift git:(main) ✗ kubectl get ns | grep sample
openshift-cluster-samples-operator                 Active   142d
```

https://github.com/openshift/cluster-samples-operator/blob/main/manifests/01-namespace.yaml#L12C9-L12C43

enxebre:
Thanks, dropped a comment https://github.com/openshift/hypershift/pull/6197#discussion_r2113692209
Looks great overall!

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -369,11 +369,13 @@ const (
 	PruneRetentionPolicy RetentionPolicy = "Prune"
 )
 
-// +kubebuilder:validation:Enum=ImageRegistry
+// +kubebuilder:validation:Enum=ImageRegistry;openshift-samples
```
csrwng, line 372: We should be consistent in the naming convention. If we already used camelcase for ImageRegistry, this should be OpenShiftSamples

csrwng (commented):

api/hypershift/v1beta1/hostedcluster_types.go
linoyaslan: Thanks for the comment :)
The name taken from configv1.ClusterVersionCapabilityOpenShiftSamples which is openshift-samples, while the ClusterVersionCapabilityImageRegistry is ImageRegistry. That's why

celebdor:

celebdor:

PR #6199, opened 2025-05-28
OCPBUGS-57450: Router publish strategy related changes for IBM Cloud platform

libesz:
When HCP contains router publish strategy for the
master services, IBM Cloud platform implementation will take care of the proper exposure of the
services, without using the actual cluster Router
(similar implementation to Azure). That is, CPO should not create any LoadBalancer or actual router deployment as part of reconciling the HCP.
The current change also makes the migration from NodePort services backward compatible, for existing clusters. That is, the NodePort services will remain as is
(they are not converted to regular ClusterIP services, resulting in permanently losing the reserved nodeports) allowing existing external clients (e.g. kubelet,
master proxy) to work as before, until they are also upgraded.

The expected behavior is to:
* Do not manage any ingress component (LB Svc, Router)
* Keep NodePort master services (if they are already existing)
* (Re)configure node agents (e.g. connectivity) to use 443 as server port

Fixes #

**Checklist**

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
@@ -1283,7 +1283,7 @@ func (r *HostedControlPlaneReconciler) reconcileOAuthServerService(ctx context.C
 	p := oauth.NewOAuthServiceParams(hcp)
 	oauthServerService := manifests.OauthServerService(hcp.Namespace)
 	if _, err := createOrUpdate(ctx, r.Client, oauthServerService, func() error {
-		return oauth.ReconcileService(oauthServerService, p.OwnerRef, serviceStrategy)
+		return oauth.ReconcileService(oauthServerService, p.OwnerRef, serviceStrategy, hcp.Spec.Platform.Type)
 	}); err != nil {
 		return fmt.Errorf("failed to reconcile OAuth service: %w", err)
 	}
@@ -1377,7 +1377,7 @@ func (r *HostedControlPlaneReconciler) reconcileOLMPackageServerService(ctx cont
 }
 
 func (r *HostedControlPlaneReconciler) reconcileHCPRouterServices(ctx context.Context, hcp *hyperv1.HostedControlPlane, createOrUpdate upsert.CreateOrUpdateFN) error {
-	if sharedingress.UseSharedIngress() {
+	if sharedingress.UseSharedIngress() || hcp.Spec.Platform.Type == hyperv1.IBMCloudPlatform {
 		return nil
 	}
 	exposeKASThroughRouter := util.IsRouteKAS(hcp)
@@ -1510,14 +1510,14 @@ func (r *HostedControlPlaneReconciler) defaultReconcileInfrastructureStatus(ctx 
 }
 
 func (r *HostedControlPlaneReconciler) reconcileInternalRouterServiceStatus(ctx context.Context, hcp *hyperv1.HostedControlPlane) (host string, needed bool, message string, err error) {
-	if !util.IsPrivateHCP(hcp) {
+	if !util.IsPrivateHCP(hcp) || hcp.Spec.Platform.Type == hyperv1.IBMCloudPlatform {
 		return
 	}
 	return r.reconcileRouterServiceStatus(ctx, manifests.PrivateRouterService(hcp.Namespace), events.NewMessageCollector(ctx, r.Client))
 }
 
 func (r *HostedControlPlaneReconciler) reconcileExternalRouterServiceStatus(ctx context.Context, hcp *hyperv1.HostedControlPlane) (host string, needed bool, message string, err error) {
-	if !util.IsPublicHCP(hcp) || !util.IsRouteKAS(hcp) || sharedingress.UseSharedIngress() {
+	if !util.IsPublicHCP(hcp) || !util.IsRouteKAS(hcp) || sharedingress.UseSharedIngress() || hcp.Spec.Platform.Type == hyperv1.IBMCloudPlatform {
 		return
 	}
 	return r.reconcileRouterServiceStatus(ctx, manifests.RouterPublicService(hcp.Namespace), events.NewMessageCollector(ctx, r.Client))
@@ -1551,12 +1551,12 @@ func (r *HostedControlPlaneReconciler) reconcileAPIServerServiceStatus(ctx conte
 		return "", 0, "", errors.New("APIServer service strategy not specified")
 	}
 
-	if sharedingress.UseSharedIngress() {
+	if sharedingress.UseSharedIngress() || hcp.Spec.Platform.Type == hyperv1.IBMCloudPlatform {
 		return sharedingress.Hostname(hcp), sharedingress.ExternalDNSLBPort, "", nil
 	}
 
 	var svc *corev1.Service
-	if serviceStrategy.Type == hyperv1.Route {
+	if serviceStrategy.Type == hyperv1.Route && hcp.Spec.Platform.Type != hyperv1.IBMCloudPlatform {
 		if util.IsPublicHCP(hcp) {
 			svc = manifests.RouterPublicService(hcp.Namespace)
 		} else {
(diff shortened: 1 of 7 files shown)


control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -1275,7 +1275,7 @@ func (r *HostedControlPlaneReconciler) reconcileOAuthServerService(ctx context.C
 	p := oauth.NewOAuthServiceParams(hcp)
 	oauthServerService := manifests.OauthServerService(hcp.Namespace)
 	if _, err := createOrUpdate(ctx, r.Client, oauthServerService, func() error {
-		return oauth.ReconcileService(oauthServerService, p.OwnerRef, serviceStrategy)
+		return oauth.ReconcileService(oauthServerService, p.OwnerRef, serviceStrategy, hcp.Spec.Platform.Type)
```
rtheis, line 1278: I think we should add unit testing if possible.

rtheis:
Please open a Jira and update this PR accordingly.

jeffnowicki:
/retitle OCPBUGS-57450: Router publish strategy related changes for IBM Cloud platform

libesz:
@rtheis I believe Konnectivity is covered here: https://github.com/openshift/hypershift/pull/6199/files#diff-61e7a39d46e3dc4071d00324eec2a347a217069cc26716d393b770ba5a5e0bafR292
But I agree, ignition was missing. Found only in cpov2, added a commit to handle.

libesz:
@sjenning @rtheis @csrwng I updated the PR and added new unit tests for the changed functions. It is ready from my side.

sjenning:
/lgtm

control-plane-operator/controllers/hostedcontrolplane/kas/service.go
```diff
@@ -87,7 +87,9 @@ func ReconcileService(svc *corev1.Service, strategy *hyperv1.ServicePublishingSt
 			portSpec.NodePort = strategy.NodePort.Port
 		}
 	case hyperv1.Route:
-		svc.Spec.Type = corev1.ServiceTypeClusterIP
+		if ((hcp.Spec.Platform.Type == hyperv1.IBMCloudPlatform) && (svc.Spec.Type != corev1.ServiceTypeNodePort)) || (hcp.Spec.Platform.Type != hyperv1.IBMCloudPlatform) {
```
sjenning: This can be simplified

```suggestion
		if hcp.Spec.Platform.Type != hyperv1.IBMCloudPlatform || svc.Spec.Type != corev1.ServiceTypeNodePort {
```

control-plane-operator/controllers/hostedcontrolplane/kas/service.go
```diff
@@ -289,7 +291,9 @@ func ReconcileKonnectivityServerService(svc *corev1.Service, ownerRef config.Own
...
```
sjenning:
```suggestion
		if hcp.Spec.Platform.Type != hyperv1.IBMCloudPlatform || svc.Spec.Type != corev1.ServiceTypeNodePort {
```

libesz:
@sjenning thanks, I committed your suggestions. Would you please LGTM this PR again?

control-plane-operator/controllers/hostedcontrolplane/v2/ignitionserver/service.go
```diff
@@ -27,7 +27,10 @@ func adaptService(cpContext component.WorkloadContext, svc *corev1.Service) erro
 			svc.Spec.Ports[0].NodePort = strategy.NodePort.Port
 		}
 	case hyperv1.Route:
-		svc.Spec.Type = corev1.ServiceTypeClusterIP
+		if (cpContext.HCP.Spec.Platform.Type != hyperv1.IBMCloudPlatform) ||
+			(cpContext.HCP.Spec.Platform.Type == hyperv1.IBMCloudPlatform && svc.Spec.Type != corev1.ServiceTypeNodePort) {
```
muraee: can you please elaborate on this part of the check
```
cpContext.HCP.Spec.Platform.Type == hyperv1.IBMCloudPlatform && svc.Spec.Type != corev1.ServiceTypeNodePort
```

muraee: I have read the PR description just now, I think this is trying to check the already existing service. This won't work as it because the passed in service obj to this func is just the static manifest. You would need to explicitly fetch the old one here and do the check.

muraee: something like:
```go
existingServiceUsesNodePort := false
existingService := &corev1.Service{}
if err := cpContext.Client.Get(cpContext.Context, client.ObjectKeyFromObject(svc), existingService); err != nil {
		return fmt.Errorf("failed to get existing ignition service: %w", err)
	}
} else {
	existingServiceUsesNodePort = existingService.Spec.Type == corev1.ServiceTypeNodePort
}

...
```

muraee: also note that the function returns early on #L14 when platform is not IBM, so you can skip the platform check here.

libesz: Thanks! I updated the logic and the tests with this dynamic fetch of the existing service.

control-plane-operator/controllers/hostedcontrolplane/v2/ignitionserver/service.go
```diff
@@ -27,7 +43,14 @@ func adaptService(cpContext component.WorkloadContext, svc *corev1.Service) erro
 			svc.Spec.Ports[0].NodePort = strategy.NodePort.Port
 		}
 	case hyperv1.Route:
-		svc.Spec.Type = corev1.ServiceTypeClusterIP
+		if existingServiceUsesNodePort {
+			svc.Spec.Type = corev1.ServiceTypeNodePort
+			if strategy.NodePort != nil {
```
muraee: this shouldn't be possible if strategy type is `Route`

control-plane-operator/controllers/hostedcontrolplane/v2/ignitionserver/service.go
```diff
@@ -8,13 +8,29 @@ import (
...
 
+	existingServiceUsesNodePort := false
+	existingService := &corev1.Service{}
+	if err := cpContext.Client.Get(cpContext.Context, client.ObjectKeyFromObject(svc), existingService); err != nil {
+		if !apierrors.IsNotFound(err) {
+			return fmt.Errorf("failed to get existing ignition service: %w", err)
+		}
+	} else {
+		if len(existingService.Spec.Ports) != 1 {
+			return fmt.Errorf("existing ignition service must have exactly one port exposed, this has: %d", len(existingService.Spec.Ports))
```
muraee: why would you block the creation/update of service if the existing one is invalid? This error can only be resolved by manually changing/deleting the existing service

control-plane-operator/controllers/hostedcontrolplane/v2/ignitionserver/service.go
libesz: Correct. Removed.

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
libesz: Added.

control-plane-operator/controllers/hostedcontrolplane/v2/ignitionserver/service.go
libesz: I wanted to make sure we are not accessing a missing 0th element later on (in case of an invalid existing service that has 0 ports). Now I updated the code to assume there is a usable existing service, only when it is 1) nodeport type and 2) has one port. No new error branches.

rtheis:
/remove-hold now that the PR comments have been resolved.

PR #6200, opened 2025-05-28
CNTRLPLANE-201: feat(shared-ingress): Bump ubi10 out of beta

celebdor:

Now that UBI 10 is no longer in beta state, we should move to the GA version.

Fixes #[CNTRLPLANE-201](https://issues.redhat.com//browse/CNTRLPLANE-201)

**Checklist**


celebdor:
/override "Red Hat Konflux / hypershift-shared-ingress-enterprise-contract / hypershift-shared-ingress-main"

PR #6206, opened 2025-05-29
AUTOSCALE-119: Add pod monitors to karpenter components

maxcao13:
This PR simply exposes karpenter and karpenter-operator existing metrics to be scrapped by prometheus in the management cluster. It does so by adding podmonitors targetting the corresponding metrics ports defined in the karpenter/karpenter-operator CPOv2 manifests. 

Fixes https://issues.redhat.com/browse/AUTOSCALE-119

**Checklist**


maxcao13:
I'm not sure if I need to enable anything else to make this work, nor do I know how to enable control-plane-component pod monitors, so please let me know. 

But I was at least able to show it works by adding the `openshift.io/cluster-monitoring: 'true'` label to the `HCP_NAMESPACE` and add rbac to the `prometheus-k8s` service account in `openshift-monitoring` to make the metrics show up on the console:

Screenshot From 2025-05-29 12-17-25

The first metric in the screenshot is from the karpenter deployment, the second (not as useful) one is from the karpenter-operator. I'm assuming it is exposed from the `controller_runtime` dependency in the operator.

maxcao13:
/test e2e-aws-techpreview

testing for regressions, but this PR shouldn't affect any core functionality.

control-plane-operator/controllers/hostedcontrolplane/v2/assets/karpenter-operator/podmonitor.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PodMonitor
+metadata:
+  name: karpenter-operator
+  namespace: HCP_NAMESPACE
+spec:
+  namespaceSelector:
+    matchNames:
+      - HCP_NAMESPACE
```
prozehna, line 9: Just curious how the actual namespace string is injected here?

prozehna:
FWIW - ACK'd the approach, thanks for the learning opportunity. So far had one question re. namespace injection.

control-plane-operator/controllers/hostedcontrolplane/v2/assets/karpenter-operator/podmonitor.yaml
maxcao13: I believe all the workload manifests get their namespaces overridden here during reconcile of the control plane component: https://github.com/openshift/hypershift/blob/94b059b126a488a85468b099ad2c753e5558f22e/support/controlplane-component/controlplane-component.go#L289

enxebre:
>But I was at least able to show it works by adding the openshift.io/cluster-monitoring: 'true' label to the HCP_NAMESPACE and add rbac to the prometheus-k8s service account in openshift-monitoring to make the metrics show up on the console:

That's exactly how it works. Underneath the HO can automate Cluster Monitoring or UWM based on some flags.
In real envs for self-hosted we automate UWM as recommended to scrape cp metrics.
In Managed envs, there are adhoc solutions that scrape those.

Cluster monitoring can be used in some envs but it's not intended for it. It's used in this CI jobs for example, so you should be able to validate the metrics are exposed as expected. Can you please add some smoke validation for them via existing [`ValidateMetrics`](https://github.com/openshift/hypershift/blob/28c182fc43a0603e0ab453b0c59959999d6a2b31/test/e2e/util/hypershift_framework.go#L161)

PR #6208, opened 2025-05-29
OCPBUGS-36389: Omit metadata for telemetry remote write and ensure _id label

csrwng:
Explicitly configures UWM to not send metadata and ensure that metrics sent to telemetry include the _id label

Fixes #[OCPBUGS-36389](https://issues.redhat.com/browse/OCPBUGS-36389)

**Checklist**

### DIFF
--- a/hypershift-operator/controllers/uwmtelemetry/uwm_telemetry.go
+++ b/hypershift-operator/controllers/uwmtelemetry/uwm_telemetry.go
@@ -192,6 +192,14 @@ func reconcileRemoteWriteSecret(src, dest *corev1.Secret, clusterID string) erro
 	return nil
 }
 
+// reconcileUWMConfigContent updates the given ConfigMap's "config.yaml" to ensure there is a Prometheus
+// remoteWrite entry for the UWM telemetry endpoint.
+//
+// It creates or replaces a remoteWrite entry that points to the hardcoded telemetry remote write URL,
+// configures authorization to use the UWM remote-write secret, enforces queue and metadata settings
+// (ensuring `metadataConfig.send` is false), and appends the provided relabelConfig to the entry's
+// WriteRelabelConfigs when non-nil. The resulting configuration is written back into cm.Data["config.yaml"].
+// An error is returned if parsing, conversion, or serialization fails.
 func reconcileUWMConfigContent(cm *corev1.ConfigMap, relabelConfig *monv1.RelabelConfig) error {
 	content := map[string]interface{}{}
 	if contentString, exists := cm.Data["config.yaml"]; exists {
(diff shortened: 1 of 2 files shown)


csrwng:
@coderabbitai review

hypershift-operator/controllers/uwmtelemetry/uwm_telemetry.go
```diff
@@ -249,14 +257,29 @@ func reconcileUWMConfigContent(cm *corev1.ConfigMap, relabelConfig *monv1.Relabe
...
 	}
 	if relabelConfig != nil {
-		telemetryRemoteWrite.WriteRelabelConfigs = []monv1.RelabelConfig{*relabelConfig}
+		telemetryRemoteWrite.WriteRelabelConfigs = append(telemetryRemoteWrite.WriteRelabelConfigs, *relabelConfig)
 	}
 	telemetryRemoteWriteMap, err := toUnstructuredMap(telemetryRemoteWrite)
 	if err != nil {
 		return fmt.Errorf("cannot convert remote write config to unstructured map: %w", err)
 	}
+	// Ensure that send=false is written out for metadata config
```
simonpasquier, line 280: why do we update again the medataConfig.enabled field?

hypershift-operator/controllers/uwmtelemetry/uwm_telemetry.go
```diff
@@ -249,14 +257,29 @@ func reconcileUWMConfigContent(cm *corev1.ConfigMap, relabelConfig *monv1.Relabe
 			MinBackoff:        ptr.To(monv1.Duration("1s")),
 			MaxBackoff:        ptr.To(monv1.Duration("256s")),
 		},
+		// Omit metadata since telemeter does not allow remote write metadata.
+		MetadataConfig: &monv1.MetadataConfig{
+			Send: false,
+		},
+		WriteRelabelConfigs: []monv1.RelabelConfig{
```
simonpasquier, line 265:
```suggestion
		// The Telemeter server will reject samples without a `_id` label. The value corresponds to the identifier of the hosted cluster.
		WriteRelabelConfigs: []monv1.RelabelConfig{
```

hypershift-operator/controllers/uwmtelemetry/uwm_telemetry.go
csrwng: Because when it's false and the config is serialized, the field is omitted

simonpasquier: ah let me fix this!

simonpasquier: hmm I realize that we can't change the field definition as it might break users relying on the existing behavior which is
* if metadataConfig isn't defined (e.g. nil), Prometheus will send metadata.
* if metadataConfig is defined but empty, Prometheus won't send metadata.

It's not super-intuitive but we'll have to deal with it. I will at least update the API documentation to mention the subtlety.

TL;DR: even if the `send` field is omitted in the serialized output, setting `metadataConfig` to a non-nil value should be enough to disable metadata sending but I agree that setting it explicitly here is more explicit.

PR #6218, opened 2025-06-02
OCPBUGS-56914: fix: Do not admit OAuth route by private router unless it has external DNS

mgencur:

* Label the OAuth Route for admission by the private router only if the OAuth endpoint has hostname (external DNS). This fixes the case when KAS is exposed as LoadBalancer and OAuth via Route without specifying external DNS.
* In order to fix the previous point, label routes for admission by the private router independently from each other (previously, all routes were labeled based on KAS Route)
* Expose private HCP router service if the endpoint access is Private or any endpoint exposes Route with external DNS. (previously it took into account just KAS Route). This fixes the missing routerCanonicalHostname on OAuth Route when KAS has LoadBalancer strategy and OAuth has Route *with* external DNS.
* Add unit test for renconciling OAuth Service and Routes

Fixes [OCPBUGS-56914](https://issues.redhat.com/browse/OCPBUGS-56914)

**Checklist**

### DIFF
--- a/support/util/expose.go
+++ b/support/util/expose.go
@@ -19,16 +19,36 @@ func IsLBKAS(hcp *hyperv1.HostedControlPlane) bool {
 }
 
 func IsRouteKAS(hcp *hyperv1.HostedControlPlane) bool {
-	apiServerService := ServicePublishingStrategyByTypeForHCP(hcp, hyperv1.APIServer)
-	return apiServerService != nil && apiServerService.Type == hyperv1.Route
+	return IsRoute(hcp, hyperv1.APIServer)
 }
 
-func UseDedicatedDNSforKAS(hcp *hyperv1.HostedControlPlane) bool {
-	apiServerService := ServicePublishingStrategyByTypeForHCP(hcp, hyperv1.APIServer)
-	return IsRouteKAS(hcp) &&
-		// When using dedicated DNS apiServerService.Route.Hostname is set explicitly
+func IsRoute(hcp *hyperv1.HostedControlPlane, svcType hyperv1.ServiceType) bool {
+	svc := ServicePublishingStrategyByTypeForHCP(hcp, svcType)
+	return svc != nil && svc.Type == hyperv1.Route
+}
+
+func UseDedicatedDNSForKAS(hcp *hyperv1.HostedControlPlane) bool {
+	return UseDedicatedDNS(hcp, hyperv1.APIServer)
+}
+
+func UseDedicatedDNSForOAuth(hcp *hyperv1.HostedControlPlane) bool {
+	return UseDedicatedDNS(hcp, hyperv1.OAuthServer)
+}
+
+func UseDedicatedDNSForKonnectivity(hcp *hyperv1.HostedControlPlane) bool {
+	return UseDedicatedDNS(hcp, hyperv1.Konnectivity)
+}
+
+func UseDedicatedDNSForIgnition(hcp *hyperv1.HostedControlPlane) bool {
+	return UseDedicatedDNS(hcp, hyperv1.Ignition)
+}
+
+func UseDedicatedDNS(hcp *hyperv1.HostedControlPlane, svcType hyperv1.ServiceType) bool {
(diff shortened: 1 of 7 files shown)


mgencur:
The primary issue reported in the JIRA is fixed and the Route looks like this:
```
ᐅ oc get route oauth -oyaml     
apiVersion: route.openshift.io/v1
kind: Route
metadata:
  annotations:
    openshift.io/host.generated: "true"
  creationTimestamp: "2025-06-02T05:38:02Z"
  name: oauth
  namespace: clusters-hc1
  ownerReferences:
  - apiVersion: hypershift.openshift.io/v1beta1
    blockOwnerDeletion: true
    controller: true
    kind: HostedControlPlane
    name: hc1
    uid: a05be54f-a4d3-46b3-b585-be9d87f72e4c
  resourceVersion: "26385"
  uid: 1307979d-c67a-4fea-ba21-ddc4876589a4
spec:
  host: oauth-clusters-hc1.apps.mgencur-mgmt.mgencur.hypershift.devcluster.openshift.com
  tls:
    insecureEdgeTerminationPolicy: None
    termination: passthrough
  to:
    kind: Service
    name: oauth-openshift
    weight: 100
  wildcardPolicy: None
status:
  ingress:
  - conditions:
    - lastTransitionTime: "2025-06-02T05:38:02Z"
      status: "True"
      type: Admitted
    host: oauth-clusters-hc1.apps.mgencur-mgmt.mgencur.hypershift.devcluster.openshift.com
    routerCanonicalHostname: router-default.apps.mgencur-mgmt.mgencur.hypershift.devcluster.openshift.com
    routerName: default
    wildcardPolicy: None
```
Note: It is admitted by the default router from the management cluster instead of the private router when KAS uses LoadBalancer AND OAuth uses Route, without defining external DNS.

support/util/expose.go
```diff
@@ -23,6 +23,21 @@ func IsRouteKAS(hcp *hyperv1.HostedControlPlane) bool {
 	return apiServerService != nil && apiServerService.Type == hyperv1.Route
 }
 
+func IsRouteOAuth(hcp *hyperv1.HostedControlPlane) bool {
```
bryan-cox: nit: do these really need to be exported functions?

mgencur: They don't have to. I just did that in the same way as for KAS.
But they can be made package private or their logic can be included in the functions where they're called.
WDYT?

bryan-cox: I think its fine either way. I would keep them separate to cut back on the return logic. 

These all still need unit tests too, no?

mgencur: Specifically, OAuth should be covered here: https://github.com/openshift/hypershift/pull/6218/files#diff-34356734f0509551aa83a787c138df16e072fb1938394cbb4c00ba7bbbc12df2R1243
Regarding Ignition and Konnectivity, there's the same logic so I thought it would be safe to have just a test for OAuth.

support/util/expose.go
```diff
@@ -31,6 +46,21 @@ func UseDedicatedDNSforKAS(hcp *hyperv1.HostedControlPlane) bool {
...
+func UseDedicatedDNSForKonnectivity(hcp *hyperv1.HostedControlPlane) bool {
+	konnService := ServicePublishingStrategyByTypeForHCP(hcp, hyperv1.Konnectivity)
+	return IsRouteKonnectivity(hcp) && konnService.Route != nil && konnService.Route.Hostname != ""
+}
+
+func UseDedicatedDNSForIgnition(hcp *hyperv1.HostedControlPlane) bool {
+	ignitionService := ServicePublishingStrategyByTypeForHCP(hcp, hyperv1.Ignition)
+	return IsRouteIgnition(hcp) && ignitionService.Route != nil && ignitionService.Route.Hostname != ""
+}
+
```
muraee: those functions could be merged into one function that takes the service type as an argument

mgencur: @muraee @bryan-cox Alright, pushed one more commit to merge those functions and remove some duplication.

mgencur:
Rebased.

PR #6219, opened 2025-06-02
OCPBUGS-56930: test/e2e: fix version gate checks on minor upgrade

sjenning:
For the `e2e-aws-upgrade` job, there is minor (y-stream) skew between `initial` and `latest` releases.

The current version checking code just uses the `latest` image for the version checks but, on `TestUpgradeControlPlane`, `inital` can be minor skew, which can result in checks not being gated properly pre-upgrade.

This PR sets the version to be that of `initial` then switches it to `latest` in `TestUpgradeControlPlane` after the initial rollout checks, but before the HCP upgrade begins.

### DIFF
--- a/test/e2e/control_plane_upgrade_test.go
+++ b/test/e2e/control_plane_upgrade_test.go
@@ -30,9 +30,15 @@ func TestUpgradeControlPlane(t *testing.T) {
 		// Sanity check the cluster by waiting for the nodes to report ready
 		guestClient := e2eutil.WaitForGuestClient(t, ctx, mgtClient, hostedCluster)
 
+		// Set the semantic version to the latest release image for version gating tests
+		err := e2eutil.SetReleaseImageVersion(testContext, globalOpts.LatestReleaseImage, globalOpts.ConfigurableClusterOptions.PullSecretFile)
+		if err != nil {
+			g.Expect(err).NotTo(HaveOccurred(), "failed to set latest release image version")
+		}
+
 		// Update the cluster image
 		t.Logf("Updating cluster image. Image: %s", globalOpts.LatestReleaseImage)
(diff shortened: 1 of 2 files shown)


sjenning:
xref https://github.com/openshift/hypershift/pull/6211

test/e2e/control_plane_upgrade_test.go
```diff
@@ -30,6 +30,9 @@ func TestUpgradeControlPlane(t *testing.T) {
 		// Sanity check the cluster by waiting for the nodes to report ready
 		guestClient := e2eutil.WaitForGuestClient(t, ctx, mgtClient, hostedCluster)
 
+		// Set the semantic version to the latest release image for version gating tests
+		_ = e2eutil.SetReleaseImageVersion(testContext, globalOpts.LatestReleaseImage, globalOpts.ConfigurableClusterOptions.PullSecretFile)
```
bryan-cox: Why explicitly ignore the err here?

jparrill:
/retest

The e2e-aws-minimal failed on teardown

test/e2e/control_plane_upgrade_test.go
jparrill: +1
```suggestion
		err = e2eutil.SetReleaseImageVersion(testContext, globalOpts.LatestReleaseImage, globalOpts.ConfigurableClusterOptions.PullSecretFile)
		if err != nil {
			g.Expect(err).NotTo(HaveOccurred(), "failed to set release image version")
	}
```

jparrill (commented):
Dropped a comment. Thanks!

test/e2e/control_plane_upgrade_test.go
sjenning: My thinking was that we should still try to proceed even if we couldn't update the semver.

But after sleeping on it, that could cause it to fail in weird ways.  I'll update it.

PR #6225, opened 2025-06-04
OCPBUGS-56966: Fix ignition server route getting constantly recreated

muraee:
Previously ignition-server and ignition-server-proxy reconciled the same route based on a predicate. However, because one of the predicates is always false, this will cause the route to be deleted while the other component with the true predicate keeps recreating it.

Fixes #

**Checklist**


sjenning:
/lgtm
/test verify

PR #6227, opened 2025-06-05
CNTRLPLANE-374: hcco: handle post-install creation of OIDC client secrets

sjenning:


csrwng:
@sjenning can you please add a commit message that describes the change? 
Also please add a unit test for the ReconcileAuthenticationConfiguration function.
Other than that, lgtm

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -1182,6 +1182,10 @@ func (r *reconciler) reconcileAuthOIDC(ctx context.Context, hcp *hyperv1.HostedC
 		// Copy OIDCClient Secrets into openshift-config namespace
 		if len(hcp.Spec.Configuration.Authentication.OIDCProviders[0].OIDCClients) > 0 {
 			for _, oidcClient := range hcp.Spec.Configuration.Authentication.OIDCProviders[0].OIDCClients {
+				// If the secret name is empty, we assume the guest cluster admin is going to create this secret manually
```
jparrill: At this point do we know that `len(hcp.Spec.Configuration.Authentication.OIDCProviders) > 0`? Just to avoid panics

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -1182,6 +1182,10 @@ func (r *reconciler) reconcileAuthOIDC(ctx context.Context, hcp *hyperv1.HostedC
...
+				if oidcClient.ClientSecret.Name == "" {
+					continue
```
jparrill: I would add something like:

```
log.Info("OIDC client secret name is empty, skipping copy. Expected to be created manually.", "component", oidcClient.ComponentName)
```
to warn the user about the potential misconfiguration

support/globalconfig/authentication.go
```diff
@@ -21,5 +23,12 @@ func ReconcileAuthenticationConfiguration(authentication *configv1.Authenticatio
 		authentication.Spec = *config.Authentication
 	}
 	authentication.Spec.ServiceAccountIssuer = issuerURL
+	for i := range authentication.Spec.OIDCProviders {
+		for j, client := range authentication.Spec.OIDCProviders[i].OIDCClients {
+			if client.ClientSecret.Name == "" {
+				authentication.Spec.OIDCProviders[i].OIDCClients[j].ClientSecret.Name = fmt.Sprintf("%s-post-install-client-secret", client.ComponentName)
```
jparrill: Set `-post-install-client-secret` as a possible constant?

jparrill (commented):

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
sjenning: Not against it, but this isn't visible to the user.  It would just be in the HCCO logs on the mgmt cluster.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
sjenning: The range will immediately exit if the len is 0 so the check is not needed.  I probably meant to check that `len(hcp.Spec.Configuration.Authentication.OIDCProviders) > 0`

devguyio:
leaving it to @jparrill to unhold 
/hold

PR #6228, opened 2025-06-05
OCPBUGS-57125: Support RHOBS when loading cpov2 monitoring manifests by replacing the API group

muraee:

This is needed when RHOBS is enabled, otherwise the monitoring manifests like PodMonitors will fail to load as they have the prometheous api group which is not registered in the scheme when RHOBS is enabled.

Fixes #

**Checklist**


csrwng:
`TestNodePool/HostedCluster0/Main/TestNodepoolMachineconfigGetsRolledout` is failing on azure e2e but it is also failing in the same way in other PRs. Overriding so we can get this fix in.
/override ci/prow/e2e-aks

PR #6236, opened 2025-06-06
CNTRLPLANE-952: New API to support specifying cluster-autoscaler flags

celebdor:

    Adding a new API to support specifying flags for the cluster-autoscaler.
    Allowing users to select the expander used by the cluster-autoscaler.
    Allowing users to specify node grouping ignored labels
    Add platform specific ignored labels (From CAO)
    Enhancing customization options for scaling node groups.

Co-Authored-by: marcbrown2024

Fixes #CNTRLPLANE-952

**Checklist**


celebdor:
I still need to add unit tests, some cli support for passing flags and improve the kubebuilder validation

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1248,9 +1248,72 @@ const (
 // +kubebuilder:validation:Enum=Karpenter
 type Provisioner string
 
+// Configures when and how to scale down cluster nodes.
+type ScaleDownConfig struct {
+	// enabled determines whether the Cluster Autoscaler should scaled down this cluster.
```
enxebre: let's document default behaviour and let the backend default to true.
A bool here will need an lint exception, see https://github.com/openshift/hypershift/blob/main/api/.golangci.yml#L118-L121
for an example

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1248,9 +1248,72 @@ const (
...
+	Enabled bool `json:"enabled"`
+
+	// delayAfterAdd sets how long after scale up the scale down evaluation resumes
```
enxebre: let's document default behaviour and an example with expected format.
MaxLength seems too generous?
We also need a minLength

enxebre: similar feedback for all the ones below

control-plane-operator/controllers/hostedcontrolplane/v2/autoscaler/deployment.go
```diff
@@ -11,34 +13,203 @@ import (
...
 const (
-	kubeconfigVolumeName = "kubeconfig"
+	leastWasteFlag string = "least-waste"
+	priorityFlag   string = "priority"
+	randomFlag     string = "random"
 )
 
+// AWS cloud provider ignore labels for the autoscaler.
+const (
+	// AwsIgnoredLabelEksctlInstanceId  is a label used by eksctl to identify instances.
```
enxebre: why would we need to exclude any eks / eksctl label?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1248,9 +1248,72 @@ const (
 // +kubebuilder:validation:Enum=Karpenter
 type Provisioner string
 
+// Configures when and how to scale down cluster nodes.
```
enxebre, line 1265: ptal @JoelSpeed

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1248,9 +1248,72 @@ const (
...
+type ScaleDownConfig struct {
+	// enabled determines whether the Cluster Autoscaler should scaled down this cluster.
+	Enabled bool `json:"enabled"`
```
JoelSpeed: Have you considered using some kind of enum to represent the state instead of a bool? Bools don't typically evolve well over time, we prefer enums with meaningful values that describe the behaviour that they configure

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1248,9 +1248,72 @@ const (
...
+
+	// delayAfterAdd sets how long after scale up the scale down evaluation resumes
+	// +kubebuilder:validation:Pattern=([0-9]*(\.[0-9]*)?[a-z]+)+
```
JoelSpeed: Are you trying to validate that this is a duration? 

Since validation for durations is complex to implement, it's preferred to include a unit and turn this into an int32, eg, `delayAfterAddSeconds` and then it's unambiguous of the unit, the value is clear to validate and it's easy to validate whether you're in go, or not

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1248,9 +1248,72 @@ const (
...
+	// +kubebuilder:validation:MaxLength=100
+	// +optional
+	DelayAfterAdd *string `json:"delayAfterAdd,omitempty"`
```
JoelSpeed: For each of the delays, what is the behaviour if I were to explicitly set `0` as the value?

api/hypershift/v1beta1/hostedcluster_types.go
JoelSpeed: If this is a required field, mark it explicitly `+required`

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1248,9 +1248,72 @@ const (
...
+	// +optional
+	DelayAfterDelete *string `json:"delayAfterDelete,omitempty"`
+
+	// delayAfterFailure sets how long after a scale down failure, scale down evaluation resumes
+	// +kubebuilder:validation:Pattern=([0-9]*(\.[0-9]*)?[a-z]+)+
+	// +kubebuilder:validation:MaxLength=100
+	// +optional
+	DelayAfterFailure *string `json:"delayAfterFailure,omitempty"`
+
+	// unneededTime establishes how long a node should be unneeded before it is eligible for scale down
```
JoelSpeed: Perhaps again this should be a duration with units? `unneededDurationSeconds`?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1248,9 +1248,72 @@ const (
...
+
 // ClusterAutoscaling specifies auto-scaling behavior that applies to all
 // NodePools associated with a control plane.
 type ClusterAutoscaling struct {
+	// scaleDown configures the behavior of the Cluster Autoscaler scale down operation.
+	//
+	// +optional
+	ScaleDown *ScaleDownConfig `json:"scaleDown,omitempty"`
+
+	// balancingIgnoredLabels sets "--balancing-ignore-label <label name>" flag on cluster-autoscaler for each listed label.
```
JoelSpeed, line 1363: This field only makes sense when using the balance similar nodes option right? So, can we validate that this list is empty when that feature is disabled?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1285,6 +1348,19 @@ type ClusterAutoscaling struct {
...
+	// Options include:
+	// * LeastWaste - selects the group with minimal idle CPU and memory after scaling.
+	// * Priority - selects the group with the highest user-defined priority.
+	// * Random - selects a group randomly.
+	// If not specified, `Random` is the default.
+	// Maximum of 3 expanders can be specified.
+	// +kubebuilder:validation:MaxItems=3
+	//
+	// +optional
+	Expanders []ExpanderString `json:"expanders,omitempty"`
```
JoelSpeed, line 1445: Should also set `MinItems` 1 to prevent an unstructured client from setting `expanders: []` which wouldn't round trip through a structured client

api/hypershift/v1beta1/hostedcluster_types.go
celebdor: I did. But I wanted to keep the API matching https://github.com/openshift/cluster-autoscaler-operator/blob/master/pkg/apis/autoscaling/v1/clusterautoscaler_types.go#L164-L164

enxebre: I'm leaning towards consistency with the existing cao API and the abstracted flag here as well.

JoelSpeed: The API review guidance changes over time. In core openshift APIs we don't allow bools at all now. Typically what happens is you end up with 1 bool, and later a second bool that is only valid if the first bool is set to a specific value, where as an enum this could have evolved just a new enum value and not a separate oddly behaved field.

We may be confident in this case that it's not changing, but if this were in o/api I would be making you change it even if it is inconsistent

enxebre: I have no strong objections and I prefer if we use the same criteria here and in o/api, so let's make it an enum.

LiangquanLi930:
I will continue working on this PR^^

api/hypershift/v1beta1/hostedcluster_types.go
LiangquanLi930: IIUC this depends on https://github.com/openshift/cluster-autoscaler-operator. When DelayAfterAdd is set to 0, tasks are added to the queue and processed immediately, without any delay.

api/hypershift/v1beta1/hostedcluster_types.go
LiangquanLi930: because we hardcode it to true in: https://github.com/openshift/hypershift/blob/main/control-plane-operator/controllers/hostedcontrolplane/v2/assets/cluster-autoscaler/deployment.yaml#L36

LiangquanLi930: I'll add the BalanceSimilarNodeGroups *bool field in a separate PR.

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1248,9 +1248,78 @@ const (
...
+// Configures when and how to scale down cluster nodes.
+type ScaleDownConfig struct {
+	// enabled determines whether the Cluster Autoscaler should scaled down this cluster.
+	//
+	// +kubebuilder:default=Enabled
+	// +kubebuilder:validation:Enum=Enabled;Disabled
+	// +optional
+	Enabled string `json:"enabled,omitempty"`
+
+	// delayAfterAddSeconds sets how long after scale up the scale down evaluation resumes in seconds
```
JoelSpeed: The limits should be documented within the godoc, users won't have access to the validations as they are written here

The default value should also be added in the doc

```
// It must be between 0 and 10000.
// When omitted, this means no opinion, and the platform is left to choose a reasonable default.
// The default value is 60.
```

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1248,9 +1248,78 @@ const (
...
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=10000
```
JoelSpeed: This is less than 3 hours, perhaps making it a 24 hours limit would be sensible? It would be 86400

api/hypershift/v1beta1/hostedcluster_types.go
JoelSpeed: What does it mean for this value to be 0? Is that actually valid? I know you responded already but we should document this as part of the godoc

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1248,9 +1248,78 @@ const (
...
+	// +kubebuilder:default=60
```
JoelSpeed: Do we really want to default this at the server level? Would you ever expect to change this default? If you think in the future we may want to adjust this value without having to go and update every pre-existing object, then not having a default here is preferable and defaulting within the controller itself. We do this a lot in configuration APIs in openshift

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1248,9 +1248,78 @@ const (
...
+	// +optional
+	DelayAfterAddSeconds *int32 `json:"delayAfterAddSeconds,omitempty"`
+
+	// delayAfterDelete sets how long after node deletion, scale down evaluation resumes, defaults to scan-interval
```
JoelSpeed: Is this one also a duration? So should be transformed to similar to the `*Seconds` examples?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1248,9 +1248,78 @@ const (
...
+	// +kubebuilder:validation:Pattern=([0-9]*(\.[0-9]*)?[a-z]+)+
+	// +kubebuilder:validation:MaxLength=100
+	// +optional
+	DelayAfterDelete *string `json:"delayAfterDelete,omitempty"`
+
+	// delayAfterFailure sets how long after a scale down failure, scale down evaluation resumes
```
JoelSpeed: As above, durations should not be used, transform to `*Seconds` pattern

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1248,9 +1248,78 @@ const (
...
+	// +kubebuilder:validation:Pattern=([0-9]*(\.[0-9]*)?[a-z]+)+
+	// +kubebuilder:validation:MaxLength=100
+	// +optional
+	DelayAfterFailure *string `json:"delayAfterFailure,omitempty"`
+
+	// unneededDurationSeconds establishes how long a node should be unneeded before it is eligible for scale down in seconds
+	// +kubebuilder:validation:Minimum=0
```
JoelSpeed, line 1301: What does it mean for this to be 0?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1248,9 +1248,78 @@ const (
...
+	// +kubebuilder:validation:Maximum=10000
```
JoelSpeed: Again, this is very low, I could imagine users wanting this much higher than 3 hours

api/hypershift/v1beta1/hostedcluster_types.go
JoelSpeed: If we hardcode it, then that's fine, lets not expose it

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1248,9 +1248,78 @@ const (
...
+	// +kubebuilder:default=600
+	// +optional
+	UnneededDurationSeconds *int32 `json:"unneededDurationSeconds,omitempty"`
+
+	// utilizationThreshold determines the node utilization level, defined as sum of requested resources divided by capacity, below which a node can be considered for scale down
+	// +kubebuilder:validation:Pattern=(0.[0-9]+)
+	// +kubebuilder:validation:MaxLength=100
+	// +optional
+	UtilizationThreshold *string `json:"utilizationThreshold,omitempty"`
```
JoelSpeed: Have you thought about converting this to a numeric respresentation of a percentage? An int32 with a range of 0-100 valid? That might be easier to express rather than expressing it a an decimal value

api/hypershift/v1beta1/hostedcluster_types.go
JoelSpeed: Should add something to the godoc along the lines of `When set to 0, this means...`

api/hypershift/v1beta1/hostedcluster_types.go
JoelSpeed: @enxebre any thoughts on this one?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1248,9 +1248,96 @@ const (
...
+	// delayAfterDeleteSeconds sets how long after node deletion, scale down evaluation resumes, defaults to scan-interval
+	// It must be between 0 and 86400.
+	// When omitted, this means no opinion, and the platform is left to choose a reasonable default.
+	// The default value is 60.
+	//
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=86400
+	// +kubebuilder:default=60
+	// +optional
+	DelayAfterDeleteSeconds *int32 `json:"delayAfterDeleteSeconds,omitempty"`
```
JoelSpeed, line 1285: Should explain in the godoc `When set to 0, this means...`

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1248,9 +1248,96 @@ const (
...
+	// delayAfterFailureSeconds sets how long after a scale down failure, scale down evaluation resumes
+	// It must be between 0 and 86400.
+	// When omitted, this means no opinion, and the platform is left to choose a reasonable default.
+	// The default value is 60.
+	//
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=86400
+	// +kubebuilder:default=60
+	// +optional
+	DelayAfterFailureSeconds *int32 `json:"delayAfterFailureSeconds,omitempty"`
```
JoelSpeed, line 1295: Should explain in the godoc `When set to 0, this means...`

api/hypershift/v1beta1/hostedcluster_types.go
JoelSpeed: Should explain in the godoc When set to 0, this means...

api/hypershift/v1beta1/hostedcluster_types.go
JoelSpeed: @enxebre thoughts on UX improvement suggestion?

api/hypershift/v1beta1/hostedcluster_types.go
LiangquanLi930: okay, remove default value.

LiangquanLi930:
For
https://github.com/openshift/hypershift/pull/6236#discussion_r2145470507
Since we need to expose `BalancingIgnoredLabels`, we should keep this API.

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1248,9 +1248,104 @@ const (
...
+	//
+	// +kubebuilder:default=Enabled
+	// +kubebuilder:validation:Enum=Enabled;Disabled
+	// +optional
+	Enabled string `json:"enabled,omitempty"`
+
+	// delayAfterAddSeconds sets how long after scale up the scale down evaluation resumes in seconds
+	// It must be between 0 and 86400.
+	// When set to 0, this means scale down evaluation will resume immediately after scale up, without any delay.
+	// When omitted, this means no opinion, and the platform is left to choose a reasonable default.
```
JoelSpeed:
```suggestion
	// When omitted, this means no opinion, and the platform is left to choose a reasonable default, which is subject to change over time.
	// The current default is ...
```

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1248,9 +1248,104 @@ const (
...
+	//
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=86400
+	// +optional
+	DelayAfterAddSeconds *int32 `json:"delayAfterAddSeconds,omitempty"`
+
+	// delayAfterDeleteSeconds sets how long after node deletion, scale down evaluation resumes, defaults to scan-interval
+	// It must be between 0 and 86400.
+	// When set to 0, this means scale down evaluation will resume immediately after node deletion, without any delay.
+	// When omitted, this means no opinion, and the platform is left to choose a reasonable default.
```
JoelSpeed:
```suggestion
	// When omitted, this means no opinion, and the platform is left to choose a reasonable default, which is subject to change over time.
```

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1248,9 +1248,104 @@ const (
...
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=86400
+	// +kubebuilder:default=60
+	// +optional
+	DelayAfterDeleteSeconds *int32 `json:"delayAfterDeleteSeconds,omitempty"`
+
+	// delayAfterFailureSeconds sets how long after a scale down failure, scale down evaluation resumes
+	// It must be between 0 and 86400.
+	// When set to 0, this means scale down evaluation will resume immediately after a scale down failure, without any delay.
+	// When omitted, this means no opinion, and the platform is left to choose a reasonable default.
```
JoelSpeed:
```suggestion
	// When omitted, this means no opinion, and the platform is left to choose a reasonable default, which is subject to change over time.
```

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1248,9 +1248,104 @@ const (
...
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=86400
+	// +kubebuilder:default=60
+	// +optional
+	DelayAfterFailureSeconds *int32 `json:"delayAfterFailureSeconds,omitempty"`
+
+	// unneededDurationSeconds establishes how long a node should be unneeded before it is eligible for scale down in seconds
+	// It must be between 0 and 86400.
+	// When set to 0, this means nodes will be considered for scale down immediately when they become unneeded, without any waiting period.
+	// When omitted, this means no opinion, and the platform is left to choose a reasonable default.
```
JoelSpeed:
```suggestion
	// When omitted, this means no opinion, and the platform is left to choose a reasonable default, which is subject to change over time.
```

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1248,9 +1248,104 @@ const (
...
+
+	// delayAfterDeleteSeconds sets how long after node deletion, scale down evaluation resumes, defaults to scan-interval
+	// It must be between 0 and 86400.
+	// When set to 0, this means scale down evaluation will resume immediately after node deletion, without any delay.
+	// When omitted, this means no opinion, and the platform is left to choose a reasonable default.
+	// The default value is 60.
+	//
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=86400
+	// +kubebuilder:default=60
```
JoelSpeed: I think we want to remove the defaults from all of these so that we can default them all controller side no?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1248,9 +1248,104 @@ const (
...
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=86400
+	// +kubebuilder:default=600
+	// +optional
+	UnneededDurationSeconds *int32 `json:"unneededDurationSeconds,omitempty"`
+
+	// utilizationThreshold determines the node utilization level, defined as sum of requested resources divided by capacity, below which a node can be considered for scale down
+	// The value represents a percentage from 0 to 100.
+	// When set to 0, this means nodes will be considered for scale down even if they are completely idle (0% utilization).
+	// When set to 100, this means nodes will only be considered for scale down when they are fully utilized (100% utilization).
```
JoelSpeed: I don't think these descriptions are correct.

IIRC, the utilization threshold is an upper bound for consideration, you've described it as a lower bound.

So a 0% means the node would have to be completely empty to be considered for scale down.

And 100% means that they will be considered for scale down always, even if they are 100% utilised.

A 50% threshold means they will be considered if their utilisation is up to, and including 50%, but a node using 51% would not be considered.

Right?

api/hypershift/v1beta1/hostedcluster_types.go
elmiko: if omitted, the autoscaler defaults to 10 minutes.

api/hypershift/v1beta1/hostedcluster_types.go
elmiko: if omitted, the autoscaler defaults to 0 seconds.

api/hypershift/v1beta1/hostedcluster_types.go
elmiko: if omitted, the autoscaler defaults to 3 minutes.

api/hypershift/v1beta1/hostedcluster_types.go
elmiko: if omitted, the autoscaler defaults to 10 minutes.

api/hypershift/v1beta1/hostedcluster_types.go
elmiko: i feel like setting this to 0 would produce some strange results. when the autoscaler decides a node is unneeded it will add a no schedule taint to the node. in the case of 0 second unneeded, it might put the taint on the node within 10 seconds of creation (one internal scan loop interval).

i don't think we need this in the godoc, just giving some context.

api/hypershift/v1beta1/hostedcluster_types.go
elmiko: correct, this is the threshold to determine when a node is considered "under-utilized". if the requests for pods on the node is below the threshold of the node's allocatable capacity, then it can be considered under-utilized.

when not specified, this defaults to "0.5" (eg 50%)

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1248,9 +1248,104 @@ const (
...
+	// A value between 0 and 100 represents the utilization threshold below which a node can be considered for scale down.
+	// When omitted, this means no opinion, and the platform is left to choose a reasonable default.
+	//
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=100
+	// +optional
+	UtilizationThreshold *int32 `json:"utilizationThreshold,omitempty"`
```
elmiko: the valid values for this are from `0.0` to `1.0`, with a default of `0.5`. if we follow the convention from the CAO, this should be a string not an int32.

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1285,6 +1380,20 @@ type ClusterAutoscaling struct {
...
 	// +optional
 	PodPriorityThreshold *int32 `json:"podPriorityThreshold,omitempty"`
+
+	// expanders guide the autoscaler in choosing node groups during scale-out.
+	// Sets the order of expanders for scaling out node groups.
+	// Options include:
+	// * LeastWaste - selects the group with minimal idle CPU and memory after scaling.
+	// * Priority - selects the group with the highest user-defined priority.
+	// * Random - selects a group randomly.
+	// If not specified, `Random` is the default.
```
elmiko: i doubled checked the cluster autoscaler, and it appears that "least waste" is the default expander now. i'm not sure if we are setting "random" somewhere, but just in case.

control-plane-operator/controllers/hostedcontrolplane/v2/autoscaler/deployment.go
```diff
@@ -11,34 +13,203 @@ import (
...
+	if sd.DelayAfterFailureSeconds != nil {
+		args = append(args, ScaleDownDelayAfterFailureArg.Value(fmt.Sprintf("%ds", *sd.DelayAfterFailureSeconds)))
+	}
+
+	if sd.UnneededDurationSeconds != nil {
+		args = append(args, ScaleDownUnneededTimeArg.Value(fmt.Sprintf("%ds", *sd.UnneededDurationSeconds)))
+	}
+
+	if sd.UtilizationThreshold != nil {
+		args = append(args, ScaleDownUtilizationThresholdArg.Value(fmt.Sprintf("%.2f", float64(*sd.UtilizationThreshold)/100.0)))
```
elmiko: i made a comment above about the api type for this field, if we decide to keep the "0-100" range and int type, this calculation should be fine.

api/hypershift/v1beta1/hostedcluster_types.go
LiangquanLi930: Okay, thanks for double-checking. I think we can change `Random` to `LeastWaste` to stay consistent with the autoscaler.

api/hypershift/v1beta1/hostedcluster_types.go
LiangquanLi930: @JoelSpeed What do you think? Should we stick with int32 or switch to string?

api/hypershift/v1beta1/hostedcluster_types.go
LiangquanLi930: done

control-plane-operator/controllers/hostedcontrolplane/v2/autoscaler/deployment.go
celebdor: Just matching what's per platform in Cluster Autoscaler Operator, but I suspect if we ever try to run on EKS it will be useful.

api/hypershift/v1beta1/hostedcluster_types.go
JoelSpeed: Floats in APIs are bad, floats embedded in strings are worse.

It was my suggestion to move to an int so that we could scale the value to be percentage. I think renaming the field to `utilizationThresholdPercent` would probably make this more obvious.

There's nothing to say we must follow the pattern of the upstream API if we believe we can provide something that maps to it that provides a better UX

LiangquanLi930: Thanks, change to UtilizationThresholdPercent

LiangquanLi930:
@JoelSpeed Thanks for your review. When you have time, could you help to review again?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1260,9 +1260,100 @@ const (
...
 type Provisioner string
 
+// Configures when and how to scale down cluster nodes.
+type ScaleDownConfig struct {
+	// enabled determines whether the Cluster Autoscaler should scaled down this cluster.
+	//
+	// +kubebuilder:default=Enabled
+	// +kubebuilder:validation:Enum=Enabled;Disabled
+	// +optional
+	Enabled string `json:"enabled,omitempty"`
```
gaol: https://github.com/openshift/cluster-autoscaler-operator/blob/51d0fd6eebcc1550ea85d8b4c446154f4763dd2d/pkg/apis/autoscaling/v1/clusterautoscaler_types.go#L164

Just my two cents: Shall we try to keep consistent with what is defined in openshift/cluster-autoscaler-operator ?

JoelSpeed: CAO was written a long time ago and would not pass today's API review, we prefer generally not to repeat previous mistakes when we now know that they are mistakes (OpenShift API review opinion)

JoelSpeed: This will stutter, `enabled: Enabled`, what other terms could we use?

```
scaleDown:
  scaleDown: Enabled | Disabled
```
Could work but also has a stutter

Would it make sense to add a field above, and validate that `scaleDown` is only set when it is set?

```
scaling: ScaleUpOnly | ScaleUpAndScaleDown 
scaleDown:
  ...
```
You could then use CEL to make sure `scaleDown` is omitted, when the `scaling` option is not `ScaleUpAndScaleDown`

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1260,9 +1260,100 @@ const (
...
+
+	// delayAfterAddSeconds sets how long after scale up the scale down evaluation resumes in seconds
```
JoelSpeed: Missing final period
```suggestion
	// delayAfterAddSeconds sets how long after scale up the scale down evaluation resumes in seconds.
```

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1260,9 +1260,100 @@ const (
...
+	// It must be between 0 and 86400.
+	// When set to 0, this means scale down evaluation will resume immediately after scale up, without any delay.
+	// When omitted, the autoscaler defaults to 600s (10 minutes).
+	//
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=86400
+	// +optional
+	DelayAfterAddSeconds *int32 `json:"delayAfterAddSeconds,omitempty"`
+
+	// delayAfterDeleteSeconds sets how long after node deletion, scale down evaluation resumes, defaults to scan-interval
```
JoelSpeed:
```suggestion
	// delayAfterDeleteSeconds sets how long after node deletion, scale down evaluation resumes, defaults to scan-interval.
```

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1260,9 +1260,100 @@ const (
...
+	// It must be between 0 and 86400.
+	// When set to 0, this means scale down evaluation will resume immediately after node deletion, without any delay.
+	// When omitted, the autoscaler defaults to 0s.
+	//
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=86400
+	// +optional
+	DelayAfterDeleteSeconds *int32 `json:"delayAfterDeleteSeconds,omitempty"`
+
+	// delayAfterFailureSeconds sets how long after a scale down failure, scale down evaluation resumes
```
JoelSpeed:
```suggestion
	// delayAfterFailureSeconds sets how long after a scale down failure, scale down evaluation resumes.
```

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1260,9 +1260,100 @@ const (
...
+	// It must be between 0 and 86400.
```
JoelSpeed: I wonder for each instance of this sentence, if it would make sense to say 
```suggestion
	// It must be between 0 and 86400 (24 hours).
```
That would be in keeping with how the defaults are written?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1260,9 +1260,100 @@ const (
...
+	// When set to 0, this means scale down evaluation will resume immediately after a scale down failure, without any delay.
+	// When omitted, the autoscaler defaults to 180s (3 minutes).
+	//
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=86400
+	// +kubebuilder:default=60
```
JoelSpeed: Remove this and let the autoscaler default an empty value?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1260,9 +1260,100 @@ const (
...
+	// +optional
+	DelayAfterFailureSeconds *int32 `json:"delayAfterFailureSeconds,omitempty"`
+
+	// unneededDurationSeconds establishes how long a node should be unneeded before it is eligible for scale down in seconds
```
JoelSpeed:
```suggestion
	// unneededDurationSeconds establishes how long a node should be unneeded before it is eligible for scale down in seconds.
```

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1260,9 +1260,100 @@ const (
...
+	// It must be between 0 and 86400.
+	// When omitted, the autoscaler defaults to 600s (10 minutes).
+	// The default value is 600.
+	//
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=86400
+	// +kubebuilder:default=600
```
JoelSpeed: Is this the autoscaler default? Do we need to specify the default in the API?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1260,9 +1260,100 @@ const (
...
+	// +optional
+	UnneededDurationSeconds *int32 `json:"unneededDurationSeconds,omitempty"`
+
+	// utilizationThresholdPercent determines the node utilization level, defined as sum of requested resources divided by capacity, below which a node can be considered for scale down
```
JoelSpeed:
```suggestion
	// utilizationThresholdPercent determines the node utilization level, defined as sum of requested resources divided by capacity, below which a node can be considered for scale down.
```

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1260,9 +1260,100 @@ const (
...
+	ScaleDown *ScaleDownConfig `json:"scaleDown,omitempty"`
+
+	// balancingIgnoredLabels sets "--balancing-ignore-label <label name>" flag on cluster-autoscaler for each listed label.
+	// This option specifies labels that cluster autoscaler should ignore when considering node group similarity.
+	// For example, if you have nodes with "topology.ebs.csi.aws.com/zone" label, you can add name of this label here
+	// to prevent cluster autoscaler from splitting nodes into different node groups based on its value.
+	//
+	//
+	// +kubebuilder:validation:MaxItems=64
+	// +kubebuilder:validation:items:MaxLength=316
```
JoelSpeed: I think this is 317 not 316? 253 for the subdomain, 1 for the `/`, and 63 for the label after the slash, makes 253 + 1 + 63 = 317?

api/hypershift/v1beta1/hostedcluster_types.go
LiangquanLi930: okay, change to `ClusterAutoscaling.Scaling`

api/hypershift/v1beta1/hostedcluster_types.go
LiangquanLi930: @elmiko What is the default value used by the autoscaler when this is omitted?

api/hypershift/v1beta1/hostedcluster_types.go
LiangquanLi930: @elmiko What is the default value used by the autoscaler when this is omitted?

LiangquanLi930: 600 is the default value for the autoscaler, so we can remove it.

api/hypershift/v1beta1/hostedcluster_types.go
LiangquanLi930: updated

LiangquanLi930:
@JoelSpeed Thanks for your review. When you have time, could you help to review again?

control-plane-operator/controllers/hostedcontrolplane/v2/autoscaler/deployment.go
```diff
@@ -11,34 +13,205 @@ import (
...
+	return args
+}
+
+func adaptDeployment(cpContext component.WorkloadContext, deployment *appsv1.Deployment) error {
+	hcp := cpContext.HCP
+
+	util.UpdateContainer(ComponentName, deployment.Spec.Template.Spec.Containers, func(c *corev1.Container) {
+		// TODO if the options for the cluster autoscaler continues to grow, we should take inspiration
+		// from the cluster-autoscaler-operator and create some utility functions for these assignments.
+		c.Args = append(c.Args, autoscalerArgs(&hcp.Spec.Autoscaling, hcp.Spec.Platform.Type, ctrl.LoggerFrom(cpContext))...)
```
gaol, line 215: `c.Args` may have already some predefined arguments defined in the deployment.yaml file, I think we need to merge the args to avoid duplication. like in my pre-merge test, I see:
```
  - args:
    - --expander=priority,least-waste
    - --cloud-provider=clusterapi
    - --node-group-auto-discovery=clusterapi:namespace=$(MY_NAMESPACE)
    - --kubeconfig=/mnt/kubeconfig/target-kubeconfig
    - --clusterapi-cloud-config-authoritative
    - --skip-nodes-with-local-storage=false
    - --alsologtostderr
    - --leader-elect-lease-duration=137s
    - --leader-elect-retry-period=26s
    - --leader-elect-renew-deadline=107s
    - --balance-similar-node-groups=true
    - --v=4
    - --balancing-ignore-label=hypershift.openshift.io/nodePool
    - --balancing-ignore-label=vpc-block-csi-driver-labels
    - --scale-down-enabled=true
    - --scale-down-delay-after-add=60s
    - --scale-down-delay-after-delete=60s
    - --scale-down-delay-after-failure=60s
    - --scale-down-unneeded-time=60s
    - --scale-down-utilization-threshold=0.50
    - --max-graceful-termination-sec=60
    - --balancing-ignore-label=topology.ebs.csi.aws.com/zone
    - --expander=priority,least-waste
    - --balancing-ignore-label=topology.ebs.csi.aws.com/zone
    - --balancing-ignore-label=lifecycle
    - --balancing-ignore-label=k8s.amazonaws.com/eniConfig
    - --balancing-ignore-label=eks.amazonaws.com/nodegroup
    - --balancing-ignore-label=alpha.eksctl.io/nodegroup-name
    - --balancing-ignore-label=alpha.eksctl.io/instance-id
    - --balancing-ignore-label=topology.k8s.aws/zone-id
```

If I update the hosted cluster with patch:
```json
{
		"scaling": "ScaleUpAndScaleDown",
			"maxPodGracePeriod": 60,
			"scaleDown": {
			    "enabled": "Enabled",
				"delayAfterAddSeconds": 60,
				"delayAfterDeleteSeconds": 60,
				"delayAfterFailureSeconds": 60,
				"unneededDurationSeconds": 60,
				"utilizationThresholdPercent": 50
			},
			"balancingIgnoredLabels": ["topology.ebs.csi.aws.com/zone"],
			"expanders": ["Priority", "LeastWaste"]
		}
```

@LiangquanLi930 FYI ^^^

LiangquanLi930: Thanks, i removed
```
- --expander=priority,least-waste
```

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1260,9 +1260,101 @@ const (
...
+
 // ClusterAutoscaling specifies auto-scaling behavior that applies to all
 // NodePools associated with a control plane.
 type ClusterAutoscaling struct {
+	// scaling defines the scaling behavior for the cluster autoscaler.
+	// ScaleUpOnly means the autoscaler will only scale up nodes, never scale down.
+	// ScaleUpAndScaleDown means the autoscaler will both scale up and scale down nodes.
+	// When set to ScaleUpAndScaleDown, the scaleDown field can be used to configure scale down behavior.
+	//
+	// +kubebuilder:default=ScaleUpAndScaleDown
```
JoelSpeed, line 1353: Is this the correct default?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1260,9 +1260,101 @@ const (
...
+	// +kubebuilder:validation:Enum=ScaleUpOnly;ScaleUpAndScaleDown
+	// +optional
+	Scaling string `json:"scaling,omitempty"`
+
+	// scaleDown configures the behavior of the Cluster Autoscaler scale down operation.
+	// This field is only valid when scaling is set to ScaleUpAndScaleDown.
+	//
+	// +optional
+	ScaleDown *ScaleDownConfig `json:"scaleDown,omitempty"`
```
JoelSpeed, line 1361: Can we use validation to make sure this field is not set when scaling is not `ScaleUpAndScaleDown`? Are we able to use CEL on this API (@enxebre ?), if we are, we have a way to check this dependency quickly

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1260,9 +1260,101 @@ const (
...
+
+	// balancingIgnoredLabels sets "--balancing-ignore-label <label name>" flag on cluster-autoscaler for each listed label.
+	// This option specifies labels that cluster autoscaler should ignore when considering node group similarity.
+	// For example, if you have nodes with "topology.ebs.csi.aws.com/zone" label, you can add name of this label here
+	// to prevent cluster autoscaler from splitting nodes into different node groups based on its value.
+	//
+	//
+	// +kubebuilder:validation:MaxItems=64
+	// +kubebuilder:validation:items:MaxLength=317
```
JoelSpeed, line 1380: Should we add validation to the ignored labels list item to say that this must be a valid label key? We might be able to use CEL, else a regex could be used

api/hypershift/v1beta1/hostedcluster_types.go
LiangquanLi930: Yes, ScaleUpAndScaleDown is correct default.

api/hypershift/v1beta1/hostedcluster_types.go
enxebre: yes, the main constraint is that this addition don't impact existing clusters behaviour in any way. That'd be the case for when Scaling empty

api/hypershift/v1beta1/hostedcluster_types.go
enxebre: the default behaviour for this api needs to be the one for already running clusters, which is scale up and down. I think the question is whether we want to convey that by defaulting here or by letting empty be it

JoelSpeed: Indeed, you could have this omitted completely (which means no change from a readers perspective after this change).

This would in theory allow you to change the default under the hood in the future if our opinions of best practice changed.

Those that had set an explicit value would of course not be affected.

I could however see an argument that this _should_ be defaulted in this way, since we don't want to change the scaling behaviour of an existing cluster maybe?

api/hypershift/v1beta1/hostedcluster_types.go
JoelSpeed: In which case we should use this https://github.com/openshift/api/blob/41708f6123394c18afbb4bef1bac526f46960769/example/v1/types_stable.go#L136

Prevent the `scaleDown` from being set when `scaling != ScaleUpAndScaleDown`, but don't require it to be set

api/hypershift/v1beta1/hostedcluster_types.go
LiangquanLi930: I think we should keep the explicit `+kubebuilder:default=ScaleUpAndScaleDown` for clarity and consistency. 

The current default behavior for existing clusters is scale up and down, and making this explicit in the API (rather than hiding it in implementation) provides better user experience and documentation. It also ensures forward compatibility for existing clusters.

api/hypershift/v1beta1/hostedcluster_types.go
LiangquanLi930: updated

api/hypershift/v1beta1/hostedcluster_types.go
LiangquanLi930: updated.

LiangquanLi930:
@JoelSpeed When you have time, Could you help to review again?

api/hypershift/v1beta1/hostedcluster_types.go
LiangquanLi930: updated

api/hypershift/v1beta1/hostedcluster_types.go
LiangquanLi930: updated

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1260,9 +1260,103 @@ const (
...
 type ClusterAutoscaling struct {
+	// scaling defines the scaling behavior for the cluster autoscaler.
+	// ScaleUpOnly means the autoscaler will only scale up nodes, never scale down.
+	// ScaleUpAndScaleDown means the autoscaler will both scale up and scale down nodes.
+	// When set to ScaleUpAndScaleDown, the scaleDown field can be used to configure scale down behavior.
+	//
+	// +kubebuilder:default=ScaleUpAndScaleDown
+	// +kubebuilder:validation:Enum=ScaleUpOnly;ScaleUpAndScaleDown
+	// +optional
+	Scaling string `json:"scaling,omitempty"`
```
JoelSpeed: Not a must, but we generally recommend to create a string alias type and then constants for the different possible values, and use that instead of a plain untyped string

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1260,9 +1260,103 @@ const (
...
+
+	// balancingIgnoredLabels sets "--balancing-ignore-label <label name>" flag on cluster-autoscaler for each listed label.
+	// This option specifies labels that cluster autoscaler should ignore when considering node group similarity.
+	// For example, if you have nodes with "topology.ebs.csi.aws.com/zone" label, you can add name of this label here
+	// to prevent cluster autoscaler from splitting nodes into different node groups based on its value.
+	//
+	//
+	// +kubebuilder:validation:MaxItems=64
+	// +kubebuilder:validation:items:MaxLength=317
+	// +kubebuilder:validation:XValidation:rule="self.all(l, l.matches('^([a-z0-9.-]{1,253}/)?[a-zA-Z0-9]([a-zA-Z0-9_.-]{0,61}[a-zA-Z0-9])?$'))", message="Each balancingIgnoredLabels item must be a valid label key"
```
JoelSpeed: I'm not sure this regex is correct, where did you get the first part `[a-z0-9.-]{1,253}`?
The correct regex for this should be more similar to the second half, as it is constructed of `.` separated labels segments

Are we sure underscores are permitted in the second part?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1297,6 +1404,20 @@ type ClusterAutoscaling struct {
...
 	// +optional
 	PodPriorityThreshold *int32 `json:"podPriorityThreshold,omitempty"`
+
+	// expanders guide the autoscaler in choosing node groups during scale-out.
+	// Sets the order of expanders for scaling out node groups.
+	// Options include:
+	// * LeastWaste - selects the group with minimal idle CPU and memory after scaling.
+	// * Priority - selects the group with the highest user-defined priority.
+	// * Random - selects a group randomly.
+	// If not specified, `LeastWaste` is the default.
```
JoelSpeed: Nit, the default should be represented a list, not a single string IMO
```suggestion
	// If not specified, `[LeastWaste]` is the default.
```

JoelSpeed (commented):
Apart from a couple of small bits and fixing the regex, I think we are good here from an API review perspective

api/hypershift/v1beta1/hostedcluster_types.go
LiangquanLi930: updated

LiangquanLi930: IIUC, the official validation allows it, as shown here:
https://github.com/kubernetes/apimachinery/blob/master/pkg/util/validation/validation.go#L175-L177

api/hypershift/v1beta1/hostedcluster_types.go
LiangquanLi930: update

LiangquanLi930:
> i have a question about the changes in `control-plane-operator/controllers/hostedcontrolplane/v2/autoscaler/deployment.go`, it seems like many of the arguments (`AutoscalerArg`) and much of the code is copied directly from the cluster-autoscaler-operator. is there a reason we don't just import the CAO module so that we can avoid duplicating the code?
> 
> i would love to see if there is a way we can share the code, even if that means we need changes on the CAO.

I agree. We should focus on ensuring the autoscaler functionality works properly and meets HyperShift's requirements first. We can address code reuse and vendoring in a follow-up PR

JoelSpeed:
LGTM from an API review perspective

elmiko:
> I agree. We should focus on ensuring the autoscaler functionality works properly and meets HyperShift's requirements first. We can address code reuse and vendoring in a follow-up PR

i think that's fair, i just don't want us to forget.

api/hypershift/v1beta1/hostedcluster_types.go
gaol: > the default behaviour for this api needs to be the one for already running clusters, which is scale up and down

@enxebre I am little confused by this comment. Is the scaling down behavior determined by the arg: `--scale-down-enabled`  in the `cluster-autoscaler` pod ?  or hypershift has it's own logic besides of this argument ? 

Now before this feature gets implemented, as long as the  `autoScaling`  in the nodepool gets enabled  without setting `autoscaling` in the hosted cluster, the `cluster-autoscaler` pod will be created with the default arguments, where I don't see the scaling down argument.

control-plane-operator/controllers/hostedcontrolplane/v2/assets/cluster-autoscaler/deployment.yaml
```diff
@@ -23,7 +23,6 @@ spec:
       automountServiceAccountToken: true
       containers:
       - args:
-        - --expander=priority,least-waste
```
gaol, line 26: Will the change of the default arguments lead to behavior change by default ? 

Like before this feature implemented, when users enable the `autoScaling` in nodepool only, the `cluster-autoscaler` pod has `--expander=priority,least-waste` specified, but after this feature, the same action will lead to the pod without having this expander arguments, is it acceptable change?

@enxebre @LiangquanLi930

control-plane-operator/controllers/hostedcontrolplane/v2/assets/cluster-autoscaler/deployment.yaml
```diff
@@ -36,9 +35,6 @@ spec:
         - --balance-similar-node-groups=true
         - --v=4
         - --balancing-ignore-label=hypershift.openshift.io/nodePool
-        - --balancing-ignore-label=topology.ebs.csi.aws.com/zone
-        - --balancing-ignore-label=topology.disk.csi.azure.com/zone
-        - --balancing-ignore-label=ibm-cloud.kubernetes.io/worker-id
```
gaol, line 41: The same concern as above about the arguments change

control-plane-operator/controllers/hostedcontrolplane/v2/autoscaler/deployment.go
```diff
@@ -11,34 +13,214 @@ import (
...
+
+	if options.PodPriorityThreshold != nil {
+		args = append(args, ExpendablePodsPriorityCutoffArg.Value(*options.PodPriorityThreshold))
+	}
+
+	for _, ignoredLabel := range options.BalancingIgnoredLabels {
+		args = append(args, BalancingIgnoreLabelArg.Value(ignoredLabel))
+	}
+
+	if len(options.Expanders) > 0 {
```
muraee, line 124: I think we should keep the current default `priority,least-waste` if `Expanders` field is not set (empty)

control-plane-operator/controllers/hostedcontrolplane/v2/assets/cluster-autoscaler/deployment.yaml
LiangquanLi930: updated https://github.com/openshift/hypershift/compare/0888d61647f1bd1e994e066b9a4e6e621c7ae232..1c0b8013c96c6dbe00ab7aa3e148f785e23c122f

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1260,9 +1260,114 @@ const (
...
+	// ScaleUpOnly means the autoscaler will only scale up nodes, never scale down.
+	ScaleUpOnly ScalingType = "ScaleUpOnly"
+
+	// ScaleUpAndScaleDown means the autoscaler will both scale up and scale down nodes.
+	ScaleUpAndScaleDown ScalingType = "ScaleUpAndScaleDown"
+)
+
 // ClusterAutoscaling specifies auto-scaling behavior that applies to all
 // NodePools associated with a control plane.
+// +kubebuilder:validation:XValidation:rule="self.scaling == 'ScaleUpAndScaleDown' ? true : !has(self.scaleDown)",message="scaleDown can only be set when scaling is ScaleUpAndScaleDown"
```
enxebre, line 1344: can we please add a case to TestOnCreateAPIUX to validate this cel rule.

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1260,9 +1260,114 @@ const (
...
 type ClusterAutoscaling struct {
+	// scaling defines the scaling behavior for the cluster autoscaler.
+	// ScaleUpOnly means the autoscaler will only scale up nodes, never scale down.
+	// ScaleUpAndScaleDown means the autoscaler will both scale up and scale down nodes.
+	// When set to ScaleUpAndScaleDown, the scaleDown field can be used to configure scale down behavior.
+	//
+	// +kubebuilder:default=ScaleUpAndScaleDown
+	// +optional
+	Scaling ScalingType `json:"scaling,omitempty"`
```
enxebre, line 1355: can we please document that this fields would only impact >= 4.20

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1260,9 +1260,114 @@ const (
...
+
+	// balancingIgnoredLabels sets "--balancing-ignore-label <label name>" flag on cluster-autoscaler for each listed label.
+	// This option specifies labels that cluster autoscaler should ignore when considering node group similarity.
+	// For example, if you have nodes with "topology.ebs.csi.aws.com/zone" label, you can add name of this label here
+	// to prevent cluster autoscaler from splitting nodes into different node groups based on its value.
+	//
+	//
+	// +kubebuilder:validation:MaxItems=64
+	// +kubebuilder:validation:items:MaxLength=317
+	// +kubebuilder:validation:XValidation:rule="self.all(l, l.matches('^([a-z0-9]([-a-z0-9]*[a-z0-9])?(\\\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*/)?[a-zA-Z0-9]([a-zA-Z0-9_.-]{0,61}[a-zA-Z0-9])?$'))", message="Each balancingIgnoredLabels item must be a valid label key"
```
enxebre, line 1381: please include a case in TestOnCreateAPIUX for failing this

control-plane-operator/controllers/hostedcontrolplane/v2/autoscaler/deployment.go
enxebre: these are labels on the guest cluster Nodes, I don't think we need them at all.
Let's drop them unless there's a reason to keep them

control-plane-operator/controllers/hostedcontrolplane/v2/autoscaler/component_test.go
```diff
@@ -115,12 +115,88 @@ func TestAdaptDeployment(t *testing.T) {
...
+				"--balancing-ignore-label=custom.label/zone",
+				"--balancing-ignore-label=custom.label/instance-type",
+			},
+			expectedReplicas: 1,
+		},
+		{
+			name: "when MaxFreeDifferenceRatioPercent is set, container has max-free-difference-ratio argument",
+			AutoscalerOptions: hyperv1.ClusterAutoscaling{
+				MaxFreeDifferenceRatioPercent: ptr.To[int32](20),
+			},
```
enxebre, line 197: Please add a unit test to validate reconciling an existing deployment with the args it has today, it remains immutable after reconciling.
You could also add an e2e for this in TestUpgradeControlPlane

control-plane-operator/controllers/hostedcontrolplane/testdata/cluster-autoscaler/zz_fixture_TestControlPlaneComponents_cluster_autoscaler_deployment.yaml
```diff
@@ -78,10 +77,14 @@ spec:
         - --balance-similar-node-groups=true
         - --v=4
         - --balancing-ignore-label=hypershift.openshift.io/nodePool
-        - --balancing-ignore-label=topology.ebs.csi.aws.com/zone
-        - --balancing-ignore-label=topology.disk.csi.azure.com/zone
-        - --balancing-ignore-label=ibm-cloud.kubernetes.io/worker-id
         - --balancing-ignore-label=vpc-block-csi-driver-labels
+        - --balancing-ignore-label=topology.ebs.csi.aws.com/zone
```
enxebre, line 81: This seems wrong. This test output should remain immutable, no?

control-plane-operator/controllers/hostedcontrolplane/testdata/cluster-autoscaler/zz_fixture_TestControlPlaneComponents_cluster_autoscaler_deployment.yaml
```diff
@@ -65,7 +65,6 @@ spec:
       automountServiceAccountToken: true
       containers:
       - args:
```
enxebre, line 67: This seems wrong. This test output should remain immutable, no?

control-plane-operator/controllers/hostedcontrolplane/v2/autoscaler/component_test.go
enxebre: We have this one indeed https://github.com/openshift/hypershift/pull/6236/files#r2185520238

api/hypershift/v1beta1/hostedcluster_types.go
LiangquanLi930: updated.

api/hypershift/v1beta1/hostedcluster_types.go
LiangquanLi930: updated.

api/hypershift/v1beta1/hostedcluster_types.go
LiangquanLi930: updated

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1260,9 +1260,116 @@ const (
...
+	Scaling ScalingType `json:"scaling,omitempty"`
+
+	// scaleDown configures the behavior of the Cluster Autoscaler scale down operation.
+	// This field is only valid when scaling is set to ScaleUpAndScaleDown.
+	//
+	// +optional
+	ScaleDown *ScaleDownConfig `json:"scaleDown,omitempty"`
+
+	// balancingIgnoredLabels sets "--balancing-ignore-label <label name>" flag on cluster-autoscaler for each listed label.
+	// This option specifies labels that cluster autoscaler should ignore when considering node group similarity.
```
enxebre, line 1364: please document our backend choosen defaults

LiangquanLi930: updated

control-plane-operator/controllers/hostedcontrolplane/testdata/cluster-autoscaler/zz_fixture_TestControlPlaneComponents_cluster_autoscaler_deployment.yaml
LiangquanLi930: The flag --expander=priority,least-waste remains unchanged; it was just reordered.

test/e2e/autoscaling_test.go
```diff
@@ -25,8 +27,35 @@ func TestAutoscaling(t *testing.T) {
 	ctx, cancel := context.WithCancel(testContext)
 	defer cancel()
 
+	e2eutil.AtLeast(t, e2eutil.Version420)
```
enxebre: why are we removing e2e test coverage for any version lower than 4.20?
If anything please scope down only new validations for >= 4.20

test/e2e/autoscaling_test.go
```diff
@@ -70,19 +176,55 @@ func TestAutoscaling(t *testing.T) {
 		g.Expect(ok).Should(BeTrue())
 
 		// Enforce max nodes creation.
-		// 60% - enough that the existing and new nodes will
+		// 50% - enough that the existing and new nodes will
 		// be used, not enough to have more than 1 pod per
 		// node.
 		workloadMemRequest := resource.MustParse(fmt.Sprintf("%v", 0.5*float32(bytes)))
-		workload := newWorkLoad(max, workloadMemRequest, "", globalOpts.LatestReleaseImage)
+		expectNodes := min*2 + 2 // default nodepool(min + 1) + additional nodepool(min + 1)
```
enxebre: why is any change to this test needed?

control-plane-operator/controllers/hostedcontrolplane/v2/autoscaler/deployment.go
```diff
@@ -11,34 +13,189 @@ import (
...
+		args = append(args, ExpanderArg.Value(strings.Join(expanders, ",")))
+	} else {
+		args = append(args, ExpanderArg.Value(strings.Join([]string{priorityFlag, leastWasteFlag}, ",")))
+	}
 
-		if options.PodPriorityThreshold != nil {
-			c.Args = append(c.Args, fmt.Sprintf("--expendable-pods-priority-cutoff=%d", *options.PodPriorityThreshold))
+	// Since we hardcode "balance-similar-node-groups" to true in the deployment yaml
+	// Append basic ignore labels for a specific cloud provider.
+	args = appendBasicIgnoreLabels(args, platformType)
```
enxebre, line 147: can we please handle all labels here programmatically consistently, i.e move the ones hardcoded in the manifest here

enxebre: also, when the value is set we should only append the ones belonging to hypershift.io, and clobber the rest assuming that's also what autoscaler does

LiangquanLi930: > can we please handle all labels here programmatically consistently, i.e move the ones hardcoded in the manifest here

updated. 

> also, when the value is set we should only append the ones belonging to hypershift.io, and clobber the rest assuming that's also what autoscaler does

Actually, Cluster Autoscaler Operator in OpenShift uses a merge-based strategy: https://github.com/openshift/cluster-autoscaler-operator/blob/08844fa4a049de944e8c914236f7ec209b056cfa/pkg/controller/clusterautoscaler/clusterautoscaler.go#L235-L252
+ It always appends cloud-specific default labels (via appendBasicIgnoreLabels),
+ And then it appends user-specified BalancingIgnoredLabels.

So it does not clobber, but adds.
IMHO, we may want to keep behavior consistent.

enxebre: ok, let's append cloud default ones as well

test/e2e/autoscaling_test.go
```diff
@@ -27,7 +28,50 @@ func TestAutoscaling(t *testing.T) {
...
+	clusterOpts.NodePoolReplicas = 1
+	var additionalNP *hyperv1.NodePool
+
+	clusterOpts.BeforeApply = func(obj crclient.Object) {
+		if nodepool, ok := obj.(*hyperv1.NodePool); ok {
+			nodepool.Spec.NodeLabels = map[string]string{
+				"custom.ignore.label": "test1",
+			}
+
+			if additionalNP == nil {
```
enxebre, line 40: fwiw you can use BuildNodePoolManifest

LiangquanLi930: Thanks, I will open another PR to take it

PR #6170, opened 2025-05-19
OCPBUGS-56458: switch termination message policy to TerminationMessageFallbackToLogsOnError for debugability

sjenning:
manual backport of https://github.com/openshift/hypershift/pull/6122 without HO/CLI/e2e changes

PR #6245, opened 2025-06-09
OCPBUGS-56930: cpov1: set terminationMessagePolicy to FallbackToLogsOnError for all HCP containers

sjenning:
https://github.com/openshift/hypershift/pull/6170 did not actually work as the codepath is only called on updates, not creates.

This sets the `terminationMessagePolicy` in each CPOv1 component individually so that it is set at create time.

This should also fix the regression on our upgrade job.


sjenning:
hold until 4.19 GA build

sjenning:
4.19 GA build is done
/hold cancel

PR #5972, opened 2025-04-03
HOSTEDCP-2059: Support capacity reservation in NodePool API 

muraee:
**What this PR does / why we need it**:
Adds new field to the NodePool AWS platform API `.placement.capacityReservation` to allow specifying a capacity reservation for the EC2 instances to be launched into.
see: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/capacity-reservation-overview.html

**Which issue(s) this PR fixes** *(optional, use `fixes #<issue_number>(, fixes #<issue_number>, ...)` format, where issue_number might be a GitHub issue, or a Jira story*:
Fixes #

**Checklist**
- [ ] Subject and description added to both, commit and PR.
- [ ] Relevant issues have been referenced.
- [ ] This change includes docs. 
- [ ] This change includes unit tests.

PR #6248, opened 2025-06-10
OCPBUGS-56861: Remove CapacityReservation validation that requires aws access

muraee:
The HO aws creds are usually different from the workload account  where capacity reservation is created, which means it can't perform those validations.

Fixes #

**Checklist**


sjenning:
reviewer note: introduced in https://github.com/openshift/hypershift/pull/5972

PR #6249, opened 2025-06-10
OCPBUGS-49812: Sanitize PowerVS image name to comply with updated naming standards

Neha-dot-Yadav:

Recent changes in IBM PowerVS have introduced stricter validation on image names, disallowing the use of dots (.)
Attempting to import an image with a name like` 418.94.202501221327-0` results in the following error:
`Bad Request: 418.94.202501221327-0 is not a valid image name. the image name does not meet the resource naming standards.`
Due to this PowerVS Hypershift deployments are failing.

This PR resolves the above mentioned issue by replacing all dots with hyphens (-) in PowerVS image name.

Fixes # https://issues.redhat.com/browse/OCPBUGS-49812

**Checklist**

### DIFF
--- a/hypershift-operator/controllers/nodepool/powervs.go
+++ b/hypershift-operator/controllers/nodepool/powervs.go
@@ -3,6 +3,7 @@ package nodepool
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
 	"github.com/openshift/hypershift/support/releaseinfo"
@@ -119,6 +120,9 @@ func getPowerVSImage(region string, releaseImage *releaseinfo.ReleaseImage) (*re
 	if !hasRegionData {
 		return nil, "", fmt.Errorf("couldn't find PowerVS image for region %q", COSRegion)
 	}
+	// PowerVS now enforces stricter validation rules on image names, disallowing dots.
+	// To ensure compatibility, transform the release name by replacing dots with hyphens.
+	regionData.Release = strings.ReplaceAll(regionData.Release, ".", "-")
 	return &regionData, COSRegion, nil
 }
 


hypershift-operator/controllers/nodepool/powervs.go
```diff
@@ -52,7 +53,7 @@ func ibmPowerVSMachineTemplateSpec(hcluster *hyperv1.HostedCluster, nodePool *hy
 	if err != nil {
 		return nil, fmt.Errorf("couldn't discover a PowerVS Image for release image: %w", err)
 	}
-	powerVSBootImage := coreOSPowerVSImage.Release
+	powerVSBootImage := strings.ReplaceAll(coreOSPowerVSImage.Release, ".", "-")
```
srivastav-abhishek: I'd suggest to modify at place where `coreOSPowerVSImage.Release` is getting populated. That way we don't have to replace it with hyphen(-) every time it's used.



Applicable for all.

Neha-dot-Yadav: Updated it.

srivastav-abhishek (approved):
/lgtm

hypershift-operator/controllers/nodepool/powervs.go
```diff
@@ -98,6 +99,9 @@ func getPowerVSImage(region string, releaseImage *releaseinfo.ReleaseImage) (*re
 	if !hasRegionData {
 		return nil, "", fmt.Errorf("couldn't find PowerVS image for region %q", COSRegion)
 	}
+	// PowerVS now enforces stricter validation rules on image names, disallowing dots.
```
enxebre, line 123: how/where does PowerVS enforce this exactly?
This input is coming from the release payload itself

Neha-dot-Yadav: CoreOS images in the IBM Cloud are hosted in the IBM Cloud Object Storage for PowerVS platform, these images needs to be imported into the PowerVS service instance needed for the machines.
IBMPowerVSImage is the spec controlled by the CAPIBM to import these images and used in the machine deployments.

Image import into the PowerVS service instance is the one failing due to new validation rules being set on the PowerVS service instance side.
Screen shot with the validation rules message :-

image

PR #6253, opened 2025-06-11
NO-JIRA: Add misspell to golangci-lint

bryan-cox:
This PR adds misspell linter to the list of enabled golanci-lint linters and fixes all the spelling mistakes.

N/A

**Checklist**


bryan-cox:
Runs in golangci lint which runs in verify

PR #6256, opened 2025-06-12
CNTRLPLANE-1020: Global PullSecret daemonSet implementation

jparrill:
- Global Pull Secret DaemonSet approach

This Pull Request introduces a mechanism that enables the _Hosted Cluster Admin persona_ to include a PullSecret for accessing container images from private registries, without requiring assistance or interaction with the _Management Cluster Administrator_ persona.

- DaemonSet, Roles, Secrets are deployed on additionalPullSecret secret creation on DataPlane: `additional-pull-secret` into `kube-system` namespace.
- Implemented a GlobalPullSecret kubernetes controller:
  - It uses the same manager as HCCO
  - Generated a different controller to avoid short circuit other controllers
  - Created informer to act on secret events during the controller reconciliation
  - Add informer to the controller watching secrets to check on additionalPullSecret
  - Add early deletion and creation during the GlobalPullSecret reconciliation
  - Removed deletion of RBAC and DaemonSet logic (rabbit hole).
- Implementation of a kubernetes controller in the global-pull-secret-syncer
  - The controller reacts to Creation, Modification and Deletion of secrets in `kube-system` namespace
- Kubelet is restarted via DBus
- Add Documentation for GlobalPullSecret
- Add E2E tests for GlobalPullSecret
- Add E2EReadOnlyPullSecret to gitleaks

- Fixes [CNTRLPLANE-1020](https://issues.redhat.com/browse/CNTRLPLANE-1020)

## Additional details
- Documentation rendered at: https://687f4c85578ee500083caa90--hypershift-docs.netlify.app/how-to/common/global-pull-secret/
- Implementation workflow diagram
```mermaid
graph TB
    %% User Input
    User[User creates additional-pull-secret] --> |kube-system namespace| AdditionalPS[additional-pull-secret Secret]
    
    %% HCCO Controller
    HCCO[Hosted Cluster Config Operator] --> |Watches kube-system secrets| GlobalPSController[Global Pull Secret Controller]
    GlobalPSController --> |Validates| AdditionalPS
    GlobalPSController --> |Gets original| OriginalPS[Original pull-secret from HCP]
    
    %% Secret Processing
    AdditionalPS --> |Validates format| ValidatePS[Validate Additional Pull Secret]
    OriginalPS --> |Extracts data| OriginalPSData[Original Pull Secret Data]
    ValidatePS --> |Extracts data| AdditionalPSData[Additional Pull Secret Data]
    
    %% Merge Process
    OriginalPSData --> MergeSecrets[Merge Pull Secrets]
    AdditionalPSData --> MergeSecrets
    MergeSecrets --> |Creates merged JSON| GlobalPSData[Global Pull Secret Data]
    
    %% Secret Creation
    GlobalPSData --> |Creates in kube-system| GlobalPSSecret[global-pull-secret Secret]
    
    %% RBAC Setup
    GlobalPSController --> |Creates RBAC| RBACSetup[Setup RBAC Resources]
    RBACSetup --> ServiceAccount[global-pull-secret-syncer ServiceAccount]
    RBACSetup --> ClusterRole[global-pull-secret-syncer ClusterRole]
    RBACSetup --> ClusterRoleBinding[global-pull-secret-syncer ClusterRoleBinding]
    
    %% DaemonSet Deployment
    GlobalPSController --> |Deploys DaemonSet| DaemonSet[global-pull-secret-syncer DaemonSet]
    DaemonSet --> |Runs on each node| DaemonSetPod[DaemonSet Pod]
    
    %% DaemonSet Pod Details
    DaemonSetPod --> |Mounts host paths| HostMounts[Host Path Mounts]
    HostMounts --> KubeletPath["/var/lib/kubelet"]
    HostMounts --> DbusPath["/var/run/dbus"]
    
    %% Container Execution
    DaemonSetPod --> |Runs command| Container[control-plane-operator Container]
    Container --> |Executes| SyncCommand[sync-global-pullsecret command]
    
    %% Sync Process
    SyncCommand --> |Watches global-pull-secret| SyncController[Global Pull Secret Reconciler]
    SyncController --> |Reads secret| ReadGlobalPS[Read global-pull-secret]
    SyncController --> |Reads original| ReadOriginalPS[Read original pull-secret]
    
    %% File Update Process
    ReadGlobalPS --> |Gets data| GlobalPSBytes[Global Pull Secret Bytes]
    ReadOriginalPS --> |Gets data| OriginalPSBytes[Original Pull Secret Bytes]
    
    %% Decision Logic
    GlobalPSBytes --> |If exists| UseGlobalPS[Use Global Pull Secret]
    OriginalPSBytes --> |If not exists| UseOriginalPS[Use Original Pull Secret]
    
    %% File Update
    UseGlobalPS --> |Updates file| UpdateKubeletConfig["Update /var/lib/kubelet/config.json"]
    UseOriginalPS --> |Updates file| UpdateKubeletConfig
    
    %% Kubelet Restart
    UpdateKubeletConfig --> |Restarts kubelet| RestartKubelet[Restart kubelet.service via systemd]
    RestartKubelet --> |Via dbus| DbusConnection[DBus Connection]
    
    %% Error Handling
    UpdateKubeletConfig --> |If restart fails| RollbackProcess[Rollback Process]
    RollbackProcess --> |Restore original| RestoreOriginal[Restore Original File Content]
    
    %% Cleanup Process
    GlobalPSController --> |If additional PS deleted| CleanupProcess[Cleanup Process]
    CleanupProcess --> |Deletes global PS| DeleteGlobalPS[Delete global-pull-secret]
    CleanupProcess --> |Removes DaemonSet| RemoveDaemonSet[Remove DaemonSet]
    
    %% Styling
    classDef userInput fill:#e1f5fe
    classDef controller fill:#f3e5f5
    classDef secret fill:#e8f5e8
    classDef process fill:#fff3e0
    classDef daemonSet fill:#fce4ec
    classDef fileSystem fill:#f1f8e9
    
    class User,AdditionalPS userInput
    class HCCO,GlobalPSController,SyncController controller
    class OriginalPS,GlobalPSSecret,ServiceAccount,ClusterRole,ClusterRoleBinding secret
    class ValidatePS,MergeSecrets,RBACSetup,UpdateKubeletConfig,RestartKubelet process
    class DaemonSet,DaemonSetPod,Container daemonSet
    class KubeletPath,DbusPath fileSystem
```


control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -98,6 +99,7 @@ web_identity_token_file = /var/run/secrets/openshift/serviceaccount/token
 sts_regional_endpoints = regional
 region = %s
 `
+	UserProvidedPullSecretNamespace = "kube-system"
```
sjenning, line 102: We'll likely need to change this, maybe to `pull-secret-user` or something like that.  We'll need interlock on docs for this.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/globalps/pullsecret.go
```diff
@@ -0,0 +1,191 @@
...
+						{
+							Name:            "global-pull-secret-syncer",
+							Image:           cpoImage,
+							ImagePullPolicy: corev1.PullIfNotPresent,
+							Command: []string{
+								"/usr/bin/control-plane-operator",
+							},
+							Args: []string{
+								"sync-global-pullsecret",
+								"--kubelet-config-json-path=/var/lib/kubelet/config.json",
```
sjenning: I don't think there is a reason to take this path as an argument

control-plane-operator/hostedclusterconfigoperator/controllers/resources/globalps/pullsecret.go
```diff
@@ -0,0 +1,191 @@
...
+								"--global-ps-secret-name=global-pull-secret",
```
sjenning: Or this

control-plane-operator/hostedclusterconfigoperator/controllers/resources/globalps/pullsecret.go
```diff
@@ -0,0 +1,191 @@
...
+								"--check-interval=10s",
```
sjenning: The daemonset should be watching both the HCP pull secret and the user defined pull secret in the guest cluster.  What do we need this for?

sjenning (commented):
Generally, I don't think we want to narrow the naming of things to just pull secret.  I can see this becoming a general "hypershift worker agent" for a number of things that we need to update on the workers without doing a NodePool rollout.

I think we should create maybe a `worker-agent` subcommand on the CPO and have the pull secret sync controller be one of potentially many controllers within it.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/manifests/pullsecret.go
```diff
@@ -20,3 +21,31 @@ func PullSecretTargetNamespaces() []string {
 		"openshift",
 	}
 }
+
```
enxebre, line 31: can we drop the ns parameter from all the signatures here, these are all well known

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -2862,6 +2869,72 @@ func (r *reconciler) reconcileAzureCloudNodeManager(ctx context.Context, image s
...
+func (r *reconciler) reconcileGlobalPullSecret(ctx context.Context, hcp *hyperv1.HostedControlPlane, cpoImage string) error {
+	var (
+		additionalPullSecretBytes []byte
+		originalPullSecretBytes   []byte
+		globalPullSecretBytes     []byte
+		err                       error
+	)
+	log := ctrl.LoggerFrom(ctx)
+
+	additionalPullSecret := manifests.UserProvidedPullSecret(UserProvidedPullSecretNamespace)
```
enxebre: why renaming from UserProvidedPullSecret into additionalPullSecret in a variable definition. Can we use userProvidedPullSecret consistently everywhere?

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -2862,6 +2869,72 @@ func (r *reconciler) reconcileAzureCloudNodeManager(ctx context.Context, image s
...
+	// Asumming hcp pull secret is valid
+	originalPullSecretBytes = originalPullSecret.Data[corev1.DockerConfigJsonKey]
+
+	// Merge the additional pull secret with the original pull secret
+	if globalPullSecretBytes, err = globalpullsecret.MergePullSecrets(originalPullSecretBytes, additionalPullSecretBytes); err != nil {
+		return fmt.Errorf("failed to merge pull secrets: %w", err)
+	}
+
+	// Create secret in the DataPlane
+	secret := manifests.GlobalPullSecret(UserProvidedPullSecretNamespace)
```
enxebre: see https://github.com/openshift/hypershift/pull/6256/files#r2154011309

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -2862,6 +2869,72 @@ func (r *reconciler) reconcileAzureCloudNodeManager(ctx context.Context, image s
...
+	if _, err := r.CreateOrUpdate(ctx, r.uncachedClient, secret, func() error {
+		secret.Data = map[string][]byte{
+			corev1.DockerConfigJsonKey: globalPullSecretBytes,
```
enxebre: what happens if someone modifies the globalPullSecret directly?

control-plane-operator/hostedclusterconfigoperator/controllers/resources/globalps/pullsecret.go
```diff
@@ -0,0 +1,191 @@
...
+				},
+			},
+			Template: corev1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: map[string]string{
+						"name": "global-pull-secret-syncer",
+					},
+				},
+				Spec: corev1.PodSpec{
+					ServiceAccountName:           "global-pull-secret-syncer",
```
enxebre: where is this SA managed?

control-plane-operator/hostedclusterconfigoperator/controllers/resources/globalps/pullsecret.go
```diff
@@ -0,0 +1,191 @@
...
+					AutomountServiceAccountToken: ptr.To(false),
+					SecurityContext: &corev1.PodSecurityContext{
+						RunAsUser: ptr.To[int64](1000),
```
enxebre: I don't think we can't guarantee a particular uid exists in the SCC that will apply to this pod https://docs.redhat.com/en/documentation/openshift_container_platform/4.19/html/authentication_and_authorization/managing-pod-security-policies#security-context-constraints-example_configuring-internal-oauth

sync-pullsecret-ds/sync-pullsecret.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+	cmd.Flags().StringVar(&opts.kubeletConfigJsonPath, "kubelet-config-json-path", opts.kubeletConfigJsonPath, "The path to the kubelet's config.json file in the DataPlane.")
+	cmd.Flags().StringVar(&opts.globalPSSecretName, "global-ps-secret-name", opts.globalPSSecretName, "The name of the global pull secret secret in the DataPlane.")
+	cmd.Flags().DurationVar(&opts.checkInterval, "check-interval", opts.checkInterval, "The interval at which the file is checked for changes.")
+	cmd.Run = func(cmd *cobra.Command, args []string) {
+		ctx, cancel := context.WithCancel(context.Background())
+		sigs := make(chan os.Signal, 1)
+		signal.Notify(sigs, syscall.SIGINT)
+		go func() {
+			<-sigs
```
enxebre: why did you choose this vs a controller that watches the secret?

sync-pullsecret-ds/sync-pullsecret.go
```diff
@@ -0,0 +1,151 @@
...
+	if err := c.Get(ctx, client.ObjectKey{Namespace: defaultGlobalPullSecretNamespace, Name: o.globalPSSecretName}, globalPullSecret); err != nil {
+		return fmt.Errorf("failed to get global pull secret: %w", err)
+	}
+
+	// Get bytes of the global pull secret
+	globalPullSecretBytes := globalPullSecret.Data[corev1.DockerConfigJsonKey]
+
+	// If file content is different, write the desired content
+	if string(existingContent) != string(globalPullSecretBytes) {
+		if err := os.WriteFile(o.kubeletConfigJsonPath, globalPullSecretBytes, 0600); err != nil {
```
enxebre: does this crio/kubelet pick this on the fly?

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
jparrill: Well, it depends on the timing, but we expect that could happen 2 things:

- The daemonSet will detects the change
- The daemonSet changes the config.json in the nodes
- The HCCO detects that the globalPullSecret is not well set and overrides it
- The daemonSet detects the change
- The daemonSet changes the config.json in the nodes

Or 
- The HCCO detects that the globalPullSecret is not well set and overrides it
- The daemonSet detects the change
- The daemonSet changes the config.json in the nodes

control-plane-operator/hostedclusterconfigoperator/controllers/resources/globalps/pullsecret.go
jparrill: Yeah, I'm on this right now :)

sync-pullsecret-ds/sync-pullsecret.go
jparrill: Initially made sense because of size of the task. From the Seth's review I was thinking that makes more sense to create a controller instead of keep this approach to handle more tasks than just manage the global pull secret. 

I can do it agnostically and expandable in order to use it as a worker agent in the DataPlane to achieve tasks to avoid NodePool rollouts. WDYT?

sync-pullsecret-ds/sync-pullsecret.go
jparrill: No, I think I need to signal the Kubelet so this part is still WIP.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/globalps/pullsecret.go
jparrill: Humm you mean in the DataPlane, check on `openshift/pull-secret` and/or `openshift-config/pull-secret` + the new user provided and do the things in a logic isolated in the DataPlane? (right now it's a mixed term between the HCCO + the daemonSet)

Right now:
- HCCO checks on the userProvidedPullSecret + the HCP PullSecret and it creates the GlobalPullSecret
- The daemonSet mounts the GlobalPullSecret, it overwrites the file and should signal the Kubelet

control-plane-operator/hostedclusterconfigoperator/controllers/resources/globalps/pullsecret.go
```diff
@@ -0,0 +1,191 @@
...
+
+func ValidateAdditionalPullSecret(pullSecret *corev1.Secret) ([]byte, error) {
+	var dockerConfigJSON credentialprovider.DockerConfigJSON
+
+	// Validate that the pull secret contains the dockerConfigJson key
+	if _, ok := pullSecret.Data[corev1.DockerConfigJsonKey]; !ok {
+		return nil, fmt.Errorf("pull secret data is not a valid docker config json")
+	}
+
+	// Validate that the pull secret is a valid Docker config JSON
```
bryan-cox: Isn't this really just validating its valid JSON vs some Docker standard?

jparrill: Yes, I was trying to unmarshall it in a DockerConfigJSON struct, but when I do that, the format changes to a user/password instead of the original with auth: "base64(user:password)". I will change that comment

jparrill: It's trying to Unmarshal it into a DockerConfigJSON struct, so that would mean that at least the format is correct.

sync-pullsecret-ds/sync-pullsecret.go
jparrill: The kubelet needs to be restarted. I pushed into this PR a signaled way to do so via dbus, same approach the MCO uses in their code.

sync-global-pullsecret/sync-global-pullsecret.go
```diff
@@ -0,0 +1,227 @@
...
+		case <-ctx.Done():
+			return nil
+		}
+	}
+}
+
+// checkAndFixFile reads the current file content and updates it if it differs from the desired content.
+// Have in mind the logic which do the merge of the pull secret is in the globalpullsecret package under the HCCO.
+func (o *syncGlobalPullSecretOptions) checkAndFixFile(ctx context.Context, c client.Client) error {
+	// TODO (jparrill):
```
bryan-cox: Does this still need done or can this be removed?

bryan-cox:
@jparrill -do you have some docs on how to perform this that maybe could be turned into upstream docs for other users and QE to use?

control-plane-operator/hostedclusterconfigoperator/controllers/resources/globalps/pullsecret.go
```diff
@@ -0,0 +1,297 @@
...
+	log.Info("Creating global pull secret RBAC resources", "exists", exists)
+
+	// Create ServiceAccount
+	sa := manifests.GlobalPullSecretServiceAccount()
+	if _, err := createOrUpdate(ctx, c, sa, func() error { return nil }); err != nil {
+		return fmt.Errorf("failed to reconcile service account: %w", err)
+	}
+
+	// Create ClusterRole
+	clusterRole := manifests.GlobalPullSecretClusterRole()
```
enxebre: please make this namespaced roles / roleBindings in the target namespaces they need to operate and use ResourceNames to scope them to the specific secrets

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -751,6 +754,12 @@ func (r *reconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Result
 		}
 	}
 
+	// Reconcile GlobalPullSecret
+	hccoImage := os.Getenv("HOSTED_CLUSTER_CONFIG_OPERATOR_IMAGE")
```
enxebre: why is the image a param to reconcileGlobalPullSecret? can't we just set it in the deployment definition?

control-plane-operator/hostedclusterconfigoperator/controllers/resources/globalps/pullsecret.go
```diff
@@ -0,0 +1,297 @@
...
+var (
+	maxUnavailable = intstr.FromString("10%")
+	maxSurge       = intstr.FromInt(0)
+)
+
+const (
+	NodePullSecretPath = "/var/lib/kubelet/config.json"
+)
+
+func ReconcileDaemonSet(ctx context.Context, daemonSet *appsv1.DaemonSet, globalPullSecretBytes []byte, c crclient.Client, createOrUpdate upsert.CreateOrUpdateFN, cpoImage string) error {
```
enxebre: where is globalPullSecretBytes used in this func?

sync-global-pullsecret/sync-global-pullsecret.go
```diff
@@ -0,0 +1,227 @@
...
+	// 	- Validate the Kubelet flags does not contain a different path for the pull secret.
+
+	// Read existing content if file exists
+	existingContent, err := os.ReadFile(o.kubeletConfigJsonPath)
+	if err != nil && !os.IsNotExist(err) {
+		return fmt.Errorf("failed to read existing file: %w", err)
+	}
+
+	// Get the global pull secret from the DataPlane
+	globalPullSecret := &corev1.Secret{}
```
enxebre: why this instead of mounting it in the pod?

sync-global-pullsecret/sync-global-pullsecret.go
```diff
@@ -0,0 +1,227 @@
...
+		return fmt.Errorf("failed to restart kubelet after %d attempts, rolled back changes: %w", maxRetries, lastErr)
+	}
+
+	return nil
+}
+
+// signalKubeletToRestartProcess signals Kubelet to reload the config by restarting the kubelet.service.
+// This is done by sending a signal to systemd via dbus.
+func signalKubeletToRestartProcess() error {
+	log.Println("Signaling Kubelet to reload config")
```
enxebre: please use logger consistently

sync-global-pullsecret/sync-global-pullsecret.go
```diff
@@ -0,0 +1,227 @@
...
+	defaultCheckInterval             = 30 * time.Second
+	defaultGlobalPullSecretNamespace = "kube-system"
+	dbusRestartUnitMode              = "replace"
+	kubeletServiceUnit               = "kubelet.service"
+)
+
+// systemd job completion states as documented in go-systemd/dbus
+const (
+	systemdJobDone       = "done"       // Job completed successfully
+	systemdJobCanceled   = "canceled"   // Job was canceled before completion
```
enxebre: I can only see where systemdJobDone is used.
Can we remove anything that's not used?

sync-global-pullsecret/sync-global-pullsecret.go
```diff
@@ -0,0 +1,227 @@
...
+		if err := os.WriteFile(o.kubeletConfigJsonPath, globalPullSecretBytes, 0600); err != nil {
+			return fmt.Errorf("failed to write file: %w", err)
+		}
+		log.Printf("Pull secret updated %s with desired content", o.kubeletConfigJsonPath)
+
+		// Attempt to restart Kubelet with retries
+		maxRetries := 3
+		var lastErr error
+		for attempt := 1; attempt <= maxRetries; attempt++ {
+			if err := signalKubeletToRestartProcess(); err != nil {
```
enxebre: Is this logic taken from mco? if not, why? if yes can include comments pointing to the original source code?

sync-pullsecret-ds/sync-pullsecret.go
enxebre: My question is why is this an infinite loop with a ticker that reconciles every 10s instead of a regular controller runtime controller that watches the user provided secret?

control-plane-operator/hostedclusterconfigoperator/controllers/resources/globalps/pullsecret.go
```diff
@@ -0,0 +1,297 @@
...
+
+	// Unmarshal additional pull secret
+	if err = json.Unmarshal(userProvidedPullSecret, &userProvidedJSON); err != nil {
+		return nil, fmt.Errorf("invalid user provided pull secret format: %w", err)
+	}
+	userProvidedAuths = userProvidedJSON["auths"].(map[string]any)
+
+	// Merge auths
+	for k, v := range userProvidedAuths {
+		originalAuths[k] = v
```
gaol: Just want to confirm that with merge, user provided pull secret may override the global pull secret, is it what we want ? I am asking because I see a
```
This feature is to provide a mechanism where the customer can place a pull secret in cluster which would then result in it being appended to the global pull secret
```
in the Jira issue description which is saying `append`.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -2862,6 +2871,78 @@ func (r *reconciler) reconcileAzureCloudNodeManager(ctx context.Context, image s
...
+		globalPullSecretBytes       []byte
+		err                         error
+	)
+	log := ctrl.LoggerFrom(ctx)
+
+	// Get the user provided pull secret
+	userProvidedPullSecret := manifests.UserProvidedPullSecret()
+	if err := r.uncachedClient.Get(ctx, client.ObjectKeyFromObject(userProvidedPullSecret), userProvidedPullSecret); err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil
```
gaol: What happens if user deleted the `userProvidedPullSecret` after the `globalpullsecret` has been synced to the worker nodes ? Will it be restored back to `originalPullSecret` ?

control-plane-operator/hostedclusterconfigoperator/controllers/resources/globalps/pullsecret.go
jparrill: Yes, these are the cases:
- HCP PS: A=1, Customer PS: B=2, Global PS: A=1, B=2
- HCP PS: A=1, C=4, Customer PS: A=3, B=2, Global PS: A=3, B=2, C=4

So, the customer PS will have precedence over the HCP PS.

sync-pullsecret-ds/sync-pullsecret.go
jparrill: Now that makes sense to have it in the PR. I will implement it, I just want to make sure that was the approach we want to follow. Thanks!

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
jparrill: Humm we are setting it in the deployment definition as an image but we don't have any way to grab the Pod name there and then bubble down to the DaemonSet. This is why I've included it as envVar.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
jparrill: Yes, that's the expected behavior. Also, the RBAC and DaemonSet deployed should be removed.

jparrill:
working on the fixes to the PR comments

sync-global-pullsecret/sync-global-pullsecret.go
jparrill: Humm not at all, they're restarting/reloading directly the systemd unit:

- https://github.com/openshift/machine-config-operator/blob/f30fb1b4ce118c59bbdf9c555a59f137b3909967/pkg/daemon/update.go#L249C85-L249C98
- https://github.com/openshift/machine-config-operator/blob/main/pkg/daemon/update.go#L78-L88

I'm doing the same but via DBus.

jparrill:
Last pushes were to adjust the documentation rendering.

Documentation rendered at: https://686cda836a0b7c0008f2b986--hypershift-docs.netlify.app/how-to/common/global-pull-secret/

sync-global-pullsecret/sync-global-pullsecret.go
```diff
@@ -0,0 +1,263 @@
...
+
+	"github.com/coreos/go-systemd/dbus"
+	"github.com/spf13/cobra"
+)
+
+// syncGlobalPullSecretOptions contains the configuration options for the sync-global-pullsecret command
+type syncGlobalPullSecretOptions struct {
+	kubeletConfigJsonPath string
+	globalPSSecretName    string
+	checkInterval         time.Duration
```
enxebre: is checkInterval used at all?

sync-global-pullsecret/sync-global-pullsecret.go
```diff
@@ -0,0 +1,263 @@
...
+func NewRunCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "sync-global-pullsecret",
+		Short: "Syncs a mixture between the user provided pull secret in DataPlane and the HostedCluster PullSecret to be deployed in the nodes of the HostedCluster",
+		Long:  `Syncs a mixture between the user provided pull secret in DataPlane and the HostedCluster PullSecret to be deployed in the nodes of the HostedCluster. The resulting pull secret is deployed in a DaemonSet in the DataPlane that updates the kubelet.config.json file with the new pull secret. If there are conflicting entries in the resulting global pull secret, the user provided pull secret will prevail.`,
+	}
+
+	opts := syncGlobalPullSecretOptions{
+		kubeletConfigJsonPath: defaultKubeletConfigJsonPath,
+		checkInterval:         defaultCheckInterval,
```
enxebre: why is there a checkInterval, can we drop it?

enxebre:
please add e2e tests for this.

sync-global-pullsecret/sync-global-pullsecret.go
jparrill: Yes, Done

sync-global-pullsecret/sync-global-pullsecret.go
jparrill: It was, not needed anymore. Dropped already. Thanks!

jparrill:
> please add e2e tests for this.

E2E Added. Thanks!. When you have a min PTAL @enxebre 🙏

test/e2e/util/util.go
```diff
@@ -1506,6 +1508,282 @@ func EnsureGuestWebhooksValidated(t *testing.T, ctx context.Context, guestClient
 	})
 }
 
+func EnsureGlobalPullSecret(t *testing.T, ctx context.Context, mgmtClient crclient.Client, guestClient crclient.Client, entryHostedCluster *hyperv1.HostedCluster) error {
+	//AtLeast(t, Version420)
```
enxebre: why is this commented?

test/e2e/util/util.go
```diff
@@ -1506,6 +1508,282 @@ func EnsureGuestWebhooksValidated(t *testing.T, ctx context.Context, guestClient
 	})
 }
```
enxebre, line 1524: where is EnsureGlobalPullSecret invoked?

test/e2e/util/util.go
```diff
@@ -1506,6 +1508,282 @@ func EnsureGuestWebhooksValidated(t *testing.T, ctx context.Context, guestClient
...
+	g := NewWithT(t)
+
+	// Create the additional-pull-secret secret in the DataPlane using the dummy pull secret.
+	// The dummy pull secret is authorized to pull restricted images.
+	err = createAdditionalPullSecret(ctx, guestClient, dummyPullSecretData, registrySecretName, registryNamespace)
+	g.Expect(err).NotTo(HaveOccurred(), "failed to create additional-pull-secret secret")
+
+	// Check if additional-pull-secret secret is in the right place at Dataplane
+	t.Run("Check if additional-pull-secret secret is in the right place at Dataplane", func(t *testing.T) {
+		t.Log("Checking if additional-pull-secret secret is in the right place at Dataplane")
```
enxebre: this log is redundant with t.run right above

test/e2e/util/util.go
```diff
@@ -1506,6 +1508,282 @@ func EnsureGuestWebhooksValidated(t *testing.T, ctx context.Context, guestClient
...
+		additionalPullSecret := hccomanifests.UserProvidedPullSecret()
```
enxebre: why are we naming differently the same thing additionalPullSecret vs UserProvidedPullSecret?
Can we please use one term consitently?

test/e2e/util/util.go
```diff
@@ -1506,6 +1508,282 @@ func EnsureGuestWebhooksValidated(t *testing.T, ctx context.Context, guestClient
...
+		g.Eventually(func() error {
+			if err := guestClient.Get(ctx, client.ObjectKey{Name: additionalPullSecret.Name, Namespace: additionalPullSecret.Namespace}, additionalPullSecret); err != nil {
+				return err
+			}
+			return nil
+		}, 20*time.Minute, 5*time.Second).Should(Succeed(), "additional-pull-secret secret is not present")
```
enxebre: is this waiting 20m for the secret to exist? why? shouldn't that be almost immediate?

test/e2e/util/util.go
jparrill: Oh, yes that was for local testing, sorry. I'll fix it

test/e2e/util/util.go
jparrill: yes, but I didn't test it with a loaded MGMT cluster. That's arbitrary, I can change it for w/e number

test/e2e/util/util.go
```diff
@@ -1506,6 +1508,282 @@ func EnsureGuestWebhooksValidated(t *testing.T, ctx context.Context, guestClient
...
 
+func EnsureGlobalPullSecret(t *testing.T, ctx context.Context, mgmtClient crclient.Client, guestClient crclient.Client, entryHostedCluster *hyperv1.HostedCluster) error {
+	//AtLeast(t, Version420)
+
+	var (
+		dummyImageTagMultiarch = "quay.io/hypershift/sleep:multiarch"
+		dummyImageTag12        = "quay.io/hypershift/sleep:1.2.0"
+		err                    error
+
+		// Additional Pull Secret
```
enxebre, line 1537: why is "additional pull secret" declared as registrySecret and then fetch as UserProvidedPullSecret. Can we please use the same term everywhere when referring to the same thing?

enxebre: also referred as dummy pull secret

test/e2e/util/util.go
```diff
@@ -1506,6 +1508,282 @@ func EnsureGuestWebhooksValidated(t *testing.T, ctx context.Context, guestClient
...
+		t.Log("Checking if additional-pull-secret secret is in the right place at Dataplane")
+		additionalPullSecret := hccomanifests.UserProvidedPullSecret()
+		g.Eventually(func() error {
+			if err := guestClient.Get(ctx, client.ObjectKey{Name: additionalPullSecret.Name, Namespace: additionalPullSecret.Namespace}, additionalPullSecret); err != nil {
+				return err
+			}
+			return nil
+		}, 20*time.Minute, 5*time.Second).Should(Succeed(), "additional-pull-secret secret is not present")
+		g.Expect(additionalPullSecret.Data).NotTo(BeEmpty(), "additional-pull-secret secret is empty")
+		g.Expect(additionalPullSecret.Data[corev1.DockerConfigJsonKey]).NotTo(BeEmpty(), "additional-pull-secret secret is empty")
```
enxebre: you can also use e2eutil.EventuallyObject for this secret and predicate which has some guardrails and logging for failures

test/e2e/util/util.go
```diff
@@ -1506,6 +1508,282 @@ func EnsureGuestWebhooksValidated(t *testing.T, ctx context.Context, guestClient
...
+	})
+
+	// Check if HCCO generates the GlobalPullSecret secret in the kube-system namespace in the DataPlane
+	t.Run("Check if GlobalPullSecret secret is in the right place at Dataplane", func(t *testing.T) {
```
enxebre, line 1556: same comments than for the secret above about e2eutil.EventuallyObject, 20min timeout and redundant logging

test/e2e/util/util.go
```diff
@@ -1506,6 +1508,282 @@ func EnsureGuestWebhooksValidated(t *testing.T, ctx context.Context, guestClient
...
+	// Check if the DaemonSet is present in the DataPlane
+	t.Run("Check if the DaemonSet is present in the DataPlane", func(t *testing.T) {
+		t.Log("Checking if the DaemonSet is present in the DataPlane")
+		g.Eventually(func() error {
+			daemonSet := hccomanifests.GlobalPullSecretDaemonSet()
+			if err := guestClient.Get(ctx, client.ObjectKey{Name: daemonSet.Name, Namespace: daemonSet.Namespace}, daemonSet); err != nil {
+				return err
+			}
+			return nil
+		}, 20*time.Minute, 5*time.Second).Should(Succeed(), "DaemonSet is not present")
```
enxebre: if any of these assumptions with 20min timeout takes more than seconds, something is wrong and I wouldn't want the test to pass, no?

test/e2e/util/util.go
```diff
@@ -2722,3 +3000,87 @@ func EnsureConsoleCapabilityDisabled(ctx context.Context, t *testing.T, g Gomega
...
+		}
+	}
+
+	// Check if all attempts failed
+	if createErr != nil {
+		t.Fatalf("Failed to create pod after %d attempts. Last error: %v", maxRetries, createErr)
+	}
+
+	t.Log("Waiting for 30 seconds to ensure the pod is created")
+	time.Sleep(30 * time.Second)
```
enxebre: why do we need additional sleep to the eventually workflow?

sync-global-pullsecret/sync-global-pullsecret.go
```diff
@@ -0,0 +1,257 @@
...
+	cmd := &cobra.Command{
+		Use:   "sync-global-pullsecret",
+		Short: "Syncs a mixture between the user provided pull secret in DataPlane and the HostedCluster PullSecret to be deployed in the nodes of the HostedCluster",
+		Long:  `Syncs a mixture between the user provided pull secret in DataPlane and the HostedCluster PullSecret to be deployed in the nodes of the HostedCluster. The resulting pull secret is deployed in a DaemonSet in the DataPlane that updates the kubelet.config.json file with the new pull secret. If there are conflicting entries in the resulting global pull secret, the user provided pull secret will prevail.`,
+	}
+
+	opts := syncGlobalPullSecretOptions{
+		kubeletConfigJsonPath: defaultKubeletConfigJsonPath,
+	}
+	cmd.Flags().StringVar(&opts.globalPSSecretName, "global-pull-secret-name", defaultGlobalPSSecretName, "The name of the global pullSecret secret in the DataPlane.")
```
enxebre, line 82: why does this global-pull-secret-name need to be a param?

test/e2e/util/util.go
jparrill: I probably override the call from a rebase.

sync-global-pullsecret/sync-global-pullsecret.go
```diff
@@ -0,0 +1,257 @@
...
+	log.Info("Reconciling GlobalPullSecret")
+
+	// Get the global pull secret
+	globalPullSecret := &corev1.Secret{}
+	if err := r.Get(ctx, req.NamespacedName, globalPullSecret); err != nil {
+		log.Error(err, "Failed to get global pull secret")
+		return reconcile.Result{}, client.IgnoreNotFound(err)
+	}
+
+	// Check and fix the file
```
enxebre: nit: this comment is redundant with the func

sync-global-pullsecret/sync-global-pullsecret.go
jparrill: for flexibility, if we need to change the secret name, it's matter of pass that new name to the controller

sync-global-pullsecret/sync-global-pullsecret.go
```diff
@@ -0,0 +1,257 @@
...
+		WithEventFilter(predicate.Funcs{
+			// Adding filters to avoid processing events that are not relevant
+			CreateFunc: func(e event.CreateEvent) bool {
+				return o.isTargetSecret(e.Object)
+			},
+			UpdateFunc: func(e event.UpdateEvent) bool {
+				return o.isTargetSecret(e.ObjectNew)
+			},
+			DeleteFunc: func(e event.DeleteEvent) bool {
+				return false
```
enxebre: why are deleted events dropped?
what happens during deletion?

sync-global-pullsecret/sync-global-pullsecret.go
enxebre: the name should never change and if it's does, how is a param more flexible than a hardcoded value providing the ds instantiation is codified itself? The only value I can think of is to run the daemon set manually for dev purposes

test/e2e/util/util.go
jparrill: I assume that depends on the MGMT cluster load, In my cluster it takes seconds but I cannot say the same on the CI. So I've set this 20 min as much. I can put that limit to 2 minutes maybe?

sync-global-pullsecret/sync-global-pullsecret.go
jparrill: This controller looks at the secrets, and concretely to GlobalPullSecret which is baked by HCCO.
- The user creates AdditionalPullSecret
- HCCO Merges with HCP one and creates GlobalPullSecret in Guest side
- The DS checks that Secret and syncs it with the nodes `config.json` and restarts the kubelet.

During deletion:
- HCCO detects the used delete the AdditionalPullSecret, then triggers the signal to restore the original config.json
- The HCCO executes the `recoverBeforeShutdown` flow. It goes through the whole process but merging the HCP pullSecret with a empty additionalPullSecret
- The DS do their work in the same way.
- In the next reconciliation of the HCCO it will delete all the things, RBAC, DS and the in place secrets generated.

So the DS controller will not need to check deletion events because the one which should look at it is HCCO.

test/e2e/util/util.go
jparrill: I think for now it's fine as it is. There are some improvements that could be done in this E2E like use the internal registry instead of external ones, use the e2eutil.EventuallyObject, etc... I think as an starting point to get the E2E rolling out it's ok. WDYT?

test/e2e/util/util.go
jparrill: changed to 2 minutes most of them, just one on 5 minutes which involves a container image pull.

test/e2e/util/util.go
jparrill: changed to 2 minutes most of them, just one on 5 minutes which involves a container image pull.

jparrill:
/retest

Flaky tests failing but not related with this PR

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -755,6 +812,15 @@ func (r *reconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Result
 		}
 	}
 
+	// Reconcile GlobalPullSecret
+	hccoImage := os.Getenv("HOSTED_CLUSTER_CONFIG_OPERATOR_IMAGE")
+	if err := r.reconcileGlobalPullSecret(ctx, hcp, hccoImage); err != nil {
```
enxebre: I would rather let this be its own controller so reconciliation is only caused by their relvant resources. We can follow up on that in a different PR now that we have test coverage.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -2862,6 +2928,128 @@ func (r *reconciler) reconcileAzureCloudNodeManager(ctx context.Context, image s
...
+	if err != nil {
+		return fmt.Errorf("failed to check if user provided pull secret exists: %w", err)
+	}
+
+	// Reconcile the RBAC for the Global Pull Secret
+	if err := globalpullsecret.ReconcileGlobalPullSecretRBAC(ctx, r.uncachedClient, r.CreateOrUpdate, hcp.Namespace); err != nil {
+		return fmt.Errorf("failed to reconcile global pull secret RBAC: %w", err)
+	}
+
+	if !exists {
```
enxebre: this should possibly use a finalizer and a deletion timestamp. We can follow up on that in a different PR now that we have test coverage.

sync-global-pullsecret/sync-global-pullsecret.go
enxebre: what happens if the global secret is deleted/updated out of band?

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -195,10 +233,17 @@ func Setup(ctx context.Context, opts *operator.HostedClusterConfigOperatorConfig
 		versions:                  opts.Versions,
 		operateOnReleaseImage:     opts.OperateOnReleaseImage,
 		ImageMetaDataProvider:     opts.ImageMetaDataProvider,
+		kubeSystemSecretInformer:  kubeSystemSecretInformer,
```
enxebre: why is this informer in the reconciler? seems unused

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -2862,6 +2928,128 @@ func (r *reconciler) reconcileAzureCloudNodeManager(ctx context.Context, image s
...
+	}
+
+	// Reconcile the RBAC for the Global Pull Secret
+	if err := globalpullsecret.ReconcileGlobalPullSecretRBAC(ctx, r.uncachedClient, r.CreateOrUpdate, hcp.Namespace); err != nil {
+		return fmt.Errorf("failed to reconcile global pull secret RBAC: %w", err)
+	}
+
+	if !exists {
+		// Early cleanup
+		if recoverBeforeShutdown {
```
enxebre: what's the point of recoverBeforeShutdown? its value will be true every reconciliation loop

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -2862,6 +2928,128 @@ func (r *reconciler) reconcileAzureCloudNodeManager(ctx context.Context, image s
...
+					corev1.DockerConfigJsonKey: globalPullSecretBytes,
+				}
+				return nil
+			}); err != nil {
+				return fmt.Errorf("failed to create global pull secret: %w", err)
+			}
+			log.Info("Original pull secret recovered, global pull secret syncer signaled to shutdown")
+			recoverBeforeShutdown = false
+
+			return fmt.Errorf("original pull secret recovered, global pull secret syncer signaled to shutdown")
```
enxebre: why is this erroring? This seems that the intent was executed

enxebre: wouldn't we reach this error every loop if the secret doesn't exist?

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -755,6 +812,15 @@ func (r *reconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Result
 		}
 	}
 
+	// Reconcile GlobalPullSecret
+	hccoImage := os.Getenv("HOSTED_CLUSTER_CONFIG_OPERATOR_IMAGE")
+	if err := r.reconcileGlobalPullSecret(ctx, hcp, hccoImage); err != nil {
+		errs = append(errs, fmt.Errorf("failed to reconcile global pull secret: %w", err))
+		if strings.Contains(err.Error(), "global pull secret syncer signaled to shutdown") {
```
enxebre: why is this RequeueAfter needed?

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
enxebre: nvm it's at the pkg level

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
enxebre: nvm about reaching the error https://github.com/openshift/hypershift/pull/6256#discussion_r2212912379

enxebre:
Some more comments to address in this PR:
https://github.com/openshift/hypershift/pull/6256#discussion_r2212874150
https://github.com/openshift/hypershift/pull/6256#discussion_r2212857374
https://github.com/openshift/hypershift/pull/6256#discussion_r2212870360

Then, In general I think this should be its own controller which doesn't increase resources.go complexity.
The deletion workflow should be driven by finalizers for both the main controller and the daemon. The former removes the finalizer when it restores back the global pull secret, the latter removes its finalizer when restores it back to disk

I'm ok with these two items being addressed in a separate PR iteratively now that we have e2e coverage.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
jparrill: With this RequeueAfter I expect the DS controller have enough time to reconcile, modify the file and restart the kubelet, which takes, a couple of seconds. But I giving it more time in case of a high load situation.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
jparrill: nice catch. Thanks

sync-global-pullsecret/sync-global-pullsecret.go
jparrill: HCCO will recreate it with the right content

sync-global-pullsecret/sync-global-pullsecret.go
jparrill: Yes, that’s a valid reason to keep it as well. I initially included it just for flexibility, but if you feel strongly about removing it, I’m happy to take it out.

jparrill:
Added https://issues.redhat.com/browse/CNTRLPLANE-1154 to track the follow up work to do

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
enxebre: requeafter is moot here since you are returning an error it would be added to the rate liming queue https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/internal/controller/controller.go#L340-L374

And even if you were returning requeafter without error, reconciliation would certainly most likely happen earlier given all the watchers this controller has.

Also we there's no guarantees, the next reconciliation loop happens after the daemonset did it's job.

We do really need to put this into it's own controller and use finalizers.

test/e2e/util/util.go
```diff
@@ -1512,6 +1514,271 @@ func EnsureGuestWebhooksValidated(t *testing.T, ctx context.Context, guestClient
...
+			}
+
+			serviceAccount := hccomanifests.GlobalPullSecretServiceAccount()
+			if err := guestClient.Get(ctx, client.ObjectKey{Name: serviceAccount.Name, Namespace: serviceAccount.Namespace}, serviceAccount); err != nil {
+				if !apierrors.IsNotFound(err) {
+					return err
+				}
+				return nil
+			}
+
```
enxebre: this should check the disk in file is as expected after deletion or that the image can't be pulled anymore

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
jparrill: tbh I'm being conservative on this, I just put a requeueAfter to give it time, but maybe that's not necessary at all. I didn't reproduced the situation when the daemonSet gets deleted and the config.json were not recovered to the original state.

enxebre:
>The deletion workflow should be driven by finalizers for both the main controller and the daemon. The former removes the finalizer when it restores back the global pull secret, the latter removes its finalizer when restores it back to disk

plus this needs a coordination mechanism to also ensure that the on delete reconciliation succeeds on every Node in the cluster. Otherwise there's no guarantees reconciliation on deletion executes on a single node and there's no guarantees it executes in all nodes either

jparrill:
Hey @enxebre I've addressed all the concerns and updated the PR description, commit, code, documentation.... PTAL when you have a moment 🙏.

For a follow up PR:
- Investigate to have ephemeral DS to save some resources in DataPlane side (Maybe switching to jobBatch).
- Small nits in documentation rendered.
- For security reasons better to mount a volume with the original HCCO and global pullSecrets
- Try to avoid give permissions to the DS to access API
- Remove privilleged true and reduce the permissions needed

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
```diff
@@ -0,0 +1,366 @@
...
+	daemonSet := manifests.GlobalPullSecretDaemonSet()
+	if err := ReconcileDaemonSet(ctx, daemonSet, secret.Name, r.hcUncachedClient, r.CreateOrUpdate, cpoImage); err != nil {
+		return fmt.Errorf("failed to reconcile global pull secret daemon set: %w", err)
+	}
+
+	return nil
+}
+
+func ReconcileDaemonSet(ctx context.Context, daemonSet *appsv1.DaemonSet, globalPullSecretName string, c crclient.Client, createOrUpdate upsert.CreateOrUpdateFN, cpoImage string) error {
+	log, err := logr.FromContext(ctx)
```
enxebre: why is this using logr instead of the controller runtime library we generally use ctrl.LoggerFrom(ctx)

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/setup.go
```diff
@@ -0,0 +1,93 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/cache"
+	crclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+func Setup(ctx context.Context, opts *operator.HostedClusterConfigOperatorConfig) error {
+	// Create a predicate for the pull-secret
```
enxebre, line 22: let's please just return early here if this is not ARO. We don't want to start running a new controller in a z stream for other envs for no reason.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/manifests/pullsecret.go
```diff
@@ -1,10 +1,20 @@
...
 	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	GlobalPullSecretDSName       = "global-pull-secret-syncer"
+	GlobalPullSecretNamespace    = "kube-system"
+	NodePullSecretPath           = "/var/lib/kubelet/config.json"
+	GlobalPullSecretFinalizer    = "hypershift.openshift.io/global-pull-secret-finalizer"
```
enxebre: I dont think GlobalPullSecretFinalizer and OriginalPullSecretAnnotation are used, can we drop them?

control-plane-operator/hostedclusterconfigoperator/controllers/resources/manifests/pullsecret.go
```diff
@@ -20,3 +30,56 @@ func PullSecretTargetNamespaces() []string {
...
+	return &corev1.ServiceAccount{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "global-pull-secret-syncer",
+			Namespace: GlobalPullSecretNamespace,
+		},
+	}
+}
+
+func GlobalPullSecretRole() *rbacv1.ClusterRole {
+	return &rbacv1.ClusterRole{
```
enxebre: this needs to be a role

control-plane-operator/hostedclusterconfigoperator/controllers/resources/manifests/pullsecret.go
```diff
@@ -20,3 +30,56 @@ func PullSecretTargetNamespaces() []string {
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "global-pull-secret-syncer",
+		},
+	}
+}
+
+func GlobalPullSecretRoleBinding() *rbacv1.ClusterRoleBinding {
+	return &rbacv1.ClusterRoleBinding{
```
enxebre: this needs to be a roleBinding?

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -99,6 +99,7 @@ web_identity_token_file = /var/run/secrets/openshift/serviceaccount/token
 sts_regional_endpoints = regional
 region = %s
 `
+	AdditionalPullSecretNamespace = "kube-system"
```
enxebre: these changes in this file seems are not needed now

control-plane-operator/hostedclusterconfigoperator/controllers/resources/manifests/pullsecret.go
jparrill: Nop, it needs to be a ClusterRole and ClusterRoleBinding. The controller checks the `pull-secret` from `openshift-config` NS in the DataPlane in order to recover the original PullSecret.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/manifests/pullsecret.go
jparrill: Same as commented in https://github.com/openshift/hypershift/pull/6256#discussion_r2221925547

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/setup.go
jparrill: GlobalPullSecret is for ROSA and ARO, should I limit the scope for now just to ARO, or should I include also ROSA?

control-plane-operator/hostedclusterconfigoperator/controllers/resources/manifests/pullsecret.go
enxebre: Then you need several roles. We don't want to expose an overprivileged SA in all Nodes.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/manifests/pullsecret.go
jparrill: spreading in multiple roles

control-plane-operator/controllers/hostedcontrolplane/v2/configoperator/deployment.go
```diff
@@ -28,6 +29,7 @@ func (h *hcco) adaptDeployment(cpContext component.WorkloadContext, deployment *
 	kubeVersion := versions["kubernetes"]
 	hcp := cpContext.HCP
 	openShiftVersion := cpContext.ReleaseImageProvider.Version()
+	hccoImage := cpContext.ReleaseImageProvider.GetImage(hostedClusterConfigOperatorImage)
```
csrwng: nit: this image is already set in the container, you could just use `c.Image` below

control-plane-operator/hostedclusterconfigoperator/cmd.go
```diff
@@ -67,6 +68,7 @@ var controllerFuncs = map[string]operator.ControllerSetupFunc{
 	"machine":                machine.Setup,
 	"drainer":                drainer.Setup,
 	hcpstatus.ControllerName: hcpstatus.Setup,
+	globalps.ControllerName:  globalps.Setup,
```
csrwng: nit on naming: `pullsecret` or `gpullsecret` would help me better understand what the package is for

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
```diff
@@ -0,0 +1,366 @@
...
+	hcpNamespace           string
+	upsert.CreateOrUpdateProvider
+}
+
+func (r *Reconciler) Reconcile(ctx context.Context, req crreconcile.Request) (crreconcile.Result, error) {
+	log := ctrl.LoggerFrom(ctx)
+	log.Info("reconciling global pull secret")
+
+	// Reconcile GlobalPullSecret
+	hccoImage := os.Getenv("HOSTED_CLUSTER_CONFIG_OPERATOR_IMAGE")
```
csrwng: This should likely be read in once where the controller is created and passed in a field. It's not going to change through the life of the controller.

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
```diff
@@ -0,0 +1,366 @@
...
+		}
+		return nil
+	}); err != nil {
+		return fmt.Errorf("failed to reconcile global pull secret syncer role binding: %w", err)
+	}
+
+	return nil
+}
+
+func AdditionalPullSecretExists(ctx context.Context, c crclient.Client) (bool, *corev1.Secret, error) {
```
csrwng: nit: no need to make this a public function. Also, just `additionalPullSecret` might be a better name

control-plane-operator/hostedclusterconfigoperator/controllers/resources/manifests/pullsecret.go
```diff
@@ -20,3 +28,58 @@ func PullSecretTargetNamespaces() []string {
 		"openshift",
 	}
 }
+
+func AdditionalPullSecret() *corev1.Secret {
+	return &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "additional-pull-secret",
+			Namespace: GlobalPullSecretNamespace,
```
csrwng, line 36: We should use the `openshift-config` namespace for the user-provided pull secret, since that's where we expect user configuration by convention. It's fine that we put the merged pull secret in `kube-system`.

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
```diff
@@ -0,0 +1,392 @@
...
+		return nil
+	}
+
+	// If the PS doesn't exist, the HCCO doesn't do anything.
+	if additionalPullSecret.Data == nil {
+		return nil
+	}
+
+	if userProvidedPullSecretBytes, err = ValidateAdditionalPullSecret(additionalPullSecret); err != nil {
+		return fmt.Errorf("failed to validate user provided pull secret: %w", err)
```
csrwng, line 86: What happens if the pull secret provided by the user is not valid? How do we communicate to them that they need to fix their pull secret? I don't think having the controller retry would help.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/manifests/pullsecret.go
```diff
@@ -20,3 +28,58 @@ func PullSecretTargetNamespaces() []string {
...
+	return &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "global-pull-secret",
+			Namespace: GlobalPullSecretNamespace,
+		},
+		Type: corev1.SecretTypeDockerConfigJson,
+	}
+}
+
+func GlobalPullSecretServiceAccount() *corev1.ServiceAccount {
```
csrwng: nit: s/GlobalPullSecretServiceAccount/GlobalPullSecretSyncerServiceAccount/

control-plane-operator/hostedclusterconfigoperator/controllers/resources/manifests/pullsecret.go
```diff
@@ -20,3 +28,58 @@ func PullSecretTargetNamespaces() []string {
...
+	return &corev1.ServiceAccount{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "global-pull-secret-syncer",
+			Namespace: GlobalPullSecretNamespace,
+		},
+	}
+}
+
+func GlobalPullSecretRole(ns string) *rbacv1.Role {
```
csrwng: nit: s/GlobalPullSecretRole/GlobalPullSecretSyncerRole/

control-plane-operator/hostedclusterconfigoperator/controllers/resources/manifests/pullsecret.go
```diff
@@ -20,3 +28,58 @@ func PullSecretTargetNamespaces() []string {
...
+	return &rbacv1.Role{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "global-pull-secret-syncer",
+			Namespace: ns,
+		},
+	}
+}
+
+func GlobalPullSecretRoleBinding(ns string) *rbacv1.RoleBinding {
```
csrwng: nit: s/GlobalPullSecretRoleBinding/GlobalPullSecretSyncerRoleBinding/

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
```diff
@@ -0,0 +1,392 @@
...
+	log.Info("reconciling global pull secret")
+
+	// Get the user provided pull secret
+	exists, additionalPullSecret, err := AdditionalPullSecretExists(ctx, r.kubeSystemSecretClient)
+	if err != nil {
+		return fmt.Errorf("failed to check if user provided pull secret exists: %w", err)
+	}
+
+	// Reconcile the RBAC for the Global Pull Secret
+	if err := ReconcileGlobalPullSecretRBAC(ctx, r.hcUncachedClient, r.kubeSystemSecretClient, r.CreateOrUpdate, hcpNamespace, "openshift-config"); err != nil {
```
csrwng: If you are merging the pull secret in the hcco, why does the daemonset need rbac access to any secrets? Wouldn't it be enough just to mount the pull secret to the daemonset?

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
```diff
@@ -0,0 +1,392 @@
...
+	if _, err := r.CreateOrUpdate(ctx, r.kubeSystemSecretClient, secret, func() error {
+		secret.Data = map[string][]byte{
+			corev1.DockerConfigJsonKey: globalPullSecretBytes,
+		}
+		return nil
+	}); err != nil {
+		return fmt.Errorf("failed to create global pull secret: %w", err)
+	}
+
+	// Use the Global Pull Secret to deploy the DaemonSet in the DataPlane.
```
csrwng: > Use the Global Pull Secret to deploy the DaemonSet in the DataPlane.

How?

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
```diff
@@ -0,0 +1,392 @@
...
+
+		return nil
+	}); err != nil {
+		return fmt.Errorf("failed to create global pull secret daemon set: %w", err)
+	}
+
+	return nil
+}
+
+func ValidateAdditionalPullSecret(pullSecret *corev1.Secret) ([]byte, error) {
```
csrwng: no need for a public function

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
```diff
@@ -0,0 +1,392 @@
...
+	return pullSecretBytes, nil
+}
+
+// MergePullSecrets merges two pull secrets into a single pull secret.
+// The additional pull secret is merged with the original pull secret.
+// If an auth entry already exists, it will be overwritten.
+// The resulting pull secret is returned as a JSON string.
+// Not using credentialprovider.DockerConfigJSON because it does not support
+// marshaling the auth field.
+func MergePullSecrets(ctx context.Context, originalPullSecret, userProvidedPullSecret []byte) ([]byte, error) {
```
csrwng: no need for public function

csrwng (commented):
Some comments. If possible, the daemonset process should have as little privilege as possible. It should be able to sync a secret that is mounted to it, instead of it directly accessing the API.

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
jparrill: it does not mount the secret into the DS:
- The HCCO creates the GlobalPullSecret into the DataPlane merging with Additional one
- The DS then checks the secrets actively using the cached client  in both namespaces.
- Then:
  - If the GlobalPullSecret exists the DS takes that as a reference to substitute the config.json
  - if not, we take the originalPullSecret from openshift-config

The checkAndFix function have the logic to override the the config.json according the needs using a payload

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
jparrill: Yeah, there are some function that comes from other packages and I forgot to unpublish them

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
```diff
@@ -0,0 +1,392 @@
...
+	log.Info("reconciling global pull secret")
+
+	// Get the user provided pull secret
+	exists, additionalPullSecret, err := AdditionalPullSecretExists(ctx, r.kubeSystemSecretClient)
+	if err != nil {
+		return fmt.Errorf("failed to check if user provided pull secret exists: %w", err)
+	}
+
+	// Reconcile the RBAC for the Global Pull Secret
+	if err := ReconcileGlobalPullSecretRBAC(ctx, r.hcUncachedClient, r.kubeSystemSecretClient, r.CreateOrUpdate, "kube-system", "openshift-config"); err != nil {
```
muraee: why reconcile the RBAC here before checking if the additionalPullSecret exist? This should be moved down just before where "ReconcileDaemonSet" is called.

jparrill: Yeah, makes sense. It had more sense before, when the RBAC and DS were deleted. Thanks!

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
```diff
@@ -0,0 +1,392 @@
...
+			if !apierrors.IsNotFound(err) {
+				return fmt.Errorf("failed to delete global pull secret: %w", err)
+			}
+		}
+		return nil
+	}
+
+	// If the PS doesn't exist, the HCCO doesn't do anything.
+	if additionalPullSecret.Data == nil {
+		return nil
```
muraee: why just return here and not delete the globalPullSecert like you do above when the secret doesn't exist? imo this should be merged with the case above.
```
if !exists || additionalPullSecret.Data == nil {
 ....
```

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
jparrill: The controller reconciles the daemon set and the uncached client takes care of the resource creation. I will remove the comment, tbh is not useful

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
jparrill: it shows a log entry saying the PS is not valid, also is documented how to create one.

control-plane-operator/controllers/hostedcontrolplane/v2/configoperator/deployment.go
jparrill: Ohh looks like that image resolves in: `hosted-cluster-config-operator`. I've reverted the change.

muraee: yea c.Image is resolved after the adapt func is executed so using c.Image for the env here wouldn't work.

sync-global-pullsecret/sync-global-pullsecret.go
```diff
@@ -0,0 +1,309 @@
...
+	err = r.uncachedClient.Get(ctx, types.NamespacedName{
+		Name:      originalPullSecretName,
+		Namespace: originalPullSecretNamespace,
+	}, originalPullSecret)
+	if err != nil {
+		return reconcile.Result{}, fmt.Errorf("failed to get original pull secret: %w", err)
+	}
+
+	err = r.cachedClient.Get(ctx, req.NamespacedName, globalPullSecret)
+	if err != nil {
```
enxebre, line 192: why is this fallback needed at all?

enxebre:
/lgtm

Thanks! lgtm as a first iteration, let's follow up in subsequent PRs.

sync-global-pullsecret/sync-global-pullsecret.go
jparrill: This is like a PullSecret selector:
- If the GlobalPullSecret does not exists: pick the Original PullSecret
- Otherwise pick the GlobalPullSecret

test/e2e/util/util.go
```diff
@@ -1512,6 +1514,248 @@ func EnsureGuestWebhooksValidated(t *testing.T, ctx context.Context, guestClient
...
+			if err := guestClient.Get(ctx, client.ObjectKey{Name: daemonSet.Name, Namespace: daemonSet.Namespace}, daemonSet); err != nil {
+				return err
+			}
+			dsImage = daemonSet.Spec.Template.Spec.Containers[0].Image
+			return nil
+		}, 20*time.Minute, 5*time.Second).Should(Succeed(), "DaemonSet is not present")
+	})
+
+	// Wait 10 seconds to ensure the DaemonSet is present
+	t.Log("Waiting 10 seconds to ensure the DaemonSet is present")
```
enxebre: please delete the sleep and rely on eventually retries

test/e2e/util/util.go
```diff
@@ -1512,6 +1514,239 @@ func EnsureGuestWebhooksValidated(t *testing.T, ctx context.Context, guestClient
...
+	t.Run("Check if we can pull restricted images, should fail", func(t *testing.T) {
+		g.Eventually(func() error {
+			globalPullSecret := hccomanifests.GlobalPullSecret()
+			if err := guestClient.Get(ctx, client.ObjectKey{Name: globalPullSecret.Name, Namespace: globalPullSecret.Namespace}, globalPullSecret); err != nil {
+				return err
+			}
+			pullSecretData := globalPullSecret.Data[corev1.DockerConfigJsonKey]
+			_, _, _, err := registryclient.GetMetadata(ctx, dummyImageTagMultiarch, pullSecretData)
+			if err != nil {
+				return err
```
enxebre: if the expectation is for this to fail, it should return nil when metadata fail no?

jparrill:
/retest

Opened a issue to Konflux: https://redhat-internal.slack.com/archives/C04PZ7H0VA8/p1753298564987559

PR #6257, opened 2025-06-12
OCPBUGS-57331: Set default security context on everything

smrtrfszm:
**What this PR does / why we need it**: It sets the default security context on everything not just deployments, so by default nothing runs as root.

For example:
The `FeatureGateGenerationJob` reconciler sets the flag to `SetDefaultSecurityContext` here: https://github.com/openshift/hypershift/blob/a7e0d051d548c0eb3765c324890fa9ef0bbcd383/control-plane-operator/controllers/hostedcontrolplane/fg/job.go#L82

However, the `ApplyToJob` does not process this flag:
https://github.com/openshift/hypershift/blob/a7e0d051d548c0eb3765c324890fa9ef0bbcd383/support/config/deployment.go#L191-L201

Only the `ApplyTo` does which only applies deployments:
https://github.com/openshift/hypershift/blob/a7e0d051d548c0eb3765c324890fa9ef0bbcd383/support/config/deployment.go#L112-L116

This results in a pod that do not have the security contexts set:
```shell
$ oc get pod featuregate-generator-h4f2p -n master-molecule1 -o=jsonpath='{.spec.securityContext}'
{}
```

So this PR makes every `ApplyTo` method respect the `SetDefaultSecurityContext` flag.

After this change:
```shell
$ oc get pod featuregate-generator-zvk42 -n master-molecule1 -o=jsonpath='{.spec.securityContext}'
{"runAsUser":1001}
```

While testing this change I noticed that the `TestReconcileFeatureGateGenerationJob` test did not test anything, so I fixed it and extended it to use the default security context as well.

Fixes #https://issues.redhat.com/browse/OCPBUGS-57331

**Checklist**


jeffnowicki:
/re-title OCPBUGS-57331: Set default security context on everything

jeffnowicki:
/retitle OCPBUGS-57331: Set default security context on everything

support/controlplane-component/defaults.go
```diff
@@ -84,11 +84,15 @@ func (c *controlPlaneWorkload[T]) defaultOptions(cpContext ControlPlaneContext,
 	enforceTerminationMessagePolicy(podTemplateSpec.Spec.Containers)
 
 	deploymentConfig := &config.DeploymentConfig{
-		SetDefaultSecurityContext: cpContext.SetDefaultSecurityContext,
 		AdditionalLabels: map[string]string{
 			hyperv1.ControlPlaneComponentLabel: c.Name(),
 		},
 	}
+	// set default security context for the pod.
```
rtheis, line 91: I'm confused as to why this isn't a separate bug that needs to be fixed independent of your changes. I see it fixed on main, but why does your PR need it? I'm not saying it isn't the right thing to do, but it may need to be done in a separate PR for a separate Jira describing this bug.

support/config/deployment.go
```diff
@@ -151,6 +146,8 @@ func (c *DeploymentConfig) ApplyTo(deployment *appsv1.Deployment) {
 }
 
 func (c *DeploymentConfig) ApplyToDaemonSet(daemonset *appsv1.DaemonSet) {
+	c.setDefaultSecurityContextForPod(&daemonset.Spec.Template.Spec)
```
rtheis, line 149: I'm concerned that you are changing behavior only for 4.19. It may be the right thing to do in general but this should be done on main branch first then cherry-picked. For this PR, I recommend that you focus solely on fixing the feature gate job and defer setting default security context on everything to a separate work item since it isn't needed. To that end, couldn't you simply fix ReconcileFeatureGateGenerationJob?

support/controlplane-component/defaults.go
smrtrfszm: Before 4.19 the only thing that is applied with the ApplyTo functions, that I modified, is the etcd, which need root. But because of this patch a `runAsUser` will be applied to the etcd and the AKS tests fail. That's way I backported this fix.

support/config/deployment.go
smrtrfszm: From 4.20 hypershift uses the CPO v2 which does not have this issue, and before 4.19 it wasn't an issue, because the only thing that was applied with these functions is the etcd, which needs root because of an AKS issue. But from 4.19 there's other things as well, so this is a 4.19 only issue.

I could separate adding the `runAsUser` to only the `ApplyToJob` function so that the featuregate-generator job run as non-root, but I don't think it make sense to make a partial fix to only that job. Adding it all ApplyTo functions makes the `SetDefaultSecurityContext` flag work as intended and matches what is on the main branch.

Also these files were removed from main branch because these are CPO v1 so I can't backport it from there.

support/controlplane-component/defaults.go
rtheis: It appears that the fix was part of the larger refactoring for CPOv2 and cherry-picking is not applicable.

support/config/deployment.go
rtheis: Thanks for clarifying. I don't have any concerns with this PR as-is.

rtheis: You all may want to check if a conflicting run as user is set on the Konnectivity agent daemonset with these changes. I see that Konnectivity agent daemonset hard codes `runAsUser` to `1000`. I don't think it will be a problem using `DefaultSecurityContextUser` instead.

smrtrfszm: This is not a problem because the Konnectivity agent daemonset doesn't set the `SetDefaultSecurityContext` flag, so the default security context won't be applied, meaning the `runAsUser` will stay `1000`.

See for example the artifacts for the `ci/prow/e2e-aks` test: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_hypershift/6257/pull-ci-openshift-hypershift-release-4.19-e2e-aks/1949893592541564928/artifacts/e2e-aks/hypershift-azure-run-e2e/artifacts/TestCreateCluster/hostedcluster.tar
In the `namespaces/kube-system/pods/konnectivity-agent-dbv96/konnectivity-agent-dbv96.yaml` file:
```yaml
  securityContext:
    runAsUser: 1000
```

sjenning:
This does not change the UID for etcd, which is the only HCP component where PV ownership could be an issue.

This has no impact for HostedClusters running on OCP and OCP SCCs set the PSC for all pods.

/approve
/label backport-risk-assessed
/label cherry-pick-approved

PR #6268, opened 2025-06-15
test

LiangquanLi930:
test


test/e2e/autoscaling_test.go
```diff
@@ -56,11 +59,49 @@ func TestAutoscaling(t *testing.T) {
 			Max: max,
 		}
 		nodepool.Spec.Replicas = nil
+
+		// Enable HostedCluster Autoscaling
```
celebdor:
```suggestion
		// Enable HostedCluster downscaling, set expanders and ignore labels
```

celebdor (changes requested):
Shouldn't it also test the ignoring of the label?

PR #6105, opened 2025-05-01
CNTRLPLANE-800: Remove code associated with Managed Identity v2 for ARO HCP

bryan-cox:
**What this PR does / why we need it**:
Overall, this PR removes code associated with Managed Identity v2. Specifically, this PR:
- updates the reconciliation of any SecretProviderClass CR to only use MIv3
- removes the ability to use ClientCertificate Authentication when reconciling the AzureClusterIdentity
- removes Azure constants associated with Managed Identity v2.

**Which issue(s) this PR fixes** *(optional, use `fixes #<issue_number>(, fixes #<issue_number>, ...)` format, where issue_number might be a GitHub issue, or a Jira story*:
Fixes #

**Checklist**
- [x] Subject and description added to both, commit and PR.
- [x] Relevant issues have been referenced.
- [ ] This change includes docs. 
- [ ] This change includes unit tests.

PR #6283, opened 2025-06-16
CNTRLPLANE-989: Initial Changes to Support Self-managed Azure

bryan-cox:
This PR contains the initial changes needed to support self-managed Azure Hosted Clusters. This PR is only tested for managed Azure cases to ensure we do not break these use cases. Further testing and development is still needed for self-managed Azure, which is outside the scope of this PR.

**Notes for Reviewers**:
This PR cannot merge until:
1. ARO HCP to move to using OCP 4.19.0
2. This PR merges first https://github.com/openshift/hypershift/pull/6105

Fixes [CNTRLPLANE-989](https://issues.redhat.com//browse/CNTRLPLANE-989)

**Checklist**


api/hypershift/v1beta1/azure.go
```diff
@@ -424,9 +424,19 @@ type AzurePlatformSpec struct {
 
 	// managedIdentities contains the managed identities needed for HCP control plane and data plane components that
 	// authenticate with Azure's API.
+	// 
+	// These are required for managed Azure, also known as ARO HCP.
```
enxebre: please enforce this validations with CEL. If one is set then other can't be set

bryan-cox: For reference for others - https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md#writing-a-union-in-go

bryan-cox: Updated

bryan-cox:
/hold

See PR description for the why

bryan-cox:
/hold cancel

The prereqs for this PR have been met.

bryan-cox:
/test all

Finished rebasing

api/hypershift/v1beta1/azure.go
```diff
@@ -636,3 +679,41 @@ type AzureKMSKey struct {
...
+	// 
+	// "ManagedIdentities" means that the Hosted Cluster is using managed identities to authenticate with Azure's API. 
+	// This is only valid for managed Azure, also known as ARO HCP.
+	//
+	// "WorkloadIdentities" means that the Hosted Cluster is using workload identities to authenticate with Azure's API. 
+	// This is only valid for self-managed Azure.
+	// 
+	// unionDiscriminator
+	// +kubebuilder:validation:Enum=ManagedIdentities;WorkloadIdentities	
+	// +required
```
enxebre, line 719: where is the cel validation for this discrimination?
also please add a fail test case within TestOnCreateAPIUX

api/hypershift/v1beta1/azure.go
```diff
@@ -479,6 +479,49 @@ type AzureResourceManagedIdentities struct {
 	// +required
 	DataPlane DataPlaneManagedIdentities `json:"dataPlane"`
 }
+type AzureWorkloadIdentities struct {
+	// ciroClientID is the client ID of a federated managed identity, associated with cluster-image-operator, used in 
+	// workload identity authentication.
```
enxebre, line 495: why does the AzureWorkloadIdentities are named different thatn the ones in AzureResourceManagedIdentities, don't we have the same set of components?

api/hypershift/v1beta1/azure.go
```diff
@@ -422,11 +422,11 @@ type AzurePlatformSpec struct {
 	// +immutable
 	SecurityGroupID string `json:"securityGroupID"`
 
-	// managedIdentities contains the managed identities needed for HCP control plane and data plane components that
-	// authenticate with Azure's API.
+	// azureAuthenticationConfig is the type of Azure authentication configuration to use to authenticate with Azure's
```
enxebre: ptal @JoelSpeed

api/hypershift/v1beta1/azure.go
```diff
@@ -636,3 +679,43 @@ type AzureKMSKey struct {
 	// +required
 	KeyVersion string `json:"keyVersion"`
 }
+
+// azureAuthenticationConfiguration is a discriminated union type that contains the Azure authentication configuration
+// for a Hosted Cluster. This configuration is used to determine how the Hosted Cluster authenticates with Azure's API,
+// either with managed identities or workload identities.
+//
+// +kubebuilder:validation:XValidation:rule="self.azureAuthenticationConfigType == 'ManagedIdentities' ? has(self.managedIdentities) : !has(self.managedIdentities)", message="managedIdentities is required when azureAuthenticationConfigType is ManagedIdentities, and forbidden otherwise"
+// +kubebuilder:validation:XValidation:rule="self.azureAuthenticationConfigType == 'WorkloadIdentities' ? has(self.workloadIdentities) : !has(self.workloadIdentities)", message="workloadIdentities is required when azureAuthenticationConfigType is WorkloadIdentities, and forbidden otherwise"
```
enxebre, line 711: is this cel validation right? doesn't seem to make ManagedIdentities struct required when the type is Type is ManagedIdentities

api/hypershift/v1beta1/azure.go
```diff
@@ -479,6 +479,49 @@ type AzureResourceManagedIdentities struct {
 	// +required
 	DataPlane DataPlaneManagedIdentities `json:"dataPlane"`
 }
+type AzureWorkloadIdentities struct {
+	// imageRegistryClientID is the client ID of a federated managed identity, associated with cluster-image-registry-operator, used in
+	// workload identity authentication.
+	// +required
+	// +kubebuilder:validation:XValidation:rule="self.matches('^[{]?[0-9a-fA-F]{8}-([0-9a-fA-F]{4}-){3}[0-9a-fA-F]{12}[}]?$')",message="the client ID of a managed identity must be a valid UUID. It should be 5 groups of hyphen separated hexadecimal characters in the form 8-4-4-4-12."
+	// +kubebuilder:validation:MaxLength=255
```
enxebre: we might want to have a dedicated workloadIdentity struct with the a field for clientID.
Let's not block the PR on this for the sake of moving forward for the managed changes. We will keep iterating for self managed

api/hypershift/v1beta1/azure.go
enxebre: nvm I missed the first block, also the e2e cover this now.

api/hypershift/v1beta1/azure.go
```diff
@@ -636,3 +679,43 @@ type AzureKMSKey struct {
...
+	// +required
+	AzureAuthenticationConfigType string `json:"azureAuthenticationConfigType"`
+
+	// managedIdentities contains the managed identities needed for HCP control plane and data plane components that
+	// authenticate with Azure's API.
+	//
+	// These are required for managed Azure, also known as ARO HCP.
+	//
+	// +optional
+	ManagedIdentities *AzureResourceManagedIdentities `json:"managedIdentities,omitempty"`
```
enxebre, line 728: we'll want to look into using omitzero and set a req on clients for go 1.24. For now this is ok

api/hypershift/v1beta1/azure.go
bryan-cox: I think so. It matches the pattern I see elsewhere like https://github.com/openshift/hypershift/blob/c628825a087587db83ad5d7b667b2c17bcef2260/api/hypershift/v1beta1/azure.go#L107 and https://github.com/openshift/hypershift/blob/5a3972d043d9e255d0abac1f9dcf1696158e6e44/api/hypershift/v1beta1/hostedcluster_types.go#L1463

api/hypershift/v1beta1/azure.go
```diff
@@ -422,11 +422,11 @@ type AzurePlatformSpec struct {
...
 	SecurityGroupID string `json:"securityGroupID"`
 
-	// managedIdentities contains the managed identities needed for HCP control plane and data plane components that
-	// authenticate with Azure's API.
+	// azureAuthenticationConfig is the type of Azure authentication configuration to use to authenticate with Azure's
+	// Cloud API.
 	//
 	// +required
-	ManagedIdentities AzureResourceManagedIdentities `json:"managedIdentities"`
+	AzureAuthenticationConfig AzureAuthenticationConfiguration `json:"azureAuthenticationConfig"`
```
JoelSpeed, line 429: Has this API shipped? You should create a tombstone that shows what `managedIdentities` looked like so that if we ever decided to reintroduce a field with this name, the serialization would be the same

api/hypershift/v1beta1/azure.go
```diff
@@ -479,6 +479,49 @@ type AzureResourceManagedIdentities struct {
 	// +required
 	DataPlane DataPlaneManagedIdentities `json:"dataPlane"`
 }
+type AzureWorkloadIdentities struct {
+	// imageRegistryClientID is the client ID of a federated managed identity, associated with cluster-image-registry-operator, used in
+	// workload identity authentication.
+	// +required
+	// +kubebuilder:validation:XValidation:rule="self.matches('^[{]?[0-9a-fA-F]{8}-([0-9a-fA-F]{4}-){3}[0-9a-fA-F]{12}[}]?$')",message="the client ID of a managed identity must be a valid UUID. It should be 5 groups of hyphen separated hexadecimal characters in the form 8-4-4-4-12."
```
JoelSpeed: If the validation is the same across all of these fields, it's best to create a string type alias and re-use that across each field no?

api/hypershift/v1beta1/azure.go
JoelSpeed, line 429: This should have `omitzero` even if it's required, are you using Go 1.24 yet?

api/hypershift/v1beta1/azure.go
JoelSpeed: Looks like the right thing to me

api/hypershift/v1beta1/azure.go
```diff
@@ -636,3 +679,43 @@ type AzureKMSKey struct {
...
+	// used to determine which identity configuration is being used. Valid values are "ManagedIdentities" and
+	// "WorkloadIdentities".
+	//
+	// "ManagedIdentities" means that the Hosted Cluster is using managed identities to authenticate with Azure's API.
+	// This is only valid for managed Azure, also known as ARO HCP.
+	//
+	// "WorkloadIdentities" means that the Hosted Cluster is using workload identities to authenticate with Azure's API.
+	// This is only valid for self-managed Azure.
+	//
+	// unionDiscriminator
```
JoelSpeed:
```suggestion
	// +unionDiscriminator
```

api/hypershift/v1beta1/azure.go
```diff
@@ -636,3 +679,43 @@ type AzureKMSKey struct {
...
+	// +kubebuilder:validation:Enum=ManagedIdentities;WorkloadIdentities
```
JoelSpeed: Generally good practice to create a string type alias with constants for the possible values, it's useful for users of the API to be able to compare to the constants

api/hypershift/v1beta1/azure.go
```diff
@@ -636,3 +679,43 @@ type AzureKMSKey struct {
...
+	// authenticate with Azure's API.
+	//
+	// These are required for managed Azure, also known as ARO HCP.
+	//
+	// +optional
+	ManagedIdentities *AzureResourceManagedIdentities `json:"managedIdentities,omitempty"`
+
+	// workloadIdentities is a slice of objects containing a component and a client ID of a federated managed identity
+	// used in workload identity authentication. These are used to authenticate with Azure cloud on both the control
+	// plane and data plane.
```
JoelSpeed: Is it a slice?

JoelSpeed (commented):
I would expect it to be advantageous to align the shape of both the worload identities and managed identities APIs, but at the moment they look very different, any reason not to make them more similar with each of the different services pointing to a struct with the clientID and potentially other future fields in?

api/hypershift/v1beta1/azure.go
bryan-cox: Reolved on Slack. I updated the names but kept the ClientID suffix for now.

api/hypershift/v1beta1/azure.go
bryan-cox: Yeah but we can follow up with that I believe - https://github.com/openshift/hypershift/pull/6283#discussion_r2219538345

bryan-cox: No, not in a customer sense yet. We are still able to make breaking changes atm.

bryan-cox:
>I would expect it to be advantageous to align the shape of both the worload identities and managed identities APIs, but at the moment they look very different, any reason not to make them more similar with each of the different services pointing to a struct with the clientID and potentially other future fields in?

@JoelSpeed - ARO HCP is kind of a special case where the OpenShift operator and operands need to use different Azure creds. We don't do that for ROSA HCP and will not for self-managed Azure when we get to that. So we don't need a delineation between control plane managed identities and data plane identities for self-managed Azure.

This PR is just to introduce the bare minimum changes needed to allow ARO HCP to continue towards GA while allowing us the new pathways of implementing self-managed Azure. The self-managed Azure is subject to change as we develop it and get it working.

api/hypershift/v1beta1/azure.go
JoelSpeed: I only semi agree with that statement.

Please consider that you have no control over which version of this repository a customer will vendor when building tooling to interact with this API.

You can make the change you are making here, because you're actually dropping a field and adding a new one, fine, but please don't assume that you can change the serialization of things, if you do this, and someone vendors a commit prior to the change, it will cause issues during encoding and decoding which is not what we want

JoelSpeed:
> So we don't need a delineation between control plane managed identities and data plane identities for self-managed Azure.

The new fields are for the self managed Azure case? 

> This PR is just to introduce the bare minimum changes needed to allow ARO HCP to continue towards GA while allowing us the new pathways of implementing self-managed Azure. The self-managed Azure is subject to change as we develop it and get it working.

I was actually thinking at a lower level than this in terms of the API, you currently have
```
fooClientID: abc
```

But in the other similar API, it looks more like:
```
foo:
  clientID: abc
```

Why not align this part of the shape (even without the control plane/data plane deliniation)

api/hypershift/v1beta1/azure.go
JoelSpeed: Have you captured following up on this in some sort of story that's been prioritised?

api/hypershift/v1beta1/azure.go
JoelSpeed: `ClientID` suffix is now gone, but these are still strings. This is very very risky.

If we ever need to evolve this API, you cannot change the serialization of the fields using the same names once this PR is merged.

IMO, these should be structs with a single field so that, if we do ever get to a point where we want to add a sibling to the clientID field, we have the ability to do so without making breaking changes/choosing new names

api/hypershift/v1beta1/azure.go
bryan-cox: No one should be vendoring the code from a self-managed Azure perspective because we don't currently support it.

Only managed services in Red Hat are vendoring the API and we are coordinating the changes with them as there are no long lived clusters yet since the managed services part is still in development.

JoelSpeed: > No one should be vendoring the code from a self-managed Azure perspective because we don't currently support it.

You cannot guarantee that no one will ever vendor a particular merged part of the code. Just because they shouldn't be doing it, doesn't stop them. Perhaps in the future you do support the self managed and someone decides to vendor, and for some reason, they pick up an older commit, boom, they're screwed

enxebre: >This should have omitzero even if it's required, are you using Go 1.24 yet?
>Yeah but we can follow up with that I believe - #6283 (comment)

I don't  think we can guarantee our consumer clients are on 1.24 hence my suggestion to postpone this.

enxebre:
/lgtm

feel free to hold if you still have more feedback @JoelSpeed

api/hypershift/v1beta1/azure.go
```diff
@@ -480,16 +480,55 @@ type AzureResourceManagedIdentities struct {
 	DataPlane DataPlaneManagedIdentities `json:"dataPlane"`
 }
 
+// AzureClientID is a string that represents the client ID of a managed identity.
+//
+// +kubebuilder:validation:XValidation:rule="self.matches('^[{]?[0-9a-fA-F]{8}-([0-9a-fA-F]{4}-){3}[0-9a-fA-F]{12}[}]?$')",message="the client ID of a managed identity must be a valid UUID. It should be 5 groups of hyphen separated hexadecimal characters in the form 8-4-4-4-12."
+// +kubebuilder:validation:MaxLength=255
```
JoelSpeed: This length isn't correct is it? Don't these IDs have a fixed length? Also, probably worth setting a minimum length too, IIUC, in this case, it would be the same value for minimum and maximum?

api/hypershift/v1beta1/azure.go
JoelSpeed: I would expect your Go deps to rely on something `k8s.io` which when updated to 1.33 would require 1.24 as a minimum go version, which will then put the same requirement on your consumers

api/hypershift/v1beta1/azure.go
bryan-cox: Yeah that makes sense

PR #6294, opened 2025-06-18
CNTRLPLANE-980: chore(contrib): add ansible playbook for setting up external OIDC

devguyio:
Add Ansible playbooks for setting up and configuring an OpenID Connect (OIDC) provider using Dex with LDAP backend. This setup can be used to test and validate OIDC authentication flows.

Check out the [README](https://github.com/openshift/hypershift/blob/a1c03ca8c11d662723595bfa3d7f68dbcdcdde40/hack/dev/oidc/README.md)

**Checklist**


contrib/oidc/playbooks/oidc_configure.yml
```diff
@@ -0,0 +1,428 @@
...
+          - "{{ dex_tls_dir }}:/etc/dex/tls:ro,Z"
+          - "{{ dex_dir }}/data:/var/dex:rw,Z"
+        ports:
+          - "80:5556"
+          - "443:5554"
+        user: "{{ dex_uid }}:{{ dex_uid }}"
+        # network: host
+        state: started
+
+    - name: Wait for Dex container to start
```
csrwng: Is this because this is the only container in the instance? Would it be more robust if you wait for a health check URL to succeed?

csrwng (commented):
Looks good to me. Just curious though... wouldn't it be easier to set these up on an existing openshift/kubernetes cluster?

contrib/oidc/playbooks/oidc_configure.yml
devguyio: good catch. fixed

devguyio:
> Looks good to me. Just curious though... wouldn't it be easier to set these up on an existing openshift/kubernetes cluster?

I guess it depends on the goal. For debugging / local testing, I don't think pods would add much value. I think one just needs a process running and a config file you can change without bothering with all the moving pieces for a kube setup (configmap, pod, volumes, service, ingress etc). 

The reason Dex is running as a container in this setup, is because Dex no longer publishes binaries, it only publishes containers.

mgencur:
LGTM

PR #6297, opened 2025-06-18
OCPBUGS-56701: Fix LookupDefaultOCPVersion to use the supported-versions ConfigMap to find a default release image

bryan-cox:
In general, the main purpose of this PR is to update the `LookupDefaultOCPVersion` so that when no release stream is passed into the function, the function will look up the latest stable multi-arch OCP release image supported by the HyperShift Operator. The version is determined by the `supported-versions` ConfigMap. Previously, the `LookupDefaultOCPVersion` would use the latest OCP version at  https://multi.ocp.releases.ci.openshift.org/api/v1/releasestream/4-stable-multi/latest which could include newer OCP versions than what is supported by the currently installed HyperShift Operator.

In addition, this PR cleans up, consolidates some code, and some other miscellaneous things:
- Cleans up the HyperShift CLI version file. Removed code was not really relevant to this CLI command and was consolidated into the existing version.go file in the support package.
- Consildated all OCP version related functions to the existing version.go file in the support package.
- Added some unit tests
- Updated existing HyperShift CLI upstream docs

Fixes [OCPBUGS-56701](https://issues.redhat.com/browse/OCPBUGS-56701)

**Checklist**

### DIFF
--- a/support/supportedversion/version.go
+++ b/support/supportedversion/version.go
@@ -250,14 +250,16 @@ type SupportedVersions struct {
 // supported versions and the server version from the specified namespace and unmarshals the versions into a
 // SupportedVersions struct. If the ConfigMap or the required keys are not found, it returns an error. The function
 // returns the supported versions, the server version, and any error encountered during the process.
-func GetSupportedOCPVersions(ctx context.Context, namespace string, client crclient.Client) (SupportedVersions, string, error) {
+func GetSupportedOCPVersions(ctx context.Context, namespace string, client crclient.Client, supportedVersions *corev1.ConfigMap) (SupportedVersions, string, error) {
 	var versions SupportedVersions
 	var serverVersion string
 
-	// Fetch the supported versions ConfigMap from the specified namespace
-	supportedVersions := manifests.ConfigMap(namespace)
-	if err := client.Get(ctx, crclient.ObjectKeyFromObject(supportedVersions), supportedVersions); err != nil {
-		return SupportedVersions{}, "", fmt.Errorf("failed to find supported versions on the server: %v", err)
+	if supportedVersions == nil {
+		// Fetch the supported versions ConfigMap from the specified namespace
+		supportedVersions := manifests.ConfigMap(namespace)
+		if err := client.Get(ctx, crclient.ObjectKeyFromObject(supportedVersions), supportedVersions); err != nil {
+			return SupportedVersions{}, "", fmt.Errorf("failed to find supported versions on the server: %v", err)
+		}
 	}
 
 	// Check if the ConfigMap contains the server version key
@@ -288,6 +290,7 @@ type ocpTags struct {
 // the latest stable release images from the provided release URL, and returns the latest supported OCP version that is
 // not a release candidate and matches the latest supported OCP version supported by the HyperShift operator.
 func retrieveSupportedOCPVersion(ctx context.Context, releaseURL string, client crclient.Client) (ocpVersion, error) {
+	var supportedVersions *corev1.ConfigMap
 	var stableOCPVersions ocpTags
 	var namespace string
 
@@ -299,6 +302,7 @@ func retrieveSupportedOCPVersion(ctx context.Context, releaseURL string, client 
 	}
 	for _, configMap := range configMapList.Items {
 		if configMap.Name == "supported-versions" {
+			supportedVersions = &configMap
 			namespace = configMap.Namespace
 			break
 		}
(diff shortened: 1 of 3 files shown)


support/supportedversion/version.go
```diff
@@ -163,3 +220,113 @@ func LookupLatestSupportedRelease(ctx context.Context, hc *hyperv1.HostedCluster
...
+}
+
+// retrieveSupportedOCPVersion retrieves the latest supported OCP version from supported versions ConfigMap, retrieves
+// the latest stable release images from the provided release URL, and returns the latest supported OCP version that is
+// not a release candidate and matches the latest supported OCP version supported by the HyperShift operator.
+func retrieveSupportedOCPVersion(releaseURL string, client crclient.Client) (ocpVersion, error) {
+	var stableOCPVersions ocpTags
+
+	// Get the latest supported OCP version from the supported versions ConfigMap
+	supportedOCPVersions, _, err := GetSupportedOCPVersions(context.TODO(), "hypershift", client)
```
bryan-cox, line 293: I'm not sure how to get around not hardcoding this to `hypershift` for the support versions ConfigMap 🤔 

This function is called by `LookupDefaultOCPVersion`, which is called in several places where we would not know the namespace where the HyperShift Operator is installed.

support/supportedversion/version_test.go
```diff
@@ -198,3 +207,202 @@ func TestGetMinSupportedVersion(t *testing.T) {
...
+			} else {
+				g.Expect(err).ToNot(HaveOccurred())
+				g.Expect(supportedVersions).To(Equal(tc.expectedVersions))
+				g.Expect(serverVersion).To(Equal(tc.expectedServerVersion))
+			}
+		})
+	}
+}
+
+func TestRetrieveSupportedOCPVersion(t *testing.T) {
```
enxebre, line 319: can we please not introduce unit tests which does http requests? this needs to be mocked

support/supportedversion/version.go
```diff
@@ -163,3 +221,131 @@ func LookupLatestSupportedRelease(ctx context.Context, hc *hyperv1.HostedCluster
...
+	var namespace string
+
+	// Find the supported versions ConfigMap since it may be in a different namespace than the default "hypershift"
+	configMapList := &corev1.ConfigMapList{}
+	err := client.List(context.TODO(), configMapList)
+	if err != nil {
+		return ocpVersion{}, fmt.Errorf("failed to list ConfigMaps to find supported versions: %v", err)
+	}
+	for _, configMap := range configMapList.Items {
+		if configMap.Name == "supported-versions" {
```
enxebre, line 304: aren't we labeling or annotating this CM with something that can be used to filter here?

support/supportedversion/version.go
```diff
@@ -163,3 +221,131 @@ func LookupLatestSupportedRelease(ctx context.Context, hc *hyperv1.HostedCluster
...
+// retrieveSupportedOCPVersion retrieves the latest supported OCP version from supported versions ConfigMap, retrieves
+// the latest stable release images from the provided release URL, and returns the latest supported OCP version that is
+// not a release candidate and matches the latest supported OCP version supported by the HyperShift operator.
+func retrieveSupportedOCPVersion(releaseURL string, client crclient.Client) (ocpVersion, error) {
+	var stableOCPVersions ocpTags
+	var namespace string
+
+	// Find the supported versions ConfigMap since it may be in a different namespace than the default "hypershift"
+	configMapList := &corev1.ConfigMapList{}
+	err := client.List(context.TODO(), configMapList)
```
muraee: we should pass a proper context to this function and use it.

support/supportedversion/version_test.go
bryan-cox: Updated

support/supportedversion/version.go
bryan-cox: Good call. I added a filter on the `client.List` to search by the label.

I did not change the following for loop. I wasn't sure if the for loop is any better than looking to see if the list has one object and assuming it's the right ConfigMap. Happy to change this though if you have a preference.

enxebre:
lgtm, I'll defer on @muraee to tag

support/supportedversion/version.go
muraee: why are we fetching the configMap here, only saving the namespace and then call `GetSupportedOCPVersions` to fetch the configMap again.

bryan-cox: `GetSupportedOCPVersions` has an additional place where it is called where the ConfigMap was not previously fetched. I've updated the function to take an existing ConfigMap; if none was passed in, it will find it as before.

bryan-cox: PR should be updated now @muraee

PR #6312, opened 2025-06-23
NO-JIRA: Add Azure DNS Zone Clean-up Instructions

bryan-cox:
This PR:
- reorganizes the Azure CI folder in contrib
- adds a README.md to the Azure CI folder in contrib
- adds the instructions with scripts to help a user delete the DNS Zone Recordsets in CI.

N/A

**Checklist**


contrib/ci/Azure/README.md
```diff
@@ -0,0 +1,65 @@
...
+2. Use the provided YAML configuration in `hypershift-install-azure.yaml` for HyperShift installation
+3. Refer to Azure cloud resource management documentation as needed
+
+## Prerequisites
+
+- OpenShift CLI (oc)
+- OCP cluster (ROSA)
+- HyperShift CLI
+- Azure credentials (available from Vault or generated following Microsoft documentation)
+- Access to OpenShift CI Vault for credential storage
```
jparrill, line 53: IIRC do we also need `ccoctl`, am I right?

bryan-cox: Hmm not yet. This is the self-managed Azure setup instructions Patryk put together when we were testing on OpenShift clusters.

PR #6317, opened 2025-06-24
ACM-21713 adding .git in .dockerignore to fix version info

rokej:
This PR adds `.git` back to the `.dockerignore` file so that vsc build information gets added to the binaries. Without this change, you get <unknown> version information for the server and the client.

Fixes #https://issues.redhat.com/browse/ACM-21713

**Checklist**


sjenning:
/hold

Until we figure out what is happening in the konflux build.  Seems like a legit break.
```
[1/2] STEP 2/4: WORKDIR /hypershift
[1/2] STEP 3/4: COPY . .
[1/2] STEP 4/4: RUN make hypershift   && make hypershift-operator   && make product-cli   && make karpenter-operator
CGO_ENABLED=0 GO111MODULE=on GOWORK=off GOFLAGS=-mod=vendor go build -gcflags=all='-N -l' -ldflags '-extldflags "-static"' -o bin/hypershift .
error obtaining VCS status: exit status 128
	Use -buildvcs=false to disable VCS stamping.
make: *** [Makefile:137: hypershift] Error 1
```

celebdor (changes requested):
How does it get version information without the .git directory?

In order to fix the build with go-toolset I found that just doing the copy with chown to the default go-toolset user fixes it:
`
-COPY . .
`
`
+COPY --chown=default . .
`

sjenning:
> How does it get version information without the .git directory?

It doesn't.  That's what this PR is fixing.  It is _not_ ignoring the `.git` directory on COPY.

celebdor:
> > How does it get version information without the .git directory?
> 
> It doesn't. That's what this PR is fixing. It is _not_ ignoring the `.git` directory on COPY.

So then the --chown flag is all that's needed, and keep using the red hat catalog image

rokej:
Verified the fix with `quay.io/redhat-user-workloads/crt-redhat-acm-tenant/hypershift-operator/hypershift-operator-main:on-pr-8c89d7d842f801ab543f537f073d8bfb704ad54e`

celebdor (approved):
/lgtm

PR #6319, opened 2025-06-25
MGMT-20826: Enable disabling of Ingress capability

linoyaslan:
This PR is part of the task for expanding support for disabling openshift capabilities in hypershift. This change adds support for disabling the Ingress capability, who managed by both CPO and CVO. Additional capabilities will be supported in future updates.

**Checklist**

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
@@ -177,7 +177,7 @@ type HostedControlPlaneReconciler struct {
 	azureCredentialsLoaded                  sync.Map
 }
 
-func (r *HostedControlPlaneReconciler) SetupWithManager(mgr ctrl.Manager, createOrUpdate upsert.CreateOrUpdateFN) error {
+func (r *HostedControlPlaneReconciler) SetupWithManager(mgr ctrl.Manager, createOrUpdate upsert.CreateOrUpdateFN, hcp *hyperv1.HostedControlPlane) error {
 	r.setup(createOrUpdate)
 	b := ctrl.NewControllerManagedBy(mgr).
 		For(&hyperv1.HostedControlPlane{}).
@@ -200,11 +200,12 @@ func (r *HostedControlPlaneReconciler) SetupWithManager(mgr ctrl.Manager, create
 
 	r.ec2Client, r.awsSession = GetEC2Client()
 
-	r.registerComponents()
+	r.registerComponents(hcp)
 	return nil
 }
 
-func (r *HostedControlPlaneReconciler) registerComponents() {
+func (r *HostedControlPlaneReconciler) registerComponents(hcp *hyperv1.HostedControlPlane) {
+
 	r.components = append(r.components,
 		pkioperatorv2.NewComponent(r.CertRotationScale),
 		etcdv2.NewComponent(),
@@ -221,7 +222,7 @@ func (r *HostedControlPlaneReconciler) registerComponents() {
 		oauthv2.NewComponent(),
 		routecmv2.NewComponent(),
 		clusterpolicyv2.NewComponent(),
-		configoperatorv2.NewComponent(r.ReleaseProvider.GetRegistryOverrides(), r.ReleaseProvider.GetOpenShiftImageRegistryOverrides()),
+		configoperatorv2.NewComponent(r.ReleaseProvider.GetRegistryOverrides(), r.ReleaseProvider.GetOpenShiftImageRegistryOverrides(), hcp.Spec.Capabilities),
 		awsccmv2.NewComponent(),
 		azureccmv2.NewComponent(),
 		kubevirtccmv2.NewComponent(),
@@ -1734,14 +1735,16 @@ func (r *HostedControlPlaneReconciler) reconcilePKI(ctx context.Context, hcp *hy
 		return fmt.Errorf("failed to reconcile root CA: %w", err)
 	}
 
-	observedDefaultIngressCert := manifests.IngressObservedDefaultIngressCertCA(hcp.Namespace)
-	if err := r.Get(ctx, client.ObjectKeyFromObject(observedDefaultIngressCert), observedDefaultIngressCert); err != nil {
-		if !apierrors.IsNotFound(err) {
-			return fmt.Errorf("failed to get observed default ingress cert: %w", err)
+	var observedDefaultIngressCert *corev1.ConfigMap
+	if capabilities.IsIngressCapabilityEnabled(hcp.Spec.Capabilities) {
+		observedDefaultIngressCert = manifests.IngressObservedDefaultIngressCertCA(hcp.Namespace)
+		if err := r.Get(ctx, client.ObjectKeyFromObject(observedDefaultIngressCert), observedDefaultIngressCert); err != nil {
+			if !apierrors.IsNotFound(err) {
+				return fmt.Errorf("failed to get observed default ingress cert: %w", err)
+			}
+			observedDefaultIngressCert = nil
 		}
-		observedDefaultIngressCert = nil
 	}
-
 	rootCAConfigMap := manifests.RootCAConfigMap(hcp.Namespace)
 	if _, err := createOrUpdate(ctx, r, rootCAConfigMap, func() error {
 		return pki.ReconcileRootCAConfigMap(rootCAConfigMap, p.OwnerRef, rootCASecret, observedDefaultIngressCert)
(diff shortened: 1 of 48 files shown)


linoyaslan:
WIP

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -1926,12 +1929,15 @@ func (r *HostedControlPlaneReconciler) reconcilePKI(ctx context.Context, hcp *hy
...
-	// Ingress Cert
-	ingressCert := manifests.IngressCert(hcp.Namespace)
-	if _, err := createOrUpdate(ctx, r, ingressCert, func() error {
-		return pki.ReconcileIngressCert(ingressCert, rootCASecret, p.OwnerRef, p.IngressSubdomain)
-	}); err != nil {
-		return fmt.Errorf("failed to reconcile ingress cert secret: %w", err)
+	// Reconcile ingress serving certificate only if Ingress capability is enabled.
+	if capabilities.IsCapabilityEnabled(hcp.Spec.Capabilities, hyperv1.IngressCapability) {
+		// Ingress Cert
+		ingressCert := manifests.IngressCert(hcp.Namespace)
```
linoyaslan, line 1942: @enxebre Hi, in the reconciliation code, I guard the creation of IngressCert and observedDefaultIngressCert behind the IngressCapability check. From what I understand, these certs are only used by the ingress controller inside the hosted cluster, so they shouldn't be needed if Ingress is disabled.
Can you confirm if that's correct — or is there any other component (maybe on the management cluster side) that still relies on them even when Ingress is off?

control-plane-operator/hostedclusterconfigoperator/controllers/machine/machine.go
```diff
@@ -65,19 +66,22 @@ func (r *reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 func (r *reconciler) findKubevirtPassthroughServices(ctx context.Context, hcp *hyperv1.HostedControlPlane) ([]corev1.Service, error) {
 	kubevirtPassthroughServices := []corev1.Service{}
 
-	// Manifests for infra/mgmt cluster passthrough service
-	cpService := hcpmanifests.IngressDefaultIngressPassthroughService(kubevirtInfraNamespace(hcp))
+	// Add ingress passthrough service only if IngressCapability is enabled
+	if capabilities.IsCapabilityEnabled(hcp.Spec.Capabilities, hyperv1.IngressCapability) {
```
linoyaslan: @enxebre is this passthrough service is required so external traffic can reach the router inside the hosted cluster? or if is it also required for anything else. Just want to make sure we’re not skipping something critical if Ingress is disabled.

linoyaslan:
**NOTES:** 

This PR is on hold due to - https://github.com/openshift/cluster-network-operator/pull/2731

EDIT: the CNO fix merged.

enxebre:
Disabling console healthchecks should be enough https://github.com/openshift/enhancements/pull/1415/files#diff-43d430f5ea26a817416c9f3026a0a5d975be1500898d782d0f48cd9cbe1df429R218

Can we make sure to sync with the enhacement author and make sure it's update to reflect reality?

This will need and e2e validation for all the resources managed by the cap
```
	// ClusterVersionCapabilityIngress manages the cluster ingress operator
	// which is responsible for running the ingress controllers (including OpenShift router).
	//
	// The following CRDs are part of the capability as well:
	// IngressController
	// DNSRecord
	// GatewayClass
	// Gateway
	// HTTPRoute
	// ReferenceGrant
	//
	// WARNING: This capability cannot be disabled on the standalone OpenShift.
	ClusterVersionCapabilityIngress ClusterVersionCapability = "Ingress"
```

linoyaslan:
@enxebre Thanks, Alberto. I shared this with @jhadvig from the console team, since, iiuc, skipping these checks should be handled in the console-operator repo, correct?

Also, regarding the e2e test: I already added one, but it currently only checks for the absence of the IngressController. I’ll update it to also verify that instances of all the listed resources are not present in the cluster.

support/capabilities/hosted_control_plane_capabilities.go
```diff
@@ -11,6 +11,19 @@ import (
...
+	if capabilities == nil {
+		return true
+	}
+	enabled := true
+	for _, disabledCap := range capabilities.Disabled {
+		if disabledCap == capability {
+			enabled = false
+		}
+	}
+	return enabled
```
muraee: nit: we can return early here
```suggestion
	for _, disabledCap := range capabilities.Disabled {
		if disabledCap == capability {
			return false
		}
	}
	return true
```

muraee:
unit textures need update

hypershift-operator/controllers/hostedcluster/network_policies.go
```diff
@@ -36,16 +36,18 @@ const (
 func (r *HostedClusterReconciler) reconcileNetworkPolicies(ctx context.Context, log logr.Logger, createOrUpdate upsert.CreateOrUpdateFN, hcluster *hyperv1.HostedCluster, hcp *hyperv1.HostedControlPlane, version semver.Version, controlPlaneOperatorAppliesManagementKASNetworkPolicyLabel bool) error {
 	controlPlaneNamespaceName := manifests.HostedControlPlaneNamespace(hcluster.Namespace, hcluster.Name)
 
-	// Reconcile openshift-ingress Network Policy
-	policy := networkpolicy.OpenshiftIngressNetworkPolicy(controlPlaneNamespaceName)
-	if _, err := createOrUpdate(ctx, r.Client, policy, func() error {
-		return reconcileOpenshiftIngressNetworkPolicy(policy)
-	}); err != nil {
-		return fmt.Errorf("failed to reconcile ingress network policy: %w", err)
+	if capabilities.IsIngressCapabilityEnabled(hcluster.Spec.Capabilities) {
```
enxebre: This is network policy for ingress pod to pod communication in the management cluster. This shouldn't be impacted by the guest cluster capability. Let's undo the change.

control-plane-operator/controllers/hostedcontrolplane/v2/ocm/config.go
```diff
@@ -96,9 +96,11 @@ func adaptConfig(cfg *openshiftcpv1.OpenShiftControllerManagerConfig, configurat
 		cfg.Build.BuildOverrides.Tolerations = buildConfig.Spec.BuildOverrides.Tolerations
 	}
 
-	// network config
-	if cidrs := configuration.GetAutoAssignCIDRs(); len(cidrs) > 0 {
-		cfg.Ingress.IngressIPNetworkCIDR = cidrs[0]
+	if capabilities.IsIngressCapabilityEnabled(caps) {
+		// network config
```
enxebre: This is for cloud provider interaction with services of type LoadBalancer, I'm not sure we want to disable it. If we want to we should probably add a CEL rule to prevent you from setting this if ingress is disabled.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/manifests/rbac.go
```diff
@@ -1,10 +1,20 @@
 package manifests
 
 import (
+	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
+
 	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+var RbacCapabilityMap = map[string]hyperv1.OptionalCapability{
```
enxebre, line 10: where's this used?

enxebre: nvm

control-plane-operator/hostedclusterconfigoperator/operator/config.go
```diff
@@ -78,11 +81,58 @@ type HostedClusterConfigOperatorConfig struct {
...
+		&operatorv1.ClusterCSIDriver{}:      allSelector,
+
+		// Needed for inplace upgrader.
+		&corev1.Node{}: allSelector,
+
+		// Needed for resource cleanup
+		&corev1.Service{}:               allSelector,
+		&corev1.PersistentVolume{}:      allSelector,
+		&corev1.PersistentVolumeClaim{}: allSelector,
+		//&operatorv1.IngressController{}: allSelector,
```
enxebre: remove commented code?

enxebre:
dropped some comments, lgtm overall, needs rebase.
/approve

linoyaslan:
> dropped some comments, lgtm overall, needs rebase.

@enxebre Thanks Alberto! All comments resolved. will appreciate your re-review

control-plane-operator/hostedclusterconfigoperator/controllers/machine/machine.go
enxebre: I think is a fair expectation for this to not work i ingress capability is not enable. cc @orenc1

linoyaslan: Great! Thanks Alberto

orenc1: i think that in case the ingress of the hosted cluster is disabled, you should also not create the ingress passthrough Route. see the code block starting at:
https://github.com/openshift/hypershift/blob/daf832300487d61dd3f89a35f60bccb55f535fb0/control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go#L1095

linoyaslan: Hi @orenc1 , the code you shared is part of reconcileIngressController, and I’ve already wrapped it with a condition that checks if ingress is enabled.

PR #6322, opened 2025-06-25
OCPBUGS-56430: Fix restarting prometheus Sts after HCP restoration

jparrill:
- During the bug verification, ROSA people find the STS monitoring pods were not restarting as expected. This is due to ReadyReplicas is not a state that STS has from the beginning but AvailableReplicas.
- Fixed another bug regarding the ResourceGeneration of the status be cause it was being updated in every reconciliation loop.

- Fixes [OCPBUGS-56430](https://issues.redhat.com/browse/OCPBUGS-56430)


bryan-cox (approved):
Looks good to me overall. The bug release note needs set on the bug ticket.

PR #6353, opened 2025-06-30
OCPBUGS-56701: Update the LookupDefaultOCPVersion function to use the multi-arch release API

bryan-cox:
Updates the LookupDefaultOCPVersion function to use the multi-arch release API when a release stream is provided. This ensures consistency between when a release stream is provided and when it is not. We should also default to multi-arch whenever possible.

Fixes https://issues.redhat.com/browse/OCPBUGS-56701

**Checklist**

### DIFF
--- a/cmd/cluster/agent/create_test.go
+++ b/cmd/cluster/agent/create_test.go
@@ -2,11 +2,14 @@ package agent
 
 import (
 	"context"
+	"net/http"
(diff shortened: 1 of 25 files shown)


jparrill (approved):
I really like this PR thank you so much @bryan-cox to put it in the table 🙏 .
/lgtm

PR #6354, opened 2025-06-30
[CNTRLPLANE-1034] sharedingress: add support for whitelist cidrs blocks

nimrodshn:
This patch provides an implementation for CIDR blocks whitelisting. 

We propogate CIDR blocks provided by client via HostedCluster's `APIServerNetworking` and utilize those in the HAProxy configuration generated in order to whitelist specific source IPs. 

~~**In case no IPs are provided we will default to `0.0.0.0/0` (i.e. "allow-all").**~~
In case no IPs are provided no ACLs are created in the HAProxy router configuration.

Fixes https://issues.redhat.com/browse/CNTRLPLANE-1034

**Checklist**

### DIFF
--- a/hypershift-operator/controllers/sharedingress/router.go
+++ b/hypershift-operator/controllers/sharedingress/router.go
@@ -66,13 +66,14 @@ func (r routeByNamespaceName) Less(i, j int) bool {
 	return r[i].Namespace+r[i].Name < r[j].Namespace+r[j].Name
 }
 
-func generateRouterConfig(svcList *corev1.ServiceList, svcsNamespaceToClusterID map[string]string, routes []routev1.Route, svcsNameToIP map[string]string) (string, error) {
+func generateRouterConfig(svcList *corev1.ServiceList, svcsNamespaceToClusterID map[string]string, routes []routev1.Route,
+	svcsNameToIP map[string]string, svcsNamespaceToApiServerNetworking map[string]*hyperv1.APIServerNetworking) (string, error) {
 	type backendDesc struct {
-		Name        string
-		SVCIP       string
-		SVCPort     int32
-		ClusterID   string
-		AllowedCIDR string
+		Name         string
+		SVCIP        string
+		SVCPort      int32
+		ClusterID    string
+		AllowedCIDRs []hyperv1.CIDRBlock
 	}
 	type ExternalDNSBackendDesc struct {
 		Name                 string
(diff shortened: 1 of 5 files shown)


enxebre:
can we please create an e2e for this?
you could include it as part of TestCreateClusterCustomConfig, right now that only runs for aws so you would need to drop the test global check and let it be granular for the specific validations

hypershift-operator/controllers/sharedingress/router_config.template
```diff
@@ -30,22 +30,31 @@ frontend dataplane-kas-svc
...
   tcp-request inspect-delay 5s
   tcp-request content accept if { req_ssl_hello_type 1 }
 
+  tcp-request content set-var(sess.src_ip) src
+
+  log-format "%{+Q}o\ src_ip = %[var(sess.src_ip)], %ci:%cp / %si:%sp -> %fi:%fp [%t] %ft %b/%s %Tw/%Tc/%Tt %B %ts %ac/%fc/%bc/%sc/%rc %sq/%bq"
+
   {{- range .ExternalDNSBackends }}
   acl is_{{ .Name }} req_ssl_sni -i {{ .HostName }}
+  acl is_{{ .Name }}_request_allowed src 91.47.40.124
```
jharrington22: What is this IP? I think this is a typo from the original snippet

hypershift-operator/controllers/sharedingress/router.go
```diff
@@ -88,10 +92,11 @@ func generateRouterConfig(svcList *corev1.ServiceList, svcsNamespaceToClusterID
...
 		p.Backends = append(p.Backends, backendDesc{
-			Name:      svc.Namespace + "-" + svc.Name,
-			SVCIP:     svc.Spec.ClusterIP,
-			SVCPort:   svc.Spec.Ports[0].Port,
-			ClusterID: svcsNamespaceToClusterID[svc.Namespace],
+			Name:         svc.Namespace + "-" + svc.Name,
+			SVCIP:        svc.Spec.ClusterIP,
+			SVCPort:      svc.Spec.Ports[0].Port,
+			ClusterID:    svcsNamespaceToClusterID[svc.Namespace],
+			AllowedCIDRs: setAllowedCIDRs(svcsNamespaceToApiServerNetworking[svc.Namespace]),
```
jharrington22, line 97: Why set this at all if there isn't a CIDR configued?

jharrington22 (changes requested):

hypershift-operator/controllers/sharedingress/router_config.template
nimrodshn: I am not sure. I will remove it.

hypershift-operator/controllers/sharedingress/router.go
nimrodshn: @jharrington22 Thanks for the review!
Done. Note that I've done this conditional formatting **inside the template** otherwise there would be redundant lines in the template.
In this manor- If no `AllowedCIDRBlocks` specified [or they where removed by the client] the template is exactly the same as it was previously. 
Without these conditionals we would likely need to write `0.0.0.0/0` to specify explicitly as in previous commit. 
I hope that makes sense.

nimrodshn:
Closing in favor of https://github.com/openshift/hypershift/pull/6366
cc @muraee

PR #6356, opened 2025-07-01
MGMT-20793: Enable Disabling NodeTuning Capability

linoyaslan:
This PR is part of the task for expanding support for disabling openshift capabilities in hypershift. This change adds support for disabling the NodeTuning capability, who managed by both CPO and CVO.

**Checklist**

### DIFF
--- a/support/capabilities/hosted_control_plane_capabilities.go
+++ b/support/capabilities/hosted_control_plane_capabilities.go
@@ -11,6 +11,19 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 )
 
+// IsNodeTuningCapabilityEnabled returns true if the NodeTuning capability is enabled, or false if disabled.
+func IsNodeTuningCapabilityEnabled(capabilities *hyperv1.Capabilities) bool {
+	if capabilities == nil {
+		return true
+	}
+	for _, disabledCap := range capabilities.Disabled {
+		if disabledCap == hyperv1.NodeTuningCapability {
+			return false
+		}
+	}
+	return true
+}
+
 // HasDisabledCapabilities returns true if any capabilities are disabled; otherwise, it returns false.
 func HasDisabledCapabilities(capabilities *hyperv1.Capabilities) bool {
 	if capabilities == nil {
--- a/test/e2e/util/util.go
+++ b/test/e2e/util/util.go
@@ -47,8 +47,11 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	kapierror "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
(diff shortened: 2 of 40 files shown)


support/capabilities/hosted_control_plane_capabilities.go
```diff
@@ -11,6 +11,19 @@ import (
...
+	if capabilities == nil {
+		return true
+	}
+	enabled := true
+	for _, disabledCap := range capabilities.Disabled {
+		if disabledCap == capability {
+			enabled = false
+		}
+	}
+	return enabled
```
enxebre: this assumes absence == enabled which is not necessarily true.
Let's change this to be IsNodeTuningCapabilityDisabled which only returns true when it's in capabilities.Disabled

linoyaslan: I’d like to make this more generic, so maybe its better to change it to IsCapabilityDisabled.

Also, this is how it was implemented for the ImageRegistry, so I assume it would need a fix there as well?

One more thing: I’m not sure how this situation can even happen. Aren’t capabilities enabled by default in HyperShift? We’re just adding the option to disable them now, right?

linoyaslan: > this assumes absence == enabled which is not necessarily true.

If I understand correctly, you’re saying that if capabilities == nil, it doesn’t explicitly mean either “enabled” or “disabled,” right?
this is an edge case I need to handle regardless of whether I’m checking if a capability is enabled or disabled. so, should we treat nil as enabled or disabled?

enxebre: >One more thing: I’m not sure how this situation can even happen. Aren’t capabilities enabled by default in HyperShift? We’re just adding the option to disable them now, right?

no, that's not necessarily the case. HCP reserves its ability to have disable caps in the backend, e.g. baremetal.

>I’d like to make this more generic, so maybe its better to change it to IsCapabilityDisabled.

we don't want to make more generic because of the above. We want to do it in cap basis.

>Also, this is how it was implemented for the ImageRegistry, so I assume it would need a fix there as well?

No it's not, registry is specific -> IsImageRegistryCapabilityEnabled -> which basically what I'm asking here.

linoyaslan: @enxebre 
Sorry for insisting on this, I just don’t understand why we wouldn’t use isCapabilityDisabled and pass the specific capability as needed. That way we have a single function for all capabilities, and we just pass the appropriate capability based on the logic.

> No it's not, registry is specific -> IsImageRegistryCapabilityEnabled -> which basically what I'm asking here.

Regarding the imageRegistry, what I meant is that the function checks **whether it’s enabled**, not whether it’s disabled. It also assumes that absence means enabled:

```
// IsImageRegistryCapabilityEnabled returns true if the Image Registry
// capability is enabled, or false if disabled.
//
// The Image Registry capability is enabled by default.
func IsImageRegistryCapabilityEnabled(capabilities *hyperv1.Capabilities) bool {
	if capabilities == nil {
		return true
	}
	enabled := true
	for _, disabledCap := range capabilities.Disabled {
		if disabledCap == hyperv1.ImageRegistryCapability {
			enabled = false
		}
	}
	return enabled
}
```

enxebre: because absence might have different meanings for different caps. We can't use a generic func

test/e2e/util/util.go
```diff
@@ -2686,3 +2689,86 @@ func EnsureConsoleCapabilityDisabled(ctx context.Context, t *testing.T, g Gomega
 		g.Expect(err.Error()).To(ContainSubstring("namespaces \"openshift-console\" not found"))
 	})
 }
+
+// EnsureNodeTuningCapabilityDisabled validates the expectations for when NodeTuningCapability is Disabled
+func EnsureNodeTuningCapabilityDisabled(ctx context.Context, t *testing.T, g Gomega, clients *GuestClients, mgmtClient crclient.Client, hostedCluster *hyperv1.HostedCluster) {
```
enxebre: does this gomegas need to be declared locally cc @muraee?

linoyaslan: As I understand it, you create g := NewWithT(t) once per test, and pass the g Gomega to helper functions to avoid creating it multiple times within the same test.

muraee: you are declaring a new sub-test with `t.Run()` not a helper function. Using gomega from the parent test will cause the parent test to fail and stop execution immediately(skipping the rest of the tests) when `g.expect()` fails and might cause unpredicted behavior.

PR #6358, opened 2025-07-01
MGMT-20948: Enable Disabling Multus

linoyaslan:
This PR adds support for disabling the Multus capability in HyperShift. When DisableMultus is set, the CNO skips deploying Multus-related resources.

**Checklist**


test/e2e/disable_multus_test.go
```diff
@@ -0,0 +1,242 @@
...
+	"k8s.io/apimachinery/pkg/util/wait"
+	crclient "sigs.k8s.io/controller-runtime/pkg/client"
+
+	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
+	networkoperator "github.com/openshift/hypershift/control-plane-operator/hostedclusterconfigoperator/controllers/resources/network"
+	"github.com/openshift/hypershift/hypershift-operator/controllers/manifests"
+	e2eutil "github.com/openshift/hypershift/test/e2e/util"
+)
+
+func TestDisableMultus(t *testing.T) {
```
enxebre: This will result in a new HC for the e2e test suite which has an infra cost we don't want to increase. Please Add this to the the createClusterCustomConfig along with the other validations

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1741,7 +1740,13 @@ type OperatorConfiguration struct {
 	// clusterVersionOperator specifies the configuration for the Cluster Version Operator in the hosted cluster.
 	//
 	// +optional
+	// +openshift:enable:FeatureGate=ClusterVersionOperatorConfiguration
 	ClusterVersionOperator *ClusterVersionOperatorSpec `json:"clusterVersionOperator,omitempty"`
+
+	// clusterNetworkOperator specifies the configuration for the Cluster Network Operator in the hosted cluster.
+	//
```
enxebre, line 1912: great! can you please drop a link to the canonical API for ref here?

linoyaslan: Is this the reference you’re referring to? - https://github.com/openshift/api/blob/master/operator/v1/types_network.go#L86 ?

api/hypershift/v1beta1/operator.go
```diff
@@ -33,3 +33,19 @@ type ClusterVersionOperatorSpec struct {
...
 	OperatorLogLevel LogLevel `json:"operatorLogLevel,omitempty"`
 }
+
+type ClusterNetworkOperatorSpec struct {
+	// disableMultus when set to true disables the Multus CNI plugin and related components
+	// in the hosted cluster. This prevents the installation of multus daemon sets in the
+	// guest cluster and the multus-admission-controller in the management cluster.
+	// Default is false (Multus is enabled).
+	// This field is immutable.
+	// This field can only be set to true when NetworkType is "Other". Setting it to true
```
enxebre, line 43: >// This field can only be set to true when NetworkType is "Other"

please enforce this with a CEL validation, you can find similar examples across the API see e.g. ServicePublishingStrategy, and then add a new validation for it in TestOnCreateAPIUX

enxebre: any reason to not name this field as in the original API https://github.com/openshift/api/blob/master/operator/v1/types_network.go#L86
I don't see it being immutable in that API, does it need to be?

linoyaslan: @enxebre "Changing the parameter value as a postinstallation cluster task has no effect."

> any reason to not name this field as in the original API

Agree, I'll change the name to be the same like in the original API

linoyaslan: @enxebre the CEL validation is sufficient right? no need the validation in Validate func anymore, correct?

api/hypershift/v1beta1/operator.go
```diff
@@ -33,3 +33,19 @@ type ClusterVersionOperatorSpec struct {
...
+	// with any other NetworkType will result in a validation error during cluster creation.
+	//
+	// +optional
+	// +kubebuilder:default:=false
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf", message="disableMultiNetwork is immutable"
```
enxebre, line 48: cc @JoelSpeed we are deliberately choosing to mirror the operator.openshift.io API here as an embedded API mechanism, hence this being bool.
Still in the original api seems to be a pointer, any preferences?

JoelSpeed: All bools should be pointers, else they won't round trip through structured clients who observe `omitempty`.

linoyaslan: Thanks, I'll fix it

cmd/cluster/core/create.go
```diff
@@ -457,6 +459,16 @@ func prototypeResources(ctx context.Context, opts *CreateOptions) (*resources, e
 	}
 	prototype.Cluster.Spec.Networking.MachineNetwork = machineNetworkEntries
 
+	if opts.DisableMultiNetwork {
```
enxebre, line 462: this will produce a failing HC providing networking won't be none, no?
I don't think we need to expose this in the cli

test/e2e/create_cluster_test.go
```diff
@@ -1146,6 +1146,69 @@ func TestOnCreateAPIUX(t *testing.T) {
...
+					name                   string
+					mutateInput            func(*hyperv1.HostedCluster)
+					expectedErrorSubstring string
+				}{
+					{
+						name: "when disableMultiNetwork is set to false it should pass",
+						mutateInput: func(hc *hyperv1.HostedCluster) {
+							hc.Spec.OperatorConfiguration = &hyperv1.OperatorConfiguration{
+								ClusterNetworkOperator: &hyperv1.ClusterNetworkOperatorSpec{
+									DisableMultiNetwork: &[]bool{false}[0],
```
enxebre: what's this notation coming from?

linoyaslan: I found it as way in Go to create a *bool with the value false/true, since you can’t take the address of a literal directly (&false isn’t valid). So we allocate a one-element slice with false, then take the address of the first element.

cmd/cluster/core/create.go
enxebre: if we expose it, the same api validation need to be added here in func (opts *RawCreateOptions) Validate(ctx context.Context) (*ValidatedCreateOptions, error) {

test/e2e/create_cluster_test.go
JoelSpeed: Might just want to use the K8s `k8s.io/utils/ptr` library, `ptr.To(false)` would be sufficient

cmd/cluster/core/create.go
linoyaslan: done

test/e2e/create_cluster_test.go
linoyaslan: Thanks @JoelSpeed I'll try that

linoyaslan: done

control-plane-operator/controllers/hostedcontrolplane/multus_disable_test.go
```diff
@@ -0,0 +1,165 @@
...
+				assert.Equal(t, *test.expectedDisableMultiNetwork, *hcp.Spec.OperatorConfiguration.ClusterNetworkOperator.DisableMultiNetwork, test.description)
+			}
+
+			// Test that the reconciler logic would make the correct decision
+			// When DisableMultiNetwork is nil, it defaults to false (multus enabled)
+			disableMultiNetwork := false
+			if hcp.Spec.OperatorConfiguration.ClusterNetworkOperator.DisableMultiNetwork != nil {
+				disableMultiNetwork = *hcp.Spec.OperatorConfiguration.ClusterNetworkOperator.DisableMultiNetwork
+			}
+			shouldProcessMultus := !disableMultiNetwork
```
csrwng: It's not clear to me what main controller logic this is testing. If this intent is to imitate the logic in lines 2309-2316, then that logic should be extracted in a separate function and the function called from the test. Otherwise, it's easy to update that logic in the future without updating the logic here in the test.

control-plane-operator/controllers/hostedcontrolplane/multus_disable_test.go
```diff
@@ -0,0 +1,165 @@
...
+				},
+			}
+
+			// Test the restart logic: restart when hasRestartAnnotation && !DisableMultiNetwork
+			// When DisableMultiNetwork is nil, it defaults to false (multus enabled)
+			disableMultiNetwork := false
+			if hcp.Spec.OperatorConfiguration.ClusterNetworkOperator.DisableMultiNetwork != nil {
+				disableMultiNetwork = *hcp.Spec.OperatorConfiguration.ClusterNetworkOperator.DisableMultiNetwork
+			}
+			shouldRestart := test.hasRestartAnnotation && !disableMultiNetwork
```
csrwng: again, for this to be a meaningful test, it should be calling code from the hostedcontrolplane_controller.go
Maybe extract that logic into function called setMultusAdmissionControllerRestartAnnotation and call that function from this test?

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -565,10 +565,17 @@ func (r *reconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Result
 		errs = append(errs, fmt.Errorf("failed to reconcile the kube apiserver service monitor: %w", err))
 	}
 
+	disableMultiNetwork := false
```
csrwng: This logic is used in 2 other places. Maybe it's worth putting it in a separate function in the /support/util package that gets called from both hostedcontrolplane_controller and here.

Something like: `util.IsMultiNetworkDisabled(hcp *hyperv1.HostedControlPlane)`

csrwng (commented):
Some comments, thx!

linoyaslan:
@csrwng Thanks for your review! I've addressed all the comments, I'd appreciate it if you could take another look.

mgencur:
Hi! I'm a little late to this but I have one question: Why does disabling multi-network handle Calico specifically? (e.g. [here](https://github.com/openshift/hypershift/pull/6358/files#diff-aecafed13dd71bf2ce40242b2480ef31a2feae92c312eaf737cea45e7706ae1fR406), or [here](https://github.com/openshift/hypershift/blob/main/cmd/cluster/core/create.go#L771))
I know in Hypershift's CLI [we support](https://github.com/openshift/hypershift/blob/main/cmd/cluster/core/create.go#L73) "Calico" as a special option but the value is simply passed to ClusterNetworkOperator and the CNO [handles](https://github.com/openshift/cluster-network-operator/blob/master/README.md?plain=1#L142-L145) all values other then "OVNKubernetes" in the same way. In other words, using "Other" or "Calico" has the very same effect for CNO. IMO, we don't need this specific Option for Calico in our CLI and I would like to remove it.
But my question is whether disabling multi network in this pull request handles Calico specifically just because we have it as a separate option or it has real functional implications for Calico (if disabling multi-network somehow affects Calico specifically, which I doubt). Thanks.

linoyaslan:
> i! I'm a little late to this but I have one question: Why does disabling multi-network handle Calico specifically?

Hi @mgencur , Calico is one of the possible values for networkType, and this PR simply adds support to disable Multus **only** when networkType=Other. Just that, nothing deeper :)

PR #6360, opened 2025-07-01
ACM-21713: fix(konflux): go-toolset runs as default(1001)

celebdor:

Since the Red Hat catalog go-toolset runs as a user with id = 1001, we need to make sure that the COPY instruction reflects that.

Fixes #

**Checklist**


celebdor:
The original #6317 missed this.

PR #3197, opened 2023-11-15
CNV-23418: Validate jsonpatch annotation + add condition on wrong patch

nunnatsa:
## What this PR does / why we need it
* validate the jsonpatch annotation in webhook. Make sure the jsonpatch string is a valid standard jsonpatch
* If the creation of a kubevirt machine template fails, add a new condition to the node pool to reflect this error.

### Which issue(s) this PR fixes
Fixes #[CNV-23418](https://issues.redhat.com//browse/CNV-23418)

### Checklist
- [x] Subject and description added to both, commit and PR.
- [x] Relevant issues have been referenced.
- [ ] This change includes docs. 
- [x] This change includes unit tests.

PR #6363, opened 2025-07-02
webhook: Replace "delete" by "remove" for jpatch

qinqon:
The json patch validation introduced at [1] is expecting the operation "delete" to skip checking if the field "value" is parth of the json patch, this is no accurate since there is no such "delete" operation json patch spec only has the "remove" operation and is the one that can be created without "value" field.

[1] https://github.com/openshift/hypershift/pull/3197

Fixes #

**Checklist**

### DIFF
--- a/hypershift-operator/controllers/hostedcluster/hostedcluster_webhook_test.go
+++ b/hypershift-operator/controllers/hostedcluster/hostedcluster_webhook_test.go
@@ -204,6 +204,14 @@ func TestValidateJsonAnnotation(t *testing.T) {
 			expectError: false,
 		},
 		{
+			name: "valid remove without value",
+			annotations: map[string]string{
+				v1beta1.JSONPatchAnnotation: `[{"op": "remove","path": "/spec/template/metadata/annotations/kubevirt.io~1allow-pod-bridge-network-live-migration"}]`,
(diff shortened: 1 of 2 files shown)


hypershift-operator/controllers/hostedcluster/hostedcluster_webhook_test.go
```diff
@@ -238,6 +246,13 @@ func TestValidateJsonAnnotation(t *testing.T) {
 			},
 			expectError: true,
 		},
+		{
+			name: "bad operation",
+			annotations: map[string]string{
+				v1beta1.JSONPatchAnnotation: `[{"op": "delete","path": "/spec/domain/cpu/cores", "value": "1"}]`,
```
nunnatsa, line 252: Is it "delete" on purpose? Because you have two issues here. Both value and delete.

qinqon: I want to check the bad operation, if I don't set "value" webhook validates the  "value" exist and fails there but not at the json patch operation validation.

qinqon:
The verify job is stuck for 13 hours straight... clearly is not related.

PR #6369, opened 2025-07-02
NO-JIRA: Update azure dev scripts

bryan-cox:
This PR updates the upstream docs and the contrib/dev scripts on the latest ways to setup ARO HCP.

N/A

**Checklist**

### DIFF
--- a/docs/content/how-to/azure/create-azure-cluster_on_aks.md
+++ b/docs/content/how-to/azure/create-azure-cluster_on_aks.md
@@ -3,8 +3,14 @@
 This document describes how to set up an Azure Hosted Cluster on an AKS management cluster with an ExternalDNS setup. 
 Azure HostedClusters on AKS are supported from OCP 4.18.0+.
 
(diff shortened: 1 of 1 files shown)


contrib/managed-azure/create_basic_hosted_cluster.sh
```diff
@@ -79,47 +104,40 @@ az identity federated-credential create --name "${CSI_FILE_CCM_MSI_NAME}"-fed-id
...
 --marketplace-publisher azureopenshift \
 --marketplace-offer aro4 \
 --marketplace-sku aro_417 \
 --marketplace-version 417.94.20240701 \
 --dns-zone-rg-name os4-common \
 --assign-service-principal-roles \
---managed-identities-file /Users/your username/aro-hcp-control-plane-MIs.json \
---data-plane-identities-file /Users/your username/aro-hcp-data-plane-MIs.json
+--managed-identities-file /Users/user/aro-hcp-control-plane-MIs.json \
+--data-plane-identities-file /Users/user/aro-hcp-data-plane-MIs.json
```
bryan-cox, line 141: Should we also add the diagnostics flag? It let's you see the serial console on the VM.

bryan-cox:
/close

In favor of https://github.com/openshift/hypershift/pull/6395

PR #6367, opened 2025-07-02
OCPBUGS-58149: fix(hcco): Don't fail for public OIDC clients with no client secret

devguyio:
**What this PR does / why we need it**:

As the client secret name will be empty for public OIDC clients, HCCO should check if the secret ref name is the empty string before attempting to copy the secret.

This PR also adds proper unit tests for reconciling Auth OIDC clients. 

This PR is based on #6365 

**Which issue(s) this PR fixes**

Fixes #OCPBUGS-58149

**Checklist**
- [x] Subject and description added to both, commit and PR.
- [x] Relevant issues have been referenced.
- [ ] This change includes docs. 
- [x] This change includes unit tests.

PR #6375, opened 2025-07-03
OCPSTRAT-2173: feat(hcco): add support for hosted OIDC client secrets

devguyio:

add support for hosted OIDC client secrets to be used as day-2 secrets. 

Fixes #[OCPSTRAT-2173](https://issues.redhat.com//browse/OCPSTRAT-2173)

This PR is based on #6367 

**Checklist**

### DIFF
--- a/api/hypershift/v1beta1/hostedcluster_types.go
+++ b/api/hypershift/v1beta1/hostedcluster_types.go
@@ -362,6 +362,19 @@ const (
 	// a process to check if the different components in the DataPlane are working as expected. Checks:
 	// - Validates the monitoring stack is properly working after restoration, if not HCCO will restart the prometheus-k8s pods.
 	HostedClusterRestoredFromBackupAnnotation = "hypershift.openshift.io/restored-from-backup"
+
+	// HostedClusterSourcedAnnotation is set to true on Secret and ConfigMap resources to designate them as
+	// hosted-cluster-sourced resources. This means that the hosted cluster version of these resources is the source of
+	// truth and the management cluster version will be just empty resources that have this annotation. This is useful
+	// to enable day-two configuration use cases where such resources are expected to be provided by the end-user after
+	// the cluster creation, and, due to certain restrictions, those resources include sensitive data that can't live
+	// on the control-plane. Setting this annotation will instruct HyperShift to skip creating this resource on the hosted
+	// cluster and to not override any changes done later on the hosted cluster version of this resource.
+	//
+	// This annotation can only be set on empty resources and currently it's only honored when set on secrets that are
+	// referenced in the HostedCluster `spec.configuration.authentication.oidcProviders[*].oidcClients[*].clientSecret`
+	// and only for the ARO-HCP platform.
+	HostedClusterSourcedAnnotation = "hypershift.openshift.io/hosted-cluster-sourced"
 )
 
 // RetentionPolicy defines the policy for handling resources associated with a cluster when the cluster is deleted.
--- a/control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
+++ b/control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
@@ -1239,6 +1239,10 @@ func (r *reconciler) reconcileAuthOIDC(ctx context.Context, hcp *hyperv1.HostedC
 						errs = append(errs, fmt.Errorf("failed to get OIDCClient secret %s: %w", oidcClient.ClientSecret.Name, err))
 						continue
 					}
+					if azureutil.IsAroHCP() && util.HasAnnotationWithValue(&src, hyperv1.HostedClusterSourcedAnnotation, "true") {
+						// This is a day-2 secret. We shouldn't copy it, instead it'll be provided by the end-user on the hosted cluster.
+						continue
+					}
 					dest := corev1.Secret{
 						ObjectMeta: metav1.ObjectMeta{
 							Name:      oidcClient.ClientSecret.Name,
@@ -1254,6 +1258,7 @@ func (r *reconciler) reconcileAuthOIDC(ctx context.Context, hcp *hyperv1.HostedC
 					})
 					if err != nil {
 						errs = append(errs, fmt.Errorf("failed to reconcile OIDCClient secret %s: %w", dest.Name, err))
+
 					}
 				}
 			}
--- a/control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources_test.go
+++ b/control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources_test.go
@@ -16,6 +16,7 @@ import (
 	"github.com/openshift/hypershift/control-plane-operator/hostedclusterconfigoperator/controllers/resources/kas"
 	"github.com/openshift/hypershift/control-plane-operator/hostedclusterconfigoperator/controllers/resources/manifests"
 	"github.com/openshift/hypershift/hypershift-operator/controllers/nodepool"
+	"github.com/openshift/hypershift/support/azureutil"
 	"github.com/openshift/hypershift/support/globalconfig"
 	fakereleaseprovider "github.com/openshift/hypershift/support/releaseinfo/fake"
 	supportutil "github.com/openshift/hypershift/support/util"
@@ -33,6 +34,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/validation"
 	"k8s.io/utils/ptr"
 
@@ -1501,6 +1503,7 @@ func TestReconcileAuthOIDC(t *testing.T) {
 		expectOIDCClientSecrets []string
 		expectErrors            bool
 		expectedErrorMessages   []string
+		setAROHCP               bool
 	}{
 		"when OAuth is enabled, should not copy OIDC resources": {
 			inputHCP: &hyperv1.HostedControlPlane{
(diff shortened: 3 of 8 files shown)


api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -362,6 +362,19 @@ const (
 	// a process to check if the different components in the DataPlane are working as expected. Checks:
 	// - Validates the monitoring stack is properly working after restoration, if not HCCO will restart the prometheus-k8s pods.
 	HostedClusterRestoredFromBackupAnnotation = "hypershift.openshift.io/restored-from-backup"
+
+	// HostedClusterSourcedAnnotation is set to true on Secret and ConfigMap resources to designate them as
+	// hosted-cluster-sourced resources. This means that the hosted cluster version of these resources in the source of
```
mgencur:
```suggestion
	// hosted-cluster-sourced resources. This means that the hosted cluster version of these resources is the source of
```

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -362,6 +362,19 @@ const (
...
+	// truth and the management cluster version will be just empty resources that have this annotation. This is useful
+	// to enable day-two configuration use cases where such resources are expected to be provided by the end-user after
+	// the cluster creation, and , duo to certain restrictions, those resources include sensitive data that can't live
+	// on the control-plane. Setting this annotation will instruct HyperShift to skip creating this resource on the hosted
+	// cluster and to not override any changes done later on the hosted cluster version of this resource.
+	//
+	// This annotation can only be set on empty resources and currently it'll only be honored when set on secrets that
```
mgencur: > can only be set on empty resources

I don't see this restriction anywhere in the code. So, this only describes how it should be used but it is not enforced, right?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -362,6 +362,19 @@ const (
...
+	// referenced in the HostedCluster `spec.configuration.authentication.oidcProviders[*].oidcClients[*].clientSecret`
+	// and only for the ARO-HCP platform.
```
mgencur, line 376: > only for the ARO-HCP platform

I supposed it will be honored on all platforms as I don't see this enforced in the code. This probably just informs the user where this might be used? Maybe I'd leave this out or rephrase it a little bit.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -1239,6 +1239,10 @@ func (r *reconciler) reconcileAuthOIDC(ctx context.Context, hcp *hyperv1.HostedC
 						errs = append(errs, fmt.Errorf("failed to get OIDCClient secret %s: %w", oidcClient.ClientSecret.Name, err))
 						continue
 					}
+					if util.HasAnnotationWithValue(&src, hyperv1.HostedClusterSourcedAnnotation, "true") {
```
mgencur: It's probably fine that the reconciler will be throwing errors until the Secret is created with this annotation (which might be for a couple of days)? It will be throwing "failed to get OIDCClient secret"

devguyio: On the management cluster, the secret needs to exist to pass a config validation done by the HO. And that's why the reconciler won't throw that error.

api/hypershift/v1beta1/hostedcluster_types.go
devguyio: nice catch! actually it's intended to be supported only on ARO-HCP and there was a check in the code that I removed during testing. will put back.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
mgencur: Alright, so the Secret with the annotation needs to exist beforehand. Cool.

api/hypershift/v1beta1/hostedcluster_types.go
devguyio: another good catch! I added that validation and a unit test for it. I also tested it in a cluster

devguyio:
## Updates

- Added a validation for empty, hosted client secrets.
- Added a unit test for the validation code.
- Addressed review comments.

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -362,6 +362,19 @@ const (
 	// a process to check if the different components in the DataPlane are working as expected. Checks:
 	// - Validates the monitoring stack is properly working after restoration, if not HCCO will restart the prometheus-k8s pods.
 	HostedClusterRestoredFromBackupAnnotation = "hypershift.openshift.io/restored-from-backup"
+
+	// HostedClusterSourcedAnnotation is set to true on Secret and ConfigMap resources to designate them as
+	// hosted-cluster-sourced resources. This means that the hosted cluster version of these resources is the source of
+	// truth and the management cluster version will be just empty resources that have this annotation. This is useful
+	// to enable day-two configuration use cases where such resources are expected to be provided by the end-user after
+	// the cluster creation, and , duo to certain restrictions, those resources include sensitive data that can't live
```
bryan-cox:
```suggestion
	// the cluster creation, and , due to certain restrictions, those resources include sensitive data that can't live
```

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -362,6 +362,19 @@ const (
...
+	// on the control-plane. Setting this annotation will instruct HyperShift to skip creating this resource on the hosted
+	// cluster and to not override any changes done later on the hosted cluster version of this resource.
+	//
+	// This annotation can only be set on empty resources and currently it'll only be honored when set on secrets that
+	// referenced in the HostedCluster `spec.configuration.authentication.oidcProviders[*].oidcClients[*].clientSecret`
```
bryan-cox, line 375:
```suggestion
	// are referenced in the HostedCluster `spec.configuration.authentication.oidcProviders[*].oidcClients[*].clientSecret`
```

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources_test.go
```diff
@@ -1952,6 +2053,12 @@ func TestReconcileAuthOIDC(t *testing.T) {
 				CreateOrUpdateProvider: &simpleCreateOrUpdater{},
 			}
 
+			if test.setAROHCP {
+				azureutil.SetAroHCP()
```
bryan-cox: I think you could drop both of these util functions and just use `t.Setenv`. You should be able to find examples of it in the code.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources_test.go
```diff
@@ -2036,6 +2143,19 @@ func TestReconcileAuthOIDC(t *testing.T) {
 					g.Expect(errors.IsNotFound(err)).To(BeTrue())
 				}
 			}
+			// Verify that no unexpected client secrets were copied
+			secretList := &corev1.SecretList{}
+			err = hcClient.List(ctx, secretList, client.InNamespace(ConfigNamespace))
```
bryan-cox, line 2147: Change the context on L2039 to use the new `t.Context()`

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -1693,6 +1693,9 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
 				if err := ensureReferencedResourceAnnotation(ctx, r.Client, hcluster.Name, sourceSecret); err != nil {
 					return ctrl.Result{}, fmt.Errorf("failed to set referenced resource annotation: %w", err)
 				}
+				if err := ensureHostedResroucesAreEmpty(ctx, r.Client, hcluster, sourceSecret); err != nil {
```
bryan-cox:
```suggestion
				if err := ensureHostedResourcesAreEmpty(ctx, r.Client, hcluster, sourceSecret); err != nil {
```

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -4727,6 +4730,20 @@ func ensureReferencedResourceAnnotation(ctx context.Context, client client.Clien
 	return nil
 }
 
+func ensureHostedResroucesAreEmpty(ctx context.Context, crclient client.Client, hcluster *hyperv1.HostedCluster, obj client.Object) error {
+	if azureutil.IsAroHCP() && hyperutil.HasAnnotationWithValue(obj, hyperv1.HostedClusterSourcedAnnotation, "true") {
```
bryan-cox: Should we not use the fail fast pattern here?
```
if !azureutil.IsAroHCP() {
     return nil
}

if !hyperutil.HasAnnotationWithValue(obj, hyperv1.HostedClusterSourcedAnnotation, "true") {
    return nil
}
....
```

hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
```diff
@@ -4290,3 +4291,112 @@ func TestReconcileComponents(t *testing.T) {
 		t.Fatalf("update loop detected: %v", err)
 	}
 }
+
+func TestEnsureHostedResroucesAreEmpty(t *testing.T) {
```
bryan-cox, line 4344: This will need the typo fixed here too. Not sure if the compiler/IDE would flag that so noting it here.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources_test.go
```diff
@@ -2036,6 +2143,19 @@ func TestReconcileAuthOIDC(t *testing.T) {
...
+			g.Expect(err).ToNot(HaveOccurred())
+			expectedSecrets := map[string]struct{}{}
+			for _, name := range test.expectOIDCClientSecrets {
+				expectedSecrets[name] = struct{}{}
+			}
```
muraee:
```suggestion
			expectedSecrets := sets.New(test.expectOIDCClientSecrets...)
```
could use a set instead of map[]struct.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources_test.go
devguyio: yeah I saw the examples and I opted for this choice to not add another hardcoded env var. I opted for encapsulating the logic in the util package, and then later on I plan to do some cleanup PRs to remove all the hardcoded env var references. I think we should avaoid that and this logic should stay in a single place only.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources_test.go
devguyio: Done

hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
devguyio: Done

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
devguyio: Done

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
devguyio: Done

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources_test.go
devguyio: Done

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources_test.go
bryan-cox: But those util functions use the `os` version. Why not use the test package's version that already does the cleanup?
```
// Setenv calls os.Setenv(key, value) and uses Cleanup to
// restore the environment variable to its original value
// after the test.
```

devguyio: whoa! I wasn't aware of `t.SetEnv`, I was actually looking for it, and asking ChatGPT about such a thing returned nothing 😲 

further more, when I did a quick "MANAGED_SERVICE" search in the existing tests, I quickly saw those results as `os.SetEnv` 🤦🏽 

Thanks for keeping me in check @bryan-cox ! Fixed!

bryan-cox: NP! I just learned of it myself not too long ago 😄

PR #6376, opened 2025-07-03
CNTRLPLANE-1042: Disable createClusterNone for azure and scope down verifyResourceGroupLocationsMatch check

enxebre:

https://github.com/openshift/release/pull/66655 enables all tests to run on aks-e2e This is failing because of the createClusterNone test. Scoping down verifyResourceGroupLocationsMatch to only run on azure platform is fix that should solve the current failure. Besides this test provides no value and we want to focus on increasing valuable coverage so this PR just disbled it for azure

Fixes #[CNTRLPLANE-1042](https://issues.redhat.com//browse/CNTRLPLANE-1042)

**Checklist**

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
@@ -880,7 +880,7 @@ func (r *HostedControlPlaneReconciler) validateConfigAndClusterCapabilities(ctx 
 		}
 	}
 
-	if hyperazureutil.IsAroHCP() {
+	if hcp.Spec.Platform.Type == hyperv1.AzurePlatform && hyperazureutil.IsAroHCP() {
 		if err := r.verifyResourceGroupLocationsMatch(ctx, hcp); err != nil {
 			return err
 		}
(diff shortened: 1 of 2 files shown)


control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -880,7 +880,7 @@ func (r *HostedControlPlaneReconciler) validateConfigAndClusterCapabilities(ctx
 		}
 	}
 
-	if hyperazureutil.IsAroHCP() {
+	if hcp.Spec.Platform.Type == hyperv1.AzurePlatform && hyperazureutil.IsAroHCP() {
```
bryan-cox, line 883: I don't think we want to check this on self-managed Azure as this is partially related to MIv3 which is ARO HCP specific. 

One would need some Azure identity in order to perform this check.

enxebre: this change is not making it run in selfhosted, it's just making the check more robust. The validation will only run when the hc is azure and is running on aro.

PR #6395, opened 2025-07-08
NO-JIRA: Fix up contrib scripts for aro like env

enxebre:

Fixes #

**Checklist**

### DIFF
--- a/contrib/managed-azure/README.md
+++ b/contrib/managed-azure/README.md
@@ -1,15 +1,64 @@
 # General
-This directory contains several developer-focused scripts and instructions related to setting up managed Azure (aka ARO 
-HCP).
+This directory contains several developer-focused scripts and instructions related to setting up an AKS management cluster and create a HostedCluster replicating what we do in ARO environments.
 
-There are no guarantees these scripts will work 100% with your setup. If you do have issues with these scripts or need 
-further help. Please reach out to #project-hypershift on Red Hat Slack.
+If you do have issues with these scripts or need further help. Please reach out to #project-hypershift on Red Hat Slack.
 
-Review each of the scripts beforehand to make sure you are accomplishing any prerequisites required for running the 
-script. If you are starting out with nothing set up, you'll want to go through the scripts in this order:
+Steps:
 
-1. [Review Steps 1-7 to Set Up Environment Variables and Such](../../docs/content/how-to/azure/create-azure-cluster_on_aks.md)
-2. [Create an AKS management cluster](setup_aks_cluster.sh)
-3. [Set up externalDNS](setup_external_dns.sh)
-4. [Install the HyperShift Operator](setup_install_ho_on_aks.sh)
-5. [Create a HostedCluster](create_basic_hosted_cluster.sh)
+Create a ServicePrincipal to let hypershift cli create the HostedCluster infra for you.
+
+```
+SP_DETAILS=$(az ad sp create-for-rbac --name "${PREFIX}-sp" --role Contributor --scopes "/subscriptions/$SUBSCRIPTION_ID" -o json)
+CLIENT_ID=$(echo "$SP_DETAILS" | jq -r '.appId')
+CLIENT_SECRET=$(echo "$SP_DETAILS" | jq -r '.password')
+
+cat <<EOF > ./azure-creds.json
+{
+   "subscriptionId": "$SUBSCRIPTION_ID",
+   "tenantId": "$TENANT_ID",
+   "clientId": "$CLIENT_ID",
+   "clientSecret": "$CLIENT_SECRET"
+}
+# EOF
+```
+
+!!! warning
+      
+    In order for your Hypershift cluster to create properly, the Microsoft Graph `Application.ReadWrite.OwnedBy` 
+    permission must be added to your Service Principal and it also must be assigned to User Access Administrator at the 
+    subscription level. 
+
+    In most cases, you'll need to submit a DPTP request to have this done.
+
+From this repo root folder:
+
+```
+mkdir dev
+cd dev
+```
+
+Create your user-vars.sh file. E.g.
+
+```
+cat <<EOF > user-vars.sh
+# User variables.
+export PREFIX="USER-management"
+export PULL_SECRET="/path/pull-secret.txt"
+export HYPERSHIFT_BINARY_PATH="/path/go/src/github.com/openshift/hypershift/bin/"
+export HYPERSHIFT_IMAGE="quay.io/hypershift/hypershift-operator:latest"
+export RELEASE_IMAGE="quay.io/openshift-release-dev/ocp-release:4.20.0-ec.3-multi"
+export LOCATION="eastus"
+export AZURE_CREDS="/path/azure-hypershift-dev.json"
+# Azure storage account names must be between 3 and 24 characters in length and may contain numbers and lowercase letters only.
+export OIDC_ISSUER_NAME="USERmanagement"
+EOF
+```
+
+Run it
+
+```
+../contrib/managed-azure/setup_all.sh
+```
+
+
+You can comment out steps from setup_all.sh to run only what you want.
--- a/contrib/managed-azure/create_basic_hosted_cluster.sh
+++ b/contrib/managed-azure/create_basic_hosted_cluster.sh
@@ -4,35 +4,29 @@ set -x
 # This creates an Azure HostedCluster with the VNET in its own RG and the NSG in its own NSG.
 # The MANAGED_RG_NAME contains all the cloud resources created by the HC creation.
 
-# Prerequisites:
+# Prerequisites.
 # 1. Two different JSON files containing control plane service principals in one and workload managed identities in the
 # other, see steps 5 & 6 in setup_dev_environment.md.
 # 2. Set up the SA Token Issuer Private Key Path and OIDC Issue URL, see steps 15 and 16 in setup_dev_environment.md.
+LOCATION="${LOCATION:-}"
+AZURE_BASE_DOMAIN=${PARENT_DNS_ZONE:-}
+PULL_SECRET="${PULL_SECRET:-}"
+HYPERSHIFT_BINARY_PATH="${HYPERSHIFT_BINARY_PATH:-}"
+CP_OUTPUT_FILE="${CP_OUTPUT_FILE:-}"
+DP_OUTPUT_FILE="${DP_OUTPUT_FILE:-}"
+OIDC_ISSUER_URL="${OIDC_ISSUER_URL:-}"
+SA_TOKEN_ISSUER_PRIVATE_KEY_PATH="${SA_TOKEN_ISSUER_PRIVATE_KEY_PATH:-}"
 
-# Constants
-PREFIX="<your-prefix>"
-RELEASE_IMAGE=quay.io/openshift-release-dev/ocp-release-nightly@sha256:96a2ae25ddc2f291de4f143ce2ba84b94c6c6fb48c34fdedc45f22a20838c752
-TAG="<your-CPO-image-tag>"
-
-LOCATION="eastus"
+# Local.
+PREFIX="${PREFIX:-}"
 MANAGED_RG_NAME="${PREFIX}-managed-rg"
 CUSTOMER_VNET_RG_NAME="${PREFIX}-customer-vnet-rg"
 CUSTOMER_NSG_RG_NAME="${PREFIX}-customer-nsg-rg"
 CUSTOMER_VNET_NAME="${PREFIX}-customer-vnet"
 CUSTOMER_VNET_SUBNET1="${PREFIX}-customer-subnet-1"
 CUSTOMER_NSG="${PREFIX}-customer-nsg"
-MGMT_DNS_ZONE_NAME="blah-blah.com"
+MGMT_DNS_ZONE_NAME="${MGMT_DNS_ZONE_NAME:-}"
 CLUSTER_NAME="${PREFIX}-hc"
-AZURE_CREDS=/Users/your-username/.azure/credentials
-AZURE_BASE_DOMAIN=blah-blah.openshift.com
-PULL_SECRET=/Users/your-username/all-the-pull-secrets.json
-HYPERSHIFT_BINARY_PATH="/Users/your-username/hypershift_ws/hypershift/bin"
-MANAGED_IDENTITIES_FILE="/Users/your-username/azure/mi-file.json"
-CSI_CCM_MSI_NAME="<name of your CSI Disk managed identity>"
-CSI_FILE_CCM_MSI_NAME="<name of your CSI File managed identity>"
-IR_MSI_NAME="<name of your image registry managed identity>"
-OIDC_ISSUER_URL="https://your OIDC issuer url"
-SA_TOKEN_ISSUER_PRIVATE_KEY_PATH="/Users/your-username/cloud-credential-operator/serviceaccount-signer.private"
 
 # Create managed resource group
 az group create --name "${MANAGED_RG_NAME}" --location ${LOCATION}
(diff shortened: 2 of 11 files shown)


contrib/managed-azure/setup_MIv3_kv.sh
```diff
@@ -146,4 +151,3 @@ cat <<EOF > "${CP_OUTPUT_FILE}"
 }
 EOF
 
-set +x
```
georgelipceanu, line 149: If this is being removed the should the `set -x` at the beginning also be removed as well?

contrib/managed-azure/setup_aks_cluster.sh
```diff
@@ -58,12 +52,6 @@ az role assignment create \
 --scope /subscriptions/${SUBSCRIPTION_ID}/resourceGroups/"${AKS_RG}" \
 --assignee-principal-type ServicePrincipal
 
-## Associate your SP with the Key Vault; this is so you can use your SP with the CLI to create certificates in the key vault
```
bryan-cox, line 61: Why was this removed?

contrib/managed-azure/setup_dataplane_identities.sh
```diff
@@ -0,0 +1,42 @@
...
+--issuer "${OIDC_ISSUER_URL}" \
+--subject system:serviceaccount:openshift-cluster-csi-drivers:azure-file-csi-driver-node-sa \
+--audience openshift
+
+az identity federated-credential create --name "${IMAGE_REGISTRY_MI_NAME}"-fed-id \
+--identity-name "${IMAGE_REGISTRY_MI_NAME}" \
+--resource-group "${PERSISTENT_RG_NAME}" \
+--issuer "${OIDC_ISSUER_URL}" \
+--subject system:serviceaccount:openshift-image-registry:registry \
+--audience openshift
```
bryan-cox: Shouldn't the `set +x` be here?

contrib/managed-azure/setup_oidc_provider.sh
```diff
@@ -0,0 +1,23 @@
...
+
+ccoctl azure create-oidc-issuer \
+--oidc-resource-group-name $PERSISTENT_RG_NAME \
+--tenant-id $TENANT_ID \
+--region ${PERSISTENT_RG_LOCATION} \
+--name $OIDC_ISSUER_NAME \
+--subscription-id $SUBSCRIPTION_ID \
+--public-key-file $SA_TOKEN_ISSUER_PUBLIC_KEY_PATH
+
+
```
bryan-cox: What's with all the blank lines here?

contrib/managed-azure/user-vars.sh
```diff
@@ -0,0 +1,9 @@
+# User variables.
```
bryan-cox: You should remove your specific naming from here

contrib/managed-azure/setup_install_ho_on_aks.sh
```diff
@@ -1,31 +1,32 @@
 #!/bin/bash
 set -x
 
-# Constants
-MGMT_DNS_ZONE_NAME="azure.blah.com"
-PULL_SECRET="/Users/your-username/all-the-pull-secrets.json"
-SERVICE_PRINCIPAL_FILEPATH="/Users/your-username/azure_mgmt.json"
-TAG="quay.io/user/hypershift:mytag1"
+# Prerequisites.
+HYPERSHIFT_BINARY_PATH="${HYPERSHIFT_BINARY_PATH:-}"
```
bryan-cox, line 5: Shouldn't these just be in contrib/managed-azure/vars.sh?

contrib/managed-azure/setup_etcd_kv.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -x
+set +x
 
 # Prerequisites: define these constants
 LOCATION="eastus"
```
bryan-cox, line 5: Shouldn't this just be in contrib/managed-azure/vars.sh?

contrib/managed-azure/setup_dataplane_identities.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
+set -x
+
+DP_OUTPUT_FILE=${DP_OUTPUT_FILE:-}
+
+AZURE_DISK_MI_NAME="azure-disk-MI-${PREFIX}"
```
bryan-cox, line 10: Shouldn't these just be in contrib/managed-azure/vars.sh?

contrib/managed-azure/setup_external_dns.sh
```diff
@@ -1,16 +1,33 @@
...
 
-# Constants
-RG="external-dns"
-LOCATION="eastus"
-MGMT_DNS_ZONE_NAME="blah-blah-blah.com"
-EXTERNAL_DNS_NEW_SP_NAME="ExternalDnsServicePrincipal"
-SERVICE_PRINCIPAL_FILEPATH="/Users/your-username/azure_mgmt.json"
+# Prerequisites.
+LOCATION="${LOCATION:-}"
+PARENT_DNS_RG_NAME="${PERSISTENT_RG_NAME}"
```
bryan-cox: Shouldn't this just be in contrib/managed-azure/vars.sh?

contrib/managed-azure/setup_aks_cluster.sh
```diff
@@ -2,39 +2,33 @@
...
-#    AKS_KUBELET_MI, in a resource group that is persistent in your tenant/subscription.
+#  setup_MIv3_kv.sh
+PREFIX="${PREFIX:-}"
+SUBSCRIPTION_ID=${SUBSCRIPTION_ID:-}
+LOCATION="${LOCATION:-}"
 
-# Constants
-PREFIX="<your-prefix-here>"
+# Local.
 AKS_RG=${PREFIX}"-aks-rg"
```
bryan-cox: Shouldn't these just be in contrib/managed-azure/vars.sh?

contrib/managed-azure/create_basic_hosted_cluster.sh
```diff
@@ -8,31 +8,25 @@ set -x
...
+PULL_SECRET="${PULL_SECRET:-}"
+HYPERSHIFT_BINARY_PATH="${HYPERSHIFT_BINARY_PATH:-}"
+CP_OUTPUT_FILE="${CP_OUTPUT_FILE:-}"
+DP_OUTPUT_FILE="${DP_OUTPUT_FILE:-}"
+OIDC_ISSUER_URL="${OIDC_ISSUER_URL:-}"
+SA_TOKEN_ISSUER_PRIVATE_KEY_PATH="${SA_TOKEN_ISSUER_PRIVATE_KEY_PATH:-}"
+
+# Local.
+PREFIX="${PREFIX:-}"
 LOCATION="eastus"
```
bryan-cox: Shouldn't these just be in contrib/managed-azure/vars.sh?

contrib/managed-azure/setup_oidc_provider.sh
```diff
@@ -0,0 +1,23 @@
...
+set -x
+
+# Prerequisites.
+OIDC_ISSUER_NAME="${OIDC_ISSUER_NAME:-}"
+
+# Local.
+SA_TOKEN_ISSUER_PUBLIC_KEY_PATH=${SA_TOKEN_ISSUER_PUBLIC_KEY_PATH:-}
+SA_TOKEN_ISSUER_PRIVATE_KEY_PATH=${SA_TOKEN_ISSUER_PRIVATE_KEY_PATH:-}
+PERSISTENT_RG_LOCATION="${PERSISTENT_RG_LOCATION:-}"
+
```
bryan-cox, line 14: Can you add a comment on the source of this tool? - Somthing like see https://github.com/nutanix-cloud-native/openshift-cloud-credential-operator/blob/master/docs/ccoctl.md#creating-rsa-keys-1 for more details.

contrib/managed-azure/setup_oidc_provider.sh
```diff
@@ -0,0 +1,23 @@
...
+ccoctl azure create-oidc-issuer \
```
bryan-cox, line 17: If this works, that's fine, but it's a deviation from what is done today
```
~/cloud-credential-operator/ccoctl azure create-all --name oidc-issuer \
                          --output-dir ~/temp-cco \
                          --region eastus \
                          --subscription-id ${SUBSCRIPTION_ID} \
                          --tenant-id ${TENANT_ID} \
                          --credentials-requests-dir ~/temp-cco/credrequests \
                          --dnszone-resource-group-name ${PARENT_DNS_RG}  
```

contrib/managed-azure/create_basic_hosted_cluster.sh
```diff
@@ -110,16 +88,13 @@ ${HYPERSHIFT_BINARY_PATH}/hypershift create cluster azure \
 --sa-token-issuer-private-key-path "${SA_TOKEN_ISSUER_PRIVATE_KEY_PATH}" \
 --oidc-issuer-url "${OIDC_ISSUER_URL}" \
 --annotations hypershift.openshift.io/pod-security-admission-label-override=baseline \
---control-plane-operator-image=<YOUR_REPO>:"${TAG}" \
---managed-identities-file ${MANAGED_IDENTITIES_FILE} \
---fips=true \
```
bryan-cox, line 115: Are you wanting to remove the FIPS part or was that accidentally taken out when you removed the other two fields?

contrib/managed-azure/create_basic_hosted_cluster.sh
```diff
@@ -64,38 +58,22 @@ GetVnetID=$(az network vnet list --query "[?name=='${CUSTOMER_VNET_NAME}'].id" -
...
+cat <<EOF > ./azure-creds.json
+{
+  "subscriptionId": "$SUBSCRIPTION_ID",
+  "tenantId": "$TENANT_ID",
+  "clientId": "$CLIENT_ID",
+  "clientSecret": "$CLIENT_SECRET"
+}
+EOF
 
 ${HYPERSHIFT_BINARY_PATH}/hypershift create cluster azure \
```
bryan-cox, line 61: These flags should be added as well - 
```
--network-plugin azure \
--network-policy azure 
```

contrib/managed-azure/vars.sh
```diff
@@ -0,0 +1,27 @@
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+# User variables.
+source "${SCRIPT_DIR}/user-vars.sh"
+
+# Internal Variables used by the scripts.
+# We export them here so any script can be commented out from setup_all.sh without breaking the other scripts.
+export PERSISTENT_RG_NAME="os4-common"
+export PARENT_DNS_ZONE="hypershift.azure.devcluster.openshift.com"
```
bryan-cox, line 18: Above this I would incorporate what we have in the docs now as reference for other teams - 
```
########################################################################################################################
# Tip - the value of `$PARENT_DNS_ZONE` may be different for different teams. Check `os4-common` resource group 
# associated with your subscription-id for pre-existing DNS zones.  If there are multiple DNS zones more than likely 
# the parent DNS zone is the common suffix among them.
########################################################################################################################
```

contrib/managed-azure/setup_external_dns.sh
```diff
@@ -23,7 +40,7 @@ az role assignment create --role "Reader" --assignee "${EXTERNAL_DNS_SP_APP_ID}"
 az role assignment create --role "Contributor" --assignee "${EXTERNAL_DNS_SP_APP_ID}" --scope "${DNS_ID}"
 
 # Creating a configuration file for our service principal
-cat <<-EOF > ${SERVICE_PRINCIPAL_FILEPATH}
+cat <<-EOF > ${EXTERNAL_DNS_SERVICE_PRINCIPAL_FILEPATH}
```
bryan-cox, line 43: At L37 (ish), I would add this tip from the current doc for other teams reference - 
```
########################################################################################################################
# Tip - If your DNS Zone is not in the same resource group where the parent DNS zone is located, you will need to add 
# these same role assignments to the resource group where the parent DNS zone is located.
########################################################################################################################
```

contrib/managed-azure/setup_install_ho_on_aks.sh
```diff
@@ -1,31 +1,32 @@
...
 --enable-conversion-webhook=false \
 --external-dns-provider=azure \
---external-dns-credentials ${SERVICE_PRINCIPAL_FILEPATH} \
+--external-dns-credentials ${EXTERNAL_DNS_SERVICE_PRINCIPAL_FILEPATH} \
 --pull-secret ${PULL_SECRET} \
 --external-dns-domain-filter ${MGMT_DNS_ZONE_NAME} \
 --managed-service ARO-HCP \
 --aro-hcp-key-vault-users-client-id "${AZURE_KEY_VAULT_AUTHORIZED_USER_ID}" \
---hypershift-image quay.io/username/hypershift:${TAG} \
+--hypershift-image ${HYPERSHIFT_IMAGE} \
```
bryan-cox, line 30: The control plane operator version of this was dropped from the create cluster script. Should it be dropped here too?

bryan-cox, line 30: I would suggest to add this flag - 
```
--limit-crd-install Azure \
```

contrib/managed-azure/README.md
```diff
@@ -1,15 +1,16 @@
...
-script. If you are starting out with nothing set up, you'll want to go through the scripts in this order:
+Steps:
 
-1. [Review Steps 1-7 to Set Up Environment Variables and Such](../../docs/content/how-to/azure/create-azure-cluster_on_aks.md)
-2. [Create an AKS management cluster](setup_aks_cluster.sh)
-3. [Set up externalDNS](setup_external_dns.sh)
-4. [Install the HyperShift Operator](setup_install_ho_on_aks.sh)
-5. [Create a HostedCluster](create_basic_hosted_cluster.sh)
\ No newline at end of file
+Set your input in user-vars.sh
```
bryan-cox: Maybe we can call this section out more to make sure users don't miss this? Something like
```
!!! tip

    The scripts may have some prerequisite expectations so please review the script in whole before running it.

    Constants will also need to be updated in each script for your exact tenant and environment.
```

contrib/managed-azure/create_basic_hosted_cluster.sh
```diff
@@ -110,16 +88,13 @@ ${HYPERSHIFT_BINARY_PATH}/hypershift create cluster azure \
...
 --marketplace-publisher azureopenshift \
 --marketplace-offer aro4 \
 --marketplace-sku aro_417 \
 --marketplace-version 417.94.20240701 \
 --dns-zone-rg-name os4-common \
 --assign-service-principal-roles \
---managed-identities-file /Users/your username/aro-hcp-control-plane-MIs.json \
---data-plane-identities-file /Users/your username/aro-hcp-data-plane-MIs.json
+--managed-identities-file $CP_OUTPUT_FILE \
+--data-plane-identities-file $DP_OUTPUT_FILE
```
bryan-cox: Should we also add the diagnostics flag? It let's you see the serial console on the VM.

contrib/managed-azure/setup_install_ho_on_aks.sh
bryan-cox: This is in contrib/managed-azure/user-vars.sh

contrib/managed-azure/setup_aks_cluster.sh
bryan-cox: Discussed on Slack, from Alberto - I put in that file vars that need to be consumed and are shared by more than one script. So you can comment out any script and it'll still work. If that's no the case I just put them behind "local".
So for each script you have prerequisists (come from user or script shared) and local.

contrib/managed-azure/setup_dataplane_identities.sh
bryan-cox: Discussed on Slack, from Alberto - I put in that file vars that need to be consumed and are shared by more than one script. So you can comment out any script and it'll still work. If that's no the case I just put them behind "local".
So for each script you have prerequisists (come from user or script shared) and local.

contrib/managed-azure/create_basic_hosted_cluster.sh
bryan-cox: Discussed on Slack, from Alberto - I put in that file vars that need to be consumed and are shared by more than one script. So you can comment out any script and it'll still work. If that's no the case I just put them behind "local".
So for each script you have prerequisists (come from user or script shared) and local.

contrib/managed-azure/setup_external_dns.sh
bryan-cox: Discussed on Slack, from Alberto - I put in that file vars that need to be consumed and are shared by more than one script. So you can comment out any script and it'll still work. If that's no the case I just put them behind "local".
So for each script you have prerequisists (come from user or script shared) and local.

contrib/managed-azure/setup_etcd_kv.sh
bryan-cox: Discussed on Slack, from Alberto - I put in that file vars that need to be consumed and are shared by more than one script. So you can comment out any script and it'll still work. If that's no the case I just put them behind "local".
So for each script you have prerequisists (come from user or script shared) and local.

contrib/managed-azure/create_basic_hosted_cluster.sh
enxebre: diagnostics-storage-account-type?

contrib/managed-azure/setup_aks_cluster.sh
enxebre: moved to miv3 script

contrib/managed-azure/create_basic_hosted_cluster.sh
bryan-cox: Without looking, yeah I think that's it. I would suggest the value of Managed.

contrib/managed-azure/create_basic_hosted_cluster.sh
enxebre: remove it for now

bryan-cox:
/retitle NO-JIRA: Fix up contrib scripts for aro like env

PR #6365, opened 2025-07-02
OCPSTRAT-2173: Revert "Merge pull request #6227 from sjenning/hcco-empty-client-secret"

devguyio:
**What this PR does / why we need it**:

The first implementation of #[OCPSTRAT-2173](https://issues.redhat.com//browse/OCPSTRAT-2173) had the side effect that when adding an OIDC client with an empty secret name, which is the case for cli clients, a default secret name will be generated. 

This behavior is now undesired and reverting this is the easiest way to fix and backport #OCPBUGS-58149.

After reverting we'll stop creating a default client secret name if no secret name is given. However #OCPBUGS-58149 will still exist, that's why this should be followed up directly by a fix for [OCPBUGS-58149](https://issues.redhat.com/browse/OCPBUGS-58149)

**Checklist**
- [x] Subject and description added to both, commit and PR.
- [x] Relevant issues have been referenced.
- [ ] This change includes docs. 
- [ ] This change includes unit tests.

PR #6296, opened 2025-06-18
[release-4.19] CNTRLPLANE-374: hcco: handle post-install creation of OIDC client secrets

sjenning:
Clean 4.19 backport of https://github.com/openshift/hypershift/pull/6227

PR #6398, opened 2025-07-08
[release-4.19] OCPBUGS-58683:fix(hcco): Don't use a default OIDC client secret name when no secret name is provided 

devguyio:

Effectively, this is  back-port of two PRs: #6365 and #6367 .

This change does the following:
- Removes the secret defaulting logic for OIDC clients with no secret name.
- Adds unit tests for the reconcileAuthOIDC function.

The functionality originally intended by #6296 will be introduced in a different way in a separate PR. 

There's currently a draft PR for it on main #6375 

Fixes OCPBUGS-58683

**Checklist**

### DIFF
--- a/control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
+++ b/control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
@@ -1150,7 +1150,8 @@ func (r *reconciler) reconcileAuthOIDC(ctx context.Context, hcp *hyperv1.HostedC
 
 		// Copy OIDCClient Secrets into openshift-config namespace
 		for _, oidcClient := range provider.OIDCClients {
-			// If the secret name is empty, we assume the guest cluster admin is going to create this secret manually
+			// Public OIDC clients have no secrets, so the secret name will be the empty string since ClientSecret is
+			// optional and not a pointer
 			if oidcClient.ClientSecret.Name == "" {
 				log.Info("OIDC client secret is empty, skipping reconciliation", "component", oidcClient.ComponentName)
 				continue
--- a/control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources_test.go
+++ b/control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources_test.go
@@ -1391,3 +1391,554 @@ func makeKubeletConfigConfigMap(name, namespace, data string) *corev1.ConfigMap 
 		},
 	}
 }
+
+func TestReconcileAuthOIDC(t *testing.T) {
+	log := controllerruntime.LoggerFrom(context.Background())
+	testNamespace := "master-cluster1"
+	testHCPName := "cluster1"
+
+	tests := map[string]struct {
+		inputHCP                *hyperv1.HostedControlPlane
+		inputCPObjects          []client.Object
+		expectIssuerCAConfigMap bool
+		expectOIDCClientSecrets []string
+		expectErrors            bool
+		expectedErrorMessages   []string
+	}{
+		"when OAuth is enabled, should not copy OIDC resources": {
+			inputHCP: &hyperv1.HostedControlPlane{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      testHCPName,
+					Namespace: testNamespace,
+				},
+				Spec: hyperv1.HostedControlPlaneSpec{
+					Configuration: &hyperv1.ClusterConfiguration{
+						OAuth: &configv1.OAuthSpec{
+							IdentityProviders: []configv1.IdentityProvider{
+								{
+									Name: "test-provider",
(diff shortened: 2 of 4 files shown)


control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -1115,7 +1115,7 @@ func (r *reconciler) reconcileIngressController(ctx context.Context, hcp *hyperv
 	return errors.NewAggregate(errs)
 }
 
-func (r *reconciler) reconcileAuthOIDC(ctx context.Context, hcp *hyperv1.HostedControlPlane, log logr.Logger) error {
+func (r *reconciler) reconcileAuthOIDC(ctx context.Context, hcp *hyperv1.HostedControlPlane) error {
 	var errs []error
 	if !util.HCPOAuthEnabled(hcp) &&
```
csrwng: I know that !HCPOAuthEnabled ensures that Configuration and Authentication are non-nil, however, that's the kind of assumption that can be lost in future refactorings. I'd prefer to explicitly check that they are non-nil before `len(hcp.Spec.Configuration.Authentication.OIDCProviders) != 0`

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -1115,7 +1115,7 @@ func (r *reconciler) reconcileIngressController(ctx context.Context, hcp *hyperv
...
 		len(hcp.Spec.Configuration.Authentication.OIDCProviders) != 0 {
```
csrwng: Following go conventions, I'd rather return early than nest all of the code within this condition.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -1149,33 +1149,32 @@ func (r *reconciler) reconcileAuthOIDC(ctx context.Context, hcp *hyperv1.HostedC
 		}
```
csrwng: Line 1130 above... if we fail to fetch the CA configmap, should we not abandon reconciliation? (same with CreateOrUpdate)

csrwng: What is the indication to the API user that there's been a failure here?

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -1149,33 +1149,32 @@ func (r *reconciler) reconcileAuthOIDC(ctx context.Context, hcp *hyperv1.HostedC
...
-			dest := corev1.Secret{
-				ObjectMeta: metav1.ObjectMeta{
-					Name:      oidcClient.ClientSecret.Name,
-					Namespace: ConfigNamespace,
-				},
-			}
-			_, err = r.CreateOrUpdate(ctx, r.client, &dest, func() error {
-				if dest.Data == nil {
-					dest.Data = map[string][]byte{}
+		if len(hcp.Spec.Configuration.Authentication.OIDCProviders[0].OIDCClients) > 0 {
```
csrwng: seems like an unnecessary check, given that you're going to loop through the OIDCClients below

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
csrwng: If we're only going to support one OIDCProvider for now (hcp.Spec.Configuration.Authentication.OIDCProviders[0]) then we should add a comment here stating so.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -1149,33 +1149,32 @@ func (r *reconciler) reconcileAuthOIDC(ctx context.Context, hcp *hyperv1.HostedC
...
+			for _, oidcClient := range hcp.Spec.Configuration.Authentication.OIDCProviders[0].OIDCClients {
+				if oidcClient.ClientSecret.Name != "" {
```
csrwng: go convention: exit/continue early

csrwng (commented):
Some comments. None of these suggestions should block the merging of this PR (that function is fixed is more important). They can be handled in a follow up.

devguyio:
> Some comments. None of these suggestions should block the merging of this PR (that function is fixed is more important). They can be handled in a follow up.

Thanks for the feedback! Yeah will definitely address that on main and backport the changes in a separate PR. I'm already working on that area and any feedback for code improvement is appreciated 🙏🏽

devguyio:
@csrwng I updated the code to make the change easier to view in the diff. Can you please take another look?

mgencur:
LGTM

PR #6411, opened 2025-07-09
CNTRLPLANE-385: programmatically enable readonlyRootFilesystem

dusk125:
Enables readonlyRootFilesystem by default on all hypershift containers to close a potential security loophole where a bad actor could edit the contents of a container for malicious purposes.

Fixes: https://issues.redhat.com/browse/CNTRLPLANE-385

Checklist

## Summary by CodeRabbit

* **New Features**
  * Control-plane containers now run with read-only root filesystems and a shared ephemeral tmp-dir mounted at /tmp; tmp-dir is marked safe-to-evict for the cluster autoscaler.

* **Tests**
  * Added e2e checks to validate read-only root filesystems, verify tmp-dir mounts, and strengthened version-gated validation of safe-to-evict annotations.

* **Chores**
  * Updated manifests and test fixtures across control-plane components to adopt tmp-dir and the new security defaults.

### DIFF
--- a/support/controlplane-component/defaults.go
+++ b/support/controlplane-component/defaults.go
@@ -121,6 +121,7 @@ func (c *controlPlaneWorkload[T]) setDefaultOptions(cpContext ControlPlaneContex
 
 	enforceTerminationMessagePolicy(podTemplateSpec.Spec.InitContainers)
 	enforceTerminationMessagePolicy(podTemplateSpec.Spec.Containers)
+	enforceReadOnlyRootFilesystem(&podTemplateSpec.Spec)
 
 	if _, exist := podTemplateSpec.Annotations[config.NeedMetricsServerAccessLabel]; exist || c.NeedsManagementKASAccess() ||
 		c.Name() == "packageserver" { // TODO: investigate why packageserver needs AutomountServiceAccountToken or set NeedsManagementKASAccess to true.
@@ -554,6 +555,33 @@ func enforceImagePullPolicy(containers []corev1.Container) error {
 		containers[i].ImagePullPolicy = corev1.PullIfNotPresent
 	}
 	return nil
+}
+
+func enforceReadOnlyRootFilesystem(podSpec *corev1.PodSpec) {
+	podSpec.Volumes = append(podSpec.Volumes, corev1.Volume{
+		Name: util.PodTmpDirMountName,
+		VolumeSource: corev1.VolumeSource{
+			EmptyDir: &corev1.EmptyDirVolumeSource{},
+		},
+	})
+	enforceReadOnlyRootFilesystemContainers(podSpec.Containers)
+}
+
+func enforceReadOnlyRootFilesystemContainers(containers []corev1.Container) {
+	for i := range containers {
+		if containers[i].SecurityContext == nil {
+			containers[i].SecurityContext = &corev1.SecurityContext{}
+		}
+		if !slices.ContainsFunc(containers[i].VolumeMounts, func(vm corev1.VolumeMount) bool {
+			return vm.MountPath == util.PodTmpDirMountPath
+		}) {
+			containers[i].VolumeMounts = append(containers[i].VolumeMounts, corev1.VolumeMount{
+				Name:      util.PodTmpDirMountName,
+				MountPath: util.PodTmpDirMountPath,
+			})
+		}
+		containers[i].SecurityContext.ReadOnlyRootFilesystem = ptr.To(true)
+	}
 }
 
 func enforceTerminationMessagePolicy(containers []corev1.Container) {
--- a/test/e2e/util/util.go
+++ b/test/e2e/util/util.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"slices"
 	"strconv"
 	"strings"
 	"testing"
@@ -1420,7 +1421,10 @@ func getExportedMetrics(ctx context.Context, log logr.Logger, c crclient.Client)
 }
 
 func EnsurePodsWithEmptyDirPVsHaveSafeToEvictAnnotations(t *testing.T, ctx context.Context, hostClient crclient.Client, hcpNs string) {
+	AtLeast(t, Version420)
+
 	t.Run("EnsurePodsWithEmptyDirPVsHaveSafeToEvictAnnotations", func(t *testing.T) {
+
 		g := NewWithT(t)
 
 		auditedAppList := map[string]string{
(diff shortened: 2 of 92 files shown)


celebdor (changes requested):
Please fill out the PR description template

dusk125:
> Please fill out the PR description template

Done

dusk125:
/hold
tests are looking good with the tmp-dir mount added by default. Need to address properly updating the test now

test/e2e/util/util.go
```diff
@@ -1461,10 +1462,22 @@ func EnsurePodsWithEmptyDirPVsHaveSafeToEvictAnnotations(t *testing.T, ctx conte
...
 				// Then continue to the next pod
-				g.Expect(pod.Annotations[suppconfig.PodSafeToEvictLocalVolumesKey]).To(BeEmpty(), "the pod  %s is not in the audited list for safe-eviction and should not contain the safe-to-evict-local-volume annotation", pod.Name)
-				continue
+				hasTmpDir := false
+				safe2EvictVolumes := strings.Split(pod.Annotations[suppconfig.PodSafeToEvictLocalVolumesKey], ",")
+				safe2EvictVolumes = slices.DeleteFunc(safe2EvictVolumes, func(s string) bool {
+					if s == "" {
+						return true
+					}
+					hasTmpDir = s == "tmp-dir"
```
csrwng: "tmp-dir" should be a constant used here and in defaults.go

test/e2e/util/util.go
```diff
@@ -1461,10 +1462,22 @@ func EnsurePodsWithEmptyDirPVsHaveSafeToEvictAnnotations(t *testing.T, ctx conte
...
 				// if the Key/Value are empty we assume that the pod is not in the auditedList,
-				// if that's the case the annotation should not exists in that pod.
+				// if that's the case the annotation should not exists in that pod (except for tmp-dir which is in every pod by default).
 				// Then continue to the next pod
-				g.Expect(pod.Annotations[suppconfig.PodSafeToEvictLocalVolumesKey]).To(BeEmpty(), "the pod  %s is not in the audited list for safe-eviction and should not contain the safe-to-evict-local-volume annotation", pod.Name)
-				continue
+				hasTmpDir := false
+				safe2EvictVolumes := strings.Split(pod.Annotations[suppconfig.PodSafeToEvictLocalVolumesKey], ",")
+				safe2EvictVolumes = slices.DeleteFunc(safe2EvictVolumes, func(s string) bool {
+					if s == "" {
```
csrwng: nit:
```
return s == "" || s == CONSTANT_FOR_tmp_dir
```

test/e2e/util/util.go
```diff
@@ -1479,6 +1492,97 @@ func EnsurePodsWithEmptyDirPVsHaveSafeToEvictAnnotations(t *testing.T, ctx conte
...
+		hcpPods := &corev1.PodList{}
+		if err := hostClient.List(ctx, hcpPods, &client.ListOptions{
+			Namespace: hcpNs,
+		}); err != nil {
+			t.Fatalf("cannot list hostedControlPlane pods: %v", err)
+		}
+
+		// a list of applications that are allowed to have Pod.Spec.Containers[*].SecurityContext.ReadOnlyRootFilesystem == false
+		// auditedAppContainersNoRORFS[pod.Name][pod.Spec.Containers[*]] indicates that particular container is allowed to be false
+		auditedAppContainersNoRORFS := map[string]map[string]struct{}{}
```
csrwng: We can't know pod names ahead of time (they're generated from a replicaset, statefulset, etc). If we're going to have an exclude list it will likely need to be via label selector.

test/e2e/util/util.go
```diff
@@ -1479,6 +1492,97 @@ func EnsurePodsWithEmptyDirPVsHaveSafeToEvictAnnotations(t *testing.T, ctx conte
...
+
+		for _, pod := range hcpPods.Items {
+			// skip etcd and feature-gate-generator pods
+			// If added to the list of audited pods,it will fail the e2e check on older release branches since e2e is ran from main.
```
csrwng, line 1578: wouldn't this entire validation fail for older releases since most pods won't have these changes?
I'd think you'd want a check like 
https://github.com/openshift/hypershift/blob/f64e73d54a6d3759b5511d61bc78d58b1b392c53/test/e2e/autoscaling_test.go#L149
at the beginning of the function

test/e2e/util/util.go
dusk125: So I intentionally separated out this condition since I need to distinguish later if the pod has tmpdir in the annotation, but we need to remove both the tmpdir annotation and any empty value from the safe2EvictVolumes slice for the check below to function properly

dusk125:
/hold cancel
tests are looking good

support/controlplane-component/defaults.go
```diff
@@ -83,6 +83,7 @@ func (c *controlPlaneWorkload[T]) setDefaultOptions(cpContext ControlPlaneContex
 
 	podTemplateSpec := c.workloadProvider.PodTemplateSpec(workloadObj)
 	enforceVolumesDefaultMode(&podTemplateSpec.Spec)
+	enforceReadOnlyRootFilesystem(&podTemplateSpec.Spec)
```
muraee: This call should be moved after konnectivity, token-minter and availabilityProber containers are injected. i.e. after defaults.go#L122.
You wouldn't need to duplicate this logic in konnecitivty-cotainer.go and token-minter-container.go then

dusk125: Good call, I've updated.

test/e2e/util/util.go
```diff
@@ -1487,6 +1501,165 @@ func EnsurePodsWithEmptyDirPVsHaveSafeToEvictAnnotations(t *testing.T, ctx conte
...
+			{label: "app", value: "azure-disk-csi-driver-controller"}: {},
+			{label: "app", value: "azure-disk-csi-driver-operator"}:   {},
+			{label: "app", value: "azure-file-csi-driver-controller"}: {},
+			{label: "app", value: "azure-file-csi-driver-operator"}:   {},
+			{label: "app", value: "aws-ebs-csi-driver-controller"}:    {},
+			{label: "app", value: "aws-ebs-csi-driver-operator"}:      {},
+			{label: "app", value: "multus-admission-controller"}:      {},
+			{label: "app", value: "network-node-identity"}:            {},
+			{label: "app", value: "ovnkube-control-plane"}:            {},
+			{label: "app", value: "packageserver"}: {
```
csrwng, line 1630: nit: pls add a comment for the reason that this is here. I assume it's because the packageserver container already had a tmpfs for /tmp and it's not named "tmp-dir". Or alternatively, you could just check that they have a volume mount for /tmp and not exactly matching the volume name "tmp-dir". 

Actually I think the latter option may be better b/c when we propagate this to the storage and cno operators, they may not use the same exact same volume name as us.

dusk125: Addressed in latest commits

PR #6420, opened 2025-07-11
OCPBUGS-58299: remove annotation is not working

zhfeng:

The JSON patch was working, but Go's map decoding behavior left stale keys in the Annotations map.
By clearing the map before decoding, the struct now matches the patched JSON, and the test passes.

Fixes #[OCPBUGS-58299](https://issues.redhat.com/browse/OCPBUGS-58299)

**Checklist**

### DIFF
--- a/hypershift-operator/controllers/nodepool/kubevirt/kubevirt_test.go
+++ b/hypershift-operator/controllers/nodepool/kubevirt/kubevirt_test.go
@@ -1121,6 +1121,60 @@ func TestJsonPatch(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "remove annotation in the nodepool",
+			nodePool: &hyperv1.NodePool{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      poolName,
+					Namespace: namespace,
+					Annotations: map[string]string{
+						hyperv1.JSONPatchAnnotation: `[{"op": "remove","path": "/spec/template/metadata/annotations/kubevirt.io~1allow-pod-bridge-network-live-migration"}]`,
+					},
+				},
+				Spec: hyperv1.NodePoolSpec{
+					ClusterName: clusterName,
+					Replicas:    nil,
+					Config:      nil,
+					Management:  hyperv1.NodePoolManagement{},
+					AutoScaling: nil,
+					Platform: hyperv1.NodePoolPlatform{
+						Type: hyperv1.KubevirtPlatform,
+						Kubevirt: generateKubevirtPlatform(
+							memoryNPOption("5Gi"),
+							coresNPOption(4),
+							imageNPOption("testimage"),
+							volumeNPOption("32Gi"),
+						),
+					},
+					Release: hyperv1.Release{},
+				},
+			},
+			hcluster: &hyperv1.HostedCluster{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "my-hostedcluster",
+					Namespace: "clusters",
+				},
+				Spec: hyperv1.HostedClusterSpec{
+					InfraID: "1234",
+				},
+			},
+
+			expected: &capikubevirt.KubevirtMachineTemplateSpec{
+				Template: capikubevirt.KubevirtMachineTemplateResource{
+					Spec: capikubevirt.KubevirtMachineSpec{
+						BootstrapCheckSpec: capikubevirt.VirtualMachineBootstrapCheckSpec{CheckStrategy: "none"},
+						VirtualMachineTemplate: *generateNodeTemplate(
+							memoryTmpltOpt("5Gi"),
+							cpuTmpltOpt(4),
+							storageTmpltOpt("32Gi"),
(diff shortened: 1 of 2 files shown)


zhfeng:
@qinqon can you take a review? and I think this is related to the fix of the following PR:
- https://github.com/openshift/hypershift/pull/6363

hypershift-operator/controllers/nodepool/kubevirt/kubevirt_test.go
```diff
@@ -1133,6 +1184,16 @@ func TestJsonPatch(t *testing.T) {
...
 			g.Expect(err).ToNot(HaveOccurred())
+
+			// Special verification for the annotation removal test
+			if tc.name == "remove annotation in the nodepool" {
+				annotations := result.Template.Spec.VirtualMachineTemplate.Spec.Template.ObjectMeta.Annotations
+				g.Expect(annotations).ToNot(HaveKey("kubevirt.io/allow-pod-bridge-network-live-migration"),
+					"Annotation should be removed by JSON patch")
+				g.Expect(annotations).To(HaveKey("cluster-autoscaler.kubernetes.io/safe-to-evict-local-volumes"),
+					"Other annotations should remain")
+			}
```
qinqon: This quite harcoded and dirty, please add enough test infra to do this kind of checks without harcoding a test name check.

zhfeng: Thanks @qinqon - yeah, I think this is only a functional unit test. And IIUC, all of them are with the mock environments, it is only to test `MachineTemplateSpec` func. I will try to add an automate test case in the openshift-test-private repo to run this checking in a real infra envirnment.

@davidvossel and @nirarg WDYT?

zhfeng: @qinqon I just remove the hardcoding of the test name. can you take a review again? Thanks!

hypershift-operator/controllers/nodepool/kubevirt/kubevirt_test.go
```diff
@@ -1133,6 +1186,16 @@ func TestJsonPatch(t *testing.T) {
...
+
+			// Special verification for the annotation removal test
+			if tc.expectAnnotationRemoved {
+				annotations := result.Template.Spec.VirtualMachineTemplate.Spec.Template.ObjectMeta.Annotations
+				g.Expect(annotations).ToNot(HaveKey("kubevirt.io/allow-pod-bridge-network-live-migration"),
+					"Annotation should be removed by JSON patch")
+				g.Expect(annotations).To(HaveKey("cluster-autoscaler.kubernetes.io/safe-to-evict-local-volumes"),
+					"Other annotations should remain")
+			}
+
```
qinqon: This is still harcoding what annotations will be missing at the test running code, that should be done at the test sepc, we want to have annoter opìon at `generateNodeTemplate` to override annotation so we can do 
```golang
expected: &capikubevirt.KubevirtMachineTemplateSpec{
				Template: capikubevirt.KubevirtMachineTemplateResource{
					Spec: capikubevirt.KubevirtMachineSpec{
						BootstrapCheckSpec: capikubevirt.VirtualMachineBootstrapCheckSpec{CheckStrategy: "none"},
						VirtualMachineTemplate: *generateNodeTemplate(false,
							memoryTmpltOpt("5Gi"),
							cpuTmpltOpt(4),
							storageTmpltOpt("32Gi"),
                                                        annotationsTmplItOpt(map[string]string{
                                                                 	suppconfig.PodSafeToEvictLocalVolumesKey:              strings.Join(LocalStorageVolumes, ","),
                                                         })

						),
					},
				},
			},
```

zhfeng: OK, I'd update the tests with your suggestions. Thanks a lot!

hypershift-operator/controllers/nodepool/kubevirt/kubevirt_test.go
```diff
@@ -1133,6 +1189,16 @@ func TestJsonPatch(t *testing.T) {
...
+			// Special verification for the annotation removal test
+			if tc.expectAnnotationRemoved {
+				annotations := result.Template.Spec.VirtualMachineTemplate.Spec.Template.ObjectMeta.Annotations
+				g.Expect(annotations).ToNot(HaveKey("kubevirt.io/allow-pod-bridge-network-live-migration"),
+					"Annotation should be removed by JSON patch")
+				g.Expect(annotations).To(HaveKey(suppconfig.PodSafeToEvictLocalVolumesKey),
+					"Other annotations should remain")
+			}
+
```
qinqon: We don't need this anymore

qinqon (changes requested):

hypershift-operator/controllers/nodepool/kubevirt/kubevirt_test.go
zhfeng: Done

PR #6445, opened 2025-07-14
Revert "OCPBUGS-56701: Update the LookupDefaultOCPVersion function to use the multi-arch release API"

bryan-cox:
A PR was originally introduced here - openshift/hypershift#6353

It is being reverted because caused the AWS e2e to fail as the AWS management cluster could not be created because no release image could be found.

This PR is now reverting openshift/hypershift#6353. Before it is remerged, we will make sure the AWS e2e passes with it.

### DIFF
--- a/cmd/cluster/agent/create_test.go
+++ b/cmd/cluster/agent/create_test.go
@@ -2,14 +2,11 @@ package agent
 
 import (
 	"context"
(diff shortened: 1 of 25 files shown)


enxebre:
can you please elaborate why this revert in the PR desc?

PR #6451, opened 2025-07-15
NO-JIRA: Update Created Azure Hosted Cluster on AKS doc

bryan-cox:
This PR overhauls Azure hosted cluster setup guide for automation-first workflow

- Replace manual shell commands with references to automation scripts in contrib/managed-azure/
- Add clear goals and script links for each setup step  
- Consolidate one-time vs per-cluster setup distinction throughout documentation
- Group one-time setup scripts together in setup_all.sh (setup_MIv3_kv.sh, setup_oidc_provider.sh, setup_dataplane_identities.sh)
- Reorder manual steps to match script execution sequence for consistency
- Add GitHub permalink links to all automation scripts for easy access
- Fix markdown formatting for proper bullet point rendering
- Remove extensive manual variable configuration in favor of simple user-vars.sh approach
- Update cluster deletion constants to match actual script variables
- Add warnings about Azure quota management and resource reuse across multiple clusters

N/A

**Checklist**


contrib/managed-azure/setup_all.sh
```diff
@@ -4,12 +4,16 @@ set -x
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 
 source "${SCRIPT_DIR}/vars.sh"
+
+# One-time setup scripts (comment out these three after first run to reuse resources)
 "${SCRIPT_DIR}/setup_MIv3_kv.sh"
```
jparrill: I would add a flag to execute these three scripts as a first time, otherwise just go with `per-cluster setup scripts`

jparrill (changes requested):
This way works perfectly, I would add a security measure to avoid the execution of the **one-time setup scripts** to be not executed as default and then add the step in the documentation.

jparrill:
Ohh and I would add another function to the `user_vars.sh` or `vars.sh` like this:

```sh
function login() {
    az login --scope https://management.core.windows.net//.default
}
```

This helps a lot just using `login` command in the CLI

PR #6301, opened 2025-06-19
OCPBUGS-57818: Reuse hosted cluster release and metadata providers

ryan-cradick:
Manual cherry-pick of https://github.com/openshift/hypershift/pull/6250

**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, use `fixes #<issue_number>(, fixes #<issue_number>, ...)` format, where issue_number might be a GitHub issue, or a Jira story*:
Fixes #

**Checklist**
- [ ] Subject and description added to both, commit and PR.
- [ ] Relevant issues have been referenced.
- [ ] This change includes docs. 
- [ ] This change includes unit tests.

PR #6454, opened 2025-07-16
Reuse hosted cluster release and metadata providers

wangke19:
Manual cherry-pick of #6301


wangke19:
/close
Dev don't think we need to back-port to 4.17.

PR #6470, opened 2025-07-19
CNTRLPLANE-1155: add ExtraArgs field to ClusterAutoscaling for custom autoscaler arguments

LiangquanLi930:
- Add ExtraArgs string field to support custom command line arguments for cluster-autoscaler.
- Users can now specify additional flags like "--cordon-node-before-terminating=true"
in a space-separated format, similar to installFlagsToAdd pattern.

[CNTRLPLANE-1155](https://issues.redhat.com/browse/CNTRLPLANE-1155)

**Checklist**


api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1443,6 +1443,15 @@ type ClusterAutoscaling struct {
...
 	Expanders []ExpanderString `json:"expanders,omitempty"`
+
+	// extraArgs allows users to specify additional command line arguments for the cluster autoscaler.
+	// These arguments will be appended to the autoscaler's command line.
+	// This field is useful for passing custom flags that are not directly supported by the API.
+	// Arguments should be space-separated (e.g., "--custom-flag=value --another-flag --third-flag=123").
+	//
+	// +kubebuilder:validation:MaxLength=1024
+	// +optional
+	ExtraArgs string `json:"extraArgs,omitempty"`
```
LiangquanLi930, line 1454: @JoelSpeed PTAL, Thanks!

JoelSpeed, line 1454: Why not a list of args, which is how they would be in the pod spec?

JoelSpeed, line 1454: Are we sure we want to support this? Doesn't this basically expose a way for customers to do unsupported and potentially odd configurations?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1443,6 +1443,15 @@ type ClusterAutoscaling struct {
 	//
 	// +optional
 	Expanders []ExpanderString `json:"expanders,omitempty"`
+
+	// extraArgs allows users to specify additional command line arguments for the cluster autoscaler.
+	// These arguments will be appended to the autoscaler's command line.
+	// This field is useful for passing custom flags that are not directly supported by the API.
+	// Arguments should be space-separated (e.g., "--custom-flag=value --another-flag --third-flag=123").
```
JoelSpeed, line 1450: Are there any args you wouldn't allow them to set?

LiangquanLi930:
@JoelSpeed Hi, Based on the team discussion, we’ll switch from using the API to annotations. Thanks!

JoelSpeed:
@LiangquanLi930 Why are you switching to annotations? PLEASE do not do that. Annotations are not an API, they cannot be validated easily, they cannot be evolved, they cannot be deprecated, they cannot be removed. 

If you need to expose this use case, then create an API, not a set of annotations, but I do still question _why_ we need to expose this use case

LiangquanLi930:
> @LiangquanLi930 Why are you switching to annotations? PLEASE do not do that. Annotations are not an API, they cannot be validated easily, they cannot be evolved, they cannot be deprecated, they cannot be removed.

> If you need to expose this use case, then create an API, not a set of annotations, but I do still question why we need to expose this use case

Hi @JoelSpeed, thank you for the quick feedback!

We totally understand your concerns about using annotations — especially regarding validation, evolution, and long-term maintainability.

The reason we considered annotations (at least for now) is that this use case is still under evaluation. Specifically, for the Adobe scenario, it’s not yet clear whether the additional flags will bring meaningful benefits, or which exact settings are truly necessary. Since introducing a new API is a long-term commitment, our team felt it would be more prudent to start with an annotation — giving us a chance to experiment and collect feedback.

JoelSpeed:
> Since introducing a new API is a long-term commitment, our team felt it would be more prudent to start with an annotation — giving us a chance to experiment and collect feedback.

Using annotations does not give you this ability. They are an API in all but name. Once you add them, and someone starts relying on them, you have to support them as if they are a tier 1 API.

If you have any links, can you please loop me in on conversations about the request from the customer, and what exactly it is that they are trying to achieve?

LiangquanLi930:
> > Since introducing a new API is a long-term commitment, our team felt it would be more prudent to start with an annotation — giving us a chance to experiment and collect feedback.
> 
> Using annotations does not give you this ability. They are an API in all but name. Once you add them, and someone starts relying on them, you have to support them as if they are a tier 1 API.
> 
> If you have any links, can you please loop me in on conversations about the request from the customer, and what exactly it is that they are trying to achieve?

We need to match the flags pointed out by customer. Here's part of the discussion: https://redhat-internal.slack.com/archives/C08UAV86AK1/p1750879807211979?thread_ts=1750866778.836959&cid=C08UAV86AK1

elmiko:
fwiw, we have specifically chosen *not* to have a field like this in standalone openshift as it opens the support matrix of possible options much wider than we have tested.

i think we need to be very careful about allowing users to input any flag options as we do not have automated tests for all of these options or the combinations they can be used with. this will create addition support burden.

PR #6475, opened 2025-07-21
NO-JIRA: test: replace context.Background/TODO with t.Context in unit tests

bryan-cox:
Replace usage of context.Background() and context.TODO() with t.Context() in test files throughout the codebase. The t.Context() method introduced in Go 1.24 provides a test-scoped context that is automatically cancelled when the test completes, improving test isolation and resource cleanup.

This change affects numerous test files across:
- hypershift-operator controllers
- cmd/cluster platform implementations
- support package tests
- control-plane-operator tests
- integration test utilities

No functional changes to test behavior, only improved context management and automatic cleanup when tests finish or are interrupted.

N/A

**Checklist**


test/e2e/upgrade_hypershift_operator_test.go
```diff
@@ -232,6 +232,6 @@ func TestUpgradeHyperShiftOperator(t *testing.T) {
 				return true
 			}, "5m", "1s").Should(gomega.BeTrue(), "Verification should consistently succeed for 5 minutes")
 		})).To(gomega.BeTrue(), "Verify upgrade invariants should succeed")
-		e2eutil.ValidateHostedClusterConditions(t, context.Background(), mgmtClient, hostedCluster, true, 5*time.Minute)
+		e2eutil.ValidateHostedClusterConditions(t, t.Context(), mgmtClient, hostedCluster, true, 5*time.Minute)
```
csrwng: looks like this should be the `ctx` passed to the function

PR #6338, opened 2025-06-27
OCPBUGS-57790: add custom kube-apiserver DNS name to shared ingress

sjenning:
**What this PR does / why we need it**:

Adds missing handling of HostedCluster.Spec.KubeAPIServerDNSName to the shared ingress controller so that traffic meant for a KAS that is exposed with a custom domain name is correctly routed.

This is necessary so that the KubeAPIExternalName feature works with platforms that use shared ingress.

**Which issue(s) this PR fixes**:

Fixes #OCPBUGS-57790

**Checklist**
- [x] Subject and description added to both, commit and PR.
- [x] Relevant issues have been referenced.
- [ ] This change includes docs. 
- [x] This change includes unit tests.

PR #6366, opened 2025-07-02
CNTRLPLANE-1034: sharedingress: Add support for kube-apiserver whitelist CIDRs

muraee:
**What this PR does / why we need it**:

This patch provides an implementation for CIDR blocks whitelisting.

We propogate CIDR blocks provided by client via HostedCluster's APIServerNetworking and utilize those in the HAProxy configuration generated in order to whitelist specific source IPs.

In case no IPs are provided no ACLs are created in the HAProxy router configuration.

supersedes https://github.com/openshift/hypershift/pull/6354
**Which issue(s) this PR fixes** *(optional, use `fixes #<issue_number>(, fixes #<issue_number>, ...)` format, where issue_number might be a GitHub issue, or a Jira story*:
Fixes #

**Checklist**
- [ ] Subject and description added to both, commit and PR.
- [ ] Relevant issues have been referenced.
- [ ] This change includes docs. 
- [ ] This change includes unit tests.

PR #6477, opened 2025-07-21
ho: exclude kube-apiserver-custom fake route from admission

sjenning:
https://github.com/openshift/hypershift/pull/6338 introduced a fake Route (i.e. not existing in the KAS) add KubeAPIServerDNSName support to the shared ingress router.  However, that list of Routes is later used to admit the Routes, which fails on the fake Route as it does not exist.

This commit removes the HCP label from the fake Route, which causes it to be skipped during the admission code.


sjenning:
superseded by https://github.com/openshift/hypershift/pull/6366

PR #6481, opened 2025-07-22
[release-4.19] OCPBUGS-59674: Initial Changes to Support Self-managed Azure

bryan-cox:
Manual backport of https://github.com/openshift/hypershift/pull/6283


bryan-cox:
We need https://github.com/openshift/hypershift/pull/6485 to pass the HO upgrade test.

PR #6491, opened 2025-07-24
OCPBUGS-59759: fix(konflux): Add missing backslash for dot escaping

celebdor:

In the renovate introducing commit, instead of two backslashes there was a single one, making the CEL checks fail.

Fixes #

**Checklist**


bryan-cox:
/retitle OCPBUGS-59759: fix(konflux): Add missing backslash for dot escaping

PR #6504, opened 2025-07-25
OCPBUGS-57336: Delete unused AdditionalTrustBundle from CP namespace

csrwng:
Resolves the bug that if .spec.additionalTrustBundle is changed to nil on a HostedCluster, the original configmap remains in the control plane namespace, causing ignition payload generation to fail.

Extracts existing code that places .spec.additionalTrustBundle into its own function so it can be unit tested and handles the bundle changing to nil by removing the corresponding ConfigMap from the control plane namespace. Adds unit test.

Fixes #[OCPBUGS-57336](https://issues.redhat.com/browse/OCPBUGS-57336)

**Checklist**

### DIFF
--- a/hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
+++ b/hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
@@ -4341,7 +4341,7 @@ func TestReconcileComponents(t *testing.T) {
 	}
 }
 
-func TestEnsureHostedResroucesAreEmpty(t *testing.T) {
+func TestEnsureHostedResourcesAreEmpty(t *testing.T) {
 	// Setup test cases
 	testCases := []struct {
 		name          string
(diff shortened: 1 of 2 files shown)


hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
```diff
@@ -4339,3 +4339,175 @@ func TestReconcileComponents(t *testing.T) {
...
+				},
+			},
+			expectError:            true,
+			expectedErrorSubstring: "must have a ca-bundle.crt key",
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			ctx := context.TODO()
```
bryan-cox: Should use new `t.Context()`

bryan-cox (commented):
The bug ticket needs the target backport versions filled out.

The ticket also needs the release note fields filled out.

bryan-cox:
/retest

For Konflux

PR #6486, opened 2025-07-23
Fix OCPBUGS-54720: HCP payload doesn't respect multiple

LiangquanLi930:
**What this PR does / why we need it**:
+ Enhance registry override to properly handle multiple mirrors
+ Add image availability verification and fallback mechanism

**Which issue(s) this PR fixes** 
Fixes # OCPBUGS-54720

**Checklist**
- [x] Subject and description added to both, commit and PR.
- [x] Relevant issues have been referenced.
- [ ] This change includes docs. 
- [x] This change includes unit tests.

PR #6222, opened 2025-06-03
OCPBUGS-54720: HCP payload doesn't respect multiple mirrors

LiangquanLi930:
**What this PR does / why we need it**:
- Enhance registry override to properly handle multiple mirrors
- Add image availability verification and fallback mechanism
- Fix namespace handling in registry override
- Update test cases for multiple mirror scenarios

**Which issue(s) this PR fixes** 
OCPBUGS-54720

**Checklist**
- [ ] Subject and description added to both, commit and PR.
- [ ] Relevant issues have been referenced.
- [ ] This change includes docs. 
- [ ] This change includes unit tests.

PR #6517, opened 2025-07-28
OCPBUGS-57141: HCP payload doesn't respect multiple mirrors

LiangquanLi930:
manual cherry-pick https://github.com/openshift/hypershift/pull/6222, https://github.com/openshift/hypershift/pull/6486

Fixes # https://issues.redhat.com/browse/OCPBUGS-57141


control-plane-operator/controllers/hostedcontrolplane/ignitionserver/ignitionserver.go
```diff
@@ -917,15 +927,23 @@ cp /tmp/manifests/99_feature-gate.yaml %[1]s/99_feature-gate.yaml
 	return fmt.Sprintf(script, workDir, featureGateYAML)
 }
 
-func lookupMappedImage(ocpOverrides map[string][]string, image string) (string, error) {
+func lookupMappedImage(ctx context.Context, ocpOverrides map[string][]string, image string, pullSecretBytes []byte) (string, error) {
```
csrwng: If the logic here is the same as the v2 one, can you move both to a utility in the support/util package and call it from both the v2 and v1 version of the code?

csrwng: To make unit testing more robust, we should pass the function that checks image availability as a parameter
`getMetadataFn func(context.Context, string, []byte) (*dockerv1client.DockerImageConfig, []distribution.Descriptor, distribution.BlobStore, error)` and pass a function we control during the unit test (we should not be reaching out to actual registries from unit tests)

control-plane-operator/controllers/hostedcontrolplane/ignitionserver/ignitionserver_test.go
```diff
@@ -147,46 +150,51 @@ func TestReconcileIgnitionServerServiceRoute(t *testing.T) {
...
+			name:        "no overrides provided",
+			overrides:   map[string][]string{},
+			image:       "quay.io/openshift-release-dev/ocp-release:4.15.0-rc.0-multi",
+			expectedImg: "quay.io/openshift-release-dev/ocp-release:4.15.0-rc.0-multi",
 		},
 		{
-			name:   "return the same image if not in the overrides",
-			in:     "registry.ci.openshift.org/ocp/4.14-2023-09-05-120503@sha256:34acb17e45b80267919ea6a2f0e2b21e3f87db3c1e2443d2b3ac9afc0455ae07",
-			expect: "registry.ci.openshift.org/ocp/4.14-2023-09-05-120503@sha256:34acb17e45b80267919ea6a2f0e2b21e3f87db3c1e2443d2b3ac9afc0455ae07",
+			name: "exact repository match found, and multiple mirrors",
```
csrwng: it'd be good to see at least one more test like this, where a different entry matches

csrwng (commented):
Some comments

control-plane-operator/controllers/hostedcontrolplane/ignitionserver/ignitionserver_test.go
LiangquanLi930: https://github.com/openshift/hypershift/blob/d76b8af19e72ee662f30d7f7cf986259e3bf4bbb/control-plane-operator/controllers/hostedcontrolplane/ignitionserver/ignitionserver_test.go#L165-L175
IIUC, I’ve already added it.

support/util/imagemapping.go
```diff
@@ -0,0 +1,52 @@
...
+
+	ctrl "sigs.k8s.io/controller-runtime"
+
+	"github.com/docker/distribution"
+)
+
+// LookupMappedImage looks up a mapped image based on OCP overrides
+// It takes a getMetadataFn parameter to make unit testing more robust
+// by allowing tests to control the image availability check without reaching out to actual registries
+func LookupMappedImage(ctx context.Context, ocpOverrides map[string][]string, image string, pullSecretBytes []byte, getMetadataFn func(context.Context, string, []byte) (*dockerv1client.DockerImageConfig, []distribution.Descriptor, distribution.BlobStore, error)) (string, error) {
```
csrwng: nit: it may make the code clearer to declare a type for the getMetadataFn

csrwng (commented):
Looks great, just one minor nit

PR #6520, opened 2025-07-28
Add support for generating a SC UID per namespace if SCC are not avalable

enxebre:
Currently when there's no SCC in the management cluster we fallback to set a default SC with a common UID for runAsUser. This is a suboptimal from security pov since it increases blast radius in case of a vulneravility. This is also problematic if a component happen to abuse inotify and hits /proc/sys/fs/inotify/max_user_instances

This PR adds a mechanism to annotate each controlplane namespace with an annotation to assign a specific uid.

Other info:

When SSCs are available this feature is implemented by https://github.com/openshift/cluster-policy-controller/blob/3e7538547c8f209c72083097a4ebaada6e9c46c5/pkg/security/controller/namespace_scc_allocation_controller.go#L148

etcd is not included in this PR to tackle this https://github.com/openshift/hypershift/blob/ae1f96b47545aad7794478adfe58f1d7017d6a83/support/controlplane-component/defaults.go#L132-L133 separately

cno and snapshotcontrollerOperator operands are also excluded to tackle separately https://github.com/openshift/hypershift/blob/ae1f96b47545aad7794478adfe58f1d7017d6a83/control-plane-operator/controllers/hostedcontrolplane/v2/cno/deployment.go#L141 https://github.com/openshift/hypershift/blob/ae1f96b47545aad7794478adfe58f1d7017d6a83/control-plane-operator/controllers/hostedcontrolplane/v2/snapshotcontroller/deployment.go#L26

Hypershift consumers can effectively opt-out to this for existing HCs by setting the annotation "hypershift.openshift.io/default-security-context-scc-uid" to 1001

Fixes #

**Checklist**


sjenning:
It seems that the UID range starts at 1001 and has no upper bound and no way to configure the UID range the HO allocates.  Is this correct?

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -201,6 +205,20 @@ func (r *HostedControlPlaneReconciler) SetupWithManager(mgr ctrl.Manager, create
 	r.ec2Client, r.awsSession = GetEC2Client()
 
 	r.registerComponents(hcp)
+
+	uidInput := os.Getenv(controlplaneoperator.DefaultSecurityContextUIDEnvVar)
+	if uidInput == "" {
+		r.Log.Info("DEFAULT_SECURITY_CONTEXT_UID is not set. This should never happen, unless you are running a HO which doesn't support this CPO")
+		r.DefaultSecurityContextUID = 0
```
csrwng: Should this fallback to 1001 which is what we have today?

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -1302,6 +1303,24 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
 			controlPlaneNamespace.Labels["openshift.io/cluster-monitoring"] = "true"
 		}
 
+		// Set a label so we can identify the control plane namespaces.
+		controlPlaneNamespace.Labels[ControlPlaneNamespaceLabelKey] = "true"
```
csrwng: We already set a label that identifies the control plane namespace:
https://github.com/openshift/hypershift/blob/f70206108efa2e208c4733f2f819ce639e1a8a75/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go#L1281

csrwng (commented):
A couple of comments. The general approach looks fine to me.

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
enxebre: we are falling back to that already via [support/controlplane-component/defaults.go](https://github.com/openshift/hypershift/pull/6520/files#diff-a194f3bdb82070261e10c3429532c0fbc35ec6b199238ccaece0d509fb788863)

This can only be reached if running this cpo with an older HO. I only added this code path because that's the case for the ho upgrade test.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -2031,6 +2056,59 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
...
+	ControlPlaneNamespaceLabelKey = "hypershift.openshift.io/hosted-control-plane"
+)
+
+var sccUIDCheckMutex sync.Mutex
+var maxSCCUIDCache int64
+
+// getNextAvailableSCCUID returns the next available SCC UID for a control plane namespace.
+// If no cache is set, it loops over existing namespaces, find the highest SCC UID and returns that + 1.
+// When SCCs are available, this feature is implemented by https://github.com/openshift/cluster-policy-controller/blob/3e7538547c8f209c72083097a4ebaada6e9c46c5/pkg/security/controller/namespace_scc_allocation_controller.go#L148
+func getNextAvailableSCCUID(ctx context.Context, c client.Client) (int64, error) {
```
rtheis: I'm wondering if we could do something similar when running on Kubernetes. Maybe looking at the hosted clusters may be a sufficient method to select a default UID.

enxebre: to clarify, this PR changes only impact HCs when running on management clusters with no SCC like in your case. I change this func name in case it was confusing.

Before this PR you always get 1001, after his PR you would get an available autodiscovered uid for each control plane namespace.

This is signaled by an annotation on each namespace. You could opt-out for this automation in your existing HCs by manually setting the annotation for each namespace to your current value i.e 1001, before this change rollsout. Otherwise the automation would automatically add the annotation for you assigning new values for each namespace and causing a fleet wise pods rollout

rtheis: Thank you @enxebre. This make sense and sounds like a nice improvement which we will plan to leverage in ROKS 4.20.

rtheis (approved):
/lgtm

PR #6523, opened 2025-07-28
CNTRLPLANE-1190: Support control plane overrides by platform

csrwng:
Modifies the CPO override file format to allow specifying a platform for the overrides. The file should contain the lower case platform name and the list of version overrides for that platform.

This is needed to support CPO overrides that are specific to ARO deployments of the HyperShift operator and are different from the ROSA deployments.

Fixes #[CNTRLPLANE-1190](https://issues.redhat.com/browse/CNTRLPLANE-1190)

**Checklist**

### DIFF
--- a/hypershift-operator/controlplaneoperator-overrides/assets/overrides.yaml
+++ b/hypershift-operator/controlplaneoperator-overrides/assets/overrides.yaml
@@ -1,218 +1,220 @@
-overrides:
-  # Beginning of OCPBUGS-48519 overrides 4.15 section
-  # Using ocp-release:4.15.45-multi
-  - version: 4.15.0
-    cpoImage: quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:10562af6cc5f0f9c61b9831792b9e371af6e660de4e3c56a5e4e8ef849219fc9
-  - version: 4.15.1
-    cpoImage: quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:10562af6cc5f0f9c61b9831792b9e371af6e660de4e3c56a5e4e8ef849219fc9
-  - version: 4.15.2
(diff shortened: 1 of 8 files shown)


hypershift-operator/controlplaneoperator-overrides/assets/overrides.yaml
```diff
@@ -1,4 +1,6 @@
-overrides:
+patforms:
```
bryan-cox:
```suggestion
platforms:
```

bryan-cox:
Do we have upstream docs on how to use this that we could provide CS, SRE, etc.

hypershift-operator/controlplaneoperator-overrides/assets/overrides.yaml
enxebre: maybe let the units consume this file to prevent such typos from happening

muraee: isn't it better to make the different platforms separate fields. so we one can't have multiple entries in the list with the same platform.
```suggestion
paltforms:
 aws:
   overrides:
    - version: x.y.z
    .....
```

bryan-cox: platforms is still spelled wrong in the suggestion 😅

csrwng:
> Do we have upstream docs on how to use this that we could provide CS, SRE, etc.

This override mechanism is not consumable by CS,SRE,etc. It is fully controlled by the HO code we deliver. However, it would not be a bad idea to document this for ourselves. I'll add a doc.

hypershift-operator/controlplaneoperator-overrides/assets/overrides.yaml
csrwng: 👍 makes sense

PR #6526, opened 2025-07-28
CNTRLPLANE-1191: add sa-token-issuer-private-key-path flag to create cluster aws

sjenning:
In order to share an OIDC provider across multiple job mgmt clusters, we need set `spec.serviceAccountSigningKey` on the `HostedCluster` at create time.  The `cluster create azure` code already has this flag.

This PR adds the flag for `cluster create aws` as well.


sjenning:
xref follow on PR in o/release https://github.com/openshift/release/pull/67630

PR #6528, opened 2025-07-29
OCPBUGS-59914: fix(snapshotcontroller): Add missing VolumeSnapshot API requirement

bryan-cox:
Add VolumeSnapshot API from snapshot.storage.k8s.io/v1 to the list of required APIs for the snapshot controller component availability prober.

OCPBUGS-59914

**Checklist**


bryan-cox:
/close

This seems to be an upstream issue. https://github.com/kubernetes-csi/external-snapshotter/blob/6b2feaaf7fc3d6d0fe21029d1032ea0eee2081e9/cmd/snapshot-controller/main.go#L93 needs to retry longer so that csi-snapshot-controller-operator has time to make the CRD.

PR #2571, opened 2023-05-17
HOSTEDCP-996: CLI: enable guest cluster dump for private clusters

csrwng:
**What this PR does / why we need it**:
Uses port-forwarding to access a kube-apiserver pod in the control plane directly. This avoids the need to have access to the external API endpoint of the guest cluster.

**Which issue(s) this PR fixes** *(optional, use `fixes #<issue_number>(, fixes #<issue_number>, ...)` format, where issue_number might be a GitHub issue, or a Jira story*:
Fixes #[HOSTEDCP-996](https://issues.redhat.com//browse/HOSTEDCP-996)

**Checklist**
- [x] Subject and description added to both, commit and PR.
- [x] Relevant issues have been referenced.

PR #6533, opened 2025-07-29
SREP-715: New option to dump data plane content through the kube API server service exposed in HCP namespaces

Nikokolas3270:

Port forwarding is not possible as debug handlers are disabled on MC clusters. As a result data plane content (including worker nodes logs) is currently failing to be dumped.

The new `--dump-guest-cluster-through-kube-service` can be used to by-pass that limitation by targeting the `kube-apiserver` service exposed by HCP namespaces.
Remark that it is only suitable to use that option when:
- Within a MC cluster (i.e. in a pod which has access to the service)
- The MC cluster has the debug handlers disabled

Contributes to [SREP-715](https://issues.redhat.com//browse/SREP-715)

Next steps are:
- Change following call to use that new option (once the `hypershift` CLI is released to be deployed on the stolostron image): https://github.com/stolostron/must-gather/blob/b7d9c20b996811aa02f5d432bbdda43fb7e37b33/collection-scripts/gather_utils#L135
- Use the stolostron image there: https://github.com/openshift/osdctl/blob/9abf1be8ec5c5029fc3ea97c4e083c30748ee278/cmd/hcp/mustgather/mustGather.go#L198

**Checklist**

### DIFF
--- a/cmd/cluster/core/dump.go
+++ b/cmd/cluster/core/dump.go
@@ -4,7 +4,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
-	"math/rand"
+	"math/rand/v2"
 	"os"
 	"os/exec"
 	"path"
@@ -167,7 +167,6 @@ func NewDumpCommand() *cobra.Command {
 	cmd.MarkFlagsMutuallyExclusive("dump-guest-cluster", "dump-guest-cluster-through-kube-service")
 
 	cmd.RunE = func(cmd *cobra.Command, args []string) error {
-		rand.New(rand.NewSource(time.Now().UnixNano()))
 		if err := DumpCluster(cmd.Context(), opts); err != nil {
 			opts.Log.Error(err, "Error")
 			return err
@@ -203,7 +202,7 @@ func dumpGuestCluster(ctx context.Context, opts *DumpOptions) error {
 	if opts.DumpGuestClusterThroughKubeService {
 		localPort = -1 // Indicates connection via kube-apiserver service instead of port-forward
 	} else {
-		localPort = rand.Intn(45000-32767) + 32767
+		localPort = rand.IntN(45000-32767) + 32767
 		podToForward, err := supportforwarder.GetRunningKubeAPIServerPod(ctx, c, cpNamespace)
 		if err != nil {
 			return fmt.Errorf("failed to get running kube-apiserver pod for guest cluster: %w", err)


cmd/cluster/core/dump.go
```diff
@@ -262,7 +226,7 @@ func createGuestKubeconfig(ctx context.Context, c client.Client, cpNamespace str
 	}
 
 	for k := range localhostKubeconfig.Clusters {
-		localhostKubeconfig.Clusters[k].Server = fmt.Sprintf("https://localhost:%d", localPort)
+		localhostKubeconfig.Clusters[k].Server = fmt.Sprintf("https://kube-apiserver.%s.svc.cluster.local:6443", cpNamespace)
```
enxebre: Fwiw this was introduced here https://github.com/openshift/hypershift/pull/2571 to support private clusters by @csrwng 

The assumption that the svc network is reachable is not always true. I think we might want to make this an opt-in flag e.g. use-service-network

Nikokolas3270: Hello,

Dumping the guest cluster is currently not working, at least in our current infrastructure (ROSA HCP) where the debug handlers are disabled. So with that scope in mind we don't see the need of an option to keep the current behaviour/code.
Also, even if the current code may work in an infrastructure outside of ROSA HCP where debug handlers are not disabled: is it worth keeping it? Having a switch between the current and the new code is introducing technical debt.

Concerning the `kube-apiserver` service eventually not accessible: can you please give examples where/when this service is not accessible?
This service is the one currently queried by RMO (Route Monitor Operator) to check for the good health of the Kube API on guest clusters. So it this service is not accessible it is likely to trigger the `api-ErrorBudgetBurn` alert.

Concerning the changes introduced by @csrwng: indeed accessing the kube API via the public URL was not a viable option with private link clusters; the new code proposed here also works with those clusters.

cmd/cluster/core/dump.go
```diff
@@ -183,7 +187,55 @@ func dumpGuestCluster(ctx context.Context, opts *DumpOptions) error {
...
 	cpNamespace := manifests.HostedControlPlaneNamespace(opts.Namespace, opts.Name)
-	localPort := rand.Intn(45000-32767) + 32767
+
+	var localPort int
+	var forwarderStop chan struct{}
+
+	if opts.DumpGuestClusterThroughKubeService {
+		localPort = -1
+	} else {
+		localPort = rand.Intn(45000-32767) + 32767
```
jparrill: I think this should be `rand.IntN` instead of `rand.Intn` (last one is good for go 1.21 and earlier), also the API should be `math/rand/v2`

Nikokolas3270:
> Hey @Nikokolas3270 In our experience the portForwarding usually is a problematic option regarding CI, many flakes were solved in the past because of this. Some questions/topics:
> 
> * Did you test this somehow Self-managed/ROSA?
> * Could we add coverage with unit tests? (I didn't check if that's possible or those are more at integration test level)

Yes, I did test this using some home made docker image on real data.

There is some `dump_test.go` file... but it is only about testing some side function; not the actual guest cluster dump. This could eventually be put in place but will cost quite some time... especially because there are now 2 modes to test: the default / existing mode using port-forwarding & the new mode using the kube-API service.

cmd/cluster/core/dump.go
Nikokolas3270: Remark that `rand.Intn(45000-32767) + 32767` is already present in the old code.
I will give a look at the new API

cmd/cluster/core/dump.go
```diff
@@ -189,7 +192,55 @@ func dumpGuestCluster(ctx context.Context, opts *DumpOptions) error {
...
+		if err := forwarder.ForwardPorts([]string{fmt.Sprintf("%d:%d", localPort, podPort)}, forwarderStop); err != nil {
+			return fmt.Errorf("cannot forward kube apiserver port: %w, output: %s", err, forwarderOutput.String())
+		}
+	}
+	defer func() {
+		if forwarderStop != nil {
+			close(forwarderStop)
+		}
+	}()
+
```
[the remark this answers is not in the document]
jparrill: Could you take a look @Nikokolas3270 ?

jparrill (commented):
/approve

I will let the `lgtm` to other peer. 

Non-blocking nit, I would like to see some examples about how to execute it using the cmd (both modes). I'm doing this more and more and really helps the UX of the CLI:

```
Σ hypershift git:(CNTRLPLANE-1691) ./bin/hypershift create backup --help
Create a backup of a hosted cluster using OADP (OpenShift API for Data Protection).

The backup command automatically detects the platform type of your HostedCluster and includes
the appropriate platform-specific resources. It validates OADP installation and creates
comprehensive backups that can be used for disaster recovery scenarios.

Examples:
  # Basic backup with defaults
  hypershift create backup --hc-name production --hc-namespace clusters

  # Backup with custom settings
  hypershift create backup --hc-name production --hc-namespace clusters --storage-location s3-backup --ttl 24h

  # Backup with custom resource selection
  hypershift create backup --hc-name production --hc-namespace clusters --included-resources hostedcluster,nodepool,secrets,configmap
...
...
...
```

cmd/cluster/core/dump.go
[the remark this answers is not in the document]
Nikokolas3270: Hello, this code was just moved in an `else` block. I did not change anything concerning the logic there.
Aim of this PR is precisely to by-pass this port forwarding section is case to the new option is set

cmd/cluster/core/dump.go
```diff
@@ -190,7 +193,55 @@ func dumpGuestCluster(ctx context.Context, opts *DumpOptions) error {
 		return fmt.Errorf("failed to get hosted cluster %s/%s: %w", opts.Namespace, opts.Name, err)
 	}
 	cpNamespace := manifests.HostedControlPlaneNamespace(opts.Namespace, opts.Name)
-	localPort := rand.Intn(45000-32767) + 32767
+
+	var localPort int
+	var forwarderStop chan struct{}
+
+	if opts.DumpGuestClusterThroughKubeService {
```
bryan-cox, line 202: Can we add some validation around the two new flags? When both --dump-guest-cluster and --dump-guest-cluster-through-kube-service are provided, the code enters dumpGuestCluster() but the behavior depends on which flag was set. The DumpGuestClusterThroughKubeService flag takes precedence silently (line 200), which may confuse users.

```
Recommendation:
  Add explicit validation or logging when both flags are used:

  // Option 1: Make flags mutually exclusive
  if opts.DumpGuestCluster && opts.DumpGuestClusterThroughKubeService {
      return fmt.Errorf("--dump-guest-cluster and --dump-guest-cluster-through-kube-service are mutually exclusive")
  }

  // Option 2: Log which mode is being used
  if opts.DumpGuestClusterThroughKubeService {
      opts.Log.Info("Dumping guest cluster through kube-apiserver service")
  } else if opts.DumpGuestCluster {
      opts.Log.Info("Dumping guest cluster through port-forwarding")
  }
```

cmd/cluster/core/dump.go
```diff
@@ -190,7 +193,55 @@ func dumpGuestCluster(ctx context.Context, opts *DumpOptions) error {
...
+		localPort = -1
```
bryan-cox: Can you add a comment here explaining why this is -1? Right now it might not be clear to a user.

Maybe something like `// Indicates connection via kube-apiserver service instead of port-forward`

cmd/cluster/core/dump.go
```diff
@@ -201,43 +252,10 @@ func dumpGuestCluster(ctx context.Context, opts *DumpOptions) error {
...
-		Out:       forwarderOutput,
-		ErrOut:    forwarderOutput,
-	}
-	podPort := supportutil.KASPodPortFromHostedCluster(hostedCluster)
-	forwarderStop := make(chan struct{})
-	if err := forwarder.ForwardPorts([]string{fmt.Sprintf("%d:%d", localPort, podPort)}, forwarderStop); err != nil {
-		return fmt.Errorf("cannot forward kube apiserver port: %w, output: %s", err, forwarderOutput.String())
+	target := filepath.Join(opts.ArtifactDir, "hostedcluster-"+opts.Name)
+	if err := os.MkdirAll(target, 0755); err != nil {
+		return err
```
bryan-cox, line 259: Can you wrap the err message here? Maybe something like `return fmt.Errorf("failed to create guest cluster dump directory %s: %w", target, err)`

cmd/cluster/core/dump.go
Nikokolas3270: Going for the mutual exclusion

cmd/cluster/core/dump.go
Nikokolas3270: Ok for the comment.
`-1` indeed indicates that `localPort` is not set.
This value is checked by the `createGuestKubeconfig` function which is creating the kube config to know whether we are going with the "port forward" mode or the "directly target the service" mode

cmd/cluster/core/dump.go
Nikokolas3270: Sorry but I don't think that wrapping this error is adding much value.
`err` will already tell that the directory cannot be created and it will tell that with more details.
Also remark that there is no error wrapping for the call to `createGuestKubeconfig`; not saying that this is an excuse to do the same here, but error wrapping should only be done:
- To add more details to an existing error
- Or when we want to change the type of an error

wewang58:
/verified by @wewang58  
I added verified steps in [SREP-715](https://issues.redhat.com//browse/SREP-715).

PR #6536, opened 2025-07-29
NO-JIRA: wrap EnsureKubeAPIDNSNameCustomCert and EnsureGlobalPullSecret in their own test

jparrill:

To have the EnsureGlobalPullSecret and EnsureKubeAPIDNSNameCustomCert make the TestCreateCluster to be skipped. This PR makes sure that would not happen anymore.

### DIFF
--- a/test/e2e/util/util.go
+++ b/test/e2e/util/util.go
@@ -1531,185 +1531,186 @@ func EnsureGuestWebhooksValidated(t *testing.T, ctx context.Context, guestClient
 }
 
 func EnsureGlobalPullSecret(t *testing.T, ctx context.Context, mgmtClient crclient.Client, entryHostedCluster *hyperv1.HostedCluster) error {
-	AtLeast(t, Version420)
+	t.Run("EnsureGlobalPullSecret", func(t *testing.T) {
+		AtLeast(t, Version419)
+		if entryHostedCluster.Spec.Platform.Type != hyperv1.AzurePlatform {
+			t.Skip("test only supported on platform ARO")
+		}
 
-	if entryHostedCluster.Spec.Platform.Type != hyperv1.AzurePlatform {
-		t.Skip("test only supported on platform ARO")
-	}
+		var (
+			dummyImageTagMultiarch = "quay.io/hypershift/sleep:multiarch"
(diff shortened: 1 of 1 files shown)


test/e2e/util/util.go
```diff
@@ -1531,185 +1531,186 @@ func EnsureGuestWebhooksValidated(t *testing.T, ctx context.Context, guestClient
...
-		additionalPullSecretDummyData       = []byte(`{"auths": {"quay.io": {"auth": "YWRtaW46cGFzc3dvcmQ="}}}`)
-		additionalPullSecretReadOnlyE2EData = []byte(`{"auths": {"quay.io": {"auth": "aHlwZXJzaGlmdCtlMmVfcmVhZG9ubHk6R1U2V0ZDTzVaVkJHVDJPREE1VVAxT0lCOVlNMFg2TlY0UkZCT1lJSjE3TDBWOFpTVlFGVE5BS0daNTNNQVAzRA=="}}}`)
-		oldglobalPullSecretData             []byte
-		dsImage                             string
-	)
-	g := NewWithT(t)
+	t.Run("EnsureGlobalPullSecret", func(t *testing.T) {
+		AtLeast(t, Version419)
+		if entryHostedCluster.Spec.Platform.Type != hyperv1.AzurePlatform {
+			t.Skip("test only supported on platform ARO")
```
sdminonne, line 1545: @jparrill this will break down the day we'll support not managed on Azure. To fix this I would consider this https://github.com/openshift/hypershift/blob/1b91a8234cc4383f59e14dfd1ae22ae52a0cb8df/support/azureutil/azureutil.go#L186

sdminonne (commented):
One `nit today. To be fixed tomorrow. Otherwwise LGTM

test/e2e/util/util.go
jparrill: Yes and no, Bryan is also working in self-managed azure, this is why I've followed the platform.Type check instead of ARO.

sdminonne: @jparrill , fair enough then

bryan-cox:
This should link to a valid Jira ticket

PR #6550, opened 2025-07-30
OCPBUGS-59886: Add selinux-warning-controller to KCM

jsafrane:
selinux-warning-controller is an controller that emits warnings about Pods with wrong SELinux labels. It's opt-in upstream, because upstream clusters typically don't run with SELinux and it would be just waste of memory and CPU.

OpenShift need the controller running, because it has SELinux enabled, and it provides better UX for our users.

And we run e2e tests that expect the controller running, see [OCPBUGS-59886](https://issues.redhat.com/browse/OCPBUGS-59886).

A similar change has been done to standalone OCP: https://github.com/openshift/cluster-kube-controller-manager-operator/pull/834

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/testdata/kube-controller-manager/TechPreviewNoUpgrade/zz_fixture_TestControlPlaneComponents_kube_controller_manager_deployment.yaml
+++ b/control-plane-operator/controllers/hostedcontrolplane/testdata/kube-controller-manager/TechPreviewNoUpgrade/zz_fixture_TestControlPlaneComponents_kube_controller_manager_deployment.yaml
@@ -92,6 +92,7 @@ spec:
         - --controllers=-ttl
         - --controllers=-bootstrapsigner
         - --controllers=-tokencleaner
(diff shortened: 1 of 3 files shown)


sjenning:
Need to update testdata with`UPDATE=true go test ./...`
/approve

jsafrane:
@sjenning tests updated

jsafrane:
/payload-job periodic-ci-openshift-hypershift-release-4.20-periodics-e2e-aws-ovn-conformance-serial

jsafrane:
All `FeatureGate:SELinuxChangePolicy` tests in the payload job above passed, i.e. the controller runs.

bryan-cox:
/test e2e-aks

There is a flake on `TestUpgradeControlPlane` we are actively trying to fix but it is not permafailing.

bryan-cox:
/override "verify"

We now require conventional commits, but that enforcement started after this PR was approved. https://github.com/openshift/hypershift/pull/6559 merged after this PR was approved.

PR #6553, opened 2025-07-31
CORENET-6229: Use networking.advertiseAddress for CNO apiserver override for IBMCloud PlatformType

bradbehle:
This tells the cluster network operator to use the HostedCluster Spec.Networking.APIServer.AdvertiseAddress and Port when connecting to the apiserver on clusters in IBM Cloud.  This is needed so that OVN (and probably other components) does not try to use the external apiserver URLs, which might either be blocked by firewall rules or using named certificates which OVN doesn't handle.

Fixes # https://issues.redhat.com/browse/CORENET-6229

**Checklist**

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/v2/cno/deployment.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/v2/cno/deployment.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"strconv"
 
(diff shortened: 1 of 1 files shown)


rtheis (approved):
/lgtm

bryan-cox:
e2e-aks is flaky at the moment due to a new test added yesterday. We are working to fix it though.

PR #6554, opened 2025-07-31
CNTRLPLANE-1209: exposing IPv4OVNKubernetesConfig for hcp's ClusterNetworking

georgelipceanu:
This change introduces an ovnKubernetesConfig field to the HC API, allowing users to configure the `internalJoinSubnet` and `internalTransitSwitchSubnet` in the HC
NOTE: These changes are primarily to `hostedcluster_types.go`, all other file changes were created after running `make api` in CLI.

Fixes #[CNTRLPLANE-1209](https://issues.redhat.com/browse/CNTRLPLANE-1209)

**Checklist**


api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1036,6 +1036,12 @@ type ClusterNetworking struct {
 	//
 	// +optional
 	APIServer *APIServerNetworking `json:"apiServer,omitempty"`
+
+	// ipv4 allows users to configure IP settings for IPv4 connections. When ommitted,
+	// this means no opinions and the default configuration is used. Check individual
+	// fields within ipv4 for details of default values.
+	// +optional
+	IPv4OVNKubernetes *IPv4OVNKubernetesConfig `json:"ipv4OVNKubernetes,omitempty"`
```
LiangquanLi930: I think we can add a struct like this:
```
OVNKubernetesConfig *OVNKubernetesConfig `json:"ovnKubernetesConfig,omitempty"`
type OVNKubernetesConfig struct {
	IPv4 *OVNIPv4Config `json:"ipv4,omitempty"`
}
```
This approach will make it easier for us to add other features in the future.

georgelipceanu: Something similar to this?
```
...............................................................
	// ipv4 allows users to configure IP settings for IPv4 connections. When ommitted,
	// this means no opinions and the default configuration is used. Check individual
	// fields within ipv4 for details of default values.
	// +optional
	OVNKubernetesConfig *OVNKubernetesConfig `json:"ovnKubernetesConfig,omitempty"`
}

type OVNKubernetesConfig struct {
	IPv4 *OVNIPv4Config `json:"ipv4,omitempty"`
}

type OVNIPv4Config struct {
	InternalTransitSwitchSubnet string `json:"internalTransitSwitchSubnet,omitempty"`
	InternalJoinSubnet          string `json:"internalJoinSubnet,omitempty"`
}
```

LiangquanLi930: yes

georgelipceanu: Pushed new changes there 🙏

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1036,6 +1036,53 @@ type ClusterNetworking struct {
 	//
 	// +optional
 	APIServer *APIServerNetworking `json:"apiServer,omitempty"`
+
+	// ovnKubernetesConfig holds OVN-Kubernetes specific configuration.
+	// This is only consumed when NetworkType is OVNKubernetes.
+	// +optional
+	OVNKubernetesConfig *OVNKubernetesConfig `json:"ovnKubernetesConfig,omitempty"`
```
LiangquanLi930: @JoelSpeed PTAL, Thanks!

JoelSpeed: Is the go.mod setting a minimum Go 1.24 yet?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1036,6 +1036,53 @@ type ClusterNetworking struct {
...
+}
+
+// OVNKubernetesConfig contains OVN-Kubernetes specific configuration options.
+type OVNKubernetesConfig struct {
```
JoelSpeed: Should add a MinProperties 1 marker to this to prevent `{}` being valid (or add a required field)

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1036,6 +1036,53 @@ type ClusterNetworking struct {
...
+	// ipv4 allows users to configure IP settings for IPv4 connections. When omitted,
+	// this means no opinions and the default configuration is used. Check individual
+	// fields within ipv4 for details of default values.
+	// +optional
+	IPv4 *OVNIPv4Config `json:"ipv4,omitempty"`
+}
+
+// OVNIPv4Config contains IPv4-specific configuration options for OVN-Kubernetes.
+type OVNIPv4Config struct {
```
JoelSpeed: Should add MinProperties or a required field

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1036,6 +1036,53 @@ type ClusterNetworking struct {
...
+	// by OVN-Kubernetes for the distributed transit switch in the OVN Interconnect
+	// architecture that connects the cluster routers on each node together to enable
+	// east west traffic. The subnet chosen should not overlap with other networks
+	// specified for OVN-Kubernetes as well as other networks used on the host.
+	// When omitted, this means no opinion and the platform is left to choose a reasonable
+	// default which is subject to change over time.
+	// The current default subnet is 100.88.0.0/16
+	// The subnet must be large enough to accommodate one IP per node in your cluster
+	// The value must be in proper IPV4 CIDR format
+	// +kubebuilder:validation:MaxLength=18
```
JoelSpeed: Please add a proper minimum length as well, I think 9?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1036,6 +1036,53 @@ type ClusterNetworking struct {
...
+	// +optional
+	InternalTransitSwitchSubnet string `json:"internalTransitSwitchSubnet,omitempty"`
+	// internalJoinSubnet is a v4 subnet used internally by ovn-kubernetes in case the
+	// default one is being already used by something else. It must not overlap with
+	// any other subnet being used by OpenShift or by the node network. The size of the
+	// subnet must be larger than the number of nodes.
+	// The current default value is 100.64.0.0/16
+	// The subnet must be large enough to accommodate one IP per node in your cluster
+	// The value must be in proper IPV4 CIDR format
+	// +kubebuilder:validation:MaxLength=18
```
JoelSpeed: Add min length as well please

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1036,6 +1036,53 @@ type ClusterNetworking struct {
...
 	// +optional
 	APIServer *APIServerNetworking `json:"apiServer,omitempty"`
+
+	// ovnKubernetesConfig holds OVN-Kubernetes specific configuration.
+	// This is only consumed when NetworkType is OVNKubernetes.
+	// +optional
+	OVNKubernetesConfig *OVNKubernetesConfig `json:"ovnKubernetesConfig,omitempty"`
+}
+
+// OVNKubernetesConfig contains OVN-Kubernetes specific configuration options.
```
enxebre: can you please include a reference to the canonical source of this API?

enxebre:
/hold

Until we clarify the rationale for this change

api/hypershift/v1beta1/hostedcluster_types.go
enxebre: this is mirroring an operatorv1 api within the block we have for configv1, I assume you were looking for OperatorConfiguration *OperatorConfiguration

georgelipceanu: The minimum is currently `go 1.23.6`
ref: https://github.com/openshift/hypershift/blob/v0.1.65/go.mod#L3

api/hypershift/v1beta1/hostedcluster_types.go
georgelipceanu: Adding these now 👍

api/hypershift/v1beta1/hostedcluster_types.go
JoelSpeed: Then this is the correct way to do this, I won't suggest different :)

api/hypershift/v1beta1/hostedcluster_types.go
georgelipceanu: Adding this now

api/hypershift/v1beta1/hostedcluster_types.go
LiangquanLi930: @tssurya Could you help to confirm?

api/hypershift/v1beta1/hostedcluster_types.go
LiangquanLi930: @tssurya Same

api/hypershift/v1beta1/hostedcluster_types.go
LiangquanLi930: okay, we need to put in `ClusterNetworkOperatorSpec`

georgelipceanu: Will `OVNKubernetesConfig` need to be removed from `ClusterNetworking`? Or will it be in both structs?

api/hypershift/v1beta1/hostedcluster_types.go
georgelipceanu: Added these references

api/hypershift/v1beta1/hostedcluster_types.go
georgelipceanu: Added `OVNKubernetesConfig` to both 👍

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1036,6 +1036,60 @@ type ClusterNetworking struct {
 	//
 	// +optional
 	APIServer *APIServerNetworking `json:"apiServer,omitempty"`
+
+	// ovnKubernetesConfig holds OVN-Kubernetes specific configuration.
+	// This is only consumed when NetworkType is OVNKubernetes.
+	// +optional
+	OVNKubernetesConfig *OVNKubernetesConfig `json:"ovnKubernetesConfig,omitempty"`
```
JoelSpeed: You should add CEL to enforce `This is only consumed when NetworkType is OVNKubernetes.`

```
+kubebuilder:validation:XValidation::rule="has(self.networkType) && self.networkType == 'OVNKubernetes' ? true : !has(self.ovnKubernetesConfig)",message="ovnKubernetesConfig is forbidden when type is not OVNKubernetes"
```

This looks like an optional union member to me

api/hypershift/v1beta1/hostedcluster_types.go
JoelSpeed: 9 is based on the shortest octet being 1 character and the shortest subnet mask being 1 character, plus the 3 dots and slash, so you have `A.B.C.D/E`, which is 9

api/hypershift/v1beta1/hostedcluster_types.go
georgelipceanu: Added this just now 🙏

api/hypershift/v1beta1/hostedcluster_types.go
LiangquanLi930: > Will OVNKubernetesConfig need to be removed from ClusterNetworking? Or will it be in both structs?

move to ClusterNetworkOperatorSpec

georgelipceanu: Will do that now

api/hypershift/v1beta1/hostedcluster_types.go
JoelSpeed: Ahh sorry, this will need to be on the parent struct level, not on the field itself, my bad, didn't mention that

georgelipceanu: No worries, adding this to the parent struct `HostedClusterSpec` now 👍

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -447,6 +447,7 @@ type Capabilities struct {
...
 // +kubebuilder:validation:XValidation:rule="!has(self.operatorConfiguration) || !has(self.operatorConfiguration.clusterNetworkOperator) || !has(self.operatorConfiguration.clusterNetworkOperator.disableMultiNetwork) || !self.operatorConfiguration.clusterNetworkOperator.disableMultiNetwork || self.networking.networkType == 'Other'",message="disableMultiNetwork can only be set to true when networkType is 'Other'"
+// +kubebuilder:validation:XValidation:rule="self.networking.networkType == 'OVNKubernetes' || !has(self.operatorConfiguration) || !has(self.operatorConfiguration.clusterNetworkOperator) || !has(self.operatorConfiguration.clusterNetworkOperator.ovnKubernetesConfig)", message="ovnKubernetesConfig is forbidden when networkType is not OVNKubernetes"
```
JoelSpeed: Is there a reason these fields are so far from each other? They appear related no?

JoelSpeed: You can use the CEL optional syntax to make this much simpler
```suggestion
// +kubebuilder:validation:XValidation:rule="self.networking.networkType == 'OVNKubernetes' || !self.?operatorConfiguration.clusterNetworkOperator.ovnKubernetesConfig.hasValue()", message="ovnKubernetesConfig is forbidden when networkType is not OVNKubernetes"
```

api/hypershift/v1beta1/operator.go
```diff
@@ -48,4 +48,58 @@ type ClusterNetworkOperatorSpec struct {
...
+}
+
+// OVNKubernetesConfig contains OVN-Kubernetes specific configuration options.
+// https://github.com/openshift/api/blob/6d3c4e25a8d3aeb57ad61649d80c38cbd27d1cc8/operator/v1/types_network.go#L400-L471
+// +kubebuilder:validation:MinProperties=1
+type OVNKubernetesConfig struct {
+	// ipv4 allows users to configure IP settings for IPv4 connections. When omitted,
+	// this means no opinions and the default configuration is used. Check individual
+	// fields within ipv4 for details of default values.
+
```
JoelSpeed: Random blank line?

georgelipceanu: Removing now

api/hypershift/v1beta1/hostedcluster_types.go
georgelipceanu: Adding this now, nice find 😄

georgelipceanu: I'm confused as to which other field you are referring to?

JoelSpeed: Well, you have a relationship between a field under the `networking` key, and then under `operatorConfiguration.clusterNetworkOperator`, so in terms of the API, they are somewhat distant, but clearly related else this condition wouldn't exist?

georgelipceanu: Ohhhhh I think I see what you mean, this was originally in `HostedClusterSpec` but was moved to `ClusterNetworkOperatorSpec`, this change was discussed in this review [here](https://github.com/openshift/hypershift/pull/6554#discussion_r2245617666). Let me know if that clears things up 🙌

LiangquanLi930:
@JoelSpeed When you have time, Could you help to review again?

JoelSpeed:
From an API perspective, I have no further feedback

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -570,7 +570,11 @@ func (r *reconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Result
 	log.Info("reconciling network operator")
 	networkOperator := networkoperator.NetworkOperator()
 	if _, err := r.CreateOrUpdate(ctx, r.client, networkOperator, func() error {
-		networkoperator.ReconcileNetworkOperator(networkOperator, hcp.Spec.Networking.NetworkType, hcp.Spec.Platform.Type, util.IsDisableMultiNetwork(hcp))
+		var ovnConfig *hyperv1.OVNKubernetesConfig
+		if hcp.Spec.OperatorConfiguration != nil && hcp.Spec.OperatorConfiguration.ClusterNetworkOperator != nil {
+			ovnConfig = hcp.Spec.OperatorConfiguration.ClusterNetworkOperator.OVNKubernetesConfig
+		}
+		networkoperator.ReconcileNetworkOperator(networkOperator, hcp.Spec.Networking.NetworkType, hcp.Spec.Platform.Type, util.IsDisableMultiNetwork(hcp), ovnConfig)
```
LiangquanLi930: The `ovnConfig` retrieval logic should be moved outside the `CreateOrUpdate` callback function. Here's the suggested change:

```go
log.Info("reconciling network operator")
networkOperator := networkoperator.NetworkOperator()

// Move config retrieval outside the callback
var ovnConfig *hyperv1.OVNKubernetesConfig
if hcp.Spec.OperatorConfiguration != nil && hcp.Spec.OperatorConfiguration.ClusterNetworkOperator != nil {
    ovnConfig = hcp.Spec.OperatorConfiguration.ClusterNetworkOperator.OVNKubernetesConfig
}

if _, err := r.CreateOrUpdate(ctx, r.client, networkOperator, func() error {
    networkoperator.ReconcileNetworkOperator(networkOperator, hcp.Spec.Networking.NetworkType, hcp.Spec.Platform.Type, util.IsDisableMultiNetwork(hcp), ovnConfig)
    return nil
}); err != nil {
    errs = append(errs, fmt.Errorf("failed to reconcile network operator: %w", err))
}
```
```

File: control-plane-operator/hostedclusterconfigoperator/controllers/resources/network/reconcile_test.go [hunk 12]
```diff
@@ -293,7 +293,7 @@ func TestReconcileDefaultIngressController(t *testing.T) {
 	for _, tc := range testsCases {
 		t.Run(tc.name, func(t *testing.T) {
 			g := NewGomegaWithT(t)
-			ReconcileNetworkOperator(tc.inputNetwork, tc.inputNetworkType, tc.inputPlatformType, tc.disableMultiNetwork)
```
LiangquanLi930, line 296: Could you add a unit test where the value is not nil?

georgelipceanu: Yes, working on this now 👍

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
georgelipceanu: At the moment, there are no other instances where `ovnConfig` is used, only inside that if block. This change visually looks easier to understand and would make sense if there was future plans for `ovnConfig` variable, so it that the case or is there a different reason for the suggested change?

LiangquanLi930: Good point! My main reason for suggesting the change is to improve clarity — extracting ovnConfig outside keeps the closure cleaner and focused only on reconciliation logic. It also makes the control flow easier to follow if the logic grows in the future.

georgelipceanu: That's fair, I'll add this outside the if block 👍

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -447,6 +447,7 @@ type Capabilities struct {
...
 // +kubebuilder:validation:XValidation:rule="!has(self.operatorConfiguration) || !has(self.operatorConfiguration.clusterNetworkOperator) || !has(self.operatorConfiguration.clusterNetworkOperator.disableMultiNetwork) || !self.operatorConfiguration.clusterNetworkOperator.disableMultiNetwork || self.networking.networkType == 'Other'",message="disableMultiNetwork can only be set to true when networkType is 'Other'"
+// +kubebuilder:validation:XValidation:rule="self.networking.networkType == 'OVNKubernetes' || !self.?operatorConfiguration.clusterNetworkOperator.ovnKubernetesConfig.hasValue()", message="ovnKubernetesConfig is forbidden when networkType is not OVNKubernetes"
```
LiangquanLi930, line 450: can we please add a case to TestOnCreateAPIUX to validate this cel rule.

georgelipceanu: Will do this now 🙌

georgelipceanu: `TestOnCreateAPIUX` is a test in `test/e2e/create_cluster_test.go` and requires a full E2E setup to run and verify. Will we not be saving this for the E2E ticket later on? Or shall we create unit test cases in the `reconcile.go` file?

LiangquanLi930: It's important to add an APIUX test.

georgelipceanu: Added these 🙏

control-plane-operator/hostedclusterconfigoperator/controllers/resources/network/reconcile_test.go
```diff
@@ -289,11 +290,245 @@ func TestReconcileDefaultIngressController(t *testing.T) {
...
+								InternalTransitSwitchSubnet: "192.168.2.0/24",
+							},
+						},
+					},
+				},
+			},
+		},
+		{
+			name:                "Only IPv4 join subnet configured",
+			inputNetwork:        NetworkOperator(),
```
LiangquanLi930: The test cases seem redundant. We can simplify them.

georgelipceanu: Removed test cases and left the following:
```
- OVN config with IPv4 subnets 
- OVN config with non-OVN network type 
- Nil OVN config
```

csrwng:
nit:  the PR description is not accurate

> This PR exposes the OVNKubernetesConfig  field in HCP's ClusterNetworking, allowing for future modifications to internal OVN IPv4 Subnets.

csrwng:
I would like to see an e2e validation that changes the cno operator config on the HostedCluster and validates that it is reflected in the hosted cluster's API and the cno doesn't report any errors via hcp conditions.

api/hypershift/v1beta1/operator.go
```diff
@@ -48,4 +48,57 @@ type ClusterNetworkOperatorSpec struct {
...
+	// internalJoinSubnet is a v4 subnet used internally by ovn-kubernetes in case the
+	// default one is being already used by something else. It must not overlap with
+	// any other subnet being used by OpenShift or by the node network. The size of the
+	// subnet must be larger than the number of nodes.
+	// The current default value is 100.64.0.0/16
+	// The subnet must be large enough to accommodate one IP per node in your cluster
+	// The value must be in proper IPV4 CIDR format
+	// +kubebuilder:validation:MaxLength=18
+	// +kubebuilder:validation:MinLength=9
+	// +kubebuilder:validation:XValidation:rule="isCIDR(self) && cidr(self).ip().family() == 4",message="Subnet must be in valid IPV4 CIDR format"
```
enxebre: this might break management clusters where isCIDR is not supported. Can you confirm since which ocp and kube version is isCIDR cel helper available?

JoelSpeed: 1.31 upstream, 4.16 downstream (edit: It was backported into earlier releases downstream)

georgelipceanu: The Kubernetes required version for Kubernetes CIDR library is stated to be **1.31** [here](https://kubernetes.io/docs/reference/using-api/cel/#cel-options-language-features-and-libraries), however
OCP version **4.16** is stated to be based on Kubenertes version 1.29, referenced [here](https://docs.redhat.com/en/documentation/openshift_container_platform/4.16/html/release_notes/ocp-4-16-release-notes#ocp-4-16-admin-ack-updating_release-notes). This feature is needed for 4.16 (found in the requirements in the description of [OCPSTRAT-1612](https://issues.redhat.com/browse/OCPSTRAT-1612)), implying that this CEL rule might not work on 4.16, or there could be something in the documentation that I'm missing?

georgelipceanu: PR for setting up CEL IP/CIDR library from 4.14 onwards https://github.com/openshift/kubernetes/pull/1913, which is inline with requirements.

rtheis: IBM manages clusters as old as 4.13 at the moment. Such clusters are unsupported but we do allow upgrades to 4.14.

rtheis: Sorry, I need to clarify that we run HyperShift operator on clusters as old as OpenShift version 4.15 and Kubernetes version 1.30.

rtheis: Per our interlock meeting, we will be pushing a PR to document IBM support levels and plan to keep the document current. Thanks for letting us know.

georgelipceanu: Added a modification to these rules that doesn't include `isCIDR` and works for versions as old as 4.13 👍

rtheis: Thank you @georgelipceanu

PR #6559, opened 2025-07-31
CNTRLPLANE-1116: build(gitlint): add gitlint to verify target

bryan-cox:
Include gitlint validation in the verify make target to ensure commit message formatting is checked as part of the standard development workflow.

Fixes #

**Checklist**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,7 +83,7 @@ lint-fix:
 	cd api && $(GOLANGCI_LINT) run --config ./.golangci.yml --fix -v
 
 .PHONY: verify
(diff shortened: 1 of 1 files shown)


muraee:
I thought we agreed to have a sperate job for this, instead of adding it to `verify`

bryan-cox:
> I thought we agreed to have a sperate job for this, instead of adding it to `verify`

@muraee while Seth was reviewing the PR to make it a separate job, he said he would prefer it to be added to verify. After looking at verify, verify doesn't take long to run. It's the images job prior to it running that take the longest.

PR #6315, opened 2025-06-24
CNTRLPLANE-935: e2e test for new Azure metrics

sdminonne:
This PR is built on top of #6180 , it adds e2e tests for HCP on managed azure (ARO)

Fixes #[CNTRLPLANE-935](https://issues.redhat.com//browse/CNTRLPLANE-935)

PR #6561, opened 2025-08-02
OCPBUGS-60376: fix token rotation causing Karpenter to stop working

maxcao13:
The first commit enables the ec2nodeclass controller to watch the karpenter userData secret
in the management cluster. This allows the userData that is embedded in the ec2nodeclass to be updated
correctly if needed whenever the associated secret is updated. 

It also allows the hostedcluster controller to watch the token secret, so that we can call `token.Reconcile` manually for the karpenter nodepool token.

This was an issue because after the token rotates, if you provision a new workload with karpenter, the node bootstrapping will reference the old token and fail to register as a new node when attempting to contact the ignition server since the userData changes did not trigger karpenter to be aware of the new token. 

---

~The second commit skips token rotation for the karpenter hyperv1.NodePool. Previously, if the token rotates, this
will cause existing Karpenter nodes to drift unnecessarily. 
This is a quick hack. The "correct" way to solve this is to make sure that a change like this do not cause drift in the karpenter-provider-aws code, (maybe as a carry since the way we pass the ignition-config to coreos nodes in hypershift seems very specialized).~

Above is removed. We will resolve that somewhere else but keep it documented in [AUTOSCALE-293](https://issues.redhat.com/browse/AUTOSCALE-293).

---

Second commit just improves some handling and logging in the e2e tests, and fixes up some test logic.

---

Third commit allows hypershift e2e tests to dump Karpenter resources from the guest cluster if they exist. I had to add the hypershift karpenter api to the global scheme to do this. Makes debugging autonode e2e tests a lot easier.

**Checklist**

## Summary by CodeRabbit

- New Features
  - Guest cluster diagnostics now include Karpenter resources when available.
  - Standardized Karpenter user data/AMI labeling and node pool naming via new constants.

- Bug Fixes
  - Karpenter token rotation secrets now reliably trigger HostedCluster reconciliation.
  - Consistent default name applied for EC2NodeClass.

- Refactor
  - EC2NodeClass reconciler watches relevant secret changes for better synchronization.
  - Finalization logs clarified and requeue intervals tuned.

- Tests
  - Stronger end-to-end checks for pod scheduling/readiness with extended timeouts.
  - Added unit tests for secret event filtering logic.

- Chores
  - Runtime scheme aligned with Karpenter APIs; certain metrics validations temporarily skipped.

### DIFF
--- a/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
+++ b/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
@@ -3361,6 +3361,16 @@ func enqueueHostedClustersFunc(metricsSet metrics.MetricsSet, operatorNamespace 
 		case *hyperv1.NodePool:
 			return []reconcile.Request{{NamespacedName: types.NamespacedName{Name: typedObj.Spec.ClusterName, Namespace: typedObj.Namespace}}}
 		case *corev1.Secret:
+			// reconcile the karpenter token rotations to the userData secret, since the ignition-server/tokensecret_controller does not control it
+			if hcAnnotation, exists := typedObj.Annotations[hyperutil.HostedClusterAnnotation]; exists {
+				if nodePoolAnnotation, exists := typedObj.Annotations[hyperkarpenterv1.TokenSecretNodePoolAnnotation]; exists {
+					// only match tokenSecret with nodepool name karpenter and ignore changes to userDataSecret to avoid duplicate updates
+					if hyperutil.ParseNamespacedName(nodePoolAnnotation).Name == hyperkarpenterv1.KarpenterNodePool && strings.HasPrefix(typedObj.Name, "token-") {
+						// we still need to map the request to it's hostedcluster to trigger the reconcile
+						return []reconcile.Request{{NamespacedName: hyperutil.ParseNamespacedName(hcAnnotation)}}
+					}
+				}
+			}
 			if typedObj.Name == manifests.KASServingCertSecret("").Name {
 				for _, ownerRef := range typedObj.OwnerReferences {
 					if ownerRef.Kind == "HostedControlPlane" {
--- a/hypershift-operator/controllers/nodepool/token.go
+++ b/hypershift-operator/controllers/nodepool/token.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
+	hyperkarpenterv1 "github.com/openshift/hypershift/api/karpenter/v1beta1"
 	"github.com/openshift/hypershift/hypershift-operator/controllers/manifests/ignitionserver"
 	"github.com/openshift/hypershift/support/backwardcompat"
 	"github.com/openshift/hypershift/support/globalconfig"
@@ -308,6 +309,12 @@ func (t *Token) reconcileTokenSecret(tokenSecret *corev1.Secret) error {
 	tokenSecret.Annotations[TokenSecretAnnotation] = "true"
 	tokenSecret.Annotations[TokenSecretNodePoolUpgradeType] = string(t.nodePool.Spec.Management.UpgradeType)
 	tokenSecret.Annotations[nodePoolAnnotation] = client.ObjectKeyFromObject(t.nodePool).String()
+	if t.hostedCluster.Spec.AutoNode != nil && t.hostedCluster.Spec.AutoNode.Provisioner.Name == hyperv1.ProvisionerKarpeneter &&
+		t.hostedCluster.Spec.AutoNode.Provisioner.Karpenter.Platform == hyperv1.AWSPlatform {
+		if t.nodePool.GetName() == hyperkarpenterv1.KarpenterNodePool {
+			tokenSecret.Annotations[supportutil.HostedClusterAnnotation] = client.ObjectKeyFromObject(t.ConfigGenerator.hostedCluster).String()
+		}
+	}
 	// active token should never be marked as expired.
 	delete(tokenSecret.Annotations, hyperv1.IgnitionServerTokenExpirationTimestampAnnotation)
 
@@ -365,9 +372,9 @@ func (t *Token) reconcileUserDataSecret(userDataSecret *corev1.Secret, token str
 	if t.hostedCluster.Spec.AutoNode != nil && t.hostedCluster.Spec.AutoNode.Provisioner.Name == hyperv1.ProvisionerKarpeneter &&
 		t.hostedCluster.Spec.AutoNode.Provisioner.Karpenter.Platform == hyperv1.AWSPlatform {
 		// TODO(alberto): prevent nodePool name collisions adding prefix to karpenter NodePool.
-		if t.nodePool.GetName() == "karpenter" {
+		if t.nodePool.GetName() == hyperkarpenterv1.KarpenterNodePool {
 			userDataSecret.Labels[hyperv1.NodePoolLabel] = fmt.Sprintf("%s-%s", t.nodePool.Spec.ClusterName, t.nodePool.GetName())
-			userDataSecret.Labels["hypershift.openshift.io/ami"] = t.userData.ami
+			userDataSecret.Labels[hyperkarpenterv1.UserDataAMILabel] = t.userData.ami
 		}
 
 	}
--- a/karpenter-operator/controllers/nodeclass/ec2_nodeclass_controller.go
+++ b/karpenter-operator/controllers/nodeclass/ec2_nodeclass_controller.go
@@ -33,14 +33,14 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/cluster"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
 const (
-	// userDataAMILabel is a label set in the userData secret generated for karpenter instances.
-	userDataAMILabel = "hypershift.openshift.io/ami"
-
 	finalizer = "hypershift.openshift.io/ec2-nodeclass-finalizer"
 )
 
@@ -85,8 +85,11 @@ func (r *EC2NodeClassReconciler) SetupWithManager(ctx context.Context, mgr ctrl.
 				return nil
 			},
 		)).
-		Watches(&awskarpenterv1.EC2NodeClass{}, &handler.EnqueueRequestForObject{})
-
+		Watches(&awskarpenterv1.EC2NodeClass{}, &handler.EnqueueRequestForObject{}).
+		// Watch secrets in the management cluster and reconcile all ec2nodeclasses
+		WatchesRawSource(source.Kind[client.Object](managementCluster.GetCache(), &corev1.Secret{},
+			handler.EnqueueRequestsFromMapFunc(r.mapToOpenShiftEC2NodeClasses),
+			r.userDataSecretPredicate()))
 	return bldr.Complete(r)
 }
 
@@ -220,7 +223,7 @@ func reconcileEC2NodeClass(ec2NodeClass *awskarpenterv1.EC2NodeClass, openshiftE
 		AMIFamily: ptr.To("Custom"),
 		AMISelectorTerms: []awskarpenterv1.AMISelectorTerm{
 			{
-				ID: string(userDataSecret.Labels[userDataAMILabel]),
+				ID: string(userDataSecret.Labels[hyperkarpenterv1.UserDataAMILabel]),
 			},
 		},
 		AssociatePublicIPAddress: openshiftEC2NodeClass.Spec.AssociatePublicIPAddress,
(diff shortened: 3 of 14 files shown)


ignition-server/controllers/tokensecret_controller.go
```diff
@@ -229,6 +229,16 @@ func (r *TokenSecretReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	if value, ok := r.PayloadStore.Get(token); ok {
 		log.Info("Payload found in cache")
 
+		// TODO(maxcao13): This is a hack. Token rotation will drift karpenter nodes, which is undesired.
```
bryan-cox: Can you include a Jira ticket that will address this TODO in the future?

karpenter-operator/controllers/nodeclass/ec2_nodeclass_controller.go
```diff
@@ -394,3 +400,27 @@ func (r *EC2NodeClassReconciler) getHCP(ctx context.Context) (*hyperv1.HostedCon
 
 	return &hcpList.Items[0], nil
 }
+
+func (r *EC2NodeClassReconciler) userDataSecretPredicate() predicate.Predicate {
```
bryan-cox, line 402: Can you add a unit test for this please?

maxcao13:
I'm probably gonna remove the second commit in here, and only present the first part because the first commit fix is fixing a whole other bug pretty much.

I'm also gonna create a ocpbugs specifically for that first commit, and leave the actual token rotation causing karpenter drift issue separate from this PR and fix it later down the line when we have feedback from upstream about a potential fix.

maxcao13:
Hm... seems like teardown isn't working for some reason and/or is very flaky. Something is blocking the hostedcluster to be torn down, but not sure what because the logs/events aren't up to date to the timeout. I believe it's the nodeclaims though.

Need to investigate further.

EDIT: cannot replicate this on my own devcluster, and i can't find enough information from the dumped guest cluster to debug further. so going to try some debug commits and retest.

EDIT2: after further investigation, I believe a lot of the test flakes are due to this bug that i've tracked here: https://issues.redhat.com/browse/AUTOSCALE-328

maxcao13:
/test e2e-aws-autonode

Let's see if it knows this exists.

enxebre:
lgtm, I'll defer to @muraee to tag

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -3414,6 +3414,12 @@ func enqueueHostedClustersFunc(metricsSet metrics.MetricsSet, operatorNamespace
 		case *hyperv1.NodePool:
 			return []reconcile.Request{{NamespacedName: types.NamespacedName{Name: typedObj.Spec.ClusterName, Namespace: typedObj.Namespace}}}
 		case *corev1.Secret:
+			// reconcile the karpenter token rotations to the userData secret, since the ignition-server/tokensecret_controller does not control it
+			if hcAnnotation, exists := typedObj.Annotations[hyperutil.HostedClusterAnnotation]; exists {
+				if hyperutil.ParseNamespacedName(hcAnnotation).Name == "karpenter" {
```
muraee: why are we checking that cluster name is "karpenter"?

karpenter-operator/controllers/nodeclass/ec2_nodeclass_controller.go
```diff
@@ -111,12 +117,12 @@ func (r *EC2NodeClassReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	}
 
 	openshiftEC2NodeClass := &hyperkarpenterv1.OpenshiftEC2NodeClass{}
-	if err := r.guestClient.Get(ctx, req.NamespacedName, openshiftEC2NodeClass); err != nil {
+	if err := r.guestClient.Get(ctx, crdNamespacedName, openshiftEC2NodeClass); err != nil {
```
muraee: this makes this controller only reconcile the default OpenshiftEC2NodeClass that we create. Users should be able to create other instances of OpenshiftEC2NodeClass

karpenter-operator/controllers/nodeclass/ec2_nodeclass_controller.go
```diff
@@ -394,3 +400,27 @@ func (r *EC2NodeClassReconciler) getHCP(ctx context.Context) (*hyperv1.HostedCon
...
+		if secret, ok := obj.(*corev1.Secret); ok {
+			if nodePoolLabel, exists := secret.Annotations[hyperv1.NodePoolLabel]; exists {
+				if util.ParseNamespacedName(nodePoolLabel).Name == "karpenter" {
+					return true
+				}
+			}
+		}
+		return false
+	}
+	userDataSecretPredicate := predicate.Funcs{
```
muraee, line 416: why not use `predicate.NewPredicateFuncs(filter)`

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
maxcao13: That was simply an checking an annotation that exists on the karpenter token secret, and only filtering on changes to that secret.

I added that annotation in this PR:
```go

	if t.hostedCluster.Spec.AutoNode != nil && t.hostedCluster.Spec.AutoNode.Provisioner.Name == hyperv1.ProvisionerKarpeneter &&
		t.hostedCluster.Spec.AutoNode.Provisioner.Karpenter.Platform == hyperv1.AWSPlatform {
		if t.nodePool.GetName() == "karpenter" {
			tokenSecret.Annotations[supportutil.HostedClusterAnnotation] = client.ObjectKeyFromObject(t.ConfigGenerator.hostedCluster).String()
		}
	}
```
Only karpenter related nodepool tokens will have that annotation with that namespaced name.

karpenter-operator/controllers/nodeclass/ec2_nodeclass_controller.go
maxcao13: Because we don't need to reconcile on delete, only on create and update. At least that's what I'm thinking.

That function applies the filter on all events.

maxcao13:
Not really sure why the tech-preview is failing, never mind the node upgrading, but just for the `releases:latest` payload image, the kubelet version of all the nodes say it is `v1.33.3` but the image itself relating to those nodes says that kubernetes is at `v1.33.4`?
```
Pull From: registry.ci.openshift.org/ocp/release@sha256:e1d8cb606dbade0b03d5bfecdf611691a37ebcc73ce7c75db43f40e5579c5189

Release Metadata:
  Version:  4.20.0-0.ci-2025-08-25-170352
  Upgrades: <none>
  Metadata:

Component Versions:
  kubectl          1.33.3         
  kubernetes       1.33.4         
  kubernetes-tests 1.33.3         
  machine-os       9.6.20250815-0 Red Hat Enterprise Linux CoreOS
```

We should probably change the way the test works so that it is not so brittle that something like this fails the test... but not in the scope of this PR.

EDIT: kubernetes version probably doesn't necessarily mean kubelet version, so I am just gonna get rid of that check.

karpenter-operator/controllers/nodeclass/ec2_nodeclass_controller.go
maxcao13: Thanks for catching that you right.

I'm instead now mapping all secret changes to reconcile all ec2nodeclasses in the guest cluster.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
muraee: you are using "client.ObjectKeyFromObject(t.ConfigGenerator.hostedCluster)" as the value here, which is the hostedCluster name not NodePool, so it wouldn't always be "karpenter" here, or am I missing something?
even if that was the case, I would rather we make it more explicit here and just set the value to "karpenter"

muraee: why not use the nodePoolAnnotation which we already set

karpenter-operator/controllers/nodeclass/ec2_nodeclass_controller.go
muraee: This is still not correct, this controller reconciles OpenshiftEC2NodeClass CRD and create EC2NodeClass CRs. users will be creating instances of OpenshiftEC2NodeClass only, they are forbidden to create EC2NodeClass CRs. 
This should stay as is getting the resource from req.NamespacedName, we also don't need the mapToEC2NodeClasses.

maxcao13: Sorry you are right about changing back to `req.NamespacedName`, for some reason I thought I did that already.

However, about the other problem:
Please correct me if I am wrong, but what I've seen is that since the controller is watching the userData secret from the management cluster, the controller will get a `reconcile.Request` from the namespace hostedcontrolplane namespace which doesn't exist in the guest cluster, so the
```
	if err := r.guestClient.Get(ctx, req.NamespacedName, openshiftEC2NodeClass); err != nil {
```
is going to fail every time.

So that's why I am mapping those requests to each existing node class. Though, it should be mapping to openshiftec2nodeclass not ec2nodeclass, so I can correct that. Does that make sense?

muraee: yes, that makes sense. thanks!

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
maxcao13: Okay never mind, I remember why we needed this.

It's because this is in the hostedcluster controller, and so like the other problem we discussed, we have to map the secret request to its hostedcluster. But the secret needs to be annotated with the details of the hostedcluster namespaced name, otherwise we don't know where it came from.

Though there is a mistake here which you pointed out. It should be something like this:
```go
// reconcile the karpenter token rotations to the userData secret, since the ignition-server/tokensecret_controller does not control it
if hcAnnotation, exists := typedObj.Annotations[hyperutil.HostedClusterAnnotation]; exists {
	if nodePoolAnnotation, exists := typedObj.Annotations[hyperv1.NodePoolLabel]; exists {
		// only match tokenSecret with nodepool name karpenter and ignore changes to userDataSecret to avoid duplicate updates
		if hyperutil.ParseNamespacedName(nodePoolAnnotation).Name == "karpenter" && strings.HasPrefix(typedObj.Name, "token-") {
			// we still need to map the request to it's hostedcluster to trigger the reconcile
			return []reconcile.Request{{NamespacedName: hyperutil.ParseNamespacedName(hcAnnotation)}}
		}
	}
}
```

Though, I don't know if it's okay for me to just apply this:
```go
tokenSecret.Annotations[supportutil.HostedClusterAnnotation] = client.ObjectKeyFromObject(t.ConfigGenerator.hostedCluster).String()
```
to all token secrets, instead of just the karpenter one.

karpenter-operator/controllers/nodeclass/ec2_nodeclass_controller.go
```diff
@@ -394,3 +400,48 @@ func (r *EC2NodeClassReconciler) getHCP(ctx context.Context) (*hyperv1.HostedCon
...
+		CreateFunc:  func(e event.CreateEvent) bool { return filterKarpenterSecret(e.Object) },
+		UpdateFunc:  func(e event.UpdateEvent) bool { return filterKarpenterSecret(e.ObjectNew) },
+		DeleteFunc:  func(e event.DeleteEvent) bool { return false },
+		GenericFunc: func(e event.GenericEvent) bool { return false },
+	}
+	return userDataSecretPredicate
+}
```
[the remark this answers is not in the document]
maxcao13: This is wrong...? it gets set here: https://github.com/maxcao13/hypershift/blob/dde53bd2a2e5f6a73c5da01728c6ccb504aacfe1/hypershift-operator/controllers/nodepool/token.go#L366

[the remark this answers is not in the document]
maxcao13: If you trace it back I'm pretty sure it's the same literal string. Just cause the variable is called `hyperv1.NodePoolLabel` doesn't mean it's set on a label.

It's getting set as an annotation, and I'm parsing it back from the annotations.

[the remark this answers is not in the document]
maxcao13: Well fair enough, we are using the same literal string as a key for an annotation and as a label on the same Secret, and the value of both are different which probably isn't great. 

I don't know which format we should stick with as I didn't write this, but I need to use the format where it's embedding the namespaced name of the `namespace/nodepoolName` so that it's the easiest for me to see if the userData secret is associated with a Karpenter hyperv1.NodePool or not. Otherwise I would have to do some string checking, like checking for `-karpenter` at the end of the string. 

But both those identical label and annotation keys on the Secret existed before this PR, so I would defer to reviewers on this. Third option which I was planning on doing, was just to ignore this and keep using both. @muraee @enxebre

karpenter-operator/controllers/nodeclass/ec2_nodeclass_controller.go
```diff
@@ -85,8 +88,11 @@ func (r *EC2NodeClassReconciler) SetupWithManager(ctx context.Context, mgr ctrl.
...
 			},
 		)).
-		Watches(&awskarpenterv1.EC2NodeClass{}, &handler.EnqueueRequestForObject{})
-
+		Watches(&awskarpenterv1.EC2NodeClass{}, &handler.EnqueueRequestForObject{}).
+		// Watch secrets in the management cluster and reconcile all ec2nodeclasses
+		WatchesRawSource(source.Kind[client.Object](managementCluster.GetCache(), &corev1.Secret{},
+			handler.EnqueueRequestsFromMapFunc(r.mapToOpenShiftEC2NodeClasses),
+			r.userDataSecretPredicate()))
 	return bldr.Complete(r)
```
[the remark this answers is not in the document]
maxcao13: Not in scope of this PR.

maxcao13:
Seems like we are failing autonode test validateMetrics because of https://github.com/openshift/hypershift/pull/6315 which seemed to remove the prometheus query, and replaced it with a query to the hypershift operator pod itself? 

Is the HO pod supposed to aggregate metrics? Not exactly sure what I should do here if we are trying _not_ to query a prometheus endpoint in the tests now.

EDIT: i've just disabled karpenter metric validation for now. I'll followup on this in a separate PR later.

karpenter-operator/controllers/nodeclass/ec2_nodeclass_controller.go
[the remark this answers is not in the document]
muraee: let's just introduce a new constant "TokenSecretNodePoolAnnotation" or smth and use it instead of `hyperv1.NodePoolLabel` which might be confusing.

karpenter-operator/controllers/nodeclass/ec2_nodeclass_controller.go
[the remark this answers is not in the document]
muraee: Open a follow-up GitHub issue for this discussion

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
muraee: this looks good to me now 👍

karpenter-operator/controllers/nodeclass/ec2_nodeclass_controller.go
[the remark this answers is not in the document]
muraee: @coderabbitai thoughts? :D

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -3361,6 +3361,16 @@ func enqueueHostedClustersFunc(metricsSet metrics.MetricsSet, operatorNamespace
 		case *hyperv1.NodePool:
 			return []reconcile.Request{{NamespacedName: types.NamespacedName{Name: typedObj.Spec.ClusterName, Namespace: typedObj.Namespace}}}
 		case *corev1.Secret:
+			// reconcile the karpenter token rotations to the userData secret, since the ignition-server/tokensecret_controller does not control it
+			if hcAnnotation, exists := typedObj.Annotations[hyperutil.HostedClusterAnnotation]; exists {
+				if nodePoolAnnotation, exists := typedObj.Annotations[hyperv1.NodePoolLabel]; exists {
```
muraee:
```suggestion
				if nodePoolAnnotation, exists := typedObj.Annotations[hyperv1.NodePoolِAnnotation]; exists {
```
can we please use the new constant here as well

hypershift-operator/controllers/nodepool/token.go
```diff
@@ -308,6 +309,12 @@ func (t *Token) reconcileTokenSecret(tokenSecret *corev1.Secret) error {
 	tokenSecret.Annotations[TokenSecretAnnotation] = "true"
 	tokenSecret.Annotations[TokenSecretNodePoolUpgradeType] = string(t.nodePool.Spec.Management.UpgradeType)
 	tokenSecret.Annotations[nodePoolAnnotation] = client.ObjectKeyFromObject(t.nodePool).String()
+	if t.hostedCluster.Spec.AutoNode != nil && t.hostedCluster.Spec.AutoNode.Provisioner.Name == hyperv1.ProvisionerKarpeneter &&
+		t.hostedCluster.Spec.AutoNode.Provisioner.Karpenter.Platform == hyperv1.AWSPlatform {
+		if t.nodePool.GetName() == hyperkarpenterv1.KarpenterNodePool {
+			tokenSecret.Annotations[supportutil.HostedClusterAnnotation] = client.ObjectKeyFromObject(t.ConfigGenerator.hostedCluster).String()
+		}
+	}
```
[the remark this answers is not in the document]
maxcao13: I'll resolve this in a separate PR.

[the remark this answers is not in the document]
maxcao13: Uh... sure if you can? :joy:

maxcao13:
If we could merge https://github.com/openshift/hypershift/pull/6714 first, then I can totally just remove this commit: https://github.com/openshift/hypershift/pull/6561/commits/aff0a64ed27805fd552afb4ccd3518f43087f065

PR #6607, opened 2025-08-05
OCPBUGS-57626: order ICSPs for determinism of OPENSHIFT_IMG_OVERRIDES

sjenning:
If there are multiple ICSPs configured in the mgmt cluster, the List is not determinstic and the ICSPs can be returned in any order.  This can lead to redeployment of the CPO when reordering on the OPENSHIFT_IMG_OVERRIDES occurs.

This change sorts the ICSPs by name so that OPENSHIFT_IMG_OVERRIDES is deterministic.

### DIFF
--- a/support/globalconfig/imagecontentsource.go
+++ b/support/globalconfig/imagecontentsource.go
@@ -3,6 +3,8 @@ package globalconfig
 import (
 	"context"
 	"fmt"
+	"sort"
+	"strings"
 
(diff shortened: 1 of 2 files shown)


csrwng:
lgtm
It'd be great if you can add a unit test for GetAllImageRegistryMirrors that ensures that it returns results in deterministic ordering.

PR #6612, opened 2025-08-06
OCPBUGS-60190: chore: Improve security context uid allocation to reuse released uids

enxebre:
This is an iterative follow up for https://github.com/openshift/hypershift/pull/6520
Before this, uids would just grow sequentially forever

Fixes #

**Checklist**

### DIFF
--- a/hypershift-operator/controllers/hostedcluster/security_context_uid.go
+++ b/hypershift-operator/controllers/hostedcluster/security_context_uid.go
@@ -1,0 +1,104 @@
+package hostedcluster
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+	"sync"
+	"time"
+
+	controlplanecomponent "github.com/openshift/hypershift/support/controlplane-component"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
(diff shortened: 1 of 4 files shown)


bryan-cox:
/retitle OCPBUGS-59736: chore: Improve security context uid allocation to reuse released uids

bryan-cox:
This is a follow on to https://github.com/openshift/hypershift/pull/6520

hypershift-operator/controllers/hostedcluster/security_context_uid.go
```diff
@@ -0,0 +1,104 @@
...
+	// lastRefresh tracks when we last refreshed from namespaces
+	lastRefresh time.Time
+}
+
+var globalUIDAllocator = &securityContextUIDAllocator{
+	allocatedUIDs: make(map[int64]struct{}),
+}
+
+// initializeFromNamespaces scans existing namespaces to initialize the allocator
+func (a *securityContextUIDAllocator) initializeFromNamespaces(ctx context.Context, c client.Client) error {
```
stevekuznetsov, line 45: @enxebre can't you just use an indexer on your lister? Shouldn't need (another) in-memory store, or the locking, etc

enxebre: @stevekuznetsov great point. I think we might still need locking to complete the allocateUID logic in a thread safety manner? but other than that yeh that would be simpler.

PR #6639, opened 2025-08-12
CNTRLPLANE-1113: feat(ci): Add renovate config to ignore go.mod/sum updates

jparrill:

Update renovate config to:
- Ignore go.mod updates in branch
- Still catch the Tekton updates from Konflux

Also ignored dev folder in .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,6 @@ tools/bin
 
 # Mocked interface generate through mockgen
 *_mock.go
+
+# Dev
+dev/
--- a/renovate.json
+++ b/renovate.json
@@ -1,0 +1,21 @@
+{
+    "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+    "enabled": true,
+    "enabledManagers": ["tekton"],
+    "packageRules": [
+        {
+            "matchManagers": ["gomod"],
+            "enabled": false
+        },
+        {
+            "matchFileNames": ["go.mod", "go.sum"],
+            "enabled": false
+        }
+    ],
+    "tekton": {
+        "branchConcurrentLimit": 10,
+        "extends": [
+            "github>konflux-ci/mintmaker-presets:update-fedora-to-stable"
+        ]
+    }
+}


.gitignore
```diff
@@ -40,3 +40,6 @@ tools/bin
 
 # Mocked interface generate through mockgen
 *_mock.go
+
+# Dev
+dev/
```
bryan-cox, line 45: Why was this needed?

renovate.json
```diff
@@ -0,0 +1,21 @@
+{
```
bryan-cox, line 1: Is there a source/document you followed to get this info for what should be here?

.gitignore
jparrill: I want to prevent the dev folder to be added to a PR by accident. I did it recently :/

renovate.json
jparrill: Renovate Official docs + claude

bryan-cox: could you link the docs here in this comment for future ref?

jparrill: Yes:
- EnableManagers: https://docs.renovatebot.com/configuration-options/#enabledmanagers
- MatchManagers: https://docs.renovatebot.com/configuration-options/#matchmanagers
- MatchFilenames: https://docs.renovatebot.com/configuration-options/#matchfilenames

renovate.json
```diff
@@ -0,0 +1,21 @@
...
+        {
+            "matchFileNames": ["go.mod", "go.sum"],
+            "enabled": false
+        }
+    ],
+    "tekton": {
+        "branchConcurrentLimit": 10,
+        "extends": [
+            "github>konflux-ci/mintmaker-presets:update-fedora-to-stable"
+        ]
```
enxebre, line 19: why does this file look different than the one in main?

sjenning:
/hold
on Alberto's comment

renovate.json
jparrill: because in main we don't block those updates, only on release branches.
We agreed to:
- Keep updates from renovate in main (all)
- Keep only Tekton updates in the rest of release branches

jparrill:
/close

We are using another approach within the team

PR #6642, opened 2025-08-12
OCPBUGS-56430: fix: Add recovery monitoring stack fixes

jparrill:
This PR include some fixes for the HCCO that will be triggered after a HostedCluster is restored:
- Use cpClient to update HCP status for restoration condition
- Get the STS pods before deletion to identify them properly
- Delete the PVCs from openshift-monitoring namespace
- Add comprehensive unit tests for RecoverMonitoringStack function which includes a concurrent test with 10 goroutines to verify thread-safety
- Use of AtomicBool to ensure we only hit the recovery once

## Test in ROSA
- Image: `quay.io/jparrill/hypershift:OCPBUGS-56430-v53`
- ✅ Verified in ROSA Int

Fixes #[OCPBUGS-56430](https://issues.redhat.com/browse/OCPBUGS-56430)

### DIFF
--- a/control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
+++ b/control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
@@ -744,35 +744,44 @@ func (r *reconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Result
 
 	// Reconcile hostedCluster recovery if the hosted cluster was restored from backup
 	if _, exists := hcp.Annotations[hyperv1.HostedClusterRestoredFromBackupAnnotation]; exists {
+		var (
+			finished bool
+			err      error
+		)
 		condition := &metav1.Condition{
 			Type:   string(hyperv1.HostedClusterRestoredFromBackup),
 			Reason: hyperv1.RecoveryFinishedReason,
 		}
 
-		if err := r.reconcileRestoredCluster(ctx, hcp); err != nil {
+		finished, err = r.reconcileRestoredCluster(ctx, hcp)
+		if err != nil {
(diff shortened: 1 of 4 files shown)


jparrill:
/hold

Until test on ROSA env

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -744,35 +744,44 @@ func (r *reconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Result
 
 	// Reconcile hostedCluster recovery if the hosted cluster was restored from backup
 	if _, exists := hcp.Annotations[hyperv1.HostedClusterRestoredFromBackupAnnotation]; exists {
+		var (
```
csrwng, line 747: We should not only rely on the annotation but also on the condition that is set in this controller. The annotation is removed by the HO, which is a separate component and could be in a non-functioning state when this code runs.

jparrill: Makes sense. Fixed!

jparrill:
/hold

until new version is tested in ROSA int

jparrill:
I’ve tested the solution in ROSA multiple times, but it’s becoming more complex than initially expected. The new fixes introduced after the PR review are not working as intended.

What I propose is:
- Revert to the version that worked well and was tested in ROSA.
- Get that merged and backported to 4.19.
- Create a follow-up task to enhance the implementation (including @csrwng's suggestion).
- Develop and test the improved implementation thoroughly, without rushing.
- Merge it and backport it to 4.19.z.

The current version (last force-pushed) in this PR is the one we have verified in ROSA int.

PR #6643, opened 2025-08-13
OCPBUGS-60381: feat(validations): add wildcard support to DNS SAN conflict detection

jparrill:
Add wildcard support to DNS SAN conflict detection:
- Add RFC-compliant wildcard matching for DNS names in SAN validation
- Implement bidirectional conflict detection with wildcard support
- Add comprehensive test coverage and refactor for better readability
- Support patterns like *.example.com matching sub.example.com

- Fixes #[OCPBUGS-60381](https://issues.redhat.com/browse/OCPBUGS-60381)


hypershift-operator/controllers/hostedcluster/validations/ocpapiserver.go
```diff
@@ -57,6 +60,16 @@ func ValidateOCPAPIServerSANs(ctx context.Context, hc *hyperv1.HostedCluster, cl
 			}
 		}
 
+		if _, exists := hc.Annotations[hyperv1.SkipKASConflicSANValidation]; exists {
+			log.Info("Skipping KAS certificate SANs validation due to annotation", "annotation", hyperv1.SkipKASConflicSANValidation)
+			return errs
+		}
+
+		if hc.Spec.Platform.AWS.EndpointAccess == hyperv1.Private || hc.Spec.Platform.AWS.EndpointAccess == hyperv1.PublicAndPrivate {
```
csrwng: pls add check for `hc.Spec.Platform.AWS != nil`

csrwng (commented):

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -375,6 +375,9 @@ const (
 	// referenced in the HostedCluster `spec.configuration.authentication.oidcProviders[*].oidcClients[*].clientSecret`
 	// and only for the ARO-HCP platform.
 	HostedClusterSourcedAnnotation = "hypershift.openshift.io/hosted-cluster-sourced"
+
+	// SkipKASCertificateConflicSANValidation allows skipping the validation of the KAS certificate SANs that not conflicts with ServicePublishingStrategy Hostname.
```
bryan-cox:
```suggestion
	// SkipKASCertificateConflicSANValidation allows skipping the validation of the KAS certificate SANs so they do not conflict with ServicePublishingStrategy Hostname.
```

bryan-cox: Could we add when you might want to do that or explain what use case when one would want to use this annotation?

hypershift-operator/controllers/hostedcluster/validations/ocpapiserver.go
```diff
@@ -57,6 +60,16 @@ func ValidateOCPAPIServerSANs(ctx context.Context, hc *hyperv1.HostedCluster, cl
...
+			log.Info("Skipping KAS certificate SANs validation due to AWS endpoint access is private or public and private", "endpointAccess", hc.Spec.Platform.AWS.EndpointAccess)
```
bryan-cox:
```suggestion
			log.Info("Skipping KAS certificate SANs validation due to AWS endpoint access value", "endpointAccess", hc.Spec.Platform.AWS.EndpointAccess)
```

hypershift-operator/controllers/hostedcluster/validations/ocpapiserver.go
```diff
@@ -150,11 +163,63 @@ func appendEntriesIfNotExists(slice []string, entries []string) []string {
 	return slice
 }
 
+// isDNSNameMatch checks if a DNS name matches a pattern, supporting wildcards.
+// Wildcards are only allowed at the beginning of a domain and can only match one label.
+// Examples:
+// - "*.example.com" matches "sub.example.com" but not "sub.sub.example.com"
+// - "*.foo.bar.com" matches "baz.foo.bar.com" but not "qux.baz.foo.bar.com"
+func isDNSNameMatch(name, pattern string) bool {
```
bryan-cox, line 171: Could we also return the reason if the DNS name doesn't match? Then we could return that to the log message on L220. I think that might be a better UX.

hypershift-operator/controllers/hostedcluster/validations/ocpapiserver.go
bryan-cox: Is this only for the AWS platform? Just wondering if we should go ahead and include Azure for self-managed 🤔

hypershift-operator/controllers/hostedcluster/validations/ocpapiserver.go
jparrill: It's enough with:

```
return fmt.Errorf("conflicting %s found in KAS SANs. kasEntry: %s, customEntry: %s. The configuration is invalid because the custom %s is conflicting with the KAS SANs", entryType, kasEntry, customEntry, entryType)
```

jparrill:
Yeah @bryan-cox it's only relevant in AWS Private and PublicAndPrivate clusters

hypershift-operator/controllers/hostedcluster/validations/ocpapiserver.go
jparrill: Yes, only relevant for AWS

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -375,6 +375,10 @@ const (
 	// referenced in the HostedCluster `spec.configuration.authentication.oidcProviders[*].oidcClients[*].clientSecret`
 	// and only for the ARO-HCP platform.
 	HostedClusterSourcedAnnotation = "hypershift.openshift.io/hosted-cluster-sourced"
+
+	// SkipKASCertificateConflicSANValidation allows skipping the validation of the KAS certificate SANs so they do not conflict with ServicePublishingStrategy Hostname.
+	// This annotation is useful as a scape hatch, that IBM could use.
```
bryan-cox:
```suggestion
	// This annotation is useful as a escape hatch, that IBM could use.
```

csrwng: "an escape hatch"

jparrill:
/jira backport release-4.19, release-4.18, release-4.17, release-4.16, release-4.15

PR #6673, opened 2025-08-18
OCPBUGS-60602: Cache Azure KMS TokenCredential for ARO HCP and clarify CPO credentials cache naming

bryan-cox:

- Caches the Azure KMS TokenCredential per namespace when running on ARO HCP to avoid repeated credential creation and external calls during reconcile, improving performance.
- Clarifies naming for the control-plane-operator’s Azure credentials cache to reduce ambiguity and improve readability.
- Behavior is unchanged for non-ARO environments.

Changes include:
- Add kmsAzureCredentials key and kmsAzureCredentialsLoaded cache in HostedControlPlaneReconciler.
- Update validateAzureKMSConfig to Load/Store the cached credential when IsAroHCP().
- Improve log messages to explicitly reference KMS and CPO flows.
- Rename constants/fields: azureCredentials → cpoAzureCredentials, azureCredentialsLoaded → cpoAzureCredentialsLoaded.

- Fixes OCPBUGS-60602

**Checklist**

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
@@ -146,6 +146,7 @@ const (
 	hcpNotReadyRequeueInterval = 15 * time.Second
 
 	cpoAzureCredentials = "CPOAzureCredentials"
+	kmsAzureCredentials = "KMSAzureCredentials"
 )
 
 type HostedControlPlaneReconciler struct {
@@ -179,6 +180,7 @@ type HostedControlPlaneReconciler struct {
 	EnableCVOManagementClusterMetricsAccess bool
 	ImageMetadataProvider                   util.ImageMetadataProvider
 	cpoAzureCredentialsLoaded               sync.Map
+	kmsAzureCredentialsLoaded               sync.Map
 }
 
 func (r *HostedControlPlaneReconciler) SetupWithManager(mgr ctrl.Manager, createOrUpdate upsert.CreateOrUpdateFN, hcp *hyperv1.HostedControlPlane) error {
@@ -3253,6 +3255,14 @@ func (r *HostedControlPlaneReconciler) validateAWSKMSConfig(ctx context.Context,
 }
 
 func (r *HostedControlPlaneReconciler) validateAzureKMSConfig(ctx context.Context, hcp *hyperv1.HostedControlPlane) {
+	var (
+		cred azcore.TokenCredential
+		ok   bool
+		err  error
+	)
+
+	log := ctrl.LoggerFrom(ctx)
+
 	if hcp.Spec.SecretEncryption == nil || hcp.Spec.SecretEncryption.KMS == nil || hcp.Spec.SecretEncryption.KMS.Azure == nil {
 		condition := metav1.Condition{
 			Type:               string(hyperv1.ValidAzureKMSConfig),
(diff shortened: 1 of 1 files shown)


jparrill (commented):
Just a question, we won't want to add cache to the TokenCreds isn't it? Otherwise LGTM

bryan-cox:
> Just a question, we won't want to add cache to the TokenCreds isn't it? Otherwise LGTM

We handled this through a Slack huddle. We cache the Azure token for the life of the pod. If the pod is restarted, it is expected we would reauthenticate once more to Azure and then reuse that token for the life of the restarted pod. The Microsoft authentication method handles the refresh of expired certificates in the background.

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -178,7 +179,8 @@ type HostedControlPlaneReconciler struct {
 	reconcileInfrastructureStatus           func(ctx context.Context, hcp *hyperv1.HostedControlPlane) (infra.InfrastructureStatus, error)
 	EnableCVOManagementClusterMetricsAccess bool
 	ImageMetadataProvider                   util.ImageMetadataProvider
-	azureCredentialsLoaded                  sync.Map
+	cpoAzureCredentialsLoaded               sync.Map
```
stevekuznetsov, line 182: super-nit: this reads like a set of booleans? `azureCredentials` or `loadedAzureCredentials` or `cachedAzureCredentials` all might make more sense

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -3253,6 +3255,14 @@ func (r *HostedControlPlaneReconciler) validateAWSKMSConfig(ctx context.Context,
 }
 
 func (r *HostedControlPlaneReconciler) validateAzureKMSConfig(ctx context.Context, hcp *hyperv1.HostedControlPlane) {
+	var (
+		cred azcore.TokenCredential
+		ok   bool
```
stevekuznetsov, line 3260: Why do you need the ok and err here? Seems like they get handled inline

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -3365,25 +3393,26 @@ func (r *HostedControlPlaneReconciler) verifyResourceGroupLocationsMatch(ctx con
 		err       error
 	)
 
-	key := hcp.Namespace + azureCredentials
+	key := hcp.Namespace + cpoAzureCredentials
```
stevekuznetsov, line 3396: Since your cache keys are specific to which credentials you are loading, you only need one sync.Map, or you could use the namespace alone as the cache key and keep the two maps

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -3266,13 +3276,31 @@ func (r *HostedControlPlaneReconciler) validateAzureKMSConfig(ctx context.Contex
...
+		if !found {
+			// Retrieve the KMS UserAssignedCredentials path
+			credentialsPath := config.ManagedAzureCredentialsPathForKMS + hcp.Spec.SecretEncryption.KMS.Azure.KMS.CredentialsSecretName
+			cred, err := dataplane.NewUserAssignedIdentityCredential(ctx, credentialsPath, dataplane.WithClientOpts(azcore.ClientOptions{Cloud: cloud.AzurePublic}))
+			if err != nil {
+				conditions.SetFalseCondition(hcp, hyperv1.ValidAzureKMSConfig, hyperv1.InvalidAzureCredentialsReason,
+					fmt.Sprintf("failed to obtain azure client credentials: %v", err))
+				return
+			}
+			r.kmsAzureCredentialsLoaded.Store(key, cred)
```
machi1990, line 3293: Some potential improvements we could consider on this;
- After HCP deletion (once the cloud resources have been removed), we can invalidate the cache by removing the key 
- when invalidating the cache, let's cancel the context that was used to create the creds; That way, we stop the background worker underneath that handles the refreshing logic and close the `fswatcher`; https://github.com/Azure/msi-dataplane/blob/main/pkg/dataplane/reloadCredentials.go#L110

bryan-cox: On HCP deletion, the CPO will be deleted, which will delete the cache.

What's the use case for invalidating the cache? The certificate behind the credentials is being watched and reloaded. 

Maybe if the customer wants to delete the managed identity tied to a control plane component? If they want to do that, they need to contact CS to change the HC and the backing credentials behind the MI. 

WDYT @stevekuznetsov?

machi1990: > On HCP deletion, the CPO will be deleted, which will delete the cache.

This is the part that I missed / didn't take onto consideration while reviewing the worker.
Can you share pointers as to where this happens? 

> What's the use case for invalidating the cache? The certificate behind the credentials is being watched and reloaded.
> Maybe if the customer wants to delete the managed identity tied to a control plane component? If they want to do that, they need to contact CS to change the HC and the backing credentials behind the MI.

I don't think this use case is supported at the moment so TBH it is not something I was not concerned about.

machi1990: > This is the part that I missed / didn't take onto consideration while reviewing the worker.
Can you share pointers as to where this happens?

@bryan-cox (thank you) pointed out to me that CPO is 1:1 to HostedCluster. So when HCP is deleted, the CPO pod is deleted; technically it is at this point where invalidation occurs.

PR #6686, opened 2025-08-20
STOR-1064: add validator image to CSO manifest

RomanBednar:
We're adding a new image and controller to CSO and it seems like I need to reflect this in hypershift manifest for CSO as well because hypershift CI jobs are currently failing due to missing image: https://github.com/openshift/cluster-storage-operator/pull/592#issuecomment-3206818015

/cc @openshift/openshift-team-hypershift


RomanBednar:
/assign @bryan-cox 

For review.

dobsonj:
zomg the retesting :facepalm:

PR #6695, opened 2025-08-25
fix(OCPBUGS-60649): add availibility-prober to azure capi-provider

sjenning:
The AWS capi-provider uses the availability-prober init container to avoid pod restarts due to the KAS not yet being available.

The PR adds the same thing for the Azure capi-provider.

### DIFF
--- a/hypershift-operator/controllers/hostedcluster/internal/platform/azure/azure.go
+++ b/hypershift-operator/controllers/hostedcluster/internal/platform/azure/azure.go
@@ -9,11 +9,13 @@ import (
 
 	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
 	"github.com/openshift/hypershift/control-plane-operator/controllers/hostedcontrolplane/cloud/azure"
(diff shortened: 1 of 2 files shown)


wallylewis:
@coderabbit review

bryan-cox:
@coderabbitai review

bryan-cox:
^ I got that command from https://docs.coderabbit.ai/guides/commands

hypershift-operator/controllers/hostedcluster/internal/platform/azure/azure.go
```diff
@@ -191,7 +195,7 @@ func (a Azure) CAPIProviderDeploymentSpec(hcluster *hyperv1.HostedCluster, _ *hy
 			},
 		)
 	}
-
```
enxebre: any reason to not do this in the v2Component definition via InjectAvailabilityProberContainer? that would also impact all other providers

sjenning: Only that that wasn't how it was done for AWS capi provider.  I can refactor to get it  into the v2 component.

sjenning:
needed for component readiness stability

PR #6707, opened 2025-08-27
OCPSTRAT-1677 | feat: allow for aws spot market options on node pools

gdbranco:

Fixes #

**Checklist**


docs/content/how-to/aws/create-aws-hosted-cluster-spot-instances.md
```diff
@@ -0,0 +1,216 @@
+---
```
enxebre: I would expect this to just be a section of https://hypershift.pages.dev/how-to/automated-machine-management/ focusing on the nodepool peculiarities, no need to redocument cluster creation

docs/content/how-to/aws/create-aws-hosted-cluster-spot-instances.md
```diff
@@ -0,0 +1,216 @@
...
+This guide demonstrates how to create a hosted cluster with node pools configured to use AWS Spot instances.
+
+## Prerequisites
+
+- [Install the latest and greatest HyperShift CLI](https://hypershift-docs.netlify.app/getting-started/#prerequisites).
+    - Make sure all prerequisites have been satisfied (Pull Secret, Hosted Zone, OIDC Bucket, etc)
+- Ensure that the AWS service-linked role for Spot is enabled in the account where the hosted cluster will be installed. This is a one-time setup per account.
+    - You can verify if the role already exists using the following command:
+```sh
+aws iam get-role --role-name AWSServiceRoleForEC2Spot
```
enxebre: this shoud be automated by the cli iam management

enxebre: This PR needs to add a new e2e for spot. Our e2e uses the same managed policies than rosa and let's you include additional perms

docs/content/how-to/aws/create-aws-hosted-cluster-spot-instances.md
```diff
@@ -0,0 +1,216 @@
...
+
+### Cost savings
+
+AWS Spot instances can provide significant cost savings, often 50-90% less than On-Demand pricing, depending on the instance type and availability.
+
+### Availability and interruptions
+
+- Spot instances are subject to interruption when AWS needs the capacity back for On-Demand customers
+- Instances receive a 2-minute warning before termination
+- HyperShift automatically handles instance replacement when Spot instances are terminated
```
enxebre: >- HyperShift automatically handles instance replacement when Spot instances are terminated

can you articulate how? what happens when an instance is terminated with and without autorepair enabled and how is that failure bubble up?
we will also need to ship a termination handler for detecting upcoming interruptions and do a best effort to handle them gracefully

api/hypershift/v1beta1/aws.go
```diff
@@ -57,6 +57,20 @@ type AWSNodePoolPlatform struct {
 	//
 	// +optional
 	Placement *PlacementOptions `json:"placement,omitempty"`
+
+	// spotMarketOptions specifies options for using AWS Spot instances.
+	// When specified, instances will be launched as Spot instances with the given configuration.
+	// +optional
+	SpotMarketOptions *AWSSpotMarketOptions `json:"spotMarketOptions,omitempty"`
```
enxebre: if this field is mutually exclusive with many of the placement options, it needs to be enforced by cel and documented

enxebre:
@coderabbitai review

docs/content/how-to/aws/create-aws-hosted-cluster-spot-instances.md
enxebre: also how about https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/service-linked-roles-spot-instance-requests.html#spot-instance-service-linked-roles-access-to-cmks

docs/content/how-to/aws/create-aws-hosted-cluster-spot-instances.md
gdbranco: My understanding is that due to auto repair setting part of the node pool management when AWS terminates the spot instance it will cause a failure in the cluster infra which in turn will cause the node pool management to replace it trying to maintain the desired replica count. But for a more graceful approach it would indeed require more integration for detecting interruptions

docs/content/how-to/aws/create-aws-hosted-cluster-spot-instances.md
gdbranco: Rosa managed policies (worker account role) should suffice in terms of permissions for spot instances already. In regards to the grant due to KMS key I would need to double check, but given ROSA already offers the possibility of using custom KMS keys I expect the customer would need to handle the grant and include the worker role as a principal

gdbranco: https://cloud.redhat.com/experts/rosa/kms/

Example documentation capturing that

api/hypershift/v1beta1/aws.go
```diff
@@ -57,6 +58,25 @@ type AWSNodePoolPlatform struct {
 	//
 	// +optional
 	Placement *PlacementOptions `json:"placement,omitempty"`
+
+	// spotMarketOptions specifies options for using AWS Spot instances.
+	// When specified, instances will be launched as Spot instances with the given configuration.
+	// Mutually exclusive with placement.capacityReservation, and placement.tenancy must be unset or set to "default".
+	// These constraints are enforced via CEL on AWSNodePoolPlatform.
+	// +optional
+	SpotMarketOptions *AWSSpotMarketOptions `json:"spotMarketOptions,omitempty"`
```
muraee: This field should be part of the `PlacementOptions` field

test/e2e/nodepool_spotmarketoptions_test.go
```diff
@@ -0,0 +1,245 @@
...
+		*nodePool.Spec.Platform.AWS.SpotMarketOptions.MaxPrice,
+		nodePool.Spec.Platform.AWS.Placement.Tenancy)
+
+	// Verify nodes were created
+	g.Expect(nodes).ToNot(BeEmpty())
+	t.Logf("Successfully provisioned %d Spot instance nodes", len(nodes))
+}
+
+// SpotMarketOptionsFailureTest tests invalid SpotMarketOptions configurations
+type SpotMarketOptionsFailureTest struct {
```
muraee: This is not needed, we do have a separate test for testing API CEL validations, please add new test cases to `TestOnCreateAPIUX` here https://github.com/openshift/hypershift/blob/main/test/e2e/create_cluster_test.go#L32 instead

hypershift-operator/controllers/nodepool/aws.go
```diff
@@ -164,6 +166,15 @@ func awsMachineTemplateSpec(infraName string, hostedCluster *hyperv1.HostedClust
 		}
 	}
 
+	if nodePool.Spec.Platform.AWS.SpotMarketOptions != nil {
+		awsMachineTemplateSpec.Template.Spec.SpotMarketOptions = &capiaws.SpotMarketOptions{
+			MaxPrice: nodePool.Spec.Platform.AWS.SpotMarketOptions.MaxPrice,
+		}
+		if awsMachineTemplateSpec.Template.Spec.MarketType == "" {
+			awsMachineTemplateSpec.Template.Spec.MarketType = capiaws.MarketTypeSpot
```
muraee: how does this work. If MarketType is set as "On-Demand" and with `SpotMarketOptions` what will happen? I supposed marketType be Spot for MaxPrice to take effect, no?

gdbranco: Well yes, but there is nothing that allows direct interaction with the Market Type at the moment and due to the CRD validation it will not allow to set capacity reservation alongside spotmarketoptions. I can also just straight up force market type to be Spot in case spot market options is populated, that way it will always guarantee the use of Spot

muraee: yes, this is better. Thanks.
Ideally we would support `Spot` market type as an option in the `markeType` field here https://github.com/openshift/hypershift/blob/main/api/hypershift/v1beta1/aws.go#L126 but we made a mistake making  that field part of the CapacityReservationOptions. We might need to reconsider this.

bryan-cox:
This shouldn't be linked to an OCPSTRAT

api/hypershift/v1beta1/aws.go
```diff
@@ -59,8 +59,21 @@ type AWSNodePoolPlatform struct {
 	Placement *PlacementOptions `json:"placement,omitempty"`
 }
 
+// AWSSpotMarketOptions defines configuration for AWS Spot instances
+type AWSSpotMarketOptions struct {
+	// maxPrice defines the maximum price (USD per instance‑hour) you are willing to pay for a Spot instance.
+	// If omitted, the On‑Demand price is used as the ceiling.
```
enxebre, line 65: >// If omitted, the On‑Demand price is used as the ceiling.

How can I express intent for this with this API?

docs/content/how-to/automated-machine-management/aws-spot-instances.md
```diff
@@ -0,0 +1,221 @@
...
+
+### Cost savings
+
+AWS Spot instances can provide significant cost savings, often 50-90% less than On-Demand pricing, depending on the instance type and availability.
+
+### Availability and interruptions
+
+- Spot instances are subject to interruption when AWS needs the capacity back for On-Demand customers
+- Instances receive a 2-minute warning before termination
+- HyperShift automatically handles instance replacement when Spot instances are terminated through the NodePool's `autoRepair` functionality
```
enxebre, line 115: >- Instances receive a 2-minute warning before termination

In standalone we run a termination handler which watches this events, mark the machines "to be terminated" and let mhc trigger draining/deletion early as a best effort right before aws terminate them. That's not plumbed here yet and so expectations should be set accordingly for consumers of this feature. Is there a jira ticket to track that as part of this epic

>- HyperShift automatically handles instance replacement when Spot instances are terminated through the NodePool's `autoRepair` functionality

autoRepair is an opt-in features, so this is only true IF autoRepair is enabled. Let's please clarify that here.

test/e2e/nodepool_spotmarketoptions_test.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+	defaultNodepool.Spec.DeepCopyInto(&nodePool.Spec)
+	nodePool.Spec.Replicas = &oneReplicas
+	if nodePool.Spec.Platform.AWS == nil {
+		nodePool.Spec.Platform.AWS = &hyperv1.AWSNodePoolPlatform{}
+	}
+	nodePool.Spec.Platform.AWS.Placement = &hyperv1.PlacementOptions{
+		Tenancy: "default",
+		SpotMarketOptions: &hyperv1.AWSSpotMarketOptions{
+			MaxPrice: ptr.To("0.50"),
```
enxebre, line 52: how likely is this to cause flakes because of spot unavailability?
Is the error message bubbled up to nodepool status via NodePoolAllMachinesReadyConditionType?
If so can we also check that and consider it part of the success criteria for the test.
If it's not bubbled up, is there a jira ticket to track that as part of this epic?

docs/content/how-to/automated-machine-management/aws-spot-instances.md
enxebre: @typeid you might want to have a look at this PR from classic vs hcp sre pov

api/hypershift/v1beta1/aws.go
enxebre: nvm, I was thinking of omitzero. I guess this is a legit case to use omitempty and not omitzero cc @JoelSpeed

enxebre:
thanks! lgtm overall pending addressing jira comments and https://github.com/openshift/hypershift/pull/6707#discussion_r2330224254

api/hypershift/v1beta1/aws.go
JoelSpeed: Historically, spot markets could exceed the on-demand price, is that still the case? 

What is the use case for `0` for the maximum price? Does AWS allow a `0` max price?

api/hypershift/v1beta1/aws.go
```diff
@@ -83,6 +96,12 @@ type PlacementOptions struct {
 	//
 	// +optional
 	CapacityReservation *CapacityReservationOptions `json:"capacityReservation,omitempty"`
+
+	// spotMarketOptions specifies options for using AWS Spot instances.
+	// When specified, instances will be launched as Spot instances with the given configuration.
+	// Mutually exclusive with capacityReservation, and tenancy must be unset or set to "default".
+	// +optional
+	SpotMarketOptions *AWSSpotMarketOptions `json:"spotMarketOptions,omitempty"`
```
JoelSpeed, line 104: This looks like it's copying a very old API which I suspect wouldn't be the way we would recommend implementing this API if we were to go through the API review process today

In particular, it's awkward to use `spotMarketOptions: {}` as a way to enable a spot instance.

Really you need a discriminated union, which would include the capacity reservation options as well.

If you look at the way MAPI implements this, we have one of the `MarketType` options as Spot

The ideal here would have been

```
placement:
  market:
    type: OnDemand | Spot | CapacityBlocks
    capacityBlocks: // Valid only when type is CapacityBlocks
      id: ...
    spot: // Valid only when type is Spot
      maxPrice: ...
```

I'm not sure why MarketType ended up inside the CapacityReservationOptions here, we really should have looked at how this looked in MAPI and predicted this need 🤔

enxebre:
/hold
until there's a path forward for the termination handler implementation and lifecycle

api/hypershift/v1beta1/aws.go
enxebre: we can either keep it as in the PR or mark CapacityReservation.MarketType as deprecated and expose a new one within PlacementOptions.

JoelSpeed: If you're happy with a deprecated field, I think it would make more sense to restructure this slightly. We should be able to validate at admission time that the two fields are  not different if the existing `CapacityReservation.MarketType` is present

api/hypershift/v1beta1/aws.go
gdbranco: "If you specify a maximum price, it must be more than USD $0.001. Specifying a value below USD $0.001 will result in an InvalidParameterValue error message."
https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_SpotMarketOptions.html

api/hypershift/v1beta1/aws.go
gdbranco: Let me know what is the direction to follow on this, I agree having the market type directly on placement makes more sense as well

api/hypershift/v1beta1/aws.go
JoelSpeed: Your pattern presently allows `0` as a valid choice, please update to remove that

enxebre:
/close
in favour of
https://github.com/openshift/hypershift/pull/7625
https://github.com/openshift/hypershift/pull/7567

PR #6708, opened 2025-08-28
NO-JIRA: fix(contrib): Make setup_aks_mi.sh executable

devguyio:

All the scripts under contrib/managed-azure are executable by default except for setup_akis_mi.sh. This makes running setup_all.sh --first-time messy since the permissions error is spit mid-way and the setup continues executing the rest of the scripts, which spits errors since setup_aks_mi.sh never run.

**Checklist**


bryan-cox:

This does not change any code and is only related to some helper dev scripts.

PR #6710, opened 2025-08-28
NO-JIRA: Handle SIGTERM in all of the commands

smrtrfszm:
Kubernetes sends a SIGTERM on termination, so pods can gracefully shut down. A few commands doesn't handle this.

This PR switches the signal handling to use `ctrl.SetupSignalHandler()` everywhere. This function returns a context that is canceled on SIGINT and also on SIGTERM. Most other commands use this.

**Checklist**

## Summary by CodeRabbit

* **Refactor**
  * Unified shutdown handling to use a standard signal-based cancellable context for the process lifecycle.
* **Bug Fixes**
  * More reliable graceful termination, reducing missed interrupts and hang scenarios.
  * Ensures consistent cleanup and exit behavior across environments.

### DIFF
--- a/etcd-defrag/main.go
+++ b/etcd-defrag/main.go
@@ -5,8 +5,6 @@ import (
 	"fmt"
 	"log"
 	"os"
-	"os/signal"
-	"syscall"
 	"time"
 
 	hyperapi "github.com/openshift/hypershift/support/api"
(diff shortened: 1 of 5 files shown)


bryan-cox:
This will fail `verify`. We require conventional commits - https://hypershift.pages.dev/contribute/#prior-to-submitting-a-pull-request.

PR #6721, opened 2025-08-29
Always compress and encode payload in token secret for inplace upgrades

enxebre:
Without compressing we are more likely to hit 1mb CM size limit. We need to follow up with something that scales better

Fixes #

**Checklist**

### DIFF
--- a/ignition-server/controllers/tokensecret_controller.go
+++ b/ignition-server/controllers/tokensecret_controller.go
@@ -307,7 +307,16 @@ func (r *TokenSecretReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	}
 
 	patch := tokenSecret.DeepCopy()
-	patch.Data[TokenSecretPayloadKey] = payload
+
+	// In inplace upgrade, compress and encode payload for inplace upgrader to consume.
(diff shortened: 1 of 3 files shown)


ignition-server/controllers/tokensecret_controller.go
```diff
@@ -307,7 +307,16 @@ func (r *TokenSecretReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	}
 
 	patch := tokenSecret.DeepCopy()
-	patch.Data[TokenSecretPayloadKey] = payload
+
+	// In inplace upgrade, compress and encode payload for inplace upgrader to consume.
```
devguyio, line 311:
```suggestion
	// In in-place upgrade, compress and encode payload for in-place upgrader to consume.
```

I think our spell checker will fail for this

ignition-server/controllers/tokensecret_controller_test.go
```diff
@@ -92,6 +92,8 @@ func TestReconcile(t *testing.T) {
 					Namespace: "test",
 					Annotations: map[string]string{
 						TokenSecretAnnotation: "true",
+						// Use inplace upgrade type to test that the payload is compressed and encoded.
```
devguyio, line 95:
```suggestion
						// Use in-place upgrade type to test that the payload is compressed and encoded.
```

same as above

PR #6726, opened 2025-08-31
OCPBUGS-60936:fix(hccp): remove immutable field from kubeletconfigmap

Tal-or:
When kubeletconfig is changing due to manual user changes, or due to performance profile changes, hccp should catch up and mirror the changes to the hosted cluster.

When setting the kubeletconfig configmap as immutable, the controller failed to update the configmap with the following error: `ConfigMap "<config-map-name>" is invalid: data: Forbidden: field is immutable when immutable is set`.

Setting the ConfigMap as immutable was redundant because in case of undesired change, hccp will reconcile the ConfigMap and update it to its desired state, so it's safe to remove it.

NOTE: It’s still possible for the ConfigMap on the hosted cluster side to be tampered with, but accepting this small risk is a reasonable trade-off to ensure the system functions correctly.

Fixes # OCPBUGS-60936

**Checklist**

## Summary by CodeRabbit

* **New Features**
  * Kubelet configuration updates now propagate to hosted clusters without requiring ConfigMap recreation, simplifying updates and day‑2 operations.

* **Bug Fixes**
  * Removed an immutability restriction on the hosted KubeletConfig mirror so updates can be applied smoothly during cluster lifecycle, reducing operational friction and update failures.

### DIFF
--- a/control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
+++ b/control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
@@ -2451,6 +2451,11 @@ func (r *reconciler) reconcileKubeletConfig(ctx context.Context) error {
 				Namespace: ConfigManagedNamespace,
 			},
 		}
+
+		if err := r.deleteImmutableConfigMapIfNeeded(ctx, log, hostedClusterCM); err != nil {
+			return err
+		}
+
 		if result, err := r.CreateOrUpdate(ctx, r.client, hostedClusterCM, func() error {
 			return mutateKubeletConfig(&cm, hostedClusterCM)
 		}); err != nil {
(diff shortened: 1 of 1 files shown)


jparrill:
Hey @Tal-or, I think this change goes against the design decisions we made regarding how a HostedCluster should propagate changes from the ControlPlane to the DataPlane.
In a normal situation, you need Admin permissions in the Management Cluster to create performance profiles. So why would we expose that option to end users and risk compromising the stability of the ControlPlane?

Would it be possible to keep that restriction and instead propagate performance changes from the ControlPlane side—similar to how we handle MachineConfig patches or tuned profiles?

Tal-or:
Hi @jparrill thank you for your comment.

Let me try to clarify:
> When kubeletconfig is changing due to manual user changes, or due to performance profile changes, hccp should catch up and mirror the changes to the hosted cluster.

The user here meant to be an Admin with permissions in the Management Cluster to create performance profiles or kubeletconfigs, no one else.

So when the admin changes the performance profile in a way that affects the kubeletconfig, or changes the kubeletconfig directly (in a case where no performance profile applied), the hccp controller should update the kubeletconfig configmap and mirrors it to he hosted-cluster.
With the current implementation the configmap is immutable, so the hccp controller cannot update it, which means that the components running on the hosted-cluster are relying on a configmap with a stale data.

jparrill:
Humm reading the code I've seen that ConfigMap contains the same as of their source of truth at the ControlPlane side. I don't believe it could harm us because the HCCO will continue overriding the content of that ConfigMap even if the user modifies the ConfigMap. I will left the approval decision to @csrwng.

jparrill:
/hold
Please @csrwng remove the hold once you decide if this could be approved.

Tal-or:
> Humm reading the code I've seen that ConfigMap contains the same as of their source of truth at the ControlPlane side. I don't believe it could harm us because the HCCO will continue overriding the content of that ConfigMap even if the user modifies the ConfigMap. I will left the approval decision to @csrwng.

Exactly, HCCO has the final call, any change from the end-user on the hosted cluster side will get overridden by it.

Tal-or:
Hi @csrwng
PTAL
Thanks!

yanirq:
this is still waiting for final approval from @csrwng  .  Do we need a revisit here ? cc @Tal-or

Tal-or:
> this is still waiting for final approval from @csrwng . Do we need a revisit here ? cc @Tal-or

It's needed and we're still waiting for approval

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -2395,7 +2395,6 @@ func (r *reconciler) reconcileKubeletConfig(ctx context.Context) error {
 }
 
 func mutateKubeletConfig(controlPlaneConfigMap, hostedClusterConfigMap *corev1.ConfigMap) error {
-	hostedClusterConfigMap.Immutable = ptr.To(true)
```
csrwng, line 2484: The function that calls createOrUpdate with this reconcile function should first get the configmap and if it's immutable, delete it so it can be created without Immutable=true

Tal-or: Just did, thanks!

PR #6730, opened 2025-09-01
CNTRLPLANE-203: feat(sharedingress): add Azure Public IP tagging support for router service

muraee:
Add support for configuring IP tags on Azure Public IP resources associated with the shared ingress router service through environment variable configuration.

- Add `SHARED_INGRESS_AZURE_PIP_IP_TAGS` environment variable constant
- Set `service.beta.kubernetes.io/azure-pip-ip-tags` annotation on router service

Fixes #

**Checklist**

## Summary by CodeRabbit

- **New Features**
  - Added support to tag the Azure Public IP used by Shared Ingress via SHARED_INGRESS_AZURE_PIP_IP_TAGS.
  - Environment value is validated (comma-separated key=value pairs; allowed tag keys enforced); invalid values prevent startup.
  - Service annotations are auto-initialized when missing so tags are applied reliably.
  - Opt-in behavior: if the env var is unset, existing configurations remain unchanged.

### DIFF
--- a/hypershift-operator/controllers/sharedingress/router.go
+++ b/hypershift-operator/controllers/sharedingress/router.go
@@ -22,6 +22,13 @@ const (
 	KASSVCLBPort      = 6443
 	ExternalDNSLBPort = 443
 
+	// AzurePipIpTagsEnvVar is the environment variable that contains the IP tags for the Azure Public IP.
+	// It is used to tag the Azure Public IP associated with the Shared Ingress.
+	// Expected format: comma separated key=value pairs with allowed keys: "FirstPartyUsage" or "RoutingPreference".
+	// Example: "RoutingPreference=Internet" or "FirstPartyUsage=SomeValue,RoutingPreference=Internet".
+	// Both keys and values must be non-empty.
+	AzurePipIpTagsEnvVar = "SHARED_INGRESS_AZURE_PIP_IP_TAGS"
+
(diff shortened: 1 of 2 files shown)


hypershift-operator/controllers/sharedingress/router.go
```diff
@@ -205,6 +210,14 @@ func ReconcileRouterService(svc *corev1.Service) error {
 	for k, v := range hcpRouterLabels() {
 		svc.Labels[k] = v
 	}
+
+	if svc.Annotations == nil {
+		svc.Annotations = map[string]string{}
+	}
+	if tags, ok := os.LookupEnv(AzurePipIpTagsEnvVar); ok {
```
enxebre: can we sanitize the env input and fail earlier?

hypershift-operator/controllers/sharedingress/router.go
```diff
@@ -22,6 +23,10 @@ const (
 	KASSVCLBPort      = 6443
 	ExternalDNSLBPort = 443
 
+	// AzurePipIpTagsEnvVar is the environment variable that contains the IP tags for the Azure Public IP.
+	// It is used to tag the Azure Public IP associated with the Shared Ingress.
```
enxebre, line 26: can we clarify in the docs and set an example with the format expectations e.g. 
comma separated key-value pairs a=b,c=d, for example RoutingPreference=Internet
https://cloud-provider-azure.sigs.k8s.io/topics/loadbalancer/

muraee: done

PR #6738, opened 2025-09-02
NO-JIRA: chore(tests): validate conditions after executing tests

enxebre:
Without this gate the following workflow might happen: A control plane pod like ocm crashes because it has no ability to be execute leader election in the upgrade test with ControlPlaneAvailabilityPolicy = string(hyperv1.HighlyAvailable) We only catch that error via ValidatePublicCluster/ValidateHostedClusterConditions. We only run that via .Before(). So the conflicting payload would pass the upgrade test because that pod failing won't prevent the upgrade from succeed and the test from going green It would only fail when using the conflicting payload as "previous" in the upgrade test

Fixes #

**Checklist**

## Summary by CodeRabbit

- **Tests**
  - Enhanced end-to-end post-run validation to detect whether guest clusters have worker nodes (private clusters assumed to have workers) and include that status in HostedCluster condition checks.
  - Skips guest-node checks for unsupported platforms; if guest cluster node listing fails, the test logs an error and continues.
  - Preserved metrics verification and extended validation timeout to 10 minutes for greater stability.

### DIFF
--- a/test/e2e/util/hypershift_framework.go
+++ b/test/e2e/util/hypershift_framework.go
@@ -28,6 +28,7 @@ import (
 	npmetrics "github.com/openshift/hypershift/hypershift-operator/controllers/nodepool/metrics"
 	karpenterassets "github.com/openshift/hypershift/karpenter-operator/controllers/karpenter/assets"
 	"github.com/openshift/hypershift/support/azureutil"
+	"github.com/openshift/hypershift/support/util"
 
 	configv1 "github.com/openshift/api/config/v1"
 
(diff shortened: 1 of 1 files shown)


test/e2e/util/hypershift_framework.go
```diff
@@ -197,6 +197,8 @@ func (h *hypershiftTest) after(hostedCluster *hyperv1.HostedCluster, platform hy
 
 		ValidateMetrics(t, context.Background(), h.client, hostedCluster, metricsToValidate, true)
 
+		numNodes := opts.NodePoolReplicas * int32(len(opts.AWSPlatform.Zones))
+		ValidateHostedClusterConditions(t, context.Background(), h.client, hostedCluster, numNodes > 0, 10*time.Minute)
```
muraee: we should really not be using `context.Background()` and use the test context `h.ctx`
I know this is already used on all the other tests in `after()`, I am not sure why

enxebre:
@coderabbitai review

enxebre:
@coderabbitai review

test/e2e/util/hypershift_framework.go
```diff
@@ -197,6 +198,17 @@ func (h *hypershiftTest) after(hostedCluster *hyperv1.HostedCluster, platform hy
 
 		ValidateMetrics(t, context.Background(), h.client, hostedCluster, metricsToValidate, true)
 
+		// Private clusters may won't be reachable from the test runner; assume workers exist.
```
bryan-cox:
```suggestion
		// Private clusters may not be reachable from the test runner; assume workers exist.
```

PR #6736, opened 2025-09-02
CNTRLPLANE-1364: feat: enable global pull secret for AWS

jparrill:
## What this PR does / why we need it
- Enable global pull secret for ROSA HCP.
- Enabled E2E for AWS plaform

## Which issue(s) this PR fixes:
- Fixes #[CNTRLPLANE-1364](https://issues.redhat.com/browse/CNTRLPLANE-1364)

## Summary by CodeRabbit

- New Features
  - Enabled global pull secret management on AWS platforms, matching existing support on Azure.

- Tests
  - Expanded end-to-end coverage to validate global pull secret behavior on AWS and Azure; test gating updated so these checks run on AWS/Azure for newer releases.

PR #6745, opened 2025-09-03
OCPBUGS-59649, CNTRLPLANE-1398: feat(globalps): security enhancements on GlobalPullSecret feature

jparrill:
This PR includes several enhancements to the GlobalPullSecret feature:
- Security:
    - **Mount pull secrets as volumes**: DaemonSet reads secrets from mounted files instead of API calls
    - **Privileged mode**: Maintained the Privileged mode on the DS, this does not work only with some capabilities enabled. Also documented appropriately. 
    - **Mount pull secrets as volumes**: DaemonSet reads secrets from mounted files instead of API calls
- Performance: 
    - **Configuration hash tracking**: Add configuration hash to DaemonSet labels to trigger pod recreation
  on content changes
    - **Reduced API footprint**: Removed controller from sync-global-pullsecret process and eliminated API
  calls in DaemonSet
- Precedence Logic:
    - **Conflict resolution**: GlobalPullSecret always preserves original pull secret entries in case of
  registry conflicts
    - **Clear documentation**: Updated documentation to clearly explain precedence behavior with examples
- Functional:
    - **Opt-in deployment strategy**: DaemonSet node selection only on appropriate labeled nodes.

## E2E adjustments

-  Fix node capacity and balancing issues
    - Increase MaxFreeDifferenceRatioPercent to 70% for more permissive balancing
    - Set m5.xlarge instance types to ensure adequate memory capacity
    - Increase MaxNodesTotal from 4 to 6 nodes for larger scale testing
    - TestAutoscalingBalancing expects at least 2+4, 3+3 or 4+2 nodes among the nodes 

Added commit from https://github.com/openshift/hypershift/pull/6736 to allow E2E testing to work as expected

- [OCPBUGS-59649](https://issues.redhat.com/browse/OCPBUGS-59649)
- [CNTRLPLANE-1398](https://issues.redhat.com/browse/CNTRLPLANE-1398)

### DIFF
--- a/test/e2e/autoscaling_test.go
+++ b/test/e2e/autoscaling_test.go
@@ -42,6 +42,11 @@ func TestAutoscaling(t *testing.T) {
 				"custom.ignore.label": "test1",
 			}
 
+			// Set instance type to m5.xlarge for autoscaling tests to increase node capacity
+			if nodepool.Spec.Platform.AWS != nil {
+				nodepool.Spec.Platform.AWS.InstanceType = "m5.xlarge"
+			}
+
 			if additionalNP == nil {
 				additionalNP = &hyperv1.NodePool{
 					ObjectMeta: metav1.ObjectMeta{
@@ -58,6 +63,11 @@ func TestAutoscaling(t *testing.T) {
 				additionalNP.Spec.AutoScaling = &hyperv1.NodePoolAutoScaling{
 					Min: 1,
 					Max: 3,
+				}
+
+				// Also set m5.xlarge for the additional NodePool
+				if additionalNP.Spec.Platform.AWS != nil {
+					additionalNP.Spec.Platform.AWS.InstanceType = "m5.xlarge"
 				}
 			}
 		}
@@ -207,8 +217,8 @@ func testAutoscalingBalancing(ctx context.Context, mgtClient crclient.Client, ho
 				BalancingIgnoredLabels: []string{
 					"custom.ignore.label",
 				},
-				MaxNodesTotal:                 ptr.To[int32](4),
-				MaxFreeDifferenceRatioPercent: ptr.To[int32](50),
+				MaxNodesTotal:                 ptr.To[int32](6),
+				MaxFreeDifferenceRatioPercent: ptr.To[int32](70),
 			}
 			return mgtClient.Update(ctx, hostedCluster)
 		})
@@ -275,8 +285,8 @@ func testAutoscalingBalancing(ctx context.Context, mgtClient crclient.Client, ho
 		// be used, not enough to have more than 1 pod per
 		// node.
 		workloadMemRequest := resource.MustParse(fmt.Sprintf("%v", 0.5*float32(bytes)))
-		expectNodes := numNodes + 2 // default nodepool(min + 1) + additional nodepool(min + 1)
-		workload := newWorkLoad(expectNodes, workloadMemRequest, "", globalOpts.LatestReleaseImage)
+		expectNodes := int32(6) // Target 6 nodes total for balancing test
+		workload := newWorkLoad(6, workloadMemRequest, "", globalOpts.LatestReleaseImage)
 		err = guestClient.Create(ctx, workload)
 		g.Expect(err).NotTo(HaveOccurred())
 		t.Logf("Created workload. Node: %s, memcapacity: %s, workload memory request: %s", nodes[0].Name, memCapacity.String(), workloadMemRequest.String())
@@ -287,7 +297,7 @@ func testAutoscalingBalancing(ctx context.Context, mgtClient crclient.Client, ho
 		_ = e2eutil.WaitForNReadyNodes(t, ctx, guestClient, expectNodes, hostedCluster.Spec.Platform.Type)
 		t.Logf("Successfully reached %d nodes", expectNodes)
 		// Check load balancing between nodepools
-		e2eutil.EventuallyObjects(t, ctx, fmt.Sprintf("both nodepools (%s and %s) to have 2 replicas each", defaultNodePool.Name, additionalNodePool.Name), func(ctx context.Context) ([]*hyperv1.NodePool, error) {
+		e2eutil.EventuallyObjects(t, ctx, fmt.Sprintf("both nodepools (%s and %s) to have reasonable distribution totaling %d nodes", defaultNodePool.Name, additionalNodePool.Name, expectNodes), func(ctx context.Context) ([]*hyperv1.NodePool, error) {
 			nodePools := []*hyperv1.NodePool{defaultNodePool, additionalNodePool}
 			for _, np := range nodePools {
 				if err := mgtClient.Get(ctx, crclient.ObjectKeyFromObject(np), np); err != nil {
@@ -299,9 +309,17 @@ func testAutoscalingBalancing(ctx context.Context, mgtClient crclient.Client, ho
 			if len(nps) != 2 {
 				return false, fmt.Sprintf("expected 2 nodepools, got %d", len(nps)), nil
 			}
-			if nps[0].Status.Replicas != 2 || nps[1].Status.Replicas != 2 {
-				return false, fmt.Sprintf("nodepools replicas are %d and %d, want both 2", nps[0].Status.Replicas, nps[1].Status.Replicas), nil
+			// Relaxing the check to allow reasonable distribution between nodepools, it's not deterministic which nodepool will get the nodes.
+			// This supports 2+4, 3+3, 4+2 configurations (each nodepool must have at least 2 nodes).
+			// With this we make sure no nodepool has ≤1 nodes and resolve flaky tests.
+			totalReplicas := nps[0].Status.Replicas + nps[1].Status.Replicas
+			if totalReplicas != 6 {
+				return false, fmt.Sprintf("total replicas is %d, want 6", totalReplicas), nil
 			}
+			if nps[0].Status.Replicas <= 1 || nps[1].Status.Replicas <= 1 {
+				return false, fmt.Sprintf("unbalanced: nodepool has ≤1 nodes (%d, %d)", nps[0].Status.Replicas, nps[1].Status.Replicas), nil
+			}
+
 			return true, fmt.Sprintf("nodepools balanced - %s: %d, %s: %d", nps[0].Name, nps[0].Status.Replicas, nps[1].Name, nps[1].Status.Replicas), nil
 		}}, nil, e2eutil.WithInterval(30*time.Second), e2eutil.WithTimeout(10*time.Minute))
 	}


muraee:
@coderabbitai review

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
```diff
@@ -260,14 +339,24 @@ func mergePullSecrets(ctx context.Context, originalPullSecret, userProvidedPullS
 	}
 	userProvidedAuths = userProvidedJSON["auths"].(map[string]any)
 
-	// Merge auths
-	for k, v := range userProvidedAuths {
-		originalAuths[k] = v
+	// If managedServices, that means the prcedence of the original pull secret is higher than the user provided pull secret so we need to merge the user provided pull secret into the original pull secret in the other case, the precedence is the opposite
```
enxebre: Why would we want to have different behavior?

jparrill: That's a good question, the input from SD is they need precedence to be in favor of the OCM's pull secret, but there is not details about the user provided one in the additional-pull-secret, so I picked to give the user the precedence when we are not in managed services.

What do you think it's better?

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
```diff
@@ -61,25 +65,56 @@ func (r *Reconciler) reconcileGlobalPullSecret(ctx context.Context) error {
 	log := ctrl.LoggerFrom(ctx)
 	log.Info("reconciling global pull secret")
 
+	// Get the original pull secret once at the beginning (used in both scenarios)
+	originalPullSecret := manifests.PullSecret(r.hcpNamespace)
+	if err := r.cpClient.Get(ctx, crclient.ObjectKeyFromObject(originalPullSecret), originalPullSecret); err != nil {
+		return fmt.Errorf("failed to get original pull secret: %w", err)
+	}
+	originalPullSecretBytes = originalPullSecret.Data[corev1.DockerConfigJsonKey]
+
```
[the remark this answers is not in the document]
jparrill: I don't think this is possible at all, but it does not harm.

sync-global-pullsecret/sync-global-pullsecret.go
```diff
@@ -98,177 +97,123 @@ func NewRunCommand() *cobra.Command {
...
+			return fmt.Errorf("failed to read original pull secret from file: %w", err)
 		}
-		log.Info("Global pull secret not found, using original pull secret")
-		chosenPullSecret = originalPullSecret.DeepCopy()
+		globalPullSecretBytes = originalPullSecretBytes
 	} else {
-		log.Info("Global pull secret found, using it")
-		chosenPullSecret = globalPullSecret.DeepCopy()
+		s.log.Info("Global pull secret content found, using it")
 	}
```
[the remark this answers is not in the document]
jparrill: I bake the GlobalPullSecret based on original + additional, but being this defensive programming, I will include it.

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
jparrill: Ok, after the discussion in slack, I've decided to maintain the precedence to always keep the originalPullSecret entry in conflict case. It's possible to the user to specify in the PullSecret a smaller scope to avoid these kind of conflicts.

support/awsutil/awsutil.go
```diff
@@ -0,0 +1,60 @@
...
+
+// IsROSAHCP returns true if the hosted control plane is a ROSA (Red Hat OpenShift Service on AWS) cluster
+// This is determined by checking for the red-hat-managed tag set to "true"
+func IsROSAHCP(hcp *hyperv1.HostedControlPlane) bool {
+	if hcp.Spec.Platform.AWS == nil {
+		return false
+	}
+
+	return HasResourceTag(hcp, "red-hat-managed", "true")
+}
```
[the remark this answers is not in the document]
jparrill: Not necessary anymore, this code is deleted

LiangquanLi930:
job failed

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
```diff
@@ -260,14 +337,17 @@ func mergePullSecrets(ctx context.Context, originalPullSecret, userProvidedPullS
 	}
 	userProvidedAuths = userProvidedJSON["auths"].(map[string]any)
 
-	// Merge auths
-	for k, v := range userProvidedAuths {
-		originalAuths[k] = v
+	for k, v := range originalAuths {
```
gaol, line 390: Maybe we need to update the doc of `mergePullSecrets` method to reflect the changes here.

gaol, line 390: What I read from https://issues.redhat.com/browse/CNTRLPLANE-1398 is that there will be cases we need to check: `managed` or `self-managed`, is that still the case ?  Maybe we need to update the Jira to reflect the real changes if not ?

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
```diff
@@ -57,29 +59,63 @@ func (r *Reconciler) reconcileGlobalPullSecret(ctx context.Context) error {
 		originalPullSecretBytes     []byte
 		globalPullSecretBytes       []byte
 		err                         error
+		ok                          bool
```
gaol, line 78: It looks the doc above in this method:
```
// If that PS exists, the HCCO deploys a DaemonSet which mounts the whole Root FS of the node, and merges the new PS with the original one.
// If the PS doesn't exist, the HCCO doesn't do anything.
```
should be update  because IMO no matter if there is `additional-pull-secret` or not, there will be a DaemonSet running in the DataPlane nodes

test/e2e/util/util.go
```diff
@@ -1761,9 +1767,8 @@ func EnsureGlobalPullSecret(t *testing.T, ctx context.Context, mgmtClient crclie
 			// Additional Pull Secret
 			additionalPullSecretName            = "additional-pull-secret"
 			additionalPullSecretNamespace       = "kube-system"
-			pullSecretNamespace                 = "openshift-config"
-			additionalPullSecretDummyData       = []byte(`{"auths": {"quay.io": {"auth": "YWRtaW46cGFzc3dvcmQ="}}}`)
-			additionalPullSecretReadOnlyE2EData = []byte(`{"auths": {"quay.io": {"auth": "aHlwZXJzaGlmdCtlMmVfcmVhZG9ubHk6R1U2V0ZDTzVaVkJHVDJPREE1VVAxT0lCOVlNMFg2TlY0UkZCT1lJSjE3TDBWOFpTVlFGVE5BS0daNTNNQVAzRA=="}}}`)
+			additionalPullSecretDummyData       = []byte(`{"auths": {"quay.io/hypershift": {"auth": "YWRtaW46cGFzc3dvcmQ="}}}`)
+			additionalPullSecretReadOnlyE2EData = []byte(`{"auths": {"quay.io/hypershift": {"auth": "aHlwZXJzaGlmdCtlMmVfcmVhZG9ubHk6R1U2V0ZDTzVaVkJHVDJPREE1VVAxT0lCOVlNMFg2TlY0UkZCT1lJSjE3TDBWOFpTVlFGVE5BS0daNTNNQVAzRA=="}}}`)
```
gaol: The job failures are related:
```
  should be able to pull other restricted images
  Expected success, but got an error:
      <*errors.errorString | 0xc003adbb90>: 
      failed to get metadata for restricted image: failed to obtain root manifest for quay.io/hypershift/sleep:1.2.0: unauthorized: access to the requested resource is not authorized
      {
          s: "failed to get metadata for restricted image: failed to obtain root manifest for quay.io/hypershift/sleep:1.2.0: unauthorized: access to the requested resource is not authorized",
      }
  --- FAIL: TestCreateCluster/Main/EnsureGlobalPullSecret (61.42s)
```
somehow it needs the host root to be specified to get manifests so also adding `quay.io` auth to `additionalPullSecretReadOnlyE2EData` can make the failure disappear. but it may expose another bug(not sure).

gaol:
One more comment is that https://github.com/openshift/hypershift/blob/main/docs/content/how-to/common/global-pull-secret.md may also be updated to reflect the changes.

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
jparrill: Yeah, the agreement now is, AWS (managed or self-managed) and ARO.

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
jparrill: I've updated the comment with the current behavior

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
jparrill: 100%, thanks!

test/e2e/util/util.go
jparrill: Fixed. This is an issue in how the underneath dep from vendor does the `keyring.Lookup(value)` basically it looks for a simplest match which ends always on quay.io, but crio does not follow that logic. So I removed part of the test that in the end is verified at HCP level raising a pod up from a restricted image.

jparrill:
> One more comment is that https://github.com/openshift/hypershift/blob/main/docs/content/how-to/common/global-pull-secret.md may also be updated to reflect the changes.

Updated the documentation and enhanced a bit.

jparrill:
/retest

Failing NTO, need to review next week...

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
```diff
@@ -49,37 +51,73 @@ func (r *Reconciler) Reconcile(ctx context.Context, req crreconcile.Request) (cr
 
 // reconcileGlobalPullSecret reconciles the original pull secret given by HCP and merges it with a new pull secret provided by the user.
 // The new pull secret is only stored in the DataPlane side so, it's not exposed in the API. It lives in the kube-system namespace of the DataPlane.
-// If that PS exists, the HCCO deploys a DaemonSet which mounts the whole Root FS of the node, and merges the new PS with the original one.
-// If the PS doesn't exist, the HCCO doesn't do anything.
+// - If that PS is created, the HCCO deploys a DaemonSet which mounts the node's kubeconfig's file, and merges the new PS with the original one.
+// - If the PS doesn't exist, the HCCO doesn't do anything.
+// - If at some point the user deletes the additional pull secret, the daemonSet will not be removed
+// - If the user never added that additional pull secret, the daemonSet will not be deployed in the DataPlane
```
gaol: From what I read in this globalps.go file, no matter the user adds the additional pull secret or not, the `reconcileDaemonSet` will be called which will create a daemonset, right ?

jparrill: Yeah you're right, let me remove that comment.

jparrill:
Looks like `TestNodePool/HostedCluster0/Main/TestNTOMachineConfigAppliedInPlace` and `TestNodePool/HostedCluster0/Main/TestNodePoolInPlaceUpgrade` are consistently failing. Need to take a look

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
```diff
@@ -260,14 +339,17 @@ func mergePullSecrets(ctx context.Context, originalPullSecret, userProvidedPullS
...
-	// Merge auths
-	for k, v := range userProvidedAuths {
-		originalAuths[k] = v
+	for k, v := range originalAuths {
+		if _, ok := userProvidedAuths[k]; ok {
+			log.Info("The registry provided in the additional-pull-secret secret already exists in the original pull secret, this is not allowed. Keeping the original pull secret registry authentication", "registry", k)
+		}
+		userProvidedAuths[k] = v
 	}
+	finalAuths = userProvidedAuths
```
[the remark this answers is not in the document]
jparrill: That's information for the user, to be read not for the developer perse.

jparrill:
/retest

Konflux

jparrill:
/retest

Konflux

csrwng (commented):
**🚨 Breaking Change Alert: DaemonSet Selector**

This selector change from `name` to `app` will cause deployment failures during upgrades since DaemonSet selectors are immutable.

**Recommendation**: 
- Add migration logic to delete the old DaemonSet before creating the new one
- Or use a different DaemonSet name to avoid conflicts  
- Add upgrade documentation

**Risk**: High - Will break existing deployments during upgrade

csrwng (commented):
**⚠️ Precedence Logic Changed**

The merge behavior has been reversed - original pull secrets now win conflicts instead of user-provided ones:

```go
// This overwrites user-provided with original
userProvidedAuths[k] = v  
```

This is a breaking behavioral change that could surprise users who expect their additional pull secrets to override defaults.

**Recommendation**: Document this change prominently in release notes and consider making it configurable.

csrwng (commented):
**🎯 Upgrade Strategy & Testing**

This PR introduces several breaking changes that need careful upgrade planning:

1. **DaemonSet selector change** - Will cause immediate failures
2. **Merge precedence change** - Changes user-visible behavior  
3. **New node labeling logic** - Could affect existing clusters

**Critical Questions**:
- How will existing clusters with the old DaemonSet handle the upgrade?
- What's the rollback strategy if this causes issues?
- Have these changes been tested on existing clusters with GlobalPullSecret already enabled?

**Recommendations**:
- Add integration tests that simulate upgrading from the previous version
- Document the upgrade procedure clearly
- Consider feature flags or gradual rollout for the behavioral changes

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
```diff
@@ -49,37 +53,83 @@ func (r *Reconciler) Reconcile(ctx context.Context, req crreconcile.Request) (cr
...
+// - If the PS doesn't exist, the HCCO doesn't do anything.
+// - If at some point the user deletes the additional pull secret, the daemonSet will not be removed
 // If the PS doesn't exist, the HCCO doesn't do anything.
+//
+// IMPORTANT: The DaemonSet is NOT deployed to nodes that belong to NodePools using InPlace upgrade strategy.
+// This prevents conflicts between the DaemonSet's kubelet config modifications and Machine Config Daemon operations during InPlace upgrades.
 func (r *Reconciler) reconcileGlobalPullSecret(ctx context.Context) error {
 	var (
 		userProvidedPullSecretBytes []byte
 		originalPullSecretBytes     []byte
```
csrwng, line 75: 🚨 **Breaking Change Alert: DaemonSet Selector**

This selector change from `name` to `app` will cause deployment failures during upgrades since DaemonSet selectors are immutable.

**Recommendation**: 
- Add migration logic to delete the old DaemonSet before creating the new one
- Or use a different DaemonSet name to avoid conflicts  

**Risk**: High - Will break existing deployments during upgrade

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
```diff
@@ -113,37 +165,128 @@ func (r *Reconciler) reconcileGlobalPullSecret(ctx context.Context) error {
...
 
-func reconcileDaemonSet(ctx context.Context, daemonSet *appsv1.DaemonSet, globalPullSecretName string, c crclient.Client, createOrUpdate upsert.CreateOrUpdateFN, hccoImage string) error {
+// labelNodesFromInPlaceNodePools labels nodes that belong to NodePools using InPlace upgrade strategy
+func (r *Reconciler) labelNodesFromInPlaceNodePools(ctx context.Context) error {
+	log := ctrl.LoggerFrom(ctx)
+
+	// Get all nodes from the hosted cluster
+	nodeList := &corev1.NodeList{}
+	if err := r.hcUncachedClient.List(ctx, nodeList); err != nil {
+		return fmt.Errorf("failed to list nodes: %w", err)
```
csrwng, line 198: ⚠️ **Precedence Logic Changed**

The merge behavior has been reversed - original pull secrets now win conflicts instead of user-provided ones:

```go
for k, v := range originalAuths {
    if _, ok := userProvidedAuths[k]; ok {
        log.Info("...Keeping the original pull secret registry authentication", "registry", k)
    }
    userProvidedAuths[k] = v  // Original overwrites user-provided
}
```

This is a breaking behavioral change that could surprise users who expect their additional pull secrets to override defaults.

**Recommendation**: Document this change prominently in release notes and consider making it configurable.

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
```diff
@@ -113,37 +165,128 @@ func (r *Reconciler) reconcileGlobalPullSecret(ctx context.Context) error {
...
+	// Create set of nodes that belong to InPlace NodePools
+	nodesFromInPlaceNodePools := make(map[string]bool)
+	for _, ms := range machineSetList.Items {
+		// Check if this MachineSet belongs to a NodePool with InPlace strategy
+		// This can be identified by the presence of InPlace-specific annotations
+		if _, hasTargetConfig := ms.Annotations["hypershift.openshift.io/nodePoolTargetConfigVersion"]; hasTargetConfig {
+			// Get Machines from this MachineSet
+			machines := &capiv1.MachineList{}
+			if err := r.cpClient.List(ctx, machines, &crclient.ListOptions{
+				Namespace:     ms.Namespace,
```
csrwng: 🔍 **Node Labeling Logic Fragility**

This function relies on annotation detection that could be fragile:

```go
if _, hasTargetConfig := ms.Annotations["hypershift.openshift.io/nodePoolTargetConfigVersion"]; hasTargetConfig {
    // Assumes this annotation identifies InPlace strategy
}
```

**Concerns**:
- Fragile dependency on annotation naming conventions
- No validation that this annotation correctly identifies InPlace strategy
- Silent failures when node labeling fails could lead to DaemonSet being deployed to wrong nodes

**Recommendation**:
- Add more robust NodePool strategy detection (check actual NodePool objects)
- Add validation and better error handling 
- Consider using explicit NodePool strategy identification rather than annotation inference

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
```diff
@@ -154,21 +297,35 @@ func reconcileDaemonSet(ctx context.Context, daemonSet *appsv1.DaemonSet, global
...
+									{
+										Name:      "original-pull-secret",
+										MountPath: "/etc/original-pull-secret",
+										ReadOnly:  true,
+									},
+								}
+								if globalPullSecretName != "" {
+									volumeMounts = append(volumeMounts, corev1.VolumeMount{
+										Name:      "global-pull-secret",
+										MountPath: "/etc/global-pull-secret",
```
csrwng: 📝 **Complex Conditional Logic**

This volume mounting logic is becoming quite complex and hard to maintain:

```go
VolumeMounts: func() []corev1.VolumeMount {
    volumeMounts := []corev1.VolumeMount{...}
    if globalPullSecretName != "" {
        volumeMounts = append(volumeMounts, ...)
    }
    return volumeMounts
}(),
```

**Issues**:
- Anonymous function makes the code harder to read
- Conditional logic embedded in struct initialization
- Similar pattern repeated for Volumes below
- Difficult to unit test in isolation

**Recommendation**: 
- Extract this to helper functions like `buildVolumeMounts(globalPullSecretName, originalPullSecretName)`
- Same for `buildVolumes()` 
- This will make the code more readable, testable, and maintainable

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
```diff
@@ -88,21 +138,23 @@ func (r *Reconciler) reconcileGlobalPullSecret(ctx context.Context) error {
...
 
-	// Get the original pull secret
-	originalPullSecret := manifests.PullSecret(r.hcpNamespace)
-	if err := r.cpClient.Get(ctx, crclient.ObjectKeyFromObject(originalPullSecret), originalPullSecret); err != nil {
-		return fmt.Errorf("failed to get original pull secret: %w", err)
-	}
-
-	// Asumming hcp pull secret is valid
-	originalPullSecretBytes = originalPullSecret.Data[corev1.DockerConfigJsonKey]
-
```
csrwng, line 99: ⚠️ **Error Handling Inconsistency**

Silent failures in node labeling could lead to inconsistent cluster state:

```go
if err := r.labelNodesFromInPlaceNodePools(ctx); err != nil {
    // Do not fail the reconciliation, just log the error
    log.Error(err, "failed to label nodes from InPlace NodePools")
}
```

**Critical Concern**: If node labeling fails, the DaemonSet might be deployed to nodes it shouldn't be, potentially causing conflicts with Machine Config Daemon during InPlace upgrades.

**This defeats the entire purpose of the InPlace NodePool avoidance logic!**

**Recommendation**: 
- Consider if this should be a hard failure to ensure cluster consistency
- Or add retry logic with exponential backoff and eventually fail if it keeps failing
- At minimum, add more visibility (metrics/events) when this fails
- The current approach silently compromises the safety mechanism

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
```diff
@@ -154,21 +297,35 @@ func reconcileDaemonSet(ctx context.Context, daemonSet *appsv1.DaemonSet, global
...
+							VolumeMounts: func() []corev1.VolumeMount {
+								volumeMounts := []corev1.VolumeMount{
+									{
+										Name:      "kubelet-config",
+										MountPath: "/var/lib/kubelet",
+									},
+									{
+										Name:      "dbus",
+										MountPath: "/var/run/dbus",
+									},
```
csrwng: ❓ **Privileged Mode Justification Needed**

The comment mentions that privileged mode is required because capabilities don't work:

```go
SecurityContext: &corev1.SecurityContext{
    Privileged: ptr.To(true),
},
```

**Questions**:
- What specific capabilities are needed that require privileged mode?
- Can you provide more details on why capabilities-based approach doesn't work?
- Have you tested with specific capabilities like `CAP_SYS_ADMIN`, `CAP_CHOWN`, `CAP_DAC_OVERRIDE`?
- What specific operations require root access to the host?

**Security Impact**: Privileged mode gives the container root access to the host, which is a significant security risk. The PR description mentions security enhancements, but this contradicts that goal.

**Recommendation**: 
- Document the specific privileged operations needed
- Test with minimal required capabilities instead of full privileged access
- If privileged mode is truly required, provide detailed justification in code comments

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
```diff
@@ -113,37 +165,128 @@ func (r *Reconciler) reconcileGlobalPullSecret(ctx context.Context) error {
...
 					Tolerations:                  []corev1.Toleration{{Operator: corev1.TolerationOpExists}},
+					// Use affinity to exclude nodes that have the InPlace NodePool label
+					Affinity: &corev1.Affinity{
+						NodeAffinity: &corev1.NodeAffinity{
+							RequiredDuringSchedulingIgnoredDuringExecution: &corev1.NodeSelector{
+								NodeSelectorTerms: []corev1.NodeSelectorTerm{
+									{
+										MatchExpressions: []corev1.NodeSelectorRequirement{
+											{
+												Key:      "hypershift.openshift.io/nodepool-inplace-strategy",
```
csrwng: 🚨 **Critical Behavioral Change in mergePullSecrets**

The precedence logic has been completely reversed in this function:

```go
for k, v := range originalAuths {
    if _, ok := userProvidedAuths[k]; ok {
        log.Info("The registry provided in the additional-pull-secret secret already exists in the original pull secret, this is not allowed. Keeping the original pull secret registry authentication", "registry", k)
    }
    userProvidedAuths[k] = v  // OVERWRITES user-provided with original!
}
finalAuths = userProvidedAuths
```

**Previously**: User-provided pull secrets would override original ones
**Now**: Original pull secrets always win conflicts

**Breaking Change Impact**:
- Users expecting their additional pull secrets to override defaults will be surprised
- Existing clusters with working additional pull secrets may suddenly fail to authenticate to registries
- This is a **major behavioral change** that needs prominent documentation

**Test Case Impact**: The test case was updated to expect this new behavior, confirming this is intentional but should be clearly communicated.

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
jparrill: Does not make sense to be configurable, the precedence should follow the rules that are set now. Original HCP PullSecret will always have precedence over user PullSecret. The comment about the release note, makes sense.

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
jparrill: Makes sense to reverse the logic, only labeled nodes will receive the DS. This way by default the DS will not be added and could not affect negatively the cluster accidentally. Added

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
jparrill: Makes sense to force the globalps controller to fail if the annotation should be added and for w/e reason the controller. Added

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
jparrill: The daemonSet needs RW permissions over the `/var/lib/kubelet/config.json` in order to work properly and there is not any CAP that allows that to work without the privileged mode enabled. I can add a note in the code for that.

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
jparrill: This is a requirement from ROSA and yes it impacts the testing. Unit + E2E were already modified in that sense and QE are already aware of it. If there is not a comment already in the code, I will add it too

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
jparrill: Reverting back to `name`

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
jparrill: Done

jparrill:
Hey @csrwng I've addressed all the comments. PTAL when you have some bandwidth

jparrill:
/retest

Infra again....

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
```diff
@@ -8,28 +8,39 @@ import (
...
+type GlobalPullSecretPodConfig struct {
+	VolumeMounts []corev1.VolumeMount
+	Volumes      []corev1.Volume
+}
+
 type Reconciler struct {
 	cpClient               crclient.Client
 	kubeSystemSecretClient crclient.Client
 	hcUncachedClient       crclient.Client
+	hcpName                string
```
csrwng: This is not needed because it is used to retrieve a hcp that is not used (see comment below)

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
```diff
@@ -49,37 +60,91 @@ func (r *Reconciler) Reconcile(ctx context.Context, req crreconcile.Request) (cr
...
+		return fmt.Errorf("original pull secret does not contain %s key", corev1.DockerConfigJsonKey)
+	}
+
 	// Get the user provided pull secret
 	exists, additionalPullSecret, err := additionalPullSecretExists(ctx, r.kubeSystemSecretClient)
 	if err != nil {
 		return fmt.Errorf("failed to check if user provided pull secret exists: %w", err)
 	}
 
+	hcp := manifests.HostedControlPlane(r.hcpNamespace, r.hcpName)
```
csrwng: This hcp is not used anywhere

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
```diff
@@ -134,19 +274,33 @@ func reconcileDaemonSet(ctx context.Context, daemonSet *appsv1.DaemonSet, global
...
 				},
 				Spec: corev1.PodSpec{
-					ServiceAccountName:           manifests.GlobalPullSecretDSName,
-					AutomountServiceAccountToken: ptr.To(true),
+					AutomountServiceAccountToken: ptr.To(false),
 					SecurityContext:              &corev1.PodSecurityContext{},
 					DNSPolicy:                    corev1.DNSDefault,
 					Tolerations:                  []corev1.Toleration{{Operator: corev1.TolerationOpExists}},
+					// Use affinity to only include nodes that are explicitly enabled for GlobalPullSecret
+					Affinity: &corev1.Affinity{
```
csrwng: A simple nodeSelector would be better than the affinity rule.

csrwng (commented):
Some comments. None blocking,  can be handled in a follow up.

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
```diff
@@ -49,37 +60,91 @@ func (r *Reconciler) Reconcile(ctx context.Context, req crreconcile.Request) (cr
...
 		return nil
 	}
 
-	// Reconcile the RBAC for the Global Pull Secret
-	if err := reconcileGlobalPullSecretRBAC(ctx, r.hcUncachedClient, r.CreateOrUpdate, "kube-system", "openshift-config"); err != nil {
-		return fmt.Errorf("failed to reconcile global pull secret RBAC: %w", err)
+	// Label nodes that should have GlobalPullSecret DaemonSet (non-InPlace nodes)
+	// This is critical to prevent DaemonSet deployment conflicts with Machine Config Daemon
+	log.Info("labeling nodes eligible for GlobalPullSecret DaemonSet")
+	if err := r.labelNodesForGlobalPullSecret(ctx); err != nil {
```
gaol, line 142: I did a round of pre-merge test and find an issue on the node labeling for the non-InPlace nodes:

Let's say we are going to add a new pull secret item:
```
    "test-additional-ps-inplace.com": {
      "auth": "dGVzdC11c2VyLWFkZGl0aW9uYWwtcHM6dGVzdC1wYXNzLWFkZGl0aW9uYWwtcHM="
    }
```

- Step 1: Create an additional-pull-secret with `test-additional-ps-inplace.com` and check all nodes should have the new pull secret item added, all good.
- Step 2: Create a new nodepool with InPlace upgrade strategy
```
hypershift create nodepool aws --name test-01 --namespace clusters --cluster-name my-hc-name  --replicas 1 --node-upgrade-type InPlace
```
- Step 3: Check the new nodes that should *NOT* have the `test-additional-ps-inplace.com` item. all good.
- Step 4: Create another nodepool with `RePlace` upgrade:
```
hypershift create nodepool aws --name test-02 --namespace clusters --cluster-name my-hc-name  --replicas 1 --node-upgrade-type RePlace
```
- Step 5: The new nodes with `RePlace` upgrade strategy should have the new `test-additional-ps-inplace.com` in `/var/lib/kubelet/config.json`, but it does not !

Now the `global-pull-secret` has the item of `test-additional-ps-inplace.com`, but the config.json file in the ndoes are not updated. I can see the damenset desired count does not increase either. The nodes with `RePlace` upgrade strategy does not have the label: `"hypershift.openshift.io/nodepool-globalps-enabled": "true"`.

NOTE: Delete the `global-pull-secret` can trigger the reconcile and make the config.json file updated in node.

So, maybe we should move the labeling action to an earlier stage so that as long as the node gets provisioned, we label them if necessary.

@jparrill @csrwng

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
jparrill: Addressed

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
jparrill: Addressed

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
jparrill: Addressed

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
jparrill: Nice catch @gaol, I've push a change that fixes that behavio, tested in my dev cluster and also covered with a unit test. This is the status after applying the patch:

image
image

I've used this image for the validation (which contains all the fixes, also the ones that Cesar were asking for): quay.io/jparrill/hypershift:OCPBUGS-59649-v66

jparrill:
/test e2e-aws

Infra...

jparrill:
/test e2e-aks 

KAS didn't show up

jparrill:
/retest

infra again....

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
gaol: I did another round of pre-test, and it works good. so this can be resolved.   :)

jparrill:
/retest

non related tests failed

jparrill:
/hold cancel

I'm tired of this CI...

jparrill:
/test e2e-aws

Autoscaling failing again...

jparrill:
Hey @gaol The only change included is the size of the nodes in the E2E + the client used to list the nodes in the GlobalPS controller. I've verified the right functioning already, but if you want to take a look again is fine.

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
```diff
@@ -8,30 +8,42 @@ import (
...
+// GlobalPullSecretPodConfig encapsulates the configuration for GlobalPullSecret DaemonSet pods
+type GlobalPullSecretPodConfig struct {
+	VolumeMounts []corev1.VolumeMount
+	Volumes      []corev1.Volume
+}
+
 type Reconciler struct {
 	cpClient               crclient.Client
 	kubeSystemSecretClient crclient.Client
+	nodeClient             crclient.Client
```
gaol, line 42: @jparrill I actually don't get it where this `nodeClient` is used  ?

jparrill: Sometimes Claude trolls me a lot..., thanks for the heads up, updating the PR

jparrill:
Last changes:
- Use nodeClient for node list
- Fix test
- Adjust the AutoscaleBalancing percentage to 70% (current percentage on 65-67% in latests E2E runs)

jparrill:
/test e2e-aks
non related test failed (NTO)

PR #6746, opened 2025-09-03
[CNTRLPLANE-408] add imagetype to nodepool api

BraeTroutman:

Add support for specifying an ImageType of `windows` when creating a nodepool. This allows for a custom mapping of such Windows nodepools to special AMIs for Windows License tracking and billing purposes.

Fixes #CNTRLPLANE-408

**Checklist**


hypershift-operator/controllers/nodepool/aws.go
```diff
@@ -317,3 +347,73 @@ func (r NodePoolReconciler) validateAWSPlatformConfig(ctx context.Context, nodeP
 
 	return nil
 }
+
+// windowsAMIMapping contains hard-coded Windows AMI mappings by region and OpenShift version.
+// This is a simplified mapping for demonstration purposes.
+var windowsAMIMapping = map[string]map[string]string{
```
BraeTroutman: Not that I'll need to change this to a specific mapping of regions + versions to AMIs, or have this lifted into some sort of external configuration that gets passed into the operator

test/e2e/nodepool_imagetype_test.go
```diff
@@ -0,0 +1,208 @@
...
+	nodePool.Spec.Replicas = &oneReplicas
+	nodePool.Spec.Management.Replace = &hyperv1.ReplaceUpgrade{
+		Strategy: hyperv1.UpgradeStrategyRollingUpdate,
+		RollingUpdate: &hyperv1.RollingUpdate{
+			MaxUnavailable: func(v intstr.IntOrString) *intstr.IntOrString { return &v }(intstr.FromInt(0)),
+			MaxSurge:       func(v intstr.IntOrString) *intstr.IntOrString { return &v }(intstr.FromInt(int(oneReplicas))),
+		},
+	}
+	nodePool.Spec.Platform.AWS.InstanceType = "m5.metal"
+
```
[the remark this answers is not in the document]
BraeTroutman: instance type must be baremetal in order to work with Windows LI

hypershift-operator/controllers/nodepool/aws.go
```diff
@@ -256,7 +264,28 @@ func (r *NodePoolReconciler) setAWSConditions(ctx context.Context, nodePool *hyp
...
+				return fmt.Errorf("couldn't discover a Windows AMI for release image: %w", err)
+			}
+			SetStatusCondition(&nodePool.Status.Conditions, hyperv1.NodePoolCondition{
+				Type:               hyperv1.NodePoolValidPlatformImageType,
+				Status:             corev1.ConditionTrue,
+				Reason:             hyperv1.AsExpectedReason,
+				Message:            fmt.Sprintf("Bootstrap Windows AMI is %q", ami),
+				ObservedGeneration: nodePool.Generation,
+			})
 		} else {
```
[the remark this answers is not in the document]
BraeTroutman: This is enforced at the validation level now, per handling of [this comment](https://github.com/openshift/hypershift/pull/6746#discussion_r2350150589)

api/hypershift/v1beta1/nodepool_types.go
```diff
@@ -218,6 +233,15 @@ type NodePoolSpec struct {
...
 	Arch string `json:"arch,omitempty"`
+
+	// imageType specifies the type of image to use for node instances.
+	// Valid values are ImageTypeDefault (empty string/default) or ImageTypeWindows ("windows").
+	// When set to ImageTypeWindows, the controller will automatically populate the AMI field
+	// with a Windows-compatible AMI based on the region and OpenShift version.
+	//
+	// +optional
+	// +kubebuilder:validation:Enum="";windows
+	ImageType ImageType `json:"imageType,omitempty"`
```
bryan-cox: This is meant to be an AWS platform specific thing? If so, it should probably go under AWSNodePoolPlatform.

CLAUDE.md
```diff
@@ -1 +1 @@
-AGENTS.md
\ No newline at end of file
+AGENTS.md
```
bryan-cox: Doesn't seem like this change should be here?

hypershift-operator/controllers/nodepool/aws.go
```diff
@@ -317,3 +345,38 @@ func (r NodePoolReconciler) validateAWSPlatformConfig(ctx context.Context, nodeP
...
+func getWindowsAMI(region string, releaseImage *releaseinfo.ReleaseImage) (string, error) {
+	if releaseImage == nil {
+		return "", fmt.Errorf("release image is nil")
+	}
+
+	if releaseImage.StreamMetadata == nil {
+		return "", fmt.Errorf("release image stream metadata is nil")
+	}
+
+	// Default to x86_64 architecture if not specified
```
bryan-cox: This comment doesn't seem correct with the code implementation.

api/hypershift/v1beta1/nodepool_types.go
```diff
@@ -27,6 +27,20 @@ const (
 	IgnitionServerTokenExpirationTimestampAnnotation = "hypershift.openshift.io/ignition-token-expiration-timestamp"
 )
 
+// ImageType specifies the type of image to use for node instances.
+type ImageType string
```
bryan-cox, line 31: This seems like it should be a union discriminator with AMI?

api/hypershift/v1beta1/nodepool_types.go
BraeTroutman: In the DDR spec the field is specified as a broader property of the general NodePool type, but I think it makes sense to put it in the AWS Platform spec, as right now it will only be supported in AWS. I'll make that move.

CLAUDE.md
BraeTroutman: My editor automatically adds newlines to files I've opened-- I'll remove this from my git commit, thanks for the catch

api/hypershift/v1beta1/nodepool_types.go
BraeTroutman: to clarify: do you mean kindof like changing the AMI field to be a new substructure that contains like the AMI itself (if specified) along with the ImageType (i.e. smth like this:
```
struct AMISpec {
   AMI string
   ImageType ImageType
}
```
or am I misunderstanding? In this situation I'm also conforming to the spec as specified in the initial [DDR](https://docs.google.com/document/d/1-sHnMTFh6SbdDhwhw2peuOVghgT4WUNO6Cex_s7VQJc/edit?usp=sharing), but looking at it now I see that Hypershift was not initially consulted in the impl of the nodepool spec changes, so I think we can adapt to make sure we're conforming to your best practices.

BraeTroutman:
with current state of e2e and comments looks like I have a bit more work to do-- thank you @bryan-cox for your initial review, I'll cycle my requests once I've acknowledged all the issues

api/hypershift/v1beta1/nodepool_types.go
bryan-cox: Yeah that's kind of the same line of thinking. Here's an example from Azure - https://github.com/openshift/hypershift/blob/a1d7e10cf90049ee95cb21c8fd4effeb19a25462/api/hypershift/v1beta1/azure.go#L106-L107

bryan-cox: We can't break existing customer's without some deprecation plan as well so maybe we can't make it a union discriminator. It just seemed like that is how it was being used in the PR. cc @muraee

bryan-cox:
@BraeTroutman you should probably move this PR to draft while you're working on it. That way it doesn't retrigger CI tests on every push.

api/hypershift/v1beta1/nodepool_types.go
BraeTroutman: For compatibility I'd like to keep it as a new field rather than changing the type of an existing one-- but I will still make the move into the AWSNodePool subobject

bryan-cox: We could add an AMI field there and deprecate the old one maybe 🤷🏻‍♂️ 

@JoelSpeed do you have any thoughts here?

muraee: imo I don't see a need to warrant going through the hassle of deprecating fields. We can add a CEL validation on the top  level struct to validate both fields are not set if that's the intended behavior

api/hypershift/v1beta1/aws.go
```diff
@@ -25,6 +25,15 @@ type AWSNodePoolPlatform struct {
...
 	AMI string `json:"ami,omitempty"`
 
+	// imageType specifies the type of image to use for node instances.
+	// Valid values are ImageTypeDefault (empty string/default) or ImageTypeWindows ("windows").
+	// When set to ImageTypeWindows, the controller will automatically populate the AMI field
+	// with a Windows-compatible AMI based on the region and OpenShift version.
+	//
+	// +optional
+	// +kubebuilder:validation:Enum="";windows
+	ImageType ImageType `json:"imageType,omitempty"`
```
JoelSpeed, line 37: Why is the empty string a choice here? You have `omitempty` and the field isn't a pointer, so currently, `""` wouldn't round trip through a structured client

Having `""` as an enum choice is not generally advised. What does `Default` actually mean in this instance anyway?

api/hypershift/v1beta1/nodepool_types.go
```diff
@@ -86,6 +100,7 @@ type NodePool struct {
 // +kubebuilder:validation:XValidation:rule="!has(oldSelf.arch) || has(self.arch)", message="Arch is required once set"
 // +kubebuilder:validation:XValidation:rule="self.arch != 'arm64' || has(self.platform.aws) || has(self.platform.azure) || has(self.platform.agent) || self.platform.type == 'None'", message="Setting Arch to arm64 is only supported for AWS, Azure, Agent and None"
 // +kubebuilder:validation:XValidation:rule="!has(self.replicas) || !has(self.autoScaling)", message="Both replicas or autoScaling should not be set"
+// +kubebuilder:validation:XValidation:rule="!has(self.platform.aws) || !has(self.platform.aws.imageType) || self.platform.aws.imageType != 'windows' || (self.arch == 'amd64' && self.platform.type == 'AWS')", message="ImageType 'windows' is only supported with arch 'amd64' and platform type 'AWS'"
```
JoelSpeed: For 4.16 and above APIs, you can simplify this. Does HCP have a minimum supported API server greater than 4.14 yet?
```suggestion
// +kubebuilder:validation:XValidation:rule="self.?platform.aws.imageType.orValue('') != 'windows' || (self.arch == 'amd64' && self.platform.type == 'AWS')", message="ImageType 'windows' is only supported with arch 'amd64' and platform type 'AWS'"
```

api/hypershift/v1beta1/nodepool_types.go
JoelSpeed, line 31: Is this really an Image type, or is it the OS? Looks like the supported values are Linux and Windows?

api/hypershift/v1beta1/aws.go
```diff
@@ -25,6 +25,15 @@ type AWSNodePoolPlatform struct {
 	// +kubebuilder:validation:MaxLength=255
 	AMI string `json:"ami,omitempty"`
 
+	// imageType specifies the type of image to use for node instances.
+	// Valid values are ImageTypeDefault (empty string/default) or ImageTypeWindows ("windows").
+	// When set to ImageTypeWindows, the controller will automatically populate the AMI field
+	// with a Windows-compatible AMI based on the region and OpenShift version.
+	//
+	// +optional
+	// +kubebuilder:validation:Enum="";windows
```
JoelSpeed: Enum values should be PascalCase

api/hypershift/v1beta1/nodepool_types.go
JoelSpeed: Surely `self.platform.aws` is not allowed unless `self.platform.type` is AWS, therefore you don't need to include that in this validation?

api/hypershift/v1beta1/nodepool_types.go
JoelSpeed: What is the actual relationship between this new ImageType field and the AMI? Can you specify both simultaneously?

api/hypershift/v1beta1/aws.go
BraeTroutman: In this situation a default ImageType (i.e. no specified ImageType) refers to the current (before this MR) default behavior of allowing Hypershift to select the AMI from the top-level releaseImage structure: [see defaultNodePoolAMI](https://github.com/openshift/hypershift/blob/main/hypershift-operator/controllers/nodepool/nodepool_controller.go#L698-L712)

When the ImageType is specified as `windows` (or `Windows` once I ack some of the other comments) breaks out of the default behavior to move the AMI selection logic to a specific set of AMIs encoded in the releaseImage metadata. 

In this particular case, is the main issue that having the empty string be a valid value along with omitempty? If that's the case, I can totally remove the empty string case, and leave the omit empty, with the idea of the field being omitted capturing the "default behavior" semantics

api/hypershift/v1beta1/nodepool_types.go
BraeTroutman: I believe for ROSA HCP 4.14 is deprecated now, so at least for our use case yeah I think we could make that assumption-- I know that ROSA/ARO are not the only ones who consume hypershift though right, so is there a possibility making that assumption could break someone's use case?

api/hypershift/v1beta1/nodepool_types.go
BraeTroutman: Right now it does map to an OS, although the specific values right now are more like "Windows" and "everything else". 

The name `ImageType` is intended to capture the fact that this is specifically controlling the amazon machine images used for creating these nodes, which may not be immediately clear (imho) if the field were named something more like "NodeOS" or something along those lines

api/hypershift/v1beta1/nodepool_types.go
BraeTroutman: 100% thank you for that catch, I was not paying good attention to the validations already covered by earlier rules-- I'll fix that.

api/hypershift/v1beta1/nodepool_types.go
BraeTroutman: this is a really good point Joel, as I've implemented it, the behavior of the AMI field overrides the ImageType field when specified.

That is to say, when the AMI field is populated, it doesn't matter what ImageType is, the specified AMI will be passed down to the AWSMachineSpec

Only when ImageType is "Windows" and the AMI is empty will we break into my newly added Windows AMI selection logic

api/hypershift/v1beta1/aws.go
JoelSpeed: So you probably want three states here:
* No opinion - The value is omitted and the platform is left to choose a default
* Windows
* Linux ? - User makes an explicit choice that a Linux AMI should be used

api/hypershift/v1beta1/nodepool_types.go
JoelSpeed: That's a question I was trying to work out with @enxebre before he went off on leave. At some point, we need to start declaring that HyperShift APIs have a supported range of releases. We currently support these APIs on every KAS from 4.14 upwards, which makes it hard to leverage new features. If we started enforcing a version skew, we could start leveraging the newer features.

@sjenning @csrwng I don't know if Alberto discussed this with either of you?

api/hypershift/v1beta1/nodepool_types.go
JoelSpeed: I'm assuming this also links to some field on the EC2 instance? Or is it completely internal to HCP? 

Can you link to the AWS field this relates to if one exists? Curious to see what they describe/how they model this

api/hypershift/v1beta1/aws.go
BraeTroutman: In this situation, both the 1st (no opinion) and 3rd (Linux ?) cases result in the exact same behavior in AMI resolution-- Windows AMIs will only be used when `Windows` enum val is specified

api/hypershift/v1beta1/nodepool_types.go
JoelSpeed: Ok, so these are siblings, and are mutually exclusive. Yes, I would agree that a discriminated union would have been the right way to model this

```
imageSelection:
  type: AMI | ???
  ami:
    id: ...
  ???:
    type: Linux | Windows
```

However, we already have AMI shipped (as a string, it probably shouldn't have been a string but that's another story). I'd be tempted to probably put a CEL validation making the two fields mutually exclusive and move on (at least one of them is required right?)
```
has(self.ami) != has(self.imageType)
```

api/hypershift/v1beta1/aws.go
JoelSpeed: > In this situation, both the 1st (no opinion) and 3rd (Linux ?) cases result in the exact same behavior in AMI resolution

That is deliberate :) We allow users to explicitly choose the default behaviour, that way, if the default were to change in the future, they would not be affected as they had explicitly chosen to keep it to the old behaviour

BraeTroutman: Perfect, that makes sense to me. Thank you Joel, I'll make that change along with the other comments you've mentioned here in a new commit. I'll squash everything once we're happy with the final results

api/hypershift/v1beta1/nodepool_types.go
BraeTroutman: Right now this field is really only internal to HCP, used to dispatch the AMI used for the underlying EC2 instance (that would be the instance field impacted here) 

The whole reason for this new field is that, in order to bill customers correctly for Windows License Included nodes, custom AMIs with special billing labels included must be used.

JoelSpeed: HCP is supported beyond ROSA right, is this just for ROSA?

What would the behaviour of this field be on a non ROSA cluster?

Is the theory that we automatically populate a recent AMI image when this field is set? Is there benefit to auto populating the linux AMIs too?

BraeTroutman: tldr; regardless of ROSA HCP or not, I would expect a nodepool with ImageType specified to dispatch to a Windows LI AMI if available

in the (trivial) case where the client is making use of a cloud provider besides AWS: the ImageType field would not be available because the AWSNodePoolPlatformSpec is not being used

in the other case where the client is making use of the AWS cloud provider for the nodepool, but the client is not pairing it with a ROSA HCP cluster, I (the client) would still expect this field to dispatch to a Windows LI AMI.

The auto-populating of an AMI is already the case for linux/default node types, the AMIs are drawn from the `release-manifests/0000_50_installer_coreos-bootimages.yaml` file in the ocp-release image used for a specific nodepool. The nodepool controller sources this file and finds the appropriate AMI for the specified OCP version and image (you can try pulling any of the specific release images from [ocp-release](https://quay.io/repository/openshift-release-dev/ocp-release?tab=tags) to see what I mean. The ImageType field causes a change in this logic when set to `Windows`, so rather than populating the AMI with the value from the release image main content, it dispatches to the metadata to look for the windows AMI corresponding to the same region and version. This metadata is not contingent on whether the nodepool is paired with a ROSA HCP cluster or not-- it is a property of the OCP release image used.

JoelSpeed: Ack, thanks for the explanation, makes sense

api/hypershift/v1beta1/aws.go
```diff
@@ -25,6 +25,15 @@ type AWSNodePoolPlatform struct {
 	// +kubebuilder:validation:MaxLength=255
 	AMI string `json:"ami,omitempty"`
 
+	// imageType specifies the type of image to use for node instances.
+	// Valid values are ImageTypeLinux ("Linux") or ImageTypeWindows ("Windows").
+	// When set to ImageTypeWindows, the controller will automatically populate the AMI field
```
JoelSpeed: If the value here is Linux, and the image in AMI is Windows, will the controller fix it?

api/hypershift/v1beta1/nodepool_types.go
```diff
@@ -86,6 +100,7 @@ type NodePool struct {
 // +kubebuilder:validation:XValidation:rule="!has(oldSelf.arch) || has(self.arch)", message="Arch is required once set"
 // +kubebuilder:validation:XValidation:rule="self.arch != 'arm64' || has(self.platform.aws) || has(self.platform.azure) || has(self.platform.agent) || self.platform.type == 'None'", message="Setting Arch to arm64 is only supported for AWS, Azure, Agent and None"
 // +kubebuilder:validation:XValidation:rule="!has(self.replicas) || !has(self.autoScaling)", message="Both replicas or autoScaling should not be set"
+// +kubebuilder:validation:XValidation:rule="self.?platform.aws.imageType.orValue(' ') != 'Windows' || (self.arch == 'amd64' && self.platform.type == 'AWS')", message="ImageType 'Windows' is only supported with arch 'amd64' and platform type 'AWS'"
```
JoelSpeed: Before this is merged, we need to confirm if 4.14 management clusters are still considered supporting, or whether we have 4.16 management clusters as the minimum now

api/hypershift/v1beta1/aws.go
BraeTroutman: No, we (nodepool controller) has no way of knowing whether the AMI is a windows one or not based on the ID alone, which is all the information that the controller is privy to: if the image type is Linux (or default) then the controller will populate the AMI exactly how it already does before this MR was made. If the ImageType is Windows, it will look for an AMI in the Windows LI metadata in the ReleaseImage, and if there isn't an image available, it will fail.

api/hypershift/v1beta1/nodepool_types.go
BraeTroutman: I've confirmed in production that all Management Clusters are configured at at least 4.16, in integration we have some legacy clusters at 4.14 or below but I need to look and see if they're even involved in any CI or anything anymore or if they're just test clusters that never got cleaned up

BraeTroutman: ahh but in some environments (FedRAMP, for example) MC is still at 4.15. I will need to adjust this query for compatibility with that apiserver

what specifically about the rule is it that is apiserver dependent? The Null value check (`self.?platform`) and the orValue method?

BraeTroutman:
@bryan-cox @JoelSpeed thank you guys so much for your review-- I've acknowledged initial comments:
- considering APIServer versions in regards to the kubebuilder directives
- updated the enum type to include more explicit options (and updated the docstrings appropriately)
- correct architecture logic when retrieving Windows Images and also enforce an amd64 arch only with ImageType in the API
- move the type within the AWSNodePoolPlatform spec to properly scope it as an AWS feature
As well as acking a couple coderabbit nitpicks

Would you guys be able to take another look at review (and possibly approval?)
I really appreciate your time so far, if I can get this MR through this week that will put me in a great place in terms of being able to start work dependent on this API being available next week, but of course don't want it to be merged if you don't think it's ready

api/hypershift/v1beta1/nodepool_types.go
JoelSpeed: >The Null value check (self.?platform) and the orValue method?

Yep, it's this "optional" semantic that was new for 4.16

api/hypershift/v1beta1/aws.go
```diff
@@ -25,6 +25,17 @@ type AWSNodePoolPlatform struct {
...
+	// with a Windows-compatible AMI based on the region and OpenShift version.
+	// When the AMI field is also set, it will take precedence over automatic selection based
+	// on imageType. Also note that Windows ImageType is only compatible with an Arch of amd64
```
JoelSpeed, line 33: This means that once the AMI is set, it is never updated right? Does hypershift have any concept of boot image upgrades automatically being applied? Would the boot image update controller know that this must be a windows image?

BraeTroutman: Yes the AMI will be a static property of the NodePool that won't change after it's been set: I don't know anything about boot image upgrades with regards to this feature, but I say it would follow the same pattern as when the AMI is just specified explicitly by the client (i.e. explicitly saying `ami: "ami-1234567..."`) in which case the AMI itself is expected to remain the same: I would expect the same behavior here.

yuqi-zhang: > Yes the AMI will be a static property of the NodePool that won't change after it's been set

I was under the impression that most hypershift deployments do not explicitly set their AMI, especially for replace upgrade nodepools. When the cluster's release image reference updates, the AMI is updated alongside the new release image, generates the new machine template, and has the MachineDeployment roll out the upgrade (via replacing the nodes). Are you saying that in windows, we would only set this once? The code below seems to indicate that we would fetch the latest off some windows stream and thereby we'd have the same behaviour (unless the user explicitly sets a AMI).

In fact, in the cases of a hard-set AMI, the user would have to update it themselves, since eventually the node's kubelet package would be out of skew, and there's no inplace mechanism like self-managed openshift to update base os content.

The bootimage update controller only runs in self-managed and does not exist in Hypershift

api/hypershift/v1beta1/nodepool_types.go
```diff
@@ -86,6 +100,7 @@ type NodePool struct {
 // +kubebuilder:validation:XValidation:rule="!has(oldSelf.arch) || has(self.arch)", message="Arch is required once set"
 // +kubebuilder:validation:XValidation:rule="self.arch != 'arm64' || has(self.platform.aws) || has(self.platform.azure) || has(self.platform.agent) || self.platform.type == 'None'", message="Setting Arch to arm64 is only supported for AWS, Azure, Agent and None"
 // +kubebuilder:validation:XValidation:rule="!has(self.replicas) || !has(self.autoScaling)", message="Both replicas or autoScaling should not be set"
+// +kubebuilder:validation:XValidation:rule="!has(self.platform.aws) || self.platform.aws.imageType != 'Windows' || self.arch == 'amd64'"
```
yuqi-zhang: Should we have validation here that specifies that since this would not work for inplace nodepools, so we reject setting Windows for inplace nodepools?

Also, would we be allowed to change this on the fly? i.e. swap the nodes from linux to windows in some replace nodepool?

bryan-cox:
> /test okd-scos-e2e-aws-ovn

@BraeTroutman this test isn't required and doesn't pass very often. Unless you are specifically looking to get it to pass, you don't need to retest it 😄

api/hypershift/v1beta1/nodepool_types.go
BraeTroutman: I would argue we take the precedent set by the ability to specify an AMI in the nodepool directly, which has no validation enforcing that this value cannot be modified on the fly, and use the same approach here for consistency. If we can customize the AMI of a nodepool on the fly directly, then modifying the ImageType on the fly should also be valid (as it's really just an indirect method of doing the same thing)

As for the enforcement of inplace modification of the ImageType, on the ROSA HCP side CS will be enforcing that users cannot modify the imageType of their nodepools day 2

api/hypershift/v1beta1/aws.go
JoelSpeed: It feels to me like the AMI field and this field should be mutually exclusive

Once the field has been set, and we populate the AMI, it is no longer observed, and we have no way to know if the AMI that is set, matches this field, this is generally a bit of a code smell.

Why do we have to populate the AMI field?

hypershift-operator/controllers/nodepool/aws.go
```diff
@@ -317,3 +345,37 @@ func (r NodePoolReconciler) validateAWSPlatformConfig(ctx context.Context, nodeP
...
+	}
+
+	archData, foundArch := releaseImage.StreamMetadata.Architectures[hyperv1.ArchAliases[specifiedArch]]
+	if !foundArch {
+		return "", fmt.Errorf("couldn't find OS metadata for architecture %q", specifiedArch)
+	}
+
+	// Default to x86_64 architecture if not specified
+	// Access the rhel-coreos-extensions aws-winli data
+	winliData := archData.RHCOS.AWSWinLi
```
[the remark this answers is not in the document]
BraeTroutman: RHCOS is not a nullable field

BraeTroutman:
@JoelSpeed @bryan-cox I'm back on this MR now, ive let it languish these past 3 weeks due an urgent move to some FedRAMP work-- would it be possible to get another review when you all get the chance? We'd love to have this feature implemented in time for AWS re:Invent this year

test/e2e/nodepool_imagetype_test.go
```diff
@@ -0,0 +1,221 @@
...
+	return nodePool, nil
+}
+
+func (it *NodePoolImageTypeTest) Run(t *testing.T, nodePool hyperv1.NodePool, nodes []corev1.Node) {
+	ctx := it.ctx
+
+	t.Run("TestDefaultImageType", func(t *testing.T) {
```
muraee: please create separate tests per nodePool. nodePool tests shouldn't be creating nodePools using the client directly.
 All of those tests seem to share the same validation steps, which you can define only once and have different nodePools config (different image-type in your case), see /test/e2e/nodepool_upgrade_test.go for example

api/hypershift/v1beta1/aws.go
```diff
@@ -25,6 +25,17 @@ type AWSNodePoolPlatform struct {
 	// +kubebuilder:validation:MaxLength=255
 	AMI string `json:"ami,omitempty"`
 
+	// imageType specifies the type of image to use for node instances.
+	// Valid values are ImageTypeLinux ("Linux") or ImageTypeWindows ("Windows").
```
muraee:
```suggestion
	// Valid values are "Linux" or "Windows"
```
I don't think we need to mention the variable names in godoc

BraeTroutman: acked this in a amend commit-- will be pushed along with the e2e changes shortly

test/e2e/nodepool_imagetype_test.go
BraeTroutman: can do, just updated the test cases. Note that the negative cases (those expecting failure) that I had can't be tested this way, because the executeNodepoolTest function expects that nodepools will always become ready before executing follow-up validations defined in the tests.

test/e2e/nodepool_test.go
```diff
@@ -129,6 +129,10 @@ func TestNodePool(t *testing.T) {
 						name: "TestMirrorConfigs",
 						test: NewMirrorConfigsTest(ctx, mgtClient, hostedCluster, hostedClusterClient),
 					},
+					{
+						name: "TestImageTypes",
```
devguyio, line 133: This shouldn't be here, this should be in a new `HostedClusterNodePoolTestCases` similar to the KubeVirt one on lines
140:152, with a setup function that skips the test for non-AWS platforms. 

The way this is currently set will skip the test after a HostedCluster is created, so for non-AWS that'd be an extra
wasted HC.

test/e2e/nodepool_imagetype_test.go
```diff
@@ -0,0 +1,114 @@
...
+		mgmtClient:          mgmtClient,
+	}
+}
+
+func (it *NodePoolImageTypeTest) Setup(t *testing.T) {
+	// Skip test for non-AWS platforms since ImageType is currently AWS-specific
+	if globalOpts.Platform != hyperv1.AWSPlatform {
+		t.Skip("test is only supported for AWS platform")
+	}
+	if e2eutil.IsLessThan(e2eutil.Version419) {
```
devguyio, line 48: This check needs to be done in the setup of a `HostedClusterNodePoolTestCases`, not here.

devguyio (approved):
/lgtm

Great work, and apologies for the mixup :)

PR #6752, opened 2025-09-03
CNTRLPLANE-1374

bryan-cox:

Fixes #

**Checklist**

### DIFF
--- a/contrib/self-managed azure/setup_all.sh
+++ b/contrib/self-managed azure/setup_all.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+set -x
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+# Check if user is logged into Azure, and log in if not
+if ! az account show >/dev/null 2>&1; then
+  echo "Not logged into Azure. Logging in now..."
+  az login --scope https://management.core.windows.net//.default
+  if [ $? -ne 0 ]; then
+    echo "Error: Azure login failed. Please check your credentials."
+    exit 1
+  fi
+  echo "Successfully logged into Azure!"
+fi
+
+source "${SCRIPT_DIR}/vars.sh"
+
+# One-time setup scripts (only run when --first-time flag is provided)
+if [[ "$1" == "--first-time" ]]; then
+  "${SCRIPT_DIR}/setup_oidc_provider.sh"
+  "${SCRIPT_DIR}/setup_workload_identities.sh"
+fi
+
+# Per-cluster setup scripts
+
+set +x
--- a/contrib/self-managed azure/setup_workload_identities.sh
+++ b/contrib/self-managed azure/setup_workload_identities.sh
@@ -1,0 +1,137 @@
+#!/bin/bash
+set -x
+
+# Prerequisites.
+PERSISTENT_RG_NAME=${PERSISTENT_RG_NAME:-}
+OIDC_ISSUER_URL=${OIDC_ISSUER_URL:-}
+
+# Local.
+WORKLOAD_IDENTITIES_FILE=${WORKLOAD_IDENTITIES_FILE:-}
+AZURE_DISK_MI_NAME="workload-azure-disk-MI-${PREFIX}"
+AZURE_FILE_MI_NAME="workload-azure-file-MI-${PREFIX}"
+IMAGE_REGISTRY_MI_NAME="workload-image-registry-MI-${PREFIX}"
+INGRESS_MI_NAME="workload-ingress-MI-${PREFIX}"
+CLOUD_PROVIDER_MI_NAME="workload-cloud-provider-MI-${PREFIX}"
+NODE_POOL_MANAGEMENT_MI_NAME="workload-nodepool-mgmt-MI-${PREFIX}"
+NETWORK_MI_NAME="workload-network-MI-${PREFIX}"
+
+AZURE_DISK_CLIENT_ID=$(az identity create --name $AZURE_DISK_MI_NAME --resource-group $PERSISTENT_RG_NAME --query clientId -o tsv)
+AZURE_FILE_CLIENT_ID=$(az identity create --name $AZURE_FILE_MI_NAME --resource-group $PERSISTENT_RG_NAME --query clientId -o tsv)
+IMAGE_REGISTRY_CLIENT_ID=$(az identity create --name $IMAGE_REGISTRY_MI_NAME --resource-group $PERSISTENT_RG_NAME --query clientId -o tsv)
+INGRESS_CLIENT_ID=$(az identity create --name $INGRESS_MI_NAME --resource-group $PERSISTENT_RG_NAME --query clientId -o tsv)
+CLOUD_PROVIDER_CLIENT_ID=$(az identity create --name $CLOUD_PROVIDER_MI_NAME --resource-group $PERSISTENT_RG_NAME --query clientId -o tsv)
+NODE_POOL_MANAGEMENT_CLIENT_ID=$(az identity create --name $NODE_POOL_MANAGEMENT_MI_NAME --resource-group $PERSISTENT_RG_NAME --query clientId -o tsv)
+NETWORK_CLIENT_ID=$(az identity create --name $NETWORK_MI_NAME --resource-group $PERSISTENT_RG_NAME --query clientId -o tsv)
+
+cat <<EOF > "${WORKLOAD_IDENTITIES_FILE}"
+{
+  "imageRegistry": {
+    "clientID": "${IMAGE_REGISTRY_CLIENT_ID}"
+  },
+  "ingress": {
+    "clientID": "${INGRESS_CLIENT_ID}"
+  },
+  "file": {
+    "clientID": "${AZURE_FILE_CLIENT_ID}"
+  },
+  "disk": {
+    "clientID": "${AZURE_DISK_CLIENT_ID}"
+  },
+  "nodePoolManagement": {
+    "clientID": "${NODE_POOL_MANAGEMENT_CLIENT_ID}"
+  },
+  "cloudProvider": {
+    "clientID": "${CLOUD_PROVIDER_CLIENT_ID}"
+  },
+  "network": {
+    "clientID": "${NETWORK_CLIENT_ID}"
+  }
+}
+EOF
+
+# Set up the federated IDs for Workload Identity
+az identity federated-credential create --name "${AZURE_DISK_MI_NAME}"-fed-id-node \
+--identity-name "${AZURE_DISK_MI_NAME}" \
+--resource-group "${PERSISTENT_RG_NAME}" \
+--issuer "${OIDC_ISSUER_URL}" \
+--subject system:serviceaccount:openshift-cluster-csi-drivers:azure-disk-csi-driver-node-sa \
+--audience openshift
+
+az identity federated-credential create --name "${AZURE_DISK_MI_NAME}"-fed-id-operator \
+--identity-name "${AZURE_DISK_MI_NAME}" \
+--resource-group "${PERSISTENT_RG_NAME}" \
+--issuer "${OIDC_ISSUER_URL}" \
+--subject system:serviceaccount:openshift-cluster-csi-drivers:azure-disk-csi-driver-operator \
+--audience openshift
+
+az identity federated-credential create --name "${AZURE_DISK_MI_NAME}"-fed-id-controller \
+--identity-name "${AZURE_DISK_MI_NAME}" \
+--resource-group "${PERSISTENT_RG_NAME}" \
+--issuer "${OIDC_ISSUER_URL}" \
+--subject system:serviceaccount:openshift-cluster-csi-drivers:azure-disk-csi-driver-controller-sa \
+--audience openshift
+
+az identity federated-credential create --name "${AZURE_FILE_MI_NAME}"-fed-id-node \
+--identity-name "${AZURE_FILE_MI_NAME}" \
+--resource-group "${PERSISTENT_RG_NAME}" \
+--issuer "${OIDC_ISSUER_URL}" \
+--subject system:serviceaccount:openshift-cluster-csi-drivers:azure-file-csi-driver-node-sa \
+--audience openshift
+
+az identity federated-credential create --name "${AZURE_FILE_MI_NAME}"-fed-id-operator \
+--identity-name "${AZURE_FILE_MI_NAME}" \
+--resource-group "${PERSISTENT_RG_NAME}" \
+--issuer "${OIDC_ISSUER_URL}" \
+--subject system:serviceaccount:openshift-cluster-csi-drivers:azure-file-csi-driver-operator \
+--audience openshift
+
+az identity federated-credential create --name "${AZURE_FILE_MI_NAME}"-fed-id-controller \
+--identity-name "${AZURE_FILE_MI_NAME}" \
+--resource-group "${PERSISTENT_RG_NAME}" \
+--issuer "${OIDC_ISSUER_URL}" \
+--subject system:serviceaccount:openshift-cluster-csi-drivers:azure-file-csi-driver-controller-sa \
+--audience openshift
+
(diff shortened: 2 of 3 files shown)


cmd/cluster/azure/create.go
```diff
@@ -260,13 +250,13 @@ func (o *CreateOptions) ApplyPlatformSpecifics(cluster *hyperv1.HostedCluster) e
 	if o.infra.WorkloadIdentities != nil {
 		// Self-managed Azure with workload identities
 		cluster.Spec.Platform.Azure.AzureAuthenticationConfig = hyperv1.AzureAuthenticationConfiguration{
-			AzureAuthenticationConfigType: hyperv1.AzureAuthenticationTypeWorkloadIdentities,
+			AzureAuthenticationConfigType: "WorkloadIdentities",
```
jparrill: Why hardcoding this?

cmd/cluster/azure/create.go
```diff
@@ -260,13 +250,13 @@ func (o *CreateOptions) ApplyPlatformSpecifics(cluster *hyperv1.HostedCluster) e
...
 			WorkloadIdentities:            o.infra.WorkloadIdentities,
 		}
 	} else {
 		// Managed Azure with managed identities
 		cluster.Spec.Platform.Azure.AzureAuthenticationConfig = hyperv1.AzureAuthenticationConfiguration{
-			AzureAuthenticationConfigType: hyperv1.AzureAuthenticationTypeManagedIdentities,
+			AzureAuthenticationConfigType: "ManagedIdentities",
```
jparrill: same question as above

cmd/cluster/azure/create.go
bryan-cox: That is a rebase mistake 🤦🏻 

Fixed.

cmd/cluster/azure/create.go
bryan-cox: That is a rebase mistake 🤦🏻 

Fixed.

cmd/infra/azure/identities.go
```diff
@@ -0,0 +1,294 @@
...
+	}
+
+	identity, err := client.CreateOrUpdate(ctx, resourceGroupName, identityName, armmsi.Identity{
+		Location: ptr.To(location),
+	}, nil)
+	if err != nil {
+		l.Error(err, "Failed to create managed identity",
+			"name", identityName,
+			"resourceGroup", resourceGroupName)
+		return "", "", "", fmt.Errorf("failed to create managed identity: %w", err)
```
jparrill: are not `l.Error` + return `fmt.Errorf` redundant at log level?

cmd/infra/azure/identities.go
```diff
@@ -0,0 +1,294 @@
...
+
+	_, err = client.CreateOrUpdate(ctx, resourceGroupName, identityName, config.CredentialName, armmsi.FederatedIdentityCredential{
+		Properties: &armmsi.FederatedIdentityCredentialProperties{
+			Issuer:    ptr.To(oidcIssuerURL),
+			Subject:   ptr.To(config.Subject),
+			Audiences: []*string{ptr.To(config.Audience)},
+		},
+	}, nil)
+	if err != nil {
+		l.Error(err, "Failed to create federated identity credential",
```
jparrill: Same as above. Maybe we can add all this info in the return of the `fmt.Errorf`?

cmd/infra/azure/identities.go
```diff
@@ -0,0 +1,294 @@
...
+		},
+		{
+			CredentialName: imageRegistryIdentityName + "-fed-id-operator",
+			Subject:        "system:serviceaccount:openshift-image-registry:cluster-image-registry-operator",
+			Audience:       "openshift",
+		},
+	}
+
+	for _, cred := range imageRegistryCredentials {
+		if err := i.createFederatedIdentityCredential(ctx, l, resourceGroupName, imageRegistryIdentityName+"-"+opts.InfraID, opts.OIDCIssuerURL, cred); err != nil {
```
jparrill: Maybe better for clarify + performance to put this `imageRegistryIdentityName+"-"+opts.InfraID` in a separated variable just above of the for?

For performance it will do the addition in every cycle of the loop, meanwhile if you have it outside it's added just once and it's clearer. Also I think to use `fmt.Sprintf` for the addition, performs better than ""+"" in a high scale.

cmd/infra/azure/identities.go
```diff
@@ -0,0 +1,294 @@
...
+	workloadIdentities.CloudProvider.ClientID = hyperv1.AzureClientID(cloudProviderClientID)
+
+	// Create federated credential for cloud provider identity
+	cloudProviderCredential := FederatedCredentialConfig{
+		CredentialName: cloudProviderIdentityName + "-fed-id",
+		Subject:        "system:serviceaccount:kube-system:azure-cloud-provider",
+		Audience:       "openshift",
+	}
+
+	if err := i.createFederatedIdentityCredential(ctx, l, resourceGroupName, cloudProviderIdentityName+"-"+opts.InfraID, opts.OIDCIssuerURL, cloudProviderCredential); err != nil {
```
jparrill: I would follow the same pattern in all the places as you did in line 237-238:
```
cloudProviderIdentityName := fmt.Sprintf("%s-cloud-provider", opts.Name)
cloudProviderClientID, _, err := i.createManagedIdentity(ctx, l, resourceGroupName, cloudProviderIdentityName, opts.InfraID, opts.Location)
```

cmd/infra/azure/identities.go
```diff
@@ -0,0 +1,294 @@
...
+	workloadIdentities.NodePoolManagement.ClientID = hyperv1.AzureClientID(nodePoolMgmtClientID)
+
+	// Create federated credential for node pool management identity
+	nodePoolMgmtCredential := FederatedCredentialConfig{
+		CredentialName: nodePoolMgmtIdentityName + "-fed-id",
+		Subject:        "system:serviceaccount:kube-system:capi-provider",
+		Audience:       "openshift",
+	}
+
+	if err := i.createFederatedIdentityCredential(ctx, l, resourceGroupName, nodePoolMgmtIdentityName+"-"+opts.InfraID, opts.OIDCIssuerURL, nodePoolMgmtCredential); err != nil {
```
jparrill: Same as above

cmd/infra/azure/identities.go
```diff
@@ -0,0 +1,294 @@
...
+	workloadIdentities.Network.ClientID = hyperv1.AzureClientID(networkClientID)
+
+	// Create federated credential for network identity
+	networkCredential := FederatedCredentialConfig{
+		CredentialName: networkIdentityName + "-fed-id",
+		Subject:        "system:serviceaccount:openshift-cloud-network-config-controller:cloud-network-config-controller",
+		Audience:       "openshift",
+	}
+
+	if err := i.createFederatedIdentityCredential(ctx, l, resourceGroupName, networkIdentityName+"-"+opts.InfraID, opts.OIDCIssuerURL, networkCredential); err != nil {
```
jparrill: Same as above

cmd/infra/azure/images.go
```diff
@@ -0,0 +1,199 @@
...
+		return "", errors.New("no storage account keys exist")
+	}
+
+	credential, err := container.NewSharedKeyCredential(storageAccountName, *storageAccountKeyResult.Keys[0].Value)
+	if err != nil {
+		return "", fmt.Errorf("failed to create shared key credentials: %w", err)
+	}
+
+	imageBlobURLPrefix := fmt.Sprintf("https://%s.blob.core.windows.net/vhd/", storageAccountName)
+
```
jparrill, line 108: nit picky: remove line

cmd/infra/azure/networking.go
```diff
@@ -0,0 +1,294 @@
...
+	return "", fmt.Errorf("could not find any DNS zones in subscription")
+}
+
+// CreateSecurityGroup creates the security group the virtual network will use
+func (n *NetworkManager) CreateSecurityGroup(ctx context.Context, resourceGroupName string, name string, infraID string, location string) (string, error) {
+	securityGroupClient, err := armnetwork.NewSecurityGroupsClient(n.subscriptionID, n.creds, nil)
+	if err != nil {
+		return "", fmt.Errorf("failed to create security group client: %w", err)
+	}
+	securityGroupFuture, err := securityGroupClient.BeginCreateOrUpdate(ctx, resourceGroupName, name+"-"+infraID+"-nsg", armnetwork.SecurityGroup{Location: &location}, nil)
```
jparrill: Separate the function call from concat for clarity

cmd/infra/azure/networking.go
```diff
@@ -0,0 +1,294 @@
...
+		Location: &location,
+		Properties: &armnetwork.VirtualNetworkPropertiesFormat{
+			AddressSpace: &armnetwork.AddressSpace{
+				AddressPrefixes: []*string{
+					ptr.To(VirtualNetworkAddressPrefix),
+				},
+			},
+			Subnets: []*armnetwork.Subnet{},
+		},
+	}
```
jparrill: I think this could be in a separated method like `NewVirtualNetwork` receiving as arg the VNetAddrPrefix

cmd/infra/azure/networking.go
```diff
@@ -0,0 +1,294 @@
...
+				AddressPrefix: ptr.To(VirtualNetworkSubnetAddressPrefix),
+				NetworkSecurityGroup: &armnetwork.SecurityGroup{
+					ID: ptr.To(securityGroupID),
+				},
+			},
+		})
+		l.Info("Creating new subnet for vnet creation")
+	}
+
+	vnetFuture, err := networksClient.BeginCreateOrUpdate(ctx, resourceGroupName, name+"-"+infraID, vnetToCreate, nil)
```
jparrill: Separate the function from concat for clarity

cmd/infra/azure/networking.go
```diff
@@ -0,0 +1,294 @@
...
+		return fmt.Errorf("failed to create new virtual network links client: %w", err)
+	}
+
+	virtualNetworkLinkParams := armprivatedns.VirtualNetworkLink{
+		Location: ptr.To(VirtualNetworkLinkLocation),
+		Properties: &armprivatedns.VirtualNetworkLinkProperties{
+			VirtualNetwork:      &armprivatedns.SubResource{ID: &vnetID},
+			RegistrationEnabled: ptr.To(false),
+		},
+	}
```
jparrill: I think this could be in a separated method like NewVirtualNetworkLink receiving as arg the VirtualNetworkLinkProperties

cmd/infra/azure/networking.go
```diff
@@ -0,0 +1,294 @@
...
+			Location: ptr.To(location),
+			Properties: &armnetwork.PublicIPAddressPropertiesFormat{
+				PublicIPAddressVersion:   ptr.To(armnetwork.IPVersionIPv4),
+				PublicIPAllocationMethod: ptr.To(armnetwork.IPAllocationMethodStatic),
+				IdleTimeoutInMinutes:     ptr.To[int32](4),
+			},
+			SKU: &armnetwork.PublicIPAddressSKU{
+				Name: ptr.To(armnetwork.PublicIPAddressSKUNameStandard),
+			},
+		},
```
jparrill: Maybe this could be a new method newArmNetworkPublicIPAddress that could be called from outside of the function call, once that is received, just use one variable parameter instead the whole chunk.


Something like:
```
armNetPublicIPAddress := newArmNetworkPublicIPAddress(infraID, location.....)
pollerResp, err := publicIPAddressClient.BeginCreateOrUpdate(
		ctx,
		resourceGroupName,
		infraID,
		armNetPublicIPAddress,
        nil,
)
```

cmd/infra/azure/networking.go
```diff
@@ -0,0 +1,294 @@
...
+func (n *NetworkManager) CreateLoadBalancer(ctx context.Context, resourceGroupName string, infraID string, location string, publicIPAddress *armnetwork.PublicIPAddress) error {
+	idPrefix := fmt.Sprintf("subscriptions/%s/resourceGroups/%s/providers/Microsoft.Network/loadBalancers", n.subscriptionID, resourceGroupName)
+	loadBalancerName := infraID
+
+	loadBalancerClient, err := armnetwork.NewLoadBalancersClient(n.subscriptionID, n.creds, nil)
+	if err != nil {
+		return fmt.Errorf("failed to create load balancer client, %w", err)
+	}
+
+	pollerResp, err := loadBalancerClient.BeginCreateOrUpdate(ctx,
```
jparrill: wow this is huge, any way to simplify it to have it clearer?

cmd/infra/azure/rbac.go
```diff
@@ -0,0 +1,307 @@
...
+// NewRBACManager creates a new RBACManager
+func NewRBACManager(subscriptionID string, creds azcore.TokenCredential) *RBACManager {
+	return &RBACManager{
+		subscriptionID: subscriptionID,
+		creds:          creds,
+	}
+}
+
+// AssignControlPlaneRoles assigns roles to control plane managed identities
+func (r *RBACManager) AssignControlPlaneRoles(ctx context.Context, opts *CreateInfraOptions, controlPlaneMIs *hyperv1.AzureResourceManagedIdentities, resourceGroupName, nsgResourceGroupName, vnetResourceGroupName string) error {
```
jparrill, line 55: The logic on AssignControlPlaneRoles and AssignWorkloadIdentities looks pretty similar, WDYT to have them together in same function? not 100% sure about this, is just to avoid repeat logic we already have in place

contrib/managed-azure/setup_external_dns.sh
```diff
@@ -8,7 +8,6 @@ PARENT_DNS_ZONE="${PARENT_DNS_ZONE:-}"
 RG="${PERSISTENT_RG_NAME:-}"
 
 # Local.
-DNS_RECORD_NAME="${PREFIX}"
```
jparrill, line 11: If this is removed, the next value will be broken, isn't it?

```
MGMT_DNS_ZONE_NAME="${DNS_RECORD_NAME}.${PARENT_DNS_ZONE}"

```

contrib/self-managed-azure/setup_hypershift_operator.sh
```diff
@@ -0,0 +1,35 @@
...
+fi
+
+if [[ ! -f "${PULL_SECRET}" ]]; then
+    echo "Error: Pull secret file not found at ${PULL_SECRET}"
+    exit 1
+fi
+
+# Install HyperShift Operator
+echo "Installing HyperShift Operator with Azure External DNS..."
+"${HYPERSHIFT_BINARY_PATH}/hypershift" install \
```
jparrill, line 23: I would check if that file exists or force to remove it (to avoid errors with older Hypershift binaries) + recompile (?)

jparrill (commented):
Dropped a couple of more things. PR review completed

contrib/managed-azure/setup_external_dns.sh
bryan-cox: This should be set here - https://github.com/openshift/hypershift/blob/e0b02cad00065478294ca430a99e688f6205e23b/contrib/self-managed-azure/user-vars.sh.example#L10

contrib/self-managed-azure/setup_hypershift_operator.sh
bryan-cox: We don't do this for the managed Azure scripts though. These are dev scripts too so I think there is some onus on the user to make sure things are up to date like their binary 🤷🏻‍♂️

cmd/infra/azure/images.go
bryan-cox: Let's wait on this. I think we might completely remove this code in a different story.

cmd/infra/azure/rbac.go
bryan-cox: Yeah they are similar but I would prefer to keep them separate since once is specifically for ARO HCP and another is specifically for self-managed Azure.

jparrill (approved):
Hey @bryan-cox looks good, I think you need to rebase the latest Konflux changes to make EC to work back again. Let me know when that happens for tagging

PR #6683, opened 2025-08-20
OSASINFRA-3675: Add credentials to openshift-manila-csi-driver namespace for Hypershift

dkokkino:
**What this PR does / why we need it**:

The credentials are already provisioned in the
openshift-manila-csi-driver namespace for standalone deployments via the clouds-credential-operator. However, Hypershift does not use the clouds-credential-operator and instead manages credentials through a reconciliation loop (reconcileCloudCredentialSecrets).

This PR ensures the secret is explicitly added to the openshift-manila-csi-driver namespace for Hypershift deployments.

**Which issue(s) this PR fixes:**
Fixes [OSASINFRA-3675](https://issues.redhat.com/browse/OSASINFRA-3675)

**Checklist**
- [x] Subject and description added to both, commit and PR.
- [x] Relevant issues have been referenced.
- [ ] This change includes docs. 
- [ ] This change includes unit tests.

PR #6777, opened 2025-09-08
OSASINFRA-3675: openstack: Sync cloud credentials to cluster namespace

stephenfin:

In https://github.com/openshift/hypershift/pull/6683, we extended the control-plane-operator to sync credentials to `openshift-manila-csi-driver / manila-cloud-credentials`. This satisfies the node driver however, unlike Cinder, the Manila controller also requires access to the cloud. These pods run in the clusters-* namespace so we also need to create the secret there.

Fixes #[OSASINFRA-3675](https://issues.redhat.com//browse/OSASINFRA-3675)

**Checklist**


stephenfin:
/close

As the :robot: says, this is already happening. Something else is up.

PR #6782, opened 2025-09-09
OCPBUGS-61462: temporarily use v4.20 catalogs

sjenning:
v4.21 catalogs have a filesystem permissions error that are causing HCP to fail to deploy.

Temporarily use v4.20 catalogs until this is fixed.

### DIFF
--- a/support/catalogs/images.go
+++ b/support/catalogs/images.go
@@ -171,6 +171,10 @@ func computeCatalogImages(releaseVersion func() (*semver.Version, error), imageE
 
 	for catalog := range catalogs {
 		catalogVersion := *version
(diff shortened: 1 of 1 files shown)


bryan-cox:
/retitle OCPBUGS-61462: fix(catalogs-4.21): temporarily use v4.20 catalogs

sjenning:
We have a chicken-egg where the Upgrade test is still using the v4.21 tags, because the CPO doesn't have this fix, and failing.

Forcing

/override ci/prow/e2e-aws
/override ci/prow/e2e-aks-4-20
/override ci/prow/e2e-aks

PR #6783, opened 2025-09-09
NO-JIRA: add permissions to role policy

sjenning:
Add missing permissions to role policy


sjenning:
/verified bypass
ci-tooling change

PR #6787, opened 2025-09-10
NO-JIRA: chore(AI): add claude agents for different SMEs

enxebre:
This introduces claude subagents for different SMEs:
- Control Plane
- Data Plane
- API
- Cloud provider
- HCP architecture

It also introduces a claude command for fetaure development levearing the SMEs

This an example with a prompt intentionally vague to observe the agents behaviour and flaws:
```
/workflows:feature-development add support for HCP in GCP aka google cloud
```

Result is https://github.com/openshift/hypershift/compare/main...enxebre:claude-poc?expand=1

Follow up:
- Include quality gates for each workflow step
- Include devops SME to deploy management cluster and hostedCluster
- Include QE SME to validate deployments
- Fuzz testing for the workflow with granular concise tasks

Fixes #

**Checklist**

### DIFF
--- a/.claude/agents/api-sme.md
+++ b/.claude/agents/api-sme.md
@@ -1,0 +1,32 @@
+---
+name: api-sme
+description: Has deep knowledge of the Kubernetes and OpenShift API best practices. It is familiar
+with all the OpenShift APIs for configuration and operators. It owns the hypershift.openshift.io APIs,
+including but not limited to hostedCluster, hostedControlPlane, nodePool and all the platform specifics.
+Makes API design decisions and enforce best practices.
+model: opus
+---
+
+You are an API subject matter expert system architect specializing in HCP.
+
+## Focus Areas
+- API design, versioning and error reporting for anything within api/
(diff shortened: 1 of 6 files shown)


enxebre:
@coderabbitai review

jparrill:
The new agents looks good. Some questions:
- In order to consume them is just matter of open Claude at the repo level?
- What if Opus model is not available?
- Those agents are loaded automatically or should we enable them someway?

enxebre:
Fixed all typos via claude.

>In order to consume them is just matter of open Claude at the repo level?
What if Opus model is not available?
Those agents are loaded automatically or should we enable them someway?

The PR includes a claude command which you can you to leverage the agents to implement a feature. See PR description example.
You can also ask claude code to use an agent explicitly or it can choose to use one based on your prompt. See https://docs.anthropic.com/en/docs/claude-code/sub-agents
If the model is not available I would expect it to fallback but I'm not sure. I believe there's a fallback_model you can use.

In any case this PR is meant to be just an starting building block to enable others to experiment.

PR #6805, opened 2025-09-12
Revert "OCPBUGS-61334: fix(kas): Set correct container port in postStart handler"

bryan-cox:
The original PR was originally introduced here:
  - #6773

  It was reverted here because:
  - The addition of the -f flag to the curl command in the kube-apiserver postStart hook causes the container to restart during normal RBAC bootstrap operations. The /readyz endpoint returns HTTP error status codes while the kube-apiserver is bootstrapping RBAC roles, which is expected behavior. However, with
   the -f flag, curl now fails immediately on these HTTP errors instead of retrying, causing the postStart hook to fail and terminating the container.

  This PR is now being reverted because:
  - Test Failures: AWS e2e tests are failing with Container kube-apiserver in pod kube-apiserver-6db5dc9d44-j6qtz has a restartCount > 0 (1) due to postStart hook failures
  - RBAC Bootstrap Incompatibility: The kube-apiserver logs show continuous errors like "unable to reconcile rolebinding.rbac.authorization.k8s.io/system:controller:bootstrap-signer: could not list rolebinding restrictions: the server could not find the requested resource (get 
  rolebindingrestrictions.authorization.openshift.io)" during startup, which cause /readyz to return failure status
  - Breaking Change: The -f flag change fundamentally altered the error tolerance behavior of the postStart hook, making it incompatible with the expected bootstrap sequence where OpenShift CRDs are not immediately available
  - Production Impact: This causes kube-apiserver instability during cluster bootstrap, particularly affecting platforms like AWS where crashToleration = 0

  The original curl command without -f properly handled the bootstrap phase by retrying on HTTP errors until the RBAC bootstrap completed successfully.


rtheis:
Per https://ibm-argonauts.slack.com/archives/C01C8502FMM/p1757679162457309, I don't think that `-f` was the problem. However, in order to restore testing, we should proceed with the revert and work on a follow-up fix.

PR #6772, opened 2025-09-08
[release-4.20] OCPBUGS-61066: fix: propagate AWSLoadBalancerTargetNodesAnnotation to HCP

mgencur:
Backport of https://github.com/openshift/hypershift/pull/6823

**What this PR does / why we need it**:

* Add AWSLoadBalancerTargetNodesAnnotation to the list of mirrored annotations in reconcileHostedControlPlaneAnnotations. This enables AWS load balancer target node configurations set on HostedCluster resources to be properly propagated to their corresponding HostedControlPlane resources.
* Propagate osd-fleet-manager.openshift.io/paired-nodes label from
request-serving nodes and set it as
service.beta.kubernetes.io/aws-load-balancer-target-node-labels
annotation on HC.

**Which issue(s) this PR fixes** *(optional, use `fixes #<issue_number>(, fixes #<issue_number>, ...)` format, where issue_number might be a GitHub issue, or a Jira story*:
Fixes [OCPBUGS-61066](https://issues.redhat.com/browse/OCPBUGS-61066)

**Checklist**
- [x] Subject and description added to both, commit and PR.
- [x] Relevant issues have been referenced.
- [ ] This change includes docs. 
- [ ] This change includes unit tests.

PR #6823, opened 2025-09-15
OCPBUGS-61719: fix: propagate AWSLoadBalancerTargetNodesAnnotation to…

mgencur:
… HCP

* Add AWSLoadBalancerTargetNodesAnnotation to the list of mirrored annotations
in reconcileHostedControlPlaneAnnotations. This enables AWS load balancer
target node configurations set on HostedCluster resources to be properly propagated to their corresponding HostedControlPlane resources.
* Propagate osd-fleet-manager.openshift.io/paired-nodes label from request-serving nodes and set it as
service.beta.kubernetes.io/aws-load-balancer-target-node-labels annotation on HC.

Fixes [OCPBUGS-61719](https://issues.redhat.com/browse/OCPBUGS-61719)

**Checklist**

### DIFF
--- a/hypershift-operator/controllers/scheduler/aws/dedicated_request_serving_nodes_test.go
+++ b/hypershift-operator/controllers/scheduler/aws/dedicated_request_serving_nodes_test.go
@@ -169,6 +169,7 @@ func TestHostedClusterScheduler(t *testing.T) {
 		checkScheduledNodes   bool
 		checkScheduledCluster bool
 		expectError           bool
+		expectedPairLabel     string
 	}{
 		{
 			name: "deleted hosted cluster",
@@ -192,6 +193,7 @@ func TestHostedClusterScheduler(t *testing.T) {
 				node("n4", "zone-c", "id2")),
 			checkScheduledNodes:   true,
 			checkScheduledCluster: true,
+			expectedPairLabel:     "id1",
 		},
 		{
 			name: "available node, existing assigned node",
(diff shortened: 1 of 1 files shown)


jparrill:
/approve 

On behalf of Cesar done for 4.20 in https://github.com/openshift/hypershift/pull/6772

hypershift-operator/controllers/scheduler/aws/dedicated_request_serving_nodes.go
```diff
@@ -262,6 +263,9 @@ func (r *DedicatedServingComponentScheduler) Reconcile(ctx context.Context, req
 			// If subnets are separated by periods, replace them with commas
 			lbSubnets = strings.ReplaceAll(lbSubnets, ".", ",")
 		}
+		if node.Labels[OSDFleetManagerPairedNodesLabel] != "" && pairLabel == "" {
+			pairLabel = node.Labels[OSDFleetManagerPairedNodesLabel]
```
jparrill, line 267: How could be different from `pairLabel == ""` if you set it up  as empty un 254?
Is this being reset in every reconciliation cycle?

jparrill (commented):
dropped a comment.

hypershift-operator/controllers/scheduler/aws/dedicated_request_serving_nodes.go
mgencur: I used the same pattern as for the variables above. And i think it's a small optimization to set it only once, even if there are multiple nodes (because we iterate over nodesToUse.

mgencur:
Added a unit test as requested on https://github.com/openshift/hypershift/pull/6772

jparrill (commented):
/lgtm

PR #6824, opened 2025-09-15
OCPBUGS-61334: fix(kas): Set correct container port in postStart handler, 2

matlaj:
Make sure the kube-apiserver `postStart` handler uses the correct container port instead of hardcoding the default.
Without this change, kube-apiserver pods will be stuck initializing if the container port is not the default 6443.

This is a follow-up attempt to fix the issue after https://github.com/openshift/hypershift/pull/6773 was reverted.
No changes are made to the `postStart` handler other than dynamically setting the port. I think any broader concerns with the existing handler logic are out of scope here and best handled in a separate PR. This PR is strictly about setting the correct port.

Fixes #[OCPBUGS-61334](https://issues.redhat.com/browse/OCPBUGS-61334)

**Checklist**


rtheis:
Based on the unit test failure, you may need to rebase the PR on main.

rtheis (approved):
/lgtm

PR #6827, opened 2025-09-15
OCPBUGS-50603: Use HCCO kubeconfig to apply the FeatureGate CR

smrtrfszm:

Currently it's not possible to change the feature gates for a CustomNoUpgrade feature set cluster, because there's a Validating Admission Policy that only allows the change if the username is `system:hosted-cluster-config` and the KAS bootstrap container tries to change it with `system:admin`. Also it uses the `CreateOrUpdate` function with an empty callback function, which causes any update to fail silently.

This PR creates a new kubeconfig that has the certs from the `hcco-kubeconfig`, but the KAS url is `localhost`, as the bootstrap container cannot access the KAS any other way. When to bootstrap script tries the apply the FeatureGate custom resource, it will use this kubeconfig.

Secondly, it replaces the `CreateOrUpdate` with server-side apply, so updates will work as expected.

Fixes #https://issues.redhat.com/browse/OCPBUGS-50603

**Checklist**


control-plane-operator/controllers/hostedcontrolplane/v2/assets/kube-apiserver/deployment.yaml
```diff
@@ -120,6 +120,8 @@ spec:
         env:
         - name: KUBECONFIG
           value: /var/secrets/localhost-kubeconfig/kubeconfig
+        - name: KUBECONFIG_HCCO
```
rtheis: Is there any particular reason why we don't always use HCCO identity for bootstrapping? I could very easily see us hitting this problem for other resources in the future and it seems wise for us to ensure that bootstrapping is not blocked by the validating admission policy.

smrtrfszm: Because this same program gives the permissions to the hcco identity. We would have to apply the `98_hcco-rolebinding.yaml` somewhere else.

rtheis: Have Red Hat engineers had a chance to look at this PR and provide their opinion?

csrwng: My preference would be to consider the rolebinding for the hcco a special thing (and even exclude it from the list of manifests). So invert what this PR does... use the old kubeconfig only for the hcco rolebinding (and only create it if not needed). The bootstrap program could just create/reconcile it inline. Once the HCCO has the power to do anything, then apply all bootstrap manifests using the hcco kubeconfig. That way future bootstrap manifests that we may disallow via VAP can still be applied.

rtheis: @csrwng I am a bit confused that you approved the PR yet would prefer things to be done differently. Can you please explain?

csrwng: per @sjenning, let's create a separate identity that is allowed to make the same updates as the HCCO. That way we can keep track of who makes the updates.

csrwng: > Can you please explain?

I removed the approved label b/c I thought of a better way to handle this ^^.

smrtrfszm: I've updated the PR with the new identity.

TwoDCube:
@coderabbitai full review

kas-bootstrap/kas_boostrap_test.go
```diff
@@ -370,58 +373,175 @@ func TestReconcileFeatureGate(t *testing.T) {
...
+}
+
+func newFakeApplyClient() *fakeApplyClient {
+	c := fake.NewClientBuilder().WithScheme(configScheme).Build()
+	return &fakeApplyClient{Client: c}
+}
+
+func (c *fakeApplyClient) Apply(ctx context.Context, obj *unstructured.Unstructured, opts ...client.PatchOption) error {
+	return c.Client.Create(ctx, obj)
+}
```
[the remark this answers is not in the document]
TwoDCube: @coderabbitai have you considered that the controller-runtime is actually being replaced with a different version in go.mod?

```
replace sigs.k8s.io/controller-runtime => sigs.k8s.io/controller-runtime v0.19.7
```

TwoDCube:
As discussed, the new identity is definitely the right direction for new releases. However, this would be a breaking change for older releases so this can't be backported as-is. The backports should impersonate the HCCO to fix the issue instead of requiring everyone who disabled pki reconciliation to create these new identities.

sjenning:
/verified by @sjenning

via e2e artifact inspection

PR #6830, opened 2025-09-15
OCPBUGS-56778: fix(kas): Disable PSA enforcement in 4.20

everettraven:
Disable PSA enforcement in 4.20

Related to https://github.com/openshift/api/pull/2475

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/v2/kas/config.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/v2/kas/config.go
@@ -117,7 +117,7 @@ func generateConfig(p KubeAPIServerConfigParams) (*kcpv1.KubeAPIServerConfig, er
 									Kind:       "PodSecurityConfiguration",
 								},
 								Defaults: podsecurityadmissionv1.PodSecurityDefaults{
-									Enforce:        "restricted",
+									Enforce:        "privileged",
 									EnforceVersion: "latest",
 									Audit:          "restricted",
 									AuditVersion:   "latest",


control-plane-operator/controllers/hostedcontrolplane/v2/kas/config.go
```diff
@@ -117,7 +117,7 @@ func generateConfig(p KubeAPIServerConfigParams) (*kcpv1.KubeAPIServerConfig, er
 									Kind:       "PodSecurityConfiguration",
 								},
 								Defaults: podsecurityadmissionv1.PodSecurityDefaults{
-									Enforce:        "restricted",
+									Enforce:        "privileged",
```
rtheis, line 120: I think that we need updates to the test fixtures too which I would think verify that we are indeed rendering the proper default PSA config for 4.20.

sjenning: I also thought this but Bryce ran `go test` with `UPDATE=true` and there was no diff.  Also `unit` didn't fail so :man_shrugging:  I guess not.

rtheis:
@everettraven verification appears to be failing on the commit title and body message.

```
1: CT1 Title does not start with one of fix, feat, chore, docs, style, refactor, perf, test, revert, ci, build: "kas: enforce privileged PSA by default"
3: B6 Body message is missing
```

control-plane-operator/controllers/hostedcontrolplane/v2/kas/config.go
rtheis: I wonder if an update to the OpenShift API go module that contains the default change would yield a unit test failure.

rtheis (approved):
/lgtm

sjenning:
`e2e-aws` failure is a real issue caused by misconfiguration in the presubmit job.

Do not retest until https://github.com/openshift/release/pull/69341 merges.

sdodson:
/retest-required
Changes to release repo should've fixed things

PR #6836, opened 2025-09-16
OCPBUGS-61776: fix(ignition-server): Don't delete user-managed cert secrets

matlaj:
Fix ignition server reconciliation deleting user-managed ignition server certificate secrets if the `disable-pki-reconciliation` annotation is present.

This prevented ignition server pods from starting up completely due to the missing secrets.

Fixes [OCPBUGS-61776](https://issues.redhat.com/browse/OCPBUGS-61776)

**Checklist**

### DIFF
--- a/support/controlplane-component/generic-adapter.go
+++ b/support/controlplane-component/generic-adapter.go
@@ -1,8 +1,13 @@
 package controlplanecomponent
 
 import (
+	"github.com/openshift/hypershift/support/config"
 	"github.com/openshift/hypershift/support/util"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+
+	capiutil "sigs.k8s.io/cluster-api/util"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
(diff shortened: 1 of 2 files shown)


bryan-cox:
looks good to me, happy to tag but I think @muraee should also take a look since this touches some CPOv2.

@matlaj do you plan on backporting this? If so the target backport versions on the Jira ticket need updated. 

Also, the release note will need updated on the Jira ticket. It's supposed to be in this format:
```
Cause – the user action or circumstances that trigger the bug
Consequence – what the user experience is when the bug occurs
Fix – what has changed to fix the bug; do not include overly technical details
Result – what happens now that the patch is applied
```

support/controlplane-component/generic-adapter.go
```diff
@@ -43,8 +50,11 @@ func (ga *genericAdapter) reconcile(cpContext ControlPlaneContext, obj client.Ob
 	workloadContext := cpContext.workloadContext()
 
 	if ga.predicate != nil && !ga.predicate(workloadContext) {
-		_, err := util.DeleteIfNeeded(cpContext, cpContext.Client, obj)
-		return err
+		if ga.keepManifest == nil || !ga.keepManifest(workloadContext) {
+			_, err := util.DeleteIfNeeded(cpContext, cpContext.Client, obj)
+			return err
+		}
+		return nil
```
muraee: I think something like this might a bit simpler and straightforward
```suggestion
		ownerRef := config.OwnerRefFrom(cpContext.HCP)
		// delete the object only if it has HCP ownerRef
		if util.HasOwnerRef(obj, ownerRef.Reference) {
			_, err := util.DeleteIfNeeded(cpContext, cpContext.Client, obj)
			return err
		}
```
wdyt?

matlaj: Agreed, see the new commit :)

muraee: thanks, can you please squash the commits

rtheis (approved):
/lgtm

rtheis (approved):
/lgtm

PR #6846, opened 2025-09-17
Implement Azure ResourceTags

fahlmant:
Implements ResourceTags for Azure to match the equivalent AWS capabilities. Specifically, tags to be set at the HostedCluster Platform level, as well as on individual NodePools. 

No Jira

N/A 

**Checklist**


bryan-cox:
Please convert this to a draft this to a draft PR until its ready for review so it doesn't trigger CI jobs on every push.

jparrill:
The `Verify` test is failing due to:
```
Linting commits from 6b857ec154a004563e0bf4df4493ecd646c32917 to 2d0334cc8d0741333d14352f3f1fa23e03c0940c (CI: PR targeting 6b857ec154a004563e0bf4df4493ecd646c32917)
1: CT1 Title does not follow ConventionalCommits.org format 'type(optional-scope): description': "Implement Azure ResourceTags"
3: B6 Body message is missing
```

bryan-cox:
> The `Verify` test is failing due to:
> 
> ```
> Linting commits from 6b857ec154a004563e0bf4df4493ecd646c32917 to 2d0334cc8d0741333d14352f3f1fa23e03c0940c (CI: PR targeting 6b857ec154a004563e0bf4df4493ecd646c32917)
> 1: CT1 Title does not follow ConventionalCommits.org format 'type(optional-scope): description': "Implement Azure ResourceTags"
> 3: B6 Body message is missing
> ```

Just adding on, this is documented in our upstream docs on how to submit a PR to our repo - https://hypershift.pages.dev/contribute/#prior-to-submitting-a-pull-request. There are some other things in there this PR still needs so I would suggest reviewing it and updating the PR. Please let us know if you have any questions or if we can help at all with that 😄 

@fahlmant I would also recommend splitting the changes out into multiple commits so its easier to review. Maybe something like:
- commit 1 - just the API change
- commit 2 - all the changes after running either `make verify`, `make update`, or `make api`
- commit 3 - the cli changes
- commit 4 - the nodepool controller changes.

api/hypershift/v1beta1/azure.go
```diff
@@ -98,6 +98,29 @@ type AzureNodePoolPlatform struct {
 	// If not specified, then Boot diagnostics will be disabled.
 	// +optional
 	Diagnostics *Diagnostics `json:"diagnostics,omitempty"`
+
+	// resourceTags is an optional list of additional tags to apply to Azure node
+	// instances. Changes to this field will be propagated in-place to Azure VM instances and their initial OS disks.
```
bryan-cox: > Changes to this field will be propagated in-place to Azure VM instances

Are we sure this is true? Has this been tested?

api/hypershift/v1beta1/azure.go
```diff
@@ -98,6 +98,29 @@ type AzureNodePoolPlatform struct {
...
+	// https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/tag-resources for
+	// information on tagging Azure resources. Most Azure resources support up to 50 tags
+	// per resource/resource group/subscription by default. Tag names can be up to 512 characters
+	// (though storage accounts limit tag names to 128 characters), tag values can be up to 256
+	// characters, and empty tag values are allowed. However, certain resource types including
+	// Automation, Content Delivery Network (CDN), Public/Private DNS zones, and Log Analytics
+	// saved searches remain limited to 15 tags total. OpenShift reserves 5 tags for its use,
+	// leaving up to 45 tags available for user use on most resources (10 tags on limited resources).
+	//
+	// +kubebuilder:validation:MaxItems=10
```
bryan-cox: Shouldn't the max items be higher than 10? 

What happens when more than 10 tags are applied to a resource that only accepts 10?

What are the 5 tags reserved for OpenShift for?

api/hypershift/v1beta1/azure.go
```diff
@@ -735,3 +780,24 @@ type AzureAuthenticationConfiguration struct {
 	// +optional
 	WorkloadIdentities *AzureWorkloadIdentities `json:"workloadIdentities,omitempty"`
 }
+
+// AzureResourceTag is a tag to apply to Azure resources created for the cluster.
+type AzureResourceTag struct {
+	// key is the key part of the tag. A tag key can have a maximum of 512 characters and cannot be empty. Key
+	// must begin with a letter, end with a letter, number or underscore, and must contain only alphanumeric
+	// characters and the following special characters `_ . -`.
+	//
```
bryan-cox: We should document in the API where these restrictions come from.

api/hypershift/v1beta1/azure.go
```diff
@@ -735,3 +780,24 @@ type AzureAuthenticationConfiguration struct {
...
+	// +required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=512
+	// +kubebuilder:validation:Pattern=`^[a-zA-Z]([0-9A-Za-z_.-]*[0-9A-Za-z_])?$`
+	Key string `json:"key"`
+	// value is the value part of the tag. A tag value can have a maximum of 256 characters and cannot be empty. Value
+	// must contain only alphanumeric characters and the following special characters `_ + , - . / : ; < = > ? @`.
```
bryan-cox, line 798: We should document in the API where these restrictions come from.

api/hypershift/v1beta1/azure.go
```diff
@@ -433,6 +456,28 @@ type AzurePlatformSpec struct {
...
+	// information on tagging Azure resources. Most Azure resources support up to 50 tags
+	// per resource/resource group/subscription by default. Tag names can be up to 512 characters
+	// (though storage accounts limit tag names to 128 characters), tag values can be up to 256
+	// characters, and empty tag values are allowed. However, certain resource types including
+	// Automation, Content Delivery Network (CDN), Public/Private DNS zones, and Log Analytics
+	// saved searches remain limited to 15 tags total. OpenShift reserves 5 tags for its use,
+	// leaving up to 45 tags available for user use on most resources (10 tags on limited resources).
+	// Changes to this field will be propagated in-place to Azure resources.
+	// These tags will be propagated to the infrastructure CR in the guest cluster, where other OCP operators might choose to honor this input to reconcile Azure resources created by them.
+	// Please consult the official documentation for a list of all Azure resources that support in-place tag updates.
```
bryan-cox: What official documentation? Azure? OpenShift?

api/hypershift/v1beta1/azure.go
```diff
@@ -98,6 +98,29 @@ type AzureNodePoolPlatform struct {
...
+	// characters, and empty tag values are allowed. However, certain resource types including
+	// Automation, Content Delivery Network (CDN), Public/Private DNS zones, and Log Analytics
+	// saved searches remain limited to 15 tags total. OpenShift reserves 5 tags for its use,
+	// leaving up to 45 tags available for user use on most resources (10 tags on limited resources).
+	//
+	// +kubebuilder:validation:MaxItems=10
+	// +listType=map
+	// +listMapKey=key
+	// +optional
+	ResourceTags []AzureResourceTag `json:"resourceTags,omitempty"`
```
bryan-cox, line 120: Will any changes to this field cause the a rollout of new nodes? If so that should be documented.

`// +rollout` should also be added here if so.

hypershift-operator/controllers/nodepool/azure.go
```diff
@@ -95,11 +95,42 @@ func azureMachineTemplateSpec(nodePool *hyperv1.NodePool) (*capiazure.AzureMachi
...
+	// Only set AdditionalTags if there are actually tags to avoid noisy diffs from empty maps
+	if tags := azureAdditionalTags(hostedCluster, nodePool); len(tags) > 0 {
+		azureMachineTemplate.Template.Spec.AdditionalTags = tags
+	}
+
 	return azureMachineTemplate, nil
 }
 
+// azureAdditionalTags merges tags from nodepool and cluster, with cluster tags taking precedence
+func azureAdditionalTags(hostedCluster *hyperv1.HostedCluster, nodePool *hyperv1.NodePool) map[string]string {
```
bryan-cox, line 107: Please add a unit test for this function.

bryan-cox, line 107: This automatically allocates a map even when there are no tags. There are efficiency gains when allocating the map to the same size as the number of tags like

```
func azureAdditionalTags(hostedCluster *hyperv1.HostedCluster, nodePool *hyperv1.NodePool) map[string]string {
	var nodePoolTags, clusterTags []hyperv1.AzureResourceTag

	// Extract tag slices
	if nodePool.Spec.Platform.Azure != nil {
		nodePoolTags = nodePool.Spec.Platform.Azure.ResourceTags
	}
	if hostedCluster.Spec.Platform.Azure != nil {
		clusterTags = hostedCluster.Spec.Platform.Azure.ResourceTags
	}

	totalSize := len(nodePoolTags) + len(clusterTags)
	if totalSize == 0 {
		return nil
	}

	// Pre-allocate with known capacity for optimal performance
	// This eliminates map resize operations and reduces memory allocations by ~73%
	tags := make(map[string]string, totalSize)

	// Add nodepool tags first
	for _, tag := range nodePoolTags {
		tags[tag.Key] = tag.Value
	}

	// Add cluster tags (which take precedence over nodepool tags)
	for _, tag := range clusterTags {
		tags[tag.Key] = tag.Value
	}

	return tags
}
```

I used a Claude golang agent to run a benchmark. This is the results of the new function vs the old:
```
Key Results:

  - Empty tags: 4.7x faster (7.69ns → 1.62ns)
  - Many tags (50): 1.7x faster with 42% less memory usage
  - Memory allocations: Reduced from 7 to 4 allocations (43% fewer)

  The Main Optimizations:

  1. Early return optimization:
  // Old: Always allocates map, checks later
  tags := make(map[string]string)
  // ... add tags ...
  if len(tags) == 0 { return nil }

  // New: Check first, allocate only if needed
  if totalSize == 0 { return nil }
  tags := make(map[string]string, totalSize)
  2. Pre-allocation with capacity:
  // Old: Map grows dynamically (expensive resizes)
  tags := make(map[string]string)

  // New: Single allocation with exact size
  tags := make(map[string]string, totalSize)

  Real-World Impact:

  - Empty tag scenarios (common): 4.7x faster
  - High tag count scenarios: 1.7x faster + 42% less memory
  - Reduced GC pressure: 43% fewer allocations
```

bryan-cox (commented):
Thanks for splitting this up in commits! That was really helpful.

I'm going to remove the NO-JIRA from the title of this PR. This change is large enough that we should have a Jira story behind this change documenting things like: who will test this change, why do we need this change, who is requesting this change, what do we gain from the change, etc.

api/hypershift/v1beta1/azure.go
fahlmant: Removed as this is not the case

api/hypershift/v1beta1/azure.go
fahlmant: Since this is for VMs in the NodePool, 49 tags are available with one being reserved by kubernetes, specifically `kubernetes.io/cluster`

api/hypershift/v1beta1/azure.go
fahlmant: Done

api/hypershift/v1beta1/azure.go
fahlmant: Done

hypershift-operator/controllers/nodepool/azure.go
fahlmant: Done

fahlmant: Agreed and changed

bryan-cox:
> Please convert this to a draft this to a draft PR until its ready for review so it doesn't trigger CI jobs on every push.

@fahlmant can you please convert this to draft when you have a moment?

fahlmant:
@bryan-cox it should be ready for review. I'm not sure why the e2e is failing as didn't touch AzureWorkloadIdentities. I rebased off of main
Edit: nvm I see what happened. Will fix

bryan-cox:
We typically strive to have all the reviews, required labels, and all tests passing on the PR before moving a PR out of draft - https://github.com/openshift/hypershift/blob/main/.github/CONTRIBUTING.md#creating-a-pull-request

fahlmant:
Things were passing locally. But sure, I'll mark as a draft. Can I tag you to review it before I switch it back?

bryan-cox:
You might need to rebase. Looks like `e2e-aks` failed because of 
```
cmd/infra/azure/create.go:306:53: result.WorkloadIdentities.Network undefined (type *"github.com/openshift/hypershift/api/hypershift/v1beta1".AzureWorkloadIdentities has no field or method Network)
```

api/hypershift/v1beta1/azure.go
```diff
@@ -98,6 +98,26 @@ type AzureNodePoolPlatform struct {
...
+	// per resource/resource group/subscription by default. Tag names can be up to 512 characters
+	// tag values can be up to 256 characters, and empty tag values are allowed.
+	// Kubernetes/OpenShift reserves 1 tag for its use,
+	// leaving up to 49 tags available for user use on instances.
+	//
+	// +kubebuilder:validation:MaxItems=49
+	// +listType=map
+	// +listMapKey=key
+	// +optional
+	ResourceTags []AzureResourceTag `json:"resourceTags,omitempty"`
```
bryan-cox, line 120: https://github.com/openshift/hypershift/pull/6846/files#r2364174798 still needs to be addressed

bryan-cox (commented):
Hey @fahlmant - thanks for the PR! Just a few things:
- Please follow the github pull request template and update the PR description
- Please identify which Jira ticket this is solving
- There seems to be one comment that wasn't addressed

Otherwise seems good to me.

api/hypershift/v1beta1/azure.go
bryan-cox: @fahlmant this comment is still unaddressed as well

api/hypershift/v1beta1/azure.go
fahlmant: The PR is still in draft, I'm testing this actively. In the OpenShift MachineSet directly, this does not cause a rollout of new nodes. I need to check within the NodePool controller if that behavior is different.

bryan-cox:
/close

Please feel free to reopen if this is still needed.

PR #6767, opened 2025-09-05
OCPBUGS-61218: test(e2e): add control plane component rollout validation to upgrade test

csrwng:
**What this PR does / why we need it**:
Add WaitForControlPlaneComponentRollout function to ensure control plane components complete rollout before proceeding with upgrade validation. This includes:
- New test step to wait for control plane component rollout completion
- Support for ControlPlaneComponent conditions in the eventually utility
- 30-minute timeout with 3-second polling interval for efficient validation

**Which issue(s) this PR fixes** *(optional, use `fixes #<issue_number>(, fixes #<issue_number>, ...)` format, where issue_number might be a GitHub issue, or a Jira story*:
Fixes #[OCPBUGS-61218](https://issues.redhat.com/browse/OCPBUGS-61218)

**Checklist**
- [x] Subject and description added to both, commit and PR.
- [x] Relevant issues have been referenced.
- [x] This change includes unit tests.

PR #6853, opened 2025-09-17
wait for component rollout in all cases

sjenning:
PR #6767 added a check for component rollout only in `TestUpgradeControlPlane` after the upgrade had been triggered.

However, we need to check this on _all_ rollouts in all tests. `WaitForImageRollout()` is the function used for this purpose.

Not waiting is causing the feature gate status to skew wrt the CVO version history.  On `TestUpgradeControlPlane`, the FeatureGate status is missing the `initial` version because the initial rollout is not completing before the upgrade begins.

This moves the new component rollout check into `WaitForImageRollout()` so that we do it for all HC rollouts in all tests.

My theory is that before #6767, both the FeatureGate status _and_ the CVO were missing the `initial` version and, thus, the test passed


wking:
/title wait for component rollout in all cases

wking:
/retitle wait for component rollout in all cases

PR #6859, opened 2025-09-18
OCPBUGS-61941: fix(ingress): add LoadBalancerSourceRanges support for external router service

muraee:

This enables CIDR restrictions for when kube-apiserver is using Router publishing strategy with External DNS where external clients access the kube-apiserver through the public router load balancer, providing the same access control via the kube-apiserver Loadbalancer service.

Changes:
- Apply LoadBalancerSourceRanges only to external (non-internal) router services
- Added comprehensive test coverage for all scenarios

Fixes #

**Checklist**

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
@@ -1393,6 +1393,7 @@ func (r *HostedControlPlaneReconciler) reconcileHCPRouterServices(ctx context.Co
 	if sharedingress.UseSharedIngress() || hcp.Spec.Platform.Type == hyperv1.IBMCloudPlatform {
 		return nil
 	}
+
 	// Create the Service type LB internal for private endpoints.
 	pubSvc := manifests.RouterPublicService(hcp.Namespace)
 	if util.IsPrivateHCP(hcp) {
--- a/control-plane-operator/controllers/hostedcontrolplane/ingress/router.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/ingress/router.go
@@ -4,6 +4,7 @@ import (
 	_ "embed"
 
 	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
+	"github.com/openshift/hypershift/support/azureutil"
 	"github.com/openshift/hypershift/support/util"
 
 	routev1 "github.com/openshift/api/route/v1"
@@ -60,6 +61,13 @@ func ReconcileRouterService(svc *corev1.Service, internal, crossZoneLoadBalancin
 			TargetPort: intstr.FromString("https"),
 			Protocol:   corev1.ProtocolTCP,
 		})
+	}
+
+	// Apply LoadBalancerSourceRanges for external router services to restrict CIDR access
+	// Only apply for external (non-internal) services and when not running on ARO HCP
+	allowedCIDRBlocks := util.AllowedCIDRBlocks(hcp)
+	if !internal && len(allowedCIDRBlocks) > 0 && !azureutil.IsAroHCP() {
+		svc.Spec.LoadBalancerSourceRanges = allowedCIDRBlocks
 	}
 
 	return nil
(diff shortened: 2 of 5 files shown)


control-plane-operator/controllers/hostedcontrolplane/ingress/router.go
```diff
@@ -62,6 +63,12 @@ func ReconcileRouterService(svc *corev1.Service, internal, crossZoneLoadBalancin
 		})
 	}
 
+	// Apply LoadBalancerSourceRanges for external router services to restrict CIDR access
+	// Only apply for external (non-internal) services and when not running on ARO HCP
+	if !internal && len(allowedCIDRBlocks) > 0 && !azureutil.IsAroHCP() {
```
jparrill, line 69: Should we just cover the AWS envs in this conditional?

control-plane-operator/controllers/hostedcontrolplane/ingress/router_test.go
```diff
@@ -50,3 +50,63 @@ func TestReconcileRouterServiceAnnotations(t *testing.T) {
 		t.Fatalf("expected target node labels annotation to be '%s', got %q", targetNodesLabel, got)
 	}
 }
+
+// Test that LoadBalancerSourceRanges is applied for external router services with allowedCIDRBlocks
+func TestReconcileRouterService_AppliesLoadBalancerSourceRanges(t *testing.T) {
+	// Given a HostedControlPlane on AWS
```
jparrill: If in the above conditional we only covers AWS, it's ok, but if not, should we add coverage for self-managed envs?

jparrill (commented):
Dropped some comments.

Thanks!

control-plane-operator/controllers/hostedcontrolplane/ingress/router.go
muraee: in ARO we use shared-ingress to restrict access

control-plane-operator/controllers/hostedcontrolplane/ingress/router_test.go
```diff
@@ -50,3 +50,63 @@ func TestReconcileRouterServiceAnnotations(t *testing.T) {
...
+			t.Fatalf("ReconcileRouterService returned error: %v", err)
+		}
+
+		// Then LoadBalancerSourceRanges should be set
+		if len(svc.Spec.LoadBalancerSourceRanges) != len(allowedCIDRs) {
+			t.Fatalf("expected LoadBalancerSourceRanges to have %d entries, got %d", len(allowedCIDRs), len(svc.Spec.LoadBalancerSourceRanges))
+		}
+
+		for i, expectedCIDR := range allowedCIDRs {
+			if svc.Spec.LoadBalancerSourceRanges[i] != expectedCIDR {
```
sjenning: nit: this depends on consistent ordering in `svc.Spec.LoadBalancerSourceRanges`

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -1393,12 +1393,19 @@ func (r *HostedControlPlaneReconciler) reconcileHCPRouterServices(ctx context.Co
 	if sharedingress.UseSharedIngress() || hcp.Spec.Platform.Type == hyperv1.IBMCloudPlatform {
 		return nil
 	}
+
+	// Extract allowedCIDRBlocks for restricting router service access
+	var allowedCIDRBlocks []string
+	for _, block := range util.AllowedCIDRBlocks(hcp) {
```
sjenning: The only other place `AllowedCIDRBlocks()` is called is here, and it does the same `CIDRBlock` to `string` slice conversion
https://github.com/openshift/hypershift/blob/168cc734abf7d727e4c446a2ff62a74798946cf2/control-plane-operator/controllers/hostedcontrolplane/kas/params.go#L22-L26

I think we should just return `[]string` from that function.  Then we can do away with this format conversion at both call sites _and_, since we already pass `hcp` to `ReconcileRouterService`, we can just call `AllowedCIDRBlocks()` inside there and not change the function signature.

muraee: done

muraee:

PR #6872, opened 2025-09-22
OCPBUGS-62056: test(e2e): fix flaky EnsureOAPIMountsTrustBundle with retry logic

jparrill:
The EnsureOAPIMountsTrustBundle test became flaky (89.47% success rate) after the removal of the kube-apiserver postStart handler in commit a9a6296ef. The handler previously provided implicit timing that allowed the certificate system to fully initialize before the test executed.

This change adds proper retry logic to wait up to 5 minutes for the `ca-bundle.crt` file to be available, eliminating the race condition and making the test more robust.

- Fixes #[OCPBUGS-62056](https://issues.redhat.com/browse/OCPBUGS-62056)

### DIFF
--- a/test/e2e/util/util.go
+++ b/test/e2e/util/util.go
@@ -876,8 +876,11 @@ func EnsureOAPIMountsTrustBundle(t *testing.T, ctx context.Context, mgmtClient c
 			}), "no volume named proxy-additional-trust-bundle found in openshift-apiserver pod")
 		}
 
-		_, err = RunCommandInPod(ctx, mgmtClient, "openshift-apiserver", hcpNs, command, "openshift-apiserver", 1*time.Minute)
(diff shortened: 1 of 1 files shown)


sjenning:
I'm not following.  The `postStart` handler was not removed, just refactored into code so the port could be modified.

https://github.com/openshift/hypershift/commit/a9a6296efefbcd2bdf3f13cca9b2f54cff2e1a18#diff-806d68cafd70a22f37b9733ffa680121e5d62f286fc00d9c5f0af9d2b2873500R114-R126

jparrill:
> The postStart handler was not removed, just refactored into code so the port could be modified

nvm, you're right. I think this was me, mixing PRs on browser. W/e the logic, in my mind look good, this + https://github.com/openshift/hypershift/pull/6843, should make it.

PR #6874, opened 2025-09-22
CNTRLPLANE-1371: feat(hostedcluster): implement service account signing key rotation with backup

muraee:
Enable safe rotation of service account signing keys by implementing backup functionality. When a signing key changes, the existing key is automatically backed up to a separate secret before the new key is applied.

Changes:
- Compare key bytes instead of just checking existence
- Create backup secret with '-backup' suffix when key rotation occurs
- Preserve existing behavior when keys haven't changed

This ensures hosted clusters can safely rotate signing keys without losing access to tokens signed by previous keys.
Fixes #

**Checklist**


devguyio:
/hold
I think this should be put on hold until we understand the requirements more from ARO-HCP

muraee:
@coderabbitai review

PR #6876, opened 2025-09-23
NO-JIRA: ci(dev-cluster): Fix dev token secret annotation

clebs:
The indentation of the annotation on the dev token secret was wrong, causing the template creation to fail.

Now the secret is created correctly when applying the template.

**Checklist**


jparrill:
/retitle NO-JIRA: ci(dev-cluster): Fix dev token secret annotation

PR #6897, opened 2025-09-25
OCPBUGS-60707: Fix user-ca-bundle not removed from guest cluster when additionalTrustBundle removed

enxebre:

This PR fixes a bug where the `user-ca-bundle` ConfigMap is not removed from the guest cluster when the `additionalTrustBundle` field is removed from the HostedCluster spec.

The issue was in the `reconcileUserCertCABundle` function in the hostedclusterconfigoperator which only handled the case where `AdditionalTrustBundle` is present, but did not clean up the ConfigMap when it's removed.

Fixes OCPBUGS-60707

- The fix adds proper cleanup logic using `util.DeleteIfNeeded` pattern consistent with other parts of the codebase
- Added comprehensive unit test coverage for the deletion scenario
- Enhanced e2e test to validate the user-ca-bundle is actually removed from the guest cluster
- The change ensures reconciliation logic properly handles both creation and cleanup scenarios

## Checklist:

🤖 Generated with [Claude Code](https://claude.ai/code) via `/jira-solve OCPBUGS-60707 enxebre`


enxebre:
/close 
in favour of https://github.com/openshift/hypershift/pull/6900

PR #6904, opened 2025-09-26
CNTRLPLANE-475: Enable Azure NodePool API happy path to pick image

bryan-cox:

This PR implements the Azure NodePool API marketplace image defaulting feature to enable explicit image selection using Azure Marketplace images for OCP >= 4.20.

**Key Features:**
- Add optional `ImageGeneration` field (Gen1/Gen2) to `AzureVMImage` struct in NodePool API
- Default to Gen2 when `ImageGeneration` is omitted
- Use "no-purchase-plan" channel from release payload marketplace metadata
- Version gating: only apply defaulting for OCP >= 4.20
- Explicit image input takes precedence over defaults
- No server-side Azure capability checks
- Full CLI support with `--image-generation` flag and validation

**Implementation:**
- Added custom `AzureVMImageGeneration` type with Gen1/Gen2 constants and kubebuilder validation
- Added `ImageGeneration *AzureVMImageGeneration` field to `AzureVMImage` struct in NodePool API
- Implemented version gating logic to only apply marketplace defaulting for OCP >= 4.20
- Added marketplace metadata parsing from real release payload CoreOSStreamMetadata
- Updated release info package to include Azure marketplace structures
- Honor client-specified Gen1/Gen2 via `ImageGeneration` field or CLI `--image-generation` flag
- CLI validation ensures only "Gen1" and "Gen2" values are accepted
- Comprehensive unit tests covering all scenarios including ARM64 architecture

**API Design:**
- `ImageGeneration` field is located within `AzureVMImage` struct (not platform level)
- Uses type-safe `AzureVMImageGeneration` string type instead of raw string
- Field is used solely by HyperShift for VM generation selection - CAPZ does not use this information
- Non-breaking change: existing explicit image configurations continue to work unchanged

Fixes [CNTRLPLANE-475](https://issues.redhat.com//browse/CNTRLPLANE-475)

- **Marketplace metadata support**: Now parses real marketplace metadata from release payload (no longer stubbed)
- **Architecture support**: Full AMD64 and ARM64 architecture support with comprehensive test coverage
- **CLI integration**: Complete CLI support with validation and error handling
- **Type safety**: Uses custom `AzureVMImageGeneration` type with proper constants
- **API placement**: `ImageGeneration` field correctly placed in `AzureVMImage` struct per feedback
- All acceptance criteria from the JIRA issue are fully implemented
- Tests verify correct behavior for version gating, generation selection, precedence rules, and both architectures

## Checklist:
- [x] This change includes docs (auto-generated API docs).
- [x] This change includes CLI support and validation.

🤖 Generated with [Claude Code](https://claude.ai/code) via `/jira-solve [CNTRLPLANE-475](https://issues.redhat.com//browse/CNTRLPLANE-475) bryan-cox`

### DIFF
--- a/hypershift-operator/controllers/nodepool/azure_test.go
+++ b/hypershift-operator/controllers/nodepool/azure_test.go
@@ -7,7 +7,11 @@ import (
 	. "github.com/onsi/gomega"
 
 	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
+	"github.com/openshift/hypershift/support/releaseinfo"
 
+	imageapi "github.com/openshift/api/image/v1"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/utils/ptr"
 
 	capiazure "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
@@ -473,7 +477,7 @@ func TestAzureMachineTemplateSpec(t *testing.T) {
 				},
 			},
 			expectedErr:    true,
-			expectedErrMsg: "either ImageID or AzureMarketplace needs to be provided for the Azure machine",
+			expectedErrMsg: "no Azure VM image configured",
 		},
 		{
 			name: "error case since a bad subnetID was provided",
@@ -510,7 +514,7 @@ func TestAzureMachineTemplateSpec(t *testing.T) {
 
 			azureSpec, err := azureMachineTemplateSpec(tc.nodePool)
 			if tc.expectedErr {
-				g.Expect(err.Error()).To(Equal(tc.expectedErrMsg))
+				g.Expect(err.Error()).To(ContainSubstring(tc.expectedErrMsg))
 			} else {
 				g.Expect(err).To(BeNil())
 				g.Expect(azureSpec).To(Equal(tc.expectedAzureMachineTemplateSpec))
@@ -523,6 +527,7 @@ func TestAzureMachineTemplate(t *testing.T) {
 	testCases := []struct {
 		name                   string
 		nodePool               *hyperv1.NodePool
+		releaseImage           *releaseinfo.ReleaseImage
 		templateNameGenerator  func(spec any) (string, error)
 		expectedTemplateName   string
 		expectedErr            bool
@@ -689,11 +694,12 @@ func TestAzureMachineTemplate(t *testing.T) {
 					},
 				},
 			},
+			releaseImage: createMockReleaseImage("4.20.0", false),
 			templateNameGenerator: func(spec any) (string, error) {
 				return "should-not-be-called", nil
 			},
 			expectedErr:    true,
-			expectedErrMsg: "failed to generate AzureMachineTemplateSpec: either ImageID or AzureMarketplace needs to be provided for the Azure machine",
+			expectedErrMsg: "no Azure VM image configured",
 		},
 		{
 			name: "When NodePool has encryption and ephemeral disk, it should create template with security configuration",
@@ -735,17 +741,26 @@ func TestAzureMachineTemplate(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			g := NewGomegaWithT(t)
 
+			// Use test-specific release image or default to one with marketplace metadata
+			releaseImg := tc.releaseImage
+			if releaseImg == nil {
+				releaseImg = createMockReleaseImage("4.20.0", true)
+			}
+
 			// Create a CAPI instance with minimal required fields
 			capi := &CAPI{
 				Token: &Token{
 					ConfigGenerator: &ConfigGenerator{
 						nodePool: tc.nodePool,
+						rolloutConfig: &rolloutConfig{
+							releaseImage: releaseImg,
+						},
 					},
 				},
 			}
 
 			// Call the method under test
-			template, err := capi.azureMachineTemplate(tc.templateNameGenerator)
+			template, err := capi.azureMachineTemplate(t.Context(), tc.templateNameGenerator)
 
 			if tc.expectedErr {
 				g.Expect(err).ToNot(BeNil())


hypershift-operator/controllers/nodepool/azure.go
```diff
@@ -99,6 +236,11 @@ func azureMachineTemplateSpec(nodePool *hyperv1.NodePool) (*capiazure.AzureMachi
 }
 
 func (c *CAPI) azureMachineTemplate(templateNameGenerator func(spec any) (string, error)) (*capiazure.AzureMachineTemplate, error) {
+	// Apply Azure Marketplace image defaults before generating machine template spec
+	if err := defaultAzureNodePoolImage(context.Background(), c.nodePool, c.ConfigGenerator.rolloutConfig.releaseImage); err != nil {
```
muraee: should pass and use a proper context here

cmd/cluster/azure/create.go
```diff
@@ -540,22 +547,12 @@ func NewCreateCommand(opts *core.RawCreateOptions) *cobra.Command {
 }
 
 func CreateInfraOptions(ctx context.Context, azureOpts *ValidatedCreateOptions, opts *core.CreateOptions) (azureinfra.CreateInfraOptions, error) {
-	rhcosImage := azureOpts.RHCOSImage
```
devguyio, line 470: what's the context behind removing this?

cmd/nodepool/azure/create_test.go
```diff
@@ -0,0 +1,264 @@
...
+			expectedError: "invalid value for --image-generation: Gen3. Supported values: Gen1, Gen2",
+		},
+		{
+			name:          "Invalid lowercase",
+			imageGen:      "gen1",
+			shouldError:   true,
+			expectedError: "invalid value for --image-generation: gen1. Supported values: Gen1, Gen2",
+		},
+		{
+			name:          "Invalid mixed case",
```
devguyio:
```suggestion
			name:          "Invalid upper case",
```

hypershift-operator/controllers/nodepool/azure.go
```diff
@@ -1,32 +1,169 @@
...
+	// Map hypershift architecture to RHCOS stream architecture
+	streamArch := arch
+	if arch == hyperv1.ArchitectureARM64 {
+		streamArch = "aarch64"
+	} else if arch == hyperv1.ArchitectureAMD64 {
+		streamArch = "x86_64"
+	}
+
+	// Extract marketplace metadata from release payload
+	azureMarketplace, err := getAzureMarketplaceMetadata(releaseImage, streamArch)
```
devguyio, line 84: I think we should pass the generation that was asked for here, and error if the releaseImage doesn't have that
generation. Unless that validation is already happening somewhere else?

cmd/cluster/azure/create.go
bryan-cox: We don't need it. Using an image uploaded to a VHD is not a production supported way of hosting the VM images in Azure. This is legacy code leftover. There is a Jira ticket to remove all related code that I will follow up on after this PR.

The production way of hosting your VM images is either marketplace or galleries.

api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/nodepools.hypershift.openshift.io/OpenStack.yaml
```diff
@@ -792,6 +792,20 @@ spec:
...
                             - version
                             type: object
+                          imageGeneration:
+                            default: Gen2
+                            description: |-
+                              imageGeneration specifies the Hyper-V generation of the Azure Marketplace image to use for the nodes.
+                              This field is only used by HyperShift to select the appropriate marketplace image
+                              from the release payload metadata. It is not passed to CAPZ (Cluster API Provider Azure).
+                              The generation information is encoded into the SKU field that CAPZ uses.
+                              This field is only relevant when Type is unset or when defaults are applied from the release payload.
```
csrwng: This says that it's only relevant when Type is unset. However, Type cannot be unset because it's a required field.

api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/nodepools.hypershift.openshift.io/AAA_ungated.yaml
```diff
@@ -816,10 +830,9 @@ spec:
                             otherwise
                           rule: 'has(self.type) && self.type == ''ImageID'' ?  has(self.imageID)
                             : !has(self.imageID)'
-                        - message: azureMarketplace is required when type is RequiredMember,
-                            and forbidden otherwise
-                          rule: 'has(self.type) && self.type == ''AzureMarketplace''
-                            ?  has(self.azureMarketplace) : !has(self.azureMarketplace)'
+                        - message: azureMarketplace is forbidden when type is ImageID
+                          rule: 'self.type == ''ImageID'' ?  !has(self.azureMarketplace)
+                            : true'
```
csrwng: With this new validation, it means that it's now possible for type == 'AzureMarketplace' and the azureMarketplace structure to not be populated. Is this intended?

bryan-cox: The validation was intentionally loosened to support a third image configuration mode for OCP 4.20+, where marketplace image metadata comes from the release payload rather than explicit CLI flags.
Three supported image configuration scenarios:

1. Explicit ImageID: User provides --image-id → imageID must be set, azureMarketplace must not be set
2. Explicit Marketplace: User provides --marketplace-publisher/offer/sku/version → azureMarketplace must be populated
3. Release Payload Default (new): User provides only --image-generation → type=AzureMarketplace but azureMarketplace is initially empty, controller populates from release payload

The CRD validation now only enforces:
- When type == ImageID, azureMarketplace must NOT exist
- When type == AzureMarketplace, azureMarketplace MAY be empty (populated later by controller)

Runtime validation (in azureMachineTemplateSpec()) ensures that after controller defaulting, either imageID or a populated azureMarketplace exists before creating CAPI resources.

Why we can't use strict bidirectional CRD validation:
The CLI sets type = AzureMarketplace even when marketplace flags aren't provided (see cmd/nodepool/azure/create.go:189-197). Strict validation would reject this valid scenario where the controller fills in marketplace data from the
release payload using the imageGeneration hint."

api/hypershift/v1beta1/azure.go
```diff
@@ -128,6 +139,18 @@ type AzureVMImage struct {
 	// +optional
 	// +unionMember
 	AzureMarketplace *AzureMarketplaceImage `json:"azureMarketplace,omitempty"`
+
+	// imageGeneration specifies the Hyper-V generation of the Azure Marketplace image to use for the nodes.
+	// This field is used by HyperShift to select the appropriate marketplace image (Gen1 or Gen2)
+	// from the release payload metadata when azureMarketplace is not explicitly provided.
```
csrwng: suggestion:
s/when azureMarketplace is not explicitly provided./when Type is AzureMarketplace and azureMarketPlace data is not provided/

api/hypershift/v1beta1/azure.go
```diff
@@ -128,6 +139,18 @@ type AzureVMImage struct {
...
+	// It is not passed to CAPZ (Cluster API Provider Azure); the generation information is
+	// encoded into the SKU field that CAPZ uses.
+	// This field has no effect when explicit imageID or azureMarketplace is set.
```
enxebre: >// This field has no effect when explicit imageID or azureMarketplace is set.

Then mutually exclusion needs to be enforced via cel.
Or should this field belong to AzureMarketplaceImage struct and be mutually exclusive with their other fields?
Or should this be part of a new struct for a new AzureVMImage.type = managed?
cc @JoelSpeed 

Also, why do we need to support gen1 at all?

api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/nodepools.hypershift.openshift.io/AAA_ungated.yaml
enxebre: the cli is completely irrelevant for discussing good API design. Let's talk api. If type= AzureMarketplace an with empty AzureMarketplaceImage struct is valid API input, this need to be documented in .type and .AzureMarketplace. How ever I question that's what we want to do https://github.com/openshift/hypershift/pull/6904/files#r2423810073

api/hypershift/v1beta1/azure.go
bryan-cox: Not all VM sizes are gen2

api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/nodepools.hypershift.openshift.io/AAA_ungated.yaml
JoelSpeed: I think what you're describing here is considered an optional union member, so the following are both valid
```
type: AzureMarketplace
```
```
type: AzureMarketplace
azureMarketplace:
  publisher: ...
  ...
```

This is something we generally support, but as Alberto says, you need to change the documentation on the fields to explain this

We also use the CEL 
```
// +kubebuilder:validation:XValidation:rule="has(self.type) && self.type == 'AzureMarketplace' ?  true : !has(self.azureMarketplace)",message="azureMarketplace is forbidden when type is not AzureMarketplace"
```
This way your CEL is tied to the `AzureMarketplace` option, and not `ImageID`. If you added a third option later, your current CEL would be invalid and need to be changed. This suggestion would be correct even for more values later

api/hypershift/v1beta1/azure.go
```diff
@@ -102,7 +113,7 @@ type AzureNodePoolPlatform struct {
 
 // AzureVMImage represents the different types of boot image sources that can be provided for an Azure VM.
 // +kubebuilder:validation:XValidation:rule="has(self.type) && self.type == 'ImageID' ?  has(self.imageID) : !has(self.imageID)",message="imageID is required when type is ImageID, and forbidden otherwise"
-// +kubebuilder:validation:XValidation:rule="has(self.type) && self.type == 'AzureMarketplace' ?  has(self.azureMarketplace) : !has(self.azureMarketplace)",message="azureMarketplace is required when type is RequiredMember, and forbidden otherwise"
+// +kubebuilder:validation:XValidation:rule="self.type == 'ImageID' ?  !has(self.azureMarketplace) : true",message="azureMarketplace is forbidden when type is ImageID"
```
JoelSpeed: See https://github.com/openshift/hypershift/pull/6904/files/26127e94d22f9185d38278ad7af2beeb4a166d45#r2425819364
```suggestion
// +kubebuilder:validation:XValidation:rule="has(self.type) && self.type == 'AzureMarketplace' ?  true : !has(self.azureMarketplace)",message="azureMarketplace is forbidden when type is not AzureMarketplace"
```

api/hypershift/v1beta1/azure.go
JoelSpeed: Looks to me like this field should be part of the `AzureMarketplaceImage` struct based on reading the description.

What's not clear right here is that users can put have only one of the three following states:
```
azureMarketplace:
  imageGeneration: ...
```
```
azureMarketplace:
  imageGeneration: ...
  publisher: ...
  offer: ...
  sku: ...
  version: ...
```
```
azureMarketplace:
  publisher: ...
  offer: ...
  sku: ...
  version: ...
```

I think what you need to do here is update the optionality on the `AzureMarketplaceImage` struct. Move this field in there, make all of the fields optional. Then add:

```
// +kubeuilder:validation:MinProperties:=1
// +kubebuilder:validation:XValidation:rule="has(self.imageGeneration) || has(self.publisher) && has(self.offer) && has(self.sku) && has(self.version)",message="must specify imageGeneration or must specify publisher, offer, sku and version"
// +kubebuilder:validation:XValidation:rule="[has(self.publisher), has(self.offer), has(self.sku), has(self.version)].filter(x, x == true).size() == 0 || [has(self.publisher), has(self.offer), has(self.sku), has(self.version)].filter(x, x == true).size() == 4",message="publisher, offer, sku and version must either be all set, or all omitted" 
```

bryan-cox:
@enxebre @JoelSpeed - I'm testing out https://github.com/openshift/hypershift/pull/6996 to address PR feedback automatically so if you see some weird comments soon, that's where it's coming from 😅

api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/nodepools.hypershift.openshift.io/AAA_ungated.yaml
bryan-cox: [AUTOGENERATED AI RESPONSE]

This feedback has been addressed in commit b7734a39d.

Changes made:
- Updated CEL validation rule to explicitly check for AzureMarketplace type  
- Changed from `self.type == 'ImageID' ? !has(self.azureMarketplace) : true`
- To: `has(self.type) && self.type == 'AzureMarketplace' ? true : !has(self.azureMarketplace)`
- This makes the validation more future-proof against additional image type additions

Please review and let me know if further changes are needed.

cmd/nodepool/azure/create.go
```diff
@@ -70,6 +71,7 @@ func bindCoreOptions(opts *RawAzurePlatformCreateOptions, flags *pflag.FlagSet)
 	flags.StringVar(&opts.DiskStorageAccountType, "disk-storage-account-type", opts.DiskStorageAccountType, "The disk storage account type for the OS disks for the VMs.")
 	flags.StringVar(&opts.SubnetID, "nodepool-subnet-id", opts.SubnetID, "The subnet id where the VMs will be placed.")
 	flags.StringVar(&opts.ImageID, "image-id", opts.ImageID, "The Image ID to boot the VMs with.")
+	flags.StringVar(&opts.ImageGeneration, "image-generation", opts.ImageGeneration, "The Hyper-V generation of the Azure VM image. Supported values: Gen1, Gen2. If unspecified, defaults to Gen2.")
```
csrwng, line 74: should flag validation fail if this is specified and the other marketplace-* options are specified as well?

test/e2e/create_cluster_test.go
```diff
@@ -1752,6 +1752,102 @@ func TestOnCreateAPIUX(t *testing.T) {
...
+										ImageGeneration: ptr.To(hyperv1.Gen2),
+									},
+								},
+								OSDisk: hyperv1.AzureNodePoolOSDisk{
+									DiskStorageAccountType: hyperv1.DiskStorageAccountTypesPremiumLRS,
+								},
+								SubnetID: "/subscriptions/12345678-1234-5678-9012-123456789012/resourceGroups/test-rg/providers/Microsoft.Network/virtualNetworks/test-vnet/subnets/test-subnet",
+							}
+						},
+						expectedErrorSubstring: "",
```
csrwng, line 1836: It looks like this should be populated if you expect an error

csrwng: It would also be good to test failures when you only specify some of the fields of the structure like Publisher and Offer but not SKU

bryan-cox:
@csrwng Re: CLI validation for --image-generation with marketplace flags:

The current behavior is intentional - `--image-generation` can be used in combination with marketplace flags. The API supports three patterns:

1. Only `--image-generation` → uses defaults from release payload
2. Only marketplace flags (publisher/offer/sku/version) → explicit image with Gen2 default  
3. Both → explicit image with explicit generation

This flexibility is documented in the API at api/hypershift/v1beta1/azure.go:151 which states marketplace images can be used "with optional imageGeneration". No validation conflict needed here.

---

bryan-cox:
@csrwng Re: Test coverage improvements (comments 2452285459 and 2452289624):

Done. Added comprehensive test cases for partial field validation:

1. **Fixed empty expectedErrorSubstring**: Changed test name from "should fail" to "should pass" since the combination is intentionally valid (test/e2e/create_cluster_test.go:1764)

2. **Added partial field validation tests** (test/e2e/create_cluster_test.go:1849-1911):
   - Only publisher+offer specified (missing sku+version) → fails
   - Only SKU specified (missing other fields) → fails  
   - Publisher+offer+SKU without version → fails

These tests verify the API's CEL validation that enforces "all-or-nothing" for the publisher/offer/sku/version quartet, ensuring users can't accidentally create invalid marketplace image references.

---

PR #6924, opened 2025-10-02
test: update external-oidc e2e tests to dynamically determine tests to run

everettraven:

This PR updates the external OIDC tests so that they dynamically run based on the testing environment in which they are run.

This more closely simulates the behavior of the standalone openshift tests where tests can be skipped or included based on feature gate enablement.

This makes it easier to have both a techpreview and default Prow job that will run these tests and makes it easier for future techpreview work on this feature to be tested without having the extra step of adding back a techpreview job while ensuring that we also capture any regressions from change introduced in TPNU before they are promoted to the Default feature set.

N/A

## Checklist:

### DIFF
--- a/test/e2e/external_oidc_test.go
+++ b/test/e2e/external_oidc_test.go
@@ -20,13 +20,12 @@ import (
 	kauthnv1typedclient "k8s.io/client-go/kubernetes/typed/authentication/v1"
 
 	crclient "sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/openshift/hypershift/control-plane-operator/featuregates"
 )
 
 func TestExternalOIDC(t *testing.T) {
 	e2eutil.AtLeast(t, e2eutil.Version419)
-	if os.Getenv("TECH_PREVIEW_NO_UPGRADE") != "true" {
-		t.Skipf("Only tested when CI sets TECH_PREVIEW_NO_UPGRADE=true and the Hypershift Operator is installed with --tech-preview-no-upgrade")
-	}
 
 	if globalOpts.ExternalOIDCProvider == "" {
 		t.Skipf("skip external OIDC test if e2e.external-oidc-provider is not provided")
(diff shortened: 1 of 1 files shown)


everettraven:
Merging as is will break existing periodic job. Holding for further discussion

/hold

sjenning:
/lgtm

everettraven:
/payload-job periodic-ci-openshift-hypershift-release-4.21-periodics-e2e-aws-external-oidc

everettraven:
/verified by periodic-ci-openshift-hypershift-release-4.21-periodics-e2e-aws-external-oidc

test/e2e/external_oidc_test.go
```diff
@@ -38,9 +37,29 @@ func TestExternalOIDC(t *testing.T) {
 	defer cancel()
 
 	clusterOpts := globalOpts.DefaultClusterOptions(t)
+	clusterOpts.NodePoolReplicas = 1
+	clusterOpts.FeatureSet = string(configv1.Default)
+
+	if os.Getenv("TECH_PREVIEW_NO_UPGRADE") == "true" {
+		clusterOpts.FeatureSet = string(configv1.TechPreviewNoUpgrade)
+	}
```
xingxingxia, line 45: @everettraven , didn't notice this PR until now. For `func TestExternalOIDC` to be used for non-tp job, not sure why explicitly adding line 43 ~ 44? CC @xiuwang FYI

everettraven: @xingxingxia This is so that the same test method can be used for testing both the default and tech preview feature set so that we have a single entrypoint for both TP and non-TP jobs.

test/e2e/external_oidc_test.go
```diff
@@ -38,9 +37,29 @@ func TestExternalOIDC(t *testing.T) {
...
+
+	featuregates.ConfigureFeatureSet(clusterOpts.FeatureSet)
```
xingxingxia, line 47: @everettraven , another comment: just found https://github.com/openshift/hypershift/blob/05a2fcdd598f4c18637177e7aeddaf9786ce2f96/control-plane-operator/featuregates/featuregates.go#L43-L47 says `ConfigureFeatureSet should only be called once on startup` : ). Calling it within a Testxxxx func modifies this `global` variable [var globalGate featuregate.FeatureGate](https://github.com/openshift/hypershift/blob/05a2fcdd598f4c18637177e7aeddaf9786ce2f96/control-plane-operator/featuregates/featuregates.go#L34 ) , i.e. it can affect other Testxxx funcs even if other funcs might not use the same gates as the `func TestExternalOIDC` tests.

PR #6917, opened 2025-09-30
OCPBUGS-62226: fix: Add configurable load balancer health probe optio…

mgencur:
…ns for AWS and Azure

## What this PR does / why we need it:

Add support for configuring load balancer health probe settings in both AWS and Azure cloud providers:

- Add ClusterServiceLoadBalancerHealthProbeMode (Shared/ServiceNodePort for AWS, shared/servicenodeport for Azure)
- Add ClusterServiceSharedLoadBalancerHealthProbePort for configuring probe port
- Add ClusterServiceSharedLoadBalancerHealthProbePath for configuring probe path
- Update CLI flags: --load-balancer-health-probe-mode, --shared-load-balancer-health-probe-port, --shared-load-balancer-health-probe-path
- Update cloud controller manager configurations for both platforms
- Regenerate CRDs with new fields
- This PR fixes OCPBUGS-62226 in the way that it is possible to configure the older mode "ServiceNodePort" which works with Cilium.

This enables fine-tuned control over health probe behavior for cluster service load balancers, allowing operators to choose between shared probe endpoints or service node port probes based on their requirements.

## Which issue(s) this PR fixes:

Fixes [OCPBUGS-62226](https://issues.redhat.com/browse/OCPBUGS-62226)

## Checklist:
- [x] Subject and description added to both, commit and PR.
- [x] Relevant issues have been referenced.
- [x] This change includes docs. 
- [ ] This change includes unit tests.

---

> [!NOTE]
> Adds fields and CLI flags to configure cluster service load balancer health probe mode/port/path for AWS and Azure, and propagates them into cloud controller manager configs with docs/CRDs/tests updated.
> 
> - **API / CRDs**:
>   - Add `ClusterServiceLoadBalancerHealthProbeMode`, `ClusterServiceSharedLoadBalancerHealthProbePort`, `ClusterServiceSharedLoadBalancerHealthProbePath` to `api/hypershift/v1beta1/{aws,azure}.go` with validation/defaults; regenerate CRDs and vendor copies.
> - **CLI**:
>   - New flags in `cmd/cluster/{aws,azure}/create.go`: `--load-balancer-health-probe-mode`, `--shared-load-balancer-health-probe-port`, `--shared-load-balancer-health-probe-path`; validate and set `HostedCluster.Spec.Platform.{AWS,Azure}`.
> - **Control Plane Operator**:
[...]

PR #6927, opened 2025-10-03
OCPBUGS-62226): fix: Use annotations for AWS and Azure load balancer health probe configuration

mgencur:
(this is an alternative to https://github.com/openshift/hypershift/pull/6917 , using annotations instead of API config)

Add support for configuring load balancer health probe settings via annotations
for both AWS and Azure platforms. This provides users with a flexible way to
customize health probe behavior without using API fields.

Changes:
- Add annotation constants for AWS and Azure health probe configuration:
  * {aws,azure}-load-balancer-health-probe-mode
  * shared-load-balancer-health-probe-path
  * shared-load-balancer-health-probe-port

- Enable annotation propagation from HostedCluster to HostedControlPlane
  for all six new annotations via mirroredAnnotations list

- Update cloud controller manager config adapters to read from annotations:
  * AWS: control-plane-operator/.../aws/config.go
  * Azure: control-plane-operator/.../azure/config.go
  * Add validation for mode values and port ranges (1-65535)
  * Path and port annotations only apply when mode is "Shared"/"shared"

Platform differences:
- AWS uses capitalized mode values: "Shared", "ServiceNodePort"
- Azure uses lowercase mode values: "shared", "servicenodeport"

Co-Authored-By: Claude <noreply@anthropic.com>

Fixes  [OCPBUGS-62226](https://issues.redhat.com/browse/OCPBUGS-62226)

## Checklist:


control-plane-operator/controllers/hostedcontrolplane/v2/cloud_controller_manager/azure/config.go
```diff
@@ -79,6 +85,40 @@ func azureConfig(cpContext component.WorkloadContext, withCredentials bool) (Azu
...
+			portNum, err := strconv.Atoi(portStr)
+			if err != nil {
+				return AzureConfig{}, fmt.Errorf("invalid value for annotation %s: %s (must be a valid port number)", hyperv1.SharedLoadBalancerHealthProbePortAnnotation, portStr)
+			}
+			if portNum < 1 || portNum > 65535 {
+				return AzureConfig{}, fmt.Errorf("invalid value for annotation %s: %d (must be between 1 and 65535)", hyperv1.SharedLoadBalancerHealthProbePortAnnotation, portNum)
+			}
+			probePort = int32(portNum)
+		}
+	}
```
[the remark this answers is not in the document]
mgencur: Added unit tests.

control-plane-operator/controllers/hostedcontrolplane/v2/assets/aws-cloud-controller-manager/config.yaml
```diff
@@ -6,7 +6,9 @@ data:
     VPC = %s
     KubernetesClusterID = %s
     SubnetID = %s
-    ClusterServiceLoadBalancerHealthProbeMode = Shared
+    ClusterServiceLoadBalancerHealthProbeMode = %s
+    ClusterServiceSharedLoadBalancerHealthProbePath = %s
```
csrwng: Having these here, and having to set a default value forces us to keep our default in sync with the values in the upstream aws provider code. Could we just not set them if they're not specified via annotation?

mgencur: We must always set the value, in the same way as [cluster-cloud-controller-manager-operator does](https://github.com/openshift/cluster-cloud-controller-manager-operator/blob/2dc5d4b1d8fb5de40563e5c38179f39293f23c80/pkg/cloud/aws/aws_config_transformer.go#L97). I believe we want the same default as standalone OpenShift.
This is because the default value in cloud-provider-aws [is still ServiceNodePort](https://github.com/openshift/cloud-provider-aws/blob/fa4e6bb70f444336072ff7c44f20afa38b2a3785/pkg/providers/v1/aws.go#L2115)

mgencur: Alright. This was about the other two annotations. Thanks for review. Fixed in a new commit.

PR #6932, opened 2025-10-06
OCPBUGS-62172: Identify OpenStack CSI apps via their correct app labels

mandre:
According to the configs in csi-operator [1] and [2], the correct app labels should be `openstack-manila-csi` and
`openstack-cinder-csi-driver-controller`.

[1] https://github.com/openshift/csi-operator/blob/12ef6e82b4cdf0cb45b43056a81dfdf9fbd833ff/assets/overlays/openstack-manila/generated/hypershift/controller.yaml#L52
[2] https://github.com/openshift/csi-operator/blob/12ef6e82b4cdf0cb45b43056a81dfdf9fbd833ff/assets/overlays/openstack-cinder/generated/hypershift/controller.yaml#L53

Fixes OCPBUGS-62172

## Checklist:


mandre:
Wrong fix.

stephenfin:
Correct fix is here https://github.com/openshift/cluster-storage-operator/pull/628. We may want something like this in the future but it should _add_ the additional pods, not replace the operator ones.

PR #6951, opened 2025-10-07
CNTRLPLANE-1581: feat(azure): relax API to make externalDNS optional for self-managed clusters

bryan-cox:

This PR removes strict Azure platform validation rules that previously required all service publishing strategies to use Route type with defined hostnames. The removed validations enforced that APIServer, OAuthServer, Konnectivity, and Ignition services must have Route service publishing strategies with hostnames, which was too restrictive for self-managed Azure clusters.

The changes include:
- Removed 4 XValidation rules from `HostedClusterSpec` that mandated Route service publishing with hostnames for Azure platform
- Allows self-managed Azure clusters to use alternative service publishing strategies (like LoadBalancer or NodePort)
- Maintains compatibility with managed Azure clusters that still use Route-based publishing
- Removes generated client listers as part of code generation cleanup

This enables more flexible deployment patterns for Azure self-managed clusters while maintaining the existing behavior for managed clusters.

Fixes [CNTRLPLANE-1581](https://issues.redhat.com//browse/CNTRLPLANE-1581)

- The validation rules were overly restrictive and prevented valid self-managed Azure cluster configurations
- This change only affects the API validation layer - existing managed clusters continue to work unchanged
- The deleted client lister files are generated code that will be regenerated during the build process
- Please verify that the removed validations don't break any existing Azure managed cluster functionality

## Checklist:


bryan-cox:
/verified by @bryan-cox 

I manually tested this for self-managed Azure. This also successfully tested with the aks e2es.

PR #6958, opened 2025-10-08
OCPBUGS-61164: Add per-pod crash toleration

clebs:
E2E tests are flaky because at times pods restart due to slow infra and the tests is then considered as failed as soon as a pod is restarted once.
In some cases a pod restart does not immediately mean that the pod is crashing, nor that the test has failed.

This PR adds the ability to set custom crash tolerations on a per-pod basis and adds the first usage for the `aws-ebs-csi-driver-controller` pod that is causing flakiness.

Fixes OCPBUGS-61164

## Checklist:


test/e2e/util/util.go
sjenning: Would be nice to collapse all this into the new logic as well.  Just set the crash toleration to some high value like 20.

clebs: Right! Should have thought of that, much cleaner.

clebs: @sjenning I have applied the requested change.

Additionally, took the chance to consolidate imports that where imported several times with different names and did some minor improvement.

muraee: > Additionally, took the chance to consolidate imports that where imported several times with different names and did some minor improvemen

@clebs Can you please seperate this into a different commit to make it easier to review the actual changes this PR introduces.

clebs: @muraee of course! I have split the commit in two, first the fix and second the cleanup/improvements.

test/e2e/util/util.go
```diff
@@ -45,13 +44,10 @@ import (
...
 	"github.com/aws/aws-sdk-go/service/route53"
 
-	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
 	k8sadmissionv1 "k8s.io/api/admissionregistration/v1"
 	appsv1 "k8s.io/api/apps/v1"
 	authenticationv1 "k8s.io/api/authentication/v1"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	kapierror "k8s.io/apimachinery/pkg/api/errors"
```
muraee: let's keep `apierrors` here, we use this import alias across all our codebase

clebs: changed thanks!

test/e2e/util/util.go
```diff
@@ -81,32 +76,59 @@ import (
...
+		regexp.MustCompile("-catalog"):           20,
+
+		// Temporary workaround for https://issues.redhat.com/browse/OCPBUGS-45182
+		regexp.MustCompile("^openstack-manila-csi-controllerplugin-"): 20,
+
+		// Temporary workaround for https://issues.redhat.com/browse/CNV-40820
+		regexp.MustCompile("^kubevirt-csi"):                  20,
+		regexp.MustCompile("^aws-ebs-csi-driver-controller"): 1,
+	}
+)
```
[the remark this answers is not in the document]
clebs: @sjenning @muraee what do you think about this suggestion?

I think that if we have overlapping regexes we already did something wrong and relying on the order of evaluation for correctness would cause very subtle bugs.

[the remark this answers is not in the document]
muraee: @clebs you are correct, but honestly I am not a fan of iterating over a map of regex as well. 
Could we match the pods based on labels instead of their name? the map keys could be the value of the pod's `app` label for example. All pods should have that label afaik.

@coderabbitai thoughts?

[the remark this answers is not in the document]
clebs: Agreed, that is a good idea. I will evaluate the possibility. My one concern is that we might not have the desired granularity: several pods with the same label could not be configured individually.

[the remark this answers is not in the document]
clebs: @muraee I have checked all labels from the pods on the list, we do not have any granularity issues and will likely never have them looking at everything in general.

I have applied the requested changes, thanks this is a better solution.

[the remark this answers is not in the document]
muraee: thanks lgtm, could you please squash your commits apart from `consolidate imports + minor improvements`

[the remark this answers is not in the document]
clebs: Done!

muraee:
lgtm, leaving it to @sjenning to tag

clebs:
/verified bypass because this introduces a concept to reduce flakiness which will be verified in time. The passing e2e tests at the very least prove that it does not introduce regressions and flakiness will stay as before in the worst case.

clebs:
/verified bypass

This introduces a concept to reduce flakiness which will be verified in time. The passing e2e tests at the very least prove that it does not introduce regressions and flakiness will stay as before in the worst case.

PR #6965, opened 2025-10-08
OCPBUGS-62851: feat(e2e): add check for Service and Pod Monitors

sjenning:
We need to catch when Service and Pod Monitor targets are down

---

> [!NOTE]
> Introduces a health check to catch down metrics targets during cluster creation.
> 
> - Adds `EnsureNoMetricsTargetsAreDown` in `test/e2e/util/util.go` to list `ServiceMonitors`/`PodMonitors`, derive jobs, and query Prometheus `up{namespace="<hcp>"}` with retries; skips autoscaler PodMonitor when no pods match
> - Registers Prometheus Operator `monitoring.v1` types in `test/e2e/util/scheme.go`
> - Invokes the new check in `TestCreateCluster` after existing validations
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3affe3825d89ac1016b97f44b85efc7b62e08243. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/test/e2e/util/util.go
+++ b/test/e2e/util/util.go
@@ -72,6 +72,7 @@ import (
 	"github.com/go-logr/logr"
 	"github.com/go-logr/zapr"
 	"github.com/google/go-cmp/cmp"
+	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	prometheusv1 "github.com/prometheus/client_golang/api/prometheus/v1"
 	dto "github.com/prometheus/client_model/go"
 	"github.com/prometheus/common/expfmt"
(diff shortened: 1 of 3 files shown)


test/e2e/util/util.go
```diff
@@ -3855,3 +3856,74 @@ func EnsureCNOOperatorConfiguration(t *testing.T, ctx context.Context, mgmtClien
...
+			if len(warnings) > 0 {
+				t.Logf("Prometheus query warnings for %s: %v", sm.Name, warnings)
+			}
+
+			// Parse the results
+			if result.Type() == model.ValVector {
+				vector := result.(model.Vector)
+				if len(vector) == 0 {
+					t.Logf("No targets found for ServiceMonitor %s (this may be expected if scraping hasn't started yet)", sm.Name)
+					continue
```
wking: looks like we're failing open here?  Can we require a successful `up` query?

test/e2e/util/util.go
```diff
@@ -3855,3 +3856,74 @@ func EnsureCNOOperatorConfiguration(t *testing.T, ctx context.Context, mgmtClien
...
+
+		// Track down targets
+		var downTargets []string
+
+		// Check each ServiceMonitor's targets
+		for _, sm := range serviceMonitors.Items {
+			// Query Prometheus for the targets associated with this ServiceMonitor
+			// The 'up' metric is a special Prometheus metric that indicates target health
+			// up == 1 means the target is healthy, up == 0 means it's down
+			query := fmt.Sprintf(`up{job="%s",namespace="%s"}`, sm.Name, hcpNamespace)
```
wking: We should be able to make a single `up{namespace="%s"}` request and then iterate over the successful result to assert that we find all the expected ServiceMonitors and that they're happy.

test/e2e/util/util.go
```diff
@@ -3855,3 +3856,74 @@ func EnsureCNOOperatorConfiguration(t *testing.T, ctx context.Context, mgmtClien
...
+
+			result, warnings, err := prometheusClient.Query(ctx, query, time.Now())
```
joshbranham: An alternative would be to use the Prometheus API to query the `targets` endpoint, but for the sake of an e2e I don't think it would be any better or worse than using an actual query like you have. https://prometheus.io/docs/prometheus/latest/querying/api/#targets

test/e2e/util/util.go
```diff
@@ -3855,3 +3856,74 @@ func EnsureCNOOperatorConfiguration(t *testing.T, ctx context.Context, mgmtClien
...
+		err := client.List(ctx, &serviceMonitors, crclient.InNamespace(hcpNamespace))
+		g.Expect(err).NotTo(HaveOccurred(), "failed to list ServiceMonitors in namespace %s", hcpNamespace)
+
+		if len(serviceMonitors.Items) == 0 {
+			t.Logf("No ServiceMonitors found in namespace %s", hcpNamespace)
+			return
+		}
+
+		// Create Prometheus client
+		prometheusClient, err := NewPrometheusClient(ctx)
```
wking, line 4272: [Looks][1] [like][2] this is connecting to the platform (`openshift-monitoring`)  `thanos-querier` Route.  That's fine if the hosted namespaces are getting `openshift.io/cluster-monitoring: true` label.  But can we rely on the test hosted namespaces having that label?  I guess [this existing test-case code][3] says we can?

[1]: https://github.com/openshift/hypershift/blob/3120e65285c4f1b5f11e70ea0c925a68f9007601/test/e2e/util/util.go#L1428-L1447
[2]: https://github.com/openshift/hypershift/blob/3120e65285c4f1b5f11e70ea0c925a68f9007601/vendor/github.com/openshift/library-go/test/library/metrics/query.go#L21-L28
[3]: https://github.com/openshift/hypershift/blob/3120e65285c4f1b5f11e70ea0c925a68f9007601/test/e2e/e2e_test.go#L335-L348

test/e2e/util/util.go
wking: And I'm not super-crisp on the labeling on `up`, but I'd expect the ServiceMonitor name to go in the `service` label, and not the `job` label.

wking: [ServiceMonitor docs][1] for a `spec.jobLabel` knob:

> `JobLabel` selects the label from the associated Kubernetes service which will be used as the job label for all metrics. For example: If in `ServiceMonitor.spec.jobLabel: foo` and in `Service.metadata.labels.foo: bar`, then the `job="bar"` label is added to all metrics. If the value of this field is empty or if the label doesn’t exist for the given Service, the `job` label of the metrics defaults to the name of the Kubernetes Service. 

I dunno if we want to set `jobLabel` to drive `job` per ServiceMonitor, but it seems easier to assume each ServiceMonitor will point at a unique Service with the same `name`, and then check for a match between the ServiceMonitor name and the `up` metric's `service` label.

[1]: https://docs.redhat.com/en/documentation/openshift_container_platform/4.12/html/monitoring_apis/servicemonitor-monitoring-coreos-com-v1#spec-7

wking: Hrm, but that conflicts with [the `TargetDown` runbook, which claims][1]:

> The label `namespace` and `job` can locate the `ServiceMonitor` or `PodMonitor` that configures Prometheus to scrape the target. The `job` label is the name of the monitor.

[The `TargetDown` alert itself][2] doesn't weigh in, just throwing both `job` and `service` into the description and letting the responding admin figure out which to use how.

[1]: https://github.com/openshift/runbooks/blob/0e7d70b93cfe1bf0d651520c5c07b5fa77fe9a28/alerts/cluster-monitoring-operator/TargetDown.md?plain=1#L48-L50
[2]: https://github.com/openshift/cluster-monitoring-operator/blob/13df5541efadb2d08428efe55a78a835b662dc72/jsonnet/rules.libsonnet#L23

sjenning:
```
: TestCreateCluster/Main/EnsureNoMetricsTargetsAreDown expand_less	0s
{Failed  === RUN   TestCreateCluster/Main/EnsureNoMetricsTargetsAreDown
    util.go:3908: No targets found for ServiceMonitor catalog-operator (this may be expected if scraping hasn't started yet)
    util.go:3908: No targets found for ServiceMonitor etcd (this may be expected if scraping hasn't started yet)
    util.go:3908: No targets found for ServiceMonitor monitor-multus-admission-controller (this may be expected if scraping hasn't started yet)
    util.go:3908: No targets found for ServiceMonitor monitor-ovn-control-plane-metrics (this may be expected if scraping hasn't started yet)
    util.go:3908: No targets found for ServiceMonitor olm-operator (this may be expected if scraping hasn't started yet)
    util.go:3924: Found 2 down metrics target(s):
        ServiceMonitor: cluster-version-operator, Target: up{_id="468b5c20-f993-4d57-9436-b148665d38b8", container="cluster-version-operator", endpoint="https", instance="10.124.0.43:8443", job="cluster-version-operator", namespace="e2e-clusters-2rgsh-create-cluster-rzknb", pod="cluster-version-operator-86d4649f6c-nrctv", prometheus="openshift-monitoring/k8s", service="cluster-version-operator"}
        ServiceMonitor: node-tuning-operator, Target: up{_id="468b5c20-f993-4d57-9436-b148665d38b8", container="cluster-node-tuning-operator", endpoint="60000", instance="10.124.0.51:60000", job="node-tuning-operator", namespace="e2e-clusters-2rgsh-create-cluster-rzknb", pod="cluster-node-tuning-operator-6c775b96d4-sxrbs", prometheus="openshift-monitoring/k8s", service="node-tuning-operator"}
        --- FAIL: TestCreateCluster/Main/EnsureNoMetricsTargetsAreDown (0.09s)
}
```

simonpasquier:
Because of target relabel configs, you have no guarantees that the job label == service name. As such the test feels brittle to me.

test/e2e/util/util.go
```diff
@@ -3855,3 +3856,92 @@ func EnsureCNOOperatorConfiguration(t *testing.T, ctx context.Context, mgmtClien
...
+			vector := result.(model.Vector)
+			if len(vector) == 0 {
+				t.Logf("No targets found in namespace %s (this may be expected if scraping hasn't started yet)", hcpNamespace)
+			}
+
+			// Create a set of monitor names for quick lookup
+			monitorSet := make(map[string]bool)
+			for _, monitor := range monitors {
+				monitorSet[monitor] = true
+			}
```
muraee: nit
```suggestion
			monitorSet := sets.New(monitors...)
```

sjenning:
/hold
until https://github.com/openshift/cluster-version-operator/pull/1243 chain merges and NTO metrics issue is debugged

test/e2e/util/util.go
wking: Orthogonally with using `sets` instead of a `map`, you may want to move this comparison-structure initialization up to where you're iterating over `serviceMonitors` and `podMonitors`.  I don't see a value in having an intermediate variable between those Kube-query results and the comparison structure you use while processing the PromQL-query results.

test/e2e/util/util.go
```diff
@@ -3855,3 +3856,92 @@ func EnsureCNOOperatorConfiguration(t *testing.T, ctx context.Context, mgmtClien
...
+
+			// Check each target in the results
+			for _, sample := range vector {
+				// Get the job label (which corresponds to the Monitor name)
+				jobLabel := string(sample.Metric["job"])
+
+				// Only check targets for our monitors
+				if monitorSet[jobLabel] {
```
wking: Instead of looping over samples and seeing if they're in our expected monitor set, can we loop over the expected monitor set and see if they're in the collected samples?  Failing if we didn't find a sample that we were expected would allow us to fail-closed and catch things like [the `job` label not matching the ServiceMonitor `name`][1] (as long as you didn't have any other monitors with that same `name` or someone else injecting an external monitor that generated the `job` label you were expecting, but those both seem like safe enough assumptions).

[1]: https://github.com/openshift/hypershift/pull/6965#issuecomment-3384526961

test/e2e/util/util.go
```diff
@@ -3855,3 +3856,92 @@ func EnsureCNOOperatorConfiguration(t *testing.T, ctx context.Context, mgmtClien
...
+					// Check if the target is down (up == 0)
+					if sample.Value == 0 {
```
wking: Just to be extra careful about catching any surprising semantics so we can think them through, can we use `!= 1` here?  Then the tests would fail and call in a maintainer to decide what needs changing if `up` grew additional values beyond its current 0 or 1.

sjenning:
> Because of target relabel configs, you have no guarantees that the job label == service name. As such the test feels brittle to me.

Yes, I already found this as pod monitor jobs are named `namespace/name` where service monitor jobs are just `name`.

However, since we (i.e. hypershift) control the SM and PM manifests, I think we can accept this limitation.

simonpasquier (commented):
I suppose that checking service/pod monitors which generate no scrape targets isn't in scope of this PR but is it something that is planned?

sjenning:
New e2e is passing after https://github.com/openshift/hypershift/pull/7086 merged.  Rebased and filtered to 4.21 only for now.

test/e2e/util/util.go
```diff
@@ -3922,3 +3923,129 @@ func ValidateConfigurationStatus(t *testing.T, ctx context.Context, mgmtClient c
...
+				g.Expect(err).NotTo(HaveOccurred(), "failed to list Services for ServiceMonitor %s/%s", sm.Namespace, sm.Name)
+
+				if len(serviceList.Items) == 0 {
+					t.Logf("No Services found for ServiceMonitor %s/%s, skipping", sm.Namespace, sm.Name)
+					continue
+				}
+
+				// Assume the first matching Service is the one we want
+				service := serviceList.Items[0]
+
```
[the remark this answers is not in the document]
sjenning: Done. Fixed to properly query Services using the selector (including matchExpressions support) and retrieve the job label value from the matched Service's labels.

---

test/e2e/util/util.go
```diff
@@ -3922,3 +3923,129 @@ func ValidateConfigurationStatus(t *testing.T, ctx context.Context, mgmtClient c
...
+		// if it has never been up, so we treat missing jobs as down.
+		var downJobs []string
+		for job, isUp := range jobs {
+			if !isUp {
+				downJobs = append(downJobs, job)
+			}
+		}
+		if len(downJobs) > 0 {
+			t.Errorf("The following Prometheus targets are down in namespace %s: %v", hcpNamespace, downJobs)
+		}
```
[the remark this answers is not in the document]
sjenning: Done. Fixed to track down targets per job and fail closed - if any instance reports down, the job is marked as down regardless of healthy instances. Added detailed failure reporting with instance names.

---

test/e2e/util/util.go
```diff
@@ -3922,3 +3923,129 @@ func ValidateConfigurationStatus(t *testing.T, ctx context.Context, mgmtClient c
...
+		// For each ServiceMonitor
+		for _, sm := range serviceMonitors.Items {
+			if sm.Spec.JobLabel != "" {
+				// Find the job label in the ServiceMonitor selector
+				// and use it as the job name directly
+				if val, ok := sm.Spec.Selector.MatchLabels[sm.Spec.JobLabel]; ok {
+					jobs[val] = false
+				} else {
+					t.Fatalf("ServiceMonitor %s/%s has jobLabel %s but it was not found in the selector labels", sm.Namespace, sm.Name, sm.Spec.JobLabel)
+				}
```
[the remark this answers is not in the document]
sjenning: Fixed. The jobLabel issue has been corrected to properly query Services using the selector and retrieve the job name from the Service's labels.

---

bryan-cox:
Are we still looking to get this PR in?

PR #6975, opened 2025-10-09
CNTRLPLANE-1622: Enable nodepool autoscaling min zero

jhjaggars:

This PR enables setting Min = 0 for NodePools to allow autoscaling to and from zero replicas.

### Key Changes:

1. **API Changes**:
   - Changed `NodePoolAutoScaling.Min` from `int32` to `*int32` (required by kubeapilinter for fields with meaningful zero values)
   - Updated validation to allow `minimum: 0` (previously `minimum: 1`)
   - Added CEL validation restricting `min=0` to AWS platform only

2. **New MachineDeployment Annotation Controller**:
   - Added new controller in `hypershift-operator/controllers/machinedeployment/`
   - Queries AWS EC2 API for instance type metadata (CPU, memory, GPU)
   - Annotates MachineDeployments with capacity information required by cluster-autoscaler
   - Enabled via `--scale-from-zero-aws-creds` flag (requires AWS credentials)
   - Critical for scale-from-zero: cluster-autoscaler needs capacity annotations to make scaling decisions
   - **Smart skip logic**: Detects if CPO natively supports scale-from-zero via image labels and automatically skips reconciliation to let CPO handle it (enables seamless migration when CPO gains native support)

3. **Controller Logic Updates**:
   - Implemented `effectiveMin` logic: enforces `min=1` for non-AWS platforms even if API validation is bypassed
   - Updated autoscaler enabled condition to check platform and version support

4. **RBAC Enhancement**:
   - Added `watch` verb to cluster-autoscaler role for `infrastructure.cluster.x-k8s.io` resources
   - Allows autoscaler's informer to use watch streams instead of polling for better performance

5. **Testing**:
   - Added E2E test `TestNodePoolAutoscalingScaleFromZero` validating scale from 0→1+ nodes
   - Added unit tests for platform-specific autoscaling logic
   - Added CreateAPIUX tests validating platform restrictions
   - Test skips when `--scale-from-zero-aws-creds` not enabled (requires MachineDeployment controller)

6. **Bug Fixes from Review**:
   - Fixed test workload to target scale-from-zero NodePool with nodeSelector
   - Fixed `mutateNodePoolAutoscalingValidation` to write spec back to schema map
   - Added nil-safe pointer dereferences using `ptr.Deref()`
   - Relaxed test assertion from exactly 2 replicas to >= 1 (pods may fit on one node)

Fixes https://issues.redhat.com/browse/CNTRLPLANE-1622

- **Scale-from-zero is AWS-only** (restricted via CEL validation and effectiveMin enforcement)
- **Requires `--enable-scale-from-zero` flag** when installing HyperShift operator
- **Requires AWS credentials** for the MachineDeployment controller to query EC2 instance types
- The new watch permission for cluster-autoscaler is optional (falls back to polling if unavailable)
- E2E test will skip in CI unless `-e2e.enable-scale-from-zero=true` is passed
- **CPO native support migration**: The MachineDeployment controller checks CPO image labels (`io.openshift.hypershift.control-plane-operator-supports-aws-autoscaling-from-zero=true`) and skips reconciliation when CPO supports it natively. The label is currently commented out in Dockerfiles until CPO implements native support - just uncomment when ready for automatic cutover.

## Checklist:

### DIFF
--- a/test/e2e/autoscaling_test.go
+++ b/test/e2e/autoscaling_test.go
@@ -17,10 +17,12 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/util/retry"
 	"k8s.io/utils/ptr"
+	capiv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	crclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
@@ -60,7 +62,7 @@ func TestAutoscaling(t *testing.T) {
 				}
 				additionalNP.Spec.Replicas = nil
 				additionalNP.Spec.AutoScaling = &hyperv1.NodePoolAutoScaling{
-					Min: 1,
+					Min: ptr.To[int32](1),
 					Max: 3,
 				}
 
@@ -107,7 +109,7 @@ func testAutoscaling(ctx context.Context, mgtClient crclient.Client, hostedClust
 
 		original := nodepool.DeepCopy()
 		nodepool.Spec.AutoScaling = &hyperv1.NodePoolAutoScaling{
-			Min: numNodes,
+			Min: ptr.To[int32](numNodes),
 			Max: max,
 		}
 		nodepool.Spec.Replicas = nil
@@ -383,4 +385,273 @@ func newWorkLoad(njobs int32, memoryRequest resource.Quantity, nodeSelector, ima
 		}
 	}
 	return job
+}
+
+func TestNodePoolAutoscalingScaleFromZero(t *testing.T) {
+	if globalOpts.Platform != hyperv1.AWSPlatform {
+		t.Skip("test only supported on platform AWS")
+	}
+
+	// Get management client to check for scale-from-zero secret
+	mgtClient, err := e2eutil.GetClient()
+	if err != nil {
+		t.Fatalf("failed to get management client: %v", err)
+	}
+
+	// Check if scale-from-zero is enabled by looking for the credentials secret
+	// The MachineDeployment controller is enabled when this secret is set
+	scaleFromZeroSecret := &corev1.Secret{}
+	err = mgtClient.Get(testContext, crclient.ObjectKey{
+		Namespace: "hypershift",
+		Name:      "hypershift-operator-scale-from-zero-aws-credentials",
+	}, scaleFromZeroSecret)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			t.Skip("test requires scale-from-zero to be enabled on the HyperShift Operator (secret hypershift-operator-scale-from-zero-aws-credentials not found)")
+		}
+		t.Fatalf("failed to check for scale-from-zero secret: %v", err)
+	}
+
+	t.Parallel()
+
+	ctx, cancel := context.WithCancel(testContext)
+	defer cancel()
+
+	clusterOpts := globalOpts.DefaultClusterOptions(t)
+	clusterOpts.NodePoolReplicas = 1
+
+	e2eutil.NewHypershiftTest(t, ctx, func(t *testing.T, g Gomega, mgtClient crclient.Client, hostedCluster *hyperv1.HostedCluster) {
+		t.Run("TestScaleFromZero", testScaleFromZero(ctx, mgtClient, hostedCluster))
+	}).Execute(&clusterOpts, globalOpts.Platform, globalOpts.ArtifactDir, "autoscaling-scale-from-zero", globalOpts.ServiceAccountSigningKey)
+}
+
+func testScaleFromZero(ctx context.Context, mgtClient crclient.Client, hostedCluster *hyperv1.HostedCluster) func(t *testing.T) {
+	return func(t *testing.T) {
+		g := NewWithT(t)
+		ctx, cancel := context.WithCancel(ctx)
+		defer cancel()
+
+		// Security context settings for workload pods
+		allowPrivilegeEscalation := false
+		runAsNonRoot := false
+		runAsUser := int64(0)
+
+		// Get guest client
+		guestClient := e2eutil.WaitForGuestClient(t, ctx, mgtClient, hostedCluster)
+
+		// Get default NodePool to copy spec
+		nodepools := &hyperv1.NodePoolList{}
+		err := mgtClient.List(ctx, nodepools, crclient.InNamespace(hostedCluster.Namespace))
+		g.Expect(err).NotTo(HaveOccurred(), "failed to list nodepools")
+		g.Expect(nodepools.Items).NotTo(BeEmpty(), "expected at least one nodepool")
+
+		// Create NodePool with scale-from-zero autoscaling
+		scaleFromZeroNP := &hyperv1.NodePool{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      hostedCluster.Name + "-scale-from-zero",
+				Namespace: hostedCluster.Namespace,
+			},
+		}
+		nodepools.Items[0].Spec.DeepCopyInto(&scaleFromZeroNP.Spec)
+		scaleFromZeroNP.Spec.Replicas = nil // Must be nil when using autoscaling
+		scaleFromZeroNP.Spec.AutoScaling = &hyperv1.NodePoolAutoScaling{
+			Min: ptr.To[int32](0), // Scale from zero
+			Max: 2,
+		}
+		// Add unique labels to nodes from this NodePool so workload can target them
+		scaleFromZeroNP.Spec.NodeLabels = map[string]string{
+			"scale-from-zero-test": "true",
+		}
+
+		err = mgtClient.Create(ctx, scaleFromZeroNP)
+		g.Expect(err).NotTo(HaveOccurred(), "failed to create scale-from-zero nodepool")
+		t.Logf("Created NodePool %s with autoscaling min=0, max=2", scaleFromZeroNP.Name)
+
+		// Verify MachineDeployment has capacity annotations
+		t.Log("Verifying MachineDeployment has capacity annotations")
+		md := &capiv1.MachineDeployment{}
+		e2eutil.EventuallyObject(t, ctx, "MachineDeployment to have capacity annotations",
+			func(ctx context.Context) (*capiv1.MachineDeployment, error) {
+				// MachineDeployment is in the hosted cluster namespace with same name as NodePool
+				err := mgtClient.Get(ctx, crclient.ObjectKey{
+					Namespace: manifests.HostedControlPlaneNamespace(hostedCluster.Namespace, hostedCluster.Name),
+					Name:      scaleFromZeroNP.Name,
+				}, md)
+				return md, err
+			},
+			[]e2eutil.Predicate[*capiv1.MachineDeployment]{
+				func(md *capiv1.MachineDeployment) (done bool, reasons string, err error) {
+					if _, ok := md.Annotations["machine.openshift.io/vCPU"]; !ok {
+						return false, "missing vCPU annotation", nil
+					}
+					if _, ok := md.Annotations["machine.openshift.io/memoryMb"]; !ok {
+						return false, "missing memoryMb annotation", nil
+					}
+					if _, ok := md.Annotations["machine.openshift.io/GPU"]; !ok {
+						return false, "missing GPU annotation", nil
+					}
+					labels, ok := md.Annotations["capacity.cluster-autoscaler.kubernetes.io/labels"]
+					if !ok {
+						return false, "missing capacity labels annotation", nil
+					}
+					if !strings.Contains(labels, "kubernetes.io/arch=") {
+						return false, "capacity labels missing architecture", nil
+					}
+					return true, "all capacity annotations present", nil
+				},
+			},
+			e2eutil.WithTimeout(5*time.Minute),
+		)
+		t.Logf("MachineDeployment has capacity annotations: vCPU=%s, memoryMb=%s, GPU=%s, labels=%s",
+			md.Annotations["machine.openshift.io/vCPU"],
+			md.Annotations["machine.openshift.io/memoryMb"],
+			md.Annotations["machine.openshift.io/GPU"],
+			md.Annotations["capacity.cluster-autoscaler.kubernetes.io/labels"])
+
+		// Verify NodePool autoscaling is enabled
+		e2eutil.EventuallyObject(t, ctx, "NodePool autoscaling to be enabled",
+			func(ctx context.Context) (*hyperv1.NodePool, error) {
+				err := mgtClient.Get(ctx, crclient.ObjectKeyFromObject(scaleFromZeroNP), scaleFromZeroNP)
+				return scaleFromZeroNP, err
+			},
+			[]e2eutil.Predicate[*hyperv1.NodePool]{
+				func(np *hyperv1.NodePool) (done bool, reasons string, err error) {
+					for _, condition := range np.Status.Conditions {
+						if condition.Type == hyperv1.NodePoolAutoscalingEnabledConditionType {
+							return condition.Status == corev1.ConditionTrue,
+								fmt.Sprintf("autoscaling enabled status is %s", condition.Status),
+								nil
+						}
+					}
+					return false, "autoscaling enabled condition not found", nil
+				},
+			},
+			e2eutil.WithTimeout(5*time.Minute),
+		)
+
+		// Verify NodePool starts with 0 replicas
+		t.Log("Verifying NodePool starts with 0 replicas")
+		err = mgtClient.Get(ctx, crclient.ObjectKeyFromObject(scaleFromZeroNP), scaleFromZeroNP)
+		g.Expect(err).NotTo(HaveOccurred())
+		g.Expect(scaleFromZeroNP.Status.Replicas).To(Equal(int32(0)), "NodePool should start with 0 replicas")
+		t.Log("Confirmed NodePool has 0 replicas")
+
+		// Create workload to trigger scale-up from 0
+		t.Log("Creating workload to trigger scale-up from 0 nodes")
+		workload := &batchv1.Job{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "scale-from-zero-workload",
+				Namespace: "default",
+			},
+			Spec: batchv1.JobSpec{
+				Template: corev1.PodTemplateSpec{
+					Spec: corev1.PodSpec{
+						Containers: []corev1.Container{
+							{
+								Name:    "workload",
+								Image:   globalOpts.LatestReleaseImage,
+								Command: []string{"sleep", "3600"},
+								Resources: corev1.ResourceRequirements{
+									Requests: corev1.ResourceList{
+										"memory": resource.MustParse("1Gi"),
+										"cpu":    resource.MustParse("500m"),
+									},
+								},
+								SecurityContext: &corev1.SecurityContext{
+									AllowPrivilegeEscalation: &allowPrivilegeEscalation,
+									Capabilities: &corev1.Capabilities{
+										Drop: []corev1.Capability{
+											"ALL",
+										},
+									},
+									RunAsNonRoot: &runAsNonRoot,
+									RunAsUser:    &runAsUser,
+									SeccompProfile: &corev1.SeccompProfile{
+										Type: corev1.SeccompProfileTypeRuntimeDefault,
+									},
+								},
+							},
+						},
+						// Target only nodes from the scale-from-zero NodePool
+						NodeSelector: map[string]string{
+							"scale-from-zero-test": "true",
+						},
+						RestartPolicy: corev1.RestartPolicyNever,
+					},
+				},
+				Completions: ptr.To[int32](2),
+				Parallelism: ptr.To[int32](2),
+			},
+		}
+
+		err = guestClient.Create(ctx, workload)
+		g.Expect(err).NotTo(HaveOccurred(), "failed to create workload")
+		t.Log("Created workload with 2 pods")
+
+		defer func() {
+			cascadeDelete := metav1.DeletePropagationForeground
+			_ = guestClient.Delete(ctx, workload, &crclient.DeleteOptions{
+				PropagationPolicy: &cascadeDelete,
+			})
+		}()
+
+		// Wait for NodePool to scale from 0 to at least 1
+		// Note: We request 2 pods but they may fit on a single node depending on instance capacity
+		t.Log("Waiting for NodePool to scale from 0 to at least 1 node")
+		e2eutil.EventuallyObject(t, ctx, "NodePool to scale from 0",
+			func(ctx context.Context) (*hyperv1.NodePool, error) {
+				err := mgtClient.Get(ctx, crclient.ObjectKeyFromObject(scaleFromZeroNP), scaleFromZeroNP)
+				return scaleFromZeroNP, err
+			},
+			[]e2eutil.Predicate[*hyperv1.NodePool]{
+				func(np *hyperv1.NodePool) (done bool, reasons string, err error) {
+					if np.Status.Replicas > 0 {
+						return true, fmt.Sprintf("NodePool scaled to %d replicas", np.Status.Replicas), nil
+					}
+					return false, fmt.Sprintf("NodePool has %d replicas, waiting for >0", np.Status.Replicas), nil
+				},
+			},
+			e2eutil.WithInterval(10*time.Second),
+			e2eutil.WithTimeout(15*time.Minute),
+		)
+		t.Logf("NodePool successfully scaled from 0 to %d replicas", scaleFromZeroNP.Status.Replicas)
+
+		// Verify pods are scheduled and running
+		t.Log("Verifying workload pods are scheduled and running")
+		e2eutil.EventuallyObjects(t, ctx, "Pods to be scheduled and running",
+			func(ctx context.Context) ([]*corev1.Pod, error) {
+				pods := &corev1.PodList{}
+				err := guestClient.List(ctx, pods,
+					crclient.InNamespace("default"),
+					crclient.MatchingLabels{"job-name": "scale-from-zero-workload"})
+				items := make([]*corev1.Pod, len(pods.Items))
+				for i := range pods.Items {
+					items[i] = &pods.Items[i]
+				}
+				return items, err
+			},
+			[]e2eutil.Predicate[[]*corev1.Pod]{
+				func(pods []*corev1.Pod) (done bool, reasons string, err error) {
+					if len(pods) < 2 {
+						return false, fmt.Sprintf("expected at least 2 pods, got %d", len(pods)), nil
+					}
+					return true, fmt.Sprintf("found %d pods (>= 2)", len(pods)), nil
+				},
+			},
+			[]e2eutil.Predicate[*corev1.Pod]{
+				e2eutil.ConditionPredicate[*corev1.Pod](e2eutil.Condition{
+					Type:   string(corev1.PodScheduled),
+					Status: metav1.ConditionTrue,
+				}),
+				e2eutil.Predicate[*corev1.Pod](func(pod *corev1.Pod) (done bool, reasons string, err error) {
+					if pod.Status.Phase == corev1.PodRunning || pod.Status.Phase == corev1.PodSucceeded {
+						return true, fmt.Sprintf("pod %s is %s", pod.Name, pod.Status.Phase), nil
+					}
+					return false, fmt.Sprintf("pod %s is %s", pod.Name, pod.Status.Phase), nil
+				}),
+			},
+			e2eutil.WithTimeout(20*time.Minute),
+		)
+		t.Log("Successfully verified scale-from-zero: workload pods are scheduled and running on scaled nodes")
+	}
 }
--- a/test/e2e/create_cluster_test.go
+++ b/test/e2e/create_cluster_test.go
@@ -25,6 +25,7 @@ import (
 	"github.com/openshift/hypershift/test/integration"
 	integrationframework "github.com/openshift/hypershift/test/integration/framework"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/client-go/tools/clientcmd"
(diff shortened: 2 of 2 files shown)


bryan-cox:
@jhjaggars - Please move this PR to draft per the contributor guidelines on the repo - https://github.com/openshift/hypershift?tab=contributing-ov-file#creating-a-pull-request

control-plane-operator/controllers/hostedcontrolplane/v2/assets/cluster-autoscaler/role.yaml
```diff
@@ -23,6 +23,7 @@ rules:
   verbs:
   - get
   - list
+  - watch
```
jhjaggars, line 26: This is for this informer:
https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_controller.go#L951

api/hypershift/v1beta1/nodepool_types.go
```diff
@@ -431,9 +431,11 @@ type NodePoolManagement struct {
 // NodePoolAutoScaling specifies auto-scaling behavior for a NodePool.
 // +kubebuilder:validation:XValidation:rule="self.max >= self.min", message="max must be equal or greater than min"
 type NodePoolAutoScaling struct {
-	// min is the minimum number of nodes to maintain in the pool. Must be >= 1 and <= .Max.
+	// min is the minimum number of nodes to maintain in the pool.
+	// Can be set to 0 for scale-from-zero.
+	// Must be >= 0 and <= .Max.
 	//
-	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Minimum=0
```
LiangquanLi930, line 454: So far, only AWS supports scale-up from zero, so I think it would be best to add a kubebuilder validation, similar to what’s done here:
https://github.com/openshift/hypershift/blob/dc604559b500f4ec35d1b0c43080170784323180/api/hypershift/v1beta1/nodepool_types.go#L87

docs/content/how-to/agent/create-agent-cluster.md
```diff
@@ -661,14 +661,30 @@ clusteroperator.config.openshift.io/storage                                    4
...
+~~~
+
+> **INFO:** With `min: 0`, the cluster will scale down to zero worker nodes when no workloads require scheduling, and automatically scale up when new workloads are created.
+
+**Important considerations for scale-from-zero:**
+
+- **Cold start delay**: When scaling from 0 to 1, there will be a delay as new agents are provisioned and join the cluster
+- **Template validation**: Ensure your NodePool configuration is valid since template issues will only be discovered during the first scale-up
+- **Monitoring**: Clusters with 0 nodes are normal in scale-from-zero scenarios, not indicative of a problem
+
```
LiangquanLi930: This part covers the installation for the agent platform. IMHO, we can move it to docs/content/how-to/automated-machine-management/autoscaling-from-zero.md.

test/e2e/autoscaling_test.go
```diff
@@ -67,6 +67,8 @@ func TestAutoscaling(t *testing.T) {
 		t.Run("TestAutoscaling", testAutoscaling(ctx, mgtClient, hostedCluster, clusterOpts.NodePoolReplicas, clusterOpts.NodePoolReplicas+2))
 
 		t.Run("TestAutoscalingBalancing", testAutoscalingBalancing(ctx, mgtClient, hostedCluster, clusterOpts.NodePoolReplicas*2, additionalNP))
+
+		t.Run("TestScaleFromZero", testScaleFromZero(ctx, mgtClient, hostedCluster))
```
LiangquanLi930: If the current solution is to use [capa-annotator](https://github.com/jhjaggars/capa-annotator), then I think we can skip adding this test for now. Once CAPA has been updated, we can bring this part of the code back.

api/hypershift/v1beta1/nodepool_types.go
jhjaggars: AWS isn't the only platform that supports scale up from zero.  Any provider can (and many do) support it.

docs/content/how-to/agent/create-agent-cluster.md
jhjaggars: On further thought, I'm not sure how helpful this documentation is at all.  How do you feel about it being omitted from this PR?

test/e2e/autoscaling_test.go
jhjaggars: Yeah, I'm not sure this sort of test makes any sense with the type of change in this PR.  I'll remove it entirely.

api/hypershift/v1beta1/nodepool_types.go
LiangquanLi930: Yes, IMHO, Agent, KubeVirt, Azure, and other platforms haven’t been systematically tested yet, so I think we can stick to AWS for now and add support for the other platforms once they’ve been verified.

docs/content/how-to/agent/create-agent-cluster.md
LiangquanLi930: look good to me

hypershift-operator/controllers/nodepool/capi.go
```diff
@@ -707,17 +707,27 @@ func setMachineDeploymentReplicas(nodePool *hyperv1.NodePool, machineDeployment
 		// The MachineDeployment replicas field should default to a value inside the (min size, max size) range based on the autoscaler annotations
 		// so the autoscaler can take control of the replicas field.
 		//
-		// 1. if it’s a new MachineDeployment, or the replicas field of the old MachineDeployment is < min size, use min size
+		// 1. if it's a new MachineDeployment, or the replicas field of the old MachineDeployment is <= min size, use min size
 		// 2. if the replicas field of the old MachineDeployment is > max size, use max size
+		//
```
jhjaggars, line 712: I want to call this section out.  I became concerned that supporting a minimum of 0 and 1 depending upon the platform might introduce subtle bugs in the case that Min gets set to 0 somehow.  This guard will ensure that Min will be treated as no less than 1 unless we are working against AWS.  This can be removed once we feel that the platform has no bearing on what the minimum should be.

api/hypershift/v1beta1/nodepool_types.go
```diff
@@ -86,6 +86,7 @@ type NodePool struct {
 // +kubebuilder:validation:XValidation:rule="!has(oldSelf.arch) || has(self.arch)", message="Arch is required once set"
 // +kubebuilder:validation:XValidation:rule="self.arch != 'arm64' || has(self.platform.aws) || has(self.platform.azure) || has(self.platform.agent) || self.platform.type == 'None'", message="Setting Arch to arm64 is only supported for AWS, Azure, Agent and None"
 // +kubebuilder:validation:XValidation:rule="!has(self.replicas) || !has(self.autoScaling)", message="Both replicas or autoScaling should not be set"
+// +kubebuilder:validation:XValidation:rule="!has(self.autoScaling) || self.autoScaling.min > 0 || has(self.platform.aws)", message="Scale-from-zero (autoScaling.min=0) is currently only supported for AWS platform"
```
bryan-cox: Could you add a test case for CreateAPIUX?

bryan-cox (commented):
Could you also add an end to end test that tests autoscaling to zero?

control-plane-operator/controllers/hostedcontrolplane/testdata/cluster-autoscaler/TechPreviewNoUpgrade/zz_fixture_TestControlPlaneComponents_cluster_autoscaler_role.yaml
```diff
@@ -32,6 +32,7 @@ rules:
   verbs:
   - get
   - list
+  - watch
```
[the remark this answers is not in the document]
jhjaggars: I think this is a false positive:
https://github.com/openshift/hypershift/pull/6975/files#diff-bb15035f378243f3b6ac2cd1419afc50596acb9512764921e54790a3cbe01601R26

cmd/install/install.go
```diff
@@ -223,6 +238,11 @@ func (o *Options) ApplyDefaults() {
 	default:
 		o.HyperShiftOperatorReplicas = 1
 	}
+
+	// Auto-enable scale-from-zero for AWS platform if not explicitly set
+	if !o.EnableScaleFromZero {
+		o.EnableScaleFromZero = isAWSPlatformEnabled(o.PlatformsToInstall)
```
csrwng: I would only enable it if creds were specified as well

cmd/install/install.go
```diff
@@ -701,26 +725,26 @@ func setupCRDs(opts Options, operatorNamespace *corev1.Namespace, operatorServic
...
-						Strategy: apiextensionsv1.WebhookConverter,
-						Webhook: &apiextensionsv1.WebhookConversion{
-							ClientConfig: &apiextensionsv1.WebhookClientConfig{
-								Service: &apiextensionsv1.ServiceReference{
-									Namespace: operatorNamespace.Name,
-									Name:      operatorService.Name,
-									Port:      ptr.To[int32](443),
-									Path:      ptr.To("/convert"),
+					// Handle conversion webhook setup
+					if opts.EnableConversionWebhook {
```
csrwng: unnecessary change

hypershift-operator/controllers/machinedeployment/controller.go
```diff
@@ -0,0 +1,279 @@
...
+
+	return ctrl.Result{}, nil
+}
+
+// getHostedClusterForMachineDeployment retrieves the HostedCluster associated with this MachineDeployment.
+func (r *Reconciler) getHostedClusterForMachineDeployment(ctx context.Context, md *clusterv1.MachineDeployment) (*hyperv1.HostedCluster, error) {
+	// Get CAPI Cluster in the same namespace via label
+	clusterName := md.Labels["cluster.x-k8s.io/cluster-name"]
+	if clusterName == "" {
+		return nil, fmt.Errorf("MachineDeployment missing cluster.x-k8s.io/cluster-name label")
```
csrwng: This is not something that should be retried. Whenever it gets the label you should get another event for it.

csrwng: I would just log it and return nil.

hypershift-operator/controllers/machinedeployment/controller.go
```diff
@@ -0,0 +1,279 @@
...
+		Name:      clusterName,
+	}, cluster)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get CAPI Cluster %s/%s: %w", md.Namespace, clusterName, err)
+	}
+
+	// Extract HostedCluster reference from annotation
+	hcKey := cluster.Annotations[supportutil.HostedClusterAnnotation]
+	if hcKey == "" {
+		return nil, fmt.Errorf("cluster %s missing %s annotation", clusterName, supportutil.HostedClusterAnnotation)
```
csrwng: same here, if the annotation is missing, I'd just return nil

hypershift-operator/controllers/machinedeployment/controller.go
```diff
@@ -0,0 +1,279 @@
...
+	}
+
+	return result, err
+}
+
+func (r *Reconciler) reconcile(ctx context.Context, machineDeployment *clusterv1.MachineDeployment) (ctrl.Result, error) {
+	klog.V(3).Infof("%v: Reconciling MachineDeployment", machineDeployment.Name)
+
+	// Check if CPO natively supports scale-from-zero
+	hc, err := r.getHostedClusterForMachineDeployment(ctx, machineDeployment)
```
csrwng: I would handle a nil err and nil cluster by just returning.

hypershift-operator/controllers/machinedeployment/controller.go
csrwng: It is also possible that the machinedeployment is for something other than a hypershift cluster.

hypershift-operator/controllers/machinedeployment/controller.go
```diff
@@ -0,0 +1,279 @@
...
+
+	if err := r.Client.Patch(ctx, machineDeployment, originalMachineDeploymentToPatch); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to patch machineDeployment: %v", err)
+	}
+
+	return result, err
+}
+
+func (r *Reconciler) reconcile(ctx context.Context, machineDeployment *clusterv1.MachineDeployment) (ctrl.Result, error) {
+	klog.V(3).Infof("%v: Reconciling MachineDeployment", machineDeployment.Name)
```
csrwng: please use the controller-runtime logger instead of klog
you can fetch the logger from the context with ctrl.LoggerFrom(ctx)

hypershift-operator/controllers/machinedeployment/controller.go
```diff
@@ -0,0 +1,279 @@
...
+		// Error reading the object - requeue the request.
+		return ctrl.Result{}, err
+	}
+
+	// Ignore deleted MachineDeployments, this can happen when foregroundDeletion
+	// is enabled
+	if !machineDeployment.DeletionTimestamp.IsZero() {
+		return ctrl.Result{}, nil
+	}
+
```
csrwng, line 282: You should ignore any machinedeployments that are outside of a control plane namespace. I would fetch the namespace of the machinedeployments and immediately return if it doesn't have the `hypershift.openshift.io/hosted-control-plane: "true"` label

hypershift-operator/controllers/nodepool/capi.go
```diff
@@ -707,17 +707,27 @@ func setMachineDeploymentReplicas(nodePool *hyperv1.NodePool, machineDeployment
...
+		// Guard: Enforce min=1 for non-AWS platforms to prevent scale-from-zero issues.
+		// Even if API validation fails or pre-existing objects exist with min=0, this prevents
+		// NodePools from being permanently stuck at 0 replicas on platforms that don't support
+		// scale-from-zero metadata.
+		effectiveMin := ptr.Deref(nodePool.Spec.AutoScaling.Min, 1)
```
muraee: why default to 1?

Containerfile.control-plane
```diff
@@ -28,3 +28,7 @@ LABEL io.openshift.hypershift.control-plane-pki-operator-signs-csrs=true
 LABEL io.openshift.hypershift.hosted-cluster-config-operator-reports-node-count=true
 LABEL io.openshift.hypershift.control-plane-operator-supports-kas-custom-kubeconfig=true
 LABEL io.openshift.hypershift.control-plane-operator.v2-isdefault=true
+# TODO: Uncomment when CPO natively adds capacity annotations to MachineDeployments for scale-from-zero.
+# When uncommented, the hypershift-operator MachineDeployment controller will automatically skip reconciliation
+# and let CPO handle annotations natively.
+# LABEL io.openshift.hypershift.control-plane-operator-supports-aws-autoscaling-from-zero=true
```
muraee: why do we need this? support from scale-from-zero is a CAPA feature, right? so it's related to the OCP release version of the HostedCluster not CPO capabilities

hypershift-operator/controllers/nodepool/capi.go
jhjaggars: The rationale (at one point) is that 1 is the more common minimum value, but the guard that follows fixes things anyhow.  Will switch to 0 as that feels more reasonable.

Containerfile.control-plane
jhjaggars: We don't need this at this point.  The strategy has changed to inspecting the awsmachinetemplates for status.capacity rather than a label on the CPO.

hypershift-operator/controllers/machinedeployment/controller.go
```diff
@@ -0,0 +1,203 @@
...
+	awsMachineTemplate, err := utils.ResolveAWSMachineTemplate(ctx, r.Client, machineDeployment)
+	if err != nil {
+		log.V(4).Info("Skipping MachineDeployment", "namespace", machineDeployment.Namespace, "name", machineDeployment.Name, "reason", err.Error())
+		// Not an error - this MachineDeployment may not be for AWS platform
+		return ctrl.Result{}, nil
+	}
+
+	// Check if AWSMachineTemplate.Status.Capacity is already populated by CPO
+	if len(awsMachineTemplate.Status.Capacity) > 0 {
+		log.V(3).Info("AWSMachineTemplate already has Status.Capacity set by CPO, skipping annotation workaround",
```
muraee: s/CPO/CAPA

hypershift-operator/controllers/machinedeployment/controller.go
```diff
@@ -0,0 +1,203 @@
...
+	InstanceTypesCache InstanceTypesCache
+
+	recorder record.EventRecorder
+	scheme   *runtime.Scheme
+}
+
+// SetupWithManager creates a new controller for a manager.
+func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
+	_, err := ctrl.NewControllerManagedBy(mgr).
+		For(&clusterv1.MachineDeployment{}).
```
muraee: should add a predicate to only watch MDs with hypershift nodePool annotation `hypershift.openshift.io/nodePool`
see https://github.com/openshift/hypershift/blob/main/hypershift-operator/controllers/nodepool/capi.go#L380

Containerfile.control-plane
muraee: ok, let's remove those comments then if it's not needed.

hypershift-operator/controllers/machinedeployment/utils/providerspec.go
```diff
@@ -0,0 +1,141 @@
...
+)
+
+const (
+	// RegionAnnotation is the fallback annotation for AWS region
+	RegionAnnotation = "capa.infrastructure.cluster.x-k8s.io/region"
+)
+
+// ResolveAWSMachineTemplate fetches the AWSMachineTemplate referenced by the MachineDeployment
+func ResolveAWSMachineTemplate(ctx context.Context, c client.Client, machineDeployment *clusterv1.MachineDeployment) (*infrav1.AWSMachineTemplate, error) {
+	// Extract infrastructureRef
```
csrwng, line 37: We should perform both of these checks (infrastructureRef.name and infraRef.kind) before attempting to fetch the namespace of the machinedeployments in the main controller body.

hypershift-operator/controllers/machinedeployment/controller.go
```diff
@@ -0,0 +1,216 @@
...
+
+func (r *Reconciler) reconcile(ctx context.Context, machineDeployment *clusterv1.MachineDeployment) (ctrl.Result, error) {
+	log := ctrl.LoggerFrom(ctx)
+	log.V(3).Info("Reconciling MachineDeployment", "machineDeployment", machineDeployment.Name)
+
+	// Resolve AWSMachineTemplate first
+	awsMachineTemplate, err := utils.ResolveAWSMachineTemplate(ctx, r.Client, machineDeployment)
+	if err != nil {
+		log.V(4).Info("Skipping MachineDeployment", "namespace", machineDeployment.Namespace, "name", machineDeployment.Name, "reason", err.Error())
+		// Not an error - this MachineDeployment may not be for AWS platform
```
csrwng: If you move the ref checks (infra ref type and name) to the main controller loop, then an error resulting from fetching the AWSMachineTemplate should be retryable and should result in a reconcile error.

hypershift-operator/controllers/machinedeployment/controller.go
```diff
@@ -0,0 +1,216 @@
...
+	originalMachineDeploymentToPatch := client.MergeFrom(machineDeployment.DeepCopy())
+
+	result, err := r.reconcile(ctx, machineDeployment)
+	if err != nil {
+		log.Error(err, "Failed to reconcile MachineDeployment")
+		r.recorder.Eventf(machineDeployment, corev1.EventTypeWarning, "ReconcileError", "%v", err)
+		// we don't return here so we want to attempt to patch the machine regardless of an error.
+	}
+
+	if err := r.Client.Patch(ctx, machineDeployment, originalMachineDeploymentToPatch); err != nil {
```
csrwng: This patch shouldn't happen unconditionally. It should only happen if there is a difference from the orignal machinedeployments.

hypershift-operator/controllers/machinedeployment/controller.go
```diff
@@ -0,0 +1,216 @@
...
+	awsClient, err := awsclient.NewValidatedClient(ctx, region, r.CredentialsFile, r.RegionCache)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("error creating aws client: %w", err)
+	}
+
+	// Get instance type information
+	instanceTypeInfo, err := r.InstanceTypesCache.GetInstanceType(ctx, awsClient, region, instanceType)
+	if err != nil {
+		log.Error(err, "Unable to set scale from zero annotations: unknown instance type", "instanceType", instanceType)
+		log.Error(nil, "Autoscaling from zero will not work. To fix this, manually populate machine annotations for your instance type", "annotations", []string{cpuKey, memoryKey, gpuKey})
```
csrwng: It seems this information might be more useful/visible in the event, rather than in the log.

hypershift-operator/controllers/machinedeployment/ec2_instance_types.go
```diff
@@ -0,0 +1,226 @@
...
+func NewInstanceTypesCache() InstanceTypesCache {
+	cache := &instanceTypesCache{}
+	cache.cache = map[string]instanceTypesRegion{}
+	cache.rwmutex = sync.RWMutex{}
+	return cache
+}
+
+// GetInstanceType retrieves InstanceType from cache by name. If the cache is stale or nil it is refreshed first from the EC2 API.
+// The fetched instance types are specific to the region of the awsClient. Using region name as cacheID is recommended.
+func (i *instanceTypesCache) GetInstanceType(ctx context.Context, awsClient awsclient.Client, cacheID string, instanceType string) (InstanceType, error) {
```
csrwng: I think the code would be clearer if you just use region throughout instead of cacheID

hypershift-operator/controllers/machinedeployment/controller.go
```diff
@@ -0,0 +1,225 @@
...
+	awsClient, err := awsclient.NewValidatedClient(ctx, region, r.CredentialsFile, r.RegionCache)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("error creating aws client: %w", err)
+	}
+
+	// Get instance type information
+	instanceTypeInfo, err := r.InstanceTypesCache.GetInstanceType(ctx, awsClient, region, instanceType)
+	if err != nil {
+		r.recorder.Eventf(machineDeployment, corev1.EventTypeWarning, "UnknownInstanceType",
+			"Unable to set scale from zero annotations: unknown instance type %s. Autoscaling from zero will not work. To fix this, manually populate machine annotations for your instance type: %v",
```
gaol: > To fix this, manually populate machine annotations for your instance type

does this mean manual update on the MachineDeployment ? if so, I think it will lead to  reconciling again which may succeed ?

hypershift-operator/controllers/machinedeployment/controller.go
```diff
@@ -0,0 +1,225 @@
...
+
+	// Resolve AWSMachineTemplate first
+	awsMachineTemplate, err := utils.ResolveAWSMachineTemplate(ctx, r.Client, machineDeployment)
+	if err != nil {
+		// Retryable error - could be transient API failure
+		return ctrl.Result{}, err
+	}
+
+	// Check if AWSMachineTemplate.Status.Capacity is already populated by CAPA
+	if len(awsMachineTemplate.Status.Capacity) > 0 {
```
gaol, line 314: maybe it is a silly question, but is it possible that the `awsMachineTemplate.Status.Capacity` gets updated after this reconciling ? 
how to test this case from QE side ?

Will setting an environment an accepted option instead of reading from runtime status to decide whether annotation workaround should be used ?

hypershift-operator/controllers/machinedeployment/controller.go
```diff
@@ -0,0 +1,225 @@
...
+	InstanceTypesCache InstanceTypesCache
+
+	recorder record.EventRecorder
+	scheme   *runtime.Scheme
+}
+
+// SetupWithManager creates a new controller for a manager.
+func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
+	_, err := ctrl.NewControllerManagedBy(mgr).
+		For(&clusterv1.MachineDeployment{}, builder.WithPredicates(predicate.NewPredicateFuncs(r.hasNodePoolAnnotation))).
```
gaol, line 61: It is possible that a HC has multiple nodepools, some nodepools don't have auto-scaling enabled, shall we filter only on the machinedeployment where the corresponding nodepool enabled auto-scaling ?

hypershift-operator/main.go
```diff
@@ -400,6 +420,20 @@ func run(ctx context.Context, opts *StartOptions, log logr.Logger) error {
 		return fmt.Errorf("unable to create controller: %w", err)
 	}
 
+	// Setup MachineDeployment controller for scale-from-zero
+	if opts.EnableScaleFromZero {
```
gaol: If this is only enabled on AWS now, shall we only set up the MachineDeployment controller if the platform is AWS ? At least there won't be miss leading logs saying the feature is enabled.

jhjaggars: Users can control if the controller starts by passing a valid credential or not.

hypershift-operator/controllers/machinedeployment/controller.go
jhjaggars: It is possible, and likely.  It is harmless to annotate deployments with this information however.

hypershift-operator/controllers/machinedeployment/controller.go
muraee: > maybe it is a silly question, but is it possible that the awsMachineTemplate.Status.Capacity gets updated after this reconciling ?

that could happen, I am thinking of an upgrade scenario from an older release to a release where CAPA supports scale-from-zero, where it will start setting this field.
I think we should remove the annotations if they exist here instead of just returning.

hypershift-operator/controllers/machinedeployment/controller.go
jhjaggars: If the instance type is unknown, then I would expect reconciliation to fail again unless AWS added that instance type in the meantime.  In the case that it exists the second time then the controller will set the proper values, else it'll emit this event again.  I think the advice to manually set the annotations is probably bad advice since the instance type is probably incorrect.  I will update the event to remove that advice.

hypershift-operator/controllers/machinedeployment/controller.go
jhjaggars: This is a good call out.  If the annotations are set and status.capacity is set, then the annotations will take precedence in the cluster-autoscaler.  This is backward from what we want.  I'm changing the controller to remove the annotations if status.capacity is set.

hypershift-operator/controllers/machinedeployment/controller.go
jhjaggars: Further, I am making sure we retry on transient failures and just emit the event if the instance type is unknown.

test/e2e/autoscaling_test.go
```diff
@@ -384,3 +385,254 @@ func newWorkLoad(njobs int32, memoryRequest resource.Quantity, nodeSelector, ima
 	}
 	return job
 }
+
+func TestNodePoolAutoscalingScaleFromZero(t *testing.T) {
+	if globalOpts.Platform != hyperv1.AWSPlatform {
+		t.Skip("test only supported on platform AWS")
+	}
+	if !globalOpts.HOInstallationOptions.EnableScaleFromZero {
```
csrwng: This may not be enough. The HOInstallationOptions are for tests that install the HO. The NodePool test uses the HO that is already installed (and configured in a previous workflow step).

You may need to fetch the HO deployment and inspect that.

hypershift-operator/main.go
```diff
@@ -124,6 +126,8 @@ type StartOptions struct {
 	EnableUWMTelemetryRemoteWrite          bool
 	EnableValidatingWebhook                bool
 	EnableDedicatedRequestServingIsolation bool
+	EnableScaleFromZero                    bool
+	AWSCreds                               string // Path to mounted credentials file (set by deployment)
```
csrwng: These creds are specifically for the new controller, so I would name something specific like:
`scale-from-zero-aws-creds` or something similar.
(same for the flag name)

api/hypershift/v1beta1/nodepool_types.go
```diff
@@ -446,11 +447,13 @@ type NodePoolManagement struct {
 // NodePoolAutoScaling specifies auto-scaling behavior for a NodePool.
 // +kubebuilder:validation:XValidation:rule="self.max >= self.min", message="max must be equal or greater than min"
 type NodePoolAutoScaling struct {
-	// min is the minimum number of nodes to maintain in the pool. Must be >= 1 and <= .Max.
+	// min is the minimum number of nodes to maintain in the pool.
+	// Can be set to 0 for scale-from-zero.
```
csrwng: `Can be set to 0 for scale-from-zero on AWS platform`

control-plane-operator/controllers/hostedcontrolplane/v2/assets/cluster-autoscaler/role.yaml
sdminonne: Sure. From https://issues.redhat.com/browse/CNTRLPLANE-1622

> RBAC Updates
> 
>     Add watch verb to cluster-autoscaler RBAC for infrastructure.cluster.x-k8s.io API group
>     Allows informer to watch rather than poll for better performance

control-plane-operator/controllers/hostedcontrolplane/testdata/cluster-autoscaler/GCP/zz_fixture_TestControlPlaneComponents_cluster_autoscaler_role.yaml
```diff
@@ -31,6 +31,7 @@ rules:
   verbs:
   - get
   - list
+  - watch
```
sdminonne, line 34: Even if GCP won't support (yet) scale from zero.  Tough but fair

control-plane-operator/controllers/hostedcontrolplane/testdata/cluster-autoscaler/IBMCloud/zz_fixture_TestControlPlaneComponents_cluster_autoscaler_role.yaml
```diff
@@ -31,6 +31,7 @@ rules:
   verbs:
   - get
   - list
+  - watch
```
sdminonne, line 34: Even if IBM won't support (yet) scale from zero. Tough but fair

hypershift-operator/controllers/machinedeployment/controller.go
```diff
@@ -0,0 +1,305 @@
...
+		}
+
+		if changed {
+			log.Info("Removing workaround annotations since CAPA provides Status.Capacity",
+				"machineDeployment", machineDeployment.Name)
+			// Continue to patch the MachineDeployment with removed annotations
+		} else {
+			// No annotations to clean up, skip
+			return ctrl.Result{}, nil
+		}
```
muraee, line 331: "// Continue to patch the MachineDeployment with removed annotations"
this comment is confusing, please remove.

the `else`? is also unnecessary, we are returning anyway right after.

jhjaggars:
/verified by @jhjaggars 

https://gist.github.com/jhjaggars/2ba3da57f5ad6929543106b78b4d16fd

jhjaggars:
/verified by @jhjaggars

https://gist.github.com/jhjaggars/d4b286d1176c8c92fe0c5df2356b0da6

@gaol is also working on additional verification.

PR #6988, opened 2025-10-10
STOR-2297: remove oVirt references from storage components

RomanBednar:
oVirt support ended in 4.14 and we're removing it from storage. It was already [removed](https://github.com/openshift/cluster-storage-operator/pull/617) from CSO, but we need to sync that change to HyperShift. 

cc @openshift/openshift-team-hypershift


muraee:
/approve
please update the commit to follow [conventional commit format](https://www.conventionalcommits.org/), otherwise `verify`  test will fail

RomanBednar:
/assign @sjenning @clebs 

For verification.

clebs:
/lgtm

Leaving to @sjenning for final check.

PR #6991, opened 2025-10-11
NO-JIRA: feat(claude): add some validation steps in jira-solve command to keep the codebase up-to-date

zhfeng:

- To check some tools before we start working
- Keep the codebase up-to-date and always create new fix branch against the upstream main
- typo fix

No JIRA

## Checklist:

### DIFF
--- a/.claude/commands/jira-solve.md
+++ b/.claude/commands/jira-solve.md
@@ -23,6 +23,13 @@ Analyze a JIRA issue and create a pull request to solve it.
 
 ## Process Flow:
 
+0. **Pre-flight Validation**:
+   - Validate that required tools are available (curl, jq, git, make)
+   - Check if current directory is a valid git repository
+   - Ensure working directory is clean or warn user about uncommitted changes
+   - Validate upstream remote is properly configured
+   - Fetch latest changes from upstream repository: `git fetch upstream`
+
 1. **Issue Analysis**: Parse JIRA URL and fetch issue details:
    - Use curl to fetch JIRA issue data: curl -s "https://issues.redhat.com/rest/api/2/issue/{$1}"
    - Parse JSON response to extract:
@@ -34,7 +41,7 @@ Analyze a JIRA issue and create a pull request to solve it.
          - Optional
             - Steps to reproduce (for bugs)
             - Expected vs actual behavior
-   - Ask the user for further issue grooming if the requried sections are missing
+   - Ask the user for further issue grooming if the required sections are missing
 
 2. **Codebase Analysis**: Search and analyze relevant code:
    - Find related files and functions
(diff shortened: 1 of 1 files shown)


.claude/commands/jira-solve.md
muraee:
```suggestion
   - Create feature branch using the jira-key $1 as the branch name. Use main branch from upstream as base. For example: "git checkout -b fix-{jira-key} upstream/main"
```

.claude/commands/jira-solve.md
```diff
@@ -23,6 +23,16 @@ Analyze a JIRA issue and create a pull request to solve it.
 
 ## Process Flow:
 
+0. **Pre-flight Validation**:
+   - Validate that required tools are available (curl, jq, git, make)
+   - Check if current directory is a valid git repository
+   - Ensure working directory is clean or warn user about uncommitted changes
+   - Validate upstream remote is properly configured
+   - Fetch latest changes from upstream repository: `git fetch upstream`
+   - Rebase current branch to latest upstream/main: `git rebase upstream/main`
```
muraee: this will cause problems. The agent should just use main as base branch when creating the new branch. See my other comment https://github.com/openshift/hypershift/pull/6991/files#r2425767804

.claude/commands/jira-solve.md
zhfeng: Thanks @muraee - I apply your suggestion.

.claude/commands/jira-solve.md
zhfeng: Yeah, I see.

muraee: still need to delete this line 
```
Rebase current branch to latest upstream/main: `git rebase upstream/main`
```

.claude/commands/jira-solve.md
```diff
@@ -23,6 +23,13 @@ Analyze a JIRA issue and create a pull request to solve it.
 
 ## Process Flow:
 
+0. **Pre-flight Validation**:
+   - Validate that required tools are available (curl, jq, git, make)
+   - Check if current directory is a valid git repository
+   - Ensure working directory is clean or warn user about uncommitted changes
+   - Validate upstream remote is properly configured
+   - Fetch latest changes from upstream repository: `git fetch upstream`
```
jparrill, line 31: This will grab all the modification related with new branches but it will not grab latest changes from the upstream main branch. For that you will need to `git pull --rebase upstream main` but that also depend on your dev workflow.

.claude/commands/jira-solve.md
```diff
@@ -65,7 +72,7 @@ Analyze a JIRA issue and create a pull request to solve it.
    - You absolutely must run `make pre-commit` and get it passing before going to the "Commit Creation"
 
 4. **Commit Creation**: 
-   - Create feature branch using the jira-key $1 as the branch name. For example: "git checkout -b fix-{jira-key}"
+   - Create feature branch using the jira-key $1 as the branch name. Use main branch from upstream as base. For example: "git checkout -b fix-{jira-key} upstream/main"
```
jparrill, line 75: If we have it done in preflight, I think is enough

.claude/commands/jira-solve.md
zhfeng: Yeah, my idea intends to make sure it keeps the codes update-to-date with the upstream repo before starting to work with the jira issue.

zhfeng: Well, I think we should create the new feature branch in this preflight step and then the codebase analysis step could grep all of the lastest codes. Does that make sense?

.claude/commands/jira-solve.md
zhfeng: This one could be moved to the preflight step?

jparrill: yeah just after the Fetch latest... Makes sense.

zhfeng:
These changes have been moved https://github.com/openshift-eng/ai-helpers/pull/55

PR #6994, opened 2025-10-13
CNTRLPLANE-1620: feat(gomaxprocs-webhook): introduce admission webhook to auto-set GOMAXPROCS

csrwng:
Add a mutating admission webhook that injects GOMAXPROCS based on CPU limits/requests to optimize Go runtime scheduling in pods. Includes:
- Implementation and configuration loader.
- Unit tests for handler and config parsing.
- Deployment manifests (service, SA, RBAC, webhook, kustomize).
- CLI entrypoint and Dockerfile.
- Documentation and Makefile.

Assisted-by: GPT-5 (via Cursor)
Fixes [CNTRLPLANE-1620](https://issues.redhat.com/browse/CNTRLPLANE-1620)

## Checklist:


csrwng:
@coderabbitai review

contrib/gomaxprocs-webhook/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+ENV GOFLAGS=-mod=mod
+COPY go.mod go.sum ./
+RUN go mod download
+
+# Copy only the gomaxprocs-webhook sources
+COPY . ./
+
+RUN make build
+
+FROM registry.access.redhat.com/ubi9:latest
```
celebdor: Maybe we can go with ubi9 minimal

contrib/gomaxprocs-webhook/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+COPY --from=builder /workspace/bin/gomaxprocs-webhook /usr/bin/gomaxprocs-webhook
+
+ENTRYPOINT ["/usr/bin/gomaxprocs-webhook", "serve"]
```
celebdor, line 18: Maybe we can run it as a regular user

contrib/gomaxprocs-webhook/cmd/serve.go
```diff
@@ -0,0 +1,178 @@
...
+		return nil
+	}
+
+	// Validate the YAML can be parsed
+	var testConfig config.Config
+	if err := yaml.Unmarshal(configYAML, &testConfig); err != nil {
+		return fmt.Errorf("configuration file %s contains invalid YAML: %w", configPath, err)
+	}
+
+	logger.V(1).Info("Configuration file validation successful", "configPath", configPath, "default", testConfig.Default, "overrides", len(testConfig.Overrides), "exclusions", len(testConfig.Exclusions))
```
celebdor: I think it would be useful to show the structured content of overrides and exclusions if run with log level debug

contrib/gomaxprocs-webhook/internal/config/loader.go
```diff
@@ -0,0 +1,174 @@
...
+		ex := cfg.Exclusions[i]
+		if strings.EqualFold(ex.WorkloadKind, workloadKind) && ex.WorkloadName == workloadName && ex.ContainerName == containerName {
+			return "", true, true
+		}
+	}
+
+	// Match overrides
+	for i := range cfg.Overrides {
+		ov := cfg.Overrides[i]
+		if strings.EqualFold(ov.WorkloadKind, workloadKind) && ov.WorkloadName == workloadName && ov.ContainerName == containerName {
```
celebdor, line 66: Should we allow not specifying the containername so it matches all the containers in a workload?

contrib/gomaxprocs-webhook/internal/webhook/pod/handler.go
```diff
@@ -0,0 +1,237 @@
...
+			return admission.Allowed("")
+		}
+		h.log.Error(err, "Failed to get namespace", "namespace", req.Namespace)
+		return admission.Errored(http.StatusInternalServerError, fmt.Errorf("failed to get namespace %s: %w", req.Namespace, err))
+	}
+
+	h.log.V(2).Info("Checking namespace labels", "namespace", req.Namespace, "labels", ns.Labels)
+	if ns.Labels == nil || ns.Labels[controlPlaneNamespaceLabel] != "true" {
+		h.log.V(2).Info("Namespace missing control plane label, allowing", "namespace", req.Namespace, "expectedLabel", controlPlaneNamespaceLabel)
+		return admission.Allowed("")
```
celebdor, line 70: Instead of all of this, couldn't we have the MutatingWebhookConfiguration with a content like:
```
  rules:
  - operations: ["CREATE"]
    apiGroups: [""]
    apiVersions: ["v1"]
    resources: ["pods"]
  namespaceSelector:
    matchLabels:
      hypershift.openshift.io/hosted-control-plane: "true"
```

Aren't you already doing that in https://github.com/openshift/hypershift/pull/6994/files#diff-c4edb749f7278e31ed38fc5e39c139e7544bdeff2987ecc5a5cb10ddd905f570 ?

contrib/gomaxprocs-webhook/manifests/role.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: gomaxprocs-webhook
+rules:
+- apiGroups: [""]
+  resources: ["namespaces", "configmaps", "pods"]
+  verbs: ["get", "list", "watch"]
```
celebdor: Are list and watch used for this resources?

Since the configmap is read from the volume mount, does it need any perms on the resource?

Don't we get the Pod from the Admission request instead of Getting/Listing it/Watching it?

In the case of namespace, if we use the mutatingwebhookconfig rules to matchlabel and we don't do defensive checks in the handler, we don't need any access to it either, right?

contrib/gomaxprocs-webhook/manifests/role.yaml
```diff
@@ -0,0 +1,17 @@
...
+- apiGroups: ["apps"]
+  resources: ["replicasets", "deployments", "statefulsets"]
+  verbs: ["get", "list", "watch"]
+- apiGroups: ["batch"]
+  resources: ["jobs", "cronjobs"]
```
celebdor: I see the following:

- Namespaces: only Get on handler.go:58
- ReplicaSets: only Get on handler.go:183
- Job: only Get on handler.go:195

I couldn't find us Getting for cronjobs, deployments and statefulsets. We get those as metadata parts of the objects we already have, right?

contrib/gomaxprocs-webhook/cmd/root.go
```diff
@@ -0,0 +1,34 @@
...
+	}
+
+	cmd.AddCommand(newServeCmd())
+	return cmd
+}
+
+// Execute runs the root command.
+func Execute(ctx context.Context) {
+	if err := NewRootCmd().ExecuteContext(ctx); err != nil {
+		fmt.Fprintln(os.Stderr, err)
```
jparrill: Better to use a logger

contrib/gomaxprocs-webhook/internal/config/loader.go
```diff
@@ -0,0 +1,174 @@
...
+		l.logger.V(1).Info("Cached configuration lost, re-parsing", "configPath", l.configPath)
+	}
+
+	// Parse the YAML
+	cfg := &Config{}
+	if rawString != "" {
+		// Compute non-sensitive identifiers for logging on error
+		contentSize := len(raw)
+		rawSHA256Sum := sha256.Sum256(raw)
+		contentHash := fmt.Sprintf("%x", rawSHA256Sum)[:12]
```
jparrill: is 12 enough quantity of chars to validate the content?

contrib/gomaxprocs-webhook/internal/config/loader_test.go
```diff
@@ -0,0 +1,351 @@
...
+	if excluded {
+		t.Error("expected not excluded for non-existent file")
+	}
+	if !ok {
+		t.Error("expected ok=true when falling back to default")
+	}
+	if value != "default-value" {
+		t.Errorf("expected default value 'default-value', got '%s'", value)
+	}
+}
```
jparrill: This test function could be integrated in the above one

contrib/gomaxprocs-webhook/internal/config/loader_test.go
```diff
@@ -0,0 +1,351 @@
...
+	if excluded {
+		t.Error("expected not excluded for empty file")
+	}
+	if !ok {
+		t.Error("expected ok=true for empty file")
+	}
+	if value != "default-value" {
+		t.Errorf("expected default value 'default-value', got '%s'", value)
+	}
+}
```
jparrill: This test function could be integrated in the above one

contrib/gomaxprocs-webhook/internal/config/loader_test.go
```diff
@@ -0,0 +1,351 @@
...
+	if excluded {
+		t.Error("expected not excluded for invalid YAML")
+	}
+	if !ok {
+		t.Error("expected ok=true when falling back to default")
+	}
+	if value != "default-value" {
+		t.Errorf("expected fallback to default value 'default-value', got '%s'", value)
+	}
+}
```
jparrill: This test function could be integrated in the above one

contrib/gomaxprocs-webhook/internal/config/loader_test.go
```diff
@@ -0,0 +1,351 @@
...
+			}
+			if ok != tt.expectedOk {
+				t.Errorf("expected ok=%v, got %v", tt.expectedOk, ok)
+			}
+			if value != tt.expectedValue {
+				t.Errorf("expected value='%s', got '%s'", tt.expectedValue, value)
+			}
+		})
+	}
+}
```
jparrill, line 292: Is this not validated already with other tests executions?

contrib/gomaxprocs-webhook/internal/webhook/pod/handler.go
jparrill: I believe this is a security measure just to avoid user modifications on the MutatingWebhookConfiguration file in the samples file.

contrib/gomaxprocs-webhook/internal/webhook/pod/handler.go
```diff
@@ -0,0 +1,237 @@
...
+		return admission.Errored(http.StatusInternalServerError, fmt.Errorf("failed to marshal patched pod: %w", err))
+	}
+	return admission.PatchResponseFromRaw(req.Object.Raw, mutatedRaw)
+}
+
+func (h *Handler) injectForPod(ctx context.Context, pod *corev1.Pod) bool {
+	// Resolve top-level workload
+	kind, name := h.resolveTopOwner(ctx, pod)
+	if kind == "" || name == "" {
+		// fall back to pod name context; still allow default application
```
jparrill: missing fallback here?

contrib/gomaxprocs-webhook/internal/webhook/pod/handler.go
```diff
@@ -0,0 +1,237 @@
...
+	currentKind := "Pod"
+	currentName := pod.Name
+	currentUID := string(pod.UID)
+	visited[currentUID] = true
+
+	// Start with the pod's first owner
+	owner := pod.OwnerReferences[0]
+	depth := 0
+
+traversalLoop:
```
jparrill, line 163: What's this?

contrib/gomaxprocs-webhook/internal/webhook/pod/handler.go
```diff
@@ -0,0 +1,237 @@
...
+			break traversalLoop
+		}
+		visited[ownerUID] = true
+
+		currentKind = owner.Kind
+		currentName = owner.Name
+
+		// Handle specific workload types that commonly have parent owners
+		var nextOwner *metav1.OwnerReference
+		switch owner.Kind {
```
jparrill, line 180: It's not clear for the reviewer which objects are being checked. the owner.Kind could be `ReplicaSet` and `Job`, but in the code's comment mentions ReplicaSet -> Deployment. From the reviewers perspective the switch will not check the Deployments nor CronJobs.

contrib/gomaxprocs-webhook/internal/webhook/pod/handler.go
```diff
@@ -0,0 +1,237 @@
...
+	h.log.V(2).Info("Owner resolution completed", "finalKind", currentKind, "finalName", currentName, "depth", depth)
+	return currentKind, currentName
+}
+
+// effectivePodName returns a non-empty identifier for the pod for logging purposes.
+// When a pod is created by a controller, the name may be empty in admission requests; in that case
+// we fall back to GenerateName. If both are empty, return a placeholder.
+func effectivePodName(pod *corev1.Pod) string {
+	if pod == nil {
+		return "<unknown>"
```
jparrill, line 228: is this name allowed by the k8s api pod admission?

contrib/gomaxprocs-webhook/internal/webhook/pod/handler_test.go
```diff
@@ -0,0 +1,525 @@
...
+			expectedName: "test-cronjob",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			scheme := runtime.NewScheme()
+			_ = corev1.AddToScheme(scheme)
+			_ = appsv1.AddToScheme(scheme)
+			_ = batchv1.AddToScheme(scheme)
```
jparrill: create a function with the right scheme for all the testing functions

contrib/gomaxprocs-webhook/internal/webhook/pod/handler_test.go
```diff
@@ -0,0 +1,525 @@
...
+
+	if kind != expectedKind {
+		t.Errorf("expected kind=%s, got %s", expectedKind, kind)
+	}
+	if name != expectedName {
+		t.Errorf("expected name=%s, got %s", expectedName, name)
+	}
+}
+
+func TestHandle_SkipsNonPod(t *testing.T) {
```
jparrill: many TestHandle... functions, maybe can we put all of them together.

contrib/gomaxprocs-webhook/Dockerfile
csrwng: fwiw it should run as a regular user on openshift by default since we randomize the UID of the pod.

contrib/gomaxprocs-webhook/internal/webhook/pod/handler.go
csrwng: What Juan said

contrib/gomaxprocs-webhook/manifests/role.yaml
csrwng: List and watch are used internally by the controller-runtime cache for the resource informer.
And yes we don't need some of these resources, will remove.

contrib/gomaxprocs-webhook/internal/config/loader.go
csrwng: I would think so, but I'll make it larger... 24 just in case

contrib/gomaxprocs-webhook/internal/config/loader_test.go
csrwng: It's more about testing the "happy path" and making sure we get the expected values from a valid configuration.

contrib/gomaxprocs-webhook/internal/webhook/pod/handler.go
csrwng: workloadKind and workloadName as passed as blank to the resolve function which should return the defaultValue if one is set in the configuration. However, the code is a no-op basically, so will remove

contrib/gomaxprocs-webhook/internal/webhook/pod/handler.go
csrwng: It's a label that allows breaking out of the loop from inside a switch statement (without the label, the code would just break out of the switch statement)

contrib/gomaxprocs-webhook/internal/webhook/pod/handler.go
csrwng: We only need to worry about refs to resources that could have an owner that we care about. In this case, ReplicaSet could have an owner of type Deployment, while Job could have an owner of type CronJob. If it's already a Deployment, StatefulSet, etc. we use the owner ref itself and don't worry about looking for another owner.
That's what the comment means:
```
// Handle specific workload types that commonly have parent owners
```
I'll explain it better in the comment.

contrib/gomaxprocs-webhook/internal/webhook/pod/handler.go
csrwng: This is only used for logging

contrib/gomaxprocs-webhook/internal/webhook/pod/handler.go
jparrill: Gotcha!, thanks for the clarification

jparrill:
Answering to myself, the webhook is working as expected, for the current workloads already running in the cluster, we need to add the annotation to restart the HCP in order to be patched by the webhook.

PR #6995, opened 2025-10-13
NO-JIRA: chore(ai): add ability for jira-solve to save a plan spec first

enxebre:
this result in jira-solve solving a spec.md file with a plan for the feature to jira to be implemented. By asking the model to plan first I would expect better final implementation. Also this enables trace back plans for restrospectives and tackle action items of a plan at different times

Fixes 

## Checklist:

### DIFF
--- a/.claude/commands/jira-solve.md
+++ b/.claude/commands/jira-solve.md
@@ -47,19 +47,22 @@ Analyze a JIRA issue and create a pull request to solve it.
       - Test files that need updates
 
 3. **Solution Implementation**:
-   - Make necessary code changes using Edit/MultiEdit tools
-   - Follow existing code patterns and conventions
-   - Add or update tests as needed
-   - Update documentation if needed within the docs/ folder
-   - If the problem is too complex consider delegating to one of the SME agents.
-   - Ensure godoc comments are generated for any newly created public functions
+   - Think hard and create a detailed, step-by-step plan to implement this feature. Save it to spec-$1.md within the contrib/ folder, for example contrib/spec-OCPBUGS-12345.md 
+   - Always ask the user to review the plan and give them the choice to modify it before start the implementation
+   - Implement the plan:
+    - Make necessary code changes using Edit/MultiEdit tools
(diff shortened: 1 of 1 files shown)


.claude/commands/jira-solve.md
```diff
@@ -47,19 +47,22 @@ Analyze a JIRA issue and create a pull request to solve it.
...
 
 3. **Solution Implementation**:
-   - Make necessary code changes using Edit/MultiEdit tools
-   - Follow existing code patterns and conventions
-   - Add or update tests as needed
-   - Update documentation if needed within the docs/ folder
-   - If the problem is too complex consider delegating to one of the SME agents.
-   - Ensure godoc comments are generated for any newly created public functions
+   - Think hard and create a detailed, step-by-step plan to implement this feature. Save it to spec-$1.md, for example spec-OCPBUGS-12345.md
+   - Always ask the user to review the plan and give them the choice to modify it before start the implementation
```
bryan-cox, line 51: How would this work with running the command in an autonomous mode thru prow for example?

.claude/commands/jira-solve.md
```diff
@@ -47,19 +47,22 @@ Analyze a JIRA issue and create a pull request to solve it.
       - Test files that need updates
 
 3. **Solution Implementation**:
-   - Make necessary code changes using Edit/MultiEdit tools
-   - Follow existing code patterns and conventions
-   - Add or update tests as needed
-   - Update documentation if needed within the docs/ folder
-   - If the problem is too complex consider delegating to one of the SME agents.
-   - Ensure godoc comments are generated for any newly created public functions
+   - Think hard and create a detailed, step-by-step plan to implement this feature. Save it to spec-$1.md, for example spec-OCPBUGS-12345.md
```
bryan-cox: Should we have it save these md plan files to the contrib folder so that we have a reference to what it did later?

.claude/commands/jira-solve.md
enxebre: I think we can tweak it then

PR #6997, opened 2025-10-13
OCPBUGS-62871: fix(e2e): refactor createHostedCluster to remove deps between option settings

sjenning:
Currently, `createHostedCluster` only sets some options related to the `Ingress` and `Authentication` configuration if the signing key is provided to the e2e binary in advance (vs being created on the fly).

There is no reason for the dependency.  The signing key options, ingress options, and authentication options have no interdependencies and should be set independently from one another.

---

> [!NOTE]
> Refactors e2e createHostedCluster to independently set AWS ingress and external OIDC auth/resources, no longer gated by the SA signing key.
> 
> - **e2e framework (`test/e2e/util/hypershift_framework.go`)**:
>   - **AWS Ingress**: Always set `HostedCluster.Spec.Configuration.Ingress` to AWS NLB via `BeforeApply` when `platform == AWS`, independent of signing key.
>   - **External OIDC**:
>     - Always create external OIDC `Secret` and CA `ConfigMap` when `opts.ExtOIDCConfig != nil`.
>     - Set `HostedCluster.Spec.Configuration.Authentication` from `ExtOIDCConfig` via its own `BeforeApply` wrapper.
>   - **SA Signing Key**: Keep signing key secret creation/assignment isolated; `BeforeApply` only sets `Spec.ServiceAccountSigningKey`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 43df45c3f10590088714b4d3cf231a36b1f9805c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/test/e2e/util/hypershift_framework.go
+++ b/test/e2e/util/hypershift_framework.go
@@ -275,6 +275,32 @@ func (h *hypershiftTest) createHostedCluster(opts *PlatformAgnosticOptions, plat
 	err := h.client.Create(h.ctx, namespace)
 	g.Expect(err).NotTo(HaveOccurred(), "failed to create namespace")
 
+	if platform == hyperv1.AWSPlatform {
+		originalBeforeApply := opts.BeforeApply
+		opts.BeforeApply = func(o crclient.Object) {
+			if originalBeforeApply != nil {
+				originalBeforeApply(o)
+			}
+
+			switch v := o.(type) {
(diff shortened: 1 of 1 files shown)


sjenning:
@coderabbitai review

PR #7024, opened 2025-10-14
OCPBUGS-62949: fix(resources): prevent indefinite blocking on cloud resource cleanup during deletion

muraee:

When ensureCloudResourcesDestroyed() attempts to clean up guest cluster resources, it queries the guest cluster's KubeAPIServer. If the KubeAPIServer is already deleted during cluster deletion, these operations fail with connection errors, causing the CloudResourcesDestroyed condition to never become True, which blocks cluster deletion indefinitely.

This fix implements two safety mechanisms to handle KubeAPIServer unavailability:

1. Early KAS check: Verify the kube-apiserver deployment exists in the control plane namespace before attempting cleanup. If not found, skip cleanup immediately as the guest cluster is already gone.

2. Connection error tracking: Track consecutive connection failures in-memory and skip cleanup after 5 attempts or 5 minutes, whichever comes first. This prevents infinite retry loops when the KubeAPIServer is unreachable.

Key implementation details:

- Added isKubeAPIServerAvailable() to check KAS deployment existence using the control plane client
- Added isConnectionError() using proper K8s API errors (IsTimeout, IsServerTimeout, IsServiceUnavailable) and Go's net.Error interface instead of string matching
- Implemented in-memory failure tracking with cleanupFailureTracker to avoid persisting state and potential API errors
- Failure tracker is NOT reset when skipping due to max failures/timeout to prevent condition flip-flopping on subsequent reconciliations
- Added comprehensive unit tests covering KAS unavailability, connection error detection, and failure tracking

The implementation ensures stable CloudResourcesDestroyed condition status, allowing cluster deletion to proceed even when the guest cluster API is unavailable.

Assisted-by: Claude 4.5 Sonnet (via Cursor)

Fixes OCPBUGS-62949

## Checklist:

### DIFF
--- a/control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
+++ b/control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
@@ -2,7 +2,9 @@ package resources
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"net"
 	"reflect"
 	"strings"
 	"sync"
@@ -67,7 +69,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/apimachinery/pkg/util/errors"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/rest"
 	apiregistrationv1 "k8s.io/kube-aggregator/pkg/apis/apiregistration/v1"
@@ -141,6 +143,7 @@ type reconciler struct {
 	versions                  map[string]string
 	operateOnReleaseImage     string
 	ImageMetaDataProvider     util.ImageMetadataProvider
+	cleanupTracker            *util.CleanupTracker
 }
 
 // eventHandler is the handler used throughout. As this controller reconciles all kind of different resources
@@ -197,6 +200,7 @@ func Setup(ctx context.Context, opts *operator.HostedClusterConfigOperatorConfig
 		versions:                  opts.Versions,
 		operateOnReleaseImage:     opts.OperateOnReleaseImage,
 		ImageMetaDataProvider:     opts.ImageMetaDataProvider,
+		cleanupTracker:            util.NewCleanupTracker(),
 	}})
 	if err != nil {
 		return fmt.Errorf("failed to construct controller: %w", err)
@@ -781,7 +785,7 @@ func (r *reconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Result
 		finished, err = r.reconcileRestoredCluster(ctx, hcp)
 		if err != nil {
 			log.Error(err, "failed to reconcile hosted cluster recovery")
-			return ctrl.Result{}, errors.NewAggregate(append(errs, err))
+			return ctrl.Result{}, utilerrors.NewAggregate(append(errs, err))
 		}
 
 		if !finished {
@@ -809,7 +813,7 @@ func (r *reconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Result
 		log.Info("successfully updated hcp status with recovery finished condition")
 	}
 
-	return ctrl.Result{}, errors.NewAggregate(errs)
+	return ctrl.Result{}, utilerrors.NewAggregate(errs)
 }
 
 func (r *reconciler) reconcileCSIDriver(ctx context.Context, hcp *hyperv1.HostedControlPlane, releaseImage *releaseinfo.ReleaseImage) error {
@@ -836,7 +840,7 @@ func (r *reconciler) reconcileCRDs(ctx context.Context) error {
 		errs = append(errs, fmt.Errorf("failed to reconcile request count crd: %w", err))
 	}
 
-	return errors.NewAggregate(errs)
+	return utilerrors.NewAggregate(errs)
 }
 
 func (r *reconciler) reconcileConfig(ctx context.Context, hcp *hyperv1.HostedControlPlane) error {
@@ -951,7 +955,7 @@ func (r *reconciler) reconcileConfig(ctx context.Context, hcp *hyperv1.HostedCon
 		errs = append(errs, fmt.Errorf("failed to reconcile apiserver config: %w", err))
 	}
 
-	return errors.NewAggregate(errs)
+	return utilerrors.NewAggregate(errs)
 }
 
 func (r *reconciler) reconcileProxyTrustedCAConfigMap(ctx context.Context, hcp *hyperv1.HostedControlPlane) error {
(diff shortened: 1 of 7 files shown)


muraee:
@coderabbitai review

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -141,6 +152,8 @@ type reconciler struct {
 	versions                  map[string]string
 	operateOnReleaseImage     string
 	ImageMetaDataProvider     util.ImageMetadataProvider
+	cleanupFailures           map[string]*cleanupFailureTracker
```
csrwng: the failure tracker (and associated mutex) should be in its own struct (likely under support/util).

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -2159,15 +2173,52 @@ func (r *reconciler) destroyCloudResources(ctx context.Context, hcp *hyperv1.Hos
 func (r *reconciler) ensureCloudResourcesDestroyed(ctx context.Context, hcp *hyperv1.HostedControlPlane) (sets.Set[string], error) {
 	log := ctrl.LoggerFrom(ctx)
 	remaining := sets.New[string]()
+	hcpKey := fmt.Sprintf("%s/%s", hcp.Namespace, hcp.Name)
+
+	// Check if KubeAPIServer deployment exists
+	kasAvailable, err := r.isKubeAPIServerAvailable(ctx, hcp)
```
csrwng: checking whether the KAS is available should be part of r.shouldSkipCleanup()

muraee:
**🤖 This reply was generated by Claude AI on behalf of the contributor**

Thank you for the review feedback! I've addressed all the comments in commit a7b8be4aa:

**@csrwng's comments:**
1. **Line 155** - Moved failure tracker and mutex to separate struct: Extracted the cleanup failure tracking logic into a new `CleanupTracker` utility in `support/util/cleanup_tracker.go`. This provides better separation of concerns and makes the tracking logic reusable.

2. **Line 2179** - Integrate KAS availability check into `shouldSkipCleanup()`: The KAS availability check is now part of the `CleanupTracker.ShouldSkipCleanup()` method, which first checks if the KubeAPIServer deployment exists before evaluating failure thresholds.

**@coderabbitai's comment:**
3. **Line 2199** - Fix CloudResourcesDestroyed condition: Modified `destroyCloudResources()` to properly distinguish between skipped cleanup (condition = False with specific skip reason) vs destroyed resources (condition = True). The `ensureCloudResourcesDestroyed()` now returns a `skipReason` string that indicates why cleanup was skipped (e.g., "KubeAPIServerUnavailable", "MaxConnectionFailuresExceeded", "ConnectionFailureTimeout").

The changes introduce:
- New `CleanupTracker` type in `support/util/cleanup_tracker.go` with thread-safe failure tracking
- Updated `ensureCloudResourcesDestroyed()` signature to return `(sets.Set[string], string, error)`
- Updated `destroyCloudResources()` to set appropriate condition status and messages based on skip reason
- Updated all tests to work with the refactored code

Please let me know if there are any other concerns!

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
muraee: Done. Extracted CleanupTracker to support/util/cleanup_tracker.go with thread-safe operations.

---

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
muraee: Done. ShouldSkipCleanup now checks KubeAPIServer deployment availability first (support/util/cleanup_tracker.go:80-90).

---

muraee:
@coderabbitai Excellent catch on the semantic issue! You're absolutely right - setting `CloudResourcesDestroyed=False` when cleanup is skipped would block deletion indefinitely (unless the 10-minute timeout is hit).

## The Fix

I've addressed this by implementing your suggestion to check for specific skip reasons in the deletion logic:

1. **Added unified constant** `CloudResourcesCleanupSkippedReason` in `api/hypershift/v1beta1/hostedcluster_conditions.go:253`

2. **Updated condition handling** in `resources.go:2131` to use this unified constant as the reason while preserving detailed skip reasons (KubeAPIServerUnavailable, MaxConnectionFailuresExceeded, ConnectionFailureTimeout) in the message for observability

3. **Updated deletion logic** in `hostedcontrolplane_controller.go:2789-2794` to check for this skip reason and **allow deletion to proceed** when cleanup was skipped due to guest cluster unavailability

## Result

Now when the guest cluster is unavailable (KAS deployment gone, connection failures, etc.), the deletion flow correctly:
- Sets `CloudResourcesDestroyed=False` with `Reason=CloudResourcesCleanupSkippedReason` 
- Includes detailed skip reason in the message for operators
- **Allows HostedControlPlane deletion to proceed** (via the new check in `removeCloudResources()`)

This prevents indefinite blocking while maintaining clear observability of why cleanup was skipped.

Thanks for the thorough review! 🙏

---

muraee:
@coderabbitai code is pushed

csrwng:
Please add unit tests to the new cleanup_tracker.go utility

muraee:
@csrwng Done. Added comprehensive unit tests for the CleanupTracker utility covering all public methods, concurrent access, and the KAS availability check.

---

support/util/cleanup_tracker.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+	tracker.count++
+}
+
+// ResetFailures resets the failure tracking for a given HCP
+func (t *CleanupTracker) ResetFailures(hcpKey string) {
+	t.mutex.Lock()
+	defer t.mutex.Unlock()
+	delete(t.failures, hcpKey)
+}
```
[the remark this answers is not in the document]
muraee: Done. Fixed the race condition by holding the write lock during map lookup/creation and increment in `RecordFailure`. Also fixed similar races in `ShouldSkipCleanup`, `GetFailureCount`, and `GetFirstFailureTime` by holding the appropriate lock during map access and field reads.

The `getFailureTracker` method has been removed as it's no longer needed. All tests pass.

---

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources_test.go
```diff
@@ -905,6 +914,163 @@ func TestDestroyCloudResources(t *testing.T) {
...
+				g.Expect(remaining.Len()).To(Equal(0))
+				g.Expect(skipReason).To(Equal("KubeAPIServerUnavailable"))
+			} else {
+				// When KAS is available, cleanup should proceed normally
+				g.Expect(remaining.Len()).To(Equal(0))
+				g.Expect(skipReason).To(BeEmpty())
+			}
+		})
+	}
+}
```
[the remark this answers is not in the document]
muraee: ensureCloudResourcesDestroyed doesn't change status of the hcp

PR #6462, opened 2025-07-17
CNTRLPLANE-1116: add gitlint support for conventional commit enforcement

bryan-cox:
**What this PR does / why we need it**:
This PR:
- adds gitlint support for conventional commit enforcement
- updates the upstream contribute page to be more concise, improve readability and flow
- adds cursor rules for generating conventional commits

**Which issue(s) this PR fixes**:
Fixes [CNTRLPLANE-1116](https://issues.redhat.com/browse/CNTRLPLANE-1116)

**Checklist**
- [x] Subject and description added to both, commit and PR.
- [x] Relevant issues have been referenced.
- [x] This change includes docs. 
- [ ] This change includes unit tests.

PR #7052, opened 2025-10-14
OCPBUGS-63172: control-plane-operator/.../ingress-operator/deployment: Declare a metrics port

wking:

We need the `metrics` declaration so the PodMonitor knows where to scrape:

```console
$ grep port: control-plane-operator/controllers/hostedcontrolplane/v2/assets/ingress-operator/podmonitor.yaml
    port: metrics
```

The port number is set to match the:

```
  - --metrics-listen-addr
  - 0.0.0.0:60000
```

arguments from earlier in the container spec.

I'm not sure why 60000 was chosen, and I have not gone digging in history to find out.  I'd have expected the port choice to be something that just had to be unique within the Pod, and for something more common like 8000 to have been chosen.  But if folks want to adjust there, that can happen in follow-up work.

~I haven't filed an OCPBUGS around this, but I can if you think it's worth backporting.~ Fixes [OCPBUGS-63172](https://issues.redhat.com/browse/OCPBUGS-63172).

No unit tests here, but #6965 is bringing in test coverage.

## Checklist:

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/testdata/ingress-operator/IBMCloud/zz_fixture_TestControlPlaneComponents_ingress_operator_deployment.yaml
+++ b/control-plane-operator/controllers/hostedcontrolplane/testdata/ingress-operator/IBMCloud/zz_fixture_TestControlPlaneComponents_ingress_operator_deployment.yaml
@@ -94,6 +94,10 @@ spec:
         image: cluster-ingress-operator
         imagePullPolicy: IfNotPresent
         name: ingress-operator
+        ports:
+        - containerPort: 60000
+          name: metrics
+          protocol: TCP
         resources:
           requests:
             cpu: 10m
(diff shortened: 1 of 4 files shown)


bryan-cox:
> I haven't filed an OCPBUGS around this, but I can if you think it's worth backporting.
@wking I think we should have a valid Jira ticket per our contributor guidelines so please file an OCPBUGS when you have a moment.

wking:
I've pushed 59edd82bd -> 878891049, picking up Seth's fixture updates from #7056, which should get the unit tests passing here.  I'm agnostic about whether y'all merge this one or #7056

wking:
And I've pushed 878891049 -> 1796e2dcc to adjust the commit message to match [the Conventional Commits spec][1] that the README recommends since 37cfe1abe (#6462), although if you want folks to notice that requirement without having to call it out by hand, you may want to grow presubmit coverage to enforce it.

[1]: https://www.conventionalcommits.org/en/v1.0.0/

PR #7056, opened 2025-10-15
NO-JIRA: fix(ingress-operator): include metrics port in deployment

sjenning:
The ingress-operator has a pod monitor that references a currently non-existent "metrics" port on pod.

Add the metrics port to the deployment definition so that the pod monitor works.

Thanks to @wking for finding this :bowing_man: 

Relates to https://issues.redhat.com//browse/OCPBUGS-62851

Supersedes https://github.com/openshift/hypershift/pull/7052


sjenning:
go with https://github.com/openshift/hypershift/pull/7052

PR #6239, opened 2025-06-07
MULTIARCH-5544: Added support to create s390x NodePool with kubevirt platform

veera-damisetti:
**What this PR does / why we need it**:

**Added support to create s390x Nodepool with  kubevirt platform.**
- Currently there is no support to create s390x nodepool with platform kubevirt
- To install hcp with kubevirt platform, on an s390x management cluster we need to have this support.

**Which issue(s) this PR fixes** *(optional, use `fixes #<issue_number>(, fixes #<issue_number>, ...)` format, where issue_number might be a GitHub issue, or a Jira story*:
Fixes #

**Checklist**
- [x] Relevant issues have been referenced.
- [x] This change includes docs. 
- [x] Subject and description added to both, commit and PR.
- [x] This change includes unit tests.

PR #7058, opened 2025-10-15
MULTIARCH-5629: Added support to pick s390x boot image for s390x Nodepool for type kubevirt

veera-damisetti:
Added support to pick s390x boot image for s390x Nodepool for type kubevirt

This adds changes to defaultImage function to pick boot image based on nodepool arch

**why we need it:**
- Currently we don't have support to create HCP with s390x computes , with type kubevirt.
- To enable this 
     - we need to add support to create s390x nodepool  -> https://github.com/openshift/hypershift/pull/6239
     - Need to change the way of picking the boot image for VMs.
- This PR will take care of changing the way of picking the boot image for VMs.

**What this PR does**

- Updated defaultImage function to pick the default boot image based on the architecture of nodepool
- This will enable user to create s390x nodepools which will have s390x nodes

## Checklist:

### DIFF
--- a/hypershift-operator/controllers/nodepool/kubevirt/kubevirt.go
+++ b/hypershift-operator/controllers/nodepool/kubevirt/kubevirt.go
@@ -38,10 +38,18 @@ var (
 	}
 )
 
-func defaultImage(releaseImage *releaseinfo.ReleaseImage) (string, string, error) {
-	arch, foundArch := releaseImage.StreamMetadata.Architectures["x86_64"]
+func defaultImage(nodePoolArch string, releaseImage *releaseinfo.ReleaseImage) (string, string, error) {
+	var archName string
+	switch nodePoolArch {
+	case hyperv1.ArchitectureS390X:
+		archName = hyperv1.ArchitectureS390X
+	default:
+		archName = hyperv1.ArchAliases[hyperv1.ArchitectureAMD64]
+	}
+	arch, foundArch := releaseImage.StreamMetadata.Architectures[archName]
+
 	if !foundArch {
-		return "", "", fmt.Errorf("couldn't find OS metadata for architecture %q", "x84_64")
+		return "", "", fmt.Errorf("couldn't find OS metadata for architecture %q", archName)
 	}
 
 	containerImage := arch.Images.Kubevirt.DigestRef
@@ -81,7 +89,7 @@ func GetImage(nodePool *hyperv1.NodePool, releaseImage *releaseinfo.ReleaseImage
 		return newBootImage(imageName, isHTTP), nil
 	}
 
-	imageName, imageHash, err := defaultImage(releaseImage)
+	imageName, imageHash, err := defaultImage(nodePool.Spec.Arch, releaseImage)
 	if err != nil && allowUnsupportedRHCOSVariants(nodePool) {
 		imageName, imageHash, err = openstack.OpenstackDefaultImage(releaseImage)
 		if err != nil {
--- a/hypershift-operator/controllers/nodepool/kubevirt/kubevirt_test.go
+++ b/hypershift-operator/controllers/nodepool/kubevirt/kubevirt_test.go
@@ -1501,3 +1501,68 @@ func generateNodeTemplate(options ...nodeTemplateOption) *capikubevirt.VirtualMa
 
 	return template
 }
+
+func TestDefaultImage(t *testing.T) {
+	// Create a mock ReleaseImage with architecture-specific Kubevirt images
+	ri := &releaseinfo.ReleaseImage{
+		StreamMetadata: &releaseinfo.CoreOSStreamMetadata{
+			Architectures: map[string]releaseinfo.CoreOSArchitecture{
+				hyperv1.ArchitectureS390X: {
+					Images: releaseinfo.CoreOSImages{
+						Kubevirt: releaseinfo.CoreOSKubevirtImages{
+							DigestRef: "quay.io/openshift/release@sha256:s390x1234",
+						},
+					},
+				},
+				hyperv1.ArchAliases[hyperv1.ArchitectureAMD64]: {
+					Images: releaseinfo.CoreOSImages{
+						Kubevirt: releaseinfo.CoreOSKubevirtImages{
+							DigestRef: "quay.io/openshift/release@sha256:x86_641234",
+						},
+					},
+				},
+			},
+		},
+	}
+
+	tests := []struct {
+		name           string
+		arch           string
+		expectedImage  string
+		expectedDigest string
+	}{
+		{
+			name:           "s390x architecture",
+			arch:           hyperv1.ArchitectureS390X,
+			expectedImage:  "quay.io/openshift/release@sha256:s390x1234",
+			expectedDigest: "sha256:s390x1234",
+		},
+		{
+			name:           "x86_64 architecture",
+			arch:           hyperv1.ArchitectureAMD64,
+			expectedImage:  "quay.io/openshift/release@sha256:x86_641234",
+			expectedDigest: "sha256:x86_641234",
+		},
+		{
+			name:           "unknown architecture falls back to x86_64",
+			arch:           "",
+			expectedImage:  "quay.io/openshift/release@sha256:x86_641234",
+			expectedDigest: "sha256:x86_641234",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			img, digest, err := defaultImage(tt.arch, ri)
+			if err != nil {
+				t.Fatalf("unexpected error for %s: %v", tt.arch, err)
+			}
+			if img != tt.expectedImage {
+				t.Errorf("got image %q, expected %q", img, tt.expectedImage)
+			}
+			if digest != tt.expectedDigest {
+				t.Errorf("got digest %q, expected %q", digest, tt.expectedDigest)
+			}
+		})
+	}
+}


veera-damisetti:
@enxebre 
This is the PR for enabling HCP installation with type kubevirt on Z , which is a requirement for https://github.com/openshift/hypershift/pull/6239

can you please look into this , thanks

hypershift-operator/controllers/nodepool/kubevirt/kubevirt.go
```diff
@@ -38,10 +38,16 @@ var (
 	}
 )
 
-func defaultImage(releaseImage *releaseinfo.ReleaseImage) (string, string, error) {
-	arch, foundArch := releaseImage.StreamMetadata.Architectures["x86_64"]
+func defaultImage(nodePool *hyperv1.NodePool, releaseImage *releaseinfo.ReleaseImage) (string, string, error) {
+	default_arch := "x86_64"
+	arch, foundArch := releaseImage.StreamMetadata.Architectures[default_arch]
+
+	if nodePool.Spec.Arch == "s390x" {
```
bryan-cox: Let's use https://github.com/openshift/hypershift/blob/dc604559b500f4ec35d1b0c43080170784323180/api/hypershift/v1beta1/nodepool_types.go#L17 instead of the string here.

hypershift-operator/controllers/nodepool/kubevirt/kubevirt.go
```diff
@@ -38,10 +38,16 @@ var (
...
 
-func defaultImage(releaseImage *releaseinfo.ReleaseImage) (string, string, error) {
-	arch, foundArch := releaseImage.StreamMetadata.Architectures["x86_64"]
+func defaultImage(nodePool *hyperv1.NodePool, releaseImage *releaseinfo.ReleaseImage) (string, string, error) {
+        default_arch := "x86_64"
+	arch, foundArch := releaseImage.StreamMetadata.Architectures[default_arch]
+
+	if nodePool.Spec.Arch == "s390x" {
+		arch, foundArch = releaseImage.StreamMetadata.Architectures["s390x"]
+	}
```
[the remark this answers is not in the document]
bryan-cox: @veera-damisetti what do you think about moving to this logic instead of the current logic?

veera-damisetti:
Hi @bryan-cox , Thanks for the quick review 
1. I can do the change to use  ArchitectureS390X   = "s390x"  
2. I didn't get  your point here _ what do you think about moving to this logic instead of the current logic?_ , can you please elaborate on which logic you are talking about

bryan-cox:
@veera-damisetti interesting, it moved my comment. This is what I was talking about - https://github.com/openshift/hypershift/pull/7058#discussion_r2433402170

veera-damisetti:
Hi @bryan-cox , 
regarding your comment https://github.com/openshift/hypershift/pull/7058#discussion_r2433402170
We cant rely on the below logic 
```
archName := "x86_64"
+	if nodePool.Spec.Arch != "" {
+		archName = nodePool.Spec.Arch
+	}
+	
+	arch, foundArch := releaseImage.StreamMetadata.Architectures[archName]
```

because, the arch format in **_nodePool.Spec.Arch_** is not always same as the format which  is available in  **_releaseImage.StreamMetadata_**.
for example, for x86 **_nodePool.Spec.Arch_** contains amd64 but **_releaseImage.StreamMetadata_** contains with x86_64

veera-damisetti:
@bryan-cox Can you please have a look here ? 
thanks

hypershift-operator/controllers/nodepool/kubevirt/kubevirt.go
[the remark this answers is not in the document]
clebs: This seems both simpler and more robust to me.
Regarding the string mismatch, it could be solved with a translation table.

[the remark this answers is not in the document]
veera-damisetti: We cant rely on the above  logic
because, the arch format in nodePool.Spec.Arch is not always same as the format which is available in releaseImage.StreamMetadata.
for example, for x86 nodePool.Spec.Arch contains amd64 but releaseImage.StreamMetadata contains with x86_64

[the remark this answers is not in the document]
clebs: I have thought about this and you are right that we need the correct constants for each arch type.
Nevertheless, the difficulty of adding more archs is a valid point.

What do you think of using a `switch`? It is a good middle ground where the issue you mention is gone, but adding a new case is easier:
```suggestion
func defaultImage(nodePool *hyperv1.NodePool, releaseImage *releaseinfo.ReleaseImage) (string, string, error) {
	var archName string
	switch nodePool.Spec.Arch {
	    case "s390x"
	    	archName = ArchitectureS390X
	    default:
	        archName = ArchitectureAmd64
	}

	arch, foundArch = releaseImage.StreamMetadata.Architectures[archName]
```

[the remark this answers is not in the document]
veera-damisetti: yeah @clebs , this approach looks good 
but we need to use archname as x86_64 for amd64 case ( which is default )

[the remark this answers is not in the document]
clebs: Sure, I did not check the specifics exactly. Please adjust values and constant names as needed.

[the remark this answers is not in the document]
veera-damisetti: Sure @clebs 
i will do the changes accordingly ... now 
thanks

hypershift-operator/controllers/nodepool/kubevirt/kubevirt.go
```diff
@@ -38,10 +38,18 @@ var (
...
 
-func defaultImage(releaseImage *releaseinfo.ReleaseImage) (string, string, error) {
-	arch, foundArch := releaseImage.StreamMetadata.Architectures["x86_64"]
+func defaultImage(nodePool *hyperv1.NodePool, releaseImage *releaseinfo.ReleaseImage) (string, string, error) {
+	var archName string
+	switch nodePool.Spec.Arch {
+	case hyperv1.ArchitectureS390X:
+		archName = hyperv1.ArchitectureS390X
+	default:
+		archName = "x86_64"
```
clebs: Does it make sense to add an `Architecture` const to `hyperv1` for this?

veera-damisetti: for x86  ? 
It is already there with hyperv1.ArchitectureAMD64
But we can't use that as we just discussed that this notation is a bit different in releaseImage.StreamMetadata.Architectures[

clebs: Ok, in that case use an unexported local constant defined in this file.

veera-damisetti: So something like this ?
const (
ArchitectureX86 = "x86_64"
)

hypershift-operator/controllers/nodepool/kubevirt/kubevirt.go
```diff
@@ -38,10 +42,18 @@ var (
 	}
 )
 
-func defaultImage(releaseImage *releaseinfo.ReleaseImage) (string, string, error) {
-	arch, foundArch := releaseImage.StreamMetadata.Architectures["x86_64"]
+func defaultImage(nodePool *hyperv1.NodePool, releaseImage *releaseinfo.ReleaseImage) (string, string, error) {
```
clebs: Sorry one last nit:
In general it is good to reduce function scope as much as possible, so I would pass `nodepool.Spec.Arch` as parameter instead of the whole nodepool.

veera-damisetti: Sure @clebs , i got your point , thanks for the input 
Updated the logic accordingly

veera-damisetti:
Thanks @clebs for the review

hypershift-operator/controllers/nodepool/kubevirt/kubevirt_test.go
```diff
@@ -1501,3 +1501,62 @@ func generateNodeTemplate(options ...nodeTemplateOption) *capikubevirt.VirtualMa
...
+			expectedImage:  "quay.io/openshift/release@sha256:s390x1234",
+			expectedDigest: "sha256:s390x1234",
+		},
+		{
+			name:           "x86_64 architecture",
+			arch:           hyperv1.ArchitectureAMD64,
+			expectedImage:  "quay.io/openshift/release@sha256:x86_641234",
+			expectedDigest: "sha256:x86_641234",
+		},
+	}
```
clebs, line 1552: Please add a case with an empty or non-existing arch to test that the function properly defaults to x86.

veera-damisetti: Sure @clebs Added one test case for this 
thanks

hypershift-operator/controllers/nodepool/kubevirt/kubevirt.go
```diff
@@ -25,6 +25,10 @@ import (
 	"kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
 )
 
+const (
+	ArchitectureX86 = "x86_64"
```
bryan-cox: Let's remove this and use https://github.com/openshift/hypershift/blob/7e919e276ff99be4dec5050778e77c099dffc239/api/hypershift/v1beta1/nodepool_types.go#L47

hypershift-operator/controllers/nodepool/kubevirt/kubevirt_test.go
```diff
@@ -1501,3 +1501,62 @@ func generateNodeTemplate(options ...nodeTemplateOption) *capikubevirt.VirtualMa
...
+				ArchitectureX86: {
+					Images: releaseinfo.CoreOSImages{
+						Kubevirt: releaseinfo.CoreOSKubevirtImages{
+							DigestRef: "quay.io/openshift/release@sha256:x86_641234",
+						},
+					},
+				},
+			},
+		},
+	}
```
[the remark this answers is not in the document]
bryan-cox: Agree

hypershift-operator/controllers/nodepool/kubevirt/kubevirt_test.go
clebs: Thanks!

PR #7065, opened 2025-10-16
OCPBUGS-62959: fix(cvo): support image overrides annotation for CVO deployment

muraee:
## Summary

This PR fixes a bug where the `hypershift.openshift.io/image-overrides` annotation was not working properly for the cluster-version-operator deployment.

## Problem

When users added an annotation like `hypershift.openshift.io/image-overrides=cluster-version-operator=some-image` on the HostedCluster, the CVO deployment continued using the default image from the release payload instead of the overridden image.

### Root Cause

The `hostedcontrolplane_controller.go` was using `p.ControlPlaneImage` (which was set to `util.HCPControlPlaneReleaseImage(hcp)`) to fetch the image digest. This bypassed the `p.ReleaseImage` field which correctly included the override from `releaseImageProvider`.

## Changes

- Removed redundant `ControlPlaneImage` field from `CVOParams` struct
- Updated `reconcileClusterVersionOperator` to use `p.ReleaseImage` which includes image overrides from the annotation
- Added clarifying comments explaining the override logic

## Test Plan

With this change:
1. Add annotation to HostedCluster: `hypershift.openshift.io/image-overrides=cluster-version-operator=custom-image:latest`
2. The CVO deployment will now correctly use `custom-image:latest` instead of the default release image

## Verification

- [x] Code compiles successfully
- [x] Change follows conventional commit standards
- [x] Logic verified through code inspection

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
@@ -3822,8 +3822,8 @@ func (r *HostedControlPlaneReconciler) reconcileClusterVersionOperator(ctx conte
 	}
 
 	var (
-		controlPlaneReleaseImage string
-		dataPlaneReleaseImage    string
+		cvoImage              string
+		dataPlaneReleaseImage string
 	)
 	// The CVO prepare-payload script needs the ReleaseImage digest for disconnected environments
 	pullSecret := common.PullSecret(hcp.Namespace)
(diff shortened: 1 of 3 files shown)


csrwng:
The fix is hard to follow because it's easy to get confused about which image is which. In my mind there's potentially 3 images involved:
1 - The release image/payload that should be used for control plane workloads
2 - The release image/payload that should be used for data plane workloads
3 - The CVO image

Here https://github.com/openshift/hypershift/pull/7065/files#diff-3d4ffda0834e09a59a4d510523932e20b9234bcb9816e40258a0fb32a87a7006R3842
It's not clear to me why we should be comparing the image you're using to override the CVO with the release image specified in the hcp resource. We should be comparing it to the control plane release image.

muraee:
Thanks for the feedback! You're absolutely correct - the comparison was confusing because it was comparing the CVO image with the data plane release image.

I've fixed this by changing the comparison to use `util.HCPControlPlaneReleaseImage(hcp)` instead of `p.ReleaseImage`. Now the code properly compares:
1. Control plane release image (which may be overridden via `spec.controlPlaneReleaseImage`)
2. Data plane release image (from `spec.releaseImage`)

This correctly determines when we need separate digests for control plane vs data plane workloads.

The change has been pushed and is now part of the latest commit.

---

csrwng:
The field `ReleaseImage` in the params struct is misleading as well. If it really means "CVO image", then we should name it as such.

muraee:
@csrwng You're absolutely right! I've renamed the `ReleaseImage` field to `CVOImage` in the `CVOParams` struct to make it clear that it represents the cluster-version-operator image, not a release image reference. This makes the code much easier to understand.

The changes include:
- Renamed `CVOParams.ReleaseImage` → `CVOParams.CVOImage`
- Updated all references to use the new field name
- Updated error messages to be more specific (e.g., "failed to get CVO image digest")

Code compiles successfully with these changes.

---

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -3831,13 +3831,16 @@ func (r *HostedControlPlaneReconciler) reconcileClusterVersionOperator(ctx conte
 		return fmt.Errorf("failed to get pull secret for namespace %s: %w", hcp.Namespace, err)
 	}
 	pullSecretBytes := pullSecret.Data[corev1.DockerConfigJsonKey]
-	_, controlPlaneReleaseImageRef, err := r.ImageMetadataProvider.GetDigest(ctx, p.ControlPlaneImage, pullSecretBytes)
+	// Use p.CVOImage which includes image overrides from the annotation if specified
+	_, controlPlaneReleaseImageRef, err := r.ImageMetadataProvider.GetDigest(ctx, p.CVOImage, pullSecretBytes)
```
csrwng: This needs further fixing/clarification so it's completely obvious in code to which of the images we're referring to. I don't think the controlPlaneReleaseImageRef should come from the CVO image.

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
csrwng: The ReconcileDeployment function may not need the controlPlaneReleaseImage but rather the CVOImage

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
muraee: Done. Renamed the variable to `cvoImage` throughout the code to clearly indicate it represents the CVO container image, not a release image reference.

---

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
muraee: Done. Updated ReconcileDeployment to use `cvoImage` parameter name instead of `controlPlaneReleaseImage`, making it clear this is the CVO container image.

---

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -3822,34 +3822,37 @@ func (r *HostedControlPlaneReconciler) reconcileClusterVersionOperator(ctx conte
...
+	// to determine if we need separate digests for each
+	if util.HCPControlPlaneReleaseImage(hcp) != hcp.Spec.ReleaseImage {
 		_, dataPlaneReleaseImageRef, err := r.ImageMetadataProvider.GetDigest(ctx, hcp.Spec.ReleaseImage, pullSecretBytes)
 		if err != nil {
 			return fmt.Errorf("failed to get release image digest %s: %w", hcp.Spec.ReleaseImage, err)
 		}
 		dataPlaneReleaseImage = dataPlaneReleaseImageRef.String()
 	} else {
-		dataPlaneReleaseImage = controlPlaneReleaseImage
+		dataPlaneReleaseImage = cvoImage
```
csrwng: This looks wrong. Shouldn't it be util.HCPControlPlaneReleaseImage(hcp) ?

muraee: Good catch! You're absolutely right. The data plane release image should always be the digest of `hcp.Spec.ReleaseImage`, not the CVO container image digest. I've simplified the code to always fetch the data plane release image digest directly, removing the unnecessary conditional logic.

---

PR #7076, opened 2025-10-21
OCPBUGS-63205: fix(e2e): correct readOnlyRootFilesystem test logic

muraee:

The EnsureReadOnlyRootFilesystem test was incorrectly enforcing exempted containers to have readOnlyRootFilesystem=false instead of simply skipping them. Fixed to properly enforce readOnlyRootFilesystem=true for all non-exempted containers.

- Replaced auditedContainersHas with findAuditedContainers function
- Fixed test logic to skip exempted containers correctly
- Improved assertion messages for clarity

Assisted-by: Claude Sonnet 4.5 (via Cursor)

Fixes 

## Checklist:

### DIFF
--- a/test/e2e/util/util.go
+++ b/test/e2e/util/util.go
@@ -1624,18 +1624,14 @@ type labelSelector struct {
 	value string
 }
 
-// auditedContainersHas checks the given map to see if the container name exists in it; if the map is empty, always return true
-func auditedContainersHas(container corev1.Container, auditedContainers map[string]struct{}) bool {
-	if auditedContainers == nil {
-		return false
+// findAuditedContainers returns the map of audited containers for a pod, or nil if the pod is not audited
+func findAuditedContainers(pod *corev1.Pod, auditedContainersSelector map[labelSelector]map[string]struct{}) map[string]struct{} {
+	for selector, containers := range auditedContainersSelector {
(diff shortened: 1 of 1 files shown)


muraee:
@coderabbitai review

muraee:
@auto-cc

test/e2e/util/util.go
```diff
@@ -1624,18 +1624,14 @@ type labelSelector struct {
...
-// auditedContainersHas checks the given map to see if the container name exists in it; if the map is empty, always return true
-func auditedContainersHas(container corev1.Container, auditedContainers map[string]struct{}) bool {
-	if auditedContainers == nil {
-		return false
-	}
-
-	if len(auditedContainers) == 0 {
-		return true
+// findAuditedContainers returns the map of audited containers for a pod, or nil if the pod is not audited
+func findAuditedContainers(pod *corev1.Pod, auditedAppContainersNoRORFS map[labelSelector]map[string]struct{}) map[string]struct{} {
```
sjenning: `auditedAppContainersNoRORFS` not a good name for this parameter since it is also called from `AppContainerNoTmpDir .  Maybe `podSelector` so something like that.

muraee: done

sjenning:
external-dns got throttled on e2e-aks

/retest-required

PR #7077, opened 2025-10-21
CNTRLPLANE-1137: Add e2e test for ARO HCP with Cilium Network Policies

wewang58:
This PR implements cilium installation and check basic network policy scenarios test in aks env.

Fixes [https://issues.redhat.com/browse/CNTRLPLANE-1137](https://issues.redhat.com/browse/CNTRLPLANE-1137)

- Create e2e test that validates Cilium Network Policies functionality on ARO HCP guest clusters
- Test should cover basic network policy scenarios (ingress/egress rules, pod-to-pod communication)
- Ensure test runs as part of ARO e2e test suite without being skipped
- Test should validate that network policies are properly enforced in the guest cluster.
- [test passed in test job](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_release/71182/rehearse-71182-periodic-ci-openshift-hypershift-release-4.21-periodics-e2e-aks/1988029793001541632) 
- InstallCilium
- WaitForCiliumPodsReady
- DeployCiliumConnectivityTest
- WaitForConnectivityTestPodsReady
- WaitForConnectivityTestCompletion

## Checklist:

### DIFF
--- a/test/e2e/cilium_network_policy_test.go
+++ b/test/e2e/cilium_network_policy_test.go
@@ -1,5 +1,4 @@
 //go:build e2e
-// +build e2e
 
 package e2e
 
--- a/test/e2e/util/cilium.go
+++ b/test/e2e/util/cilium.go
@@ -14,9 +14,9 @@ import (
 	"github.com/openshift/hypershift/support/azureutil"
 	hyperutil "github.com/openshift/hypershift/support/util"
 
+	configv1 "github.com/openshift/api/config/v1"
 	securityv1 "github.com/openshift/api/security/v1"
 
-	configv1 "github.com/openshift/api/config/v1"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -289,7 +289,9 @@ func InstallCilium(t *testing.T, ctx context.Context, guestClient crclient.Clien
 
 				t.Logf("CiliumConfig has correct clusterPoolIPv4MaskSize: %d", maskSize)
 				return true
-			}, CiliumShortTimeout, CiliumDefaultPollInterval).Should(BeTrue(), "CiliumConfig should have correct configuration")			// Wait for operator to create DaemonSet
+			}, CiliumShortTimeout, CiliumDefaultPollInterval).Should(BeTrue(), "CiliumConfig should have correct configuration")
+
+			// Wait for operator to create DaemonSet
 			t.Log("Waiting for Cilium DaemonSet to be created by operator")
 			var ciliumDaemonSet *appsv1.DaemonSet
 			g.Eventually(func() bool {
@@ -459,24 +461,24 @@ func EnsureCiliumConnectivityTestResources(t *testing.T, ctx context.Context, gu
 		for _, pod := range podList.Items {
 			if pod.Status.Phase != corev1.PodRunning {
 				failedPods = append(failedPods, fmt.Sprintf("%s (phase: %s)", pod.Name, pod.Status.Phase))
-			
 			}
 		}
 
-					if len(failedPods) > 0 {
-						t.Errorf("Found %d pods not in Running phase: %v", len(failedPods), failedPods)
-					} else {
-						t.Logf("All %d connectivity test pods are running successfully", len(podList.Items))
-					}
-				}) // Closes the anonymous function for "WaitForConnectivityTestCompletion"
-		
-			t.Log("Cilium connectivity test completed successfully")
-		
-			return func() {
-				CleanupCiliumConnectivityTestResources(ctx, t, guestClient)
-			}
+		if len(failedPods) > 0 {
+			t.Errorf("Found %d pods not in Running phase: %v", len(failedPods), failedPods)
+		} else {
+			t.Logf("All %d connectivity test pods are running successfully", len(podList.Items))
 		}
-// getCiliumNetworkConfig extracts pod CIDR and host prefix from guest cluster
+	}) // Closes the anonymous function for "WaitForConnectivityTestCompletion"
+
+	t.Log("Cilium connectivity test completed successfully")
+
+	return func() {
+		CleanupCiliumConnectivityTestResources(ctx, t, guestClient)
+	}
+}
+
+// GetCiliumNetworkConfig extracts pod CIDR and host prefix from guest cluster
 // In dual-stack environments, this function ensures we return the IPv4 network
 func GetCiliumNetworkConfig(ctx context.Context, guestClient crclient.Client) (podCIDR string, hostPrefix int32) {
 	podCIDR = "10.132.0.0/14"
@@ -527,21 +529,21 @@ func CleanupCiliumConnectivityTestResources(ctx context.Context, t *testing.T, g
 		t.Logf("Warning: failed to delete namespace: %v", err)
 	}
 
-		t.Log("Cleanup completed")
+	t.Log("Cleanup completed")
 }
 
 func ptrBool(b bool) *bool {
 	return &b
 }
 
-// createCiliumConfig creates a CiliumConfig custom resource
+// CreateCiliumConfig creates a CiliumConfig custom resource
 func CreateCiliumConfig(podCIDR string, hostPrefix int32) *unstructured.Unstructured {
 	ciliumConfig := &unstructured.Unstructured{
 		Object: map[string]interface{}{
 			"apiVersion": fmt.Sprintf("%s/%s", CiliumConfigGroup, CiliumConfigVersion),
-			"kind": CiliumConfigKind,
+			"kind":       CiliumConfigKind,
 			"metadata": map[string]interface{}{
-				"name": CiliumConfigName,
+				"name":      CiliumConfigName,
 				"namespace": CiliumNamespace,
 			},
 			"spec": map[string]interface{}{
--- a/test/e2e/util/util.go
+++ b/test/e2e/util/util.go
@@ -36,6 +36,7 @@ import (
 	"github.com/openshift/hypershift/support/releaseinfo/registryclient"
 	hyperutil "github.com/openshift/hypershift/support/util"
 
+	configv1 "github.com/openshift/api/config/v1"
 	operatorv1 "github.com/openshift/api/operator/v1"
 	routev1 "github.com/openshift/api/route/v1"
 	configv1client "github.com/openshift/client-go/config/clientset/versioned"
@@ -53,7 +54,12 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
+	k8syaml "k8s.io/apimachinery/pkg/util/yaml"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	kubeclient "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
@@ -72,13 +78,6 @@ import (
 	"github.com/prometheus/common/expfmt"
 	"github.com/prometheus/common/model"
 	"go.uber.org/zap/zaptest"
-
-	configv1 "github.com/openshift/api/config/v1"
-	k8syaml "k8s.io/apimachinery/pkg/util/yaml"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-	"k8s.io/apimachinery/pkg/labels"
-	"k8s.io/apimachinery/pkg/runtime/schema"
-	"k8s.io/apimachinery/pkg/types"
 )
 
 var (
@@ -1201,8 +1200,6 @@ func EnsureNetworkPolicies(t *testing.T, ctx context.Context, c crclient.Client,
 	})
 }
 
-
-
 // ApplyYAMLFromURL downloads YAML content from a URL and applies it to the cluster.
 // If defaultNamespace is provided, it will be used for namespaced resources that don't have a namespace set.
 func ApplyYAMLFromURL(ctx context.Context, c crclient.Client, url string, defaultNamespace ...string) error {
@@ -1235,33 +1232,44 @@ func ApplyYAMLFromURL(ctx context.Context, c crclient.Client, url string, defaul
 			return fmt.Errorf("failed to decode YAML: %w", err)
 		}
 
-		if obj.GetKind() == "" {
+		if len(obj.Object) == 0 {
 			continue
 		}
 
 		// Set default namespace if provided and the resource doesn't have one
 		if len(defaultNamespace) > 0 && obj.GetNamespace() == "" {
-			// Check if this is a namespaced resource (not cluster-scoped like ClusterRole)
-			kind := obj.GetKind()
-			isClusterScoped := kind == "ClusterRole" || kind == "ClusterRoleBinding" ||
-				kind == "CustomResourceDefinition" || kind == "Namespace"
+			isNamespaced := false
+			gvk := obj.GroupVersionKind()
 
-			if !isClusterScoped {
+			// Check if the resource is namespaced using RESTMapper
+			mapping, err := c.RESTMapper().RESTMapping(gvk.GroupKind(), gvk.Version)
+			if err != nil {
+				return fmt.Errorf("failed to get mapping for %s: %w", gvk, err)
+			}
+
+			if mapping.Scope.Name() == meta.RESTScopeNameNamespace {
+				isNamespaced = true
+			}
+
+			if isNamespaced {
 				obj.SetNamespace(defaultNamespace[0])
 			}
 		}
 
-		// Create the resource
-		err = c.Create(ctx, obj)
-		if err != nil && !apierrors.IsAlreadyExists(err) {
-			return fmt.Errorf("failed to create %s/%s: %w", obj.GetKind(), obj.GetName(), err)
+		// Apply the resource using Server-Side Apply
+		patchOpts := []crclient.PatchOption{
+			crclient.ForceOwnership,
+			crclient.FieldOwner("hypershift-e2e"),
+		}
+
+		err = c.Patch(ctx, obj, crclient.Apply, patchOpts...)
+		if err != nil {
+			return fmt.Errorf("failed to apply %s/%s: %w", obj.GetKind(), obj.GetName(), err)
 		}
 	}
 
 	return nil
 }
-
-
 
 // EnsureNodesRuntime ensures that all nodes in the NodePool have the expected runtime handlers.
 // This is only supported on 4.18+ when the default runtime is changed to crun.
--- a/.tekton/pipelines/common-operator-build.yaml
+++ b/.tekton/pipelines/common-operator-build.yaml
@@ -89,7 +89,7 @@ spec:
       - name: name
         value: init
       - name: bundle
-        value: quay.io/konflux-ci/tekton-catalog/task-init:0.2@sha256:bbf313b09740fb39b3343bc69ee94b2a2c21d16a9304f9b7c111c305558fc346
+        value: quay.io/konflux-ci/tekton-catalog/task-init:0.2@sha256:75b88ee5e134a22ee35eb974808dfe6a63693115fa445208a9060a7175b448cf
       - name: kind
         value: task
       resolver: bundles
@@ -110,7 +110,7 @@ spec:
       - name: name
         value: git-clone-oci-ta
       - name: bundle
-        value: quay.io/konflux-ci/tekton-catalog/task-git-clone-oci-ta:0.1@sha256:3a920a83fc0135aaae2730fe9d446eb2da2ffc9d63a34bceea04afd24653bdee
+        value: quay.io/konflux-ci/tekton-catalog/task-git-clone-oci-ta:0.1@sha256:0a89e1a6304076525e9766f63a4cd006763d21d5aca6863281fc427537a23c6f
       - name: kind
         value: task
       resolver: bundles
@@ -141,7 +141,7 @@ spec:
       - name: name
         value: prefetch-dependencies-oci-ta
       - name: bundle
-        value: quay.io/konflux-ci/tekton-catalog/task-prefetch-dependencies-oci-ta:0.2@sha256:970285e3b0495961199523b566e0dd92ec2e29bedbcf61d8fc67106b06d0f923
+        value: quay.io/konflux-ci/tekton-catalog/task-prefetch-dependencies-oci-ta:0.2@sha256:3fa0204a481044b21f0e784ce39cbd25e8fb49c664a5458f3eef351fff1c906e
       - name: kind
         value: task
       resolver: bundles
@@ -183,7 +183,7 @@ spec:
       - name: name
         value: buildah-remote-oci-ta
       - name: bundle
-        value: quay.io/konflux-ci/tekton-catalog/task-buildah-remote-oci-ta:0.5@sha256:4f348fa6e0b5d7f976ae6cbb75f4e93bc0be1188f074e39bcae3b5fd71796a3f
+        value: quay.io/konflux-ci/tekton-catalog/task-buildah-remote-oci-ta:0.7@sha256:b24359805297760c87cbce7b4c378267bc83aa1b9a3ac8431829f80bc26ed5d7
       - name: kind
         value: task
       resolver: bundles
@@ -212,7 +212,7 @@ spec:
       - name: name
         value: build-image-index
       - name: bundle
-        value: quay.io/konflux-ci/tekton-catalog/task-build-image-index:0.1@sha256:79784d53749584bc5a8de32142ec4e2f01cdbf42c20d94e59280e0b927c8597d
+        value: quay.io/konflux-ci/tekton-catalog/task-build-image-index:0.1@sha256:d455c28a6ae9f6ed40504e65e69b75ab147bb685c9fd606e6ffc3bad6f722bf4
       - name: kind
         value: task
       resolver: bundles
@@ -235,7 +235,7 @@ spec:
       - name: name
         value: deprecated-image-check
       - name: bundle
-        value: quay.io/konflux-ci/tekton-catalog/task-deprecated-image-check:0.5@sha256:f59175d9a0a60411738228dfe568af4684af4aa5e7e05c832927cb917801d489
+        value: quay.io/konflux-ci/tekton-catalog/task-deprecated-image-check:0.5@sha256:462baed733dfc38aca5395499e92f19b6f13a74c2e88fe5d86c3cffa2f899b57
       - name: kind
         value: task
       resolver: bundles
@@ -257,7 +257,7 @@ spec:
       - name: name
         value: clair-scan
       - name: bundle
-        value: quay.io/konflux-ci/tekton-catalog/task-clair-scan:0.3@sha256:a7cc183967f89c4ac100d04ab8f81e54733beee60a0528208107c9a22d3c43af
+        value: quay.io/konflux-ci/tekton-catalog/task-clair-scan:0.3@sha256:ee558db6af779ab162163ec88f288a5c1b2d5f70c3361f3690a474866e3bdc74
       - name: kind
         value: task
       resolver: bundles
@@ -281,7 +281,7 @@ spec:
       - name: name
         value: ecosystem-cert-preflight-checks
       - name: bundle
-        value: quay.io/konflux-ci/tekton-catalog/task-ecosystem-cert-preflight-checks:0.2@sha256:9568c51a5158d534248908b9b561cf67d2826ed4ea164ffd95628bb42380e6ec
+        value: quay.io/konflux-ci/tekton-catalog/task-ecosystem-cert-preflight-checks:0.2@sha256:04f75593558f79a27da2336400bc63d460bf0c5669e3c13f40ee2fb650b1ad1e
       - name: kind
         value: task
       resolver: bundles
@@ -307,7 +307,7 @@ spec:
       - name: name
         value: sast-snyk-check-oci-ta
       - name: bundle
-        value: quay.io/konflux-ci/tekton-catalog/task-sast-snyk-check-oci-ta:0.4@sha256:181d63c126e3119a9d57b8feed4eb66a875b5208c3e90724c22758e65dca8733
+        value: quay.io/konflux-ci/tekton-catalog/task-sast-snyk-check-oci-ta:0.4@sha256:8ad28b7783837a24acbc9a8494c935e796e591ce476085ad5899bebd7e53f077
       - name: kind
         value: task
       resolver: bundles
@@ -333,7 +333,7 @@ spec:
       - name: name
         value: clamav-scan
       - name: bundle
-        value: quay.io/konflux-ci/tekton-catalog/task-clamav-scan:0.3@sha256:b0bd59748cda4a7abf311e4f448e6c1d00c6b6d8c0ecc1c2eb33e08dc0e0b802
+        value: quay.io/konflux-ci/tekton-catalog/task-clamav-scan:0.3@sha256:f3d2d179cddcc07d0228d9f52959a233037a3afa2619d0a8b2effbb467db80c3
       - name: kind
         value: task
       resolver: bundles
@@ -394,7 +394,7 @@ spec:
       - name: name
         value: coverity-availability-check
       - name: bundle
-        value: quay.io/konflux-ci/tekton-catalog/task-coverity-availability-check:0.2@sha256:db2b267dc15e4ed17f704ee91b8e9b38068e1a35b1018a328fdca621819d74c6
+        value: quay.io/konflux-ci/tekton-catalog/task-coverity-availability-check:0.2@sha256:36400873d3031df128c55aa71ee11d322c3e55fd8f13dc5779098fbc117c0aa3
       - name: kind
         value: task
       resolver: bundles
@@ -420,7 +420,7 @@ spec:
       - name: name
         value: sast-shell-check-oci-ta
       - name: bundle
-        value: quay.io/konflux-ci/tekton-catalog/task-sast-shell-check-oci-ta:0.1@sha256:bf7bdde00b7212f730c1356672290af6f38d070da2c8a316987b5c32fd49e0b9
+        value: quay.io/konflux-ci/tekton-catalog/task-sast-shell-check-oci-ta:0.1@sha256:d44336d7bcbd1f7cedee639357a493bd1f661e2859e49e11a34644bdf6819c4e
       - name: kind
         value: task
       resolver: bundles
@@ -446,7 +446,7 @@ spec:
       - name: name
         value: sast-unicode-check-oci-ta
       - name: bundle
-        value: quay.io/konflux-ci/tekton-catalog/task-sast-unicode-check-oci-ta:0.3@sha256:a2bde66f6b4164620298c7d709b8f08515409404000fa1dc2260d2508b135651
+        value: quay.io/konflux-ci/tekton-catalog/task-sast-unicode-check-oci-ta:0.3@sha256:e5a8d3e8e7be7246a1460385b95c084ea6e8fe7520d40fe4389deb90f1bf5176
       - name: kind
         value: task
       resolver: bundles
@@ -468,7 +468,7 @@ spec:
       - name: name
         value: apply-tags
       - name: bundle
-        value: quay.io/konflux-ci/tekton-catalog/task-apply-tags:0.2@sha256:f44be1bf0262471f2f503f5e19da5f0628dcaf968c86272a2ad6b4871e708448
+        value: quay.io/konflux-ci/tekton-catalog/task-apply-tags:0.2@sha256:e4017ec351a0891ef95989f35bd20b8c3f091fa1a3da364c4d4e975e99f3063c
       - name: kind
         value: task
       resolver: bundles
@@ -491,7 +491,7 @@ spec:
       - name: name
         value: push-dockerfile-oci-ta
       - name: bundle
-        value: quay.io/konflux-ci/tekton-catalog/task-push-dockerfile-oci-ta:0.1@sha256:14fba04580b236e4206a904b86ee2fd8eeaa4163f7619a9c2602d361e4f74c51
+        value: quay.io/konflux-ci/tekton-catalog/task-push-dockerfile-oci-ta:0.1@sha256:08bba4a659ecd48f871bef00b80af58954e5a09fcbb28a1783ddd640c4f6535e
       - name: kind
         value: task
       resolver: bundles
@@ -508,7 +508,7 @@ spec:
       - name: name
         value: rpms-signature-scan
       - name: bundle
-        value: quay.io/konflux-ci/tekton-catalog/task-rpms-signature-scan:0.2@sha256:1b6c20ab3dbfb0972803d3ebcb2fa72642e59400c77bd66dfd82028bdd09e120
+        value: quay.io/konflux-ci/tekton-catalog/task-rpms-signature-scan:0.2@sha256:20eb21c60522a12198205f70b9c58cb5d71db561a255a3ba1ced56ae7b4af270
       - name: kind
         value: task
       resolver: bundles


wewang58:
Now :   Pod cilium-olm-844cdcc94b-tc9xh: Phase=Running, Ready=true is running, but  DaemonSet cilium status: Desired=2, Current=0, Ready=0, Available=0, reason is nodes are not ready, should check more.

test/e2e/util/util.go
```diff
@@ -1167,6 +1193,454 @@ func EnsureNetworkPolicies(t *testing.T, ctx context.Context, c crclient.Client,
 	})
 }
 
+// InstallCiliumNetworkPolicies validates that Cilium network policies are properly enforced
+// in ARO HCP guest clusters. This test covers:Verifying Cilium installation
+func InstallCiliumNetworkPolicies(t *testing.T, ctx context.Context, guestClient crclient.Client, hostedCluster *hyperv1.HostedCluster) {
+	t.Run("InstallCiliumNetworkPolicies", func(t *testing.T) {
+		if hostedCluster.Spec.Platform.Type != hyperv1.AzurePlatform {
```
bryan-cox: I think we could just drop this if statement since the one on L1204 should be sufficient.

test/e2e/util/util.go
```diff
@@ -1167,6 +1193,454 @@ func EnsureNetworkPolicies(t *testing.T, ctx context.Context, c crclient.Client,
...
+					t.Logf("Failed to get DaemonSet status: %v", err)
+					return false
+				}
+
+				// Check if at least one pod is ready
+				if ciliumDaemonSet.Status.NumberReady > 0 {
+					return true
+				}
+				// // Diagnose pods if any exist but aren't ready
+				// podList := &corev1.PodList{}
```
bryan-cox: This should all be removed if its not needed.

test/e2e/cilium_network_policy_test.go
```diff
@@ -0,0 +1,262 @@
...
+			} else {
+				t.Logf("All %d connectivity test pods are running successfully", len(podList.Items))
+			}
+		})
+
+		t.Log("Cilium connectivity test completed successfully")
+	}).Execute(&clusterOpts, globalOpts.Platform, globalOpts.ArtifactDir, "cilium-connectivity", globalOpts.ServiceAccountSigningKey)
+}
+
+// ptrBool returns a pointer to a bool
```
bryan-cox: I would drop this and use the k8s library for this. There are examples in our codebase for this.

test/e2e/cilium_network_policy_test.go
```diff
@@ -0,0 +1,262 @@
...
+	if globalOpts.Platform != hyperv1.AzurePlatform {
+		t.Skip("Skipping test because it requires Azure platform")
+	}
+
+	e2eutil.NewHypershiftTest(t, ctx, func(t *testing.T, g Gomega, mgtClient crclient.Client, hostedCluster *hyperv1.HostedCluster) {
+		if !azureutil.IsAroHCP() {
+			t.Skip("test only supported on ARO HCP clusters")
+		}
+		//Skip if NetworkType is not Other (Cilium)
+		if hostedCluster.Spec.Networking.NetworkType != hyperv1.Other {
```
mgencur: Where is this NetworkType set on HostedCluster? For conformance tests, we do that by passing `--network-type` to `hypershift cluster cluster` but this is E2E test suite.

test/e2e/cilium_network_policy_test.go
```diff
@@ -0,0 +1,262 @@
...
+			}
+			if err := guestClient.Delete(ctx, ns); err != nil && !apierrors.IsNotFound(err) {
+				t.Logf("Warning: failed to delete namespace: %v", err)
+			}
+
+			t.Log("Cleanup completed")
+		}
+		defer cleanup()
+
+		t.Run("WaitForCiliumPodsReady", func(t *testing.T) {
```
mgencur: Could this wait be moved to the place where Cilium is installed? It's better to fail early when installing. Here we can probably just check once that the pods are there.

test/e2e/util/hypershift_framework.go
```diff
@@ -124,7 +124,7 @@ func (h *hypershiftTest) Execute(opts *PlatformAgnosticOptions, platform hyperv1
 // runs before each test.
 func (h *hypershiftTest) before(hostedCluster *hyperv1.HostedCluster, opts *PlatformAgnosticOptions, platform hyperv1.PlatformType) {
 	h.Run("ValidateHostedCluster", func(t *testing.T) {
-		if platform != hyperv1.NonePlatform {
+		if platform != hyperv1.NonePlatform && hostedCluster.Spec.Networking.NetworkType != hyperv1.Other {
```
mgencur, line 128: I would like to know where the NetworkType is set (see my previous comment). 
My feeling is that there could be a flag to install Cilium instead of the default CNI. Then we could get rid of these specific checks for azureutil.IsAroHCP() and platform == hyperv1.AzurePlatform and 		if platform == hyperv1.AzurePlatform && azureutil.IsAroHCP() && hostedCluster.Spec.Networking.NetworkType == hyperv1.Other
We'd simply install Cilium when required.
I also think that this test can be usable on other platforms, not just Azure/ARO. This test could be run also on AWS, etc. So, having the option to choose Cilium more speficically would be useful. Also, we might need  to add Tigera Calico. This will also have NetworkType==Other. The current checks would then allow running the Cilium test with Tigera Calico, which is wrong.

test/e2e/util/util.go
```diff
@@ -1167,6 +1193,454 @@ func EnsureNetworkPolicies(t *testing.T, ctx context.Context, c crclient.Client,
 	})
 }
 
+// InstallCiliumNetworkPolicies validates that Cilium network policies are properly enforced
+// in ARO HCP guest clusters. This test covers:Verifying Cilium installation
+func InstallCiliumNetworkPolicies(t *testing.T, ctx context.Context, guestClient crclient.Client, hostedCluster *hyperv1.HostedCluster) {
```
mgencur: This installs the whole Cilium, not just network policies. Maybe this function could be called just "InstallCilium" ?

test/e2e/util/util.go
```diff
@@ -77,6 +81,28 @@ import (
 	"go.uber.org/zap/zaptest"
 )
 
+const (
+	// CiliumVersion is the version of Cilium used for installation and connectivity tests
+	CiliumVersion = "1.15.1"
```
mgencur: Maybe there's a better way to put this constant. I just can't find any.

test/e2e/util/hypershift_framework.go
mgencur: OK. I see it here: https://github.com/openshift/release/pull/71182/files
In that case, the "Other" network stack will be used for all hosted clusters that are created in the E2E test suite. Is this intentional? I thought just this single hosted cluster could install Cilium and run the test.

wewang58: This is just my debug pr:https://github.com/openshift/release/pull/71182/files , will close it after debugging it.

wewang58:
@mgencur  @bryan-cox  Updated codes, and all jobs passed, can you review again? thanks a lot.

mgencur:
I just saw the test name change but my remaining comments are not answered. Please see above.

wewang58:
> I just saw the test name change but my remaining comments are not answered. Please see above.

@mgencur  I am not familiar with Tigera Calico, let me research and check your idea about using flag, thanks

test/e2e/util/hypershift_framework.go
heliubj18: @mgencur I don’t quite understand what you mean by this comment. I think this check is correct and necessary — for HostedClusters with the Other network type, we should skip the validation to avoid waiting for node readiness.

test/e2e/util/util.go
heliubj18: Yes, I think it’s better to read the value from the `CILIUM_VERSION` environment variable first, and use this constant only as a fallback. This way, when we update the version in the Prow job, we only need to modify the e2e job there, without requiring a new PR in Hypershift.

test/e2e/util/hypershift_framework.go
mgencur: Sorry. This should have been on line 140 below.

Unrelated comment that is applicable for line 127: The user can still choose hyperv1.Calico but the condition just checks for hyperv1.Other. So, currently it would be validating the cluster for Calico, and fail.

wewang58:
@mgencur  @bryan-cox @heliubj18  Updated and [cilium test passed](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_release/71182/rehearse-71182-periodic-ci-openshift-hypershift-release-4.21-periodics-e2e-aks/1988946534208638976 ) in my debug env, can you review again, thanks

test/e2e/cilium_network_policy_test.go
mgencur: I suppose this check is now redundant when we already check globalOpts.ExternalCNIProvider above?

test/e2e/util/hypershift_framework.go
```diff
@@ -137,6 +137,16 @@ func (h *hypershiftTest) before(hostedCluster *hyperv1.HostedCluster, opts *Plat
 				ValidateAuthenticationSpec(t, h.ctx, h.client, hostedCluster, opts.ExtOIDCConfig)
 			}
 		}
+		if hostedCluster.Spec.Networking.NetworkType == hyperv1.Other {
```
mgencur: Should we use the `globalOpts.ExternalCNIProvider` to check if Cillium is really desired?

test/e2e/util/util.go
```diff
@@ -77,6 +81,38 @@ import (
...
+var (
+	// CiliumVersion is the version of Cilium used for installation and connectivity tests
+	// Must be set via CILIUM_VERSION environment variable
+	CiliumVersion string
+)
+
+func init() {
+	CiliumVersion = os.Getenv("CILIUM_VERSION")
+	if CiliumVersion == "" {
+		panic("CILIUM_VERSION environment variable is not set")
```
mgencur: Hmm. So now the test suite always needs CILIUM_VERSION to be set? Since this is in "init" it looks like the test suite will fail before we even check if we want to install Cilium or not. Or am I mistaken?

mgencur: CodeRabbit had a similar comment.

test/e2e/util/util.go
```diff
@@ -1167,6 +1203,442 @@ func EnsureNetworkPolicies(t *testing.T, ctx context.Context, c crclient.Client,
...
+	t.Run("InstallCilium", func(t *testing.T) {
+		// if hostedCluster.Spec.Platform.Type != hyperv1.AzurePlatform {
+		// 	t.Skipf("test only supported on Azure platform, saw %s", hostedCluster.Spec.Platform.Type)
+		// }
+
+		if !azureutil.IsAroHCP() {
+			t.Skip("test only supported on ARO HCP clusters")
+		}
+
+		if hostedCluster.Spec.Networking.NetworkType != hyperv1.Other {
```
mgencur: Maybe we can remove this check? We already check it here: https://github.com/openshift/hypershift/pull/7077/files#diff-65fe136e12a7c2756931254c4f076ef0b5bc990b0fc13a8407647f421ddfa328R140
But if we want to keep this check, should we use the globalOpts.ExternalCNIProvider to check if Cillium is really desired?

wewang58:
> @wewang58 Thank you for the changes. Left some comments. There's still this basic question: Do we want to install Cilium just for that single test or for all tests? If we want to use it for all tests we will need to create another CI job to run it, IMO. If the setup was just for that single test we could run it side-by-side with other tests in existing jobs.

wewang58:
> @wewang58 Thank you for the changes. Left some comments. There's still this basic question: Do we want to install Cilium just for that single test or for all tests? If we want to use it for all tests we will need to create another CI job to run it, IMO. If the setup was just for that single test we could run it side-by-side with other tests in existing jobs.
@mgencur I think we can run it with other tests in existing jobs, @heliubj18 do you have any opinion?

test/e2e/util/util.go
wewang58: You are right, updated.

test/e2e/util/util.go
wewang58: Updated.

test/e2e/util/hypershift_framework.go
wewang58: Updated.

test/e2e/cilium_network_policy_test.go
wewang58: Deleted.

mgencur:
> @mgencur I think we can run it with other tests in existing jobs

Thanks for the updates! It looks good to me now. I just think wee need a new job for that because when `opts.ExternalCNIProvider == "cilium"` it will install Cillium in `before` phase for all tests. Do you have any specific job in mind where this could be added? Thanks

wewang58:
> > @mgencur I think we can run it with other tests in existing jobs
> 
> Thanks for the updates! It looks good to me now. I just think wee need a new job for that because when `opts.ExternalCNIProvider == "cilium"` it will install Cillium in `before` phase for all tests. Do you have any specific job in mind where this could be added? Thanks

I draft a job: https://github.com/openshift/release/pull/71383, but haven't had time to test yet， can you and @heliubj18  review it first？ thanks

wewang58:
@mgencur  @heliubj18  I use debug pr: https://github.com/openshift/release/pull/71182/files to run all tests，[ 22/228 cases failed](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_release/71182/rehearse-71182-periodic-ci-openshift-hypershift-release-4.21-periodics-e2e-aks/1989239426324107264), not sure if it's related the pr, can you help to check too? thanks

mgencur:
@wewang58 The failures are related to Cilium but it will take longer to investigate the root cause. For some of them it's clear:
*  TestCreateClusterCustomConfig/Main/EnsureCNOOperatorConfiguration: `Invalid value: "object": ovnKubernetesConfig is forbidden when networkType is not OVNKubernetes`
*  TestUpgradeControlPlane/Main/Wait_for_control_plane_components_to_complete_rollout: `RolloutComplete=False: WaitingForOperands(failed to get deployment ovnkube-control-plane: Deployment.apps "ovnkube-control-plane" not found`
* TestNodePool/HostedCluster2/ValidateHostedCluster/InstallCilium/InstallCilium - since these are node pool tests it looks like the nodes are not ready yet when we're trying to install Cilium in `before` phase. So Cilium installation fails.
* TestHAEtcdChaos/ValidateHostedCluster - this test uses NodePoolReplicas == 0 so it's the same root cause like TestNodePool
* AutoscalingTest - not sure

wewang58:
@mgencur Submitted prs to fix existed e2e issues
> * TestCreateClusterCustomConfig/Main/EnsureCNOOperatorConfiguration: `Invalid value: "object": ovnKubernetesConfig is forbidden when networkType is not OVNKubernetes`
Fixed pr: https://github.com/openshift/hypershift/pull/7244/files

> * TestUpgradeControlPlane/Main/Wait_for_control_plane_components_to_complete_rollout: `RolloutComplete=False: WaitingForOperands(failed to get deployment ovnkube-control-plane: Deployment.apps "ovnkube-control-plane" not found`
Fixed pr: https://github.com/openshift/hypershift/pull/7243/files

> * AutoscalingTest - not sure
- TestCreateClusterCustomConfig/Main/EnsureCNOOperatorConfiguration: Invalid value: "object": ovnKubernetesConfig is forbidden when networkType is not OVNKubernetes, fixed pr: https://github.com/openshift/hypershift/pull/7244/files
-  TestUpgradeControlPlane/Main/Wait_for_control_plane_components_to_complete_rollout: `RolloutComplete=False: WaitingForOperands(failed to get deployment ovnkube-control-plane: Deployment.apps "ovnkube-control-plane" not found`, fixed pr: https://github.com/openshift/hypershift/pull/7243/files
- * AutoscalingTest - Maybe the pr fixed the issue: https://github.com/openshift/hypershift/pull/6745

Continue debug left 2 issues.

wewang58:
@mgencur  Now finished all failure e2e test  analysis in the [job](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_release/71182/rehearse-71182-periodic-ci-openshift-hypershift-release-4.21-periodics-e2e-aks/1989239426324107264)
related fixed pr: https://github.com/openshift/hypershift/pull/7243; https://github.com/openshift/hypershift/pull/7244; https://github.com/openshift/hypershift/pull/7263, do you think if it's ok for pr to merge? thanks in advance.

test/e2e/util/hypershift_framework.go
```diff
@@ -137,6 +138,21 @@ func (h *hypershiftTest) before(hostedCluster *hyperv1.HostedCluster, opts *Plat
 				ValidateAuthenticationSpec(t, h.ctx, h.client, hostedCluster, opts.ExtOIDCConfig)
 			}
 		}
+		if opts.ExternalCNIProvider == "cilium" {
+			// Only install Cilium when there are worker nodes configured.
+			// The cilium-olm deployment requires worker nodes to schedule its pods.
+			// TestNodePool sets NodePoolReplicas=0 and creates NodePools later in individual tests,
+			// so we skip Cilium installation during the initial cluster validation phase.
+			if !util.IsPrivateHC(hostedCluster) && opts.NodePoolReplicas > 0 {
```
mgencur: Hmm. This condition seems to be new? 
Can you highlight what changes you did in the last "force-push"? I may have missed something. It would be better to push new commits incrementally to make it easier for the reviewer. Force-pushing to the same commit makes me review everything again :-/
Anyway, maybe it would be better to make the test fail fast if NodePoolReplicas is 0. Something like this:
```
if !util.IsPrivateHC(hostedCluster) {
   if opts.NodePoolReplicas == 0 {
      t.Fatal("NodePool replicas must be positive for Cilium to install")
   }
   guestClient := WaitForGuestClient(t, context.Background(), h.client, hostedCluster)
   ...
}
```
This would prevent the situation when we want cilium but it's not installed and we would find out only later through various tests failing.

wewang58: Sorry, confused you, update this reason is NodePoolReplicas=0  in https://github.com/openshift/hypershift/blob/main/test/e2e/nodepool_test.go#L211,  we need at least one worker node to install cilium-olm operator before individual tests run, so I updated another pr to fix  TestNodePool issue: https://github.com/openshift/hypershift/pull/7263/files

wewang58: @mgencur I misunderstood your comments before, now updated according to your [comments](https://github.com/openshift/hypershift/pull/7077#discussion_r2544768276), can you review it again, thanks.

mgencur: It looks good to me now. Except for the commit message. There's now a single commit with title: "test(e2e): Prompt failure when NodePoolReplicas is 0" . Should it be like that?

wewang58: @mgencur Fixed no body in committed message issue, please review again, thanks a lot.

test/e2e/cilium_network_policy_test.go
```diff
@@ -0,0 +1,257 @@
...
+		cleanup := func() {
+			t.Log("Cleaning up Cilium connectivity test resources")
+
+			// Delete SCC
+			scc := &securityv1.SecurityContextConstraints{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "cilium-test",
+				},
+			}
+			if err := guestClient.Delete(ctx, scc); err != nil && !apierrors.IsNotFound(err) {
```
jparrill: I think you can use `util.DeleteIfNeeded` function for this part.

test/e2e/cilium_network_policy_test.go
```diff
@@ -0,0 +1,257 @@
...
+	e2eutil.NewHypershiftTest(t, ctx, func(t *testing.T, g Gomega, mgtClient crclient.Client, hostedCluster *hyperv1.HostedCluster) {
+		if !azureutil.IsAroHCP() {
+			t.Skip("test only supported on ARO HCP clusters")
+		}
+		guestClient := e2eutil.WaitForGuestClient(t, ctx, mgtClient, hostedCluster)
+
+		testNamespace := "cilium-test"
+
+		// Cleanup function
+		cleanup := func() {
```
jparrill: Instead of putting an anonymous function here, I would create their own cleanup function under `test/e2e/util`. And not just for that, I would create an EnsureXXXXX function for the whole non-cleanup function, so the code would be more readable and the cleanup function could be reused.

test/e2e/cilium_network_policy_test.go
```diff
@@ -0,0 +1,257 @@
...
+				t.Logf("Warning: failed to delete SCC: %v", err)
+			}
+
+			// Delete namespace
+			ns := &corev1.Namespace{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: testNamespace,
+				},
+			}
+			if err := guestClient.Delete(ctx, ns); err != nil && !apierrors.IsNotFound(err) {
```
jparrill: Same as above

test/e2e/cilium_network_policy_test.go
```diff
@@ -0,0 +1,257 @@
...
+				t.Logf("Warning: failed to delete namespace: %v", err)
+			}
+
+			t.Log("Cleanup completed")
+		}
+		defer cleanup()
+
+		t.Run("CheckCiliumPodsRunning", func(t *testing.T) {
```
jparrill: As I mentioned in another comment, Isolate the whole E2E without the cleanup in their own function and call it EnsureXXXXXX.

test/e2e/cilium_network_policy_test.go
```diff
@@ -0,0 +1,257 @@
...
+			g := NewWithT(t)
+
+			t.Log("Check cilium pods to be running")
+			g.Eventually(func() bool {
+				podList := &corev1.PodList{}
+				err := guestClient.List(ctx, podList, crclient.InNamespace("cilium"))
+				if err != nil {
```
jparrill: - Use gomega for conditional validations.
- Create const objects for Cilium verbs

test/e2e/cilium_network_policy_test.go
```diff
@@ -0,0 +1,257 @@
...
+
+			t.Log("Cleanup completed")
+		}
+		defer cleanup()
+
+		t.Run("CheckCiliumPodsRunning", func(t *testing.T) {
+			g := NewWithT(t)
+
+			t.Log("Check cilium pods to be running")
+			g.Eventually(func() bool {
```
jparrill: maybe you can use here `waitForDeploymentAvailable` function from test/e2e/utils?

test/e2e/cilium_network_policy_test.go
```diff
@@ -0,0 +1,257 @@
...
+						if condition.Type == corev1.PodReady && condition.Status == corev1.ConditionTrue {
+							ready = true
+							break
+						}
+					}
+					if !ready {
+						notReadyPods = append(notReadyPods, fmt.Sprintf("%s (phase: %s)", pod.Name, pod.Status.Phase))
+					}
+				}
+				if len(notReadyPods) > 0 {
```
jparrill: The deployment verification could pass with at least 1 replica running, what about if other replicas are not available? It's there any functional issue at Cilium level?

test/e2e/cilium_network_policy_test.go
```diff
@@ -0,0 +1,257 @@
...
+				AllowHostNetwork:         true,
+				AllowHostDirVolumePlugin: false,
+				AllowHostIPC:             false,
+				AllowHostPID:             false,
+				AllowPrivilegeEscalation: ptrBool(false),
+				AllowPrivilegedContainer: false,
+				ReadOnlyRootFilesystem:   false,
+				RequiredDropCapabilities: []corev1.Capability{},
+				RunAsUser:                securityv1.RunAsUserStrategyOptions{Type: securityv1.RunAsUserStrategyMustRunAsRange},
+				SELinuxContext:           securityv1.SELinuxContextStrategyOptions{Type: securityv1.SELinuxStrategyMustRunAs},
```
jparrill: Are any of these suitable for parameterization?

test/e2e/cilium_network_policy_test.go
```diff
@@ -0,0 +1,257 @@
...
+				Users:                    []string{"system:serviceaccount:cilium-test:default"},
```
jparrill: SA name should be part of the const.

test/e2e/cilium_network_policy_test.go
```diff
@@ -0,0 +1,257 @@
...
+					Labels: map[string]string{
+						"security.openshift.io/scc.podSecurityLabelSync": "false",
+						"pod-security.kubernetes.io/enforce":             "privileged",
+						"pod-security.kubernetes.io/audit":               "privileged",
+						"pod-security.kubernetes.io/warn":                "privileged",
+					},
+				},
+			}
+
+			err := guestClient.Create(ctx, ns)
```
jparrill: maybe it's worth to make `createIfNotExists` public and use it in the multiple parts of the code?

test/e2e/cilium_network_policy_test.go
```diff
@@ -0,0 +1,257 @@
...
+			}
+
+			t.Log("Test namespace created successfully")
+		})
+
+		t.Run("DeployCiliumConnectivityTest", func(t *testing.T) {
+			g := NewWithT(t)
+
+			t.Logf("Deploying Cilium connectivity test from version %s", e2eutil.CiliumVersion)
+			connectivityTestURL := fmt.Sprintf("https://raw.githubusercontent.com/cilium/cilium/%s/examples/kubernetes/connectivity-check/connectivity-check.yaml", e2eutil.CiliumVersion)
```
jparrill: The URL should be a const/var. Maybe the version a const and the var a mixture of the URL + version.

test/e2e/cilium_network_policy_test.go
```diff
@@ -0,0 +1,257 @@
...
+			g.Expect(err).ToNot(HaveOccurred(), "failed to apply connectivity test manifest")
+
+			t.Log("Connectivity test manifests applied successfully")
+		})
+
+		t.Run("WaitForConnectivityTestPodsReady", func(t *testing.T) {
+			g := NewWithT(t)
+
+			t.Log("Waiting for connectivity test pods to be ready")
+			g.Eventually(func() bool {
```
jparrill: `waitForDeploymentAvailable` maybe?

test/e2e/cilium_network_policy_test.go
```diff
@@ -0,0 +1,257 @@
...
+
+		t.Run("WaitForConnectivityTestCompletion", func(t *testing.T) {
+			g := NewWithT(t)
+
+			t.Logf("Waiting %v for connectivity tests to run", e2eutil.CiliumConnectivityTestDuration)
+			time.Sleep(e2eutil.CiliumConnectivityTestDuration)
+
+			t.Log("Verifying all test pods are still running")
+			podList := &corev1.PodList{}
+			err := guestClient.List(ctx, podList, crclient.InNamespace(testNamespace))
```
jparrill: Same as above.

test/e2e/util/util.go
```diff
@@ -77,6 +81,30 @@ import (
...
+	CiliumDaemonSetReadyTimeout       = 20 * time.Minute
+	CiliumDaemonSetReadyPollInterval  = 15 * time.Second
+
+	// Cilium connectivity test timeouts
+	CiliumPodsReadyTimeout                 = 10 * time.Minute
+	CiliumPodsReadyPollInterval            = 10 * time.Second
+	CiliumConnectivityTestPodsTimeout      = 10 * time.Minute
+	CiliumConnectivityTestPodsPollInterval = 15 * time.Second
+	CiliumConnectivityTestDuration         = 60 * time.Second
+)
```
jparrill: Could you simplify these timeouts?

test/e2e/util/util.go
```diff
@@ -1167,6 +1195,434 @@ func EnsureNetworkPolicies(t *testing.T, ctx context.Context, c crclient.Client,
...
+				// Namespace doesn't exist, create it with labels
+				t.Log("Creating cilium namespace")
+				ciliumNs = &corev1.Namespace{}
+				ciliumNs.Name = "cilium"
+				ciliumNs.Labels = map[string]string{
+					"security.openshift.io/scc.podSecurityLabelSync": "false",
+					"pod-security.kubernetes.io/enforce":             "privileged",
+					"pod-security.kubernetes.io/audit":               "privileged",
+					"pod-security.kubernetes.io/warn":                "privileged",
+				}
```
jparrill: Create a `manifest.CilliumNamespace` (or something similar) function with all the details to be called from here.

test/e2e/util/util.go
```diff
@@ -1167,6 +1195,434 @@ func EnsureNetworkPolicies(t *testing.T, ctx context.Context, c crclient.Client,
...
+				g.Expect(err).ToNot(HaveOccurred(), "unexpected error checking cilium namespace")
+
+				t.Log("Updating existing cilium namespace with PodSecurity labels")
+				if ciliumNs.Labels == nil {
+					ciliumNs.Labels = make(map[string]string)
+				}
+				ciliumNs.Labels["security.openshift.io/scc.podSecurityLabelSync"] = "false"
+				ciliumNs.Labels["pod-security.kubernetes.io/enforce"] = "privileged"
+				ciliumNs.Labels["pod-security.kubernetes.io/audit"] = "privileged"
+				ciliumNs.Labels["pod-security.kubernetes.io/warn"] = "privileged"
```
jparrill: Just grab the manifest function once created and apply the content if that's not equal.

test/e2e/util/util.go
```diff
@@ -1167,6 +1195,434 @@ func EnsureNetworkPolicies(t *testing.T, ctx context.Context, c crclient.Client,
...
+					securityv1.FSTypeEmptyDir,
+					securityv1.FSTypeSecret,
+					securityv1.FSTypeConfigMap,
+					securityv1.FSProjected,
+				},
+				Users: []string{
+					"system:serviceaccount:cilium:cilium",
+					"system:serviceaccount:cilium:cilium-operator",
+				},
+			}
```
jparrill: I would isolate this into a manifest function.

test/e2e/util/util.go
```diff
@@ -1167,6 +1195,434 @@ func EnsureNetworkPolicies(t *testing.T, ctx context.Context, c crclient.Client,
...
+
+			t.Log("Cilium SecurityContextConstraints created successfully")
+		})
+
+		t.Run("InstallCilium", func(t *testing.T) {
+			g := NewWithT(t)
+			t.Logf("Installing Cilium operator version %s", CiliumVersion)
+			// Install Cilium operator manifests
+			manifestURLs := []string{
```
jparrill: Isolate into a function and call it from here. Also, you can do the same as suggested above, use the version as const and the urls as vars concatenated with the version at their isolated manifest function

test/e2e/util/util.go
```diff
@@ -1167,6 +1195,434 @@ func EnsureNetworkPolicies(t *testing.T, ctx context.Context, c crclient.Client,
...
+			// Wait for cilium-olm deployment to be ready
+			t.Log("Waiting for cilium-olm deployment to be ready")
+			g.Eventually(func() bool {
+				deployment := &appsv1.Deployment{}
+				err := guestClient.Get(ctx, crclient.ObjectKey{Name: "cilium-olm", Namespace: "cilium"}, deployment)
+				if err != nil {
+					t.Logf("Failed to get cilium-olm deployment: %v", err)
+					return false
+				}
+				return deployment.Status.ReadyReplicas > 0
```
jparrill: `waitForDeploymentAvailable`?

test/e2e/util/util.go
```diff
@@ -1167,6 +1195,434 @@ func EnsureNetworkPolicies(t *testing.T, ctx context.Context, c crclient.Client,
...
+					return false
+				}
+				if maskSize != int64(hostPrefix) {
+					t.Logf("CiliumConfig clusterPoolIPv4MaskSize is %d, expected %d. Updating...", maskSize, hostPrefix)
+					// Update it again if the operator overwrote it
+					desiredConfig := createCiliumConfig(podCIDR, hostPrefix)
+					ciliumConfig.Object["spec"] = desiredConfig.Object["spec"]
+					err = guestClient.Update(ctx, ciliumConfig)
+					if err != nil {
+						t.Logf("Failed to update CiliumConfig: %v", err)
```
jparrill: If the update fails it will continue, I think this is not desired maybe?.

test/e2e/util/util.go
```diff
@@ -1167,6 +1195,434 @@ func EnsureNetworkPolicies(t *testing.T, ctx context.Context, c crclient.Client,
...
+					return false
+				}
+
+				t.Logf("CiliumConfig has correct clusterPoolIPv4MaskSize: %d", maskSize)
+				return true
+			}, CiliumConfigVerifyTimeout, CiliumConfigVerifyPollInterval).Should(BeTrue(), "CiliumConfig should have correct configuration")
+
+			// Wait for operator to create DaemonSet
+			t.Log("Waiting for Cilium DaemonSet to be created by operator")
+			var ciliumDaemonSet *appsv1.DaemonSet
```
jparrill: You can use here: `waitForDaemonSetsReady`

test/e2e/util/util.go
```diff
@@ -1167,6 +1195,434 @@ func EnsureNetworkPolicies(t *testing.T, ctx context.Context, c crclient.Client,
...
+					}
+				}
+
+				t.Log("Cilium DaemonSet not created by operator yet")
+				return false
+			}, CiliumDaemonSetCreateTimeout, CiliumDaemonSetCreatePollInterval).Should(BeTrue(), "cilium-olm operator should create Cilium DaemonSet")
+
+			// Now wait for DaemonSet pods to be ready
+			t.Log("Waiting for Cilium agent pods from DaemonSet to be ready")
+			g.Eventually(func() bool {
```
jparrill: Build a map with the desired DS to be verified. Something like:
```
    daemonSetsToCheck := []DaemonSetManifest{
		{GetFunc: OpenshiftOVNKubeDaemonSet, AllowPartialNodes: false},
		{GetFunc: hccomanifests.GlobalPullSecretDaemonSet, AllowPartialNodes: false},
		{GetFunc: hccomanifests.KonnectivityAgentDaemonSet, AllowPartialNodes: false},
		{GetFunc: KubeletConfigVerifierDaemonSet, AllowPartialNodes: true},
	}

	err = waitForDaemonSetsReady(t, ctx, guestClient, daemonSetsToCheck, availableNodesCount)
	g.Expect(err).NotTo(HaveOccurred(), "failed to wait for DaemonSets to be ready")
```

test/e2e/util/util.go
```diff
@@ -1167,6 +1195,434 @@ func EnsureNetworkPolicies(t *testing.T, ctx context.Context, c crclient.Client,
...
+
+	yamlContent, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return fmt.Errorf("failed to read manifest content: %w", err)
+	}
+
+	// Split multi-document YAML using yaml decoder
+	decoder := k8syaml.NewYAMLOrJSONDecoder(bytes.NewReader(yamlContent), 4096)
+
+	for {
```
jparrill, line 1225: Could we change this infinite loop by any other loop with finite resources?

test/e2e/util/util.go
```diff
@@ -1167,6 +1195,434 @@ func EnsureNetworkPolicies(t *testing.T, ctx context.Context, c crclient.Client,
...
+
+		// Set default namespace if provided and the resource doesn't have one
+		if len(defaultNamespace) > 0 && obj.GetNamespace() == "" {
+			// Check if this is a namespaced resource (not cluster-scoped like ClusterRole)
+			kind := obj.GetKind()
+			isClusterScoped := kind == "ClusterRole" || kind == "ClusterRoleBinding" ||
+				kind == "CustomResourceDefinition" || kind == "Namespace"
+
+			if !isClusterScoped {
+				obj.SetNamespace(defaultNamespace[0])
```
jparrill, line 1255: What's the default namespace? It shouldn't be a const?

test/e2e/util/util.go
```diff
@@ -1167,6 +1195,434 @@ func EnsureNetworkPolicies(t *testing.T, ctx context.Context, c crclient.Client,
...
+	}
+
+	return nil
+}
+
+// getCiliumNetworkConfig extracts pod CIDR and host prefix from guest cluster
+// In dual-stack environments, this function ensures we return the IPv4 network
+func getCiliumNetworkConfig(ctx context.Context, guestClient crclient.Client) (podCIDR string, hostPrefix int32) {
+	podCIDR = "10.132.0.0/14"
+	hostPrefix = 23
```
jparrill: Const?

test/e2e/util/util.go
```diff
@@ -1167,6 +1195,434 @@ func EnsureNetworkPolicies(t *testing.T, ctx context.Context, c crclient.Client,
...
+
+// createCiliumConfig creates a CiliumConfig custom resource
+func createCiliumConfig(podCIDR string, hostPrefix int32) *unstructured.Unstructured {
+	ciliumConfig := &unstructured.Unstructured{
+		Object: map[string]interface{}{
+			"apiVersion": "cilium.io/v1alpha1",
+			"kind":       "CiliumConfig",
+			"metadata": map[string]interface{}{
+				"name":      "cilium",
+				"namespace": "cilium",
```
jparrill: const

test/e2e/util/util.go
```diff
@@ -1167,6 +1195,434 @@ func EnsureNetworkPolicies(t *testing.T, ctx context.Context, c crclient.Client,
...
+						"enabled": false,
+					},
+				},
+				"sessionAffinity": true,
+				"tolerations": []map[string]interface{}{
+					{
+						"operator": "Exists",
+					},
+				},
+			},
```
jparrill: Not sure how customizable are these fields.

jparrill (commented):
Dropped some comments. Thanks

test/e2e/util/util.go
wewang58: It's according to the script: https://github.com/openshift/release/blob/master/ci-operator/step-registry/cucushift/hypershift-extended/cilium/cucushift-hypershift-extended-cilium-commands.sh

test/e2e/util/util.go
wewang58: If install cilium, default namespace is cilium, if install connectivity-check, default namespace is cilium-test, so cannot be const.

test/e2e/cilium_network_policy_test.go
wewang58: @mgencur  and i discussed  the issue before: https://github.com/openshift/hypershift/pull/7077#discussion_r2514597244

test/e2e/cilium_network_policy_test.go
wewang58: We need confirm all pods( cilium-olm, cilium-operator and cilium) pods are all running, then cni can work correctly.

wewang58:
@jparrill  Submitted a new commit for [Refactor codes](https://github.com/openshift/hypershift/pull/7077/commits/f6d098f9a68a49e337565aa723fd1fe0e62fe55e) according to comments, test : TestCiliumConnectivity passed in [e2e-aks job](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_release/71182/rehearse-71182-periodic-ci-openshift-hypershift-release-4.21-periodics-e2e-aks/1993502032803139584), can you review again, thanks.

test/e2e/util/util.go
wewang58: yes

wewang58:
> @jparrill Submitted a new commit for [Refactor codes](https://github.com/openshift/hypershift/pull/7077/commits/f6d098f9a68a49e337565aa723fd1fe0e62fe55e) according to comments, test : TestCiliumConnectivity passed in [e2e-aks job](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_release/71182/rehearse-71182-periodic-ci-openshift-hypershift-release-4.21-periodics-e2e-aks/1993502032803139584), can you review again, thanks.

@jparrill  Can you review my new update, when you have time? i am plan to test the pr: https://github.com/openshift/release/pull/71383 when the pr is ready, thanks.

test/e2e/util/cilium.go
```diff
@@ -0,0 +1,310 @@
...
+}
+
+// EnsureCiliumConnectivityTestResources performs the Cilium connectivity tests.
+// It returns a cleanup function.
+func EnsureCiliumConnectivityTestResources(t *testing.T, ctx context.Context, guestClient crclient.Client) func() {
+	t.Run("CheckCiliumPodsRunning", func(t *testing.T) {
+		g := NewWithT(t)
+
+		t.Log("Check cilium pods to be running")
+		g.Eventually(func() bool {
```
LiangquanLi930: for
```
g.Eventually(func()
```
We can use
```
g.Eventually(func(g Gomega) {
```
This makes the code cleaner and better aligned with the current project’s style.

wewang58: Updated, thanks

test/e2e/util/cilium.go
```diff
@@ -0,0 +1,294 @@
...
+				for _, condition := range pod.Status.Conditions {
+					if condition.Type == corev1.PodReady && condition.Status == corev1.ConditionTrue {
+						ready = true
+						break
+					}
+				}
+				if !ready {
+					notReadyPods = append(notReadyPods, fmt.Sprintf("%s (phase: %s)", pod.Name, pod.Status.Phase))
+				}
+			}
```
LiangquanLi930: For loop-based checks, we can use HaveEach，E.g
```
  g.Expect(podList.Items).To(HaveEach(
      HaveField("Status.Conditions", ContainElement(
          And(
              HaveField("Type", corev1.PodReady),
              HaveField("Status", corev1.ConditionTrue),
          ),
      )),
  ))
```

test/e2e/util/cilium.go
```diff
@@ -0,0 +1,294 @@
...
+					if condition.Type == corev1.PodReady && condition.Status == corev1.ConditionTrue {
+						ready = true
+						break
+					}
+				}
+				if !ready {
+					notReadyPods = append(notReadyPods, fmt.Sprintf("Pod %s is not ready yet, phase: %s", pod.Name, pod.Status.Phase))
+				}
+			}
+			g.Expect(notReadyPods).To(BeEmpty(), "some pods are not ready")
```
LiangquanLi930: same

test/e2e/util/cilium.go
wewang58: have learned more.

test/e2e/util/cilium.go
```diff
@@ -0,0 +1,282 @@
...
+				"pod-security.kubernetes.io/enforce":             "privileged",
+				"pod-security.kubernetes.io/audit":               "privileged",
+				"pod-security.kubernetes.io/warn":                "privileged",
+			},
+		},
+	}
+}
+
+// CiliumSCCManifest returns the SecurityContextConstraints for Cilium.
+func CiliumSCCManifest() *securityv1.SecurityContextConstraints {
```
jparrill, line 80: This needs to be public?

test/e2e/util/cilium.go
```diff
@@ -0,0 +1,282 @@
...
+	crclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	// CiliumNamespace is the namespace where Cilium agent pods run.
+	CiliumNamespace = "cilium"
+	// CiliumTestNamespace is the namespace created for Cilium connectivity tests.
+	CiliumTestNamespace = "cilium-test"
+	// CiliumTestServiceAccount is the name of the service account used for Cilium connectivity tests.
+	CiliumTestServiceAccount = "default"
```
jparrill, line 53: This needs to be public?

test/e2e/util/cilium.go
```diff
@@ -0,0 +1,282 @@
...
+)
+
+// CiliumNamespaceManifest returns the cilium namespace with the required PodSecurity labels.
+func CiliumNamespaceManifest() *corev1.Namespace {
```
jparrill, line 65: This needs to be public?

test/e2e/util/cilium.go
```diff
@@ -0,0 +1,282 @@
...
+		},
+		Users: []string{
+			"system:serviceaccount:cilium:cilium",
+			"system:serviceaccount:cilium:cilium-operator",
+		},
+	}
+}
+
+// CiliumManifestURLs returns the list of Cilium manifest URLs for a given version.
+func CiliumManifestURLs(version string) []string {
```
jparrill, line 116: This needs to be public?

test/e2e/util/util.go
```diff
@@ -77,6 +80,21 @@ import (
 	"go.uber.org/zap/zaptest"
 )
 
+var (
+	// CiliumVersion is read from CILIUM_VERSION at runtime. When empty, the Cilium tests should skip.
+	CiliumVersion = os.Getenv("CILIUM_VERSION")
```
jparrill: - This needs to be public?
- Could we move the cilium related vars to the cilium.go file?

test/e2e/util/util.go
```diff
@@ -77,6 +80,21 @@ import (
...
+const (
+	// Generic timeouts and intervals for Cilium tests
+	CiliumDefaultTimeout           = 10 * time.Minute
+	CiliumLongTimeout              = 20 * time.Minute
+	CiliumShortTimeout             = 2 * time.Minute
+	CiliumDefaultPollInterval      = 10 * time.Second
+	CiliumLongPollInterval         = 15 * time.Second
+	CiliumConnectivityWaitDuration = 60 * time.Second
+)
+
```
jparrill: - This needs to be public?
- Could we move the cilium related vars to the cilium.go file?

test/e2e/util/util.go
jparrill: PTAL

jparrill (commented):
More comments. Thanks

test/e2e/util/cilium.go
wewang58: I think no need.

test/e2e/util/cilium.go
wewang58: I think no need.

test/e2e/util/cilium.go
wewang58: I think no need.

test/e2e/util/cilium.go
wewang58: I think no need.

test/e2e/util/util.go
wewang58: Yes, moved the vars to cilium.go

test/e2e/util/util.go
wewang58: Updated.

wewang58:
@jparrill can you help to review my pr when you have time? thanks.

PR #7094, opened 2025-10-23
OCPBUGS-63370: fix: Make the hypershift CLI binary FIPS-compliant

rokej:

- Default FIPS compliance: Main binaries are now FIPS-compliant by default
- Backward compatibility: Non-FIPS option available via hypershift-no-cgo for any automation scripts that rely on non-FIPS binary.

Fixes https://issues.redhat.com//browse/OCPBUGS-63370

## Checklist:


jparrill:
/retitle OCPBUGS-63370: fix: Make the hypershift CLI binary FIPS-compliant

PR #7138, opened 2025-10-31
OCPBUGS-64609:fix:(hcp-aws-create): remove the default imagestram 

heliubj18:
Issue:
When using the hcp client, if `--release-image` is not specified, it always uses the latest image (for example, 4.20.2-multi), even in a 4.19 HO environment.

Fix:
Removed the default imagestream set in hcp aws create to ensure that when `spec.release.image` is unset, the system uses the latest supported stable release image instead of the absolute latest.

Note:  
- When neither an imagestream nor `--release-image` is provided, the hostedcluster `spec.release.image` field in the rendered create file is left empty.
- The webhook subsequently injects the default latest supported stable release image into the HostedCluster resource after `oc apply`

### DIFF
--- a/product-cli/cmd/cluster/aws/create.go
+++ b/product-cli/cmd/cluster/aws/create.go
@@ -5,7 +5,6 @@ import (
 
 	hypershiftaws "github.com/openshift/hypershift/cmd/cluster/aws"
 	"github.com/openshift/hypershift/cmd/cluster/core"
-	"github.com/openshift/hypershift/support/config"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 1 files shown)


heliubj18:
@bryan-cox could you please help review it ? thanks!

bryan-cox:
Hey @heliubj18 - I think we might be working on two similar tickets. Wouldn't this bug be gone once https://github.com/openshift/hypershift/pull/7228 is in?

heliubj18:
@bryan-cox yes, I will close this one. thanks

PR #7139, opened 2025-10-31
OCPBUGS-34078: Prevent duplicate AWS resource tag keys in HostedCluster and NodePool

bryan-cox:

Prevents duplicate AWS resource tag keys in HostedCluster and NodePool resources. Previously, users could specify the same tag key multiple times without receiving an error, leading to undefined behavior.

Fixes OCPBUGS-34078

## How the fix works:

**Three levels of validation:**

1. **CRD Schema Validation** (`api/hypershift/v1beta1/aws.go`): Uses `+listType=map` with `+listMapKey=key` on both:
   - `HostedCluster.Spec.Platform.AWS.ResourceTags`
   - `NodePool.Spec.Platform.AWS.ResourceTags`

   This ensures users receive clear "Duplicate value" errors when applying YAML with duplicate tag keys.

2. **CLI Validation** (`cmd/util/client.go`): The `ParseAWSTags` function now returns an error when duplicate keys are detected: `duplicate tag key: "foo"`

3. **Idempotent Tag Addition** (`cmd/util/client.go`): Added `AddUniqueTag()` helper to safely add tags without creating duplicates. This fixes e2e test failures where `Complete()` was called multiple times on the same options object.

## Testing:

- Unit tests added for `ParseAWSTags` duplicate key detection
- Unit tests added for `AddUniqueTag` helper function
- E2E tests added for both HostedCluster and NodePool validation

## Checklist:

### DIFF
--- a/api/hypershift/v1beta1/aws.go
+++ b/api/hypershift/v1beta1/aws.go
@@ -61,6 +61,8 @@ type AWSNodePoolPlatform struct {
 	// for the user.
 	//
 	// +kubebuilder:validation:MaxItems=25
+	// +listType=map
+	// +listMapKey=key
 	// +optional
 	ResourceTags []AWSResourceTag `json:"resourceTags,omitempty"`
 
@@ -314,6 +316,8 @@ type AWSPlatformSpec struct {
 	// These take precedence over tags defined out of band (i.e., tags added manually or by other tools outside of HyperShift) in AWS in case of conflicts.
 	//
 	// +kubebuilder:validation:MaxItems=25
+	// +listType=map
+	// +listMapKey=key
 	// +optional
 	ResourceTags []AWSResourceTag `json:"resourceTags,omitempty"`
 
--- a/api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/hostedclusters.hypershift.openshift.io/AAA_ungated.yaml
+++ b/api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/hostedclusters.hypershift.openshift.io/AAA_ungated.yaml
@@ -3800,6 +3800,9 @@ spec:
                           type: object
                         maxItems: 25
                         type: array
+                        x-kubernetes-list-map-keys:
+                        - key
+                        x-kubernetes-list-type: map
                       rolesRef:
                         description: |-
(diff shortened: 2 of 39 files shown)


api/hypershift/v1beta1/aws.go
```diff
@@ -61,6 +61,9 @@ type AWSNodePoolPlatform struct {
 	// for the user.
 	//
 	// +kubebuilder:validation:MaxItems=25
+	// +kubebuilder:validation:XValidation:rule="self.all(t, size(t.key) > 0)", message="spec.platform.aws.resourceTags key cannot be empty"
+	// +kubebuilder:validation:XValidation:rule="self.all(t, size(t.value) > 0)", message="spec.platform.aws.resourceTags value cannot be empty"
```
muraee: afaik AWS allows empty tag values, why are we adding this validation?

bryan-cox: That is true but this appears to be a limitation we are under - [OCPBUGS-32932](https://issues.redhat.com/browse/OCPBUGS-32932). 

As you pointed out as well, there's existing validation here - seems the key, value fields on the AWSResrouceTag struct have validations already https://github.com/openshift/hypershift/blob/main/api/hypershift/v1beta1/aws.go#L394-L409 so this seems to be an old invalid bug.

bryan-cox:
The AWS tests were failing before this last push because 
```
     Root Cause: The Complete() method in cmd/cluster/aws/create.go appends
     red-hat-managed=true to AdditionalTags every time it's called. In the e2e test
     framework, core.CreateCluster() is called twice with the same platform options
     object:

     1. test/e2e/util/fixture.go:195 - Render mode (to generate manifests)
     2. test/e2e/util/fixture.go:213 - Create mode (to apply manifests)

     Each call triggers Complete(), resulting in duplicate red-hat-managed=true entries.
```

bryan-cox:
/override "Red Hat Konflux / enterprise-contract-mce-211 / hypershift-cli-mce-211"

bryan-cox:

bryan-cox:

enxebre:
supersed by https://github.com/openshift/hypershift/pull/8937

PR #7140, opened 2025-10-31
OCPBUGS-32932: Validate AWS resource tags to reject empty values

bryan-cox:

This PR adds validation to the NodePool webhook to prevent empty tag keys or values in AWS resource tags.

Previously, empty tag values could be set on NodePools, but they would fail later during manifest application to the guest cluster with a CRD validation error. This delayed failure provided a poor user experience.

With this change:
- Users receive immediate feedback when attempting to create or update NodePools with invalid tag values
- The validation is consistent with the existing CRD validation (minLength: 1)
- Better error messages guide users to fix the problem

Fixes OCPBUGS-32932

- The validation logic is straightforward - it checks that both tag keys and values are non-empty
- Added comprehensive unit tests covering:
  - Empty tag values (rejected)
  - Empty tag keys (rejected)
  - Valid tags (accepted)
  - No tags (accepted)
  - Mixed valid/invalid tags (rejected)
- Tests follow the project's "When...it should..." naming convention

## Checklist:
- [ ] This change includes docs. (Not needed - internal validation logic)

---

🤖 Generated with [Claude Code](https://claude.com/claude-code) via `/jira:solve OCPBUGS-32932 origin`

### DIFF
--- a/hypershift-operator/controllers/hostedcluster/hostedcluster_webhook.go
+++ b/hypershift-operator/controllers/hostedcluster/hostedcluster_webhook.go
@@ -227,6 +227,8 @@ func (v nodePoolValidator) ValidateCreate(ctx context.Context, obj runtime.Objec
 	switch np.Spec.Platform.Type {
 	case hyperv1.KubevirtPlatform:
 		return v.validateCreateKubevirtNodePool(ctx, np)
+	case hyperv1.AWSPlatform:
+		return v.validateCreateAWSNodePool(ctx, np)
 	default:
 		return nil, nil // no validation needed
 	}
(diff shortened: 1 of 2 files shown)


hypershift-operator/controllers/hostedcluster/hostedcluster_webhook.go
```diff
@@ -275,6 +282,39 @@ func (v nodePoolValidator) validateUpdateKubevirtNodePool(ctx context.Context, o
 	return nil
 }
 
+func (v nodePoolValidator) validateCreateAWSNodePool(ctx context.Context, np *hyperv1.NodePool) (admission.Warnings, error) {
+	if np.Spec.Platform.AWS != nil {
```
clebs, line 286: Is `np.Spec.Platform.AWS == nil` a valid value for admission?

bryan-cox: No, I don't think so when the platform type is AWS on L254. cc @csrwng

bryan-cox:
Closing in favor of #7139 which combines both empty value validation (OCPBUGS-32932) and duplicate key validation (OCPBUGS-34078) into a single comprehensive solution. The combined approach provides better maintainability with all AWS resource tag validations in one place.

PR #7141, opened 2025-10-31
OCPBUGS-20470: Validate OAuth accessTokenInactivityTimeout minimum value

bryan-cox:

This PR fixes the validation for `spec.configuration.oauth.tokenConfig.accessTokenInactivityTimeout` in HostedClusters to properly reject values less than 300 seconds, making it consistent with OpenShift management cluster behavior.

**Approach:** This fix uses CEL (Common Expression Language) validation directly in the CRD, which is enforced at the API server level. The validation is placed on the `ClusterConfiguration` struct rather than the individual `OAuth` field, allowing controller-gen to properly generate the validation in CRDs.

Without this validation, users can configure invalid timeout values that are accepted but don't work as expected. Management clusters reject values < 300s with the error:
```
spec.tokenConfig.accessTokenInactivityTimeout: Invalid value: v1.Duration{Duration:100000000000}: the minimum acceptable token timeout value is 300 seconds
```

## Changes Made:

**File: `api/hypershift/v1beta1/hostedcluster_types.go`**
- Added CEL validation to the `ClusterConfiguration` struct (line 1865)
- Removed non-functional validation from the `OAuth` field (was at line 1909)
- The validation rule: `!has(self.oauth) || !has(self.oauth.tokenConfig) || !has(self.oauth.tokenConfig.accessTokenInactivityTimeout) || duration(self.oauth.tokenConfig.accessTokenInactivityTimeout).getSeconds() >= 300`

**Generated files:**
- All CRD manifests now include the validation at the `configuration` object level
- Validation appears in `hostedclusters-Default.crd.yaml` at lines 2753-2758

Fixes [OCPBUGS-20470](https://issues.redhat.com/browse/OCPBUGS-20470)

### Why CEL validation instead of webhook validation?

Initially, it appeared that CEL validations couldn't work for this case because `OAuth` references an external type (`configv1.OAuthSpec`). However, **CEL validations DO work when placed on the parent struct** rather than the external type field itself.

**Advantages of this CEL approach:**
- ✅ Simpler - no additional validation code needed
- ✅ More robust - validation enforced directly by API server
- ✅ No webhook dependency - works even if webhook is unavailable
- ✅ Declarative - validation is part of the API schema

The CEL validation is now properly embedded in all generated CRDs and will be enforced by Kubernetes API server admission.

### Verification

You can verify the validation exists in the generated CRD:
```bash
grep -A 5 "accessTokenInactivityTimeout minimum acceptable" \
  cmd/install/assets/hypershift-operator/zz_generated.crd-manifests/hostedclusters-Default.crd.yaml
```

## Test Plan:

- [x] `make lint-fix` passes
- [x] `make verify` passes
- [x] Unit tests pass
- [ ] Manual testing: Deploy HyperShift with updated CRDs and verify that creating a HostedCluster with `accessTokenInactivityTimeout: 100s` is rejected with the validation error

## Checklist:
- [ ] This change includes docs. (Not needed - validation error message is self-explanatory)
- [x] This change includes validation in CRDs (auto-generated from code)

🤖 Generated with [Claude Code](https://claude.com/claude-code) via `/jira:solve OCPBUGS-20470`

### DIFF
--- a/api/hypershift/v1beta1/hostedcluster_types.go
+++ b/api/hypershift/v1beta1/hostedcluster_types.go
@@ -1862,6 +1862,7 @@ type ConfigurationStatus struct {
 //
 // The API for individual configuration items is at:
 // https://docs.openshift.com/container-platform/4.7/rest_api/config_apis/config-apis-index.html
+// +kubebuilder:validation:XValidation:rule="!has(self.oauth) || !has(self.oauth.tokenConfig) || !has(self.oauth.tokenConfig.accessTokenInactivityTimeout) || duration(self.oauth.tokenConfig.accessTokenInactivityTimeout).getSeconds() >= 300", message="spec.configuration.oauth.tokenConfig.accessTokenInactivityTimeout minimum acceptable token timeout value is 300 seconds"
 type ClusterConfiguration struct {
 	// apiServer holds configuration (like serving certificates, client CA and CORS domains)
 	// shared by all API servers in the system, among them especially kube-apiserver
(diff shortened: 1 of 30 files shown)


bryan-cox:
Closing this PR as OCPBUGS-20470 is a duplicate bug ticket. The OAuth accessTokenInactivityTimeout validation has been working correctly since commit a3e0d2081 (November 2, 2023), where the CEL validation rule was fixed to properly check for null tokenConfig before accessing accessTokenInactivityTimeout.

The validation exists on the OAuth field at api/hypershift/v1beta1/hostedcluster_types.go:1909 and is correctly generated into all CRD files. No changes are needed.

PR #7143, opened 2025-10-31
OCPBUGS-34027: Fix cleanup of orphaned OIDC resources in guest cluster

bryan-cox:

This PR fixes an issue where ConfigMaps and Secrets created for external authentication (OIDC providers) in the guest cluster's `openshift-config` namespace were not being deleted when the OIDC configuration was removed from the HostedControlPlane spec.

**UPDATE after testing:** The initial implementation caused a race condition where resources were deleted before the guest cluster's Authentication status was updated, causing components to report errors. This has been fixed with an additional safety check.

The fix:
1. Tracks expected ConfigMaps and Secrets based on current HCP configuration
2. Adds managed labels to OIDC resources for identification
3. **Waits for Authentication status to clear before deleting resources** (prevents race condition)
4. Deletes orphaned resources when safe to do so

Cleanup handles these scenarios:
- Complete removal of OIDC provider configuration
- Removal of CA certificate reference from the provider
- Removal of individual OIDCClients from a provider

Fixes OCPBUGS-34027

### Race Condition Fix (Critical)

The initial implementation deleted resources immediately when OIDC config was removed, but the Authentication operator hadn't finished updating the guest cluster's Authentication status yet. This caused:
- Console/CLI components still expecting OIDC resources
- Errors: `secret 'authid-console-openshift-console' not found`
- Errors: `configmap 'keycloak-oidc-ca' not found`

**Solution:** Added safety check that prevents deletion until `Authentication.status.oidcClients` is empty. Resources are cleaned up on the next reconciliation loop after the Authentication operator finishes processing the change.

### Other Considerations

1. **Backwards Compatibility**: Resources created before this change won't have the managed label but can still be cleaned up
2. **ARO HCP Day-2 Secrets**: Secrets with `hypershift.openshift.io/hosted-cluster-sourced: "true"` are preserved
3. **Test Coverage**: Comprehensive unit tests covering all scenarios including the race condition

## Checklist:
- [ ] This change includes docs. (No docs needed - internal implementation detail)

🤖 Generated with [Claude Code](https://claude.com/claude-code) via `/jira:solve OCPBUGS-34027 origin`

### DIFF
--- a/control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
+++ b/control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
@@ -96,7 +96,11 @@ const (
 	ConfigNamespace        = "openshift-config"
 	ConfigManagedNamespace = "openshift-config-managed"
 	CloudProviderCMName    = "cloud-provider-config"
-	awsCredentialsTemplate = `[default]
+	// OIDCProviderManagedLabel is used to identify ConfigMaps and Secrets in the guest cluster
+	// that are managed by the OIDC provider reconciliation logic. Resources with this label
+	// will be subject to cleanup when they are no longer referenced in the HCP configuration.
+	OIDCProviderManagedLabel = "hypershift.openshift.io/oidc-provider-managed"
+	awsCredentialsTemplate   = `[default]
 role_arn = %s
 web_identity_token_file = /var/run/secrets/openshift/serviceaccount/token
 sts_regional_endpoints = regional
@@ -1238,6 +1242,11 @@ func (r *reconciler) reconcileIngressController(ctx context.Context, hcp *hyperv
 
 func (r *reconciler) reconcileAuthOIDC(ctx context.Context, hcp *hyperv1.HostedControlPlane) error {
 	var errs []error
+
+	// Track expected resources that should exist based on current HCP configuration
+	expectedConfigMaps := sets.New[string]()
+	expectedSecrets := sets.New[string]()
+
 	if !util.HCPOAuthEnabled(hcp) &&
 		len(hcp.Spec.Configuration.Authentication.OIDCProviders) != 0 {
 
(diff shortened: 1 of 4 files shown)


.work/jira/solve/spec-OCPBUGS-34027.md
```diff
@@ -0,0 +1,162 @@
+# Implementation Specification: OCPBUGS-34027
```
sjenning: did you mean to commit this?

bryan-cox: Hmm I was struggling with that. Should we commit those?
I was thinking we might want to keep a record of how those /jira-solved PRs were created with Claude.

bryan-cox: Decided to drop that commit

xiuwang:
Failed to pre-merge test on this pr, see details in [bug comment](https://issues.redhat.com/browse/OCPBUGS-34027?focusedId=28403395&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-28403395)

bryan-cox:
## Update: Fixed Authentication Type Issue

The tester confirmed that secrets and configmaps were still not being cleaned up even with the race condition fix. The root cause was discovered:

**Problem:** When OIDC configuration is removed from the HCP, we were clearing `Authentication.Spec` to an empty struct, but the `Type` field was being set to its zero value (empty string) instead of explicitly being set to `IntegratedOAuth`. The authentication-operator requires an explicit type change to properly clear the `Authentication.Status.OIDCClients` field.

**Solution:** Now explicitly setting `Type: configv1.AuthenticationTypeIntegratedOAuth` when OIDC configuration is removed. This tells the authentication-operator to:
1. Switch back to integrated OAuth mode
2. Clear the OIDC status fields properly  
3. Allow our cleanup logic to proceed

**What Changed:**
```go
// Before:
authentication.Spec = configv1.AuthenticationSpec{}

// After:
authentication.Spec = configv1.AuthenticationSpec{
    Type: configv1.AuthenticationTypeIntegratedOAuth,
}
```

This ensures the cluster properly falls back to OAuth when OIDC is removed, which is the expected day-2 behavior.

bryan-cox:
Moved back to draft since this is having issues in pre-merge testing.

bryan-cox:
/close

Not actively maintaining this PR

PR #7151, opened 2025-11-03
OCPBUGS-63353: Fix ValidAWSIdentityProvider status when KAS is unavailable

muraee:

This PR fixes the `ValidAWSIdentityProvider` condition to show `Unknown` status instead of `False` when the Kube API Server (KAS) is not available.

The AWS identity provider validation requires minting tokens via KAS. When KAS is not up (during cluster provisioning or when API server pods are down), the validation cannot be performed. Previously, the condition would show as `False` with reason `InvalidIdentityProvider` and message `WebIdentityErr`, which was misleading since the actual validation never occurred.

### Changes:
1. **Check KAS availability first**: Before attempting AWS validation, check if the `KubeAPIServerAvailable` condition is `True`
2. **Set condition to Unknown when validation cannot be performed**: 
   - When KAS is not available
   - When EC2 client is `nil` (token minting failed)
3. **Only validate when KAS is available**: Perform actual AWS identity provider validation only when KAS is confirmed to be available

This ensures accurate status reporting and prevents misleading alerts for service providers.

Fixes OCPBUGS-63353

- The metric `hypershift_cluster_invalid_aws_creds` handling will be addressed in a separate PR
- Unit tests have been added to validate the condition logic
- The change only affects the condition status when KAS is not available - existing validation logic remains unchanged

## Checklist:
- [ ] This change includes docs. (No doc changes needed - behavior fix)

🤖 Generated with [Claude Code](https://claude.com/claude-code) via `/jira:solve OCPBUGS-63353`

### DIFF
--- a/control-plane-operator/controllers/healthcheck/aws_test.go
+++ b/control-plane-operator/controllers/healthcheck/aws_test.go
@@ -1,0 +1,110 @@
+package healthcheck
+
+import (
+	"testing"
+
+	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
+
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// TestAWSHealthCheckIdentityProviderConditionLogic tests the condition setting logic
+// when KAS is not available. This validates that the function correctly sets the
+// ValidAWSIdentityProvider condition to Unknown when it cannot perform validation.
+func TestAWSHealthCheckIdentityProviderConditionLogic(t *testing.T) {
+	testCases := []struct {
+		name            string
+		kasCondition    *metav1.Condition
+		expectedStatus  metav1.ConditionStatus
+		expectedReason  string
+		expectedMessage string
+	}{
+		{
+			name:            "KAS not available - condition missing",
+			kasCondition:    nil,
+			expectedStatus:  metav1.ConditionUnknown,
+			expectedReason:  hyperv1.StatusUnknownReason,
+			expectedMessage: "Cannot validate AWS identity provider while KubeAPIServer is not available",
+		},
+		{
+			name: "KAS not available - condition False",
+			kasCondition: &metav1.Condition{
+				Type:   string(hyperv1.KubeAPIServerAvailable),
+				Status: metav1.ConditionFalse,
+			},
+			expectedStatus:  metav1.ConditionUnknown,
+			expectedReason:  hyperv1.StatusUnknownReason,
+			expectedMessage: "Cannot validate AWS identity provider while KubeAPIServer is not available",
+		},
+		{
+			name: "KAS not available - condition Unknown",
+			kasCondition: &metav1.Condition{
+				Type:   string(hyperv1.KubeAPIServerAvailable),
+				Status: metav1.ConditionUnknown,
+			},
+			expectedStatus:  metav1.ConditionUnknown,
+			expectedReason:  hyperv1.StatusUnknownReason,
+			expectedMessage: "Cannot validate AWS identity provider while KubeAPIServer is not available",
+		},
+		{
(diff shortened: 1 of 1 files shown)


muraee:
@coderabbitai review

control-plane-operator/controllers/healthcheck/aws_test.go
```diff
@@ -0,0 +1,118 @@
...
+
+			// Verify the function returns without error when KAS is not available
+			if tc.shouldReturnEarly && err != nil {
+				t.Errorf("Expected no error when returning early, got: %v", err)
+			}
+
+			// Verify the ValidAWSIdentityProvider condition was set correctly
+			condition := meta.FindStatusCondition(hcp.Status.Conditions, string(hyperv1.ValidAWSIdentityProvider))
+			if condition == nil {
+				t.Fatal("ValidAWSIdentityProvider condition was not set")
```
sdminonne, line 89: t.Error(...)
continue

muraee: Thank you for the review! I've improved the test code based on CodeRabbit's suggestions (renamed variable and added message verification).

Regarding the `t.Fatal` vs `t.Error + continue`: I kept `t.Fatal` because:

1. We're inside a `t.Run()` sub-test, so `t.Fatal` only stops the current sub-test, not the entire test suite
2. All subsequent assertions (lines 101-115) depend on `condition` being non-nil - continuing would cause nil pointer panics
3. Using `continue` inside the `t.Run` function literal would be a compile error (can't continue outer loop from nested function)

If the concern is that `t.Fatal` is too aggressive, we could use `t.Error + return` instead, which would exit only the current sub-test. However, `t.Fatal` is semantically correct here since the test cannot proceed without a valid condition.

What do you think?

---

muraee:
@coderabbitai Thanks for the detailed review! I've addressed both nitpick suggestions:

1. ✅ Added `expectedMessage` to the "KAS available" test case to verify the EC2 client unavailable message
2. ✅ Renamed `expectedMessagePart` to `expectedMessage` for clarity and updated the error message to reflect exact matching

Changes pushed in commit 940d047. All tests pass.

---

control-plane-operator/controllers/healthcheck/aws_test.go
sdminonne: fair enough

control-plane-operator/controllers/healthcheck/aws.go
```diff
@@ -15,8 +15,33 @@ import (
...
+			ObservedGeneration: hcp.Generation,
+			Status:             metav1.ConditionUnknown,
+			Message:            "Cannot validate AWS identity provider while KubeAPIServer is not available",
+			Reason:             hyperv1.StatusUnknownReason,
+		}
+		meta.SetStatusCondition(&hcp.Status.Conditions, condition)
+		return nil
+	}
+
 	ec2Client, _ := hostedcontrolplane.GetEC2Client()
```
sdminonne, line 34: @muraee this is more a question and I know you did not modify this code. But is it normal we shallow eventual error here?

sdminonne (commented):
A simple question otherwise /lgtm

control-plane-operator/controllers/healthcheck/aws_test.go
sdminonne: unit test env changed a lot since I used the first time :)

control-plane-operator/controllers/healthcheck/aws.go
muraee: the 2nd returned value is not an error, it's aws session

muraee:
/override "e2e-aks"
this PR introduces changes to an AWS controller only

muraee:
/override "ci/prow/e2e-aks"
this PR introduces changes to an AWS controller only

PR #7160, opened 2025-11-05
CNTRLPLANE-1857: fix(azure): reduce external-dns Azure DNS API call frequency

bryan-cox:
## Summary

Upgrades ExternalDNS to v0.19.0 and configures it to reduce Azure DNS API throttling in e2e tests.

## Changes

1. **Upgrade ExternalDNS image**: `registry.k8s.io/external-dns/external-dns:v0.19.0`
   - Previous: Red Hat edo image based on v0.13.1
   - Reason: v0.19.0 includes the `--azure-maxretries-count` flag

2. **Add `--min-event-sync-interval=1m`**
   - Reduces event-based sync from 5s to 1m
   - Decreases Azure DNS API calls by ~12x (720/hour → 60/hour)

3. **Add `--azure-maxretries-count=10`**
   - Increases retry attempts from default 3 to 10
   - Improves resilience against transient Azure throttling

## Why

Azure AKS e2e tests were failing due to:
- HTTP 429 "Throttled" errors from Azure DNS API
- HTTP 409 "Conflict" errors from stale ETags after throttling
- DNS records not created, preventing nodes from becoming ready

These changes reduce API call frequency to stay below Azure rate limits and increase retry resilience.

### DIFF
--- a/cmd/install/assets/hypershift_operator.go
+++ b/cmd/install/assets/hypershift_operator.go
@@ -319,6 +319,7 @@ func (o ExternalDNSDeployment) Build() *appsv1.Deployment {
 		deployment.Spec.Template.Spec.Containers[0].Args = append(deployment.Spec.Template.Spec.Containers[0].Args,
 			"--azure-config-file=/etc/provider/credentials",
 			"--min-event-sync-interval=1m",
+			"--azure-maxretries-count=10",
 		)
 	}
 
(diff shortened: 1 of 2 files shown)


bryan-cox:
/close

 It looks like the problem was too many recordsets in our Azure DNS Zone.

PR #7193, opened 2025-11-07
Platform Provider Architecture Proposal (HOv2)

muraee:
This document first presents a detailed analysis of the current architectural challenges in HyperShift, then outlines a comprehensive re-architecture plan to separate platform-specific logic from core components. The goal is to enable easier maintenance of existing platforms and simpler addition of new platforms while eliminating cross-platform contamination risks.


bryan-cox:
This should be moved to o/e under HyperShift

PR #7195, opened 2025-11-07
OCPBUGS-64817: Add CPO hotfix arm64 builds

celebdor:

Adds arm64 builds for the konflux hotfix pipeline.

Fixes: OCPBUGS-64817

## Checklist:


celebdor:
Marking valid bug due to this being a release-4.19 issue only

PR #7207, opened 2025-11-10
GCP-196: Implement GCPPrivateServiceObserver controller

cristianoveiga:
This PR implements the GCPPrivateServiceObserver controller which watches private-router Service and creates GCPPrivateServiceConnect CR.

Commits:
- Add LoadBalancerIP field to GCPPrivateServiceConnect CRD
- Implement GCPPrivateServiceObserver controller with comprehensive tests

[GCP-196](https://issues.redhat.com//browse/GCP-196) 

## Checklist:

### DIFF
--- a/vendor/github.com/openshift/hypershift/api/hypershift/v1beta1/gcpprivateserviceconnect_types.go
+++ b/vendor/github.com/openshift/hypershift/api/hypershift/v1beta1/gcpprivateserviceconnect_types.go
@@ -42,7 +42,7 @@ type GCPPrivateServiceConnectSpec struct {
 	// Populated by the observer from service status
 	// This value must be a valid IPv4 or IPv6 address.
 	// +required
-	// +kubebuilder:validation:XValidation:rule="isIP(self)",message="loadBalancerIP must be a valid IPv4 or IPv6 address"
+	// +kubebuilder:validation:XValidation:rule="self.matches('^((\\\\d{1,3}\\\\.){3}\\\\d{1,3})$') || self.matches('^(([0-9a-fA-F]{1,4}:){7}[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,7}:|([0-9a-fA-F]{1,4}:){1,6}:[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,5}(:[0-9a-fA-F]{1,4}){1,2}|([0-9a-fA-F]{1,4}:){1,4}(:[0-9a-fA-F]{1,4}){1,3}|([0-9a-fA-F]{1,4}:){1,3}(:[0-9a-fA-F]{1,4}){1,4}|([0-9a-fA-F]{1,4}:){1,2}(:[0-9a-fA-F]{1,4}){1,5}|[0-9a-fA-F]{1,4}:((:[0-9a-fA-F]{1,4}){1,6})|:((:[0-9a-fA-F]{1,4}){1,7}|:))$')",message="loadBalancerIP must be a valid IPv4 or IPv6 address"
 	// +kubebuilder:validation:MaxLength=45
 	LoadBalancerIP string `json:"loadBalancerIP"`
 
@@ -101,7 +101,7 @@ type GCPPrivateServiceConnectStatus struct {
 	// endpointIP is the reserved IP address for the PSC endpoint
 	// This value must be a valid IPv4 or IPv6 address.
 	// +optional
-	// +kubebuilder:validation:XValidation:rule="self == '' || isIP(self)",message="endpointIP must be a valid IPv4 or IPv6 address"
+	// +kubebuilder:validation:XValidation:rule="self == '' || self.matches('^((\\\\d{1,3}\\\\.){3}\\\\d{1,3})$') || self.matches('^(([0-9a-fA-F]{1,4}:){7}[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,7}:|([0-9a-fA-F]{1,4}:){1,6}:[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,5}(:[0-9a-fA-F]{1,4}){1,2}|([0-9a-fA-F]{1,4}:){1,4}(:[0-9a-fA-F]{1,4}){1,3}|([0-9a-fA-F]{1,4}:){1,3}(:[0-9a-fA-F]{1,4}){1,4}|([0-9a-fA-F]{1,4}:){1,2}(:[0-9a-fA-F]{1,4}){1,5}|[0-9a-fA-F]{1,4}:((:[0-9a-fA-F]{1,4}){1,6})|:((:[0-9a-fA-F]{1,4}){1,7}|:))$')",message="endpointIP must be a valid IPv4 or IPv6 address"
 	// +kubebuilder:validation:MaxLength=45
 	EndpointIP string `json:"endpointIP,omitempty"`
 


control-plane-operator/controllers/gcpprivateserviceconnect/observer_test.go
```diff
@@ -0,0 +1,311 @@
...
+					Platform: hyperv1.PlatformSpec{
+						Type: hyperv1.GCPPlatform,
+						GCP: &hyperv1.GCPPlatformSpec{
+							Project: "",
+						},
+					},
+				},
+			},
+			expected: []string{""},
+		},
```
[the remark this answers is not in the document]
cristianoveiga: The function is returning an empty slice, as expected. The empty project is out of the scope for this unit test.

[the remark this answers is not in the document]
cristianoveiga: I will remove the invalid test case.

control-plane-operator/controllers/gcpprivateserviceconnect/observer_test.go
```diff
@@ -0,0 +1,311 @@
...
+				ip := tt.service.Status.LoadBalancer.Ingress[0].IP
+				hasValidIP = ip != ""
+				if tt.expectLoadBalancerIP != "" {
+					g.Expect(ip).To(Equal(tt.expectLoadBalancerIP))
+				}
+			}
+
+			g.Expect(hasValidIP).To(Equal(tt.expectHasValidIP))
+		})
+	}
```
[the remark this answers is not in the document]
cristianoveiga: addressed.

control-plane-operator/controllers/gcpprivateserviceconnect/observer_test.go
```diff
@@ -0,0 +1,311 @@
...
+			for _, ownerRef := range tt.ownerRefs {
+				if ownerRef.Kind == "HostedControlPlane" && ownerRef.APIVersion == hyperv1.GroupVersion.String() {
+					hcpName = ownerRef.Name
+					break
+				}
+			}
+
+			g.Expect(hcpName).To(Equal(tt.expectHCPName))
+			g.Expect(hcpName != "").To(Equal(tt.expectHasOwner))
+		})
```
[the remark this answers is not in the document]
cristianoveiga: addressed.

control-plane-operator/controllers/gcpprivateserviceconnect/observer.go
```diff
@@ -0,0 +1,209 @@
...
+// getConsumerAcceptList extracts the consumer accept list from HostedControlPlane
+func getConsumerAcceptList(hcp *hyperv1.HostedControlPlane) []string {
+	if hcp.Spec.Platform.GCP == nil {
+		return nil
+	}
+
+	// Use the GCP project as the consumer accept list entry
+	// This allows the service attachment to be accessed by the same project
+	return []string{hcp.Spec.Platform.GCP.Project}
+}
```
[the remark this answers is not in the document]
cristianoveiga: Not applicable, project is required.

control-plane-operator/controllers/gcpprivateserviceconnect/observer.go
```diff
@@ -0,0 +1,209 @@
...
+			Kind:       "HostedControlPlane",
+			Name:       hcp.Name,
+			UID:        hcp.UID,
+			Controller: pointer.Bool(true),
+		}}
+
+		// Set spec fields
+		gcpPSC.Spec.LoadBalancerIP = loadBalancerIP
+		gcpPSC.Spec.ConsumerAcceptList = consumerAcceptList
+		// ForwardingRuleName is left empty - populated by hypershift-operator reconciler
```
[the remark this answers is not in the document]
cristianoveiga: N/A

control-plane-operator/main.go
```diff
@@ -520,6 +521,23 @@ func NewStartCommand() *cobra.Command {
 			}
 		}
 
+		if hcp.Spec.Platform.Type == hyperv1.GCPPlatform && util.IsPrivateHCP(hcp) && mgmtClusterCaps.Has(capabilities.CapabilityRoute) {
+			controllerName := "GCPPrivateServiceObserver"
```
cblecker, line 525: Would we not want to name this controller "PrivateKubeAPIServerServiceObserver" like the AWS one? Only one of them will be invoked because the platform type is different. Will we need a second one for the ingress later?

control-plane-operator/controllers/gcpprivateserviceconnect/observer.go
```diff
@@ -0,0 +1,229 @@
...
+	gcpPSC := &hyperv1.GCPPrivateServiceConnect{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      r.ServiceName,
+			Namespace: r.HCPNamespace,
+		},
+	}
+
+	_, err := r.CreateOrUpdate(ctx, r.Client, gcpPSC, func() error {
+		// Set OwnerReference to HostedControlPlane for lifecycle management
+		gcpPSC.OwnerReferences = []metav1.OwnerReference{{
```
cblecker, line 130: Note: it doesn't look like the AWS controller does this. This seems more correct, but could it cause problems?

control-plane-operator/controllers/gcpprivateserviceconnect/observer.go
```diff
@@ -0,0 +1,229 @@
...
+
+	// Find HostedControlPlane from service OwnerReference
+	hcpName := extractHostedControlPlaneOwnerName(svc.OwnerReferences)
+	if hcpName == "" {
+		return ctrl.Result{}, fmt.Errorf("service does not have HostedControlPlane owner reference")
+	}
+
+	// Fetch HostedControlPlane for ConsumerAcceptList
+	hcp := &hyperv1.HostedControlPlane{}
+	if err := r.Get(ctx, types.NamespacedName{Namespace: r.HCPNamespace, Name: hcpName}, hcp); err != nil {
```
cblecker, line 99: AWS does a list, and defends against multiple HCPs in the same namespace. Is that something we want to do too? https://github.com/openshift/hypershift/blob/d6c61ae619d8844a851c5be68e69555f392577b8/control-plane-operator/controllers/awsprivatelink/awsprivatelink_controller.go#L140-L153

control-plane-operator/controllers/gcpprivateserviceconnect/observer.go
cristianoveiga: Yes, I also noticed that the AWS controller doesn't do that. A explicit link/relationship between the PSC CR and the HCP can help with lookups (and keep things clean during deletion).

> but could it cause problems ?

Are you concerned with anything in particular ? I can't think of any issues...

The ultimate protection against deletion will be provided by the finalizers, which I will add in the next PRs when we create the actual GCP resources.

control-plane-operator/controllers/gcpprivateserviceconnect/observer.go
cristianoveiga: I think the AWS implementation does that because they don't have the owner reference (that we have been discussing in the another thread).

In the GCP approach, we perform a direct lookup (which should perform better anyway).

I can't think in a scenario where we have multiple HCPs in the namespace - and this probably should be enforced at another level.

control-plane-operator/main.go
cristianoveiga: My understanding is that we only need one Observer controller and perhaps it's a good idea to keep its name generic. Here is why:

Looking at the AWS implementation, I see that there are two observer controllers:

- PrivateKubeAPIServerServiceObserver
- PrivateIngressServiceObserver

But only the second one is actually creating a awsendpointservice CR because (the first doesn't have a load balancer associated to it):
https://github.com/openshift/hypershift/blob/46c0dcde8b94f3599a206653d2f4cea2620c856f/control-plane-operator/controllers/awsprivatelink/awsprivatelink_controller.go#L160-L163

Hence, this CR seems to be used for both. The actual configuration of PrivateIngress and happens in the AWSEndpointServiceReconciler, which we will implement the GCP equivalent in the next card, GCP-197.

Thoughts?

control-plane-operator/controllers/gcpprivateserviceconnect/observer.go
cblecker: No specific concerns, but I wanted to ask the question if perhaps there is a reason that there isn't an owner reference (e.g. resources being cleaned up in the wrong order on cluster uninstallation).

control-plane-operator/controllers/gcpprivateserviceconnect/observer.go
cblecker: No disagreement, but if it isn't enforced at another level today, could this leave us with no enforcement?

control-plane-operator/main.go
cblecker: What if the customer wants public ingress but a private API server? Is that why there is two for AWS?

cristianoveiga: That should be possible, but it doesn't seem it's to be controlled by the `PrivateKubeAPIServerServiceObserver`. Instead, this is achieved by the combination of the `endpointAccess` and the `servicePublishingStrategy` in the HostedCluster spec.

I didn't find any `kube-apiserver-private` service in our ROSA Clusters (even in private clusters) - that's why I think the first observer might not be required in our GCP PSC scenario (the observer that watches the private-router service would be enough).

Please let me know if you want me to rename it to `PrivateIngressServiceObserver` anyway. Otherwise, we can revisit this  in the future (during the implementation of GCP 'fully' Private clusters).

control-plane-operator/controllers/gcpprivateserviceconnect/observer.go
cristianoveiga: I tested this and I confirmed that the deletion is going to work without any issues.

We will implement a deletion mechanism for the gcpprivateserviceconnect CRs that will be triggered by the hypershift operator, similar to the AWS - see logic here: 

https://github.com/openshift/hypershift/blob/e83d907270ba8545b371edfa7b69989da2903214/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go#L3105-L3144

https://github.com/openshift/hypershift/blob/e83d907270ba8545b371edfa7b69989da2903214/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go#L3235

The HO basically deletes the awsendpointservice and waits until the resources are cleaned up before deleting the HCPs.

With that, the Owner references will not play a part on the deletion process.

control-plane-operator/controllers/gcpprivateserviceconnect/observer.go
cristianoveiga: Agreed. We could end up without this protection. But on the other hand, should this be a responsibility of the PSC observer?

I am fine replicating the AWS logic here (as I have done for most of the other parts). I just want to make sure that we keep the logic clean and avoid unnecessary checks.

api/hypershift/v1beta1/gcpprivateserviceconnect_types.go
```diff
@@ -38,12 +38,20 @@ const (
...
 // GCPPrivateServiceConnectSpec defines the desired state of PSC infrastructure
 type GCPPrivateServiceConnectSpec struct {
-	// forwardingRuleName is the name of the Internal Load Balancer forwarding rule
+	// loadBalancerIP is the IP address of the Internal Load Balancer
+	// Populated by the observer from service status
 	// +required
-	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MinLength=7
+	// +kubebuilder:validation:MaxLength=15
+	// +kubebuilder:validation:Pattern=`^((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$`
```
csrwng: could this ever be ipv6?

control-plane-operator/controllers/gcpprivateserviceconnect/observer.go
```diff
@@ -0,0 +1,230 @@
...
+	gcpInternalLoadBalancerType   = "Internal"
+	gcpLoadBalancerTypeAnnotation = "networking.gke.io/load-balancer-type"
+)
+
+// GCPPrivateServiceObserver watches a router Service with Internal Load Balancer
+// and reconciles a GCPPrivateServiceConnect CR representation for it.
+type GCPPrivateServiceObserver struct {
+	client.Client
+
+	clientset *kubeclient.Clientset
```
csrwng: why do you need a separate kubernetes clientset?

csrwng: You can just watch for service events and filter them for the service you want to focus on in the event handler.

control-plane-operator/controllers/gcpprivateserviceconnect/observer.go
```diff
@@ -0,0 +1,230 @@
...
+	}))
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (r *GCPPrivateServiceObserver) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	r.log.Info("reconciling")
+
```
csrwng, line 67: if req.Name != r.ServiceName, return

cristianoveiga:
> Some comments. More generally though, I would like to better understand how the link will be setup for GCP clusters. For AWS we have 2 separate controllers operating on the AWSEndpointService resource. On the HO side, we create the ServiceEndpoint (the provider side of the equation), and only when that's created, the Endpoint is created in the customer account. Each side has different credentials to do what it needs to do. Will there be a similar split for GCP?

Exactly - this is what we are going to implement in the next cards ([GCP-197](https://issues.redhat.com//browse/GCP-197) and [GCP-198](https://issues.redhat.com//browse/GCP-198)). I already tested the whole flow.

Here are more details:

The three controllers:

-   Observer (Control-Plane-Operator) - control-plane-operator/controllers/gcpprivateserviceconnect/observer.go:1

  Purpose: Watches the Internal Load Balancer service and creates the initial GCPPrivateServiceConnect CR

  Key Responsibilities:
  - Monitors the router service with annotation networking.gke.io/load-balancer-type: Internal
  - Waits for LoadBalancer to get an IP address from GCP
  - Creates the initial GCPPrivateServiceConnect CR with:
    - LoadBalancerIP from service status
    - ConsumerAcceptList extracted from HostedControlPlane (using the customer project ID)
    - OwnerReference to HostedControlPlane for lifecycle management

  Location: Runs in control-plane-operator pods, activated when hcp.Spec.Platform.Type == hyperv1.GCPPlatform && util.IsPrivateHCP(hcp)

-   Management-Side Controller (HyperShift-Operator) - hypershift-operator/controllers/platform/gcp/privateserviceconnect_controller.go:1

  Purpose: Manages the GCP Service Attachment in the management project/VPC

  Key Responsibilities:
  1. Lookup ForwardingRule: Uses LoadBalancerIP to find the corresponding GCP ForwardingRule name via GCP Compute API
  2. Discover NAT Subnet: Finds available PRIVATE_SERVICE_CONNECT purpose subnets that aren't already in use
  3. Create Service Attachment: Creates GCP Service Attachment with:
    - Target: ForwardingRule URL
    - ConsumerAcceptList: Projects allowed to connect
    - NATSubnet: Dedicated subnet for NAT operations
    - Connection preference: ACCEPT_MANUAL (requires approval)
  4. Update Status: Populates ServiceAttachmentName and ServiceAttachmentURI in the CR status

  Authentication: Uses management cluster's (hypershift-operator) GCP credentials

-   Customer-Side Controller (Control-Plane-Operator) - control-plane-operator/controllers/gcpprivateserviceconnect/psc_endpoint_controller.go:1

  Purpose: Creates PSC endpoint and DNS configuration in the customer project/VPC

  Key Responsibilities:
  1. Wait for Service Attachment: Checks for GCPServiceAttachmentAvailable condition
  2. Reserve IP Address: Creates static internal IP in customer's PSC subnet
  3. Create PSC Endpoint: Creates ForwardingRule that:
    - Points to the management-side Service Attachment URI
    - Uses reserved IP address
    - Connects customer VPC to management VPC
  4. Update Conditions: Sets GCPEndpointAvailable condition based on endpoint state
     
  Authentication: will use customer credentials (via WIF integration )

control-plane-operator/controllers/gcpprivateserviceconnect/observer.go
cristianoveiga: hmm, I followed the AWS pattern here, which also uses the kubernetes clientset to set up the informerFactory and filter events for a specific service (in a namespace).

https://github.com/openshift/hypershift/blob/545830b770850780ac952a51ff6f442ba7485544/control-plane-operator/controllers/awsprivatelink/awsprivatelink_controller.go#L60-L63

And 

https://github.com/openshift/hypershift/blob/545830b770850780ac952a51ff6f442ba7485544/control-plane-operator/controllers/awsprivatelink/awsprivatelink_controller.go#L105-L116

Do you see any issues with this approach ?

api/hypershift/v1beta1/gcpprivateserviceconnect_types.go
cristianoveiga: Good point, I think it can (although it will probably require additional changes in the flow).

I've update the validation to use [isIP](https://kubernetes.io/docs/reference/using-api/cel/#kubernetes-ip-address-library), which can accept both.

control-plane-operator/controllers/gcpprivateserviceconnect/observer.go
cristianoveiga: Understood, but this won't be required if we keep the informer setup (since the filtering is already handled at the informer level).

control-plane-operator/controllers/gcpprivateserviceconnect/observer.go
csrwng: The main thing is that it duplicates memory in a separate informer which is not really needed. The reason the aws controller does what it does is because at the time it was written, the manager cache was scoped to the cluster, and a separate informer was more efficient. That's no longer the case, since now the cache is scoped to the control plane namespace. I think it makes more sense to use the normal controller-runtime client and a Watch with event filter for resource name.

control-plane-operator/controllers/gcpprivateserviceconnect/observer.go
cristianoveiga: added the check/return statement.

control-plane-operator/controllers/gcpprivateserviceconnect/observer.go
cristianoveiga: I see... thanks for the additional context :)

I've removed the additional clientSet/informer set up and added the name filtering in Reconcile method.

api/hypershift/v1beta1/gcpprivateserviceconnect_types.go
csrwng: Unfortunately we can't use isIP because older vanilla kube management clusters don't support it.
See https://github.com/openshift/hypershift/pull/6032

cristianoveiga: Thanks @csrwng , I was trying to avoid a complex regular expression. 

I've updated it.

PR #7219, opened 2025-11-12
GCP-205: feat(gcp): ensure restricted PodSecurity compliance for control plane components

patjlm:
## Summary

Apply GCP-specific PodSecurity compliance settings to all control plane component containers when deployed on GKE management clusters.

## Problem Statement

GKE enforces Kubernetes PodSecurity "restricted:latest" policy requiring container-level security context fields:
- `allowPrivilegeEscalation: false`
- `runAsNonRoot: true`
- `capabilities.drop: ["ALL"]`

Without these settings, control plane pods fail to start with PodSecurity violations.

## Solution

Apply restricted security context to all containers (including sidecars) after they are injected into the pod spec. The implementation automatically preserves any `capabilities.add` fields already declared in deployment templates, allowing a declarative approach where containers specify their capability requirements directly in their manifests.

## Changes

**Modified Files:**
- `support/controlplane-component/defaults.go` - Apply security context for GCP platform after sidecar injection
- `support/util/containers.go` - Security context function with automatic capability preservation
- `support/util/containers_test.go` - Comprehensive test coverage

**Security Context Applied:**
- `AllowPrivilegeEscalation: false`
- `RunAsNonRoot: true`
- `Capabilities.Drop: ["ALL"]`
- Preserves existing `Capabilities.Add` from deployment templates (e.g., haproxy's `NET_BIND_SERVICE`)

## Benefits

- ✅ Eliminates PodSecurity violations on GKE clusters
- ✅ Applies to all containers including dynamically injected sidecars
- ✅ Declarative capability management via deployment templates
- ✅ Platform-specific (only affects GCP deployments)
- ✅ Comprehensive unit test coverage

## Testing

- ✅ Built successfully with `make build`
- ✅ Pre-commit hooks passed
- ✅ Unit tests added and passing
- ✅ Tested on GCP GKE management cluster
- ✅ Verified no PodSecurity violations
- ✅ All control plane pods start successfully

## JIRA

[GCP-205](https://issues.redhat.com/browse/GCP-205)

Co-Authored-By: Claude <noreply@anthropic.com>


support/util/containers_test.go
```diff
@@ -0,0 +1,334 @@
...
+							AllowPrivilegeEscalation: ptr.To(false),
+							RunAsNonRoot:             ptr.To(true),
+							Capabilities: &corev1.Capabilities{
+								Drop: []corev1.Capability{"ALL"},
+							},
+						},
+					},
+				},
+			},
+		},
```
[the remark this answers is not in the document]
cristianoveiga: +1

@patjlm , I think we can extend the coverage of the tests here to confirm that we are overriding `AllowPrivilegeEscalation`, `RunAsNonRoot` and existing capabilities.

cristianoveiga:
@patjlm Thanks for this fix. One comment about the test coverage, but lgtm otherwise.

support/util/containers_test.go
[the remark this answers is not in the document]
patjlm: sounds good, fixed in latest commit. @coderabbitai, please have a second look

patjlm:
I just spotted an other exception for NET_BIND_SERVICE. I will need to rework this a bit.

/hold

support/util/containers.go
```diff
@@ -139,3 +140,52 @@ func WithOptions(opts *AvailabilityProberOpts) AvailabilityProberOpt {
...
+
+		caps := &corev1.Capabilities{
+			Drop: []corev1.Capability{"ALL"},
+		}
+		// Preserve existing capabilities.add if present
+		if podSpec.Containers[i].SecurityContext.Capabilities != nil &&
+			len(podSpec.Containers[i].SecurityContext.Capabilities.Add) > 0 {
+			caps.Add = podSpec.Containers[i].SecurityContext.Capabilities.Add
+		}
+		podSpec.Containers[i].SecurityContext.Capabilities = caps
```
cristianoveiga: This logic is duplicated between regular containers and init containers - can we extract it into a function and reuse it ?

support/util/containers.go
```diff
@@ -139,3 +140,52 @@ func WithOptions(opts *AvailabilityProberOpts) AvailabilityProberOpt {
...
+		if podSpec.Containers[i].SecurityContext == nil {
+			podSpec.Containers[i].SecurityContext = &corev1.SecurityContext{}
+		}
+		podSpec.Containers[i].SecurityContext.AllowPrivilegeEscalation = ptr.To(false)
+		podSpec.Containers[i].SecurityContext.RunAsNonRoot = ptr.To(true)
+
+		caps := &corev1.Capabilities{
+			Drop: []corev1.Capability{"ALL"},
+		}
+		// Preserve existing capabilities.add if present
```
cristianoveiga: Consider having a white list of capabilities that can be applied. 

I had claude generating a categorized list of capabiltiies and this is what they came up with:

```
  ✅ SAFE - Generally OK for containers:

  var safeCapabilities = []corev1.Capability{
      "NET_BIND_SERVICE",    // Bind to privileged ports (< 1024)
      "SETGID",             // Change group ID - needed for some apps
      "SETUID",             // Change user ID - needed for some apps
      "CHOWN",              // Change file ownership - limited use cases
      "DAC_OVERRIDE",       // Bypass file perms - careful use only
      "FOWNER",             // Bypass file ownership checks
      "AUDIT_WRITE",        // Write to audit log
      "NET_RAW",            // Raw sockets - for network tools only
  }

  ⚠️ DANGEROUS - Use with extreme caution:

  var dangerousCapabilities = []corev1.Capability{
      "SYS_ADMIN",          // "god mode" - can break out of containers
      "SYS_MODULE",         // Load kernel modules
      "SYS_PTRACE",         // Debug other processes
      "SYS_BOOT",          // Reboot system
      "SYS_TIME",          // Change system time
      "NET_ADMIN",         // Network configuration
      "SYS_RAWIO",         // Raw hardware access
      "DAC_READ_SEARCH",   // Bypass all file read permissions
  }

  🚫 NEVER - Absolutely forbidden in containers:

  var forbiddenCapabilities = []corev1.Capability{
      "SYS_MODULE",        // Kernel module loading
      "SYS_RAWIO",         // Direct hardware access
      "SYS_BOOT",          // System reboot
      "SYS_TIME",          // System clock manipulation
      "WAKE_ALARM",        // System wakeup
      "BLOCK_SUSPEND",     // Block system suspend
      "SYS_TTY_CONFIG",    // TTY manipulation
      "SYS_PACCT",         // Process accounting
  }
```

patjlm: From https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted only `NET_BIND_SERVICE` should be allowed, so I restricted to this.

support/util/containers.go
patjlm: done

support/util/containers.go
```diff
@@ -139,3 +140,49 @@ func WithOptions(opts *AvailabilityProberOpts) AvailabilityProberOpt {
...
+	container.SecurityContext.AllowPrivilegeEscalation = ptr.To(false)
+	container.SecurityContext.RunAsNonRoot = ptr.To(true)
+
+	var existingAdd []corev1.Capability
+	if container.SecurityContext.Capabilities != nil {
+		existingAdd = container.SecurityContext.Capabilities.Add
+		// Validate capabilities against restricted pod security standards
+		for _, cap := range existingAdd {
+			if cap != "NET_BIND_SERVICE" {
+				return fmt.Errorf("container %q: capability %q is not allowed by restricted pod security standards (only NET_BIND_SERVICE is permitted)", container.Name, cap)
```
muraee, line 160: so this will fail in runtime if any container defines this cap in it's manifest, right?
Ideally this should be caught beforehand in unit test. 
Please add a test case for GCP platform in `TestControlPlaneComponents` similar to IBM here: 
https://github.com/openshift/hypershift/blob/95835694eb99c9afc6207738d64a30db9f37e83b/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller_test.go#L1799-L1803

patjlm: Yes that's the idea. Well it should fail if any other capability than `NET_BIND_SERVICE` tries to be added

If we don't fail, GKE will reject the pod/deployment anyways later.

I'll add this test.

patjlm: done in https://github.com/openshift/hypershift/pull/7219/commits/eff3a3114e0c9a0da947774ba0936de931f6ae65

Thanks for your feedback!

muraee: lgtm, can you please squash the commits, except for the last one adding the test and fixtures

patjlm: woops missed that last message, sorry. Done now.

apahim (approved):
/lgtm

PR #7222, opened 2025-11-12
OCPBUGS-64628: add hypershift-no-cgo to the latest operator conta…

rokej:

Non-FIPS compliant hypershift CLI (hypershift-no-cgo) needs to be in the operator container image so that QE test pipelines can extract this binary to run test cases outside the container.

Fixes https://issues.redhat.com/browse/OCPBUGS-64628

## Checklist:


jparrill (commented):
/lgtm

PR #7223, opened 2025-11-13
AROSLSRE-218: allow hypershift install cmd to accept env vars to be passed to the hypershift operator deployment

ashishmax31:
This PR allows the install cmd to accept env vars as args and pass it to the HO deployment.
https://issues.redhat.com/browse/AROSLSRE-218

## Checklist:

### DIFF
--- a/cmd/install/assets/hypershift_operator.go
+++ b/cmd/install/assets/hypershift_operator.go
@@ -3,6 +3,8 @@ package assets
 import (
 	_ "embed"
 	"fmt"
+	"slices"
+	"strings"
 	"time"
 
 	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
@@ -397,6 +399,7 @@ type HyperShiftOperatorDeployment struct {
 	EnableSizeTagging                       bool
 	EnableEtcdRecovery                      bool
 	EnableCPOOverrides                      bool
+	AdditionalOperatorEnvVars               map[string]string
 	AROHCPKeyVaultUsersClientID             string
 	TechPreviewNoUpgrade                    bool
 	RegistryOverrides                       string
@@ -444,6 +447,23 @@ func (o HyperShiftOperatorDeployment) Build() *appsv1.Deployment {
 			Name:  "CERT_ROTATION_SCALE",
 			Value: o.CertRotationScale.String(),
 		},
+	}
+
+	// Add any additional environment variables specified if they don't already exist.
+	for key, value := range o.AdditionalOperatorEnvVars {
+		trimmedKey := strings.TrimSpace(key)
+		trimmedValue := strings.TrimSpace(value)
+
+		if trimmedKey != "" &&
+			trimmedValue != "" &&
+			!slices.ContainsFunc(envVars, func(e corev1.EnvVar) bool {
+				return e.Name == trimmedKey
+			}) {
+			envVars = append(envVars, corev1.EnvVar{
+				Name:  trimmedKey,
+				Value: trimmedValue,
+			})
+		}
 	}
 
 	// Add the new HYPERSHIFT_FEATURESET env var if TPNU is set.
--- a/cmd/install/assets/hypershift_operator_test.go
+++ b/cmd/install/assets/hypershift_operator_test.go
@@ -20,6 +20,7 @@ func TestHyperShiftOperatorDeployment_Build(t *testing.T) {
 		expectedVolumeMounts []corev1.VolumeMount
 		expectedVolumes      []corev1.Volume
 		expectedArgs         []string
+		expectedEnvVars      []corev1.EnvVar
 	}{
 		"empty oidc parameters result in no volume mounts": {
 			inputBuildParameters: HyperShiftOperatorDeployment{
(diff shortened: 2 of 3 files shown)


cmd/install/install.go
```diff
@@ -277,6 +278,7 @@ func NewCommand() *cobra.Command {
...
 	cmd.PersistentFlags().StringVar(&opts.RegistryOverrides, "registry-overrides", "", "registry-overrides contains the source registry string as a key and the destination registry string as value. Images before being applied are scanned for the source registry string and if found the string is replaced with the destination registry string. Format is: sr1=dr1,sr2=dr2")
 	cmd.PersistentFlags().StringSliceVar(&opts.PlatformsToInstall, "limit-crd-install", opts.PlatformsToInstall, "Used to limit the CRDs that are installed to a per platform basis (example: --limit-crd-install=AWS,Azure). If this flag is not specified, all CRDs for all platforms will be installed. Valid, case-insensitive values are: AWS, Azure, IBMCloud, KubeVirt, Agent, OpenStack.")
+	cmd.PersistentFlags().StringToStringVar(&opts.OperatorEnvironmentVariables, "operator-env-vars", opts.OperatorEnvironmentVariables, "A set of key-value pairs to be set as environment variables on the HyperShift Operator and shared ingress controller deployment.")
```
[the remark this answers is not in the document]
bryan-cox: Agree - @ashishmax31

bryan-cox:
/retitle [AROSLSRE-218](https://issues.redhat.com//browse/AROSLSRE-218): allow hypershift install cmd to accept env vars to be passed to the hypershift operator deployment

cmd/install/install.go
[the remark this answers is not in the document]
ashishmax31: Sorry that was a mistake. Fixed

cmd/install/assets/hypershift_operator.go
```diff
@@ -446,6 +448,16 @@ func (o HyperShiftOperatorDeployment) Build() *appsv1.Deployment {
 		},
 	}
 
+	// Add any additional environment variables specified
+	for key, value := range o.EnvironmentVariables {
+		if strings.TrimSpace(key) != "" && strings.TrimSpace(value) != "" {
```
jparrill: This needs to be changed, if you modify one of the current env vars already set in the hypershift operator deployment you can shoot in the foot pretty easily. The expected behavior should be:
- If the additional env var is new, append it
- If the additional env var already exists, it gets rejected and the original one is kept

This validation should be done because if you add duplicated keys, the apimachinery allows that behavior but the final status of the container will take the last entry from the pod definition.

cmd/install/assets/hypershift_operator_test.go
```diff
@@ -337,6 +338,48 @@ func TestHyperShiftOperatorDeployment_Build(t *testing.T) {
 				fmt.Sprintf("--private-platform=%s", string(hyperv1.NonePlatform)),
 			},
 		},
+		"When OperatorEnvironmentVariables are set, they are included as env vars in the HO deployment": {
```
jparrill: Please add test cases to cover conflicting env var cases. This is an example of the envs in the HO:
```
...
...
        env:
        - name: MY_NAMESPACE
          valueFrom:
            fieldRef:
              apiVersion: v1
              fieldPath: metadata.namespace
        - name: MY_NAME
          valueFrom:
            fieldRef:
              apiVersion: v1
              fieldPath: metadata.name
        - name: METRICS_SET
          value: All
        - name: CERT_ROTATION_SCALE
          value: 24h0m0s
        - name: ENABLE_ETCD_RECOVERY
          value: "1"
        - name: AWS_SHARED_CREDENTIALS_FILE
          value: /etc/provider/credentials
        - name: AWS_REGION
          value: us-west-2
        - name: AWS_SDK_LOAD_CONFIG
          value: "1"
...
...
```

cmd/install/install.go
```diff
@@ -107,6 +107,7 @@ type Options struct {
 	ExternalDNSDomainFilter                   string
 	ExternalDNSTxtOwnerId                     string
 	ExternalDNSImage                          string
+	OperatorEnvironmentVariables              map[string]string
```
jparrill: I would rename it as AdditionalOperatorEnvVars or something like that because there are already existing env vars in the current HO deployment.

cmd/install/install.go
```diff
@@ -347,6 +349,7 @@ func NewInstallOptionsWithDefaults() Options {
 	opts.OIDCStorageProviderS3CredentialsSecretKey = "credentials"
 	opts.PrivatePlatform = string(hyperv1.NonePlatform)
 	opts.ImagePullPolicy = "IfNotPresent"
+	opts.OperatorEnvironmentVariables = map[string]string{}
```
jparrill: Same as above, to fit the reality and make this match with the name of the variable (EG `opts.AdditionalOperatorEnvVars`)

ashishmax31: Done

cmd/install/install.go
ashishmax31: Done

cmd/install/assets/hypershift_operator.go
ashishmax31: Makes sense, fixed.

cmd/install/assets/hypershift_operator.go
```diff
@@ -446,6 +449,18 @@ func (o HyperShiftOperatorDeployment) Build() *appsv1.Deployment {
...
+	// Add any additional environment variables specified if they don't already exist.
+	for key, value := range o.AdditionalOperatorEnvVars {
+		if strings.TrimSpace(key) != "" &&
+			strings.TrimSpace(value) != "" &&
+			!slices.ContainsFunc(envVars, func(e corev1.EnvVar) bool { return e.Name == key }) {
+			envVars = append(envVars, corev1.EnvVar{
+				Name:  key,
+				Value: value,
+			})
+		}
```
jparrill: Small nit: 

WDYT about this? If you adds an space, that will be in the deployment too, this would save you some time. Otherwise LGTM

```
for key, value := range o.AdditionalOperatorEnvVars {
      trimmedKey := strings.TrimSpace(key)
      trimmedValue := strings.TrimSpace(value)

      if trimmedKey != "" &&
          trimmedValue != "" &&
          !slices.ContainsFunc(envVars, func(e corev1.EnvVar) bool {
              return e.Name == trimmedKey
          }) {
          envVars = append(envVars, corev1.EnvVar{
              Name:  trimmedKey,
              Value: trimmedValue,
          })
      }
  }
```

ashishmax31: @jparrill Done

PR #7228, opened 2025-11-13
OCPBUGS-56701: Filter RC versions and validate against supported OCP versions

bryan-cox:

This PR fixes an issue where the CLI could select Release Candidate (RC) versions that are newer than what the installed HyperShift Operator supports when creating hosted clusters without specifying `--release-image` or `--release-stream`.

### The Problem
When users run `hcp create cluster` without specifying a release image, the CLI fetches from the latest release stream. If that stream contains RC versions (e.g., 4.20.0-rc.3) that are newer than what the HyperShift Operator supports (e.g., 4.18), cluster creation fails with a version mismatch error.

### The Solution
This PR updates the `LookupDefaultOCPVersion` function to:

1. **Detect stream type**: Distinguishes between multi-arch streams (ending in `-multi`) and single-arch streams
2. **Apply RC filtering universally**: Uses `retrieveSupportedOCPVersion` for both code paths (previously only used when no stream was provided)
3. **Handle both endpoint types**: Updates `retrieveSupportedOCPVersion` to parse responses from both `/tags` (multi-arch) and `/latest` (single-arch) endpoints
4. **Validate against supported versions**: Ensures selected versions are compatible with the installed HyperShift Operator

### Why This Approach vs PR #6353
The original PR #6353 was reverted because it tried to use multi-arch endpoints for ALL release streams. This caused AWS e2e tests to fail because single-arch streams like "4-stable" only exist on the amd64 endpoint, not the multi endpoint.

This PR solves the issue correctly by:
- ✅ Routing multi-arch streams (`4-stable-multi`) to the multi endpoint
- ✅ Routing single-arch streams (`4-stable`) to the amd64 endpoint  
- ✅ Applying RC filtering and version validation to both
- ✅ Maintaining backward compatibility with existing e2e tests

Fixes https://issues.redhat.com/browse/OCPBUGS-56701

1. **Tests Added**: New comprehensive test cases in `TestIsMultiArchStream` and `TestRetrieveSupportedOCPVersionWithRCFiltering` specifically test:
   - Multi-arch stream detection
   - RC filtering for both `/tags` and `/latest` endpoints
   - Version validation against the supported-versions ConfigMap
   - Error handling for unsupported versions

2. **Backward Compatibility**: This implementation preserves the existing behavior for e2e tests:
   - AWS e2e tests use `--release-stream 4-stable` → routes to amd64 endpoint (unchanged)
   - No changes needed to test infrastructure or CI configuration

3. **Testing Done**:
   - All unit tests pass: `go test ./support/supportedversion/...`
   - New tests validate the exact scenario from the JIRA issue

## Checklist:

🤖 Generated with [Claude Code](https://claude.com/claude-code) via `/jira:solve OCPBUGS-56701`

### DIFF
--- a/support/supportedversion/version_test.go
+++ b/support/supportedversion/version_test.go
@@ -2,6 +2,7 @@ package supportedversion
 
 import (
 	"encoding/json"
+	"fmt"
 	"net/http"
 	"net/http/httptest"
 	"testing"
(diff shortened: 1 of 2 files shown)


support/supportedversion/version_test.go
```diff
@@ -500,3 +500,215 @@ func TestRetrieveSupportedOCPVersion(t *testing.T) {
...
+
+func TestIsMultiArchStream(t *testing.T) {
+	testCases := []struct {
+		name           string
+		releaseStream  string
+		expectedResult bool
+	}{
+		{
+			name:           "When stream ends with -multi, it should be identified as multi-arch",
+			releaseStream:  "4-stable-multi",
```
jparrill, line 601: Are those valid releaseStreams?

bryan-cox: The 4.19-stable-multi is not. The others are:

- https://multi.ocp.releases.ci.openshift.org/#4-stable-multi
- https://amd64.ocp.releases.ci.openshift.org/

jparrill: Not necessary to remove anything, I just want to confirm :)

wewang58:
@bryan-cox Can you help to rebase it? let me test more, thanks.

PR #7250, opened 2025-11-17
CNTRLPLANE-1942: Add connectivity conditions to HostedControlPlane CR

sdminonne:

This PR adds two new condition types to the HostedControlPlane API to signal network connectivity state between the control plane and data plane:

- `DataPlaneToControlPlaneConnectivity` - indicates whether data plane nodes can successfully reach control plane components
- `ControlPlaneToDataPlaneConnectivity` - indicates whether the control plane can successfully reach data plane components

Both conditions follow the True=healthy, False=unhealthy pattern.

This change also creates a dedicated `hostedcontrolplane_conditions.go` file to match the organizational pattern used by HostedCluster, moving existing condition constants to this new file for better consistency and maintainability.

This is an API-only change that provides the foundation for monitoring and reporting connectivity issues. The actual controller logic to update these conditions will be implemented in future work.

Fixes [CNTRLPLANE-1942](https://issues.redhat.com/browse/CNTRLPLANE-1942)

- This PR only adds the condition type constants to the API
- No controller logic is included in this change
- No CRD schema changes are needed as conditions are already part of the HostedControlPlane status
- Documentation has been regenerated to include the new condition types
- This mirrors the approach taken in PR #7242 for HostedCluster

## Checklist:
- [ ] This change includes unit tests. (N/A - API-only change, no logic to test)

🤖 Generated with [Claude Code](https://claude.com/claude-code) via `/jira:solve [CNTRLPLANE-1942](https://issues.redhat.com//browse/CNTRLPLANE-1942) origin`

### DIFF
--- a/docs/content/reference/api.md
+++ b/docs/content/reference/api.md
@@ -4864,7 +4864,10 @@ created in the guest VPC</p>
 has been created for the specified NLB in the management VPC</p>
 </td>
 </tr><tr><td><p>&#34;CVOScaledDown&#34;</p></td>
-<td></td>
+<td><p>CVOScaledDown indicates whether the Cluster Version Operator has been scaled down.
+When True, the CVO is scaled down (typically during maintenance operations).
+When False, the CVO is running normally.</p>
+</td>
 </tr><tr><td><p>&#34;CloudResourcesDestroyed&#34;</p></td>
 <td><p>CloudResourcesDestroyed bubbles up the same condition from HCP. It signals if the cloud provider infrastructure created by Kubernetes
 in the consumer cloud provider account was destroyed.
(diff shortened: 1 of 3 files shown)


api/hypershift/v1beta1/hostedcontrolplane_conditions.go
```diff
@@ -0,0 +1,43 @@
+package v1beta1
+
+// "Condition values may change back and forth, but some condition transitions may be monotonic, depending on the resource and condition type.
```
mgencur, line 3: Personally, I think this comment is unnecessary. We define many more conditions in other files and this comment is not there. It also requires much more knowledge than just this sentence for a person to be able to work with conditions and understand them.

Otherwise this PR looks good to me.

PR #7257, opened 2025-11-18
OCPBUGS-65687: fix(cpov2): prevent azure component access on non-Azure platforms

jhjaggars:

Prevents control-plane-operator from attempting to access Azure-specific resources on non-Azure platforms.

**The Problem:**
The azure-cloud-controller-manager component was attempting to access SecretProviderClass resources on ALL platforms (AWS, KubeVirt, OpenStack, etc.), not just Azure. This caused the controller to create informers for resources that should never exist on those platforms, resulting in continuous RBAC forbidden errors that blocked control plane reconciliation.

**Root Cause:**
The azure-cloud-controller-manager component only used `.WithPredicate()` which triggers cleanup mode on non-Azure platforms. In cleanup mode, the framework calls `Client.Get()` to check for existing resources before deletion. When called with platform-specific resource types like SecretProviderClass, controller-runtime automatically creates informers that continuously attempt to LIST/WATCH those resources, even though they will never exist on non-Azure platforms.

**The Fix:**
Add `.WithPlatformPredicate()` to azure-cloud-controller-manager, which causes the component to skip entirely on non-Azure platforms without any API calls or cleanup attempts.

Both predicates are now present:
- `WithPlatformPredicate` (platform filtering) - checked first, exits early on wrong platform
- `WithPredicate` (configuration-based) - checked second, for disabling on correct platform

The generic-adapter implementation enforces this evaluation order automatically.

**Impact:**
- Control-plane-operator no longer attempts to access Azure resources on non-Azure platforms
- No unnecessary informer creation for platform-specific resource types
- Control plane reconciliation proceeds normally on all platforms

Fixes OCPBUGS-65687

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/v2/assets/assets.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/v2/assets/assets.go
@@ -85,7 +85,7 @@ func LoadManifestInto(componentName string, fileName string, into client.Object)
 	return obj.(client.Object), gvk, err
 }
 
-func ForEachManifest(componentName string, action func(manifestName string) error) error {
+func ForEachManifest(componentName string, manifestFilter func(manifestName string) bool, action func(manifestName string) error) error {
 	return fs.WalkDir(manifestsAssets, componentName, func(path string, d fs.DirEntry, err error) error {
 		if err != nil {
 			return err
@@ -95,6 +95,12 @@ func ForEachManifest(componentName string, action func(manifestName string) erro
 		}
 		manifestName := d.Name()
 		if manifestName == deploymentManifest || manifestName == statefulSetManifest || manifestName == cronJobManifest || manifestName == jobManifest {
+			return nil
+		}
+
+		// Apply filter if provided - skip manifest if filter returns false
+		// This prevents LoadManifest from being called, avoiding informer creation
+		if manifestFilter != nil && !manifestFilter(manifestName) {
 			return nil
 		}
 
--- a/control-plane-operator/controllers/hostedcontrolplane/v2/cloud_controller_manager/azure/component.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/v2/cloud_controller_manager/azure/component.go
@@ -37,7 +37,7 @@ func (c *azureOptions) NeedsManagementKASAccess() bool {
 func NewComponent() component.ControlPlaneComponent {
 	return component.NewDeploymentComponent(ComponentName, &azureOptions{}).
 		WithAdaptFunction(adaptDeployment).
-		WithPredicate(predicate).
+		WithPlatformPredicate(predicate).
 		WithManifestAdapter(
 			"serviceaccount.yaml",
 			component.WithAdaptFunction(adaptServiceAccount),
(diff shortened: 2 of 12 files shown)


bryan-cox:
You'll probably need some work around this in the PR as well -https://github.com/openshift/hypershift/blob/076b764d7615eab0336b228a04e8b5d648600114/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go#L275

jhjaggars:
> You'll probably need some work around this in the PR as well -
> 
> https://github.com/openshift/hypershift/blob/076b764d7615eab0336b228a04e8b5d648600114/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go#L275

I'm not sure that I understand.  In the above case all the RBAC will have been applied at this point right?

csrwng:
@jhjaggars looks good, but my preference would be to fix the cpo v2 insistence on touching code that it should completely skip. So right now we use the component predicate functions to skip a component, but even if the predicate says to skip, we still try to go and fetch the resource and potentially delete it.
See: 
https://github.com/openshift/hypershift/blob/545830b770850780ac952a51ff6f442ba7485544/support/controlplane-component/generic-adapter.go#L51-L67
and
https://github.com/openshift/hypershift/blob/545830b770850780ac952a51ff6f442ba7485544/support/controlplane-component/controlplane-component.go#L158-L166

The problem is that we're using predicate to indicate 2 different things:
* completely ignore this because it's the wrong platform
* this should not exist based on current settings (but could come back at some point if you change settings)

My proposal would be to add another predicate (call it "platformPredicate") that if false, means just ignore this component/resource completely, don't attempt to delete it, don't attempt to fetch it.

wdyt?

bryan-cox:
> > You'll probably need some work around this in the PR as well -
> > https://github.com/openshift/hypershift/blob/076b764d7615eab0336b228a04e8b5d648600114/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go#L275
> 
> I'm not sure that I understand. In the above case all the RBAC will have been applied at this point right?

@jhjaggars - Good question. You're right that the RBAC will be applied. However, I'm wondering if we should also make the *watching* of SecretProviderClass unconditional for consistency, not just the RBAC permissions.

  Currently, line 275-277 only adds SecretProviderClass to managedResources when IsAroHCP() is true, which means the hostedcluster controller only actively watches these resources in ARO HCP mode.

  Since we're now granting RBAC unconditionally (to handle cases where the CRD is installed), should we also watch them unconditionally? Or is there a reason to keep the watching conditional while making RBAC unconditional?

jhjaggars:
@csrwng I added some commits on top to explore what adding a platformPredicate to the framework might look like.  It's fairly invasive and due to what happens at schema registration time, I think it's insufficient (and probably not worth going this route without some more thought)

Edit: After some more digging, I discovered an order of operations issue
(we actually have to register the platform predicate on the builder not in the manifest adapter).  It's still pretty invasive, but probably worth checking out now.

support/api/scheme.go
```diff
@@ -136,7 +136,11 @@ func init() {
 		_ = machinev1beta1.AddToScheme(scheme)
 		_ = capiopenstackv1alpha1.AddToScheme(scheme)
 		_ = capiopenstackv1beta1.AddToScheme(scheme)
-		_ = secretsstorev1.AddToScheme(scheme)
+		// SecretProviderClass is only used by ARO-HCP managed service
+		// Conditionally register to prevent informer creation on other platforms
```
jhjaggars: This is the most unfortunate part.  Due to the way this works if we don't skip it this way we'll still try to fetch resoruces we don't have permission for and fail to provision the control plane components.

xiuwang:
/verified by @xiuwang

Test comment in [OCPBUGS-65687](https://issues.redhat.com/browse/OCPBUGS-65687?focusedId=28584181&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-28584181)

PR #7258, opened 2025-11-18
OCPBUGS-63452: Add finalizer protection to control plane workloads

clebs:

This PR adds finalizer protection for critical control plane deployments to prevent orphaned cloud resources when workloads are accidentally deleted.

**Problem**: Control plane deployments (capi-provider, cluster-api, cloud-controller-manager-*) lack finalizers. If these deployments are accidentally deleted before HostedCluster deletion, the controller pods stop running before they can process finalizers on their managed CAPI resources (MachineDeployment, Machine, platform-specific machines). This leads to orphaned cloud resources (EC2 instances, VMs, disks, NICs, load balancers) that continue running and generating costs.

**Solution**: Add a finalizer to all control plane workloads that require it. The finalizer ensures:
1. Deployments aren't deleted immediately when deletion is requested, leaving orphaned resources.
2. CAPI resources can be marked for deletion and their finalizers processed by still-running controllers
3. Only after the underlying infrastructure is deleted are the finalizers removed, allowing safe removal of the HC namespace.

**Affected components**:
- cluster-api
- capi-provider
- cloud-controller-manager (for each provider)
- karpenter-operator
- cluster-autoscaler

Fixes OCPBUGS-63452

- **Implementation approach**: Framework-level (support/controlplane-component) rather than per-component to ensure consistent protection across all workloads
- **Test fixture updates**: Test fixtures for affected components have been updated to reflect the new finalizer being added to deployments.
- **Deletion flow**: Finalizers are removed from the deployments after the underlying infrastructure is deleted.

**What reviewers should verify**:
- Code changes in `support/controlplane-component/controlplane-component.go` and `hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go`.
- Test fixtures for affected components components include the finalizer in metadata.finalizers.

## Checklist:

### DIFF
--- a/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
+++ b/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
@@ -144,8 +144,11 @@ const (
 	referencedResourceAnnotationPrefix    = "referenced-resource.hypershift.openshift.io/"
 )
 
-// NoopReconcile is just a default mutation function that does nothing.
-var NoopReconcile controllerutil.MutateFn = func() error { return nil }
+var (
+	// NoopReconcile is just a default mutation function that does nothing.
+	NoopReconcile  controllerutil.MutateFn = func() error { return nil }
+	CAPIComponents                         = []string{capimanagerv2.ComponentName, capiproviderv2.ComponentName}
+)
 
 // HostedClusterReconciler reconciles a HostedCluster object
 type HostedClusterReconciler struct {
@@ -4885,10 +4888,9 @@ func (r *HostedClusterReconciler) reconcileCAPIFinalizers(ctx context.Context, h
 		return nil
 	}
 
-	capiComponents := []string{capimanagerv2.ComponentName, capiproviderv2.ComponentName}
 	namespace := manifests.HostedControlPlaneNamespace(hc.Namespace, hc.Name)
 
-	for _, name := range capiComponents {
+	for _, name := range CAPIComponents {
 		deployment := &appsv1.Deployment{
 			ObjectMeta: metav1.ObjectMeta{
 				Name:      name,
--- a/test/e2e/util/util.go
+++ b/test/e2e/util/util.go
@@ -24,6 +24,7 @@ import (
 	cpomanifests "github.com/openshift/hypershift/control-plane-operator/controllers/hostedcontrolplane/manifests"
 	hccokasvap "github.com/openshift/hypershift/control-plane-operator/hostedclusterconfigoperator/controllers/resources/kas"
 	hccomanifests "github.com/openshift/hypershift/control-plane-operator/hostedclusterconfigoperator/controllers/resources/manifests"
+	hcc "github.com/openshift/hypershift/hypershift-operator/controllers/hostedcluster"
 	hcmetrics "github.com/openshift/hypershift/hypershift-operator/controllers/hostedcluster/metrics"
 	"github.com/openshift/hypershift/hypershift-operator/controllers/manifests"
 	controlplaneoperatoroverrides "github.com/openshift/hypershift/hypershift-operator/controlplaneoperator-overrides"
@@ -2136,7 +2137,6 @@ func waitForDaemonSetsReady(t *testing.T, ctx context.Context, guestClient crcli
 				return true, nil
 			}
 		})
-
 		if err != nil {
 			return fmt.Errorf("failed to wait for DaemonSet %s to be ready: %w", dsName, err)
 		}
--- a/test/e2e/create_cluster_test.go
+++ b/test/e2e/create_cluster_test.go
@@ -2348,6 +2348,7 @@ func TestCreateCluster(t *testing.T) {
 		e2eutil.EnsureCustomTolerations(t, ctx, mgtClient, hostedCluster)
 		e2eutil.EnsureAppLabel(t, ctx, mgtClient, hostedCluster)
 		e2eutil.EnsureFeatureGateStatus(t, ctx, guestClient)
+		e2eutil.EnsureCAPIFinalizers(t, ctx, mgtClient, hostedCluster)
 
 		// ensure KAS DNS name is configured with a KAS Serving cert
 		e2eutil.EnsureKubeAPIDNSNameCustomCert(t, ctx, mgtClient, hostedCluster)


clebs:
/verified remove

Must be manually verified by attempting to reproduce the bug as described.

support/controlplane-component/controlplane-component.go
```diff
@@ -326,8 +337,78 @@ func (c *controlPlaneWorkload[T]) reconcileWorkload(cpContext ControlPlaneContex
 		return err
 	}
 
+	// Ensure the workload has a finalizer for cleanup
```
rtheis: Per discussion with @csrwng during our IBM/Red Hat interlock meeting, I think that we need this scoped to the components of interest since applying the finalizer to all components may break how ROKS handles cluster deletion.

clebs: Yes, I am currently reworking this to correctly apply to the relevant deployments.

support/controlplane-component/controlplane-component.go
```diff
@@ -14,15 +16,21 @@ import (
...
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
+const ControlPlaneComponentFinalizer = "hypershift.openshift.io/component-finalizer"
+
+var managedComponents = []string{"cluster-api", "capi-provider", "cloud-controller-manager", "karpenter-operator", "cluster-autoscaler"}
```
muraee: I think we are only interested in "cluster-api" and "capi-provider". Why do we need the rest?
- `cloud-controller-manager` is deployed by the CPO, so it's protected from deletion by `ownerReference`
- `cluster-autoscaler` doesn't create or manage any resources.
- `karpenter-operator`  doesn't depend on CAPI resources.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -3202,6 +3202,11 @@ func (r *HostedClusterReconciler) delete(ctx context.Context, hc *hyperv1.Hosted
 			}
 		}
 
+		// once infra is deleted remove finalizers from managed deployments.
+		if err := r.removeControlPlaneDeploymentFinalizers(ctx, controlPlaneNamespace); err != nil {
+			return false, fmt.Errorf("failed to remove finalizer from managed deployments: %w", err)
+		}
```
muraee:
```suggestion
      if !exist {
		  if err := r.removeControlPlaneDeploymentFinalizers(ctx, controlPlaneNamespace); err != nil {
			  return false, fmt.Errorf("failed to remove finalizer from managed deployments: %w", err)
		  }
	   }
```
should check if the cluster doesn't exist before removing the finalizers

support/controlplane-component/controlplane-component.go
clebs: I added them since they are mentioned in the Jira issue as potentially also needing it, but if it is not the case I will leave it at the 2 that really do need it.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
clebs: Of course, good catch! Fixed thanks!

support/controlplane-component/controlplane-component.go
```diff
@@ -315,6 +326,19 @@ func (c *controlPlaneWorkload[T]) reconcileWorkload(cpContext ControlPlaneContex
 		}
 	}
 
+	if c.requiresFinalizer() {
+		// add finalizer to deployments managed by the hosted cluster controller.
+		// Finalizer removal happens during hc deletion reconcile.
+		if deployment, ok := any(workloadObj).(*appsv1.Deployment); ok {
```
muraee: now that we only add finalizers for capi which is deployed by the HO, let's move the logic to the HO directly instead of adding it to the framework.

should be easier to track when adding/deleting the finalizers is in one place.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -4857,3 +4864,26 @@ func (r *HostedClusterReconciler) reconcileAdditionalTrustBundle(ctx context.Con
...
 	return nil
 }
+
+// removeControlPlaneDeploymentFinalizers removes finalizers from control plane deployments
+// after CAPI resources have been cleaned up to allow deployment deletion
+func (r *HostedClusterReconciler) removeControlPlaneDeploymentFinalizers(ctx context.Context, namespace string) error {
+	log := ctrl.LoggerFrom(ctx)
+
+	deploymentList := &appsv1.DeploymentList{}
+	if err := r.List(ctx, deploymentList, client.InNamespace(namespace)); err != nil {
```
muraee: no need to list all deployments, we just need  `capi` and `capi-provider`

Also note that when the `DisableMachineManagement` annotation is set, we don't deploy capi/capi-provider. So, no need to add/remove finalizers if the annotation is set

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -1207,6 +1213,16 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
 		}
 	}
 
+	// Ensure the CAPI deployments have finalizers
+	if hcp != nil {
+		_, machineManagementDisabled := hcp.Annotations[hyperv1.DisableMachineManagement]
+		if !machineManagementDisabled {
```
muraee: we could pass the `hcluster` object to the function and let it return early if the annotation is set, instead of checking it every time before calling it.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -4857,3 +4882,47 @@ func (r *HostedClusterReconciler) reconcileAdditionalTrustBundle(ctx context.Con
 
 	return nil
 }
+
+// reconcileCAPIFinalizers adds or removes finalizers from CAPI deployments managed by the controller.
+func (r *HostedClusterReconciler) reconcileCAPIFinalizers(ctx context.Context, namespace string, remove bool) error {
+	log := ctrl.LoggerFrom(ctx)
+
+	capiDeployments := []*appsv1.Deployment{}
```
muraee: why do we need this intermediate slice? why not everything in one loop?

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -4857,3 +4882,47 @@ func (r *HostedClusterReconciler) reconcileAdditionalTrustBundle(ctx context.Con
...
+				controllerutil.RemoveFinalizer(deployment, ControlPlaneComponentFinalizer)
+			}
+		} else {
+			if !controllerutil.ContainsFinalizer(deployment, ControlPlaneComponentFinalizer) {
+				log.Info("adding finalizer to CAPI deployment")
+				controllerutil.AddFinalizer(deployment, ControlPlaneComponentFinalizer)
+			}
+		}
+
+		if err := r.Update(ctx, deployment); err != nil {
```
muraee: we should track above if we actually added/removed the finalizer and only call `update` if we did.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
clebs: I considered that, wanted to keep it single responsibility. But yes, we can have that in the function to make sure it never runs if the annotation is set.

Does the `hcluster` have the annotation same as the `hcp`? Also if we do it like this I am assuming we do not need to nil-check the `hcp`.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -145,8 +146,13 @@ const (
...
 )
 
-// NoopReconcile is just a default mutation function that does nothing.
-var NoopReconcile controllerutil.MutateFn = func() error { return nil }
+var (
+	// NoopReconcile is just a default mutation function that does nothing.
+	NoopReconcile controllerutil.MutateFn = func() error { return nil }
+
+	// capiComponents contains the list of CAPI deployments managed by the controller that need to be prorected against deletion.
+	capiComponents = []string{"cluster-api", "capi-provider"}
```
muraee: I don't think this needs to be a global var, could be defined inside the function that uses it.
Also, please use the constants from the components. 
```suggestion
	capiComponents = []string{capimanagerv2.ComponentName, capiproviderv2.ComponentName}
```
https://github.com/openshift/hypershift/blob/main/control-plane-operator/controllers/hostedcontrolplane/v2/capi_manager/component.go#L9

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
clebs: We can do it in one go yes, I kept it like this mostly because of readability.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
clebs: Same as before, readability and avoid duplication, but yes, we can call update on those 2 conditional blocks to only update if there has been a change.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
muraee: All annotation are set on the HostedCluster and then copied to the HCP, nothing is set directly on the hcp by users.

> Also if we do it like this I am assuming we do not need to nil-check the hcp.

yep

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
muraee: I meant to have a variable to track if we should update or not that we set to `true` in the conditional blocks. not to duplicate the update call

clebs: Right, that is actually a great idea, since `Add/RemoveFinalizer` returns wether the finalizer list has been updated. We can use that return bool to precisely know if we need to send an update request.

clebs:
/remove-area control-plane-operator

The new approach only adds changes to the hypershift operator.

test/e2e/util/util.go
```diff
@@ -1012,6 +1012,29 @@ func EnsureFeatureGateStatus(t *testing.T, ctx context.Context, guestClient crcl
 	})
 }
 
+func EnsureCAPIFinalizers(t *testing.T, ctx context.Context, client crclient.Client, hostedCluster *hyperv1.HostedCluster) {
+	t.Run("EnsureCustomLabels", func(t *testing.T) {
```
muraee:
```suggestion
	t.Run("EnsureCAPIFinalizers", func(t *testing.T) {
```

rtheis (approved):
/lgtm from IBM Cloud perspective

devguyio:
/uncc @devguyio @jparrill @bryan-cox @csrwng 
(doing this to test auto-cc)

wewang58:
> /assign @wewang58

Yes, I am building hypershift operator image and launching a cluster to test it.

PR #7259, opened 2025-11-18
GCP-232: refactor(rbac): apply extension-apiserver-authentication rolebinding unconditionally

cblecker:
## Summary

This PR simplifies the creation of the `extension-apiserver-authentication-reader` RoleBinding by applying it unconditionally to all clusters at installation time, removing platform-specific conditional logic. Previously this RoleBinding was only created for ARO HCP clusters.

## Changes

- **Renamed Type**: `KubeSystemRoleBinding` → `HyperShiftExtensionAPIServerAuthenticationReaderRoleBinding` (follows HyperShift naming conventions)
- **Removed Conditional Logic**: Applied to all clusters instead of only ARO HCP (`opts.ManagedService == hyperv1.AroHCP`)
- **Updated RoleBinding Name**: `authentication-reader-for-authenticated-users` → `hypershift:extension-apiserver-authentication-reader`
- **Changed Subject**: `system:authenticated` → `system:serviceaccounts` (more secure, least privilege)
- **Use Kubernetes Constant**: `metav1.NamespaceSystem` instead of hardcoded `"kube-system"`
- **Removed Namespace Field**: Always kube-system, no configuration needed
- **Added Comprehensive Documentation**: Explains why this RoleBinding is needed and why it's safe

## Motivation

The `extension-apiserver-authentication-reader` Role provides read access to the client CA bundle used to verify client certificates for authentication. HyperShift components that serve metrics endpoints need this information to properly authenticate requests.

### Why Unconditional Application is Safe

1. **Read-only access**: The Role only grants read access to authentication configuration metadata, not write permissions
2. **Standard practice**: OpenShift clusters include this RoleBinding out of the box via `cluster-kube-apiserver-operator`
3. **Required for metrics**: HyperShift instantiates multiple metrics-serving operators dynamically across many namespaces as hosted clusters are created
4. **No harm to OpenShift**: Clusters that already have this RoleBinding are unaffected by creating another one
5. **Fixes gaps**: Non-OpenShift platforms (vanilla K8s, AKS, GKE) need this for metrics to work correctly

### Why This is Better Than Conditional Logic

The previous approach only created this for ARO HCP. This unconditional approach:
- **Simpler**: No platform detection or conditionals needed
- **More maintainable**: Less code, clearer intent
- **Future-proof**: Any new platform automatically gets necessary permissions
- **Consistent**: All clusters have the same RBAC configuration

## Testing

- [x] Build succeeds (`make hypershift`)
- [x] Verified RoleBinding structure and naming
- [x] Confirmed subject uses `system:serviceaccounts`
- [x] Verified namespace uses `metav1.NamespaceSystem` constant

### DIFF
--- a/cmd/install/install.go
+++ b/cmd/install/install.go
@@ -989,14 +989,11 @@ func setupRBAC(opts Options, operatorNamespace *corev1.Namespace) (*corev1.Servi
 	}.Build()
 	objects = append(objects, operatorRoleBinding)
 
-	// In OpenShift management clusters, this RoleBinding is brought in through openshift/cluster-kube-apiserver-operator.
-	// In ARO HCP, Hosted Clusters are running on AKS management clusters, so we need to provide this through the HO.
-	if opts.ManagedService == hyperv1.AroHCP {
-		roleBinding := assets.KubeSystemRoleBinding{
-			Namespace: "kube-system",
-		}.Build()
-		objects = append(objects, roleBinding)
-	}
+	// HyperShift components that serve metrics need access to the client CA bundle
+	// for request authentication. This RoleBinding grants service accounts read access
+	// to the extension-apiserver-authentication-reader Role in kube-system namespace.
+	roleBinding := assets.HyperShiftExtensionAPIServerAuthenticationReaderRoleBinding{}.Build()
+	objects = append(objects, roleBinding)
 
 	if opts.EnableAdminRBACGeneration {
 		clientObjs := setupAdminRBAC(operatorNamespace)
--- a/cmd/install/assets/hypershift_operator.go
+++ b/cmd/install/assets/hypershift_operator.go
@@ -2004,30 +2004,40 @@ func (o HyperShiftPullSecret) Build() *corev1.Secret {
 	}
 }
 
-type KubeSystemRoleBinding struct {
-	Namespace string
-}
+// HyperShiftExtensionAPIServerAuthenticationReaderRoleBinding creates a RoleBinding in the
+// kube-system namespace that grants service accounts read access to the
+// extension-apiserver-authentication-reader Role.
+//
+// This RoleBinding provides access to the client CA bundle used to verify client certificates
+// for authentication. HyperShift components that serve metrics endpoints need this information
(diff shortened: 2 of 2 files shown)


cmd/install/install.go
patjlm, line 295: i might be mistaken, but i am not sure this function is called when installing hypershift from YAML/template/helm. Is it?

If it is not, we will not run through the detection code being added here.

cmd/install/install_render.go
```diff
@@ -109,7 +110,9 @@ func RenderHyperShiftOperator(cmdOut io.Writer, opts *Options) error {
 		}
 		objects = []crclient.Object{templateObject}
 	} else {
-		crds, objects, err = hyperShiftOperatorManifests(*opts)
+		// For rendering, assume no capabilities (vanilla Kubernetes) as the safest default
+		mgmtClusterCaps := &fake.FakeSupportNoCapabilities{}
+		crds, objects, err = hyperShiftOperatorManifests(*opts, mgmtClusterCaps)
```
patjlm: so this means the rendered YAML/template will contain the rolebinding, even when being applied on Openshift.

In the end all installs will end up with the rolebinding, except when running directly `hypershift install` on an openshift cluster. Right?

patjlm: You might need an equivalent for helm rendering (`hypershift install helm`)

patjlm: I'd actually be fine with this, but then we don't need any detection mechanism and we just always add this rolebinding.

cmd/install/install.go
cblecker: Yeah if doing a render install, this logic won't trip.

cmd/install/install_render.go
cblecker: That's correct. If you're rendering, we assume non-OpenShift

bryan-cox: > If you're rendering, we assume non-OpenShift

I'm not sure that is always the case though, but I also don't know that this would hurt anything either 🤔

csrwng:
/override ci/prow/e2e-aks 
node timeout on aks, not related to this PR

PR #7260, opened 2025-11-19
CNTRLPLANE-1956: add logic to check control-plane to data-plane connectivity 

sdminonne:
We need this PR to create and update a status.condition in hosted-control-plane CR to show the status of connection between control-plane and data-plane. 
It lists the konnecivity-agent PODs and it get the logs.

Fixes [CNTRLPLANE-1956](https://issues.redhat.com//browse/CNTRLPLANE-1956)

## Checklist:

### DIFF
--- a/api/hypershift/v1beta1/hostedcluster_conditions.go
+++ b/api/hypershift/v1beta1/hostedcluster_conditions.go
@@ -195,6 +195,15 @@ const (
 	// This condition is used to track the status of the recovery process and to determine if the HostedCluster
 	// is ready to be used after restoration.
 	HostedClusterRestoredFromBackup ConditionType = "HostedClusterRestoredFromBackup"
+
+	// DataPlaneConnectionAvailable indicates whether the control plane has a successful
+	// network connection to the data plane components.
+	// **True** means the control plane can successfully reach the data plane nodes.
+	// **False** means there are network connection issues preventing the control plane from reaching the data plane.
+	// A failure here suggests potential issues such as: network policy restrictions,
+	// firewall rules, missing data plane nodes, or problems with infrastructure
+	// components like the konnectivity-agent workload.
+	DataPlaneConnectionAvailable ConditionType = "DataPlaneConnectionAvailable"
 )
 
 // Reasons.
@@ -250,7 +259,15 @@ const (
 
 	RecoveryFinishedReason = "RecoveryFinished"
 
+	ReconcileErrorReason = "ReconcileError"
+
 	CloudResourcesCleanupSkippedReason = "CloudResourcesCleanupSkipped"
+
+	DataPlaneConnectionNoKonnectivityAgentPodsNotFoundReason = "KonnectivityAgentPodsNotFound"
+
+	DataPlaneConnectionLogsAccessFailedReason = "LogsAccessFailed"
+
+	DataPlaneConnectionNoWorkerNodesAvailableReason = "NoWorkerNodesAvailable"
 )
 
 // Messages.
--- a/control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
+++ b/control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
@@ -71,6 +71,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
+	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	apiregistrationv1 "k8s.io/kube-aggregator/pkg/apis/apiregistration/v1"
 	"k8s.io/utils/ptr"
@@ -127,6 +128,7 @@ exec /bin/azure-cloud-node-manager \
 type reconciler struct {
 	client         client.Client
 	uncachedClient client.Client
+	clientSet      *clientset.Clientset
 	upsert.CreateOrUpdateProvider
 	platformType              hyperv1.PlatformType
 	rootCA                    string
@@ -144,6 +146,18 @@ type reconciler struct {
 	operateOnReleaseImage     string
 	ImageMetaDataProvider     util.ImageMetadataProvider
 	cleanupTracker            *util.CleanupTracker
+
+	// exposed for unit test since GetLogs looks hard to be mocked
+	GetPodLogs func(context context.Context, clientset *clientset.Clientset, namespace, name, container string) ([]byte, error)
+}
+
+func getPodLogs(ctx context.Context, clientSet *clientset.Clientset, namespace, name, container string) ([]byte, error) {
+	limit := int64(1024)
+	opts := &corev1.PodLogOptions{
+		Container:  container,
+		LimitBytes: &limit,
+	}
+	return clientSet.CoreV1().Pods(namespace).GetLogs(name, opts).DoRaw(ctx)
 }
 
 // eventHandler is the handler used throughout. As this controller reconciles all kind of different resources
@@ -181,9 +195,15 @@ func Setup(ctx context.Context, opts *operator.HostedClusterConfigOperatorConfig
 		return fmt.Errorf("failed to create kubevirt infra uncached client: %w", err)
 	}
 
+	clientset, err := clientset.NewForConfig(opts.Manager.GetConfig())
+	if err != nil {
+		return fmt.Errorf("failed to initialize kubeClient from config: %w", err)
+	}
+
 	c, err := controller.New(ControllerName, opts.Manager, controller.Options{Reconciler: &reconciler{
 		client:                    opts.Manager.GetClient(),
 		uncachedClient:            uncachedClient,
+		clientSet:                 clientset,
 		CreateOrUpdateProvider:    opts.TargetCreateOrUpdateProvider,
 		platformType:              opts.PlatformType,
 		rootCA:                    opts.InitialCA,
@@ -201,6 +221,7 @@ func Setup(ctx context.Context, opts *operator.HostedClusterConfigOperatorConfig
 		operateOnReleaseImage:     opts.OperateOnReleaseImage,
 		ImageMetaDataProvider:     opts.ImageMetaDataProvider,
 		cleanupTracker:            util.NewCleanupTracker(),
+		GetPodLogs:                getPodLogs,
 	}})
 	if err != nil {
 		return fmt.Errorf("failed to construct controller: %w", err)
@@ -537,6 +558,11 @@ func (r *reconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Result
 	log.Info("reconciling konnectivity agent")
 	if err := r.reconcileKonnectivityAgent(ctx, hcp, releaseImage); err != nil {
 		errs = append(errs, fmt.Errorf("failed to reconcile konnectivity agent: %w", err))
+	}
+
+	log.Info("reconciling control-Plane to data-Plane status conditions")
+	if err := r.reconcileControlPlaneDataPlaneConnectivityConditions(ctx, hcp, log); err != nil {
+		errs = append(errs, fmt.Errorf("failed to update ControlPlaneToDataPlaneConnectivity condition: %w", err))
 	}
 
 	log.Info("reconciling openshift apiserver apiservices")
@@ -1387,6 +1413,79 @@ func (r *reconciler) reconcileClusterVersion(ctx context.Context, hcp *hyperv1.H
 	}
 
 	return nil
+}
+
+func (r *reconciler) reconcileControlPlaneDataPlaneConnectivityConditions(ctx context.Context, hcp *hyperv1.HostedControlPlane, log logr.Logger) error {
+	patchHCPWithCondition := func(hcp *hyperv1.HostedControlPlane, condition *metav1.Condition) error {
+		originalHCP := hcp.DeepCopy()
+		if !meta.SetStatusCondition(&hcp.Status.Conditions, *condition) {
+			return nil // No status change; avoid unnecessary API call.
+		}
+		if err := r.cpClient.Status().Patch(ctx, hcp, client.MergeFrom(originalHCP)); err != nil {
+			return fmt.Errorf("failed to update HostedControlPlane status with %s condition: %w", condition.Type, err)
+		}
+		log.Info(string(hyperv1.DataPlaneConnectionAvailable) + " updated")
+		return nil
+	}
+
+	condition := &metav1.Condition{
+		Type:   string(hyperv1.DataPlaneConnectionAvailable),
+		Status: metav1.ConditionFalse, // False by default
+	}
+	totalNodes, err := util.CountAvailableNodes(ctx, r.client)
+	if err != nil {
+		condition.Status = metav1.ConditionUnknown
+		condition.Reason = hyperv1.ReconcileErrorReason
+		condition.Message = "Unable to count worker nodes: " + err.Error()
+		return patchHCPWithCondition(hcp, condition)
+	}
+	if totalNodes == 0 {
+		condition.Status = metav1.ConditionUnknown
+		condition.Reason = hyperv1.DataPlaneConnectionNoWorkerNodesAvailableReason
+		condition.Message = "No worker nodes available"
+		return patchHCPWithCondition(hcp, condition)
+	}
+	var podList corev1.PodList
+	if err := r.uncachedClient.List(ctx, &podList,
+		client.MatchingLabels{"app": "konnectivity-agent"}, client.InNamespace("kube-system")); err != nil {
+		condition.Reason = hyperv1.ReconciliationErrorReason
+		condition.Message = "Couldn't list konnectivity-agent PODs in kube-system namespace: " + err.Error()
+		return patchHCPWithCondition(hcp, condition)
+	}
+
+	logsFound := false
+	runningPodsFound := false
+	for _, pod := range podList.Items {
+		if pod.Status.Phase != corev1.PodRunning {
+			continue
+		}
+		runningPodsFound = true
+		data, err := r.GetPodLogs(ctx, r.clientSet, pod.Namespace, pod.Name, "konnectivity-agent")
+		if err != nil {
+			log.Error(err,
+				fmt.Sprintf("failed to get logs for konnectivity-agent pod %s/%s", pod.Namespace, pod.Name))
+			continue
+		}
+		if len(data) > 0 {
+			logsFound = true
+			break
+		}
+	}
+	if !runningPodsFound {
+		condition.Reason = hyperv1.DataPlaneConnectionNoKonnectivityAgentPodsNotFoundReason
+		condition.Message = "Couldn't find any konnectivity-agent running in data plane"
+	} else {
+		if !logsFound {
+			condition.Reason = hyperv1.DataPlaneConnectionNoKonnectivityAgentPodsNotFoundReason
+			condition.Message = "failed to read konnectivity-agent logs from data plane"
+		} else {
+			condition.Status = metav1.ConditionTrue
+			condition.Reason = hyperv1.AsExpectedReason
+			condition.Message = hyperv1.AllIsWellMessage
+		}
+	}
+
+	return patchHCPWithCondition(hcp, condition)
 }
 
 func (r *reconciler) reconcileOpenshiftAPIServerAPIServices(ctx context.Context, hcp *hyperv1.HostedControlPlane) error {
--- a/control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources_test.go
+++ b/control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources_test.go
@@ -38,6 +38,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/validation"
+	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/utils/ptr"
 
 	controllerruntime "sigs.k8s.io/controller-runtime"
@@ -163,7 +164,7 @@ func TestReconcileErrorHandling(t *testing.T) {
 			CreateOrUpdateProvider: &simpleCreateOrUpdater{},
 			platformType:           hyperv1.NonePlatform,
 			clusterSignerCA:        "foobar",
-			cpClient:               fake.NewClientBuilder().WithScheme(api.Scheme).WithObjects(cpObjects...).Build(),
+			cpClient:               fake.NewClientBuilder().WithScheme(api.Scheme).WithObjects(cpObjects...).WithStatusSubresource(&hyperv1.HostedControlPlane{}).Build(),
 			hcpName:                "foo",
 			hcpNamespace:           "bar",
 			releaseProvider:        &fakereleaseprovider.FakeReleaseProvider{},
@@ -2344,6 +2345,210 @@ func TestReconcileAuthOIDC(t *testing.T) {
 			for _, secret := range secretList.Items {
 				if !expectedSecrets.Has(secret.Name) {
 					t.Errorf("unexpected OIDC client secret copied: %s", secret.Name)
+				}
+			}
+		})
+	}
+}
+
+func newCondition(conditionType string, status metav1.ConditionStatus, reason, message string) *metav1.Condition {
+	return &metav1.Condition{
+		Type:    conditionType,
+		Reason:  reason,
+		Status:  status,
+		Message: message,
+	}
+}
+
+func Test_reconciler_reconcileControlPlaneDataPlaneConnectivityConditions(t *testing.T) {
+	newKonnectivityAgentPod := func(name string, phase corev1.PodPhase) corev1.Pod {
+		return corev1.Pod{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      name,
+				Namespace: "kube-system",
+				Labels:    map[string]string{"app": "konnectivity-agent"},
+			},
+			Status: corev1.PodStatus{
+				Phase: phase,
+			},
+		}
+	}
+	newRunningNode := func(name string) corev1.Node {
+		return corev1.Node{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: name,
+			},
+			Spec: corev1.NodeSpec{
+				Unschedulable: false,
+			},
+			Status: corev1.NodeStatus{
+				Conditions: []corev1.NodeCondition{
+					{Type: corev1.NodeReady,
+						Status: corev1.ConditionTrue,
+					},
+				},
+			},
+		}
+	}
+
+	tests := []struct {
+		name              string
+		hcp               *hyperv1.HostedControlPlane
+		wantErr           bool
+		expectedCondition *metav1.Condition
+		pods              []corev1.Pod
+		nodes             []corev1.Node
+		mockedGetPodLogs  func(context context.Context, clientet *clientset.Clientset, namespace, name, container string) ([]byte, error)
+	}{
+		{
+			name:    "no worker nodes Condition Unknown",
+			hcp:     fakeHCP(),
+			wantErr: false,
+			expectedCondition: newCondition(string(hyperv1.DataPlaneConnectionAvailable),
+				metav1.ConditionUnknown, hyperv1.DataPlaneConnectionNoWorkerNodesAvailableReason, "No worker nodes available"),
+			nodes: []corev1.Node{}, // no Nodes
+			mockedGetPodLogs: func(context context.Context,
+				clientet *clientset.Clientset,
+				namespace, name,
+				container string) ([]byte, error) {
+				return nil, nil
+			},
+		},
+		{
+			name:    "no konnectivity-agent PODs condition False",
+			hcp:     fakeHCP(),
+			wantErr: false,
+			expectedCondition: newCondition(string(hyperv1.DataPlaneConnectionAvailable),
+				metav1.ConditionFalse, hyperv1.DataPlaneConnectionNoKonnectivityAgentPodsNotFoundReason, "Couldn't find any konnectivity-agent running in data plane"),
+			nodes: []corev1.Node{newRunningNode("node1")},
+			pods:  []corev1.Pod{}, // no Pods
+			mockedGetPodLogs: func(context context.Context,
+				clientet *clientset.Clientset,
+				namespace, name,
+				container string) ([]byte, error) {
+				return nil, nil
+			},
+		},
+		{
+			name:    "only one pending POD condition False",
+			hcp:     fakeHCP(),
+			wantErr: false,
+			expectedCondition: newCondition(string(hyperv1.DataPlaneConnectionAvailable),
+				metav1.ConditionFalse, hyperv1.DataPlaneConnectionNoKonnectivityAgentPodsNotFoundReason, "Couldn't find any konnectivity-agent running in data plane"),
+			nodes: []corev1.Node{newRunningNode("node1")},
+			pods:  []corev1.Pod{newKonnectivityAgentPod("konnectivity-agent-rdax", corev1.PodPending)},
+			mockedGetPodLogs: func(context context.Context,
+				clientet *clientset.Clientset,
+				namespace, name,
+				container string) ([]byte, error) {
+				return nil, nil
+			},
+		},
+		{
+			name:    "one konnectivity-agent PODs running condition OK",
+			hcp:     fakeHCP(),
+			wantErr: false,
+			expectedCondition: newCondition(string(hyperv1.DataPlaneConnectionAvailable),
+				metav1.ConditionTrue, hyperv1.AsExpectedReason, hyperv1.AllIsWellMessage),
+			nodes: []corev1.Node{newRunningNode("node1")},
+			pods:  []corev1.Pod{newKonnectivityAgentPod("konnectivity-agent-rdax", corev1.PodRunning)},
+			mockedGetPodLogs: func(context context.Context,
+				clientet *clientset.Clientset,
+				namespace, name,
+				container string) ([]byte, error) {
+				return []byte("this is the log my friend"), nil
+			},
+		},
+		{
+			name:    "may konnectivity-agent PODs only one running condition OK",
+			hcp:     fakeHCP(),
+			wantErr: false,
+			expectedCondition: newCondition(string(hyperv1.DataPlaneConnectionAvailable),
+				metav1.ConditionTrue, hyperv1.AsExpectedReason, hyperv1.AllIsWellMessage),
+			nodes: []corev1.Node{newRunningNode("node1")},
+			pods: []corev1.Pod{newKonnectivityAgentPod("konnectivity-agent-rdax1", corev1.PodPending),
+				newKonnectivityAgentPod("konnectivity-agent-rdax2", corev1.PodPending),
+				newKonnectivityAgentPod("konnectivity-agent-rdax3", corev1.PodRunning),
+				newKonnectivityAgentPod("konnectivity-agent-rdax4", corev1.PodPending),
+				newKonnectivityAgentPod("konnectivity-agent-rdax5", corev1.PodPending)},
+			mockedGetPodLogs: func(context context.Context,
+				clientet *clientset.Clientset,
+				namespace, name,
+				container string) ([]byte, error) {
+				return []byte("this is the log my friend"), nil
+			},
+		},
+		{
+			name:    "one konnectivity-agent PODs running bad since error getting LOG",
+			hcp:     fakeHCP(),
+			wantErr: false,
+			expectedCondition: newCondition(string(hyperv1.DataPlaneConnectionAvailable),
+				metav1.ConditionFalse, hyperv1.DataPlaneConnectionNoKonnectivityAgentPodsNotFoundReason,
+				"failed to read konnectivity-agent logs from data plane"),
+			nodes: []corev1.Node{newRunningNode("node1")},
+			pods:  []corev1.Pod{newKonnectivityAgentPod("konnectivity-agent-rdax", corev1.PodRunning)},
+			mockedGetPodLogs: func(context context.Context,
+				clientet *clientset.Clientset,
+				namespace, name,
+				container string) ([]byte, error) {
+				return nil, fmt.Errorf("this time we fail")
+			},
+		},
+		{
+			name:    "one konnectivity-agent PODs running bad since no LOG", // unsure this is possible
+			hcp:     fakeHCP(),
+			wantErr: false,
+			expectedCondition: newCondition(string(hyperv1.DataPlaneConnectionAvailable),
+				metav1.ConditionFalse, hyperv1.DataPlaneConnectionNoKonnectivityAgentPodsNotFoundReason,
+				"failed to read konnectivity-agent logs from data plane"),
+			nodes: []corev1.Node{newRunningNode("node1")},
+			pods:  []corev1.Pod{newKonnectivityAgentPod("konnectivity-agent-rdax", corev1.PodRunning)},
+			mockedGetPodLogs: func(context context.Context,
+				clientet *clientset.Clientset,
+				namespace, name,
+				container string) ([]byte, error) {
+				return nil, nil
+			},
+		},
+	}
+	log := zapr.NewLogger(zaptest.NewLogger(t))
+	ctx := logr.NewContext(context.Background(), log)
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			var r reconciler
+			nodeList := &corev1.NodeList{
+				Items: tt.nodes,
+			}
+			podList := &corev1.PodList{
+				Items: tt.pods,
+			}
+			r.client = fake.NewClientBuilder().WithLists(nodeList).Build()
+			r.uncachedClient = fake.NewClientBuilder().WithLists(podList).Build()
+			r.cpClient = fake.NewClientBuilder().WithScheme(api.Scheme).WithObjects(tt.hcp).WithStatusSubresource(&hyperv1.HostedControlPlane{}).Build()
+			r.GetPodLogs = tt.mockedGetPodLogs
+
+			gotErr := r.reconcileControlPlaneDataPlaneConnectivityConditions(ctx, tt.hcp, log)
+			if gotErr != nil {
+				if !tt.wantErr {
+					t.Errorf("reconcileControlPlaneDataPlaneConnectivityConditions() failed: %v", gotErr)
+				}
+				return
+			}
+			if tt.wantErr {
+				t.Fatal("reconcileControlPlaneDataPlaneConnectivityConditions() succeeded unexpectedly")
+			}
+			if tt.expectedCondition != nil {
+				found := false
+				for _, c := range tt.hcp.Status.Conditions {
+					if tt.expectedCondition.Type == c.Type &&
+						tt.expectedCondition.Message == c.Message &&
+						tt.expectedCondition.Status == c.Status &&
+						tt.expectedCondition.Reason == c.Reason {
+						found = true
+					}
+				}
+				if !found {
+					t.Fatal("couldn't find expected condition")
 				}
 			}
 		})
--- a/docs/content/reference/api.md
+++ b/docs/content/reference/api.md
@@ -4901,6 +4901,15 @@ underlying cluster&rsquo;s ClusterVersion.</p>
 </tr><tr><td><p>&#34;RolloutComplete&#34;</p></td>
 <td><p>ControlPlaneComponentRolloutComplete indicates whether the ControlPlaneComponent has completed its rollout.</p>
 </td>
+</tr><tr><td><p>&#34;DataPlaneConnectionAvailable&#34;</p></td>
+<td><p>DataPlaneConnectionAvailable indicates whether the control plane has a successful
+network connection to the data plane components.
+<strong>True</strong> means the control plane can successfully reach the data plane nodes.
+<strong>False</strong> means there are network connection issues preventing the control plane from reaching the data plane.
+A failure here suggests potential issues such as: network policy restrictions,
+firewall rules, missing data plane nodes, or problems with infrastructure
+components like the konnectivity-agent workload.</p>
+</td>
 </tr><tr><td><p>&#34;EtcdAvailable&#34;</p></td>
 <td><p>EtcdAvailable bubbles up the same condition from HCP. It signals if etcd is available.
 A failure here often means a software bug or a non-stable cluster.</p>
--- a/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
+++ b/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
@@ -825,6 +825,7 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
 			hyperv1.ValidReleaseInfo,
 			hyperv1.ValidIDPConfiguration,
 			hyperv1.HostedClusterRestoredFromBackup,
+			hyperv1.DataPlaneConnectionAvailable,
 		}
 
 		for _, conditionType := range hcpConditions {
--- a/vendor/github.com/openshift/hypershift/api/hypershift/v1beta1/hostedcluster_conditions.go
+++ b/vendor/github.com/openshift/hypershift/api/hypershift/v1beta1/hostedcluster_conditions.go
@@ -195,6 +195,15 @@ const (
 	// This condition is used to track the status of the recovery process and to determine if the HostedCluster
 	// is ready to be used after restoration.
 	HostedClusterRestoredFromBackup ConditionType = "HostedClusterRestoredFromBackup"
+
+	// DataPlaneConnectionAvailable indicates whether the control plane has a successful
+	// network connection to the data plane components.
+	// **True** means the control plane can successfully reach the data plane nodes.
+	// **False** means there are network connection issues preventing the control plane from reaching the data plane.
+	// A failure here suggests potential issues such as: network policy restrictions,
+	// firewall rules, missing data plane nodes, or problems with infrastructure
+	// components like the konnectivity-agent workload.
+	DataPlaneConnectionAvailable ConditionType = "DataPlaneConnectionAvailable"
 )
 
 // Reasons.
@@ -250,7 +259,15 @@ const (
 
 	RecoveryFinishedReason = "RecoveryFinished"
 
+	ReconcileErrorReason = "ReconcileError"
+
 	CloudResourcesCleanupSkippedReason = "CloudResourcesCleanupSkipped"
+
+	DataPlaneConnectionNoKonnectivityAgentPodsNotFoundReason = "KonnectivityAgentPodsNotFound"
+
+	DataPlaneConnectionLogsAccessFailedReason = "LogsAccessFailed"
+
+	DataPlaneConnectionNoWorkerNodesAvailableReason = "NoWorkerNodesAvailable"
 )
 
 // Messages.


control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -1389,6 +1411,44 @@ func (r *reconciler) reconcileClusterVersion(ctx context.Context, hcp *hyperv1.H
...
+		if err == nil {
+			condition.Status = metav1.ConditionTrue
+			condition.Reason = "ControlPlaneToDataPlaneConnectivityOK"
+			condition.Message = "At least a konnectivity-agent is running on data plane"
+			break
+		}
+	}
+	meta.SetStatusCondition(&hcp.Status.Conditions, *condition)
+	return nil
+
```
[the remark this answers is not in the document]
sdminonne: Thanks @coderabbitai I know! That's why it's a draft :)

api/hypershift/v1beta1/hostedcontrolplane_conditions.go
```diff
@@ -0,0 +1,43 @@
...
+	// HostedControlPlaneAvailable indicates whether the HostedControlPlane has a healthy control plane.
+	// When True, the control plane is operational and serving requests.
+	// When False, the control plane is experiencing issues that prevent normal operation.
+	HostedControlPlaneAvailable ConditionType = "Available"
+
+	// HostedControlPlaneDegraded indicates whether the HostedControlPlane is encountering
+	// an error that may require user intervention to resolve.
+	// When True, the control plane is degraded and experiencing issues.
+	// When False, the control plane is operating normally without degradation.
+	HostedControlPlaneDegraded ConditionType = "Degraded"
```
jparrill: Don't we already have those verbs already in `controlplanecomponent_types.go`

api/hypershift/v1beta1/hostedcontrolplane_conditions.go
```diff
@@ -0,0 +1,43 @@
+package v1beta1
+
+// "Condition values may change back and forth, but some condition transitions may be monotonic, depending on the resource and condition type.
+// However, conditions are observations and not, themselves, state machines."
+// https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties
```
jparrill: Are those comments necessary?

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -1389,6 +1411,44 @@ func (r *reconciler) reconcileClusterVersion(ctx context.Context, hcp *hyperv1.H
 	return nil
 }
 
+func (r *reconciler) updateControlPlaneDatapPlaneConnectivityConditions(ctx context.Context, hcp *hyperv1.HostedControlPlane) error {
+	var podList corev1.PodList
+	if err := r.client.List(ctx, &podList,
```
jparrill: Is that the right client to do the request? We need to be careful with how many request we do against the guest cluster with non-watched objects.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -1389,6 +1411,44 @@ func (r *reconciler) reconcileClusterVersion(ctx context.Context, hcp *hyperv1.H
...
+				return err
+			}
+			if len(data) > 0 { // found something
+				return nil
+			}
+			return fmt.Errorf("no logs for %s/%s container %s", pod.Namespace, pod.Name, "konnectivity-agen")
+		}()
+		if err == nil {
+			condition.Status = metav1.ConditionTrue
+			condition.Reason = "ControlPlaneToDataPlaneConnectivityOK"
```
jparrill: Create a `const` with the new `Reasons`.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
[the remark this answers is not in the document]
jparrill: Agree with the feedback :)

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources_test.go
```diff
@@ -313,16 +313,16 @@ func TestReconcileOLM(t *testing.T) {
 	}
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
-			g := NewWithT(t)
+			g := gomega.NewWithT(t)
```
jparrill: why? I don't think these changes are necessary at all

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources_test.go
```diff
@@ -2331,13 +2331,13 @@ func TestReconcileAuthOIDC(t *testing.T) {
...
 					}, caConfigMap)
-					g.Expect(apierrors.IsNotFound(err)).To(BeTrue())
+					g.Expect(apierrors.IsNotFound(err)).To(gomega.BeTrue())
 				}
 			}
 			// Verify that no unexpected client secrets were copied
 			secretList := &corev1.SecretList{}
 			err = hcClient.List(ctx, secretList, client.InNamespace(ConfigNamespace))
-			g.Expect(err).ToNot(HaveOccurred())
+			g.Expect(err).ToNot(gomega.HaveOccurred())
```
jparrill, line 2342: I think all the Gomega renames are not necessary (until here)

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources_test.go
```diff
@@ -2349,3 +2349,109 @@ func TestReconcileAuthOIDC(t *testing.T) {
...
+func initCondition(conditionType string, status metav1.ConditionStatus, reason, message string) *metav1.Condition {
+	return &metav1.Condition{
+		Type:    conditionType,
+		Reason:  reason,
+		Status:  status,
+		Message: message,
+	}
+}
+
+func initPod() corev1.Pod {
```
jparrill: The signature of the function does not match the expectations. I would rename it with a most appropriated name

support/konnectivityproxy/proxy_dialer.go
```diff
@@ -17,7 +17,7 @@ func newHTTPDialer(proxyURL *url.URL) proxy.Dialer {
 }
 
 // Everything below is a copied from https://github.com/fasthttp/websocket/blob/2f8e79d2aac1e8e5a06518870e872b15608cea90/proxy.go
-// as the golang.org/x/net/proxy package only supports socks5 proxies, but does allow registering additional protocols.
+// as the golang.org/x/xnet/proxy package only supports socks5 proxies, but does allow registering additional protocols.
```
jparrill: this pkg does not exists

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
sdminonne: I don't know, to be honest. This is the main reason this PR has been sent even if not 100% ready.
 Is the `resources` reconciler the right place for this?

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources_test.go
sdminonne: I can get rid of all of them and restore back the dot import but which now start to be deprecated (and never seen as a good practice).

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources_test.go
sdminonne: OK, I'll try to do better :)

support/konnectivityproxy/proxy_dialer.go
sdminonne: Unsure who touched this maybe the first PR (claude made) to introduce APIs... I'll amend it. Thanks good catch!

sdminonne:
@jparrill many many thanks!

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources_test.go
sdminonne: Restored

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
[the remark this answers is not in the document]
sdminonne: @CodeRabbitAI done! PTAL

api/hypershift/v1beta1/hostedcontrolplane_conditions.go
sdminonne: This is specifically about connectivity

sdminonne:
@coderabbitai review

api/hypershift/v1beta1/hostedcontrolplane_conditions.go
```diff
@@ -0,0 +1,43 @@
...
+	// When True, the CVO is scaled down (typically during maintenance operations).
+	// When False, the CVO is running normally.
+	CVOScaledDown ConditionType = "CVOScaledDown"
+
+	// DataPlaneToControlPlaneConnectivityHealthy indicates whether the data plane can successfully
+	// reach the control plane components.
+	// When True, data plane nodes have healthy connectivity to control plane services.
+	// When False, there are network connectivity issues preventing data plane from reaching the control plane.
+	// A failure here may indicate network policy issues, firewall rules, or infrastructure problems.
+	DataPlaneToControlPlaneConnectivityHealthy ConditionType = "DataPlaneToControlPlaneConnectivityHealthy"
```
muraee: what is this condition for? it's not set anywhere

sdminonne: Originally the idea was to have a condition ControlPlane->DataPlane and a condition DataPlane->ControlPlane but if GetLogs works both have to work (otherwise log won't be streamed)

sdminonne: In a second step we'll refine running a sidecar container in dataPlane updating the condition. That's why I didn't remove it

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -181,9 +191,15 @@ func Setup(ctx context.Context, opts *operator.HostedClusterConfigOperatorConfig
 		return fmt.Errorf("failed to create kubevirt infra uncached client: %w", err)
 	}
 
+	clientset, err := clientset.NewForConfig(opts.Config)
```
muraee: this is not the correct config. This is for the mgmt cluster
use `opts.Manager.GetConfig()`  to get the guest cluster config.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -1389,6 +1411,59 @@ func (r *reconciler) reconcileClusterVersion(ctx context.Context, hcp *hyperv1.H
...
+			condition.Reason = ControlPlaneToDataPlaneReasonConnectivityOK
+			condition.Message = "At least a konnectivity-agent is running on data plane"
+			break
+		}
+	}
+
+	original := hcp.DeepCopy()
+	meta.SetStatusCondition(&hcp.Status.Conditions, *condition)
+
+	if equality.Semantic.DeepEqual(original.Status.Conditions, hcp.Status.Conditions) {
```
muraee: `meta.SetStatusCondition()` returns a bool indicating if the condition changed or not. You should just check that instead of using `DeepEqual`

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -1389,6 +1411,59 @@ func (r *reconciler) reconcileClusterVersion(ctx context.Context, hcp *hyperv1.H
...
+		return nil // No status change; avoid unnecessary API call.
+	}
+	if err := r.cpClient.Status().Update(ctx, hcp); err != nil {
```
muraee: please use `Patch` instead of `Update`
```
r.cpClient.Status().Patch(ctx, hcp, client.MergeFrom(originalHCP))
```

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -1389,6 +1411,59 @@ func (r *reconciler) reconcileClusterVersion(ctx context.Context, hcp *hyperv1.H
 	return nil
 }
 
+const (
+	ControlPlaneToDataplaneReasonNoKonnectivityAgentPodsFound = "NoKonnectivityAgentPodsFound"
+	ControlPlaneToDataPlaneReasonConnectivityOK               = "ControlPlaneToDataPlaneConnectivityOK"
```
muraee:
```suggestion
	ControlPlaneToDataplaneNoKonnectivityAgentPodsFoundReason = "NoKonnectivityAgentPodsFound"
	ControlPlaneToDataPlaneConnectivityOKReason             = "ControlPlaneToDataPlaneConnectivityOK"
```
Also, those consts should be defined in the same file as the condition, see https://github.com/openshift/hypershift/blob/main/api/hypershift/v1beta1/hostedcluster_conditions.go#L200

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -1389,6 +1411,59 @@ func (r *reconciler) reconcileClusterVersion(ctx context.Context, hcp *hyperv1.H
...
+)
+
+func (r *reconciler) updateControlPlaneDatapPlaneConnectivityConditions(ctx context.Context, hcp *hyperv1.HostedControlPlane) error {
+	var podList corev1.PodList
+	if err := r.client.List(ctx, &podList,
+		&client.ListOptions{LabelSelector: labels.SelectorFromValidatedSet(labels.Set{"app": "konnectivity-agent"}),
+			FieldSelector: fields.SelectorFromSet(fields.Set{"status.phase": "Running"}),
```
muraee: this field `status.phase`  needs to be indexed for this to work, I don't believe we are doing that.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
sdminonne: @muraee the konnectivity-agents runs on HostedCluster not on Management Cluster

sdminonne: Ok taking back after our call

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
sdminonne: I removed the fieldSelector and I added a manul check in the Post list iteration

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
sdminonne: Done thanks!

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
sdminonne: Done thanks

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
jparrill: I assume so, yes. The thing is, in the HCCO we have multiple clients:
- The ControlPlane one
- The DataPlane with cached resources from Informers (cheap)
- The DataPlane with uncached resources (expensive)
- The controller specifics for separated flows (not used in resources.go).

You'll need to decide which one use in which case to avoid extra usage of the KAS :)

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -1389,6 +1415,58 @@ func (r *reconciler) reconcileClusterVersion(ctx context.Context, hcp *hyperv1.H
 	return nil
 }
 
+func (r *reconciler) updateControlPlaneDatapPlaneConnectivityConditions(ctx context.Context, hcp *hyperv1.HostedControlPlane) error {
```
mgencur:
```suggestion
func (r *reconciler) updateControlPlaneDataPlaneConnectivityConditions(ctx context.Context, hcp *hyperv1.HostedControlPlane) error {
```

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -1389,6 +1415,58 @@ func (r *reconciler) reconcileClusterVersion(ctx context.Context, hcp *hyperv1.H
...
+			if err != nil {
+				condition.Status = metav1.ConditionFalse
+				condition.Reason = hyperv1.ControlPlaneToDataPlaneReasonLogAccessFailed
+				condition.Message = "konnectivity-agent PODs are running but no Logs accessible"
+				return err
+			}
+			if len(data) > 0 { // found something
+				return nil
+			}
+			return fmt.Errorf("no logs for %s/%s container %s", pod.Namespace, pod.Name, "konnectivity-agen")
```
mgencur:
```suggestion
			return fmt.Errorf("no logs for %s/%s container %s", pod.Namespace, pod.Name, "konnectivity-agent")
```

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -1389,6 +1415,58 @@ func (r *reconciler) reconcileClusterVersion(ctx context.Context, hcp *hyperv1.H
...
+		}()
+		if err == nil {
+			condition.Status = metav1.ConditionTrue
+			condition.Reason = hyperv1.ControlPlaneToDataPlaneReasonConnectivityOK
+			condition.Message = "At least a konnectivity-agent is running on data plane and logs are accessible"
```
mgencur: Just curious what is meant by this message? It it that at least "one" konnectivity-agent is running? Or, what are the other things that should be running. Just thinking if this error message is clear enough. I don't know.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources_test.go
```diff
@@ -2349,3 +2347,116 @@ func TestReconcileAuthOIDC(t *testing.T) {
...
+			Namespace: "kube-system",
+			Labels:    map[string]string{"app": "konnectivity-agent"},
+		},
+		Status: corev1.PodStatus{
+			Phase: corev1.PodRunning,
+		},
+	}
+}
+
+func Test_reconciler_updateControlPlaneDatapPlaneConnectivityConditions(t *testing.T) {
```
mgencur:
```suggestion
func Test_reconciler_updateControlPlaneDataPlaneConnectivityConditions(t *testing.T) {
```

api/hypershift/v1beta1/hostedcluster_conditions.go
```diff
@@ -251,6 +251,12 @@ const (
 	RecoveryFinishedReason = "RecoveryFinished"
 
 	CloudResourcesCleanupSkippedReason = "CloudResourcesCleanupSkipped"
+
+	ControlPlaneToDataplaneReasonNoKonnectivityAgentPodsFound = "NoKonnectivityAgentPodsFound"
```
mgencur:
```suggestion
	ControlPlaneToDataPlaneReasonNoKonnectivityAgentPodsFound = "NoKonnectivityAgentPodsFound"
```
(this would also need update in other places where this is used.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
sdminonne: One konnectivity agent is enough to have connection between control plane and data plane.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources_test.go
sdminonne: Yeah in a follow-up

api/hypershift/v1beta1/hostedcluster_conditions.go
sdminonne: Same, in a follow-up

api/hypershift/v1beta1/hostedcontrolplane_conditions.go
muraee: let's remove it please, if it's not being introduced in this PR

api/hypershift/v1beta1/hostedcontrolplane_conditions.go
muraee: Please move those to hostedcluster_conditions.go.
There are no condition specific for HCP, all of them are surfaced at the HostedCluster. no need for this file

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -1389,6 +1415,58 @@ func (r *reconciler) reconcileClusterVersion(ctx context.Context, hcp *hyperv1.H
 	return nil
 }
 
+func (r *reconciler) updateControlPlaneDatapPlaneConnectivityConditions(ctx context.Context, hcp *hyperv1.HostedControlPlane) error {
+	var podList corev1.PodList
+	if err := r.client.List(ctx, &podList,
+		&client.ListOptions{LabelSelector: labels.SelectorFromValidatedSet(labels.Set{"app": "konnectivity-agent"}),
+			Namespace: "kube-system"}); err != nil {
```
muraee:
```suggestion
		client.MatchingLabels{
			"app": "konnectivity-agent",
		}, client.InNamespace("kube-system")); err != nil {
```

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -1389,6 +1415,58 @@ func (r *reconciler) reconcileClusterVersion(ctx context.Context, hcp *hyperv1.H
...
+		if err == nil {
+			condition.Status = metav1.ConditionTrue
+			condition.Reason = hyperv1.ControlPlaneToDataPlaneReasonConnectivityOK
+			condition.Message = "At least a konnectivity-agent is running on data plane and logs are accessible"
+			break
+		}
+	}
+
+	originalHCP := hcp.DeepCopy()
+	if set := meta.SetStatusCondition(&hcp.Status.Conditions, *condition); !set {
```
muraee:
```suggestion
	if !meta.SetStatusCondition(&hcp.Status.Conditions, *condition) {
```
no need for the scoped var.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources_test.go
```diff
@@ -2349,3 +2347,116 @@ func TestReconcileAuthOIDC(t *testing.T) {
 		})
 	}
 }
+
+func initCondition(conditionType string, status metav1.ConditionStatus, reason, message string) *metav1.Condition {
```
muraee: `NewCondition` or `CreateCondition` might be better names for this.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources_test.go
```diff
@@ -2349,3 +2347,116 @@ func TestReconcileAuthOIDC(t *testing.T) {
...
+	return &metav1.Condition{
+		Type:    conditionType,
+		Reason:  reason,
+		Status:  status,
+		Message: message,
+	}
+}
+
+func initKonnectivyAgentRunningPod() corev1.Pod {
```
muraee: same for the nameing. Also, this function definition can be moved inside `Test_reconciler_updateControlPlaneDatapPlaneConnectivityConditions` and defined as a variable that you can use.

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
muraee, line 290: why this change? this doesn't seem relevant to this PR

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
muraee: getting the logs is an implementation detail, users don't need to know about it.
A simple `hyperv1.AllIsWellMessage` should be enough for this case.

sdminonne:
~~Unsure why `e2e-aks` is failing now. Yesterday it was passing and I simply modified a commit to make verify passing ~~ :disappointed: 

Taking back I didn't see the @muraee comment before this

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
sdminonne: I don't know

sdminonne: But sure it's not relevant to this PR

mgencur:
I have tested this on my dev cluster and getting these statuses in HC and HCP:
HCP:
```
  - lastTransitionTime: "2025-11-21T14:30:36Z"
    message: Couldn't find an konnectivity-agent running in data plane
    reason: NoKonnectivityAgentPodsFound
    status: "False"
    type: ControlPlaneToDataPlaneConnectivityHealthy
```
    
HC:
```
  - lastTransitionTime: "2025-11-21T14:27:45Z"
    message: Condition not found in the HCP
    observedGeneration: 3
    reason: StatusUnknown
    status: Unknown
    type: DataPlaneToControlPlaneConnectivityHealthy
  - lastTransitionTime: "2025-11-21T14:30:36Z"
    message: Couldn't find an konnectivity-agent running in data plane
    observedGeneration: 3
    reason: NoKonnectivityAgentPodsFound
    status: "False"
    type: ControlPlaneToDataPlaneConnectivityHealthy
 ```
The konnectivity pods are there on guest cluster:
```
 ᐅ oc get pods -n kube-system                  
NAME                                                READY   STATUS    RESTARTS   AGE
global-pull-secret-syncer-hfn99                     1/1     Running   0          47m
global-pull-secret-syncer-ssr82                     1/1     Running   0          47m
global-pull-secret-syncer-zvpzb                     1/1     Running   0          16m
konnectivity-agent-2p5xg                            1/1     Running   0          47m
konnectivity-agent-l7js2                            1/1     Running   0          47m
konnectivity-agent-v8r4z                            1/1     Running   0          47m
kube-apiserver-proxy-ip-10-0-131-0.ec2.internal     1/1     Running   0          47m
kube-apiserver-proxy-ip-10-0-138-153.ec2.internal   1/1     Running   0          47m
kube-apiserver-proxy-ip-10-0-142-80.ec2.internal    1/1     Running   0          47m
```

My latest commit is "d8041d649 - Salvatore Dario Minonne, 22 hours ago : fix(CNTRPLANE-1956): address Mulham feedback". Not sure there were any changes recently that could fix it.

**Comment by sdminonne** [comment]
Unxpected if you don't have anything but only default setting

**Comment by sdminonne** [comment]
Looking at it

**Comment by sdminonne** [comment]
@coderabbit review

File: control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go [hunk 25]
```diff
@@ -1389,6 +1415,62 @@ func (r *reconciler) reconcileClusterVersion(ctx context.Context, hcp *hyperv1.H
...
+	if totalNodes > 0 {
+		condition.Status = metav1.ConditionFalse
+		condition.Reason = hyperv1.ControlPlaneToDataPlaneReasonNoKonnectivityAgentPodsFound
+		condition.Message = "Couldn't find an konnectivity-agent running in data plane"
+		var podList corev1.PodList
+		if err := r.client.List(ctx, &podList,
+			client.MatchingLabels{"app": "konnectivity-agent"}, client.InNamespace("kube-system")); err != nil {
+			log.Error(err, "unable to list konnectivity-agent PODs in kube-system namespace")
+		}
+		for _, pod := range podList.Items {
```
mgencur: Thinking about it a little more. Should we verify that "all" konnectivity-agent Pods are running? It's running a DeamonSet and it's required on each node. The connectivity is not fully established until the konnectivity-pod is running on each Node. Say, we successfully start a konnectivity-agent Pod on one node but the other ones keep failing. Then we'll get success and the message "At least a konnectivity-agent is running on data plane and logs are accessible" even though konnectivity on other nodes is not yet available (and it might not be ever).

sdminonne: @mgencur one konnectivity-agent POD running grant network connectivty between control-plane and data-plane

api/hypershift/v1beta1/hostedcontrolplane_conditions.go
sdminonne: going to do. Thanks

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
sdminonne: Going to do

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources_test.go
sdminonne: modified the name

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
sdminonne: Done

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
sdminonne: Done

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
sdminonne: I've added nodes to cache and using uncached to get PODs.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -222,6 +243,7 @@ func Setup(ctx context.Context, opts *operator.HostedClusterConfigOperatorConfig
 		&corev1.Secret{},
 		&corev1.Service{},
 		&corev1.Endpoints{},
+		&corev1.Node{},
```
muraee: why do we need to watch nodes?

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -1389,6 +1416,65 @@ func (r *reconciler) reconcileClusterVersion(ctx context.Context, hcp *hyperv1.H
...
+		}
+	}
+
+	originalHCP := hcp.DeepCopy()
+	if !meta.SetStatusCondition(&hcp.Status.Conditions, *condition) {
+		return nil // No status change; avoid unnecessary API call.
+	}
+
+	if err := r.cpClient.Status().Patch(ctx, hcp, client.MergeFrom(originalHCP)); err != nil {
+		log.Error(err, "failed to update HostedControlPlane status with "+condition.Type)
```
muraee: no need for log when we are returning the error

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -1389,6 +1416,65 @@ func (r *reconciler) reconcileClusterVersion(ctx context.Context, hcp *hyperv1.H
...
+		Status:  metav1.ConditionUnknown,
+		Reason:  hyperv1.ControlPlaneToDataPlaneReasonNoWorkerNodesAvailable,
+		Message: "No worker nodes available",
+	}
+	totalNodes, err := util.CountAvailableNodes(ctx, r.client)
+	if err != nil {
+		condition.Message = "Unable to count worker nodes: " + err.Error()
+		log.Error(err, "Unable to count worker nodes")
+	}
+	if totalNodes > 0 {
```
muraee: the way the condition is set before the code path is confusing and not clear. We should set the condition inside the code path and return early. e.g.
```go
totalNodes, err := util.CountAvailableNodes(ctx, r.client)
if err != nil {
      condition.Message = "Failed to count worker nodes: " + err.Error()
      condition.Status =  metav1.ConditionUnknown
      return patchHCPWithCondition(hcp, condition)
}

if totalNodes < 0 {
      condition.Message = "No worker nodes"
      condition.Status =  metav1.ConditionUnknown
      return patchHCPWithCondition(hcp, condition)
}

// same for the rest of the code
...

```

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -1389,6 +1416,65 @@ func (r *reconciler) reconcileClusterVersion(ctx context.Context, hcp *hyperv1.H
...
+		for _, pod := range podList.Items {
+			if pod.Status.Phase != corev1.PodRunning {
+				continue
+			}
+			condition.Status = metav1.ConditionFalse
+			condition.Reason = hyperv1.ControlPlaneToDataPlaneReasonLogAccessFailed
+			data, err := r.GetPodLogs(ctx, r.clientSet, pod.Namespace, pod.Name, "konnectivity-agent")
+			if err == nil && len(data) > 0 { // found something
+				condition.Status = metav1.ConditionTrue
+				condition.Reason = hyperv1.ControlPlaneToDataPlaneReasonConnectivityOK
```
muraee: please use `hyperv1.AsExpectedReason` when the condition is true

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
sdminonne: We need to list nodes and at the moment we're using cached client. If I'm not wrong we need to watch them

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
sdminonne: OK.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
sdminonne: OK

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
muraee: this won't affect the cache. This will trigger a reconcile loop on every change to the node resource.

mgencur:
I have tested the negative scenario by using `--annotations=hypershift.openshift.io/konnectivity-agent-image=quay.io/mgencur/non-existent:latest`:
Initial condition was this when starting the cluster:
```
~/hypershift/clusters/mgmt ᐅ oc get hc mgencur-hc1-2 -n clusters-2 -ojsonpath='{.status.conditions[?(@.type=="ControlPlaneToDataPlaneConnectivityHealthy")]}' | jq .
{
  "lastTransitionTime": "2025-11-25T11:33:34Z",
  "message": "No worker nodes available",
  "observedGeneration": 3,
  "reason": "NoWorkerNodesAvailable",
  "status": "Unknown",
  "type": "ControlPlaneToDataPlaneConnectivityHealthy"
}
```
Later it changed to:
```
ᐅ oc get hc mgencur-hc1-2 -n clusters-2 -ojsonpath='{.status.conditions[?(@.type=="ControlPlaneToDataPlaneConnectivityHealthy")]}' | jq .
{
  "lastTransitionTime": "2025-11-25T11:44:45Z",
  "message": "Couldn't find any konnectivity-agent running in data plane",
  "observedGeneration": 3,
  "reason": "NoKonnectivityAgentPodsFound",
  "status": "False",
  "type": "ControlPlaneToDataPlaneConnectivityHealthy"
}
```
This is from data plane:
```
ᐅ oc get pods -l app=konnectivity-agent -n kube-system
NAME                       READY   STATUS             RESTARTS   AGE
konnectivity-agent-ghw8q   0/1     ImagePullBackOff   0          2m51s
konnectivity-agent-rc7tt   0/1     ImagePullBackOff   0          3m44s
konnectivity-agent-xrqpk   0/1     ImagePullBackOff   0          3m45s
```
I think this looks alright. Maybe just the condition could be named `NoRunningKonnectivityAgentPodsFound` instead of `NoKonnectivityAgentPodsFound`. The Pods are there but not running. That's probably not so important, though.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
sdminonne: Ok. It's there 'cause I had the cached client without it I couldn't list nodes properly.  I'm going to get rid of it, i

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
sdminonne: @muraee done, PTAL

mgencur:
Re-tested with the last code from https://github.com/openshift/hypershift/pull/7260/commits/3b0b583043078c1bc6a729b0b8b04feada03e2b5:

For the happy scenario, the HC status goes from:
```{
  "lastTransitionTime": "2025-11-25T15:42:40Z",
  "message": "No worker nodes available",
  "observedGeneration": 3,
  "reason": "NoWorkerNodesAvailable",
  "status": "Unknown",
  "type": "ControlPlaneToDataPlaneConnectivityHealthy"
}
```
to
```
{
  "lastTransitionTime": "2025-11-25T15:53:48Z",
  "message": "All is well",
  "observedGeneration": 3,
  "reason": "AsExpected",
  "status": "True",
  "type": "ControlPlaneToDataPlaneConnectivityHealthy"
}
```

For the failure scenarios when konnectiviry-agent Pods don't run, it goes from:
```
{
  "lastTransitionTime": "2025-11-25T16:36:52Z",
  "message": "No worker nodes available",
  "observedGeneration": 3,
  "reason": "NoWorkerNodesAvailable",
  "status": "Unknown",
  "type": "ControlPlaneToDataPlaneConnectivityHealthy"
}

```
to
```
{
  "lastTransitionTime": "2025-11-25T16:47:52Z",
  "message": "Couldn't find any konnectivity-agent running in data plane",
  "observedGeneration": 3,
  "reason": "NoKonnectivityAgentPodsFound",
  "status": "False",
  "type": "ControlPlaneToDataPlaneConnectivityHealthy"
}

```

sdminonne:
@mgencur the first modification is what @muraee requested, and the second one looks more accurate. NoPod are correctly reported as NoPod.

api/hypershift/v1beta1/hostedcontrolplane_conditions.go
sdminonne: Done

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
sdminonne: Done

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -1389,6 +1415,75 @@ func (r *reconciler) reconcileClusterVersion(ctx context.Context, hcp *hyperv1.H
...
+			// we set temporary the message but we keep looping
+			condition.Message = "konnectivity-agent PODs are running but no logs accessible"
+			continue
+		}
+		// found something
+		condition.Status = metav1.ConditionTrue
+		condition.Reason = hyperv1.AsExpectedReason
+		condition.Message = hyperv1.AllIsWellMessage
+		return patchHCPWithCondition(hcp, condition)
+	}
```
muraee:
```go
    logsFound := false
	for _, pod := range podList.Items {
		if pod.Status.Phase != corev1.PodRunning {
			continue
		}
		data, err := r.GetPodLogs(ctx, r.clientSet, pod.Namespace, pod.Name, "konnectivity-agent")
		if err != nil  {
			log.Error(err,
				fmt.Sprintf("failed to get logs for konnectivity-agent pod %s/%s", pod.Namespace, pod.Name))
			continue
		}
		if len(data) > 0 {
		    logsFound = true
			break
		}
	}
	
    if !logsFound {
       condition.Reason = hyperv1.ControlPlaneToDataPlaneReasonNoKonnectivityAgentPodsFound
       condition.Message = "failed to read konnectivity-agent logs from data plane"
    } else {
       condition.Status = metav1.ConditionTrue
	   condition.Reason = hyperv1.AsExpectedReason
	   condition.Message = hyperv1.AllIsWellMessage
    }

    return patchHCPWithCondition(hcp, condition)
  }
```

I think something like this is more clear.
we don't need to include the error in the condition message as we are logging it.

api/hypershift/v1beta1/hostedcluster_conditions.go
```diff
@@ -195,6 +195,13 @@ const (
 	// This condition is used to track the status of the recovery process and to determine if the HostedCluster
 	// is ready to be used after restoration.
 	HostedClusterRestoredFromBackup ConditionType = "HostedClusterRestoredFromBackup"
+
+	// ControlPlaneToDataPlaneConnectivityHealthy indicates whether the control plane can successfully
+	// reach the data plane components.
+	// When True, control plane has healthy connectivity to data plane nodes.
+	// When False, there are network connectivity issues preventing control plane from reaching the data plane.
+	// A failure here may indicate network policy issues, firewall rules, or infrastructure problems.
+	ControlPlaneToDataPlaneConnectivityHealthy ConditionType = "ControlPlaneToDataPlaneConnectivityHealthy"
```
devguyio: I know naming is hard, and subjective, but this condition name is VERY long. I typically refrain from adding strong opinions about names, except for user-facing primitives like API, CLI args etc.

In this case I'd suggest to have a more concise and simple name for this condition like `DataPlaneConnectionHealthy`:

1) This implies that it's "ControlPlaneToDataPlane" but much shorter.
2) Connection is a simpler term than "Connectivity".

api/hypershift/v1beta1/hostedcluster_conditions.go
```diff
@@ -251,6 +258,12 @@ const (
 	RecoveryFinishedReason = "RecoveryFinished"
 
 	CloudResourcesCleanupSkippedReason = "CloudResourcesCleanupSkipped"
+
+	ControlPlaneToDataPlaneReasonNoKonnectivityAgentPodsFound = "NoKonnectivityAgentPodsFound"
```
devguyio:
```suggestion
	KonnectivityAgentPodsNotFoundReason = "KonnectivityAgentPodsNotFound"
```

1) When possible, not tying the reason to the condition name makes in more concise and reusable.
2) reason consts typically end with `Reason` suffix
3) `NotFound` is the typical way to reflect not finding something.

api/hypershift/v1beta1/hostedcluster_conditions.go
```diff
@@ -251,6 +258,12 @@ const (
...
+
+	ControlPlaneToDataPlaneReasonLogAccessFailed = "ControlPlaneToDataPlaneLogAccessFailed"
```
devguyio:
```suggestion
	KonnectivityPodsLogAccessFailedReason = "KonnectivityPodsLogAccessFailed"
```

api/hypershift/v1beta1/hostedcluster_conditions.go
```diff
@@ -251,6 +258,12 @@ const (
...
+
+	ControlPlaneToDataPlaneReasonNoWorkerNodesAvailable = "NoWorkerNodesAvailable"
```
devguyio:
```suggestion
	NoWorkerNodesAvailableReason = "NoWorkerNodesAvailable"
```

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -1389,6 +1415,75 @@ func (r *reconciler) reconcileClusterVersion(ctx context.Context, hcp *hyperv1.H
...
+
+	condition := &metav1.Condition{
+		Type:   string(hyperv1.ControlPlaneToDataPlaneConnectivityHealthy),
+		Status: metav1.ConditionFalse, // False by default
+	}
+	totalNodes, err := util.CountAvailableNodes(ctx, r.client)
+	if err != nil {
+		condition.Status = metav1.ConditionUnknown
+		condition.Reason = hyperv1.ControlPlaneToDataPlaneReasonNoWorkerNodesAvailable
```
devguyio: I think having `NoWorkerNodesAvailable` as the reason might be a little misleading. I'd say some generic
`ReconcileError` reason

api/hypershift/v1beta1/hostedcluster_conditions.go
sdminonne: The point that this is only half of the job.
Ths represents ControlPlane -> DataPlane connection

When I'll have more time I'll add another condition to represet DataPlane -> ControlPlane connection.

api/hypershift/v1beta1/hostedcluster_conditions.go
sdminonne: The reason is expressed like that 'cause the connection issue  may come from multiple problems. 
No nodes (in that case is Unknown)
No Pods or error returning Logs. the idea was to give customer some info about the source of problem

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
sdminonne: Same, I'm just trying to give info to customer. 
ReoncileError is too generic and doesn't help the customers. 

Our internal customers (ROSA-HCP sreps) clearly asked to save some time in konnectivity agent troubleshooting (see Testing the Konnectivity server with curl at page https://hypershift-docs.netlify.app/reference/konnectivity/). They perform it quite often :disappointed:

sdminonne:
@devguyio I'm totally OK to rename conditions and messages but there're some requirement we need to satisfy, thanks for your help! Really appreciated

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
devguyio: I think we're on the same page about the use-case and what we're trying to achieve. 

>  ReoncileError is too generic and doesn't help the customers.

Please note that I'm only referring to reason you're on line 1438. What you have here is:
1. You're trying to list the total nodes
2. You received an error. 
3. You're setting the condition status to `ConditionUnknown` , and the reason `NoWorkerNodesAvaialable`. 

So this would show as 
```
  - lastTransitionTime: "2025-11-14T22:45:08Z"
    message: "Couldn't list konnectivity-agent PODs in kube-system namespace: <error here>"
    reason: "NoWorkerNodesAvailable"
    status: "Unknown"
    type: "ControlPlaneToDataPlaneConnectivityHealthy
```

Which shows that the status is unknown, because there are no worker nodes available. While the reality is that the status is unknown because there was an error getting the worker nodes.

So I'm suggesting that we change the reason in this case to something like `ErrorEncountered` or `ReconcileError`.

That would look like

```
  - lastTransitionTime: "2025-11-14T22:45:08Z"
    message: "Couldn't list konnectivity-agent PODs in kube-system namespace: <error here>"
    reason: "ErrorEncountered"
    status: "Unknown"
    type: "ControlPlaneToDataPlaneConnectivityHealthy
```

Don't you think this helps customers more?

devguyio:
> @devguyio I'm totally OK to rename conditions and messages but there're some requirement we need to satisfy, thanks for your help! Really appreciated

Let's chat about this over a call tomorrow.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
sdminonne: OK. Going to do.

sdminonne:
@devguyio PTAL

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
sdminonne: Done

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
sdminonne: Done this as well. Added an if to make unit test passing

api/hypershift/v1beta1/hostedcluster_conditions.go
sdminonne: Postfixing Reason in cost name done.

api/hypershift/v1beta1/hostedcluster_conditions.go
sdminonne: Preserving ControlPlaneToDataPlane naming here

sdminonne:
@devguyio PTAL

api/hypershift/v1beta1/hostedcluster_conditions.go
```diff
@@ -250,7 +257,15 @@ const (
 
 	RecoveryFinishedReason = "RecoveryFinished"
 
+	ReconcileErrorReason = "ReconcileError"
+
 	CloudResourcesCleanupSkippedReason = "CloudResourcesCleanupSkipped"
+
+	ControlPlaneToDataPlaneNoKonnectivityAgentPodsFoundReason = "NoKonnectivityAgentPodsFound"
```
devguyio: Nit: One last comment that wasn't addressed aside from renaming is this: "KonnectivityAgentPodsNotFound", to match the existing convention.

api/hypershift/v1beta1/hostedcluster_conditions.go
```diff
@@ -195,6 +195,14 @@ const (
 	// This condition is used to track the status of the recovery process and to determine if the HostedCluster
 	// is ready to be used after restoration.
 	HostedClusterRestoredFromBackup ConditionType = "HostedClusterRestoredFromBackup"
+
+	// DataPlaneConnectionAvailable indicates whether the control plane can successfully
+	// reach the data plane components.
+	// When True, control plane has available connection to data plane nodes.
+	// When False, there are network connection issues preventing control plane from reaching the data plane.
+	// A failure here may indicate network policy issues, firewall rules, or infrastructure problems
+	// like issues in data plane: missing nodes or in konnectivity-agent workload
```
devguyio:
```suggestion
	// DataPlaneConnectionAvailable indicates whether the control plane has a successful
	// network connection to the data plane components.
	//
	// **True** means the control plane can successfully reach the data plane nodes.
	// **False** means there are network connection issues preventing the control plane from reaching the data plane.
	//
	// A failure here suggests potential issues such as: network policy restrictions,
	// firewall rules, missing data plane nodes, or problems with infrastructure
	// components like the konnectivity-agent workload.
```

The last sentence needed a bit of restructure so Gemini suggested the above edit.

api/hypershift/v1beta1/hostedcluster_conditions.go
```diff
@@ -250,7 +258,15 @@ const (
 
 	RecoveryFinishedReason = "RecoveryFinished"
 
+	ReconcileErrorReason = "ReconcileError"
+
 	CloudResourcesCleanupSkippedReason = "CloudResourcesCleanupSkipped"
+
+	DataPlaneConnectionNoKonnectivityAgentPodsFoundReason = "NoKonnectivityAgentPodsFound"
```
devguyio:
```suggestion
	DataPlaneConnectionNoKonnectivityAgentPodsFoundReason = "KonnectivityAgentPodsNotFound"
```

api/hypershift/v1beta1/hostedcluster_conditions.go
sdminonne: OK but I'm removing the `//` empty lines. Aesthetic is nice to have but empty lines take space in git repo and empty lines in comments... Well, it's too much :)

api/hypershift/v1beta1/hostedcluster_conditions.go
sdminonne: Oups, thanks!

sdminonne: Also the suggested one is wrong  Missing `Not` in const name 

~~`DataPlaneConnectionNoKonnectivityAgentPodsFoundReason`~~ `= "KonnectivityAgentPodsNotFound"`  => `DataPlaneConnectionNoKonnectivityAgentPodsNotFoundReason= KonnectivityAgentPodsNotFound"` 

Goint to fix it

devguyio:
I'm approving this PR, feel free to unhold after you the few NITs above are addressed and leaving it for @muraee to tag
/approve
/hold

mgencur:
/verified by unit tests

I verified the E2E functionality in the previous run. The latest changes are specific for condition names and they're covered by unit tests.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -1389,6 +1415,79 @@ func (r *reconciler) reconcileClusterVersion(ctx context.Context, hcp *hyperv1.H
...
+	} else {
+		if !logsFound {
+			condition.Reason = hyperv1.DataPlaneConnectionNoKonnectivityAgentPodsNotFoundReason
+			condition.Message = "failed to read konnectivity-agent logs from data plane"
+		} else {
+			condition.Status = metav1.ConditionTrue
+			condition.Reason = hyperv1.AsExpectedReason
+			condition.Message = hyperv1.AllIsWellMessage
+		}
+	}
```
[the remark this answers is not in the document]
sdminonne: @coderabbitai ignore it

api/hypershift/v1beta1/hostedcluster_conditions.go
```diff
@@ -250,7 +259,15 @@ const (
 
 	RecoveryFinishedReason = "RecoveryFinished"
 
+	ReconcileErrorReason = "ReconcileError"
+
 	CloudResourcesCleanupSkippedReason = "CloudResourcesCleanupSkipped"
+
+	DataPlaneConnectionNoKonnectivityAgentPodsNotFoundReason = "NoKonnectivityAgentPodsNotFound"
```
muraee:
```suggestion
	DataPlaneConnectionKonnectivityAgentPodsNotFoundReason = "NoKonnectivityAgentPodsNotFound"
```
nit: double negative

sdminonne: Thanks

sdminonne: fixed

PR #7274, opened 2025-11-21
GCP-251: feat(install): Add GCP provider support for external-dns with Workload Identity

patjlm:
## Summary

Adds Google Cloud Platform provider support to Hypershift's external-dns integration using principal-based Workload Identity Federation.

Complements gcp-hcp-infra changes in https://github.com/openshift-online/gcp-hcp-infra/pull/209

### GCP Provider Support

- Added `GCPExternalDNSProvider` constant ("google")
- Added `GoogleProject` field to `ExternalDNSDeployment` struct
- Added GCP case in provider switch with `--google-project` arg
- Added `--external-dns-google-project` CLI flag

### Optional Credentials for GCP (Workload Identity)

- Made credentials optional when `--external-dns-provider=google`
- Validation: Credentials required for AWS/Azure, optional for GCP
- Conditional volume/volumeMount: Only added when `CredentialsSecret != nil`

### Optional google-project Parameter

- `--external-dns-google-project` is optional
- When omitted: external-dns falls back to `EXTERNAL_DNS_GOOGLE_PROJECT` env var → metadata server auto-detection
- When provided: Added as `--google-project` arg to external-dns deployment
- Supports ArgoCD env var injection pattern

### Template/Helm Rendering Support

- Added `ExternalDNSGoogleProject` to `TemplateParams` struct
- OpenShift Template: Conditional `${EXTERNAL_DNS_GOOGLE_PROJECT}` parameter
- Helm: Added `.Values.externaldns.googleProject` support
- Conditional field wrapping: Only creates template vars for provided fields

### Test Coverage

Added 3 test cases:
- GCP provider without credentials (Workload Identity)
- GCP provider with credentials (optional)
- GCP provider without google-project (uses env var fallback)

## Authentication Flow

**Principal-Based Workload Identity (no annotations):**

```bash
# IAM binding grants permissions directly to Kubernetes SA principal
gcloud projects add-iam-policy-binding DNS_PROJECT_ID \
  --role=roles/dns.admin \
  --member=principal://iam.googleapis.com/projects/PROJECT_NUMBER/locations/global/workloadIdentityPools/PROJECT_ID.svc.id.goog/subject/ns/hypershift/sa/external-dns
```

**Installation:**
```bash
hypershift install \
  --external-dns-provider=google \
  --external-dns-domain-filter=example.com
  # --external-dns-google-project is optional (uses env var or auto-detect)
```

## Fallback Chain

When `--google-project` arg is omitted:
1. Checks `EXTERNAL_DNS_GOOGLE_PROJECT` environment variable
2. Falls back to GCP metadata server (auto-detects cluster project)
3. Fails if neither available

## Backward Compatibility

- AWS/Azure behavior unchanged (credentials still required)
- Existing external-dns deployments continue to work
- GCP users can provide credentials if needed (both auth methods supported)

## Testing

- [x] Unit tests pass (16/16)
- [x] Build successful
- [x] Validated all install modes: `install`, `install render`, `install render --template`, `install helm`
- [ ] Integration testing with GKE cluster + Workload Identity

Jira: https://issues.redhat.com/browse/GCP-251

### DIFF
--- a/cmd/install/assets/hypershift_operator.go
+++ b/cmd/install/assets/hypershift_operator.go
@@ -180,6 +180,7 @@ type ExternalDNSProvider string
 const (
 	AWSExternalDNSProvider   ExternalDNSProvider = "aws"
 	AzureExternalDNSProvider ExternalDNSProvider = "azure"
+	GCPExternalDNSProvider   ExternalDNSProvider = "google"
 )
 
 type ExternalDNSDeployment struct {
@@ -191,6 +192,7 @@ type ExternalDNSDeployment struct {
 	CredentialsSecret *corev1.Secret
 	TxtOwnerId        string
 	Proxy             *configv1.Proxy
+	GoogleProject     string
 }
 
 func (o ExternalDNSDeployment) Build() *appsv1.Deployment {
@@ -270,12 +272,6 @@ func (o ExternalDNSDeployment) Build() *appsv1.Deployment {
 								Privileged:             &privileged,
 							},
 							TerminationMessagePolicy: corev1.TerminationMessageFallbackToLogsOnError,
-							VolumeMounts: []corev1.VolumeMount{
-								{
-									Name:      "credentials",
-									MountPath: "/etc/provider",
-								},
-							},
 						},
 					},
 					ImagePullSecrets: []corev1.LocalObjectReference{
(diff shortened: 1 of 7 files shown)


cmd/install/install_test.go
```diff
@@ -83,6 +83,33 @@ func TestOptions_Validate(t *testing.T) {
 			},
 			expectError: true,
 		},
+		"when external-dns GCP provider is set without credentials it succeeds (Workload Identity)": {
+			inputOptions: Options{
+				PrivatePlatform:          string(hyperv1.NonePlatform),
```
cristianoveiga: For clarlity, should we use GCPPlatorm here ? I know that it's not available in main yet, but perhaps it's a good time to include GCP in the list of supported private platforms (I will also need that for GCP-197)

cristianoveiga: Although, looking at the implementation, it looks like that the `ExternalDNSProvider=google` is what drives the google-specific logic, not the privatePlatform attribute, so it doesn't matter to much (except for clarity in the tests, as I mentioned).

cmd/install/install.go
cristianoveiga, line 149: @patjlm if we want to add GCP as private platform, we will need to update this validation function here.

cristianoveiga: @patjlm also here - add GCP to the list of supported private platforms

cmd/install/assets/hypershift_operator.go
```diff
@@ -321,6 +330,13 @@ func (o ExternalDNSDeployment) Build() *appsv1.Deployment {
...
 			"--azure-config-file=/etc/provider/credentials",
 		)
+	case GCPExternalDNSProvider:
+		if len(o.GoogleProject) > 0 {
+			deployment.Spec.Template.Spec.Containers[0].Args = append(
+				deployment.Spec.Template.Spec.Containers[0].Args,
+				fmt.Sprintf("--google-project=%s", o.GoogleProject),
+			)
+		}
 	}
```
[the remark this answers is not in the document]
patjlm: No, for GCP we don't use credentials files, we only rely on WIF

PR #7277, opened 2025-11-21
NO-ISSUE: test(install): validate DynamicResourceAllocation field exists in CRD

csrwng:
Add unit test to verify that the field
spec.configuration.scheduler.profileCustomizations.dynamicResourceAllocation exists in the generated HostedCluster CRD manifests.

The test validates the field exists in all CRD variants (Default, TechPreviewNoUpgrade, CustomNoUpgrade) by traversing the CRD schema. This ensures that a bump of openshift/api does not accidentally remove this field from HyperShift without proper deprecation.

The test uses schema traversal with cycle detection and depth limits to safely navigate the CRD schema structure.

### DIFF
--- a/cmd/install/crd_validation_test.go
+++ b/cmd/install/crd_validation_test.go
@@ -1,0 +1,125 @@
+package install
+
+import (
+	"bytes"
+	"io"
+	"strings"
+	"testing"
+
+	. "github.com/onsi/gomega"
+
+	"github.com/openshift/hypershift/cmd/install/assets"
+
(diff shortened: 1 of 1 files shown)


JoelSpeed:
@csrwng Can we shift this problem left somehow? I'm concerned that this is protecting a vendor bump and not a merge at the source (o/api). It would be preferable to catch this and prevent the merge at o/api

csrwng:
sgtm @JoelSpeed. I'll create a test there and close this one.
/close

PR #7278, opened 2025-11-24
NVIDIA-314: Support Node CIDR Allocation for networkType=Other Using AllocateNodesCIDRs

linoyaslan:
This PR introduces a new field, `AllocateNodesCIDRs`, which allows the kube-controller-manager to allocate node CIDRs when networkType=Other. It’s recommended to enable this field when using the Flannel CNI.

## Checklist:
- [X] This change includes docs. 


api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1076,6 +1077,16 @@ type ClusterNetworking struct {
...
 	APIServer *APIServerNetworking `json:"apiServer,omitempty"`
+
+	// isFlannel – Set this variable to "true" when using networkType=Other with Flannel as the CNI.
+	// This enables the KCM to handle node CIDR allocation.
+	// Default is false.
+	// This field can only be set to true when NetworkType is "Other". Setting it to true
+	// with any other NetworkType will result in a validation error during cluster creation.
+	//
+	// +optional
+	// +kubebuilder:default:=false
```
jparrill: The field should be immutable:
```
	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="isFlannel is immutable and cannot be modified once set."
```

linoyaslan:
Hi @jparrill, thanks for reviewing! :) 
I get your concerns, but I think these are the same types of risks we face anytime we add a new API field, no?

Here’s how I see the problematic cases:
Scenario 1: HO rollback
User creates HostedCluster CR with isFlannel=true -> Older HO version doesn’t know the field, reads it as nil -> HCP gets nil as well -> CPO interprets nil as "don’t enable flannel logic".

Scenario 2: CPO vendoring mismatch
let's say HCP CR has isFlannel: true -> CPO using older API and read HCP  and unmarshals the json -> IsFlannel missing -> nil -> flannel logic disabled

api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/hostedcontrolplanes.hypershift.openshift.io/AutoNodeKarpenter.yaml
```diff
@@ -2659,6 +2659,19 @@ spec:
...
+                      isFlannel – Set this variable to "true" when using networkType=Other with Flannel as the CNI.
+                      This enables the KCM to handle node CIDR allocation.
+                      Default is false.
+                      This field can only be set to true when NetworkType is "Other". Setting it to true
+                      with any other NetworkType will result in a validation error during cluster creation.
+                    type: boolean
+                    x-kubernetes-validations:
+                    - message: isFlannel is immutable and cannot be modified once
+                        set.
+                      rule: self == oldSelf
```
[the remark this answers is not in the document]
linoyaslan: that's all the idea, we want it to be immutable

muraee:
Hey @linoyaslan, thanks for the PR.
I think we discussed this in the office meeting. We don't want to expose an API field for flannel, because this will imply that it's a supported CNI, which is not. 

I believe we want to expose an annotation (or possibly even an API field) that will configure the KCM with the flag `-allocate-node-cidrs=true`. This will be a generic configuration option for KCM not specifically related to flannel.
cc @csrwng

linoyaslan:
Hi @muraee :) 

> I think we discussed this in the office meeting. We don't want to expose an API field for flannel

Oh, I wasn’t invited to that meeting, so I didn’t participate. Based on the RFE, I understood that what I implemented was the expected behavior (though I may have misunderstood). According to what Ramon said:

> "Approved but we won't expose the flag outside of the Flannel use case, i.e. if flannel (potentially a new config flag to enable it) is specified hcp will change the flag (but not deploy flannel)"

api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/hostedclusters.hypershift.openshift.io/IngressControllerLBSubnetsAWS.yaml
```diff
@@ -2615,6 +2615,20 @@ spec:
...
+                      allocation. When using networkType=Other, it is recommended to set this field
+                      to "true" if Flannel is used as the CNI, as it relies on this behavior.
+                      Default is false.
+                      This field can only be set to true when NetworkType is "Other". Setting it to true
+                      with any other NetworkType will result in a validation error during cluster creation.
+                    type: boolean
+                    x-kubernetes-validations:
+                    - message: allocateNodeCidrs is immutable and cannot be modified
+                        once set.
+                      rule: self == oldSelf
```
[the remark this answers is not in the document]
linoyaslan: I think we have a default for hostPrefix right? so why we need to validate that?

api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/hostedcontrolplanes.hypershift.openshift.io/ExternalOIDCWithUIDAndExtraClaimMappings.yaml
```diff
@@ -3013,6 +3013,20 @@ spec:
...
+                      allocation. When using networkType=Other, it is recommended to set this field
+                      to "true" if Flannel is used as the CNI, as it relies on this behavior.
+                      Default is false.
+                      This field can only be set to true when NetworkType is "Other". Setting it to true
+                      with any other NetworkType will result in a validation error during cluster creation.
+                    type: boolean
+                    x-kubernetes-validations:
+                    - message: allocateNodeCidrs is immutable and cannot be modified
+                        once set.
+                      rule: self == oldSelf
```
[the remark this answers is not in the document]
linoyaslan: I put this validation in HostedControlPlaneSpec and HostedClusterSpec, isn't it sufficient?

api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/hostedclusters.hypershift.openshift.io/NetworkDiagnosticsConfig.yaml
```diff
@@ -2767,6 +2767,20 @@ spec:
...
+                      allocation. When using networkType=Other, it is recommended to set this field
+                      to "true" if Flannel is used as the CNI, as it relies on this behavior.
+                      Default is false.
+                      This field can only be set to true when NetworkType is "Other". Setting it to true
+                      with any other NetworkType will result in a validation error during cluster creation.
+                    type: boolean
+                    x-kubernetes-validations:
+                    - message: allocateNodeCidrs is immutable and cannot be modified
+                        once set.
+                      rule: self == oldSelf
```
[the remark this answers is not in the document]
linoyaslan: @muraee do you prefer allocateNodeCIDRs above allocateNodeCidrs?

api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/hostedclusters.hypershift.openshift.io/IngressControllerLBSubnetsAWS.yaml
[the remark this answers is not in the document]
linoyaslan: what about this:

```
func ReconcileNetworkConfig(cfg *configv1.Network, hcp *hyperv1.HostedControlPlane) error {
	var clusterNetwork []configv1.ClusterNetworkEntry
	for _, entry := range hcp.Spec.Networking.ClusterNetwork {
		hostPrefix := uint32(entry.HostPrefix)
		if hostPrefix == 0 {
			ipv4, err := util.IsIPv4CIDR(entry.CIDR.String())
			if err != nil {
				return fmt.Errorf("the CIDR %s included in the cluster network spec is not valid: %w", entry.CIDR.String(), err)
			}

			if ipv4 {
				hostPrefix = defaultIPv4HostPrefix
			} else {
				hostPrefix = defaultIPv6HostPrefix
			}
		}
		clusterNetwork = append(clusterNetwork, configv1.ClusterNetworkEntry{
			CIDR:       entry.CIDR.String(),
			HostPrefix: hostPrefix,
		})
	}
	cfg.Spec.ClusterNetwork = clusterNetwork
...
```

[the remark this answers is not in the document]
linoyaslan: @jparrill @muraee I'd appreciate your opinion on it. Does the defaults are enough indeed? or do we need validation for host-prefix? Thank you!

linoyaslan:
@jparrill @muraee I'd appreciate your re-review please. Thank you! :)

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1080,6 +1081,18 @@ type ClusterNetworking struct {
...
+
+	// allocateNodeCidrs is a flag that controls whether the KCM manages node CIDR
+	// allocation. When using networkType=Other, it is recommended to set this field
+	// to "true" if Flannel is used as the CNI, as it relies on this behavior.
+	// Default is false.
+	// This field can only be set to true when NetworkType is "Other". Setting it to true
+	// with any other NetworkType will result in a validation error during cluster creation.
+	//
+	// +optional
+	// +kubebuilder:default:=false
```
muraee: avoid setting defaults on the API level, default in code instead and mention the current default in the field description

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1080,6 +1081,18 @@ type ClusterNetworking struct {
...
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="allocateNodeCidrs is immutable and cannot be modified once set."
+	AllocateNodeCidrs *bool `json:"allocateNodeCidrs,omitempty"` // nolint:kubeapilinter
```
muraee: using bools is not recommended in APIs, use an enum(string type) instead. 
something like `Enabled` and `Disabled` enum values

api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/hostedclusters.hypershift.openshift.io/IngressControllerLBSubnetsAWS.yaml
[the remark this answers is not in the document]
muraee: I think this needs to be tested if it works or not without setting a host-prefix.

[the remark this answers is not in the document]
linoyaslan: I already tested this on a HostedCluster using Flannel as the CNI, and the CIDRs were allocated as expected. I didn’t specify a hostPrefix:

```
# oc get nodes -o yaml
      ...
  spec:
    podCIDR: 10.132.0.0/24
    podCIDRs:
    - 10.132.0.0/24
  status:
    addresses:
    - address: 10.6.135.212
      type: InternalIP
...
```

linoyaslan:
@jparrill @muraee Hi again :) I resolved @muraee comments. I'd appreciate your re-review please. Thank you! :)

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1091,6 +1081,17 @@ type ClusterNetworking struct {
 	//
 	// +optional
 	APIServer *APIServerNetworking `json:"apiServer,omitempty"`
+
+	// allocateNodeCidrs controls whether the KCM manages node CIDR allocation.
+	// When using networkType=Other, it is recommended to set this field to "Enabled"
+	// if Flannel is used as the CNI, as it relies on this behavior.
+	// Default is "Disabled" (set in code, not at API level).
```
muraee:
```suggestion
	// Default is "Disabled".
```

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1091,6 +1081,17 @@ type ClusterNetworking struct {
 	//
 	// +optional
 	APIServer *APIServerNetworking `json:"apiServer,omitempty"`
+
+	// allocateNodeCidrs controls whether the KCM manages node CIDR allocation.
```
muraee:
```suggestion
	// allocateNodeCidrs controls whether the kube-controller-manager manages node CIDR allocation.
```

control-plane-operator/controllers/hostedcontrolplane/v2/kcm/deployment.go
```diff
@@ -40,6 +40,27 @@ func adaptDeployment(cpContext component.WorkloadContext, deployment *appsv1.Dep
 			c.Args = append(c.Args, fmt.Sprintf("--cloud-provider=%s", "external"))
 		}
 
+		// Enable node CIDR allocation for Flannel CNI
```
muraee:
```suggestion
		// Enable node CIDR allocation if `.Spec.Networking.AllocateNodeCidrs` is enabled
```

control-plane-operator/controllers/hostedcontrolplane/v2/kcm/deployment.go
```diff
@@ -40,6 +40,27 @@ func adaptDeployment(cpContext component.WorkloadContext, deployment *appsv1.Dep
...
+						c.Args[i] = "--allocate-node-cidrs=true"
+					}
+					// If it's already true, keep it as is
+					found = true
+					break
+				}
+			}
+			// If not found, add it
+			if !found {
+				c.Args = append(c.Args, "--allocate-node-cidrs=true")
```
muraee: why are we doing this? you can just append the arg.
This deployment object is loaded from the static manifest, so it wouldn't have the arg always

api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/hostedcontrolplanes.hypershift.openshift.io/ExternalOIDC.yaml
```diff
@@ -2859,6 +2859,22 @@ spec:
...
+                      This field can only be set to "Enabled" when NetworkType is "Other". Setting it to "Enabled"
+                      with any other NetworkType will result in a validation error during cluster creation.
+                    enum:
+                    - Enabled
+                    - Disabled
+                    type: string
+                    x-kubernetes-validations:
+                    - message: allocateNodeCidrs is immutable and cannot be modified
+                        once set.
+                      rule: self == oldSelf
```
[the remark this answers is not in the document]
muraee: @linoyaslan changing the name to `allocateNodeCIDRs` makes sense. not the rest though :)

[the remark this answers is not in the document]
linoyaslan: > changing the name to allocateNodeCIDRs

ok, sure! will do

control-plane-operator/controllers/hostedcontrolplane/v2/kcm/deployment.go
linoyaslan: according to that - https://github.com/openshift/hypershift/blob/main/control-plane-operator/controllers/hostedcontrolplane/v2/assets/kube-controller-manager/deployment.yaml#L26
The static manifest sets --allocate-node-cidrs=false.
and based on my testing, this action is necessary to prevent a duplicate result from being generated:
--allocate-node-cidrs=false, --allocate-node-cidrs=true

muraee: I see. I think it's better to remove the flag from the manifest then. And set it always in code, either to true or false based on the API.

linoyaslan: So iiuc, you mean,  removing it from the manifest I shared above, and instead handling it in the code like this:

```
if hcp.Spec.Networking.AllocateNodeCidrs != nil && *hcp.Spec.Networking.AllocateNodeCidrs == hyperv1.AllocateNodeCidrsEnabled {
    c.Args[i] = "--allocate-node-cidrs=true"
} else {
    c.Args[i] = "--allocate-node-cidrs=false"
}

```

And that’s it?

muraee: exactly

linoyaslan:
Hi @muraee, I resolved all the comments. I'd appreciate your re-review please. Thank you! :)

api/hypershift/v1beta1/hosted_controlplane.go
```diff
@@ -41,6 +41,7 @@ type HostedControlPlane struct {
 // HostedControlPlaneSpec defines the desired state of HostedControlPlane
 // +kubebuilder:validation:XValidation:rule="self.platform.type == 'IBMCloud' ? size(self.services) >= 3 : size(self.services) >= 4",message="spec.services in body should have at least 4 items or 3 for IBMCloud"
 // +kubebuilder:validation:XValidation:rule="!has(self.operatorConfiguration) || !has(self.operatorConfiguration.clusterNetworkOperator) || !has(self.operatorConfiguration.clusterNetworkOperator.disableMultiNetwork) || !self.operatorConfiguration.clusterNetworkOperator.disableMultiNetwork || self.networking.networkType == 'Other'",message="disableMultiNetwork can only be set to true when networkType is 'Other'"
+// +kubebuilder:validation:XValidation:rule="!has(self.networking) || !has(self.networking.allocateNodeCIDRs) || self.networking.allocateNodeCIDRs == 'Disabled' || self.networking.networkType == 'Other'",message="allocateNodeCIDRs can only be set to Enabled when networkType is 'Other'"
```
jparrill: Does it make sense to have `allocateNodeCIDRs` with another networkType? Do you think this is more accurate:

```
!has(self.networking) || !has(self.networking.allocateNodeCIDRs) || self.networking.allocateNodeCIDRs != 'Enabled' || self.networking.networkType == 'Other'
```

Yours is more permissive I want to make sure we don't have gaps in the API :)

linoyaslan: My rule already rejects the invalid state (Enabled + non-Other). Why do you think a change is needed?

muraee: the current validation seems correct. However, I just noticed that we don't need to put it at the root spec level, the CEL validation can be put at `networking` object directly.

api/hypershift/v1beta1/hostedcluster_types.go
muraee, line 1045:
```suggestion
// +kubebuilder:validation:XValidation:rule="has(self.allocateNodeCIDRs) && self.allocateNodeCIDRs == 'Enabled' ? self.networkType == 'Other' : true",message="allocateNodeCIDRs can only be set to Enabled when networkType is 'Other'"
type ClusterNetworking struct {
```

linoyaslan: doesn't it sufficient to put this validation in the HCP/ HC spec (like I already did) ?

linoyaslan:
Hi @muraee, I resolved your new comment. I'd appreciate your re-review please. Thank you! :)

PR #7281, opened 2025-11-24
OCPBUGS-62944: forceful termination of instances during HCP AutoNode deletion

maxcao13:
When a HostedControlPlane is deleted, Karpenter NodeClaims may get stuck during deletion due to stuck pods, finalizers, or volume detachment issues. This blocks and AutoNode-enabled cluster deletion indefinitely.

When a HostedControlPlane is deleted, Karpenter NodeClaims may get stuck
during deletion due to stuck pods, finalizers, or volume detachment issues.
This blocks and AutoNode-enabled cluster deletion indefinitely.

This change implements forceful instance termination to unblock deletions for AWS AutoNode clusters.
When HCP is deleted, we set an annotation on each NodeClaim, signaling Karpenter to force delete the underlying nodes and instances

Fixes: https://issues.redhat.com/browse/OCPBUGS-62944

~I'm not completely convinced that this is the correct approach. I am basically emulating what karpenter-provider-aws does when using a `terminationGracePeriod=0` on a NodePool. The AWS Karpenter provider will directly terminate EC2 instances if for example PDBs or volume attachments blocking Node deletions.~

~Maybe a safer approach would to re-implement a node drain + volume deletion, but my opinion is that we can instead forcefully terminate instances, and let Karpenter do the rest of the work in deleting the Node objects. In the future, if we implementation autonode disablement, then it would best to add a configurable timeout for graceful deletion first.~
~_However_, this is assuming that we as a service provider do not give guarantees on data persistence or graceful shutdown _because_ the user, or whoever, chose to delete the cluster themselves. This is probably a knowledge gap on my end on how HCP consumers like ROSA wants to behave.~

Above is obsolete, we allow Karpenter to forcefully delete stuck nodeclaims in a forceful shutdown situation.

### DIFF
--- a/karpenter-operator/controllers/karpenter/karpenter_controller.go
+++ b/karpenter-operator/controllers/karpenter/karpenter_controller.go
@@ -23,7 +23,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/utils/ptr"
 
@@ -41,6 +40,13 @@ import (
 
 const (
 	karpenterFinalizer = "hypershift.openshift.io/karpenter-finalizer"
+
+	// NodeClaimDeletionTimeout is the timeout for the deletion of a NodeClaim during cluster deletion.
+	// If the timeout is reached, the NodeClaim will be forcefully deleted by setting the termination timestamp annotation.
+	NodeClaimDeletionTimeout = 3 * time.Minute
+
+	// KarpenterDeletionRequeueInterval is the interval at which the controller will requeue deletion of Karpenter resources during a hosted cluster deletion.
+	KarpenterDeletionRequeueInterval = 15 * time.Second
 )
 
 var (
@@ -143,10 +149,11 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 	hcp, err := r.getHCP(ctx)
 	if err != nil {
-		if apierrors.IsNotFound(err) {
-			return ctrl.Result{}, nil
-		}
 		return ctrl.Result{}, err
+	}
+	if hcp == nil {
+		log.Info("HostedControlPlane not found")
+		return ctrl.Result{}, nil
 	}
 
 	// Setup for ControlPlaneContext and the Karpenter control plane v2 component.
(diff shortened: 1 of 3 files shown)


maxcao13:
/hold

Did not expect that many changes to be in this PR. Gonna tone down the vendoring.

karpenter-operator/controllers/karpenter/karpenter_controller_test.go
```diff
@@ -193,3 +198,383 @@ func TestKarpenterDeletion(t *testing.T) {
...
+				nodeClaim := &karpenterv1.NodeClaim{}
+				err := fakeGuestClient.Get(ctx, client.ObjectKey{Name: nodeClaimName}, nodeClaim)
+				g.Expect(err).NotTo(HaveOccurred())
+
+				hasAnnotation := nodeClaim.Annotations[KarpenterInstanceTerminationAttemptedAnnotation] == "true"
+				g.Expect(hasAnnotation).To(Equal(shouldHaveAnnotation))
+			}
+		})
+	}
+}
```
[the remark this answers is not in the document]
maxcao13: The NodeClaim has a finalizer so it won't actually get deleted in the mocked env.

maxcao13:
Reverted the changes that were using aws sdk 2 back to 1 so we don't change a bunch of dependencies.

Also pushed with unit tests + e2e validation step, and some changes to satisfy code rabbit.

karpenter-operator/controllers/karpenter/karpenter_controller.go
```diff
@@ -329,3 +375,110 @@ func (r *Reconciler) getHCP(ctx context.Context) (*hyperv1.HostedControlPlane, e
...
+	}
+
+	client, err := r.EC2ClientFactory(region)
+	if err != nil {
+		return nil, err
+	}
+
+	r.ec2Client = client
+	return client, nil
+}
```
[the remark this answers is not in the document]
maxcao13: HCP clusters are confined to one region only.

muraee:
can we just set NodeClaim’s `spec.expireAfter` field to `0s` or something to force deletion automatically by karpenter?
https://karpenter.sh/docs/concepts/disruption/#expiration

maxcao13:
> can we just set NodeClaim’s spec.expireAfter field to 0s or something to force deletion automatically by karpenter?
https://karpenter.sh/docs/concepts/disruption/#expiration

We cannot because NodeClaims are immutable, as much as I'd like Karpenter to do all the work for us. I wish there was a way to set terminationGracePeriod to 0 as well, but setting it on the NodePool will cause drift. Defaulting NodePool terminationGracePeriod 0 doesn't solve the problem either if a user wants to custom set it to infinity.

muraee:
@maxcao13 ack, thanks.
/approve

maxcao13:
I'm gonna adjust this PR to have zero grace period when hosted cluster is deleted, since ROSA HCP prefers forceful deletion when this happens.

I will keep the path open in case we ever want to support autonode disablement however (which would probably involve some sort of tunable grace period).

karpenter-operator/controllers/karpenter/karpenter_controller_test.go
```diff
@@ -193,3 +198,336 @@ func TestKarpenterDeletion(t *testing.T) {
...
+						Finalizers: []string{"test-finalizer"},
+						// No DeletionTimestamp
+					},
+					Status: karpenterv1.NodeClaimStatus{
+						NodeName:   "test-node-orphaned",
+						ProviderID: "aws:///us-east-1a/i-orphaned123456789",
+					},
+				},
+			},
+			expectedTerminations: 0,
```
elmiko: just curious why this doesn't result in a termination?

elmiko (commented):
i think this PR looks mostly good, i just have a question about one of the tests.

karpenter-operator/controllers/karpenter/karpenter_controller_test.go
maxcao13: Because the code returns in this code path instead of continuing the reconcilliation loop, and termination in this context means ec2 api instance termination.

Ref: https://github.com/openshift/hypershift/pull/7281/files#diff-8c5ef8eaf4d0c36cdaeae363f8b4e54b1eca3e8771b7546a5dbb284721d32a69R236-R241

elmiko: got it, thanks!

maxcao13:
Waiting for hypershfit qe to verify this now: https://redhat-external.slack.com/archives/C01C8502FMM/p1767829436482759

karpenter-operator/controllers/karpenter/karpenter_controller.go
```diff
@@ -324,8 +368,115 @@ func (r *Reconciler) getHCP(ctx context.Context) (*hyperv1.HostedControlPlane, e
...
+
+func parseEC2InstanceIDFromProviderID(providerID string) string {
+	// providerID format: aws:///<availability-zone>/<instance-id>
+	// Example: aws:///us-east-1a/i-0123456789abcdef0
+
+	// Validate scheme
+	if !strings.HasPrefix(providerID, "aws://") {
+		return ""
+	}
```
jimma: can this check be removed as it only supports aws platform ?

karpenter-operator/controllers/karpenter/karpenter_controller.go
```diff
@@ -324,8 +368,115 @@ func (r *Reconciler) getHCP(ctx context.Context) (*hyperv1.HostedControlPlane, e
...
 }
+
+// handleNodeClaimDeletion handles the timeout of a NodeClaim during cluster deletion.
+func (r *Reconciler) handleNodeClaimDeletion(ctx context.Context, hcp *hyperv1.HostedControlPlane, nodeClaim *karpenterv1.NodeClaim) error {
+	log := ctrl.LoggerFrom(ctx)
+
+	// Check if we've already attempted termination
+	if nodeClaim.Annotations[KarpenterInstanceTerminationAttemptedAnnotation] == "true" {
+		log.Info("Instance termination already attempted, skipping", "nodeClaim", nodeClaim.Name)
+		return nil
```
jimma, line 376: Does it only allow to delete NodeClaim for one time?  If there is error from ec2 client, can it try for another time?

maxcao13: Yes it will try again and _not_ put that annotation on if there is an error from ec2 client: https://github.com/openshift/hypershift/pull/7281/files#diff-8c5ef8eaf4d0c36cdaeae363f8b4e54b1eca3e8771b7546a5dbb284721d32a69R395-R399

karpenter-operator/controllers/karpenter/karpenter_controller.go
maxcao13: I just removed the whole function, and used a similar pattern here: https://github.com/openshift/hypershift/blob/fe3f21d746594fc4b56079769a16953490f2867b/karpenter-operator/controllers/karpenter/machine_approver.go#L209

karpenter-operator/controllers/karpenter/karpenter_controller.go
jimma: Thanks @maxcao13.

karpenter-operator/controllers/karpenter/karpenter_controller.go
jimma: Thanks @maxcao13 .

jimma:
I tested this PR and I didn't see the blocker issue and other failures when delete cluster.  
```
{"level":"info","ts":"2026-01-20T20:58:48+08:00","msg":"Deleted role","role":"ema-autonode422-5gwxn-worker-role"}
{"level":"info","ts":"2026-01-20T20:58:48+08:00","msg":"Deleting Secrets","namespace":"clusters"}
{"level":"info","ts":"2026-01-20T20:58:50+08:00","msg":"Deleted CLI generated secrets"}
{"level":"info","ts":"2026-01-20T20:58:51+08:00","msg":"Finalized hosted cluster","namespace":"clusters","name":"ema-autonode422"}
{"level":"info","ts":"2026-01-20T20:58:51+08:00","msg":"Successfully destroyed cluster and infrastructure","namespace":"clusters","name":"ema-autonode422","infraID":"ema-autonode422-5gwxn"}
```

enxebre:
can we see the tech-preview test passing here?

enxebre:
shouldn't we be able to achieve this via https://github.com/kubernetes-sigs/karpenter/blob/3c00d41c0e290b1cac96efcb6a125c8b3d95b236/pkg/apis/v1/nodepool.go#L225-L242 without the need to remove finalizers and ec2?

maxcao13:
> shouldn't we be able to achieve this via https://github.com/kubernetes-sigs/karpenter/blob/3c00d41c0e290b1cac96efcb6a125c8b3d95b236/pkg/apis/v1/nodepool.go#L225-L242 without the need to remove finalizers and ec2?


I indirectly answered that here, but you cant retroactively set terminationGracePeriod on a NodePool, otherwise that causes drift first. And you can't set it on a NodeClaim individually cause they are immutable, and it would probably be undefined behaviour if we allowed that part to be mutable. I've also talked about this on Slack, and I didn't see any alternatives: https://kubernetes.slack.com/archives/C02SFFZSA2K/p1764271296619679

maxcao13:
From[ Jason's comment upstream](https://github.com/kubernetes-sigs/karpenter/issues/2815#issuecomment-3786305516), I've completely overhauled the instance termination stuff, and just put an annotation to simulate a terminationGracePeriod set to 0 on all the nodeclaims. This allows Karpenter code to do all the forceful termination for us without re-implementing it here. We'll come back to this code once that issue is resolved and implemented.

karpenter-operator/controllers/karpenter/karpenter_controller.go
```diff
@@ -41,6 +40,13 @@ import (
 
 const (
 	karpenterFinalizer = "hypershift.openshift.io/karpenter-finalizer"
+
+	// NodeClaimDeletionTimeout is the timeout for the deletion of a NodeClaim during cluster deletion.
+	// If the timeout is reached, the NodeClaim will be forcefully deleted by setting the termination timestamp annotation.
+	NodeClaimDeletionTimeout = 3 * time.Minute
```
enxebre, line 46: why wait at all?

enxebre:
@coderabbitai review

karpenter-operator/controllers/karpenter/karpenter_controller.go
maxcao13: I don't think that code is actually being used right now, because I have `force` hardcoded to `true` right now. I was preparing in case we ever have to do autonode disablement, in which case we probably would some node termination grace period, if we decide that's what we want to do.

maxcao13:
For posterity reasons, this is what the annotation does in the code:
annotation itself: https://github.com/openshift/kubernetes-sigs-karpenter/blob/ee4e950462f88532a8474a73bf3157625e398406/pkg/controllers/node/termination/controller.go#L382
where it's used in business logic: https://github.com/openshift/kubernetes-sigs-karpenter/blob/ee4e950462f88532a8474a73bf3157625e398406/pkg/controllers/node/termination/controller.go#L128

The annotation gets passed to `awaitDrain`, `awaitVolumeDetachment`, and `awaitInstanceTermination` in that order, [which run in a loop](https://github.com/openshift/kubernetes-sigs-karpenter/blob/ee4e950462f88532a8474a73bf3157625e398406/pkg/controllers/node/termination/controller.go#L145-L149), only continuing if the previous function didn't error out.
1. For `awaitDrain`, it's a bit complicated in the code, but the gist is that Karpenter itself attempts to force delete pods when the timestamp is reached. Then, if Karpenter notices the pods are stuck deleting with [this function](https://github.com/openshift/kubernetes-sigs-karpenter/blob/1b5824bc141d9cd3a24ecac241204937c3145d22/pkg/utils/pod/scheduling.go#L76-L83) `IsStuckTerminating`, then eventually Karpenter skips these pods and moves on to the next functions.
2. For `awaitVolumeDetachment`, we simply just skip the process if the timestamp is reached.
3. For  `awaitInstanceTermination`, the timestamp is not used, we simply just terminate the instances using ec2 client (in AWS Karpenter code) if the previous functions in the loop don't return an error.

So if the timestamp is hit, Karpenter should never block deletion in recoverable state. In a unrecoverable state, there will be a followup PR to handle this.

PR #7294, opened 2025-11-26
OCPBUGS-63717: Fix flaky EnsureGlobalPullSecret test race condition

bryan-cox:

Fixes a race condition in the `EnsureGlobalPullSecret` E2E test that caused flaky failures.

The `waitForDaemonSetsReady` function in `AllowPartialNodes` mode was comparing against a stale `nodeCount` value captured before node availability changed. When additional nodes became available during test execution, the DaemonSet would have more pods (e.g., 2) than expected (e.g., 1), causing the polling loop to fail indefinitely with messages like:
```
DaemonSet global-pull-secret-syncer not ready: 2/1 pods ready (based on available nodes)
```

This timeout would eventually trigger a Gomega assertion failure, and since the Gomega instance was created with the parent test's `*testing.T` rather than the subtest's, it caused a panic:
```
testing.go:1679: test executed panic(nil) or runtime.Goexit: subtest may have called FailNow on a parent test
```

The fix uses the DaemonSet's own `DesiredNumberScheduled` field which always reflects the current cluster state, eliminating the race condition entirely.

Fixes https://issues.redhat.com/browse/OCPBUGS-63717

The change is minimal and focused - only modifying the `AllowPartialNodes` mode in `waitForDaemonSetsReady` to use `DesiredNumberScheduled` instead of a potentially stale `nodeCount`.

## Checklist:

🤖 Generated with [Claude Code](https://claude.com/claude-code) via `/jira:solve OCPBUGS-63717`

### DIFF
--- a/test/e2e/util/util.go
+++ b/test/e2e/util/util.go
@@ -2065,7 +2065,7 @@ func waitForDaemonSetsReady(t *testing.T, ctx context.Context, guestClient crcli
 		allowPartialNodes := dsManifest.AllowPartialNodes
 
 		if allowPartialNodes {
-			t.Logf("Waiting for %s DaemonSet to be ready with ≤%d available nodes", dsName, nodeCount)
+			t.Logf("Waiting for %s DaemonSet to be ready (using DesiredNumberScheduled)", dsName)
 		} else {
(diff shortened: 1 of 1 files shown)


bryan-cox:
@coderabbitai review

jparrill:
nice catch

PR #7295, opened 2025-11-26
GCP-199: add createDnsZones field to GCPPlatformSpec

cristianoveiga:

This PR adds a new `createDnsZones` boolean field to the `GCPPlatformSpec` API to provide control over DNS zone management for GCP HyperShift clusters.

**Key changes:**
- Adds `createDnsZones` field to `api/hypershift/v1beta1/gcp.go`
- Generates all associated API artifacts (CRDs, apply configurations, client code)
- Adds linter exclusion following existing project patterns
- Uses simple boolean semantics with clear default behavior

**Why this is needed:**
- Enables users to choose between automatic DNS zone creation or using pre-existing zones
- Provides flexibility for different deployment scenarios (managed vs. customer-managed DNS)
- Follows established patterns from other platform implementations
- Supports both greenfield deployments (auto-create zones) and brownfield deployments (use existing zones)

**API Design:**
- `createDnsZones: true` → control-plane-operator creates Cloud DNS zones for cluster ingress and internal communication (hypershift.local)
- `createDnsZones: false` (default) → assumes DNS zones are externally managed
- Simple boolean field with `+kubebuilder:default=false` for safe defaults

Pre-work of [GCP-199](https://issues.redhat.com//browse/GCP-199)

- This PR only adds the API field - the actual DNS zone creation logic will be implemented in a follow-up PR
- Linter exclusion was added to `.golangci.yml` following the established pattern for other boolean fields
- All generated code (CRDs, apply configurations, documentation) has been updated via `make update`
- Default behavior is conservative: `false` means no automatic zone creation

## Checklist:

### DIFF
--- a/api/.golangci.yml
+++ b/api/.golangci.yml
@@ -120,7 +120,7 @@ issues:
       linters:
         - kubeapilinter
     - path: "hypershift/v1beta1/gcp.go"
-      text: "nobools: field CreateDnsZones should not use a bool. Use a string type with meaningful constant values as an enum."
+      text: "nobools: field CreateDNSZones should not use a bool. Use a string type with meaningful constant values as an enum."
       linters:
         - kubeapilinter
     - path: "hypershift/v1beta1/hostedcluster_types.go"
--- a/api/hypershift/v1beta1/gcp.go
+++ b/api/hypershift/v1beta1/gcp.go
@@ -87,5 +87,5 @@ type GCPPlatformSpec struct {
 	// and internal communication (hypershift.local). Defaults to false.
 	// +kubebuilder:default=false
 	// +optional
-	CreateDnsZones bool `json:"createDnsZones,omitempty"`
+	CreateDNSZones bool `json:"createDnsZones,omitempty"`
 }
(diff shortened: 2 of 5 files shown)


cmd/install/assets/hypershift-operator/zz_generated.crd-manifests/hostedcontrolplanes-CustomNoUpgrade.crd.yaml
```diff
@@ -5640,6 +5640,13 @@ spec:
                     description: gcp specifies configuration for clusters running
                       on Google Cloud Platform.
                     properties:
+                      createDnsZones:
+                        default: false
+                        description: |-
+                          createDnsZones enables automatic DNS zone creation in the customer project.
+                          When true, the control-plane-operator creates Cloud DNS zones for cluster ingress
+                          and internal communication (hypershift.local). Defaults to false.
+                        type: boolean
```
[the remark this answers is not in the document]
cristianoveiga: createDnsZones is a new field.

api/.golangci.yml
```diff
@@ -119,6 +119,10 @@ issues:
       text: "nobools: field MultiArch should not use a bool. Use a string type with meaningful constant values as an enum."
       linters:
         - kubeapilinter
+    - path: "hypershift/v1beta1/gcp.go"
+      text: "nobools: field CreateDNSZones should not use a bool. Use a string type with meaningful constant values as an enum."
```
muraee, line 123: why are we adding an exception? bool API fields should be avoided.

cristianoveiga: I understand the general rule of thumb to avoid bools, but this really sounded like a scenario where it would be okay/appropriate to have it as bool, so I added the exception.

I can switch it to enum, if there's a strong recommendation around that.

cristianoveiga:
As per the discussion with the Hypershift team, we'll track the self-managed support for GCP as a separate effort.

PR #7298, opened 2025-11-27
OCPBUGS-66067: fix(kas): apply LoadBalancerSourceRanges only for LoadBalancer service type

muraee:

when `hc.spec.allowedCIDRBlocks` is set, only set `LoadBalancerSourceRanges` filed on KAS service  when service is of type LoadBalancer.

Fixes: https://issues.redhat.com/browse/OCPBUGS-66067

## Checklist:


muraee:
/override "ci/prow/e2e-aks"
changes don't affect aks

PR #7302, opened 2025-11-28
GCP-197: add GCPPrivateServiceConnect controller for managing Service Attachments for Private Service Connect

cristianoveiga:

This PR introduces the GCPPrivateServiceConnect controller for managing Service Attachments for Private Service Connect in the management VPC. It also includes a bug fix for the install command to properly scope AWS credentials volume mounting.

Key changes:

1. GCPPrivateServiceConnectReconciler controller - Manages Service Attachments for Private Service Connect:
  - Lookup ForwardingRule by LoadBalancer IP using GCP API filters
  - Discover available PSC-purpose NAT subnets with usage tracking
  - Create Service Attachments with proper target service and subnet URLs
  - Status management with proper Service Attachment URI format
  - Error handling with appropriate retry logic for GCP API failures
  - Finalizer-based cleanup on resource deletion
2. Dependency update - Updated google.golang.org/api to include the compute v1 API required for the controller
3. Bug fix for install command - The install command previously mounted AWS credentials volume for all non-None private platforms. With the addition of GCP as a private platform, this caused failures when GCP was selected (as AWSPrivateSecret is not configured). The fix scopes the AWS credentials volume mounting to only the AWS platform case.

Fixes [GCP-197](https://issues.redhat.com//browse/GCP-197)

  - This is the management-side counterpart to the GCPPrivateServiceObserver added in [GCP-196](https://issues.redhat.com//browse/GCP-196)
  - The controller follows HyperShift AWS controller patterns for consistency in authentication, logging, and resource management
  - The install command bug was discovered while testing GCP private platform support

## Checklist:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -78,7 +78,7 @@ require (
 	go.uber.org/zap v1.27.1
 	golang.org/x/crypto v0.46.0
 	golang.org/x/net v0.48.0
-	golang.org/x/oauth2 v0.30.0
+	golang.org/x/oauth2 v0.34.0
 	golang.org/x/sync v0.19.0
 	golang.org/x/time v0.14.0
 	google.golang.org/api v0.258.0
@@ -271,7 +271,6 @@ require (
 	go.yaml.in/yaml/v3 v3.0.4 // indirect
 	golang.org/x/exp v0.0.0-20250305212735-054e65f0b394 // indirect
 	golang.org/x/mod v0.30.0 // indirect
-	golang.org/x/oauth2 v0.34.0 // indirect
 	golang.org/x/sys v0.39.0 // indirect
 	golang.org/x/term v0.38.0 // indirect
 	golang.org/x/text v0.32.0 // indirect


hypershift-operator/controllers/platform/gcp/privateserviceconnect_controller.go
```diff
@@ -0,0 +1,615 @@
...
+				log.Info("Found available PSC subnet", "subnet", subnet.Name)
+				return subnet.Name, nil
+			}
+
+			log.V(1).Info("Subnet already in use, trying next", "subnet", subnet.Name)
+		}
+	}
+
+	return "", fmt.Errorf("no available PRIVATE_SERVICE_CONNECT subnet found in region %s", r.Region)
+}
```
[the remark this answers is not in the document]
cristianoveiga: This is a valid concern, which we will address eventually - the current approach is sufficient for the time being.

hypershift-operator/controllers/platform/gcp/privateserviceconnect_controller.go
```diff
@@ -0,0 +1,615 @@
...
+	}
+
+	// 4. Wait for operation completion
+	if op.Status == "RUNNING" {
+		log.Info("Service Attachment creation in progress", "operation", op.Name)
+		return ctrl.Result{RequeueAfter: time.Second * 30}, nil
+	}
+
+	return ctrl.Result{}, nil
+}
```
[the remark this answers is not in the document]
cristianoveiga: Added the check for operation errors.

hypershift-operator/controllers/platform/gcp/privateserviceconnect_controller.go
```diff
@@ -0,0 +1,615 @@
...
+	// Total format: up to 15 + 1 + 8 + 1 + 20 + 7 = 52 chars (well within 63 limit)
+	serviceAttachmentName := fmt.Sprintf("%s-%s-%s-psc-sa", pscName, clusterID, clusterName)
+
+	// Ensure we don't exceed GCP's 63-character limit
+	if len(serviceAttachmentName) > 63 {
+		serviceAttachmentName = serviceAttachmentName[:63]
+	}
+
+	return serviceAttachmentName
+}
```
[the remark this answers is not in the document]
cristianoveiga: Simplified the logic here using the cluster ID only

hypershift-operator/controllers/platform/gcp/privateserviceconnect_controller.go
```diff
@@ -0,0 +1,594 @@
...
+		return "", fmt.Errorf("failed to list subnets: %w", err)
+	}
+
+	// 3. Find first available PRIVATE_SERVICE_CONNECT subnet that's not already in use
+	// We want subnet isolation between different Service Attachments for security
+	//
+	// Note: This approach still allows race conditions if multiple controllers
+	// are creating Service Attachments simultaneously (especially across different
+	// management clusters). Future improvements could include GCP-native subnet
+	// reservation mechanisms or external coordination systems.
```
patjlm: note we plan a single management cluster per project

apahim (commented):
## Missing Status Update After Service Attachment Creation

### Issue

In `reconcileServiceAttachment()` around lines 320-358, when a Service Attachment is successfully created (`op.Status == "DONE"`), the function returns without updating the CR status:

```go
// 4. Check operation status - check for errors first, then check if still running
if op.Error != nil && len(op.Error.Errors) > 0 {
    opErr := fmt.Errorf("operation failed: %s", op.Error.Errors[0].Message)
    return r.handleGCPError(ctx, gcpPSC, "ServiceAttachmentCreationFailed", opErr)
}

if op.Status == "RUNNING" {
    log.Info("Service Attachment creation in progress", "operation", op.Name)
    return ctrl.Result{RequeueAfter: time.Second * 30}, nil
}

return ctrl.Result{}, nil  // ❌ Status not updated when op.Status == "DONE"
```

### Impact

When the Service Attachment creation completes:
1. ✅ Resource is created in GCP
2. ❌ CR status remains empty (`ServiceAttachmentName`, `ServiceAttachmentURI`, conditions not set)
3. ✅ Next reconciliation loop updates status (when `existingServiceAttachment != nil` at line 310)

This creates a temporary inconsistency where the GCP resource exists but the CR status doesn't reflect it.

### Suggested Fix

After successful operation completion, fetch the created Service Attachment and update status:

```go
// 4. Check operation status - check for errors first, then check if still running
if op.Error != nil && len(op.Error.Errors) > 0 {
    opErr := fmt.Errorf("operation failed: %s", op.Error.Errors[0].Message)
    return r.handleGCPError(ctx, gcpPSC, "ServiceAttachmentCreationFailed", opErr)
}

if op.Status == "RUNNING" {
    log.Info("Service Attachment creation in progress", "operation", op.Name)
    return ctrl.Result{RequeueAfter: time.Second * 30}, nil
}

// Operation completed successfully - fetch the created Service Attachment and update status
log.Info("Service Attachment created successfully", "name", serviceAttachmentName)
createdServiceAttachment, err := r.GcpClient.ServiceAttachments.Get(r.ProjectID, r.Region, serviceAttachmentName).Context(ctx).Do()
if err != nil {
    return ctrl.Result{}, fmt.Errorf("failed to get created Service Attachment: %w", err)
}

return r.updateStatusFromServiceAttachment(ctx, gcpPSC, createdServiceAttachment)
```

### Alternative Approach

If you prefer to avoid an extra GCP API call, you could construct a minimal Service Attachment object from the operation result:

```go
if op.Status == "DONE" {
    // Construct Service Attachment info from what we know
    createdServiceAttachment := &compute.ServiceAttachment{
        Name:          serviceAttachmentName,
        TargetService: r.constructForwardingRuleURL(gcpPSC.Spec.ForwardingRuleName),
        NatSubnets:    []string{r.constructSubnetURL(gcpPSC.Spec.NATSubnet)},
    }
    return r.updateStatusFromServiceAttachment(ctx, gcpPSC, createdServiceAttachment)
}
```

However, I'd recommend the first approach (fetching the resource) to ensure the status reflects the actual GCP state.

---

**Priority**: Medium - The controller works correctly, but status updates are delayed by one reconciliation cycle.

hypershift-operator/controllers/platform/gcp/privateserviceconnect_controller.go
cristianoveiga: Updated the comment here with this information.

cristianoveiga:
Thanks, I addressed this.

> ## Missing Status Update After Service Attachment Creation
> ### Issue
> In `reconcileServiceAttachment()` around lines 320-358, when a Service Attachment is successfully created (`op.Status == "DONE"`), the function returns without updating the CR status:
> 
> ```go
> // 4. Check operation status - check for errors first, then check if still running
> if op.Error != nil && len(op.Error.Errors) > 0 {
>     opErr := fmt.Errorf("operation failed: %s", op.Error.Errors[0].Message)
>     return r.handleGCPError(ctx, gcpPSC, "ServiceAttachmentCreationFailed", opErr)
> }
> 
> if op.Status == "RUNNING" {
>     log.Info("Service Attachment creation in progress", "operation", op.Name)
>     return ctrl.Result{RequeueAfter: time.Second * 30}, nil
> }
> 
> return ctrl.Result{}, nil  // ❌ Status not updated when op.Status == "DONE"
> ```
> 
> ### Impact
> When the Service Attachment creation completes:
> 
> 1. ✅ Resource is created in GCP
> 2. ❌ CR status remains empty (`ServiceAttachmentName`, `ServiceAttachmentURI`, conditions not set)
> 3. ✅ Next reconciliation loop updates status (when `existingServiceAttachment != nil` at line 310)
> 
> This creates a temporary inconsistency where the GCP resource exists but the CR status doesn't reflect it.
> 
> ### Suggested Fix
> After successful operation completion, fetch the created Service Attachment and update status:
> 
> ```go
> // 4. Check operation status - check for errors first, then check if still running
> if op.Error != nil && len(op.Error.Errors) > 0 {
>     opErr := fmt.Errorf("operation failed: %s", op.Error.Errors[0].Message)
>     return r.handleGCPError(ctx, gcpPSC, "ServiceAttachmentCreationFailed", opErr)
> }
> 
> if op.Status == "RUNNING" {
>     log.Info("Service Attachment creation in progress", "operation", op.Name)
>     return ctrl.Result{RequeueAfter: time.Second * 30}, nil
> }
> 
> // Operation completed successfully - fetch the created Service Attachment and update status
> log.Info("Service Attachment created successfully", "name", serviceAttachmentName)
> createdServiceAttachment, err := r.GcpClient.ServiceAttachments.Get(r.ProjectID, r.Region, serviceAttachmentName).Context(ctx).Do()
> if err != nil {
>     return ctrl.Result{}, fmt.Errorf("failed to get created Service Attachment: %w", err)
> }
> 
> return r.updateStatusFromServiceAttachment(ctx, gcpPSC, createdServiceAttachment)
> ```
> 
> ### Alternative Approach
> If you prefer to avoid an extra GCP API call, you could construct a minimal Service Attachment object from the operation result:
> 
> ```go
> if op.Status == "DONE" {
>     // Construct Service Attachment info from what we know
>     createdServiceAttachment := &compute.ServiceAttachment{
>         Name:          serviceAttachmentName,
>         TargetService: r.constructForwardingRuleURL(gcpPSC.Spec.ForwardingRuleName),
>         NatSubnets:    []string{r.constructSubnetURL(gcpPSC.Spec.NATSubnet)},
>     }
>     return r.updateStatusFromServiceAttachment(ctx, gcpPSC, createdServiceAttachment)
> }
> ```
> 
> However, I'd recommend the first approach (fetching the resource) to ensure the status reflects the actual GCP state.
> 
> **Priority**: Medium - The controller works correctly, but status updates are delayed by one reconciliation cycle.

hypershift-operator/controllers/platform/gcp/privateserviceconnect_controller.go
```diff
@@ -0,0 +1,592 @@
...
+
+// reconcileGCPPrivateServiceConnectSpec reconciles the GCPPrivateServiceConnect spec fields
+func (r *GCPPrivateServiceConnectReconciler) reconcileGCPPrivateServiceConnectSpec(ctx context.Context, gcpPSC *hyperv1.GCPPrivateServiceConnect, hc *hyperv1.HostedCluster) error {
+	// Set ForwardingRuleName if not already populated
+	if gcpPSC.Spec.ForwardingRuleName == "" {
+		forwardingRuleName, err := r.lookupForwardingRuleName(ctx, gcpPSC)
+		if err != nil {
+			return fmt.Errorf("failed to lookup ForwardingRule: %w", err)
+		}
+		if forwardingRuleName != "" {
```
csrwng: nit: unnecessary check

hypershift-operator/controllers/platform/gcp/privateserviceconnect_controller.go
```diff
@@ -0,0 +1,592 @@
...
+	"google.golang.org/api/googleapi"
+	"google.golang.org/api/option"
+)
+
+const (
+	finalizer = "hypershift.openshift.io/gcp-private-service-connect"
+)
+
+// RBAC permissions for GCPPrivateServiceConnect controller
+// +kubebuilder:rbac:groups=hypershift.openshift.io,resources=gcpprivateserviceconnects,verbs=get;list;watch;update;patch
```
csrwng, line 38: nit: these are not used elsewhere and might give the impression that they cause some code to be generated. They're great as documentation though. I would just addd a note saying that they are for doc only.

hypershift-operator/controllers/platform/gcp/privateserviceconnect_controller.go
```diff
@@ -0,0 +1,592 @@
...
+
+// lookupForwardingRuleName finds the ForwardingRule name by LoadBalancer IP
+func (r *GCPPrivateServiceConnectReconciler) lookupForwardingRuleName(ctx context.Context, gcpPSC *hyperv1.GCPPrivateServiceConnect) (string, error) {
+	log := r.Log.WithValues("gcpprivateserviceconnect", gcpPSC.Name, "loadBalancerIP", gcpPSC.Spec.LoadBalancerIP)
+
+	// Use GCP Compute API filter syntax to find the forwarding rule by IP
+	filter := fmt.Sprintf("IPAddress eq %s", gcpPSC.Spec.LoadBalancerIP)
+
+	req := r.GcpClient.ForwardingRules.List(r.ProjectID, r.Region).Filter(filter)
+	resp, err := req.Context(ctx).Do()
```
csrwng: General question, do these calls have a default timeout? If not, it might not be a bad idea to pass a context with a timeout.

hypershift-operator/controllers/platform/gcp/privateserviceconnect_controller.go
```diff
@@ -0,0 +1,592 @@
...
+	req := r.GcpClient.Subnetworks.List(r.ProjectID, r.Region)
+	resp, err := req.Context(ctx).Do()
+	if err != nil {
+		return "", fmt.Errorf("failed to list subnets: %w", err)
+	}
+
+	// Find first available PRIVATE_SERVICE_CONNECT subnet not already in use.
+	// Race conditions are not a concern since we use a single management cluster per GCP project.
+	for _, subnet := range resp.Items {
+		if subnet.Purpose == "PRIVATE_SERVICE_CONNECT" {
```
csrwng, line 258: question: what will create these subnets?

hypershift-operator/controllers/platform/gcp/privateserviceconnect_controller.go
```diff
@@ -0,0 +1,592 @@
...
+	}
+
+	return "", fmt.Errorf("no available PRIVATE_SERVICE_CONNECT subnet found in region %s", r.Region)
+}
+
+// reconcileServiceAttachment manages Service Attachment lifecycle
+func (r *GCPPrivateServiceConnectReconciler) reconcileServiceAttachment(ctx context.Context, gcpPSC *hyperv1.GCPPrivateServiceConnect, hc *hyperv1.HostedCluster) (ctrl.Result, error) {
+	log := r.Log.WithValues("gcpprivateserviceconnect", gcpPSC.Name)
+
+	// 1. Construct unique Service Attachment name using PSC name + cluster ID + cluster name
```
csrwng: nit: this comment seems wrong. It looks like it's only using cluster ID

hypershift-operator/controllers/platform/gcp/privateserviceconnect_controller.go
```diff
@@ -0,0 +1,592 @@
...
+	}
+
+	// 3. Add finalizer if not present
+	if !controllerutil.ContainsFinalizer(gcpPSC, finalizer) {
+		controllerutil.AddFinalizer(gcpPSC, finalizer)
+		return ctrl.Result{}, r.Update(ctx, gcpPSC)
+	}
+
+	// 4. Find the hosted control plane and cluster (for normal reconciliation)
+	hcp, err := r.hostedControlPlane(ctx, gcpPSC.Namespace)
```
csrwng: It seems you only get the hcp so you can get the hostedcluster. It might be more efficient to put the same hc annotation on the gcpPSC as the hcp so you can just get the hostedcluster from the gcpPSC.

csrwng (commented):
some comments/questions

hypershift-operator/controllers/platform/gcp/privateserviceconnect_controller.go
cristianoveiga: The subnets are created via terraform during the Management Project set up.

hypershift-operator/controllers/platform/gcp/privateserviceconnect_controller.go
cristianoveiga: Agreed, removed.

hypershift-operator/controllers/platform/gcp/privateserviceconnect_controller.go
cristianoveiga: Added a comment about the documentation only purpose.

hypershift-operator/controllers/platform/gcp/privateserviceconnect_controller.go
cristianoveiga: They don't have a default timeout - followed your suggestion and added a context with timeout for the client calls.

hypershift-operator/controllers/platform/gcp/privateserviceconnect_controller.go
cristianoveiga: Indeed, this was outdated - I also updated this logic to ensure that the Service Attachment name starts with letters, not numbers (required by GCP)

hypershift-operator/controllers/platform/gcp/privateserviceconnect_controller.go
cristianoveiga: Indeed. I've updated the control plane operator controller to add the annotation to the gcpPSC CR and updated the hypershift operator code to read from it. It's much cleaner now (the original implementation was based on the AWS approach).

hypershift-operator/controllers/platform/gcp/privateserviceconnect_controller.go
```diff
@@ -0,0 +1,582 @@
...
+	if len(resp.Items) > 1 {
+		// This shouldn't happen, but handle gracefully
+		log.Info("Multiple ForwardingRules found for LoadBalancer IP, using first one", "count", len(resp.Items))
+	}
+
+	forwardingRule := resp.Items[0]
+	log.Info("Found ForwardingRule for LoadBalancer IP", "forwardingRule", forwardingRule.Name)
+
+	return forwardingRule.Name, nil
+}
```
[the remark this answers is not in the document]
cristianoveiga: Are you sure that `100.0.0.1` would match ?

hypershift-operator/controllers/platform/gcp/privateserviceconnect_controller.go
```diff
@@ -0,0 +1,582 @@
...
+	}
+
+	if op != nil && op.Status == "RUNNING" {
+		log.Info("Service Attachment deletion in progress", "operation", op.Name)
+		return false, nil // Not completed yet
+	}
+
+	log.Info("Service Attachment deletion completed", "name", serviceAttachmentName)
+	return true, nil // Deletion completed
+}
```
cristianoveiga, line 463: addressed.

hypershift-operator/controllers/platform/gcp/privateserviceconnect_controller_test.go
```diff
@@ -0,0 +1,533 @@
...
+			}
+			if !test.expectError && err != nil {
+				t.Errorf("unexpected error: %v", err)
+			}
+			if result != test.expected {
+				t.Errorf("expected %q, got %q", test.expected, result)
+			}
+		})
+	}
+}
```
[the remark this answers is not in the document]
cristianoveiga: Removed the test - it didn't have enough value, only env var checks.

hypershift-operator/controllers/platform/gcp/privateserviceconnect_controller.go
```diff
@@ -0,0 +1,587 @@
...
+	// 3. Create new Service Attachment
+	serviceAttachment := &compute.ServiceAttachment{
+		Name:                 serviceAttachmentName,
+		Description:          fmt.Sprintf("Service Attachment for HyperShift cluster %s", gcpPSC.Name),
+		TargetService:        r.constructForwardingRuleURL(gcpPSC.Spec.ForwardingRuleName),
+		ConnectionPreference: "ACCEPT_MANUAL",
+		ConsumerAcceptLists:  r.buildConsumerAcceptLists(gcpPSC.Spec.ConsumerAcceptList),
+		NatSubnets:           []string{r.constructSubnetURL(gcpPSC.Spec.NATSubnet)},
+		EnableProxyProtocol:  false,
+	}
```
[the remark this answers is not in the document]
cristianoveiga: This is not a valid scenario - at this point, the forwarding rule should be present.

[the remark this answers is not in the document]
cristianoveiga: The GCPPrivateServiceConnect CR is created by the ControlPlaneOperator when the LoadBalancer becomes available - see control-plane-operator/controllers/gcpprivateserviceconnect/observer.go.

we can still add the validation as a safety net for the isolated function, but this is not a valid real world scenario.

PR #7304, opened 2025-11-28
NO-JIRA: fix(rbac): add missing CAPI manager ClusterRole permissions for v1.10

apahim:

The cluster-api manager ClusterRole was missing required permissions for the crdmigrator controller introduced in CAPI v1.10.0. This caused permission errors when the controller attempted to patch CRDs and manage IPAM resources.

The crdmigrator controller replaces the deprecated clusterctl upgrade CRD storage version migration and requires additional RBAC permissions:
- patch and update verbs for CRDs (previously only get/list/watch)
- full CRUD permissions for ipam.cluster.x-k8s.io resources
- cluster-scoped permissions for cluster.x-k8s.io resources

This is a pre-existing bug that became visible after upgrading from CAPI v1.9.4 to v1.10.4 in commit 82af36570. The original ClusterRole implementation dates back to March 2021 (commit ed606fb50d) and affects all platforms (AWS, Azure, GCP, etc.), not just GCP.

Errors fixed:
- failed to patch CustomResourceDefinition: "cluster-api" cannot patch resource "customresourcedefinitions"
- failed to list IPAddress/IPAddressClaim: "cluster-api" cannot list resource "ipaddresses/ipaddressclaims" in API group "ipam.cluster.x-k8s.io"

## Checklist:

### DIFF
--- a/cmd/install/assets/hypershift_operator.go
+++ b/cmd/install/assets/hypershift_operator.go
@@ -1218,7 +1218,7 @@ func (o HyperShiftOperatorClusterRole) Build() *rbacv1.ClusterRole {
 			{
 				APIGroups: []string{"ipam.cluster.x-k8s.io"},
 				Resources: []string{"ipaddresses", "ipaddresses/status"},
-				Verbs:     []string{"create", "delete", "get", "list", "update", "watch"},
+				Verbs:     []string{"create", "delete", "get", "list", "patch", "update", "watch"},
 			},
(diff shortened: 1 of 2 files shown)


muraee:
[@apahim](https://redhat.enterprise.slack.com/team/UR94E41B8) we are aware of this. afaik those error logs don't cause any operational errors.
We don't want to run the migrator on Hypershift because it could mess up things that we are unaware of. The plan is to use `--skip-crd-migration-phases` flag instead

PR #7310, opened 2025-12-01
CNTRLPLANE-1814: Add support for setting max expiration duration for KAS service account tokens

muraee:

This PR adds support for setting the maximum expiration duration for service account tokens issued by the kube-apiserver. This is essential for graceful service account signing key rotation.

During SA signing key rotation, users need to enforce a limited token lifetime to ensure that all tokens are periodically refreshed and signed with the current signing key. This allows for a graceful rotation period where old tokens are still valid but eventually expire and get re-issued with the new key.

### Changes:

1. **API Change**: Added `KubeAPIServerServiceAccountTokenMaxExpiration` annotation constant that users can set on HostedCluster or HostedControlPlane resources.

2. **Control-Plane-Operator**: When the annotation is set, the CPO passes the configured duration value to kube-apiserver as the `--service-account-max-token-expiration` flag.

### Usage Example:

```yaml
apiVersion: hypershift.openshift.io/v1beta1
kind: HostedCluster
metadata:
  name: example
  annotations:
    hypershift.openshift.io/kube-apiserver-service-account-token-max-expiration: "24h"
spec:
  # ...
```

This will configure the kube-apiserver with `--service-account-max-token-expiration=24h`.

Fixes [CNTRLPLANE-1814](https://issues.redhat.com//browse/CNTRLPLANE-1814)

- The annotation follows the existing pattern used for other KAS configuration options (like `KubeAPIServerMaximumRequestsInFlight`, `KubeAPIServerGoAwayChance`, etc.)
- The value must be a valid Go duration string (e.g., "24h", "168h", "720h")
- Minimum value is 600s (10 minutes) per Kubernetes API server requirements
- Invalid values will cause the kube-apiserver to fail to start, which will be surfaced through HCP conditions

## Checklist:

🤖 Generated with [Claude Code](https://claude.com/claude-code) via `/jira:solve [CNTRLPLANE-1814](https://issues.redhat.com//browse/CNTRLPLANE-1814)`

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/v2/kas/config_test.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/v2/kas/config_test.go
@@ -570,6 +570,17 @@ func TestGenerateConfig(t *testing.T) {
 				},
 			),
 		},
+		{
+			name: "with service account max token expiration",
+			params: KubeAPIServerConfigParams{
+				ServiceAccountMaxTokenExpiration: "24h",
+			},
+			expected: modifyKasConfig(defaultKASConfig(),
+				func(kasc *kcpv1.KubeAPIServerConfig) {
+					kasc.APIServerArguments["service-account-max-token-expiration"] = kcpv1.Arguments{"24h"}
+				},
+			),
+		},
 	}
 
 	for _, tc := range testcases {
--- a/control-plane-operator/controllers/hostedcontrolplane/v2/kas/params_test.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/v2/kas/params_test.go
@@ -286,6 +286,22 @@ func TestNewConfigParams(t *testing.T) {
 			},
 		},
 		{
+			name: "with service account token max expiration annotation",
+			hcp: func() *hyperv1.HostedControlPlane {
+				hcp := createDefaultHostedControlPlane()
+				hcp.Annotations = map[string]string{
+					hyperv1.KubeAPIServerServiceAccountTokenMaxExpiration: "24h",
+				}
+				return hcp
+			}(),
+			expected: func(hcp *hyperv1.HostedControlPlane, featureGates []string) KubeAPIServerConfigParams {
+				params := defaultKubeAPIServerConfigParams()
+				params.FeatureGates = featureGates
+				params.ServiceAccountMaxTokenExpiration = "24h"
+				return params
+			},
+		},
+		{
 			name: "with full configuration",
 			hcp: func() *hyperv1.HostedControlPlane {
 				hcp := createDefaultHostedControlPlane()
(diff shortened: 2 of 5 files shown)


muraee:
@coderabbitai review

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -393,6 +393,13 @@ const (
 	// KubeAPIServerGoAwayChance allows the --goaway-chance parameter of the kube-apiserver to be overridden from its default of 0
 	KubeAPIServerGoAwayChance = "hypershift.openshift.io/kube-apiserver-goaway-chance"
 
+	// KubeAPIServerServiceAccountTokenMaxExpiration allows setting the maximum expiration duration
+	// for service account tokens issued by the kube-apiserver. This is useful during service account
+	// signing key rotation to enforce a limited token lifetime, ensuring tokens are re-issued with
+	// the new signing key. The value must be a valid Go duration string (e.g., "24h", "168h", "720h").
+	// Minimum value is 600s (10 minutes) per Kubernetes requirements.
```
jparrill, line 400: How are we validating the minimum value?

control-plane-operator/controllers/hostedcontrolplane/v2/kas/config_test.go
```diff
@@ -570,6 +570,17 @@ func TestGenerateConfig(t *testing.T) {
 				},
 			),
 		},
+		{
```
jparrill, line 573: I would add a failing test with a value under minimum allowed (E.G 9min).

control-plane-operator/controllers/hostedcontrolplane/v2/kas/params_test.go
```diff
@@ -285,6 +285,22 @@ func TestNewConfigParams(t *testing.T) {
...
+				}
+				return hcp
+			}(),
+			expected: func(hcp *hyperv1.HostedControlPlane, featureGates []string) KubeAPIServerConfigParams {
+				params := defaultKubeAPIServerConfigParams()
+				params.FeatureGates = featureGates
+				params.ServiceAccountMaxTokenExpiration = "24h"
+				return params
+			},
+		},
```
jparrill, line 303: I would add a failing test with a value under minimum allowed (E.G 9min).

api/hypershift/v1beta1/hostedcluster_types.go
muraee: there is no validation as all other API configs we pass to the KAS though annotations. the KAS would fail which will be reflected in the conditions.
I didn't want to break the pattern of all other fields, and introduce a validation for this one specifically. Although, it makes sense that we do that at some point

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -393,6 +393,13 @@ const (
...
+	KubeAPIServerServiceAccountTokenMaxExpiration = "hypershift.openshift.io/kube-apiserver-service-account-token-max-expiration"
```
rtheis, line 401: We should ensure that the annotation name matches the option. The annotation uses `TokenMax` and the option is `max-token`.

muraee: hey @rtheis, sorry I didn't catch this before the PR merged.

rtheis: No worries @muraee. It's not a big deal.

PR #7311, opened 2025-12-01
OCPBUGS-44436: fix(HostedClusterReconciler): check service nodeport address to avoid a panic

ingvagabund:
getNodePortIP does not check whether .nodeport is set. If not set accessing Address field produces a panic.

Fixes

## Checklist:
- [-] Relevant issues have been referenced.
- [-] This change includes docs. 


ingvagabund:
Adding a simple unit test (to make sure a panic is not raised when the nodeport is not set)

rtheis (approved):
/lgtm

ingvagabund:
This fix improves the stability

PR #7315, opened 2025-12-02
CNTRLPLANE-2123: feat: update latest supported version from 4.21 to 4.22

jparrill:
## Summary
- Update LatestSupportedVersion from 4.21.0 to 4.22.0
- Add Kubernetes 1.34.0 mapping for OCP 4.21.0

## Changes
- **Version constants**: Updated supported version range to 4.14-4.22
- **Kubernetes mappings**: Added 4.21.0 → 1.34.0 mapping
- **Test updates**: Updated all tests to reflect new version constraints
- **NodePool tests**: Fixed version skew test with compatible versions

## Fixes
- [CNTRLPLANE-2123](https://issues.redhat.com/browse/CNTRLPLANE-2123)

## Test plan
- [x] Unit tests pass with updated version ranges
- [x] `hypershift version` command shows 4.22.0 as latest supported
- [x] Version skew validation works correctly with new minimum versions
- [x] All existing functionality preserved

### DIFF
--- a/hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
+++ b/hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
@@ -1652,7 +1652,7 @@ func TestHostedClusterWatchesEverythingItCreates(t *testing.T) {
 				},
 			},
 			Data: map[string]string{
-				"supported-versions": "{\"versions\":[\"4.20\",\"4.19\",\"4.18\",\"4.17\",\"4.16\",\"4.15\",\"4.14\"]}",
+				"supported-versions": "{\"versions\":[\"4.21\",\"4.20\",\"4.19\",\"4.18\",\"4.17\",\"4.16\",\"4.15\",\"4.14\"]}",
 				"server-version":     "some-fake-server-version",
 			},
 		},
--- a/support/supportedversion/version.go
+++ b/support/supportedversion/version.go
@@ -33,7 +33,7 @@ const (
 // NOTE: The .0 (z release) should be ignored. It's only here to support
 // semver parsing.
 var (
-	LatestSupportedVersion      = semver.MustParse("4.21.0")
+	LatestSupportedVersion      = semver.MustParse("4.22.0")
 	MinSupportedVersion         = semver.MustParse("4.14.0")
 	IBMCloudMinSupportedVersion = semver.MustParse("4.14.0")
 )
@@ -48,6 +48,7 @@ var ocpVersionToKubeVersion = map[string]semver.Version{
 	"4.18.0": semver.MustParse("1.31.0"),
 	"4.19.0": semver.MustParse("1.32.0"),
 	"4.20.0": semver.MustParse("1.33.0"),
+	"4.21.0": semver.MustParse("1.34.0"),
 }
 
 func GetKubeVersionForSupportedVersion(supportedVersion semver.Version) (*semver.Version, error) {
--- a/support/supportedversion/version_test.go
+++ b/support/supportedversion/version_test.go
@@ -25,7 +25,7 @@ import (
 
 func TestSupportedVersions(t *testing.T) {
 	g := NewGomegaWithT(t)
-	g.Expect(Supported()).To(Equal([]string{"4.21", "4.20", "4.19", "4.18", "4.17", "4.16", "4.15", "4.14"}))
+	g.Expect(Supported()).To(Equal([]string{"4.22", "4.21", "4.20", "4.19", "4.18", "4.17", "4.16", "4.15", "4.14"}))
 }
 
 func TestIsValidReleaseVersion(t *testing.T) {
@@ -59,7 +59,7 @@ func TestIsValidReleaseVersion(t *testing.T) {
 				Major: LatestSupportedVersion.Major,
 				Minor: LatestSupportedVersion.Minor + 1,
 			},
-			latestVersionSupported: v("4.20.0"),
+			latestVersionSupported: v("4.21.0"),
 			minVersionSupported:    v("4.14.0"),
 			expectError:            true,
 			platform:               hyperv1.NonePlatform,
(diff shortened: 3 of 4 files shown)


jparrill:
/hold

Until branching happens this friday

sdminonne:
As already mentioned in PM, we may need another pair of eyes for this one but. to me...

hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
```diff
@@ -1651,7 +1651,7 @@ func TestHostedClusterWatchesEverythingItCreates(t *testing.T) {
 				},
 			},
 			Data: map[string]string{
-				"supported-versions": "{\"versions\":[\"4.20\",\"4.19\",\"4.18\",\"4.17\",\"4.16\",\"4.15\",\"4.14\"]}",
+				"supported-versions": "{\"versions\":[\"4.20\",\"4.19\",\"4.18\",\"4.17\",\"4.16\",\"4.15\"]}",
```
devguyio: 1. why not add 4.21?
2. if we're keeping IBM min supported version to 4.14, should we still remove 4.14 from here?

hypershift-operator/controllers/nodepool/nodepool_controller_test.go
```diff
@@ -1825,22 +1825,22 @@ func TestSupportedVersionSkewCondition(t *testing.T) {
...
 			releaseProvider: &fakereleaseprovider.FakeReleaseProvider{
-				Version: "4.14.0",
+				Version: "4.16.0",
 			},
 			expectedCondition: &hyperv1.NodePoolCondition{
 				Type:               hyperv1.NodePoolSupportedVersionSkewConditionType,
 				Status:             corev1.ConditionFalse,
 				Reason:             hyperv1.NodePoolUnsupportedSkewReason,
-				Message:            "NodePool minor version 4.14 is less than 4.15, which is the minimum NodePool version compatible with the 4.18 HostedCluster",
+				Message:            "NodePool minor version 4.16 is less than 4.17, which is the minimum NodePool version compatible with the 4.20 HostedCluster",
```
devguyio: shouldn't this be nodePool set to 4.18 and hostedcluster set to 4.22 ? or 4.21 if it must be an existing release? I didn't look deeply but the name of the test is giving me that indication

support/supportedversion/version_test.go
```diff
@@ -24,7 +24,7 @@ import (
 
 func TestSupportedVersions(t *testing.T) {
 	g := NewGomegaWithT(t)
-	g.Expect(Supported()).To(Equal([]string{"4.21", "4.20", "4.19", "4.18", "4.17", "4.16", "4.15", "4.14"}))
+	g.Expect(Supported()).To(Equal([]string{"4.22", "4.21", "4.20", "4.19", "4.18", "4.17", "4.16", "4.15"}))
```
devguyio: same comment about keeping IBM version at 4.14, should we keep 4.14 here still?

test/e2e/util/version.go
```diff
@@ -15,14 +15,14 @@ import (
...
 	// y-stream versions supported by e2e in main
+	Version422 = semver.MustParse("4.22.0")
 	Version421 = semver.MustParse("4.21.0")
 	Version420 = semver.MustParse("4.20.0")
 	Version419 = semver.MustParse("4.19.0")
 	Version418 = semver.MustParse("4.18.0")
 	Version417 = semver.MustParse("4.17.0")
 	Version416 = semver.MustParse("4.16.0")
 	Version415 = semver.MustParse("4.15.0")
-	Version414 = semver.MustParse("4.14.0")
```
devguyio, line 25: should we leave it if we're keeping IBM at 4.14?

devguyio (changes requested):
Thanks for doing this!

Left a few comments

support/supportedversion/version_test.go
jparrill: humm I would say no, talking yesterday with Cesar he mentioned that we should report 4.15 as minimum version supported for all the customers but keep 4.14 for IBM, in order to let the take the initiative to put a PR. w/e the case I will talk in the IBM sync meeting this Thursday to make sure they know we are following this path or if they agree remove the 4.14 completely.

test/e2e/util/version.go
jparrill: I don't think so, we don't have any reference to this part of the code and I don't think we will add it tbh.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
jparrill: Adding it or not it will not cause a difference, this is just a fixed test.

devguyio: I'm a little confused, so why are we removing the 4.14 if it won't make a difference?

support/supportedversion/version.go
```diff
@@ -32,8 +32,8 @@ const (
 // NOTE: The .0 (z release) should be ignored. It's only here to support
 // semver parsing.
 var (
-	LatestSupportedVersion      = semver.MustParse("4.21.0")
-	MinSupportedVersion         = semver.MustParse("4.14.0")
+	LatestSupportedVersion      = semver.MustParse("4.22.0")
+	MinSupportedVersion         = semver.MustParse("4.15.0")
```
devguyio: This might break IBM clusters if they're using the supported versions CM, no?

jparrill: It should not, the function `GetMinSupportedVersion` contains a switch which select per platform, in this case we return another minimum for IBM which in this case is `IBMCloudMinSupportedVersion = semver.MustParse("4.14.0")`.

```
    switch hc.Spec.Platform.Type {
	case hyperv1.IBMCloudPlatform:
		return IBMCloudMinSupportedVersion
	default:
		return defaultMinVersion
	}
```

So I would say that would not happen.

sjenning:
Something wrong with loki

sjenning:
/hold

We need to remove support for 4.14 as a separate step.  There are CI complications doing it all at once.

See comment here
https://github.com/openshift/release/pull/72122#discussion_r2615968255

PR #7320, opened 2025-12-02
CNTRLPLANE-2152: chore: override capi images

muraee:

This is needed to unblock capi rebass to v1.11 which introduces a new API version v1beta2. Without this, controller can't run as Hypershift doesn't install the new API version yet.

Fixes 

## Checklist:

### DIFF
--- a/hypershift-operator/controllers/hostedcluster/internal/platform/platform.go
+++ b/hypershift-operator/controllers/hostedcluster/internal/platform/platform.go
@@ -105,6 +105,13 @@ func GetPlatform(ctx context.Context, hcluster *hyperv1.HostedCluster, releasePr
 				return nil, fmt.Errorf("failed to fetch payload version: %w", err)
 			}
 		}
+
+		// temporary override for 4.21 to unblock CAPI bump to 1.11 which introduces a new API version.
+		// used image from multiArch release payload 4.20.5 (quay.io/openshift-release-dev/ocp-release@sha256:1f2c28ac126453a3b9e83b349822b9f1fb7662973a212f936b90fdc40e06eb58)
+		// TODO(https://issues.redhat.com/browse/CNTRLPLANE-1200): Remove this override once Hypershift installs the CAPI v1beta2 API version
+		if payloadVersion != nil && payloadVersion.Major == 4 && payloadVersion.Minor == 21 {
+			capiImageProvider = "quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:b8881eafb5fdb31bdaffe03d2cb74ee6729a4776ce879c550e0ea26c489e2cf6"
+		}
 		platform = aws.New(utilitiesImage, capiImageProvider, payloadVersion)
 	case hyperv1.IBMCloudPlatform:
 		platform = &ibmcloud.IBMCloud{}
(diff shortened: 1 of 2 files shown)


hypershift-operator/controllers/hostedcluster/internal/platform/platform.go
```diff
@@ -105,6 +105,12 @@ func GetPlatform(ctx context.Context, hcluster *hyperv1.HostedCluster, releasePr
 				return nil, fmt.Errorf("failed to fetch payload version: %w", err)
 			}
 		}
+
+		// temporary override for 4.21 to unblock CAPI bump to 1.11 which introduces a new API version.
+		// used image from release payload quay.io/openshift-release-dev/ocp-release:4.20.6-x86_64
```
bryan-cox: We should use a multi-arch image for these no?

muraee: yes, I was looking for it but couldn't find it :)
do you have a link?

celebdor:
```
❯ curl -s "https://api.openshift.com/api/upgrades_info/v1/graph?channel=stable-4.20&arch=multi" | \
  jq -r '.nodes | sort_by(.version|split(".")|map(tonumber)) | .[-1] | [.payload, .version]'
[
  "quay.io/openshift-release-dev/ocp-release@sha256:1f2c28ac126453a3b9e83b349822b9f1fb7662973a212f936b90fdc40e06eb58",
  "4.20.5"
]
```

celebdor: Note that 4.20.6 AFAICT has not reached stable

bryan-cox: https://multi.ocp.releases.ci.openshift.org/#4-stable-multi

bryan-cox: 4.20.6 is listed in the stable channel 🤷🏻‍♂️

celebdor: Maybe it's just not in the upgrades?

muraee: thanks, updated

hypershift-operator/controllers/hostedcluster/internal/platform/platform.go
```diff
@@ -105,6 +105,13 @@ func GetPlatform(ctx context.Context, hcluster *hyperv1.HostedCluster, releasePr
...
+
+		// temporary override for 4.21 to unblock CAPI bump to 1.11 which introduces a new API version.
+		// used image from multiArch release payload 4.20.5 (quay.io/openshift-release-dev/ocp-release@sha256:1f2c28ac126453a3b9e83b349822b9f1fb7662973a212f936b90fdc40e06eb58)
+		// TODO(https://issues.redhat.com/browse/CNTRLPLANE-1200): Remove this override once Hypershift installs the CAPI v1beta2 API version
+		if payloadVersion != nil && payloadVersion.Major == 4 && payloadVersion.Minor == 21 {
+			capiImageProvider = "quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:b8881eafb5fdb31bdaffe03d2cb74ee6729a4776ce879c550e0ea26c489e2cf6"
+		}
 		platform = aws.New(utilitiesImage, capiImageProvider, payloadVersion)
 	case hyperv1.IBMCloudPlatform:
 		platform = &ibmcloud.IBMCloud{}
```
damdo, line 117: @muraee I think this will need changing too, we own that as a cluster infra team and have a PR up for it https://github.com/openshift/cluster-api-provider-ibmcloud/pull/127

damdo: As a matter of fact, it might be worth pinning all providers to the same so we don't need to follow up with a separate PR for those once they need to do the same in the following days.

damdo (commented):
TY, left a comment

hypershift-operator/controllers/hostedcluster/internal/platform/platform.go
muraee: ibm deploys their own manifests.

damdo: Concerns addressed here: https://redhat-internal.slack.com/archives/C05KZA3NVU6/p1764694845958369?thread_ts=1759837562.930379&cid=C05KZA3NVU6

LGTM

damdo: Resolved.

damdo (commented):
Thanks 
/lgtm

gaol:
The `quay.io/prometheus/busybox:latest` has been updated hours ago, so the expected digest needs to be updated as well or we should change to other tags -> digest test.

PR #7321, opened 2025-12-02
NO-JIRA: docs(aws): migrate getting-started to STS-based authentication

bryan-cox:

Updates the AWS getting-started documentation to use STS (Security Token Service) credentials instead of static AWS credentials files. This reflects the modern authentication approach for the HyperShift CLI.

Changes include:
- Replace `--aws-creds` flag with `--role-arn` and `--sts-creds` flags
- Add comprehensive section on setting up IAM roles with trust policies
- Include complete IAM policy document with required permissions for EC2, ELB, IAM, Route53, and S3 operations
- Document STS credential generation using `aws sts get-session-token`
- Update cluster destruction command with new authentication flags
- Add notes explaining the STS credential workflow

Fixes 

This is a documentation-only change. No code changes are included.

## Checklist:

### DIFF
--- a/cmd/infra/aws/create_cli_role.go
+++ b/cmd/infra/aws/create_cli_role.go
@@ -115,7 +115,9 @@ const (
 					"iam:DeleteRolePolicy",
 					"iam:UpdateRole",
 					"iam:DeleteOpenIDConnectProvider",
-					"iam:GetRolePolicy"
+					"iam:GetRolePolicy",
+					"iam:ListAttachedRolePolicies",
+					"iam:DetachRolePolicy"
 				],
 				"Resource": "*"
 			},


docs/content/getting-started.md
```diff
@@ -98,25 +100,69 @@ hypershift install \
...
+
+The HyperShift CLI uses AWS STS (Security Token Service) for authentication. You need:
+
+1. An IAM role with the [required permissions](./reference/infrastructure/aws.md) to create cluster infrastructure
+2. Temporary STS credentials to authenticate
+
+Create an IAM role using the HyperShift CLI:
+
+```shell linenums="1"
+AWS_CREDS="$HOME/.aws/credentials"
```
clebs, line 117: These are the credentials from the IAM user created previously right? It might be good to make the distinction between your IAM user (used by AWS CLI) and the one hypershift creates for STS.

docs/content/getting-started.md
```diff
@@ -98,25 +100,69 @@ hypershift install \
...
+The command will output the role ARN. Save this value for use in subsequent commands:
+
+```shell
+ROLE_ARN="arn:aws:iam::123456789012:role/hypershift-cli-role"
+```
+
+!!! note
+
+    The `hypershift create iam cli-role` command creates an IAM role named `hypershift-cli-role` with all
+    required permissions. You can specify a different name using the `--name` flag.
```
clebs: Is this command idem potent and is it safe for many people using the same IAM user? 

Otherwise it might be safer to not have a default name and make the `--name` flag mandatory.

docs/content/getting-started.md
bryan-cox: Done. Added clarifying comments that `AWS_CREDS` refers to your existing personal IAM user credentials (used by AWS CLI), and that HyperShift uses these to create the new role. Also updated the example to show a unique role name.

---

docs/content/getting-started.md
bryan-cox: The command is idempotent - it checks if the role exists and reuses it, updating the permissions policy if needed. However, you're right about the multi-user concern: the trust policy is tied to the IAM identity that created the role (via `GetCallerIdentity`), so sharing a role between different IAM users won't work.

Updated the docs to:
1. Make `--name` explicit in the example with a user-specific name
2. Add an important note explaining that each user should create their own uniquely-named role
3. Document the idempotent behavior in the note

---

clebs:
/lgtm

We should have a followup story to change the CLI behavior to not have a default the role name and make `--name` required.
That would greatly reduce the risk of the multiple user problem (One would have to explicitly use someone else's name).

PR #7322, opened 2025-12-03
NO-JIRA: Update expected digest for busybox:latest in TestGetDigest

patjlm:

This PR fixes a flaky test failure in `TestGetDigest` caused by an outdated expected digest for `quay.io/prometheus/busybox:latest`.

The test was using a hardcoded digest that became stale when the `:latest` image was updated. Since `:latest` is a mutable tag, the underlying image can change at any time, causing the digest to change.

**This is a quick fix. See PR #7323 for a more robust alternative.**

**Changes:**
- Updated the expected digest in the test case "Image not present in overridden registry, falling back to original imageRef"
- Changed from: `sha256:dfa54ef35e438b9e71ac5549159074576b6382f95ce1a434088e05fd6b730bc4`
- Changed to: `sha256:fe9080e7b47dd0751812e64eb2510b0444b3706fbbe426c4e9372dd8a77141ee`

**Trade-offs:**
- ✅ Minimal change - only updates one digest value
- ❌ Will break again when upstream updates the `:latest` image
- ❌ Requires periodic maintenance

**Comparison with PR #7323:**
- **This PR**: Quick fix, updates hardcoded digest (will need updates in the future)
- **PR #7323**: Uses digest reference instead of `:latest` tag (permanent fix)

Fixes flaky unit test in `support/util/imagemetadata_test.go`

**Note**: Only ONE of this PR or PR #7323 should be merged, not both.

- This is a minimal change - only updates the hardcoded digest to match current upstream image
- The new digest was verified by running `skopeo inspect docker://quay.io/prometheus/busybox:latest`
- Test passes locally after this change
- Choose this PR if you prefer a quick minimal fix, or choose PR #7323 for a permanent solution

## Checklist:

### DIFF
--- a/support/util/imagemetadata_test.go
+++ b/support/util/imagemetadata_test.go
@@ -236,7 +236,7 @@ func TestGetDigest(t *testing.T) {
 			pullSecret:     pullSecret,
 			expectedErr:    false,
 			validateCache:  true,
-			expectedDigest: "sha256:dfa54ef35e438b9e71ac5549159074576b6382f95ce1a434088e05fd6b730bc4",
+			expectedDigest: "sha256:fe9080e7b47dd0751812e64eb2510b0444b3706fbbe426c4e9372dd8a77141ee",
 		},
 	}
 


patjlm:
I think we could also avoid using `:latest` and rather use a direct sha256 ref:
```go
{
      name:           "Image not present in overridden registry, falling back to original imageRef",
      imageRef:       "quay.io/prometheus/busybox@sha256:fe9080e7b47dd0751812e64eb2510b0444b3706fbbe426c4e9372dd8a77141ee",
      pullSecret:     pullSecret,
      expectedErr:    false,
      validateCache:  true,
      expectedDigest: "sha256:fe9080e7b47dd0751812e64eb2510b0444b3706fbbe426c4e9372dd8a77141ee",
  },
```

WDYT @muraee ?

support/util/imagemetadata_test.go
patjlm, line 235:
```suggestion
			imageRef:       "quay.io/prometheus/busybox@sha256:fe9080e7b47dd0751812e64eb2510b0444b3706fbbe426c4e9372dd8a77141ee",
```

support/util/imagemetadata_test.go
```diff
@@ -236,7 +236,7 @@ func TestGetDigest(t *testing.T) {
 			pullSecret:     pullSecret,
 			expectedErr:    false,
 			validateCache:  true,
-			expectedDigest: "sha256:dfa54ef35e438b9e71ac5549159074576b6382f95ce1a434088e05fd6b730bc4",
+			expectedDigest: "sha256:fe9080e7b47dd0751812e64eb2510b0444b3706fbbe426c4e9372dd8a77141ee",
```
jparrill, line 239: Hey Patrick, I don't know how urgent is this for you, I assume pretty much. For me this is a quick patch that maybe for the urgency, I think make sense to implement now, but this could be solved in 2 ways:
- Modify the test to use a mocked response (It's a unit test and should not need communication with the final registry)
- Put that test under test/integration folder and keep it as it is.

As I said, if this is urgent I will tag it. If not, do you mind update the PR have a generated Mocked response?

patjlm: It's currently blocking all pull requests , so id say it's pretty urgent. However I can try having a look at these suggestions in a subsequent PR if you want

PR #7324, opened 2025-12-03
CNTRLPLANE-1768: feat(api): add support for graceful service account signing key rotation 

muraee:

Add documentation and constant for the optional "old-key.pub" secret key
in the ServiceAccountSigningKey secret reference. This key allows users
to provide the previous signing key's public key during key rotation,
enabling the kube-apiserver to validate tokens signed with either the
current or previous key.

Changes:
- Add ServiceAccountOldPublicKeySecretKey constant ("old-key.pub")
- Update ServiceAccountSigningKey field documentation to describe
  the key rotation workflow

Fixes 

## Checklist:

### DIFF
--- a/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
+++ b/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
@@ -4464,28 +4464,9 @@ func (r *HostedClusterReconciler) validateServiceAccountSigningKey(ctx context.C
 		return fmt.Errorf("the IssuerURL must be set when specifying a service account signing key")
 	}
 
-	privateBytes, _, err := r.serviceAccountSigningKeyBytes(ctx, hc)
-	if err != nil {
-		return err
-	}
-	controlPlaneNamespace := manifests.HostedControlPlaneNamespace(hc.Namespace, hc.Name)
-	cpSigningKeySecret := controlplaneoperator.ServiceAccountSigningKeySecret(controlPlaneNamespace)
-	if err := r.Client.Get(ctx, client.ObjectKeyFromObject(cpSigningKeySecret), cpSigningKeySecret); err != nil {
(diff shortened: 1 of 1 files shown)


muraee:
depends on https://github.com/openshift/hypershift/pull/6874

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -4425,10 +4416,11 @@ func (r *HostedClusterReconciler) reconcileServiceAccountSigningKey(ctx context.
 	}
 	cpSigningKeySecret := controlplaneoperator.ServiceAccountSigningKeySecret(targetNamespace)
 	_, err = createOrUpdate(ctx, r.Client, cpSigningKeySecret, func() error {
-		// Only set the signing key when the key does not already exist
-		if _, hasKey := cpSigningKeySecret.Data[controlplaneoperator.ServiceSignerPrivateKey]; hasKey {
+		existingKeyBytes, hasKey := cpSigningKeySecret.Data[controlplaneoperator.ServiceSignerPrivateKey]
```
csrwng: would we require that a new signing key secret be referenced? If not, then I would think we want to watch the referenced secret so we can react to changes to it.

muraee: we already watch all referenced resources after this PR https://github.com/openshift/hypershift/pull/6103, no?

devguyio:
I didn't have time to review this due to the firefighting we had on Friday. I took the liberty tagging it without reviewing since this PR is overdue, it had multiple eyes on it already, and it was verified by QE.

muraee:
/override "Red Hat Konflux / enterprise-contract-mce-211 / hypershift-cli-mce-211"

PR #7335, opened 2025-12-04
AUTOSCALE-473:  Allow role/instanceprofile manipulation in autonode via OpenShiftEC2NodeClass

jkyros:

- Design documents intended instanceprofile/node to be adjustable by users, but we hadn't added the plumbing to OpenShiftEc2NodeClass to do that yet.

- This just:
  -  adds the fields to OpenShiftEc2NodeClass 
  -  adds the sync plumbing to the controller in the karpenter operator
  - adds a VAP to protect ONLY the default `OpenShiftEC2NodeClass` so the user can get feedback that they can't change it rather than it being silently discarded
  - adds an e2e test to enforce the instanceProfile and Role functionality (using `Role` results in auto-creation of an instaceProfile by Karpenter) 
- This comes with the caveat that if `Role` is used, and karpenter auto-generates instance profiles (it will), those instances profiles will _not_ be garbage collected automatically until we get to Karpenter 1.7 (we are currently on <= 1.4): https://karpenter.sh/docs/upgrading/upgrade-guide/#upgrading-to-170

Fixes 

## Checklist:

### DIFF
--- a/test/e2e/assets/karpenter-nodepool.yaml
+++ b/test/e2e/assets/karpenter-nodepool.yaml
@@ -4,6 +4,15 @@ metadata:
   name: on-demand
 spec:
   template:
+    metadata:
+      labels:
+        # This label enables the GlobalPullSecret DaemonSet to run on Karpenter-managed nodes.
+        # The GlobalPullSecret feature allows users to add custom registry credentials by creating
+        # an additional-pull-secret in kube-system, which is then merged with the cluster's pull
+        # secret and synced to kubelet on each node.
+        # TODO(jkyros): When OpenshiftNodePool wrapper is implemented, this label should be auto-injected
+        # similar to how OpenshiftEC2NodeClass wraps EC2NodeClass.
+        hypershift.openshift.io/nodepool-globalps-enabled: "true"
     spec:
       requirements:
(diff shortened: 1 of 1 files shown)


jkyros:
hmm the bot didn't make it go
/test e2e-aws-autonode

jkyros:
Autoscale team talked about this in planning -- we think default ec2nodeclass was intended to be "immutable" but any other ec2nodeclass was supposed to sync user changes from their corresponding openshiftec2nodeclass (at least for supported fields). 

While that is a weird UX corner for the default  -- "Yes, you can manipulate the default openshiftec2nodeclass successfully, but we quietly discard your changes" --  the default one being immutable (even via openshiftec2nodeclass) seemed to be the desire. So I'll adjust this accordingly and try to find a decent way via validation or something to convey to the user that "no, the default openshiftec2nodeclass is just for show" :smile:

jkyros:
- Reverted allowing changes to default ec2nodeclass 
- Added VAP to protect default `OpenShiftEC2NodeClass` so it's clear the user shouldn't touch it (it nudges them toward creating a custom one) 
- Added e2e tests to verify this functionality

jkyros:
Closing this, we will come back to it if we want to expose it like this someday

PR #7363, opened 2025-12-10
NO-JIRA: fix(install): skip IPAM CRDs if they already exist

muraee:
## Summary

- Skip installing cluster-api IPAM CRDs (ipaddressclaims, ipaddresses) if they already exist in the cluster
- These CRDs may already be installed by other components (e.g., by CVO) and overwriting them can cause conflicts
- The render command also skips existing IPAM CRDs when a kubeconfig is available

## Test plan

- [ ] Run `hypershift install` on a cluster with existing IPAM CRDs and verify they are skipped
- [ ] Run `hypershift install` on a cluster without IPAM CRDs and verify they are installed
- [ ] Run `hypershift install render` and verify IPAM CRDs are included when no kubeconfig is available
- [ ] Run existing unit tests: `go test ./cmd/install/...`

### DIFF
--- a/cmd/install/install.go
+++ b/cmd/install/install.go
@@ -307,7 +307,12 @@ func InstallHyperShiftOperator(ctx context.Context, out io.Writer, opts Options)
 		return err
 	}
 
-	crds, objects, err := hyperShiftOperatorManifests(opts)
+	client, err := util.GetClient()
+	if err != nil {
+		return err
+	}
+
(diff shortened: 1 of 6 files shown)


devguyio:
Thanks! I'm thinking you probably preferred the detection logic over a flag-based skipping so that existing users of the CLI work out of the box without having to change CI jobs, scripts etc?

muraee:
/override "ci/prow/e2e-aws"
not related

PR #7371, opened 2025-12-11
NO-JIRA: chore(azure): bump AKS Kubernetes version to 1.33.0

bryan-cox:

Updates the AKS cluster setup script to use Kubernetes version 1.33.0, upgrading from 1.31.1.

N/A - version maintenance

Simple version bump for the managed Azure AKS cluster setup script.

## Checklist:


bryan-cox:
/verified bypass

We don't need to verify this.

PR #7375, opened 2025-12-11
feat(etcd): implement multi-shard etcd support

jhjaggars:
## Summary

This PR implements multi-shard etcd support for HyperShift, enabling hosted clusters to distribute etcd data across multiple independent etcd clusters for improved scalability and performance.

### Key Features

- **Configurable etcd sharding**: Support for defining multiple etcd shards with independent resource prefixes
- **Named shards**: Pre-defined shard configurations (main, events, leases)
- **CLI support**: New flags for specifying etcd shard configuration during cluster creation
- **Flexible resource distribution**: Ability to map Kubernetes resource types to specific etcd shards
- **Independent scaling**: Each shard can be configured with different replica counts and storage settings

### Changes

- Implemented etcd shard API types and effective shard resolution
- Added CLI flags for etcd sharding configuration (`--etcd-shard-*`)
- Integrated etcd sharding with control-plane-operator reconciliation
- Updated Kube API Server to connect to multiple etcd endpoints
- Fixed OAuth APIServer NO_PROXY configuration for sharded etcd
- Added comprehensive CLI documentation
- Fixed regression: restored ServiceAccountMaxTokenExpiration field

### Related Issues

This work is part of the etcd sharding initiative to improve hosted cluster scalability.

## Summary by CodeRabbit

* **New Features**
  * Full etcd sharding (managed + unmanaged): per-shard names, prefixes, priority, replicas, storage, backup schedules; CLI flags for file/inline shard config; shard-aware services, StatefulSets, ServiceMonitors, PDBs, init/restore flow, aggregated health reporting, and orphaned-shard cleanup.
* **Documentation**
  * New etcd sharding guide with examples, CLI usage, validation and operational guidance.
* **Tests**
  * Unit tests for shard parsing, reconciliation, status aggregation, cleanup, and API integrations.


api/hypershift/v1beta1/hostedcluster_helpers.go
```diff
@@ -0,0 +1,55 @@
+package v1beta1
+
+import (
+	"k8s.io/utils/ptr"
+)
```
JoelSpeed: Bad practice to include dependencies like this in your API types. Impacts users building projects against your APIs. Prefer to have utils like this somewhere separate and not as methods on the types themselves please

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1574,8 +1574,82 @@ type EtcdSpec struct {
 // HyperShift.
 type ManagedEtcdSpec struct {
 	// storage specifies how etcd data is persisted.
+	// When shards are specified, this serves as the default for all shards
+	// unless overridden per-shard.
```
JoelSpeed, line 1904: Could also make it so that this field cannot be set, forcing a choice in each shard to be explicit

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1574,8 +1574,82 @@ type EtcdSpec struct {
...
 	// +required
 	Storage ManagedEtcdStorageSpec `json:"storage"`
+
+	// shards configures etcd sharding by Kubernetes resource kind.
+	// When not specified, a default single shard accepting all prefixes is used.
+	// When specified, exactly one shard must have "/" in its resourcePrefixes.
+	// +optional
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=10
```
JoelSpeed, line 1921: Any particular decision process to get to 10 as a maximum?

Should document in the godoc that this list accepts at most 10 items

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1574,8 +1574,82 @@ type EtcdSpec struct {
...
+	// +listType=map
+	// +listMapKey=name
+	// +kubebuilder:validation:XValidation:rule="self.exists(s, '/' in s.resourcePrefixes)",message="exactly one shard must have '/' prefix"
```
JoelSpeed, line 1924: The rule itself checks that at least 1 item has `/` in the resourcePrefixes, but the message says `exactly one`, perhaps you want
```suggestion
	// +kubebuilder:validation:XValidation:rule="self.exists_one(s, '/' in s.resourcePrefixes)",message="exactly one shard must have '/' prefix"
```

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1574,8 +1574,82 @@ type EtcdSpec struct {
...
+	// +kubebuilder:validation:XValidation:rule="self.all(s, s.resourcePrefixes.all(p, p == '/' || p.endsWith('#')))",message="non-default prefixes must end with '#'"
```
JoelSpeed, line 1925: Why is this rule here vs on the resourcePrefixes field itself?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1574,8 +1574,82 @@ type EtcdSpec struct {
...
+)
+
+// ManagedEtcdShardSpec defines configuration for a single managed etcd shard
+type ManagedEtcdShardSpec struct {
+	// name is the unique identifier for this shard
+	// Must be DNS-1035 compliant (lowercase alphanumeric + hyphens)
+	// Used for resource naming: etcd-{name}, etcd-{name}-client, etc.
+	// +required
+	// +kubebuilder:validation:Pattern=`^[a-z]([-a-z0-9]*[a-z0-9])?$`
+	// +kubebuilder:validation:MaxLength=15
```
JoelSpeed, line 1948: Why 15? DNS1035 is typically 63 capped. Looks like we are prefixing this, is it always `etcd-` or are we using other prefixes?

JoelSpeed, line 1948: Should add `MinLength=1` too

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1574,8 +1574,82 @@ type EtcdSpec struct {
...
+	EtcdShardPriorityHigh     EtcdShardPriority = "High"
+	EtcdShardPriorityMedium   EtcdShardPriority = "Medium"
+	EtcdShardPriorityLow      EtcdShardPriority = "Low"
+)
+
+// ManagedEtcdShardSpec defines configuration for a single managed etcd shard
+type ManagedEtcdShardSpec struct {
+	// name is the unique identifier for this shard
+	// Must be DNS-1035 compliant (lowercase alphanumeric + hyphens)
+	// Used for resource naming: etcd-{name}, etcd-{name}-client, etc.
```
JoelSpeed, line 1944: For consistency with how we describe this validation on other APIs
```suggestion
	// name is the unique identifier for this shard.
	// Must be at most 15 characters in length and consist only of lowercase
	// alphanumeric characters and hyphens, and must start with an alphabetic
	// character and end with an alphanumeric character.
	// Used for resource naming: etcd-{name}, etcd-{name}-client, etc.
```

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1574,8 +1574,82 @@ type EtcdSpec struct {
...
+	// +required
+	// +kubebuilder:validation:Pattern=`^[a-z]([-a-z0-9]*[a-z0-9])?$`
```
JoelSpeed: Nit, if you use CEL instead, you can provide a more helpful error message that explains the validation. Tend to prefer not to dump regex out for folk in error messages within openshift APIs where possible

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1574,8 +1574,82 @@ type EtcdSpec struct {
...
+	// +kubebuilder:validation:MaxLength=15
+	Name string `json:"name"`
+
+	// resourcePrefixes specifies which Kubernetes resources are stored in this shard
+	// Format: "group/resource#" or "/" for default (catch-all)
+	// Examples: "/events#", "/coordination.k8s.io/leases#", "/"
+	// Exactly one shard must have "/" as a prefix
```
JoelSpeed, line 1954: Formatting here won't be preserved when this is generated into documentation. Prefer to use complete sentences.

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1574,8 +1574,82 @@ type EtcdSpec struct {
...
+	// +required
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=50
+	// +kubebuilder:validation:items:MaxLength=255
+	// +listType=set
+	ResourcePrefixes []string `json:"resourcePrefixes"`
```
JoelSpeed: The rule for checking the end # would make more sense here no?
```suggestion
	// +kubebuilder:validation:XValidation:rule="self.all(x, x == "/" || x.endsWith('#')
	ResourcePrefixes []string `json:"resourcePrefixes"`
```

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1574,8 +1574,82 @@ type EtcdSpec struct {
...
+
+	// priority determines operational importance and default backup frequency
+	// Critical: Default backup every 30 minutes
+	// High: Default backup hourly
+	// Medium/Low: Default backup disabled
+	// +optional
+	// +kubebuilder:default=Medium
+	Priority EtcdShardPriority `json:"priority,omitempty"`
```
JoelSpeed, line 1969: Is this implemented? Do we actually care about this as a feature?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1574,8 +1574,82 @@ type EtcdSpec struct {
...
+	// If not specified, inherits from ManagedEtcdSpec.Storage
+	// +optional
+	Storage *ManagedEtcdStorageSpec `json:"storage,omitempty"`
+
+	// replicas is the number of etcd replicas for this shard
+	// Must be 1 or 3. If not specified, defaults based on cluster's
+	// ControllerAvailabilityPolicy (1 for SingleReplica, 3 for HighlyAvailable)
+	// +optional
+	// +kubebuilder:validation:Enum=1;3
+	Replicas *int32 `json:"replicas,omitempty"`
```
JoelSpeed, line 1981: Defaulting based on another resource is typically considered bad practice. I think this is what those utils were for, where/when is the defaulting called?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1574,8 +1574,82 @@ type EtcdSpec struct {
...
+
+	// backupSchedule is the cron schedule for backups (standard cron format)
+	// If empty, uses priority-based default or disables backups
```
JoelSpeed, line 1984: AFAICT there is no way to have the priority field empty currently (there's a default) so I don't know how I would actually represent "please disable backups" in this API 🤔

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1645,20 +1719,68 @@ type PersistentVolumeEtcdStorageSpec struct {
...
 type UnmanagedEtcdSpec struct {
-	// endpoint is the full etcd cluster client endpoint URL. For example:
-	//
-	//     https://etcd-client:2379
-	//
-	// If the URL uses an HTTPS scheme, the TLS field is required.
-	//
+	// endpoint is the full etcd cluster client endpoint URL.
+	// Used only when shards is not specified (legacy single-etcd mode).
+	// When shards are specified, this field is ignored.
```
JoelSpeed, line 2065: Would be better to add an XValidation to prevent them from both being set at the same time?

api/hypershift/v1beta1/hostedcluster_types.go
JoelSpeed, line 1883: You currently have two different use cases in the unmanaged etcd config - one is what we have today, the other is shards is the only field specified and the rest of the struct must be empty - Have you considered adding a new value to this enum and having a separate union member for sharded unmanaged etcd?

```
managementType: ShardedUnmanagedEtcd
shardedUnmanagedEtcd:
  shards: []
```

cmd/cluster/core/create.go
```diff
@@ -91,6 +91,8 @@ func bindCoreOptions(opts *RawCreateOptions, flags *pflag.FlagSet) {
 	flags.BoolVar(&opts.GenerateSSH, "generate-ssh", opts.GenerateSSH, "If true, generate SSH keys")
 	flags.StringVar(&opts.EtcdStorageClass, "etcd-storage-class", opts.EtcdStorageClass, "The persistent volume storage class for etcd data volumes")
 	flags.StringVar(&opts.EtcdStorageSize, "etcd-storage-size", opts.EtcdStorageSize, "The storage size for etcd data volume. Example: 8Gi")
+	flags.StringVar(&opts.EtcdShardingConfig, "etcd-sharding-config", opts.EtcdShardingConfig, "Path to YAML/JSON file containing etcd sharding configuration. Mutually exclusive with --etcd-shard.")
```
JoelSpeed, line 94: How would I know what the format is for the file I need to pass here? Is there precedence for this kind of flag in this cli?

Appears this is for managed etcd sharding, but what about unmanaged? Would it make sense to be able to configure unmanaged here?

cmd/cluster/core/create.go
```diff
@@ -1159,3 +1192,269 @@ func validateVersion(ctx context.Context, versionCLI string, client crclient.Cli
...
+	}
+	if !hasPrefixes {
+		return shard, fmt.Errorf("prefixes is required")
+	}
+
+	return shard, nil
+}
+
+// applyGlobalDefaults applies global etcd configuration defaults to shards that don't have specific values.
+func applyGlobalDefaults(shards []hyperv1.ManagedEtcdShardSpec, opts *CreateOptions) {
```
JoelSpeed, line 1369: This function name doesn't feel like it will age well, not linked to etcd shards in any way

control-plane-operator/controllers/hostedcontrolplane/etcd/cleanup.go
```diff
@@ -0,0 +1,101 @@
...
+
+	appsv1 "k8s.io/api/apps/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/util/sets"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// CleanupOrphanedShards removes resources for shards that are no longer in the spec
+// This is important when shards are removed from the configuration to prevent resource leaks
```
JoelSpeed, line 19: Should shards be allowed to be removed? At the moment, do we know what happens if you removed a shard? We have no migration strategy to move data back from one etcd to another right? I assume removing a shard means you lose all data from that shard

control-plane-operator/controllers/hostedcontrolplane/etcd/cleanup.go
```diff
@@ -0,0 +1,101 @@
...
+func CleanupOrphanedShards(
+	ctx context.Context,
+	hcp *hyperv1.HostedControlPlane,
+	activeShards []hyperv1.ManagedEtcdShardSpec,
+	client client.Client,
+) error {
+	// Get all etcd StatefulSets in namespace
+	stsList := &appsv1.StatefulSetList{}
+	if err := client.List(ctx, stsList); err != nil {
```
JoelSpeed, line 28: Use a list option here to list only statefulsets in the desired namespace

control-plane-operator/controllers/hostedcontrolplane/etcd/cleanup.go
```diff
@@ -0,0 +1,101 @@
...
+		return err
+	}
+
+	// Filter for etcd StatefulSets in the HCP namespace
+	// Check both metadata labels and name prefix since old v2 etcd may not have the label
```
JoelSpeed, line 33: Was the label added in this PR? Any reference to what an old v2 etcd deployment looks like vs whatever we are on now?

control-plane-operator/controllers/hostedcontrolplane/etcd/cleanup.go
```diff
@@ -0,0 +1,101 @@
...
+	if resourceName == "etcd" {
+		return "default"
+	}
+	return strings.TrimPrefix(resourceName, "etcd-")
+}
+
+// resourceNameForShard generates resource names for etcd shards
+// For backward compatibility, the default shard uses the base name without suffix
+// Named shards use the pattern: baseName-shardName
+func resourceNameForShard(baseName, shardName string) string {
```
JoelSpeed, line 96: Basename is always `etcd`? Do we really need to pass it as an argument? If we do, shouldn't extractShardName be consistent and take a basename argument?

control-plane-operator/controllers/hostedcontrolplane/etcd/pdb.go
```diff
@@ -0,0 +1,34 @@
...
+// This function modifies the PDB in place and should be used with createOrUpdate pattern
+func ReconcilePodDisruptionBudget(
+	pdb *policyv1.PodDisruptionBudget,
+	hcp *hyperv1.HostedControlPlane,
+	shard hyperv1.ManagedEtcdShardSpec,
+) error {
+	// Set minimum available replicas to maintain quorum
+	// For a 3-node etcd cluster, we need at least 2 available (quorum = n/2 + 1)
+	minAvailable := intstr.FromInt(1)
+	pdb.Spec.MinAvailable = &minAvailable
```
JoelSpeed, line 21: The comment and the code do not agree here. Was it supposed to be `MaxUnavailable: 1` maybe?

control-plane-operator/controllers/hostedcontrolplane/etcd/shards.go
JoelSpeed, line 1: A lot of functions in this package are exported, do they need to be? Thinking about the reconcile functions which appear to be used here, but maybe don't need to be exported?

control-plane-operator/controllers/hostedcontrolplane/etcd/shards_test.go
```diff
@@ -0,0 +1,332 @@
...
+
+	// Verify events shard StatefulSet
+	eventsSts := manifests.EtcdStatefulSetForShard(hcp.Namespace, "events")
+	err = fakeClient.Get(ctx, client.ObjectKeyFromObject(eventsSts), eventsSts)
+	if err != nil {
+		t.Errorf("Expected events shard StatefulSet to be created, got error: %v", err)
+	}
+	if eventsSts.Name != "etcd-events" {
+		t.Errorf("Expected events shard StatefulSet name to be 'etcd-events', got: %s", eventsSts.Name)
+	}
```
JoelSpeed, line 180: What does this actually test? L169 is generated a statefulset struct which will have the name populated? Then we do a get, which, won't change anything about the name field, and then we check the name is good still? Not really sure what we are trying to achieve

control-plane-operator/controllers/hostedcontrolplane/etcd/shards_test.go
```diff
@@ -0,0 +1,332 @@
...
+	fakeClient := fake.NewClientBuilder().WithScheme(scheme).WithObjects(orphanedSts).Build()
+
+	err := CleanupOrphanedShards(ctx, hcp, activeShards, fakeClient)
+	if err != nil {
+		t.Fatalf("CleanupOrphanedShards failed: %v", err)
+	}
+
+	// Verify orphaned StatefulSet was deleted
+	err = fakeClient.Get(ctx, client.ObjectKeyFromObject(orphanedSts), orphanedSts)
+	if err == nil {
```
JoelSpeed, line 321: What about other errors? This could pass for any error currently

control-plane-operator/controllers/hostedcontrolplane/etcd/statefulset_test.go
```diff
@@ -0,0 +1,270 @@
...
+		},
+	}
+
+	err := ReconcileStatefulSet(sts, hcp, shard, params)
+	if err != nil {
+		t.Fatalf("ReconcileStatefulSet failed: %v", err)
+	}
+
+	// Verify StatefulSet name includes shard name
+	if sts.Name != "etcd-events" {
```
JoelSpeed, line 161: What does this actually achieve? We set the sts name on L141?

control-plane-operator/controllers/hostedcontrolplane/etcd/status.go
```diff
@@ -0,0 +1,170 @@
...
+	hcp *hyperv1.HostedControlPlane,
+	shards []hyperv1.ManagedEtcdShardSpec,
+	client client.Client,
+	version string,
+) error {
+	cpc := &hyperv1.ControlPlaneComponent{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "etcd",
+			Namespace: hcp.Namespace,
+		},
```
JoelSpeed, line 31: Doing this is redundant since you do a Get with a  key later?

control-plane-operator/controllers/hostedcontrolplane/etcd/status.go
```diff
@@ -0,0 +1,170 @@
...
+				APIVersion: hyperv1.GroupVersion.String(),
+				Kind:       "HostedControlPlane",
+				Name:       hcp.Name,
+				UID:        hcp.UID,
+			},
+		}
+		if err := client.Create(ctx, cpc); err != nil {
+			return err
+		}
+		// Need to Get it again to update status
```
JoelSpeed, line 69: Why? The Create call should dump the response form the API server into the object, so cpc should be up to date before you do this get here

control-plane-operator/controllers/hostedcontrolplane/etcd/status.go
```diff
@@ -0,0 +1,170 @@
...
+		if err := client.Get(ctx, types.NamespacedName{Namespace: hcp.Namespace, Name: "etcd"}, cpc); err != nil {
+			return err
+		}
+	}
+
+	// Update status (works for both newly created and existing resources)
+	now := metav1.Now()
```
JoelSpeed, line 76: You can't assume last transition time is always now, it should only change when the status flicks true to false on the condition. If you keep reconciling this status you'll just keep ticking the time up for no reason

control-plane-operator/controllers/hostedcontrolplane/etcd/status.go
```diff
@@ -0,0 +1,170 @@
...
+			ObservedGeneration: hcp.Generation,
+		},
+		{
+			Type:               "RolloutComplete",
+			Status:             condition.Status,
+			Reason:             condition.Reason,
+			Message:            condition.Message,
+			LastTransitionTime: now,
+			ObservedGeneration: hcp.Generation,
+		},
```
JoelSpeed, line 95: Why would you have two conditions here with the same content? That doesn't seem particularly helpful

control-plane-operator/controllers/hostedcontrolplane/etcd/status.go
```diff
@@ -0,0 +1,170 @@
...
+		{
+			Type:               "RolloutComplete",
+			Status:             condition.Status,
+			Reason:             condition.Reason,
+			Message:            condition.Message,
+			LastTransitionTime: now,
+			ObservedGeneration: hcp.Generation,
```
JoelSpeed, line 94: ObservedGeneration should be the cpc generation no?

control-plane-operator/controllers/hostedcontrolplane/etcd/status.go
```diff
@@ -0,0 +1,170 @@
...
+				messages = append(messages, fmt.Sprintf("shard %s StatefulSet not found", shard.Name))
+				continue
+			}
+			return nil, err
+		}
+
+		requiredReplicas := *sts.Spec.Replicas
+		readyReplicas := sts.Status.ReadyReplicas
+
+		// Critical shards need quorum
```
JoelSpeed, line 125: All shards should need quorum no?

control-plane-operator/controllers/hostedcontrolplane/etcd/status.go
```diff
@@ -0,0 +1,170 @@
...
+
+		if readyReplicas < requiredReplicas {
+			messages = append(messages, fmt.Sprintf("shard %s: %d/%d replicas ready", shard.Name, readyReplicas, requiredReplicas))
+		}
+	}
+
+	// All CRITICAL shards must have quorum
+	if criticalTotal > 0 && criticalReady < criticalTotal/2+1 {
+		return &metav1.Condition{
+			Type:    string(hyperv1.EtcdAvailable),
```
JoelSpeed, line 146: This gets thrown away in the parent scope where this function is called, why are we creating other conditions based on this one vs this being the condition we set?

control-plane-operator/controllers/hostedcontrolplane/manifests/etcd.go
```diff
@@ -105,3 +109,68 @@ func EtcdBackupCronJob(hcpNamespace string) *batchv1.CronJob {
 		},
 	}
 }
+
+// resourceNameForShard generates resource names for etcd shards
+// For backward compatibility, the default shard uses the base name without suffix
+// Named shards use the pattern: baseName-shardName
+func resourceNameForShard(baseName, shardName string) string {
```
JoelSpeed, line 195: This is duplicated in at least one other location in this PR, is there a way to avoid the duplication/possibility of drift

control-plane-operator/controllers/hostedcontrolplane/v2/etcd/service.go
```diff
@@ -0,0 +1,69 @@
...
+func adaptClientService(cpContext component.WorkloadContext, svc *corev1.Service) error {
+	hcp := cpContext.HCP
+	managedEtcdSpec := hcp.Spec.Etcd.Managed
+
+	// Use EffectiveShards to get normalized shard configuration
+	shards := managedEtcdSpec.EffectiveShards(hcp)
+	if len(shards) == 0 {
+		return fmt.Errorf("no etcd shards configured")
+	}
+	defaultShard := shards[0]
```
JoelSpeed, line 21: Is this a good assumption to make?

control-plane-operator/controllers/hostedcontrolplane/v2/kas/config.go
```diff
@@ -211,6 +212,23 @@ func generateConfig(p KubeAPIServerConfigParams) (*kcpv1.KubeAPIServerConfig, er
 	args.Set("etcd-keyfile", cpath(etcdClientCertVolumeName, pki.EtcdClientKeyKey))
 	args.Set("etcd-prefix", "kubernetes.io")
 	args.Set("etcd-servers", p.EtcdURL)
+
+	// Build etcd-servers-overrides if shards are configured
+	// Format per Kubernetes docs: /group/resource#url1;url2,/group2/resource2#url3
```
JoelSpeed, line 217: Why do we require the # in the API vs just constructing it here? Seems to be passing complexity to end users vs something that would be relatively straightforward for us to handle?

docs/content/how-to/etcd-sharding/example-3-shard.yaml
```diff
@@ -0,0 +1,65 @@
...
+              size: 8Gi
+              storageClassName: fast-ssd
+
+        - name: events
+          # This shard handles only Events
+          resourcePrefixes:
+            - "/events#"
+          priority: Low
+          replicas: 1
+          # No backupSchedule = no backups for this shard
```
JoelSpeed, line 48: Docs on the field says it falls back to priority based backup?

JoelSpeed:
The changes to [‎control-plane-operator/controllers/hostedcontrolplane/v2/etcd/component.go](https://github.com/openshift/hypershift/pull/7375/changes#diff-d1c31e6eb34486fc420b4f2f55758e04f7ab3561cbd893213d166daf70311171) in this PR don't actually achieve anything since the functions there are not currently called.

Importantly, this means that the defrag bits are not deployed if you are trying to use an HA control plane for testing. Will need to fix this, for now will work around this by deploying the components manually

jhjaggars:
Closing this in favor of a different implementation.

PR #7379, opened 2025-12-12
AUTOSCALE-473: Set default instanceProfile on ec2nodeclasses so things like ECR access work 

jkyros:

This applies the default `instanceProfile` to Karpenter  `ec2nodeclasses`  similar to how the "CAPI nodepools" handle this, and adds e2e test verification that it actually works and the nodes created get the proper roles.  

As for why: 
- We have users who are testing karpenter/autonode that can't access the ECR because the nodes don't have a default `instanceProfile` like the CAPI nodes do.  
- We think https://github.com/openshift/hypershift/pull/7335 requires more discussion and we don't need it yet, we just need to unblock testing, and this won't hurt anybody regardless of what we decide to do over there 

Fixes 

The whole "we will get %s-worker but if we don't, we use "%s-worker-profile" was cooking my brain: https://github.com/openshift/hypershift/blob/382e87ce76b32797717eae673e650d009ae29dff/hypershift-operator/controllers/nodepool/aws.go#L114 but it looks like that fallback never really gets used because the default profile https://github.com/openshift/hypershift/blob/382e87ce76b32797717eae673e650d009ae29dff/cmd/infra/aws/iam.go#L805 gets created, so that's what I went with. 

## Checklist:

### DIFF
--- a/karpenter-operator/controllers/nodeclass/ec2_nodeclass_controller.go
+++ b/karpenter-operator/controllers/nodeclass/ec2_nodeclass_controller.go
@@ -163,7 +163,7 @@ func (r *EC2NodeClassReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	}
 
 	// Use the default instance profile created by IAM (matches cmd/infra/aws/iam.go:DefaultProfileName)
-	instanceProfile := fmt.Sprintf("%s-worker", hcp.Spec.InfraID)
+	instanceProfile := getInstanceProfileName(hcp)
 
 	if _, err := r.CreateOrUpdate(ctx, r.guestClient, ec2NodeClass, func() error {
 		return reconcileEC2NodeClass(ec2NodeClass, openshiftEC2NodeClass, hcp, userDataSecret, instanceProfile)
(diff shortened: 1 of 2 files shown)


karpenter-operator/controllers/nodeclass/ec2_nodeclass_controller.go
```diff
@@ -162,8 +162,11 @@ func (r *EC2NodeClassReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		return ctrl.Result{}, err
 	}
 
+	// Use the default instance profile created by IAM (matches cmd/infra/aws/iam.go:DefaultProfileName)
+	instanceProfile := fmt.Sprintf("%s-worker", hcp.Spec.InfraID)
```
muraee: that would probably not work in ROSA, as they do not use our CLI to create infra.
CS /rosa cli, is creating the infra and they probably have a different name for the default instance profile.

jkyros: Thanks for looking! I added "rosa namegen" to this now. I need...to find a way to test it e2e though... 

We've probably got 4 options for ROSA: 
- What I did here where our name gen matches ROSA name gen until we figure something out
- Snarf it off the CAPI nodepool since we can see that, but we'd have to: 
	- be sure we got the right nodepool, 
	- add RBAC for that and
	- be mindful that since users are testing Cluster Autoscaler to Karpenter migrations, they might mess with the CAPI nodepools 
- Do https://github.com/openshift/hypershift/pull/7335 first to unblock testing since SRE can set it to whatever it needs to be on custom ones and come back for this? 
- Make ROSA create the ec2nodeclass like they do for the day1 CAPI nodepools?

jkyros:
Rebased, e2e-aws-autonode passed (it's in teardown). Pulling this back out of draft mode.

karpenter-operator/controllers/nodeclass/ec2_nodeclass_controller.go
muraee: a simpler solution would be to add an annotation to the HostedCluster resource to set the default instance profile for karpenter. The code would read the  annotation value and set the profile for EC2NodeClasses.
SREs can then set this annotation to whatever they want.

jkyros: Thanks! I put that up over here https://github.com/openshift/hypershift/pull/7392 and I'll close this one.

I'm not sure that was simpler,  but it is easier to test given how we have ROSA/OCM set up. :stuck_out_tongue:

jkyros:
Closing in favor of https://github.com/openshift/hypershift/pull/7392

PR #7382, opened 2025-12-12
OCPBUGS-62760: Pre-pull images to avoid startup delays and flaky e2e

clebs:
This PR pre-pulls images via `InitContainers` to avoid delayed startup of KAS and bootstrap containers, which would cause flakiness in the e2e tests.

[OCPBUGS-62760](https://issues.redhat.com/browse/OCPBUGS-62760)

- Added one init container per regular container.
- Pre-pull init containers are prepended to other init containers to make sure they run at the very start.
- Added an e2e test to verify the above.

## Checklist:

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/v2/kas/deployment.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/v2/kas/deployment.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
-	"slices"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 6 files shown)


clebs:
/uncc @jparrill 

Already selected as approver.

test/e2e/util/util.go
```diff
@@ -4155,3 +4154,61 @@ func ExtractVersionFromReleaseImage(releaseImage string) string {
...
+	// Validate we have one pre-pull init container per regular container image
+	g.Expect(len(prePullInitContainers)).To(Equal(len(regularContainerImages)),
+		"expected one pre-pull init container per regular container image")
+
+	// Validate that pre-pull init containers use images from regular containers
+	for _, prePullContainer := range prePullInitContainers {
+		g.Expect(regularContainerImages[prePullContainer.Image]).To(BeTrue(),
+			"pre-pull init container %s uses image %s which is not in regular containers",
+			prePullContainer.Name, prePullContainer.Image)
+	}
```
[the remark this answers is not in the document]
clebs: The goal is to have one pre-pull container per unique image, not per container, therefore this is the right behavior.

test/e2e/util/util.go
```diff
@@ -4155,3 +4154,61 @@ func ExtractVersionFromReleaseImage(releaseImage string) string {
 	// No known architecture suffix found, return the tag as-is
 	return tag
 }
+
+// EnsureKubeAPIServerPrePullInitContainers validates that the kube-apiserver deployment has
```
sjenning: IMHO, this should be checked by unit tests, not e2e

clebs: Replaced this e2e test with an equivalent unit test. PTAL!

control-plane-operator/controllers/hostedcontrolplane/v2/kas/deployment.go
```diff
@@ -295,6 +300,35 @@ func buildKASAuditWebhookConfigFileVolume(auditWebhookRef *corev1.LocalObjectRef
...
+// the regular containers start, reducing startup time.
+func addImagePrePullInitContainers(podSpec *corev1.PodSpec) {
+	imagesToPrePull := []string{}
+
+	// Collect images from regular containers
+	for _, container := range podSpec.Containers {
+		if container.Image != "" && !slices.Contains(imagesToPrePull, container.Image) {
+			imagesToPrePull = append(imagesToPrePull, container.Image)
+		}
+	}
```
csrwng, line 322: we should remove any images that are already in initContainers from `imagesToPrePull`

clebs: I have added the logic to remove the images from other init containers from the final set of pre-pull containers.

Also added a unit test for it.

PR #7386, opened 2025-12-15
CNTRLPLANE-2214: Migrate STS to AWS SDK v2

LiangquanLi930:
Migrate STS to AWS SDK v2

[CNTRLPLANE-2214](https://issues.redhat.com/browse/CNTRLPLANE-2214)

## Checklist:

### DIFF
--- a/cmd/infra/aws/util/sts.go
+++ b/cmd/infra/aws/util/sts.go
@@ -1,26 +1,47 @@
 package util
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"os"
 
 	supportawsutil "github.com/openshift/hypershift/support/awsutil"
 
+	awsv2 "github.com/aws/aws-sdk-go-v2/aws"
+	awsmiddleware "github.com/aws/aws-sdk-go-v2/aws/middleware"
+	configv2 "github.com/aws/aws-sdk-go-v2/config"
+	v2credentials "github.com/aws/aws-sdk-go-v2/credentials"
+	// AWS SDK v1
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/request"
 	"github.com/aws/aws-sdk-go/aws/session"
+	stsv1 "github.com/aws/aws-sdk-go/service/sts"
+	"github.com/aws/smithy-go/middleware"
 
 	"k8s.io/utils/ptr"
 )
 
 func NewSTSSession(agent, rolenArn, region string, assumeRoleCreds *credentials.Credentials) (*session.Session, error) {
 	assumeRoleSession := session.Must(session.NewSession(aws.NewConfig().WithCredentials(assumeRoleCreds)))
-	creds, err := supportawsutil.AssumeRole(assumeRoleSession, agent, rolenArn)
+
+	// Use v1 STS to assume role
+	stsClient := stsv1.New(assumeRoleSession)
+	assumeRoleOutput, err := stsClient.AssumeRole(&stsv1.AssumeRoleInput{
+		RoleArn:         aws.String(rolenArn),
+		RoleSessionName: aws.String(agent),
+	})
 	if err != nil {
 		return nil, err
 	}
+
+	// Create credentials from assumed role
+	creds := credentials.NewStaticCredentials(
+		*assumeRoleOutput.Credentials.AccessKeyId,
+		*assumeRoleOutput.Credentials.SecretAccessKey,
+		*assumeRoleOutput.Credentials.SessionToken,
+	)
 
 	awsSessionOpts := session.Options{
 		Config: aws.Config{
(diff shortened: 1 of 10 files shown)


support/awsutil/sts.go
```diff
@@ -1,55 +1,58 @@
...
-	resp, err := svc.AssumeRoleWithWebIdentity(input)
+		WebIdentityToken: aws.String(token),
+	})
 	if err != nil {
 		return nil, err
 	}
 	if resp.Credentials == nil {
-		return nil, credentials.ErrStaticCredentialsEmpty
-
+		return nil, ErrEmptyCredentials
```
sdminonne, line 29: @LiangquanLi930 just for my personal understanding. Are Did they remove `ErrStaticCredentialsEmpty` hence we have to define our own?

support/awsutil/sts.go
```diff
@@ -1,55 +1,58 @@
...
+	client := stsv2.NewFromConfig(cfg)
+	resp, err := client.AssumeRole(ctx, &stsv2.AssumeRoleInput{
 		RoleArn:         aws.String(roleArn),
 		RoleSessionName: aws.String(roleSessionName),
-	}
-
-	resp, err := svc.AssumeRole(input)
+	})
 	if err != nil {
 		return nil, err
```
sdminonne: I know that you did not modify this but do you agree to add some context to the `err` here? 
Something like `fmt.Errorf("could not assume role: %v", err)` or similar

support/awsutil/sts.go
LiangquanLi930: Yes — AWS SDK for Go v2 removed the global error variable ErrStaticCredentialsEmpty.

support/awsutil/sts.go
LiangquanLi930: sure

sdminonne:
thanks @LiangquanLi930 !

cmd/infra/aws/util/sts.go
```diff
@@ -1,27 +1,48 @@
...
+	"github.com/aws/smithy-go/middleware"
 
 	"k8s.io/utils/ptr"
 )
 
 func NewSTSSession(agent, rolenArn, region string, assumeRoleCreds *credentials.Credentials) (*session.Session, error) {
 	assumeRoleSession := session.Must(session.NewSession(aws.NewConfig().WithCredentials(assumeRoleCreds)))
-	creds, err := supportawsutil.AssumeRole(assumeRoleSession, agent, rolenArn)
+
+	// Use v1 STS to assume role
```
jparrill, line 29: Why should we still keep the V1 in the code? or we are keeping it until the deprecation comes to a real thing and then deprecate it?

cmd/infra/aws/util/sts.go
```diff
@@ -65,6 +126,27 @@ func ParseSTSCredentialsFile(credentialsFile string) (*credentials.Credentials,
 	return creds, nil
 }
 
+// ParseSTSCredentialsFileV2 parses STS credentials file and returns v2 credentials
```
jparrill, line 134: Same as above, why should we keep the `ParseSTSCredentialsFile` function if we will use the v2?

cmd/infra/aws/util/sts.go
```diff
@@ -40,6 +61,46 @@ func NewSTSSession(agent, rolenArn, region string, assumeRoleCreds *credentials.
 	return awsSession, nil
 }
 
+func NewSTSSessionV2(agent, roleArn, region string, assumeRoleCreds *awsv2.Credentials) (*awsv2.Config, error) {
```
jparrill: Please add `validations` for `roleArn` and agent in case they are empty.

cmd/infra/aws/util/util.go
```diff
@@ -105,6 +142,27 @@ func NewSession(agent, credentialsFile, credKey, credSecretKey, region string) *
 	return awsSession
 }
 
+func NewSessionV2(agent, credentialsFile, credKey, credSecretKey, region string) *awsv2.Config {
+	ctx := context.Background()
```
jparrill: This context should not be inherited from the function which is calling this NewSessionV2 instead of use `context.Background()`?

cmd/infra/aws/util/util.go
```diff
@@ -105,6 +142,27 @@ func NewSession(agent, credentialsFile, credKey, credSecretKey, region string) *
 	return awsSession
 }
 
+func NewSessionV2(agent, credentialsFile, credKey, credSecretKey, region string) *awsv2.Config {
```
jparrill: Could you please add unit tests for this function?

cmd/infra/aws/util/sts.go
```diff
@@ -65,6 +126,27 @@ func ParseSTSCredentialsFile(credentialsFile string) (*credentials.Credentials,
 	return creds, nil
 }
 
+// ParseSTSCredentialsFileV2 parses STS credentials file and returns v2 credentials
+func ParseSTSCredentialsFileV2(credentialsFile string) (*awsv2.Credentials, error) {
```
jparrill, line 135: Could you please add unit tests for this function?

cmd/infra/aws/util/sts.go
```diff
@@ -40,6 +61,46 @@ func NewSTSSession(agent, rolenArn, region string, assumeRoleCreds *credentials.
...
+			Value: *assumedCreds,
+		},
+	))
+	if region != "" {
+		loadOptions = append(loadOptions, configv2.WithRegion(region))
+	}
+	loadOptions = append(loadOptions, configv2.WithAPIOptions([]func(*middleware.Stack) error{
+		awsmiddleware.AddUserAgentKeyValue("openshift.io hypershift", agent),
+	}))
+	finalCfg, err := configv2.LoadDefaultConfig(ctx, loadOptions...)
```
jparrill, line 101: This function is being called twice, is that necessary?

cmd/infra/aws/util/sts.go
LiangquanLi930: Yes, we're keeping v1 code for now because this PR uses an incremental migration strategy.

The `NewSTSSession()` v1 function is still needed because other AWS services (EC2, IAM, Route53, etc.) haven't been migrated to v2 yet and still require v1 sessions. You can see this in files like `cmd/infra/aws/create.go`, `destroy.go`, `create_iam.go`, etc.

cmd/infra/aws/util/sts.go
LiangquanLi930: Same

cmd/infra/aws/util/sts.go
LiangquanLi930: IMHO, both calls are necessary and this follows the same pattern as the v1 `NewSTSSession` function.

The v1 version also creates two sessions:
 - First session with initial credentials to call AssumeRole
 - Final session with the new assumed role credentials + region + user agent

The v2 version preserves this logic - we can't consolidate because each `LoadDefaultConfig` uses different credentials.

jparrill:
/lgtm cancel
Misclick, lgtm should be added by other peer from the team

PR #7393, opened 2025-12-16
OCPBUGS-66211: fix(azure): preserve defaulted values in Infrastructure status

bryan-cox:

Fixes an issue where the Azure platform reconciliation in HyperShift was replacing the entire `AzurePlatformStatus` struct, which inadvertently reverted API-defaulted fields like `CloudLoadBalancerConfig` and `IPFamily` to nil/empty values.

This caused the HCCO to continuously update the Infrastructure resource as the defaulting controller fought with the reconciliation loop, resulting in errors like:
```
Object got updated more than one time without a no-op update, this indicates hypershift incorrectly reverting defaulted values
```

### Changes:
1. Only initialize the `AzurePlatformStatus` struct if it's nil, preserving any defaulted values
2. Read the cloud name from `hcp.Spec.Platform.Azure.Cloud` instead of hardcoding `AzurePublicCloud`, falling back to `AzurePublicCloud` if not specified
3. Only update the specific fields HyperShift explicitly manages (`CloudName` and `ResourceGroupName`)

A part of OCPBUGS-66211

The change is minimal and targeted - instead of replacing the entire struct, we now:
1. Only initialize the struct if it's nil
2. Properly derive CloudName from the HCP spec with a sensible default
3. Update only the fields HyperShift explicitly manages

## Checklist:

### DIFF
--- a/support/globalconfig/infrastructure.go
+++ b/support/globalconfig/infrastructure.go
@@ -82,7 +82,11 @@ func ReconcileInfrastructure(infra *configv1.Infrastructure, hcp *hyperv1.Hosted
 		if infra.Status.PlatformStatus.Azure == nil {
 			infra.Status.PlatformStatus.Azure = &configv1.AzurePlatformStatus{}
 		}
-		infra.Status.PlatformStatus.Azure.CloudName = configv1.AzurePublicCloud
+		cloudName := configv1.AzureCloudEnvironment(hcp.Spec.Platform.Azure.Cloud)
+		if cloudName == "" {
+			cloudName = configv1.AzurePublicCloud
+		}
+		infra.Status.PlatformStatus.Azure.CloudName = cloudName
 		infra.Status.PlatformStatus.Azure.ResourceGroupName = hcp.Spec.Platform.Azure.ResourceGroupName
 	case hyperv1.PowerVSPlatform:
 		infra.Status.PlatformStatus.PowerVS = &configv1.PowerVSPlatformStatus{


support/globalconfig/infrastructure.go
```diff
@@ -79,10 +79,11 @@ func ReconcileInfrastructure(infra *configv1.Infrastructure, hcp *hyperv1.Hosted
...
 	case hyperv1.AzurePlatform:
 		infra.Spec.CloudConfig.Name = "cloud.conf"
-		infra.Status.PlatformStatus.Azure = &configv1.AzurePlatformStatus{
-			CloudName:         configv1.AzurePublicCloud,
-			ResourceGroupName: hcp.Spec.Platform.Azure.ResourceGroupName,
+		if infra.Status.PlatformStatus.Azure == nil {
+			infra.Status.PlatformStatus.Azure = &configv1.AzurePlatformStatus{}
 		}
+		infra.Status.PlatformStatus.Azure.CloudName = configv1.AzurePublicCloud
+		infra.Status.PlatformStatus.Azure.ResourceGroupName = hcp.Spec.Platform.Azure.ResourceGroupName
```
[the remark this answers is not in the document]
bryan-cox: Yeah I'll change that

support/globalconfig/infrastructure.go
```diff
@@ -82,7 +82,11 @@ func ReconcileInfrastructure(infra *configv1.Infrastructure, hcp *hyperv1.Hosted
 		if infra.Status.PlatformStatus.Azure == nil {
 			infra.Status.PlatformStatus.Azure = &configv1.AzurePlatformStatus{}
 		}
-		infra.Status.PlatformStatus.Azure.CloudName = configv1.AzurePublicCloud
+		cloudName := configv1.AzureCloudEnvironment(hcp.Spec.Platform.Azure.Cloud)
```
csrwng, line 85: These are not the same constants. I believe you'll need to add a mapping function.

bryan-cox: The direct type conversion is actually valid here - both types use the same underlying string values. Looking at both APIs:

- **HyperShift API** (`hyperv1.AzurePlatformSpec.Cloud`): `string` with enum `AzurePublicCloud;AzureUSGovernmentCloud;AzureChinaCloud;AzureGermanCloud;AzureStackCloud`
- **configv1 API** (`configv1.AzureCloudEnvironment`): `type AzureCloudEnvironment string` with enum `"";AzurePublicCloud;AzureUSGovernmentCloud;AzureChinaCloud;AzureGermanCloud;AzureStackCloud`

The string values are identical between both APIs. The only difference is configv1 also allows empty string, which we handle on lines 86-88 by defaulting to `AzurePublicCloud`.

Since both are backed by `string` and use the exact same values, the type conversion `configv1.AzureCloudEnvironment(hcp.Spec.Platform.Azure.Cloud)` works correctly.

I did check for an existing mapping function - `GetAzureCloudConfiguration` in `support/azureutil/azureutil.go` maps to Azure SDK's `cloud.Configuration`, not to `configv1.AzureCloudEnvironment`. Happy to add an explicit mapping function if you'd prefer that for clarity, but the current approach is type-safe.

---

PR #7399, opened 2025-12-16
OCPBUGS-69447: feat(updates): enable CVO metrics access with RHOBS monitoring flag

Chee-Lu:
## What this PR does:
When `--rhobs-monitoring=true` is set (for ROSA HCP), enable CVO access to OBO Prometheus for conditional update risk evaluation. 

Aldo add `--cvo-prometheus-url` flag to allow overriding the default Prometheus endpoint. This provides flexibility for future changes (e.g., if ROSA changes the service name) or for platforms with different monitoring architectures (e.g., ARO HCP's self-managed Prometheus). When not specified, platform-appropriate defaults are used.

The CVO deployment logic routes to different metrics endpoints based on the monitoring stack:
- RHOBS stack (ROSA HCP): `http://hypershift-monitoring-stack-prometheus.openshift-observability-operator.svc:9090`
- CoreOS stack (Self-managed HyperShift on OpenShift or ARP HCP): `https://thanos-querier.openshift-monitoring.svc:9092`
- Custom URL: Configurable via `--cvo-prometheus-url` parameter

Key changes:
- CVO deployment enables metrics access when either `--rhobs-monitoring` (for ROSA HCP) or `--enable-cvo-management-cluster-metrics-access` (for self-managed HyperShift on OpenShift or ARO HCP) is set
- Network policies updated to allow egress to the appropriate monitoring endpoint based on stack configuration
- Add `--cvo-prometheus-url` flag to configure CVO Prometheus endpoint

fixes https://issues.redhat.com//browse/OCM-10395
fixes https://issues.redhat.com//browse/OCM-20970

### Backport Requirements
This change should be backported to 4.17.z, 4.18.z, 4.19.z, 4.20.z and 4.21.z to benefit customers on that version.

### DIFF
--- a/cmd/install/install.go
+++ b/cmd/install/install.go
@@ -116,6 +116,7 @@ type Options struct {
 	WaitUntilAvailable                        bool
 	WaitUntilEstablished                      bool
 	RHOBSMonitoring                           bool
+	CVOPrometheusURL                          string
 	SLOsAlerts                                bool
 	MonitoringDashboards                      bool
 	CertRotationScale                         time.Duration
@@ -225,6 +226,10 @@ func (o *Options) Validate() error {
 		errs = append(errs, fmt.Errorf("when invoking this command with the --rhobs-monitoring flag, the --enable-cvo-management-cluster-metrics-access flag is not supported "))
 	}
 
+	if len(o.CVOPrometheusURL) > 0 && !o.RHOBSMonitoring && !o.EnableCVOManagementClusterMetricsAccess {
+		errs = append(errs, fmt.Errorf("--cvo-prometheus-url requires either --rhobs-monitoring or --enable-cvo-management-cluster-metrics-access to be enabled"))
+	}
+
 	if len(o.ManagedService) > 0 && o.ManagedService != hyperv1.AroHCP {
 		errs = append(errs, fmt.Errorf("not a valid managed service type: %s", o.ManagedService))
 	}
@@ -305,7 +310,8 @@ func NewCommand() *cobra.Command {
 	cmd.PersistentFlags().BoolVar(&opts.EnableCVOManagementClusterMetricsAccess, "enable-cvo-management-cluster-metrics-access", opts.EnableCVOManagementClusterMetricsAccess, "If true, the hosted CVO will have access to the management cluster metrics server to evaluate conditional updates (supported for OpenShift management clusters)")
 	cmd.Flags().BoolVar(&opts.WaitUntilAvailable, "wait-until-available", opts.WaitUntilAvailable, "If true, pauses installation until hypershift operator has been rolled out and its webhook service is available (if installing the webhook)")
 	cmd.Flags().BoolVar(&opts.WaitUntilEstablished, "wait-until-crds-established", opts.WaitUntilEstablished, "If true, pauses installation until all custom resource definitions are established before applying other manifests.")
-	cmd.PersistentFlags().BoolVar(&opts.RHOBSMonitoring, "rhobs-monitoring", opts.RHOBSMonitoring, "If true, HyperShift will generate and use the RHOBS version of monitoring resources (ServiceMonitors, PodMonitors, etc)")
+	cmd.PersistentFlags().BoolVar(&opts.RHOBSMonitoring, "rhobs-monitoring", opts.RHOBSMonitoring, "If true, HyperShift will generate and use the RHOBS version of monitoring resources (ServiceMonitors, PodMonitors, etc). For ROSA HCP, this also enables the Cluster Version Operator to query the RHOBS Prometheus for conditional update evaluation. Use --cvo-prometheus-url to override the default Prometheus endpoint.")
+	cmd.PersistentFlags().StringVar(&opts.CVOPrometheusURL, "cvo-prometheus-url", opts.CVOPrometheusURL, "Prometheus URL for the Cluster Version Operator to query metrics for conditional update risk evaluation. Only effective when --rhobs-monitoring or --enable-cvo-management-cluster-metrics-access is enabled. If not specified, defaults to the RHOBS monitoring stack for ROSA HCP, or the Thanos querier for self-managed HyperShift.")
 	cmd.PersistentFlags().BoolVar(&opts.SLOsAlerts, "slos-alerts", opts.SLOsAlerts, "If true, HyperShift will generate and use the prometheus alerts for monitoring HostedCluster and NodePools")
 	cmd.PersistentFlags().BoolVar(&opts.MonitoringDashboards, "monitoring-dashboards", opts.MonitoringDashboards, "If true, HyperShift will generate a monitoring dashboard for every HostedCluster that it creates")
 	cmd.PersistentFlags().DurationVar(&opts.CertRotationScale, "cert-rotation-scale", opts.CertRotationScale, "The scaling factor for certificate rotation. It is not supported to set this to anything other than 24h.")
@@ -899,6 +905,7 @@ func setupOperatorResources(opts Options, userCABundleCM *corev1.ConfigMap, trus
 		IncludeVersion:                          !opts.Template,
 		UWMTelemetry:                            opts.EnableUWMTelemetryRemoteWrite,
 		RHOBSMonitoring:                         opts.RHOBSMonitoring,
+		CVOPrometheusURL:                        opts.CVOPrometheusURL,
 		MonitoringDashboards:                    opts.MonitoringDashboards,
 		CertRotationScale:                       opts.CertRotationScale,
 		EnableCVOManagementClusterMetricsAccess: opts.EnableCVOManagementClusterMetricsAccess,
@@ -1066,6 +1073,7 @@ func setupRBAC(opts Options, operatorNamespace *corev1.Namespace) (*corev1.Servi
 
 	operatorClusterRole := assets.HyperShiftOperatorClusterRole{
 		EnableCVOManagementClusterMetricsAccess: opts.EnableCVOManagementClusterMetricsAccess,
+		RHOBSMonitoring:                         opts.RHOBSMonitoring,
 		ManagedService:                          opts.ManagedService,
 		EnableAuditLogPersistence:               opts.EnableAuditLogPersistence,
 	}.Build()
--- a/control-plane-operator/controllers/hostedcontrolplane/v2/cvo/deployment.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/v2/cvo/deployment.go
@@ -3,15 +3,18 @@ package cvo
 import (
 	"encoding/json"
 	"fmt"
+	"os"
 	"path"
 	"strings"
 
 	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
 	"github.com/openshift/hypershift/control-plane-operator/controllers/hostedcontrolplane/common"
 	hyperapi "github.com/openshift/hypershift/support/api"
+	"github.com/openshift/hypershift/support/awsutil"
 	"github.com/openshift/hypershift/support/capabilities"
 	"github.com/openshift/hypershift/support/config"
 	component "github.com/openshift/hypershift/support/controlplane-component"
+	"github.com/openshift/hypershift/support/rhobsmonitoring"
 	"github.com/openshift/hypershift/support/util"
 
 	configv1 "github.com/openshift/api/config/v1"
@@ -27,11 +30,20 @@ import (
 )
 
 func (cvo *clusterVersionOperator) adaptDeployment(cpContext component.WorkloadContext, deployment *appsv1.Deployment) error {
-	if cvo.enableCVOManagementClusterMetricsAccess {
+	enableMetricsAccess := cvo.isManagementClusterMetricsAccessEnabled(cpContext)
+
+	if enableMetricsAccess {
 		if deployment.Spec.Template.Labels == nil {
 			deployment.Spec.Template.Labels = map[string]string{}
 		}
 		deployment.Spec.Template.Labels[config.NeedMetricsServerAccessLabel] = "true"
+
+		// Set annotation to enable automountServiceAccountToken for metrics endpoint access
+		if deployment.Spec.Template.Annotations == nil {
+			deployment.Spec.Template.Annotations = map[string]string{}
+		}
+		deployment.Spec.Template.Annotations[config.NeedMetricsServerAccessLabel] = "true"
+
 		deployment.Spec.Template.Spec.ServiceAccountName = ComponentName
 	}
 
@@ -93,10 +105,24 @@ func (cvo *clusterVersionOperator) adaptDeployment(cpContext component.WorkloadC
 		if updateService := cpContext.HCP.Spec.UpdateService; updateService != "" {
 			c.Args = append(c.Args, "--update-service", string(updateService))
 		}
-		if cvo.enableCVOManagementClusterMetricsAccess {
+
+		if enableMetricsAccess {
 			c.Args = append(c.Args, "--use-dns-for-services=true")
+
+			// Determine metrics URL based on monitoring stack.
+			cvoPrometheusURL := os.Getenv(config.CVOPrometheusURLEnvVar)
+			if cvoPrometheusURL == "" {
+				if os.Getenv(rhobsmonitoring.EnvironmentVariable) == "1" && awsutil.IsROSAHCP(cpContext.HCP) {
+					// RHOBS monitoring stack (ROSA HCP) - currently uses HTTP without TLS
+					cvoPrometheusURL = "http://hypershift-monitoring-stack-prometheus.openshift-observability-operator.svc:9090"
+				} else {
+					// Self-managed HyperShift - OCP Thanos Query uses HTTPS with service CA
+					cvoPrometheusURL = fmt.Sprintf("https://thanos-querier.openshift-monitoring.svc:9092?namespace=%s", cpContext.HCP.Namespace)
+				}
+			}
+
 			c.Args = append(c.Args, "--metrics-ca-bundle-file=/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt")
-			c.Args = append(c.Args, fmt.Sprintf("--metrics-url=https://thanos-querier.openshift-monitoring.svc:9092?namespace=%s", cpContext.HCP.Namespace))
+			c.Args = append(c.Args, fmt.Sprintf("--metrics-url=%s", cvoPrometheusURL))
 		}
 	})
 
--- a/hypershift-operator/controllers/hostedcluster/network_policies.go
+++ b/hypershift-operator/controllers/hostedcluster/network_policies.go
@@ -5,13 +5,16 @@ import (
 	"fmt"
 	"net"
 	"net/netip"
+	"os"
 
 	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
 	"github.com/openshift/hypershift/hypershift-operator/controllers/manifests"
 	"github.com/openshift/hypershift/hypershift-operator/controllers/manifests/networkpolicy"
 	"github.com/openshift/hypershift/hypershift-operator/controllers/sharedingress"
+	"github.com/openshift/hypershift/support/awsutil"
 	"github.com/openshift/hypershift/support/capabilities"
 	"github.com/openshift/hypershift/support/config"
+	"github.com/openshift/hypershift/support/rhobsmonitoring"
 	"github.com/openshift/hypershift/support/upsert"
 
 	configv1 "github.com/openshift/api/config/v1"
@@ -84,10 +87,12 @@ func (r *HostedClusterReconciler) reconcileNetworkPolicies(ctx context.Context, 
 		}
 
 		// Allow egress communication to the HCP metrics server for pods that have a known annotation.
-		if r.EnableCVOManagementClusterMetricsAccess {
+		// Enable if either RHOBS monitoring is enabled for ROSA HCP or the explicit flag is set.
+		enableMetricsAccess := r.EnableCVOManagementClusterMetricsAccess || (os.Getenv(rhobsmonitoring.EnvironmentVariable) == "1" && awsutil.IsROSAHCP(hcp))
+		if enableMetricsAccess {
 			policy = networkpolicy.MetricsServerNetworkPolicy(controlPlaneNamespaceName)
 			if _, err := createOrUpdate(ctx, r.Client, policy, func() error {
-				return reconcileMetricsServerNetworkPolicy(policy)
+				return reconcileMetricsServerNetworkPolicy(policy, hcp)
 			}); err != nil {
 				return fmt.Errorf("failed to reconcile metrics server network policy: %w", err)
 			}
@@ -831,35 +836,71 @@ func kasEndpointsToCIDRs(kubernetesEndpoint *corev1.Endpoints) []string {
 }
 
 // reconcileMetricsServerNetworkPolicy selects pods having NeedMetricsServerAccessLabel.
-// It allows egress traffic to the HCP metrics server that is available for self-managed
-// HyperShift running on an OpenShift management cluster.
-func reconcileMetricsServerNetworkPolicy(policy *networkingv1.NetworkPolicy) error {
+// It allows egress traffic to the HCP metrics server that is available for self-managed HyperShift and ROSA HCP.
+// Depending on the monitoring stack:
+// - RHOBS (ROSA HCP only): targets OBO Prometheus at port 9090 in openshift-observability-operator namespace
+// - CoreOS (self-managed HyperShift): targets Thanos Querier at port 9092 in openshift-monitoring namespace
+func reconcileMetricsServerNetworkPolicy(policy *networkingv1.NetworkPolicy, hcp *hyperv1.HostedControlPlane) error {
 	protocol := corev1.ProtocolTCP
-	port := intstr.FromInt(9092)
-	policy.Spec.Egress = []networkingv1.NetworkPolicyEgressRule{
-		{
-			To: []networkingv1.NetworkPolicyPeer{
-				{
-					PodSelector: &metav1.LabelSelector{
-						MatchLabels: map[string]string{
-							"app.kubernetes.io/instance": "thanos-querier",
-							"app.kubernetes.io/name":     "thanos-query",
-						}},
-					NamespaceSelector: &metav1.LabelSelector{
-						MatchLabels: map[string]string{
-							"network.openshift.io/policy-group": "monitoring",
+	var egressRules []networkingv1.NetworkPolicyEgressRule
+
+	if os.Getenv(rhobsmonitoring.EnvironmentVariable) == "1" && awsutil.IsROSAHCP(hcp) {
+		// RHOBS Prometheus configuration (ROSA HCP specific)
+		port := intstr.FromInt(9090)
+		egressRules = []networkingv1.NetworkPolicyEgressRule{
+			{
+				To: []networkingv1.NetworkPolicyPeer{
+					{
+						PodSelector: &metav1.LabelSelector{
+							MatchLabels: map[string]string{
+								"app.kubernetes.io/instance": "hypershift-monitoring-stack",
+								"app.kubernetes.io/name":     "prometheus",
+							}},
+						NamespaceSelector: &metav1.LabelSelector{
+							MatchLabels: map[string]string{
+								"network.openshift.io/policy-group": "monitoring",
+							},
 						},
 					},
 				},
-			},
-			Ports: []networkingv1.NetworkPolicyPort{
-				{
-					Protocol: &protocol,
-					Port:     &port,
+				Ports: []networkingv1.NetworkPolicyPort{
+					{
+						Protocol: &protocol,
+						Port:     &port,
+					},
 				},
 			},
-		},
+		}
+	} else {
+		// CoreOS Thanos configuration
+		port := intstr.FromInt(9092)
+		egressRules = []networkingv1.NetworkPolicyEgressRule{
+			{
+				To: []networkingv1.NetworkPolicyPeer{
+					{
+						PodSelector: &metav1.LabelSelector{
+							MatchLabels: map[string]string{
+								"app.kubernetes.io/instance": "thanos-querier",
+								"app.kubernetes.io/name":     "thanos-query",
+							}},
+						NamespaceSelector: &metav1.LabelSelector{
+							MatchLabels: map[string]string{
+								"network.openshift.io/policy-group": "monitoring",
+							},
+						},
+					},
+				},
+				Ports: []networkingv1.NetworkPolicyPort{
+					{
+						Protocol: &protocol,
+						Port:     &port,
+					},
+				},
+			},
+		}
 	}
+
+	policy.Spec.Egress = egressRules
 
 	policy.Spec.PolicyTypes = []networkingv1.PolicyType{networkingv1.PolicyTypeEgress}
 	policy.Spec.PodSelector = metav1.LabelSelector{
--- a/cmd/install/assets/hypershift_operator.go
+++ b/cmd/install/assets/hypershift_operator.go
@@ -432,6 +432,7 @@ type HyperShiftOperatorDeployment struct {
 	IncludeVersion                          bool
 	UWMTelemetry                            bool
 	RHOBSMonitoring                         bool
+	CVOPrometheusURL                        string
 	MonitoringDashboards                    bool
 	CertRotationScale                       time.Duration
 	EnableCVOManagementClusterMetricsAccess bool
@@ -719,6 +720,13 @@ func (o HyperShiftOperatorDeployment) Build() *appsv1.Deployment {
 		envVars = append(envVars, corev1.EnvVar{
 			Name:  rhobsmonitoring.EnvironmentVariable,
 			Value: "1",
+		})
+	}
+
+	if o.CVOPrometheusURL != "" {
+		envVars = append(envVars, corev1.EnvVar{
+			Name:  config.CVOPrometheusURLEnvVar,
+			Value: o.CVOPrometheusURL,
 		})
 	}
 
@@ -1093,6 +1101,7 @@ func (o HyperShiftOperatorServiceAccount) Build() *corev1.ServiceAccount {
 
 type HyperShiftOperatorClusterRole struct {
 	EnableCVOManagementClusterMetricsAccess bool
+	RHOBSMonitoring                         bool
 	ManagedService                          string
 	EnableAuditLogPersistence               bool
 }
@@ -1380,7 +1389,7 @@ func (o HyperShiftOperatorClusterRole) Build() *rbacv1.ClusterRole {
 			},
 		},
 	}
-	if o.EnableCVOManagementClusterMetricsAccess {
+	if o.EnableCVOManagementClusterMetricsAccess || o.RHOBSMonitoring {
 		role.Rules = append(role.Rules,
 			rbacv1.PolicyRule{
 				APIGroups: []string{"metrics.k8s.io"},
--- a/control-plane-operator/controllers/hostedcontrolplane/v2/controlplaneoperator/role.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/v2/controlplaneoperator/role.go
@@ -3,9 +3,11 @@ package controlplaneoperator
 import (
 	"os"
 
+	"github.com/openshift/hypershift/support/awsutil"
 	"github.com/openshift/hypershift/support/azureutil"
 	"github.com/openshift/hypershift/support/config"
 	component "github.com/openshift/hypershift/support/controlplane-component"
+	"github.com/openshift/hypershift/support/rhobsmonitoring"
 	"github.com/openshift/hypershift/support/util"
 
 	rbacv1 "k8s.io/api/rbac/v1"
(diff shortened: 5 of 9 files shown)


sdminonne:
@coderabbitai review

cmd/install/install.go
```diff
@@ -269,7 +269,7 @@ func NewCommand() *cobra.Command {
...
 	cmd.Flags().BoolVar(&opts.WaitUntilAvailable, "wait-until-available", opts.WaitUntilAvailable, "If true, pauses installation until hypershift operator has been rolled out and its webhook service is available (if installing the webhook)")
 	cmd.Flags().BoolVar(&opts.WaitUntilEstablished, "wait-until-crds-established", opts.WaitUntilEstablished, "If true, pauses installation until all custom resource definitions are established before applying other manifests.")
-	cmd.PersistentFlags().BoolVar(&opts.RHOBSMonitoring, "rhobs-monitoring", opts.RHOBSMonitoring, "If true, HyperShift will generate and use the RHOBS version of monitoring resources (ServiceMonitors, PodMonitors, etc)")
+	cmd.PersistentFlags().BoolVar(&opts.RHOBSMonitoring, "rhobs-monitoring", opts.RHOBSMonitoring, "If true, HyperShift will generate and use the RHOBS version of monitoring resources (ServiceMonitors, PodMonitors, etc) and enable CVO access to OBO Prometheus for conditional update risk evaluation")
```
simonpasquier: OBO is an internal code name.
```suggestion
	cmd.PersistentFlags().BoolVar(&opts.RHOBSMonitoring, "rhobs-monitoring", opts.RHOBSMonitoring, "If true, HyperShift will generate and use the RHOBS version of monitoring resources (ServiceMonitors, PodMonitors, etc) and enable the Cluster Version Operator to query  Prometheus for conditional update risk evaluation")
```

control-plane-operator/controllers/hostedcontrolplane/v2/cvo/deployment.go
```diff
@@ -93,10 +98,24 @@ func (cvo *clusterVersionOperator) adaptDeployment(cpContext component.WorkloadC
...
-			c.Args = append(c.Args, "--metrics-ca-bundle-file=/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt")
-			c.Args = append(c.Args, fmt.Sprintf("--metrics-url=https://thanos-querier.openshift-monitoring.svc:9092?namespace=%s", cpContext.HCP.Namespace))
+
+			// Configure metrics endpoint based on monitoring stack
+			var metricsURL string
+			if os.Getenv(rhobsmonitoring.EnvironmentVariable) == "1" {
+				// RHOBS Prometheus uses HTTP without TLS
+				metricsURL = fmt.Sprintf("http://hypershift-monitoring-stack-prometheus.openshift-observability-operator.svc:9090?namespace=%s", cpContext.HCP.Namespace)
+			} else {
+				// CoreOS Thanos uses HTTPS with service CA
```
simonpasquier:
```suggestion
				// OCP Thanos Query uses HTTPS with service CA
```

cmd/install/install.go
Chee-Lu: Thanks for the feedback, I've replace internal code names with public terminology

control-plane-operator/controllers/hostedcontrolplane/v2/cvo/deployment.go
Chee-Lu: Updated!

control-plane-operator/controllers/hostedcontrolplane/v2/cvo/deployment.go
```diff
@@ -93,10 +98,24 @@ func (cvo *clusterVersionOperator) adaptDeployment(cpContext component.WorkloadC
...
+		if enableMetricsAccess {
 			c.Args = append(c.Args, "--use-dns-for-services=true")
-			c.Args = append(c.Args, "--metrics-ca-bundle-file=/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt")
-			c.Args = append(c.Args, fmt.Sprintf("--metrics-url=https://thanos-querier.openshift-monitoring.svc:9092?namespace=%s", cpContext.HCP.Namespace))
+
+			// Configure metrics endpoint based on monitoring stack
+			var metricsURL string
+			if os.Getenv(rhobsmonitoring.EnvironmentVariable) == "1" {
+				// RHOBS Prometheus uses HTTP without TLS
+				metricsURL = fmt.Sprintf("http://hypershift-monitoring-stack-prometheus.openshift-observability-operator.svc:9090?namespace=%s", cpContext.HCP.Namespace)
```
simonpasquier: I'm not sure that in ROSA land, the monitoring stack honors the `namespace` parameter (in fact I'd bet that it doesn't). Also hardcoding the URL might not work for other environments (e.g. GCP/ARO).

Chee-Lu: Thanks for the feedback!

> Also hardcoding the URL might not work for other environments (e.g. GCP/ARO).

ARO and GCP have separate tickets for their implementations([Ticket for ARO](https://issues.redhat.com/browse/ARO-22519)). 

How should I scope this to ROSA only while leaving room for ARO/GCP teams to add their platform-specific configurations later? Or do you have other suggestions on how to handle this?

simonpasquier: > How should I scope this to ROSA only while leaving room for ARO/GCP teams to add their platform-specific configurations later? Or do you have other suggestions on how to handle this?

I'm not an HyperShift engineer so I can't really tell what's the best option. But anyway I've got bigger concerns about the design itself as I'm not sure that it will work as expected...

control-plane-operator/controllers/hostedcontrolplane/v2/cvo/deployment.go
```diff
@@ -93,10 +98,24 @@ func (cvo *clusterVersionOperator) adaptDeployment(cpContext component.WorkloadC
 		if updateService := cpContext.HCP.Spec.UpdateService; updateService != "" {
 			c.Args = append(c.Args, "--update-service", string(updateService))
 		}
-		if cvo.enableCVOManagementClusterMetricsAccess {
+
+		// Enable CVO metrics access if either RHOBS monitoring is enabled or the explicit flag is set
+		enableMetricsAccess := os.Getenv(rhobsmonitoring.EnvironmentVariable) == "1" || cvo.enableCVOManagementClusterMetricsAccess
```
LiangquanLi930: This code has already been defined on line 33.

control-plane-operator/controllers/hostedcontrolplane/v2/cvo/deployment.go
LiangquanLi930: We can determine this using a check like the following:
```
func isROSAHCP(hc *hyperv1.HostedCluster) bool {
	if hc.Spec.Platform.AWS == nil {
		return false
	}

	for _, tag := range hc.Spec.Platform.AWS.ResourceTags {
		if tag.Key == "red-hat-managed" && tag.Value == "true" {
			return true
		}
	}
	return false
}

```

simonpasquier: And about the URL, my gut feeling is that it isn't an HyperShift concern but rather something that would be provided via configuration (e.g what if ROSA decides to change the service name?).

control-plane-operator/controllers/hostedcontrolplane/v2/cvo/deployment.go
Chee-Lu: Removed

control-plane-operator/controllers/hostedcontrolplane/v2/cvo/deployment.go
Chee-Lu: Thanks, added the checking to CVO and network policies. Scoped the changes to ROSA only.

hypershift-operator/controllers/hostedcluster/network_policies.go
```diff
@@ -826,36 +843,72 @@ func kasEndpointsToCIDRs(kubernetesEndpoint *corev1.Endpoints) []string {
...
 
 // reconcileMetricsServerNetworkPolicy selects pods having NeedMetricsServerAccessLabel.
-// It allows egress traffic to the HCP metrics server that is available for self-managed
-// HyperShift running on an OpenShift management cluster.
-func reconcileMetricsServerNetworkPolicy(policy *networkingv1.NetworkPolicy) error {
+// It allows egress traffic to the HCP metrics server that is available for self-managed HyperShift running on an OpenShift management cluster.
+// Depending on the monitoring stack:
+// - RHOBS (ROSA HCP only): targets OBO Prometheus at port 9090 in openshift-observability-operator namespace
+// - CoreOS (self-managed HyperShift): targets Thanos Querier at port 9092 in openshift-monitoring namespace
+func reconcileMetricsServerNetworkPolicy(policy *networkingv1.NetworkPolicy, hcp *hyperv1.HostedControlPlane) error {
```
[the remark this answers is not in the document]
Chee-Lu: Fixed comments

hypershift-operator/controllers/hostedcluster/network_policies.go
```diff
@@ -33,6 +35,19 @@ const (
...
+		return false
+	}
+
+	for _, tag := range hcp.Spec.Platform.AWS.ResourceTags {
+		if tag.Key == "red-hat-managed" && tag.Value == "true" {
+			return true
+		}
+	}
+	return false
+}
```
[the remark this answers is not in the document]
Chee-Lu: fixed, moved IsROSAHCP to shared awsutil package

control-plane-operator/controllers/hostedcontrolplane/v2/cvo/deployment.go
simonpasquier: regarding my previous comment:

> And about the URL, my gut feeling is that it isn't an HyperShift concern but rather something that would be provided via configuration.

IMHO the most flexible solution is to expose the URL as a configuration option of the HyperShift operator. For ARO, it would probably require no further change to HyperShift itself (https://github.com/Azure/ARO-HCP/blob/d1a8c74011cd8bb9a7627b6e44331af99b4644ee/config/config.yaml#L208).

@csrwng @sjenning for input (or redirect the question to the right person).

sjenning:
/approve

waiting to apply ok-to-test until CI is working again

muraee:
please squash the commits.
happy to tag afterwards.

control-plane-operator/controllers/hostedcontrolplane/v2/cvo/deployment.go
```diff
@@ -93,10 +99,22 @@ func (cvo *clusterVersionOperator) adaptDeployment(cpContext component.WorkloadC
...
 			c.Args = append(c.Args, "--use-dns-for-services=true")
-			c.Args = append(c.Args, "--metrics-ca-bundle-file=/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt")
-			c.Args = append(c.Args, fmt.Sprintf("--metrics-url=https://thanos-querier.openshift-monitoring.svc:9092?namespace=%s", cpContext.HCP.Namespace))
+
+			// Configure metrics endpoint based on monitoring stack
+			var metricsURL string
+			if os.Getenv(rhobsmonitoring.EnvironmentVariable) == "1" && awsutil.IsROSAHCP(cpContext.HCP) {
+				// RHOBS Prometheus uses HTTP without TLS (ROSA HCP specific)
+				c.Args = append(c.Args, "--metrics-ca-bundle-file=")
+				metricsURL = "http://hypershift-monitoring-stack-prometheus.openshift-observability-operator.svc:9090"
```
simonpasquier: I had an old comment on this part earlier for which I never got a reply from HyperShift maintainers: does it really make sense to hardcode the endpoint value which is specific to ROSA and may change in the future?

My personal feeling is that it should be a configuration option of the HyperShift operator for greater flexibility (but happy to be in disagreement).

cc @celebdor @muraee

muraee: missed that comment. Agreed, I think it makes sense to make this configurable rather than hardcoding. ARO/GCP can then use the same knob if needed to configure the correct url.

LiangquanLi930:
need to rebase this PR, then should be fix Konflux jobs

control-plane-operator/controllers/hostedcontrolplane/v2/cvo/deployment.go
```diff
@@ -93,10 +99,22 @@ func (cvo *clusterVersionOperator) adaptDeployment(cpContext component.WorkloadC
...
+		if enableMetricsAccess {
 			c.Args = append(c.Args, "--use-dns-for-services=true")
-			c.Args = append(c.Args, "--metrics-ca-bundle-file=/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt")
-			c.Args = append(c.Args, fmt.Sprintf("--metrics-url=https://thanos-querier.openshift-monitoring.svc:9092?namespace=%s", cpContext.HCP.Namespace))
+
+			// Configure metrics endpoint based on monitoring stack
+			var metricsURL string
+			if os.Getenv(rhobsmonitoring.EnvironmentVariable) == "1" && awsutil.IsROSAHCP(cpContext.HCP) {
+				// RHOBS Prometheus uses HTTP without TLS (ROSA HCP specific)
+				c.Args = append(c.Args, "--metrics-ca-bundle-file=")
```
heliubj18: I have a minor suggestion: For better readability, could we keep the related command-line arguments `--metrics-ca-bundle-file` and `--metrics-url` together? For example, either place both inside the same if/else block or keep both outside

cmd/install/install.go
```diff
@@ -291,7 +269,7 @@ func NewCommand() *cobra.Command {
...
 	cmd.Flags().BoolVar(&opts.WaitUntilAvailable, "wait-until-available", opts.WaitUntilAvailable, "If true, pauses installation until hypershift operator has been rolled out and its webhook service is available (if installing the webhook)")
 	cmd.Flags().BoolVar(&opts.WaitUntilEstablished, "wait-until-crds-established", opts.WaitUntilEstablished, "If true, pauses installation until all custom resource definitions are established before applying other manifests.")
-	cmd.PersistentFlags().BoolVar(&opts.RHOBSMonitoring, "rhobs-monitoring", opts.RHOBSMonitoring, "If true, HyperShift will generate and use the RHOBS version of monitoring resources (ServiceMonitors, PodMonitors, etc)")
+	cmd.PersistentFlags().BoolVar(&opts.RHOBSMonitoring, "rhobs-monitoring", opts.RHOBSMonitoring, "If true, HyperShift will generate and use the RHOBS version of monitoring resources (ServiceMonitors, PodMonitors, etc) and enable the Cluster Version Operator to query Prometheus for conditional update risk evaluation")
```
simonpasquier: not true anymore if we expose a parameter to configure the CVO's Prometheus URL in HyperShift operator.

control-plane-operator/controllers/hostedcontrolplane/v2/cvo/deployment.go
Chee-Lu: Thanks! placed them both inside the same block.

control-plane-operator/controllers/hostedcontrolplane/v2/cvo/deployment.go
Chee-Lu: > I'm not sure that in ROSA land, the monitoring stack honors the namespace parameter

Forgot to reply to this comment. This is correct, already removed namespace parameter from the URL.

control-plane-operator/controllers/hostedcontrolplane/v2/cvo/deployment.go
Chee-Lu: Thanks for the feedback @simonpasquier  @muraee.

I've added environment variable `RHOBS_PROMETHEUS_URL` to make the Prometheus URL configurable, with a default value of `http://hypershift-monitoring-stack-prometheus.openshift-observability-operator.svc:9090` in case it's not set.

Just wondering if the default is necessary though - if ARO/GCP need different URLs in the future, the default wouldn't apply to them anyway.

cmd/install/install.go
Chee-Lu: Are you referring to a specific part of the description that needs updating now that the URL is configurable? Could you elaborate on that a bit more?

Chee-Lu:
Rebased this PR

cmd/install/install.go
simonpasquier: Sorry for being dense! If there's a default CVO URL tied to this parameter, we should probably document it here. But it also depends on the decision about having a default or not.

control-plane-operator/controllers/hostedcontrolplane/v2/cvo/deployment.go
simonpasquier: I'm not sure that we should tie the value of the CVO URL to the use of the `monitoring.rhobs` API group. From what I understand today:
* ROSA uses `monitoring.rhobs` API group with metrics being ingested by [COO](github.com/rhobs/observability-operator) monitoring stack.
* On-prem HCP is supposed to use the `monitoring.coreos.com` API group with metrics being ingested by the OCP user-defined monitoring stack (though there's nothing technically preventing someone to configure RHOBSMonitoring=true with a custom COO stack manually ala ROSA).
* ARO HCP uses the `monitoring.coreos.com` API group because it deploys a self-managed Prometheus stack (https://github.com/Azure/ARO-HCP/blob/main/docs/monitoring.md and https://github.com/Azure/ARO-HCP/blob/500e8c1792cc913140f3a4ab4967ff7ffa172b9c/hypershiftoperator/deploy/templates/installer.job.yaml#L23-L34).

If I'm not wrong, to support ARO-HCP we need to override the CVO URL even in the case of RHOBSMonitoring=false.

ping @tony-schndr @stevekuznetsov to get the ARO-HCP perspective.

Chee-Lu: @simonpasquier Thanks for clarifying the monitoring stack differences across platforms!

You're right that tying the CVO URL to RHOBS API group usage wouldn't work for ARO HCP. Based on your input, I've decoupled the CVO URL configuration from RHOBS monitoring:

- Renamed `--rhobs-prometheus-url` to `--cvo-prometheus-url`
- The `--cvo-prometheus-url` parameter now works independently of the `--rhobs-monitoring` flag

This way, ARO HCP (using `monitoring.coreos.com` API group) can override the CVO URL via the `--cvo-prometheus-url` parameter with requiring `RHOBSMonitoring=true`.

Chee-Lu: Given that we want to deliver this for ROSA HCP soon and we haven't received specific requirements from the ARO team yet, I suggest we proceed with the current implementation if there are no other major concerns. The design provides flexibility to address ARO-specific changes later once their plans are clearer.

cc @tony-schndr @stevekuznetsov

cmd/install/install.go
Chee-Lu: Good point! I reverted the `--rhobs-monitoring` flag description, it focus solely on API group selection, while `--cvo-prometheus-url` now handles the CVO Prometheus URL configuration separately.

I kept platform-aware defaults (Thanos for self-managed, RHOBS stack for ROSA HCP) for backward compatibility, so most on-prem HCP users won't need to configure anything. Platforms like ARO HCP (or even ROSA HCP) can override the URL via `--cvo-prometheus-url` when needed.

Chee-Lu:
@sjenning @celebdor I've rebased and squashed the commits. Could you please take another look and approve/comment /ok-to-test if everything looks good? Thank you!

control-plane-operator/controllers/hostedcontrolplane/v2/cvo/deployment.go
tony-schndr: Thanks for calling this out @simonpasquier.  I am not aware of any specific requirements for ARO-HCP at this time but it doesn't mean there aren't any.  This does sound like a feature that ARO-HCP would want in the future for upgrades, we typically seek feature parity with ROSA HCP.

Chee-Lu:
> Just to recap, we now have the following environment variables which are somehow related:
> 
> * `RHOBS_MONITORING`
> * `ENABLE_CVO_MANAGEMENT_CLUSTER_METRICS_ACCESS`
> * `CVO_PROMETHEUS_URL`
> 
> ENABLE_CVO_MANAGEMENT_CLUSTER_METRICS_ACCESS=0	ENABLE_CVO_MANAGEMENT_CLUSTER_METRICS_ACCESS=1
> CVO_PROMETHEUS_URL not defined	CVO doesn't run PromQL risk evals except if rhobs-monitoring is on in which case it uses the default ROSA URL	CVO runs PromQL risk evals against the OCP Thanos querier (if not rhobs-monitoring) or the default ROSA Prometheus URL
> CVO_PROMETHEUS_URL defined	same as above	CVO runs PromQL risk evals against the configured URL
> Should we prevent some combinations to improve the user experience? For instance, setting `CVO_PROMETHEUS_URL` while not setting `ENABLE_CVO_MANAGEMENT_CLUSTER_METRICS_ACCESS` doesn't make lots of sense? Or should we consider that setting `CVO_PROMETHEUS_UR`L automatically enables CVO's metrics access?
> 
> Also the impact of the "rhobs-monitoring" setting on the final CVO configuration isn't very obvious IMHO. I'd rather have the rhobs-monitoring setting influence only which API group to use for monitoring resources

hi @simonpasquier , Thanks for raising these points. I think this touches on some historical design decisions that are worth discussing.

> Also the impact of the "rhobs-monitoring" setting on the final CVO configuration isn't very obvious IMHO. 
1. Yes, from the name of flag, the impact of --rhobs-monitoring on CVO configuration isn't immediately obvious. This stems from a limitation: --rhobs-monitoring and --enable-cvo-management-cluster-metrics-access were mutually exclusive because CVO accessing RHOBS metrics was not supported at that time. Additionally, these two flags had semantically conflicting names.

I see three possible approaches to address this:

**Option 1: Rename for Separation of Concerns**

Rename --enable-cvo-management-cluster-metrics-access → --enable-cvo-metrics-access
--rhobs-monitoring: Only controls monitoring stack API group
--enable-cvo-metrics-access: Controls CVO metrics access for both CoreOS and RHOBS
Pros: Clear separation, no semantic conflict
Cons: Breaking change, requires deprecation period and backward compatibility handling

**Option 2: Add another New Flag --enable-rhobs-cvo-metrics-access (No Breaking Changes)**

--rhobs-monitoring: Only controls monitoring stack API group
Keep existing --enable-cvo-management-cluster-metrics-access for CoreOS/self-managed
Add new --enable-rhobs-cvo-metrics-access for RHOBS-specific CVO configuration
Pros: No breaking changes, explicit control
Cons: Adds complexity, ROSA users need to use --rhobs-monitoring + --enable-rhobs-cvo-metrics-access

**Option 3: Current Implementation**

This is also based on @wking 's feedback - he preferred keeping it simple with a single --rhobs-monitoring flag
--rhobs-monitoring: Controls both monitoring API and implicitly enables CVO metrics access for ROSA HCP
Keep existing --enable-cvo-management-cluster-metrics-access for self-managed HyperShift
Pros: Simple for users (one flag does everything for RHOBS), no breaking changes
Cons: The impact on CVO configuration isn't very obvious from the flag name, but we can add descriptions 

IMO, this is acceptable, So I've added back the description to --rhobs-monitoring explicitly mention that it enables CVO metrics access.


> Should we prevent some combinations to improve the user experience?
2. Yes, I've added validation in this PR: setting --cvo-prometheus-url without either --rhobs-monitoring or --enable-cvo-management-cluster-metrics-access now produces a clear error message.

> Should we consider that setting CVO_PROMETHEUS_URL automatically enables CVO's metrics access?
3. I don't think this would work well. Like if a user sets a RHOBS URL but hasn't enabled RHOBS (still using CoreOS), there would be no data available and it would fail. It's better to require explicit enablement.

Let me raise this in our meeting today.

simonpasquier:
> I don't think this would work well. Like if a user sets a RHOBS URL but hasn't enabled RHOBS (still using CoreOS), there would be no data available and it would fail. It's better to require explicit enablement.

As I've written before, ARO HCP will likely use upstream (coreos) CRDs and will also need to customize the CVO Prometheus URL. This case should be supported, my point is that if you set an explicit CVO URL then why would you need also to enable CVO metrics access.

control-plane-operator/controllers/hostedcontrolplane/v2/cvo/deployment.go
```diff
@@ -27,12 +30,26 @@ import (
...
+	enableMetricsAccess := (enableRHOBSMonitoring && isROSAHCP) || cvo.enableCVOManagementClusterMetricsAccess
+
+	if enableMetricsAccess {
 		if deployment.Spec.Template.Labels == nil {
 			deployment.Spec.Template.Labels = map[string]string{}
 		}
 		deployment.Spec.Template.Labels[config.NeedMetricsServerAccessLabel] = "true"
-		deployment.Spec.Template.Spec.ServiceAccountName = ComponentName
+
+		if cvo.enableCVOManagementClusterMetricsAccess {
```
csrwng: Setting the service account blank results in us using the default service account, which I don't think is better.

We should just create a service account and set it in the deployment template.

Then determine whether we need management cluster credentials and set that here:
https://github.com/openshift/hypershift/blob/a4c53d1a99dae14f2745a8abc6696882a5c0d956/control-plane-operator/controllers/hostedcontrolplane/v2/cvo/component.go#L31-L33

otherwise, setting one service account or the other makes no difference. Right now because we're always returning false for `NeedsManagementKASAccess`, then the pod template spec has automountServiceAccountToken set to false. Which means that there is no token available to send to the metrics endpoint.

control-plane-operator/controllers/hostedcontrolplane/v2/cvo/deployment.go
```diff
@@ -93,10 +110,29 @@ func (cvo *clusterVersionOperator) adaptDeployment(cpContext component.WorkloadC
...
-			c.Args = append(c.Args, "--metrics-ca-bundle-file=/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt")
-			c.Args = append(c.Args, fmt.Sprintf("--metrics-url=https://thanos-querier.openshift-monitoring.svc:9092?namespace=%s", cpContext.HCP.Namespace))
+
+			cvoPrometheusURL := os.Getenv(config.CVOPrometheusURLEnvVar)
+
+			// Configure metrics endpoint based on monitoring stack
+			if enableRHOBSMonitoring && isROSAHCP {
+				// RHOBS monitoring stack (ROSA HCP) - no TLS
+				if cvoPrometheusURL == "" {
+					cvoPrometheusURL = "http://hypershift-monitoring-stack-prometheus.openshift-observability-operator.svc:9090"
```
csrwng, line 117: Will this always be the case? (that the default endpoint is unencrypted?) 
My understanding is that we want all communication to be encrypted within a cluster.

csrwng: If we always pass the --metrics-token-file and the --metrics-ca-bundle from the service account, does it hurt anything? That way in the future if we have an endpoint that is tls and authenticates the user, then it can be done by just changing the URL.

control-plane-operator/controllers/hostedcontrolplane/v2/cvo/deployment.go
csrwng: Updated above comment (looked at the wrong template yaml)

control-plane-operator/controllers/hostedcontrolplane/v2/cvo/deployment.go
Chee-Lu: Hi Cesar, I'm reaching out to the RHOBS team to ask if there are any plans to enable TLS for the metrics endpoint. cc @moadz. Will update you once I hear back.

nelsonjean: Hi @Chee-Lu , did you discuss with the RHOBS team?

control-plane-operator/controllers/hostedcontrolplane/v2/cvo/deployment.go
Chee-Lu: Thanks for the feedback!

> Setting the service account blank results in us using the default service account, which I don't think is better.

1. Updated to use the existing cluster-version-operator SA instead of default.

> Right now because we're always returning false for NeedsManagementKASAccess, then the pod template spec has automountServiceAccountToken set to false. Which means that there is no token available to send to the metrics endpoint.

2. Good point. It seems this might be a pre-existing issue that with `automountServiceAccountToken` always be `false`, the `--metrics-ca-bundle-file` path wouldn't exist for Thanos. 

3. I fixed this `automountServiceAccountToken` always be `false`, by set `Annotations[config.NeedMetricsServerAccessLabel] = "true"` which will also triggers `automountServiceAccountToken=true` here:

 https://github.com/openshift/hypershift/blob/52cb5ad1d/support/controlplane-component/defaults.go#L126-L131 

I didn't modify NeedsManagementKASAccess() because:
 - It's static and doesn't have runtime context, so if `--rhobs-monitoring` and `--enable-cvo-management-cluster-metrics-access` were both disabled, then CVO should not have the metrics access.
 - Semantically, CVO needs metrics access, not Management KAS access

control-plane-operator/controllers/hostedcontrolplane/v2/cvo/deployment.go
Chee-Lu: > Will this always be the case? (that the default endpoint is unencrypted?)
> My understanding is that we want all communication to be encrypted within a cluster.

Yes, RHOBS currently uses HTTP without TLS. No certificates configured, and the liveness probes use HTTP. Based on discussions with the RHOBS team @moadz , query the prometheus over TLS is technically possible, like adding an annotation "service.beta.openshift.io/serving-cert-secret-name", but it would require changes on the RHOBS side, which seems unlikely to happen in the short term.

> If we always pass the --metrics-token-file and the --metrics-ca-bundle from the service account, does it hurt anything? That way in the future if we have an endpoint that is tls and authenticates the user, then it can be done by just changing the URL.

Following your suggestion, I now always pass --metrics-ca-bundle-file from the service account, and [cvo will have a default --metrics-token-file path](https://github.com/openshift/cluster-version-operator/blob/95c76c7c2acdc4d5c595eda98f4058f0fb88f7a3/pkg/clusterconditions/clusterconditions.go#L123-L134). even though they're not currently needed for HTTP. As long as these files exist, passing them is safe. After fixing  `automountServiceAccountToken` always be `false` it should be good.

control-plane-operator/controllers/hostedcontrolplane/v2/cvo/deployment.go
```diff
@@ -27,11 +30,24 @@ import (
...
 
 func (cvo *clusterVersionOperator) adaptDeployment(cpContext component.WorkloadContext, deployment *appsv1.Deployment) error {
-	if cvo.enableCVOManagementClusterMetricsAccess {
+	enableRHOBSMonitoring := os.Getenv(rhobsmonitoring.EnvironmentVariable) == "1"
+	isROSAHCP := awsutil.IsROSAHCP(cpContext.HCP)
+
+	// Enable CVO metrics access if either RHOBS monitoring is enabled or the explicit flag is set
+	enableMetricsAccess := (enableRHOBSMonitoring && isROSAHCP) || cvo.enableCVOManagementClusterMetricsAccess
+
+	if enableMetricsAccess {
```
simonpasquier, line 35: (nit) isn't this the same as calling `cvo.isManagementClusterMetricsAccessEnabled()`?

control-plane-operator/controllers/hostedcontrolplane/v2/cvo/deployment.go
```diff
@@ -93,10 +109,29 @@ func (cvo *clusterVersionOperator) adaptDeployment(cpContext component.WorkloadC
...
+			cvoPrometheusURL := os.Getenv(config.CVOPrometheusURLEnvVar)
+
+			// Configure metrics endpoint based on monitoring stack.
+			if enableRHOBSMonitoring && isROSAHCP {
+				// RHOBS monitoring stack (ROSA HCP) - currently uses HTTP without TLS, but passing CA and token allows switching to a TLS-authenticated endpoint in the future by changing the URL.
+				if cvoPrometheusURL == "" {
+					cvoPrometheusURL = "http://hypershift-monitoring-stack-prometheus.openshift-observability-operator.svc:9090"
+				}
+				c.Args = append(c.Args, "--metrics-ca-bundle-file=/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt")
+				c.Args = append(c.Args, "--metrics-token-file=/var/run/secrets/kubernetes.io/serviceaccount/token")
```
simonpasquier: (nit) isn't it the same as the default value used by CVO?

control-plane-operator/controllers/hostedcontrolplane/v2/cvo/deployment.go
```diff
@@ -93,10 +109,29 @@ func (cvo *clusterVersionOperator) adaptDeployment(cpContext component.WorkloadC
...
+
+		if enableMetricsAccess {
 			c.Args = append(c.Args, "--use-dns-for-services=true")
-			c.Args = append(c.Args, "--metrics-ca-bundle-file=/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt")
-			c.Args = append(c.Args, fmt.Sprintf("--metrics-url=https://thanos-querier.openshift-monitoring.svc:9092?namespace=%s", cpContext.HCP.Namespace))
+
+			cvoPrometheusURL := os.Getenv(config.CVOPrometheusURLEnvVar)
+
+			// Configure metrics endpoint based on monitoring stack.
+			if enableRHOBSMonitoring && isROSAHCP {
```
simonpasquier: it still doesn't account for the future ARO (and GCP) HCP cases which would not configure RHOBS monitoring but want to use a custom Prometheus URL.

simonpasquier: I stand corrected, it should work but the code could be simplified.

Chee-Lu:
hi @simonpasquier , I addressed your comments, please let me know if you have other concerns

cmd/install/install.go
```diff
@@ -306,7 +310,8 @@ func NewCommand() *cobra.Command {
...
 	cmd.Flags().BoolVar(&opts.WaitUntilAvailable, "wait-until-available", opts.WaitUntilAvailable, "If true, pauses installation until hypershift operator has been rolled out and its webhook service is available (if installing the webhook)")
 	cmd.Flags().BoolVar(&opts.WaitUntilEstablished, "wait-until-crds-established", opts.WaitUntilEstablished, "If true, pauses installation until all custom resource definitions are established before applying other manifests.")
-	cmd.PersistentFlags().BoolVar(&opts.RHOBSMonitoring, "rhobs-monitoring", opts.RHOBSMonitoring, "If true, HyperShift will generate and use the RHOBS version of monitoring resources (ServiceMonitors, PodMonitors, etc)")
+	cmd.PersistentFlags().BoolVar(&opts.RHOBSMonitoring, "rhobs-monitoring", opts.RHOBSMonitoring, "If true, HyperShift will generate and use the RHOBS version of monitoring resources (ServiceMonitors, PodMonitors, etc) and enable the Cluster Version Operator to query the RHOBS Prometheus for conditional update evaluation. Use --cvo-prometheus-url to override the default Prometheus endpoint.")
```
simonpasquier: (nit) `enable the Cluster Version Operator to query the RHOBS Prometheus` => it's only true if the operator detects that the cluster is a ROSA HCP cluster?

hypershift-operator/controllers/hostedcluster/network_policies.go
```diff
@@ -831,36 +836,72 @@ func kasEndpointsToCIDRs(kubernetesEndpoint *corev1.Endpoints) []string {
 }
 
 // reconcileMetricsServerNetworkPolicy selects pods having NeedMetricsServerAccessLabel.
-// It allows egress traffic to the HCP metrics server that is available for self-managed
-// HyperShift running on an OpenShift management cluster.
-func reconcileMetricsServerNetworkPolicy(policy *networkingv1.NetworkPolicy) error {
+// It allows egress traffic to the HCP metrics server that is available for self-managed HyperShift and ROSA HCP.
+// Depending on the monitoring stack:
+// - RHOBS (ROSA HCP only): targets OBO Prometheus at port 9090 in openshift-observability-operator namespace
+// - CoreOS (self-managed HyperShift): targets Thanos Querier at port 9092 in openshift-monitoring namespace
```
simonpasquier, line 842: Here we may have a gap for other managed environments which aren't ROSA and use network policies.  I'm not sure how it can be solved though?

cmd/install/install.go
Chee-Lu: Updated the description

hypershift-operator/controllers/hostedcluster/network_policies.go
Chee-Lu: I think it makes sense to extend this when the requirements from ARO or GCP become clearer. What do you think?

Chee-Lu:
@csrwng @sjenning  @celebdor  Hi, could you please take another look and approve if everything looks good to you? Thank you!

hypershift-operator/controllers/hostedcluster/network_policies.go
simonpasquier: makes sense to me

Chee-Lu:
Hi @celebdor @csrwng, all tests for this PR have passed! Could you PTAL if we need an ok-to-test or an approval before it can be merged?

tdrozdowski:
> Hi @celebdor @csrwng, all tests for this PR have passed! Could you PTAL if we need an ok-to-test or an approval before it can be merged?

@celebdor / @csrwng  - just a bump on this after the weekend.  thx.

PR #7407, opened 2025-12-17
CNTRLPLANE-1988: Add Azure support to product CLI for self-managed customers

bryan-cox:

Adds Azure platform support to the product CLI (`hcp`) for self-managed Azure customers. This enables customers to create and manage Azure HostedClusters and NodePools using the productized CLI.

**New commands added:**
- `hcp create cluster azure` - Creates a HostedCluster on Azure
- `hcp destroy cluster azure` - Destroys a HostedCluster and its Azure infrastructure
- `hcp create nodepool azure` - Creates an Azure NodePool
- `hcp destroy nodepool` - Destroys a NodePool (platform-agnostic)

**Key design decisions:**
- Added `BindProductFlags` functions to expose only self-managed Azure flags, excluding ARO HCP-specific flags like `--managed-identities-file` and `--data-plane-identities-file`
- NodePool destroy is platform-agnostic since no cloud infrastructure cleanup is required
- Follows the established AWS product CLI pattern with thin wrappers

Fixes [CNTRLPLANE-1988](https://issues.redhat.com//browse/CNTRLPLANE-1988)

The `BindProductFlags` pattern differs slightly from AWS which uses `BindOptions` directly. This is intentional because Azure has distinct flag sets for self-managed (workload identities) vs managed/ARO HCP (managed identities), and the product CLI targets only self-managed customers.

I tested the commands out. Everything looked good to me - https://issues.redhat.com/browse/CNTRLPLANE-1988?focusedId=28746778&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-28746778.

## Checklist:

### DIFF
--- a/cmd/cluster/azure/create.go
+++ b/cmd/cluster/azure/create.go
@@ -31,11 +31,11 @@ const (
 
 var _ core.Platform = (*CreateOptions)(nil)
 
-func DefaultOptions() (*RawCreateOptions, error) {
+func DefaultOptions() *RawCreateOptions {
 	return &RawCreateOptions{
 		Location:     "eastus",
 		NodePoolOpts: azurenodepool.DefaultOptions(),
-	}, nil
+	}
 }
 
 func NewCreateCommand(opts *core.RawCreateOptions) *cobra.Command {
@@ -45,11 +45,7 @@ func NewCreateCommand(opts *core.RawCreateOptions) *cobra.Command {
 		SilenceUsage: true,
 	}
 
-	azureOpts, err := DefaultOptions()
-	if err != nil {
-		opts.Log.Error(err, "Failed to create default options")
-		return nil
-	}
+	azureOpts := DefaultOptions()
 	BindOptions(azureOpts, cmd.Flags())
 	_ = cmd.MarkPersistentFlagRequired("pull-secret")
 
@@ -78,31 +74,32 @@ func BindOptions(opts *RawCreateOptions, flags *pflag.FlagSet) {
 
 func bindCoreOptions(opts *RawCreateOptions, flags *pflag.FlagSet) {
 	// ARO HCP (managed Azure) only flags; these flags should not be used for self-managed Azure
-	flags.StringVar(&opts.KMSUserAssignedCredsSecretName, "kms-credentials-secret-name", opts.KMSUserAssignedCredsSecretName, "The name of a secret, in Azure KeyVault, containing the JSON UserAssignedIdentityCredentials used in KMS to authenticate to Azure.")
-	flags.StringVar(&opts.ManagedIdentitiesFile, "managed-identities-file", opts.ManagedIdentitiesFile, "Path to a file containing the managed identities configuration in json format.")
-	flags.StringVar(&opts.DataPlaneIdentitiesFile, "data-plane-identities-file", opts.DataPlaneIdentitiesFile, "Path to a file containing the client IDs of the managed identities for the data plane configured in json format.")
-	flags.BoolVar(&opts.AssignCustomHCPRoles, "assign-custom-hcp-roles", opts.AssignCustomHCPRoles, "Assign custom roles to HCP identities")
+	flags.StringVar(&opts.KMSUserAssignedCredsSecretName, "kms-credentials-secret-name", opts.KMSUserAssignedCredsSecretName, util.KMSCredentialsSecretNameDescription)
+	flags.StringVar(&opts.ManagedIdentitiesFile, "managed-identities-file", opts.ManagedIdentitiesFile, util.ManagedIdentitiesFileDescription)
+	flags.StringVar(&opts.DataPlaneIdentitiesFile, "data-plane-identities-file", opts.DataPlaneIdentitiesFile, util.DataPlaneIdentitiesFileDescription)
+	flags.BoolVar(&opts.AssignCustomHCPRoles, "assign-custom-hcp-roles", opts.AssignCustomHCPRoles, util.AssignCustomHCPRolesDescription)
 
 	// self-managed Azure only flags; these flags should not be used for ARO HCP (managed Azure)
-	flags.StringVar(&opts.WorkloadIdentitiesFile, "workload-identities-file", opts.WorkloadIdentitiesFile, "Path to a file containing the workload identity client IDs configuration in json format for self-managed Azure.")
+	flags.StringVar(&opts.WorkloadIdentitiesFile, "workload-identities-file", opts.WorkloadIdentitiesFile, util.WorkloadIdentitiesFileDescription)
 
 	// flags used for both ARO HCP and self-managed Azure
-	// In ARO HCP, it assigns roles to managed identities; in self-managed Azure, it assigns roles to workload identities. The self-managed HCP CLI version of this flage is named assign-identity-roles.
-	flags.BoolVar(&opts.AssignServicePrincipalRoles, "assign-service-principal-roles", opts.AssignServicePrincipalRoles, "Assign required Azure RBAC roles to identities (managed identities for ARO HCP, workload identities for self-managed).")
+	// In ARO HCP, it assigns roles to managed identities; in self-managed Azure, it assigns roles to workload identities.
+	// The self-managed HCP CLI version of this flag is named assign-identity-roles.
+	flags.BoolVar(&opts.AssignServicePrincipalRoles, "assign-service-principal-roles", opts.AssignServicePrincipalRoles, util.AssignServicePrincipalRolesDescription)
 
 	// general flags used for both managed and self-managed Azure
-	flags.StringVar(&opts.CredentialsFile, "azure-creds", opts.CredentialsFile, "Path to an Azure credentials file (required). This file is used to extract the subscription ID, tenant ID, and its credentials are used to create the necessary Azure resources for the HostedCluster.")
-	flags.StringVar(&opts.Location, "location", opts.Location, "Location for the HostedCluster. This is also used as the location for the Azure resources created for the HostedCluster.")
-	flags.StringVar(&opts.EncryptionKeyID, "encryption-key-id", opts.EncryptionKeyID, "etcd encryption key identifier in the form of https://<vaultName>.vault.azure.net/keys/<keyName>/<keyVersion> used to set up KMSv2 for etcd encryption.")
-	flags.StringSliceVar(&opts.AvailabilityZones, "availability-zones", opts.AvailabilityZones, "The availability zones in which NodePools will be created. Must be left unspecified if the region does not support AZs. If set, one nodepool per zone will be created.")
-	flags.StringVar(&opts.ResourceGroupName, "resource-group-name", opts.ResourceGroupName, "The resource group name to create the HostedCluster infrastructure resources under. If not provided, a new resource group will be created.")
-	flags.StringVar(&opts.VnetID, "vnet-id", opts.VnetID, "An existing VNET ID. If not provided, a new VNET will be created.")
-	flags.StringVar(&opts.NetworkSecurityGroupID, "network-security-group-id", opts.NetworkSecurityGroupID, "The Network Security Group ID to use in the default NodePool. If not provided, a new Network Security Group will be created.")
-	flags.StringToStringVarP(&opts.ResourceGroupTags, "resource-group-tags", "t", opts.ResourceGroupTags, "Additional tags to apply to the resource group created (e.g. 'key1=value1,key2=value2')")
-	flags.StringVar(&opts.SubnetID, "subnet-id", opts.SubnetID, "The subnet ID where the VMs will be placed. If not provided, a new subnet will be created.")
-	flags.StringVar(&opts.IssuerURL, "oidc-issuer-url", "", "The OIDC provider issuer URL.")
-	flags.StringVar(&opts.ServiceAccountTokenIssuerKeyPath, "sa-token-issuer-private-key-path", "", "The file to the private key for the service account token issuer.")
-	flags.StringVar(&opts.DNSZoneRGName, "dns-zone-rg-name", opts.DNSZoneRGName, "The name of the resource group where the DNS Zone resides. This is needed for the ingress controller. This is just the name and not the full ID of the resource group.")
+	flags.StringVar(&opts.CredentialsFile, "azure-creds", opts.CredentialsFile, util.AzureCredsDescription)
+	flags.StringVar(&opts.Location, "location", opts.Location, util.LocationDescription)
+	flags.StringVar(&opts.EncryptionKeyID, "encryption-key-id", opts.EncryptionKeyID, util.EncryptionKeyIDDescription)
+	flags.StringSliceVar(&opts.AvailabilityZones, "availability-zones", opts.AvailabilityZones, util.AvailabilityZonesDescription)
+	flags.StringVar(&opts.ResourceGroupName, "resource-group-name", opts.ResourceGroupName, util.ResourceGroupNameDescription)
+	flags.StringVar(&opts.VnetID, "vnet-id", opts.VnetID, util.VnetIDDescription)
+	flags.StringVar(&opts.NetworkSecurityGroupID, "network-security-group-id", opts.NetworkSecurityGroupID, util.NetworkSecurityGroupIDDescription)
+	flags.StringToStringVarP(&opts.ResourceGroupTags, "resource-group-tags", "t", opts.ResourceGroupTags, util.ResourceGroupTagsDescription)
+	flags.StringVar(&opts.SubnetID, "subnet-id", opts.SubnetID, util.SubnetIDDescription)
+	flags.StringVar(&opts.IssuerURL, "oidc-issuer-url", "", util.OIDCIssuerURLDescription)
+	flags.StringVar(&opts.ServiceAccountTokenIssuerKeyPath, "sa-token-issuer-private-key-path", "", util.SATokenIssuerKeyPathDescription)
+	flags.StringVar(&opts.DNSZoneRGName, "dns-zone-rg-name", opts.DNSZoneRGName, util.DNSZoneRGNameDescription)
 }
 
 // BindDeveloperOptions binds developer/development only options for the Azure create cluster command
@@ -129,7 +126,7 @@ func BindProductFlags(opts *RawCreateOptions, flags *pflag.FlagSet) {
 	flags.StringVar(&opts.WorkloadIdentitiesFile, "workload-identities-file", opts.WorkloadIdentitiesFile, util.WorkloadIdentitiesFileDescription)
 	flags.StringVar(&opts.IssuerURL, "oidc-issuer-url", "", util.OIDCIssuerURLDescription)
 	flags.StringVar(&opts.ServiceAccountTokenIssuerKeyPath, "sa-token-issuer-private-key-path", "", util.SATokenIssuerKeyPathDescription)
-	flags.BoolVar(&opts.AssignServicePrincipalRoles, "assign-identity-roles", opts.AssignServicePrincipalRoles, util.AssignIdentityRolesDescription)
+	flags.BoolVar(&opts.AssignServicePrincipalRoles, "auto-assign-roles", opts.AssignServicePrincipalRoles, util.AutoAssignRolesDescription)
 
 	// Encryption
 	flags.StringVar(&opts.EncryptionKeyID, "encryption-key-id", opts.EncryptionKeyID, util.EncryptionKeyIDDescription)
--- a/cmd/nodepool/azure/create.go
+++ b/cmd/nodepool/azure/create.go
@@ -28,7 +28,6 @@ type AzurePlatformCreateOptions struct {
 	EnableEphemeralOSDisk         bool
 	DiskStorageAccountType        string
 	SubnetID                      string
-	ImageID                       string
 	ImageGeneration               string
 	Arch                          string
 	EncryptionAtHost              string
@@ -147,7 +146,7 @@ func (o *RawAzurePlatformCreateOptions) Validate() (*ValidatedAzurePlatformCreat
 	}
 
 	if o.EncryptionAtHost != "" && o.EncryptionAtHost != "Enabled" && o.EncryptionAtHost != "Disabled" {
-		return nil, fmt.Errorf("flag --enable-encryption-at-host has an invalid value; accepted values are 'Enabled' and 'Disabled'")
+		return nil, fmt.Errorf("flag --encryption-at-host has an invalid value; accepted values are 'Enabled' and 'Disabled'")
 	}
 
 	if !slices.Contains([]string{"", "1", "2", "3"}, o.AvailabilityZone) {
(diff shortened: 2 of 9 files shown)


wewang58:
/verified by @wewang58 
@bryan-cox the pr is already tested, we merge the pr first? then test the pr: https://github.com/openshift/hypershift/pull/7436, right?

product-cli/main.go
```diff
@@ -26,9 +26,16 @@ import (
 	"github.com/openshift/hypershift/product-cli/cmd/destroy"
 
 	"github.com/spf13/cobra"
+	"go.uber.org/zap/zapcore"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 )
```
devguyio: we do k8s imports in a separate group

product-cli/main.go
```diff
@@ -26,9 +26,16 @@ import (
...
 
 func main() {
+	ctrl.SetLogger(zap.New(zap.JSONEncoder(func(o *zapcore.EncoderConfig) {
```
devguyio: I definitely support JSON structured logging, however in a CLI don't you think a normal console output is more user friendly?

also this is inconsistent with other commands like destroy command.

product-cli/cmd/cluster/azure/create.go
```diff
@@ -0,0 +1,50 @@
...
+
+	azureOpts, err := hypershiftazure.DefaultOptions()
+	if err != nil {
+		opts.Log.Error(err, "Failed to create default options")
+		cmd.RunE = func(cmd *cobra.Command, args []string) error {
+			return fmt.Errorf("failed to initialize Azure options: %w", err)
+		}
+		return cmd
+	}
+
```
devguyio: Is this pattern correct? DefaultOptions() will never return an error, and the error handling here is to set a command that fails? I think it's better to simplify this and follow the other commands' pattern where DefaultOptions doesn't return an error

product-cli/cmd/cluster/azure/create.go
```diff
@@ -0,0 +1,50 @@
...
+	hypershiftazure.BindProductFlags(azureOpts, cmd.Flags())
+
```
devguyio, line 24: in [aws-cli](https://github.com/openshift/hypershift/blob/main/product-cli/cmd/cluster/aws/create.go#L20-L21) we set the `opts.ReleaseStream`, should we do that here too?

cmd/cluster/azure/create.go
```diff
@@ -107,6 +110,34 @@ func BindDeveloperOptions(opts *RawCreateOptions, flags *pflag.FlagSet) {
 	bindCoreOptions(opts, flags)
 }
 
+// BindProductFlags binds customer-facing flags for self-managed Azure in the product CLI
+func BindProductFlags(opts *RawCreateOptions, flags *pflag.FlagSet) {
+	// Required credentials
+	flags.StringVar(&opts.CredentialsFile, "azure-creds", opts.CredentialsFile, util.AzureCredsDescription)
```
devguyio, line 113: so my understanding is that here description is moved to a struct because description gets to be a long string, however on the downsides are:
1. Less code readability vs having the option name right next to the description, thus understanding the option from a first glance.
2. Mixed convention, because even a few lines above the options follow the other convention.

I'd suggest we stay with the old convention OR move all azure options to as struct where both the name and the description live together, and hoping we'd unify the rest of the commands later.

cmd/cluster/azure/create.go
```diff
@@ -107,6 +110,34 @@ func BindDeveloperOptions(opts *RawCreateOptions, flags *pflag.FlagSet) {
...
+	flags.StringVar(&opts.SubnetID, "subnet-id", opts.SubnetID, util.SubnetIDDescription)
+	flags.StringVar(&opts.NetworkSecurityGroupID, "network-security-group-id", opts.NetworkSecurityGroupID, util.NetworkSecurityGroupIDDescription)
+	flags.StringToStringVarP(&opts.ResourceGroupTags, "resource-group-tags", "t", opts.ResourceGroupTags, util.ResourceGroupTagsDescription)
+	flags.StringVar(&opts.DNSZoneRGName, "dns-zone-rg-name", opts.DNSZoneRGName, util.DNSZoneRGNameDescription)
+
+	// Self-managed Azure identity flags
+	flags.StringVar(&opts.WorkloadIdentitiesFile, "workload-identities-file", opts.WorkloadIdentitiesFile, util.WorkloadIdentitiesFileDescription)
+	flags.StringVar(&opts.IssuerURL, "oidc-issuer-url", "", util.OIDCIssuerURLDescription)
+	flags.StringVar(&opts.ServiceAccountTokenIssuerKeyPath, "sa-token-issuer-private-key-path", "", util.SATokenIssuerKeyPathDescription)
+	flags.BoolVar(&opts.AssignServicePrincipalRoles, "assign-identity-roles", opts.AssignServicePrincipalRoles, util.AssignIdentityRolesDescription)
```
devguyio: I understand why the `assign-identity-roles` and `assign-service-principal-roles` have different names, but using the same opt field for both is confusing IMHO and carries an assumption that can make debuging harder (i.e. code reader must know that this opt, is the one that comes from a different flag name only for this case). 

I think it's simple if we just assign the same abstract-name for both the flags and opts field across both ARO and self-managed (e.g. `auto-assign-roles`), and deprecate the `assign-service-principal-roles` role to not break the CI

cmd/nodepool/azure/create.go
```diff
@@ -83,6 +82,32 @@ func BindDeveloperOptions(opts *RawAzurePlatformCreateOptions, flags *pflag.Flag
...
+	flags.StringVar(&opts.InstanceType, "instance-type", opts.InstanceType, util.InstanceTypeDescription)
+	flags.Int32Var(&opts.DiskSize, "root-disk-size", opts.DiskSize, util.RootDiskSizeDescription)
+	flags.StringVar(&opts.AvailabilityZone, "availability-zone", opts.AvailabilityZone, util.AvailabilityZoneDescription)
+	flags.StringVar(&opts.SubnetID, "nodepool-subnet-id", opts.SubnetID, util.SubnetIDDescription)
+
+	// Disk configuration
+	flags.StringVar(&opts.DiskStorageAccountType, "disk-storage-account-type", opts.DiskStorageAccountType, util.DiskStorageAccountTypeDescription)
+	flags.StringVar(&opts.DiskEncryptionSetID, "disk-encryption-set-id", opts.DiskEncryptionSetID, util.DiskEncryptionSetIDDescription)
+	flags.BoolVar(&opts.EnableEphemeralOSDisk, "enable-ephemeral-disk", opts.EnableEphemeralOSDisk, util.EnableEphemeralOSDiskDescription)
+	flags.StringVar(&opts.EncryptionAtHost, "encryption-at-host", opts.EncryptionAtHost, util.EncryptionAtHostDescription)
```
devguyio, line 96: I noticed the error message in line 150 has the wrong flag name `--enable-encryption-at-host`

cmd/nodepool/azure/create.go
```diff
@@ -70,7 +70,6 @@ func bindCoreOptions(opts *RawAzurePlatformCreateOptions, flags *pflag.FlagSet)
 	flags.BoolVar(&opts.EnableEphemeralOSDisk, "enable-ephemeral-disk", opts.EnableEphemeralOSDisk, "If enabled, the Azure VMs in the NodePool will be setup with ephemeral OS disks")
 	flags.StringVar(&opts.DiskStorageAccountType, "disk-storage-account-type", opts.DiskStorageAccountType, "The disk storage account type for the OS disks for the VMs.")
 	flags.StringVar(&opts.SubnetID, "nodepool-subnet-id", opts.SubnetID, "The subnet id where the VMs will be placed.")
-	flags.StringVar(&opts.ImageID, "image-id", opts.ImageID, "The Image ID to boot the VMs with.")
 	flags.StringVar(&opts.ImageGeneration, "image-generation", opts.ImageGeneration, "The Hyper-V generation of the Azure VM image. Supported values: Gen1, Gen2. If unspecified, defaults to Gen2.")
```
devguyio, line 72: with the `--image-id` flag removed, we need to remove the opts field (line 31) and the code in line 208

product-cli/cmd/cluster/azure/destroy.go
```diff
@@ -0,0 +1,36 @@
...
+func NewDestroyCommand(opts *core.DestroyOptions) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:          "azure",
+		Short:        "Destroys a HostedCluster and its associated infrastructure on Azure",
+		SilenceUsage: true,
+	}
+
+	opts.AzurePlatform.Location = "eastus"
+	cmd.Flags().StringVar(&opts.AzurePlatform.CredentialsFile, "azure-creds", opts.AzurePlatform.CredentialsFile, util.AzureCredsDescription)
+	cmd.Flags().StringVar(&opts.AzurePlatform.Location, "location", opts.AzurePlatform.Location, util.LocationDescription)
```
devguyio: seeing how the codepath of destroy infers the location if the hosted cluster exists, I think this description should be something like

```markdown
Azure region of the cluster. Inferred from the HostedCluster if it exists; only required if the cluster resource has already been deleted.
```

product-cli/main.go
devguyio: This is how it would like 

```text
⬢ [go-dev] ❯ ./bin/hcp create cluster aws
{"level":"error","ts":"2026-01-20T12:37:40+01:00","msg":"Failed to create cluster","error":"--name is required","stacktrace":"github.com/openshift/hypershift/product-cli/cmd/cluster/aws.NewCreateCommand.func1\n\t/home/ahmed/GolandProjects/github.com/openshift/hypershift_ws/capi-bump/hypershift/product-cli/cmd/cluster/aws/create.go:34\ngithub.com/spf13/cobra.(*Command).execute\n\t/home/ahmed/GolandProjects/github.com/openshift/hypershift_ws/capi-bump/hypershift/vendor/github.com/spf13/cobra/command.go:1015\ngithub.com/spf13/cobra.(*Command).ExecuteC\n\t/home/ahmed/GolandProjects/github.com/openshift/hypershift_ws/capi-bump/hypershift/vendor/github.com/spf13/cobra/command.go:1148\ngithub.com/spf13/cobra.(*Command).Execute\n\t/home/ahmed/GolandProjects/github.com/openshift/hypershift_ws/capi-bump/hypershift/vendor/github.com/spf13/cobra/command.go:1071\ngithub.com/spf13/cobra.(*Command).ExecuteContext\n\t/home/ahmed/GolandProjects/github.com/openshift/hypershift_ws/capi-bump/hypershift/vendor/github.com/spf13/cobra/command.go:1064\nmain.main\n\t/home/ahmed/GolandProjects/github.com/openshift/hypershift_ws/capi-bump/hypershift/product-cli/main.go:66\nruntime.main\n\t/home/ahmed/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.11.linux-amd64/src/runtime/proc.go:283"}
Error: --name is required
--name is required
```

devguyio: after digging deeper, there are two separate issues here:
1. Let's switch to using the Log from cmd/log/log.go because that's the convention across the repo.
2. We need to improve the UX. But that's irrelevant of this PR I believe, unless we wanna start self-managed with a good UX.

product-cli/cmd/cluster/azure/create.go
```diff
@@ -0,0 +1,50 @@
...
+	_ = cmd.MarkFlagRequired("azure-creds")
+	_ = cmd.MarkPersistentFlagRequired("pull-secret")
+
+	cmd.RunE = func(cmd *cobra.Command, args []string) error {
```
devguyio, line 28: (Not specific for Azure)

Our UX for error handling could use some love. Currently:
1. We use RunE, which expects that the run func returns the error, and Cobra will print it nor log it. 
2. Yet we log it (line 43).
3. And we print it again in [main.go](https://github.com/openshift/hypershift/blob/main/product-cli/main.go#L55)
```go
	if err := cmd.ExecuteContext(ctx); err != nil {
		_, _ = fmt.Fprintf(os.Stderr, "%v\n", err)
		os.Exit(1)
	}
```


so in case of an error, the error is printed three time 😞

```text
⬢ [go-dev] ❯ ./bin/hcp create cluster azure
{"level":"error","ts":"2026-01-20T12:37:17+01:00","msg":"Failed to create cluster","error":"--name is required","stacktrace":"github.com/openshift/hypershift/product-cli/cmd/cluster/azure.NewCreateCommand.func2\n\t/home/ahmed/GolandProjects/github.com/openshift/hypershift_ws/capi-bump/hypershift/product-cli/cmd/cluster/azure/create.go:43\ngithub.com/spf13/cobra.(*Command).execute\n\t/home/ahmed/GolandProjects/github.com/openshift/hypershift_ws/capi-bump/hypershift/vendor/github.com/spf13/cobra/command.go:1015\ngithub.com/spf13/cobra.(*Command).ExecuteC\n\t/home/ahmed/GolandProjects/github.com/openshift/hypershift_ws/capi-bump/hypershift/vendor/github.com/spf13/cobra/command.go:1148\ngithub.com/spf13/cobra.(*Command).Execute\n\t/home/ahmed/GolandProjects/github.com/openshift/hypershift_ws/capi-bump/hypershift/vendor/github.com/spf13/cobra/command.go:1071\ngithub.com/spf13/cobra.(*Command).ExecuteContext\n\t/home/ahmed/GolandProjects/github.com/openshift/hypershift_ws/capi-bump/hypershift/vendor/github.com/spf13/cobra/command.go:1064\nmain.main\n\t/home/ahmed/GolandProjects/github.com/openshift/hypershift_ws/capi-bump/hypershift/product-cli/main.go:66\nruntime.main\n\t/home/ahmed/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.11.linux-amd64/src/runtime/proc.go:283"}
Error: --name is required
--name is required
```

product-cli/cmd/nodepool/destroy.go
```diff
@@ -0,0 +1,61 @@
...
+
+	"github.com/spf13/cobra"
+)
+
+type DestroyNodePoolOptions struct {
+	Name      string
+	Namespace string
+}
+
+func NewDestroyCommand() *cobra.Command {
```
devguyio, line 21: so we're now introducing a new "destroy nodepool" CLI for all platforms?

product-cli/cmd/nodepool/destroy.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+
+	nodePool := &hyperv1.NodePool{}
+	if err := client.Get(ctx, types.NamespacedName{Name: o.Name, Namespace: o.Namespace}, nodePool); err != nil {
+		return fmt.Errorf("failed to get NodePool %s/%s: %w", o.Namespace, o.Name, err)
+	}
+
+	if err := client.Delete(ctx, nodePool); err != nil {
+		return fmt.Errorf("failed to delete NodePool %s/%s: %w", o.Namespace, o.Name, err)
+	}
```
devguyio:
```suggestion
	if err := client.Delete(ctx, nodePool); err != nil {
        if apierrors.IsNotFound(err) {
            log.Log.Info("NodePool already deleted or not found", ...)
            return nil 
        }
		return fmt.Errorf("failed to delete NodePool %s/%s: %w", o.Namespace, o.Name, err)
	}
```

I think that's a more common pattern to avoid races between Get and Delete, and also avoid a duplicate API request.

product-cli/cmd/cluster/azure/create.go
```diff
@@ -0,0 +1,50 @@
...
+			return fmt.Errorf("failed to initialize Azure options: %w", err)
+		}
+		return cmd
+	}
+
+	hypershiftazure.BindProductFlags(azureOpts, cmd.Flags())
+
+	_ = cmd.MarkFlagRequired("azure-creds")
+	_ = cmd.MarkPersistentFlagRequired("pull-secret")
+
```
devguyio, line 27: We should make sure those two flags have a `(required)` suffix in the description.

product-cli/cmd/destroy/destroy.go
```diff
@@ -14,6 +15,7 @@ func NewCommand() *cobra.Command {
 	}
 
 	destroyCmd.AddCommand(cluster.NewDestroyCommands())
+	destroyCmd.AddCommand(nodepool.NewDestroyCommand())
```
devguyio, line 18: we'll need to update the `Short` description of the destroy command in line 13 to be 

```text
Commands for destroying HostedClusters and NodePools
```

product-cli/cmd/cluster/azure/destroy.go
```diff
@@ -0,0 +1,36 @@
...
+
+func NewDestroyCommand(opts *core.DestroyOptions) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:          "azure",
+		Short:        "Destroys a HostedCluster and its associated infrastructure on Azure",
+		SilenceUsage: true,
+	}
+
+	opts.AzurePlatform.Location = "eastus"
+	cmd.Flags().StringVar(&opts.AzurePlatform.CredentialsFile, "azure-creds", opts.AzurePlatform.CredentialsFile, util.AzureCredsDescription)
```
devguyio: Many of the flags description need to change because they describe the creation behavior. 

```text
⬢ [go-dev] ❯ ./bin/hcp destroy cluster azure --help 
Destroys a HostedCluster and its associated infrastructure on Azure

Usage:
  hcp destroy cluster azure [flags]

Flags:
      --azure-creds string           Path to an Azure credentials file (JSON format) containing subscription ID, tenant ID, client ID, and client secret. These credentials are used to create and manage Azure resources for the HostedCluster.
  -h, --help                         help for azure
      --location string              Azure region where the cluster and its resources will be created (e.g. eastus, westus2, northeurope). (default "eastus")
      --preserve-resource-group      Keep the resource group after cluster deletion. Only cluster-specific resources within the group will be removed.
      --resource-group-name string   Name of an existing resource group where HostedCluster infrastructure resources will be created. If omitted, a new resource group will be created.
```

cmd/cluster/azure/create.go
bryan-cox: If we do this, I'd prefer to have it as a follow up. The o/r uses this field so changing it now risks breaking the CI tests.

product-cli/main.go
bryan-cox: Done. Fixed the import grouping - k8s imports are now separated from other imports.

---

product-cli/main.go
bryan-cox: Done. Removed the JSON structured logging setup entirely. The product CLI now uses the default logging behavior consistent with other commands.

---

product-cli/cmd/cluster/azure/create.go
bryan-cox: Done. Simplified the pattern - `DefaultOptions()` no longer returns an error since it never fails. This follows the AWS command pattern.

---

product-cli/cmd/cluster/azure/create.go
bryan-cox: Done. Added `opts.ReleaseStream = config.DefaultReleaseStream` to match the AWS create command.

---

cmd/cluster/azure/create.go
bryan-cox: The flag descriptions are now centralized in `cmd/util/azure_flag_descriptions.go` with descriptive constant names. This keeps both the flag name and description together in a single location, and allows the same descriptions to be shared across cluster and nodepool commands. The approach follows the pattern suggested to have both the name and description together in a struct-like organization.

---

cmd/nodepool/azure/create.go
bryan-cox: Done. Fixed the error message - changed from `--enable-encryption-at-host` to `--encryption-at-host` to match the actual flag name.

---

cmd/nodepool/azure/create.go
bryan-cox: Done. Removed the `ImageID` field from `AzurePlatformCreateOptions` and the corresponding code in `NodePoolPlatform()` that handled the ImageID case. The code now always uses the marketplace image path.

---

product-cli/cmd/cluster/azure/destroy.go
bryan-cox: Done. Added destroy-specific flag descriptions:
- `LocationDestroyDescription`: "Azure region of the cluster. Inferred from the HostedCluster if it exists; only required if the cluster resource has already been deleted."
- `AzureCredsDestroyDescription`: "Path to an Azure credentials file (JSON format) used to authenticate and delete Azure resources."
- `ResourceGroupNameDestroyDescription`: Similar pattern for resource group.

---

product-cli/main.go
bryan-cox: Acknowledged. The JSON logging setup has been removed. The broader UX issue with error messages being printed multiple times is outside the scope of this PR, but noted for future improvement.

---

bryan-cox: Done. Switched to using the default logging (removed the JSON encoder setup). This is consistent with `cmd/log/log.go` convention.

The broader UX improvement (reducing error duplication) is noted as a separate improvement outside this PR's scope.

---

product-cli/cmd/cluster/azure/create.go
bryan-cox: Acknowledged. The triple error printing is a broader UX issue that exists across the product-cli commands. This is outside the scope of this PR but noted as a good follow-up improvement.

---

product-cli/cmd/nodepool/destroy.go
bryan-cox: Yes, this introduces a new `destroy nodepool` CLI command for all platforms. It's a platform-agnostic command that simply deletes the NodePool resource by name/namespace. This provides a consistent way to delete nodepools regardless of the underlying cloud platform.

---

product-cli/cmd/nodepool/destroy.go
bryan-cox: Done. Optimized the delete to avoid the Get+Delete race condition. The code now directly calls Delete without first getting the resource, and handles `NotFound` gracefully:

```go
if err := client.Delete(ctx, nodePool); err != nil {
    if apierrors.IsNotFound(err) {
        log.Log.Info("NodePool already deleted or not found", ...)
        return nil
    }
    return fmt.Errorf("failed to delete NodePool %s/%s: %w", ...)
}
```

---

product-cli/cmd/cluster/azure/create.go
bryan-cox: Done. Added "(required)" suffix to the `AzureCredsDescription` constant. The flags are also marked as required via `MarkFlagRequired` which Cobra enforces.

---

product-cli/cmd/destroy/destroy.go
bryan-cox: Done. Updated the Short description to "Commands for destroying HostedClusters and NodePools".

---

product-cli/cmd/cluster/azure/destroy.go
bryan-cox: Done. Added destroy-specific flag descriptions in `cmd/util/azure_flag_descriptions.go`:
- `LocationDestroyDescription`
- `AzureCredsDestroyDescription`
- `ResourceGroupNameDestroyDescription`

The destroy command now uses these descriptions which accurately reflect their purpose in a destroy context.

---

cmd/cluster/azure/create.go
bryan-cox: Agreed that unifying the flag name would be cleaner. As Bryan mentioned, this will be addressed as a follow-up to avoid breaking the existing CI tests that use the `--assign-service-principal-roles` flag. The current approach keeps backwards compatibility while we plan the deprecation.

---

cmd/cluster/azure/create.go
devguyio: this comment wasn't addressed

cmd/cluster/azure/create.go
devguyio: we can definitely work on deprecating the `--assign-service-principal-roles` in a follow-up PR, but can we introduce the correct, final flag for hcp in this PR? 

What do you think of `--auto-assign-identity-roles` ?

bryan-cox: I really don't see a value in changing it

bryan-cox: If it helps to get this PR moved along though I will.

product-cli/cmd/cluster/azure/create.go
devguyio: I think we can within the scope of this PR since it's introducing a customer-facing CLI for the first time and the changes are minor, let's at least remove the log on line 37 since the contract of RunE is that errors will be logged. Thus reducing one redundant logging, and leaving this code clean from day 0. 
Cause then the next cleanup would be in the future to cleanup the main.go

cmd/cluster/azure/create.go
devguyio: I just think now is the good time to introduce the best cli option for our customers. If you are saying that you prefer `--assign-identity-roles` over `--auto-assign-identity-roles` then sure, I know naming is hard 😄 and the current option is good enough for me too 👍🏽

bryan-cox: Yeah it makes sense to change it now while we are fixing other things. I will make it so.

cmd/cluster/azure/create.go
bryan-cox: Done. Centralized all Azure flag descriptions in `cmd/util/azure_flag_descriptions.go` and updated `bindCoreOptions` to use these constants, ensuring consistent convention across both functions.

---

product-cli/cmd/cluster/azure/create.go
bryan-cox: Done. Removed the redundant `opts.Log.Error` calls from both `product-cli/cmd/cluster/azure/create.go` and `product-cli/cmd/cluster/azure/destroy.go`. Errors are now simply returned, letting Cobra handle the logging per the RunE contract.

---

cmd/cluster/azure/create.go
bryan-cox: Done. Renamed `--assign-identity-roles` to `--auto-assign-roles` in the product CLI for a cleaner, more consistent naming convention.

---

PR #7412, opened 2025-12-18
Add instrumentation for contextual logging to include service provider keys

deads2k:
We will use this to correlate our logs and have more efficient debugging capability between teams

/assign @sjenning 

if this is agreeable, I'll search out all controllers and update them.  There will likely be a followup later to log using JSON so we can more easily extract the keys.


hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -332,6 +330,22 @@ func pauseHostedControlPlane(ctx context.Context, c client.Client, hcp *hyperv1.
 	return nil
 }
 
+// addAnnotationContext takes annotations, extracts context desired by the service provider, and adds that context to the logger.
+// This is useful when the service provider has keys like resourceGroupName, resourceName, hcpClusterName, clusterServiceID
+// and wants to be able to select all log lines that contain those keys.
+// We use annotations because they can hold more values and are applicable to all resource types.
+func addAnnotationContext(log logr.Logger, annotations map[string]string) logr.Logger {
```
deads2k: Per Cesar, move to support/logcontext

PR #7417, opened 2025-12-22
OCPBUGS-65687: fix(controlplane-component): prevent informer creation for unused platform resources

muraee:

When components use .WithPredicate() (e.g., azure-cloud-controller-manager), non-matching platforms trigger cleanup mode which calls Client.Get() to check for existing resources before deletion. For platform-specific resource types like SecretProviderClass, this causes controller-runtime to create informers that continuously LIST/WATCH resources that will never exist on non-Azure platforms.

Track whether resources have been successfully applied and skip delete operations entirely for resources that were never created, avoiding the Client.Get() call that triggers unnecessary informer creation.

Fixes 

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved component state tracking to prevent deletion attempts on resources that haven't been successfully applied.
  * Enhanced deletion logic to safely handle unapplied resources and avoid deletion errors.

### DIFF
--- a/support/controlplane-component/controlplane-component.go
+++ b/support/controlplane-component/controlplane-component.go
@@ -144,6 +144,9 @@ type controlPlaneWorkload[T client.Object] struct {
 
 	customOperandsRolloutCheck   func(cpContext WorkloadContext) (bool, error)
 	monitorOperandsRolloutStatus bool
+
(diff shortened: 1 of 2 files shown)


muraee:
superseded by https://github.com/openshift/hypershift/pull/7819

PR #6780, opened 2025-09-09
[release-4.19] OCPBUGS-61407: fix: Do not admit OAuth route by private router unless…

mgencur:
… it has external DNS

Manual cherry-pick of https://github.com/openshift/hypershift/pull/6218

**What this PR does / why we need it**:

* Label the OAuth Route for admission by the private router only if the OAuth endpoint has hostname (external DNS). This fixes the case when KAS is exposed as LoadBalancer and OAuth via Route without specifying external DNS.
* In order to fix the previous point, label routes for admission by the private router independently from each other (previously, all routes were labeled based on KAS Route)
* Expose private HCP router service if the endpoint access is Private or any endpoint exposes Route with external DNS. (previously it took into account just KAS Route). This fixes the missing routerCanonicalHostname on OAuth Route when KAS has LoadBalancer strategy and OAuth has Route *with* external DNS.
* Add unit test for renconciling OAuth Service and Routes

**Which issue(s) this PR fixes** *(optional, use `fixes #<issue_number>(, fixes #<issue_number>, ...)` format, where issue_number might be a GitHub issue, or a Jira story*:
Fixes #

**Checklist**
- [ ] Subject and description added to both, commit and PR.
- [ ] Relevant issues have been referenced.
- [ ] This change includes docs. 
- [ ] This change includes unit tests.

PR #7418, opened 2025-12-23
OCPBUGS-70147: skip HCP router LB when routes use apps domain

celebdor:

Fix regression in 4.19 where HCP router LoadBalancer service was created unexpectedly for configurations using Route hostnames under the management cluster's apps domain.

After upgrading HCP clusters from 4.18 to 4.19, a `router` service of type LoadBalancer is created unexpectedly. On platforms with limited LoadBalancer IPs (e.g., BareMetal with exhausted MetalLB IPAddressPool), the service stays in `<pending>` state and blocks the upgrade.

The fix for OCPBUGS-56914 (PR #6780) changed the condition for creating the public router service from checking only if KAS uses Route to checking if ANY service uses Route with hostname. This caused the router service to be created for configurations that previously didn't need it.

This PR adds `IsPublicWithExternalDNS()` function that checks if any service uses a Route with a hostname that is **external to** the management cluster's default ingress domain. When hostnames are subdomains of the apps domain (e.g., `oauth.apps.mgmt-cluster.example.com`), the management cluster's default router can serve them via wildcard DNS, so no dedicated HCP router LoadBalancer service is needed.

This preserves the OCPBUGS-56914 fix for external DNS users while preventing unnecessary router service creation for apps-domain routes.

Fixes OCPBUGS-70147

**Checklist**

🤖 Generated with [Claude Code](https://claude.com/claude-code) via `/jira:solve OCPBUGS-70147 celebdor`


celebdor:
Closing in favor of https://github.com/openshift/hypershift/pull/7605

PR #7427, opened 2026-01-05
OCPBUGS-70320: enable scale-up from zero with InPlace upgrade

LiangquanLi930:
Enable scale-up from zero with InPlace upgrade. Remove machinedeployment controller and refactor scale-from-zero using instancetype provider abstraction.

Fixes: OCPBUGS-70320

## Checklist:

### DIFF
--- a/cmd/install/install.go
+++ b/cmd/install/install.go
@@ -20,7 +20,6 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"slices"
 	"strconv"
 	"strings"
 	"time"
@@ -142,6 +141,11 @@ type Options struct {
 func (o *Options) Validate() error {
 	var errs []error
 
+	o.ScaleFromZeroProvider = strings.TrimSpace(o.ScaleFromZeroProvider)
+	if len(o.ScaleFromZeroProvider) != 0 {
+		o.ScaleFromZeroProvider = strings.ToLower(o.ScaleFromZeroProvider)
+	}
+
 	switch hyperv1.PlatformType(o.PrivatePlatform) {
 	case hyperv1.AWSPlatform:
 		if (len(o.AWSPrivateCreds) == 0 && len(o.AWSPrivateCredentialsSecret) == 0) || len(o.AWSPrivateRegion) == 0 {
@@ -191,7 +195,7 @@ func (o *Options) Validate() error {
 	}
 
 	// Validate scale-from-zero credentials
-	supportedProviders := []string{"aws"}
+	supportedProviders := set.New("aws")
 	if len(o.ScaleFromZeroCreds) != 0 || len(o.ScaleFromZeroCredentialsSecret) != 0 {
 		// Check mutual exclusivity - only one of file or secret should be provided
 		if len(o.ScaleFromZeroCreds) != 0 && len(o.ScaleFromZeroCredentialsSecret) != 0 {
@@ -201,8 +205,8 @@ func (o *Options) Validate() error {
 		// Provider is required when using scale-from-zero credentials
 		if len(o.ScaleFromZeroProvider) == 0 {
 			errs = append(errs, fmt.Errorf("--scale-from-zero-provider is required when using scale-from-zero credentials"))
-		} else if !slices.Contains(supportedProviders, strings.ToLower(o.ScaleFromZeroProvider)) {
-			errs = append(errs, fmt.Errorf("invalid --scale-from-zero-provider: %s (must be one of: %v)", o.ScaleFromZeroProvider, supportedProviders))
+		} else if !supportedProviders.Has(o.ScaleFromZeroProvider) {
+			errs = append(errs, fmt.Errorf("invalid --scale-from-zero-provider: %s (must be one of: %v)", o.ScaleFromZeroProvider, supportedProviders.UnsortedList()))
 		}
 
 		// Validate credentials file exists and is accessible if provided
--- a/cmd/install/install_test.go
+++ b/cmd/install/install_test.go
@@ -142,14 +142,18 @@ func TestOptions_Validate(t *testing.T) {
 		},
 		"when scale-from-zero provider is missing but creds provided it errors": {
 			inputOptions: Options{
-				PrivatePlatform:    string(hyperv1.NonePlatform),
+				PrivatePlatform:    string(hyperv1.AWSPlatform),
+				AWSPrivateCreds:    "/dev/null",
+				AWSPrivateRegion:   "us-east-1",
 				ScaleFromZeroCreds: "/path/to/creds",
 			},
 			expectError: true,
 		},
 		"when scale-from-zero provider is invalid it errors": {
 			inputOptions: Options{
-				PrivatePlatform:       string(hyperv1.NonePlatform),
+				PrivatePlatform:       string(hyperv1.AWSPlatform),
+				AWSPrivateCreds:       "/dev/null",
+				AWSPrivateRegion:      "us-east-1",
 				ScaleFromZeroProvider: "gcp",
 				ScaleFromZeroCreds:    "/path/to/creds",
 			},
@@ -157,7 +161,9 @@ func TestOptions_Validate(t *testing.T) {
 		},
 		"when scale-from-zero both creds and secret provided it errors": {
 			inputOptions: Options{
-				PrivatePlatform:                string(hyperv1.NonePlatform),
+				PrivatePlatform:                string(hyperv1.AWSPlatform),
+				AWSPrivateCreds:                "/dev/null",
+				AWSPrivateRegion:               "us-east-1",
 				ScaleFromZeroProvider:          "aws",
 				ScaleFromZeroCreds:             "/path/to/creds",
 				ScaleFromZeroCredentialsSecret: "my-secret",
@@ -166,7 +172,9 @@ func TestOptions_Validate(t *testing.T) {
 		},
 		"when scale-from-zero provider is aws with creds file there is no error": {
 			inputOptions: Options{
-				PrivatePlatform:       string(hyperv1.NonePlatform),
+				PrivatePlatform:       string(hyperv1.AWSPlatform),
+				AWSPrivateCreds:       "/dev/null",
+				AWSPrivateRegion:      "us-east-1",
 				ScaleFromZeroProvider: "aws",
 				ScaleFromZeroCreds:    "/dev/null", // Use /dev/null as it always exists
 			},
@@ -174,7 +182,9 @@ func TestOptions_Validate(t *testing.T) {
 		},
 		"when scale-from-zero provider is aws with secret reference there is no error": {
 			inputOptions: Options{
-				PrivatePlatform:                   string(hyperv1.NonePlatform),
+				PrivatePlatform:                   string(hyperv1.AWSPlatform),
+				AWSPrivateCreds:                   "/dev/null",
+				AWSPrivateRegion:                  "us-east-1",
 				ScaleFromZeroProvider:             "aws",
 				ScaleFromZeroCredentialsSecret:    "my-secret",
 				ScaleFromZeroCredentialsSecretKey: "credentials",
--- a/hypershift-operator/controllers/nodepool/instancetype/aws/provider.go
+++ b/hypershift-operator/controllers/nodepool/instancetype/aws/provider.go
@@ -4,10 +4,13 @@ import (
 	"context"
 	"fmt"
 
+	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
 	"github.com/openshift/hypershift/hypershift-operator/controllers/nodepool/instancetype"
 
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/aws/aws-sdk-go/service/ec2/ec2iface"
+
+	"k8s.io/utils/ptr"
 )
 
 // Provider implements the instancetype.Provider interface for AWS.
@@ -58,11 +61,11 @@ func transformInstanceTypeInfo(rawInstanceType *ec2.InstanceTypeInfo) (*instance
 	if rawInstanceType == nil {
 		return nil, fmt.Errorf("rawInstanceType is nil")
 	}
-	if rawInstanceType.InstanceType == nil {
-		return nil, fmt.Errorf("instance type name is missing")
-	}
 
-	instanceTypeName := *rawInstanceType.InstanceType
+	instanceTypeName := ptr.Deref(rawInstanceType.InstanceType, "")
+	if instanceTypeName == "" {
+		return nil, fmt.Errorf("instance type name is missing or empty")
+	}
 	info := &instancetype.InstanceTypeInfo{
 		InstanceType: instanceTypeName,
 	}
@@ -88,25 +91,23 @@ func transformInstanceTypeInfo(rawInstanceType *ec2.InstanceTypeInfo) (*instance
 	info.MemoryMb = memoryMb
 
 	// Extract GPU information (optional, defaults to 0)
-	if rawInstanceType.GpuInfo != nil && len(rawInstanceType.GpuInfo.Gpus) > 0 {
-		info.GPU = getGpuCount(rawInstanceType.GpuInfo)
-	}
+	info.GPU = getGpuCount(rawInstanceType.GpuInfo)
 
-	// Extract and normalize CPU architecture (defaults to amd64)
-	if rawInstanceType.ProcessorInfo != nil && len(rawInstanceType.ProcessorInfo.SupportedArchitectures) > 0 {
-		architecture := *rawInstanceType.ProcessorInfo.SupportedArchitectures[0]
-		switch architecture {
-		case ec2.ArchitectureTypeX8664:
-			info.CPUArchitecture = instancetype.ArchAMD64
-		case ec2.ArchitectureTypeArm64:
-			info.CPUArchitecture = instancetype.ArchARM64
-		default:
-			// Default to amd64 for unknown architectures
-			info.CPUArchitecture = instancetype.ArchAMD64
-		}
-	} else {
-		// Default to amd64 if architecture information is not available
-		info.CPUArchitecture = instancetype.ArchAMD64
+	// Extract and normalize CPU architecture
+	if rawInstanceType.ProcessorInfo == nil || len(rawInstanceType.ProcessorInfo.SupportedArchitectures) == 0 {
+		return nil, fmt.Errorf("missing CPU architecture information for instance type %q", instanceTypeName)
+	}
+	if rawInstanceType.ProcessorInfo.SupportedArchitectures[0] == nil {
+		return nil, fmt.Errorf("CPU architecture is nil for instance type %q", instanceTypeName)
+	}
+	architecture := *rawInstanceType.ProcessorInfo.SupportedArchitectures[0]
+	switch architecture {
+	case ec2.ArchitectureTypeX8664:
+		info.CPUArchitecture = hyperv1.ArchitectureAMD64
+	case ec2.ArchitectureTypeArm64:
+		info.CPUArchitecture = hyperv1.ArchitectureARM64
+	default:
+		return nil, fmt.Errorf("unsupported CPU architecture %q for instance type %q", architecture, instanceTypeName)
 	}
 
 	return info, nil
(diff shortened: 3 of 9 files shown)


hypershift-operator/controllers/nodepool/instancetype/instancetype.go
```diff
@@ -0,0 +1,45 @@
+package instancetype
+
+import (
+	"context"
+)
+
+// CPUArch represents a normalized CPU architecture type.
+type CPUArch string
+
+const (
```
bryan-cox: These are already defined in the API so no need to redefine them here

hypershift-operator/controllers/nodepool/nodepool_controller.go
```diff
@@ -1096,6 +1112,98 @@ func deleteConfigByLabel(ctx context.Context, c client.Client, lbl map[string]st
 	return nil
 }
 
+// reconcileScaleFromZeroAnnotations sets scale-from-zero annotations on MachineDeployment/MachineSet for AWS platform.
+func (r *NodePoolReconciler) reconcileScaleFromZeroAnnotations(ctx context.Context, nodePool *hyperv1.NodePool, controlPlaneNamespace string) error {
```
bryan-cox:
```
 1. DRY Violation: Duplicated Code in reconcileScaleFromZeroAnnotations

  File: hypershift-operator/controllers/nodepool/nodepool_controller.go:1115-1204

  Description: The MachineDeployment and MachineSet handling logic is nearly identical (90%+ duplication). Both branches:
  1. Get the resource by key
  2. Initialize annotations map
  3. Check infrastructure ref kind
  4. Fetch AWSMachineTemplate
  5. Set annotations
  6. Update the resource

  Current Code Pattern:
  if nodePool.Spec.Management.UpgradeType == hyperv1.UpgradeTypeInPlace {
      // InPlace mode uses MachineSet - 40+ lines of code
      ms := &capiv1.MachineSet{}
      // ... same pattern ...
  } else {
      // Replace mode uses MachineDeployment - 40+ lines of code
      md := &capiv1.MachineDeployment{}
      // ... same pattern ...
  }

  Recommendation: Extract common logic into a helper function using generics or an interface:

  type machineResource interface {
      client.Object
      GetAnnotations() map[string]string
      SetAnnotations(map[string]string)
      GetInfrastructureRef() *corev1.ObjectReference
  }

  func (r *NodePoolReconciler) setScaleFromZeroAnnotations(
      ctx context.Context,
      resource machineResource,
      nodePool *hyperv1.NodePool,
      controlPlaneNamespace string,
  ) error {
      // Unified logic here
  }

  Effort: Medium
```

hypershift-operator/controllers/nodepool/nodepool_controller.go
```diff
@@ -1096,6 +1112,98 @@ func deleteConfigByLabel(ctx context.Context, c client.Client, lbl map[string]st
...
+		templateKey := client.ObjectKey{
+			Namespace: cmp.Or(ms.Spec.Template.Spec.InfrastructureRef.Namespace, controlPlaneNamespace),
+			Name:      ms.Spec.Template.Spec.InfrastructureRef.Name,
+		}
+		if err := r.Get(ctx, templateKey, awsMachineTemplate); err != nil {
+			return fmt.Errorf("failed to get AWSMachineTemplate: %w", err)
+		}
+
+		// Set annotations on the MachineSet
+		if err := setScaleFromZeroAnnotationsOnObject(ctx, r.InstanceTypeProvider, nodePool, ms.Annotations, awsMachineTemplate); err != nil {
```
bryan-cox:
```
  2. Missing Error Context in reconcileScaleFromZeroAnnotations

  File: hypershift-operator/controllers/nodepool/nodepool_controller.go:419

  Description: When reconcileScaleFromZeroAnnotations fails, the error is logged but the reconciliation still returns nil, silently ignoring the failure:

  if err = r.reconcileScaleFromZeroAnnotations(ctx, nodePool, controlPlaneNamespace); err != nil {
      log.Error(err, "Failed to set scale-from-zero annotations")
      // Error is swallowed - no return, no condition update
  }

  Impact: Scale-from-zero failures are silent and won't trigger retries or surface in NodePool conditions.

  Recommendation: Either:
  1. Return the error to trigger a retry, OR
  2. Update a NodePool condition to surface the issue, OR
  3. Use RequeueAfter for graceful retry

  if err = r.reconcileScaleFromZeroAnnotations(ctx, nodePool, controlPlaneNamespace); err != nil {
      log.Error(err, "Failed to set scale-from-zero annotations, will retry")
      return ctrl.Result{RequeueAfter: 30 * time.Second}, nil
  }

  Effort: Easy
```

hypershift-operator/main.go
```diff
@@ -409,6 +421,22 @@ func run(ctx context.Context, opts *StartOptions, log logr.Logger) error {
 
 	npmetrics.CreateAndRegisterNodePoolsMetricsCollector(mgr.GetClient(), ec2Client)
 
+	var instanceTypeProvider instancetype.Provider
+
+	if opts.ScaleFromZeroCreds != "" && opts.ScaleFromZeroProvider != "" {
+		switch strings.ToLower(opts.ScaleFromZeroProvider) {
+		case "aws":
+			awsSession := awsutil.NewSession("hypershift-operator-scale-from-zero", opts.ScaleFromZeroCreds, "", "", "")
```
bryan-cox, line 429:
```
  3. Potential Race Condition with EC2 Client Creation

  File: hypershift-operator/main.go:424-433

  Description: A new EC2 client is created for scale-from-zero without specifying a region:

  awsSession := awsutil.NewSession("hypershift-operator-scale-from-zero", opts.ScaleFromZeroCreds, "", "", "")
  awsConfig := awsutil.NewConfig()
  scaleFromZeroEC2Client := ec2.New(awsSession, awsConfig)

  Impact: The DescribeInstanceTypes API call may use the wrong region, or fail if no default region is configured.

  Note: The old MachineDeployment controller had a RegionCache that extracted region from the AWSMachineTemplate. This PR appears to rely on the credentials file having region or default SDK behavior.

  Recommendation: Document the region selection behavior or add explicit region handling.

  Effort: Medium
```

hypershift-operator/controllers/nodepool/nodepool_controller.go
```diff
@@ -406,9 +408,23 @@ func (r *NodePoolReconciler) reconcile(ctx context.Context, hcluster *hyperv1.Ho
...
+		if err = r.reconcileScaleFromZeroAnnotations(ctx, nodePool, controlPlaneNamespace); err != nil {
+			log.Error(err, "Failed to set scale-from-zero annotations")
+		}
+	}
+
 	return ctrl.Result{}, nil
 }
 
+// supportedScaleFromZeroPlatform checks if the platform supports scale-from-zero functionality.
+func supportedScaleFromZeroPlatform(platform hyperv1.PlatformType) bool {
```
bryan-cox, line 424:
```
  4. SOLID: Open/Closed Principle - supportedScaleFromZeroPlatform hardcoded

  File: hypershift-operator/controllers/nodepool/nodepool_controller.go:422-424

  Description: Platform support check is hardcoded:

  func supportedScaleFromZeroPlatform(platform hyperv1.PlatformType) bool {
      return platform == hyperv1.AWSPlatform
  }

  Current Pattern: When adding Azure support, this function must be modified.

  Recommendation: Consider a registry pattern for extensibility:

  var scaleFromZeroPlatforms = map[hyperv1.PlatformType]bool{
      hyperv1.AWSPlatform: true,
      // hyperv1.AzurePlatform: true,  // Easy to add
  }

  func supportedScaleFromZeroPlatform(platform hyperv1.PlatformType) bool {
      return scaleFromZeroPlatforms[platform]
  }

  Effort: Trivial
```

hypershift-operator/controllers/nodepool/scale_from_zero.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+	// Sort for deterministic output
+	sort.Strings(parts)
+	return strings.Join(parts, ",")
+}
+
+// setScaleFromZeroAnnotationsOnObject sets scale-from-zero workaround annotations on MachineDeployment or MachineSet.
+// This is called during resource creation/update to provide capacity information
+// for cluster-autoscaler when CAPA doesn't support Status.Capacity natively.
+func setScaleFromZeroAnnotationsOnObject(ctx context.Context, provider instancetype.Provider, nodePool *hyperv1.NodePool, annotations map[string]string, awsMachineTemplate *infrav1.AWSMachineTemplate) error {
```
bryan-cox:
```
  5. setScaleFromZeroAnnotationsOnObject is AWS-specific

  File: hypershift-operator/controllers/nodepool/scale_from_zero.go:60

  Description: The function signature is tightly coupled to AWS:

  func setScaleFromZeroAnnotationsOnObject(ctx context.Context, provider instancetype.Provider,
      nodePool *hyperv1.NodePool, annotations map[string]string,
      awsMachineTemplate *infrav1.AWSMachineTemplate) error {

  Impact: When adding Azure support, this function would need significant changes or duplication.

  Recommendation: Consider a more generic interface:

  type MachineTemplateAccessor interface {
      GetInstanceType() string
      HasStatusCapacity() bool
  }

  Effort: Medium
```

hypershift-operator/controllers/nodepool/scale_from_zero.go
```diff
@@ -0,0 +1,120 @@
...
+	// 0. Check if CAPA already provides Status.Capacity (prefer native support)
+	if len(awsMachineTemplate.Status.Capacity) > 0 {
+		// Clean up workaround annotations (if they exist)
+		annotationKeys := []string{cpuKey, memoryKey, gpuKey, labelsKey, taintsKey}
```
bryan-cox, line 94:
```
  6. Inconsistent annotation key cleanup logic

  File: hypershift-operator/controllers/nodepool/scale_from_zero.go:63-70

  Description: When CAPA provides Status.Capacity, annotations are deleted individually:

  annotationKeys := []string{cpuKey, memoryKey, gpuKey, labelsKey, taintsKey}
  for _, key := range annotationKeys {
      delete(annotations, key)
  }

  But at line 116, only taintsKey is conditionally deleted:

  if len(nodePool.Spec.Taints) > 0 {
      annotations[taintsKey] = taintsAnnotation
  } else {
      delete(annotations, taintsKey)
  }

  Impact: Minor inconsistency; GPU annotation would remain even if GPU=0 in the new state.

  Effort: Trivial
```

hypershift-operator/controllers/nodepool/instancetype/aws/provider.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+// transformInstanceTypeInfo converts EC2 InstanceTypeInfo to our common InstanceTypeInfo structure.
+// It extracts and validates CPU, memory, GPU, and architecture information from the AWS response.
+func transformInstanceTypeInfo(rawInstanceType *ec2.InstanceTypeInfo) (*instancetype.InstanceTypeInfo, error) {
+	// Validate raw instance type data
+	if rawInstanceType == nil {
+		return nil, fmt.Errorf("rawInstanceType is nil")
+	}
+	if rawInstanceType.InstanceType == nil {
```
bryan-cox:
```
  8. Missing validation for empty instanceType string in transformInstanceTypeInfo

  File: hypershift-operator/controllers/nodepool/instancetype/aws/provider.go:65

  Description: The code checks if rawInstanceType.InstanceType == nil but not if it's an empty string:

  if rawInstanceType.InstanceType == nil {
      return nil, fmt.Errorf("instance type name is missing")
  }
  instanceTypeName := *rawInstanceType.InstanceType  // Could be ""

  Effort: Trivial
```

cmd/install/install.go
```diff
@@ -189,19 +191,27 @@ func (o *Options) Validate() error {
 	}
 
 	// Validate scale-from-zero credentials
-	if len(o.ScaleFromZeroAWSCreds) != 0 || len(o.ScaleFromZeroAWSCredentialsSecret) != 0 {
+	supportedProviders := []string{"aws"}
```
muraee: should use a `Set` instead of a slice

hypershift-operator/controllers/nodepool/nodepool_controller.go
muraee: looks good as is, imo.

hypershift-operator/controllers/nodepool/nodepool_controller.go
LiangquanLi930: Thanks @bryan-cox 
Agree there's duplication, but I think the current code is clearer for only 2 cases. If we add a third upgrade type, that would be a good time to refactor.

hypershift-operator/main.go
LiangquanLi930: Thanks @bryan-cox!

IMHO, this is acceptable because instance type specs (CPU/memory/GPU) are consistent across regions - we're only querying metadata, not creating resources. If no region is configured, the AWS SDK will fail with a clear error during reconciliation.

cmd/install/assets/hypershift_operator.go
```diff
@@ -154,20 +154,21 @@ func (o HyperShiftOperatorOIDCProviderS3Secret) Build() *corev1.Secret {
 	return secret
 }
 
-type ScaleFromZeroAWSCredentialsSecret struct {
+type ScaleFromZeroCredentialsSecret struct {
 	Namespace  *corev1.Namespace
 	CredsBytes []byte
 	CredsKey   string
+	Provider   string
```
muraee, line 161: why do we need this? I prefer if we keep the old `-scale-aws-creds` flag, and if we ever wanted to add support for other providers, we will add new creds flag like `--scale-azure-creds`, etc.

hypershift-operator/controllers/nodepool/nodepool_controller.go
muraee: agreed, no need for generics. However, the current code can still be refactored to avoid some duplication. There is an existing function that does this https://github.com/openshift/hypershift/blob/main/hypershift-operator/controllers/nodepool/capi.go#L1195

```go
func (r *NodePoolReconciler) reconcileScaleFromZeroAnnotations(ctx context.Context, nodePool *hyperv1.NodePool, capi *CAPI) error {
  awsMachineTemplate := &capiaws.AWSMachineTemplate{}
  if err := capi.GetExistingMachineTemplate(ctx, awsMachineTemplate); err != nil {
    return err
  }
  
  // setScaleFromZeroAnnotationsOnObject()
  // update()
....
}
```

cmd/install/assets/hypershift_operator.go
LiangquanLi930: When we switch platforms—for example, to Azure: we can use
```
--scale-from-zero-provider azure --scale-from-zero-creds xxxx
```
This can significantly reduce the number of flags; currently, --external-dns-xxxx is handled in the same way.

cmd/install/install_test.go
```diff
@@ -140,6 +140,47 @@ func TestOptions_Validate(t *testing.T) {
 			},
 			expectError: true,
 		},
+		"when scale-from-zero provider is missing but creds provided it errors": {
```
bryan-cox, line 143: Shouldn't these all test against the AWS platform vs none?

hypershift-operator/controllers/nodepool/instancetype/aws/provider.go
```diff
@@ -0,0 +1,128 @@
...
+
+	// Extract and normalize CPU architecture (defaults to amd64)
+	if rawInstanceType.ProcessorInfo != nil && len(rawInstanceType.ProcessorInfo.SupportedArchitectures) > 0 {
+		architecture := *rawInstanceType.ProcessorInfo.SupportedArchitectures[0]
+		switch architecture {
+		case ec2.ArchitectureTypeX8664:
+			info.CPUArchitecture = hyperv1.ArchitectureAMD64
+		case ec2.ArchitectureTypeArm64:
+			info.CPUArchitecture = hyperv1.ArchitectureARM64
+		default:
```
bryan-cox: We should not hit this but we should return an error instead imo rather than assuming amd64.

hypershift-operator/controllers/nodepool/instancetype/aws/provider.go
```diff
@@ -0,0 +1,128 @@
...
+			// Default to amd64 for unknown architectures
+			info.CPUArchitecture = hyperv1.ArchitectureAMD64
+		}
+	} else {
```
bryan-cox: Same here, we should return an error if the info is not available.

hypershift-operator/controllers/nodepool/nodepool_controller.go
```diff
@@ -1096,6 +1112,73 @@ func deleteConfigByLabel(ctx context.Context, c client.Client, lbl map[string]st
 	return nil
 }
 
+// reconcileScaleFromZeroAnnotations sets scale-from-zero annotations on MachineDeployment/MachineSet for AWS platform.
```
bryan-cox: Given this is nodepool_controller.go, the AWS specific bits should be obfuscated away a bit from this function imo into an aws specific provider rather than hardcoding this whole function to just work for AWS.

hypershift-operator/controllers/nodepool/scale_from_zero.go
```diff
@@ -0,0 +1,145 @@
...
+		statusCapacity = template.Status.Capacity
+	// Future platform support can be added here:
+	// case *capiazure.AzureMachineTemplate:
+	//     instanceType = template.Spec.Template.Spec.VMSize
+	//     statusCapacity = template.Status.Capacity
+	default:
+		return fmt.Errorf("unsupported machine template type: %T", machineTemplate)
+	}
+
+	// 1. Check if CAPA already provides Status.Capacity (prefer native support)
```
bryan-cox: Shouldn't we remove the CAPA vernacular from the comment here since other platforms could be used in the future?

hypershift-operator/controllers/nodepool/scale_from_zero.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+	// Sort for deterministic output
+	sort.Strings(parts)
+	return strings.Join(parts, ",")
+}
+
+// setScaleFromZeroAnnotationsOnObject sets scale-from-zero workaround annotations on MachineDeployment or MachineSet.
+// This is called during resource creation/update to provide capacity information
+// for cluster-autoscaler when CAPA doesn't support Status.Capacity natively.
+func setScaleFromZeroAnnotationsOnObject(ctx context.Context, provider instancetype.Provider, nodePool *hyperv1.NodePool, annotations map[string]string, machineTemplate interface{}) error {
```
muraee: passing only the annotations to this function and manipulating it inplace in very confusing and error prone. please pass the complete object here.
```suggestion
func setScaleFromZeroAnnotationsOnObject(ctx context.Context, provider instancetype.Provider, nodePool *hyperv1.NodePool, object client.Object, machineTemplate interface{}) error {
```

hypershift-operator/controllers/nodepool/nodepool_controller.go
```diff
@@ -1096,6 +1112,73 @@ func deleteConfigByLabel(ctx context.Context, c client.Client, lbl map[string]st
...
+	awsMachineTemplate := &capiaws.AWSMachineTemplate{}
+	if err := capi.getExistingMachineTemplate(ctx, awsMachineTemplate); err != nil {
+		if apierrors.IsNotFound(err) {
+			// Machine template doesn't exist yet, skip annotation (it will be reconciled later)
+			return nil
+		}
+		return fmt.Errorf("failed to get AWSMachineTemplate: %w", err)
+	}
+
+	if nodePool.Spec.Management.UpgradeType == hyperv1.UpgradeTypeInPlace {
```
muraee, line 1149:
```go
var obj client.Object
if nodePool.Spec.Management.UpgradeType == hyperv1.UpgradeTypeInPlace {
  ms := capi.machineSet()
   // get ms
   obj = ms
} else {
   // get md 
   obj = md
}

// setScaleFromZeroAnnotationsOnObject(obj)
// Update()
```

cmd/install/install.go
```diff
@@ -189,19 +192,27 @@ func (o *Options) Validate() error {
 	}
 
 	// Validate scale-from-zero credentials
-	if len(o.ScaleFromZeroAWSCreds) != 0 || len(o.ScaleFromZeroAWSCredentialsSecret) != 0 {
+	supportedProviders := map[string]struct{}{"aws": {}}
```
muraee:
```suggestion
	supportedProviders := sets.New("aws")
```

hypershift-operator/controllers/nodepool/nodepool_controller.go
LiangquanLi930: I’ve updated reconcileScaleFromZeroAnnotations, and it now supports multiple platforms.

cmd/install/assets/hypershift_operator.go
jhjaggars: Why do we need to specify the provider at all?  Is that for cases where the platform is otherwise unspecified?

PR #7431, opened 2026-01-06
feat(kubevirt): OCPBUGS-55974: add CPU model configuration for NodePool VMs

orenc1:
Allow customers to specify the CPU model for KubeVirt VMs via a new Model field on `KubevirtCompute`. Currently supports `host-passthrough`, which exposes the host CPU model to the guest VM, enabling full MMIO
space needed by large or multiple vGPUs.

**API Changes:**
- Added `CpuModelType` enum type with `host-passthrough` value
- Added Model field to KubevirtCompute struct

**Controller Changes:**
- Thread the CPU model through to the VirtualMachineInstance template for both Burstable and Guaranteed QoS paths
- Set `KubeVirtNodesLiveMigratable` condition to False when `host-passthrough` is configured, warning users about the live migration limitation
- Guard against setting Cores to zero in the CPU struct when only model is specified

Test Coverage:
- Serialization compatibility test for KubevirtCompute (N-1/N+1)
- CPU model with Burstable QoS
- CPU model with Guaranteed QoS
- CPU model without explicit cores
- CPU model combined with host devices (vGPU use case)

Fixes 

## Checklist:

## New Features
- Added support for selecting the KubeVirt VM CPU model in NodePool compute settings.
- Added the `host-passthrough` option, allowing VMs to use the host’s CPU model for improved performance.
- CPU model settings work with or without explicit core counts and resource guarantees.

## Documentation
- Documented CPU model configuration and its performance and migration implications.

## Important
- VMs using `host-passthrough` may not support live migration.


api/hypershift/v1beta1/kubevirt.go
```diff
@@ -12,6 +12,16 @@ const (
 
 type QoSClass string
 
+// CpuModelType represents the CPU model for KubeVirt VMs.
+//
+// +kubebuilder:validation:Enum=host-passthrough
```
JoelSpeed: In k8s we use PascalCase for enum values. This should be `HostPassthrough`

api/hypershift/v1beta1/kubevirt.go
```diff
@@ -34,6 +44,13 @@ type KubevirtCompute struct {
 	// +kubebuilder:validation:Enum=Burstable;Guaranteed
 	// +kubebuilder:default=Burstable
 	QosClass *QoSClass `json:"qosClass,omitempty"`
+
+	// model specifies the CPU model for the KubeVirt VirtualMachineInstance.
+	// When set to "host-passthrough", the VM will use the same CPU model as the node,
```
JoelSpeed: What happens when I don't set anything for this field?

api/hypershift/v1beta1/kubevirt.go
```diff
@@ -34,6 +44,13 @@ type KubevirtCompute struct {
...
+	// which provides the best performance but may limit live migration compatibility.
+	//
+	// +optional
+	Model *CpuModelType `json:"model,omitempty"`
```
JoelSpeed: `""` is not a valid user choice, so you don't actually need a pointer here

api/hypershift/v1beta1/kubevirt.go
orenc1: when the field is not set, no explicit CPU model is configured on the `VirtualMachineInstance` and kubevirt uses its default behavior. i've updated the godoc to clarify this:
```
// When not set, no explicit CPU model is configured and KubeVirt will use
// its default behavior.
```

api/hypershift/v1beta1/kubevirt.go
orenc1: thanks, fixed.
changed `*CpuModelType` to `CpuModelType`.
since the enum only allows `HostPassthrough`, the empty string zero value is sufficient to distinguish "not set" from "set"

hypershift-operator/controllers/nodepool/kubevirt/kubevirt.go
```diff
@@ -204,10 +220,20 @@ func virtualMachineTemplateBase(nodePool *hyperv1.NodePool, bootImage BootImage)
 
 		template.Spec.Template.Spec.Domain.Resources.Requests = podResources
 		template.Spec.Template.Spec.Domain.Resources.Limits = podResources
+
+		// In the guaranteed QoS path, cores are specified via resource requests/limits
+		// rather than the CPU struct. Only Model belongs in CPU here.
+		if cpuModel != "" {
+			template.Spec.Template.Spec.Domain.CPU = &kubevirtv1.CPU{Model: cpuModel}
+		}
```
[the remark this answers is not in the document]
orenc1: Thanks for flagging this edge case. However, this behavior pre-dates this PR - the `cores` > 0 guard on line 217 and the resulting "Guaranteed QoS without CPU resources" scenario existed before the CPU model feature was added. The model field doesn't change the cores/resource handling at all.

In practice this is also very unlikely to hit because the Compute struct defaults to `{memory: "8Gi", cores: 2}` via the CRD-level default. Improving validation for the Guaranteed + no Cores combination would be a worthwhile follow-up but is out of scope for this PR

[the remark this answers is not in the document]
orenc1: You make a fair point that this PR makes the no-cores path more relevant. I've added a test case ("When CPU model and Guaranteed QoS are set without explicit cores, it should set guaranteed memory and CPU model only") that explicitly documents this behavior as intentional — the VM receives guaranteed memory and the CPU model, but no CPU resource guarantees. The cores > 0 guard is pre-existing behavior and changing it would affect all Guaranteed QoS configurations, not just those with a model. If we want to validate or default cores in the Guaranteed path, that should be a separate PR.

api/hypershift/v1beta1/kubevirt.go
```diff
@@ -34,6 +44,16 @@ type KubevirtCompute struct {
 	// +kubebuilder:validation:Enum=Burstable;Guaranteed
 	// +kubebuilder:default=Burstable
 	QosClass *QoSClass `json:"qosClass,omitempty"`
+
+	// model specifies the CPU model for the KubeVirt VirtualMachineInstance.
```
JoelSpeed, line 48: I'd recommend adding an explicit list of valid values in here
```suggestion
	// model specifies the CPU model for the KubeVirt VirtualMachineInstance.
	// Valid values are "HostPassthrough" and omitted.
```

orenc1: makes sense, thanks.
i've added it.

nunnatsa (approved):
/lgtm

hypershift-operator/controllers/nodepool/kubevirt/kubevirt.go
```diff
@@ -142,12 +142,50 @@ func PlatformValidation(nodePool *hyperv1.NodePool) error {
...
+
+	condition := hyperv1.NodePoolCondition{
+		Type:               hyperv1.NodePoolKubeVirtLiveMigratableType,
+		Status:             corev1.ConditionFalse,
+		Reason:             hyperv1.NodePoolKubeVirtLiveMigratableReason,
+		Message:            "CPU model host-passthrough is configured; VMs using host-passthrough may not be live-migratable",
+		ObservedGeneration: nodePool.Generation,
+	}
+
+	for i, c := range nodePool.Status.Conditions {
```
jparrill: This reimplements condition find-or-update logic that `SetStatusCondition` already handles. The manual loop doesn't set `LastTransitionTime`, so conditions created here will have a zero timestamp — this propagates to HostedCluster level via `hostedcluster_controller.go:5146`.

Could you use `SetStatusCondition` instead? If the package boundary is the issue, one option is to return the condition from this function and let the caller call `SetStatusCondition`.

jparrill (commented):
Overall the PR is well-structured — clean API addition, solid N-1/N+1 serialization tests, correct rollout behavior via template hash change. Main concern is condition management around `ApplyLiveMigrationWarning`. See inline comments for details.

hypershift-operator/controllers/nodepool/kubevirt.go
```diff
@@ -104,6 +104,8 @@ func (r *NodePoolReconciler) setKubevirtConditions(ctx context.Context, nodePool
 
 	r.addKubeVirtCacheNameToStatus(kubevirtBootImage, nodePool)
 
+	kubevirt.ApplyLiveMigrationWarning(nodePool)
```
jparrill: When the model is changed from `HostPassthrough` to empty and replicas are 0, neither `setAllMachinesLMCondition` (no machines, returns early without touching the condition) nor `ApplyLiveMigrationWarning` (model empty, returns early) transitions the stale `KubeVirtNodesLiveMigratable=False` to True.

The root cause is in `setAllMachinesLMCondition` — it does an early return when there are no machines instead of setting the condition to True. Small fix — could you change the early return to set `ConditionTrue` with reason `AsExpected`?

jparrill: This and `setAllMachinesLMCondition` (called earlier in the signal conditions loop) both write to `KubeVirtNodesLiveMigratable`. The current ordering works because this runs second and wins, but it's undocumented. A small comment here noting the ordering dependency would prevent accidental regressions.

hypershift-operator/controllers/nodepool/kubevirt/kubevirt_test.go
```diff
@@ -1622,3 +1924,143 @@ func TestDefaultImage(t *testing.T) {
...
+		t.Run(tt.name, func(t *testing.T) {
+			result := CpuModelToKubevirt(tt.input)
+			if result != tt.expected {
+				t.Errorf("CpuModelToKubevirt(%q) = %q, want %q", tt.input, result, tt.expected)
+			}
+		})
+	}
+}
+
+func TestApplyLiveMigrationWarning(t *testing.T) {
```
jparrill: Missing a test case for the transition: condition was set to False (model was HostPassthrough), then model changed to empty — condition should be cleared. Also in "should update an existing condition", asserting `len(nodePool.Status.Conditions) == 1` would catch duplicate-append bugs.

hypershift-operator/controllers/nodepool/kubevirt.go
orenc1: Changed `setAllMachinesLMCondition` to set `ConditionTrue` with reason `AsExpected` when there are no machines, instead of early-returning without touching the condition. This ensures a stale False condition is cleared when the model changes from `HostPassthrough` to empty and replicas are 0

orenc1: Added a comment at the call site explaining that `LiveMigrationWarningCondition` runs after `setAllMachinesLMCondition` intentionally, because the spec-based warning takes precedence over per-machine status aggregation

hypershift-operator/controllers/nodepool/kubevirt/kubevirt_test.go
orenc1: - Added a test case for the transition scenario (model changed from `HostPassthrough` to empty — should return nil)
- Renamed the test function from `TestApplyLiveMigrationWarning` to `TestLiveMigrationWarningCondition` to match the new function name
- Tests now verify the returned condition directly rather than inspecting NodePool status, since the function no longer mutates it

hypershift-operator/controllers/nodepool/kubevirt/kubevirt.go
orenc1: - Renamed `ApplyLiveMigrationWarning` to `LiveMigrationWarningCondition` which now returns a `*hyperv1.NodePoolCondition` (or nil) instead of directly mutating the NodePool
- The caller in `setKubevirtConditions` now uses ``SetStatusCondition` to apply it, which properly handles LastTransitionTime

jparrill (commented):
/approve

Looks like envtests are failing, do you mind take a look?

orenc1:
> /approve
> 
> Looks like envtests are failing, do you mind take a look?

i don't see any clear reason for it to fail. it's just saying one of the lanes was cancelled. hopefully this can be re-run.
/retest

jparrill:
/override ci/prow/e2e-v2-gke

It's and infra issue, the test already passed in older executions.

jparrill:
/lgtm 

exception just in favor of @nunnatsa which right now it's OoO

PR #7442, opened 2026-01-08
OCPBUGS-72397: Fix Konflux EC voilation, update deprecated base …

rokej:

Base images are deprecated causing Konflux image validations and EC violations. 

Fixes https://issues.redhat.com/browse/OCPBUGS-72397

## Checklist:

### DIFF
--- a/Containerfile.cli
+++ b/Containerfile.cli
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi9/go-toolset:1.25.3-1765311584 AS builder
+FROM registry.redhat.io/rhel9/go-toolset:1.24.6-1763548439 AS builder
 
 WORKDIR /hypershift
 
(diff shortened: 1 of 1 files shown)


rokej:
The failed checks are not related to this change. Approvers, please override to skip them.

Containerfile.cli
```diff
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi9/go-toolset:1.25.3-1765311584 AS builder
+FROM registry.redhat.io/rhel9/go-toolset:1.24.6-1763548439 AS builder
```
celebdor, line 1: Could we stay with the unauthenticated ubi9 image? Maybe bump to `1.25-1767889151`

rokej: @celebdor Actually we are supposed to use brew.registry.redhat.io/rh-osbs/openshift-golang-builder for product containers. I will change to that.

PR #7447, opened 2026-01-08
OCPBUGS-72411: fix(cno): use brackets only for IPv6 in server URL

devguyio:

The rewrite-config init container was unconditionally wrapping `KUBERNETES_SERVICE_HOST` in brackets, creating URLs like `https://[172.29.0.1]:443.`

Go 1.24.8+ (CVE-2025-47912) now rejects IPv4 addresses in brackets per RFC 3986.

This PR fixes the init container shell to detect IPv6 (contains colon) and only add brackets for those addresses.

Fixes: OCPBUGS-72411

I added a unit test that verifies the expected behavior of init-container bash. The comment in the file explains how it works. Here's its output failing before applying the fix:

```
⬢ [go-dev] ❯ go test -v .
=== RUN   TestRewriteConfigInitContainer
=== RUN   TestRewriteConfigInitContainer/When_KUBERNETES_SERVICE_HOST_is_IPv4_it_should_format_URL_without_brackets
    deployment_init_container_test.go:169: kubectl log mismatch (-want +got):
          strings.Join({
          	"--kubeconfig /configs/management config set clusters.default.ser",
          	"ver https://",
        + 	"[",
          	"172.29.0.1",
        + 	"]",
          	":443\n--kubeconfig /configs/management config set clusters.defaul",
          	"t.certificate-authority /var/run/secrets/kubernetes.io/serviceac",
          	... // 1016 identical bytes
          }, "")
=== RUN   TestRewriteConfigInitContainer/When_KUBERNETES_SERVICE_HOST_is_private_IPv4_it_should_format_URL_without_brackets
    deployment_init_container_test.go:169: kubectl log mismatch (-want +got):
          strings.Join({
          	"--kubeconfig /configs/management config set clusters.default.ser",
          	"ver https://",
        + 	"[",
          	"10.0.0.1",
        + 	"]",
          	":443\n--kubeconfig /configs/management config set clusters.defaul",
          	"t.certificate-authority /var/run/secrets/kubernetes.io/serviceac",
          	... // 1016 identical bytes
          }, "")
=== RUN   TestRewriteConfigInitContainer/When_KUBERNETES_SERVICE_HOST_is_IPv6_loopback_it_should_format_URL_with_brackets
=== RUN   TestRewriteConfigInitContainer/When_KUBERNETES_SERVICE_HOST_is_IPv6_ULA_it_should_format_URL_with_brackets
=== RUN   TestRewriteConfigInitContainer/When_KUBERNETES_SERVICE_HOST_is_full_IPv6_it_should_format_URL_with_brackets
=== RUN   TestRewriteConfigInitContainer/When_KUBERNETES_SERVICE_HOST_is_IPv6_mapped_IPv4_it_should_format_URL_with_brackets
=== RUN   TestRewriteConfigInitContainer/When_KUBERNETES_SERVICE_HOST_is_hostname_it_should_format_URL_without_brackets
    deployment_init_container_test.go:169: kubectl log mismatch (-want +got):
          strings.Join({
          	"--kubeconfig /configs/management config set clusters.default.ser",
          	"ver https://",
        + 	"[",
          	"kubernetes.default.svc",
        + 	"]",
          	":443\n--kubeconfig /configs/management config set clusters.defaul",
          	"t.certificate-authority /var/run/secrets/kubernetes.io/serviceac",
          	... // 1016 identical bytes
          }, "")
=== RUN   TestRewriteConfigInitContainer/When_KUBERNETES_SERVICE_HOST_is_hostname_with_subdomain_it_should_format_URL_without_brackets
    deployment_init_container_test.go:169: kubectl log mismatch (-want +got):
          strings.Join({
          	"--kubeconfig /configs/management config set clusters.default.ser",
          	"ver https://",
        + 	"[",
          	"api.example.com",
        + 	"]",
          	":6443\n--kubeconfig /configs/management config set clusters.defau",
          	"lt.certificate-authority /var/run/secrets/kubernetes.io/servicea",
          	... // 1017 identical bytes
          }, "")
--- FAIL: TestRewriteConfigInitContainer (0.01s)
    --- FAIL: TestRewriteConfigInitContainer/When_KUBERNETES_SERVICE_HOST_is_IPv4_it_should_format_URL_without_brackets (0.00s)
    --- FAIL: TestRewriteConfigInitContainer/When_KUBERNETES_SERVICE_HOST_is_private_IPv4_it_should_format_URL_without_brackets (0.00s)
    --- PASS: TestRewriteConfigInitContainer/When_KUBERNETES_SERVICE_HOST_is_IPv6_loopback_it_should_format_URL_with_brackets (0.00s)
    --- PASS: TestRewriteConfigInitContainer/When_KUBERNETES_SERVICE_HOST_is_IPv6_ULA_it_should_format_URL_with_brackets (0.00s)
    --- PASS: TestRewriteConfigInitContainer/When_KUBERNETES_SERVICE_HOST_is_full_IPv6_it_should_format_URL_with_brackets (0.00s)
    --- PASS: TestRewriteConfigInitContainer/When_KUBERNETES_SERVICE_HOST_is_IPv6_mapped_IPv4_it_should_format_URL_with_brackets (0.00s)
    --- FAIL: TestRewriteConfigInitContainer/When_KUBERNETES_SERVICE_HOST_is_hostname_it_should_format_URL_without_brackets (0.00s)
    --- FAIL: TestRewriteConfigInitContainer/When_KUBERNETES_SERVICE_HOST_is_hostname_with_subdomain_it_should_format_URL_without_brackets (0.00s)
FAIL
FAIL	github.com/openshift/hypershift/control-plane-operator/controllers/hostedcontrolplane/v2/cno	0.055s
FAIL

```

## Checklist:

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/testdata/cluster-network-operator/GCP/zz_fixture_TestControlPlaneComponents_cluster_network_operator_deployment.yaml
+++ b/control-plane-operator/controllers/hostedcontrolplane/testdata/cluster-network-operator/GCP/zz_fixture_TestControlPlaneComponents_cluster_network_operator_deployment.yaml
@@ -339,7 +339,13 @@ spec:
           set -xeuo pipefail
 
           kc=/configs/management
-          kubectl --kubeconfig $kc config set clusters.default.server "https://[${KUBERNETES_SERVICE_HOST}]:${KUBERNETES_SERVICE_PORT}"
+          # Use brackets only for IPv6 addresses (contain colons), not for IPv4 or hostnames
+          # This is required for CVE-2025-47912 compliance: Go's net/url rejects IPv4 in brackets
+          case "${KUBERNETES_SERVICE_HOST}" in
+            *:*) server="https://[${KUBERNETES_SERVICE_HOST}]:${KUBERNETES_SERVICE_PORT}" ;;
+            *)   server="https://${KUBERNETES_SERVICE_HOST}:${KUBERNETES_SERVICE_PORT}" ;;
+          esac
+          kubectl --kubeconfig $kc config set clusters.default.server "$server"
           kubectl --kubeconfig $kc config set clusters.default.certificate-authority /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
           kubectl --kubeconfig $kc config set users.admin.tokenFile /var/run/secrets/kubernetes.io/serviceaccount/token
           kubectl --kubeconfig $kc config set contexts.default.cluster default
--- a/control-plane-operator/controllers/hostedcontrolplane/testdata/cluster-network-operator/IBMCloud/zz_fixture_TestControlPlaneComponents_cluster_network_operator_deployment.yaml
+++ b/control-plane-operator/controllers/hostedcontrolplane/testdata/cluster-network-operator/IBMCloud/zz_fixture_TestControlPlaneComponents_cluster_network_operator_deployment.yaml
@@ -307,7 +307,13 @@ spec:
           set -xeuo pipefail
 
           kc=/configs/management
-          kubectl --kubeconfig $kc config set clusters.default.server "https://[${KUBERNETES_SERVICE_HOST}]:${KUBERNETES_SERVICE_PORT}"
+          # Use brackets only for IPv6 addresses (contain colons), not for IPv4 or hostnames
+          # This is required for CVE-2025-47912 compliance: Go's net/url rejects IPv4 in brackets
+          case "${KUBERNETES_SERVICE_HOST}" in
+            *:*) server="https://[${KUBERNETES_SERVICE_HOST}]:${KUBERNETES_SERVICE_PORT}" ;;
+            *)   server="https://${KUBERNETES_SERVICE_HOST}:${KUBERNETES_SERVICE_PORT}" ;;
+          esac
+          kubectl --kubeconfig $kc config set clusters.default.server "$server"
           kubectl --kubeconfig $kc config set clusters.default.certificate-authority /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
           kubectl --kubeconfig $kc config set users.admin.tokenFile /var/run/secrets/kubernetes.io/serviceaccount/token
           kubectl --kubeconfig $kc config set contexts.default.cluster default
--- a/control-plane-operator/controllers/hostedcontrolplane/testdata/cluster-network-operator/TechPreviewNoUpgrade/zz_fixture_TestControlPlaneComponents_cluster_network_operator_deployment.yaml
+++ b/control-plane-operator/controllers/hostedcontrolplane/testdata/cluster-network-operator/TechPreviewNoUpgrade/zz_fixture_TestControlPlaneComponents_cluster_network_operator_deployment.yaml
@@ -306,7 +306,13 @@ spec:
           set -xeuo pipefail
 
           kc=/configs/management
-          kubectl --kubeconfig $kc config set clusters.default.server "https://[${KUBERNETES_SERVICE_HOST}]:${KUBERNETES_SERVICE_PORT}"
+          # Use brackets only for IPv6 addresses (contain colons), not for IPv4 or hostnames
+          # This is required for CVE-2025-47912 compliance: Go's net/url rejects IPv4 in brackets
+          case "${KUBERNETES_SERVICE_HOST}" in
+            *:*) server="https://[${KUBERNETES_SERVICE_HOST}]:${KUBERNETES_SERVICE_PORT}" ;;
+            *)   server="https://${KUBERNETES_SERVICE_HOST}:${KUBERNETES_SERVICE_PORT}" ;;
+          esac
+          kubectl --kubeconfig $kc config set clusters.default.server "$server"
           kubectl --kubeconfig $kc config set clusters.default.certificate-authority /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
           kubectl --kubeconfig $kc config set users.admin.tokenFile /var/run/secrets/kubernetes.io/serviceaccount/token
           kubectl --kubeconfig $kc config set contexts.default.cluster default
--- a/control-plane-operator/controllers/hostedcontrolplane/testdata/cluster-network-operator/zz_fixture_TestControlPlaneComponents_cluster_network_operator_deployment.yaml
+++ b/control-plane-operator/controllers/hostedcontrolplane/testdata/cluster-network-operator/zz_fixture_TestControlPlaneComponents_cluster_network_operator_deployment.yaml
@@ -306,7 +306,13 @@ spec:
           set -xeuo pipefail
 
           kc=/configs/management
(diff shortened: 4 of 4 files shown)


sjenning:
/payload-job periodic-ci-openshift-hypershift-release-4.22-periodics-e2e-aws-ovn-conformance

control-plane-operator/controllers/hostedcontrolplane/v2/cno/deployment_init_container_test.go
```diff
@@ -0,0 +1,205 @@
...
+
+	"github.com/pmezard/go-difflib/difflib"
+)
+
+// expectedKubectlLogTemplate is a template of the expected calls of kubectl. The {{.MgmtServer}} placeholder is the dynamic
+// part calculated by the shell script. It will be filled by expectedKubectlLog() for each test case.
+//
+// We'll be mocking kubectl via a function that echos all the args its receives to a file and use this template to
+// compare the expected output vs the actual output.
+const expectedKubectlLogTemplate = `--kubeconfig /configs/management config set clusters.default.server {{.MgmtServer}}
```
muraee, line 39: why do we need to compare all the logs. It should be just enough to grep the expected IP and check it exists

devguyio: The idea of the test is that it verifies what the shell script conceptually is doing. That verification can happen in a lot of way, here are some that I considered:

**Shell script goal:** Build a proper kubeconfig with a context pointing to the  hosted cluster and another for the management cluster

#### Ideas to verify
1. Verify the generated kubeconfig
    * This needs either a complex kubectl mock function or using kubectl itself in a unit test
 2. Grep the expected IP and check it exits
     * I didn't go there because this focuses only on that part of the shell script, so if it was modified by some other ways, this unit test will still succeed and whoever modified it might not be incentivized to add unit tests validating the changes
 3. Verify all the kubectl calls
    * Since all what the shell script does is call kubectl, if we verify all its calls, we basically verify all the logic. 
    * This has a drawback that the shell script can still change by adding none kubectl calls, without the unit tests failing, but given how I couldn't find an simple way to verify ALL what the shell script does (e.g. what if creates some files, or call other commands, etc) , this seemed to me like the sweet spot between verifying all the logic and not overcomplicating things.

muraee: > I didn't go there because this focuses only on that part of the shell script, so if it was modified by some other ways, this unit test will still succeed and whoever modified it might not be incentivized to add unit tests validating the change

well, the rest of script is just static, the only variable is the server. So, I see 2 disadvantages:
1. obviously the test is too complex now, because we are trying to match all the logs
2. we are forcing static changes to be made in 2 places, in the yaml and in the unit test, which is just adding more work .

The unit test fixtures,are there to test that everything renders correctly. I don't think we need to replicate that for the static parts.

devguyio: I see your point, and there's no single approach that doesn't add work. However:
1. Because fixtures are auto-generated, they don't test the "intention" of the shell script. E.g. here's a change that can be skipped by that test:
    * someone adds a 3rd context, and uses the same wrong `[SOME_HOST]` technique for both IPV4 and IPV6, the fixture test won't catch that, the unit test to grep the expected IP won't catch that either. However, the unit test in this PR will fail, the user will "codify" the expected args, thinking kubectl will be passed SOME_HOST but it'll fail too, cause they'll get [SOME_HOST]. 
2. I understand it adds works and complexity, but i don't think it's really not much work to expect if anyone changes the kubectl logic, please change that. It's a pretty standard way of doing tests as you know. 

If this was go, I'd have not have that much strong opinion about it, but because it's shell, I'm of the opinion that this added extra is worth it as a safety guard.

control-plane-operator/controllers/hostedcontrolplane/v2/cno/deployment_init_container_test.go
```diff
@@ -0,0 +1,205 @@
...
+			}
+
+			actualLog, err := os.ReadFile(kubectlLog)
+			if err != nil {
+				t.Fatalf("failed to read kubectl log: %v", err)
+			}
+
+			expectedLog := expectedKubectlLog(tt.expectedServer)
+			if string(actualLog) != expectedLog {
+				diff, _ := difflib.GetUnifiedDiffString(difflib.UnifiedDiff{
```
muraee: can we at least please use `cmp.Diff` as we do across the code base to stay consistent.

control-plane-operator/controllers/hostedcontrolplane/v2/cno/deployment_init_container_test.go
muraee: fair enough 👍

one other comment then :) https://github.com/openshift/hypershift/pull/7447#discussion_r2676600551

control-plane-operator/controllers/hostedcontrolplane/v2/cno/deployment_init_container_test.go
devguyio: I actually implemented it first using `cmp.Diff` but when I saw the failure  output looked like this because it was multiline output:

```
    deployment_init_container_test.go:170: kubectl log mismatch (-want +got):
          []string{
          	strings.Join({
          		"--kubeconfig /configs/management config set clusters.default.ser",
          		"ver https://",
        + 		"[",
          		"kubernetes.default.svc",
        + 		"]",
          		":443",
          	}, ""),
          	"--kubeconfig /configs/management config set clusters.default.cer"...,
          	"--kubeconfig /configs/management config set users.admin.tokenFil"...,
          	... // 12 identical elements
          }
```

working with Claude Code to find a more git diff style output it suggested to use `github.com/pmezard/go-difflib/difflib` which was already an indirect dependency so that seemed reasonable given how the diff looks like 

```
        --- want
        +++ got
        @@ -1,4 +1,4 @@
        ---kubeconfig /configs/management config set clusters.default.server https://10.0.0.1:443
        +--kubeconfig /configs/management config set clusters.default.server https://[10.0.0.1]:443
         --kubeconfig /configs/management config set clusters.default.certificate-authority /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
         --kubeconfig /configs/management config set users.admin.tokenFile /var/run/secrets/kubernetes.io/serviceaccount/token
         --kubeconfig /configs/management config set contexts.default.cluster default
```

But I had no strong opinion. Reverting back to `cmp.Diff`

devguyio: @muraee Done

devguyio:
another needed rebase to get a required tekton bump

celebdor:
Added the tag that was lost due to the konflux task rebase (which doesn't change in any way the code)

sjenning:
minimal is failing because it also needs https://github.com/openshift/cluster-network-operator/pull/2871

sjenning:
e2e's will not pass without https://github.com/openshift/cluster-network-operator/pull/2871 but that PR will not pass without this PR.  Breaking the cycle.

/override ci/prow/e2e-aks
/override ci/prow/e2e-aks-4-21
/override ci/prow/e2e-aws
/override ci/prow/e2e-aws-upgrade-hypershift-operator

PR #7454, opened 2026-01-10
OCPBUGS-72559: fix(sync-global-pullsecret): merge existing auths instead of overwriting

joshbranham:
The sync-global-pullsecret code was previously overwriting the entire /var/lib/kubelet/config.json file, which broke scenarios where external systems had already injected auths that HyperShift was unaware of.

This change updates the code to merge auths from the existing on-disk file with the desired pull secret, preserving any external modifications. Existing auths unknown to Hypershift always take precedence over HyperShift's desired auths, allowing external systems to safely modify the file.

An example:
* The file `/var/lib/kubelet/config.json` contains an auth for `1234567.dkr.ecr.us-east-1.amazonaws.com`
* Hypershift and the additional-pull-secret do not have an auth for `1234567.dkr.ecr.us-east-1.amazonaws.com`
* When generating the config.json, Hypershift will merge in it's values, leaving the ECR auth in place
* If the user adds an auth for `1234567.dkr.ecr.us-east-1.amazonaws.com` to the additional-pull-secret, Hypershift _will_ squash over it

This allows existing workarounds for pull secrets to work, while providing a path for those auths to be "owned" by Hypershift and the cluster when possible.

Changes:
- Added dockerConfigJSON struct and parsing functions
- Implemented mergeDockerConfigs to merge existing and desired auths
- Updated checkAndFixFile to use merge logic instead of overwrite
- Added comprehensive test coverage for merge scenarios
- Logs when external auths are discovered and preserved

Fixes OCPBUGS-72559

## Checklist:

### DIFF
--- a/sync-global-pullsecret/sync-global-pullsecret.go
+++ b/sync-global-pullsecret/sync-global-pullsecret.go
@@ -47,10 +47,6 @@ const (
 	originalPullSecretFilePath = "/etc/original-pull-secret/.dockerconfigjson"
 	globalPullSecretFilePath   = "/etc/global-pull-secret/.dockerconfigjson"
 
-	// Cache file path for tracking last-synced pull secret state
-	// This allows us to distinguish between external auths (preserve) and removed HyperShift auths (delete)
-	pullSecretCachePath = "/var/lib/kubelet/.hypershift-pullsecret-cache.json"
-
 	tickerPace = 30 * time.Second
 
 	// systemd job completion state as documented in go-systemd/dbus
@@ -62,6 +58,11 @@ const (
 )
 
 var (
+	// Cache file path for tracking last-synced pull secret state
+	// This allows us to distinguish between external auths (preserve) and removed HyperShift auths (delete)
+	// Made a variable instead of const to allow test overrides
+	pullSecretCachePath = "/var/lib/kubelet/.hypershift-pullsecret-cache.json"
+
 	// writeFileFunc is a variable that holds the function used to write files.
 	// This allows tests to inject custom write functions for testing rollback scenarios.
 	writeFileFunc = writeAtomic
@@ -69,6 +70,10 @@ var (
 	// readFileFunc is a variable that holds the function used to read files.
 	// This allows tests to inject custom read functions for testing.
 	readFileFunc = os.ReadFile
+
+	// signalKubeletToRestartFunc is a variable that holds the function used to restart kubelet.
+	// This allows tests to inject custom restart functions for testing.
+	signalKubeletToRestartFunc = signalKubeletToRestartProcess
 )
 
 // NewRunCommand creates a new cobra.Command for the sync-global-pullsecret command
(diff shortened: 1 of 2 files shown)


sync-global-pullsecret/sync-global-pullsecret.go
```diff
@@ -189,21 +190,50 @@ func (s *GlobalPullSecretSyncer) checkAndFixFile(pullSecretBytes []byte) error {
...
 		return fmt.Errorf("failed to read existing file: %w", err)
 	}
 
-	// Preserve trailing newline if it exists in the original file
-	contentToWrite := pullSecretBytes
-	if len(existingContent) > 0 && existingContent[len(existingContent)-1] == '\n' {
-		if len(pullSecretBytes) == 0 || pullSecretBytes[len(pullSecretBytes)-1] != '\n' {
-			contentToWrite = append(pullSecretBytes, '\n')
+	// Determine if we need to preserve trailing newline
+	preserveNewline := len(existingContent) > 0 && existingContent[len(existingContent)-1] == '\n'
```
joshbranham, line 257: I am not sure we need this anymore given we are now marshalling to JSON and back to disk, but unsure what motivation led to it being added in the first place (maybe it was Claude all along). Happy to remove if desired.

sync-global-pullsecret/sync-global-pullsecret.go
```diff
@@ -189,21 +190,50 @@ func (s *GlobalPullSecretSyncer) checkAndFixFile(pullSecretBytes []byte) error {
...
+	}
+
+	// Preserve trailing newline if it existed in the original file
+	contentToWrite := mergedBytes
+	if preserveNewline && (len(mergedBytes) == 0 || mergedBytes[len(mergedBytes)-1] != '\n') {
+		contentToWrite = append(mergedBytes, '\n')
+	}
+
+	// If file content is different, write the merged content
 	if string(existingContent) != string(contentToWrite) {
```
joshbranham: This preserves the existing behavior: after merging auths on disk with auths in-cluster, we compare if anything has changed, and proceed to write the contents and restart kubelet. This has the benefit of also recognizing when an external process has updated the auths on disk.

sync-global-pullsecret/sync-global-pullsecret_test.go
```diff
@@ -49,6 +49,32 @@ func TestCheckAndFixFile(t *testing.T) {
 			expectedFinalContent: `{"auths":{"old.registry.com":{"auth":"b2xkOnRlc3Q="}}}`,
 			expectError:          true,
 		},
+		{
```
joshbranham, line 52: TODO: Fixup these tests

sync-global-pullsecret/sync-global-pullsecret_test.go
```diff
@@ -148,6 +174,45 @@ func TestCheckAndFixFile(t *testing.T) {
...
+			name:               "multiple external auths preserved",
+			description:        "file has multiple external auths, all should be preserved in merge",
+			initialContent:     `{"auths":{"external1.com":{"auth":"ZXh0MQ=="},"external2.com":{"auth":"ZXh0Mg=="}}}`,
+			secretContent:      `{"auths":{"hypershift.io":{"auth":"aHlwZXI="}}}`,
+			rollbackShouldFail: false,
+			expectedErrorContains: []string{
+				"failed to restart kubelet after 3 attempts",
+				"rolled back changes",
+			},
+			expectedFinalContent: `{"auths":{"external1.com":{"auth":"ZXh0MQ=="},"external2.com":{"auth":"ZXh0Mg=="}}}`,
```
joshbranham, line 201: TODO: Fix

sdminonne:
@coderabbitai review

sync-global-pullsecret/sync-global-pullsecret.go
```diff
@@ -189,21 +190,50 @@ func (s *GlobalPullSecretSyncer) checkAndFixFile(pullSecretBytes []byte) error {
...
-	if len(existingContent) > 0 && existingContent[len(existingContent)-1] == '\n' {
-		if len(pullSecretBytes) == 0 || pullSecretBytes[len(pullSecretBytes)-1] != '\n' {
-			contentToWrite = append(pullSecretBytes, '\n')
+	// Determine if we need to preserve trailing newline
+	preserveNewline := len(existingContent) > 0 && existingContent[len(existingContent)-1] == '\n'
+
+	// Merge with existing config if it exists
+	var mergedConfig *dockerConfigJSON
+	if len(existingContent) > 0 {
+		existingConfig, err := parseDockerConfigJSON(existingContent)
```
jparrill, line 269: Users configure this PR expecting external auths to always be preserved, but if the file becomes corrupted, they're lost silently. It's ok to not hard fail but I would warn the user about the lost entries

```
if err != nil {
      s.log.Warn("Existing kubelet config corrupted - external auths will be lost",
                 "file", s.kubeletConfigJsonPath,
                 "error", err,
                 "action", "using cluster-provided config only")
      mergedConfig = desiredConfig
  }
```

sync-global-pullsecret/sync-global-pullsecret.go
```diff
@@ -181,6 +181,7 @@ func (s *GlobalPullSecretSyncer) syncPullSecret() error {
 }
 
 // checkAndFixFile reads the current file content and updates it if it differs from the desired content (global pull secret content).
+// This function merges auths from the existing on-disk file with the desired pull secret to preserve any external modifications.
 func (s *GlobalPullSecretSyncer) checkAndFixFile(pullSecretBytes []byte) error {
```
jparrill, line 222: I see a potential race. 

Scenario:
---
- Time T0: sync-global-pullsecret reads file: `{"auths": {"reg1": "auth1"}}`
- Time T1: external operator adds new auth entry: `{"auths": {"reg1": "auth1", "external": "auth2"}}`
- Time T2: sync-global-pullsecret writes merged content, potentially LOSING `"external": "auth2"`

Maybe we can use `flock` to avoid this issue? WDYT?

jparrill (commented):
I would be great to have E2E support on the new behaviour.
This is what I've found for now, I will take another look later.

jparrill:
I forgot to mention a potential problem. The scenario when an entry is added by w/e process and not removed. This will end on a multiple orphaned entries that maybe are not legit. 

I think this scenario could be problematic. It's there any external process that reviews those files/entries?

joshbranham:
> I forgot to mention a potential problem. The scenario when an entry is added by w/e process and not removed. This will end on a multiple orphaned entries that maybe are not legit.
> 
> I think this scenario could be problematic. It's there any external process that reviews those files/entries?

I agree, however my assumption is external processes modifying the auths would be an edge case, and not something we would expect users to use. They should/would instead use this feature as intended, with the `additional-pull-secret`. With that in mind, my thought is that manual auth injection is a "use at your own risk" workaround, and if you leak auths into it, that is on your external system to manage.

sync-global-pullsecret/sync-global-pullsecret.go
joshbranham: Yeah that was one thing I was thinking of but didn't have a solution for. I think a lock could work fine, pulled in `flock` which is used similarly elsewhere in the project. I can back that out if you'd prefer and use `x/sys/unix`, but `gofrs/flock` was "cleaner".

sync-global-pullsecret/sync-global-pullsecret.go
```diff
@@ -180,7 +181,33 @@ func (s *GlobalPullSecretSyncer) syncPullSecret() error {
...
 }
 
+// acquireFileLock acquires an exclusive lock on the file at the given path.
+// It returns a *flock.Flock instance that should be used to release the lock when done.
+func acquireFileLock(path string) (*flock.Flock, error) {
+	fileLock := flock.New(path)
+
+	// Try to acquire exclusive lock with timeout and retry
+	const lockTimeout = 30 * time.Second
+	const retryInterval = 100 * time.Millisecond
```
jparrill: I would move those to a section in the file for these constants

sync-global-pullsecret/sync-global-pullsecret.go
jparrill: sgtm, agree on prefer simplicity in this case.

sync-global-pullsecret/sync-global-pullsecret.go
jparrill: It's ok to keep it 👍

sync-global-pullsecret/sync-global-pullsecret.go
```diff
@@ -279,6 +349,11 @@ func readPullSecretFromFile(filePath string) ([]byte, error) {
 	return content, nil
 }
 
+// dockerConfigJSON represents the structure of a Docker config.json file
+type dockerConfigJSON struct {
+	Auths map[string]interface{} `json:"auths"`
+}
+
```
jparrill, line 433: I missed this on the first review. Could we maybe use 	`github.com/openshift/hypershift/support/thirdparty/kubernetes/pkg/credentialprovider`, more concrete `DockerConfigJSON` struct ? We have that dep already in the code and the validation would be more strict

test/e2e/util/globalps.go
```diff
@@ -265,3 +266,228 @@ func OpenshiftOVNKubeDaemonSet() *appsv1.DaemonSet {
...
+				},
+			},
+		},
+	}
+
+	return guestClient.Create(ctx, daemonSet)
+}
+
+// VerifyMockAuthPreservedAfterSync verifies that the mock registry auth is still present after sync
+func VerifyMockAuthPreservedAfterSync(t *testing.T, ctx context.Context, guestClient crclient.Client, dsImage string) {
```
jparrill: I think this is too much, I would change `VerifyKubeletConfigWithDaemonSet` function which is basically the same as you are doing but adding your requirements.

test/e2e/util/util.go
```diff
@@ -2011,6 +2052,12 @@ func EnsureGlobalPullSecret(t *testing.T, ctx context.Context, mgmtClient crclie
 		g.Expect(err).NotTo(HaveOccurred(), "failed to wait for DaemonSets to be ready")
 	})
 
+	// Verify that mock on-disk auth was preserved after global pull secret sync
+	t.Run("Verify mock on-disk auth preserved after sync", func(t *testing.T) {
```
jparrill: Is this necessary? I would just check at the end on `VerifyKubeletConfigWithDaemonSet`

sync-global-pullsecret/sync-global-pullsecret.go
joshbranham: I don't think so. That type is for Docker, and writes auths with a `Username|Password|Email` field. The `/var/lib/kubelet/config.json` uses a `auth` key with the `username:password` value base64 encoded.

jparrill:
you will need to rebase, due to EC failing from Konflux

joshbranham:
Looks like docs preview is failing for:
> error: failed to create/update secret: failed to update secret: Request entity too large: limit is 3145728

test/e2e/util/globalps.go
```diff
@@ -265,3 +275,93 @@ func OpenshiftOVNKubeDaemonSet() *appsv1.DaemonSet {
...
+					"name": "mock-auth-injector",
+				},
+			},
+			Template: corev1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: map[string]string{
+						"name": "mock-auth-injector",
+					},
+				},
+				Spec: corev1.PodSpec{
```
gaol, line 302: global-pull-secret-syncer ds runs on InPlace nodes only, it would be good for the mock ds keep consistent with it, otherwise, it pollutes other nodes, which overrides the config.json file and no global-pull-secret-syncer ds does the recovery

gaol (commented):
I did a pre-merge test, and I see a problem that the registries added to the `additional-pull-secret` won't get deleted after the `additional-pull-secret` gets deleted.

```
[🎩 lgao@lgao-mac hypershift-qe-utils]$ oc create secret docker-registry additional-pull-secret -n kube-system --docker-server=test2.registry.io --docker-username=myuser --docker-password=mypass
secret/additional-pull-secret created
[🎩 lgao@lgao-mac hypershift-qe-utils]$ oc debug "node/$node" -q -- chroot /host  cat /var/lib/kubelet/config.json |jq -r '.auths["test2.registry.io"]'
{
  "auth": "bXl1c2VyOm15cGFzcw==",
  "password": "mypass",
  "username": "myuser"
}
[🎩 lgao@lgao-mac hypershift-qe-utils]$ oc delete secret additional-pull-secret -n kube-system
secret "additional-pull-secret" deleted

# wait or run the following command multiple times, you still see it exists:

[🎩 lgao@lgao-mac hypershift-qe-utils]$ oc debug "node/$node" -q -- chroot /host  cat /var/lib/kubelet/config.json |jq -r '.auths["test2.registry.io"]'
{
  "auth": "bXl1c2VyOm15cGFzcw==",
  "password": "mypass",
  "username": "myuser"
}
```
The new added items in the additional-pull-secret is expected to be deleted, otherwise, the config.json file will keep growing.

sync-global-pullsecret/sync-global-pullsecret.go
```diff
@@ -189,58 +227,167 @@ func (s *GlobalPullSecretSyncer) checkAndFixFile(pullSecretBytes []byte) error {
...
-	// If file content is different, write the desired content
-	if string(existingContent) != string(contentToWrite) {
-		s.log.Info("file content is different, updating it")
-		// Save original content for potential rollback
-		originalContent := existingContent
+	// Marshal the merged config
+	mergedBytes, err := json.Marshal(mergedConfig)
+	if err != nil {
+		return fmt.Errorf("failed to marshal merged config: %w", err)
+	}
```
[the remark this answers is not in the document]
joshbranham: The existing solution already does not honor these fields, opting to not address that during this PR.

joshbranham:
The merge behavior is proving to be too risky, and QE has validated that. I am closing this in favor of https://github.com/openshift/hypershift/pull/7594

PR #7466, opened 2026-01-13
OCPBUGS-73721: Enable autoscaler from/to zero on Hypershift

LiangquanLi930:
manual cherry-pick https://github.com/openshift/hypershift/pull/6975 and https://github.com/openshift/hypershift/pull/7427


gaol:
/verified by @gaol 
I did some tests using [the on-going scripts](https://github.com/gaol/openshift-release/blob/autoscaling_from_zero/ci-operator/step-registry/cucushift/hypershift-extended/autoscaling-from-zero/cucushift-hypershift-extended-autoscaling-from-zero-commands.sh), and it passed both Replace and InPlace np node upgrade type.

zhfeng:
## Scale-from-Zero Verification Results — PASS ✅

| Step | Result | Duration |
|------|--------|----------|
| ipi-install-rbac | ✅ SUCCESS | 7s |
| create-management-cluster | ✅ SUCCESS | 16m29s |
| hypershift-install | ✅ SUCCESS | 1m32s |
| hypershift-aws-run-e2e-external (TestNodePoolAutoscalingScaleFromZero) | ✅ SUCCESS | 36m9s |
| dump-management-cluster | ✅ SUCCESS | 3m14s |
| hypershift-dump-extra | ✅ SUCCESS | 14s |
| destroy-management-cluster | ✅ SUCCESS | 5m31s |

### Evidence
- **MachineDeployment annotations verified:** `vCPU=2, memoryMb=8192, labels=kubernetes.io/arch=amd64,scale-from-zero-test=true`
- **NodePool started at 0 replicas**, scaled to 1 in ~6min after workload creation
- **Workload pods scheduled and running** on scaled nodes

### Links
- [Rehearsal job](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_release/76959/rehearse-76959-pull-ci-openshift-hypershift-release-4.21-e2e-aws-scale-from-zero/2037456343287205888)
- [CI PR: openshift/release#76959](https://github.com/openshift/release/pull/76959)

/verified by zhfeng

api/hypershift/v1beta1/nodepool_types.go
```diff
@@ -101,7 +101,8 @@ type NodePool struct {
 // +kubebuilder:validation:XValidation:rule="self.arch != 'arm64' || has(self.platform.aws) || has(self.platform.azure) || has(self.platform.agent) || self.platform.type == 'None'", message="Setting Arch to arm64 is only supported for AWS, Azure, Agent and None"
 // +kubebuilder:validation:XValidation:rule="!has(self.replicas) || !has(self.autoScaling)", message="Both replicas or autoScaling should not be set"
 // +kubebuilder:validation:XValidation:rule="self.arch != 's390x' || has(self.platform.kubevirt)", message="s390x is only supported on KubeVirt platform"
-// +kubebuilder:validation:XValidation:rule="(has(self.platform.aws) && has(self.platform.aws.imageType) && self.platform.aws.imageType == 'Windows') ? self.arch == 'amd64' : true", message="ImageType 'Windows' requires arch 'amd64' (AWS only)"
+// +kubebuilder:validation:XValidation:rule="!has(self.platform.aws) || !has(self.platform.aws.imageType) || self.platform.aws.imageType != 'Windows' || self.arch == 'amd64'", message="ImageType 'Windows' requires arch 'amd64' (AWS only)"
```
JoelSpeed, line 104: Why did you change this rule?

api/hypershift/v1beta1/nodepool_types.go
```diff
@@ -446,11 +447,13 @@ type NodePoolManagement struct {
...
-	// min is the minimum number of nodes to maintain in the pool. Must be >= 1 and <= .Max.
+	// min is the minimum number of nodes to maintain in the pool.
+	// Can be set to 0 for scale-from-zero for AWS platform.
+	// Must be >= 0 and <= .Max.
 	//
-	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Minimum=0
 	// +required
-	Min int32 `json:"min"`
+	Min *int32 `json:"min"`
```
JoelSpeed, line 456: Hmm, so this is/could be a breaking change. If you want to keep this as a pointer, it **needs** `omitempty` too.

For any client that already is coded against this object and the go types, when they update their client library, they would need to update now to handle the fact that this is a pointer. It would be fairly obvious as their code wouldn't compile. But in places where they are reading the value, this may not be so obvious. The benefit is that this field is required, so there _will_ always be a non-nil value for a reading client.

Alternatively, don't change this to a pointer. Because there is no `omitempty`, this not being a pointer means that `0` can round trip through a structured client. The awkwardness comes from the fact that for a structured client this effectively makes this field optional. If the user doesn't set an opinion explicitly, the `0` will serialize and they will (potentially) accidentally enable scale from zero when they didn't intend to.

PR #7471, opened 2026-01-13
CNTRLPLANE-2049: Review and address kube-api-linter exclusions

Nirshal:

This PR progressively reviews and addresses kube-api-linter exclusions in api/.golangci.yml. The linter configuration currently contains numerous exclusions that were added temporarily to allow the linter to pass. This work categorizes each exclusion as either fixable (non-API-breaking) or requiring permanent exception with justification.

This initial draft processes the following exclusion categories:
- `arrayofstruct`: Removed 4 exclusions by adding +listType markers to array fields
- `defaultorrequired`: Removed 8 exclusions by fixing invalid +default markers and adding +optional where needed
- `statusoptional`: Removed 2 exclusions by adding +optional markers to status subresource fields
- `ssatags`: Removed 40 exclusions by adding +listType markers with detailed rationale for atomic vs map choices

Additional exclusion categories will be processed in follow-up commits to this same PR while in draft mode.

**Benefits**:
- Cleaner, more maintainable linter configuration
- API types better aligned with Kubernetes API conventions
- Improved Server-Side Apply behavior through explicit listType markers
- No backward compatibility breaks

Ref https://issues.redhat.com/browse/CNTRLPLANE-2049

- This PR is opened in draft mode for incremental work on remaining exclusion categories
- Each category has been processed in a separate commit with detailed commit messages
- Conservative choices (atomic over map) were made when backward compatibility was uncertain
- All changes follow Kubernetes SSA best practices and maintain API compatibility

## Checklist:

### DIFF
--- a/api/hypershift/v1beta1/azure.go
+++ b/api/hypershift/v1beta1/azure.go
@@ -241,7 +241,6 @@ const (
 type Diagnostics struct {
 	// storageAccountType determines if the storage account for storing the diagnostics data
 	// should be disabled (Disabled), provisioned by Azure (Managed) or by the user (UserManaged).
-	// +kubebuilder:validation:Enum=Managed;UserManaged;Disabled
 	// +kubebuilder:default:=Disabled
 	// +unionDiscriminator
 	// +optional
@@ -576,9 +575,6 @@ type ManagedIdentity struct {
 	// The default value is utf-8.
 	//
 	// See this for more info - https://github.com/Azure/secrets-store-csi-driver-provider-azure/blob/master/website/content/en/getting-started/usage/_index.md
-	//
-	// +kubebuilder:validation:Enum:=utf-8;hex;base64
-	// +kubebuilder:default:="utf-8"
 	ObjectEncoding ObjectEncodingFormat `json:"objectEncoding"`
 
 	// credentialsSecretName is the name of an Azure Key Vault secret. This field assumes the secret contains the JSON
--- a/api/hypershift/v1beta1/hostedcluster_types.go
+++ b/api/hypershift/v1beta1/hostedcluster_types.go
@@ -1352,7 +1352,6 @@ type AutoNode struct {
 type ProvisionerConfig struct {
 	// name specifies the name of the provisioner to use.
 	// +required
-	// +kubebuilder:validation:Enum=Karpenter
 	Name Provisioner `json:"name"`
 	// karpenter specifies the configuration for the Karpenter provisioner.
 	// +optional
--- a/api/.golangci.yml
+++ b/api/.golangci.yml
@@ -29,6 +29,7 @@ linters:
               - defaultorrequired
               - statusoptional
               - ssatags
+              - duplicatemarkers
           lintersConfig:
             conditions:
               isFirstField: Warn
@@ -152,24 +153,6 @@ linters:
           - kubeapilinter
         path: hypershift/v1beta1/nodepool_types.go
         text: 'conditions: Conditions field in NodePoolStatus must be a slice of metav1.Condition'
-
-      # duplicatemarkers (4 issues)
-      - linters:
-          - kubeapilinter
-        path: hypershift/v1beta1/azure.go
-        text: 'duplicatemarkers: Diagnostics.StorageAccountType has duplicated markers kubebuilder:validation:Enum'
-      - linters:
-          - kubeapilinter
-        path: hypershift/v1beta1/azure.go
-        text: 'duplicatemarkers: ManagedIdentity.ObjectEncoding has duplicated markers kubebuilder:default'
-      - linters:
-          - kubeapilinter
-        path: hypershift/v1beta1/azure.go
-        text: 'duplicatemarkers: ManagedIdentity.ObjectEncoding has duplicated markers kubebuilder:validation:Enum'
-      - linters:
-          - kubeapilinter
-        path: hypershift/v1beta1/hostedcluster_types.go
-        text: 'duplicatemarkers: ProvisionerConfig.Name has duplicated markers kubebuilder:validation:Enum'
 
       # integers (6 issues)
       - linters:


api/hypershift/v1beta1/aws.go
bryan-cox, line 1: How did you know what listType to pick for fields?

api/hypershift/v1beta1/azure.go
```diff
@@ -580,7 +579,6 @@ type ManagedIdentity struct {
 	//
 	// +kubebuilder:validation:Enum:=utf-8;hex;base64
 	// +kubebuilder:default:="utf-8"
-	// +required
```
bryan-cox, line 583: This one should be required

api/hypershift/v1beta1/azure.go
bryan-cox, line 1: When removing the default values, did you ensure we set a default value in the CLI and/or a controller?

api/hypershift/v1beta1/hosted_controlplane.go
```diff
@@ -315,17 +315,15 @@ type HostedControlPlaneStatus struct {
 	// ready denotes that the HostedControlPlane API Server is ready to
 	// receive requests
 	// This satisfies CAPI contract https://github.com/kubernetes-sigs/cluster-api/blob/cd3a694deac89d5ebeb888307deaa61487207aa0/controllers/cluster_controller_phases.go#L226-L230
-	// +required
-	// +kubebuilder:default=false
```
bryan-cox, line 319: I would keep the default here

api/hypershift/v1beta1/hosted_controlplane.go
```diff
@@ -315,17 +315,15 @@ type HostedControlPlaneStatus struct {
...
+	// +optional
 	Ready bool `json:"ready"`
 
 	// initialized denotes whether or not the control plane has
 	// provided a kubeadm-config.
 	// Once this condition is marked true, its value is never changed. See the Ready condition for an indication of
 	// the current readiness of the cluster's control plane.
 	// This satisfies CAPI contract https://github.com/kubernetes-sigs/cluster-api/blob/cd3a694deac89d5ebeb888307deaa61487207aa0/controllers/cluster_controller_phases.go#L238-L252
-	// +required
-	// +kubebuilder:default=false
```
bryan-cox, line 328: I would keep the default here

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -610,7 +610,7 @@ type HostedClusterSpec struct {
 
 	// networking specifies network configuration for the hosted cluster.
 	// Defaults to OVNKubernetes with a cluster network of cidr: "10.132.0.0/14" and a service network of cidr: "172.31.0.0/16".
-	// +required
+	// +optional
```
bryan-cox, line 613: I don't think we'd put the +optional here and below

api/hypershift/v1beta1/kubevirt.go
bryan-cox, line 1: I don't think we'd put the required field here

api/hypershift/v1beta1/aws.go
Nirshal: - For securityGroup it was a bit difficult because map can't be used because the two field of the underlying object are mutually-exclusive so there is no key. Set can't be used because the use of set for complex objects is deprecate. Atomic seemed a conservative option because reading around and asking both claude and gemini when no marker is specified "atomic" is the implicit behavior. I was a bit confused because I found the method "WithSecurityGroups(...)" in the client/applyconfiguration/hypershift/v1beta1/awsnodepoolplatform.go:87 that implied that new items were appended to the slice instead. 
- The ones I put map on seemed at the beginning quite easy, because as an example for ResourceTags you don't want tags on a resource to be duplicated and the undelying object structure already suggests the key to use. Nonetheless when I was about to finish I realized that if in your yaml for some particular reason you have some duplicates, then this change is going to make your request be rejected, so it can be seen as a breaking change, for this reason I had a personal note to ask you your opinion about it.
- Filter.Values I was a bit unsure on putting "Set", because even if I suppose the filter in theory should have unique values inside, I could not determine if in the past the field was used as atomic (and in theory it should, because it is default behavior)... I had a note to ask this to you. (The same applies to AWSPlatformSpec.AdditionalAllowedPrincipals)

api/hypershift/v1beta1/azure.go
Nirshal: Almost all the default values have been removed because of "duplicatemarkers" linter where the same marker is specified both on the field and the undelying type. Default behavior is that when this happens the field mark is ignored and the one on the type should be preserved (this is also the suggested resolution in the linter docs). In short: they should Inherit the same default from the underlying field type, for instance in the case of the Diagnostic.StorageAccountType we have:
`// AzureDiagnosticsStorageAccountType specifies the type of storage account for storing Azure VM diagnostics data.
// +kubebuilder:validation:Enum=Managed;UserManaged;Disabled
type AzureDiagnosticsStorageAccountType string`
so the default is already inherited.
There is another specific case where I removed the default, but I will comment in the specific point you asked :)

api/hypershift/v1beta1/azure.go
Nirshal: This field is a tricky one, because there were several linters complaining about it. The final result seems a bit drastic so let me go step by step:
- when I was processing "defaultorrequired" linter, I removed the "required" because it was in contradiction with the default value (this is what the linter was complaining about), and it seemed more conservative to keep the default than the required, because for what I have read, the defaulting webhook always run before the validating webhook, so today even if you leave the field empty the validation never notices that, because the defaulting populates the field with the utf-8 value.
- this was also subject to a double "duplicatemarkers" and I applied the same rationale as per comment above, where I kept the markers on the type, and removed them from the field. As a reference, the type is in the same file at row 480:
`// +kubebuilder:validation:Enum:=utf-8;hex;base64
// +kubebuilder:default:="utf-8"
type ObjectEncodingFormat string`

api/hypershift/v1beta1/hosted_controlplane.go
Nirshal: I was indeed unsure about this, if you check our slack DMs I asked you about it but in the "wall of text" I usually write (sorry for that) it went unnoticed  😆 
The rationale behind this is that bool's zero-value for golang is already "false", so you don't need to initialize a bools like for an address type if the value you want is false. I wanted to ask because I'm not yet very clear if some deserialization issues may rise in golang.

api/hypershift/v1beta1/hosted_controlplane.go
Nirshal: same as comment above

api/hypershift/v1beta1/hostedcluster_types.go
Nirshal: same consideration about conflicting default and required (I will fix it accordingly when we discuss the azure's ManagedIdentity.ObjectEncoding)

api/hypershift/v1beta1/kubevirt.go
Nirshal: Can't see what row this comment is about :-/

Nirshal:
@bryan-cox after our internal discussions with @enxebre I think it's better if I close this in favor of a new PR with revised approach - only fixing GCP/OCI violations, keeping exclusions for GA API.

PR #7445, opened 2026-01-08
API-1835: Scaffold OpenShiftManager controller

p0lyn0mial:
Scaffolds a Reconciler which will be responsible for integration with OpenShiftManager.
This controller is experimental and will be **disabled** by default.

For more information on OpenShiftManager, see
https://github.com/openshift/enhancements/blob/master/dev-guide/multi-operator-manager.md.

PR #7477, opened 2026-01-14
API-1835: Scaffold dynamic informers

p0lyn0mial:
Adds an input resource initializer runnable for OpenShiftManager that discovers
input resources and wires informers, and registers it in the controller setup.

Notes: discovery is still a stub and we still need to register informer handlers and add filtering.

requires: https://github.com/openshift/hypershift/pull/7445


control-plane-operator/controllers/openshiftmanager/input_resource_initializer.go
```diff
@@ -0,0 +1,118 @@
...
+		if ctx.Err() != nil {
+			return ctx.Err()
+		}
+		return fmt.Errorf("caches did not sync")
+	}
+	return nil
+}
+
+func (r *inputResourceInitializer) discoverInputResources() (map[string]*libraryinputresources.InputResources, error) {
+	return nil, fmt.Errorf("not implemented")
```
p0lyn0mial: @benluddy this is where we could use the file sytem queue client to discover the input-res per operator.

control-plane-operator/controllers/openshiftmanager/input_resource_initializer.go
```diff
@@ -0,0 +1,118 @@
...
+		registeredGVK := sets.NewString()
+		for _, exactResource := range resources.ApplyConfigurationResources.ExactResources {
+			gvr := schema.GroupVersionResource{Group: exactResource.Group, Version: exactResource.Version, Resource: exactResource.Resource}
+			gvk, err := r.managementClusterRESTMapper.KindFor(gvr)
+			if err != nil {
+				return fmt.Errorf("unable to find Kind for %#v, for %s operator, err: %w", exactResource, operator, err)
+			}
+
+			if registeredGVK.Has(gvk.String()) {
+				continue
```
p0lyn0mial, line 66: so that we don't register multiple event handlers for the same gvk

control-plane-operator/controllers/openshiftmanager/input_resource_initializer.go
```diff
@@ -0,0 +1,118 @@
...
+		return err
+	}
+
+	for operator, resources := range inputResources {
+		// note that for the POC we are only interested in ApplyConfigurationResources.ExactResources
+		// the checkSupportedInputResources ensures no other resources were provided.
+		//
+		// TODO: in the future we need to extend to full list
+		registeredGVK := sets.NewString()
+		for _, exactResource := range resources.ApplyConfigurationResources.ExactResources {
```
p0lyn0mial, line 58: for the poc we only support `ApplyConfigurationResources.ExactResources` in the future we need to cover the entire list

control-plane-operator/controllers/openshiftmanager/input_resource_initializer.go
```diff
@@ -0,0 +1,118 @@
...
+			registeredGVK.Insert(gvk.String())
+		}
+	}
+
+	if !r.managementClusterCache.WaitForCacheSync(ctx) {
+		if ctx.Err() != nil {
+			return ctx.Err()
+		}
+		return fmt.Errorf("caches did not sync")
+	}
```
p0lyn0mial, line 84: i forgot to add a TODO but at the end we need to send a signal to the main controller the caches are ready.

control-plane-operator/controllers/openshiftmanager/input_resource_initializer.go
```diff
@@ -0,0 +1,118 @@
...
+
+			if registeredGVK.Has(gvk.String()) {
+				continue
+			}
+
+			_, err = r.managementClusterCache.GetInformerForKind(ctx, gvk, cache.BlockUntilSynced(true))
+			if err != nil {
+				return err
+			}
+			// TODO: register informer event handlers
```
p0lyn0mial, line 73: events will be passed to a different component that will be responsible for mapping resources to an operator and trigger the main controller. something like "if X resource(s) change then call/rigger this operator"

control-plane-operator/controllers/openshiftmanager/input_resource_initializer.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func newInputResourceInitializer(mgmtClusterRESTMapper meta.RESTMapper, mgmtClusterCache cache.Cache) *inputResourceInitializer {
+	return &inputResourceInitializer{
+		managementClusterRESTMapper: mgmtClusterRESTMapper,
+		managementClusterCache:      mgmtClusterCache,
+	}
+}
+
+func (r *inputResourceInitializer) Start(ctx context.Context) error {
```
p0lyn0mial, line 36: checking what unit tests could be added.

control-plane-operator/controllers/openshiftmanager/input_resource_initializer.go
```diff
@@ -0,0 +1,118 @@
+package openshiftmanager
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/openshift/multi-operator-manager/pkg/library/libraryinputresources"
```
p0lyn0mial, line 7: we will also pull a new dep

csrwng:
/approve

Looks like you need to fix your commit messages.

control-plane-operator/controllers/openshiftmanager/input_resource_initializer.go
benluddy: Do modules that depend on multi-operator-manager need to keep their Kube dependencies in sync with multi-operator-manager's Kube dependencies? I don't know if we will (in the future) want/need to either (a) avoid a specific dependency in multi-operator-manager or (b) split multi-operator-manager into separate modules for e.g. parent, child, API.

control-plane-operator/controllers/openshiftmanager/input_resource_initializer.go
```diff
@@ -0,0 +1,125 @@
...
+			if err != nil {
+				return err
+			}
+			// TODO: register informer event handlers
+
+			registeredGVK.Insert(gvk.String())
+		}
+	}
+
+	if !r.managementClusterCache.WaitForCacheSync(ctx) {
```
benluddy, line 79: Eventually, I think we will need to handle cases where there's a mix of synced and unsynced informer caches. If all of the caches that a certain child controller needs have synced, the parent can still trigger that controller without also triggering another controller for which all informer caches have _not_ synced.

control-plane-operator/controllers/openshiftmanager/input_resource_initializer.go
```diff
@@ -0,0 +1,125 @@
...
+		}
+		if len(resList.GeneratedNameResources) > 0 {
+			return fmt.Errorf("%v.GeneratedNameResources are unsupported for now", fieldPath)
+		}
+		if len(resList.LabelSelectedResources) > 0 {
+			return fmt.Errorf("%v.LabelSelectedResources are unsupported for now", fieldPath)
+		}
+		if len(resList.ResourceReferences) > 0 {
+			return fmt.Errorf("%v.ResourceReferences are unsupported for now", fieldPath)
+		}
```
benluddy, line 108: Not important here, just an idea:

```go
if !semantic.DeepEqual(resList, ResourceList{ExactResources: resList.ExactResources) {
```

Also, should we be treating types like these as conventional versioned Kubernetes-style APIs?

control-plane-operator/controllers/openshiftmanager/input_resource_initializer.go
p0lyn0mial: yeah, I think the modules need to keep their Kube deps in sync with the multi-operator-manager’s Kube deps. 

this is a similar situation to what we already have today with library-go, which is used by HyperShift and also depends on a specific Kube version.

moreover, multi-operator-manager also uses library-go. 

because of these intertwined dependencies, I think that we will be forced to keep all of the deps at the same level.

control-plane-operator/controllers/openshiftmanager/input_resource_initializer.go
p0lyn0mial: I think we could have something like that per operator. 

Right now, `startAndWaitForInformersFor` is for a single operator, but in the future we would need a new method that takes the inputs for multiple operators into account.

control-plane-operator/controllers/openshiftmanager/input_resource_initializer.go
p0lyn0mial: > Also, should we be treating types like these as conventional versioned Kubernetes-style APIs?

This is an interesting question. In theory, the types exposed by a library are also APIs. 
I think versioning would make sense if we wanted to offer some level of backward compatibility. 

I’d say let’s start simple and evaluate the requirements in the future.

p0lyn0mial: decided to use `semantic.DeepEqual` - it can help detect unknown (new) fields in the future.
decided to keep the specific checks so that the error msg is clear without using the reflection.

control-plane-operator/controllers/openshiftmanager/input_resource_initializer.go
```diff
@@ -0,0 +1,130 @@
...
+				return fmt.Errorf("unable to find Kind for %#v, for %s operator, err: %w", exactResource, operator, err)
+			}
+
+			if registeredGVK.Has(gvk.String()) {
+				continue
+			}
+
+			_, err = r.managementClusterCache.GetInformerForKind(ctx, gvk, cache.BlockUntilSynced(true))
+			if err != nil {
+				return err
```
bertinatto: nit: maybe add some context for this error?

control-plane-operator/controllers/openshiftmanager/input_resource_initializer.go
bertinatto: We could register multiple even handlers once we have more than one operator, right? Would make sense to move the declaration of `registeredGVK` out of the loop?

p0lyn0mial: I think that gvk can only be registered once. For operators requiring the same gvk we will register multiple filters.

control-plane-operator/controllers/openshiftmanager/input_resource_initializer.go
p0lyn0mial: I can do that in a follow-up.

p0lyn0mial: Had to rebase (Konflux) the PR and decided to add more context.

PR #7473, opened 2026-01-13
AUTOSCALE-494: add OpenShift release image version field

maxcao13:
## What this PR does / why we need it:

This change introduces the ability to specify an OpenShift release image version per OpenshiftEC2NodeClass, allowing users to launch nodes with different OpenShift versions independent of the control plane release version.

API Changes:
- Add `openShiftReleaseImageVersion` field to OpenshiftEC2NodeClassSpec, enabling users to specify which OpenShift release image nodes should use

Token Generation Migration:
- Move token/userData secret generation from hypershift-operator to karpenter-operator
- The ec2nodeclass controller now generates tokens for each OpenshiftEC2NodeClass, allowing each NodeClass to have its own release version
- Add config version tracking via annotation on OpenshiftEC2NodeClass to enable proper cleanup of outdated token secrets

RBAC Changes:
- Add new HostedCluster-scoped Role and RoleBinding for karpenter-operator to access HostedCluster resources needed for token generation
- Grant karpenter-operator permissions to manage secrets in the HCP namespace

Label/Annotation Changes:
- Introduce `hypershift.openshift.io/managed-for-karpenter` label to identify secrets managed by the karpenter-operator
- Use `hypershift.openshift.io/currentConfigVersion` annotation on OpenshiftEC2NodeClass to track config versions across reconciles
- Remove unused NodePoolLabel from karpenter-managed secrets

Code Organization:
- Extract token reconciliation into helper functions for better testability: hcpNodePool(), buildConfigGenerator(), reconcileToken(), updateConfigVersionAnnotation()
- Migrate and update tests from hypershift-operator to karpenter-operator
- Remove obsolete karpenter_test.go from hypershift-operator

This enables users to run nodes with different OpenShift versions within the same cluster, supporting use cases like canary deployments and gradual version rollouts.

Commit-Message-Assisted-by: Cursor Agent (via Cursor IDE)
## Which issue(s) this PR fixes:
Fixes [AUTOSCALE-492 ](https://issues.redhat.com/browse/AUTOSCALE-492) and [AUTOSCALE-494](https://issues.redhat.com/browse/AUTOSCALE-494)

## Checklist:
- [x] Subject and description added to both, commit and PR.
- [x] Relevant issues have been referenced.
- [ ] This change includes docs. 
- [x] This change includes unit tests.

PR #7478, opened 2026-01-14
AUTOSCALE-492: migrate token generation to karpenter-operator

maxcao13:
This change moves the token/userData secret generation from hypershift-operator
to karpenter-operator, enabling the ec2nodeclass controller to manage node
bootstrap tokens directly.

Token Generation Migration:
- Move token/userData secret generation from hypershift-operator to karpenter-operator
- The ec2nodeclass controller now generates tokens for each OpenshiftEC2NodeClass
- Add config version tracking via annotation on OpenshiftEC2NodeClass to enable
  proper cleanup of outdated token secrets

Architecture Changes:
- Refactor karpenter-operator to use HostedControlPlane instead of fetching
  HostedCluster directly
- Introduce hostedClusterFromHCP() utility to create a synthetic in-memory
  HostedCluster from HCP for reusing existing ConfigGenerator code paths
- Pass ignition endpoint via command-line flag instead of fetching from
  HostedCluster status
- Add SyntheticHostedClusterCPNSAnnotation to support/karpenter for control
  plane namespace override in synthetic HostedCluster objects

Label/Annotation Changes:
- Introduce `hypershift.openshift.io/managed-by-karpenter` label to identify
  secrets managed by the karpenter-operator
- Use `hypershift.openshift.io/currentConfigVersion` annotation on
  OpenshiftEC2NodeClass to track config versions across reconciles
- Add `hypershift.openshift.io/synthetic-hc-cpns` internal annotation for
  karpenter-operator's synthetic HostedCluster objects

Code Organization:
- Extract token reconciliation into helper functions for better testability:
  hcpNodePool(), buildConfigGenerator(), reconcileToken(), updateConfigVersionAnnotation()
- Migrate and update tests from hypershift-operator to karpenter-operator

Commit-Message-Assisted-by: Cursor Agent (via Cursor IDE)

We need to be able to decouple the hosted control plane openshift version from the guest cluster data plane openshift version. Tying each OpenshiftEC2NodeClass to their own token, similar to the hypershift NodePool, and moving this token generation logic to the karpenter-operator seems like the best way to do this, to keep separation of concerns, and allow us to allow the user to specify the openshift version for their karpenter nodes separate from the control plane, and separate for each nodeclass.

The existing control plane upgrade and drift e2e test for karpenter should validate this.

A followup PR will come after this to introduce the UX to allow users to actually do this. For now, the intention of this PR is to preserve the previous behaviour (NodeClass versions still follow the control plane), but migrate the token generation logic to keep things separate and allow easier review. That PR will be here: https://github.com/openshift/hypershift/pull/7473 and be reworked and rebased once this merges and gets ok'd.

Fixes [AUTOSCALE-492](https://issues.redhat.com/browse/AUTOSCALE-492)

## Checklist:


maxcao13:
/test e2e-aws-autonode
/test e2e-aws-techpreview

These tests should both pass, if I did this right.

karpenter-operator/controllers/nodeclass/ec2_nodeclass_controller.go
```diff
@@ -407,17 +439,198 @@ func (r *EC2NodeClassReconciler) getHCP(ctx context.Context) (*hyperv1.HostedCon
...
+
+// buildConfigGenerator creates a ConfigGenerator for the NodePool by looking up the release image,
+// generating HAProxy configuration, and combining all node configuration sources.
+func (r *EC2NodeClassReconciler) buildConfigGenerator(ctx context.Context, hcp *hyperv1.HostedControlPlane, np *hyperv1.NodePool, openshiftEC2NodeClass *hyperkarpenterv1.OpenshiftEC2NodeClass) (*nodepool.ConfigGenerator, error) {
+	// Convert HCP to a barebones HostedCluster for use with existing HostedCluster-based code paths.
+	hostedCluster := hostedClusterFromHCP(hcp)
+	hostedCluster.Status.IgnitionEndpoint = r.IgnitionEndpoint
+
+	pullSecret := corev1.Secret{}
+	if err := r.managementClient.Get(ctx, types.NamespacedName{Namespace: hcp.Namespace, Name: hcp.Spec.PullSecret.Name}, &pullSecret); err != nil {
```
muraee: pull-secret name in the HCP namespace is static 
see https://github.com/openshift/hypershift/blob/main/control-plane-operator/controllers/hostedcontrolplane/common/manifests.go#L10

maxcao13:
test is failing because the upcoming drifted node immediately get's disrupted due to diskpressure. The extra latency of the time between upgrade -> drift is causing this flake to actually start failing the test now...

We may have to hardcode a default blockdevicemapping for all our openshfitec2nodeclasses, since [Custom AMIFamily doens't have any defaults.](https://karpenter.sh/docs/concepts/nodeclasses/#custom-1)

maxcao13:
Ignore the verify and konflux failures, it was just bad import order. I will push a new change to fix this in the next round of reviews.

Latest change is to just add a new default for block storage in the ec2nodeclass since the aws-Karpenter default for Custom AMIFamily is like 20GB for root volume storage, which isn't enough and causes diskpressure. I can farm that out to a new PR if we really want to, but I just included it here to hopefully unflake the test.

/test e2e-aws-autonode

karpenter-operator/controllers/nodeclass/ec2_nodeclass_controller.go
```diff
@@ -407,17 +439,198 @@ func (r *EC2NodeClassReconciler) getHCP(ctx context.Context) (*hyperv1.HostedCon
...
+			Arch: hyperv1.ArchitectureAMD64, // used to find default AMI
+		},
+	}
+}
+
+// hostedClusterFromHCP creates a barebones in-memory HostedCluster from a HostedControlPlane.
+// This allows reusing HostedCluster-based code paths to build a ConfigGenerator without needing the actual HostedCluster object.
+// NOTE: The Namespace field is set to hcp.Namespace (the HCP namespace) rather than the original HC namespace.
+// This ensures secret lookups work correctly since the operator is only allowed to read secrets in the HCP namespace.
+// The SyntheticHostedClusterCPNSAnnotation is set to hcp.Namespace to bypass the HostedControlPlaneNamespace()
```
enxebre: why do we need the annotation at all if the same info is already stored in the .namespace?

Also can we avoid this completely? e.g by passing the namespace in the signature of GenerateHAProxyRawConfig? or even consider the implications of letting the original funcs to consume the cp namespace as well

karpenter-operator/controllers/nodeclass/ec2_nodeclass_controller.go
```diff
@@ -407,17 +439,198 @@ func (r *EC2NodeClassReconciler) getHCP(ctx context.Context) (*hyperv1.HostedCon
...
+		HypershiftOperatorImage: r.ControlPlaneOperatorImage,
+		ReleaseProvider:         r.ReleaseProvider,
+		ImageMetadataProvider:   r.ImageMetadataProvider,
+	}
+	haproxyRawConfig, err := haproxyClient.GenerateHAProxyRawConfig(ctx, hostedCluster)
+	if err != nil {
+		return nil, fmt.Errorf("failed to generate HAProxy config: %w", err)
+	}
+
+	cg, err := nodepool.NewConfigGenerator(ctx, r.managementClient, hostedCluster, np, releaseImage, haproxyRawConfig)
```
enxebre: I would expect we move the karpenter operator within its own repo and install in hcp via payload. This coupling makes it a bit more convoluted.

enxebre:
@maxcao13 let's please move the PR to "draft" until we are good with the changes and we see e2e-aws-techpreview going green, so we avoid running all infra resources for every test

karpenter-operator/controllers/nodeclass/ec2_nodeclass_controller.go
maxcao13: Yeah, I can just change the signature of the function to take in control-plane-namespace to avoid all this, thanks for the feedback :+1:

karpenter-operator/controllers/nodeclass/ec2_nodeclass_controller.go
maxcao13: Agreed here.

~~The problem is that we need to get some ignition config from somewhere, in order to feed it to the nodeclass, as I'm sure you're aware. In HCP, we can get it from the ignition server. In standalone, I don't think such a thing exists, but I could be wrong.~~

~~This definitely requires a lot more architecture and discussion on how to grab ignition configurations without simply utilizing the CAPI node bootstrap/upgrade flow in order to make the operator work on standalone ocp, so I hear you and I 100% agree, but I don't want this to be a blocker in our tight timelines.~~

~~EDIT: The only thing I can think of right now, is if we create real hyperv1.NodePool objects in the management cluster, and have the nodepool controller watch the guest cluster to maintain karpenter-only NodePool objects so that the extracted Karpenter-Operator does not have to deal with generating it's own tokens and we would have to find away for the operator to obtain the tokens to query the ignition server in HCP. And to generalize, we could use this interface for any server that serves ignition outside of HCP.~~

~~To me this seems way broader than we can do design and do right now.~~

See slack for my updated opinion.

maxcao13:
superseded by https://github.com/openshift/hypershift/pull/7582

PR #7482, opened 2026-01-15
CNTRLPLANE-2262: enable scale-from-zero for Azure platform

LiangquanLi930:
Add Azure to the list of platforms supporting scale-from-zero (autoScaling.min=0) in NodePool validation with instance type provider implementation.

Fixes 

## Checklist:

### DIFF
--- a/cmd/install/install.go
+++ b/cmd/install/install.go
@@ -197,7 +197,7 @@ func (o *Options) Validate() error {
 	}
 
 	// Validate scale-from-zero credentials
-	supportedProviders := set.New("aws")
+	supportedProviders := set.New("aws", "azure")
 	if len(o.ScaleFromZeroCreds) != 0 || len(o.ScaleFromZeroCredentialsSecret) != 0 {
 		// Check mutual exclusivity - only one of file or secret should be provided
 		if len(o.ScaleFromZeroCreds) != 0 && len(o.ScaleFromZeroCredentialsSecret) != 0 {
@@ -330,7 +330,7 @@ func NewCommand() *cobra.Command {
 	cmd.PersistentFlags().StringSliceVar(&opts.PlatformsToInstall, "limit-crd-install", opts.PlatformsToInstall, "Used to limit the CRDs that are installed to a per platform basis (example: --limit-crd-install=AWS,Azure). If this flag is not specified, all CRDs for all platforms will be installed. Valid, case-insensitive values are: AWS, Azure, IBMCloud, KubeVirt, Agent, OpenStack, GCP.")
 	cmd.PersistentFlags().StringToStringVar(&opts.AdditionalOperatorEnvVars, "additional-operator-env-vars", opts.AdditionalOperatorEnvVars, "Set of additional environment variables to be set on the HyperShift Operator deployment.")
 	cmd.PersistentFlags().BoolVar(&opts.EnableAuditLogPersistence, "enable-audit-log-persistence", opts.EnableAuditLogPersistence, "If true, enables persistent audit logs with automatic snapshots for kube-apiserver pods")
-	cmd.PersistentFlags().StringVar(&opts.ScaleFromZeroProvider, "scale-from-zero-provider", opts.ScaleFromZeroProvider, "Platform type for scale-from-zero autoscaling (aws)")
+	cmd.PersistentFlags().StringVar(&opts.ScaleFromZeroProvider, "scale-from-zero-provider", opts.ScaleFromZeroProvider, "Platform type for scale-from-zero autoscaling (aws, azure)")
 	cmd.PersistentFlags().StringVar(&opts.ScaleFromZeroCreds, "scale-from-zero-creds", opts.ScaleFromZeroCreds, "Path to credentials file for scale-from-zero instance type queries")
 	cmd.PersistentFlags().StringVar(&opts.ScaleFromZeroCredentialsSecret, "scale-from-zero-secret", opts.ScaleFromZeroCredentialsSecret, "Name of existing secret containing scale-from-zero credentials (alternative to --scale-from-zero-creds)")
 	cmd.PersistentFlags().StringVar(&opts.ScaleFromZeroCredentialsSecretKey, "scale-from-zero-secret-key", opts.ScaleFromZeroCredentialsSecretKey, "Key within the scale-from-zero credentials secret (default: credentials)")
--- a/cmd/install/install_test.go
+++ b/cmd/install/install_test.go
@@ -191,6 +191,15 @@ func TestOptions_Validate(t *testing.T) {
 			},
 			expectError: false,
 		},
+		"when scale-from-zero provider is azure with secret reference there is no error": {
+			inputOptions: Options{
+				PrivatePlatform:                   string(hyperv1.NonePlatform),
+				ScaleFromZeroProvider:             "azure",
+				ScaleFromZeroCredentialsSecret:    "my-azure-secret",
+				ScaleFromZeroCredentialsSecretKey: "credentials",
+			},
+			expectError: false,
+		},
 	}
 	for name, test := range tests {
 		t.Run(name, func(t *testing.T) {
--- a/hypershift-operator/controllers/nodepool/capi.go
+++ b/hypershift-operator/controllers/nodepool/capi.go
@@ -753,12 +753,12 @@ func setMachineDeploymentReplicas(nodePool *hyperv1.NodePool, machineDeployment 
 		// 1. if it's a new MachineDeployment, or the replicas field of the old MachineDeployment is <= min size, use min size
 		// 2. if the replicas field of the old MachineDeployment is > max size, use max size
 		//
-		// Guard: Enforce min=1 for non-AWS platforms to prevent scale-from-zero issues.
+		// Guard: Enforce min=1 for platforms that don't support scale-from-zero.
 		// Even if API validation fails or pre-existing objects exist with min=0, this prevents
 		// NodePools from being permanently stuck at 0 replicas on platforms that don't support
 		// scale-from-zero metadata.
 		effectiveMin := ptr.Deref(nodePool.Spec.AutoScaling.Min, 0)
-		if effectiveMin == 0 && nodePool.Spec.Platform.Type != hyperv1.AWSPlatform {
+		if effectiveMin == 0 && !supportedScaleFromZeroPlatform(nodePool.Spec.Platform.Type) {
 			effectiveMin = 1
 		}
 
@@ -1059,12 +1059,12 @@ func setMachineSetReplicas(nodePool *hyperv1.NodePool, machineSet *capiv1.Machin
 		// 1. if it's a new MachineSet, or the replicas field of the old MachineSet is <= min size, use min size
 		// 2. if the replicas field of the old MachineSet is > max size, use max size
 		//
-		// Guard: Enforce min=1 for non-AWS platforms to prevent scale-from-zero issues.
+		// Guard: Enforce min=1 for platforms that don't support scale-from-zero.
 		// Even if API validation fails or pre-existing objects exist with min=0, this prevents
 		// NodePools from being permanently stuck at 0 replicas on platforms that don't support
 		// scale-from-zero metadata.
 		effectiveMin := ptr.Deref(nodePool.Spec.AutoScaling.Min, 0)
-		if effectiveMin == 0 && nodePool.Spec.Platform.Type != hyperv1.AWSPlatform {
+		if effectiveMin == 0 && !supportedScaleFromZeroPlatform(nodePool.Spec.Platform.Type) {
 			effectiveMin = 1
 		}
 
--- a/hypershift-operator/controllers/nodepool/conditions.go
+++ b/hypershift-operator/controllers/nodepool/conditions.go
@@ -165,17 +165,7 @@ func (r *NodePoolReconciler) autoscalerEnabledCondition(_ context.Context, nodeP
 	if isAutoscalingEnabled(nodePool) {
 		// Check if autoscaling from zero (min=0) is being used
 		if nodePool.Spec.AutoScaling.Min != nil && *nodePool.Spec.AutoScaling.Min == 0 {
-			// Check platform-specific support
-			var supported bool
-			switch nodePool.Spec.Platform.Type {
-			case hyperv1.AWSPlatform:
-				// AWS supports scale-from-zero either natively (when CPO supports it)
-				// or via MachineDeployment controller workaround annotations
-				supported = true
-			default:
-				// Other platforms don't support autoscaling from zero yet
-				supported = false
-			}
+			supported := supportedScaleFromZeroPlatform(nodePool.Spec.Platform.Type)
 
 			if !supported {
 				SetStatusCondition(&nodePool.Status.Conditions, hyperv1.NodePoolCondition{
--- a/hypershift-operator/controllers/nodepool/instancetype/azure/provider.go
+++ b/hypershift-operator/controllers/nodepool/instancetype/azure/provider.go
@@ -1,0 +1,193 @@
+package azure
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+	"strings"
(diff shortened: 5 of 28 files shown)


gaol:
/verified by @gaol 

Using [azure-ipi-ovn-hypershift-mgmt-f7](azure-ipi-ovn-hypershift-mgmt-f7) job which has ho installed with:
```
--scale-from-zero-provider=azure
--scale-from-zero-creds=/etc/hypershift-ext-dns-app-azure/credentials
```
worker nodes can scale up from zero and scale down to zero.

hypershift-operator/controllers/nodepool/instancetype/azure/provider_test.go
```diff
@@ -0,0 +1,203 @@
...
+func (m *mockSKUClient) NewListPager(options *armcompute.ResourceSKUsClientListOptions) *runtime.Pager[armcompute.ResourceSKUsClientListResponse] {
+	if m.newListPagerFunc != nil {
+		return m.newListPagerFunc(options)
+	}
+	return nil
+}
+
+// Test helper functions
+
+func ptr[T any](v T) *T {
```
bryan-cox: Could you just use the k8s ptr function here?

hypershift-operator/controllers/nodepool/instancetype/azure/provider_test.go
bryan-cox, line 1: Test names are not following When it should format

hypershift-operator/main.go
```diff
@@ -432,6 +435,17 @@ func run(ctx context.Context, opts *StartOptions, log logr.Logger) error {
 			scaleFromZeroEC2Client := ec2.New(awsSession, awsConfig)
 			instanceTypeProvider = awsinstancetype.NewProvider(scaleFromZeroEC2Client)
 			log.Info("Instance type provider initialized", "provider", opts.ScaleFromZeroProvider)
+		case "azure":
```
bryan-cox: Shouldn't equalFolds be used here?

hypershift-operator/controllers/nodepool/capi.go
```diff
@@ -753,12 +753,12 @@ func setMachineDeploymentReplicas(nodePool *hyperv1.NodePool, machineDeployment
...
 		// 2. if the replicas field of the old MachineDeployment is > max size, use max size
 		//
-		// Guard: Enforce min=1 for non-AWS platforms to prevent scale-from-zero issues.
+		// Guard: Enforce min=1 for platforms that don't support scale-from-zero.
 		// Even if API validation fails or pre-existing objects exist with min=0, this prevents
 		// NodePools from being permanently stuck at 0 replicas on platforms that don't support
 		// scale-from-zero metadata.
 		effectiveMin := ptr.Deref(nodePool.Spec.AutoScaling.Min, 0)
-		if effectiveMin == 0 && nodePool.Spec.Platform.Type != hyperv1.AWSPlatform {
+		if effectiveMin == 0 && nodePool.Spec.Platform.Type != hyperv1.AWSPlatform && nodePool.Spec.Platform.Type != hyperv1.AzurePlatform {
```
bryan-cox: replace with supportedScaleFromZeroPlatform()

hypershift-operator/controllers/nodepool/capi.go
```diff
@@ -1059,12 +1059,12 @@ func setMachineSetReplicas(nodePool *hyperv1.NodePool, machineSet *capiv1.Machin
...
 		// 2. if the replicas field of the old MachineSet is > max size, use max size
 		//
-		// Guard: Enforce min=1 for non-AWS platforms to prevent scale-from-zero issues.
+		// Guard: Enforce min=1 for platforms that don't support scale-from-zero.
 		// Even if API validation fails or pre-existing objects exist with min=0, this prevents
 		// NodePools from being permanently stuck at 0 replicas on platforms that don't support
 		// scale-from-zero metadata.
 		effectiveMin := ptr.Deref(nodePool.Spec.AutoScaling.Min, 0)
-		if effectiveMin == 0 && nodePool.Spec.Platform.Type != hyperv1.AWSPlatform {
+		if effectiveMin == 0 && nodePool.Spec.Platform.Type != hyperv1.AWSPlatform && nodePool.Spec.Platform.Type != hyperv1.AzurePlatform {
```
bryan-cox: replace with supportedScaleFromZeroPlatform()

hypershift-operator/controllers/nodepool/conditions.go
bryan-cox: replace with supportedScaleFromZeroPlatform()

hypershift-operator/controllers/nodepool/scale_from_zero.go
bryan-cox: This is missing a unit test

hypershift-operator/controllers/nodepool/nodepool_controller.go
```diff
@@ -428,7 +428,7 @@ func (r *NodePoolReconciler) reconcile(ctx context.Context, hcluster *hyperv1.Ho
 
 // supportedScaleFromZeroPlatform checks if the platform supports scale-from-zero functionality.
 func supportedScaleFromZeroPlatform(platform hyperv1.PlatformType) bool {
```
bryan-cox, line 430: This is missing a unit test

hypershift-operator/controllers/nodepool/conditions.go
bryan-cox, line 164: This is missing a unit test

hypershift-operator/controllers/nodepool/nodepool_controller.go
bryan-cox, line 1149: This is missing a unit test

cmd/install/install.go
```diff
@@ -197,7 +197,7 @@ func (o *Options) Validate() error {
 	}
 
 	// Validate scale-from-zero credentials
-	supportedProviders := set.New("aws")
+	supportedProviders := set.New("aws", "azure")
```
bryan-cox, line 200: No test verifying azure is in the supportedProviders

hypershift-operator/controllers/nodepool/scale_from_zero.go
bryan-cox: Shouldn't interface{} be `any`?

hypershift-operator/controllers/nodepool/scale_from_zero.go
```diff
@@ -77,13 +78,13 @@ func setScaleFromZeroAnnotationsOnObject(ctx context.Context, provider instancet
...
 		instanceType = template.Spec.Template.Spec.InstanceType
 		statusCapacity = template.Status.Capacity
-	// Future platform support can be added here:
-	// case *capiazure.AzureMachineTemplate:
-	//     instanceType = template.Spec.Template.Spec.VMSize
-	//     statusCapacity = template.Status.Capacity
+	case *capz.AzureMachineTemplate:
+		instanceType = template.Spec.Template.Spec.VMSize
+		// Azure CAPI provider doesn't support Status.Capacity, so statusCapacity remains nil
+		statusCapacity = nil
```
bryan-cox: this could be removed since its redundant

hypershift-operator/controllers/nodepool/instancetype/azure/provider_test.go
bryan-cox, line 1: t.Parallel() should be used in this file and in the other files adding unit tests in this PR.

cmd/install/install.go
bryan-cox, line 200: This set is duplicated in both here and in main.go - could be a shared constant or set in `support/`

hypershift-operator/main.go
```diff
@@ -432,6 +435,17 @@ func run(ctx context.Context, opts *StartOptions, log logr.Logger) error {
...
+			azureCred, subscriptionID, cloudCfg, err := azureutil.NewScaleFromZeroCredential(opts.ScaleFromZeroCreds)
+			if err != nil {
+				return fmt.Errorf("failed to create Azure credentials for scale-from-zero: %w", err)
+			}
+			azureSKUClient, err := armcompute.NewResourceSKUsClient(subscriptionID, azureCred, azureutil.NewARMClientOptions(cloudCfg))
```
bryan-cox: This is inadvertently using CPOUserAgent here when it should not. This should use an operator specific user agent.

hypershift-operator/main.go
```diff
@@ -432,6 +435,17 @@ func run(ctx context.Context, opts *StartOptions, log logr.Logger) error {
...
+			if err != nil {
+				return fmt.Errorf("failed to create Azure SKU client for scale-from-zero: %w", err)
+			}
+			instanceTypeProvider = azureinstancetype.NewProvider(azureSKUClient)
+			log.Info("Instance type provider initialized", "provider", opts.ScaleFromZeroProvider, "subscriptionID", subscriptionID)
```
bryan-cox: Probably shouldn't log the subscription ID here

hypershift-operator/controllers/nodepool/instancetype/azure/provider.go
```diff
@@ -0,0 +1,180 @@
...
+		skuClient: skuClient,
+	}
+}
+
+// GetInstanceTypeInfo queries Azure Resource SKUs API for VM size specifications.
+// This information is used to populate cluster autoscaler capacity annotations
+// for scaling from zero replicas.
+// Note: VM size specifications (CPU, memory, GPU) are consistent across all Azure locations,
+// so we don't need to filter by location.
+func (p *Provider) GetInstanceTypeInfo(ctx context.Context, vmSize string) (*instancetype.InstanceTypeInfo, error) {
```
bryan-cox, line 45: Every time this function is called, it creates a new pager (line 44) and iterates through all Azure Resource SKUs — which can be thousands of entries across all resource types (VMs, disks, snapshots, etc.) —  just to find a single VM size.

This is called in a reconcile so I'm concerned about the performance of this. Maybe a sync.Map or cache to the Provider struct could be used once a VM is looked up, subsequent reconciles would return the cached result instead of re-looking up the specs again.

hypershift-operator/controllers/nodepool/nodepool_controller.go
bryan-cox, line 419: Copied this from a code reviewer, worth checking out - 

```
Line 419 is inside the main Reconcile method of the NodePoolReconciler — the standard controller-runtime reconcile loop at line 182. It runs every time a NodePool resource is created or updated.                                                                    
                                                                                                                                                                                                                                                                        
  So the flow is:                                                                                                                                                                                                                                                       
   
  1. Reconcile() (line 182) — triggered by NodePool create/update                                                                                                                                                                                                       
  2. CAPI resources are reconciled (line 409) — creates MachineDeployment, MachineSet, AzureMachineTemplate, etc.
  3. Line 419 — checks if scale-from-zero annotations should be set

  The concern is that the condition check (autoscalerEnabledCondition in conditions.go) runs in a separate code path earlier in the reconcile and reports "everything is fine" for Azure with min=0, but the actual annotation-setting at line 419 silently no-ops when
  InstanceTypeProvider is nil. There's a disconnect — the condition says "supported" but the runtime silently skips the work needed to make it actually function.
```

cmd/install/install.go
LiangquanLi930: IMHO, With only two providers, keeping them inline is more straightforward than introducing a shared constant.

hypershift-operator/controllers/nodepool/nodepool_controller.go
LiangquanLi930: @bryan-cox  I'll add the InstanceTypeProvider check to the condition in a follow-up PR, along with documentation on the required operator configuration flags.

cmd/install/install_test.go
```diff
@@ -191,6 +191,15 @@ func TestOptions_Validate(t *testing.T) {
 			},
 			expectError: false,
 		},
+		"when scale-from-zero provider is azure with secret reference there is no error": {
+			inputOptions: Options{
+				PrivatePlatform:                   string(hyperv1.NonePlatform),
```
bryan-cox, line 196: Should this not be Azure?

hypershift-operator/controllers/nodepool/nodepool_controller_test.go
```diff
@@ -2324,3 +2326,163 @@ func TestNodePoolReconciler_reconcile(t *testing.T) {
...
+
+	tests := []struct {
+		name         string
+		platform     hyperv1.PlatformType
+		objects      []client.Object
+		provider     instancetype.Provider
+		expectErr    bool
+		errSubstring string
+		validate     func(g Gomega, c client.Client)
+	}{
```
bryan-cox, line 2391: Should there not be an Azure test case here?

hypershift-operator/main.go
```diff
@@ -427,14 +430,24 @@ func run(ctx context.Context, opts *StartOptions, log logr.Logger) error {
...
-		case "aws":
+		if strings.EqualFold(opts.ScaleFromZeroProvider, "aws") {
 			awsSession := awsutil.NewSession("hypershift-operator-scale-from-zero", opts.ScaleFromZeroCreds, "", "", "")
 			awsConfig := awsutil.NewConfig()
 			scaleFromZeroEC2Client := ec2.New(awsSession, awsConfig)
 			instanceTypeProvider = awsinstancetype.NewProvider(scaleFromZeroEC2Client)
 			log.Info("Instance type provider initialized", "provider", opts.ScaleFromZeroProvider)
-		default:
+		} else if strings.EqualFold(opts.ScaleFromZeroProvider, "azure") {
+			azureCred, subscriptionID, cloudCfg, err := azureutil.NewScaleFromZeroCredential(opts.ScaleFromZeroCreds)
```
bryan-cox, line 440: Who provides these credentials from a managed service perspective? Is this on the Red Hat side or the customer side?

hypershift-operator/controllers/nodepool/instancetype/azure/provider.go
```diff
@@ -0,0 +1,191 @@
...
+	cacheKey := strings.ToLower(vmSize)
+
+	// Return cached result if available
+	if cached, ok := p.cache.Load(cacheKey); ok {
+		return cached.(*instancetype.InstanceTypeInfo), nil
+	}
+
+	// Cache miss — query Azure Resource SKUs API
+	pager := p.skuClient.NewListPager(&armcompute.ResourceSKUsClientListOptions{
+		Filter: nil, // We'll filter in code since API doesn't support filtering by name
```
bryan-cox: Could we not at least add something like filter := "resourceType eq 'virtualMachines'"?

hypershift-operator/main.go
LiangquanLi930: It should be on the Red Hat side, this should be consistent with the design of ROSA-HCP.

cmd/install/install_test.go
LiangquanLi930: IIUC, private-platform does not support Azure. In real ARO HCP deployments, --private-platform is not passed at all (it defaults to None). Therefore, using NonePlatform here is correct.

hypershift-operator/main.go
bryan-cox: To date, the HO has not needed to authenticate with Azure. If this comes from the Red Hat side, is the expectation the same credentials will be used with different customer's hosted clusters running from the same management cluster?

Is CS providing those credentials? Are they aware of that or this PR?

LiangquanLi930: We currently provide two approaches: one is to modify the CAPI providers directly; the other is to add some code (an annotator). This PR falls into the second category. If ARO-HCP does not need to support lower versions, then we can skip this configuration.

>  If this comes from the Red Hat side, is the expectation the same credentials will be used with different customer's hosted clusters running from the same management cluster?

For this point, I think as long as these credentials can access and view the instance information, that should be sufficient.

bryan-cox:
/hold

This still needs discussed with ARO HCP.

bryan-cox:
This PR also seems to have broken self-managed Azure e2e https://github.com/openshift/hypershift/pull/7482#issuecomment-3997424770

PR #7483, opened 2026-01-15
OCPBUGS-73847: fix(cli): use /tags endpoint for all release streams to enable RC filtering

bryan-cox:

This PR fixes an issue where the CLI failed when the latest version in a release stream was a release candidate (RC). 

### The Problem

When using release streams like `4-stable` or `4-stable-arm64`, the CLI was using the `/latest` endpoint which only returns a single version. If that version happened to be an RC (e.g., `4.21.0-rc.0`), the CLI would fail with:

```
release image is required when unable to lookup default OCP version: failed to get OCP version from release URL https://amd64.ocp.releases.ci.openshift.org/api/v1/releasestream/4-stable/latest: the latest version in stream is a release candidate (4.21.0-rc.0), which is not supported by this HyperShift Operator
```

### The Solution

This PR updates the release stream lookup to:

1. **Use `/tags` endpoint for all streams**: The `/tags` endpoint returns a list of all versions, enabling proper RC filtering
2. **Add architecture detection**: Detects the architecture from the stream suffix and routes to the correct endpoint:
   - `-multi` → `multi.ocp.releases.ci.openshift.org`
   - `-arm64` → `arm64.ocp.releases.ci.openshift.org`
   - `-ppc64le` → `ppc64le.ocp.releases.ci.openshift.org`
   - `-s390x` → `s390x.ocp.releases.ci.openshift.org`
   - default → `amd64.ocp.releases.ci.openshift.org`

3. **Leverage existing RC filtering**: The existing `findLatestSupportedVersion()` function already handles RC filtering - it now works for all streams

### Previous Fix (PR #7228)

PR #7228 attempted to fix this issue but only addressed multi-arch streams. Single-arch streams continued to use the `/latest` endpoint which cannot provide fallback versions when the latest is an RC.

Fixes the regression from PR #7228 where single-arch streams (4-stable, 4-stable-arm64, etc.) still failed when the latest version was an RC.

1. **Tests Added**: Updated `TestGetArchFromStream` to test architecture detection for all supported architectures (multi, arm64, ppc64le, s390x, and default amd64)

2. **Tests Updated**: `TestRetrieveSupportedOCPVersionWithRCFiltering` now tests RC filtering for multi-arch, amd64, and arm64 streams

3. **Backward Compatibility**: This change is fully backward compatible - the behavior for multi-arch streams remains the same, and single-arch streams now get proper RC filtering

## Checklist:

### DIFF
--- a/support/supportedversion/version.go
+++ b/support/supportedversion/version.go
@@ -24,8 +24,7 @@ import (
 
 // https://docs.ci.openshift.org/docs/getting-started/useful-links/#services
 const (
-	multiArchReleaseURLTemplate = "https://multi.ocp.releases.ci.openshift.org/api/v1/releasestream/%s/tags"
-	releaseURLTemplate          = "https://amd64.ocp.releases.ci.openshift.org/api/v1/releasestream/%s/latest"
+	releaseURLTemplate = "https://%s.ocp.releases.ci.openshift.org/api/v1/releasestream/%s/tags"
 )
 
 // LatestSupportedVersion is the latest minor OCP version supported by the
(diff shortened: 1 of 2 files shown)


bryan-cox:

bryan-cox:

PR #7484, opened 2026-01-15
NO-JIRA: test(autoscaling): make GPU annotation optional in scale-from-zero test

jhjaggars:

This PR fixes a test failure in `TestNodePoolAutoscalingScaleFromZero` that was failing when testing with non-GPU instance types.

The test unconditionally expected the `machine.openshift.io/GPU` annotation to be present on the MachineDeployment. However, the scale-from-zero controller only sets this annotation when the instance type has GPUs (see `hypershift-operator/controllers/nodepool/scale_from_zero.go:126-132`). When testing with non-GPU instances (like m5.large), the GPU annotation is intentionally not set, causing the test to fail.

**Changes:**
1. Removed mandatory GPU annotation check - The test now only requires vCPU, memoryMb, and labels annotations (which are always set)
2. Updated logging to gracefully handle missing GPU annotation in log output

Fixes test failures seen in https://github.com/openshift/release/pull/72805 which enables scale-from-zero testing in CI.

- The controller behavior is correct - verified from CI artifacts that vCPU, memoryMb, and labels annotations are properly set
- GPU annotation is correctly omitted for non-GPU instances per the implementation in `scale_from_zero.go`
- `make staticcheck` passes

## Checklist:
- [ ] This change includes docs. (N/A - test-only change)
- [ ] This change includes unit tests. (N/A - fixes existing e2e test)

### DIFF
--- a/test/e2e/autoscaling_test.go
+++ b/test/e2e/autoscaling_test.go
@@ -487,9 +487,7 @@ func testScaleFromZero(ctx context.Context, mgtClient crclient.Client, hostedClu
 					if _, ok := md.Annotations["machine.openshift.io/memoryMb"]; !ok {
 						return false, "missing memoryMb annotation", nil
 					}
-					if _, ok := md.Annotations["machine.openshift.io/GPU"]; !ok {
-						return false, "missing GPU annotation", nil
-					}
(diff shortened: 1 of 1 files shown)


bryan-cox:
/lgtm

Just rebased

PR #7489, opened 2026-01-15
CNTRLPLANE-2203: add ControlPlaneConnectionAvailable condition

sdminonne:

It introduces ControlPlaneConnectionAvailable condition. It detect connection problem executing a curl command. The curl is a simple GET to ` https://kubernetes.default.svc/api`

Fixes https://issues.redhat.com/browse/CNTRLPLANE-2203

This change includes e2e. Upgrade e2e is temporary modified since we're upgrading from 4.22+newCondition -> 4.22 and lack os condition is problematic. 
See https://github.com/openshift/hypershift/pull/7489/changes#diff-cebde6b83b4b372162b2c6a22710a4ab0b1838576120535593a66a3a9bdfc170R2944

## Checklist:

### DIFF
--- a/api/hypershift/v1beta1/hostedcluster_conditions.go
+++ b/api/hypershift/v1beta1/hostedcluster_conditions.go
@@ -214,7 +214,18 @@ const (
 	// A failure here suggests potential issues such as: network policy restrictions,
 	// firewall rules, missing data plane nodes, or problems with infrastructure
 	// components like the konnectivity-agent workload.
+	// **Unknown** means the status cannot be determined (e.g., no worker nodes available or unable to inspect).
 	DataPlaneConnectionAvailable ConditionType = "DataPlaneConnectionAvailable"
+
+	// ControlPlaneConnectionAvailable indicates whether data plane workloads have a successful
+	// network connection to the control plane components. This condition is computed per-node using
+	// a DaemonSet that performs reachability checks from each worker node.
+	// **True** means all data plane nodes can successfully reach the control plane (per-node reachability check passes for every node).
+	// **False** means there are connectivity failures preventing some or all data plane nodes from reaching the control plane,
+	// or required infrastructure components (such as kube-apiserver-proxy pods or the connection checker DaemonSet) are missing or unavailable.
+	// **Unknown** means the status cannot be determined due to true inability to inspect (e.g., no worker nodes exist or inspection cannot be performed),
+	// not due to missing required components.
+	ControlPlaneConnectionAvailable ConditionType = "ControlPlaneConnectionAvailable"
 )
 
 // Reasons.
@@ -279,6 +290,12 @@ const (
 	DataPlaneConnectionLogsAccessFailedReason = "LogsAccessFailed"
 
 	DataPlaneConnectionNoWorkerNodesAvailableReason = "NoWorkerNodesAvailable"
+
+	ControlPlaneConnectionKASAccessFailedReason = "KASAccessFailed"
+
+	ControlPlaneConnectionDaemonSetNotFoundReason = "DaemonSetNotFound"
+
+	ControlPlaneConnectionNoWorkerNodesAvailableReason = "NoWorkerNodesAvailable"
 
 	ControlPlaneComponentsNotAvailable = "ComponentsNotAvailable"
 )
--- a/control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
+++ b/control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
@@ -67,6 +67,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/apimachinery/pkg/util/sets"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
@@ -500,9 +501,26 @@ func (r *reconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Result
 		errs = append(errs, fmt.Errorf("failed to reconcile konnectivity agent: %w", err))
 	}
 
-	log.Info("reconciling control-Plane to data-Plane status conditions")
-	if err := r.reconcileControlPlaneDataPlaneConnectivityConditions(ctx, hcp, log); err != nil {
+	log.Info("reconciling KAS connection checker daemonset")
+	// The "pod" component is a standard component in OpenShift release payloads
+	// that provides this lightweight pause container from the release itself.
+	pauseImage, ok := releaseImage.ComponentImages()["pod"]
+	if !ok {
+		errs = append(errs, fmt.Errorf("failed to find pod image in release"))
+	} else {
+		if err := r.reconcileKASConnectionCheckerDaemonSet(ctx, hcp, pauseImage); err != nil {
+			errs = append(errs, fmt.Errorf("failed to reconcile KAS connection checker daemonset: %w", err))
+		}
+	}
+
+	log.Info("reconciling data plane connection available condition")
+	if err := r.reconcileDataPlaneConnectionAvailable(ctx, hcp, log); err != nil {
 		errs = append(errs, fmt.Errorf("failed to update ControlPlaneToDataPlaneConnectivity condition: %w", err))
+	}
+
+	log.Info("reconciling control plane connection available condition")
+	if err := r.reconcileControlPlaneConnectionAvailable(ctx, hcp); err != nil {
+		errs = append(errs, fmt.Errorf("failed to update ControlPlaneConnectionAvailable condition: %w", err))
 	}
 
 	log.Info("reconciling openshift apiserver apiservices")
@@ -1355,19 +1373,7 @@ func (r *reconciler) reconcileClusterVersion(ctx context.Context, hcp *hyperv1.H
 	return nil
 }
 
-func (r *reconciler) reconcileControlPlaneDataPlaneConnectivityConditions(ctx context.Context, hcp *hyperv1.HostedControlPlane, log logr.Logger) error {
-	patchHCPWithCondition := func(hcp *hyperv1.HostedControlPlane, condition *metav1.Condition) error {
-		originalHCP := hcp.DeepCopy()
-		if !meta.SetStatusCondition(&hcp.Status.Conditions, *condition) {
-			return nil // No status change; avoid unnecessary API call.
-		}
-		if err := r.cpClient.Status().Patch(ctx, hcp, client.MergeFrom(originalHCP)); err != nil {
-			return fmt.Errorf("failed to update HostedControlPlane status with %s condition: %w", condition.Type, err)
-		}
-		log.Info(string(hyperv1.DataPlaneConnectionAvailable) + " updated")
-		return nil
-	}
-
+func (r *reconciler) reconcileDataPlaneConnectionAvailable(ctx context.Context, hcp *hyperv1.HostedControlPlane, log logr.Logger) error {
 	condition := &metav1.Condition{
 		Type:   string(hyperv1.DataPlaneConnectionAvailable),
 		Status: metav1.ConditionFalse, // False by default
@@ -1377,20 +1383,20 @@ func (r *reconciler) reconcileControlPlaneDataPlaneConnectivityConditions(ctx co
 		condition.Status = metav1.ConditionUnknown
 		condition.Reason = hyperv1.ReconcileErrorReason
 		condition.Message = "Unable to count worker nodes: " + err.Error()
-		return patchHCPWithCondition(hcp, condition)
+		return r.patchHCPStatusCondition(ctx, hcp, condition)
 	}
 	if totalNodes == 0 {
 		condition.Status = metav1.ConditionUnknown
 		condition.Reason = hyperv1.DataPlaneConnectionNoWorkerNodesAvailableReason
 		condition.Message = "No worker nodes available"
-		return patchHCPWithCondition(hcp, condition)
+		return r.patchHCPStatusCondition(ctx, hcp, condition)
 	}
 	var podList corev1.PodList
 	if err := r.uncachedClient.List(ctx, &podList,
 		client.MatchingLabels{"app": "konnectivity-agent"}, client.InNamespace("kube-system")); err != nil {
 		condition.Reason = hyperv1.ReconciliationErrorReason
 		condition.Message = "Couldn't list konnectivity-agent PODs in kube-system namespace: " + err.Error()
-		return patchHCPWithCondition(hcp, condition)
+		return r.patchHCPStatusCondition(ctx, hcp, condition)
 	}
 
 	logsFound := false
@@ -1425,7 +1431,132 @@ func (r *reconciler) reconcileControlPlaneDataPlaneConnectivityConditions(ctx co
 		}
 	}
 
-	return patchHCPWithCondition(hcp, condition)
+	return r.patchHCPStatusCondition(ctx, hcp, condition)
+}
+
+// getKASHealthCheckEndpoint returns the appropriate KAS health check endpoint based on the platform type.
+// IBM Cloud uses a different liveness endpoint that excludes etcd and log checks.
+func getKASHealthCheckEndpoint(platformType hyperv1.PlatformType) string {
+	if platformType == hyperv1.IBMCloudPlatform {
+		return "/livez?exclude=etcd&exclude=log"
+	}
+	return "/version"
+}
+
+// patchHCPStatusCondition patches the HostedControlPlane status with the provided condition.
+// It only performs the API call if the condition actually changed.
+func (r *reconciler) patchHCPStatusCondition(ctx context.Context, hcp *hyperv1.HostedControlPlane, condition *metav1.Condition) error {
+	log := ctrl.LoggerFrom(ctx)
+	originalHCP := hcp.DeepCopy()
+	if !meta.SetStatusCondition(&hcp.Status.Conditions, *condition) {
+		return nil // No status change; avoid unnecessary API call.
+	}
+	if err := r.cpClient.Status().Patch(ctx, hcp, client.MergeFrom(originalHCP)); err != nil {
+		return fmt.Errorf("failed to update HostedControlPlane status with %s condition: %w", condition.Type, err)
+	}
+	log.Info(string(condition.Type) + " condition updated")
+	return nil
+}
+
+func (r *reconciler) reconcileKASConnectionCheckerDaemonSet(ctx context.Context, hcp *hyperv1.HostedControlPlane, pauseImage string) error {
+	kasAdvertiseAddress := util.GetAdvertiseAddress(hcp, config.DefaultAdvertiseIPv4Address, config.DefaultAdvertiseIPv6Address)
+	kasPort := util.KASPodPort(hcp)
+	endpoint := getKASHealthCheckEndpoint(hcp.Spec.Platform.Type)
+
+	daemonSet := manifests.KASConnectionCheckerDaemonSet()
+	if _, err := r.CreateOrUpdate(ctx, r.client, daemonSet, func() error {
+		// Configure the DaemonSet spec
+		daemonSet.Spec.Selector = &metav1.LabelSelector{
+			MatchLabels: map[string]string{
+				"app": manifests.KASConnectionCheckerName,
+			},
+		}
+
+		daemonSet.Spec.Template.ObjectMeta.Labels = map[string]string{
+			"app": manifests.KASConnectionCheckerName,
+		}
+
+		// Use a lightweight pause container from the release payload
+		daemonSet.Spec.Template.Spec.Containers = []corev1.Container{
+			{
+				Name:  "connection-checker",
+				Image: pauseImage,
+				ReadinessProbe: &corev1.Probe{
+					ProbeHandler: corev1.ProbeHandler{
+						HTTPGet: &corev1.HTTPGetAction{
+							Path:   endpoint,
+							Port:   intstr.FromInt(int(kasPort)),
+							Scheme: corev1.URISchemeHTTPS,
+							Host:   kasAdvertiseAddress,
+						},
+					},
+					InitialDelaySeconds: 5,
+					PeriodSeconds:       10,
+					TimeoutSeconds:      5,
+					SuccessThreshold:    1,
+					FailureThreshold:    3,
+				},
+			},
+		}
+
+		// Tolerate all taints so it runs on all nodes
+		daemonSet.Spec.Template.Spec.Tolerations = []corev1.Toleration{
+			{
+				Operator: corev1.TolerationOpExists,
+			},
+		}
+
+		return nil
+	}); err != nil {
+		return fmt.Errorf("failed to reconcile kas-connection-checker daemonset: %w", err)
+	}
+
+	return nil
+}
+
+func (r *reconciler) reconcileControlPlaneConnectionAvailable(ctx context.Context, hcp *hyperv1.HostedControlPlane) error {
+	condition := &metav1.Condition{
+		Type:   string(hyperv1.ControlPlaneConnectionAvailable),
+		Status: metav1.ConditionUnknown,
+		Reason: hyperv1.StatusUnknownReason,
+	}
+
+	// Get the KAS connection checker DaemonSet
+	daemonSet := manifests.KASConnectionCheckerDaemonSet()
+
+	if err := r.client.Get(ctx, client.ObjectKeyFromObject(daemonSet), daemonSet); err != nil {
+		if apierrors.IsNotFound(err) {
+			condition.Reason = hyperv1.ControlPlaneConnectionDaemonSetNotFoundReason
+			condition.Message = fmt.Sprintf("KAS connection checker DaemonSet %s/%s not found; the hosted cluster config operator may not have reconciled it yet",
+				manifests.KASConnectionCheckerNamespace, manifests.KASConnectionCheckerName)
+			return r.patchHCPStatusCondition(ctx, hcp, condition)
+		}
+		condition.Reason = hyperv1.ReconcileErrorReason
+		condition.Message = fmt.Sprintf("Failed to get KAS connection checker DaemonSet %s/%s: %v",
+			manifests.KASConnectionCheckerNamespace, manifests.KASConnectionCheckerName, err)
+		return r.patchHCPStatusCondition(ctx, hcp, condition)
+	}
+
+	// Check if there are any nodes to schedule pods on
+	if daemonSet.Status.DesiredNumberScheduled == 0 {
+		condition.Reason = hyperv1.ControlPlaneConnectionNoWorkerNodesAvailableReason
+		condition.Message = "KAS connection checker DaemonSet has 0 desired pods scheduled; no worker nodes are available to verify control plane connectivity"
+		return r.patchHCPStatusCondition(ctx, hcp, condition)
+	}
+
+	// Check if all desired pods are ready
+	if daemonSet.Status.NumberReady == daemonSet.Status.DesiredNumberScheduled {
+		condition.Status = metav1.ConditionTrue
+		condition.Reason = hyperv1.AsExpectedReason
+		condition.Message = hyperv1.AllIsWellMessage
+	} else {
+		condition.Status = metav1.ConditionFalse
+		condition.Reason = hyperv1.ControlPlaneConnectionKASAccessFailedReason
+		condition.Message = fmt.Sprintf("Data plane to control plane connection is not available: %d/%d pods ready",
+			daemonSet.Status.NumberReady, daemonSet.Status.DesiredNumberScheduled)
+	}
+
+	return r.patchHCPStatusCondition(ctx, hcp, condition)
 }
 
 func (r *reconciler) reconcileOpenshiftAPIServerAPIServices(ctx context.Context, hcp *hyperv1.HostedControlPlane) error {
--- a/control-plane-operator/hostedclusterconfigoperator/controllers/resources/manifests/kasconnectionchecker.go
+++ b/control-plane-operator/hostedclusterconfigoperator/controllers/resources/manifests/kasconnectionchecker.go
@@ -1,0 +1,23 @@
+package manifests
+
+import (
+	appsv1 "k8s.io/api/apps/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	// KASConnectionCheckerName is the name of the KAS connection checker DaemonSet
+	KASConnectionCheckerName = "kas-connection-checker"
+	// KASConnectionCheckerNamespace is the namespace where the DaemonSet is deployed
+	KASConnectionCheckerNamespace = "kube-system"
+)
+
+// KASConnectionCheckerDaemonSet returns an empty DaemonSet object for the KAS connection checker
+func KASConnectionCheckerDaemonSet() *appsv1.DaemonSet {
+	return &appsv1.DaemonSet{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      KASConnectionCheckerName,
+			Namespace: KASConnectionCheckerNamespace,
+		},
+	}
+}
--- a/control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources_test.go
+++ b/control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources_test.go
@@ -17,6 +17,7 @@ import (
 	"github.com/openshift/hypershift/control-plane-operator/hostedclusterconfigoperator/controllers/resources/manifests"
 	"github.com/openshift/hypershift/hypershift-operator/controllers/nodepool"
 	"github.com/openshift/hypershift/support/azureutil"
+	"github.com/openshift/hypershift/support/config"
 	"github.com/openshift/hypershift/support/globalconfig"
 	fakereleaseprovider "github.com/openshift/hypershift/support/releaseinfo/fake"
 	supportutil "github.com/openshift/hypershift/support/util"
@@ -91,6 +92,7 @@ var initialObjects = []client.Object{
 	manifests.ValidatingAdmissionPolicyBinding(fmt.Sprintf("%s-binding", kas.AdmissionPolicyNameInfra)),
 
 	fakeOperatorHub(),
+	manifests.KASConnectionCheckerDaemonSet(),
 }
 
 func shouldNotError(key client.ObjectKey) bool {
@@ -165,8 +167,12 @@ func TestReconcileErrorHandling(t *testing.T) {
 			cpClient:               fake.NewClientBuilder().WithScheme(api.Scheme).WithObjects(cpObjects...).WithStatusSubresource(&hyperv1.HostedControlPlane{}).Build(),
 			hcpName:                "foo",
 			hcpNamespace:           "bar",
-			releaseProvider:        &fakereleaseprovider.FakeReleaseProvider{},
-			ImageMetaDataProvider:  &imageMetaDataProvider,
+			releaseProvider: &fakereleaseprovider.FakeReleaseProvider{
+				Components: map[string]string{
+					"pod": "registry.k8s.io/pause:3.9",
+				},
+			},
+			ImageMetaDataProvider: &imageMetaDataProvider,
 		}
 		_, err := r.Reconcile(ctx, controllerruntime.Request{})
 		if err != nil {
@@ -195,8 +201,12 @@ func TestReconcileErrorHandling(t *testing.T) {
 			cpClient:               fake.NewClientBuilder().WithScheme(api.Scheme).WithObjects(cpObjects...).Build(),
 			hcpName:                "foo",
 			hcpNamespace:           "bar",
-			releaseProvider:        &fakereleaseprovider.FakeReleaseProvider{},
-			ImageMetaDataProvider:  &imageMetaDataProvider,
+			releaseProvider: &fakereleaseprovider.FakeReleaseProvider{
+				Components: map[string]string{
+					"pod": "registry.k8s.io/pause:3.9",
+				},
+			},
+			ImageMetaDataProvider: &imageMetaDataProvider,
 		}
 		_, err := r.Reconcile(ctx, controllerruntime.Request{})
 		if err != nil {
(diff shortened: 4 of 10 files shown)


sdminonne:
@coderabbit-ai review

sdminonne:
@coderabbit-ai ptal

sdminonne:
@coderabbit-ai both major issues should be fixed now. PTAL

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -160,6 +166,35 @@ func getPodLogs(ctx context.Context, clientSet *clientset.Clientset, namespace,
...
+
+	httpStatusCode, stderr, err := util.ExecCommandInPod(ctx, restConfig, clientSet, namespace, name, container, []string{
+		"sh", "-c",
+		"curl -o /dev/null -s -w \"%{http_code}\n\" --connect-timeout 3 --max-time 5 --cacert /var/run/secrets/kubernetes.io/serviceaccount/ca.crt --header \"Authorization: Bearer $(cat /var/run/secrets/kubernetes.io/serviceaccount/token)\" https://kubernetes.default.svc/api",
+	})
+	if err != nil {
+		return false, fmt.Errorf("failed to execute curl command in %s/%s pod container %s, stderr: %s: %v", namespace, name, container, stderr, err)
+	}
+	return httpStatusCode == "200", nil
+}
```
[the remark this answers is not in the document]
sdminonne: Thanks @coderabbitai I've fixed it

sdminonne:
@coderabbitai review

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -1416,77 +1454,83 @@ func (r *reconciler) reconcileClusterVersion(ctx context.Context, hcp *hyperv1.H
...
 	}
 	var podList corev1.PodList
 	if err := r.uncachedClient.List(ctx, &podList,
 		client.MatchingLabels{"app": "konnectivity-agent"}, client.InNamespace("kube-system")); err != nil {
-		condition.Reason = hyperv1.ReconciliationErrorReason
-		condition.Message = "Couldn't list konnectivity-agent PODs in kube-system namespace: " + err.Error()
-		return patchHCPWithCondition(hcp, condition)
+		dataPlaneConnectionCondition.Reason = hyperv1.ReconcileErrorReason
+		dataPlaneConnectionCondition.Message = "Couldn't list konnectivity-agent PODs in kube-system namespace: " + err.Error()
+		return r.PatchHCPStatusConditions(ctx, r.cpClient, hcp, dataPlaneConnectionCondition, controlPlaneConnectionCondition)
```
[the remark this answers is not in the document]
sdminonne: @coderabbitai should be fixed now

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -172,7 +172,7 @@ func checkPodKasConnection(ctx context.Context, restConfig *rest.Config, clientS
 
 	httpStatusCode, stderr, err := util.ExecCommandInPod(ctx, restConfig, clientSet, namespace, name, container, []string{
 		"sh", "-c",
-		"curl -o /dev/null -s -w \"%{http_code}\" --connect-timeout 3 --max-time 5 --cacert /var/run/secrets/kubernetes.io/serviceaccount/ca.crt --header \"Authorization: Bearer $(cat /var/run/secrets/kubernetes.io/serviceaccount/token)\" https://kubernetes.default.svc/api",
+		"curl -k -o /dev/null -s -w \"%{http_code}\" --connect-timeout 3 --max-time 5 https://172.20.0.1:6443/version",
```
mgencur: The address and port might be different. The logic is here:
https://github.com/openshift/hypershift/blob/main/hypershift-operator/controllers/nodepool/apiserver-haproxy/haproxy.go#L122-L135
We should probably use something similar to determine the address and port to check.
This is just one comment, I will review later when the PR is ready.

sdminonne: I modified that code. @mgencur sorry. Please, mind wait for my green light before reviewing?
TY!

sdminonne:
@coderabbitai reiew

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -160,6 +166,41 @@ func getPodLogs(ctx context.Context, clientSet *clientset.Clientset, namespace,
...
 
+func checkPodKasConnection(ctx context.Context, restConfig *rest.Config, clientSet clientset.Interface, namespace, name, container string, hcp *hyperv1.HostedControlPlane) (bool, error) {
+	ctx, cancel := context.WithTimeout(ctx, 60*time.Second)
+	defer cancel()
+
+	// Use the advertise address where kube-apiserver-proxy actually listens
+	// This is the IP configured on the loopback interface of each node (172.20.0.1 for IPv4 or fd00::1 for IPv6)
+	kasIP := util.GetAdvertiseAddress(hcp, config.DefaultAdvertiseIPv4Address, config.DefaultAdvertiseIPv6Address)
+	kasPort := util.KASPodPort(hcp)
+	kasURL := "https://" + net.JoinHostPort(kasIP, fmt.Sprintf("%d", kasPort)) + "/version"
```
mgencur: The endpoint should be different on IBM cloud. See https://github.com/openshift/hypershift/blob/main/hypershift-operator/controllers/nodepool/apiserver-haproxy/haproxy.go#L254
Ideally, we extract this into a helper function and choose the endpoint based on cloud in all places using this function.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -561,8 +605,8 @@ func (r *reconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Result
 		errs = append(errs, fmt.Errorf("failed to reconcile konnectivity agent: %w", err))
 	}
 
-	log.Info("reconciling control-Plane to data-Plane status conditions")
-	if err := r.reconcileControlPlaneDataPlaneConnectivityConditions(ctx, hcp, log); err != nil {
+	log.Info("reconciling control data plane connection status conditions")
```
mgencur: Nit: Maybe "reconciling control plane and data plane connection status conditions" ?

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -1416,77 +1460,84 @@ func (r *reconciler) reconcileClusterVersion(ctx context.Context, hcp *hyperv1.H
...
-	condition := &metav1.Condition{
+func (r *reconciler) reconcileConnectionConditions(ctx context.Context, hcp *hyperv1.HostedControlPlane, log logr.Logger) error {
+	dataPlaneConnectionCondition := &metav1.Condition{
 		Type:   string(hyperv1.DataPlaneConnectionAvailable),
 		Status: metav1.ConditionFalse, // False by default
 	}
+	controlPlaneConnectionCondition := &metav1.Condition{
+		Type:    string(hyperv1.ControlPlaneConnectionAvailable),
+		Reason:  hyperv1.ReconcileErrorReason,
+		Status:  metav1.ConditionUnknown, // Unknown by default
```
mgencur: Could we unify this initial state for both `dataPlaneConnectionCondition` and `controlPlaneConnectionCondition`? 
The `ConditionUnknown` is not documented in docs/content/reference/api.md for any of these DataPlaneConnectionAvailable and ControlPlaneConnectionAvailable.
It depends what we want but I would prefer something like this:
```
dataPlaneConnectionCondition := &metav1.Condition{
		Type:      string(hyperv1.DataPlaneConnectionAvailable),
		Status:   metav1.ConditionUnknown,
        Reason: hyperv1.StatusUnknownReason
	}
controlPlaneConnectionCondition := &metav1.Condition{
		Type:         string(hyperv1.ControlPlaneConnectionAvailable),
		Status:      metav1.ConditionUnknown,
		Reason: hyperv1.StatusUnknownReason
	}
```
...and document the Unknown state in API docs possibly.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -1416,77 +1460,84 @@ func (r *reconciler) reconcileClusterVersion(ctx context.Context, hcp *hyperv1.H
...
+			}
+			if kasConnectionFound {
+				controlPlaneConnectionCondition.Status = metav1.ConditionTrue
+				controlPlaneConnectionCondition.Reason = hyperv1.AsExpectedReason
+				controlPlaneConnectionCondition.Message = hyperv1.AllIsWellMessage
+				continue
+			}
+			controlPlaneConnectionCondition.Status = metav1.ConditionFalse
+			controlPlaneConnectionCondition.Reason = hyperv1.ControlPlaneConnectionKASAccessFailedReason
+			controlPlaneConnectionCondition.Message = "Failed to access KAS from data plane"
```
mgencur: The connection should be there from all konnectivity pods (one on each Node). When we reach this point and kas connection is not found we should probably `break` the cycle. Otherwise, if the last Pod returns `True` the final condition will be True (nothing can change it anymore).
Example:
1) Konnectivity pod on Node 1 returns False
2) Konnectivity pod on Node 2 returns False
3) Konnectivity pod on Node 3 returns True
4) The final state is True even though the first two nodes can't reach the control plane.

Does it make sense?

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -1416,77 +1460,84 @@ func (r *reconciler) reconcileClusterVersion(ctx context.Context, hcp *hyperv1.H
...
 		if pod.Status.Phase != corev1.PodRunning {
 			continue
 		}
 		runningPodsFound = true
-		data, err := r.GetPodLogs(ctx, r.clientSet, pod.Namespace, pod.Name, "konnectivity-agent")
-		if err != nil {
-			log.Error(err,
-				fmt.Sprintf("failed to get logs for konnectivity-agent pod %s/%s", pod.Namespace, pod.Name))
-			continue
+		if dataPlaneConnectionCondition.Status != metav1.ConditionTrue { // skip operation in case logs already detected
```
mgencur: See my comment below. This is similar. What if the first node reports True and the other two report False? I believe we want to report overall True (control plane reachable) only if all three Nodes can reach it?

api/hypershift/v1beta1/hostedcluster_conditions.go
```diff
@@ -215,6 +215,12 @@ const (
 	// firewall rules, missing data plane nodes, or problems with infrastructure
 	// components like the konnectivity-agent workload.
 	DataPlaneConnectionAvailable ConditionType = "DataPlaneConnectionAvailable"
+
+	// ControlPlaneConnectionAvailable indicates whether data plane workloads have a successful
+	// network connection to the control plane components.
+	// **True** means data plane workloads can successfully reach the control plane.
+	// **False** means there are issues preventing data plane workloads from connecting to the control plane.
```
mgencur: Should we document the Unknown state as well? For `DataPlaneConnectionAvailable` as well.

mgencur:
The TestCreateClusterProxy test fails in [HCCO](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_hypershift/7489/pull-ci-openshift-hypershift-main-e2e-aws/2015881028446457856/artifacts/e2e-aws/hypershift-aws-run-e2e-external/artifacts/TestCreateClusterProxy/namespaces/e2e-clusters-rxn6z-proxy-npwcb/core/pods/logs/hosted-cluster-config-operator-5d4d7cd5fd-r5dks-hosted-cluster-config-operator.log) with this error:
```
{"level":"error","ts":"2026-01-26T22:22:16Z","msg":"failed to check connectivity from Pod to kas","controller":"resources","object":{"name":""},"namespace":"","name":"","reconcileID":"a31cd9de-1431-49b2-9cd5-abb9d9508fbc","error":"failed to execute curl command in kube-system/konnectivity-agent-n96th pod container konnectivity-agent, stderr: : command terminated with exit code 28","stacktrace":"github.com/openshift/hypershift/control-plane-operator/hostedclusterconfigoperator/controllers/resources.(*reconciler).reconcileConnectionConditions\n\t/hypershift/control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go:1521\ngithub.com/openshift/hypershift/control-plane-operator/hostedclusterconfigoperator/controllers/resources.(*reconciler).Reconcile\n\t/hypershift/control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go:609\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Reconcile\n\t/hypershift/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:116\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).reconcileHandler\n\t/hypershift/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:303\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).processNextWorkItem\n\t/hypershift/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:263\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Start.func2.2\n\t/hypershift/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:224"}
```
I think the commit message for the commit that fixes the case with HTTP_PROXY is helpful: https://github.com/openshift/hypershift/commit/228f256756499d377c04342ed60e9155ff26d64a

mgencur:
For the case with HTTP_PROXY, I think it will need to use HTTP CONNECT. Something like the HA proxy itself does: https://github.com/openshift/hypershift/blob/main/kubernetes-default-proxy/kubernetes_default_proxy.go#L98

api/hypershift/v1beta1/hostedcluster_conditions.go
sdminonne: Yes, I think we should. Good catch!

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
sdminonne: As soon one connection controlPlane to dataPlane is succesful the connection is available.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
sdminonne: My point is that as soon one Konnectivity pod can reach the KAS the connection is possible. I implmeneted in this way. 
Should the curl be successful for every POD? 
I think we need @csrwng opinion.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
sdminonne: yeah, to be honest I told my-self to do multiple time and I left like that.
Going to do. Thanks

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
sdminonne: totally

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
sdminonne: Thanks, looking and for sure I'm going to extract an helper function for this

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
mgencur: My point is, the konnectivity (or kube-apiserver-proxy) might run on some nodes and on some nodes it may fail.
The user probably expects (IMO!) that reporting ControlPlaneConnectionAvailable and DataPlaneConnectionAvailable as "True" means there's no problem in connection from all expected nodes.
Below is a non-comprehensive list of possible reasons why the pod might not run on certain nodes:
* The container image might fail to pull on certain nodes:                   
  - Network issues on specific nodes                                         
  - Private registry authentication missing on some nodes                    
  - Different image architectures (amd64 vs arm64)                           
  - Disk space issues preventing image download
* Resource constraints
* Pod security policies

mgencur:
I have tested a couple of scenarios with Pods failing in the guest cluster:
* Konnectivity agent pods failing in kube-system namespace on the guest cluster

Pods in guest cluster:
```
ᐅ oc get pod -l app=konnectivity-agent -n clusters-mgencur-hc1       
NAME                                 READY   STATUS             RESTARTS   AGE
konnectivity-agent-bb98d9889-5fdct   0/1     ImagePullBackOff   0          2m32s
```

HC conditions as expected:
```
  - lastTransitionTime: "2026-01-28T08:26:45Z"
    message: Couldn't find any konnectivity-agent running in data plane
    observedGeneration: 3
    reason: KonnectivityAgentPodsNotFound
    status: "False"
    type: DataPlaneConnectionAvailable
  - lastTransitionTime: "2026-01-28T08:19:20Z"
    message: Unable to inspect data plane to control plane connection
    observedGeneration: 3
    reason: ReconcileError
    status: Unknown
    type: ControlPlaneConnectionAvailable
```
    
HCP conditions same as HC conditions.

* Konnectivity agent pods running but kube-apiserver-proxy pods failing

Pods in guest cluster
```
ᐅ oc get pods -n kube-system                               
NAME                                                READY   STATUS             RESTARTS   AGE
global-pull-secret-syncer-nf9vp                     1/1     Running            0          5m28s
global-pull-secret-syncer-vrhlf                     1/1     Running            0          11m
konnectivity-agent-r5zcc                            1/1     Running            0          5m28s
konnectivity-agent-snc8c                            1/1     Running            0          21m
konnectivity-agent-wqzs2                            1/1     Running            0          11m
kube-apiserver-proxy-ip-10-0-128-95.ec2.internal    0/1     ImagePullBackOff   0          11m
kube-apiserver-proxy-ip-10-0-129-175.ec2.internal   0/1     ErrImagePull       0          21m
kube-apiserver-proxy-ip-10-0-131-243.ec2.internal   0/1     ImagePullBackOff   0          5m28s
```

HC conditions:
```
  - lastTransitionTime: "2026-01-28T09:32:06Z"
    message: All is well
    observedGeneration: 3
    reason: AsExpected
    status: "True"
    type: DataPlaneConnectionAvailable
    
  - lastTransitionTime: "2026-01-28T09:56:28Z"
    message: Unable to inspect data plane to control plane connection
    observedGeneration: 3
    reason: ReconcileError
    status: Unknown
    type: ControlPlaneConnectionAvailable
```

HCP conditions same as HC conditions.

So, the propagation works as expected when testing this end-to-end. 

Question: Should the status be "False" when the kube-apiserver-proxy Pods are not running? We generally put "Unknown" when there's a connection error via the `curl` command. But maybe for this control plane connection it should return "False" if there's an error. There's one case where konnectivity pods are not running and we properly return "Unknown" in this case because there's no way to check the connection to control plane in this case.
Anyway, if kube-apiserver-proxy pods are not running the control-plane connection should clearly be "False" and not "Unknown" because kube-apiserver-proxy is the key component in this communication.

mgencur:
/test e2e-kubevirt-aws-ovn

This original JIRA comes from environments like Kubevirt and Baremetal. Let's test kubevirt here through the E2E tests.

mgencur:
@sdminonne Thanks for the changes. It looks much better now. The only two concerns I have now:
* The conditions should return "True" only if all Nodes can reach the KAS. See my point in [this comment](https://github.com/openshift/hypershift/pull/7489#discussion_r2735798167)
* When kube-apiserver-proxy is not available the ControlPlaneConnectionAvailable should be False. Currently, it's "Unknown". See my tests and explanation in [another comment](https://github.com/openshift/hypershift/pull/7489#issuecomment-3810315120)

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
sdminonne: Hmm unfortunately I disagree, as soon one konnectivity agent is running the dataPlaneConnection has to be available. 'cause one konnectivity-agent is enough to grant connection. At least this is what I got from my preliminary discussion with @csrwng 

Concerning the controlPlaneConnectionAvailable I don't have a strong opinion even if I'm inclined to think that if one curl works then network connection could be up and running and having one curl running is definitiely more efficient than having tens (or worst hudreds) of curl running.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
sdminonne: Got it, good point. It's a matter of trying to guessing user (hcp owner) expectation but I would be careful trying to fit too many (even if legit and possible) errors behind one single condition.
To me these conditions means:

1. dataPlaneConnectionAvailable: no network config or error is blocking routing from control plane to data plane
2. controlPlaneConnectionAvaileble: no network config error is blocking routing from data plane to control plane

Do you agree? Do you think that the code is reflecting it?
Thanks!

mgencur: I think the code is reflecting that. I just don't know how likely it is that some specific nodes would have network issues. In this case it would be good to check all nodes instead of just one. But checking just one might be good enough. Not sure.

sdminonne:
- lastTransitionTime: "2026-02-02T10:22:20Z"
    message: All is well
    observedGeneration: 3
    reason: AsExpected 
    status: "True"
    type: DataPlaneConnectionAvailable
  - lastTransitionTime: "2026-02-02T10:22:17Z"
    message: All is well
    observedGeneration: 3
    reason: AsExpected
    status: "True"
    type: ControlPlaneConnectionAvailable

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -561,11 +562,26 @@ func (r *reconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Result
 		errs = append(errs, fmt.Errorf("failed to reconcile konnectivity agent: %w", err))
 	}
 
-	log.Info("reconciling control-Plane to data-Plane status conditions")
+	log.Info("reconciling KAS connection checker daemonset")
+	pauseImage, ok := releaseImage.ComponentImages()["pod"]
```
mgencur, line 507: Nit. Extract as a const and give it some description

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -561,11 +562,26 @@ func (r *reconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Result
...
+	if !ok {
+		errs = append(errs, fmt.Errorf("failed to find pod image in release"))
```
mgencur, line 509: Nit: Highlight that "pod image" is actually a release image for "pod" component, not just a pod image which is a general term.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -1417,18 +1433,6 @@ func (r *reconciler) reconcileClusterVersion(ctx context.Context, hcp *hyperv1.H
 }
 
 func (r *reconciler) reconcileControlPlaneDataPlaneConnectivityConditions(ctx context.Context, hcp *hyperv1.HostedControlPlane, log logr.Logger) error {
```
mgencur: Just curious if it makes sense to rename this to `reconcileDataPlaneConnectionAvailable`, i.e. exclude the "control plane" from the name as we have the condition "DataPlaneConnectionAvailable" which also doesn't include "control plane" in the name. That might prevent confusion in the future.
The other function is now called `reconcileControlPlaneConnectionAvailable`. So we can align them.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -1486,7 +1490,135 @@ func (r *reconciler) reconcileControlPlaneDataPlaneConnectivityConditions(ctx co
...
+	}
+	if err := r.cpClient.Status().Patch(ctx, hcp, client.MergeFrom(originalHCP)); err != nil {
+		return fmt.Errorf("failed to update HostedControlPlane status with %s condition: %w", condition.Type, err)
+	}
+	log.Info(conditionTypeName + " condition updated")
+	return nil
+}
+
+func (r *reconciler) reconcileKASConnectionCheckerDaemonSet(ctx context.Context, hcp *hyperv1.HostedControlPlane, pauseImage string) error {
+	kasAdvertiseAddress := util.GetAdvertiseAddress(hcp, "172.20.0.1", "fd00::1")
```
mgencur: Why not to use the constants? `config.DefaultAdvertiseIPv4Address`, `config.DefaultAdvertiseIPv6Address`

mgencur (commented):
@sdminonne Looks great! Just a few minor comments here and there.

I'm also wondering why the E2E test with http proxy doesn't fail, I'm going to check this myself separately.

mgencur:
> I'm also wondering why the E2E test with http proxy doesn't fail, I'm going to check this myself separately.

The specific kube-apiserver-proxy is only used for HTTPS PROXY, not HTTP PROXY. See [this line](https://github.com/openshift/hypershift/blob/main/hypershift-operator/controllers/nodepool/apiserver-haproxy/haproxy.go#L138) which sets the `apiserverProxy` variable which is later used [here](https://github.com/openshift/hypershift/blob/main/hypershift-operator/controllers/nodepool/apiserver-haproxy/haproxy.go#L281) to decide if the specific kube-apiserver-proxy should be used.
Our E2E tests only verify HTTP PROXY, not HTTPS.
I'm trying to verify with my Public cluster on AWS but getting this error from the Node's AWS machine:
```
[     563.728794] ignition[697]: GET error: Get "https://ignition-server-clusters-mgencur-hc1.apps.mgencur-mgmt.mgencur.hypershift.devcluster.openshift.com/ignition": proxyconnect tcp: tls: failed to verify certificate: x509: certificate signed by unknown authority
```
So, I'm not able to verify now. Will continue later but I suppose the problem with HTTPS proxy is there.

Update: Found this old issue: https://issues.redhat.com/browse/OCPBUGS-57060 which prevents the verification with HTTPS proxy.
Apparently, this whole functionality of the HTTPS proxy with public clusters on AWS is broken. But, I'm pretty sure the DaemonSet in this PR needs to use a different way to check the connection to KAS in this setup. We should probably create a separate issue to implement this later, and make it depend on https://issues.redhat.com/browse/OCPBUGS-57060

mgencur:
Just checked the negative scenario with kube-apiserver-proxy Pod not starting and results are as expected.
This is without HTTPS or HTTP proxy.
Pods in kube-system namespace on guest cluster:
```
kas-connection-checker-6lqjk                        0/1     Running             0          14s
kas-connection-checker-bjr5v                        1/1     Running             0          16m
kas-connection-checker-d9zf2                        1/1     Running             0          16m
kube-apiserver-proxy-ip-10-0-134-112.ec2.internal   1/1     Running             0          16m
kube-apiserver-proxy-ip-10-0-135-65.ec2.internal    0/1     ErrImagePull        0          6m53s
kube-apiserver-proxy-ip-10-0-141-225.ec2.internal   1/1     Running             0          16m
```

HC conditions:
```
ᐅ oc get hc $CLUSTER_NAME -n $NAMESPACE -ojsonpath='{.status.conditions[?(@.type=="DataPlaneConnectionAvailable")]}' | jq .   
{
  "lastTransitionTime": "2026-02-03T16:12:39Z",
  "message": "All is well",
  "observedGeneration": 3,
  "reason": "AsExpected",
  "status": "True",
  "type": "DataPlaneConnectionAvailable"
}
ᐅ oc get hc $CLUSTER_NAME -n $NAMESPACE -ojsonpath='{.status.conditions[?(@.type=="ControlPlaneConnectionAvailable")]}' | jq .
{
  "lastTransitionTime": "2026-02-03T16:20:44Z",
  "message": "Data plane to control plane connection is not available: 2/3 pods ready",
  "observedGeneration": 3,
  "reason": "KASAccessFailed",
  "status": "False",
  "type": "ControlPlaneConnectionAvailable"
}
```

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
sdminonne: added a comment

mgencur:
/lgtm

Pending: Create a JIRA to fix the case with HTTPS proxy as mentioned in [this comment](https://github.com/openshift/hypershift/pull/7489#issuecomment-3842032525) and make it depend on [OCPBUGS-57060](https://issues.redhat.com/browse/OCPBUGS-57060)

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
sdminonne: done

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
sdminonne: added a comment `pod` is needed

mgencur:
Correction: with HTTPS proxy, we're hitting [OCPBUGS-56896](https://issues.redhat.com/browse/OCPBUGS-56896), not [OCPBUGS-57060](https://issues.redhat.com/browse/OCPBUGS-57060). Discussing on slack.

sdminonne:
@mgencur added https://issues.redhat.com/browse/OCPBUGS-75883 to track the proxy issue

mgencur:
/test e2e-kubevirt-aws-ovn

This original JIRA comes from environments like Kubevirt and Baremetal. Let's test kubevirt here through the E2E tests.

mgencur:
I think the run e2e-kubevirt-aws-ovn looks good. One test failed but it failed in teardown phase. Looks unrelated to this PR.

api/hypershift/v1beta1/hostedcluster_conditions.go
```diff
@@ -214,7 +214,18 @@ const (
 	// A failure here suggests potential issues such as: network policy restrictions,
 	// firewall rules, missing data plane nodes, or problems with infrastructure
 	// components like the konnectivity-agent workload.
+	// **Unknown** means the status cannot be determined (e.g., no worker nodes available or unable to inspect).
```
devguyio, line 217: I wonder if it's a better option to differentiate between a state when we KNOW that we have no workers (i.e all nodepool replicas are 0) and an unknown state.

devguyio: To be precise, if we know that the NodePool replicas are all 0, then IMHO we should set the condition to `False` with a reason like `NodePoolsScaledToZero` or something like that.

sdminonne: @devguyio I know it's confusing. In this PR there's no code related to `dataPlaneConnectionAvailable` condition. I documented the Unknwown value following a @mgencur suggestion in one previous feedback/review https://github.com/openshift/hypershift/pull/7489#discussion_r2731529034

`dataPlaneConnectionAvailable` condition are set to `unknown` but the message should be different
* https://github.com/openshift/hypershift/blob/e62906a22225b5072bee325b2a916ad701abc00f/control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go#L1440
* https://github.com/openshift/hypershift/blob/e62906a22225b5072bee325b2a916ad701abc00f/control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go#L1446
* https://github.com/openshift/hypershift/blob/e62906a22225b5072bee325b2a916ad701abc00f/control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go#L1453

Hope it helps

sdminonne:
```
  E2E-AWS Failure Analysis for PR #7489
                                                                                                                             
  Summary                                

  The latest ci/prow/e2e-aws run (build 2020797473169084416) had 7 test failures, all within the TestNodePool hierarchy. None
   of the failures are caused by the PR's ControlPlaneConnectionAvailable condition changes. They are infrastructure/timing
  issues.
```

sdminonne:
@mgencur may you put another lgtm?

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -1425,7 +1431,139 @@ func (r *reconciler) reconcileControlPlaneDataPlaneConnectivityConditions(ctx co
...
+	log := ctrl.LoggerFrom(ctx)
+	originalHCP := hcp.DeepCopy()
+	if !meta.SetStatusCondition(&hcp.Status.Conditions, *condition) {
+		return nil // No status change; avoid unnecessary API call.
+	}
+	if err := r.cpClient.Status().Patch(ctx, hcp, client.MergeFrom(originalHCP)); err != nil {
+		return fmt.Errorf("failed to update HostedControlPlane status with %s condition: %w", condition.Type, err)
+	}
+	log.Info(conditionTypeName + " condition updated")
+	return nil
```
muraee:
```suggestion
func (r *reconciler) patchHCPStatusCondition(ctx context.Context, hcp *hyperv1.HostedControlPlane, condition *metav1.Condition) error {
	log := ctrl.LoggerFrom(ctx)
	originalHCP := hcp.DeepCopy()
	if !meta.SetStatusCondition(&hcp.Status.Conditions, *condition) {
		return nil // No status change; avoid unnecessary API call.
	}
	if err := r.cpClient.Status().Patch(ctx, hcp, client.MergeFrom(originalHCP)); err != nil {
		return fmt.Errorf("failed to update HostedControlPlane status with %s condition: %w", condition.Type, err)
	}
	log.Info(condition.Type + " condition updated")
	return nil
```

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -1425,7 +1431,139 @@ func (r *reconciler) reconcileControlPlaneDataPlaneConnectivityConditions(ctx co
...
+					PeriodSeconds:       10,
+					TimeoutSeconds:      5,
+					SuccessThreshold:    1,
+					FailureThreshold:    3,
+				},
+			},
+		}
+
+		// Use host network to access the KAS advertise address via loopback
+		daemonSet.Spec.Template.Spec.HostNetwork = true
```
muraee: why try to access via loopback?
This should simulate normal workloads which connect via the kubernetes svc.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -1425,7 +1431,139 @@ func (r *reconciler) reconcileControlPlaneDataPlaneConnectivityConditions(ctx co
...
+		Status: metav1.ConditionUnknown,
+		Reason: hyperv1.StatusUnknownReason,
+	}
+
+	// Get the KAS connection checker DaemonSet
+	daemonSet := &appsv1.DaemonSet{}
+	daemonSetKey := client.ObjectKey{
+		Namespace: manifests.KASConnectionCheckerNamespace,
+		Name:      manifests.KASConnectionCheckerName,
+	}
```
muraee:
```suggestion
	daemonSet := manifests.KASConnectionCheckerDaemonSet()
```

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
sdminonne: after a chat with @csrwng and @mgencur we ended up realizing that creating one workload per node can not scale for clusters with many nodes, avoiding hitting kas too much

mgencur: I suppose it's a left over from the time when this was implemented as a `curl` command running in konnectivity-agent pod - that one uses HostNetwork and the kubernetes.default.svc didn't work.
But now that this is a new DeamonSet, it could possibly work if the pods were not running on HostNetwork.

sdminonne: Unsure about this. @muraee @mgencur. Are you arguing I should remove the `HostNetwork = true` ?

mgencur: OK. if I look at kube-apiserver-proxy which runs on hostnetwork, there is an instance running on each Node and all of them refer to the same host (172.20.0.1):
```
      livenessProbe:
        failureThreshold: 3
        httpGet:
          host: 172.20.0.1
```
And the endpoint slice for kubernetes.default.svc points to it too:
```
ᐅ oc get endpoints kubernetes -oyaml
Warning: v1 Endpoints is deprecated in v1.33+; use discovery.k8s.io/v1 EndpointSlice
apiVersion: v1
kind: Endpoints
metadata:
  creationTimestamp: "2026-01-21T08
20Z"
  labels:
    endpointslice.kubernetes.io/skip-mirror: "true"
    hypershift.openshift.io/managed: "true"
  name: kubernetes
  namespace: default
  resourceVersion: "1181"
  uid: 75c3c25a-8a88-41fb-b3db-e5d4c6ef4bf5
subsets:
- addresses:
  - ip: 172.20.0.1
  ports:
  - name: https
    port: 6443
    protocol: TCP
```
I think removing HostNetwork and using the kubernetes.default.svc in the readiness probe instead of of the IP address could work and it would cover a little more (the SVC, and Endpoints).
There are some small downsides:
* unnecessary network/DNS/iptables dependencies
* slower probe execution (network vs local)

muraee: exactly as Martin suggested, this is the typical scenario for workloads, which is what we want to test here.

muraee: > we ended up realizing that creating one workload per node can not scale for clusters with many nodes, avoiding hitting kas too much

This is still true as you are creating a DaemonSet. Shouldn't this be a deployment instead with node anti-affinity, so that each replica lands on a different node?

sdminonne: Hmm, Ok got the point but still I think `[DaemonSet](https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/)` is the proper choice.

sdminonne:
converting to draft

sdminonne:
@mgencur @muraee PTAL

mgencur:
LGTM

PR #7541, opened 2026-01-18
OCPBUGS-73940: fix(bug): AWS failures on OCP 4.19 and earlier for the shared role

heliubj18:

  This PR fixes a bug causing AWS CI test failures when using `--shared-role` with HostedControlPlane on payload versions 4.19 and earlier.

  **Problem:**
  When creating a hosted cluster with `--shared-role` enabled, the ingress cloud-credentials secret was not being properly
  updated/synced for payload versions 4.19 and earlier, due to missing logic in the credential reconciliation loop.

  **Root Cause:**
 On the AWS platform, when multiple components (such as ingress and storage) use a shared role ARN, the
 reconcileCloudCredentialSecrets function in HCCO used a map structure with the ARN as the key. Since map keys must be unique, when IngressARN == StorageARN, only one entry is retained in the map, causing the other component's credentials to be skipped.

  **Solution:**
  The fix ensures that both ingress and storage credential secrets are properly synchronized when using a shared role ARN, regardless of the payload version.

  Fixes OCPBUGS-73940

  - The test failure occurs when HO is the latest version with `--shared-role` enabled, and payload is 4.19 or earlier
  - Currently, `hcp` client does not support `--shared-role`, so this issue only affects CI tests using the `hypershift` client
  - This proactive fix also prevents future issues when hcp adds `--shared-role` support (e.g., rosa hcp)
  - We could verify the fix against the failing AWS CI jobs

  ## Checklist:

### DIFF
--- a/control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
+++ b/control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
@@ -1537,15 +1537,22 @@ func (r *reconciler) reconcileCloudCredentialSecrets(ctx context.Context, hcp *h
 			}
 			return nil
 		}
-		for arn, secret := range map[string]*corev1.Secret{
-			hcp.Spec.Platform.AWS.RolesRef.IngressARN: manifests.AWSIngressCloudCredsSecret(),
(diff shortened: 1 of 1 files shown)


heliubj18:
replaced by https://github.com/openshift/release/pull/73696
/close

PR #7543, opened 2026-01-19
API-1835: Scaffold event dispatcher 

p0lyn0mial:
requires https://github.com/openshift/hypershift/pull/7477

```
// inputResourceDispatcher is a simple dispatcher that applies GVK scoped filters
// and forwards matching events.
//
// Each GVK has its own set of filters. Today these
// may include name/namespace checks, and in the future label selectors.
//
// Longer term, this dispatcher is expected to track which input resources are
// associated with which operator.
```

### DIFF
--- a/control-plane-operator/controllers/openshiftmanager/input_resource_dispatcher.go
+++ b/control-plane-operator/controllers/openshiftmanager/input_resource_dispatcher.go
@@ -1,0 +1,49 @@
+package openshiftmanager
+
+import (
+	"k8s.io/apimachinery/pkg/runtime/schema"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+)
+
+type inputResourceEventFilter func(obj client.Object) bool
+
+// inputResourceDispatcher is a simple dispatcher that applies GVK scoped filters
+// and forwards matching events.
+//
+// Each GVK has its own set of filters. Today these
+// may include name/namespace checks, and in the future label selectors.
+//
+// Longer term, this dispatcher is expected to track which input resources are
+// associated with which operator.
+type inputResourceDispatcher struct {
+	eventsCh chan event.GenericEvent
+	filters  map[schema.GroupVersionKind][]inputResourceEventFilter
+}
+
+func newInputResourceDispatcher(filters map[schema.GroupVersionKind][]inputResourceEventFilter) *inputResourceDispatcher {
+	return &inputResourceDispatcher{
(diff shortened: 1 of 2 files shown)


control-plane-operator/controllers/openshiftmanager/input_resource_dispatcher.go
```diff
@@ -0,0 +1,49 @@
...
+// Longer term, this dispatcher is expected to track which input resources are
+// associated with which operator.
+type inputResourceDispatcher struct {
+	eventsCh chan event.GenericEvent
+	filters  map[schema.GroupVersionKind][]inputResourceEventFilter
+}
+
+func newInputResourceDispatcher(filters map[schema.GroupVersionKind][]inputResourceEventFilter) *inputResourceDispatcher {
+	return &inputResourceDispatcher{
+		eventsCh: make(chan event.GenericEvent, 10),
```
p0lyn0mial: I think I will change it to an unbuffered channel.
(let's start simple)

control-plane-operator/controllers/openshiftmanager/input_resource_dispatcher.go
```diff
@@ -0,0 +1,49 @@
...
+		eventsCh: make(chan event.GenericEvent),
+		filters:  filters,
+	}
+}
+
+func (d *inputResourceDispatcher) Handle(gvk schema.GroupVersionKind, cObj client.Object) {
+	filters := d.filters[gvk]
+	if len(filters) == 0 {
+		d.eventsCh <- event.GenericEvent{Object: cObj}
+		return
```
bertinatto: Why dispatch the event if there are no filters? Is this an oversight?

p0lyn0mial: Filter will be added in https://github.com/openshift/hypershift/pull/7552

p0lyn0mial: ok, so i misunderstood you question.

don't have a good answer here, i can change it to return if there a no filters and we can always adjust the code.

p0lyn0mial: it seems like, the initial list will contain "exact resources" only, xref: https://github.com/openshift/cluster-authentication-operator/blob/master/pkg/cmd/mom/input_resources_command.go#L18

so, it looks like we will always require some filters.

bertinatto (commented):
/lgmt

bryan-cox:
/close

Seems like the PR hasn't been touched in some time. Please feel free to reopen if we need this.

PR #7574, opened 2026-01-23
WIP: add OpenShiftVersion field to OpenshiftEC2NodeClass

maxcao13:
WIP
Currently based upon https://github.com/openshift/hypershift/pull/7478

Introduces `openshiftVersion` field to OpenshiftEC2NodeClass. The field respects version skew between the nodeclass node versions and the HostedControlPlane version.

Fixes https://issues.redhat.com/browse/AUTOSCALE-494

This is what it looks like:
```yaml
apiVersion: karpenter.hypershift.openshift.io/v1beta1
kind: OpenshiftEC2NodeClass
metadata:
  name: example-nodeclass
spec:
  blockDeviceMappings:
  - deviceName: /dev/xvda
    ebs:
      deleteOnTermination: true
      encrypted: true
      iops: 3000
      throughput: 125
      volumeSize: 75Gi
      volumeType: gp3
  openshiftVersion: 4.20.1
  securityGroupSelectorTerms:
  - tags:
      karpenter.sh/discovery: macao-karpenter-421-v5fmf
  subnetSelectorTerms:
  - tags:
      karpenter.sh/discovery: macao-karpenter-421-v5fmf
```
watch nodeclaims and nodes:
```bash
NAME                                             TYPE        CAPACITY   ZONE         NODE
  READY   AGE
nodeclaim.karpenter.sh/ondemand-and-spot-ckppf   t3.xlarge   spot       us-east-1a   ip-10-0-142-248.ec2.internal
  True    18m

NAME                                STATUS   ROLES    AGE   VERSION
node/ip-10-0-142-248.ec2.internal   Ready    worker   14m   v1.33.5
node/ip-10-0-143-25.ec2.internal    Ready    worker   81m   v1.34.1
```

## Checklist:


maxcao13:
Definitely needs some feedback. Like if we go down this route, do we want a validating webhook to constrain which values can be fed into openshiftVersion? And would we use [IsValidReleaseVersion](https://github.com/openshift/hypershift/blob/ef03ac5090edeccf53e61625fd506c0cd1932256/support/supportedversion/version.go#L106) in there as well as versionSkew function?

Also, is there a better way to lookup valid releaseImage image specs from an x.y.z version? Right now, I've copied the format of some of the other functions in the version package, which http get from https://multi.ocp.releases.ci.openshift.org/ which doesn't seem amazing. I don't have enough knowledge on HCP to know if there's some sort of internal cache of valid release images to pull from at startup without relying on the internet.

maxcao13:
Gonna close this for now. We can revisit in the future.

PR #7575, opened 2026-01-23
OCPBUGS-74247: CAPI image overrides aware of registry config

clebs:
The current behavior of the CAPI image overrides is not aware of registry configuration and does not apply it, causing CAPI images to be downloaded from quay.io even if a different registry is configured.

To fix the issue:
- CAPI image overrides now apply registry overrides by using the registry Provider instead of being hardcoded.
- the override version has been pinned to 4.20.10 instead of a hash.
- Moved all the logic to the `support/backwardscompat` package for better isolation and future cleanup.

Fixes [OCPBUGS-74247](https://issues.redhat.com/browse/OCPBUGS-74247)

## Checklist:

### DIFF
--- a/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
+++ b/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
@@ -52,6 +52,7 @@ import (
 	kvinfra "github.com/openshift/hypershift/kubevirtexternalinfra"
 	"github.com/openshift/hypershift/support/api"
 	"github.com/openshift/hypershift/support/azureutil"
+	"github.com/openshift/hypershift/support/backwardcompat"
 	"github.com/openshift/hypershift/support/capabilities"
 	"github.com/openshift/hypershift/support/certs"
 	"github.com/openshift/hypershift/support/config"
@@ -2446,11 +2447,16 @@ func (r *HostedClusterReconciler) reconcileCAPIManager(cpContext controlplanecom
 
 	imageOverride := hcluster.Annotations[hyperv1.ClusterAPIManagerImage]
 
-	// temporary override for 4.21 to unblock CAPI bump to 1.11 which introduces a new API version.
-	// used image from multiArch release payload 4.20.5 (quay.io/openshift-release-dev/ocp-release@sha256:1f2c28ac126453a3b9e83b349822b9f1fb7662973a212f936b90fdc40e06eb58)
-	// TODO(https://issues.redhat.com/browse/CNTRLPLANE-1200): Remove this override once Hypershift installs the CAPI v1beta2 API version
-	if imageOverride == "" && releaseVersion.GTE(semver.MustParse("4.21.0-0")) {
-		imageOverride = "quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:f82ee3586bba6ea0479e3154f959123ef8ebdb2594c4ed9e5899b7ce728a3eb2"
+	if imageOverride == "" {
+		pullSecret, err := hyperutil.GetPullSecretBytes(cpContext, r.Client, hcluster)
+		if err != nil {
+			return err
+		}
+
+		imageOverride, err = backwardcompat.GetBackwardCompatibleCAPIImage(cpContext, pullSecret, r.RegistryProvider.GetReleaseProvider(), releaseVersion, ImageStreamCAPI)
+		if err != nil {
+			return err
+		}
 	}
 
 	capiManager := capimanagerv2.NewComponent(imageOverride)
--- a/support/backwardcompat/backwardcompat.go
+++ b/support/backwardcompat/backwardcompat.go
@@ -1,8 +1,14 @@
 package backwardcompat
 
 import (
+	"context"
+	"fmt"
+
 	"github.com/openshift/hypershift/api/hypershift/v1beta1"
+	"github.com/openshift/hypershift/support/releaseinfo"
 	supportutil "github.com/openshift/hypershift/support/util"
+
+	"github.com/blang/semver"
 )
 
 const ImageStreamImportModeField = "imageStreamImportMode"
@@ -24,4 +30,27 @@ func GetBackwardCompatibleConfigHash(config *v1beta1.ClusterConfiguration) (stri
 	// We need to drop the field when it shows up as empty in the marshaled string to keep backward compatibility.
 	// Implementing this at the marshal operation level might result in undesired impact as we might potentially modify other fields and ordering is not deterministic
 	return supportutil.HashStructWithJSONMapper(config, supportutil.NewOmitFieldIfEmptyJSONMapper(ImageStreamImportModeField))
+}
+
+// GetBackwardCompatibleCAPIImage returns a CAPI image pinned to a CAPI 1.10 compatible version.
+// If the releaseVersion is 4.21 or higher. Otherwise an empty string is returned.
+func GetBackwardCompatibleCAPIImage(ctx context.Context, pullSecret []byte, releaseProvider releaseinfo.Provider, releaseVersion semver.Version, component string) (string, error) {
+	// TODO(https://issues.redhat.com/browse/CNTRLPLANE-1200): Remove this override once Hypershift installs the CAPI v1beta2 API version
+	// temporary override for 4.21 to unblock CAPI bump to 1.11 which introduces a new API version.
+	// The images returned are pinned to version 4.20.10.
+	const (
+		pinnedRelease = "quay.io/openshift-release-dev/ocp-release@sha256:7f183e9b5610a2c9f9aabfd5906b418adfbe659f441b019933426a19bf6a5962"
+		minRelease    = "4.21.0-0"
+	)
+
+	if releaseVersion.GTE(semver.MustParse(minRelease)) {
+		imageOverride, err := supportutil.GetPayloadImageFromRelease(ctx, releaseProvider, pinnedRelease, component, pullSecret)
+		if err != nil {
+			return "", fmt.Errorf("error getting backwards compatible image for %s:%s: %w", component, pinnedRelease, err)
+		}
+
+		return imageOverride, nil
+	}
+
+	return "", nil
 }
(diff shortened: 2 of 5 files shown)


hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -2434,10 +2434,23 @@ func (r *HostedClusterReconciler) reconcileCAPIManager(cpContext controlplanecom
...
+		pullSecret, err := hyperutil.GetPullSecretBytes(cpContext, r.Client, hcluster)
+		if err != nil {
+			return err
+		}
+
+		imageOverride, err = hyperutil.GetPayloadImageFromRelease(cpContext, r.RegistryProvider.GetReleaseProvider(), version, component, pullSecret)
+		if err != nil {
+			return fmt.Errorf("error getting image for %s:%s override: %w", component, version, err)
+		}
 	}
```
devguyio, line 2460: 1. there's a package called `support/backwardcompat` that I once introduce to hold such backward-compatibility fixes that need to be remove in the future, so that we have them in one place and can easily track them.

Can you please move all the image-lookup logic that's introduced due to the pinning to that package? 

Ideally, only the code in the backwardcompat package should be talking about 4.21 temp override etc etc.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -2434,10 +2434,23 @@ func (r *HostedClusterReconciler) reconcileCAPIManager(cpContext controlplanecom
...
 
 	// temporary override for 4.21 to unblock CAPI bump to 1.11 which introduces a new API version.
-	// used image from multiArch release payload 4.20.5 (quay.io/openshift-release-dev/ocp-release@sha256:1f2c28ac126453a3b9e83b349822b9f1fb7662973a212f936b90fdc40e06eb58)
+	// used image from multiArch release payload 4.20.12
 	// TODO(https://issues.redhat.com/browse/CNTRLPLANE-1200): Remove this override once Hypershift installs the CAPI v1beta2 API version
 	if imageOverride == "" && releaseVersion.GTE(semver.MustParse("4.21.0-0")) {
-		imageOverride = "quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:f82ee3586bba6ea0479e3154f959123ef8ebdb2594c4ed9e5899b7ce728a3eb2"
+		const (
+			version   = "4.20.12"
+			component = "ose-cluster-api-container"
```
bryan-cox: Where does this come from? This is not the name I normally see in the payload.

clebs: I took this from the component name label on `oc image info` but it is wrong. Correct name is `cluster-capi-controllers`, which already has a constant here https://github.com/openshift/hypershift/blob/14b8508b8ec7185c9ec80b2b0555555bbf9ad0a1/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go?plain=1#L122

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
clebs: Thanks for the suggestion! I have moved the logic there leaving minimal change on the reconciliation code.

support/backwardcompat/backwardcompat.go
```diff
@@ -25,3 +30,26 @@ func GetBackwardCompatibleConfigHash(config *v1beta1.ClusterConfiguration) (stri
 	// Implementing this at the marshal operation level might result in undesired impact as we might potentially modify other fields and ordering is not deterministic
 	return supportutil.HashStructWithJSONMapper(config, supportutil.NewOmitFieldIfEmptyJSONMapper(ImageStreamImportModeField))
 }
+
+// GetBackwardCompatibleCAPIImageOverride returns a CAPI image pinned to a CAPI 1.10 compatible version.
+// If the releaseVersion is 4.21 or higher. Otherwise an empty string is returned.
+func GetBackwardCompatibleCAPIImageOverride(ctx context.Context, pullSecret []byte, releaseProvider releaseinfo.Provider, releaseVersion semver.Version, component string) (string, error) {
```
bryan-cox: Can we drop the override naming from this function? It's not always getting an override image when the release is greater than 4.21.

support/backwardcompat/backwardcompat.go
```diff
@@ -25,3 +30,26 @@ func GetBackwardCompatibleConfigHash(config *v1beta1.ClusterConfiguration) (stri
...
+	// TODO(https://issues.redhat.com/browse/CNTRLPLANE-1200): Remove this override once Hypershift installs the CAPI v1beta2 API version
+	// temporary override for 4.21 to unblock CAPI bump to 1.11 which introduces a new API version.
+	// Used image from multiArch release payload 4.20.12.
```
bryan-cox: Can we drop this comment or can you help me understand what it is pointing to? I don't see what image its talking about.

support/backwardcompat/backwardcompat.go
```diff
@@ -25,3 +31,26 @@ func GetBackwardCompatibleConfigHash(config *v1beta1.ClusterConfiguration) (stri
 	// Implementing this at the marshal operation level might result in undesired impact as we might potentially modify other fields and ordering is not deterministic
 	return supportutil.HashStructWithJSONMapper(config, supportutil.NewOmitFieldIfEmptyJSONMapper(ImageStreamImportModeField))
 }
+
+// GetBackwardCompatibleCAPIImageOverride returns a CAPI image pinned to a CAPI 1.10 compatible version.
```
bryan-cox: This needs a unit test

support/util/imagemetadata.go
```diff
@@ -486,7 +481,11 @@ func GetRegistryOverrides(ctx context.Context, ref reference.DockerImageReferenc
 }
 
 func GetPayloadImage(ctx context.Context, releaseImageProvider releaseinfo.Provider, hc *hyperv1.HostedCluster, component string, pullSecret []byte) (string, error) {
-	releaseImage, err := releaseImageProvider.Lookup(ctx, HCControlPlaneReleaseImage(hc), pullSecret)
+	return GetPayloadImageFromRelease(ctx, releaseImageProvider, HCControlPlaneReleaseImage(hc), component, pullSecret)
```
bryan-cox, line 484: Why was this change needed?

clebs: Because the current signature has a HC as parameter and extracts the version from there, but we want to be able to use an arbitrary release version instead so that we can pass `4.20.12` on the backwards compatibility function.

support/backwardcompat/backwardcompat.go
clebs: Ok this might need some rephrasing, I want to document that this pins the release payload to 4.20.12 and the given component image will be fetched for that release, not the current release version on the HC.

support/backwardcompat/backwardcompat.go
clebs: Yes, in the scope of this function this is not an override, just an image. If it is then used as an override elsewhere is not relevant here.

devguyio (commented):
LGTM given the comments above will be addressed.

support/util/imagemetadata.go
```diff
@@ -486,7 +481,11 @@ func GetRegistryOverrides(ctx context.Context, ref reference.DockerImageReferenc
...
+}
+
+func GetPayloadImageFromRelease(ctx context.Context, releaseImageProvider releaseinfo.Provider, release string, component string, pullSecret []byte) (string, error) {
```
enxebre, line 487: why this change?

clebs: https://github.com/openshift/hypershift/pull/7575#discussion_r2721485468

enxebre:
/approve
change looks good pending addressing comments

support/backwardcompat/backwardcompat.go
clebs: Function has been renamed!

support/backwardcompat/backwardcompat.go
clebs: I made this line more specific to what is actually happening inside. I am also fine removing it since the constant inside makes it clear too.

support/backwardcompat/backwardcompat.go
clebs: Unit test added!

devguyio:
Should we also do the rest of the platforms in this PR? I think yes

/uncc @sjenning

support/backwardcompat/backwardcompat.go
```diff
@@ -25,3 +31,26 @@ func GetBackwardCompatibleConfigHash(config *v1beta1.ClusterConfiguration) (stri
...
 	return supportutil.HashStructWithJSONMapper(config, supportutil.NewOmitFieldIfEmptyJSONMapper(ImageStreamImportModeField))
 }
+
+// GetBackwardCompatibleCAPIImage returns a CAPI image pinned to a CAPI 1.10 compatible version.
+// If the releaseVersion is 4.21 or higher. Otherwise an empty string is returned.
+func GetBackwardCompatibleCAPIImage(ctx context.Context, pullSecret []byte, releaseProvider releaseinfo.Provider, releaseVersion semver.Version, component string) (string, error) {
+	// TODO(https://issues.redhat.com/browse/CNTRLPLANE-1200): Remove this override once Hypershift installs the CAPI v1beta2 API version
+	// temporary override for 4.21 to unblock CAPI bump to 1.11 which introduces a new API version.
+	// The images returned are pinned to version 4.20.10.
+	const (
```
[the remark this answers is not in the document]
clebs: The hash corresponds exactly to version 4.20.10, the comment documents that so that there is no need to check what version the hash corresponds to.

mgencur:
@clebs Do we plan to ship this in 4.21? If yes, we will probably need to document how to use the product in disconnected environment. It is using an image from an older release so users will have to call `oc adm release mirror` with this older release too so as to get the image mirrored/available.  I had to use a workaround with the previous solution and this will need a similar one, see my attempts in https://github.com/openshift/release/pull/73933/changes#diff-e12a43ad1798ff5e7b7d10a8439c3470f48ffe5a281f96afbf565adf21ef71a3R165 I'm mirroring just a single image but for a user it might be easier to just mirror all image from the release.
We're already hitting https://issues.redhat.com/browse/OCPBUGS-74263 in CI. This new solution is better but the image is still not available in disconnected env.

clebs:
@mgencur this override is already on 4.21, this PR fixes the issue where it is not reflecting registry overrides.
And yes, 4.20 images need to be mirrored and it needs to be documented.

I did align with ACM and ARO to make sure they were aware of this and @lahinson took up the item to document this requirement.

PR #7582, opened 2026-01-24
AUTOSCALE-492: karpenter-ignition controller

maxcao13:

This commit adds a controller to the karpenter-operator. The controller watches OpenshiftEC2NodeClass resources in the guest cluster and generates ignition userData + token secrets. These secrets are then used by the karpenter-operator container to pass to the EC2NodeClass, or any other future nodeclasses.

Overall, this allows us to generate the necessary secrets on a per karpenter nodeclass basis.

Note that the control plane and karpenter data plane nodes are still coupled in this commit.

Fixes [AUTOSCALE-492](https://issues.redhat.com/browse/AUTOSCALE-492)
Supersedes: https://github.com/openshift/hypershift/pull/7478

In this implementation I'm introducing a new controller in the control-plane-operator binary, and putting the code in the karpenter-operator controllers. When we eventually refactor the operator business logic out, I expect the karpenter-ignition controller to stay in HCP somewhere.

Also, similar to https://github.com/openshift/hypershift/pull/, I am transforming a HCP into a barebones in-memory hostedcluster object in order to do configgenerator operations and lookups in the sidecar controller.

## Checklist:


karpenter-operator/controllers/karpenterignition/karpenterignition_controller.go
```diff
@@ -0,0 +1,329 @@
+package karpenterignition
+
```
enxebre, line 2: this file seems duplicated in the parent folder?

karpenter-operator/controllers/karpenterignition/controller.go
```diff
@@ -0,0 +1,329 @@
...
+
+	np := r.createInMemoryNodePool(hcp, openshiftEC2NodeClass)
+
+	// TODO(maxcao13): When openshiftVersion field is added to OpenshiftEC2NodeClassSpec, use it here
+	// For now, use the control plane's release image, which means control plane and data plane is still coupled.
+	pullSpec := hcp.Spec.ReleaseImage
+	np.Spec.Release.Image = pullSpec
+
+	controlPlaneNamespace := hostedCluster.Namespace
+	cg, err := r.buildConfigGenerator(ctx, hostedCluster, np, controlPlaneNamespace)
```
enxebre: why is this passing the hc namespace as the cp namespace?

karpenter-operator/controllers/karpenterignition/controller.go
```diff
@@ -0,0 +1,329 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to generate HAProxy raw config: %w", err)
+	}
+
+	return nodepool.NewConfigGenerator(ctx, r.ManagementClient, hostedCluster, np, releaseImage, haproxyRawConfig, controlPlaneNamespace)
+}
+
+// hostedClusterFromHCP creates a barebones in-memory HostedCluster from a HostedControlPlane.
+// NOTE: The Namespace field is set to hcp.Namespace (the HCP namespace) rather than the original HC namespace.
+// This ensures secret lookups work correctly since the operator is only allowed to read secrets in the HCP namespace.
```
enxebre: can we clarify which secrets lookups is this refering to?

karpenter-operator/controllers/karpenterignition/karpenterignition_controller.go
```diff
@@ -0,0 +1,329 @@
...
+	if err := r.ManagementClient.List(ctx, hcpList, client.InNamespace(r.Namespace)); err != nil {
+		return nil, err
+	}
+	if len(hcpList.Items) == 0 {
+		return nil, apierrors.NewNotFound(hyperv1.SchemeGroupVersion.WithResource("hostedcontrolplanes").GroupResource(), "")
+	}
+	return &hcpList.Items[0], nil
+}
+
+func (r *KarpenterIgnitionReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
enxebre, line 75: can this controller have some unit tests?

enxebre:
@coderabbit review

karpenter-operator/controllers/karpenterignition/karpenterignition_controller.go
maxcao13: Oh yeah, whoops. An artifact of when I moved the code as it's own controller instead of a sidecar I think.

karpenter-operator/controllers/karpenterignition/controller.go
maxcao13: Technically, they are the same because I built the hosted cluster using the hcp and passed the namespace over, but I'll change this to hcp.Namespace to make it make more sense.

EDIT: In fact, we don't really need the extra argument at all, we can just use hc.Namespace in the function, but I'll keep it to be explicit, unless there are opinions about this.

karpenter-operator/controllers/karpenterignition/karpenterignition_controller.go
```diff
@@ -0,0 +1,322 @@
...
+}
+
+func (r *KarpenterIgnitionReconciler) createInMemoryNodePool(
+	hcp *hyperv1.HostedControlPlane,
+	openshiftEC2NodeClass *hyperkarpenterv1.OpenshiftEC2NodeClass,
+) *hyperv1.NodePool {
+	return &hyperv1.NodePool{
+		ObjectMeta: metav1.ObjectMeta{
+			// The nodepool name must end with "karpenter" so the resulting secrets don't get cleaned up by the secret janitor
+			Name:        fmt.Sprintf("%s-%s", openshiftEC2NodeClass.Name, hyperkarpenterv1.KarpenterNodePool),
```
muraee: this should be extracted into a function an reused in `getUserData()` https://github.com/openshift/hypershift/pull/7582/changes#diff-bf586010fe886072d1c63888abf4ff11c99c738214ffb9983e183a5202935477R402

maxcao13: Done, thanks!

I also changed the logic for ignoring karpenter secrets that have an non-existent nodepool from secret-janitor by using the new label I added instead of the name prefix "karpenter". Hopefully that's not contentious :-)

PR #7583, opened 2026-01-24
AUTOSCALE-535: default block device mappings

maxcao13:

Default block device mappings if the user doesn't set them in OpenshiftEC2NodeClass.

We've seen a lot of flakes in the e2e-test-techpreview and e2e-aws-autonode prow jobs where what happens is that nodes churn because of DiskPressure.

I've also seen it in manual smoketesting as well, and to get around this I've created a custom nodeclass with more block device root storage.

Fixes 

## Checklist:

### DIFF
--- a/karpenter-operator/controllers/nodeclass/ec2_nodeclass_controller.go
+++ b/karpenter-operator/controllers/nodeclass/ec2_nodeclass_controller.go
@@ -22,6 +22,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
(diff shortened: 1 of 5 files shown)


karpenter-operator/controllers/nodeclass/ec2_nodeclass_controller.go
```diff
@@ -281,6 +282,20 @@ func reconcileEC2NodeClass(ec2NodeClass *awskarpenterv1.EC2NodeClass, openshiftE
 	}
 	ec2NodeClass.Spec.SecurityGroupSelectorTerms = securityGroupSelectorTerms
 
+	// Set default BlockDeviceMappings if not specified
```
enxebre: this doesn't honour the current consumer facing API doc 
```
	// BlockDeviceMappings to be applied to provisioned nodes.
	// +kubebuilder:validation:XValidation:message="must have only one blockDeviceMappings with rootVolume",rule="self.filter(x, has(x.rootVolume)?x.rootVolume==true:false).size() <= 1"
	// +kubebuilder:validation:MaxItems:=50
	// +optional
	BlockDeviceMappings []*BlockDeviceMapping `json:"blockDeviceMappings,omitempty"`
```

Please update that accordingly to clarify that the backend would preserve the ability to set a sane default subject to change if none provided

karpenter-operator/controllers/karpenter/karpenter_controller.go
```diff
@@ -303,6 +304,16 @@ func (r *Reconciler) reconcileOpenshiftEC2NodeClassDefault(ctx context.Context,
 					},
 				},
 			},
+			BlockDeviceMappings: []*hyperkarpenterv1.BlockDeviceMapping{
+				{
+					DeviceName: ptr.To("/dev/xvda"),
+					EBS: &hyperkarpenterv1.BlockDevice{
+						VolumeSize: ptr.To(resource.MustParse("75Gi")),
```
enxebre: any rationale we can add in a comment to pick this number?

maxcao13: I just added a comment that the CLI has code that picks 120 for default hcp nodepools. Seems reasonable to me: https://github.com/openshift/hypershift/blob/8be1d9c6f8f79106444e48f2b7d0069b942ba0d7/cmd/nodepool/aws/create.go#L27-L32

enxebre:
My preference would be to default only the service owned default, for the rest let the user make a deliberate choice

/approve
/lgtm

maxcao13:
/hold

some ci issue.

PR #7587, opened 2026-01-26
CNTRLPLANE-418: fix shared ingress AllowedCIDRs to use fc_src for PROXY protocol

enxebre:
## Summary
- Change the HAProxy AllowedCIDRs ACL to use `fc_src` instead of `src` for checking allowed CIDRs
- When PROXY protocol is enabled (`accept-proxy`), `src` results in the internal IP of the connection rather than the unique outbound IP
- Using `fc_src` results in the unique outbound IP (e.g., Azure LB) which is what should be checked against the allowed CIDRs

## Problem
When traffic flows through a proxy chain (e.g., data plane -> shared ingress), the `src` variable in HAProxy results in the internal IP of the connection (e.g., `169.254.0.1`, `172.20.0.1`, or pod IPs like `10.132.0.x`), not the unique outbound IP.


enxebre:
/area hypershift-operator
/hold

to qe pre-merge

PR #7590, opened 2026-01-27
CNTRLPLANE-2207: Upgrade to CAPI 1.11

clebs:

Bumps hypershift to use CAPI `v1.11` including the following tasks:
- [x] Update CAPI and all providers to a `v1.11` compatible version in `go.mod`.
- [x] Removes @csrwng's fork containing a temporary fix.
- [x] Update controller-gen goal in Makefile.
- [x] Update install assets: CAPI CRDs.
- [x] Patch CAPI CRDs to use `v1beta1` as storage version.
- [x] Adds conversion webhooks for `v1beta1` <-> `v1beta2`.
- [x] Removes the temporary CAPI image overrides (#7575).
- [x] Check in updated vendored dependencies.

Fixes [CNTRLPLANE-2207](https://issues.redhat.com/browse/CNTRLPLANE-2207)

## Checklist:

## Release Notes

* **Chores**
  * Updated Go module dependencies to newer versions for improved stability and security.
  * Adjusted build configuration for CRD generation to optimize the build process.
  * Updated linter configuration to address deprecated package deprecations.


cmd/install/install.go
```diff
@@ -781,28 +781,43 @@ func setupCRDs(ctx context.Context, client crclient.Client, opts Options, operat
...
-									Namespace: operatorNamespace.Name,
-									Name:      operatorService.Name,
-									Port:      ptr.To[int32](443),
-									Path:      ptr.To("/convert"),
-								},
+					needsConversion = true
+					conversionReviewVersions = []string{"v1beta1", "v1alpha1"}
+				} else if override, ok := assets.CAPICRDOverrides[crd.Name]; ok && override.NeedsConversion {
+					needsConversion = true
+					conversionReviewVersions = []string{"v1"}
```
csrwng: Looks like this line should be ~~swapped with line 794~~

csrwng: Actually should be v1beta1 and v1beta2, no?

clebs: That's right. I guess it works now because v1 is a catch all.

LiangquanLi930:
@coderabbitai can you help analyze why this job failed?

LiangquanLi930:
@clebs @enxebre, Once this PR is merged, I will continue working on upgrading the HO clients to v1beta2.

api/go.mod
```diff
@@ -23,37 +23,45 @@ require (
...
+	github.com/go-openapi/swag/conv v0.25.4 // indirect
+	github.com/go-openapi/swag/fileutils v0.25.4 // indirect
+	github.com/go-openapi/swag/jsonname v0.25.4 // indirect
+	github.com/go-openapi/swag/jsonutils v0.25.4 // indirect
+	github.com/go-openapi/swag/loading v0.25.4 // indirect
+	github.com/go-openapi/swag/mangling v0.25.4 // indirect
+	github.com/go-openapi/swag/netutils v0.25.4 // indirect
+	github.com/go-openapi/swag/stringutils v0.25.4 // indirect
+	github.com/go-openapi/swag/typeutils v0.25.4 // indirect
+	github.com/go-openapi/swag/yamlutils v0.25.4 // indirect
```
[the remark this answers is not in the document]
clebs: This is likely a false positive, after removing the extra `github.com/go-openapi/swag/*` packages, running `go mod tidy` adds them back in.

api/go.mod
```diff
@@ -74,22 +82,21 @@ require (
...
 	k8s.io/apiextensions-apiserver v0.34.3 // indirect
 	k8s.io/client-go v0.34.3 // indirect
 	k8s.io/cloud-provider v0.34.1 // indirect
 	k8s.io/component-base v0.34.3 // indirect
-	k8s.io/component-helpers v0.34.2 // indirect
+	k8s.io/component-helpers v0.34.3 // indirect
 	k8s.io/csi-translation-lib v0.34.1 // indirect
 	k8s.io/klog/v2 v2.130.1 // indirect
-	k8s.io/kube-openapi v0.0.0-20250710124328-f3f2b991d03b // indirect
+	k8s.io/kube-openapi v0.0.0-20251125145642-4e65d59e963e // indirect
```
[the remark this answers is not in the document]
clebs: Caused by rebase, fixed now.

everettraven:
Approving the API changes as they are gomod/vendor only updates.

/approve

LiangquanLi930:
As we discussed, I'll continue working on `Upgrade HO clients to use v1beta2` once this PR is merged.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -2563,18 +2562,6 @@ func (r *HostedClusterReconciler) reconcileCAPIManager(cpContext controlplanecom
 
 	imageOverride := hcluster.Annotations[hyperv1.ClusterAPIManagerImage]
 
-	if imageOverride == "" {
-		pullSecret, err := hyperutil.GetPullSecretBytes(cpContext, r.Client, hcluster)
-		if err != nil {
-			return err
-		}
-
-		imageOverride, err = backwardcompat.GetBackwardCompatibleCAPIImage(cpContext, pullSecret, r.RegistryProvider.GetReleaseProvider(), releaseVersion, ImageStreamCAPI)
```
enxebre, line 2544: can this func declaration be removed: GetBackwardCompatibleCAPIImage?

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -3000,6 +3000,11 @@ func reconcileCAPIManagerClusterRoleBinding(binding *rbacv1.ClusterRoleBinding,
 			Name:      sa.Name,
 			Namespace: sa.Namespace,
 		},
+		{
```
enxebre, line 2979: can you point me to where the provider need access to this?

Should we instead create a dedicated targeted clusterRole with only what's needed here hypershift/control-plane-operator/controllers/hostedcontrolplane/v2/assets/capi-provider?

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
clebs: This can be removed yes.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
clebs: It' s been a while but IIRC I added this so the provider would have access to CRDs to be able to do conversions.

I will look into creating a separate role + binding + SA for this.

hypershift-operator/controllers/hostedcluster/hostedcluster_webhook.go
```diff
@@ -150,6 +153,30 @@ func SetupWebhookWithManager(mgr ctrl.Manager, imageMetaDataProvider *hyperutil.
 	if err != nil {
 		return fmt.Errorf("unable to register hostedcontrolplane webhook: %w", err)
 	}
+
+	// Initialize CAPI v1beta1 conversion support.
+	// CAPI v1beta1 types need an apiVersionGetter to convert object references
+	// from v1beta2 (Hub) ContractVersionedObjectReference back to v1beta1 corev1.ObjectReference.
+	// The getter resolves GroupKind to the preferred API version string using the scheme.
```
enxebre: which object references to core objects might we have in capi resouces?

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
clebs: I have given a second look to this and the Role we are binding `capi-provider` to (`hypershift-cluster-api`) is already minimal. It only grants get, watch and list of CRDs and is only used by `capi-manager` SA and `capi-provider` SA.

I am not sure we would benefit of having a dedicated role here.

hypershift-operator/controllers/hostedcluster/hostedcluster_webhook.go
clebs: As discussed, this block will be removed because it actually always falls back to the default and therefore has no effect.

hypershift-operator/controllers/nodepool/nodepool_controller.go
```diff
@@ -774,12 +776,54 @@ func defaultNodePoolGCPImage(specifiedArch string, releaseImage *releaseinfo.Rel
...
+		} `json:"spec"`
+		Status struct {
+			ObservedGeneration int64  `json:"observedGeneration"`
+			Replicas           *int32 `json:"replicas"`
+			AvailableReplicas  *int32 `json:"availableReplicas"`
+			UpToDateReplicas   *int32 `json:"upToDateReplicas"`
+		} `json:"status"`
+	}
+	if err := json.Unmarshal([]byte(raw), &v1beta2MD); err != nil {
+		return true
```
jparrill, line 819: nit: `json.Unmarshal` failure here returns `true` silently — a corrupt annotation would make the controller consider a rolling update complete prematurely. Consider at least logging a warning before the fallback.

hypershift-operator/controllers/hostedcluster/hostedcluster_webhook.go
```diff
@@ -150,6 +153,30 @@ func SetupWebhookWithManager(mgr ctrl.Manager, imageMetaDataProvider *hyperutil.
...
+	// Since we use v1beta1 as the storage version for CAPI CRDs, we prefer v1beta1 when available.
+	capiv1beta1.SetAPIVersionGetter(func(gk schema.GroupKind) (string, error) {
```
jparrill, line 161: You mentioned this block always falls back to the default. If confirmed, it's dead code in a critical webhook path — worth removing.

hypershift-operator/controllers/nodepool/nodepool_controller.go
```diff
@@ -774,12 +776,54 @@ func defaultNodePoolGCPImage(specifiedArch string, releaseImage *releaseinfo.Rel
...
+		return false
+	}
+	return machineDeploymentCompleteFromConversionData(deployment)
+}
+
+// machineDeploymentCompleteFromConversionData parses the v1beta2 conversion-data annotation
+// and verifies that the native v1beta2 status also indicates completion. If the annotation
+// is absent (e.g. CAPI < v1.11), returns true to preserve backwards compatibility.
+func machineDeploymentCompleteFromConversionData(deployment *capiv1.MachineDeployment) bool {
+	raw, ok := deployment.Annotations["cluster.x-k8s.io/conversion-data"]
```
jparrill: nit: CAPI exports `conversion.DataAnnotation` in `sigs.k8s.io/cluster-api/util/conversion` — using it instead of the raw string avoids typo-induced bugs and tracks upstream renames.

jparrill (commented):

jparrill:
nit: `GetBackwardCompatibleCAPIImage` still exists in `support/backwardcompat/` but all call sites are removed in this PR — can be cleaned up.

clebs:
> nit: `GetBackwardCompatibleCAPIImage` still exists in `support/backwardcompat/` but all call sites are removed in this PR — can be cleaned up.

@jparrill yes this will be removed, agreed upon above on @enxebre's comment.

hypershift-operator/controllers/nodepool/nodepool_controller.go
clebs: Good point! Will use the constant instead.

hypershift-operator/controllers/hostedcluster/hostedcluster_webhook.go
clebs: yes, we discussed this last week and I will remove it.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
enxebre: It might be minimal atm, if you need to open it up tomorrow, then you transparently relax permissions for the provider as well.

>It' s been a while but IIRC I added this so the provider would have access to CRDs to be able to do conversions.

At minimum can we please prove this and // document this change with that reasoning and evidence?

hypershift-operator/controllers/nodepool/nodepool_controller.go
clebs: I think this is the intended behavior, as the godoc above states: if the annotation is missing it means there was no conversion and we assume we are on an older CAPI, in which case we return true to keep backwards compatibility.

But it is true that if the key is present in the map and we were not able to unmarshal something is not right, we will still return true as fallback but should log this.

hypershift-operator/controllers/hostedcluster/hostedcluster_webhook.go
clebs: After removing the block conversion was failing. Reintroduced it simplified and better clarified its need on the comment.

hypershift-operator/controllers/nodepool/nodepool_controller.go
clebs: Added a log entry when unmarshal fails and a unit test for this branch.

enxebre:
/close 
in favour of https://github.com/openshift/hypershift/pull/8594

PR #7594, opened 2026-01-27
OCPBUGS-72559: feat(globalps): add support for preserving registries from existing config.json

joshbranham:
Add a new feature that allows users to preserve specific registry credentials from the existing /var/lib/kubelet/config.json when the sync-global-pullsecret process runs. This is controlled via a ConfigMap named `globalps-preserve-registries` in the kube-system namespace.

Users can create the ConfigMap with a list of registries to preserve:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: globalps-preserve-registries
  namespace: kube-system
data:
  registries: |
    my-legacy-registry.example.com
    another-registry.io/v2
```

Merge priority (lowest to highest):
1. User-provided additional-pull-secret (base layer)
2. Preserved registries (from existing config.json)
3. Original HCP pull secret (always wins)

Changes:
- Add PreserveRegistriesConfigMap manifest function
- Add ConfigMap existence check and optional volume/mount in DaemonSet
- Add preserve-registries reading, parsing, and merge logic in syncer
- Add unit tests for all new functionality
- Add e2e tests for preserve-registries feature

Fixes: OCPBUGS-72559

## Checklist:

### DIFF
--- a/control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
+++ b/control-plane-operator/hostedclusterconfigoperator/controllers/globalps/globalps.go
@@ -132,7 +132,7 @@ func (r *Reconciler) reconcileGlobalPullSecret(ctx context.Context) error {
 		preserveRegistriesName := ""
 		preserveExists, err := preserveRegistriesConfigMapExists(ctx, r.kubeSystemSecretClient)
 		if err != nil {
-			return fmt.Errorf("failed to check if preserve-registries ConfigMap exists: %w", err)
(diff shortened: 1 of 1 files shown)


joshbranham:
After a bit more thinking, this feature won't work as designed. Currently, users can pass additional mirrors in ROSA HCP clusters, and those mirrors can be for repos that are mirrored into ECR. Therefore, they rely on the same auth flow that we have now to automatically pull those. By using a ConfigMap, we would need to merge user defined alongside service provider defined ECR mirrors and their auth, which can get messy. I need to keep thinking on this one.

PR #7600, opened 2026-01-28
GCP-176: feat(gcp): add WIF credentials support for control plane op…

ckandag:

- Add ControlPlaneOperatorCredsSecret for CPO WIF credentials
- Update ReconcileCredentials to create secrets for both NodePool and ControlPlane
- Update buildGCPWorkloadIdentityCredentials to accept service account email
- Add CPO deployment configuration for GCP WIF

Signed-off-by: Chaitanya Kandagatla <ckandaga@redhat.com>

https://issues.redhat.com/browse/GCP-176

## Checklist:

### DIFF
--- a/hypershift-operator/controllers/hostedcluster/internal/platform/gcp/gcp_test.go
+++ b/hypershift-operator/controllers/hostedcluster/internal/platform/gcp/gcp_test.go
@@ -20,6 +20,12 @@ import (
 	"github.com/blang/semver"
 )
 
+const (
+	// Test service account emails used across GCP tests
+	testNodePoolGSA     = "test-capg-sa@test-project.iam.gserviceaccount.com"
+	testControlPlaneGSA = "test-control-plane-sa@test-project.iam.gserviceaccount.com"
+)
+
 // testCreateOrUpdate is a test helper that implements createOrUpdate functionality
 // for testing without requiring the actual upsert package dependencies.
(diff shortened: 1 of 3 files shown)


hypershift-operator/controllers/hostedcluster/internal/platform/gcp/gcp_test.go
```diff
@@ -381,7 +381,7 @@ func TestBuildGCPWorkloadIdentityCredentials(t *testing.T) {
 		},
 	}
 
-	credentials, err := buildGCPWorkloadIdentityCredentials(wif)
+	credentials, err := buildGCPWorkloadIdentityCredentials(wif, wif.ServiceAccountsEmails.NodePool)
```
cblecker, line 392: Are we only testing one of the operators?

ckandag: in this context, it doesnt matter which email/sa (i could just  say xxx@..),  
the testcase is testing the support method that builds credentials json which is the same for all.

I added an additional testcase for checking the existing of controlplane op

cblecker: That makes sense. Should we have a comment explaining that this is just an example one, and don't need to individually check each one?

ckandag: ok, added a comment.
Also cleaned up the sa-email strings and added the same comment in other similar places

PR #7603, opened 2026-01-29
GCP-198: add PSC endpoint controller for customer-side management

cristianoveiga:

  Adds a new PSC endpoint controller in the control-plane-operator that manages Private Service Connect (PSC) endpoints in the customer's GCP project. This is the customer-side counterpart to the management-side PSC controller that creates Service Attachments.

  **Key features:**
  - Reserves static internal IP addresses for PSC endpoints in the customer's subnet
  - Creates ForwardingRules targeting the management-side Service Attachment
  - Handles cleanup of GCP resources on deletion with proper finalizers
  - Waits for WIF token availability before adding finalizers (prevents cleanup failures)
  - Implements 30-second timeout on all GCP API calls to prevent hung reconcilers
  - Propagates `GCPEndpointAvailable` and `GCPServiceAttachmentAvailable` conditions to HostedCluster status

  Fixes https://issues.redhat.com/browse/GCP-198

  - DNS implementation is intentionally excluded and will be handled separately in [GCP-199](https://issues.redhat.com/browse/GCP-199)
  - The controller uses WIF (Workload Identity Federation) credentials to authenticate with the customer's GCP project
  - Timeout logic matches the pattern used in the management-side PSC controller (`hypershift-operator/controllers/platform/gcp/privateserviceconnect_controller.go`)
  - Tested end-to-end in the `gcp-hcp-pilot` branch

  ## Checklist:

### DIFF
--- a/control-plane-operator/controllers/gcpprivateserviceconnect/psc_endpoint_controller.go
+++ b/control-plane-operator/controllers/gcpprivateserviceconnect/psc_endpoint_controller.go
@@ -231,8 +231,9 @@ func (r *GCPPrivateServiceConnectReconciler) Reconcile(ctx context.Context, req 
 
 // isServiceAttachmentReady checks if the management-side Service Attachment is ready
 func (r *GCPPrivateServiceConnectReconciler) isServiceAttachmentReady(gcpPSC *hyperv1.GCPPrivateServiceConnect) bool {
-	// Check if management-side has populated ServiceAttachmentURI
-	if gcpPSC.Status.ServiceAttachmentURI == "" {
+	// Check if management-side has populated both ServiceAttachmentURI and ServiceAttachmentName.
+	// Both are required: URI for the PSC endpoint target, Name for constructing GCP resource names.
+	if gcpPSC.Status.ServiceAttachmentURI == "" || gcpPSC.Status.ServiceAttachmentName == "" {
 		return false
 	}
 
--- a/control-plane-operator/controllers/gcpprivateserviceconnect/psc_endpoint_controller_test.go
+++ b/control-plane-operator/controllers/gcpprivateserviceconnect/psc_endpoint_controller_test.go
@@ -160,25 +160,38 @@ func TestIsServiceAttachmentReady(t *testing.T) {
 			name: "When ServiceAttachmentURI is empty it should return false",
 			gcpPSC: &hyperv1.GCPPrivateServiceConnect{
 				Status: hyperv1.GCPPrivateServiceConnectStatus{
-					ServiceAttachmentURI: "",
+					ServiceAttachmentURI:  "",
+					ServiceAttachmentName: "test-sa",
 				},
 			},
 			expected: false,
 		},
 		{
-			name: "When ServiceAttachmentURI exists but condition is missing it should return false",
+			name: "When ServiceAttachmentName is empty it should return false",
 			gcpPSC: &hyperv1.GCPPrivateServiceConnect{
 				Status: hyperv1.GCPPrivateServiceConnectStatus{
-					ServiceAttachmentURI: "projects/mgmt-project/regions/us-central1/serviceAttachments/test-sa",
+					ServiceAttachmentURI:  "projects/mgmt-project/regions/us-central1/serviceAttachments/test-sa",
+					ServiceAttachmentName: "",
 				},
 			},
 			expected: false,
 		},
 		{
-			name: "When ServiceAttachmentURI exists but condition is False it should return false",
+			name: "When both URI and Name exist but condition is missing it should return false",
 			gcpPSC: &hyperv1.GCPPrivateServiceConnect{
 				Status: hyperv1.GCPPrivateServiceConnectStatus{
-					ServiceAttachmentURI: "projects/mgmt-project/regions/us-central1/serviceAttachments/test-sa",
+					ServiceAttachmentURI:  "projects/mgmt-project/regions/us-central1/serviceAttachments/test-sa",
+					ServiceAttachmentName: "test-sa",
+				},
+			},
+			expected: false,
+		},
+		{
+			name: "When both URI and Name exist but condition is False it should return false",
+			gcpPSC: &hyperv1.GCPPrivateServiceConnect{
+				Status: hyperv1.GCPPrivateServiceConnectStatus{
+					ServiceAttachmentURI:  "projects/mgmt-project/regions/us-central1/serviceAttachments/test-sa",
+					ServiceAttachmentName: "test-sa",
 					Conditions: []metav1.Condition{
 						{
 							Type:   string(hyperv1.GCPServiceAttachmentAvailable),
@@ -190,10 +203,11 @@ func TestIsServiceAttachmentReady(t *testing.T) {
 			expected: false,
 		},
 		{
-			name: "When ServiceAttachmentURI exists and condition is True it should return true",
+			name: "When both URI and Name exist and condition is True it should return true",
 			gcpPSC: &hyperv1.GCPPrivateServiceConnect{
 				Status: hyperv1.GCPPrivateServiceConnectStatus{
-					ServiceAttachmentURI: "projects/mgmt-project/regions/us-central1/serviceAttachments/test-sa",
+					ServiceAttachmentURI:  "projects/mgmt-project/regions/us-central1/serviceAttachments/test-sa",
+					ServiceAttachmentName: "test-sa",
 					Conditions: []metav1.Condition{
 						{
 							Type:   string(hyperv1.GCPServiceAttachmentAvailable),


cristianoveiga:
@coderabbitai review

control-plane-operator/controllers/gcpprivateserviceconnect/psc_endpoint_controller.go
```diff
@@ -0,0 +1,581 @@
...
+
+		// Always remove finalizer regardless of cleanup success (following AWS pattern)
+		if controllerutil.ContainsFinalizer(gcpPSC, pscEndpointFinalizer) {
+			controllerutil.RemoveFinalizer(gcpPSC, pscEndpointFinalizer)
+			if err := r.Update(ctx, gcpPSC); err != nil {
+				return ctrl.Result{}, fmt.Errorf("failed to remove finalizer: %w", err)
+			}
+		}
+		return ctrl.Result{}, nil
+	}
```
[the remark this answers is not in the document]
cristianoveiga: Added the suggestion.

control-plane-operator/controllers/gcpprivateserviceconnect/psc_endpoint_controller.go
```diff
@@ -0,0 +1,581 @@
...
+	ipName := r.constructIPAddressName(gcpPSC)
+	endpoint := &compute.ForwardingRule{
+		Name:        endpointName,
+		Description: fmt.Sprintf("PSC endpoint for HyperShift cluster %s", gcpPSC.Name),
+		Network:     r.constructNetworkURL(hcp.Spec.Platform.GCP.NetworkConfig.Network.Name, customerProject),
+		Subnetwork:  r.constructSubnetURL(hcp.Spec.Platform.GCP.NetworkConfig.PrivateServiceConnectSubnet.Name, customerProject, region),
+		Target:      gcpPSC.Status.ServiceAttachmentURI,                     // From management-side
+		IPAddress:   r.constructAddressURL(ipName, customerProject, region), // Reserved IP resource URL
+		// LoadBalancingScheme not set for PSC endpoints - it's implicit and setting it causes API errors
+	}
```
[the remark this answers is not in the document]
cristianoveiga: They should be validated elsewhere - prior to reaching the controller.

control-plane-operator/controllers/gcpprivateserviceconnect/psc_endpoint_controller.go
```diff
@@ -0,0 +1,581 @@
...
+		}
+		// IP was deleted, need to allocate new one
+		log.Info("Previously allocated IP no longer exists, allocating new one")
+	}
+
+	pscSubnet := hcp.Spec.Platform.GCP.NetworkConfig.PrivateServiceConnectSubnet.Name
+	if pscSubnet == "" {
+		return ctrl.Result{}, fmt.Errorf("PrivateServiceConnectSubnet not specified in HostedControlPlane")
+	}
+
```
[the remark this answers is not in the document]
cristianoveiga: Similar to the another comment: They should be validated elsewhere - prior to reaching the controller.

control-plane-operator/controllers/gcpprivateserviceconnect/psc_endpoint_controller.go
```diff
@@ -0,0 +1,592 @@
...
+	return InitCustomerGCPClient(ctx, b.customerProject, b.controlPlaneOperatorGSA)
+}
+
+// GCPPrivateServiceConnectReconciler manages PSC endpoints in customer projects
+type GCPPrivateServiceConnectReconciler struct {
+	client.Client
+	upsert.CreateOrUpdateProvider
+	gcpClientBuilder gcpClientBuilder
+}
+
```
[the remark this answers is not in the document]
cristianoveiga: The CPO runs per-HCP namespace with a single GCPPrivateServiceConnect CR, so there's no concurrent reconciliation across different clusters. Additionally, the GCP project and region are immutable fields in the HCP spec, making the caching appropriate.

  This pattern follows the existing AWS implementation in control-plane-operator/controllers/awsprivatelink/awsprivatelink_controller.go (lines 193-310), which uses the same reconciler-level clientBuilder with an initialized guard.

  The second commit adds re-initialization during deletion to handle controller restarts, which is actually an improvement over the AWS pattern.

control-plane-operator/controllers/gcpprivateserviceconnect/psc_endpoint_controller.go
```diff
@@ -0,0 +1,592 @@
...
+	customerProject         string
+	region                  string
+	controlPlaneOperatorGSA string
+}
+
+func (b *gcpClientBuilder) initializeWithHCP(hcp *hyperv1.HostedControlPlane) {
+	if !b.initialized {
+		b.setFromHCP(hcp)
+		b.initialized = true
+	}
```
cblecker, line 76: **Race condition risk with concurrent reconciles**

The `gcpClientBuilder` struct is accessed by multiple goroutines (`MaxConcurrentReconciles: 10`) but lacks mutex protection.

The AWS PrivateLink controller's equivalent `clientBuilder` uses a `sync.Mutex`:

```go
// control-plane-operator/controllers/awsprivatelink/awsprivatelink_controller.go:199-205
type clientBuilder struct {
    mu                             sync.Mutex
    initialized                    bool
    // ...
}
```

Consider adding mutex protection to prevent race conditions when multiple reconciles call `initializeWithHCP()` concurrently.

control-plane-operator/controllers/gcpprivateserviceconnect/psc_endpoint_controller.go
```diff
@@ -0,0 +1,592 @@
...
+		b.customerProject = hcp.Spec.Platform.GCP.Project
+		b.region = hcp.Spec.Platform.GCP.Region
+		// TODO: Extract controlPlaneOperatorGSA when RolesRef is available
+		b.controlPlaneOperatorGSA = ""
+	} else {
+		b.customerProject = ""
+		b.region = ""
+		b.controlPlaneOperatorGSA = ""
+	}
+}
```
cblecker, line 87: **Question: Is there a tracking issue for this TODO?**

The `controlPlaneOperatorGSA` is never used. Is this needed for WIF authentication, and if so, is there a Jira tracking completion of this work?

control-plane-operator/controllers/gcpprivateserviceconnect/psc_endpoint_controller.go
```diff
@@ -0,0 +1,592 @@
...
+
+	// 10. Reconcile PSC Endpoint
+	return r.reconcilePSCEndpoint(ctx, gcpPSC, hcp, customerGCPClient, customerProject, region, log)
+}
+
+// isServiceAttachmentReady checks if the management-side Service Attachment is ready
+func (r *GCPPrivateServiceConnectReconciler) isServiceAttachmentReady(gcpPSC *hyperv1.GCPPrivateServiceConnect) bool {
+	// Check if management-side has populated ServiceAttachmentURI
+	if gcpPSC.Status.ServiceAttachmentURI == "" {
+		return false
```
cblecker, line 237: **Missing periodic requeue for drift detection**

The AWS PrivateLink controller always requeues after successful reconciliation to catch out-of-band changes:

```go
// control-plane-operator/controllers/awsprivatelink/awsprivatelink_controller.go:483-485
// always requeue to catch and report out of band changes in AWS
return ctrl.Result{RequeueAfter: 5 * time.Minute}, nil
```

Without this, if someone manually deletes the GCP ForwardingRule or Address outside of Kubernetes, this controller won't detect or remediate it until something else triggers a reconcile.

Consider adding periodic requeue to align with the AWS pattern.

control-plane-operator/controllers/gcpprivateserviceconnect/psc_endpoint_controller.go
```diff
@@ -0,0 +1,592 @@
...
+func (r *GCPPrivateServiceConnectReconciler) constructEndpointName(gcpPSC *hyperv1.GCPPrivateServiceConnect) string {
+	// Use service attachment name as base - it's unique and within GCP naming limits
+	return fmt.Sprintf("%s-endpoint", gcpPSC.Status.ServiceAttachmentName)
+}
+
+func (r *GCPPrivateServiceConnectReconciler) constructIPAddressName(gcpPSC *hyperv1.GCPPrivateServiceConnect) string {
+	// Use service attachment name as base - it's unique and within GCP naming limits
+	return fmt.Sprintf("%s-ip", gcpPSC.Status.ServiceAttachmentName)
+}
+
```
cblecker, line 539: **Potential invalid resource names if ServiceAttachmentName is empty**

If `gcpPSC.Status.ServiceAttachmentName` is empty (before management-side populates it), this returns `-endpoint` and `constructIPAddressName` returns `-ip`, which would likely fail GCP API validation.

While `isServiceAttachmentReady()` should prevent this path, adding defensive validation here would be safer:

```go
func (r *GCPPrivateServiceConnectReconciler) constructEndpointName(gcpPSC *hyperv1.GCPPrivateServiceConnect) string {
    if gcpPSC.Status.ServiceAttachmentName == "" {
        return "" // or panic, depending on how you want to handle this
    }
    return fmt.Sprintf("%s-endpoint", gcpPSC.Status.ServiceAttachmentName)
}
```

control-plane-operator/controllers/gcpprivateserviceconnect/psc_endpoint_controller.go
```diff
@@ -0,0 +1,592 @@
...
+// handleGCPError handles GCP API errors with appropriate retry logic
+func (r *GCPPrivateServiceConnectReconciler) handleGCPError(ctx context.Context, gcpPSC *hyperv1.GCPPrivateServiceConnect, reason string, err error) (ctrl.Result, error) {
+	log, _ := logr.FromContext(ctx)
+
+	var requeueAfter time.Duration
+	var message string
+
+	if googleErr, ok := err.(*googleapi.Error); ok {
+		switch googleErr.Code {
```
cblecker, line 579: **Logger error not checked**

The bool return from `logr.FromContext` is discarded. AWS checks this properly:

```go
// control-plane-operator/controllers/awsprivatelink/awsprivatelink_controller.go:357-360
log, err := logr.FromContext(ctx)
if err != nil {
    return ctrl.Result{}, fmt.Errorf("logger not found: %w", err)
}
```

While unlikely to fail in practice, checking the return maintains consistency with the AWS pattern.

control-plane-operator/controllers/gcpprivateserviceconnect/psc_endpoint_controller.go
```diff
@@ -0,0 +1,592 @@
...
+			RateLimiter:             workqueue.NewTypedItemExponentialFailureRateLimiter[reconcile.Request](3*time.Second, 30*time.Second),
+			MaxConcurrentReconciles: 10,
+		}).
+		Build(r)
+	if err != nil {
+		return fmt.Errorf("failed setting up with a controller manager: %w", err)
+	}
+	r.Client = mgr.GetClient()
+
+	return nil
```
cblecker, line 118: **Question: Is HCP watching intentionally omitted?**

AWS watches the HostedControlPlane for endpoint access changes:

```go
// control-plane-operator/controllers/awsprivatelink/awsprivatelink_controller.go:319
Watches(&hyperv1.HostedControlPlane{}, handler.Funcs{UpdateFunc: r.enqueueOnAccessChange(mgr)})
```

This allows re-reconciling when certain HCP fields change. Is this not needed for GCP, or should it be added for parity?

control-plane-operator/controllers/gcpprivateserviceconnect/psc_endpoint_controller.go
cristianoveiga: The AWS implementation has some historical patterns that may not be optimal today. As noted in a previous review (PR #7207), the AWS controller used separate informers because the manager cache was previously scoped to the cluster - that's no longer the case since the cache is now scoped to the control plane namespace.

  Similarly, the mutex in AWS's clientBuilder may be defensive for patterns that don't apply to our simpler GCP implementation:

  1. The CPO runs per-HCP namespace with a single GCPPrivateServiceConnect CR
  2. The GCP project and region are immutable fields
  3. Even concurrent reconciles for the same CR would set identical values

  I would like to keep it simple without mutex for now. We can add it if the design evolves to require it.

control-plane-operator/controllers/gcpprivateserviceconnect/psc_endpoint_controller.go
cristianoveiga: Thanks for the suggestion. I considered this, but I think we can keep it simple here:

  1. The management-side controller populates ServiceAttachmentURI and ServiceAttachmentName atomically in the same status update
  2. isServiceAttachmentReady() blocks the normal reconciliation flow until the URI and condition are set
  3. During deletion, if ServiceAttachmentName is somehow empty, the CR was never properly initialized - indicating a larger issue that should be fixed at the source

  Adding defensive validation here would mask a root cause bug in the management-side controller. If this ever fails, the GCP API error (-endpoint invalid name) would clearly point to the issue for debugging.

  Let me know if you feel strongly about adding it - happy to reconsider.

control-plane-operator/controllers/gcpprivateserviceconnect/psc_endpoint_controller.go
cristianoveiga: Good catch - this is left over from the initial implementation (before we introduced WIF support). I will remove it.

control-plane-operator/controllers/gcpprivateserviceconnect/psc_endpoint_controller.go
cristianoveiga: Intentionally omitted for now. The HCP watch in AWS is used to immediately react when EndpointAccess changes (to create/remove ExternalName services for private DNS routing).

I think it makes more sense add the HCP watch along with the DNS handling logic in the next PR (for https://issues.redhat.com/browse/GCP-199. I added a note on the Jira ticket about that)

control-plane-operator/controllers/gcpprivateserviceconnect/psc_endpoint_controller.go
cristianoveiga: Removed!

control-plane-operator/controllers/gcpprivateserviceconnect/psc_endpoint_controller.go
cristianoveiga: Added check

control-plane-operator/controllers/gcpprivateserviceconnect/psc_endpoint_controller.go
cristianoveiga: Added periodic requeue

control-plane-operator/controllers/gcpprivateserviceconnect/psc_endpoint_controller.go
```diff
@@ -0,0 +1,611 @@
...
+	// 3. Add finalizer only if WIF token is accessible
+	// The token is created by the token minter after credential secrets are ready.
+	// If we add the finalizer before the token exists, cleanup on deletion will fail.
+	if !controllerutil.ContainsFinalizer(gcpPSC, pscEndpointFinalizer) {
+		if !isWIFTokenAccessible() {
+			log.Info("WIF token not yet accessible, waiting before adding finalizer")
+			return ctrl.Result{RequeueAfter: time.Second * 30}, nil
+		}
+		controllerutil.AddFinalizer(gcpPSC, pscEndpointFinalizer)
+		return ctrl.Result{}, r.Update(ctx, gcpPSC)
```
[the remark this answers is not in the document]
cristianoveiga: Good catch. The comment was misleading - this controller only supports Workload Identity Federation (WIF) credentials, which are set up by the token minter in the control-plane-operator. Service account keys aren't part of this architecture.

  Updated the comments in InitCustomerGCPClient to explicitly state the WIF requirement and clarify the relationship with the finalizer gate.

control-plane-operator/controllers/gcpprivateserviceconnect/psc_endpoint_controller.go
```diff
@@ -0,0 +1,613 @@
...
+			log.Info("PSC endpoint not found, deletion already completed", "name", endpointName)
+		} else {
+			return false, fmt.Errorf("failed to delete PSC endpoint: %w", err)
+		}
+	} else {
+		// Check operation status (following management-side pattern)
+		if op != nil && op.Status == "RUNNING" {
+			log.Info("PSC endpoint deletion in progress", "operation", op.Name)
+			return false, nil // Not completed yet
+		}
```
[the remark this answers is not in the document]
cristianoveiga: addressed

control-plane-operator/controllers/gcpprivateserviceconnect/psc_endpoint_controller.go
```diff
@@ -0,0 +1,616 @@
...
+		})
+	}
+
+	if err := r.Status().Patch(ctx, gcpPSC, patch); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// Always requeue to catch and report out-of-band changes in GCP (drift detection)
+	return ctrl.Result{RequeueAfter: driftDetectionRequeueInterval}, nil
+}
```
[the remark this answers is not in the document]
cristianoveiga: I considered this but believe the validation is unnecessary for the following reasons:

  1. GCP ForwardingRules are immutable - The target and IPAddress fields cannot be modified after creation. Any "drift" would require deleting and recreating the resource.
  2. Controller recreates on deletion - If someone deletes the endpoint, our controller detects this via the periodic requeue (drift detection) and recreates it with the correct values on the next reconcile.
  3. Name collision prevents replacement - The endpoint name is derived from the service attachment name. If someone tries to create a different endpoint with the same name pointing elsewhere, GCP would reject it as a conflict.
  4. No remediation path - Even if we detected a mismatch, we only set the condition to False. We don't actually delete and recreate the endpoint, so the user would see "unhealthy" with no automatic fix.

  The simple check if endpoint.IPAddress != "" && endpoint.Target != "" is sufficient because if the endpoint exists with those fields populated, it was created by this controller with the correct values.

control-plane-operator/controllers/gcpprivateserviceconnect/psc_endpoint_controller.go
```diff
@@ -0,0 +1,616 @@
...
+		_, err := customerGCPClient.Addresses.Delete(customerProject, region, ipName).Context(ipDeleteCtx).Do()
+		if err != nil && !isNotFoundError(err) {
+			log.Error(err, "failed to delete reserved IP, continuing with cleanup")
+		} else {
+			log.Info("Reserved IP address deleted", "name", ipName)
+		}
+	}
+
+	return true, nil // Deletion completed
+}
```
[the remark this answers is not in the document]
cristianoveiga: addressed

control-plane-operator/controllers/gcpprivateserviceconnect/psc_endpoint_controller.go
```diff
@@ -0,0 +1,648 @@
...
+	op, err := customerGCPClient.Addresses.Insert(customerProject, region, ipAddress).Context(insertCtx).Do()
+	if err != nil {
+		return r.handleGCPError(ctx, gcpPSC, "IPReservationFailed", err)
+	}
+
+	// GCP operations can be PENDING, RUNNING, or DONE
+	if op.Status != "DONE" {
+		log.Info("IP reservation in progress", "operation", op.Name, "status", op.Status)
+		return ctrl.Result{RequeueAfter: time.Second * 15}, nil
+	}
```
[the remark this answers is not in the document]
cristianoveiga: addressed

control-plane-operator/controllers/gcpprivateserviceconnect/psc_endpoint_controller.go
```diff
@@ -0,0 +1,656 @@
...
+
+	// Check if GCPServiceAttachmentAvailable condition is True
+	for _, condition := range gcpPSC.Status.Conditions {
+		if condition.Type == string(hyperv1.GCPServiceAttachmentAvailable) {
+			return condition.Status == metav1.ConditionTrue
+		}
+	}
+
+	return false
+}
```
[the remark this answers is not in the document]
cristianoveiga: addressed

PR #7613, opened 2026-01-30
CNTRLPLANE-418: Aro swift private

enxebre:
Adds support for swift in ARO:
- Introduces ARO Swift annotation
- isPrivate returns true now when ARO and swift.
- Router component runs when ARO and swift and listens on 443
- Data plane HAProxy: Updates nodepool apiserver-haproxy to handle Swift scenarios with different ports and proxy protocol settings
- Shared ingress: Modified to skip dataplane-kas-service backend when Swift is enabled
- Add a bunch of unit tests

In this PR we still support the no swift path to not break CI.

Fixes 

## Checklist:


sharedingress-config-generator/config.go
```diff
@@ -122,13 +122,16 @@ func getBackendsForHostedCluster(ctx context.Context, hc hyperv1.HostedCluster,
...
 
-	backends = append(backends, backendDesc{
-		Name:         kasService.Namespace + "-" + kasService.Name,
-		SVCIP:        kasService.Spec.ClusterIP,
-		SVCPort:      kasService.Spec.Ports[0].Port,
-		ClusterID:    hc.Spec.ClusterID,
-		AllowedCIDRs: allowedCIDRs,
-	})
+	// When is private (ARO with Swift), we don't need to create a backend for the dataplane-kas-svc frontend.
+	if util.IsPrivateHC(&hc) {
```
csrwng: should this be `if !util.IsPrivateHC(&hc)`?

control-plane-operator/controllers/hostedcontrolplane/v2/router/config.go
```diff
@@ -116,9 +120,14 @@ func generateRouterConfig(routeList *routev1.RouteList, svcsNameToIP map[string]
 			p.Backends = append(p.Backends, backendDesc{Name: "metrics_forwarder", HostName: route.Spec.Host, DestinationServiceIP: svcsNameToIP[route.Spec.To.Name], DestinationPort: route.Spec.Port.TargetPort.IntVal})
 		}
 	}
-	if p.HasKubeAPI {
-		p.KASSVCPort = config.KASSVCPort
+
+	p.ListenPort = 8443
+	if azureutil.IsAroHCP() && util.IsPrivateHCP(hcp) {
+		// in ARO with Swift there is no service for the router as connections go directly to the router pod.
+		// so we should listen on HTTPS port 443.
```
csrwng: why does it need to be port 443 though?

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -1162,22 +1162,6 @@ func (r *HostedControlPlaneReconciler) reconcileCPOV2(ctx context.Context, hcp *
 	return utilerrors.NewAggregate(errs)
 }
 
-// useHCPRouter returns true if a dedicated common router is created for a HCP to handle ingress for the managed endpoints.
```
csrwng, line 1169: 👍 I didn't reuse the router component function in my PR, but it does make sense to consolidate and not repeat here. I'll do the same there.

sharedingress-config-generator/config.go
enxebre: nice catch, thanks! update the check and the fixtures rendered by the unit

control-plane-operator/controllers/hostedcontrolplane/v2/router/config.go
enxebre: Good point. I assumed this was to play nicely with the guest infra LB that forwards to swift, cc @jharrington22

jharrington22: No need for it to be 443, we can change the LB ports in Azure

sharedingress-config-generator/config.go
```diff
@@ -163,14 +166,14 @@ func getBackendsForHostedCluster(ctx context.Context, hc hyperv1.HostedCluster,
 		case ignitionserver.Route("").Name:
 			externalDNSBackends = append(externalDNSBackends, externalDNSBackendDesc{
 				Name:         route.Namespace + "-ignition",
-				HostName:     route.Spec.Host,
+				HostName:     route.Spec.Host, // TODO(alberto): This will be set to hypershift.local by reconcileInternalRoute. We need to have another external route or get from the spec don't expose it at all.
```
muraee, line 169: should we resolve this before merging the PR? public hostname should be set in the publishing strategy spec

enxebre: kas would be resolvable via shared ingress. For anything else I would expect we follow up to just drop those backends. I'd rather don't introduce anymore behavioral changes to this PR.

sharedingress-config-generator/config.go
```diff
@@ -122,13 +122,16 @@ func getBackendsForHostedCluster(ctx context.Context, hc hyperv1.HostedCluster,
...
+	// When is private (ARO with Swift), we don't need to create a backend for the dataplane-kas-svc frontend.
+	if !util.IsPrivateHC(&hc) {
+		backends = append(backends, backendDesc{
+			Name:         kasService.Namespace + "-" + kasService.Name,
+			SVCIP:        kasService.Spec.ClusterIP,
+			SVCPort:      kasService.Spec.Ports[0].Port,
+			ClusterID:    hc.Spec.ClusterID,
+			AllowedCIDRs: allowedCIDRs,
+		})
+	}
```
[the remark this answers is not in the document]
enxebre: this is intentional, shared ingress only deploy on azure for now

PR #7625, opened 2026-02-03
CNTRLPLANE-1388: Add Spot instance support in the API

enxebre:
Follow up for https://github.com/openshift/hypershift/pull/7567

## API Changes

- Add `marketType` field to `PlacementOptions` with values: OnDemand, Spot, CapacityBlocks
- Add `SpotOptions` struct with optional `maxPrice` field for spot instance configuration
- Add `terminationHandlerQueueURL` field to `AWSPlatformSpec` for SQS queue configuration
- Deprecate `capacityReservation.marketType` in favor of `placement.marketType`
- Add CEL validation rules to ensure proper spot configuration:
  - Spot instances cannot combine with Capacity Reservations
  - Spot instances require default tenancy
  - CapacityBlocks requires capacityReservation to be specified
  - Spot options can only be specified when marketType is Spot

## Controller Changes

- Update `awsMachineTemplateSpec` to handle new marketType field in placement options
- Configure CAPA `SpotMarketOptions` when marketType is Spot
- Update `isSpotEnabled` to check API spec first, then annotation (for backward compatibility)
- Refactor AWS Node Termination Handler to read queue URL from API instead of annotation
- Update `isAWSNodeTerminationHandlerNeeded` to detect spot NodePools via API

## Testing

- Add unit tests for `isSpotEnabled` function
- Add unit tests for spot market type in `awsMachineTemplateSpec`
- Add unit tests for `getTerminationHandlerQueueURL`
- Add e2e test for spot instances with API configuration
- Update existing tests to use new API fields

Fixes 

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * AWS spot support with selectable market types (OnDemand, Spot, CapacityBlocks), spot options including max-price, and an automatic spot-interruption remediation controller.

* **Bug Fixes / Validation**
  * Expanded cross-field validations and clearer constraints for spot, tenancy, capacity reservations, and max-price interactions.

* **Chores**
  * Termination-handler queue URL moved from annotations into the AWS platform spec; RBAC adjusted to allow machine deletions; extensive test coverage added.

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/testdata/aws-node-termination-handler/AROSwift/zz_fixture_TestControlPlaneComponents_aws_node_termination_handler_deployment.yaml
+++ b/control-plane-operator/controllers/hostedcontrolplane/testdata/aws-node-termination-handler/AROSwift/zz_fixture_TestControlPlaneComponents_aws_node_termination_handler_deployment.yaml
@@ -111,7 +111,7 @@ spec:
           value: /etc/aws/credentials
         - name: ENABLE_KUBERNETES_EVENTS
           value: "true"
-        image: public.ecr.aws/aws-ec2/aws-node-termination-handler:v1.25.3
+        image: aws-node-termination-handler
         imagePullPolicy: IfNotPresent
         name: aws-node-termination-handler
         resources:
--- a/control-plane-operator/controllers/hostedcontrolplane/testdata/aws-node-termination-handler/GCP/zz_fixture_TestControlPlaneComponents_aws_node_termination_handler_deployment.yaml
+++ b/control-plane-operator/controllers/hostedcontrolplane/testdata/aws-node-termination-handler/GCP/zz_fixture_TestControlPlaneComponents_aws_node_termination_handler_deployment.yaml
@@ -111,7 +111,7 @@ spec:
           value: /etc/aws/credentials
         - name: ENABLE_KUBERNETES_EVENTS
           value: "true"
-        image: public.ecr.aws/aws-ec2/aws-node-termination-handler:v1.25.3
+        image: aws-node-termination-handler
         imagePullPolicy: IfNotPresent
         name: aws-node-termination-handler
         resources:
--- a/control-plane-operator/controllers/hostedcontrolplane/testdata/aws-node-termination-handler/IBMCloud/zz_fixture_TestControlPlaneComponents_aws_node_termination_handler_deployment.yaml
+++ b/control-plane-operator/controllers/hostedcontrolplane/testdata/aws-node-termination-handler/IBMCloud/zz_fixture_TestControlPlaneComponents_aws_node_termination_handler_deployment.yaml
@@ -111,7 +111,7 @@ spec:
           value: /etc/aws/credentials
         - name: ENABLE_KUBERNETES_EVENTS
           value: "true"
-        image: public.ecr.aws/aws-ec2/aws-node-termination-handler:v1.25.3
+        image: aws-node-termination-handler
         imagePullPolicy: IfNotPresent
         name: aws-node-termination-handler
         resources:
--- a/control-plane-operator/controllers/hostedcontrolplane/testdata/aws-node-termination-handler/TechPreviewNoUpgrade/zz_fixture_TestControlPlaneComponents_aws_node_termination_handler_deployment.yaml
+++ b/control-plane-operator/controllers/hostedcontrolplane/testdata/aws-node-termination-handler/TechPreviewNoUpgrade/zz_fixture_TestControlPlaneComponents_aws_node_termination_handler_deployment.yaml
@@ -111,7 +111,7 @@ spec:
           value: /etc/aws/credentials
         - name: ENABLE_KUBERNETES_EVENTS
           value: "true"
-        image: public.ecr.aws/aws-ec2/aws-node-termination-handler:v1.25.3
+        image: aws-node-termination-handler
         imagePullPolicy: IfNotPresent
         name: aws-node-termination-handler
         resources:
--- a/control-plane-operator/controllers/hostedcontrolplane/testdata/aws-node-termination-handler/zz_fixture_TestControlPlaneComponents_aws_node_termination_handler_deployment.yaml
+++ b/control-plane-operator/controllers/hostedcontrolplane/testdata/aws-node-termination-handler/zz_fixture_TestControlPlaneComponents_aws_node_termination_handler_deployment.yaml
@@ -111,7 +111,7 @@ spec:
           value: /etc/aws/credentials
         - name: ENABLE_KUBERNETES_EVENTS
           value: "true"
-        image: public.ecr.aws/aws-ec2/aws-node-termination-handler:v1.25.3
+        image: aws-node-termination-handler
         imagePullPolicy: IfNotPresent
         name: aws-node-termination-handler
         resources:
--- a/control-plane-operator/controllers/hostedcontrolplane/testdata/hosted-cluster-config-operator/AROSwift/zz_fixture_TestControlPlaneComponents_hosted_cluster_config_operator_role.yaml
+++ b/control-plane-operator/controllers/hostedcontrolplane/testdata/hosted-cluster-config-operator/AROSwift/zz_fixture_TestControlPlaneComponents_hosted_cluster_config_operator_role.yaml
@@ -98,5 +98,6 @@ rules:
   - get
   - patch
   - update
+  - delete
   - list
   - watch


control-plane-operator/hostedclusterconfigoperator/controllers/spotremediation/spotremediation.go
```diff
@@ -0,0 +1,126 @@
...
+		}
+		return ctrl.Result{}, fmt.Errorf("failed to get Node: %w", err)
+	}
+
+	taintKey := nthTaintKey(node)
+	if taintKey == "" {
+		return ctrl.Result{}, nil
+	}
+
+	log.Info("Node has NTH taint, looking up Machine", "taintKey", taintKey)
```
muraee: s/NTH/node-termination-handler, seeing `NTH` in the logs might be confusing for people not familiar with the abbreviation.

control-plane-operator/hostedclusterconfigoperator/controllers/spotremediation/spotremediation.go
```diff
@@ -0,0 +1,126 @@
...
+	spotInterruptionSignalAnnotation = "hypershift.openshift.io/spot-interruption-signal"
+)
+
+type reconciler struct {
+	client             client.Client
+	guestClusterClient client.Client
+}
+
+func (r *reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
+	log := ctrl.LoggerFrom(ctx)
```
muraee: can we also include the node name in the logs `WithValues("node", req.NamespacedName)`

control-plane-operator/hostedclusterconfigoperator/controllers/spotremediation/spotremediation.go
```diff
@@ -0,0 +1,126 @@
...
+	if machine.Annotations == nil {
+		machine.Annotations = map[string]string{}
+	}
+	machine.Annotations[spotInterruptionSignalAnnotation] = taintKey
+	if err := r.client.Update(ctx, machine); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to annotate Machine %s: %w", machine.Name, err)
+	}
+
+	log.Info("Deleting Machine for spot interruption remediation", "machine", machine.Name, "namespace", machine.Namespace, "taintKey", taintKey)
+	if err := r.client.Delete(ctx, machine); err != nil {
```
muraee, line 82: should the nodePool controller handle the deletion since it's own the lifecycle of CAPI resources? HCCO deleting CAPI machines is going cross it's boundary imo.

Also, why do we need this controller? shouldn't the MHC be able to detect the drained nodes and replace them?

enxebre: >should the nodePool controller handle the deletion since it's own the lifecycle of CAPI resources? HCCO deleting CAPI machines is going cross it's boundary imo.

I see this the other way around. To decide which machine to delete, the controller needs access the guest cluster to watch the node, which the ho/nodepool controller is not empowered to do.

>Also, why do we need this controller? shouldn't the MHC be able to detect the drained nodes and replace them?

That's a good point. The MHC will only react when the ec2 instance is actually terminated and once the machine CR reports that. So having this controllers speeds things up, for spot we know the instance is doomed. The MHC is still there for robustness and cover other scenarios.

Alternatives considered:
1 - Let the controller put a condition in the Node instead. Then let the MHC watch that condition and trigger deletion.
2 - Completely rewrite own own nth which does the above (set a condition or delete the node).

I opted for current approach to avoid the additional extra steps in 1 workflow, and discarded 2 for reusability of the component.

muraee: > the controller needs access the guest cluster to watch the node, which the ho/nodepool controller is not empowered to do.

right, what I was thinking is since this controller already marks machines to be deleted with `deletedspotInterruptionSignalAnnotation`, the nodePool controller can take it from there and delete machines with this annotation as part of the `reconcileAWSMachines()` here https://github.com/openshift/hypershift/blob/3b8c325b6af34efa4364e69bc4da8a436f0cb81f/hypershift-operator/controllers/nodepool/aws.go#L259

api/hypershift/v1beta1/aws.go
```diff
@@ -87,24 +98,74 @@ type PlacementOptions struct {
...
 	Tenancy string `json:"tenancy,omitempty"`
 
+	// marketType specifies the EC2 instance purchasing model.
+	//
+	// Possible values:
+	// - "OnDemand": Standard on-demand instances (default if unset)
+	// - "Spot": Spot instances using spare EC2 capacity at reduced prices but may be interrupted.
+	//           Requires spot options and terminationHandlerQueueURL on the HostedCluster.
+	// - "CapacityBlocks": Scheduled pre-purchased compute capacity. Recommended for GPU/ML workloads.
+	//                     Requires capacityReservation with a specific reservation ID.
```
JoelSpeed: Might want to consider making this prose instead of bullets. The exact formatting doesn't translate well across the different formats so bullets can end up concatenated and look weird

api/hypershift/v1beta1/aws.go
```diff
@@ -87,24 +98,74 @@ type PlacementOptions struct {
...
+
+	// spot configures Spot instance options.
+	// Required when marketType is "Spot".
+	//
+	// Spot instances use spare EC2 capacity at reduced prices but may be interrupted
+	// with a 2-minute warning. Requires terminationHandlerQueueURL to be set on the
+	// HostedCluster's AWS platform spec for graceful handling of interruptions.
+	//
+	// +optional
+	Spot *SpotOptions `json:"spot,omitempty"`
```
JoelSpeed: You probably don't need a pointer here unless yo want `spot: {}` to be a valid user choice, distinct from being omitted

api/hypershift/v1beta1/aws.go
```diff
@@ -87,24 +98,74 @@ type PlacementOptions struct {
...
+
 	// capacityReservation specifies Capacity Reservation options for the NodePool instances.
 	//
 	// Cannot be specified when tenancy is set to "host" as Dedicated Hosts
 	// do not support Capacity Reservations. Compatible with "default" and "dedicated" tenancy.
 	//
+	// Required when marketType is "CapacityBlocks".
```
JoelSpeed, line 127: New validation? Or just docs update?

api/hypershift/v1beta1/aws.go
```diff
@@ -87,24 +98,74 @@ type PlacementOptions struct {
...
+	// For example, "0.50" means 50 cents per hour.
+	//
+	// Note: AWS recommends NOT setting maxPrice to reduce interruption frequency.
+	// When omitted, you pay the current Spot price (capped at On-Demand price).
+	// AWS minimum allowed value is $0.001.
+	//
+	// +optional
+	// +kubebuilder:validation:Pattern=`^[0-9]+(\.[0-9]+)?$`
+	// +kubebuilder:validation:MaxLength=20
+	MaxPrice *string `json:"maxPrice,omitempty"`
```
JoelSpeed: Is the empty string a valid value? If not, probably doesn't need to be a pointer

api/hypershift/v1beta1/aws.go
```diff
@@ -87,24 +98,74 @@ type PlacementOptions struct {
...
+	// The value should be a decimal number representing the price per hour in USD.
+	// For example, "0.50" means 50 cents per hour.
+	//
+	// Note: AWS recommends NOT setting maxPrice to reduce interruption frequency.
+	// When omitted, you pay the current Spot price (capped at On-Demand price).
+	// AWS minimum allowed value is $0.001.
+	//
+	// +optional
+	// +kubebuilder:validation:Pattern=`^[0-9]+(\.[0-9]+)?$`
+	// +kubebuilder:validation:MaxLength=20
```
JoelSpeed, line 150: Minlength?

api/hypershift/v1beta1/aws.go
```diff
@@ -87,24 +98,74 @@ type PlacementOptions struct {
...
 	// +optional
 	CapacityReservation *CapacityReservationOptions `json:"capacityReservation,omitempty"`
 }
 
-// MarketType describes the market type of the CapacityReservation for an Instance.
+// SpotOptions configures options for Spot instances.
+//
+// Spot instances use spare EC2 capacity at reduced prices but may be interrupted
+// with a 2-minute warning when EC2 needs the capacity back.
+type SpotOptions struct {
```
JoelSpeed, line 137: Should this have a required field or a MinProperties to prevent `spot: {}` being set but not mean anything?

api/hypershift/v1beta1/aws.go
```diff
@@ -352,6 +416,22 @@ type AWSPlatformSpec struct {
...
+	// and EC2 Instance Rebalance Recommendations via EventBridge rules.
+	// The AWS Node Termination Handler will poll this queue and cordon/drain nodes
+	// before they are terminated, providing a best effort for graceful shutdown.
+	//
+	// Supports both standard and FIFO queues (FIFO queues end with .fifo suffix).
+	//
+	// +optional
+	// +kubebuilder:validation:Pattern=`^https://sqs\.[a-z0-9-]+\.amazonaws\.com/[0-9]{12}/[a-zA-Z0-9_-]+(\.fifo)?$`
+	// +kubebuilder:validation:MaxLength=512
+	TerminationHandlerQueueURL *string `json:"terminationHandlerQueueURL,omitempty"`
```
JoelSpeed: Min length here too? Probably doesn't need to be a pointer if the empty string is not a valid choice

api/hypershift/v1beta1/aws.go
```diff
@@ -352,6 +416,22 @@ type AWSPlatformSpec struct {
...
+	//
+	// The queue should be configured to receive EC2 Spot Instance Interruption Warnings
+	// and EC2 Instance Rebalance Recommendations via EventBridge rules.
+	// The AWS Node Termination Handler will poll this queue and cordon/drain nodes
+	// before they are terminated, providing a best effort for graceful shutdown.
+	//
+	// Supports both standard and FIFO queues (FIFO queues end with .fifo suffix).
+	//
+	// +optional
+	// +kubebuilder:validation:Pattern=`^https://sqs\.[a-z0-9-]+\.amazonaws\.com/[0-9]{12}/[a-zA-Z0-9_-]+(\.fifo)?$`
```
JoelSpeed: Can you use CEL to implement this? It will allow a better error message and might make it easier

E.g. checking the hostname
```
isURL(self) && url(self).getHost().matches('^sqs\.[a-z0-9-]+\.amazonaws\.com$')
```
I'd then add a second for the path with `getEscapedPath()`

api/hypershift/v1beta1/aws.go
enxebre: spot block is omitzero. MarketType=spot and absence of this block express intent for spot with default price

JoelSpeed: With `omitzero`, `spot: {}` won't round trip through a structured client.

If we have 
```
marketType: Spot
```
What is the semantic difference then with
```
marketType: Spot
spot: {}
```

I think from what you've said, nothing? If that's the case, we want to prevent `spot: {}` from being persisted by an unstructured client, so we need the min properties or to turn the `maxPrice` to required

If there is a semantic difference, `spot` needs a pointer and omitempty so we can tell the difference and round trip through a structure client

hypershift-operator/controllers/nodepool/aws.go
```diff
@@ -168,21 +180,40 @@ func awsMachineTemplateSpec(infraName string, hostedCluster *hyperv1.HostedClust
...
+				switch capacityReservation.MarketType {
+				case hyperv1.MarketTypeCapacityBlock:
 					awsMachineTemplateSpec.Template.Spec.MarketType = capiaws.MarketTypeCapacityBlock
+				case hyperv1.MarketTypeOnDemand:
+					awsMachineTemplateSpec.Template.Spec.MarketType = capiaws.MarketTypeOnDemand
+				default:
+					if placement.Tenancy != "host" && capacityReservation.ID != nil {
+						// if the tenancy is not host and the ID is set, default the market type to CapacityBlock
+						awsMachineTemplateSpec.Template.Spec.MarketType = capiaws.MarketTypeCapacityBlock
+					}
```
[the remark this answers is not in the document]
enxebre: this is preexisting behaviour

test/e2e/nodepool_spot_termination_handler_test.go
```diff
@@ -88,16 +86,15 @@ func (s *SpotTerminationHandlerTest) Setup(t *testing.T) {
...
 func (s *SpotTerminationHandlerTest) BuildNodePoolManifest(defaultNodepool hyperv1.NodePool) (*hyperv1.NodePool, error) {
 	nodePool := &hyperv1.NodePool{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      s.hostedCluster.Name + "-" + "test-spot-termination",
 			Namespace: s.hostedCluster.Namespace,
+			// We use the annotation to enable spot instances for the e2e test
+			// since real spot instances are not reliable for CI.
 			Annotations: map[string]string{
 				AnnotationEnableSpot: "true",
 			},
```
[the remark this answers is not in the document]
enxebre: this is intended

test/e2e/nodepool_spot_termination_handler_test.go
```diff
@@ -88,16 +85,15 @@ func (s *SpotTerminationHandlerTest) Setup(t *testing.T) {
 	if globalOpts.Platform != hyperv1.AWSPlatform {
 		t.Skip("test only supported on platform AWS")
 	}
-	if e2eutil.IsLessThan(e2eutil.Version422) {
-		t.Skip("test only supported on version 4.22 and above")
-	}
```
muraee, line 93: why remove this? the feature is only added in 4.22 no?

PR #7630, opened 2026-02-03
test(NO-JIRA): update scale-from-zero test for CAPI 1.11 native capacity

jhjaggars:

Updates the `TestScaleFromZero` e2e test to support CAPI 1.11's native `Status.Capacity` feature on `AWSMachineTemplate`.

In CAPI 1.11, the cluster-api-provider-aws now populates `Status.Capacity` directly on the `AWSMachineTemplate`, making the workaround annotations unnecessary. The HyperShift controller detects this and skips setting annotations when `Status.Capacity` is present (see `scale_from_zero.go:92-102`).

The test previously only checked for annotations on the `MachineDeployment`, which caused it to fail with CAPI 1.11 since those annotations are no longer set.

## Changes:

- Check `AWSMachineTemplate.Status.Capacity` first (CAPI 1.11+ native support)
- Fall back to `MachineDeployment` annotations if `Status.Capacity` is empty (pre-CAPI 1.11)
- Log the capacity source for debugging
- Backward compatible with both CAPI 1.10 and 1.11+

Fixes test failures in #7590 (CAPI 1.11 upgrade)

## Test plan:

This fix makes the test pass with CAPI 1.11 while remaining compatible with CAPI 1.10.

### DIFF
--- a/test/e2e/autoscaling_test.go
+++ b/test/e2e/autoscaling_test.go
@@ -22,6 +22,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/util/retry"
 	"k8s.io/utils/ptr"
+	capiaws "sigs.k8s.io/cluster-api-provider-aws/v2/api/v1beta2"
 	capiv1 "sigs.k8s.io/cluster-api/api/v1beta1"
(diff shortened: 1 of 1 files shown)


clebs:
@jhjaggars I have cherry-picked your commit into #7590.

PR #7631, opened 2026-02-04
GCP-216: feat(nodepool): add GCP platform support

apahim:

This PR adds GCP platform support for NodePools, enabling users to create and manage worker nodes on Google Cloud Platform for HyperShift hosted clusters.

### Key Features Implemented:

**API Types (`api/hypershift/v1beta1/gcp.go`):**
- `GCPNodePoolPlatform` with configuration for machine type, zone, boot disk, service accounts, and more
- Support for custom images, disk encryption (CMEK), preemptible instances, and resource labels
- Network tags for firewall rule integration
- Host maintenance policy configuration

**Controller Implementation (`hypershift-operator/controllers/nodepool/gcp.go`):**
- Full CAPI GCP integration for machine template generation
- Automatic image resolution from release metadata
- Subnet resolution for Private Service Connect networking
- Service account configuration with appropriate scopes
- Boot disk configuration (size, type, encryption)
- Resource labels and network tags management
- Maintenance behavior based on instance type (preemptible vs standard)

**Additional Components:**
- Release metadata integration for GCP RHCOS images
- CAPI GCP machine deployment reconciliation
- Client code generation and deepcopy implementation
- Feature gate integration for TechPreviewNoUpgrade

### Configuration Options:

Users can configure:
- Machine type (e.g., `n2-standard-4`)
- Zone placement (e.g., `us-central1-a`)
- Custom boot images or auto-selection from release metadata
- Boot disk size, type (`pd-standard`, `pd-ssd`, `pd-balanced`), and KMS encryption
- Preemptible instances with automatic maintenance policy handling
- Service account and OAuth scopes
- Resource labels for organization
- Network tags for firewall rules

[Fixes [GCP-216](https://issues.redhat.com//browse/GCP-216)](https://issues.redhat.com/browse/GCP-216)

- **Feature Gate**: GCP platform support is behind the `GCPPlatform` feature gate and requires `--tech-preview-no-upgrade` installation flag
- **Subnet Resolution**: The implementation only provides subnet names to CAPG, which automatically prepends the full GCP path
(`projects/{project}/regions/{region}/subnetworks/{subnet}`)
- **Pointer Types**: Optional fields use pointer types to distinguish unset values from zero values, following Kubernetes API conventions
- **Preemptible Instances**: When `preemptible: true`, the maintenance policy is automatically set to `TERMINATE` as required by GCP
- **Testing**: Comprehensive unit test coverage for all GCP-specific functionality including machine template generation, image resolution, and
configuration handling

## Checklist:
- [x] This change includes docs (API documentation generated).
- [x] This change includes unit tests (7 test cases covering all GCP NodePool functionality).

### DIFF
--- a/docs/content/reference/aggregated-docs.md
+++ b/docs/content/reference/aggregated-docs.md
@@ -34055,6 +34055,25 @@ Common scopes include:
 </tr>
 </tbody>
 </table>
+###GCPOnHostMaintenance { #hypershift.openshift.io/v1beta1.GCPOnHostMaintenance }
+<p>
+<p>GCPOnHostMaintenance defines the behavior when a host maintenance event occurs.</p>
+</p>
+<table>
+<thead>
+<tr>
+<th>Value</th>
+<th>Description</th>
+</tr>
+</thead>
+<tbody><tr><td><p>&#34;MIGRATE&#34;</p></td>
+<td><p>GCPOnHostMaintenanceMigrate causes Compute Engine to live migrate an instance during host maintenance.</p>
+</td>
+</tr><tr><td><p>&#34;TERMINATE&#34;</p></td>
+<td><p>GCPOnHostMaintenanceTerminate causes Compute Engine to stop an instance during host maintenance.</p>
+</td>
+</tr></tbody>
+</table>
 ###GCPPlatformSpec { #hypershift.openshift.io/v1beta1.GCPPlatformSpec }
 <p>
 (<em>Appears on:</em>
--- a/docs/content/reference/api.md
+++ b/docs/content/reference/api.md
@@ -6306,6 +6306,25 @@ Common scopes include:
 </tr>
 </tbody>
 </table>
+###GCPOnHostMaintenance { #hypershift.openshift.io/v1beta1.GCPOnHostMaintenance }
+<p>
+<p>GCPOnHostMaintenance defines the behavior when a host maintenance event occurs.</p>
+</p>
+<table>
+<thead>
+<tr>
+<th>Value</th>
+<th>Description</th>
+</tr>
+</thead>
+<tbody><tr><td><p>&#34;MIGRATE&#34;</p></td>
+<td><p>GCPOnHostMaintenanceMigrate causes Compute Engine to live migrate an instance during host maintenance.</p>
+</td>
+</tr><tr><td><p>&#34;TERMINATE&#34;</p></td>
+<td><p>GCPOnHostMaintenanceTerminate causes Compute Engine to stop an instance during host maintenance.</p>
+</td>
+</tr></tbody>
+</table>
 ###GCPPlatformSpec { #hypershift.openshift.io/v1beta1.GCPPlatformSpec }
 <p>
 (<em>Appears on:</em>


apahim:
## CI Triage: e2e-aks Failure Analysis

The `e2e-aks` job failed with a **blocking issue** that needs a code fix.

### Error
```
CustomResourceDefinition.apiextensions.k8s.io "nodepools.hypershift.openshift.io" is invalid: 
spec.validation.openAPIV3Schema.properties[spec].x-kubernetes-validations[1].rule: 
compilation failed: ERROR: undefined field 'gcp'
```

### Root Cause
The validation rule in `nodepool_types.go:101` references `has(self.platform.gcp)`, but the `gcp` field is behind the `GCPPlatform` feature gate. When the feature gate is disabled (as in e2e-aks), the field doesn't exist in the CRD schema, causing CEL validation compilation to fail.

### Suggested Fix
Use a feature-gated validation marker. Change:
```go
// +kubebuilder:validation:XValidation:rule="self.arch \!= 'arm64' || has(self.platform.aws) || has(self.platform.azure) || has(self.platform.agent) || has(self.platform.gcp) || self.platform.type == 'None'", message="Setting Arch to arm64 is only supported for AWS, Azure, Agent, GCP and None"
```

To:
```go
// Keep original rule without GCP
// +kubebuilder:validation:XValidation:rule="self.arch \!= 'arm64' || has(self.platform.aws) || has(self.platform.azure) || has(self.platform.agent) || self.platform.type == 'None'", message="Setting Arch to arm64 is only supported for AWS, Azure, Agent and None"
// Add feature-gated rule for GCP
// +openshift:validation:FeatureGateAwareXValidation:featureGate=GCPPlatform,rule="self.arch \!= 'arm64' || has(self.platform.aws) || has(self.platform.azure) || has(self.platform.agent) || has(self.platform.gcp) || self.platform.type == 'None'", message="Setting Arch to arm64 is only supported for AWS, Azure, Agent, GCP and None"
```

This follows the same pattern used in `openshift/api` for feature-gated validations.

---
*Automated CI triage*

apahim:
@coderabbitai Thanks for the thorough review! I've addressed the following issues:

**1. Spot VM ProvisioningModel (Critical)**
Fixed in `gcp.go:114`. Now correctly maps hypershift's `ProvisioningModel` to CAPG's `ProvisioningModel` field. Spot VMs will now be properly provisioned instead of falling back to Standard.

**2. Conflicting arm64 validation rules (Major)**
Fixed in `nodepool_types.go:101-102`. Consolidated the two overlapping validation rules into a single rule that uses `self.platform.type == 'GCP'` enum check, which safely allows arm64+GCP without the conflicting feature-gated rule.

Added test coverage for the Spot VM provisioning model configuration.

---

hypershift-operator/controllers/nodepool/gcp.go
```diff
@@ -0,0 +1,322 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to resolve GCP image: %w", err)
+	}
+
+	// Configure network placement with PSC subnet
+	subnet, err := resolveGCPSubnet(hcGCPPlatform)
+	if err != nil {
+		return nil, fmt.Errorf("failed to resolve GCP subnet: %w", err)
+	}
+
```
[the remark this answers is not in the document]
apahim: Done. Updated `resolveGCPSubnet` to honor NodePool-specified subnet with clear priority:
1. NodePool `spec.platform.gcp.subnet` (if set)
2. HostedCluster PSC subnet (if configured)
3. "default" fallback

---

cmd/infra/gcp/iam.go
```diff
@@ -566,6 +566,8 @@ func (c *IAMManager) formatIssuerUri() string {
 }
 
 // formatServiceAccountID returns the service account ID for a given component name.
+// Format: {infraID}-{componentName}
+// The infraID is provided by the user via the CLI and must comply with GCP naming requirements.
```
cristianoveiga, line 570: Are we validating the infraID somewhere in HyperShift? Or is the plan to rely on validation from the CLI/CLM only?

hypershift-operator/controllers/hostedcluster/internal/platform/gcp/gcp.go
```diff
@@ -151,17 +151,21 @@ func (p GCP) reconcileGCPCluster(gcpCluster *capigcp.GCPCluster, hcluster *hyper
 			gcpCluster.Spec.AdditionalLabels = make(map[string]string)
 		}
 		for _, label := range gcpSpec.ResourceLabels {
-			gcpCluster.Spec.AdditionalLabels[label.Key] = ptr.Deref(label.Value, "")
+			if label.Value != nil {
+				gcpCluster.Spec.AdditionalLabels[label.Key] = *label.Value
+			} else {
+				gcpCluster.Spec.AdditionalLabels[label.Key] = ""
+			}
```
cristianoveiga: Any specific reason not to use ptr.Defef here ? The one-liner is more concise and it appears to do the same thing.

hypershift-operator/controllers/hostedcluster/internal/platform/gcp/gcp.go
```diff
@@ -151,17 +151,21 @@ func (p GCP) reconcileGCPCluster(gcpCluster *capigcp.GCPCluster, hcluster *hyper
...
 
 	// Add HyperShift-specific labels for resource identification
 	if gcpCluster.Spec.AdditionalLabels == nil {
 		gcpCluster.Spec.AdditionalLabels = make(map[string]string)
 	}
-	gcpCluster.Spec.AdditionalLabels[toGCPLabel("hypershift.openshift.io/cluster")] = hcluster.Name
+	gcpCluster.Spec.AdditionalLabels["hypershift-openshift-io-cluster"] = hcluster.Name
 	if hcluster.Spec.InfraID != "" {
-		gcpCluster.Spec.AdditionalLabels[toGCPLabel("hypershift.openshift.io/infra-id")] = hcluster.Spec.InfraID
+		gcpCluster.Spec.AdditionalLabels["hypershift-openshift-io-infra-id"] = hcluster.Spec.InfraID
```
cristianoveiga: These label keys are duplicated in nodepool/gcp.go. Consider defining them as constants to avoid inconsistency.

hypershift-operator/controllers/nodepool/gcp.go
```diff
@@ -0,0 +1,328 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	capigcp "sigs.k8s.io/cluster-api-provider-gcp/api/v1beta1"
+	capiv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// gcpMachineTemplate creates a GCPMachineTemplate for the given NodePool.
+// This follows the AWS and Azure patterns for CAPI machine template generation.
+func (c *CAPI) gcpMachineTemplate(ctx context.Context, templateNameGenerator func(spec any) (string, error)) (client.Object, error) {
```
cristianoveiga, line 21: The ctx parameter is unused.

hypershift-operator/controllers/nodepool/gcp.go
```diff
@@ -0,0 +1,328 @@
...
+		tags = append(tags, fmt.Sprintf("%s-worker", infraID))
+	}
+
+	return tags
+}
+
+// configureGCPMaintenanceBehavior determines the host maintenance behavior.
+func configureGCPMaintenanceBehavior(userMaintenance *string, provisioningModel *hyperv1.GCPProvisioningModel) capigcp.HostMaintenancePolicy {
+	if userMaintenance != nil && *userMaintenance != "" {
+		if *userMaintenance == "TERMINATE" {
```
cristianoveiga: Minor issue: "TERMINATE" is a magic string - could use a constant.

hypershift-operator/controllers/nodepool/gcp.go
```diff
@@ -0,0 +1,328 @@
...
+	)
+	if err != nil {
+		return nil, fmt.Errorf("failed to generate GCP machine template spec: %w", err)
+	}
+
+	// Create hash of the template spec for naming
+	templateName, err := templateNameGenerator(templateSpec)
+	if err != nil {
+		return nil, fmt.Errorf("failed to generate template name: %w", err)
+	}
```
cristianoveiga: GCP labels are included in the template name hash. This means if a user updates resourceLabels on the HostedCluster or NodePool it would trigger a rolling upgrade of all nodes.

AWS excludes AdditionalTags from the hash to avoid this (see aws.go lines 224-227). Should we do the same for GCP's AdditionalLabels?

support/globalconfig/infrastructure.go
```diff
@@ -106,5 +106,11 @@ func ReconcileInfrastructure(infra *configv1.Infrastructure, hcp *hyperv1.Hosted
 			APIServerInternalIPs: []string{},
 			IngressIPs:           []string{},
 		}
+	case hyperv1.GCPPlatform:
+		if infra.Status.PlatformStatus.GCP == nil {
+			infra.Status.PlatformStatus.GCP = &configv1.GCPPlatformStatus{}
+		}
+		infra.Status.PlatformStatus.GCP.ProjectID = hcp.Spec.Platform.GCP.Project
+		infra.Status.PlatformStatus.GCP.Region = hcp.Spec.Platform.GCP.Region
 	}
```
cristianoveiga, line 131: ResourceLabels not propagated to Infrastructure status

AWS propagates ResourceTags to infra.Status.PlatformStatus.AWS.ResourceTags (lines 67-79), but GCP doesn't propagate ResourceLabels.

Should we add similar logic for GCP?

hypershift-operator/controllers/hostedcluster/internal/platform/gcp/gcp.go
apahim: Done in fa67af23d. Updated both hostedcluster and nodepool locations.

hypershift-operator/controllers/hostedcluster/internal/platform/gcp/gcp.go
apahim: Done in fa67af23d. Extracted `GCPLabelCluster` and `GCPLabelInfraID` constants to `support/util/util.go` and referenced from both files.

hypershift-operator/controllers/nodepool/gcp.go
apahim: Keeping `ctx` for consistency with AWS and Azure `*MachineTemplate` methods, and forward-compatibility — it is common to pass `ctx` through even if not currently used, making future additions (logging, tracing, API calls) possible without a signature change.

hypershift-operator/controllers/nodepool/gcp.go
apahim: Done in fa67af23d. Defined `GCPOnHostMaintenanceTerminate` and `GCPOnHostMaintenanceMigrate` typed constants in `api/hypershift/v1beta1/gcp.go`.

hypershift-operator/controllers/nodepool/gcp.go
apahim: Done in fa67af23d. Following the AWS pattern — deep-copy the spec and nil out `AdditionalLabels` before hashing. CAPG webhooks explicitly allow in-place updates to `additionalLabels` (they exclude the field from immutability checks), so label changes will update existing templates without triggering node replacement.

cmd/infra/gcp/iam.go
apahim: `infraID` is validated server-side in the NodePool controller via `validateInfraID()` (nodepool_controller.go:951, called at line 310). However, it is intentionally platform-agnostic — adding GCP-specific naming validation would break the platform-neutral design. GCP naming compliance is the responsibility of the user/CLI when providing the infraID value.

support/globalconfig/infrastructure.go
apahim: Done in fa67af23d. Now propagating `ResourceLabels` to `infra.Status.PlatformStatus.GCP.ResourceLabels`, following the AWS `ResourceTags` pattern. Labels with the `kubernetes-io` prefix are filtered out to avoid breaking downstream components (same approach as AWS).

PR #7637, opened 2026-02-04
fix(sync-global-pullsecret): compare kubelet config content ignoring trailing newlines

devguyio:

The global pull secret syncer tried to preserve new lines in kubelet config but that caused an E2E mismatch. 

This PR fixes that and enhances the sync-global-pullsecret unit tests. 

Fixes 

## Checklist:


devguyio:
/close
gonna introduce this first on main then cherry-pick

PR #7638, opened 2026-02-04
OCPBUGS-78195: fix(sync-global-pullsecret): compare content ignoring trailing newlines

devguyio:

The global pull secret syncer tried to preserve new lines in kubelet config but that caused an E2E mismatch.

This PR fixes that and enhances the sync-global-pullsecret unit tests.

Fixes OCPBUGS-67262

## Checklist:

### DIFF
--- a/sync-global-pullsecret/sync-global-pullsecret.go
+++ b/sync-global-pullsecret/sync-global-pullsecret.go
@@ -3,6 +3,7 @@ package syncglobalpullsecret
 // sync-global-pullsecret syncs the pull secret from the user provided pull secret in DataPlane and appends it to the HostedCluster PullSecret to be deployed in the nodes of the HostedCluster.
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
@@ -25,16 +26,30 @@ type syncGlobalPullSecretOptions struct {
 	kubeletConfigJsonPath string
 }
 
-//go:generate ../hack/tools/bin/mockgen -destination=sync-global-pullsecret_mock.go -package=syncglobalpullsecret . dbusConn
+//go:generate ../hack/tools/bin/mockgen -destination=sync-global-pullsecret_mock.go -package=syncglobalpullsecret . dbusConn,KubeletRestarter
 type dbusConn interface {
 	RestartUnit(name string, mode string, ch chan<- string) (int, error)
 	Close()
+}
+
+// KubeletRestarter is an interface for restarting the kubelet service.
+// This allows tests to inject a mock implementation.
+type KubeletRestarter interface {
+	Restart() error
+}
+
+// realKubeletRestarter implements KubeletRestarter using systemd dbus.
+type realKubeletRestarter struct{}
+
+func (r *realKubeletRestarter) Restart() error {
+	return signalKubeletToRestartProcess()
 }
 
 // GlobalPullSecretSyncer handles the synchronization of pull secrets
 type GlobalPullSecretSyncer struct {
 	kubeletConfigJsonPath string
 	log                   logr.Logger
+	kubeletRestarter      KubeletRestarter
 }
 
 const (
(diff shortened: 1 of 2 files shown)


devguyio:
@coderabbitai review

devguyio:
all three failures were infra failures
/retest

sync-global-pullsecret/sync-global-pullsecret.go
```diff
@@ -25,16 +26,30 @@ type syncGlobalPullSecretOptions struct {
...
-//go:generate ../hack/tools/bin/mockgen -destination=sync-global-pullsecret_mock.go -package=syncglobalpullsecret . dbusConn
+//go:generate ../hack/tools/bin/mockgen -destination=sync-global-pullsecret_mock.go -package=syncglobalpullsecret . dbusConn,KubeletRestarter
 type dbusConn interface {
 	RestartUnit(name string, mode string, ch chan<- string) (int, error)
 	Close()
 }
 
+// KubeletRestarter is an interface for restarting the kubelet service.
+// This allows tests to inject a mock implementation.
+type KubeletRestarter interface {
```
jparrill, line 37: Is this interface necessary at all? I mean:
- The interfaces usually are there to expose a function in a contract for a generic behavior. In this case this is pretty specific
- The problem statement is the content comparison between files, could we change the code in that sense?

devguyio: The interface is necessary to allow for mocking Kubelet restarting. What alternative are you proposing?

jparrill: Not at all, the issue this PR is trying to address (IMHO) is the comparison between the content processed in the HCCO vs the file content in the Kubelet. Alternative for kubelet restart, we can do the same we have already, something like:

```
// Global variable for testing
var kubeletRestartFunc = signalKubeletToRestartProcess

// In tests
func TestCheckAndFixFile(t *testing.T) {
  originalFunc := kubeletRestartFunc
  defer func() { kubeletRestartFunc = originalFunc }()

  kubeletRestartFunc = func() error {
     return nil // or whatever error you want to test
  }
}
```

devguyio: >  the issue this PR is trying to address (IMHO) 

agreed, and that includes improving the unit tests of the function where that logic no? 

the existing unit tests were confusing because even in the happy path cases, it expected errors:

```go
{
	name:               "file exists with different content",
	description:        "file exists with different content, kubelet restart fails, rollback succeeds",
	initialContent:     `{"auths":{"old.registry.com":{"auth":"b2xkOnRlc3Q="}}}`,
	secretContent:      `{"auths":{"test.registry.com":{"auth":"dGVzdDp0ZXN0"}}}`,
	rollbackShouldFail: false,
	expectedErrorContains: []string{
		"failed to restart kubelet after 3 attempts",
		"rolled back changes",
	},
	expectedFinalContent: `{"auths":{"old.registry.com":{"auth":"b2xkOnRlc3Q="}}}`,
	expectError:          true,
},
```

if I'm reading that unit test, I would think that if the file exists with different content, the sync function should succeed and and there should be no errors, so why is `expectError` set to `true?

I understand this was done because kubelet restart wasn't mocked. So here's my thought process:

There are two ways that are the convention in this repo: 

1. Use a hand-made mock or a toggles (as you mentioned) 
```go
type kubeletRestarter = func() error
``` 
2. Use mocking library for generating mocks.


My understanding is that we started recently adopting the gomock approach. For gomock, you need an interface. In both cases we need an extra type, whether it's an interface or an extra function type it depends IMHO on what you're trying to achieve with that extra type. 

I'm fine switching to a hand-made mock and drop the gomock if that'll unblock this PR

devguyio:
`e2e-ak-4-21` failed to acquire lease

```
{  failed to acquire lease for "hypershift-aks-quota-slice": resources not found}
```

/retest

PR #7641, opened 2026-02-05
OCPBUGS-75930: [release-4.21] fix(cpo): Correct route labeling logic for HCP router infrastructure

csrwng:
Manual backport of #7605

### DIFF
--- a/support/util/visibility.go
+++ b/support/util/visibility.go
@@ -1,10 +1,17 @@
 package util
 
 import (
+	"os"
+
 	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
 
 	"k8s.io/utils/ptr"
 )
+
+func UseSharedIngress() bool {
+	managedService, _ := os.LookupEnv("MANAGED_SERVICE")
+	return managedService == hyperv1.AroHCP
+}
 
 func IsPrivateHCP(hcp *hyperv1.HostedControlPlane) bool {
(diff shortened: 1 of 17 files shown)


support/util/visibility.go
```diff
@@ -55,16 +62,104 @@ func IsPublicHC(hc *hyperv1.HostedCluster) bool {
...
+		// 1. Cluster has no public access (Private-only), OR
+		// 2. Public cluster with dedicated DNS for KAS (KAS uses Route with hostname)
+		//
+		// For PublicAndPrivate clusters using LoadBalancer for KAS:
+		// - Internal routes (Konnectivity, Ignition) are served by internal HCP router
+		// - External routes (OAuth) use the management cluster router (no public HCP router needed)
+		// This avoids creating an unnecessary public LoadBalancer service.
+		return !IsPublicHCP(hcp) || UseDedicatedDNSForKAS(hcp)
+
+	case hyperv1.AgentPlatform, hyperv1.KubevirtPlatform, hyperv1.OpenStackPlatform, hyperv1.NonePlatform:
```
enxebre, line 138: don't we need to include https://github.com/openshift/hypershift/pull/7789/changes#diff-803aa540bd449a684ada7b96a65ae59c2f5b263f619096c98d6c05de57f8a83d here?

csrwng: no, because self-managed is not part of 4.21 afaik

PR #7648, opened 2026-02-05
OCPBUGS-76337: feat(scale-from-zero): add DescribeInstanceTypes IAM permission

clebs:
Scale from zero does not work properly for ROSA HCP clusters because the nodepool IAM policies are missing the `ec2:DescribeInstanceTypes` permission.  
For scale from zero to work correctly, the `AWSMachineTemplate.Status.Capacity` fields need to be populated with information from EC2 instances.
Without the permission, this information cannot be retrieved.

Previously, capacities were set by the nodepool controller (which already has this permission) as annotations.
With the latest changes in CAPA, this is reconciled by the provider instead, and therefore, it needs the same permission to gather the data from the EC2 instances.

## Checklist:

### DIFF
--- a/cmd/infra/aws/iam.go
+++ b/cmd/infra/aws/iam.go
@@ -466,6 +466,7 @@ var (
         "ec2:DescribeAvailabilityZones",
         "ec2:DescribeImages",
         "ec2:DescribeInstances",
+        "ec2:DescribeInstanceTypes",
         "ec2:DescribeInternetGateways",
         "ec2:DescribeNatGateways",
         "ec2:DescribeNetworkInterfaces",


clebs:
/retitle OCPBUGS-76337: feat(scale-from-zero): add DescribeInstanceTypes IAM permission

cmd/infra/aws/iam.go
```diff
@@ -466,6 +466,7 @@ var (
         "ec2:DescribeAvailabilityZones",
         "ec2:DescribeImages",
         "ec2:DescribeInstances",
+        "ec2:DescribeInstanceTypes",
```
bryan-cox, line 469: What's the reason behind why this permission is now needed? Ex. Is this a new thing needed by AWS, is this because it's needed because of a new feature, etc.

clebs: This is needed because of the new scale from zero feature. Until now, capacities were set by the nodepool controller (which already has this permission) as annotations.
With the changes in CAPA, there is now the `AWSMachineTemplate.Status.Capacity` field that the provider reconciles and therefore needs the same permission to gather the data from the EC2 instances.

I have improved the description on the linked issue as well as the PR.

sjenning:
/approve
per slack discussion

PR #7652, opened 2026-02-05
OCPBUGS-65824: Add dynamic NodePort range validation to prevent cluster creation failures

jparrill:

This PR adds validation for NodePort.Port to ensure ports fall within the cluster's configured ServiceNodePortRange. This prevents late failures during cluster provisioning when the control-plane-operator rejects ports outside the acceptable range.

**The problem**: When creating a hosted cluster with NodePort set to values like 10000, the creation fails because the port is outside the acceptable range (typically 30000-32767), but this validation only happens late in the process after resources are already created.

**The solution**: 
- Adds dynamic validation that reads cluster-specific ServiceNodePortRange configuration
- Validates NodePort.Port against the configured range (supports custom ranges, not just default)
- Allows port 0 (dynamic assignment)
- Provides clear error messages with the actual configured range
- Ensures early failure with immediate feedback

Fixes https://issues.redhat.com/browse/OCPBUGS-65824

- The validation is **dynamic** - it reads the cluster's ServiceNodePortRange configuration instead of hardcoding 30000-32767
- This supports customers who may have custom port ranges configured
- No CRD changes were made - validation is runtime-only to support configurable ranges
- One existing test was updated (port 4433 → 30443) because it was using an invalid NodePort
- All existing tests continue to pass

## Checklist:
- [x] This change includes docs. (Function documentation explains the validation logic)
- [x] This change includes unit tests. (Added comprehensive tests for range parsing and validation)

🤖 Generated with Claude Code via `/jira:solve OCPBUGS-65824 origin`

### DIFF
--- a/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
+++ b/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
@@ -4135,7 +4135,11 @@ func parseNodePortRange(rangeStr string) (min, max int32, err error) {
 		return 0, 0, fmt.Errorf("invalid maximum port: %s", parts[1])
 	}
 
-	return int32(minVal), int32(maxVal), nil
+	minPort, maxPort := int32(minVal), int32(maxVal)
+	if minPort > maxPort {
+		return 0, 0, fmt.Errorf("invalid range: minimum port %d must be less than or equal to maximum port %d", minPort, maxPort)
+	}
+	return minPort, maxPort, nil
 }
 
 // validateNodePortPortRange validates that NodePort.Port is within the configured ServiceNodePortRange.
--- a/hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
+++ b/hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
@@ -5133,6 +5133,18 @@ func TestParseNodePortRange(t *testing.T) {
 			expectedMax: 99999,
 			expectError: false, // parseNodePortRange doesn't validate port limits
 		},
+		{
+			name:        "min greater than max - invalid range",
+			rangeStr:    "32767-30000",
+			expectError: true,
+		},
+		{
+			name:        "min equals max - valid single port range",
+			rangeStr:    "31000-31000",
+			expectedMin: 31000,
+			expectedMax: 31000,
+			expectError: false,
+		},
 	}
 
 	for _, tc := range testCases {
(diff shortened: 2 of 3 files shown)


jparrill:
@coderabbitai review

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -4110,6 +4111,83 @@ func validateNodePortVsServiceNetwork(hc *hyperv1.HostedCluster) field.ErrorList
 	return errs
 }
 
+// parseNodePortRange parses a ServiceNodePortRange string like "30000-32767" into min/max values.
+// This range defines which ports on worker nodes are available for NodePort services.
+// The range is configurable per cluster via spec.configuration.network.serviceNodePortRange,
+// with a default of 30000-32767 (following Kubernetes standard).
+func parseNodePortRange(rangeStr string) (min, max int32, err error) {
```
sdminonne, line 4118: @jparrill may I say that this `parseNodePortRange` func doesn't check that `min` is smaller than `max`?
Or at least I cannot see imeediately.

jparrill: Thanks for pointing this out :). Added on line 4139 🙏

hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
```diff
@@ -5077,6 +5065,330 @@ func TestValidateNodePortVsServiceNetwork(t *testing.T) {
...
+			rangeStr:    "abc-32767",
+			expectError: true,
+		},
+		{
+			name:        "invalid maximum port",
+			rangeStr:    "30000-xyz",
+			expectError: true,
+		},
+		{
+			name:        "negative port",
```
sdminonne, line 5125: @coderabbitai here

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -4110,6 +4111,87 @@ func validateNodePortVsServiceNetwork(hc *hyperv1.HostedCluster) field.ErrorList
...
+	if err != nil {
+		return 0, 0, fmt.Errorf("invalid maximum port: %s", parts[1])
+	}
+
+	minPort, maxPort := int32(minVal), int32(maxVal)
+	if minPort > maxPort {
+		return 0, 0, fmt.Errorf("invalid range: minimum port %d must be less than or equal to maximum port %d", minPort, maxPort)
+	}
+	return minPort, maxPort, nil
+}
```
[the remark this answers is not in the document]
sdminonne: That check is performed elsewhere with unit test included Line 5125 in file hostedcluter_controller_test.go file

PR #7673, opened 2026-02-06
OCPBUGS-67240: Ignition generalize image download errors

clebs:
This PR remove layer-specific sha256 digests from registry errors in the ignition to prevent cascading reconciliations caused by hash changes in resource status.

Mitigates [OCPBUGS-67240](https://issues.redhat.com/browse/OCPBUGS-67240)

The bug causes high resource usage through cascading reconciliations. This change mitigates the problem but is not completely fixing it.
A proper fix will require a redesign on the igniton server which will follow up as an enhancement proposal.

## Checklist:

## Summary by CodeRabbit

* **Refactor**
  * Error messages from image downloads are now sanitized to remove technical digest values, yielding cleaner, user-friendly messages on failure.

* **Tests**
  * Added tests validating that download error messages are consistently formatted and do not expose raw digest hashes.

### DIFF
--- a/ignition-server/controllers/image_file_cache_test.go
+++ b/ignition-server/controllers/image_file_cache_test.go
@@ -83,6 +83,64 @@ func TestImageFileCache(t *testing.T) {
 	t.Log("failure message returned:", err5)
 }
 
+func TestDownloadImageErrorHashStripping(t *testing.T) {
+	tests := []struct {
+		name        string
+		regErr      string
+		wantContain string
+		wantExclude string
+	}{
+		{
+			name:        "When error contains a sha256 hash it should remove only the hash",
+			regErr:      "unable to access the source layer sha256:a3ed95caeb02ffe68cdd9fd84406680ae93d633cb16422d00e8a7c22955b46d4: connection refused",
+			wantContain: "unable to access the source layer: connection refused",
+			wantExclude: "sha256:",
+		},
+		{
+			name:        "When error contains multiple sha256 hashes it should remove all of them",
+			regErr:      "layer sha256:a3ed95caeb02ffe68cdd9fd84406680ae93d633cb16422d00e8a7c22955b46d4 and sha256:bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb failed",
+			wantContain: "layer and failed",
+			wantExclude: "sha256:",
+		},
+		{
+			name:        "When error contains no sha256 hash it should preserve the message unchanged",
+			regErr:      "connection refused",
+			wantContain: "connection refused",
+		},
+		{
+			name:        "When error contains a short hex string after sha256 prefix it should not strip it",
+			regErr:      "unexpected sha256:abcdef value",
+			wantContain: "sha256:abcdef",
+		},
+		{
+			name:        "When error contains sha256 with uppercase hex it should still strip it",
+			regErr:      "layer sha256:A3ED95CAEB02FFE68CDD9FD84406680AE93D633CB16422D00E8A7C22955B46D4 gone",
+			wantContain: "layer gone",
+			wantExclude: "sha256:",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			g := NewGomegaWithT(t)
+
+			regClientFuncMock := func(ctx context.Context, imageRef string, pullSecret []byte, file string, out io.Writer) error {
+				return errors.New(tt.regErr)
+			}
+
+			cacheDir := t.TempDir()
+			_, err := downloadImageFile(t.Context(), regClientFuncMock, "quay.io/test/image:latest", []byte("pull-secret"), "usr/bin/test", cacheDir)
+
+			g.Expect(err).To(HaveOccurred())
+			g.Expect(err.Error()).To(ContainSubstring("failed to extract image file"))
+			g.Expect(err.Error()).To(ContainSubstring(tt.wantContain))
+			if tt.wantExclude != "" {
+				g.Expect(err.Error()).NotTo(ContainSubstring(tt.wantExclude))
+			}
+		})
+	}
+}
+
 func getImageFile(t *testing.T, sut *imageFileCache, imageRef string, imageFile string) (string, error) {
 	var buff bytes.Buffer
 	sutErr := sut.extractImageFile(t.Context(), imageRef, []byte("pull-secret"), imageFile, &buff)
@@ -97,7 +155,7 @@ func simulateFileCorruption(t *testing.T, cacheDir string) {
 	fileInfo := files[0] // only one file is supposed to be in the directory
 	cachedFileName := filepath.Join(cacheDir, fileInfo.Name())
 	t.Log("adding some garbage into file:", cachedFileName)
-	cachedFile, err := os.OpenFile(cachedFileName, os.O_APPEND|os.O_WRONLY, 0644)
+	cachedFile, err := os.OpenFile(cachedFileName, os.O_APPEND|os.O_WRONLY, 0o644)
 	if err != nil {
 		t.Fatal("failed to open cache file", err)
 	}


enxebre:
these replicas are not just controllers, they are servers behind a router. If only one is active, ignition http requests will fail 1/3 of the times, no?

rtheis:
> these replicas are not just controllers, they are servers behind a router. If only one is active, ignition http requests will fail 1/3 of the times, no?

Agreed, we only want the stateful controllers that are writing data have leader election. The serving side should respond to ignition requests from all pods.

clebs:
@enxebre @rtheis thanks for the feedback.

I am not sure I understand correctly, looking at the ignition-server code I could not find any stateful controller and ignitions server pods are designed to be stateless and all reconciling concurrently to keep their stores in sync.

~Could you please point me to what should be the target of the leader election?~

EDIT: this requires a fundamental design change in how the ignition server is deployed currently to make it (partially?) stateful.
Keeping this as WIP.

rtheis:
> @enxebre @rtheis thanks for the feedback.
> 
> I am not sure I understand correctly, looking at the ignition-server code I could not find any stateful controller and ignitions server pods are designed to be stateless and all reconciling concurrently to keep their stores in sync.
> 
> ~Could you please point me to what should be the target of the leader election?~
> 
> EDIT: this requires a fundamental design change in how the ignition server is deployed currently to make it (partially?) stateful. Keeping this as WIP.

@clebs our testing shows that all ignition servers write the node pool token secrets. I have not looked into the code further to determine if there are other concurrent resource reconciliations.

clebs:
@rtheis I have briefly discussed this and have scheduled a meeting to discuss the re-design.

ignition-server/controllers/image_file_cache.go
```diff
@@ -41,8 +43,10 @@ type imageFileCache struct {
 	mutex     sync.RWMutex
 }
 
-type checksum []byte
-type regClient func(ctx context.Context, imageRef string, pullSecret []byte, file string, out io.Writer) error
+type (
+	checksum  []byte
+	regClient func(ctx context.Context, imageRef string, pullSecret []byte, file string, out io.Writer) error
+)
```
muraee: can you please revert this change. unlike consts, grouping types is not great for readability

ignition-server/controllers/image_file_cache.go
```diff
@@ -150,7 +156,10 @@ func downloadImageFile(ctx context.Context, regClient regClient, imageRef string
 	}()
 
 	if err = regClient(ctx, imageRef, pullSecret, imageFile, newFile); err != nil {
-		return "", fmt.Errorf("failed to extract image file: %w", err)
+		// Generalize the error by removing any layer specific sha256 values to avoid
+		// cascading reconciliations caused by small hash changes in the resource status repeatedly.
+		genericErrStr := shaRegex.ReplaceAllString(err.Error(), "")
+		return "", fmt.Errorf("failed to extract image file: %w", errors.New(genericErrStr))
```
muraee, line 160: I think should we fix the original function https://github.com/openshift/hypershift/blob/main/support/releaseinfo/registryclient/client.go#L50 to not include shas in the errors instead. So any future usages of that function don't cause the same issue

ignition-server/controllers/image_file_cache.go
clebs: Sure, this was my fmt automatically.

ignition-server/controllers/image_file_cache.go
clebs: I intentionally applied this change to ignition only to prevent the reconciliation issues, but having hashes on the errors in other scenarios is probably useful.
I am not sure that extending this to every usage is what we want.

ignition-server/controllers/image_file_cache.go
clebs: Fixed thanks!

ignition-server/controllers/image_file_cache.go
```diff
@@ -136,6 +138,8 @@ func returnCacheFile(fullCacheFileName string, out io.Writer) error {
 	return nil
 }
 
+var shaRegex = regexp.MustCompile(`\ssha256:[a-fA-F0-9]{64}`)
```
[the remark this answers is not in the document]
clebs: That case will not happen and we need to strip the space at the preceding the hash to have a properly formatted error message.

ignition-server/controllers/image_file_cache.go
```diff
@@ -150,7 +154,10 @@ func downloadImageFile(ctx context.Context, regClient regClient, imageRef string
 	}()
 
 	if err = regClient(ctx, imageRef, pullSecret, imageFile, newFile); err != nil {
-		return "", fmt.Errorf("failed to extract image file: %w", err)
+		// Generalize the error by removing any layer specific sha256 values to avoid
```
rtheis, line 157: It's not clear to me how this will mitigate https://redhat.atlassian.net/browse/OCPBUGS-67144 which is the primary reconciliation problem that we have see at scale.

clebs: Hi @rtheis,
quoting @sjenning from when this was discussed and proposed a solution to the heavy load:
> The issue they are having is the ignition-servers are fighting over updating the nodepool secret (particularly the error) when they are all active. (...) All the ignition-servers trying to update the secret in the same way wouldn't necessarily be an issue, but the problem is they temporarily overload their image registry when this happens and the error recorded in the secret from each ignition-server can differ.
Another solution could just be to change the ignition-server so it reports generic errors without image layer hashes and the like.  As long as all the ignition-server reports, "failed to pull image: registry reported error" vs "failed to pull image layer < sha256 >" and all of them failing on different layers and overwriting each others error messages.

So this prevents a "reconciliation storm" caused by small variations on the resource, which should reduce the load.

sjenning: @rtheis there are two problems as I understand it

Problem 1 is that the mgmt cluster is under load when the ignition-servers all try to update the Secret with their unqiue error msg.  The PR might help with that as a mitigation.

Problem 2, and bigger issue, is that while we want all the ignition-servers in an HA config back the HTTPS service, we do _not_ want all ignition-servers to pull NodePool image and generate configs.  This leads to registry pressure that, in turn, results in errors on pulls that result in problem 1.  That part of the ignition-server should be active-passive in an HA setup and only done by one ignition-server at a time.

Solving the second issue fixes the first.  Fixing the first does not fix the second.

It is arguable whether this PR would help.  If the ignition-servers get some combination of registry errors and connection timeouts, the problem still exists.

Honestly, I'd rather just fix the bigger issue.  I need to talk with the team about this.  We might separate the nodepool config generation logic from the ignition-server and run this new component with leader election.

FYI @enxebre

clebs: From the conversation we had about this 2 items came out:
- Mitigate the immediate problem (this issue)
- Create an enhancement proposal to redesign ignition so that we can have leader election.

Th second item is still pending and is a larger effort.

enxebre: >We might separate the nodepool config generation logic from the ignition-server and run this new component with leader election.

This would be my perference. ignition-server can then sync the payload generated by the controller via http pull/push. Would this work @rtheis

rtheis: Hi folks, we discussed this during today's IBM/Red Hat HyperShift interlock and agree that splitting ignition into two components (nodepool config and server) is the best long term solution. In addition, we (IBM) really need the image download errors to be very generic which I don't think this PR gives us. If you review https://redhat.atlassian.net/browse/OCPBUGS-67144, we would need to use very generic error messages that would basically turn `failed to generate payload: ... : toomanyrequests` errors into  `failed to generate payload: toomanyrequests`. I don't recall seeing any `failed to extract image file` errors which this PR is updating.

rtheis: Also, we could consider another mitigation of only running 2 instead of 3 ignition server pods until such time as ignition is split into two components.

clebs: @rtheis I have analyzed further based on your concern. The error you mention (`failed to generate payload: ... : toomanyrequests`) originates [here](https://github.com/openshift/hypershift/blob/cd1e5f9dfb5835ee6957958fb370e7f129ec8e5f/ignition-server/controllers/tokensecret_controller.go?plain=1#L288). A few lines above we can see this error wraps the error produced by `GetPayload()` (called [here](https://github.com/openshift/hypershift/blob/cd1e5f9dfb5835ee6957958fb370e7f129ec8e5f/ignition-server/controllers/tokensecret_controller.go?plain=1#L276)).

`GetPayload()` underneath calls the piece of code this PR changes and bubbles its errors up. This means that the `...` omitted from the full error above will contain image hashes when those kinds of errors happen (which is what @sjenning found when doing the RCA).

In conclusion, this change will not make all errors fully generic and identical, but will greatly reduce variance by generalizing the segment that changes the most; and therefore reduce load on api-server.

rtheis: Interesting side note ... I found that 4.20 and later deploy ignition server already runs with 2 replicas rather than 3. This appears to be due to the use of CPOv2. So when ignition server is refactored, I suspect we'll want to fix that to restore the server component to 3 replicas while the controller can remain 2.

rtheis: Thank you @clebs and @sjenning. This looks good to me as a temporary mitigation.

PR #7678, opened 2026-02-07
CNTRLPLANE-2215: feat(aws): migrate S3 to AWS SDK v2

LiangquanLi930:
Replace aws-sdk-go S3 client with aws-sdk-go-v2 implementation. Updated delegating client generator to support both v1 and v2 services, created support/awsapi package for v2 interfaces, and migrated all S3 consumers.

Fixes [CNTRLPLANE-2215](https://issues.redhat.com/browse/CNTRLPLANE-2215)

## Checklist:


LiangquanLi930:
@coderabbitai review

cmd/infra/aws/destroy.go
```diff
@@ -274,46 +283,67 @@ func (o *DestroyInfraOptions) DestroyInfra(ctx context.Context) error {
...
 	return errs
 }
 
-func emptyBucket(ctx context.Context, client s3iface.S3API, name string) error {
-	iter := s3manager.NewDeleteListIterator(client, &s3.ListObjectsInput{
-		Bucket: aws.String(name),
+// emptyBucket deletes all objects in an S3 bucket.
+// Note: AWS SDK Go v2 does not provide a BatchDelete utility like v1's s3manager.BatchDelete.
+// This is a known limitation tracked in https://github.com/aws/aws-sdk-go-v2/issues/1463
+// We manually implement batch deletion using ListObjectsV2Paginator + DeleteObjects API.
```
sdminonne, line 320: Great. Thanks @LiangquanLi930 !

sdminonne (changes requested):
@LiangquanLi930 my only real remark is in emptyBucket func. May you also some unit-tests for `emptyBucket`?
Otherwise it's a very nice work! TY!

cmd/infra/aws/destroy.go
```diff
@@ -274,46 +283,67 @@ func (o *DestroyInfraOptions) DestroyInfra(ctx context.Context) error {
...
+			return err
+		}
+		if len(page.Contents) == 0 {
+			continue
+		}
+		var objects []s3types.ObjectIdentifier
+		for _, obj := range page.Contents {
+			objects = append(objects, s3types.ObjectIdentifier{Key: obj.Key})
+		}
+		if _, err := client.DeleteObjects(ctx, &s3.DeleteObjectsInput{
```
sdminonne: Here we check only the golang `err` right? But I guess we may look inside the returned object. https://docs.aws.amazon.com/code-library/latest/ug/go_2_s3_code_examples.html

What do you thing?

sdminonne (commented):
Here is the code where I would add another check on error in response

cmd/infra/aws/destroy.go
LiangquanLi930: updated

cmd/infra/aws/destroy_test.go
```diff
@@ -0,0 +1,309 @@
...
+}
+
+func (m *mockS3Client) DeleteObjects(ctx context.Context, input *s3.DeleteObjectsInput, optFns ...func(*s3.Options)) (*s3.DeleteObjectsOutput, error) {
+	if m.deleteObjectsFunc != nil {
+		return m.deleteObjectsFunc(ctx, input, optFns...)
+	}
+	return &s3.DeleteObjectsOutput{}, nil
+}
+
+func TestEmptyBucket_EmptyBucket(t *testing.T) {
```
jparrill: Talking in general and regarding the whole file not this function in concrete. These kind of tests are not maintainable at all. There are plenty of functions that they could be reduced to a test mapping as we already have in the repo, with the format `When … it should …`. Maybe with one or a couple (to avoid very complex tests) of functions and a struct + testMap you can cover most of the cases.

etcd-backup/etcdbackup.go
```diff
@@ -104,34 +105,39 @@ func run(ctx context.Context, opts options) error {
...
-	output, err := uploader.UploadWithContext(ctx, &s3manager.UploadInput{
-		Bucket:  aws.String(opts.s3BucketName),
-		Key:     aws.String(key),
+	uploader := transfermanager.New(s3Client, transfermanager.Options{})
+	_, err = uploader.PutObject(ctx, &transfermanager.PutObjectInput{
+		Bucket:  opts.s3BucketName,
+		Key:     key,
 		Body:    f,
-		Tagging: mapToTags(opts.s3ObjectTags),
+		Tagging: aws.ToString(mapToTags(opts.s3ObjectTags)),
```
jparrill, line 134: Verify that `mapToTags` produces a URL-encoded tag string (e.g. key1=value1&key2=value2). If the v1 implementation used another format, add a test to lock in and document the expected behavior.

jparrill (commented):
Thanks for the PR 👍 . This is a big one so maybe I missed something already in the review. Meanwhile I've dropped some comments.

cmd/infra/aws/destroy_test.go
LiangquanLi930: done

etcd-backup/etcdbackup.go
LiangquanLi930: done

support/awsapi/mock_s3.go
```diff
@@ -0,0 +1,422 @@
+// Code generated by MockGen. DO NOT EDIT.
+// Source: s3.go
+//
+// Generated by this command:
+//
+//	mockgen -source=s3.go -package=awsapi -destination=mock_s3.go
+//
+
+// Package awsapi is a generated GoMock package.
+package awsapi
```
jparrill: Generated mock files should be part of the code? I mean these are not generated during the build instead of store them at git level?

jparrill:
/approve

Please @sdminonne, PTAL the PR when you have time 🙏

sdminonne:
@jparril when I wrote this mockfiles were not stored in git and `gitignored`

PR #7679, opened 2026-02-08
Add Swift networking to KAS, Add DNS side car to route MSFT DNS reque…

jharrington22:
…sts via customer vNet

## What this PR does / why we need it:

This PR implements private keyvault for ARO HCP clusters. It does this by leveraging Swift networking, to make this work its required that DNS resolution is completed via the customer vNet.

This PR introduces a DNS sidecar to accomplish the above it also means that we need to disable the wait-for-etcd init container. 

## Which issue(s) this PR fixes:
This PR accompanies changes made in https://github.com/openshift/hypershift/pull/7613
Fixes 

## Special notes for your reviewer:

I don't have a way yet to discover the kube DNS service for the AKS cluster on which hypershift runs currently its hard coded or overwritten by an annotation. I could use some thoughts here.

## Checklist:
- [ ] Subject and description added to both, commit and PR.
- [ ] Relevant issues have been referenced.
- [ ] This change includes docs. 
- [ ] This change includes unit tests.

PR #7692, opened 2026-02-10
Private kv via proxy

jharrington22:

- Implements a HTTP CONNECT azure-dns-proxy to handle DNS resolution for Private KV
- Runs the azure-dns-proxy as a container in the router
- Reorders predicates so that the router starts as a dependency of the kube-apiserver (ARO HCP Swift only)

Alternative to DNS sidecar in https://github.com/openshift/hypershift/pull/7679

### DIFF
--- a/api/hypershift/v1beta1/hostedcluster_types.go
+++ b/api/hypershift/v1beta1/hostedcluster_types.go
@@ -327,6 +327,12 @@ const (
 	// When set to "servicenodeport", each service gets its own dedicated health probe.
 	AzureLoadBalancerHealthProbeModeAnnotation = "hypershift.openshift.io/azure-load-balancer-health-probe-mode"
 
+	// SwiftPodNetworkInstanceAnnotation specifies the Swift pod network instance ID for Azure CNI Swift networking.
+	// When set on a HostedCluster, it enables Swift networking for control plane pods that need access to
+	// customer VNet resources (e.g., Azure Private Link endpoints for Key Vault).
+	// The value should be the ID of the Swift pod network instance resource in Azure.
+	SwiftPodNetworkInstanceAnnotation = "hypershift.openshift.io/swift-pod-network-instance"
+
 	// DisableClusterAutoscalerAnnotation allows disabling the cluster autoscaler for a hosted cluster.
 	// This annotation is only set by the hypershift-operator on HosterControlPlanes.
 	// It is not set by the end-user.
--- a/azure-dns-proxy/main.go
+++ b/azure-dns-proxy/main.go
@@ -1,0 +1,229 @@
+package azurednsproxy
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"os"
+	"os/signal"
+	"strings"
+	"syscall"
+	"time"
+
+	"github.com/go-logr/logr"
+	"github.com/openshift/hypershift/support/supportedversion"
+	"github.com/spf13/cobra"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+	"go.uber.org/zap/zapcore"
+)
+
+const (
+	azureDNSServer = "168.63.129.16:53"
+)
+
+func NewStartCommand() *cobra.Command {
+	l := log.Log.WithName("azure-dns-proxy")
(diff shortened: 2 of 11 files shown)


azure-dns-proxy/main.go
```diff
@@ -0,0 +1,229 @@
...
+	"github.com/go-logr/logr"
+	"github.com/openshift/hypershift/support/supportedversion"
+	"github.com/spf13/cobra"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+	"go.uber.org/zap/zapcore"
+)
+
+const (
+	azureDNSServer = "168.63.129.16:53"
```
jparrill, line 24: What's this ip?

azure-dns-proxy/main.go
```diff
@@ -0,0 +1,229 @@
...
+
+	cmd := &cobra.Command{
+		Use:   "azure-dns-proxy",
+		Short: "Runs the Azure DNS HTTP CONNECT proxy server.",
+	}
+
+	var listenAddr string
+	var requestTimeout time.Duration
+	var connectTimeout time.Duration
+	var tunnelIdleTimeout time.Duration
```
jparrill, line 41: Group all the vars plz

azure-dns-proxy/main.go
```diff
@@ -0,0 +1,229 @@
...
+
+	cmd.Flags().StringVar(&listenAddr, "listen-addr", "0.0.0.0:8888", "Address to listen on for HTTP CONNECT requests")
+	cmd.Flags().DurationVar(&requestTimeout, "request-timeout", 30*time.Second, "Timeout for proxy request handling")
+	cmd.Flags().DurationVar(&connectTimeout, "connect-timeout", 10*time.Second, "Timeout for establishing connections to target hosts")
+	cmd.Flags().DurationVar(&tunnelIdleTimeout, "tunnel-idle-timeout", 5*time.Minute, "Idle timeout for connection tunnels")
+
+	cmd.Run = func(cmd *cobra.Command, args []string) {
+		l.Info("Starting Azure DNS HTTP CONNECT proxy", "version", supportedversion.String())
+
+		azureDomains := []string{
```
jparrill, line 51: This should be a Global Var above.

azure-dns-proxy/main.go
```diff
@@ -0,0 +1,229 @@
...
+		}
+
+		<-ctx.Done()
+		l.Info("Azure DNS proxy stopped")
+	}
+
+	return cmd
+}
+
+type AzureDNSProxy struct {
```
jparrill, line 102: The types should be placed at the top of the file

azure-dns-proxy/main.go
```diff
@@ -0,0 +1,229 @@
...
+	log               logr.Logger
+	azureDNSServer    string
+	azureDomains      []string
+	connectTimeout    time.Duration
+	tunnelIdleTimeout time.Duration
+}
+
+func (p *AzureDNSProxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
```
jparrill, line 110: Include context on the workflow

jparrill: Another comment, will be there a HTTPS support too?

jparrill (changes requested):
Thanks for the PR :).

I think this is still on going, it's that the case? If that so, please put it on draft mode to avoid consume innecessary CI resources. 
Also I think the PR is still green to be reviewed, please let me know when the PR is ready for review @jharrington22.

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -327,6 +327,12 @@ const (
 	// When set to "servicenodeport", each service gets its own dedicated health probe.
 	AzureLoadBalancerHealthProbeModeAnnotation = "hypershift.openshift.io/azure-load-balancer-health-probe-mode"
 
+	// SwiftPodNetworkInstanceAnnotation specifies the Swift pod network instance ID for Azure CNI Swift networking.
+	// When set on a HostedCluster, it enables Swift networking for control plane pods that need access to
+	// customer VNet resources (e.g., Azure Private Link endpoints for Key Vault).
+	// The value should be the ID of the Swift pod network instance resource in Azure.
+	SwiftPodNetworkInstanceAnnotation = "hypershift.openshift.io/swift-pod-network-instance"
```
JoelSpeed, line 334: If this is azure specific, should azure be a prefix to the annotation value? `azure-swift-...`?

PR #7589, opened 2026-01-27
OCPBUGS-53083: ci: Improves the `isLeaderElectionFailure` function

nunnatsa:
## What this PR does / why we need it:
The `EnsureNoCrashingPods` function should ignore restarts that happened as result of leader election lost. But this is not always happening, and there is no way to understand why. For example, in this case, the function missed the leader election lost failure, and failed the test because of the pod restart; see [here](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-api-provider-kubevirt/347/pull-ci-openshift-cluster-api-provider-kubevirt-main-e2e-hypershift-kubevirt/2015780962863943680/artifacts/e2e-hypershift-kubevirt/run-e2e-local/artifacts/TestCreateCluster/namespaces/e2e-clusters-hf726-create-cluster-7975c/core/pods/logs/kubevirt-cloud-controller-manager-5bb7949f94-f8mmb-cloud-controller-manager-previous.log)

This PR adds logging from the `isLeaderElectionFailure` function, to help debugging the issue. It also changes the algorithm, to use the `bufio.Scanner` from the standard library, in order to search the log for leader election lost, guessing that maybe reading the whole stream, maybe miss newer output.

PR #7696, opened 2026-02-10
NO-JIRA: ci: fix bug in leader election detection

nunnatsa:
The `isLeaderElectionFailure` function uses the guest cluster, to read logs of management cluster pods. We always get "not found" error when trying to read these logs.

This PR replaces the guest cluster client, with the management cluster one.

Note: it seems that the leader election detection has never worked, but after merging #7589, we can see why, an it always because the pod is not found. 

## Checklist:

### DIFF
--- a/test/e2e/util/util.go
+++ b/test/e2e/util/util.go
@@ -746,17 +746,15 @@ func EnsureNoCrashingPods(t *testing.T, ctx context.Context, client crclient.Cli
 			defaultCrashToleration = 0
 		}
 
-		guestKubeConfigSecretData := WaitForGuestKubeConfig(t, ctx, client, hostedCluster)
-		guestConfig, err := clientcmd.RESTConfigFromKubeConfig(guestKubeConfigSecretData)
-		g.Expect(err).NotTo(HaveOccurred(), "couldn't load guest kubeconfig")
-		guestClient := kubeclient.NewForConfigOrDie(guestConfig)
+		cfg, err := GetConfig()
+		g.Expect(err).ToNot(HaveOccurred(), "failed to get client configuration")
+		k8sClient := kubeclient.NewForConfigOrDie(cfg)
 
 		namespace := manifests.HostedControlPlaneNamespace(hostedCluster.Namespace, hostedCluster.Name)
 
 		var podList corev1.PodList
-		if err := client.List(ctx, &podList, crclient.InNamespace(namespace)); err != nil {
-			t.Fatalf("failed to list pods in namespace %s: %v", namespace, err)
-		}
+		g.Expect(client.List(ctx, &podList, crclient.InNamespace(namespace))).To(Succeed())
+
 		for _, pod := range podList.Items {
 			crashToleration := defaultCrashToleration
 			if toleration, ok := podCrashTolerations[getComponentName(&pod)]; ok {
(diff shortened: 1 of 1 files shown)


test/e2e/util/util.go
```diff
@@ -746,17 +746,15 @@ func EnsureNoCrashingPods(t *testing.T, ctx context.Context, client crclient.Cli
 			defaultCrashToleration = 0
 		}
 
-		guestKubeConfigSecretData := WaitForGuestKubeConfig(t, ctx, client, hostedCluster)
-		guestConfig, err := clientcmd.RESTConfigFromKubeConfig(guestKubeConfigSecretData)
-		g.Expect(err).NotTo(HaveOccurred(), "couldn't load guest kubeconfig")
-		guestClient := kubeclient.NewForConfigOrDie(guestConfig)
+		cfg, err := GetConfig()
+		g.Expect(err).ToNot(HaveOccurred(), "failed to get client configuration")
+		k8sClient := kubeclient.NewForConfigOrDie(cfg)
```
devguyio, line 751: This is not needed, we already have the mgmt cluster passed in the argument `client`

test/e2e/util/util.go
```diff
@@ -746,17 +746,15 @@ func EnsureNoCrashingPods(t *testing.T, ctx context.Context, client crclient.Cli
...
 
 		namespace := manifests.HostedControlPlaneNamespace(hostedCluster.Namespace, hostedCluster.Name)
 
 		var podList corev1.PodList
-		if err := client.List(ctx, &podList, crclient.InNamespace(namespace)); err != nil {
-			t.Fatalf("failed to list pods in namespace %s: %v", namespace, err)
-		}
+		g.Expect(client.List(ctx, &podList, crclient.InNamespace(namespace))).To(Succeed())
```
devguyio, line 756: nice 👍🏽

test/e2e/util/util.go
```diff
@@ -773,7 +771,7 @@ func EnsureNoCrashingPods(t *testing.T, ctx context.Context, client crclient.Cli
 						}
 					}
 
-					if isLeaderElectionFailure(ctx, guestClient, &pod, containerStatus.Name, t) {
+					if isLeaderElectionFailure(ctx, k8sClient, &pod, containerStatus.Name, t) {
```
devguyio, line 774:
```suggestion
					if isLeaderElectionFailure(ctx, client, &pod, containerStatus.Name, t) {
```

That's the correct client.

devguyio (changes requested):
nice catch! just a quick change and this should be good! 
/cc @sjenning 
/assign @sjenning

clebs:
@nunnatsa great finding! Thanks for fixing this.

test/e2e/util/util.go
nunnatsa: @devguyio , sorry, but we do need it. The `client` variable is a controller-runtime client, while for log streaming, we need the `kubernetes.Clientset` client. The controller runtime version does not support log streaming.

test/e2e/util/util.go
nunnatsa: Thanks, but this is not accurate. see my response above.

test/e2e/util/util.go
devguyio: thank you! TIL!

PR #7697, opened 2026-02-11
GCP-295: Add GCP platform support to E2E test framework

cristianoveiga:

Adds GCP platform support to the HyperShift E2E test framework and completes the cluster creation CLI, enabling automated testing of GCP hosted clusters.

**E2E framework:**
- Register GCP-specific test flags (`--e2e.gcp-project`, `--e2e.gcp-region`, `--e2e.gcp-network`, `--e2e.gcp-psc-subnet`, `--e2e.gcp-wif-*`, `--e2e.gcp-*-sa`)
- Add `GCPPlatform` to `PlatformAgnosticOptions` and wire `ConfigurableClusterOptions` to `DefaultGCPOptions`
- Add platform-agnostic `ExpectedNodeCount` helper to fix node count calculation for non-AWS platforms
- Add `--e2e.disable-cluster-capabilities` flag for platform compatibility
- Skip unsupported tests on GCP (break-glass credentials, alertSLOs)
- Use `ExternalDNSDomain` for GCP service domain in KAS custom cert test

**Cluster creation command (`cmd/cluster/gcp/create.go`):**
- Add `--service-account-signing-key-path`, `--endpoint-access`, `--oidc-issuer-url` flags
- Add `--machine-type`, `--zone`, `--subnet`, `--boot-image` flags for NodePool configuration
- Populate `nodePool.Spec.Platform.GCP` in `GenerateNodePools` with defaults (n2-standard-4, {region}-a)
- Set default `UpgradeType` to `Replace`, consistent with all other cloud platforms
- Add external DNS route hostname support for APIServer and OAuthServer
- Konnectivity and Ignition routes are omitted — workers access these via Private Service Connect

**GCPCluster controller:**
- Preserve server-defaulted fields (Mtu, Purpose, StackType) in GCPCluster reconciliation to avoid unnecessary updates and reconciliation loops

Fixes [GCP-295](https://issues.redhat.com/browse/GCP-295)

- GCP platform is TechPreview
- This PR provides the test framework; CI infrastructure to run these tests is in [openshift/release PR #74552](https://github.com/openshift/release/pull/74552) ([GCP-401](https://issues.redhat.com/browse/GCP-401))

## Checklist:

### DIFF
--- a/hypershift-operator/controllers/hostedcluster/internal/platform/gcp/gcp_test.go
+++ b/hypershift-operator/controllers/hostedcluster/internal/platform/gcp/gcp_test.go
@@ -11,6 +11,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/utils/ptr"
 
 	capigcp "sigs.k8s.io/cluster-api-provider-gcp/api/v1beta1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -18,7 +19,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	"github.com/blang/semver"
-	"k8s.io/utils/ptr"
 )
 
 const (


test/e2e/util/gcp.go
```diff
@@ -0,0 +1,59 @@
...
+		return fmt.Errorf("GCP nodepool service account is required (--e2e.gcp-nodepool-sa)")
+	}
+	if opts.ControlPlaneServiceAccount == "" {
+		return fmt.Errorf("GCP control plane service account is required (--e2e.gcp-controlplane-sa)")
+	}
+	if opts.CloudControllerServiceAccount == "" {
+		return fmt.Errorf("GCP cloud controller service account is required (--e2e.gcp-cloudcontroller-sa)")
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
cristianoveiga: Fixed

test/e2e/util/options.go
```diff
@@ -178,6 +180,25 @@ type ConfigurableClusterOptions struct {
...
+	GCPNodePoolServiceAccount        string
+	GCPControlPlaneServiceAccount    string
+	GCPCloudControllerServiceAccount string
+	GCPServiceAccountSigningKeyPath  string
+	GCPEndpointAccess                string
+	GCPIssuerURL                     string
+	GCPMachineType                   string
+	GCPZone                          string
+	GCPSubnet                        string
+	GCPBootImage                     string
```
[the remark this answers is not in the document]
cristianoveiga: Added

cmd/cluster/gcp/create.go
```diff
@@ -213,19 +284,65 @@ func (o *CreateOptions) ApplyPlatformSpecifics(hostedCluster *hyperv1.HostedClus
...
+	if len(o.ServiceAccountSigningKeyPath) > 0 {
+		privateKey, err := os.ReadFile(o.ServiceAccountSigningKeyPath)
+		if err != nil {
+			return nil, fmt.Errorf("failed to read service account signing key file: %w", err)
+		}
+
+		saSecret := serviceAccountTokenIssuerSecret(o.namespace, o.name)
+		saSecret.Data = map[string][]byte{
+			"key": privateKey,
+		}
```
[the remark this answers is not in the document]
cristianoveiga: The constant `ServiceAccountSigningKeySecretKey = "key"` exists, but the hardcoded `"key"` string is the established pattern across all platforms — AWS uses the same approach in `cmd/cluster/aws/create.go:463`. Changing it only for GCP would be inconsistent. If this should be updated, it should be done as a separate cross-platform cleanup.

cmd/cluster/gcp/create.go
```diff
@@ -204,9 +246,42 @@ func (o *CreateOptions) ApplyPlatformSpecifics(hostedCluster *hyperv1.HostedClus
...
+				hostedCluster.Spec.Services[i].Route = &hyperv1.RoutePublishingStrategy{
+					Hostname: fmt.Sprintf("konnectivity-%s.%s", hostedCluster.Name, o.externalDNSDomain),
+				}
+			case hyperv1.Ignition:
+				hostedCluster.Spec.Services[i].Route = &hyperv1.RoutePublishingStrategy{
+					Hostname: fmt.Sprintf("ignition-%s.%s", hostedCluster.Name, o.externalDNSDomain),
+				}
+			}
+		}
+	}
```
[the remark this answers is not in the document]
cristianoveiga: AWS uses the same direct cast pattern in cmd/cluster/aws/create.go:258 without validation. Invalid values are caught by CRD validation on the API server. Keeping consistent with the established approach.

test/e2e/e2e_test.go
```diff
@@ -161,6 +162,25 @@ func TestMain(m *testing.M) {
...
+	flag.StringVar(&globalOpts.ConfigurableClusterOptions.GCPCloudControllerServiceAccount, "e2e.gcp-cloudcontroller-sa", "", "Service Account for Cloud Controller Manager")
+	flag.StringVar(&globalOpts.ConfigurableClusterOptions.GCPServiceAccountSigningKeyPath, "e2e.gcp-sa-signing-key-path", "", "Path to the private key file for the GCP service account token issuer")
+	flag.StringVar(&globalOpts.ConfigurableClusterOptions.GCPEndpointAccess, "e2e.gcp-endpoint-access", "Private", "GCP endpoint access type: Private or PublicAndPrivate")
+	flag.StringVar(&globalOpts.ConfigurableClusterOptions.GCPIssuerURL, "e2e.gcp-oidc-issuer-url", "", "The OIDC provider issuer URL for GCP")
+	flag.StringVar(&globalOpts.ConfigurableClusterOptions.GCPMachineType, "e2e.gcp-machine-type", "", "GCP machine type for node instances. Defaults to n2-standard-4")
+	flag.StringVar(&globalOpts.ConfigurableClusterOptions.GCPZone, "e2e.gcp-zone", "", "GCP zone for node instances. Defaults to {region}-a")
+	flag.StringVar(&globalOpts.ConfigurableClusterOptions.GCPSubnet, "e2e.gcp-subnet", "", "Subnet for node instances. Defaults to --private-service-connect-subnet value")
```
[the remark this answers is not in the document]
cristianoveiga: addressed

hypershift-operator/controllers/hostedcluster/internal/platform/gcp/gcp.go
```diff
@@ -126,23 +126,19 @@ func (p GCP) reconcileGCPCluster(gcpCluster *capigcp.GCPCluster, hcluster *hyper
...
-					Region: gcpSpec.Region,
-					// CAPG will discover the network CIDR automatically
-				},
+			if len(gcpCluster.Spec.Network.Subnets) == 0 {
+				gcpCluster.Spec.Network.Subnets = []capigcp.SubnetSpec{{}}
 			}
+			gcpCluster.Spec.Network.Subnets[0].Name = gcpSpec.NetworkConfig.PrivateServiceConnectSubnet.Name
+			gcpCluster.Spec.Network.Subnets[0].Region = gcpSpec.Region
 		}
 	}
```
cristianoveiga, line 143: @apahim PTAL

test/e2e/util/hypershift_framework.go
cblecker, line 1: **BUG: Private cluster detection in `before()` is AWS-only — GCP private clusters will run wrong validation path**

The `before()` method at line 150 determines private vs. public cluster validation using only `opts.AWSPlatform.EndpointAccess`:

```go
if opts.AWSPlatform.EndpointAccess == string(hyperv1.Private) {
    ValidatePrivateCluster(...)
} else {
    ValidatePublicCluster(...)
}
```

GCP defaults to `Private` endpoint access (set at `e2e_test.go:178`), but that value lives in `opts.GCPPlatform.EndpointAccess`, not the AWS field. So GCP private clusters will always be routed to `ValidatePublicCluster`, which calls `WaitForGuestClient` — a call that will hang/timeout when there is no connectivity to a private API server.

**Suggested fix:** Use `hyperutil.IsPrivateHC(hostedCluster)` from `support/util/visibility.go`, which already handles GCP correctly (lines 29-33, 59-62). This function is already imported and used elsewhere in the same file (lines 167, 253).

This was independently flagged by 4 of 9 review sub-agents (Cloud Provider SME, Data Plane SME, HCP Architect SME, SOLID review).

test/e2e/util/fixture.go
```diff
@@ -171,6 +174,14 @@ func createCluster(ctx context.Context, hc *hyperv1.HostedCluster, opts *Platfor
 		return renderCreate(ctx, &opts.RawCreateOptions, &opts.PowerVSPlatform, manifestsFile, renderLogFile, createLogFile)
 	case hyperv1.OpenStackPlatform:
 		return renderCreate(ctx, &opts.RawCreateOptions, &opts.OpenStackPlatform, manifestsFile, renderLogFile, createLogFile)
+	case hyperv1.GCPPlatform:
+		completer, err := opts.GCPPlatform.Validate(ctx, coreOpts)
+		if err != nil {
+			return fmt.Errorf("failed to validate GCP platform options: %w", err)
+		}
+		_ = completer.(*gcp.ValidatedCreateOptions)
```
cblecker: **Dead code with panic-prone bare type assertion**

The `Validate()` call is redundant since `renderCreate -> core.CreateCluster` re-runs validation internally. Additionally, the bare type assertion `completer.(*gcp.ValidatedCreateOptions)` will panic if the type doesn't match, and the result is discarded with `_ =`.

Compare with the OpenStack case (lines 175-176) which calls `renderCreate` directly without explicit validation. Or compare with AWS (line 108) where the type assertion result is actually _used_ for infra provisioning.

**Suggested fix:** Either simplify to match the OpenStack pattern:
```go
case hyperv1.GCPPlatform:
    return renderCreate(ctx, &opts.RawCreateOptions, &opts.GCPPlatform, manifestsFile, renderLogFile, createLogFile)
```

Or if this is scaffolding for future infra provisioning, add a TODO comment and use the comma-ok type assertion idiom:
```go
case hyperv1.GCPPlatform:
    // TODO: use validatedOpts for GCP infra provisioning when implemented
    completer, err := opts.GCPPlatform.Validate(ctx, coreOpts)
    if err != nil {
        return fmt.Errorf("failed to validate GCP platform options: %w", err)
    }
    if _, ok := completer.(*gcp.ValidatedCreateOptions); !ok {
        return fmt.Errorf("unexpected completer type for GCP platform")
    }
    return renderCreate(ctx, &opts.RawCreateOptions, &opts.GCPPlatform, manifestsFile, renderLogFile, createLogFile)
```

cmd/cluster/gcp/create.go
```diff
@@ -213,19 +275,86 @@ func (o *CreateOptions) ApplyPlatformSpecifics(hostedCluster *hyperv1.HostedClus
...
+		machineType = "n2-standard-4"
+	}
+	zone := o.Zone
+	if zone == "" {
+		zone = o.Region + "-a"
+	}
+	subnet := o.Subnet
+	if subnet == "" {
+		subnet = o.PrivateServiceConnectSubnet
+	}
```
cblecker, line 330: nit: GCP enforces that PSC subnets have `purpose: PRIVATE_SERVICE_CONNECT`. Worker node VMs require a subnet with `purpose: PRIVATE`. If the `--private-service-connect-subnet` flag refers to a dedicated PSC NAT subnet, worker node creation would fail here. Can you confirm this is a shared-purpose subnet in the CI infrastructure, or should `--subnet` be required when creating NodePools?

cmd/cluster/gcp/create.go
```diff
@@ -213,19 +275,86 @@ func (o *CreateOptions) ApplyPlatformSpecifics(hostedCluster *hyperv1.HostedClus
...
 func (o *CreateOptions) GenerateNodePools(constructor core.DefaultNodePoolConstructor) []*hyperv1.NodePool {
 	nodePool := constructor(hyperv1.GCPPlatform, "")
+	if nodePool.Spec.Management.UpgradeType == "" {
+		nodePool.Spec.Management.UpgradeType = hyperv1.UpgradeTypeReplace
+	}
+
+	machineType := o.MachineType
+	if machineType == "" {
+		machineType = "n2-standard-4"
+	}
```
cblecker, line 322: nit: Consider defining a named constant for this default (e.g., `const DefaultGCPMachineType = "n2-standard-4"`). It appears in code, the flag help text (line 119), and field comments (line 90).

hypershift-operator/controllers/hostedcluster/internal/platform/gcp/gcp.go
```diff
@@ -126,23 +126,19 @@ func (p GCP) reconcileGCPCluster(gcpCluster *capigcp.GCPCluster, hcluster *hyper
...
 		if gcpSpec.NetworkConfig.PrivateServiceConnectSubnet.Name != "" {
-			// CAPG handles subnets differently, configure as needed for PSC
-			// This will be used by CAPG for node creation
-			gcpCluster.Spec.Network.Subnets = []capigcp.SubnetSpec{
-				{
-					Name:   gcpSpec.NetworkConfig.PrivateServiceConnectSubnet.Name,
-					Region: gcpSpec.Region,
-					// CAPG will discover the network CIDR automatically
-				},
+			if len(gcpCluster.Spec.Network.Subnets) == 0 {
```
cblecker, line 137: The refactoring here to preserve server-defaulted fields (`Mtu`, `Purpose`, `StackType`) is the correct approach to avoid the CAPG loop detector. However, this core behavioral change lacks a regression test.

Consider adding a test that:
1. Pre-populates `gcpCluster.Spec.Network.Subnets[0]` with server-defaulted fields (e.g., `Mtu: 1460`, `Purpose: ptr.To("PRIVATE_RFC_1918")`)
2. Calls `reconcileGCPCluster`
3. Asserts that `Name` and `Region` are updated but `Mtu`, `Purpose`, `StackType` are preserved

test/e2e/e2e_test.go
```diff
@@ -161,6 +162,26 @@ func TestMain(m *testing.M) {
...
+	flag.StringVar(&globalOpts.ConfigurableClusterOptions.GCPWorkloadIdentityProviderID, "e2e.gcp-wif-provider-id", "", "Workload Identity Provider ID")
+	flag.StringVar(&globalOpts.ConfigurableClusterOptions.GCPNodePoolServiceAccount, "e2e.gcp-nodepool-sa", "", "Service Account for NodePool CAPG controllers")
+	flag.StringVar(&globalOpts.ConfigurableClusterOptions.GCPControlPlaneServiceAccount, "e2e.gcp-controlplane-sa", "", "Service Account for Control Plane Operator")
+	flag.StringVar(&globalOpts.ConfigurableClusterOptions.GCPCloudControllerServiceAccount, "e2e.gcp-cloudcontroller-sa", "", "Service Account for Cloud Controller Manager")
+	flag.StringVar(&globalOpts.ConfigurableClusterOptions.GCPStorageServiceAccount, "e2e.gcp-storage-sa", "", "Service Account for GCP PD CSI Driver")
+	flag.StringVar(&globalOpts.ConfigurableClusterOptions.GCPServiceAccountSigningKeyPath, "e2e.gcp-sa-signing-key-path", "", "Path to the private key file for the GCP service account token issuer")
+	flag.StringVar(&globalOpts.ConfigurableClusterOptions.GCPEndpointAccess, "e2e.gcp-endpoint-access", "Private", "GCP endpoint access type: Private or PublicAndPrivate")
```
cblecker: nit: Use `string(hyperv1.GCPEndpointAccessPrivate)` instead of the literal `"Private"` for type safety. The CLI code at `create.go:117` already does this correctly.

test/integration/control_plane_pki_operator.go
```diff
@@ -42,6 +42,12 @@ func RunTestControlPlanePKIOperatorBreakGlassCredentials(t *testing.T, ctx conte
 		if hostedCluster.Spec.Platform.Type == hypershiftv1beta1.KubevirtPlatform {
 			t.Skip("skipping break-glass credentials test on KubeVirt platform")
 		}
+		// TODO(GCP): Revisit. GKE Warden blocks CSRs with system: prefix in CN
+		// and system:masters in Organization. The pre-generated certs work but
+		// the CSR flow and revocation tests fail.
+		if hostedCluster.Spec.Platform.Type == hypershiftv1beta1.GCPPlatform {
+			t.Skip("skipping break-glass credentials test on GCP platform - GKE Warden blocks system: CSRs")
```
cblecker, line 49: nit: This skip is overly broad. The "direct fetch" sub-test (line 71) reads an already-issued certificate from a secret and should work on GCP since it bypasses the CSR flow that GKE Warden blocks. Consider moving the skip inside only the CSR flow and revocation sub-tests.

cmd/cluster/gcp/create.go
cristianoveiga: The --private-service-connect-subnet refers to the subnet on the customer (hosted cluster) side where the PSC endpoint is created. This is a regular subnet (not a dedicated PSC NAT subnet with purpose: PRIVATE_SERVICE_CONNECT), so it can also host worker node VMs. The fallback here is intentional - when --subnet is not explicitly provided, we reuse this subnet for the NodePool.

test/integration/control_plane_pki_operator.go
cristianoveiga: Agreed that "direct fetch" should work on GCP. However, the test structure makes it difficult to skip only the CSR flow and revocation paths cleanly, and we're not fully confident in which sub-tests would pass on GCP. Keeping the broad skip for now and will revisit once we have more clarity.

test/e2e/util/hypershift_framework.go
cristianoveiga: Fixed - replaced opts.AWSPlatform.EndpointAccess == string(hyperv1.Private) with !util.IsPublicHC(hostedCluster) which handles both AWS and GCP. Used !IsPublicHC rather than IsPrivateHC because PublicAndPrivate clusters have a reachable API server and should use ValidatePublicCluster.

test/e2e/util/fixture.go
cristianoveiga: Fixed — simplified to match the OpenStack pattern since GCP does not have an infra provisioning step. The Validate() call and bare type assertion were unused scaffolding.

cmd/cluster/gcp/create.go
cristianoveiga: Addressed — extracted defaultGCPMachineType constant and updated the flag help text and runtime default to reference it.

hypershift-operator/controllers/hostedcluster/internal/platform/gcp/gcp.go
cristianoveiga: Added TestReconcileGCPClusterPreservesServerDefaultedFields which verifies that Mtu, Purpose, and StackType are preserved after reconciliation while network name, subnet name, and region are correctly updated.

test/e2e/e2e_test.go
cristianoveiga: Fixed - now uses string(hyperv1.GCPEndpointAccessPrivate) instead of the literal "Private".

PR #7701, opened 2026-02-11
CNTRLPLANE-2801: add Key Vault private endpoint TCP relay via private router

muraee:
## Summary
- When Azure KMS etcd encryption uses a Key Vault behind a private endpoint, the `azure-kms-provider` sidecar in the KAS pod cannot resolve the vault FQDN (CoreDNS on the AKS VNet lacks the `privatelink.vaultcore.azure.net` zone).
- Adds a TCP passthrough relay via the private router (which runs on the customer VNet via Swift): HAProxy forwards Key Vault connections using SNI routing on port 8443, and KAS uses a `hostAlias` to redirect the vault hostname to the `private-router` Service ClusterIP.
- TLS passes through end-to-end — no termination, no new ports, no new Services.

## Changes
| File | Change |
|------|--------|
| `support/azureutil/azureutil.go` | Add `GetKeyVaultFQDN()` shared helper |
| `support/azureutil/azureutil_test.go` | Add unit tests for `GetKeyVaultFQDN()` |
| `.../v2/router/router_config.template` | Add `resolvers azure_dns`, keyvault SNI ACL + backend |
| `.../v2/router/config.go` | Pass Key Vault FQDN to template when `IsAroHCP()` |
| `.../v2/router/config_test.go` | Add test case verifying keyvault backend in HAProxy config |
| `.../v2/kas/deployment.go` | Add `hostAlias` mapping vault FQDN → private-router ClusterIP |

All changes are gated behind `azureutil.IsAroHCP()`.

## Test plan
- [x] Unit tests for `GetKeyVaultFQDN()` (public cloud, gov cloud, nil/empty error cases)
- [x] Unit test for router config with keyvault backend (fixture-based)
- [x] Existing tests pass (`go test ./control-plane-operator/controllers/hostedcontrolplane/...`)
- [x] `TestControlPlaneComponents` passes
- [ ] E2E: Deploy ARO HCP cluster with KMS + private Key Vault, verify etcd encryption works

## Summary by CodeRabbit

* **New Features**
  * Added Azure Key Vault access mode (Public/Private) and private-route support for Key Vault-backed secret encryption on ARO hosted control planes.
  * Integrated Key Vault backend into router and kube-apiserver configurations, plus SecretProviderClass for CSI-driven Key Vault secrets and an encryption provider sidecar.

* **Documentation**
  * API and docs updated to document keyVaultAccess field and validation rules.

* **Tests**
  * Added tests and fixtures covering private Key Vault scenarios.

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/v2/kas/deployment.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/v2/kas/deployment.go
@@ -115,7 +115,7 @@ func adaptDeployment(cpContext component.WorkloadContext, deployment *appsv1.Dep
 	// sidecar resolves the Key Vault FQDN to the private-router Service ClusterIP.
 	// The private router has access to the customer VNet (via Swift) and can reach the
 	// Key Vault's private endpoint, acting as a TCP passthrough relay.
-	if azureutil.IsAroHCP() {
+	if azureutil.IsAroHCP() && azureutil.IsPrivateKeyVault(hcp) {
 		kvFQDN, err := azureutil.GetKeyVaultFQDN(hcp)
 		if err != nil {
 			return fmt.Errorf("failed to get Key Vault FQDN for hostAlias: %w", err)
--- a/control-plane-operator/controllers/hostedcontrolplane/v2/router/config.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/v2/router/config.go
@@ -62,7 +62,7 @@ func adaptConfig(cpContext component.WorkloadContext, cm *corev1.ConfigMap) erro
 	}
 
 	var keyVaultFQDN string
-	if azureutil.IsAroHCP() {
+	if azureutil.IsAroHCP() && azureutil.IsPrivateKeyVault(cpContext.HCP) {
 		kvFQDN, err := azureutil.GetKeyVaultFQDN(cpContext.HCP)
 		if err != nil {
 			return fmt.Errorf("failed to get Key Vault FQDN: %w", err)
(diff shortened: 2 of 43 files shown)


control-plane-operator/controllers/hostedcontrolplane/v2/router/config.go
```diff
@@ -60,7 +61,16 @@ func adaptConfig(cpContext component.WorkloadContext, cm *corev1.ConfigMap) erro
 		svcsNameToIP[route.Spec.To.Name] = svc.Spec.ClusterIP
 	}
 
-	routerConfig, err := generateRouterConfig(routeList, svcsNameToIP)
+	var keyVaultFQDN string
+	if azureutil.IsAroHCP() {
```
enxebre: we should use isARO and isPrivate here so this doesn't impact our current CI job. https://github.com/openshift/hypershift/pull/7613/changes/948f703c01d4bb86143f76b5242efbeed7aa54df#diff-803aa540bd449a684ada7b96a65ae59c2f5b263f619096c98d6c05de57f8a83dR18-R21

muraee: addressed with the new isPrivateKeyVault API field

control-plane-operator/controllers/hostedcontrolplane/v2/kas/deployment.go
```diff
@@ -108,6 +111,30 @@ func adaptDeployment(cpContext component.WorkloadContext, deployment *appsv1.Dep
 	// Avoids flakiness in tests: https://issues.redhat.com/browse/OCPBUGS-62760
 	addImagePrePullInitContainers(&deployment.Spec.Template.Spec)
 
+	// When running on ARO HCP, add a hostAlias so the azure-kms-provider
+	// sidecar resolves the Key Vault FQDN to the private-router Service ClusterIP.
+	// The private router has access to the customer VNet (via Swift) and can reach the
+	// Key Vault's private endpoint, acting as a TCP passthrough relay.
+	if azureutil.IsAroHCP() {
```
enxebre, line 118: this need to check if IsPrivateHCP as well, otherwise it would break current CI which has no swift https://github.com/openshift/hypershift/blob/130436bcb77dd5237fb91e860671c50867a600b7/support/util/visibility.go#L17-L20

enxebre: I see that was addressed in next commit

enxebre:
thanks! code wise lgtm, feels a bit weird you can choose a public vault access for private clusters. As we model this into topologies, maybe this should be implicit? is there a real valid use case for a HC in aro to use a public vault, providing we don't support no swift clusters?

PR #7702, opened 2026-02-11
GCP-386: feat(gcp): enable DNS delegation for hosted cluster ingress via DNSEndpoint

patjlm:
## Summary

Enable DNS delegation for hosted cluster ingress zones on GCP using external-dns DNSEndpoint CRDs. This completes the DNS delegation chain: regional DNS zone → customer public ingress zone, making hosted cluster ingress automatically resolvable.

Resolves: [GCP-386](https://issues.redhat.com//browse/GCP-386)

## What does this PR do?

This PR implements automatic NS record delegation from regional DNS zones to hosted cluster ingress zones by:

1. **Control-plane-operator (PSC controller)**:
   - Creates DNSEndpoint resources after successful DNS zone reconciliation
   - Uses PublicIngressNSRecords from DNS setup as NS targets
   - Implements graceful error handling (CRD missing, webhooks, permissions, etc.)
   - Adds explicit cleanup in `cleanupDNS()`

2. **RBAC permissions**:
   - hypershift-operator ClusterRole: DNSEndpoint permissions for delegation
   - control-plane-operator Role: Namespace-scoped DNSEndpoint CRUD
   - external-dns ClusterRole: DNSEndpoint list/watch/status permissions

3. **External-DNS configuration**:
   - Adds `--source=crd` to enable DNSEndpoint resource processing
   - Adds `--managed-record-types` for A, AAAA, CNAME, NS records
   - Enables NS record management for zone delegation

4. **Testing**:
   - Adds 5 new unit tests covering DNSEndpoint logic and error handling
   - All existing tests pass

## Dependency

**Critical Dependency:** This PR depends on external-dns support for DNSEndpoint on Google Cloud Platform, which is **not yet available** in the current deployed version.

- **Required Fix**: https://github.com/openshift/external-dns/pull/125 (by @alebedev87)
- **Current Status**: Not yet released - **GCP-HCP project team is using a home-built image**
- **Expected Release**: Approximately March 2026 with the next OpenShift release
- **Follow-up Action**: Once released, we'll create another PR to bump the openshift-external-dns version, including CRD deployment

## How to test

1. Deploy a hosted cluster on GCP with this change
2. Verify DNSEndpoint resource is created in the management cluster namespace:
   ```bash
   oc get dnsendpoint -n clusters-{cluster-name}
   ```
3. Verify NS delegation records are created in regional DNS zone pointing to customer ingress zone nameservers
4. Verify hosted cluster ingress becomes resolvable through the delegation chain

## Error Handling

DNSEndpoint creation is best-effort. Any error (CRD missing, validation webhook, permissions, timeout) is logged but doesn't fail PSC reconciliation. The DNSEndpoint will be retried on next reconciliation.

---

cc: @alebedev87 for external-dns DNSEndpoint GCP support

### DIFF
--- a/control-plane-operator/controllers/gcpprivateserviceconnect/psc_endpoint_controller.go
+++ b/control-plane-operator/controllers/gcpprivateserviceconnect/psc_endpoint_controller.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"strings"
 	"time"
 
 	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
@@ -17,6 +18,8 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/client-go/util/workqueue"
@@ -53,6 +56,12 @@ const (
 	// out-of-band changes to GCP resources. Matches the AWS private link controller pattern.
 	driftDetectionRequeueInterval = 5 * time.Minute
 )
+
+var dnsEndpointGVK = schema.GroupVersionKind{
+	Group:   "externaldns.k8s.io",
+	Version: "v1alpha1",
+	Kind:    "DNSEndpoint",
+}
 
 // isWIFTokenAccessible checks if the WIF service account token file exists and is accessible.
 // The token is written by the token minter after the credential secret is created.
@@ -348,6 +357,24 @@ func (r *GCPPrivateServiceConnectReconciler) reconcileDNS(ctx context.Context, g
 		},
 	}
 
+	// Create DNSEndpoint for NS delegation to region DNS zone
+	// This is a best-effort operation - any failure is logged but doesn't fail the reconciliation
+	if len(dnsResult.PublicIngressNSRecords) > 0 {
+		log.Info("Creating DNSEndpoint for ingress zone delegation",
+			"dnsName", dnsResult.IngressDNSName,
+			"nameservers", dnsResult.PublicIngressNSRecords)
+
+		if err := r.reconcileDNSEndpoint(ctx, hcp, dnsResult.IngressDNSName, dnsResult.PublicIngressNSRecords); err != nil {
+			// Log the error but don't fail reconciliation
+			// DNSEndpoint creation can fail for various reasons (CRD missing, validation webhook,
+			// permissions, network issues, etc.) and shouldn't break PSC reconciliation
+			log.Info("DNSEndpoint creation failed, will retry on next reconciliation",
+				"error", err.Error())
+		} else {
+			log.Info("DNSEndpoint created successfully for ingress delegation")
+		}
+	}
+
 	// Set GCPDNSAvailable condition to True
 	meta.SetStatusCondition(&gcpPSC.Status.Conditions, metav1.Condition{
 		Type:               string(hyperv1.GCPDNSAvailable),
@@ -367,8 +394,27 @@ func (r *GCPPrivateServiceConnectReconciler) reconcileDNS(ctx context.Context, g
 	return ctrl.Result{}, nil
 }
 
-// cleanupDNS cleans up DNS zones using zone names stored in PSC status
+// cleanupDNS cleans up DNS zones and DNSEndpoint using zone names stored in PSC status
 func (r *GCPPrivateServiceConnectReconciler) cleanupDNS(ctx context.Context, gcpPSC *hyperv1.GCPPrivateServiceConnect) error {
+	log := ctrl.LoggerFrom(ctx)
+
+	// Clean up DNSEndpoint first (even if we don't have zone info)
+	hcp, err := r.getHostedControlPlane(ctx, gcpPSC)
+	if err == nil {
+		// Delete DNSEndpoint if it exists
+		dnsEndpoint := &unstructured.Unstructured{}
+		dnsEndpoint.SetGroupVersionKind(dnsEndpointGVK)
+		dnsEndpoint.SetName(dnsEndpointName(hcp.Name))
+		dnsEndpoint.SetNamespace(hcp.Namespace)
+
+		if err := r.Client.Delete(ctx, dnsEndpoint); err != nil && !apierrors.IsNotFound(err) {
+			log.Error(err, "Failed to delete DNSEndpoint during cleanup")
+			// Don't fail cleanup just because DNSEndpoint deletion failed
+		} else if err == nil {
+			log.Info("Deleted DNSEndpoint during cleanup", "name", dnsEndpoint.GetName())
+		}
+	}
+
 	// Check if we have zone information in status
 	if len(gcpPSC.Status.DNSZones) == 0 {
 		return nil // No DNS zones to clean up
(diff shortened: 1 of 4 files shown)


control-plane-operator/controllers/gcpprivateserviceconnect/psc_endpoint_controller.go
```diff
@@ -932,3 +972,94 @@ func InitCustomerGCPClient(ctx context.Context) (*compute.Service, error) {
...
+		nsTargets[i] = strings.TrimSuffix(ns, ".")
+	}
+
+	// Create DNSEndpoint as an unstructured object
+	dnsEndpoint := &unstructured.Unstructured{
+		Object: map[string]interface{}{
+			"apiVersion": "externaldns.k8s.io/v1alpha1",
+			"kind":       "DNSEndpoint",
+			"metadata": map[string]interface{}{
+				"name":      hcp.Name + "-ingress-delegation",
```
cristianoveiga: nit: This name construction (`hcp.Name + "-ingress-delegation"`) is duplicated at the `cleanupDNS` method (line 401) and in the test file (line 651). Consider extracting to a small helper to keep it consistent:

```go
func dnsEndpointName(hcpName string) string {
    return hcpName + "-ingress-delegation"
}
```

Then all three call sites use `dnsEndpointName(hcp.Name)` and the tests can reference the same function instead of reimplementing the logic inline.

control-plane-operator/controllers/gcpprivateserviceconnect/psc_endpoint_controller_test.go
```diff
@@ -528,3 +531,230 @@ func TestReconcileExternalServiceGCP(t *testing.T) {
...
+		t.Run(tt.name, func(t *testing.T) {
+			// Test the trimming logic used in reconcileDNSEndpoint
+			dnsName := tt.ingressDNS
+			if len(dnsName) > 0 && dnsName[len(dnsName)-1] == '.' {
+				dnsName = dnsName[:len(dnsName)-1]
+			}
+			assert.Equal(t, tt.expectedDNS, dnsName, tt.description)
+		})
+	}
+}
```
cristianoveiga, line 582: Question: Some of the new tests reimplement the trimming, naming, and error handling logic inline rather than calling actual code. If the implementation changes, these tests could still pass while the real code breaks. 

Maybe we should consider:
                                                                                                                                                                                                                                                                                                                                                                                                                               
1. Extract small helpers (e.g., trimDNSName, trimNameservers, dnsEndpointName) from the inline logic in reconcileDNSEndpoint and test those directly. 
2. Test reconcileDNSEndpoint directly using fake.NewClientBuilder() with an injected client.

Thoughts ?

patjlm: Good points, thanks!

I've extracted `dnsEndpointName()`, `trimDNSName()`, and `trimNameservers()` as helpers and replaced all call sites.

For the tests, I added `TestReconcileDNSEndpoint` which calls the real method with a `fake.NewClientBuilder()` client, verifying the created/updated `DNSEndpoint` has correct fields, trimmed values, and owner references. The previous tests that reimplemented the logic inline were redundant with this, so I removed them.

control-plane-operator/controllers/gcpprivateserviceconnect/psc_endpoint_controller.go
```diff
@@ -932,3 +972,105 @@ func InitCustomerGCPClient(ctx context.Context) (*compute.Service, error) {
...
+
+	// Update existing DNSEndpoint
+	log.Info("Updating existing DNSEndpoint",
+		"name", dnsEndpoint.GetName(),
+		"namespace", dnsEndpoint.GetNamespace(),
+		"dnsName", dnsName,
+		"nameservers", nameservers)
+
+	dnsEndpoint.SetResourceVersion(existing.GetResourceVersion())
+	if err := r.Client.Update(ctx, dnsEndpoint); err != nil {
```
muraee: need to check if an update is actually needed, i.e. compare with existing obj.
Otherwise, this will keep making unnecessary update API calls every reconcile loop. You should be able to reuse `upsert.ApplyProvider` to help with that

patjlm: Great catch, thanks!

I've updated `reconcileDNSEndpoint` to use the `CreateOrUpdateProvider` that's already embedded on the reconciler. This uses `equality.Semantic.DeepEqual` to compare existing vs desired state and only calls Update when something actually changed — matching the pattern already used for `ExternalName` services in this same controller (line 464).

I went with `CreateOrUpdate` rather than `ApplyProvider` since the reconciler already embeds `CreateOrUpdateProvider` and uses it elsewhere. Both seem to achieve the same goal of avoiding unnecessary updates — let me know if you'd prefer `ApplyManifest` instead.

alebedev87 (commented):
LGTM

control-plane-operator/controllers/gcpprivateserviceconnect/psc_endpoint_controller.go
```diff
@@ -932,3 +979,78 @@ func InitCustomerGCPClient(ctx context.Context) (*compute.Service, error) {
...
+	result, err := r.CreateOrUpdate(ctx, r, dnsEndpoint, func() error {
+		if err := controllerutil.SetControllerReference(hcp, dnsEndpoint, r.Scheme()); err != nil {
+			return fmt.Errorf("failed to set controller reference on DNSEndpoint: %w", err)
+		}
+		dnsEndpoint.Object["spec"] = map[string]interface{}{
+			"endpoints": []interface{}{
+				map[string]interface{}{
+					"dnsName":    dnsName,
+					"recordType": "NS",
+					"targets":    nsTargets,
```
apahim, line 1040: being nsTargets a string, I guess it will always trigger an "update" api call on every reconcile... I mean, the JSON unmarshalling will produce an `[]interface{}` for arrays, so the `Deepequal` (https://github.com/openshift/hypershift/blob/main/support/upsert/upsert.go) will always think it's different...
long story short, this should probably be an interface...
```
targets := make([]interface{}, len(nsTargets))
...
      "targets": targets,
...
```

control-plane-operator/controllers/gcpprivateserviceconnect/psc_endpoint_controller.go
```diff
@@ -932,3 +979,78 @@ func InitCustomerGCPClient(ctx context.Context) (*compute.Service, error) {
...
+					"recordTTL":  int64(300),
```
apahim: and this a `float64`, for the same reason

control-plane-operator/controllers/gcpprivateserviceconnect/psc_endpoint_controller.go
```diff
@@ -367,8 +394,28 @@ func (r *GCPPrivateServiceConnectReconciler) reconcileDNS(ctx context.Context, g
...
+// cleanupDNS cleans up DNS zones and DNSEndpoint using zone names stored in PSC status
 func (r *GCPPrivateServiceConnectReconciler) cleanupDNS(ctx context.Context, gcpPSC *hyperv1.GCPPrivateServiceConnect) error {
+	log := ctrl.LoggerFrom(ctx)
+
+	// Clean up DNSEndpoint first (even if we don't have zone info)
+	hcp, err := r.getHostedControlPlane(ctx, gcpPSC)
+	if err == nil {
+		// Delete DNSEndpoint if it exists
+		dnsEndpoint := &unstructured.Unstructured{}
+		dnsEndpoint.SetAPIVersion("externaldns.k8s.io/v1alpha1")
```
apahim: should we use the dnsEndpointGVK var here as well?

apahim:
I think the types thing is the critical one, to avoid an upsert on every loop... the reuse of the variable is just a nitpick

control-plane-operator/controllers/gcpprivateserviceconnect/psc_endpoint_controller.go
patjlm: Thanks! Updated in latest push

control-plane-operator/controllers/gcpprivateserviceconnect/psc_endpoint_controller.go
patjlm: Thanks! Updated in latest push

control-plane-operator/controllers/gcpprivateserviceconnect/psc_endpoint_controller.go
patjlm: Thanks! Updated in latest push

PR #7711, opened 2026-02-12
OCPBUGS-76645: Make generated container names k8s compliant

clebs:
On the KAS deployment pre-pull container names are generated based on the image they pull, this is sometimes not compliant with kubernetes naming rules. Names must be:
- RFC 1123 compliant (only alphanumeric characters and hyphens).
- max length of 63 characters.

This PR fixes the issue by greatly simplifying the approach:
- Only kube-apiserver and bootstrap images are pre-pulled (the rest are not needed).
- Contianer names for those images are now hardcoded and compliant with kubernetes naming.

Fixes [OCPBUGS-76645](https://issues.redhat.com/browse/OCPBUGS-76645)

## Checklist:

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/testdata/kube-apiserver/AROSwift/zz_fixture_TestControlPlaneComponents_kube_apiserver_deployment.yaml
+++ b/control-plane-operator/controllers/hostedcontrolplane/testdata/kube-apiserver/AROSwift/zz_fixture_TestControlPlaneComponents_kube_apiserver_deployment.yaml
@@ -357,37 +357,10 @@ spec:
       - command:
         - /bin/sh
         - -c
-        - 'echo ''Image pre-pulled: hyperkube''; exit 0'
+        - 'echo ''Image for kube-apiserver pre-pulled: hyperkube''; exit 0'
         image: hyperkube
         imagePullPolicy: IfNotPresent
-        name: pre-pull-image-hyperkube
-        resources: {}
-        terminationMessagePolicy: FallbackToLogsOnError
-      - command:
-        - /bin/sh
-        - -c
-        - 'echo ''Image pre-pulled: controlplane-operator''; exit 0'
-        image: controlplane-operator
-        imagePullPolicy: IfNotPresent
-        name: pre-pull-image-controlplane-operator
-        resources: {}
-        terminationMessagePolicy: FallbackToLogsOnError
-      - command:
-        - /bin/sh
-        - -c
-        - 'echo ''Image pre-pulled: apiserver-network-proxy''; exit 0'
-        image: apiserver-network-proxy
-        imagePullPolicy: IfNotPresent
-        name: pre-pull-image-apiserver-network-proxy
-        resources: {}
-        terminationMessagePolicy: FallbackToLogsOnError
-      - command:
-        - /bin/sh
-        - -c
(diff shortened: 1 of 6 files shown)


support/infraid/infraid.go
```diff
@@ -16,24 +16,13 @@ const (
...
-	re = regexp.MustCompile(`-{2,}`)
-	base = re.ReplaceAllString(base, "-")
-
-	// truncate to maxBaseLen
-	if len(base) > maxBaseLen {
-		base = base[:maxBaseLen]
-	}
-	base = strings.TrimRight(base, "-")
+	// Sanitize the base to be RFC 1123 label compliant
+	base = util.SanitizeForRFC1123Label(base, maxBaseLen)
```
heliubj18: @clebs Could you clarify why we need to modify the infra ID generation method here?

clebs: it has not been modified, I extracted the common logic into `SanitizeForRFC1123Label` so that the KAS deployment can reuse it for pre-pull container name generation.

sjenning:
/hold

This is getting way too complicated for what the initial issue was; that is "pre-pull the kube-apiserver image".

I should have picked up on this in https://github.com/openshift/hypershift/pull/7382, but we should not pre-pull all the images.  The kube-apiserver container is the only one that we want some determinism about how fast it starts.

clebs:
@sjenning you have a good point, I will take a step back and simplify this as much as possible.
In the end IIRC we only need the apiserver and bootstrap images prepulled.

clebs:
@sjenning I have applied your feedback and simplified this as much as possible. PTAL!

control-plane-operator/controllers/hostedcontrolplane/v2/kas/deployment.go
```diff
@@ -299,36 +299,28 @@ func buildKASAuditWebhookConfigFileVolume(auditWebhookRef *corev1.LocalObjectRef
 	return v
 }
 
-// addImagePrePullInitContainers adds init containers to pre-pull images for all regular containers
+// prePullContainerNames defines which containers should have their images pre-pulled for the most critical components.
+var prePullContainerNames = []string{ComponentName, "bootstrap"}
```
sjenning: We don't even need to prepull the bootstrap image.  The prompt starting of the KAS is the think we are looking to ensure.  We only need to prepull the KAS image.

clebs: I have applied the change, PTAL!

sjenning:
Very nice :+1: 
/lgtm

sjenning:
flaked on DNS
/override ci/prow/e2e-aks

clebs:
/assign @heliubj18

For verification.

sjenning:
/lgtm

https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_hypershift/7711/pull-ci-openshift-hypershift-main-e2e-aws/2023842424597516288/artifacts/e2e-aws/hypershift-aws-run-e2e-external/artifacts/TestCreateCluster/namespaces/e2e-clusters-g7fn6-create-cluster-msv6m/core/pods/logs/kube-apiserver-7655c65779-w2cww-pre-pull-image-kube-apiserver.log

PR #7713, opened 2026-02-12
OCPBUGS-60093: fix(upsert): add desired-state hash to detect spec field removals

muraee:
## Summary

- Adds a `hypershift.openshift.io/desired-state-hash` annotation containing a SHA-256 hash of the desired manifest state
- On each reconcile, compares the hash against the stored value to detect changes that `DeepDerivative` cannot detect (field removals, trailing slice items, zero-value clearing)
- Retains `DeepDerivative` as a fallback for drift detection (external modifications) and migration (objects without the annotation)

## Details

`DeepDerivative` treats nil/empty fields as "don't care", making it impossible to detect when spec fields like `nodeSelector`, `tolerations`, or container args are explicitly removed. This has been a recurring class of bugs (e.g., OCPBUGS-65879).

This replaces the earlier full-JSON `last-applied-configuration` approach which was held due to etcd size explosion at fleet scale and security concerns from embedding Secret data in annotations. A hash (~64 bytes) has none of these costs.

The existing label-count removal workaround is preserved since hash-based detection cannot drive three-way metadata merges.

## Test plan

- [x] Unit tests for hash stamping on create
- [x] Unit tests for trailing slice removal detection
- [x] Unit tests for nil/empty slice removal detection
- [x] Unit tests for idempotency (no-op on identical state)
- [x] Unit tests for migration force-stamp and subsequent no-op
- [x] Unit tests for non-self-referential hash content
- [x] Unit tests for deterministic JSON output
- [x] Unit tests for loop detector compatibility
- [x] Unit tests for `toUnstructured` field stripping
- [x] Fixture regeneration for CPO and hypershift-operator testdata
- [x] NodePool `MachineTemplateBuilders` test updated to account for hash annotation

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reconciliation for manifest-based updates by adding desired-state hashing, reducing unnecessary changes and drift when only read-only metadata differs.
  * Refined reconciliation test behavior to account for automatically stamped desired-state hashes.
  * Updated generated manifests and fixtures to better reflect real deployed ownership and tracking metadata across components and cloud providers.
  * Enhanced startup behavior for some components, including token and certificate preparation, and expanded control-plane initialization manifests.


support/upsert/apply.go
rtheis, line 1: This appears to be an extremely costly update to a large fleet that is already being stressed with etcd size explosion caused by HyperShift node pools. I don't think IBM Cloud can afford having `hypershift.openshift.io/last-applied-configuration` added. I would like to discuss with you all at our interlock meeting next week.

rtheis (changes requested):

rtheis (approved):
/lgtm

Thank you!

mgencur:
Here's my analysis of this PR:
* This requires regenerating the manifests again as Unit tests fail (`UPDATE=true go test ./...`)
* It should probably have a valid JIRA because this feature is quite substantial. It was earlier retitled to "NO-JIRA: ..." but it somehow disappeared. But I think this would deserve a JIRA.
* I ran the code review skill with overall "PASS" with recommendations. The "low" priority items seems unnecessary but there's one "medium": 
```
Gap A (Medium): No test for external drift detection via DeepDerivative fallback. The PR
  description says "Retains DeepDerivative as a fallback for drift detection (external
  modifications)" but no test exercises the scenario where hashes match yet someone externally
  modified the cluster object's spec. A test should: create with hash, externally mutate the existing
  object's spec without touching the hash annotation, reconcile, and assert OperationResultUpdated.
```
I'm adding the test in a PR that targets @muraee 's branch. PTAL: https://github.com/muraee/hypershift/pull/1
* I've generated a test plan and will create a separate "throw-away" PR shortly that will test it E2E. Link: [https://github.com/openshift/hypershift/pull/9236](https://github.com/openshift/hypershift/pull/9236)

mgencur:
/verified by mgencur

This was verified in another PR, see my comment on https://github.com/openshift/hypershift/pull/9236#issuecomment-5204524395

mgencur:
/hold

For merging the additional unit test from https://github.com/muraee/hypershift/pull/1

rtheis (approved):
/lgtm

PR #7724, opened 2026-02-13
fix(ho): make e2e HostedCluster interactions platform-aware

devguyio:

Removes hardcoded AWS platform assumptions from e2e tests so they can run correctly on non-AWS platforms (GCP, Azure, etc.).

- Add a `platformType` field to each test case group in `TestOnCreateAPIUX` so platform intent is explicit. GCP and Azure groups declare their platform; platform-agnostic groups fall back to `globalOpts.Platform`.
- Remove redundant `Platform.Type` assignments from individual `mutateInput` functions.
- The GCP skip condition now checks `tc.platformType` instead of `hostedCluster.Spec.Platform.Type`, preventing platform-agnostic tests from being incorrectly skipped on GCP environments.
- Add `numExpectedNodes` helper that computes node count based on platform instead of always multiplying by `AWSPlatform.Zones`.
- Add nil guard for `Platform.AWS` in the metrics collector to prevent nil pointer dereference on non-AWS hosted clusters.
- Fix pre-existing nil-pointer warning in `sts_test.go`.

## Why:

When running e2e tests on non-AWS environments (e.g. e2e-aks), the test framework was creating HostedCluster objects with `platform.type: AWS` hardcoded in the YAML template. The HO would then receive these HostedClusters with an AWS platform type but no `Platform.AWS` spec populated. This triggered unexpected code paths in the metrics collector, which assumed `Platform.AWS` was non-nil whenever `Platform.Type == AWS`, causing a nil pointer dereference. Additionally, the test harness unconditionally multiplied node counts by `AWSPlatform.Zones` and checked `AWSPlatform.EndpointAccess` regardless of the actual platform, leading to incorrect node expectations and wrong cluster validation paths on non-AWS runs.

Fixes

The `hostedcluster-base.yaml` asset still has `type: AWS` — this is intentional since our loop always overrides it, and changing the asset would affect `test/e2e/v2/tests/api_ux_validation_test.go` which is out of scope.## Checklist:

## Release Notes

* **Tests**
  * Improved test assertions for credential parsing validation.
  * Enhanced platform-specific test case handling for cluster creation scenarios.
  * Added helper utilities for consistent node count calculations across different cloud platforms.
  * Refined test setup to ensure accurate platform-specific configuration validation.

### DIFF
--- a/test/e2e/util/hypershift_framework.go
+++ b/test/e2e/util/hypershift_framework.go
@@ -85,6 +85,23 @@ func (h *hypershiftTest) WithAssetReader(reader assets.AssetReader) *hypershiftT
 	return h
 }
 
+// numExpectedNodes returns the number of nodes expected for a cluster based on platform.
+// On AWS and Azure, nodes are spread across zones, so the count is replicas * zones.
+// On all other platforms, the count is just replicas.
+func numExpectedNodes(opts *PlatformAgnosticOptions, platform hyperv1.PlatformType) int32 {
+	switch platform {
+	case hyperv1.AWSPlatform:
+		if len(opts.AWSPlatform.Zones) > 0 {
+			return opts.NodePoolReplicas * int32(len(opts.AWSPlatform.Zones))
+		}
+	case hyperv1.AzurePlatform:
+		if len(opts.AzurePlatform.AvailabilityZones) > 0 {
+			return opts.NodePoolReplicas * int32(len(opts.AzurePlatform.AvailabilityZones))
+		}
+	}
+	return opts.NodePoolReplicas
+}
+
 func (h *hypershiftTest) Execute(opts *PlatformAgnosticOptions, platform hyperv1.PlatformType, artifactDir, name string, serviceAccountSigningKey []byte) {
 	artifactDir = filepath.Join(artifactDir, artifactSubdirFor(h.T))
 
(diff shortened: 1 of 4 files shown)


test/e2e/util/hypershift_framework.go
```diff
@@ -84,6 +84,16 @@ func (h *hypershiftTest) WithAssetReader(reader assets.AssetReader) *hypershiftT
...
 }
 
+// numExpectedNodes returns the number of nodes expected for a cluster based on platform.
+// On AWS, nodes are spread across zones, so the count is replicas * zones.
+// On all other platforms, the count is just replicas.
+func numExpectedNodes(opts *PlatformAgnosticOptions, platform hyperv1.PlatformType) int32 {
+	if platform == hyperv1.AWSPlatform && len(opts.AWSPlatform.Zones) > 0 {
+		return opts.NodePoolReplicas * int32(len(opts.AWSPlatform.Zones))
+	}
+	return opts.NodePoolReplicas
```
devguyio, line 102: this doesn't look right, checking

test/e2e/util/hypershift_framework.go
```diff
@@ -133,7 +150,8 @@ func (h *hypershiftTest) Execute(opts *PlatformAgnosticOptions, platform hyperv1
 func (h *hypershiftTest) before(hostedCluster *hyperv1.HostedCluster, opts *PlatformAgnosticOptions, platform hyperv1.PlatformType) {
 	h.Run("ValidateHostedCluster", func(t *testing.T) {
 		if platform != hyperv1.NonePlatform && hostedCluster.Spec.Networking.NetworkType != hyperv1.Other {
-			if opts.AWSPlatform.EndpointAccess == string(hyperv1.Private) {
+			isPrivate := platform == hyperv1.AWSPlatform && opts.AWSPlatform.EndpointAccess == string(hyperv1.Private)
```
devguyio, line 154: this doesn't look right

devguyio: ah right, we don't have azure private yet.

PR #7725, opened 2026-02-13
ROSAENG-94: feat(karpenter): add Version field to OpenshiftEC2NodeClass for independent data plane versioning

joshbranham:

Add a `version` field to OpenshiftEC2NodeClassSpec that accepts a semver
string (e.g., "4.20.1"). When set, the karpenter ignition controller
resolves it to a release image by querying the Cincinnati graph API
(api.openshift.com) for the multi-arch release payload, enabling
Karpenter-managed nodes to run a different OpenShift version than the
control plane.

Key changes:
- Add `version` to spec and `releaseImage` to status in CRD types
- Add `VersionResolved` condition for observability when a `version` is set
- Create CincinnatiVersionResolver with in-memory cache (10-min TTL)
- Validate version skew against control plane using existing n-3 policy
- Preserve ignition-controller-managed status fields during ec2nodeclass status sync

Fixes [ROSAENG-94](https://issues.redhat.com/browse/ROSAENG-94)

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Specify an OpenShift version on node classes; the system resolves and records the corresponding release image and supports inheriting the control-plane image when unset.
  * Semantic version validation and version-skew checks applied; supports resolution via a configurable resolver.
* **Bug Fixes**
  * Resolved release image and VersionResolved condition are preserved across status updates.
* **Status**
  * Status now exposes the resolved release image and a VersionResolved condition reflecting resolution state.
* **Tests**
  * Added tests covering version resolution, error paths, and status updates.

### DIFF
--- a/karpenter-operator/controllers/karpenterignition/karpenterignition_controller.go
+++ b/karpenter-operator/controllers/karpenterignition/karpenterignition_controller.go
@@ -114,7 +114,7 @@ func (r *KarpenterIgnitionReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	}
 
 	releaseImage := hcp.Spec.ReleaseImage
-	version := hostedCluster.Status.Version.Desired.Version
+	version := currentClusterVersion(hostedCluster)
 
 	// When the user requests a version for the OpenshiftEC2NodeClass we perform further validation and lookup of the release image.
 	// We only detect skew and if the version is valid in this case, as under normal circumstances we can assume the release image
@@ -248,6 +248,44 @@ func (r *KarpenterIgnitionReconciler) createInMemoryNodePool(
 			Arch: hyperv1.ArchitectureAMD64, // used to find default AMI
 		},
 	}
+}
+
+// currentClusterVersion returns the version of the most recently completed update from the
+// HostedCluster's version history. It searches history entries for the one with state=Completed
+// and the most recent CompletionTime. If no completed entries exist and there is exactly one
+// history entry, it falls back to the desired version. This handles the case where a cluster
+// is still rolling out its initial version.
+func currentClusterVersion(hostedCluster *hyperv1.HostedCluster) string {
+	if hostedCluster.Status.Version == nil {
+		return ""
+	}
+
+	var latest *configv1.UpdateHistory
+	for i := range hostedCluster.Status.Version.History {
+		entry := &hostedCluster.Status.Version.History[i]
+		if entry.State != configv1.CompletedUpdate {
+			continue
+		}
+		if latest == nil {
+			latest = entry
+			continue
+		}
+		if entry.CompletionTime != nil && latest.CompletionTime != nil && entry.CompletionTime.After(latest.CompletionTime.Time) {
+			latest = entry
+		}
+	}
+
+	if latest != nil {
+		return latest.Version
+	}
+
+	// If there are no completed entries but exactly one history entry exists, the cluster
+	// is likely still rolling out its first version. Fall back to the desired version.
+	if len(hostedCluster.Status.Version.History) == 1 {
+		return hostedCluster.Status.Version.Desired.Version
+	}
+
+	return hostedCluster.Status.Version.Desired.Version
 }
 
 // validateVersion checks whether the requested version is valid for the given HostedCluster.
@@ -434,6 +472,7 @@ func hostedClusterFromHCP(hcp *hyperv1.HostedControlPlane, ignitionEndpoint stri
 
 	if hcp.Status.VersionStatus != nil {
 		hc.Status.Version.Desired.Version = hcp.Status.VersionStatus.Desired.Version
+		hc.Status.Version.History = hcp.Status.VersionStatus.History
 	}
 
 	if hcp.Spec.ControlPlaneReleaseImage != nil {
--- a/karpenter-operator/controllers/karpenterignition/karpenterignition_controller_test.go
+++ b/karpenter-operator/controllers/karpenterignition/karpenterignition_controller_test.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"strings"
 	"testing"
+	"time"
 
 	. "github.com/onsi/gomega"
 
@@ -112,6 +113,13 @@ func TestReconcile(t *testing.T) {
 			VersionStatus: &hyperv1.ClusterVersionStatus{
 				Desired: configv1.Release{
 					Version: "4.17.0",
+				},
+				History: []configv1.UpdateHistory{
+					{
+						State:          configv1.CompletedUpdate,
+						Version:        "4.17.0",
+						CompletionTime: &metav1.Time{Time: time.Now()},
+					},
 				},
 			},
 			KubeConfig: &hyperv1.KubeconfigSecretRef{
@@ -421,6 +429,13 @@ func TestReconcileVersionResolution(t *testing.T) {
 					Desired: configv1.Release{
 						Version: "4.17.0",
 					},
+					History: []configv1.UpdateHistory{
+						{
+							State:          configv1.CompletedUpdate,
+							Version:        "4.17.0",
+							CompletionTime: &metav1.Time{Time: time.Now()},
+						},
+					},
 				},
 				KubeConfig: &hyperv1.KubeconfigSecretRef{
 					Name: "admin-kubeconfig",
@@ -708,6 +723,13 @@ func TestResolveVersion(t *testing.T) {
 					Desired: configv1.Release{
 						Version: "4.17.0",
 					},
+					History: []configv1.UpdateHistory{
+						{
+							State:          configv1.CompletedUpdate,
+							Version:        "4.17.0",
+							CompletionTime: &metav1.Time{Time: time.Now()},
+						},
+					},
 				},
 			},
 		}
@@ -779,6 +801,13 @@ func TestResolveVersion(t *testing.T) {
 		hcp.Status.Version = "4.20.0"
 		hcp.Status.VersionStatus = &hyperv1.ClusterVersionStatus{
 			Desired: configv1.Release{Version: "4.20.0"},
+			History: []configv1.UpdateHistory{
+				{
+					State:          configv1.CompletedUpdate,
+					Version:        "4.20.0",
+					CompletionTime: &metav1.Time{Time: time.Now()},
+				},
+			},
 		}
 		hcp.Spec.ReleaseImage = "quay.io/openshift-release-dev/ocp-release:4.20.0-x86_64"
 		hc := baseHostedCluster(hcp)
@@ -821,6 +850,13 @@ func TestResolveVersion(t *testing.T) {
 		hcp.Status.Version = "invalid"
 		hcp.Status.VersionStatus = &hyperv1.ClusterVersionStatus{
 			Desired: configv1.Release{Version: "invalid"},
+			History: []configv1.UpdateHistory{
+				{
+					State:          configv1.CompletedUpdate,
+					Version:        "invalid",
+					CompletionTime: &metav1.Time{Time: time.Now()},
+				},
+			},
 		}
 		hc := baseHostedCluster(hcp)
 
@@ -950,6 +986,148 @@ func TestUpdateVersionStatus(t *testing.T) {
 		err = r.updateVersionStatus(ctx, nc, hcpImage, "4.17.0", nil)
 		g.Expect(err).NotTo(HaveOccurred())
 	})
+}
+
+func TestCurrentClusterVersion(t *testing.T) {
+	completedTime1 := metav1.Now()
+	completedTime2 := metav1.NewTime(completedTime1.Add(time.Hour))
+
+	testCases := []struct {
+		name            string
+		hostedCluster   *hyperv1.HostedCluster
+		expectedVersion string
+	}{
+		{
+			name: "When there is a single completed history entry it should return that version",
+			hostedCluster: &hyperv1.HostedCluster{
+				Status: hyperv1.HostedClusterStatus{
+					Version: &hyperv1.ClusterVersionStatus{
+						Desired: configv1.Release{Version: "4.17.0"},
+						History: []configv1.UpdateHistory{
+							{
+								State:          configv1.CompletedUpdate,
+								Version:        "4.17.0",
+								CompletionTime: &completedTime1,
+							},
+						},
+					},
+				},
+			},
+			expectedVersion: "4.17.0",
+		},
+		{
+			name: "When there are multiple completed entries it should return the one with the most recent CompletionTime",
+			hostedCluster: &hyperv1.HostedCluster{
+				Status: hyperv1.HostedClusterStatus{
+					Version: &hyperv1.ClusterVersionStatus{
+						Desired: configv1.Release{Version: "4.18.0"},
+						History: []configv1.UpdateHistory{
+							{
+								State:          configv1.CompletedUpdate,
+								Version:        "4.17.0",
+								CompletionTime: &completedTime1,
+							},
+							{
+								State:          configv1.CompletedUpdate,
+								Version:        "4.18.0",
+								CompletionTime: &completedTime2,
+							},
+						},
+					},
+				},
+			},
+			expectedVersion: "4.18.0",
+		},
+		{
+			name: "When there is a partial entry and a completed entry it should return the completed entry",
+			hostedCluster: &hyperv1.HostedCluster{
+				Status: hyperv1.HostedClusterStatus{
+					Version: &hyperv1.ClusterVersionStatus{
+						Desired: configv1.Release{Version: "4.18.0"},
+						History: []configv1.UpdateHistory{
+							{
+								State:   configv1.PartialUpdate,
+								Version: "4.18.0",
+							},
+							{
+								State:          configv1.CompletedUpdate,
+								Version:        "4.17.0",
+								CompletionTime: &completedTime1,
+							},
+						},
+					},
+				},
+			},
+			expectedVersion: "4.17.0",
+		},
+		{
+			name: "When there are no completed entries and exactly one history entry it should fall back to Desired.Version",
+			hostedCluster: &hyperv1.HostedCluster{
+				Status: hyperv1.HostedClusterStatus{
+					Version: &hyperv1.ClusterVersionStatus{
+						Desired: configv1.Release{Version: "4.17.0"},
+						History: []configv1.UpdateHistory{
+							{
+								State:   configv1.PartialUpdate,
+								Version: "4.17.0",
+							},
+						},
+					},
+				},
+			},
+			expectedVersion: "4.17.0",
+		},
+		{
+			name: "When there are no completed entries and multiple history entries it should fall back to Desired.Version",
+			hostedCluster: &hyperv1.HostedCluster{
+				Status: hyperv1.HostedClusterStatus{
+					Version: &hyperv1.ClusterVersionStatus{
+						Desired: configv1.Release{Version: "4.18.0"},
+						History: []configv1.UpdateHistory{
+							{
+								State:   configv1.PartialUpdate,
+								Version: "4.18.0",
+							},
+							{
+								State:   configv1.PartialUpdate,
+								Version: "4.17.0",
+							},
+						},
+					},
+				},
+			},
+			expectedVersion: "4.18.0",
+		},
+		{
+			name: "When history is empty it should fall back to Desired.Version",
+			hostedCluster: &hyperv1.HostedCluster{
+				Status: hyperv1.HostedClusterStatus{
+					Version: &hyperv1.ClusterVersionStatus{
+						Desired: configv1.Release{Version: "4.17.0"},
+						History: []configv1.UpdateHistory{},
+					},
+				},
+			},
+			expectedVersion: "4.17.0",
+		},
+		{
+			name: "When Version is nil it should return empty string",
+			hostedCluster: &hyperv1.HostedCluster{
+				Status: hyperv1.HostedClusterStatus{
+					Version: nil,
+				},
+			},
+			expectedVersion: "",
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			g := NewWithT(t)
+			version := currentClusterVersion(tc.hostedCluster)
+			g.Expect(version).To(Equal(tc.expectedVersion))
+		})
+	}
 }
 
 func findCondition(conditions []metav1.Condition, condType string) *metav1.Condition {


joshbranham:
@coderabbitai review

support/releaseinfo/cincinnati_version_resolver_test.go
```diff
@@ -0,0 +1,160 @@
...
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"sync/atomic"
+	"testing"
+	"time"
+)
+
+func TestCincinnatiVersionResolver_WhenValidResponse_ItShouldReturnReleaseImage(t *testing.T) {
```
joshbranham, line 12: I used a local test that makes a real api call vs a mock to test behavior as well. I suppose we could commit it, but felt odd having unit tests make external calls.

```
// TestCincinnatiVersionResolver_LiveAPI calls the real Cincinnati API.
// This test is for manual verification only — delete or comment out before committing.
func TestCincinnatiVersionResolver_LiveAPI(t *testing.T) {
	resolver := NewCincinnatiVersionResolver()
	version := "4.20.1"
	expectedImage := "quay.io/openshift-release-dev/ocp-release@sha256:4e903dead99fc02fc64c5f599aafe066f99c33a5b375d2fe85e85f0ca64f3aaa"

	image, err := resolver.Resolve(t.Context(), version)
	if err != nil {
		t.Fatalf("live API resolve failed: %v", err)
	}

	t.Logf("Resolved 4.20.1 -> %s", image)

	if image == "" {
		t.Fatal("expected non-empty release image")
	}

	if image != expectedImage {
		t.Fatalf("got %q, want %q", image, expectedImage)
	}
}
```

Result
```
=== RUN   TestCincinnatiVersionResolver_LiveAPI
    cincinatti_version_resolver_test.go:174: Resolved 4.20.1 -> quay.io/openshift-release-dev/ocp-release@sha256:4e903dead99fc02fc64c5f599aafe066f99c33a5b375d2fe85e85f0ca64f3aaa
--- PASS: TestCincinnatiVersionResolver_LiveAPI (0.55s)
PASS
```

jhjaggars: Mocking seems like the best choice.  At best you can expect a real api call to be flakey.

joshbranham: Totally. I just shared the test I wrote and deleted to show the logic works 👍

api/karpenter/v1beta1/karpenter_types.go
```diff
@@ -104,6 +108,21 @@ type OpenshiftEC2NodeClassSpec struct {
...
+
+	// Version is an OpenShift version (e.g., "4.20.1") specifying the release version
+	// for nodes managed by this NodeClass. When set, the controller resolves this to a
+	// release image via the Cincinnati graph API. When not set, nodes use the control plane's
+	// release image.
+	// +kubebuilder:validation:Pattern=`^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)$`
+	// +optional
+	Version *string `json:"version,omitempty"`
+
+	// Channel specifies the full Cincinnati channel (e.g., "stable-4.20", "candidate-4.20", "eus-4.16")
```
joshbranham: Question: Do we want to expose this in docs as a Cincinatti sourced channel, or should this just be `OpenShift channel`? Same comment apples to all other usages of Cincinatti.

karpenter-operator/controllers/karpenterignition/karpenterignition_controller.go
```diff
@@ -296,6 +339,88 @@ func (r *KarpenterIgnitionReconciler) updateConfigVersionAnnotation(ctx context.
 	return nil
 }
 
+// updateVersionStatus updates the OpenshiftEC2NodeClass status with the resolved release image
+// and sets the VersionResolved condition based on whether resolution succeeded.
+func (r *KarpenterIgnitionReconciler) updateVersionStatus(ctx context.Context, openshiftEC2NodeClass *hyperkarpenterv1.OpenshiftEC2NodeClass, resolvedImage string, resolveErr error) error {
+	// Only update version-related status if spec.version is set
```
maxcao13: If version is not set, the image will be set to the control plane image correct? Do we want to expose the image status anyways even if they don't set `version`?

karpenter-operator/controllers/karpenterignition/karpenterignition_controller.go
```diff
@@ -296,6 +339,88 @@ func (r *KarpenterIgnitionReconciler) updateConfigVersionAnnotation(ctx context.
...
+	for i, c := range openshiftEC2NodeClass.Status.Conditions {
+		if c.Type == hyperkarpenterv1.ConditionTypeVersionResolved {
+			openshiftEC2NodeClass.Status.Conditions[i] = condition
+			found = true
+			break
+		}
+	}
+	if !found {
+		openshiftEC2NodeClass.Status.Conditions = append(openshiftEC2NodeClass.Status.Conditions, condition)
+	}
```
maxcao13: Can we use `meta.SetStatusCondition` instead here?

karpenter-operator/controllers/karpenterignition/karpenterignition_controller.go
```diff
@@ -116,27 +129,57 @@ func (r *KarpenterIgnitionReconciler) reconcileNodeClassToken(
 	hcp *hyperv1.HostedControlPlane,
 	hostedCluster *hyperv1.HostedCluster,
 	openshiftEC2NodeClass *hyperkarpenterv1.OpenshiftEC2NodeClass,
-) error {
+) (string, error) {
```
maxcao13: Do we want to use https://github.com/openshift/hypershift/blob/409618b6aed98ae6eac6dbace6fa9c2f0ced67b7/support/supportedversion/version.go#L106 as well here? This function gets called when validating hyperv1.NodePool releaseImages, but it might be restrictive for ROSA purposes? Not sure, but just wanted to bring it up.

api/karpenter/v1beta1/karpenter_types.go
```diff
@@ -104,6 +108,21 @@ type OpenshiftEC2NodeClassSpec struct {
 	// DetailedMonitoring controls if detailed monitoring is enabled for instances that are launched
 	// +optional
 	DetailedMonitoring *bool `json:"detailedMonitoring,omitempty"`
+
+	// Version is an OpenShift version (e.g., "4.20.1") specifying the release version
+	// for nodes managed by this NodeClass. When set, the controller resolves this to a
+	// release image via the Cincinnati graph API. When not set, nodes use the control plane's
+	// release image.
+	// +kubebuilder:validation:Pattern=`^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)$`
+	// +optional
```
maxcao13, line 143: Is the UX that its going to follow the control plane, or just the initial version will be set to whatever version the control plane was at that time?

joshbranham: The intent would be that `version=""` means the `OpenshiftEC2NodeClass` tied nodes just track the control plane. We won't set this value programmatically then, it would be purely for the user to set if they want to drift the version. Additionally, setting the value to that of the control plane should result in a no-op.

karpenter-operator/controllers/karpenterignition/karpenterignition_controller.go
joshbranham: Correct. I believe exposing the resolved image in the status is helpful for cluster administrators to know what image the nodes under the nodeclass are using.

karpenter-operator/controllers/karpenterignition/karpenterignition_controller.go
joshbranham: So looks like `IsValidReleaseVersion` does some checks for common known misconfigurations, but doesn't detect skew on it's own. However, I think those checks are still valid, so added that before the skew check 👍

karpenter-operator/controllers/karpenterignition/karpenterignition_controller.go
maxcao13: Cool, from what I'm reading in the code right now it seems like if `version == ""` then, `status.ReleaseImage == ""` which is not what it seems like we are intending. From what I am reading, what is supposed to happen is if `version ==""`, then `status.ReleaseImage==controlPlaneReleaseImage` should happen, or am I misunderstanding something?

joshbranham: Yeah good point, in this case, I think given the added complexity to support this, I will omit it. If another review feels it's warranted, I can plumb it through.

api/karpenter/v1beta1/karpenter_types.go
```diff
@@ -104,6 +108,14 @@ type OpenshiftEC2NodeClassSpec struct {
...
+	Version *string `json:"version,omitempty"`
```
enxebre: this doesn't need to be a pointer unless we want to differentiate nil vs "" which is not the case.
Do we need to update the kas linter to catch this cc @JoelSpeed @bryan-cox?

karpenter-operator/controllers/karpenterignition/karpenterignition_controller.go
```diff
@@ -296,6 +357,49 @@ func (r *KarpenterIgnitionReconciler) updateConfigVersionAnnotation(ctx context.
...
+		Type:               hyperkarpenterv1.ConditionTypeVersionResolved,
+		ObservedGeneration: openshiftEC2NodeClass.Generation,
+		LastTransitionTime: metav1.Now(),
+	}
+
+	if openshiftEC2NodeClass.Spec.Version == nil {
+		condition.Status = metav1.ConditionTrue
+		condition.Reason = "VersionNotSpecified"
+		condition.Message = "No version specified, using control plane release image"
+	} else if resolveErr != nil {
```
enxebre, line 523: I find this resolveErr naming miss leading. I would appear is referring to the new resolving logic but what you are actually passing here is the returning error from reconcileNodeClassToken which involves more than the resolving version logic.
Should we decouple the resolving step from reconcileNodeClassToken and signal only that in the condition? or should we rename this variable and possibly the condition so is clear what the intent is?
I think the latter makes sense cc @maxcao13 @muraee

karpenter-operator/controllers/karpenterignition/karpenterignition_controller.go
```diff
@@ -296,6 +357,49 @@ func (r *KarpenterIgnitionReconciler) updateConfigVersionAnnotation(ctx context.
 	return nil
 }
 
+// updateVersionStatus updates the OpenshiftEC2NodeClass status with the resolved release image
+// and sets the VersionResolved condition based on whether resolution succeeded.
+func (r *KarpenterIgnitionReconciler) updateVersionStatus(ctx context.Context, openshiftEC2NodeClass *hyperkarpenterv1.OpenshiftEC2NodeClass, resolvedImage string, resolveErr error) error {
```
enxebre: can we please have a unit test for updateVersionStatus

karpenter-operator/controllers/karpenterignition/karpenterignition_controller.go
enxebre: I agree with Max, status should report consistently here so we don't degrade UX. Also this might result in stale info when changing version intent from something to ""

karpenter-operator/controllers/karpenterignition/karpenterignition_controller.go
```diff
@@ -103,40 +108,91 @@ func (r *KarpenterIgnitionReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+		); err != nil {
+			return "", fmt.Errorf("failed to validate if version %q is valid: %w", *openshiftEC2NodeClass.Spec.Version, err)
+		}
+
+		// Validate the release version does not exceed allowed skew between control plane and data plane
+		if err = supportedversion.ValidateVersionSkew(&hostedClusterVersion, &nodeClassVersion); err != nil {
+			return "", fmt.Errorf("version skew validation failed: %w", err)
+		}
+
+		resolved, err := r.VersionResolver.Resolve(ctx, *openshiftEC2NodeClass.Spec.Version, hcp.Spec.Channel)
```
enxebre: can we not tie to any specific channel?

karpenter-operator/controllers/karpenterignition/karpenterignition_controller.go
```diff
@@ -103,40 +108,91 @@ func (r *KarpenterIgnitionReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+	var resolvedImage string
+	if openshiftEC2NodeClass.Spec.Version != nil {
+		nodeClassVersion, err := semver.Parse(*openshiftEC2NodeClass.Spec.Version)
+		if err != nil {
+			return "", fmt.Errorf("failed to parse OpenshiftEC2NodeClass version %q: %w", *openshiftEC2NodeClass.Spec.Version, err)
+		}
+
+		hostedClusterVersion, err := semver.Parse(hostedCluster.Status.Version.Desired.Version)
+		if err != nil {
+			return "", fmt.Errorf("failed to parse HostedCluster version %q: %w", *openshiftEC2NodeClass.Spec.Version, err)
```
enxebre: this error message says openshiftEC2NodeClass.Spec.Version but that's not what's being parsed

karpenter-operator/controllers/karpenterignition/karpenterignition_controller.go
```diff
@@ -103,40 +108,91 @@ func (r *KarpenterIgnitionReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+) (string, error) {
 	log := ctrl.LoggerFrom(ctx).WithValues("nodeclass", openshiftEC2NodeClass.Name)
 
 	np := r.createInMemoryNodePool(hcp, openshiftEC2NodeClass)
 
-	// TODO(maxcao13): When we have a way to gate karpenter node upgrades, use that logic here instead
-	// For now, use the control plane's release image, which means control plane and data plane is still coupled.
-	pullSpec := hcp.Spec.ReleaseImage
+	// Determine the release image to use for this NodeClass.
+	// When spec.version is set, resolve it to a release image via the Cincinnati graph API
```
enxebre: can we move all this stuff about the image discovering and validation into its own unit tested function? reconciletoken already does too much

api/karpenter/v1beta1/karpenter_types.go
```diff
@@ -19,6 +19,10 @@ const (
 
 	// UserDataAMILabel is a label set in the userData secret generated for karpenter instances.
 	UserDataAMILabel = "hypershift.openshift.io/ami"
+
+	// ConditionTypeVersionResolved indicates whether the spec.version was successfully
+	// resolved to a release image via the Quay registry API.
```
enxebre: quay registry seems left over

karpenter-operator/controllers/karpenterignition/karpenterignition_controller.go
```diff
@@ -103,40 +108,91 @@ func (r *KarpenterIgnitionReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+		hostedClusterVersion, err := semver.Parse(hostedCluster.Status.Version.Desired.Version)
+		if err != nil {
+			return "", fmt.Errorf("failed to parse HostedCluster version %q: %w", *openshiftEC2NodeClass.Spec.Version, err)
+		}
+
+		// Validate the release version is valid (basic checks not including version skew)
+		minSupportedVersion := supportedversion.GetMinSupportedVersion(hostedCluster)
+		if err = supportedversion.IsValidReleaseVersion(
+			&nodeClassVersion,
+			&hostedClusterVersion,
```
enxebre: currenversion passed here, should be the curren nodeclass version not, the hc

api/karpenter/v1beta1/karpenter_types.go
```diff
@@ -242,6 +254,12 @@ type OpenshiftEC2NodeClassStatus struct {
 	// +optional
 	SecurityGroups []SecurityGroup `json:"securityGroups,omitempty"`
 
+	// ReleaseImage is the fully qualified release image resolved from spec.version,
+	// in the format "quay.io/openshift-release-dev/ocp-release@sha256:<digest>".
+	// Only set when spec.version is specified.
+	// +optional
+	ReleaseImage string `json:"releaseImage,omitempty"`
```
muraee: this should be set always, even when defaulting to the controlPlane version

karpenter-operator/controllers/karpenterignition/karpenterignition_controller.go
muraee: +1 `.status.ReleaseImage` should always be set

karpenter-operator/controllers/karpenterignition/karpenterignition_controller.go
```diff
@@ -296,6 +357,49 @@ func (r *KarpenterIgnitionReconciler) updateConfigVersionAnnotation(ctx context.
...
+		condition.Status = metav1.ConditionFalse
+		condition.Reason = "ResolutionFailed"
+		condition.Message = fmt.Sprintf("Failed to resolve version %q: %v", *openshiftEC2NodeClass.Spec.Version, resolveErr)
+	} else {
+		condition.Status = metav1.ConditionTrue
+		condition.Reason = "VersionResolved"
+		condition.Message = fmt.Sprintf("Version %q resolved to %s", *openshiftEC2NodeClass.Spec.Version, resolvedImage)
+	}
+
+	meta.SetStatusCondition(&openshiftEC2NodeClass.Status.Conditions, condition)
```
muraee: check returned value here to decide whether to do a patch or not. If the condition hasn't changes, this will return false

karpenter-operator/controllers/karpenterignition/karpenterignition_controller.go
muraee: agreed, resolving the image looks like a very different step from reconciling the token. I think these 2 steps should be separate and reported separately (possible in 2 different conditions)

api/karpenter/v1beta1/karpenter_types.go
JoelSpeed: Enable both `minlength` and `maxlength` linters, they'll force you to set appropriate limits. Without the minimum length set here, the linter cannot determine if the zero value is valid or not, so to be safe, assumes it is and tells you that it needs to be a pointer

Also make sure the pointer policy for the `optionalfields` linter is set to `WhenRequired`

bryan-cox: > Do we need to update the kas linter to catch this

We probably do. We do not have most of the default linters enabled. 

@Nirshal - just checking in, have you been able to progress your kube api linter PR further?

Nirshal: @bryan-cox not yet, its been put to lower priority for a while, due to release rotation and self-managed azure work. I can prio that if you need it, just let me know.

karpenter-operator/controllers/karpenterignition/karpenterignition_controller.go
joshbranham: We are tying this version resolution to the channel of the HCP. Is that what you mean, or were you wanting to see something different?

We also discussed adding channel as configurable, but for now, opted to exclude it so nodes and the control plane can't be on different channels.

api/karpenter/v1beta1/karpenter_types.go
joshbranham: Are you saying this should be a required field, even in a status type? Will implement 👍

api/karpenter/v1beta1/karpenter_types.go
joshbranham: Has been updated to default to `""`

karpenter-operator/controllers/karpenterignition/karpenterignition_controller.go
joshbranham: Good point, broke the logic out into two distinct steps now.

joshbranham:
> We should fix/enhance the logic in the karpenter e2e test, particularly the drift upgrade test case:
> 
> https://github.com/openshift/hypershift/blob/f3252860a237bc6e80731714ecd5f7369969289f/test/e2e/karpenter_test.go#L136
> 
> We can run only the karpenter suite with `/test e2e-aws-autonode` but we'd like to see `/test e2e-aws-techpreview` to pass altogether which is a superset of the karpenter suite.
> 
> /test e2e-aws-autonode

Makes sense, took a stab at that in https://github.com/openshift/hypershift/pull/7725/commits/d0da7fed4ef09caa48077a1b450b9015023fb0de. One piece that is interesting is how we handle setting a version different than the control plane in CI where I am not sure how the versions are fed in.

karpenter-operator/controllers/karpenterignition/karpenterignition_controller.go
joshbranham: I looked around a bit, as did claude, and it seems like in order to prevent users from trying to downgrade a OpenshiftEC2NodeClass, we would need the current version string. I think the _best_ way to do that is when we reconcile and figure out the version (wether from `version:` or from the control plane) we then store that as a status condition alongside the release image. Then we can reference that when calling `supportedversion.IsValidReleaseVersion()`. Thoughts?

karpenter-operator/controllers/karpenterignition/karpenterignition_controller.go
```diff
@@ -103,29 +110,40 @@ func (r *KarpenterIgnitionReconciler) Reconcile(ctx context.Context, req ctrl.Re
 
 	hostedCluster := hostedClusterFromHCP(hcp, r.IgnitionEndpoint)
 
-	if err := r.reconcileNodeClassToken(ctx, hcp, hostedCluster, openshiftEC2NodeClass); err != nil {
+	releaseImage, err := r.resolveVersion(ctx, hcp, hostedCluster, openshiftEC2NodeClass.Spec.Version)
```
joshbranham: This will run on every reconcile, however since the `releaseinfo.VersionResolver` comes with a 10min cache, we will at least reduce the calls that way.

@wking can you confirm something for me: will cincinatti ever _change_ the `payload` field for an existing version. For example, if I hit cincinatti with `stable-4.21` and ask for version `4.21.1`, will the response be consistent? One risk I see is if that is not true, we could unintentionally act on the changed value and upgrade worker nodes without warning.

karpenter-operator/controllers/karpenterignition/karpenterignition_controller.go
joshbranham: I did not add a new status condition for the token generation, but will defer to @maxcao13 if he wants to add that in a follow up.

api/karpenter/v1beta1/karpenter_types.go
joshbranham: I tried to make this required but it appears to break updates to the status before a `ReleaseImage` is calculated and set. Maybe I am doing something else wrong or misunderstanding the intended behavior?

JoelSpeed: > // Only set when spec.version is specified.

This doesn't sound required to me

karpenter-operator/controllers/karpenterignition/karpenterignition_controller.go
```diff
@@ -192,6 +218,57 @@ func (r *KarpenterIgnitionReconciler) createInMemoryNodePool(
...
+		nil,               // No current version — NodeClass sets a version, not an upgrade transition
+		&maxSupportedVersion,
+		&minSupportedVersion,
+		hostedCluster.Spec.Networking.NetworkType,
+		hostedCluster.Spec.Platform.Type,
+	); err != nil {
+		return "", fmt.Errorf("failed to validate if version %q is valid: %w", version, err)
+	}
+
+	if err = supportedversion.ValidateVersionSkew(&hostedClusterVersion, &nodeClassVersion); err != nil {
```
enxebre: This would leave all pools not operational if e.g. a cp upgrade causes nodes skew.  For regular nodepools we don't hard fail, we set a condition NodePoolSupportedVersionSkewConditionType. I suggest we do something similar here.

karpenter-operator/controllers/karpenterignition/karpenterignition_controller.go
enxebre: yes, we can store the semver as we store the release in status. I'm fine leaving this as it is and allowing recreate against an older version. If you violate skew that should be signaled https://github.com/openshift/hypershift/pull/7725#discussion_r2845443194

support/releaseinfo/cincinnati_version_resolver.go
```diff
@@ -0,0 +1,127 @@
+package releaseinfo
```
enxebre, line 1: this files have a typo cincinatti / cincinnati

karpenter-operator/controllers/karpenterignition/karpenterignition_controller.go
```diff
@@ -266,6 +343,11 @@ func hostedClusterFromHCP(hcp *hyperv1.HostedControlPlane, ignitionEndpoint stri
 		},
 		Status: hyperv1.HostedClusterStatus{
 			IgnitionEndpoint: ignitionEndpoint,
+			Version: &hyperv1.ClusterVersionStatus{
+				Desired: configv1.Release{
+					Version: hcp.Status.Version,
```
enxebre: why not read from versionStatus.desired.version here?

enxebre: Also for the record, what happens when this is empty, e.g new cluster? resolveVersion fails and we signal that in the condition.

enxebre:
lgtm other than
https://github.com/openshift/hypershift/pull/7725#discussion_r2845443194
https://github.com/openshift/hypershift/pull/7725#discussion_r2845455093
We'll want to see e2e-aws-autonode and e2e-aws-techpreview passing.

api/karpenter/v1beta1/karpenter_types.go
muraee: I meant the code should always set this status field. The comment should also reflect that.

JoelSpeed: If `spec.version` is not specified, what is the content of this field? I would assume it is empty and therefore omitted, right?

muraee: If spec.version is not specified, the controller will default to the HostedCluster version, which should be reflected in this status field imo.

JoelSpeed: What is enacting the defaulting of the spec? Will it always be defaulted before the controller tries to write the status? E.g. if an existing resource is read, will the spec be defaulted on clusters today?

muraee: there is no spec defaulting, the field on `OpenshiftEC2NodeClass` stays unset. The controller default to the hostedCluster version when it reconciles the `EC2NodeClass`.

JoelSpeed: Ok, so the controller will always know, by the time it's coming to write status, what the value is.
I think your godoc line might want to be tweaked as it implies this would be empty when the spec is empty, but I don't think that's what you've said in this thread

So thinking this through, if you were to make this required, then there will be a period during upgrade where the new CRD is enforced, but the old controller is leader. This controller won't be able to write status, it will be broken.

In theory, the new controller will take over and then always write this field, so it will recover.

Technically this would be a breaking change to make this required, for now it seems best to leave it as optional, even though you expect the controller to always set it

joshbranham: So is `+optional` and allowing it to default to `""` acceptable? Then when the controller reconciles the first time after resolving the image, it will update it.

karpenter-operator/controllers/karpenterignition/karpenterignition_controller.go
joshbranham: I fixed the ref to be from `VersionStatus`

karpenter-operator/controllers/karpenterignition/karpenterignition_controller.go
joshbranham: I add a new `Version` to the `OpenshiftEC2NodeClass` status type, which we can use to feed into `IsValidReleaseVersion`

api/karpenter/v1beta1/karpenter_types.go
JoelSpeed: Optional with omitempty is correct, field looks good as currently pushed

karpenter-operator/controllers/karpenterignition/karpenterignition_controller.go
wking: tl;dr: OCP (version, architecture, hashAlgorithm) tuples should reliably refer to a single release-image digest for all time.

long form:

Cincinnati is getting the version information out of the payload metadata, for example:

```console
$ oc adm release info -o json quay.io/openshift-release-dev/ocp-release:4.21.1-multi | jq '{digest, listDigest, metadata}'
{
  "digest": "sha256:9464de2e22c82939eb237b8a2f3115c972b1575c3960d054bec0de4adcb12363",
  "listDigest": "sha256:011b1362aff9e82a3b7ca720456618eaebcaf4582fda007ee9c5271d6716ef65",
  "metadata": {
    "kind": "cincinnati-metadata-v0",
    "version": "4.21.1",
    "previous": [
      "4.20.13",
      "4.20.14",
      "4.21.0"
    ],
    "metadata": {
      "release.openshift.io/architecture": "multi",
      "url": "https://access.redhat.com/errata/RHSA-2026:2129"
    }
  }
}
```

Those releases are signed, currently with both OpenPGP/GPG (at the manifest-list level) and Sigstore (at the per-arch-shard level):

```console
$ curl -s 'https://mirror.openshift.com/pub/openshift-v4/signatures/openshift/release/sha256=011b1362aff9e82a3b7ca720456618eaebcaf4582fda007ee9c5271d6716ef65/signature-1' | gpg --decrypt | jq .
gpg: Signature made Thu 05 Feb 2026 10:39:18 PM PST
gpg:                using RSA key 199E2F91FD431D51
gpg: Good signature from "Red Hat, Inc. (release key 2) <security@redhat.com>" [unknown]
gpg: WARNING: This key is not certified with a trusted signature!
gpg:          There is no indication that the signature belongs to the owner.
Primary key fingerprint: 567E 347A D004 4ADE 55BA  8A5F 199E 2F91 FD43 1D51
{
  "critical": {
    "image": {
      "docker-manifest-digest": "sha256:011b1362aff9e82a3b7ca720456618eaebcaf4582fda007ee9c5271d6716ef65"
    },
    "type": "atomic container signature",
    "identity": {
      "docker-reference": "quay.io/openshift-release-dev/ocp-release:4.21.1-multi"
    }
  },
  "optional": {
    "creator": "Red Hat OpenShift Signing Authority 0.0.1"
  }
}
$ oc image info -o json quay.io/openshift-release-dev/ocp-release:sha256-9464de2e22c82939eb237b8a2f3115c972b1575c3960d054bec0de4adcb12363.sig | grep meiaType
  "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "mediaType": "application/vnd.dev.cosign.simplesigning.v1+json",
      "mediaType": "application/vnd.dev.cosign.simplesigning.v1+json",
      "mediaType": "application/vnd.dev.cosign.simplesigning.v1+json",
      "mediaType": "application/vnd.dev.cosign.simplesigning.v1+json",
```

ART (building OCP releases) should never publish multiple signatures for a given (version, architecture, hashAlgorithm) tuple.  Cincinnati is just passing along the ART-published information, and you can reach around Cincinnati and look at the ART-published signatures if you want to guard against Cincinnati misbehavior.

And you have the claimed (version, architecture, hashAlgorithm) in the release payload itself, if you want to no-op or otherwise guard against ART misbehavior if a pullspec you're seing diverges from a pullspec you previously saw for that (version, architecture, hashAlgorithm).

And currently, all OCP releases use `sha256` as a hash algorithm.  I'm including `hashAlgorithm` in my tuples here, because there's a chance that future OCP ship both `sha256`- and `sha512`-hashed releases in parallel.  Or possibly there will be one OCP release that's dual-hashed somehow (labels on digest references with alternative-algorithm hashes?).  It's not clear at the moment.  If that does happen, Cincinnati is likely to have separate channels, or a new query parameter, or distinct versoin names, or similar for each hash, because the Cincinnati/ClusterVersion format doesn't allow for clear distinctions between multiple entries claiming the same version name.

karpenter-operator/controllers/karpenterignition/karpenterignition_controller.go
```diff
@@ -296,6 +408,80 @@ func (r *KarpenterIgnitionReconciler) updateConfigVersionAnnotation(ctx context.
...
+		condition.Status = metav1.ConditionTrue
+		condition.Reason = "VersionNotSpecified"
+		condition.Message = "No version specified, using control plane release image"
+	} else if resolveErr != nil {
+		condition.Status = metav1.ConditionFalse
+		condition.Reason = "ResolutionFailed"
+		condition.Message = fmt.Sprintf("Failed to resolve version %q: %v", openshiftEC2NodeClass.Spec.Version, resolveErr)
+	} else {
+		condition.Status = metav1.ConditionTrue
+		condition.Reason = "VersionResolved"
```
enxebre: nit: can we move this reasons as constants at the api level file

karpenter-operator/controllers/karpenterignition/karpenterignition_controller.go
```diff
@@ -192,6 +230,75 @@ func (r *KarpenterIgnitionReconciler) createInMemoryNodePool(
...
+		currentVersion,
+		&maxSupportedVersion,
+		&minSupportedVersion,
+		hostedCluster.Spec.Networking.NetworkType,
+		hostedCluster.Spec.Platform.Type,
+	); err != nil {
+		return "", nil, fmt.Errorf("failed to validate if version %q is valid: %w", version, err)
+	}
+
+	skewErr = supportedversion.ValidateVersionSkew(&hostedClusterVersion, &nodeClassVersion)
```
enxebre: I'd rather keep this func scoped to resolve the version and invoke this ValidateVersionSkew outside, right after updateVersionSkewStatus

karpenter-operator/controllers/karpenterignition/karpenterignition_controller.go
```diff
@@ -192,6 +230,75 @@ func (r *KarpenterIgnitionReconciler) createInMemoryNodePool(
...
+// currentStatusVersion is the NodeClass's current status.version, used to detect y-stream downgrades.
+func (r *KarpenterIgnitionReconciler) resolveVersion(
+	ctx context.Context,
+	hcp *hyperv1.HostedControlPlane,
+	hostedCluster *hyperv1.HostedCluster,
+	version string,
+	currentStatusVersion string,
+) (releaseImage string, skewErr error, err error) {
+	if version == "" {
+		return hcp.Spec.ReleaseImage, nil, nil
```
enxebre: this should technically resolve to whatever the hcp has in the state update history as last completed upgrade, otherwise there's room for a node to break policy with the kas.

karpenter-operator/controllers/karpenterignition/karpenterignition_controller.go
```diff
@@ -266,6 +373,11 @@ func hostedClusterFromHCP(hcp *hyperv1.HostedControlPlane, ignitionEndpoint stri
 		},
 		Status: hyperv1.HostedClusterStatus{
 			IgnitionEndpoint: ignitionEndpoint,
+			Version: &hyperv1.ClusterVersionStatus{
+				Desired: configv1.Release{
+					Version: hcp.Status.VersionStatus.Desired.Version,
```
enxebre: This pointer dereference needs a nil check or it might panic before this is set

karpenter-operator/controllers/nodeclass/ec2_nodeclass_controller.go
```diff
@@ -321,7 +321,19 @@ func (r *EC2NodeClassReconciler) reconcileStatus(ctx context.Context, ec2NodeCla
 
 	originalObj := openshiftNodeClass.DeepCopy()
 
+	// Preserve fields managed by the ignition controller before resetting status
```
enxebre: instead of hardcoding this we can stop resetting the entire status struct and only update the fields owned by the ec2nodeclass controller and properly merge conditions.

enxebre: at minimum let's please add unit test coverage for reconcileStatus

support/releaseinfo/cincinnati_version_resolver.go
```diff
@@ -0,0 +1,127 @@
...
+	"fmt"
+	"io"
+	"net/http"
+	"sync"
+	"time"
+)
+
+const (
+	defaultArch          = "multi"
+	defaultCincinnatiURL = "https://api.openshift.com/api/upgrades_info/v1/graph"
```
enxebre, line 15: what's the story here for disconnected environments?

enxebre: orthogonally we'll want to inject proxy.SetEnvVars into the karpenter operator

karpenter-operator/controllers/nodeclass/ec2_nodeclass_controller.go
muraee: +1 
`openshiftNodeClass.Status = hyperkarpenterv1.OpenshiftEC2NodeClassStatus{}` this line should be removed

karpenter-operator/controllers/karpenterignition/karpenterignition_controller.go
```diff
@@ -192,6 +230,75 @@ func (r *KarpenterIgnitionReconciler) createInMemoryNodePool(
...
+// The returned skewErr is non-nil when the version falls outside the supported skew policy;
+// this is reported as a status condition but does not block reconciliation.
+// currentStatusVersion is the NodeClass's current status.version, used to detect y-stream downgrades.
+func (r *KarpenterIgnitionReconciler) resolveVersion(
+	ctx context.Context,
+	hcp *hyperv1.HostedControlPlane,
+	hostedCluster *hyperv1.HostedCluster,
+	version string,
+	currentStatusVersion string,
+) (releaseImage string, skewErr error, err error) {
```
muraee: this function is very overloaded and confusing. It should only be resolving the image and be rename to something like `resolveReleaseImage(version)`

karpenter-operator/controllers/karpenterignition/karpenterignition_controller.go
```diff
@@ -103,29 +110,59 @@ func (r *KarpenterIgnitionReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+	// Determine the resolved version string for status.version.
+	// When spec.version is set, it is the version. When not set, use the HCP's version.
+	resolvedVersion := openshiftEC2NodeClass.Spec.Version
+	if resolvedVersion == "" {
+		resolvedVersion = hostedCluster.Status.Version.Desired.Version
+	}
+
+	if openshiftEC2NodeClass.Spec.Version != "" {
+		log.Info("Resolved version to release image", "version", openshiftEC2NodeClass.Spec.Version, "channel", hcp.Spec.Channel, "releaseImage", releaseImage)
+	}
```
muraee: something like this could be easier to follow:
```go
releaseImage := hcp.Spec.ReleaseImage
version :=  hostedCluster.Status.Version.Desired.Version
if openshiftEC2NodeClass.Spec.Version != "" {
  releaseImage = resolveReleaseImage(version)
  version := openshiftEC2NodeClass.Spec.Version

}

supportedversion.IsValidReleaseVersion()
validateVersionSkew()  
...
```

karpenter-operator/controllers/karpenterignition/karpenterignition_controller.go
muraee: see my other comment https://github.com/openshift/hypershift/pull/7725/changes#r2863407154

karpenter-operator/controllers/karpenterignition/karpenterignition_controller.go
joshbranham: This brings up a good point. The default behavior should be that the nodeclass upgrade is triggered only after a control plane has been upgraded (when the user doesn't explicitly set a version.)

What is the canonical place to get the version and release image that can confidently signal that the control plane is on that version and using that release image?

support/releaseinfo/cincinnati_version_resolver.go
joshbranham: How do we handle release payload pulling in Hypershift for disconnected environments? I presume we have more than one place where we make an external call that could fail.

joshbranham: Opened a standalone PR injecting the proxy env vars for the operator and karpenter itself, ptal https://github.com/openshift/hypershift/pull/7831

karpenter-operator/controllers/karpenterignition/karpenterignition_controller.go
joshbranham: I think what you are suggesting is we should enumerate `hcp.Status.VersionStatus.UpdateHistory` and find the one with the most recent `CompletionTime` and use that version as the default for non-versioned `OpenshiftEC2NodeClass`?

joshbranham: Covered in https://github.com/openshift/hypershift/pull/7725/commits/de553b20a26d7b8e8c7f9c59965aba628e9c9586

joshbranham:
This is ready for a final review and merge, I believe all feedback has been addressed and e2e tests are passing with the new logic.

karpenter-operator/controllers/karpenterignition/karpenterignition_controller.go
```diff
@@ -192,6 +250,143 @@ func (r *KarpenterIgnitionReconciler) createInMemoryNodePool(
 	}
 }
 
+// currentClusterVersion returns the version of the most recently completed update from the
+// HostedCluster's version history. It searches history entries for the one with state=Completed
+// and the most recent CompletionTime. If no completed entries exist and there is exactly one
+// history entry, it falls back to the desired version. This handles the case where a cluster
+// is still rolling out its initial version.
+func currentClusterVersion(hostedCluster *hyperv1.HostedCluster) string {
```
enxebre, line 258: can we please unit test this func?

karpenter-operator/controllers/karpenterignition/karpenterignition_controller.go
```diff
@@ -192,6 +250,143 @@ func (r *KarpenterIgnitionReconciler) createInMemoryNodePool(
...
+	}
+
+	return hostedCluster.Status.Version.Desired.Version
+}
+
+// validateVersion checks whether the requested version is valid for the given HostedCluster.
+// It validates semver parsing, range bounds, and y-stream downgrade detection.
+// currentStatusVersion is the NodeClass's current status.version, used to detect y-stream downgrades.
+// Returns nil when version is empty (nodes use the control plane image).
+func validateVersion(hostedCluster *hyperv1.HostedCluster, version string, currentStatusVersion string) error {
```
enxebre, line 295: can we please unit test this func

karpenter-operator/controllers/karpenterignition/karpenterignition_controller.go
joshbranham: This is tested by nature of the full public func `Reconcile` being in unit tests. I can add a standalone unit test if you'd like? Do you want to see that in this PR or can it be in a follow up?

karpenter-operator/controllers/karpenterignition/karpenterignition_controller.go
joshbranham: This is tested by nature of the full public func `Reconcile` being in unit tests. I can add a standalone unit test if you'd like? Do you want to see that in this PR or can it be in a follow up?

jkyros:
The failure modes here all seem acceptable. If cincinatti is down we can't provision new "pinned" nodes at `spec.version` but we if we removed the version we could still provision nodes at the controlPlane's version. 

Also, we can deal with this later, but are we concerned at all about the "condition aggregation" here? Like, for example, say:
-  `spec.version` is set on OpenShiftEC2Nodeclass and  
- we fail to resolve a version (Cincinnati is dead or something)

Presumably:
- ec2nodeclass will be `Ready=True` (which is fine, we don't control that) 
- OpenShiftEC2Nodeclass will have `VersionResolved=False` but
- OpenShiftEC2Nodeclass will still be`Ready=True`

So I could see a user who is wondering why they aren't getting any nodes being confused. Like I think they'd eventually stumble upon `VersionResolved=False` but the `Ready=True` would seem to indicate that that version resolution failure would be non-fatal even though it is totally preventing us from creating new nodes? 

/lgtm

PR #7726, opened 2026-02-15
CNTRLPLANE-2216: feat(aws): migrate route53 to AWS SDK v2

LiangquanLi930:
migrate route53 to AWS SDK v2

Fixes [CNTRLPLANE-2216](https://issues.redhat.com/browse/CNTRLPLANE-2216)

## Checklist:


cmd/infra/aws/route53.go
```diff
@@ -378,9 +378,6 @@ func retryRoute53WithBackoff(ctx context.Context, fn func() error) error {
 		Factor:   1.5,
 	}
 	retriable := func(e error) bool {
-		if !awsutil.IsErrorRetryable(e) {
```
sdminonne, line 381: Any reason this is removed? Isn't there an equivalent v2?

control-plane-operator/controllers/awsprivatelink/route53.go
```diff
@@ -94,47 +106,41 @@ func fqdn(name string) string {
...
-		return false
-	})
-
+	resp, err := client.ListResourceRecordSets(ctx, input)
 	if err != nil {
 		return nil, err
 	}
-	return record, nil
+	if len(resp.ResourceRecordSets) == 0 {
+		return nil, nil
```
sdminonne, line 123: @LiangquanLi930 may you explain me why we're not returning resp.ResourceRecordSets ?

support/awsapi/route53.go
```diff
@@ -0,0 +1,34 @@
+// Code generated by delegatingclientgenerator. DO NOT EDIT.
+
+package awsapi
+
+//go:generate ../../hack/tools/bin/mockgen -source=route53.go -package=awsapi -destination=route53_mock.go
```
sdminonne, line 5: We've this but I couldn't find unit tests. Am I missing anything?

sdminonne (approved):
Nice! I've a couple of questions though.

cmd/infra/aws/util/util.go
```diff
@@ -162,15 +162,14 @@ func NewSessionV2(ctx context.Context, agent, credentialsFile, credKey, credSecr
...
-	awsRoute53Config := NewConfig()
-	awsRoute53Config.Retryer = client.DefaultRetryer{
-		NumMaxRetries:    10,
-		MinRetryDelay:    5 * time.Second,
-		MinThrottleDelay: 10 * time.Second,
+// NewRoute53ConfigV2 creates a v2 retryer with more conservative retry timing for Route53.
+func NewRoute53ConfigV2() func() awsv2.Retryer {
+	return func() awsv2.Retryer {
+		return retry.NewStandard(func(o *retry.StandardOptions) {
+			o.MaxAttempts = 21                                              // 1 initial + 20 retries
```
jparrill, line 169: Is this arbitrary or is there any logic under the hood?

cmd/infra/aws/route53.go
```diff
@@ -7,19 +7,19 @@ import (
...
+	route53types "github.com/aws/aws-sdk-go-v2/service/route53/types"
 
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/util/retry"
 
 	"github.com/go-logr/logr"
 )
 
-func (o *CreateInfraOptions) LookupPublicZone(ctx context.Context, logger logr.Logger, client route53iface.Route53API) (string, error) {
+func (o *CreateInfraOptions) LookupPublicZone(ctx context.Context, logger logr.Logger, client awsapi.ROUTE53API) (string, error) {
```
jparrill, line 22: Do we have plans to make integration tests on these functions?
- LookupPublicZone
- LookupZone
- CreatePrivateZone
- DestroyDNS
- DestroyPrivateZones
- CleanupPublicZone

support/awsapi/route53.go
jparrill: Good catch!

jparrill (commented):

cmd/infra/aws/route53.go
LiangquanLi930: IsErrorRetryable guards against retrying v1 `awserr.Error` with code `SharedCredsLoad`. In v2, credential resolution happens at `config.LoadDefaultConfig()` time before any API call, so credential load errors never reach this retry loop. No v2 equivalent is needed.

control-plane-operator/controllers/awsprivatelink/route53.go
LiangquanLi930: This piece of code is at line 136 in the new file, so this should just be a display issue.

support/awsapi/route53.go
LiangquanLi930: You're right, there are no unit tests using the Route53 mock yet. This is the same as v1 — Route53 operations were only covered by e2e tests. The mock (route53_mock.go) is generated and ready to use.

cmd/infra/aws/route53.go
LiangquanLi930: Adding dedicated unit tests using the generated MockROUTE53API is a good follow-up but out of scope for this SDK migration PR. We could track it as a separate card.

open: https://issues.redhat.com/browse/CNTRLPLANE-2077

cmd/infra/aws/util/util.go
LiangquanLi930: Not arbitrary. In v1, NumMaxRetries was 10 with a 5s base delay and no backoff cap override (so it defaulted to 300s), resulting in a worst-case total wait of ~25 minutes. In v2, we use more retries (21) but a much lower backoff cap (30s), bringing the worst-case down to ~8 minutes — more retries with faster feedback when hitting Route53 throttling. WDYT?

Or should I tweak these values to something else?

jparrill: Looks good, the only thing I can think of is to document this somehow (maybe a comment in the code should be sufficient)

cmd/infra/aws/route53.go
jparrill: Being a follow up card, it's Ok

cmd/infra/aws/route53.go
sdminonne: TIL! Greant thanks

support/awsapi/route53.go
sdminonne: OK I won't block on this but even if ready to use it can be misleading... May you ask to claude in a follow-up to generate some unit tests? TY!

LiangquanLi930: https://github.com/openshift/hypershift/pull/7726#discussion_r2813103321
Is ToDO list

control-plane-operator/controllers/awsprivatelink/awsprivatelink_controller.go
```diff
@@ -229,7 +236,12 @@ func (b *clientBuilder) getClients() (ec2iface.EC2API, route53iface.Route53API,
 	if err != nil {
 		return nil, nil, err
 	}
-	awsRoute53Session, err := b.awsSession()
+	ctx := context.Background()
```
enxebre: this context need to be passed in the func signature

PR #7731, opened 2026-02-17
GCP-367: enable multizone support in CCM cloud config

cristianoveiga:

The GCP CCM discovers its zone from the GCE metadata server, which returns the management cluster node's zone. In HCP, this causes the CCM to only search that zone for guest cluster worker nodes, failing to find nodes in other zones within the same region.

Add multizone = true to the cloud config so the CCM searches all zones in the region for instances.

Fixes [GCP-367](https://issues.redhat.com/browse/GCP-367)

## Checklist:


cristianoveiga:
@coderabbitai review

cblecker (approved):

PR #7744, opened 2026-02-18
OCPBUGS-62177: fix(pki): use pod container port instead of kubeconfig service port

enxebre:
## Summary

- Fix certificate revocation controller to extract the KAS port from the pod container spec instead of the kubeconfig service URL
- The kubeconfig contains the service port which can differ from the container port (e.g. IBMCloud uses service port 2040 while the container always listens on 6443)
- When connecting directly to pod IPs, the container port must be used
- Add `containerPort()` helper with fallback to `KASPodDefaultPort` (6443)
- Add unit tests for the new helper and populate existing test pods with container port specs

## Test plan

- [x] Unit tests pass (`go test ./control-plane-pki-operator/certificaterevocationcontroller/...`)
- [ ] Verify certificate revocation flow works on IBMCloud clusters with service port 2040


sdminonne (approved):
Thanks Alberto

cblecker:
@coderabbit full review

PR #7749, opened 2026-02-19
CNTRLPLANE-2775: Expose KAS availability and latency metrics from the control-plane-operator

enxebre:
## Description

Instruments the existing `healthCheckKASEndpoint()` function in the control-plane-operator to expose two new Prometheus metrics for KAS health monitoring:

| Metric | Type | Description |
|--------|------|-------------|
| `hypershift_kube_apiserver_available` | Gauge | 1 if `/healthz` returns HTTP 200, 0 otherwise |
| `hypershift_kube_apiserver_request_duration_seconds` | Histogram | Latency of the `/healthz` probe (buckets: 0.01–10s) |

### Why

HCP offerings (ROSA HCP, ARO HCP) need to monitor customer API endpoint availability and latency for SLA purposes. ROSA HCP currently relies on an external tool ([route-monitor-operator](https://github.com/openshift/route-monitor-operator)) solely for this. These native metrics eliminate that dependency.

### How

- Metrics are registered with the controller-runtime metrics registry and automatically scraped by the existing PodMonitor for the CPO — **no new monitoring infrastructure required**
- Each CPO pod runs in its own HCP namespace, so metrics are naturally scoped per hosted cluster
- The existing `HostedControlPlaneAvailable` condition logic is unchanged — metrics are a side-effect, not a replacement
- Works across all endpoint topologies: private clusters, public with Route, public with LoadBalancer, shared ingress (ARO HCP / ROSA HCP)

### Key files

- `control-plane-operator/controllers/hostedcontrolplane/kas/metrics.go` — metric definitions and registration
- `control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go` — instrumented health check
- `control-plane-operator/main.go` — metrics initialization

## Testing

- Unit tests verify gauge and histogram are correctly set for success (200), failure (503), and unreachable scenarios
- E2e test validates metric presence on the CPO pod using `GetMetricsFromPod`/`ValidateMetricPresence` (Karpenter pattern)
- All existing tests pass (`make test` exits 0)

## Jira

[CNTRLPLANE-2775](https://issues.redhat.com/browse/CNTRLPLANE-2775)

🤖 Generated with [Claude Code](https://claude.com/claude-code) via `/jira:solve [CNTRLPLANE-2775](https://issues.redhat.com/browse/CNTRLPLANE-2775)`

## Summary by CodeRabbit

* **New Features**
  * Added KAS health metrics (availability gauge and request-duration histogram) and registered them for collection.

* **Tests**
  * Added unit tests for KAS health metrics and end-to-end validation to check metrics are emitted during runs.

* **Chores**
  * Metrics instrumentation initialized at startup so controller exposes KAS health metrics for monitoring.

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
@@ -180,6 +180,7 @@ type HostedControlPlaneReconciler struct {
 	OperateOnReleaseImage                   string
 	DefaultIngressDomain                    string
 	MetricsSet                              metrics.MetricsSet
+	KASHealthMetrics                        *kas.KASHealthMetrics
 	SREConfigHash                           string
 	ec2Client                               awsapi.EC2API
 	awsSession                              *aws.Config
@@ -944,9 +945,9 @@ func (r *HostedControlPlaneReconciler) healthCheckKASLoadBalancers(ctx context.C
 		// When the cluster is private, checking the load balancers will depend on whether the load balancer is
 		// using the right subnets. To avoid uncertainty, we'll limit the check to the service endpoint.
 		if hcp.Spec.Platform.Type == hyperv1.IBMCloudPlatform {
-			return healthCheckKASEndpoint(manifests.KubeAPIServerService("").Name, config.KASSVCIBMCloudPort)
+			return healthCheckKASEndpoint(manifests.KubeAPIServerService("").Name, config.KASSVCIBMCloudPort, r.KASHealthMetrics)
 		}
-		return healthCheckKASEndpoint(manifests.KubeAPIServerService("").Name, config.KASSVCPort)
+		return healthCheckKASEndpoint(manifests.KubeAPIServerService("").Name, config.KASSVCPort, r.KASHealthMetrics)
 	case serviceStrategy.Type == hyperv1.Route:
 		if hcp.Spec.Platform.Type != hyperv1.IBMCloudPlatform {
 			externalRoute := manifests.KubeAPIServerExternalPublicRoute(hcp.Namespace)
@@ -958,7 +959,7 @@ func (r *HostedControlPlaneReconciler) healthCheckKASLoadBalancers(ctx context.C
 			if err != nil {
 				return err
 			}
-			return healthCheckKASEndpoint(endpoint, port)
+			return healthCheckKASEndpoint(endpoint, port, r.KASHealthMetrics)
 		}
 	case serviceStrategy.Type == hyperv1.LoadBalancer:
 		svc := manifests.KubeAPIServerService(hcp.Namespace)
@@ -987,17 +988,33 @@ func (r *HostedControlPlaneReconciler) healthCheckKASLoadBalancers(ctx context.C
 		} else if LBIngress.IP != "" {
 			ingressPoint = LBIngress.IP
 		}
-		return healthCheckKASEndpoint(ingressPoint, port)
+		return healthCheckKASEndpoint(ingressPoint, port, r.KASHealthMetrics)
 	}
 	return nil
 }
 
-func healthCheckKASEndpoint(ingressPoint string, port int) error {
+func healthCheckKASEndpoint(ingressPoint string, port int, m *kas.KASHealthMetrics) error {
 	healthEndpoint := fmt.Sprintf("https://%s:%d/healthz", ingressPoint, port)
 
 	httpClient := util.InsecureHTTPClient()
 	httpClient.Timeout = 10 * time.Second
+
+	start := time.Now()
 	resp, err := httpClient.Get(healthEndpoint)
+	duration := time.Since(start).Seconds()
+	if resp != nil {
+		defer resp.Body.Close()
+	}
+
+	if m != nil {
+		m.RequestDuration.Observe(duration)
+		if err == nil && resp != nil && resp.StatusCode == http.StatusOK {
+			m.Available.Set(1)
+		} else {
+			m.Available.Set(0)
+		}
+	}
+
 	if err != nil {
 		return err
 	}
--- a/control-plane-operator/controllers/hostedcontrolplane/kas/metrics.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/kas/metrics.go
@@ -1,0 +1,42 @@
+package kas
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const (
+	KASAvailableMetricName       = "hypershift_kube_apiserver_available"
+	KASRequestDurationMetricName = "hypershift_kube_apiserver_request_duration_seconds"
+)
+
+// kasRequestDurationBuckets defines the histogram bucket boundaries for KAS
+// health check latency measurements, ranging from 10ms to 10s.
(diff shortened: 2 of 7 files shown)


typeid:
Just to clarify a bit further, while these metrics are great, in ROSA HCP we intend to probe KAS availability externally via RHOBS synthetic monitoring ([SREP-333](https://issues.redhat.com/browse/SREP-333)), where Blackbox Exporter runs on RHOBS cells outside the management cluster. This gives us the advantage of testing the actual customer-facing network path (only partially for private API), including DNS resolution, load balancer health, and regional routing, rather than probing from within the MC's own network.

I understand ARO HCP wants to avoid the RMO dependency, and these metrics help with that. However, since the CPO probe originates from within the MC, it's not a full replacement for external synthetic monitoring for SLA purposes IMO.

That said, these CPO-local metrics are definitely useful even for the ROSA side for faster internal detection of control plane issues, for example catching KAS pod crashes or pinpointing network issues to in-cluster networking failures. 

LGTM & thanks for the addition!

typeid:
Also cc @dustman9000 as a FYI that this exists and is now being extended with latency as well :)

muraee:
lgtm

dustman9000:
This is a great addition! Having a native KAS health signal per HCP from inside the CPO will be very useful.

One thing I want to flag from the ROSA HCP SRE monitoring side: the probe frequency here is tied to the reconcile loop (1 min when healthy, 15s when unhealthy), and the probe runs from inside the management cluster. We currently have two layers of external synthetic monitoring that validate the customer-facing network path (DNS, LB, ingress):

- route-monitor-operator: creates Prometheus blackbox exporter probes per HCP endpoint on the management cluster
- RHOBS synthetics agent: external blackbox probes monitoring HCP API endpoints from outside the MC, feeding into RHOBS for SLO tracking

Both probe at a higher frequency and validate reachability from outside the control plane. For SLO/SLA calculations like KubeAPIErrorBudgetBurn, that external perspective and tighter sampling interval matter.

So this is complementary to our synthetic monitoring, not a replacement. If we can, I'd suggest softening the "eliminate that dependency" language in the description to avoid confusion downstream. Something like "these native metrics reduce reliance on external probing for internal health checks" would be more accurate.

control-plane-operator/controllers/hostedcontrolplane/kas/metrics.go
```diff
@@ -0,0 +1,44 @@
...
+			Help: "1 if the KAS /healthz endpoint returns HTTP 200, 0 otherwise.",
+		}),
+		RequestDuration: prometheus.NewHistogram(prometheus.HistogramOpts{
+			Name:    KASRequestDurationMetricName,
+			Help:    "Latency of the KAS /healthz health check probe in seconds.",
+			Buckets: KASRequestDurationBuckets,
+		}),
+	}
+
+	crmetrics.Registry.MustRegister(m.Available, m.RequestDuration)
```
jparrill: Consider accepting `prometheus.Registerer` instead of hard-coding the global registry. This eliminates the 3 duplicate manual constructions in tests and avoids `MustRegister` panic on double-registration:
```go
func NewKASHealthMetrics(reg prometheus.Registerer) *KASHealthMetrics {
```

control-plane-operator/controllers/hostedcontrolplane/kas/metrics.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+const (
+	KASAvailableMetricName       = "hypershift_kube_apiserver_available"
+	KASRequestDurationMetricName = "hypershift_kube_apiserver_request_duration_seconds"
+)
+
+// KASRequestDurationBuckets defines the histogram bucket boundaries for KAS
+// health check latency measurements, ranging from 10ms to 10s.
+var KASRequestDurationBuckets = []float64{0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10}
```
jparrill: nit: Exported mutable slice — any caller can corrupt the histogram buckets. Consider unexported `kasRequestDurationBuckets`.

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -897,20 +898,34 @@ func (r *HostedControlPlaneReconciler) healthCheckKASLoadBalancers(ctx context.C
...
+			m.Available.Set(1)
+		} else {
+			m.Available.Set(0)
+		}
+	}
+
 	if err != nil {
 		return err
 	}
+	defer resp.Body.Close()
```
jparrill: Good fix for the pre-existing `resp.Body` leak. Placement is non-idiomatic though — `resp` is used in the `if m != nil` block before `err` is checked. Safer:
```go
if resp != nil {
    defer resp.Body.Close()
}
```
placed right after `httpClient.Get()`, before the metrics block.

jparrill (commented):

control-plane-operator/controllers/hostedcontrolplane/kas/metrics_test.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"testing"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/testutil"
+)
+
+func TestNewKASHealthMetrics(t *testing.T) {
+	t.Run("When creating KAS health metrics, it should register both metrics", func(t *testing.T) {
```
jparrill, line 15: This never calls `NewKASHealthMetrics()` — it tests Prometheus primitives, not the constructor. Rename to `TestKASHealthMetricsOperations`, or rewrite to call the real constructor (easier with the `Registerer` param suggested above). Also: missing `t.Parallel()`, uses `t.Errorf` instead of gomega (project convention).

test/e2e/util/hypershift_framework.go
```diff
@@ -230,6 +230,7 @@ func (h *hypershiftTest) after(hostedCluster *hyperv1.HostedCluster, platform hy
 		}
 
 		ValidateMetrics(t, context.Background(), h.client, hostedCluster, metricsToValidate, true)
+		ValidateCPOMetrics(t, context.Background(), h.client, hostedCluster)
```
jparrill: Runs in every test's `after()` — if metrics emission fails on any platform, all E2E tests fail. Suggest starting as a dedicated test, then promoting to `after()` once stable in CI.

test/e2e/util/util.go
```diff
@@ -2708,6 +2709,43 @@ func ValidateMetrics(t *testing.T, ctx context.Context, client crclient.Client,
...
+		}
+
+		kasMetrics := []string{
+			kas.KASAvailableMetricName,
+			kas.KASRequestDurationMetricName,
+		}
+		hcpNamespace := manifests.HostedControlPlaneNamespace(hc.Namespace, hc.Name)
+
+		err := wait.PollUntilContextTimeout(ctx, 10*time.Second, 5*time.Minute, true, func(ctx context.Context) (bool, error) {
+			mf, err := GetMetricsFromPod(ctx, c, "control-plane-operator", "control-plane-operator", hcpNamespace, "8080")
```
jparrill: nit: Magic string `"8080"` — CPO metrics port from `main.go:207`. Extract to a constant.

PR #2719, opened 2023-06-22
HOSTEDCP-1027: Add Create kubeconfig for HCP CLI

bryan-cox:
**What this PR does / why we need it**:
Add Create kubeconfig for HCP CLI

**Which issue(s) this PR fixes**:
Fixes [HOSTEDCP-1027](https://issues.redhat.com/browse/HOSTEDCP-1027)

**Checklist**
- [x] Subject and description added to both, commit and PR.
- [x] Relevant issues have been referenced.

PR #2487, opened 2023-04-26
ACM-5173 [backport 4.13] get pull secret instead of dockerconfigjson from mce credentials

o-farag:
Backport of https://github.com/openshift/hypershift/commit/142d91e6b81858adbfd50907674c59d9f016c447
Issue: https://issues.redhat.com/browse/ACM-5173

PR #7774, opened 2026-02-23
 CNTRLPLANE-2551: Hypershift integration for API Auth Config changes

ShazaAldawamneh:
This PR updates to support the newly introduced feature-gated fields in the Authentication configuration API added by [#2487](https://github.com/openshift/api/pull/2487) and [#2719](https://github.com/openshift/api/pull/2719). It bumps the openshift/api dependency to consume the updated API types and ensures the new fields are correctly pulled into the HyperShift APIs.

To prevent the feature from automatically going GA, this PR introduces a new TechPreviewNoUpgrade (TPNU) feature gate in HyperShift. This includes updating the feature-gate manifests used for API generation and wiring the new feature gate through to the Control Plane Operator so that the additional Kubernetes API Server deployment configuration logic is conditionally enabled. The implementation ensures that the new behavior is strictly gated and only activated when the TechPreview feature gate is explicitly enabled.

Fixes 
https://issues.redhat.com/browse/CNTRLPLANE-2551 
## Checklist:

## Summary by CodeRabbit

* **New Features**
  * External OIDC authentication with upstream parity support (Tech Preview)
  * CEL-based validation rules for token/claim validation in OIDC providers
  * OIDC discovery URL override for issuer configuration
  * User identity validation rules for OIDC identity providers


control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth.go
```diff
@@ -94,10 +94,15 @@ func generateJWTForProvider(ctx context.Context, provider configv1.OIDCProvider,
 		return out, fmt.Errorf("generating claim validation rules: %v", err)
 	}
 
+	userValidationRules, err := generateUserValidationRules(provider.UserValidationRules...)
+	if err != nil {
+		return out, fmt.Errorf("generating userValidationRules for provider %q: %v", provider.Name, err)
+	}
```
everettraven: Needs to be feature gated

api/hypershift/v1beta1/featuregates/featureGate-Hypershift-Default.yaml
```diff
@@ -43,6 +43,9 @@
                     {
                         "name": "ExternalOIDCWithUIDAndExtraClaimMappings"
                     },
+                    {
+                        "name": "ExternalOIDCWithUpstreamParity"
+                    },
```
everettraven: This should not be enabled by default. Please move this to `disabled` for this file.

api/hypershift/v1beta1/featuregates/featureGate-SelfManagedHA-Default.yaml
```diff
@@ -43,6 +43,9 @@
                     {
                         "name": "ExternalOIDCWithUIDAndExtraClaimMappings"
                     },
+                    {
+                        "name": "ExternalOIDCWithUpstreamParity"
+                    },
```
everettraven: Move to disabled.

control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth.go
everettraven, line 1: Just a general note that we should ensure the generation process here matches that of the cluster-authentication-operator.

control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth_types.go
```diff
@@ -56,6 +56,7 @@ type JWTAuthenticator struct {
 }
 
 // Issuer provides the configuration for a external provider specific settings.
+// +openshift:validation:FeatureGateAwareXValidation:featureGate=ExternalOIDCWithUpstreamParity,rule="self.?discoveryURL.orValue(\"\").size() > 0 ? (self.issuerURL.size() == 0 || self.discoveryURL.find('^.+[^/]') != self.issuerURL.find('^.+[^/]')) : true",message="discoveryURL must be different from issuerURL"
```
everettraven: This isn't necessary. These types are internal only and are not used to generate CRDs.

control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth_types.go
```diff
@@ -94,6 +95,25 @@ type Issuer struct {
...
+	//
+	// +optional
+	// +openshift:enable:FeatureGate=ExternalOIDCWithUpstreamParity
+	// +kubebuilder:validation:XValidation:rule="isURL(self)",message="discoveryURL must be a valid URL"
+	// +kubebuilder:validation:XValidation:rule="url(self).getScheme() == 'https'",message="discoveryURL must be a valid https URL"
+	// +kubebuilder:validation:XValidation:rule="url(self).getQuery().size() == 0",message="discoveryURL must not contain query parameters"
+	// +kubebuilder:validation:XValidation:rule="self.matches('^[^#]*$')",message="discoveryURL must not contain fragments"
+	// +kubebuilder:validation:XValidation:rule="!self.matches('^https://.+:.+@.+/.*$')",message="discoveryURL must not contain user info"
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=2048
```
everettraven: These are not necessary either

control-plane-operator/featuregates/featuregates.go
```diff
@@ -11,19 +11,21 @@ import (
...
 	ExternalOIDCWithUIDAndExtraClaimMappings featuregate.Feature = "ExternalOIDCWithUIDAndExtraClaimMappings"
+	ExternalOIDCWithUpstreamParity           featuregate.Feature = "ExternalOIDCWithUpstreamParity"
 )
 
 // Initialize new features here
 var (
 	allFeatures = featuregates.NewFeatureSetAwareFeatures()
 
 	externalOIDCWithUIDAndExtraClaimMappingsFeature = featuregates.NewFeature(ExternalOIDCWithUIDAndExtraClaimMappings, featuregates.WithEnableForFeatureSets(configv1.TechPreviewNoUpgrade, configv1.Default))
+	externalOIDCWithUpstreamParityFeature           = featuregates.NewFeature(ExternalOIDCWithUpstreamParity, featuregates.WithEnableForFeatureSets(configv1.TechPreviewNoUpgrade, configv1.Default))
```
everettraven: This should not be enabled by default

```suggestion
	externalOIDCWithUpstreamParityFeature           = featuregates.NewFeature(ExternalOIDCWithUpstreamParity, featuregates.WithEnableForFeatureSets(configv1.TechPreviewNoUpgrade))
```

control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth.go
ShazaAldawamneh: What do you mean by the generation process ? @everettraven

everettraven:
Adding a hold as this depends on https://github.com/openshift/api/pull/2719

/hold

cmd/install/assets/hypershift-operator/zz_generated.crd-manifests/hostedclusters-Hypershift-Default.crd.yaml
everettraven, line 1: This file should not have had any changes. I suspect that this is because of the fact that the new feature gate was marked as enabled by default in the first iteration of this PR and this is a stale generated artifact.

Can you try regenerating the CRD manifests to make sure that they match up with what we would expect here?

cmd/install/assets/hypershift-operator/zz_generated.crd-manifests/hostedcontrolplanes-Hypershift-Default.crd.yaml
everettraven, line 1: Same comment here as https://github.com/openshift/hypershift/pull/7774/changes#r2878497027

control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth.go
everettraven: The logic to generate the structured authentication configuration. The only difference here would be how we validate the CEL expressions because of the fact that HyperShift may be configuring an older Kubernetes API server that doesn't have newer CEL capabilities, but that is already handled in this code so you don't need to do anything here.

control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth.go
```diff
@@ -111,6 +118,12 @@ func generateIssuer(ctx context.Context, issuer configv1.TokenIssuer, client crc
 		out.Audiences = append(out.Audiences, string(audience))
 	}
 
+	if featuregates.Gate().Enabled(featuregates.ExternalOIDCWithUpstreamParity) {
+		if len(issuer.DiscoveryURL) > 0 {
+			out.DiscoveryURL = issuer.DiscoveryURL
+		}
```
everettraven, line 149: Should we validate the discovery url is valid like we did in the cluster-authentication-operator?

It will be validated regardless after all the generation logic has completed, but I think failing early if we can tell the discovery url is invalid before we do a more computationally expensive validation step is a reasonable thing to do.

go.mod
```diff
@@ -123,6 +123,8 @@ require (
 	sigs.k8s.io/yaml v1.6.0
 )
 
+replace github.com/openshift/api => github.com/ShazaAldawamneh/api v0.0.0-20260105154902-5470873fa1dc
```
everettraven: Just a standing reminder that this replacement must be removed in favor of a proper update to the commit that includes the changes from https://github.com/openshift/api/pull/2719 (once merged).

everettraven:
@ShazaAldawamneh Would you mind also updating the PR description to include more information as to the changes here?

It also looks like we are missing tests for the new changes - can we make sure to add some?

control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth.go
everettraven: IIRC, this PR is also based on https://github.com/openshift/api/pull/2719 so we should make sure any changes related to supporting these new fields are added here. I don't see anything that configures the `expression` fields for the username/group claim mappings yet.

everettraven, line 392: Claim validation rule configuration for a CEL-based claim validation rule looks to be missing?

ShazaAldawamneh: yes i didn't implemented those yet, I planned to add those after the PR get reviewed and get close to being merged, so now I will work on adding them.

control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth_test.go
```diff
@@ -980,6 +980,253 @@ func TestAdaptAuthConfig(t *testing.T) {
...
+							{
+								Expression: "", // invalid: empty expression
+								Message:    "must have a valid expression",
+							},
+						},
+					},
+				},
+			},
+			shouldError: true,
+		},
```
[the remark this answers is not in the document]
ShazaAldawamneh: good point

control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth.go
```diff
@@ -351,3 +442,18 @@ func HCPAuthConfigToAPIServerAuthConfig(authConfig *AuthenticationConfiguration)
...
+func validateCELExpression(expressionAccessor authenticationcel.ExpressionAccessor) error {
+	_, err := authenticationcel.NewDefaultCompiler().CompileClaimsExpression(expressionAccessor)
+	return err
+}
+
+// validateUserCELExpression validates a user CEL expression using the user.* scope.
+func validateUserCELExpression(expressionAccessor authenticationcel.ExpressionAccessor) error {
+	_, err := authenticationcel.NewDefaultCompiler().CompileUserExpression(expressionAccessor)
+	return err
+}
```
everettraven: We don't need to copy these over. The existing `validateAuthConfig` method will perform the validation we care about on the CEL expressions.

For the CEL expressions we can just perform the necessary conversion logic.

control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth.go
```diff
@@ -196,16 +239,17 @@ func generateUsernameClaimMapping(username configv1.UsernameClaimMapping, issuer
 	return out, nil
 }
 
-func generateGroupsClaimMapping(groups configv1.PrefixedClaimMapping) PrefixedClaimOrExpression {
+func generateGroupsClaimMapping(groups configv1.PrefixedClaimMapping) (PrefixedClaimOrExpression, error) {
 	out := PrefixedClaimOrExpression{}
+	if featuregates.Gate().Enabled(featuregates.ExternalOIDCWithUpstreamParity) && groups.Expression != "" {
+		out.Expression = groups.Expression
+		return out, nil
+	}
```
everettraven: I have the same questions here as I did in the cluster-authentication-operator PR: https://github.com/openshift/cluster-authentication-operator/pull/850#discussion_r2905602237

Let's continue the discussion in the CAO PR since the overall scope of changes there is smaller and copy over the structure to this PR once we are happy with it.

api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/hostedclusters.hypershift.openshift.io/AAA_ungated.yaml
```diff
@@ -580,11 +580,14 @@ spec:
                               ciphers:
                                 description: |-
                                   ciphers is used to specify the cipher algorithms that are negotiated
-                                  during the TLS handshake.  Operators may remove entries their operands
-                                  do not support.  For example, to use DES-CBC3-SHA  (yaml):
+                                  during the TLS handshake. Operators may remove entries that their operands
+                                  do not support. For example, to use only ECDHE-RSA-AES128-GCM-SHA256 (yaml):
```
ehearne-redhat, line 598: Hey @ShazaAldawamneh  - I'm a little confused about the cipher changes as I didn't see it in the referenced PRs in the description. Is it an update coming from one of the repos?

api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/hostedclusters.hypershift.openshift.io/ExternalOIDCWithUIDAndExtraClaimMappings.yaml
```diff
@@ -1858,6 +1797,152 @@ spec:
                                 x-kubernetes-list-type: atomic
                             type: object
                         type: object
+                      networkDiagnostics:
```
ehearne-redhat, line 1814: I'm assuming it is brought in when updating the spec (related to previous comment) ?

control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth.go
ehearne-redhat: Yeah, my same comments here from the reference PR above. :)

control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth_test.go
```diff
@@ -980,6 +980,450 @@ func TestAdaptAuthConfig(t *testing.T) {
...
+				JWT: []JWTAuthenticator{
+					{
+						Issuer: Issuer{
+							URL:                 "https://test.com",
+							AudienceMatchPolicy: AudienceMatchPolicyMatchAny,
+							Audiences:           []string{"one", "two"},
+						},
+						ClaimMappings: ClaimMappings{
+							Username: PrefixedClaimOrExpression{
+								Expression: "claims.email",
```
ehearne-redhat, line 1242: Looking at [the unit test](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_hypershift/7774/pull-ci-openshift-hypershift-main-unit/2031005614867484672) , we need to change this from `claims.email` to `claims.email_verified` . Hopefully that resolves the issue.

```bash
FAIL: TestAdaptAuthConfig (0.05s)
    --- FAIL: TestAdaptAuthConfig/authn_spec_provided,_username_expression_set,_no_error,_successful_mapping (0.00s)
        auth_test.go:1464: unexpected error: validating generated authentication config: jwt[0].claimMappings.username.expression: Invalid value: "claims.email": claims.email_verified must be used in claimMappings.username.expression or claimMappings.extra[*].valueExpression or claimValidationRules[*].expression when claims.email is used in claimMappings.username.expression
```

ehearne-redhat:
Hey @ShazaAldawamneh - looking at [the verify test](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_hypershift/7774/pull-ci-openshift-hypershift-main-verify/2031005617396649984) we need to follow the commit message convention.

The commit message can be changed to something like (Note the new lines - new lines are required to count as commit message description):

```bash
feat(api): add cel expression support to user and group claim mapping

Updates are made to support the newly introduced feature-gated fields in the Authentication configuration API added by https://github.com/openshift/api/pull/2487 and https://github.com/openshift/api/pull/2719. It bumps the openshift/api dependency to consume the updated API types and ensures the new fields are correctly pulled into the HyperShift APIs.

To prevent the feature from automatically going GA, this PR introduces a new TechPreviewNoUpgrade (TPNU) feature gate in HyperShift. This includes updating the feature-gate manifests used for API generation and wiring the new feature gate through to the Control Plane Operator so that the additional Kubernetes API Server deployment configuration logic is conditionally enabled. The implementation ensures that the new behavior is strictly gated and only activated when the TechPreview feature gate is explicitly enabled.
```

We may need to shorten this as I believe the verify test will fail if the commit message length is too long, but this will resolve the current test failure. :)


```bash
Linting commits from 806fc3b347dc128a7cb58d5231799a9f0465686f to 1f9a984a0a7441f57a0065f16c57f49543baf6ce (CI: PR targeting 806fc3b347dc128a7cb58d5231799a9f0465686f)
1: CT1 Title does not follow ConventionalCommits.org format 'type(optional-scope): description': "adding cel expression support to user and group claim mapping"
make: *** [Makefile:394: run-gitlint] Error 1
```

api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/hostedclusters.hypershift.openshift.io/AAA_ungated.yaml
ShazaAldawamneh: these kind of files changed automatically so it's not manual changes

api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/hostedclusters.hypershift.openshift.io/ExternalOIDCWithUIDAndExtraClaimMappings.yaml
ShazaAldawamneh: yes I think so

control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth_test.go
ShazaAldawamneh: locally it passed without making this change, so will see the outcome now with the new changes if that fixed the issue

ehearne-redhat: There is an alternative which I will comment on separately.

> claimValidationRules[*].expression when claims.email is used in claimMappings.username.expression

control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth_test.go
```diff
@@ -980,6 +980,450 @@ func TestAdaptAuthConfig(t *testing.T) {
...
+							},
+							Groups: PrefixedClaimOrExpression{
+								Prefix: ptr.To(""),
+								Claim:  "",
+							},
+							UID:   ClaimOrExpression{Claim: "sub"},
+							Extra: []ExtraMapping{},
+						},
+						ClaimValidationRules: []ClaimValidationRule{},
```
ehearne-redhat: Maybe we could add a claim validation rule such as 

```
ClaimValidationRules: []ClaimValidationRule{
							{
								Expression: "claims.email_verified",
								Message:    "email must be verified",
							},
						},
```

It should satisfy this verification (see in resolved comment). 

> claimValidationRules[*].expression when claims.email is used in claimMappings.username.expression

control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth_test.go
```diff
@@ -980,6 +980,450 @@ func TestAdaptAuthConfig(t *testing.T) {
...
+								Prefix: ptr.To("https://test.com#"),
+								Claim:  "username",
+							},
+							Groups: PrefixedClaimOrExpression{
+								Expression: "claims.groups",
+							},
+							UID:   ClaimOrExpression{Claim: "sub"},
+							Extra: []ExtraMapping{},
+						},
+						ClaimValidationRules: []ClaimValidationRule{},
```
ehearne-redhat, line 1375: Same here.

control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth_test.go
ehearne-redhat: Seen in https://github.com/openshift/hypershift/pull/7774#discussion_r2910671658 .

ShazaAldawamneh: where did you see that this is what causing the issue ? because I don't seem to find it in the unit tests logs.

ehearne-redhat:
The error in the [unit test](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_hypershift/7774/pull-ci-openshift-hypershift-main-unit/2031707848081149952) is because we do not have any check for CEL type. 

--- FAIL: TestAdaptAuthConfig (0.03s)
    --- FAIL: TestAdaptAuthConfig/authn_spec_provided,_username_expression_set,_no_error,_successful_mapping (0.00s)
        auth_test.go:1478: unexpected error: failed to generate authentication config: generating JWT authenticator for provider "test": generating claim validation rules: unknown claimValidationRule type "CEL"

This should be added in this file.

https://github.com/openshift/hypershift/blob/5271846bd0b82b480e38da151c0a9e11d9f9cd0a/control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth.go#L282-L301

Something along the lines of 

```golang
if claimValidationRule.RequiredClaim == nil || claimValidationRule.CEL == nil { 
 			return out, fmt.Errorf("claimValidationRule.type is %s and requiredClaim is not set", configv1.TokenValidationRuleTypeRequiredClaim) 
 		} 
```

should resolve the test issue there.

control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth.go
```diff
@@ -155,30 +214,93 @@ func generateClaimMappings(claimMappings configv1.TokenClaimMappings, issuerURL
...
+		return out, nil, fmt.Errorf("claim or expression field is empty in username claim mapping")
+	}
+
+	if len(username.Expression) > 0 && len(username.Claim) > 0 {
+		return out, nil, fmt.Errorf("username claim mapping must not set both claim and expression")
+	}
+
+	// TODO: update the API-side validations to ensure that prefixPolicy is set to
+	// NoPrefix or omitted when expression is set, for stronger admission time
+	// validation prior to GA.
```
everettraven: Remove this?

control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth.go
```diff
@@ -23,12 +25,24 @@ import (
...
 	certificateAuthorityConfigMapKey = "ca-bundle.crt"
 )
 
+// oidcGenerationState holds compilation results gathered during JWT generation
+// that are needed for cross-field validation (e.g. email_verified enforcement).
+type oidcGenerationState struct {
+	usernameResult         *authenticationcel.CompilationResult
+	extraResults           []authenticationcel.CompilationResult
+	claimValidationResults []authenticationcel.CompilationResult
+}
```
everettraven: This is where differences between how HyperShift has done the validation and how the CAO has done the validation differs a bit - I don't think we need to actually copy this logic over because HyperShift ends up using the upstream validation logic to run validation on the converted type [1] - which means any CEL compilations we do earlier in the chain end up getting repeated.

Because of this, we have only been trying to fail out early on things that are relatively cheap to evaluate before we proceed to the more expensive validation step later in the generation cycle.

[1]: https://github.com/openshift/hypershift/blob/6766813331de049b44f4d08efff3887cd090996c/control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth.go#L309-L333

control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth.go
```diff
@@ -155,30 +214,93 @@ func generateClaimMappings(claimMappings configv1.TokenClaimMappings, issuerURL
...
+	}
+
+	if len(username.Expression) > 0 {
+		compiler := authenticationcel.NewDefaultCompiler()
+		result, err := compiler.CompileClaimsExpression(&authenticationcel.ClaimMappingExpression{
+			Expression: username.Expression,
+		})
+		if err != nil {
+			return out, nil, fmt.Errorf("invalid CEL expression: %v", err)
+		}
```
everettraven: For the same reasons outlined in https://github.com/openshift/hypershift/pull/7774/changes#r2946886415 - I'd remove the compilation step here and defer that to the more expensive validation step we do in the future.

control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth.go
```diff
@@ -242,9 +376,19 @@ func generateExtraClaimMapping(extras ...configv1.ExtraMapping) ([]ExtraMapping,
...
+		if len(extra.ValueExpression) > 0 {
+			compiler := authenticationcel.NewDefaultCompiler()
+			result, err := compiler.CompileClaimsExpression(&authenticationcel.ExtraMappingExpression{
+				Key:        extra.Key,
+				Expression: extra.ValueExpression,
+			})
+			if err == nil {
+				compilationResults = append(compilationResults, result)
+			}
+		}
```
everettraven: For the same reasons outlined in https://github.com/openshift/hypershift/pull/7774/changes#r2946886415 I don't think we need this added.

control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth.go
```diff
@@ -288,18 +435,67 @@ func generateClaimValidationRule(claimValidationRule configv1.TokenClaimValidati
...
+		if len(claimValidationRule.CEL.Expression) == 0 {
+			return out, nil, fmt.Errorf("claimValidationRule.type is %s and expression is not set", configv1.TokenValidationRuleTypeCEL)
+		}
+		compiler := authenticationcel.NewDefaultCompiler()
+		result, err := compiler.CompileClaimsExpression(&authenticationcel.ClaimValidationCondition{
+			Expression: claimValidationRule.CEL.Expression,
+		})
+		if err != nil {
+			return out, nil, fmt.Errorf("invalid CEL expression: %v", err)
+		}
```
everettraven: https://github.com/openshift/hypershift/pull/7774/changes#r2946886415

control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth.go
```diff
@@ -288,18 +435,67 @@ func generateClaimValidationRule(claimValidationRule configv1.TokenClaimValidati
...
+func generateUserValidationRule(rule configv1.TokenUserValidationRule) (UserValidationRule, error) {
+	if len(rule.Expression) == 0 {
+		return UserValidationRule{}, fmt.Errorf("userValidationRule expression must be non-empty")
+	}
+
+	if _, err := authenticationcel.NewDefaultCompiler().CompileUserExpression(&authenticationcel.UserValidationCondition{
+		Expression: rule.Expression,
+	}); err != nil {
+		return UserValidationRule{}, fmt.Errorf("invalid CEL expression: %v", err)
+	}
```
everettraven: https://github.com/openshift/hypershift/pull/7774/changes#r2946886415

control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth.go
```diff
@@ -351,3 +547,138 @@ func HCPAuthConfigToAPIServerAuthConfig(authConfig *AuthenticationConfiguration)
...
+	return id != nil && id.Name == operand
+}
+
+func isConstField(exp *exprpb.Expr, field string) bool {
+	if exp == nil || len(field) == 0 {
+		return false
+	}
+	c := exp.GetConstExpr()
+	return c != nil && c.GetStringValue() == field
+}
```
everettraven: Because of https://github.com/openshift/hypershift/pull/7774/changes#r2946886415 - we shouldn't need to carry this here.

control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth_test.go
```diff
@@ -980,6 +980,547 @@ func TestAdaptAuthConfig(t *testing.T) {
...
+								"two",
+							},
+						},
+					},
+				},
+			},
+			shouldError: true,
+		},
+		{
+			name:   "authn spec provided, username expression set with claims.email and claims.email_verified in claimValidationRule, no error, successful mapping",
```
everettraven, line 1223: Just a note - all these tests should stay to make sure that the behavior of the later validation stage I called out in https://github.com/openshift/hypershift/pull/7774/changes#r2946886415 is behaving the way we expect.

control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth.go
ShazaAldawamneh: that's actually correct, I missed that, since HyperShift already runs full upstream validation via validateAuthConfig which repeats the CEL compilation anyway, the compilation we added earlier is redundant work. Will revert the CEL compilation logic and keep only the cheap structural checks before the upstream validation step.

control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth_test.go
ShazaAldawamneh: understood, keeping all the tests to ensure the upstream validation step in validateAuthConfig is behaving as expected

ehearne-redhat:
Hey @ShazaAldawamneh - you might want to squash your commits so that `verify` doesn't fail. The first commit contains a description but the second one doesn't. They seem to have the same first line. :)

ShazaAldawamneh:
> Hey @ShazaAldawamneh - you might want to squash your commits so that `verify` doesn't fail. The first commit contains a description but the second one doesn't. They seem to have the same first line. :)

@ehearne-redhat I don't think it's failing because of that, because it's complaining about the untracked file in the log of the verify test, but thanks for reminding me to squash the 2 commits :)

control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth.go
```diff
@@ -168,11 +209,63 @@ func generateClaimMappings(claimMappings configv1.TokenClaimMappings, issuerURL
...
+	if len(username.Expression) == 0 && len(username.Claim) == 0 {
+		return out, fmt.Errorf("claim or expression field is empty in username claim mapping")
+	}
+
+	if len(username.Expression) > 0 && len(username.Claim) > 0 {
+		return out, fmt.Errorf("username claim mapping must not set both claim and expression")
+	}
+
+	if len(username.Expression) > 0 && username.PrefixPolicy == configv1.Prefix {
+		return out, fmt.Errorf("username claim mappings cannot have a prefix set when using an expression based mapping. If you want to set a prefix while using an expression mapping, set the prefix in the expression")
```
everettraven: Non-blocking nit: I know I had suggested something like this, but on a second pass something like:

```suggestion
		return out, fmt.Errorf("username claim mappings cannot have a prefix set when using an expression based mapping. When using an expression mapping, set the prefix in the expression")
```

reads a bit better. Won't block on this though - we can always update this later.

control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth.go
everettraven, line 391: Need to drop this comment now?

api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/hostedclusters.hypershift.openshift.io/NetworkDiagnosticsConfig.yaml
```diff
@@ -1,7020 +0,0 @@
-apiVersion: apiextensions.k8s.io/v1
```
ehearne-redhat, line 1: This file and similar are removed because of api changes made in https://github.com/openshift/api/pull/2672. Nothing to be concerned with :)

ehearne-redhat:
/lgtm

No new major changes between last lgtm and now since this was a rebase to resolve merge conflict. :)

PR #7781, opened 2026-02-23
GCP-386: scope externaldns RBAC to GCP platform only

cristianoveiga:

PR #7702 added externaldns.k8s.io/dnsendpoints permissions unconditionally to the CPO role. On non-GCP platforms, the HO does not hold these permissions, causing RBAC escalation errors during reconciliation.

Move the externaldns rule from the static role.yaml to the adaptRole function in role.go, gated on GCPPlatform, consistent with how ARO HCP and ROSA HCP conditional rules are handled.

[GCP-386](https://issues.redhat.com/browse/GCP-386)

## Checklist:

## Summary by CodeRabbit

* **Chores**
  * Tightened ExternalDNS RBAC by removing general DNSEndpoints permissions.
  * Restored limited DNSEndpoints permissions only when ExternalDNS is configured for Google/GCP.

### DIFF
--- a/cmd/install/install.go
+++ b/cmd/install/install.go
@@ -37,6 +37,7 @@ import (
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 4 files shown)


muraee:
could we please also scope the ExternalDNSClusterRole changes to GCP/google

cmd/install/install.go
```diff
@@ -964,6 +965,21 @@ func setupExternalDNS(opts Options, operatorNamespace *corev1.Namespace) ([]crcl
 	objects = append(objects, externalDNSServiceAccount)
 
 	externalDNSClusterRole := assets.ExternalDNSClusterRole{}.Build()
+	if hyperv1.PlatformType(opts.PrivatePlatform) == hyperv1.GCPPlatform {
```
muraee: should this check `opts.ExternalDNSProvider == "google"` instead?

cristianoveiga: I was actually debating between the two options, but I think it makes sense to check for google here (following the same logic we use to set --source=crd)

cristianoveiga: I've updated it.

muraee:
/approve
@cristianoveiga could you please squash the commits

PR #7786, opened 2026-02-24
CNTRLPLANE-2840: feat(CPO): add endpoint-resolver control plane component

muraee:
## Summary

- Adds the endpoint-resolver as a new CPOv2 control plane component that resolves control plane
  component names to their backing ready pod IPs using the
  `hypershift.openshift.io/control-plane-component` label, enabling the metrics proxy to discover
  and scrape individual pods directly
- Implements an HTTPS server with TLS, self-signed CA/serving certificate management, and full
  RBAC (deployment, service, serviceaccount, role, rolebinding)
- Registers the endpoint-resolver in the control-plane-operator main entrypoint

This is part of the [hosted control plane metrics exposure enhancement](https://github.com/openshift/enhancements/blob/master/enhancements/hypershift/hosted-control-plane-metrics-exposure.md).

## Test plan

- [ ] Verify generated test fixtures match expected output across all platform variants (default, AROSwift, GCP, IBMCloud, TechPreviewNoUpgrade)
- [ ] Run `go test ./control-plane-operator/endpoint-resolver/...` to validate server unit tests
- [ ] Run `make test` for full unit test suite
- [ ] Run `make verify` to ensure code generation and formatting are correct

## Summary by CodeRabbit

* **New Features**
  * Introduced an endpoint-resolver service: deployable control-plane component with a CLI subcommand, TLS-backed HTTP endpoints (/resolve/, /healthz), RBAC, ServiceAccount, Service and Deployment, and certificate handling.
  * Added a HostedCluster annotation to opt into metrics forwarding; reconciliation now propagates this annotation to the hosted control plane.

* **Tests**
  * Added unit tests and comprehensive multi-platform test fixtures for the endpoint resolver.

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
@@ -42,6 +42,7 @@ import (
 	kubevirtcsiv2 "github.com/openshift/hypershift/control-plane-operator/controllers/hostedcontrolplane/v2/csi/kubevirt"
 	cvov2 "github.com/openshift/hypershift/control-plane-operator/controllers/hostedcontrolplane/v2/cvo"
 	dnsoperatorv2 "github.com/openshift/hypershift/control-plane-operator/controllers/hostedcontrolplane/v2/dnsoperator"
+	endpointresolverv2 "github.com/openshift/hypershift/control-plane-operator/controllers/hostedcontrolplane/v2/endpoint_resolver"
 	etcdv2 "github.com/openshift/hypershift/control-plane-operator/controllers/hostedcontrolplane/v2/etcd"
 	fgv2 "github.com/openshift/hypershift/control-plane-operator/controllers/hostedcontrolplane/v2/fg"
 	ignitionserverv2 "github.com/openshift/hypershift/control-plane-operator/controllers/hostedcontrolplane/v2/ignitionserver"
@@ -265,6 +266,7 @@ func (r *HostedControlPlaneReconciler) registerComponents(hcp *hyperv1.HostedCon
 		konnectivityv2.NewComponent(),
 		ignitionserverv2.NewComponent(r.ReleaseProvider, r.DefaultIngressDomain),
 		ignitionproxyv2.NewComponent(r.DefaultIngressDomain),
+		endpointresolverv2.NewComponent(),
 	)
 	r.components = append(r.components,
 		olmv2.NewComponents(r.ManagementClusterCapabilities.Has(capabilities.CapabilityImageStream))...,
--- a/control-plane-operator/controllers/hostedcontrolplane/v2/endpoint_resolver/component.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/v2/endpoint_resolver/component.go
@@ -1,0 +1,50 @@
+package endpointresolver
+
+import (
+	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
+	component "github.com/openshift/hypershift/support/controlplane-component"
+)
+
+const (
+	ComponentName = "endpoint-resolver"
+)
+
+var _ component.ComponentOptions = &endpointResolver{}
+
+type endpointResolver struct{}
+
+func (r *endpointResolver) IsRequestServing() bool {
+	return false
+}
+
+func (r *endpointResolver) MultiZoneSpread() bool {
+	return false
+}
+
+func (r *endpointResolver) NeedsManagementKASAccess() bool {
+	return true
+}
+
+func NewComponent() component.ControlPlaneComponent {
+	return component.NewDeploymentComponent(ComponentName, &endpointResolver{}).
+		WithPredicate(predicate).
+		WithManifestAdapter(
+			"ca-cert.yaml",
+			component.WithAdaptFunction(adaptCACertSecret),
+			component.DisableIfAnnotationExist(hyperv1.DisablePKIReconciliationAnnotation),
+			component.ReconcileExisting(),
+		).
+		WithManifestAdapter(
+			"serving-cert.yaml",
+			component.WithAdaptFunction(adaptServingCertSecret),
+			component.DisableIfAnnotationExist(hyperv1.DisablePKIReconciliationAnnotation),
+			component.ReconcileExisting(),
+		).
+		Build()
+}
+
+func predicate(cpContext component.WorkloadContext) (bool, error) {
+	_, disableMonitoring := cpContext.HCP.Annotations[hyperv1.DisableMonitoringServices]
+	_, enableMetricsForwarding := cpContext.HCP.Annotations[hyperv1.EnableMetricsForwarding]
+	return !disableMonitoring && enableMetricsForwarding, nil
+}
--- a/control-plane-operator/endpoint-resolver/cmd.go
+++ b/control-plane-operator/endpoint-resolver/cmd.go
@@ -1,0 +1,125 @@
+package endpointresolver
+
+import (
+	"context"
+	"crypto/tls"
+	"fmt"
+	"net/http"
+	"os"
+	"os/signal"
+	"syscall"
+	"time"
+
+	"github.com/openshift/hypershift/support/supportedversion"
+
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+
+	"sigs.k8s.io/controller-runtime/pkg/log"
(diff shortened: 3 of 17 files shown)


control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -42,6 +42,7 @@ import (
 	kubevirtcsiv2 "github.com/openshift/hypershift/control-plane-operator/controllers/hostedcontrolplane/v2/csi/kubevirt"
 	cvov2 "github.com/openshift/hypershift/control-plane-operator/controllers/hostedcontrolplane/v2/cvo"
 	dnsoperatorv2 "github.com/openshift/hypershift/control-plane-operator/controllers/hostedcontrolplane/v2/dnsoperator"
+	endpointresolverv2 "github.com/openshift/hypershift/control-plane-operator/controllers/hostedcontrolplane/v2/endpoint_resolver"
```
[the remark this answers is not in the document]
muraee: this is meant to be only used with metrics-proxy component introduced here https://github.com/openshift/hypershift/pull/7752.

control-plane-operator/endpoint-resolver/server.go
```diff
@@ -0,0 +1,99 @@
...
+
+	if len(pods) == 0 {
+		http.Error(w, fmt.Sprintf("no ready pods found for component %s", componentName), http.StatusNotFound)
+		return
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	if err := json.NewEncoder(w).Encode(ResolveResponse{Pods: pods}); err != nil {
+		http.Error(w, fmt.Sprintf("failed to encode response: %v", err), http.StatusInternalServerError)
+	}
```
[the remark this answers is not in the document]
muraee: Done. Changed to marshal to a buffer first via `json.Marshal`, so `http.Error` can reliably return a 500 if encoding fails.

---

control-plane-operator/endpoint-resolver/server_test.go
```diff
@@ -0,0 +1,169 @@
...
+		lister := newFakePodLister("test-namespace", nil)
+		handler := NewResolverHandler(lister)
+		req := httptest.NewRequest(http.MethodGet, "/resolve/", nil)
+		rec := httptest.NewRecorder()
+
+		handler.ServeHTTP(rec, req)
+
+		if rec.Code != http.StatusBadRequest {
+			t.Errorf("expected status code %d, got %d", rec.Code, http.StatusBadRequest)
+		}
```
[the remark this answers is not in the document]
muraee: Done. Fixed the reversed format arguments so the error message correctly reads "expected <expected>, got <actual>".

---

control-plane-operator/endpoint-resolver/cmd.go
```diff
@@ -0,0 +1,123 @@
...
+		sigChan := make(chan os.Signal, 1)
+		signal.Notify(sigChan, syscall.SIGTERM, syscall.SIGINT)
+		go func() {
+			sig := <-sigChan
+			l.Info("Received shutdown signal", "signal", sig)
+			cancel()
+			if err := server.Shutdown(context.Background()); err != nil {
+				l.Error(err, "Error during server shutdown")
+			}
+		}()
```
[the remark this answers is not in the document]
muraee: Done. Added a 30-second timeout to the shutdown context to prevent indefinite blocking if connections hang.

---

enxebre:
the desc link to the enhancement is pointing to the fork

control-plane-operator/endpoint-resolver/server.go
```diff
@@ -0,0 +1,102 @@
...
+	// Extract component name from /resolve/<component-name>
+	path := strings.TrimPrefix(r.URL.Path, "/resolve/")
+	componentName := strings.TrimSuffix(path, "/")
+	if componentName == "" {
+		http.Error(w, "component name is required", http.StatusBadRequest)
+		return
+	}
+
+	selector := labels.SelectorFromSet(labels.Set{
+		hyperv1.ControlPlaneComponentLabel: componentName,
```
enxebre: shouldn't we be filtering here by pod/service Monitors?

muraee: could you elaborate more?

control-plane-operator/controllers/hostedcontrolplane/v2/endpoint_resolver/component.go
```diff
@@ -0,0 +1,49 @@
...
+			"serving-cert.yaml",
+			component.WithAdaptFunction(adaptServingCertSecret),
+			component.DisableIfAnnotationExist(hyperv1.DisablePKIReconciliationAnnotation),
+			component.ReconcileExisting(),
+		).
+		Build()
+}
+
+func predicate(cpContext component.WorkloadContext) (bool, error) {
+	_, disableMonitoring := cpContext.HCP.Annotations[hyperv1.DisableMonitoringServices]
```
csrwng, line 47: Should we have a separate "feature flag" to enable/disable pushing metrics into the hosted cluster?

control-plane-operator/endpoint-resolver/server.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+// ResolveResponse is the JSON response returned by the resolve endpoint.
+type ResolveResponse struct {
+	Pods []PodEndpoint `json:"pods"`
+}
+
+// ResolverHandler handles endpoint resolution requests by looking up Pods
+// using the hypershift.openshift.io/control-plane-component label.
+type ResolverHandler struct {
```
csrwng: nit: this doesn't need to be public

control-plane-operator/endpoint-resolver/server.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+func (h *ResolverHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	if r.Method != http.MethodGet {
+		http.Error(w, "method not allowed", http.StatusMethodNotAllowed)
+		return
+	}
+
+	// Extract component name from /resolve/<component-name>
+	path := strings.TrimPrefix(r.URL.Path, "/resolve/")
```
csrwng: same literal used in cmd.go for creating the mux, consider making it a constant

control-plane-operator/endpoint-resolver/server.go
csrwng: so doing a listing of pods in a control plane for pods that don't have the control-plane-component label, I get:
```
NAME                                            
aws-ebs-csi-driver-controller-8684c4ffdd-gmz24  
aws-ebs-csi-driver-operator-5dddcd9544-zhhhq    
cloud-network-config-controller-7455c88989-s2jbw
csi-snapshot-controller-5c47877f7b-4nnrk        
multus-admission-controller-66b9b94595-zhc7w    
```

would we ever want to collect metrics from those, and if so, then we'd need to either add the label or use something other than the control-plane-component selector.

I think what @enxebre was referring to is that instead of matching the name with a component, you'd match the name with either a podmonitor/servicemonitor, then follow the label selector chain from there.

control-plane-operator/endpoint-resolver/server.go
muraee: Done. Made `ResolverHandler` unexported (`resolverHandler`) along with its constructor.

---

control-plane-operator/endpoint-resolver/server.go
muraee: Done. Extracted `/resolve/` into a `resolvePathPrefix` constant shared between `server.go` and `cmd.go`.

---

control-plane-operator/controllers/hostedcontrolplane/v2/endpoint_resolver/component.go
muraee: Good point. We'll add an opt-in annotation so this component isn't deployed for all clusters immediately. The predicate will check both: monitoring must not be disabled (`DisableMonitoringServices`) AND the new opt-in annotation must be present.

---

control-plane-operator/endpoint-resolver/server.go
muraee: Agreed, the current approach would miss those unlabeled pods. We're going to redesign the resolver to accept arbitrary label selectors via a POST request body instead of relying on the `control-plane-component` label. The metrics-proxy will be responsible for resolving the ServiceMonitor → Service → pod label selector chain (or PodMonitor → pod label selector), and pass the final pod labels to the resolver. This keeps the resolver thin while covering all monitored pods.

---

csrwng:
/lgtm
/hold
in case there are any additional review comments

PR #7789, opened 2026-02-24
OCPBUGS-77224: fix(routing): add Azure platform to LabelHCPRoutes() switch

bryan-cox:

PR #7605 replaced `IsPublicWithDNS()` with `LabelHCPRoutes()` as the single source of truth for route labeling decisions, but omitted `hyperv1.AzurePlatform` from the switch statement. Azure fell through to `default: return false`, which prevented HCP router infrastructure from being created for self-managed Azure clusters using KAS Route.

Meanwhile, `kas/service.go:reconcileExternalRoute()` unconditionally labels the kube-apiserver route with `hypershift.openshift.io/hosted-control-plane`. The default IngressController's `routeSelector` excludes routes with that label. With no HCP router created and the default IngressController rejecting the route, the kube-apiserver route was never admitted, ExternalDNS had no target, and the cluster could not progress.

This PR adds `hyperv1.AzurePlatform` to the same case as Agent, KubeVirt, OpenStack, and None platforms in `LabelHCPRoutes()`, which returns `UseDedicatedDNSForKAS(hcp)`. This ensures the HCP router is created when KAS uses Route with a hostname on Azure.

**Empirical evidence from a live cluster confirmed the fix resolves the issue.**

Fixes https://issues.redhat.com/browse/OCPBUGS-77224

This is a one-line fix. The regression was introduced by PR #7605 (`csrwng/fix_unnecessary_router`, OCPBUGS-70152) which added `LabelHCPRoutes()` but did not include a case for `AzurePlatform`.

Before PR #7605, `IsPublicWithDNS()` was platform-agnostic — it checked if any service had dedicated DNS regardless of platform, so Azure worked. After PR #7605, Azure falls through to the `default` case which returns `false`.

Evidence from the live cluster:

| Route | HCP Label | Admitted | DNS Record |
|-------|-----------|----------|------------|
| oauth | No | Yes | Yes |
| konnectivity | No | Yes | Yes |
| kube-apiserver | Yes (always) | No | No |

After applying the fix and redeploying the CPO, the HCP router was created, the kube-apiserver route was admitted, and the cluster began progressing.

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Azure platform now follows the same routing and DNS behaviour as other supported clouds, ensuring consistent control-plane routing and hostname handling.

* **Tests**
  * Added tests and test fixtures covering Azure self-managed route scenarios (with and without KAS hostnames), verifying labeling, external-router requirements, and status reconciliation.

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/infra/infra_test.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/infra/infra_test.go
@@ -414,6 +414,30 @@ func TestReconcileInfrastructure(t *testing.T) {
 				NeedExternalRouter:    false,
 			},
 		},
+		// Azure self-managed test cases
+		{
+			name: "When Azure self-managed cluster has KAS Route with hostname, it should need an external router",
+			hcp: withServices(
+				baseAzureHCP(),
+				allServicesRouteWithHostnames(),
+			),
+			expectError: false,
+			// For Azure self-managed with Route:
+			// - LabelHCPRoutes returns true (KAS uses Route with hostname)
+			// - External router is needed to admit labeled routes
+			expectedStatus: &InfrastructureStatus{
+				APIHost:               testKASHostname,
+				APIPort:               443,
+				OAuthEnabled:          true,
+				OAuthHost:             testOAuthHostname,
+				OAuthPort:             443,
+				KonnectivityHost:      testKonnectivityHost,
+				KonnectivityPort:      443,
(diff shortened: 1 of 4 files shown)


enxebre:
you might want to include a testcase for this here https://github.com/openshift/hypershift/blob/main/control-plane-operator/controllers/hostedcontrolplane/infra/infra_test.go#L255

PR #7791, opened 2026-02-24
AUTOSCALE-559: PrivateLink plumbing for karpenter ec2nodeclass subnets 

jkyros:

I don't love this, but this is basically: 
- Nodeclass controller grabs all subnets from non-default ec2nodeclasses 
- Stuffs those subnets into a configmap that gets read by the private link controller
- Private link controller adds them to its list when it runs 
- e2e to test that scenerio 

Fixes [AUTOSCALE-559](https://issues.redhat.com/browse/AUTOSCALE-559)

Things that are not great: 
- UX: There is no feedback to the user/ec2nodeclass on whether it worked or not up at the privatelink controller level, and since it's just a configmap and the private link controller is in the control plane, so the user can't see the logs. If it goes bad, they will just find out when their node doesn't join :disappointed: 
- Tests: I wanted to test it against a cluster with no existing nodepool, so I did that test separately and stuffed it in before the rest of the karpenter tests run 

## Checklist:

## Release Notes

* **New Features**
  * Added automatic subnet discovery and synchronization between Karpenter and the operator via ConfigMap.
  * Subnets are now aggregated from multiple sources and dynamically propagated.

* **Tests**
  * Expanded test coverage for subnet management, ConfigMap handling, and Karpenter integration scenarios.

### DIFF
--- a/hypershift-operator/controllers/platform/aws/controller.go
+++ b/hypershift-operator/controllers/platform/aws/controller.go
@@ -2,9 +2,9 @@ package aws
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
-	"sort"
 	"strings"
 	"time"
 
@@ -12,6 +12,7 @@ import (
 	awsutil "github.com/openshift/hypershift/cmd/infra/aws/util"
 	"github.com/openshift/hypershift/control-plane-operator/controllers/hostedcontrolplane/manifests"
 	"github.com/openshift/hypershift/support/awsapi"
+	karpenterutil "github.com/openshift/hypershift/support/karpenter"
 	"github.com/openshift/hypershift/support/upsert"
 	supportutil "github.com/openshift/hypershift/support/util"
 
@@ -25,6 +26,7 @@ import (
 	elbv2types "github.com/aws/aws-sdk-go-v2/service/elasticloadbalancingv2/types"
 	"github.com/aws/smithy-go"
 
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
@@ -34,11 +36,13 @@ import (
 	"k8s.io/client-go/util/workqueue"
 
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	"github.com/go-logr/logr"
@@ -97,6 +101,16 @@ func (r *AWSEndpointServiceReconciler) SetupWithManager(mgr ctrl.Manager) error 
 			DeleteFunc: r.enqueueOnNodePoolDelete(mgr),
 		}).
 		Watches(&hyperv1.HostedCluster{}, handler.Funcs{UpdateFunc: r.enqueueOnHostedClusterChange(mgr)}).
+		// We can't filter this to just our configmaps at the cache level because there are
+		// other things (e.g. nodepool reconciler) that are registered with this same manager
+		// and share the infrastructure, but what we can do is limit the events with a predicate
+		Watches(&corev1.ConfigMap{}, handler.Funcs{
+			CreateFunc: r.enqueueOnKarpenterConfigMapCreate(mgr),
+			UpdateFunc: r.enqueueOnKarpenterConfigMapChange(mgr),
+		}, builder.WithPredicates(predicate.NewPredicateFuncs(func(o client.Object) bool {
+			return o.GetName() == karpenterutil.KarpenterSubnetsConfigMapName &&
+				o.GetLabels()["hypershift.openshift.io/managed-by"] == "karpenter"
+		}))).
 		WithOptions(controller.Options{
 			RateLimiter:             workqueue.NewTypedItemExponentialFailureRateLimiter[reconcile.Request](3*time.Second, 30*time.Second),
 			MaxConcurrentReconciles: 10,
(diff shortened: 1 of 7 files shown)


karpenter-operator/controllers/nodeclass/ec2_nodeclass_controller.go
```diff
@@ -349,6 +361,104 @@ func (r *EC2NodeClassReconciler) reconcileStatus(ctx context.Context, ec2NodeCla
...
+func (r *EC2NodeClassReconciler) reconcileKarpenterSubnetsConfigMap(ctx context.Context, hcp *hyperv1.HostedControlPlane) error {
+	log := ctrl.LoggerFrom(ctx)
+
+	// List all OpenshiftEC2NodeClass resources in guest cluster
+	openshiftEC2NodeClassList := &hyperkarpenterv1.OpenshiftEC2NodeClassList{}
+	if err := r.guestClient.List(ctx, openshiftEC2NodeClassList); err != nil {
+		return fmt.Errorf("failed to list OpenshiftEC2NodeClass: %w", err)
+	}
+
+	// Aggregate subnet IDs from user-defined NodeClasses only.
```
enxebre: >		// SubnetSelectorTerms == nil is the signal that this is the managed default NodeClass, // not a user-defined one.

SubnetSelectorTerms is an optional field, any api consumer can create a class and omit it, no?

>// NodeClasses without SubnetSelectorTerms are managed defaults that use the
	// discovery-tag selector; their subnets are already covered by the control-plane

how is this tied to the control-plane? An infra admin persona can add "karpenter.sh/discovery": hcp.Spec.InfraID to any subnet, which then would be discovered by
any user created class with no SubnetSelectorTerms, correct?


>// terms set (ID or tag). We skip the default because its subnets are already handled
		// by the control-plane subnet configuration and re-injecting them here can cause AWSEndpointService modifications to fail if those subnets
	// land in AZs unsupported by the endpoint service NLB).

I think looking at the status is the right thing to do. The service endpoint controller underlying logic should consolidate the input if there's any overlap between karpenter subnets ande regular nodepool subnets.
To ensure private link zone matching requirement regardless of AZ alias differences, the management cluster must have subnets and worker nodes present in all zones of its region

hypershift-operator/controllers/platform/aws/controller.go
```diff
@@ -309,7 +346,7 @@ func (r *AWSEndpointServiceReconciler) Reconcile(ctx context.Context, req ctrl.R
 	return ctrl.Result{RequeueAfter: 5 * time.Minute}, nil
 }
 
-func reconcileAWSEndpointService(ctx context.Context, c client.Client, awsEndpointService *hyperv1.AWSEndpointService, hc *hyperv1.HostedCluster) error {
+func reconcileAWSEndpointService(ctx context.Context, c client.Client, ec2Client ec2iface.EC2API, awsEndpointService *hyperv1.AWSEndpointService, hc *hyperv1.HostedCluster) error {
```
enxebre: ec2Client ec2iface.EC2API seems unused

hypershift-operator/controllers/platform/aws/controller.go
```diff
@@ -97,6 +100,9 @@ func (r *AWSEndpointServiceReconciler) SetupWithManager(mgr ctrl.Manager) error
 			DeleteFunc: r.enqueueOnNodePoolDelete(mgr),
 		}).
 		Watches(&hyperv1.HostedCluster{}, handler.Funcs{UpdateFunc: r.enqueueOnHostedClusterChange(mgr)}).
+		Watches(&corev1.ConfigMap{}, handler.Funcs{
```
enxebre, line 107: can we leverage 
```
Cache: cache.Options{
			ByObject:  {
                Label: targetLabel,
		},
```
to let the informer of this manager watch only this specific configMap?

enxebre:
>If it goes bad, they will just find out when their node doesn't join

AWSEndpointServiceStatus is already accessible for Service Provider persona, we could push some of that info down into the nodeclass if we choose to and make it accessible to the Cluster Admin persona.

hypershift-operator/controllers/platform/aws/controller.go
jkyros: Assuming I am not an idiot, I don' t think we can :disappointed:

 I put a predicate on it so we weren't triggering on everything, but that configMap informer/cache looks like it's shared by the other things on the manager (e.g. NodePool reconciler) and they are using them to cast a much wider net for configmaps than we are here ( so if we limit it for this, we break them). 

(And even if we did split out a separate manager/cache the "old" one would still have all the stuff in it so we'd be limiting "our" cache while "their" cache still filled up with a bunch of configmaps, wouldn't it?)

jkyros:
Rebased, addressed review feedback (removed nil guarding logic, reduce configmap watch triggers), adjusted e2e test to create an arbitrary AWS subnet for existing cluster vs spinning up a whole separate cluster (42mins -> 12mins).

jkyros:
My e2e test did not take into account that which AWSEndpointService the subnet ended up in depended on how we configured routing, now it does :sweat_smile: 
/test e2e-aws-autonode

hypershift-operator/controllers/platform/aws/controller.go
jkyros: Thanks, removed

jkyros:
We passed, but we didn't wait for the subnet to be removed totally from the endpoint service before we deleted it, so we hit `InvalidSubnetId.NotFound` and the "did we break it?" checker at the end of the test failed us. Now I'm waiting for it to come totally out. 
/test e2e-aws-autonode

jkyros:
Pass rate on that is like 50/50, I can't imagine I broke it, it just timed out on cleanup
/test e2e-aws

karpenter-operator/controllers/nodeclass/ec2_nodeclass_controller.go
jkyros: I adjusted this to just take all the subnets from statuses and send them up, they get deduped. That AWS private link controller doesn't protect itself though. Like if the user tags a subnet that would be a duplicate/overlapping range we get a `DuplicateSubnetsInSameZone` error and it barfs the controller until we take it out.

To your comment below: 
     
>AWSEndpointServiceStatus is already accessible for Service Provider persona, we could push some of that info down into the nodeclass if we choose to and make it accessible to the Cluster Admin persona.

That feels like what we need to get to, e.g. some conditions feedback on OpenShiftEC2Nodeclass that says "you fed us this subnet that broke it, take it out" but we don't have the granularity yet.

jkyros:
Really, quay? Really? 
```
--- FAIL: TestGetImageArchitecture (0.42s)
    --- FAIL: TestGetImageArchitecture/Get_ppc64le_from_ppc64le_image;_no_err (0.12s)
        util_test.go:601: 
            Expected
                <*fmt.wrapError | 0xc000ae0400>: 
                failed to look up image metadata for quay.io/openshift-release-dev/ocp-release:4.16.11-ppc64le: failed to obtain root manifest for quay.io/openshift-release-dev/ocp-release:4.16.11-ppc64le: received unexpected HTTP status: 502 Bad Gateway
                {
                    msg: "failed to look up image metadata for quay.io/openshift-release-dev/ocp-release:4.16.11-ppc64le: failed to obtain root manifest for quay.io/openshift-release-dev/ocp-release:4.16.11-ppc64le: received unexpected HTTP status: 502 Bad Gateway",
                    err: <*fmt.wrapError | 0xc000ae03e0>{
                        msg: "failed to obtain root manifest for quay.io/openshift-release-dev/ocp-release:4.16.11-ppc64le: received unexpected HTTP status: 502 Bad Gateway",
                        err: <*client.UnexpectedHTTPStatusError | 0xc00081a4e0>{
                            Status: "502 Bad Gateway",
                        },
                    },
                }
            to be nil
FAIL
```
/test unit

enxebre:
/test e2e-aws-autonode
/test e2e-aws-techpreview
/approve

hypershift-operator/controllers/platform/aws/controller.go
```diff
@@ -344,22 +409,60 @@ func listNodePools(ctx context.Context, c client.Client, nodePoolNamespace strin
...
+	karpenterSubnets, err := listKarpenterSubnetIDs(ctx, c, hcpNamespace)
+	if err != nil {
+		// Log but don't fail - ConfigMap might not exist yet
+		ctrl.LoggerFrom(ctx).V(4).Info("Failed to get Karpenter subnets, continuing with NodePool subnets only", "error", err)
+	} else if len(karpenterSubnets) > 0 {
+		subnetIDSet.Insert(karpenterSubnets...)
+	}
+
+	subnetIDs := subnetIDSet.List()
 	sort.Strings(subnetIDs)
```
muraee:
```suggestion
	subnetIDs := subnetIDSet.List()
```
`.List()` returns a sorted list already.

karpenter-operator/controllers/nodeclass/ec2_nodeclass_controller.go
```diff
@@ -408,6 +420,90 @@ func (r *EC2NodeClassReconciler) computeReadyCondition(openshiftNodeClass *hyper
...
+	// If there are no OpenshiftEC2NodeClass resources with resolved subnets,
+	// delete the ConfigMap (no NodeClasses exist, or none have subnets in status yet).
+	// The ConfigMap is also cleaned up automatically via owner reference when HCP is deleted.
+	if subnetIDSet.Len() == 0 {
+		configMap := &corev1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      karpenterutil.KarpenterSubnetsConfigMapName,
+				Namespace: r.Namespace,
+			},
+		}
```
muraee: move the definition of the configMap outside the if statement to avoid duplicate definition afterwards in #L467

karpenter-operator/controllers/nodeclass/ec2_nodeclass_controller.go
```diff
@@ -408,6 +420,90 @@ func (r *EC2NodeClassReconciler) computeReadyCondition(openshiftNodeClass *hyper
...
+		err := r.managementClient.Delete(ctx, configMap)
+		if err != nil && !apierrors.IsNotFound(err) {
+			return fmt.Errorf("failed to delete karpenter subnets configmap: %w", err)
+		}
```
muraee: can use `util.DeleteIfNeeded()`

karpenter-operator/controllers/nodeclass/ec2_nodeclass_controller.go
jkyros: Moved, thanks!

hypershift-operator/controllers/platform/aws/controller.go
jkyros: Indeed it does, removed.

karpenter-operator/controllers/nodeclass/ec2_nodeclass_controller.go
jkyros: Adjusted to use`util.DeleteIfNeeded()`

jkyros:
Adderssed Mulham's feedback. Thank you! 

I think we need https://github.com/openshift/hypershift/pull/8037 in order to pass [e2e-aws-techpreview](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_hypershift/7791/pull-ci-openshift-hypershift-main-e2e-aws-techpreview/2036382777829494784) , we're probably perma-fail until then? 
EDIT: yay that fix merged!

jkyros:
Rebased to fix aws migration conflict
/test e2e-aws-autonode
/test e2e-aws-techpreview

jkyros:
Not a good day for aks infrastructure
/test e2e-aks

jkyros:
More infra issues, unrelated. 
/test e2e-aws-4-21

PR #7803, opened 2026-02-25
OCPBUGS-77254: fix(globalps): watch Machine updates to fix node labeling race condition

judexzhu:
## Summary

- Fixes a ~15-minute delay in scheduling the `global-pull-secret-syncer` DaemonSet pod on newly created nodes
- Adds a watch on `capiv1.Machine` updates on the management cluster, triggering reconciliation when `Machine.Status.NodeRef` transitions from nil to non-nil
- Adds a unit test verifying that nodes with nil `Machine.Status.NodeRef` are correctly skipped during labeling

## Problem

The globalps controller had a race condition: when a new Node was created on the hosted cluster, the controller received the Node CREATE event and ran `labelNodesForGlobalPullSecret()`. However, the CAPI Machine controller on the management cluster had not yet set `Machine.Status.NodeRef`, so the new node could not be mapped to its Machine and was skipped for labeling with `hypershift.openshift.io/nodepool-globalps-enabled: "true"`.

With no Machine watch, no Node update handling, and no `RequeueAfter`, the node remained unlabeled until an unrelated event (e.g., a kube-system Secret change) triggered a reconcile -- often ~15 minutes later.

## Solution

Watch `capiv1.Machine` on the CP cluster cache with a predicate that only fires when `Status.NodeRef` transitions from nil to non-nil. This ensures the controller reconciles at the exact moment CAPI links the Machine to the Node, so `labelNodesForGlobalPullSecret` can label the node and the DaemonSet pod is scheduled promptly.

## Test plan

- [x] New unit test: "When Machine.Status.NodeRef is nil it should skip the node for labeling"
- [x] All existing globalps unit tests pass with race detection
- [x] `make lint-fix` passes with 0 issues

## Fixes

- [OCPBUGS-77254](https://issues.redhat.com/browse/OCPBUGS-77254)

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **Tests**
  * Added unit tests covering node-labeling behavior and event predicates, including cases where machines lack node references and when node references appear.

* **Improvements**
  * Controller now watches machine state changes and reconciles when a machine becomes linked to a node, ensuring timely labeling and consistent cluster state.

### DIFF
--- a/control-plane-operator/hostedclusterconfigoperator/controllers/globalps/setup.go
+++ b/control-plane-operator/hostedclusterconfigoperator/controllers/globalps/setup.go
@@ -11,6 +11,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/rest"
 
+	capiv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	crclient "sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
@@ -156,5 +157,40 @@ func Setup(ctx context.Context, opts *operator.HostedClusterConfigOperatorConfig
 		return fmt.Errorf("failed to watch nodes: %w", err)
 	}
 
+	// Watch for Machine updates on the management cluster.
+	// When CAPI sets Machine.Status.NodeRef (linking a Machine to a Node),
+	// we need to reconcile so that labelNodesForGlobalPullSecret can label
+	// the newly linked node. This closes a race where the Node CREATE event
+	// fires before NodeRef is set, causing the node to be skipped.
+	if err := c.Watch(source.Kind(opts.CPCluster.GetCache(), &capiv1.Machine{},
+		// The reconciler does not use the request key; it always runs a full reconciliation.
+		// We enqueue the Machine's name/namespace for logging and observability only.
+		handler.TypedEnqueueRequestsFromMapFunc(func(ctx context.Context, o *capiv1.Machine) []crreconcile.Request {
+			return []crreconcile.Request{{NamespacedName: types.NamespacedName{Name: o.GetName(), Namespace: o.GetNamespace()}}}
+		}),
+		machineNodeRefPredicate(),
+	)); err != nil {
+		return fmt.Errorf("failed to watch Machines: %w", err)
+	}
+
 	return nil
+}
+
+// machineNodeRefPredicate returns a predicate that only triggers on Machine
+// updates where Status.NodeRef transitions from nil to non-nil.
+func machineNodeRefPredicate() predicate.TypedPredicate[*capiv1.Machine] {
+	return predicate.TypedFuncs[*capiv1.Machine]{
+		CreateFunc: func(e event.TypedCreateEvent[*capiv1.Machine]) bool {
+			return false
+		},
+		UpdateFunc: func(e event.TypedUpdateEvent[*capiv1.Machine]) bool {
+			return e.ObjectOld.Status.NodeRef == nil && e.ObjectNew.Status.NodeRef != nil
+		},
+		DeleteFunc: func(e event.TypedDeleteEvent[*capiv1.Machine]) bool {
+			return false
+		},
+		GenericFunc: func(e event.TypedGenericEvent[*capiv1.Machine]) bool {
+			return false
+		},
+	}
 }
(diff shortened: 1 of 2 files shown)


enxebre:
why are we labeling machines and let it propagate? why don't label nodes directly?
edit:
why do we need labelNodesForGlobalPullSecret at all? we can propagate this label down at machineDeployment reconciliation.

judexzhu:
Hi @enxebre . I have no context why `labelNodesForGlobalPullSecret` there at the first place, I think this PR is a simple and safe bug fix for the currently architect design.

I think your idea is a better long-term design, (correct me if I understand wrong), when nodepool created, NodePool controller check if the upgrade Replace strategy, if true, directly add the label into the node template.

However this looks might need update changes the labeling mechanism cross multiple controller/operators and touches code paths used by every NodePool. I think if could, we could open another story for follow up architecture design improvement

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/setup.go
```diff
@@ -156,5 +157,37 @@ func Setup(ctx context.Context, opts *operator.HostedClusterConfigOperatorConfig
 		return fmt.Errorf("failed to watch nodes: %w", err)
 	}
 
+	// Watch for Machine updates on the management cluster.
+	// When CAPI sets Machine.Status.NodeRef (linking a Machine to a Node),
+	// we need to reconcile so that labelNodesForGlobalPullSecret can label
+	// the newly linked node. This closes a race where the Node CREATE event
+	// fires before NodeRef is set, causing the node to be skipped.
+	if err := c.Watch(source.Kind[crclient.Object](opts.CPCluster.GetCache(), &capiv1.Machine{},
```
jparrill: why not directly watch over `*capiv1.Machine`?

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/setup.go
```diff
@@ -156,5 +157,37 @@ func Setup(ctx context.Context, opts *operator.HostedClusterConfigOperatorConfig
...
+				if !ok {
+					return false
+				}
+				newMachine, ok := e.ObjectNew.(*capiv1.Machine)
+				if !ok {
+					return false
+				}
+				return oldMachine.Status.NodeRef == nil && newMachine.Status.NodeRef != nil
+			},
+			DeleteFunc: func(e event.DeleteEvent) bool {
```
jparrill: I recommend to add `GenericFunc` and return `false` explicitly, in order to avoid unnecessary reconciliations.

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/setup.go
```diff
@@ -156,5 +157,37 @@ func Setup(ctx context.Context, opts *operator.HostedClusterConfigOperatorConfig
...
 	}
 
+	// Watch for Machine updates on the management cluster.
+	// When CAPI sets Machine.Status.NodeRef (linking a Machine to a Node),
+	// we need to reconcile so that labelNodesForGlobalPullSecret can label
+	// the newly linked node. This closes a race where the Node CREATE event
+	// fires before NodeRef is set, causing the node to be skipped.
+	if err := c.Watch(source.Kind[crclient.Object](opts.CPCluster.GetCache(), &capiv1.Machine{},
+		handler.EnqueueRequestsFromMapFunc(func(ctx context.Context, o crclient.Object) []crreconcile.Request {
+			return []crreconcile.Request{{NamespacedName: types.NamespacedName{Name: o.GetName(), Namespace: o.GetNamespace()}}}
```
jparrill, line 169: Do you mind add a comment on this line clarifying that the reconciler does not use the request key? Something like:

```
// The reconciler does not use the request key; it always runs a full reconciliation.
// We enqueue the Machine's name/namespace for logging and observability only.
```

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/setup.go
```diff
@@ -156,5 +157,37 @@ func Setup(ctx context.Context, opts *operator.HostedClusterConfigOperatorConfig
...
+		}),
+		predicate.Funcs{
```
jparrill: Do you mind cover the predicate functions with unit testing?

control-plane-operator/hostedclusterconfigoperator/controllers/globalps/setup.go
judexzhu: Done

PR #7815, opened 2026-02-26
AUTOSCALE-557: Allow configuration of CapacityReservationSelectorTerms on OpenShiftEC2NodeClass

jkyros:

This PR: 
- Exposes CapacityReservationSelectorTerms on OpenShiftEC2Nodeclass the same way they are on ec2nodeclass
- Adds plumbing to reconcile CapacityReservationSelectorTerms to the underlying ec2nodeclass
- Adds `iam:ec2:DescribeCapacityReservations` so Karpenter can see the capacity reservations and use the ReservedCapacity feature 
- Adds an e2e test to verify that a node will be created and join under a capacity reservation configured via OpenShiftEC2Nodeclass

Because: 
- The ReservedCapacity gate is beta in 1.8 (we've rebased to it) and on by default, and we want to be able to actually use it in hypershift 

Fixes [AUTOSCALE-557](https://issues.redhat.com/browse/AUTOSCALE-557)

## Checklist:
- [] This change includes docs. 

## Summary by CodeRabbit

* **New Features**
  * Capacity reservation selection for OpenShift EC2 NodeClasses and visibility of observed capacity reservations in status.
  * Declarative builder/configuration support for capacity reservation objects and selector terms.

* **Tests**
  * Unit tests for spec/status propagation and end-to-end tests that create and verify EC2 capacity reservations and instance placement.

* **Chores**
  * AWS IAM policy extended to allow reading capacity reservation details.


jkyros:
rebased
/test e2e-aws-autonode

api/karpenter/v1beta1/karpenter_types.go
```diff
@@ -111,6 +111,13 @@ type OpenshiftEC2NodeClassSpec struct {
 	// +optional
 	SecurityGroupSelectorTerms []SecurityGroupSelectorTerm `json:"securityGroupSelectorTerms,omitempty"`
 
+	// CapacityReservationSelectorTerms is a list of capacity reservation selector terms. The terms are ORed.
+	// +kubebuilder:validation:XValidation:message="expected at least one, got none, ['tags', 'id', 'instanceMatchCriteria']",rule="self.all(x, has(x.tags) || has(x.id) || has(x.instanceMatchCriteria))"
+	// +kubebuilder:validation:XValidation:message="'id' is mutually exclusive, cannot be set along with other fields in a capacity reservation selector term",rule="!self.all(x, has(x.id) && (has(x.tags) || has(x.ownerID) || has(x.instanceMatchCriteria)))"
+	// +kubebuilder:validation:MaxItems:=30
+	// +optional
+	CapacityReservationSelectorTerms []CapacityReservationSelectorTerm `json:"capacityReservationSelectorTerms,omitempty"`
```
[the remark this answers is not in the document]
jkyros: That's a 1:1 copy from upstream karpenter's type though, isn't it?  We're going for parity so we need to be "broken" the same way the upstream API is, right?

test/e2e/karpenter_test.go
```diff
@@ -655,6 +656,216 @@ func TestKarpenter(t *testing.T) {
...
+				Spec: hyperkarpenterv1.OpenshiftEC2NodeClassSpec{
+					CapacityReservationSelectorTerms: []hyperkarpenterv1.CapacityReservationSelectorTerm{
+						{ID: crID},
+					},
+					// Mirror the cluster-wide public IP setting so the node can reach the ignition
+					// server. Without this, nodes in subnets with MapPublicIpOnLaunch=false get no
+					// public IP and no route out, causing ignition fetch to time out. The default
+					// OpenshiftEC2NodeClass gets this set automatically by the operator; user-created
+					// NodeClasses like this one do not.
+					AssociatePublicIPAddress: ptr.To(hostedCluster.Annotations[hyperv1.AWSMachinePublicIPs] == "true"),
```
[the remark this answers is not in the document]
jkyros: I took this out, this was just in there from before we adjusted the test setup + subnets

test/e2e/karpenter_test.go
```diff
@@ -655,6 +655,209 @@ func TestKarpenter(t *testing.T) {
...
+			// Using a separate object avoids contaminating the shared "default" class used by other sub-tests.
+			crNodeClass := &hyperkarpenterv1.OpenshiftEC2NodeClass{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "capacity-reservation-test",
+				},
+				Spec: hyperkarpenterv1.OpenshiftEC2NodeClassSpec{
+					SubnetSelectorTerms: []hyperkarpenterv1.SubnetSelectorTerm{
+						{Tags: map[string]string{"karpenter.sh/discovery": hostedCluster.Spec.InfraID}},
+					},
+					SecurityGroupSelectorTerms: []hyperkarpenterv1.SecurityGroupSelectorTerm{
```
maxcao13: Is this gonna have to change when https://github.com/openshift/hypershift/pull/7933 merges? Or are both the new default subnetselectorterms and the old defaults going to match the same set of subnets that this doesn't matter?

test/e2e/karpenter_test.go
```diff
@@ -655,6 +655,209 @@ func TestKarpenter(t *testing.T) {
...
+			}()
+
+			// Verify the downstream EC2NodeClass has the CapacityReservationSelectorTerms propagated.
+			t.Logf("Waiting for EC2NodeClass capacity-reservation-test to have CapacityReservationSelectorTerms set")
+			g.Eventually(func(g Gomega) {
+				ec2nc := &awskarpenterv1.EC2NodeClass{}
+				g.Expect(guestClient.Get(ctx, crclient.ObjectKey{Name: "capacity-reservation-test"}, ec2nc)).To(Succeed())
+				g.Expect(ec2nc.Spec.CapacityReservationSelectorTerms).To(HaveLen(1))
+				g.Expect(ec2nc.Spec.CapacityReservationSelectorTerms[0].ID).To(Equal(crID))
+			}).WithTimeout(2 * time.Minute).WithPolling(5 * time.Second).Should(Succeed())
```
maxcao13: Could we use `e2eutil.EventuallyObject` instead?
Same with the openshiftec2nodeclass check.

test/e2e/karpenter_test.go
```diff
@@ -655,6 +655,209 @@ func TestKarpenter(t *testing.T) {
...
+					"key":      "node.kubernetes.io/instance-type",
+					"operator": "In",
+					"values":   []interface{}{"t3.large"},
+				},
+				map[string]interface{}{
+					"key":      "karpenter.sh/capacity-type",
+					"operator": "In",
+					"values":   []interface{}{"reserved"},
+				},
+			}
```
maxcao13, line 901: We should probably stop using unstructured 😛 . That'll be part of my refactor later.

test/e2e/karpenter_test.go
```diff
@@ -655,6 +655,209 @@ func TestKarpenter(t *testing.T) {
...
+			defer func() {
+				if err := guestClient.Delete(ctx, crWorkload); err != nil {
+					t.Logf("warning: failed to delete workload capacity-reservation-web-app: %v", err)
+				}
+			}()
+			// Retry the create to tolerate transient admission plugin timeouts on the guest API server.
+			g.Eventually(func(g Gomega) {
+				crWorkload.SetResourceVersion("")
+				g.Expect(guestClient.Create(ctx, crWorkload)).To(Succeed())
+			}).WithTimeout(2 * time.Minute).WithPolling(10 * time.Second).Should(Succeed())
```
maxcao13: Have you seen these failures happen? We should probably do this for all creates if so, but not in this PR.

test/e2e/karpenter_test.go
```diff
@@ -655,6 +655,209 @@ func TestKarpenter(t *testing.T) {
...
+			t.Logf("Created workload capacity-reservation-web-app to trigger node provisioning")
+
+			// Wait for the node to be ready.
+			crNodeLabels := map[string]string{
+				"karpenter.sh/nodepool": crNodePool.GetName(),
+			}
```
maxcao13: I think we can re-use this for the nodeSelector as well.

test/e2e/karpenter_test.go
jkyros: _Something_ was happening, that was the theory at least but I never 100% root caused it.  I had > 1 guest cluster running on my master when I was testing that a couple weeks ago and I was getting inconsistent behavior until I added that. I should probably take it out.

test/e2e/karpenter_test.go
jkyros: I think the `PublicOnly=False` prevents the public subnets from getting discovery tagged, which is why e.g. the OpenShift Version test which sets the same selectors also passed #7933. 

I would expect if we undo `PublicOnly=False` we would need to change these selectors to match the default private-only selectors because we might have a public subnet.  

I set this up around the time we were still working through the #7933 business so I think I can/should let these default now.

test/e2e/karpenter_test.go
jkyros: _When in Rome.._ 
( I would love to have structured types here, I really would.  )

test/e2e/karpenter_test.go
jkyros: excellent catch

test/e2e/karpenter_test.go
jkyros: Removed.

test/e2e/karpenter_test.go
jkyros: Yes, did them both, thanks!

test/e2e/karpenter_test.go
jkyros: Selectors removed, using the default selectors now.

jkyros:
Rebased, review fixes per Max (thanks Max :smile: )

jkyros:
This passed locally on my clusters, but I did change the selectors, so just in case...
/test e2e-aws-autonode

karpenter-operator/controllers/nodeclass/ec2_nodeclass_controller.go
```diff
@@ -363,6 +364,20 @@ func (r *EC2NodeClassReconciler) reconcileStatus(ctx context.Context, ec2NodeCla
...
+			AvailabilityZone:      cr.AvailabilityZone,
+			EndTime:               cr.EndTime,
+			ID:                    cr.ID,
+			InstanceMatchCriteria: cr.InstanceMatchCriteria,
+			InstanceType:          cr.InstanceType,
+			OwnerID:               cr.OwnerID,
+			ReservationType:       string(cr.ReservationType),
+			State:                 string(cr.State),
+		})
+	}
```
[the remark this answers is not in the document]
jkyros: Fixed + added a unit test case. Surprised we didn't see that one earlier.

test/e2e/util/aws.go
```diff
@@ -176,3 +178,60 @@ func CleanupOIDCBucketObjects(ctx context.Context, log logr.Logger, s3Client aws
...
+
+	crID := aws.StringValue(result.CapacityReservation.CapacityReservationId)
+
+	if err := wait.PollUntilContextTimeout(ctx, 5*time.Second, 3*time.Minute, true, func(ctx context.Context) (bool, error) {
+		desc, err := ec2Client.DescribeCapacityReservationsWithContext(ctx, &ec2.DescribeCapacityReservationsInput{
+			CapacityReservationIds: []*string{aws.String(crID)},
+		})
+		if err != nil {
+			return false, nil // transient error, retry
+		}
```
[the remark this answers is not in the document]
jkyros: I'm unconvinced. There could be propagation lag for perms since we're setting them up in the test,  and it's only 3 minutes. I'd wait 3 minutes to find out our code was broken.

test/e2e/karpenter_test.go
```diff
@@ -700,6 +700,218 @@ func TestKarpenter(t *testing.T) {
...
 		})
 
+		t.Run("Capacity reservation selector propagation", func(t *testing.T) {
+			g := NewWithT(t)
+
+			// Ensure the HostedCluster has the karpenter role ARN configured.
+			if hostedCluster.Spec.AutoNode == nil ||
+				hostedCluster.Spec.AutoNode.Provisioner.Karpenter == nil ||
+				hostedCluster.Spec.AutoNode.Provisioner.Karpenter.AWS == nil {
+				t.Skip("HostedCluster does not have a Karpenter AWS role configured, skipping capacity reservation test")
```
enxebre, line 786: Maybe worth a comment this should never happen since this is required the API level.

test/e2e/karpenter_test.go
```diff
@@ -700,6 +700,218 @@ func TestKarpenter(t *testing.T) {
...
+			}
+			karpenterRoleARN := hostedCluster.Spec.AutoNode.Provisioner.Karpenter.AWS.RoleARN
+
+			// Temporarily grant the karpenter role permission to create and cancel capacity reservations.
+			// ec2:DescribeCapacityReservations is already permanently in the policy (added to AllowRegionalReadActions).
+			crPolicy := `{
```
enxebre: We run the tests using the public ROSA managed policies to catch issues earlier.
We do this pattern in other tests when the manage policies doesn't have a permission yet that we want to exercise ahead.

In this case there's no karpenter managed policy, the test run without own. So you can just add the new needed permission there https://github.com/openshift/hypershift/blob/4f13e843100baf7bd9df1658f91ffb02d1d59a76/cmd/infra/aws/iam.go#L239

And we need to make sure this is included in the public karpenter policy.
We also need to update the link above to start using the public managed policy in these e2e

jkyros: @enxebre I was only granting Create/Cancel so I could create CapacityReservations for the test (but I don't think I even needed to do that because the test credentials can).  All Karpenter really needs in order to use this feature is Describe, the CapacityReservation needs to already exist. 

Are you saying we should add Create/Cancel to the Karpenter (and eventually ROSA) policy because we know ROSA will need to use them in order to create the CapacityReservations that Karpenter will use (Even if Karpenter itself does not use them) ? Or do I misunderstand?

jkyros: I removed the Create/Cancel from the test karpenter policy as it appeared to be unnecessary, and I have confirmed with the ROSA folks that the managed policy has the required perms for this to function, including Create/Cancel: https://github.com/openshift/managed-cluster-config/blob/903d32846d2b8f6d4a6ddd8b4a5543ec7871c869/resources/sts/hypershift/openshift_hcp_karpenter_controller_credentials_policy.json#L9

test/e2e/karpenter_test.go
jkyros: Commented. I can pull that whole check if you think it's extra.

jkyros:
Rebased and pulled out the extra Create/Cancel CapacityReservation perms from the test, did verify with ROSA that they are present in the managed policy.

api/karpenter/v1beta1/karpenter_types.go
```diff
@@ -203,6 +203,15 @@ type OpenshiftEC2NodeClassSpec struct {
 	// +optional
 	SecurityGroupSelectorTerms []SecurityGroupSelectorTerm `json:"securityGroupSelectorTerms,omitempty"`
 
+	// capacityReservationSelectorTerms is a list of capacity reservation selector terms. The terms are ORed.
+	// +kubebuilder:validation:XValidation:message="expected at least one, got none, ['tags', 'id', 'instanceMatchCriteria']",rule="self.all(x, has(x.tags) || has(x.id) || has(x.instanceMatchCriteria))"
+	// +kubebuilder:validation:XValidation:message="'id' is mutually exclusive, cannot be set along with other fields in a capacity reservation selector term",rule="!self.all(x, has(x.id) && (has(x.tags) || has(x.ownerID) || has(x.instanceMatchCriteria)))"
```
enxebre, line 211: can we express all these validations as part of the gdoc above so it's readable for api consumers

api/karpenter/v1beta1/karpenter_types.go
```diff
@@ -307,6 +316,38 @@ type SecurityGroupSelectorTerm struct {
...
+	// ownerID is the owner account id for the capacity reservation.
+	// +kubebuilder:validation:XValidation:rule="self.matches('^[0-9]{12}$')",message="ownerID must be a 12-digit AWS account ID"
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=12
+	// +optional
+	OwnerID string `json:"ownerID,omitempty"`
+
+	// instanceMatchCriteria specifies how instances are matched to capacity reservations.
+	// +kubebuilder:validation:Enum:={open,targeted}
+	// +optional
```
enxebre, line 363: nit: please use pascalCase for enums, Open/Targeted.
can we document what's the behaviour when ommited?

api/karpenter/v1beta1/karpenter_types.go
```diff
@@ -400,10 +441,54 @@ type BlockDevice struct {
...
+type CapacityReservation struct {
+	// availabilityZone is the availability zone the capacity reservation is available in.
+	// +required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=64
+	AvailabilityZone string `json:"availabilityZone,omitempty"`
+	// endTime is the time at which the capacity reservation expires. Once expired, the reserved capacity is released and Karpenter
+	// will no longer be able to launch instances into that reservation.
+	// +optional
+	EndTime *metav1.Time `json:"endTime,omitempty"`
```
enxebre: should this be not a pointer and use omitzero?

enxebre:
Thanks! dropped some minor comments that needs addressing before merging
/approve

Let's make sure to run 
e2e-aws-autonode
e2e-aws-techpreview
after pushing changes

api/karpenter/v1beta1/karpenter_types.go
```diff
@@ -400,10 +434,40 @@ type BlockDevice struct {
...
+	InstanceType string `json:"instanceType"`
+	// The ID of the AWS account that owns the capacity reservation.
+	// +required
+	OwnerID string `json:"ownerID"`
+	// The type of capacity reservation.
+	// +optional
+	ReservationType string `json:"reservationType,omitempty"`
+	// The state of the capacity reservation.
+	// +optional
+	State string `json:"state,omitempty"`
```
muraee: seems like both fields should be enums with defined values.

jkyros: enum'd both of these

api/karpenter/v1beta1/karpenter_types.go
jkyros: converted to pointer + omitzero

api/karpenter/v1beta1/karpenter_types.go
jkyros: pascal'd and added some doc on what happens if omitted (it selects both open and targeted, doesn't filter)

api/karpenter/v1beta1/karpenter_types.go
jkyros: added some doc to explain the validations

jkyros:
Now that this is fixed, we can: 
/test e2e-aws-techpreview

jkyros:
@jkyros thought it would be fun to do aws stuff in the middle of an aws migration. _Now_ we should be able to run techpreview. 
/test e2e-aws-autonode
/test e2e-aws-techpreview

jkyros:
Network pod restarted, looked like infra failure, I didn't crash it 
/test e2e-aws-techpreview

api/karpenter/v1beta1/karpenter_types.go
enxebre: The gdoc when looking at the API docs would look like
"// CapacityReservationSelectorTerms is a list of capacity reservation selector terms. The terms are ORed."
I'm suggesting we add the additional context you included in the XValidation for the docs as well, e.g.
"// CapacityReservationSelectorTerms is a list of capacity reservation selector terms. The terms are ORed. One of 'tags', 'id', 'instanceMatchCriteria' is required. id is mutually exclusive"

enxebre: also // CapacityReservationSelectorTerms shuold be start with lower case.

mm why is the linter not cathing this one?

api/karpenter/v1beta1/karpenter_types.go
```diff
@@ -203,6 +203,13 @@ type OpenshiftEC2NodeClassSpec struct {
 	// +optional
 	SecurityGroupSelectorTerms []SecurityGroupSelectorTerm `json:"securityGroupSelectorTerms,omitempty"`
 
+	// CapacityReservationSelectorTerms is a list of capacity reservation selector terms. The terms are ORed.
+	// +kubebuilder:validation:XValidation:message="expected at least one, got none, ['tags', 'id', 'instanceMatchCriteria']",rule="self.all(x, has(x.tags) || has(x.id) || has(x.instanceMatchCriteria))"
+	// +kubebuilder:validation:XValidation:message="'id' is mutually exclusive, cannot be set along with other fields in a capacity reservation selector term",rule="!self.all(x, has(x.id) && (has(x.tags) || has(x.ownerID) || has(x.instanceMatchCriteria)))"
+	// +kubebuilder:validation:MaxItems:=30
```
enxebre, line 213: minItems?

api/karpenter/v1beta1/karpenter_types.go
enxebre: ignore this. I was looking at older commit. All seems adressed. Thanks

api/karpenter/v1beta1/karpenter_types.go
enxebre: ignore this. I was looking at older commit. All seems adressed. Thanks

jkyros:
Metadata made it in, but that conflicted us. Had to rebase.

jkyros:
Had to rebase after privatelink merged
/test e2e-aws-autonode
/test e2e-aws-techpreview

jkyros:
looks like teardown didn't finish for some AWS resources (not related to this)
/test e2e-aws

PR #7824, opened 2026-02-27
GCP-431: feat: Add CNCC support for GCP WIF in HyperShift

apahim:
**What this PR does / why we need it:**                                                                                                                                      
                                                                                                                                                                       
Adds CNCC (Cloud Network Config Controller) support for GCP Workload Identity Federation in HyperShift HCP mode. Currently HyperShift does not provision WIF credentials 
for CNCC on GCP, so CNCC cannot authenticate to GCP APIs to manage egress IPs and subnets.                                                                               
                                                                                                                                                                         
This PR:                                                                                                                                                                 
- Adds a Network field to GCPServiceAccountsEmails API for the CNCC service account                                                                                      
- Adds a cloud-network service account entry to iam-bindings.json with fine-grained CNCC permissions (compute.instances.get, compute.subnetworks.use, etc.)              
- Provisions a cloud-network-config-controller-creds secret in ReconcileCredentials when the Network GSA email is configured                                             
- Sets GCP_CNCC_CREDENTIALS_FILE env var on the CNO deployment for GCP platform, so CNO can wire GOOGLE_APPLICATION_CREDENTIALS on the CNCC container                    

**Which issue(s) this PR fixes:**

Fixes [GCP-431](https://issues.redhat.com/browse/GCP-431)

**Special notes for your reviewer:**

This PR is part of a cross-repo effort:
- **CNO** (https://issues.redhat.com/browse/GCP-430): Reads GCP_CNCC_CREDENTIALS_FILE and sets GOOGLE_APPLICATION_CREDENTIALS on the CNCC container, plus adds
--token-audience=openshift to the cloud-token minter
- **CNCC** (https://github.com/openshift/cloud-network-config-controller/pull/206): Adds WIF auto-detection via readGCPCredentialsConfig()
- **This PR**: Provisions the credential secret and env var that connect the other two

The Network field is +required, consistent with the other GSA fields. The iam-bindings.json entry uses predefined roles (roles/compute.instanceAdmin.v1 + roles/compute.networkUser), following the same pattern as the other service accounts.

**Checklist:**

- Subject and description added to both, commit and PR.
- Relevant issues have been referenced.
- This change includes docs.
- This change includes unit tests.


<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **New Features**
  * Added a GCP network service-account field for hosted clusters with immutability and project-scoped validation.
  * Enabled Cloud Network Config Controller (CNCC) support: new cloud-network service account, automatic CNCC credential provisioning, and a control-plane env var to expose CNCC credentials.
  * Added a CLI flag to supply the network service account during cluster creation.
* **Tests**
  * Added serialization/compatibility and controller tests for the new network field.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/api/hypershift/v1beta1/gcp.go
+++ b/api/hypershift/v1beta1/gcp.go
@@ -105,6 +105,7 @@ type GCPNetworkConfig struct {
 // +kubebuilder:validation:XValidation:rule="self.workloadIdentity.serviceAccountsEmails.cloudController.contains('@') && self.workloadIdentity.serviceAccountsEmails.cloudController.endsWith('@' + self.project + '.iam.gserviceaccount.com')",message="cloudController service account must belong to the same project"
 // +kubebuilder:validation:XValidation:rule="self.workloadIdentity.serviceAccountsEmails.storage.contains('@') && self.workloadIdentity.serviceAccountsEmails.storage.endsWith('@' + self.project + '.iam.gserviceaccount.com')",message="storage service account must belong to the same project"
 // +kubebuilder:validation:XValidation:rule="self.workloadIdentity.serviceAccountsEmails.imageRegistry.contains('@') && self.workloadIdentity.serviceAccountsEmails.imageRegistry.endsWith('@' + self.project + '.iam.gserviceaccount.com')",message="imageRegistry service account must belong to the same project"
+// +kubebuilder:validation:XValidation:rule="self.workloadIdentity.serviceAccountsEmails.network.endsWith('@' + self.project + '.iam.gserviceaccount.com')",message="network service account must belong to the same project"
 type GCPPlatformSpec struct {
 	// project is the GCP project ID.
 	// A valid project ID must satisfy the following rules:
@@ -357,6 +358,22 @@ type GCPServiceAccountsEmails struct {
 	// +immutable
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="ImageRegistry is immutable"
 	ImageRegistry GCPServiceAccountEmail `json:"imageRegistry,omitempty"`
+
+	// network is the Google Service Account email for the Cloud Network Config Controller
+	// that manages cloud-level network configurations (egress IPs, subnets).
+	// This GSA requires the following IAM roles:
+	// - roles/compute.instanceAdmin.v1 (Compute Instance Admin - for managing network interfaces)
+	// - roles/compute.networkUser (Compute Network User - for using subnets)
+	// See cmd/infra/gcp/iam-bindings.json for the authoritative role definitions.
+	// Format: service-account-name@project-id.iam.gserviceaccount.com
+	//
+	// This is a user-provided value referencing a pre-created Google Service Account.
+	// Typically obtained from the output of `hypershift infra create gcp` which creates
+	// the required service accounts with appropriate IAM roles and WIF bindings.
+	//
+	// +required
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Network is immutable"
+	Network GCPServiceAccountEmail `json:"network,omitempty"`
 }
 
 // GCPOnHostMaintenance defines the behavior when a host maintenance event occurs.
--- a/cmd/cluster/gcp/create_test.go
+++ b/cmd/cluster/gcp/create_test.go
@@ -39,6 +39,7 @@ func TestCreateOptionsApplyPlatformSpecifics(t *testing.T) {
 						CloudControllerServiceAccount: "cloudcontroller@test-project-123.iam.gserviceaccount.com",
 						StorageServiceAccount:         "storage@test-project-123.iam.gserviceaccount.com",
 						ImageRegistryServiceAccount:   "imageregistry@test-project-123.iam.gserviceaccount.com",
+						NetworkServiceAccount:         "network@test-project-123.iam.gserviceaccount.com",
 					},
 				},
 			},
@@ -63,6 +64,7 @@ func TestCreateOptionsApplyPlatformSpecifics(t *testing.T) {
 	g.Expect(hostedCluster.Spec.Platform.GCP.WorkloadIdentity.ServiceAccountsEmails.CloudController).To(Equal(hyperv1.GCPServiceAccountEmail("cloudcontroller@test-project-123.iam.gserviceaccount.com")))
 	g.Expect(hostedCluster.Spec.Platform.GCP.WorkloadIdentity.ServiceAccountsEmails.Storage).To(Equal(hyperv1.GCPServiceAccountEmail("storage@test-project-123.iam.gserviceaccount.com")))
 	g.Expect(hostedCluster.Spec.Platform.GCP.WorkloadIdentity.ServiceAccountsEmails.ImageRegistry).To(Equal(hyperv1.GCPServiceAccountEmail("imageregistry@test-project-123.iam.gserviceaccount.com")))
+	g.Expect(hostedCluster.Spec.Platform.GCP.WorkloadIdentity.ServiceAccountsEmails.Network).To(Equal(hyperv1.GCPServiceAccountEmail("network@test-project-123.iam.gserviceaccount.com")))
 }
 
 func TestValidateGCPOptions(t *testing.T) {
@@ -81,6 +83,7 @@ func TestValidateGCPOptions(t *testing.T) {
 		CloudControllerServiceAccount: "cloudcontroller@test-project-123.iam.gserviceaccount.com",
 		StorageServiceAccount:         "storage@test-project-123.iam.gserviceaccount.com",
 		ImageRegistryServiceAccount:   "imageregistry@test-project-123.iam.gserviceaccount.com",
+		NetworkServiceAccount:         "network@test-project-123.iam.gserviceaccount.com",
 	}
 
 	tests := map[string]struct {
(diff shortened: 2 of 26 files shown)


cristianoveiga:
The cloud-network-config-controller pod is still stuck because the e2e test doesn't pass the network service account. The PR adds --network-service-account to the CLI but the e2e pipeline is       
missing the plumbing. You'd need changes in three places:
                                                                                                                                                                                                     
1. hypershift repo (test/e2e/e2e_test.go) — add e2e.gcp-network-sa flag following the pattern at https://github.com/openshift/hypershift/blob/main/test/e2e/e2e_test.go#L175-L179                    
2. hypershift repo (test/e2e/util/options.go) — add GCPNetworkServiceAccount field and wire it to NetworkServiceAccount                                                                              
3. openshift/release repo — the https://github.com/openshift/release/tree/main/ci-operator/step-registry/hypershift/gcp/hosted-cluster-setup step needs to create the network SA and write it to     
${SHARED_DIR}/network-sa, and https://github.com/openshift/release/blob/main/ci-operator/step-registry/hypershift/gcp/run-e2e/hypershift-gcp-run-e2e-commands.sh needs to load it and pass           
--e2e.gcp-network-sa="${NETWORK_SA}"                                                                                                                                                                 
                                                                                                                                                                                                     
One way to move forward is splitting this into three PRs:                                                                                                                                                           
- hypershift PR 1: the hypershift infra create gcp changes (IAM bindings are already here) and the e2e plumbing (1, 2). This is additive and harmless when the network SA field is empty.
- openshift/release PR: the CI step changes (3).                                                                                                                                                     
- hypershift PR 2 (this one): the API type, secret creation, and CNO env var. When it merges, everything connects and e2e-gke starts passing.
                                                                                                                                                                                                     
Alternatively, you could include 1 and 2 in this PR and follow up with only the openshift/release PR (3). 

The downside is that e2e-gke won't have signal until 3 lands, but it's not a required check and is already      
failing on other PRs.

api/hypershift/v1beta1/gcp_test.go
```diff
@@ -0,0 +1,103 @@
+package v1beta1
+
+import (
+	"encoding/json"
+	"testing"
+)
+
+// gcpServiceAccountsEmailsNMinus1 represents the previous version of
+// GCPServiceAccountsEmails before the Network field was added.
+// Used to verify N-1/N+1 serialization compatibility.
```
JoelSpeed: You have `omitempty` on the new field, I don't think we need a serialization test. You are testing the json library, not your own code

api/hypershift/v1beta1/gcp.go
```diff
@@ -357,6 +358,26 @@ type GCPServiceAccountsEmails struct {
...
+	// Format: service-account-name@project-id.iam.gserviceaccount.com
+	//
+	// This is a user-provided value referencing a pre-created Google Service Account.
+	// Typically obtained from the output of `hypershift infra create gcp` which creates
+	// the required service accounts with appropriate IAM roles and WIF bindings.
+	//
+	// +optional
+	// +immutable
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Network is immutable"
+	Network GCPServiceAccountEmail `json:"network,omitempty"`
```
cristianoveiga: We are adding this as optional: is that purely for backwards compatibility (existing clusters that don't have it yet), or will there be GCP deployments that intentionally never configure a Network SA?

(the implementation plan refers to it as `required`)

control-plane-operator/controllers/hostedcontrolplane/v2/cno/deployment.go
```diff
@@ -162,5 +162,14 @@ func buildCNOEnvVars(cpContext component.WorkloadContext) ([]corev1.EnvVar, erro
...
 	}
 
+	// For GCP deployments, we pass the credentials filename for the Cloud Network Config Controller.
+	// The CNO uses this to configure the CNCC deployment it manages in the management cluster.
+	if hcp.Spec.Platform.Type == hyperv1.GCPPlatform {
+		cnoEnv = append(cnoEnv, corev1.EnvVar{
+			Name:  "GCP_CNCC_CREDENTIALS_FILE",
+			Value: "application_default_credentials.json",
+		})
+	}
```
cristianoveiga, line 169: If the Network service account stays optional, consider gating this.

The credentials secret is only created when Network is set (gcp.go:ReconcileCredentials), but this env var is always set for GCP. CNO will render GOOGLE_APPLICATION_CREDENTIALS pointing at the credential file, and without the backing secret the CNCC pod gets a FailedMount.

api/hypershift/v1beta1/gcp.go
apahim: made it required

control-plane-operator/controllers/hostedcontrolplane/v2/cno/deployment.go
apahim: made it required

cristianoveiga:
Since the new field is `required`, can you please update the GCP docs under `docs/content/how-to/gcp/`  with the new flag?

api/hypershift/v1beta1/gcp.go
```diff
@@ -357,6 +358,26 @@ type GCPServiceAccountsEmails struct {
...
+	// - compute.zoneOperations.get
+	// See cmd/infra/gcp/iam-bindings.json for the authoritative permission definitions.
+	// Format: service-account-name@project-id.iam.gserviceaccount.com
+	//
+	// This is a user-provided value referencing a pre-created Google Service Account.
+	// Typically obtained from the output of `hypershift infra create gcp` which creates
+	// the required service accounts with appropriate IAM roles and WIF bindings.
+	//
+	// +required
+	// +immutable
```
JoelSpeed: This tag doesn't actually do anything, you should remove it

api/hypershift/v1beta1/gcp.go
```diff
@@ -105,6 +105,7 @@ type GCPNetworkConfig struct {
...
 // +kubebuilder:validation:XValidation:rule="self.workloadIdentity.serviceAccountsEmails.storage.contains('@') && self.workloadIdentity.serviceAccountsEmails.storage.endsWith('@' + self.project + '.iam.gserviceaccount.com')",message="storage service account must belong to the same project"
 // +kubebuilder:validation:XValidation:rule="self.workloadIdentity.serviceAccountsEmails.imageRegistry.contains('@') && self.workloadIdentity.serviceAccountsEmails.imageRegistry.endsWith('@' + self.project + '.iam.gserviceaccount.com')",message="imageRegistry service account must belong to the same project"
+// +kubebuilder:validation:XValidation:rule="self.workloadIdentity.serviceAccountsEmails.network.contains('@') && self.workloadIdentity.serviceAccountsEmails.network.endsWith('@' + self.project + '.iam.gserviceaccount.com')",message="network service account must belong to the same project"
```
JoelSpeed: The first half of this is redundant no? The validation of the `GCPServiceAccountEmail` type would enforce this?

api/hypershift/v1beta1/gcp.go
apahim: sure, removing

api/hypershift/v1beta1/gcp.go
apahim: I was following the pattern, but indeed it's redundant

test/e2e/v2/tests/api_ux_validation_test.go
```diff
@@ -418,6 +419,11 @@ var _ = Describe("API UX Validation", Label("API"), func() {
 						spec.WorkloadIdentity.ServiceAccountsEmails.ImageRegistry = "invalid-image-registry-email"
 					},
 					"imageRegistry in body"),
+				Entry("it should reject invalid Network service account email",
```
enxebre: This should be covered by envtest now, let's drop it.
https://github.com/openshift/hypershift/pull/8210

hypershift-operator/controllers/hostedcluster/internal/platform/gcp/gcp.go
```diff
@@ -426,6 +427,17 @@ func ImageRegistryCredsSecret(controlPlaneNamespace string) *corev1.Secret {
 	}
 }
 
+// CNCCCredsSecret returns the secret containing Workload Identity Federation credentials
+// for the Cloud Network Config Controller to manage cloud-level network configurations.
+func CNCCCredsSecret(controlPlaneNamespace string) *corev1.Secret {
```
muraee, line 434: don't you need to mount this secret in the CNO deployment?

apahim: CNO only needs the filename to set GOOGLE_APPLICATION_CREDENTIALS on the CNCC container it creates. The secret is not mounted in the CNO deployment. The CNCC deployment (created by CNO) mounts it, as as with all other platforms.

test/e2e/v2/tests/api_ux_validation_test.go
apahim: Done

hypershift-operator/controllers/hostedcluster/internal/platform/gcp/gcp.go
muraee: how does the CNCC know the secret name? is it hard-coded in CNCC?

apahim: 1. https://github.com/openshift/cluster-network-operator/pull/2915
2. https://github.com/openshift/cloud-network-config-controller/pull/206

muraee: it's hardcoded here
https://github.com/openshift/cluster-network-operator/blob/bc5af87f8957262a504b225591593c92db1c8bc2/bindata/cloud-network-config-controller/managed/controller.yaml#L253

muraee: can we please add a comment with the link to explain that this name is hardcoded there

apahim: Done!

muraee:
/approve
@apahim can you please squash the commits

cmd/infra/gcp/iam-bindings.json
```diff
@@ -66,6 +66,22 @@
...
+      "name": "cloud-network",
+      "displayName": "Cloud Network Config Controller Service Account",
+      "description": "Service account for cloud network configuration (egress IP, subnet management)",
+      "permissions": [
+        "compute.instances.updateNetworkInterface",
+        "compute.instances.get",
+        "compute.subnetworks.use",
+        "compute.subnetworks.get",
+        "compute.zoneOperations.get"
+      ],
```
cblecker, line 77: This entry uses `"permissions"` but `ServiceAccountDefinition` in `iam.go` only has `Roles []string` with a `json:"roles"` tag. There's no `Permissions` field on the struct, so this entire list gets silently dropped during JSON unmarshaling.

The result is that `CreateServiceAccounts()` creates the `cloud-network` SA and sets up the WIF binding, but `len(def.Roles) > 0` is false so no IAM permissions are ever assigned. Every other SA entry uses `"roles"` and gets permissions assigned correctly — this one ends up with zero permissions.

CNCC will authenticate fine (WIF works), but every GCP API call will fail with permission denied at runtime.

Either the struct needs a `Permissions` field with custom role creation logic, or this should use predefined `"roles"` like the other entries (though I see you noted these permissions don't map cleanly to a single predefined role).

hypershift-operator/controllers/hostedcluster/internal/platform/gcp/gcp_test.go
cblecker, line 1: `TestValidateWorkloadIdentityConfiguration` has explicit "missing" test cases for every SA email (nodePool, controlPlane, storage, cloudController, imageRegistry) but not for network. The validation at `gcp.go:591-593` that checks `wif.ServiceAccountsEmails.Network == ""` is untested.

cmd/cluster/gcp/create_test.go
```diff
@@ -81,6 +83,7 @@ func TestValidateGCPOptions(t *testing.T) {
 		CloudControllerServiceAccount: "cloudcontroller@test-project-123.iam.gserviceaccount.com",
 		StorageServiceAccount:         "storage@test-project-123.iam.gserviceaccount.com",
 		ImageRegistryServiceAccount:   "imageregistry@test-project-123.iam.gserviceaccount.com",
+		NetworkServiceAccount:         "network@test-project-123.iam.gserviceaccount.com",
```
cblecker, line 86: `TestValidateGCPOptions` has "missing" test cases for cloud-controller, storage, and image-registry service accounts but not for network-service-account. Also worth noting: the existing "missing" test cases don't include `NetworkServiceAccount` in their opts, so each one is now testing with two missing required fields at once (relying on validation check ordering for the expected error).

cblecker (changes requested):
Nit: the PR description still says "The Network field is +optional" but the code has `+required` (changed per Cristiano's feedback). Might want to update the description to match.

cmd/cluster/gcp/create_test.go
apahim: done

hypershift-operator/controllers/hostedcluster/internal/platform/gcp/gcp_test.go
apahim: done

cmd/infra/gcp/iam-bindings.json
apahim: oh good catch, tks, done

apahim:
## Code Coverage Note

The `codecov/patch` drop is expected. The uncovered lines are in `cmd/infra/gcp/iam.go` — specifically the `createCustomRole()` and `deleteCustomRole()` methods and their call sites in `CreateServiceAccounts`/`DeleteServiceAccounts`. These methods make direct GCP IAM API calls (`iamService.Projects.Roles.Create/Delete/Get/Patch/Undelete`) that cannot be unit tested without mocking the GCP IAM service client.

This is consistent with the existing pattern in the same file: `CreateServiceAccounts`, `DeleteServiceAccounts`, `assignRoles`, `removeRoles`, `createWorkloadIdentityBinding`, and all other GCP API-calling methods are also not unit tested — they are exercised by e2e tests (`e2e-gke`, `e2e-v2-gke`).

The pure logic helpers (`formatCustomRoleID`, `formatCustomRoleName`) and the JSON deserialization of the `Permissions` field are covered by unit tests.

cmd/infra/gcp/iam.go
```diff
@@ -51,7 +51,11 @@ type ServiceAccountDefinition struct {
 	Description string `json:"description"`
 
 	// Roles are the GCP IAM roles to assign to this GSA
-	Roles []string `json:"roles"`
+	Roles []string `json:"roles,omitempty"`
+
+	// Permissions are individual GCP IAM permissions used to create a custom role.
+	// When specified, a project-level custom role is created and assigned to this GSA.
+	Permissions []string `json:"permissions,omitempty"`
```
cblecker: I'm concerned about the per-cluster custom role approach at scale. GCP has a hard limit of 300 custom roles per project (non-adjustable), and soft-deleted roles count against that quota for 7 days. Role IDs can't be reused for ~44 days after deletion. In a managed service scenario with cluster churn, this could exhaust the quota quickly.

Since all CNCC instances need the same 5 permissions, I'd suggest switching to predefined roles instead — `roles/compute.instanceAdmin.v1` + `roles/compute.networkUser` covers all 5 permissions and follows the same pattern as every other SA in iam-bindings.json. That would let us drop all the custom role lifecycle code (create, undelete, patch, delete) and the `Permissions` field on the struct.

For reference, `nodepool-mgmt` already has `instanceAdmin.v1` + `networkAdmin` + `serviceAccountUser`, so `instanceAdmin.v1` + `networkUser` for CNCC is less privileged than what we already grant to other SAs.

cblecker (commented):
Thanks for addressing the previous findings — the test gaps and the permissions bug are all resolved.

One remaining concern: the PR description still says "+optional" but the code has `+required` (per Cristiano's earlier feedback).

cmd/infra/gcp/iam.go
apahim: Replaced custom roles with predefined roles — the IAM bindings commit, now:                                                   
  - Uses "roles": ["roles/compute.instanceAdmin.v1", "roles/compute.networkUser"] instead of "permissions": [...]                           
  - No Permissions field on the struct, no custom role lifecycle code

api/hypershift/v1beta1/gcp.go
```diff
@@ -357,6 +358,25 @@ type GCPServiceAccountsEmails struct {
...
+
+	// network is the Google Service Account email for the Cloud Network Config Controller
+	// that manages cloud-level network configurations (egress IPs, subnets).
+	// This GSA requires a custom role with the following permissions:
+	// - compute.instances.updateNetworkInterface
+	// - compute.instances.get
+	// - compute.subnetworks.use
+	// - compute.subnetworks.get
+	// - compute.zoneOperations.get
+	// See cmd/infra/gcp/iam-bindings.json for the authoritative permission definitions.
```
cblecker: The comment says "custom role" but after the switch to predefined roles this is no longer accurate — `iam-bindings.json` now assigns `roles/compute.instanceAdmin.v1` and `roles/compute.networkUser`. The permissions listed here are what those roles cover, which is useful context, but the "requires a custom role" framing is stale from the earlier iteration. Worth updating to say something like "This GSA requires the following permissions (granted via predefined roles in `iam-bindings.json`):" so it stays accurate.

cmd/infra/gcp/iam-bindings.json
cblecker, line 1: nit: file is missing a trailing newline.

cblecker (commented):
Took a look — overall this is solid. The cross-repo wiring (HyperShift → CNO → CNCC) is well thought out and the implementation is consistent with how the other GCP service accounts are provisioned. A couple of small things below.

api/hypershift/v1beta1/gcp.go
apahim: Addressed in commit 1882865e00 — updated to say "granted via predefined roles in iam-bindings.json".

cmd/infra/gcp/iam-bindings.json
apahim: Addressed in commit 1882865e00 — added trailing newline.

api/hypershift/v1beta1/gcp.go
```diff
@@ -357,6 +358,25 @@ type GCPServiceAccountsEmails struct {
...
+
+	// network is the Google Service Account email for the Cloud Network Config Controller
+	// that manages cloud-level network configurations (egress IPs, subnets).
+	// This GSA requires the following permissions (granted via predefined roles in iam-bindings.json):
+	// - compute.instances.updateNetworkInterface
+	// - compute.instances.get
+	// - compute.subnetworks.use
+	// - compute.subnetworks.get
+	// - compute.zoneOperations.get
+	// See cmd/infra/gcp/iam-bindings.json for the authoritative permission definitions.
```
cblecker: nit: Every other field in this struct lists IAM roles in its comment (e.g. `roles/compute.loadBalancerAdmin`, `roles/storage.admin`), but this one lists individual permissions. Since the `iam-bindings.json` entry assigns `roles/compute.instanceAdmin.v1` and `roles/compute.networkUser`, the comment should reference those roles to be consistent with the sibling fields. Also the last line here says "authoritative permission definitions" while all the others say "authoritative role definitions".

docs/content/how-to/gcp/create-gcp-iam.md
```diff
@@ -58,6 +58,7 @@ The `hypershift create iam gcp` command creates WIF resources in the hosted clus
   - `cloud-controller` — Cloud Controller Manager (load balancer admin, security admin, compute viewer)
   - `storage` — GCP PD CSI Driver (storage admin, instance admin)
   - `image-registry` — Image Registry Operator (storage admin)
+  - `cloud-network` — Cloud Network Config Controller (network interface, subnet, zone operations)
```
cblecker: nit: The other entries use informal role names in parentheses — `(load balancer admin, security admin, compute viewer)`, `(storage admin, instance admin)`, `(storage admin)`. This should follow the same pattern: `(instance admin, network user)` to match the predefined roles `roles/compute.instanceAdmin.v1` and `roles/compute.networkUser`.

PR #7840, opened 2026-03-02
GCP-368: add GCP CCM v2 e2e tests

cristianoveiga:

Adds v2 e2e tests validating GCP Cloud Controller Manager node initialization. Changes:

- **Workload registry**: Register `gcp-cloud-controller-manager` deployment so existing workload tests (resource requests, security contexts, etc.) automatically cover it
- **Guest cluster client**: Add `GetGuestClient()` to `TestContext` for tests that need to inspect guest cluster state
- **GCP CCM tests**: Add `GCPCloudControllerManagerTest` to `control_plane_workloads_test.go` validating:
  - ProviderID assignment (`gce://<project>/<zone>/<instance>`)
  - Zone/region topology labels
  - Uninitialized taint removal

Tests are GCP-specific (skipped on other platforms via `BeforeEach` guard), confirmed working on AWS CI run (properly skipped, 0 failures).

Fixes [GCP-368](https://issues.redhat.com/browse/GCP-368)

- LoadBalancer service provisioning tests were descoped to a follow-up card

## Checklist:

### DIFF
--- a/test/e2e/v2/internal/test_context.go
+++ b/test/e2e/v2/internal/test_context.go
@@ -38,15 +38,15 @@ type TestContextGetter func() *TestContext
 // TestContext holds the test context including clients and hosted cluster reference
 type TestContext struct {
 	context.Context
-	MgmtClient            crclient.Client
-	ClusterName           string
-	ClusterNamespace      string
-	ControlPlaneNamespace string
-	ArtifactDir           string
-	hostedCluster         *hyperv1.HostedCluster
-	hostedClusterOnce     sync.Once
-	guestClient           crclient.Client
-	guestClientOnce       sync.Once
+	MgmtClient              crclient.Client
+	ClusterName             string
+	ClusterNamespace        string
+	ControlPlaneNamespace   string
+	ArtifactDir             string
+	hostedCluster           *hyperv1.HostedCluster
+	hostedClusterOnce       sync.Once
+	hostedClusterClient     crclient.Client
+	hostedClusterClientOnce sync.Once
 }
 
 // GetHostedCluster returns the HostedCluster associated with this test context.
@@ -78,13 +78,13 @@ func (tc *TestContext) GetHostedCluster() *hyperv1.HostedCluster {
 	return tc.hostedCluster
 }
 
-// GetGuestClient returns a controller-runtime client for the guest cluster.
+// GetHostedClusterClient returns a controller-runtime client for the hosted cluster.
 // It reads the kubeconfig from the secret referenced by the HostedCluster status.
 // The client is lazily initialized and cached.
 // Returns nil if the HostedCluster is not available or its KubeConfig status is not set.
 // Panics on any other initialization failure (e.g., kubeconfig secret not found, invalid kubeconfig data).
-func (tc *TestContext) GetGuestClient() crclient.Client {
-	tc.guestClientOnce.Do(func() {
+func (tc *TestContext) GetHostedClusterClient() crclient.Client {
+	tc.hostedClusterClientOnce.Do(func() {
 		hc := tc.GetHostedCluster()
 		if hc == nil || hc.Status.KubeConfig == nil {
 			return
@@ -111,11 +111,11 @@ func (tc *TestContext) GetGuestClient() crclient.Client {
 
 		client, err := crclient.New(restConfig, crclient.Options{Scheme: hyperapi.Scheme})
 		if err != nil {
-			panic(fmt.Sprintf("failed to create guest cluster client: %v", err))
+			panic(fmt.Sprintf("failed to create hosted cluster client: %v", err))
 		}
-		tc.guestClient = client
+		tc.hostedClusterClient = client
 	})
-	return tc.guestClient
+	return tc.hostedClusterClient
 }
 
 var (
--- a/test/e2e/v2/tests/control_plane_workloads_test.go
+++ b/test/e2e/v2/tests/control_plane_workloads_test.go
@@ -816,78 +816,6 @@ func SecurityContextUIDTest(getTestCtx internal.TestContextGetter) {
 	})
 }
 
-// GCPCloudControllerManagerTest registers tests for GCP CCM node initialization validation.
-// These tests verify that the CCM workload is producing the expected effects on the guest cluster:
-// providerID assignment, topology labels, and taint removal.
-func GCPCloudControllerManagerTest(getTestCtx internal.TestContextGetter) {
-	Context("GCP Cloud Controller Manager", Label("GCP", "CCM"), func() {
-		BeforeEach(func() {
-			testCtx := getTestCtx()
-			hc := testCtx.GetHostedCluster()
-			if hc == nil || hc.Spec.Platform.Type != hyperv1.GCPPlatform {
-				Skip("GCP Cloud Controller Manager test is only for GCP platform")
-			}
-		})
-
-		Context("When nodes are initialized by the CCM", func() {
-			var nodes *corev1.NodeList
-
-			BeforeEach(func() {
-				testCtx := getTestCtx()
-				guestClient := testCtx.GetGuestClient()
-				Expect(guestClient).NotTo(BeNil(), "guest client is nil; HostedCluster may not have KubeConfig status set")
-
-				nodes = &corev1.NodeList{}
-				Expect(guestClient.List(context.Background(), nodes)).To(Succeed())
-				Expect(nodes.Items).NotTo(BeEmpty(), "cluster should have nodes")
-			})
-
-			It("should set providerID on all nodes", func() {
-				testCtx := getTestCtx()
-				hc := testCtx.GetHostedCluster()
-				Expect(hc.Spec.Platform.GCP).NotTo(BeNil(), "GCP platform spec must be set for GCP HostedCluster %s/%s", hc.Namespace, hc.Name)
-				gcpProject := hc.Spec.Platform.GCP.Project
-
-				for _, node := range nodes.Items {
-					Expect(node.Spec.ProviderID).NotTo(BeEmpty(),
-						"node %s should have providerID set", node.Name)
-					// GCP providerID format: gce://<project>/<zone>/<instance-name>
-					Expect(node.Spec.ProviderID).To(HavePrefix("gce://"+gcpProject+"/"),
-						"node %s providerID should reference project %s", node.Name, gcpProject)
-					parts := strings.Split(node.Spec.ProviderID, "/")
-					Expect(parts).To(HaveLen(5),
-						"node %s providerID should have format gce://<project>/<zone>/<instance-name>", node.Name)
-				}
-			})
-
-			It("should set zone and region topology labels on all nodes", func() {
-				for _, node := range nodes.Items {
-					zone, ok := node.Labels["topology.kubernetes.io/zone"]
-					Expect(ok).To(BeTrue(),
-						"node %s should have topology.kubernetes.io/zone label", node.Name)
-					Expect(zone).NotTo(BeEmpty(),
-						"node %s zone label should not be empty", node.Name)
-
-					region, ok := node.Labels["topology.kubernetes.io/region"]
-					Expect(ok).To(BeTrue(),
-						"node %s should have topology.kubernetes.io/region label", node.Name)
-					Expect(region).NotTo(BeEmpty(),
-						"node %s region label should not be empty", node.Name)
-				}
-			})
-
-			It("should remove the uninitialized taint from all nodes", func() {
-				for _, node := range nodes.Items {
-					for _, taint := range node.Spec.Taints {
-						Expect(taint.Key).NotTo(Equal("node.cloudprovider.kubernetes.io/uninitialized"),
-							"node %s should not have the cloud provider uninitialized taint", node.Name)
-					}
-				}
-			})
-		})
-	})
-}
-
 // RegisterControlPlaneWorkloadsTests registers all control plane workloads tests
 func RegisterControlPlaneWorkloadsTests(getTestCtx internal.TestContextGetter) {
 	WorkloadRegistryValidationTest(getTestCtx)
@@ -902,7 +830,6 @@ func RegisterControlPlaneWorkloadsTests(getTestCtx internal.TestContextGetter) {
 	ServiceAccountTokenMountingTest(getTestCtx)
 	PodAffinitiesAndTolerationsTest(getTestCtx)
 	SecurityContextUIDTest(getTestCtx)
-	GCPCloudControllerManagerTest(getTestCtx)
 }
 
 var _ = Describe("Control Plane Workloads", Label("control-plane-workloads"), func() {
--- a/test/e2e/v2/tests/hosted_cluster_ccm_test.go
+++ b/test/e2e/v2/tests/hosted_cluster_ccm_test.go
@@ -1,0 +1,118 @@
+//go:build e2ev2
+
+/*
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package tests
+
+import (
+	"context"
+	"strings"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
+	"github.com/openshift/hypershift/test/e2e/v2/internal"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+// GCPCloudControllerManagerTest registers tests for GCP CCM node initialization validation.
+// These tests verify that the CCM workload is producing the expected effects on the hosted cluster:
+// providerID assignment, topology labels, and taint removal.
+func GCPCloudControllerManagerTest(getTestCtx internal.TestContextGetter) {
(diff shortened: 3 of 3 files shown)


test/e2e/v2/internal/test_context.go
```diff
@@ -72,6 +77,45 @@ func (tc *TestContext) GetHostedCluster() *hyperv1.HostedCluster {
...
+		}
+
+		client, err := crclient.New(restConfig, crclient.Options{Scheme: hyperapi.Scheme})
+		if err != nil {
+			return
+		}
+		tc.guestClient = client
+	})
+	return tc.guestClient
+}
```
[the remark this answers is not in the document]
cristianoveiga: The v2 framework assumes the hosted cluster is fully operational before tests run. The v1 framework uses a different approach (`WaitForGuestClient` with active polling), but that pattern doesn't apply here.

Instead of adding retry complexity with `sync.Mutex`, we've made `GetGuestClient()` fail-loud (panic with descriptive messages) to match the `GetHostedCluster()` pattern already established in the same file. If something is misconfigured, the test fails immediately with a clear error rather than silently returning nil.

test/e2e/v2/tests/cloud_integration_test.go
```diff
@@ -0,0 +1,189 @@
...
+				Expect(nodes.Items).NotTo(BeEmpty())
+
+				for _, node := range nodes.Items {
+					for _, taint := range node.Spec.Taints {
+						Expect(taint.Key).NotTo(Equal("node.cloudprovider.kubernetes.io/uninitialized"),
+							"node %s should not have the cloud provider uninitialized taint", node.Name)
+					}
+				}
+			})
+		})
```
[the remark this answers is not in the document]
cristianoveiga: The established pattern for stateless v2 tests is direct assertions — no `Eventually()` is used in any of the existing control plane workloads or infrastructure tests. The only `Eventually()` in the v2 suite is in `backup_restore_test.go`, which is a lifecycle test that mutates state.

The CCM initializes nodes during registration, so by the time a node is listed and Ready, providerID, topology labels, and taints are already set. Adding `Eventually()` here would diverge from the pattern every other stateless v2 test follows.

test/e2e/v2/tests/cloud_integration_test.go
```diff
@@ -0,0 +1,189 @@
...
+
+		Context("When a LoadBalancer service is created", func() {
+			const (
+				testNamespace   = "default"
+				testServiceName = "ccm-lb-test"
+			)
```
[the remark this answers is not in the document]
cristianoveiga: The LoadBalancer service provisioning test was descoped from this PR and is tracked separately in a follow-up card. There's no fixed service name in the current code.

cristianoveiga:
@coderabbitai review

test/e2e/v2/internal/test_context.go
```diff
@@ -73,6 +78,45 @@ func (tc *TestContext) GetHostedCluster() *hyperv1.HostedCluster {
 	return tc.hostedCluster
 }
 
+// GetGuestClient returns a controller-runtime client for the guest cluster.
+// It reads the kubeconfig from the secret referenced by the HostedCluster status.
+// The client is lazily initialized and cached.
+// Returns nil if the guest client cannot be created (e.g., HostedCluster not ready).
```
cblecker: This docstring is inaccurate. The method panics on most error paths (secret fetch failure, missing kubeconfig key, REST config creation, client creation) — it only returns nil when `hc == nil` or `hc.Status.KubeConfig == nil`.

Suggest updating to match the actual behavior (and mirror the `GetHostedCluster()` pattern):

```go
// Returns nil if the HostedCluster is not available or its KubeConfig status is not set.
// Panics on any other initialization failure (e.g., kubeconfig secret not found, invalid kubeconfig data).
```

test/e2e/v2/internal/test_context.go
```diff
@@ -73,6 +78,45 @@ func (tc *TestContext) GetHostedCluster() *hyperv1.HostedCluster {
...
+func (tc *TestContext) GetGuestClient() crclient.Client {
+	tc.guestClientOnce.Do(func() {
+		hc := tc.GetHostedCluster()
+		if hc == nil || hc.Status.KubeConfig == nil {
+			return
+		}
+
+		var kubeconfigSecret corev1.Secret
+		err := tc.MgmtClient.Get(context.Background(), crclient.ObjectKey{
```
cblecker: `GetHostedCluster()` uses `tc.Context` for its API call (line 57), but this uses `context.Background()`. This means the kubeconfig secret fetch won't respect test timeout/cancellation.

Suggest using `tc.Context` for consistency:

```go
err := tc.MgmtClient.Get(tc.Context, crclient.ObjectKey{
```

test/e2e/v2/tests/control_plane_workloads_test.go
```diff
@@ -816,6 +816,88 @@ func SecurityContextUIDTest(getTestCtx internal.TestContextGetter) {
...
+				testCtx := getTestCtx()
+				guestClient := testCtx.GetGuestClient()
+				Expect(guestClient).NotTo(BeNil(), "guest client is required")
+
+				nodes := &corev1.NodeList{}
+				Expect(guestClient.List(context.Background(), nodes)).To(Succeed())
+				Expect(nodes.Items).NotTo(BeEmpty(), "cluster should have nodes")
+
+				hc := testCtx.GetHostedCluster()
+				gcpProject := hc.Spec.Platform.GCP.Project
```
cblecker: `GCP` is a pointer field (`*GCPPlatformSpec`). While the `BeforeEach` guard checks `Platform.Type == GCPPlatform`, a nil `GCP` field would cause a raw nil pointer panic here with no useful diagnostic. Adding a defensive check produces a clear failure message:

```go
Expect(hc.Spec.Platform.GCP).NotTo(BeNil(), "GCP platform spec must be set for GCP HostedCluster %s/%s", hc.Namespace, hc.Name)
gcpProject := hc.Spec.Platform.GCP.Project
```

test/e2e/v2/internal/workload_registry.go
cblecker, line 1: Pre-existing issue, but worth fixing in this PR since the file is being edited: the file header (lines 3-4) says `"This file is generated. Do not edit manually."` and references a script at `/tmp/generate_workloads.go` that doesn't exist in the repository. The output filename referenced (`generated_workloads.go`) also doesn't match the actual filename (`workload_registry.go`). The file has been manually edited in multiple commits including this one.

Suggest removing those two lines or replacing with something accurate like:
```go
// This file defines the control plane workload registry.
// Add new workload entries here when onboarding new components.
```

test/e2e/v2/tests/control_plane_workloads_test.go
```diff
@@ -816,6 +816,88 @@ func SecurityContextUIDTest(getTestCtx internal.TestContextGetter) {
...
+			if hc == nil || hc.Spec.Platform.Type != hyperv1.GCPPlatform {
+				Skip("Test requires a GCP HostedCluster")
+			}
+		})
+
+		Context("When nodes are initialized by the CCM", func() {
+			It("should set providerID on all nodes", func() {
+				testCtx := getTestCtx()
+				guestClient := testCtx.GetGuestClient()
+				Expect(guestClient).NotTo(BeNil(), "guest client is required")
```
cblecker: suggestion: The setup block (get testCtx, get guestClient, assert not nil, list nodes, assert not empty) is repeated identically in all 3 `It` blocks. Consider lifting the shared setup into a `BeforeEach` in this Context, which is the idiomatic Ginkgo pattern used by other tests in this file (e.g., `SecurityContextUIDTest`).

Also, if `GetGuestClient()` returns nil, the assertion message `"guest client is required"` doesn't help diagnose *why*. Something like `"guest client is nil; HostedCluster may not have KubeConfig status set"` would save debugging time.

test/e2e/v2/tests/control_plane_workloads_test.go
```diff
@@ -816,6 +816,88 @@ func SecurityContextUIDTest(getTestCtx internal.TestContextGetter) {
...
+
+				for _, node := range nodes.Items {
+					Expect(node.Spec.ProviderID).NotTo(BeEmpty(),
+						"node %s should have providerID set", node.Name)
+					// GCP providerID format: gce://<project>/<zone>/<instance-name>
+					Expect(node.Spec.ProviderID).To(HavePrefix("gce://"+gcpProject+"/"),
+						"node %s providerID should reference project %s", node.Name, gcpProject)
+					parts := strings.Split(node.Spec.ProviderID, "/")
+					Expect(parts).To(HaveLen(5),
+						"node %s providerID should have format gce://project/zone/instance", node.Name)
```
cblecker: nit: The error message says `gce://project/zone/instance` but the inline comment on line 848 uses the more precise `gce://<project>/<zone>/<instance-name>`. Consider matching the comment format in the error message for clarity during failure triage:

```go
"node %s providerID should have format gce://<project>/<zone>/<instance-name>", node.Name)
```

test/e2e/v2/internal/test_context.go
cristianoveiga: Good catch - this was left over from the initial implementation. I will update it.

test/e2e/v2/tests/control_plane_workloads_test.go
cristianoveiga: addressed

test/e2e/v2/internal/test_context.go
cristianoveiga: Done!

test/e2e/v2/internal/test_context.go
cristianoveiga: Good call - updated it.

test/e2e/v2/internal/workload_registry.go
cristianoveiga: Done! I'm guessing this was an "one-time" generator just to get the first version of this file in place/migrated?

test/e2e/v2/tests/control_plane_workloads_test.go
cristianoveiga: Done!

test/e2e/v2/tests/control_plane_workloads_test.go
cristianoveiga: Done! Moved the duplicated code into BeforeEach and improved the assertion message.

test/e2e/v2/internal/workload_registry.go
cblecker: That's my suspicion too from when @csrwng created it

cblecker:
Follow-up: a couple of minor pattern consistency items I noticed after looking at the updated diff more closely.

**1. Labels on Context node**

The new test uses `Label("GCP", "CCM")` on its Context:
```go
Context("GCP Cloud Controller Manager", Label("GCP", "CCM"), func() {
```
No other test function in this file uses Labels on Context nodes — `PodAffinitiesAndTolerationsTest` (AWS) and `SecurityContextUIDTest` (Azure) both use plain `Context(...)`. Labels in this file only appear on the top-level `Describe` and one `It` block (`"Informing"`). Consider removing the labels for consistency, or if they're intentional for filtering, that's fine too — just flagging the deviation.

**2. Skip message format**

Existing platform-skip messages follow a consistent pattern:
- `"Pod affinities and tolerations test is only for AWS platform"`
- `"Security context UID test is only for Azure platform"`

The new code uses:
```
"Test requires a GCP HostedCluster"
```

Consider matching the existing format, e.g.: `"GCP Cloud Controller Manager test is only for GCP platform"`

Both are minor — the functional changes all look good.

cblecker:
One more question: the other tests registered via `RegisterControlPlaneWorkloadsTests` all validate properties of control plane workloads (deployments/pods in the management cluster) — things like resource requests, pull policy, security contexts, etc. The GCP CCM test is different in that it validates guest cluster node state (providerID, topology labels, taints).

Was there a deliberate reason to put it in `control_plane_workloads_test.go` rather than a separate test file? It works fine here, just curious if you considered splitting it out since it's testing a different layer (guest cluster effects vs. management cluster workload properties).

cristianoveiga:
> Follow-up: a couple of minor pattern consistency items I noticed after looking at the updated diff more closely.
> 
> **1. Labels on Context node**
> 
> The new test uses `Label("GCP", "CCM")` on its Context:
> 
> ```go
> Context("GCP Cloud Controller Manager", Label("GCP", "CCM"), func() {
> ```
> 
> No other test function in this file uses Labels on Context nodes — `PodAffinitiesAndTolerationsTest` (AWS) and `SecurityContextUIDTest` (Azure) both use plain `Context(...)`. Labels in this file only appear on the top-level `Describe` and one `It` block (`"Informing"`). Consider removing the labels for consistency, or if they're intentional for filtering, that's fine too — just flagging the deviation.


The labels are intentional, yes. I used them to run the tests locally (against a pre-provisioned MC that I had).

They were useful to filter: --ginkgo.label-filter="GCP && CCM").


> 
> **2. Skip message format**
> 
> Existing platform-skip messages follow a consistent pattern:
> 
> * `"Pod affinities and tolerations test is only for AWS platform"`
> * `"Security context UID test is only for Azure platform"`
> 
> The new code uses:
> 
> ```
> "Test requires a GCP HostedCluster"
> ```
> 
> Consider matching the existing format, e.g.: `"GCP Cloud Controller Manager test is only for GCP platform"`
> 

Updated the message to match the existing format.


> Both are minor — the functional changes all look good.

cristianoveiga:
I initially had it in a separate `cloud_integration_test.go`, but I decided to include it in the existing file because I felt we didn't have similar v2 tests yet to determine the ideal file structure for guest-cluster-level validations.  So I intentionally deferred that decision until we have more tests migrated to v2.

That said, I'm happy to move this to a new file now if you have a specific preference.


> One more question: the other tests registered via `RegisterControlPlaneWorkloadsTests` all validate properties of control plane workloads (deployments/pods in the management cluster) — things like resource requests, pull policy, security contexts, etc. The GCP CCM test is different in that it validates guest cluster node state (providerID, topology labels, taints).
> 
> Was there a deliberate reason to put it in `control_plane_workloads_test.go` rather than a separate test file? It works fine here, just curious if you considered splitting it out since it's testing a different layer (guest cluster effects vs. management cluster workload properties).

cblecker:
Feedback on naming: the v2 framework is a clean slate and doesn't use "guest cluster" terminology anywhere (the only "guest" reference is the AWS env var `AWS_GUEST_INFRA_CREDENTIALS_FILE` which comes from external convention). The v1 framework has `WaitForGuestClient`/`guestClient` heavily, but the project's preferred terminology is "hosted cluster" and "control plane" — see AGENTS.md which consistently uses these terms and never says "guest cluster."

Since v2 is the chance to get this right, I'd suggest renaming:
- `GetGuestClient()` → `GetHostedClusterClient()`
- `guestClient` / `guestClientOnce` fields → `hostedClusterClient` / `hostedClusterClientOnce`
- The docstring: "guest cluster" → "hosted cluster"
- Variable names in the tests: `guestClient` → `hostedClusterClient`

Separately, per my earlier question about file organization — I'd recommend moving `GCPCloudControllerManagerTest` out of `control_plane_workloads_test.go` into a new `hosted_cluster_ccm_test.go` file. The tests in `control_plane_workloads_test.go` all validate properties of workloads running in the control plane namespace (management cluster side), but the CCM tests validate node state on the hosted cluster — a different layer entirely.

A feature-scoped file (rather than a monolithic `hosted_cluster_test.go`) sets a good convention as more hosted-cluster-side tests get added. `control_plane_workloads_test.go` is already 850 lines with 13 test functions, and the v1 framework's approach of smaller domain-specific files (karpenter, autoscaling, OLM, etc.) has scaled better than large monoliths. If other platform CCM tests or the descoped LoadBalancer tests land later, they can share this file or get their own.

The structure would follow the existing v2 pattern:
- `RegisterHostedClusterCCMTests(getTestCtx)` registration function
- `var _ = Describe("Hosted Cluster CCM", Label("hosted-cluster-ccm"), ...)` top-level block
- Platform-specific test functions nested inside with `BeforeEach` skip guards

The GCP CCM workload registry entry should stay in `workload_registry.go` — that's the right place for it. Only the behavioral test function would move.

cristianoveiga:
Both good suggestions - Implemented.

> Feedback on naming: the v2 framework is a clean slate and doesn't use "guest cluster" terminology anywhere (the only "guest" reference is the AWS env var `AWS_GUEST_INFRA_CREDENTIALS_FILE` which comes from external convention). The v1 framework has `WaitForGuestClient`/`guestClient` heavily, but the project's preferred terminology is "hosted cluster" and "control plane" — see AGENTS.md which consistently uses these terms and never says "guest cluster."
> 
> Since v2 is the chance to get this right, I'd suggest renaming:
> 
> * `GetGuestClient()` → `GetHostedClusterClient()`
> * `guestClient` / `guestClientOnce` fields → `hostedClusterClient` / `hostedClusterClientOnce`
> * The docstring: "guest cluster" → "hosted cluster"
> * Variable names in the tests: `guestClient` → `hostedClusterClient`
> 
> Separately, per my earlier question about file organization — I'd recommend moving `GCPCloudControllerManagerTest` out of `control_plane_workloads_test.go` into a new `hosted_cluster_ccm_test.go` file. The tests in `control_plane_workloads_test.go` all validate properties of workloads running in the control plane namespace (management cluster side), but the CCM tests validate node state on the hosted cluster — a different layer entirely.
> 
> A feature-scoped file (rather than a monolithic `hosted_cluster_test.go`) sets a good convention as more hosted-cluster-side tests get added. `control_plane_workloads_test.go` is already 850 lines with 13 test functions, and the v1 framework's approach of smaller domain-specific files (karpenter, autoscaling, OLM, etc.) has scaled better than large monoliths. If other platform CCM tests or the descoped LoadBalancer tests land later, they can share this file or get their own.
> 
> The structure would follow the existing v2 pattern:
> 
> * `RegisterHostedClusterCCMTests(getTestCtx)` registration function
> * `var _ = Describe("Hosted Cluster CCM", Label("hosted-cluster-ccm"), ...)` top-level block
> * Platform-specific test functions nested inside with `BeforeEach` skip guards
> 
> The GCP CCM workload registry entry should stay in `workload_registry.go` — that's the right place for it. Only the behavioral test function would move.

PR #7845, opened 2026-03-03
CNTRLPLANE-2801: fix(router): deploy router services for private ARO and remove KAS/OAPI dependencies

muraee:
## Summary
- Move `UseHCPRouter()` to a shared `router/util` package to break the import cycle, allowing `infra.go` to use it directly for determining whether router services should be deployed
- Fix an issue where router services were not deployed alongside the router deployment for private ARO clusters (Swift enabled) — the condition in `infra.go` was inconsistent with `UseHCPRouter()`
- Remove the router component's explicit dependency on OAPI and exclude it from the automatic kube-apiserver dependency, since the router doesn't require either to start

## Test plan
- [ ] Verify private ARO HCP clusters deploy router services correctly
- [ ] Verify public ARO HCP clusters do not deploy unnecessary router services
- [ ] Verify AWS private/public-and-private clusters still deploy router services
- [ ] Verify router deployment starts without waiting for KAS and OAPI
- [ ] Run unit tests: `go test ./control-plane-operator/controllers/hostedcontrolplane/v2/router/...`

## Summary by CodeRabbit

* **Refactor**
  * Centralized router decision logic into a shared util and simplified routing branch logic.

* **Bug Fixes**
  * When the HCP router is not used, both public and private router services are removed early to avoid stale services.

* **Chores**
  * Excluded the router from automatic kube-apiserver dependency checks and updated related test fixtures.


devguyio:
/lgtm

retagging since the change is only missing fixtures.

PR #7850, opened 2026-03-04
OCPBUGS-55916: fix Azure load balancer scope not updating on IngressController change

jparrill:

Adds missing Azure platform support for load balancer internal/external annotation handling in HyperShift.

On ARO HCP (Azure AKS HyperShift) clusters, when changing an IngressController's `endpointPublishingStrategy.loadBalancer.scope` from Internal to External (or vice versa), the load balancer service's IP was not updated because:

1. **`ReconcileRouterService`** handled AWS (`aws-load-balancer-internal`) and GCP (`networking.gke.io/load-balancer-type`) platform-specific annotations but had no Azure handling. The `service.beta.kubernetes.io/azure-load-balancer-internal` annotation is now set/removed based on the `internal` parameter.

2. **`ReconcileDefaultIngressController`** had no explicit Azure case, falling through to a default that created a `LoadBalancerServiceStrategyType` without setting the `loadBalancerScope`. Azure now gets the same treatment as IBM Cloud, with the scope properly set.

Fixes https://issues.redhat.com/browse/OCPBUGS-55916

- The `ReconcileRouterService` change also handles the scope change case (Internal→External) by explicitly removing the `azure-load-balancer-internal` annotation via `delete()`.
- `ReconcileDefaultIngressController` only runs on initial creation (early return if `ResourceVersion != ""`), so the Azure case ensures the default IngressController is created with the correct scope from the start.
- All pre-existing `make test` failures are unrelated (missing mock types in other packages).

## Checklist:

🤖 Generated with [Claude Code](https://claude.com/claude-code) via `/jira:solve OCPBUGS-55916`

## Release Notes

* **New Features**
  * Added Azure platform support for ingress routing with configurable internal/external load balancer options, aligning with existing cloud provider functionality.

* **Tests**
  * Added comprehensive test coverage for Azure load balancer configuration across various scenarios.

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/ingress/router.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/ingress/router.go
@@ -45,6 +45,17 @@ func ReconcileRouterService(svc *corev1.Service, internal, crossZoneLoadBalancin
 		}
 	}
 
+	if hcp.Spec.Platform.Type == hyperv1.AzurePlatform {
+		if svc.Annotations == nil {
+			svc.Annotations = map[string]string{}
+		}
+		if internal {
+			svc.Annotations["service.beta.kubernetes.io/azure-load-balancer-internal"] = "true"
+		} else {
+			delete(svc.Annotations, "service.beta.kubernetes.io/azure-load-balancer-internal")
+		}
+	}
+
 	if svc.Labels == nil {
 		svc.Labels = map[string]string{}
 	}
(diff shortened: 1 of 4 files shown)


jparrill:
/close
After a discussion with @bryan-cox, we agree that the issue is not related with this PR but more on the CI configuration side [thread](https://redhat-internal.slack.com/archives/G01QS0P2F6W/p1772624844016979)

PR #7837, opened 2026-03-02
NO-JIRA: feat(e2e): add OIDC IAM fix and NodePool validation to backup/restore 

mgencur:
## What this PR does / why we need it:

Add RunFixDrOidcIam CLI wrapper, ValidateConditions helper, and
integrate both into the backup/restore e2e test with unit tests.

## Which issue(s) this PR fixes:

Related to https://issues.redhat.com/browse/CNTRLPLANE-2031
Implements an E2E test for https://github.com/openshift/hypershift/pull/7210

## Checklist:
- [x] Subject and description added to both, commit and PR.
- [x] Relevant issues have been referenced.
- [ ] This change includes docs. 
- [x] This change includes unit tests.

---

> [!NOTE]
> **Medium Risk**
> Expands the backup/restore E2E flow to run a DR OIDC/IAM remediation command and to assert NodePool condition stability across restore, which can introduce new test flake/failure modes and requires AWS credentials. Also changes the `fix dr-oidc-iam` CLI to patch (not update) HostedCluster annotations, slightly altering how it writes cluster state.
> 
> **Overview**
> Adds a new backup/restore E2E post-restore flow that runs `hypershift fix dr-oidc-iam` (using a new `RunFixDrOidcIam` wrapper) and validates that `NodePool` conditions observed pre-backup are restored to the same state after recovery.
> 
> Introduces shared helpers to support this: `internal.ValidateConditions`, a defaulted `AWS_GUEST_INFRA_CREDENTIALS_FILE` env var, improved control-plane workload wait error reporting, and a cleanup tweak to also wait for hosted cluster namespace deletion. Minor fixes include switching `dr-oidc-iam` HostedCluster updates to `Patch`, updating backup/restore docs for ADP 1.5+ plugin configuration, and correcting a path typo in `Dockerfile.e2e`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c6f30d4470211e31f918bd5adc0ea5a7d4248df4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Added a comprehensive backup & restore E2E framework: CLI helpers, concurrent probers, Velero orchestration, control-plane workload readiness checks, exportable helper utilities, and configurable artifact output.

* **Documentation**
  * Detailed README covering backup/restore E2E objectives, prerequisites, AWS setup, execution, artifacts, and limitations.

[...]

PR #7851, opened 2026-03-04
OCPBUGS-77621: fix(nodepool): preserve ignition-reached annotation on token secret after restore

jparrill:
## Summary

- During backup/restore with OADP, the secret janitor deletes the token secret before the NodePool is restored. The NodePool controller then creates a new token secret **without** the `ignition-reached` annotation, causing `ReachedIgnitionEndpoint=False` and blocking `MachineHealthCheck` creation and `AutoRepair`.
- When the HostedCluster has the `hypershift.openshift.io/restored-from-backup` annotation (already set by the OADP plugin), the token secret is now created with `ignition-reached=True` so the condition is preserved.

## Root Cause

The restore sequence triggers a race condition:

1. Velero restores the token secret (with `ignition-reached=True`)
2. Secret janitor deletes it (NodePool doesn't exist yet)
3. Velero restores the NodePool
4. NodePool controller creates a **new** token secret without `ignition-reached`
5. Nodes are already running and won't contact the ignition endpoint again

## Fix

In `reconcileTokenSecret`, when the HostedCluster has the `restored-from-backup` annotation, set `ignition-reached=True` on the token secret. This follows the same pattern used for InPlace upgrades where nodes also don't re-contact the ignition endpoint.

Ref: https://issues.redhat.com/browse/OCPBUGS-77621

## Test plan

- [ ] New unit test `TestTokenReconcile` case validates that `ignition-reached` annotation is set when HostedCluster has `restored-from-backup` annotation
- [ ] Existing test case validates that `ignition-reached` is NOT set without the restore annotation (no regression)
- [ ] E2E: Run backup/restore tests from https://github.com/openshift/hypershift/pull/7837 using `make test-backup-restore` and verify NodePool conditions after restore

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserve ignition-reached state when a hosted cluster is restored from backup so node pools and machine health checks continue to be created and operate correctly.
* **Tests**
  * Added tests verifying the ignition-reached state is preserved on restore, ensuring expected reconciliation behavior.

### DIFF
--- a/hypershift-operator/controllers/nodepool/token.go
+++ b/hypershift-operator/controllers/nodepool/token.go
@@ -313,6 +313,16 @@ func (t *Token) reconcileTokenSecret(tokenSecret *corev1.Secret) error {
 	// active token should never be marked as expired.
 	delete(tokenSecret.Annotations, hyperv1.IgnitionServerTokenExpirationTimestampAnnotation)
 
+	// During a backup/restore the token secret may be deleted by the secret janitor
+	// before the NodePool is restored, causing the NodePool controller to create a new
+	// token secret without the ignition-reached annotation. Since the nodes are already
+	// running and won't contact the ignition endpoint again, the annotation must be
+	// carried over so that ReachedIgnitionEndpoint remains True and MachineHealthChecks
+	// continue to be created.
+	if _, restored := t.hostedCluster.Annotations[hyperv1.HostedClusterRestoredFromBackupAnnotation]; restored {
+		tokenSecret.Annotations[TokenSecretIgnitionReachedAnnotation] = "True"
+	}
+
 	if tokenSecret.Data == nil {
 		// 2. - Reconcile towards expected state of the world.
 		compressedConfig, err := t.CompressedAndEncoded()
--- a/hypershift-operator/controllers/nodepool/token_test.go
+++ b/hypershift-operator/controllers/nodepool/token_test.go
@@ -627,6 +627,68 @@ func TestTokenReconcile(t *testing.T) {
 				DecompressAndDecodeConfig: true,
 			},
 		},
+		{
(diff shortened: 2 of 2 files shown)


mgencur:
Nice. Will verify this shortly.

mgencur:
/verified by @mgencur 

Verified with a modified version of https://github.com/openshift/hypershift/pull/7837 where I check all conditions and don't delete hyperv1.NodePoolReachedIgnitionEndpoint and hyperv1.NodePoolAutorepairEnabledConditionType from the expected conditions map.

The condition after restore is:
```
    - lastTransitionTime: "2026-03-04T16:54:15Z"
      observedGeneration: 1
      reason: AsExpected
      status: "True"
      type: ReachedIgnitionEndpoint
```

hypershift-operator/controllers/nodepool/token.go
```diff
@@ -313,6 +313,17 @@ func (t *Token) reconcileTokenSecret(tokenSecret *corev1.Secret) error {
 	// active token should never be marked as expired.
 	delete(tokenSecret.Annotations, hyperv1.IgnitionServerTokenExpirationTimestampAnnotation)
 
+	// During a backup/restore the token secret may be deleted by the secret janitor
+	// before the NodePool is restored, causing the NodePool controller to create a new
+	// token secret without the ignition-reached annotation. Since the nodes are already
+	// running and won't contact the ignition endpoint again, the annotation must be
+	// carried over so that ReachedIgnitionEndpoint remains True and MachineHealthChecks
+	// continue to be created.
+	// See: https://issues.redhat.com/browse/OCPBUGS-77621
```
bryan-cox: nit: couldn't this be dropped since this is the same bug you are fixing the problem?

jparrill:
/verified by @mgencur

Removed the comment mentioned by @bryan-cox

PR #7853, opened 2026-03-04
AUTOSCALE-571: fix(karpenter): don't associate public IP addresses for the default OpenshiftEC2NodeClass

joshbranham:
Previously, Karpenter would reconcile the default named OpenshiftEC2NodeClass to use public IPs if the annotation was set. This annotation was used for testing purposes, and was likely added as the initial e2e test suite had public subnets only.

The default behavior for this OpenshiftEC2NodeClass should be that it works in private subnets. However, if you use a public subnet, you should not expect to get a public IP.

If a user wants to use public subnets with a public IP for nodes, they should make their own OpenshiftEC2NodeClass and set `spec.AssociatePublicIPAddress: true` there.

Fixes [AUTOSCALE-571](https://issues.redhat.com/browse/AUTOSCALE-571)

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed automatic public IP address assignment for OpenShift Karpenter EC2 node classes based on HCP annotations. The `AssociatePublicIPAddress` setting now defaults to unset during reconciliation, changing how public IP addresses are managed for EC2 nodes.

### DIFF
--- a/karpenter-operator/controllers/karpenter/karpenter_controller.go
+++ b/karpenter-operator/controllers/karpenter/karpenter_controller.go
@@ -342,9 +342,6 @@ func (r *Reconciler) reconcileOpenshiftEC2NodeClassDefault(ctx context.Context, 
 			},
 		}
 
-		if hcp.Annotations[hyperv1.AWSMachinePublicIPs] == "true" {
-			ec2NodeClass.Spec.IPAddressAssociation = hyperkarpenterv1.IPAddressAssociationPublic
-		}
 		return nil
 	})
 	if err != nil {


karpenter-operator/controllers/karpenter/karpenter_controller.go
```diff
@@ -343,9 +343,6 @@ func (r *Reconciler) reconcileOpenshiftEC2NodeClassDefault(ctx context.Context,
 			},
 		}
 
-		if hcp.Annotations[hyperv1.AWSMachinePublicIPs] == "true" {
-			ec2NodeClass.Spec.AssociatePublicIPAddress = ptr.To(true)
-		}
```
jkyros, line 347: Hmmm, I think we need this as long as:
-  the `AWSMachinePublicIPs` annotation is a thing and/or
-  there is the possibility that the ec2NodeClass contains a selector that grabs a public subnet (e.g. the `o.PublicOnly` option that I'm not sure if anyone uses in Hypershift)
 
Don't we?  

Like say there is a public subnet somewhere tagged with `karpenter.sh/discovery`, nothing prevents Karpenter from launching a node in that IPless public subnet that got selected, it will just quietly fail to join. 

Taking this out right now is just "make sure any node launched on a public subnet is broken because it won't get an IP" , right? 😛

enxebre: There's no real use case for a node on having a public IP. The annotation only exists because that's what we exercise in our CI. We do it that way in CI to save $$ in infra creation (no need for nat gateways). 

@joshbranham without this, our e2e won't pass unless they rely on another class with the public IP API knob enable

joshbranham: ROSA does not support instances in public subnets.

I already changed the e2e test suite in the decoupled Version PR to not use public subnets, so they already pass without this set 👍

joshbranham: https://github.com/openshift/hypershift/blob/8f670794c7055903c60a39911d436d27330d7418/test/e2e/karpenter_test.go#L55

joshbranham: @jkyros are we moving to exclude public subnets in the selection for the default nodeclass, and therefore, should we merge this? ROSA HCP won't support public, so would be good to come to a consensus here, let me know if you want to dig in deeper on Slack.

jkyros: Yes, as soon as https://github.com/openshift/hypershift/pull/7933 gets in we should do this

jkyros:
Now that https://github.com/openshift/hypershift/pull/7933 is in, this should be able to come out since we're explicitly selecting private networks by default, a public one shouldn't be able to be in the list and need this.    
/lgtm

enxebre:
This aligns with https://github.com/openshift/hypershift/blob/4585fa11d8263470dd3e51d5ccfe3a27ac3e3a4e/test/e2e/karpenter_test.go#L54
We'll want to revaluate this to not degrade infra cost in presubmits.

/approve
/lgtm

enxebre:
/overreide "Red Hat Konflux / hypershift-operator-main-enterprise-contract / hypershift-operator-main"
/pipeline required

joshbranham:
Looks like I need https://github.com/openshift/hypershift/commit/5830cf10787e0171e814a3305ad5f63f987d8d3c but don't want to rebase and retest if I don't have to. Can you try to override again @enxebre 

/test e2e-aws-techpreview

cristianoveiga:
> /test e2e-aws-techpreview

hey @joshbranham , I believe you will need the fix on https://github.com/openshift/hypershift/pull/8037.

The e2e-aws-techpreview will fail without that.

enxebre:
/approve
/lgtm
/test e2e-aws-techpreview
/verified by e2e @joshbranham 
/hold
on the test passing

PR #7858, opened 2026-03-05
[release-4.19] OCPBUGS-77367,OCPBUGS-77844: Fix ignition-server pod restarts

jparrill:
## Summary
- Manual backport of #7842 to release-4.19
- Implements a mirror availability cache with differential TTL to prevent `MIRRORED_RELEASE_IMAGE` env var flapping that caused massive deployment regeneration
- Removes `MIRRORED_RELEASE_IMAGE` from the ignition-server deployment (dead code not consumed at runtime)
- Removes `getRegistryOverrides()` which performed non-deterministic live HTTP registry checks during every CPO reconciliation, replacing it with static registry-level overrides from the HostedCluster spec

## Fixes
- [OCPBUGS-77367](https://issues.redhat.com/browse/OCPBUGS-77367)
- [OCPBUGS-77844](https://issues.redhat.com/browse/OCPBUGS-77844)

## Backport adaptations
The following changes were made beyond the pure cherry-picks to adapt to release-4.19:

1. **`support/util/imagemetadata_test.go`**: Removed unused `"fmt"` and `"os"` imports that were residual from the original commit but not referenced in the test code.
2. **`control-plane-operator/controllers/hostedcontrolplane/v2/ignitionserver/deployment_test.go`**: Replaced `t.Context()` with `context.Background()` (adding the `"context"` import), since `t.Context()` requires Go 1.24 and release-4.19 uses Go 1.23.

All merge conflicts were resolved by accepting the incoming commit version (removal of `getRegistryOverrides()` and the `controller-runtime/pkg/client` import).

## Test plan
- [ ] Verify ignition-server deployment no longer sets `MIRRORED_RELEASE_IMAGE` env var
- [ ] Verify mirror availability cache works with differential TTL (5min available, 1min unavailable)
- [ ] Verify deployment remains stable without flapping in disconnected environments
- [ ] Run unit tests: `go test ./support/util/... ./control-plane-operator/controllers/hostedcontrolplane/v2/ignitionserver/...`

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
@@ -1329,7 +1329,6 @@ func (r *HostedControlPlaneReconciler) reconcile(ctx context.Context, hostedCont
 				r.ManagementClusterCapabilities.Has(capabilities.CapabilitySecurityContextConstraint),
 				config.OwnerRefFrom(hostedControlPlane),
 				openShiftTrustedCABundleConfigMapForCPOExists,
-				r.ReleaseProvider.GetMirroredReleaseImage(),
 				util.LabelHCPRoutes(hostedControlPlane),
 			); err != nil {
 				return fmt.Errorf("failed to reconcile ignition server: %w", err)
--- a/control-plane-operator/controllers/hostedcontrolplane/ignitionserver/ignitionserver.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/ignitionserver/ignitionserver.go
@@ -7,7 +7,6 @@ import (
 	"net"
 
 	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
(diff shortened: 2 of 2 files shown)


support/util/imagemetadata.go
```diff
@@ -456,12 +525,33 @@ func SeekOverride(ctx context.Context, openshiftImageRegistryOverrides map[strin
...
+						return ref
+					} else {
+						log.V(2).Info("Mirror unavailable (cached), skipping", "mirror", mirrorURL)
+						continue
+					}
+				}
+
+				// Cache miss - verify mirror availability with 15s timeout
+				verifyCtx, cancel := context.WithTimeout(ctx, 15*time.Second)
+				defer cancel()
```
csrwng, line 543: nit: it might make sense to make the change from https://github.com/openshift/hypershift/pull/7846 so we don't keep propagating the bug.

jparrill: Make sense to me, I will propagate the fix ASAP, but we need this to be in 4.18 also ASAP. So the plan I have in my mind is to push this until 4.18 and then backport the #7846. This is because the customer needs these changes urgently.

PR #7859, opened 2026-03-05
NO-JIRA: api: fix AutoNode types to comply with OpenShift API conventions

enxebre:
- Add +union, +unionDiscriminator, +unionMember markers to ProvisionerConfig and KarpenterConfig discriminated unions
- Add CEL validation rules enforcing union member presence matches the discriminator value
- Add field-level +kubebuilder:validation:Enum=AWS on KarpenterConfig.Platform to restrict to supported platforms
- Change AutoNode.Provisioner from pointer to value (required field)
- Improve type-level and field-level Godoc on AutoNode, ProvisionerConfig, KarpenterConfig, and KarpenterAWSConfig

Fixes 

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Karpenter-based automatic node provisioning added with AWS-targeted options.

* **Documentation**
  * Expanded docs and API descriptions for autoNode and provisioner usage; clarified platform semantics and examples.

* **Validation**
  * Stronger CRD/schema validations and cross-field constraints; IAM role ARN pattern and extended max length supported.

* **Refactor / API**
  * AutoNode provisioner initialization/signature updated—may require minor client/test adjustments.


api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1355,46 +1357,302 @@ type Release struct {
...
 }
 
-// We expose here internal configuration knobs that won't be exposed to the service.
+// AutoNode specifies the configuration for automatic node provisioning and lifecycle management.
 type AutoNode struct {
-	// provisionerConfig is the implementation used for Node auto provisioning.
+	// provisionerConfig specifies the provisioner used for automatic node management.
 	// +required
-	Provisioner *ProvisionerConfig `json:"provisionerConfig"`
+	Provisioner ProvisionerConfig `json:"provisionerConfig"`
```
JoelSpeed: How long have these types existed? Who is this likely to break?

JoelSpeed: To meet current conventions, all fields should have `omitempty` or `omitzero` for structs

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1355,46 +1357,302 @@ type Release struct {
...
+	// 			"Effect": "Allow",
+	// 			"Resource": "*",
+	// 			"Action": "iam:ListInstanceProfiles"
+	// 		}
+	// 	]
+	// }
+	//
 	// +required
-	// +kubebuilder:validation:MaxLength=255
+	// +kubebuilder:validation:Pattern:=`^arn:(aws|aws-cn|aws-us-gov):iam::[0-9]{12}:role\/.*$`
```
JoelSpeed: We prefer to use CEL to implement patterns like this as you can then customise the error message to be helpful, with this it'll just dump the regex which is less user friendly generally

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1355,46 +1357,302 @@ type Release struct {
...
+	// +kubebuilder:validation:MaxLength=2048
```
JoelSpeed, line 1646: How did you come up with the extended maximum length here?

enxebre: I was just being conservative

api/hypershift/v1beta1/hostedcluster_types.go
enxebre: this not exposed yet.

api/hypershift/v1beta1/hostedcluster_types.go
JoelSpeed: You can probably keep this tighter and relax the validation if you find issues later. I would be looking to see if there is a length limit imposed by AWS on the variable part (role name?) and then the rest we can calculate the length of to come up with a suitable max length

enxebre: It seems 2048 is actually accurate https://docs.aws.amazon.com/IAM/latest/APIReference/API_Role.html#API_Role_Contents.

api/hypershift/v1beta1/hostedcluster_types.go
muraee: struct is required with 1 required field,  `omitzero` is never used. Why do we need to add it? It's a bit confusing

JoelSpeed: It is used. Imagine a client builds a structured object and doesn't populate anything within here. With `omitzero` the key `provisionerConfig` is not marshalled, and they will get `provisioner is required` within the error

If you didn't have omitzero, they would marshal `provisioner: {}` and potentially get other errors, or just mess

We updated the API conventions last year to try and bring the behaviour of structured clients and unstructured clients closer together and this is an important part of that alignment

muraee: I see

PR #7864, opened 2026-03-05
OCPBUGS-77514: E2E filter rolling upgrade verification to current MachineSet

clebs:
The rolling upgrade E2E test fails when bumping CAPI to 1.11 because old machines are still deleting when `MachineDeploymentComplete()` signals that the deployment is done. Then the test proceeds to check that all machines are on the new template, which is not true for the machine still in deletion.

Until the cause of this flake is found upstream, this PR adds filtering to the e2e test so that it verifies the current `MachineSet` only.

Fixes [OCPBUGS-77514](https://issues.redhat.com/browse/OCPBUGS-77514)

There is a detailed RCA in [OCPBUGS-77514](https://issues.redhat.com/browse/OCPBUGS-77514). Please have a look at it to get the full picture.

## Checklist:

## Summary by CodeRabbit

* **Tests**
  * Tightened rolling-upgrade validations to target the current machine set during node pool upgrades
  * Added utilities to identify and constrain checks to the active machine set for both AWS and Azure rotations
  * Added non-empty assertions for cloud-provider machine lists and unified namespace handling to improve test reliability and clarity

### DIFF
--- a/test/e2e/nodepool_rolling_upgrade_test.go
+++ b/test/e2e/nodepool_rolling_upgrade_test.go
@@ -5,6 +5,7 @@ package e2e
 import (
 	"context"
 	"fmt"
+	"strconv"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


test/e2e/nodepool_rolling_upgrade_test.go
```diff
@@ -131,26 +132,68 @@ func (k *RollingUpgradeTest) Run(t *testing.T, nodePool hyperv1.NodePool, nodes
 		e2eutil.WithTimeout(k.getRollingUpgradeTimeout()),
 	)
 
+	controlPlaneNamespace := manifests.HostedControlPlaneNamespace(k.hostedCluster.Namespace, k.hostedCluster.Name)
+
+	// Find the current (latest revision) MachineSet for the MachineDeployment.
+	// After a rolling upgrade completes, the old MachineSet may still have machines
+	// that are being scaled down. We must only verify machines belonging to the
+	// current MachineSet to avoid false failures from stale machines.
+	currentMachineSetName := currentMachineSetForDeployment(t, k.ctx, k.mgmtClient, controlPlaneNamespace, nodePool.Name)
```
clebs, line 144: Another solution I considered to this issue is to modify `MachineDeploymentComplete()` to only report completion when all old machines are done deleting (in addition to the current condition).

I decided against it because discarded machines taking longer to delete should not affect this condition and the test accounting for it seems right.

test/e2e/nodepool_rolling_upgrade_test.go
```diff
@@ -131,26 +132,68 @@ func (k *RollingUpgradeTest) Run(t *testing.T, nodePool hyperv1.NodePool, nodes
...
+	machineLabels := crclient.MatchingLabels{
+		capiv1.MachineDeploymentNameLabel: nodePool.Name,
+		capiv1.MachineSetNameLabel:        currentMachineSetName,
+	}
```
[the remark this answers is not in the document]
clebs: Proposed fix applied.

test/e2e/nodepool_rolling_upgrade_test.go
```diff
@@ -131,26 +133,70 @@ func (k *RollingUpgradeTest) Run(t *testing.T, nodePool hyperv1.NodePool, nodes
...
+	g.Expect(err).ToNot(HaveOccurred(), "failed to list MachineSets")
+	g.Expect(machineSets.Items).ToNot(BeEmpty(), "expected at least one MachineSet for MachineDeployment %s", machineDeploymentName)
+
+	var current string
+	var maxRevision int64 = -1
+	for i := range machineSets.Items {
+		rev, err := strconv.ParseInt(machineSets.Items[i].Annotations[capiv1.RevisionAnnotation], 10, 64)
+		if err != nil {
+			continue
+		}
```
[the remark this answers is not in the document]
clebs: This is intentional to get the latest revision.

test/e2e/nodepool_rolling_upgrade_test.go
```diff
@@ -131,26 +133,70 @@ func (k *RollingUpgradeTest) Run(t *testing.T, nodePool hyperv1.NodePool, nodes
 		e2eutil.WithTimeout(k.getRollingUpgradeTimeout()),
 	)
 
+	controlPlaneNamespace := manifests.HostedControlPlaneNamespace(k.hostedCluster.Namespace, k.hostedCluster.Name)
+
+	// Find the current (latest revision) MachineSet for the MachineDeployment.
+	// Currently in CAPI v1.11, after a rolling upgrade completes, the old MachineSet may still have machines
+	// that are being scaled down. As a temporary workaround We only verify machines belonging to the
+	// current MachineSet to avoid false failures from stale machines.
+	// TODO(bclement): Revert this change once the root cause is found and fixed:
```
bryan-cox, line 142: Why are we punting this vs trying to solve the issue here?

I've been trying to reduce our TODOs in our codebase lately so I'm a little sensitive to adding new TODOs 😅

clebs: I have not been able to find the cause for this change in behavior in CAPI 1.11 vs 1.10 so far and estimate it will take quite some time to solve.

I would rather be able to update our CAPI and k8s versions first (+ removing the 2 hacks we have in place) and afterwards investigate the issue upstream further.

But of course it is a matter of just priority.

bryan-cox (approved):
/lgtm

clebs:
This workaround can be removed once we can solve https://github.com/kubernetes-sigs/cluster-api/issues/11401

PR #7896, opened 2026-03-09
GCP-410: feat(gcp): add HCCO credential propagation for GCP image registry

cblecker:
## Summary

Enables the GCP image registry operator to authenticate to GCS in HyperShift-hosted
clusters by propagating a WIF credential secret into the guest cluster's
`openshift-image-registry` namespace.

- Add `support/gcputil` package exposing `BuildWorkloadIdentityCredentials` and the
  associated `ExternalAccountCredential` types so both the HO and HCCO can share the
  credential builder without an import cycle
- Update the HO's internal GCP platform to call `gcputil.BuildWorkloadIdentityCredentials`
  (identical behaviour, no functional change to the HO side)
- Add `control-plane-operator/hostedclusterconfigoperator/controllers/resources/gcp`
  package with `SetupOperandCredentials`, following the Azure HCCO pattern; currently
  manages a single credential — `openshift-image-registry/installer-cloud-credentials`
  populated with a WIF JSON blob for the image registry GSA
- Wire GCP into `reconcileCloudCredentialSecrets` in the HCCO resources controller
- Guard secret upsert on namespace existence to handle the bootstrap race (namespace may
  not yet exist on first reconcile); mirrors the same pattern used for AWS
- Skip credential creation when the `ImageRegistry` capability is explicitly disabled

The `service_account.json` data key matches the CCO-provisioned format expected by
`cluster-image-registry-operator`. Only image registry is wired through HCCO for GCP;
other GCP operators receive their credentials directly from HO-provisioned secrets.

## Test plan

- [x] `go test ./support/gcputil/...` — unit tests covering happy path and all validation
  error branches (`ProjectNumber`, `PoolID`, `ProviderID`, `ServiceAccountEmail` empty)
- [x] `go test ./control-plane-operator/hostedclusterconfigoperator/controllers/resources/gcp/...`
  — unit tests covering: capability enabled (secret created with correct WIF JSON structure),
  capability disabled (secret skipped), target namespace absent (skipped without error)
- [x] `make test` — full unit test suite
- [ ] E2E: create a GCP HCP cluster with `imageRegistry` GSA set, verify
  `openshift-image-registry/installer-cloud-credentials` exists in the guest cluster with
  a valid `service_account.json` key, and confirm the image registry operator reaches
  `Available` (tracked in [GCP-413](https://redhat.atlassian.net/browse/GCP-413))


control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -1831,6 +1832,10 @@ func (r *reconciler) reconcileCloudCredentialSecrets(ctx context.Context, hcp *h
 				errs = append(errs, fmt.Errorf("failed to reconcile powervs image registry cloud credentials secret %w", err))
 			}
 		}
+	case hyperv1.GCPPlatform:
+		if hcp.Spec.Platform.GCP != nil {
+			errs = gcpresources.SetupOperandCredentials(ctx, r.client, r.CreateOrUpdateProvider, hcp)
```
ckandag: NP, not blocking.
Consider using errs = append(errs, gcpresources.SetupOperandCredentials(...)...) for consistency with the other platform cases in this function. The current errs = assignment is correct as-is since nothing else populates errs before the switch, but if anything is ever added before it, those errors would be silently dropped.

ckandag:
LGTM 
Please consider testing this in a live environment if you havent already.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
cblecker: Good catch — updated to use `append` for consistency with other platform cases.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/gcp/gcp_test.go
```diff
@@ -0,0 +1,188 @@
...
+				g.Expect(json.Unmarshal(sec.Data["service_account.json"], &cred)).To(Succeed())
+				g.Expect(cred.Type).To(Equal("external_account"))
+				g.Expect(cred.Audience).To(ContainSubstring(testProjectNumber))
+				g.Expect(cred.Audience).To(ContainSubstring(testPoolID))
+				g.Expect(cred.Audience).To(ContainSubstring(testProviderID))
+				g.Expect(cred.ServiceAccountImpersonationURL).To(ContainSubstring(testImageRegistryGSA))
+				g.Expect(cred.CredentialSource.File).To(Equal("/var/run/secrets/openshift/serviceaccount/token"))
+			} else {
+				g.Expect(err).To(HaveOccurred())
+			}
```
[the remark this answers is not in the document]
cblecker: Good catch — tightened the assertion to check for NotFound specifically.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/gcp/gcp.go
```diff
@@ -0,0 +1,134 @@
...
+) []error {
+	var errs []error
+
+	for _, cfg := range configs {
+		if cfg.capabilityChecker != nil && !cfg.capabilityChecker(hcp.Spec.Capabilities) {
+			continue
+		}
+
+		secret := cfg.manifestFunc()
+
```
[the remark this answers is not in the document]
cblecker: Thanks for the suggestion. All three platforms (AWS, Azure, GCP) currently use the same skip-only pattern — none delete secrets when capabilities are disabled. Adding deletion logic only to GCP would create cross-platform inconsistency. If this behavior should change, it should be addressed holistically across all platforms in a separate PR.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/gcp/gcp.go
```diff
@@ -0,0 +1,134 @@
...
+		SubjectTokenType:               "urn:ietf:params:oauth:token-type:jwt",
+		TokenURL:                       "https://sts.googleapis.com/v1/token",
+		ServiceAccountImpersonationURL: fmt.Sprintf("https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/%s:generateAccessToken", serviceAccountEmail),
+		CredentialSource: gcpCredentialSource{
+			File: "/var/run/secrets/openshift/serviceaccount/token",
+			Format: gcpCredentialSourceFormat{
+				Type: "text",
+			},
+		},
+	}
```
cristianoveiga: This credential JSON references /var/run/secrets/openshift/serviceaccount/token as the token source. On the management cluster side, the token-minter sidecar mints a SA token (with the audience matching the WIF provider's AllowedAudiences) and writes it to this path for the cluster-image-registry-operator pod.

However, the installer-cloud-credentials secret created by this code is consumed by the image-registry pods running on guest worker nodes, which don't have a token-minter sidecar.

For AWS, the aws-pod-identity-webhook (a MutatingWebhookConfiguration installed on the guest cluster) injects a projected SA token at this path. I don't see an equivalent mechanism for GCP.

How will the registry pods on the guest cluster get a valid token at this path? Is this covered by a follow-up card?

control-plane-operator/hostedclusterconfigoperator/controllers/resources/gcp/gcp.go
```diff
@@ -0,0 +1,134 @@
...
+	"github.com/openshift/hypershift/control-plane-operator/hostedclusterconfigoperator/controllers/resources/manifests"
+	"github.com/openshift/hypershift/support/capabilities"
+	"github.com/openshift/hypershift/support/upsert"
+
+	corev1 "k8s.io/api/core/v1"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// WIF credential JSON structs — same format as HO's gcp.go, duplicated because that code is in an internal package.
```
cristianoveiga: In addition to the structs, we are also duplicating the `buildGCPWorkloadIdentityCredentials` function. 

Azure solved the same problem by extracting shared credential logic into support/azureutil/. 

Would it make sense to create a support/gcputil/ package with these types and the builder function? This would prevent drift as more GCP credential secrets are added (CSI, ingress, etc.).

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -1831,6 +1832,10 @@ func (r *reconciler) reconcileCloudCredentialSecrets(ctx context.Context, hcp *h
 				errs = append(errs, fmt.Errorf("failed to reconcile powervs image registry cloud credentials secret %w", err))
 			}
 		}
+	case hyperv1.GCPPlatform:
+		if hcp.Spec.Platform.GCP != nil {
+			errs = append(errs, gcpresources.SetupOperandCredentials(ctx, r.client, r.CreateOrUpdateProvider, hcp)...)
+		}
 	}
```
cristianoveiga, line 2221: The AWS implementation (line 1667-1672 above) checks whether the target namespace exists before creating the secret, logging a warning and skipping gracefully if not found. This avoids unnecessary error noise and requeues during early cluster bootstrap.

The GCP case doesn't do this. If ImageRegistry is enabled but openshift-image-registry hasn't been created yet, the upsert will fail and trigger repeated requeues. Consider adding a similar namespace guard, either here or inside SetupOperandCredentials.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/gcp/gcp.go
```diff
@@ -0,0 +1,134 @@
...
+	ctx context.Context,
+	c client.Client,
+	upsertProvider upsert.CreateOrUpdateProvider,
+	hcp *hyperv1.HostedControlPlane,
+) []error {
+	configs := []gcpCredentialConfig{
+		{
+			manifestFunc:        manifests.GCPImageRegistryCloudCredsSecret,
+			serviceAccountEmail: hcp.Spec.Platform.GCP.WorkloadIdentity.ServiceAccountsEmails.ImageRegistry,
+			capabilityChecker:   capabilities.IsImageRegistryCapabilityEnabled,
```
cristianoveiga, line 40: Have you been able to validate this on a dev cluster with ImageRegistry enabled? Since GCP E2E tests aren't available yet (https://issues.redhat.com/browse/GCP-401), unit tests are the only automated coverage for this code path.                                                                                                                                                                                         
   
I'm wondering what's the plan here - enable the ImageRegistry capability in integration (it's currently disabled) to get some coverage sooner, or wait for GCP E2E tests and remove it from the disabled capabilities list in the release repo at that point?

control-plane-operator/hostedclusterconfigoperator/controllers/resources/gcp/gcp.go
cblecker: The `cluster-image-registry-operator` (CIRO) runs on the management cluster as a CPOv2 component with `InjectTokenMinterContainer(CloudAndAPIServerToken)` — the `cloud-token-minter` sidecar mints a token and writes it to `/var/run/secrets/openshift/serviceaccount/token` for the operator's own GCP API calls.

For the image registry pods on the guest cluster: the upstream CIRO creates the registry deployment and mounts both the `installer-cloud-credentials` secret and a projected SA token volume. The guest cluster's kube-apiserver OIDC issuer is configured to match the GCP WIF provider, so standard Kubernetes projected SA tokens work — this is [standard GCP WIF for non-GKE Kubernetes clusters](https://docs.cloud.google.com/iam/docs/workload-identity-federation-with-kubernetes#kubernetes). No equivalent of `aws-pod-identity-webhook` is needed because projected SA tokens are a native Kubernetes feature that GCP WIF consumes directly.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/gcp/gcp.go
cblecker: Not yet validated on a live cluster, but I plan to before merging. GCP E2E tests aren't available yet (GCP-401), so unit tests are the current automated coverage.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/gcp/gcp.go
ckandag: Control plane pods (CAPG, CCM, etc.) run on the management cluster (GKE) and receive a management cluster SA token by default. However, WIF pool we setup is configured to trust the guest cluster's OIDC issuer — not the management cluster's GKE issuer. This is why token-minter is needed as a sidecar: it calls the guest cluster's TokenRequest API to mint a guest-OIDC-signed JWT and writes it to /var/run/secrets/openshift/serviceaccount/token, which the WIF credential JSON then exchanges at GCP STS.

For pods like the Image Registry Operator that run directly in the guest cluster, their service account tokens are natively issued by the guest cluster's kube-apiserver and already trusted by the WIF pool — so agree that no token-minter would be required.

cblecker:
@coderabbitai resume

PR #7346, opened 2025-12-05
CNTRLPLANE-2049: Update kube-api-linter dependency

bryan-cox:
## What this PR does / why we need it:

Updates the kube-api-linter dependency to the latest version and adjusts the golangci-lint configuration to work with the new linter behavior.

Key changes:
- Update sigs.k8s.io/kube-api-linter dependency
- Vendor the updated dependency
- Disable all default kube-api-linter analyzers and explicitly enable only specific ones (conditions, maxlength, nobools, nomaps, statussubresource). This prevents unexpected failures when upgrading kube-api-linter in the future.
- Update exclusion text patterns to match the new error message format which now includes struct name prefixes (e.g., "Volume.Encrypted" instead of "Encrypted")
- Add exclusion for statussubresource error on CertificateSigningRequestApproval type

## Which issue(s) this PR fixes:

[CNTRLPLANE-2049](https://issues.redhat.com//browse/CNTRLPLANE-2049)

Supersedes #7178

## Special notes for your reviewer:

The new version of kube-api-linter changed its error message format to include struct names in field references. All exclusion patterns have been updated to match.

## Checklist:
- [x] Subject and description added to both, commit and PR.
- [x] Relevant issues have been referenced.
- [ ] This change includes docs.
- [ ] This change includes unit tests.

PR #7333, opened 2025-12-03
CNTRLPLANE-2049: chore(lint): remove unused exclusion rules

cblecker:
## Summary
Removed unused golangci-lint exclusion rules from both root and API configuration files that were not matching any issues, as identified by running `make lint` with `warn-unused: true` enabled.

## Changes
- **`.golangci.yml`**: Removed 2 unused ST1019 exclusion rules
- **`api/.golangci.yml`**: Removed 1 unused ForwardingRuleName exclusion rule
- **Preserved**: Path exclusions for `third_party$`, `builtin$`, and `examples$`

## Motivation
- Clean up linting configuration by removing rules that don't serve a purpose
- Reduce noise in the configuration files
- Make the linting setup more maintainable

## Testing
- Ran `make lint` to identify unused rules (showed "Skipped 0 issues" warnings)
- Removed the unused rules
- Re-ran `make lint` to verify:
  - 0 linting issues
  - No unused rule warnings (except for path patterns)
  - No new errors introduced

## Additional Notes
This is a pure cleanup change with no functional impact on linting behavior. Path exclusions were intentionally preserved as defensive patterns.

PR #7903, opened 2026-03-10
CNTRLPLANE-2049: Review and address kube-api-linter exclusions

Nirshal:

This PR reviews and addresses all kube-api-linter exclusions in api/.golangci.yml
as part of the alignment with openshift/api linter configuration.

### Plugin bump and config alignment
- Bump kube-api-linter plugin to 20260114 to match openshift/api
- Enable new linters: minlength, defaults, preferredmarkers, noreferences
- Align all linter settings with openshift/api configuration

### GA API types (exclusions)
GA API types cannot be modified without breaking changes. All violations
from 15 linter categories are excluded with per-field granular rules:
arrayofstruct, conditions, defaultorrequired, defaults, integers,
maxlength, minlength, nobools, nodurations, noreferences, optionalfields,
preferredmarkers, requiredfields, ssatags, statussubresource.

### Non-GA GCP types (code fixes)
GCP types (gcp.go, gcpprivateserviceconnect_types.go) are still evolving
and violations are fixed directly in the code:
- Replace +kubebuilder:validation:Pattern with XValidation (CEL) for
  better error messages
- Use +default instead of +kubebuilder:default (preferred short form)
- Add omitempty/omitzero to required fields with valid zero values
- Add MinLength, MinItems validation markers where applicable
- Add +listType=set to ConsumerAcceptList

9 optionalfields "pointer" violations on GCP types are excluded pending
further review on whether the pointer usage is intentional or removable.

3 minlength violations are excluded where MinProperties is not applicable
(pointer or non-pointer structs with omitempty).

Ref https://issues.redhat.com/browse/CNTRLPLANE-2049

- Commit structure: plugin bump → config alignment → GA exclusions →
  non-GA code fixes → non-GA exclusions → make update
- GA types get per-field exclusions; non-GA GCP types get code fixes
- 9 GCP optionalfields pointer exclusions are pending team discussion

## Checklist:

## Summary by CodeRabbit

* **Enhancements**
  * Strengthened and clarified GCP field validations with more descriptive error messages and tighter length/format checks.
  * Many GCP fields are now optional in serialization to reduce payload noise.

* **Documentation**
  * Updated API reference to reflect new validation rules and optional/omitzero behavior for GCP platform fields.

* **Chores**
  * Bumped tooling dependency versions for improved validation support.


api/.golangci.yml
```diff
@@ -449,6 +449,1071 @@ linters:
           - kubeapilinter
         path: hypershift/v1beta1/hostedcluster_types.go
         text: 'maxlength: field ClusterVersionStatus.History must have a maximum items'
+      # minlength
+      - linters:
+          - kubeapilinter
+        path: hypershift/v1beta1/agent.go
+        text: 'minlength: field AgentPlatformSpec.AgentNamespace'
```
enxebre, line 456: this field is required, we can possibly adjust this. Is there any valid use case for this to be "" cc @eranco74 @CrystalChun

api/.golangci.yml
```diff
@@ -449,6 +449,1071 @@ linters:
...
+      - linters:
+          - kubeapilinter
+        path: hypershift/v1beta1/aws.go
+        text: 'minlength: field AWSNodePoolPlatform.InstanceType'
+      - linters:
+          - kubeapilinter
+        path: hypershift/v1beta1/aws.go
+        text: 'minlength: field AWSNodePoolPlatform.InstanceProfile'
```
enxebre, line 464: InstanceProfile is optional, it needs adjustment so when is set it has a minlength

api/.golangci.yml
```diff
@@ -449,6 +449,1071 @@ linters:
...
         text: 'maxlength: field ClusterVersionStatus.History must have a maximum items'
+      # minlength
+      - linters:
+          - kubeapilinter
+        path: hypershift/v1beta1/agent.go
+        text: 'minlength: field AgentPlatformSpec.AgentNamespace'
+      - linters:
+          - kubeapilinter
+        path: hypershift/v1beta1/aws.go
+        text: 'minlength: field AWSNodePoolPlatform.InstanceType'
```
enxebre, line 460: InstanceType is required, the fact we currently allow "" is a bug, it needs adjustment

enxebre:
Let's merge this to prevent upcoming changes from breaking conventions and then follow up for removing unnecessary exclusions gradually
/approve

enxebre:
@cblecker  PTAL for the GCP markers

bryan-cox (approved):
/lgtm

api/.golangci.yml
CrystalChun: This field should always have a value and never be empty if Agent platform is used

api/hypershift/v1beta1/gcp.go
```diff
@@ -45,7 +45,7 @@ type GCPResourceLabel struct {
 	// +optional
 	// +kubebuilder:validation:MinLength=0
 	// +kubebuilder:validation:MaxLength=63
-	// +kubebuilder:validation:Pattern=`^$|^[a-z]([_a-z0-9-]{0,61}[a-z0-9])?$`
+	// +kubebuilder:validation:XValidation:rule="self.matches('^[a-z]([_a-z0-9-]{0,61}[a-z0-9])?$')",message="value must match GCP label format: lowercase letters, digits, underscores, or hyphens"
```
cblecker, line 48: The original Pattern `^$|^[a-z]([_a-z0-9-]{0,61}[a-z0-9])?$` allowed empty strings via the `^$|` prefix. This XValidation drops that prefix, so `value: ""` now fails validation.

[GCP documentation](https://cloud.google.com/compute/docs/labeling-resources) confirms: label values CAN be empty (0-63 chars). The field comment also says "Empty values are allowed by GCP" and `MinLength=0` is set.

Since Value is `*string`, nil (omitted) works, but explicitly sending `value: ""` is a regression from the original behavior.

The `endpointIP` field in `gcpprivateserviceconnect_types.go` already handles this correctly with `self == '' || self.matches(...)`. Suggest the same pattern here:

```suggestion
	// +kubebuilder:validation:XValidation:rule="self == '' || self.matches('^[a-z]([_a-z0-9-]{0,61}[a-z0-9])?$')",message="value must match GCP label format: lowercase letters, digits, underscores, or hyphens"
```

api/hypershift/v1beta1/gcp.go
```diff
@@ -120,9 +120,9 @@ type GCPPlatformSpec struct {
 	// +immutable
 	// +kubebuilder:validation:MinLength=6
 	// +kubebuilder:validation:MaxLength=30
-	// +kubebuilder:validation:Pattern=`^[a-z]([a-z0-9-]{4,28}[a-z0-9])$`
+	// +kubebuilder:validation:XValidation:rule="self.matches('^[a-z]([_a-z0-9-]{0,61}[a-z0-9])?$')",message="project must match GCP label format: lowercase letters, digits, underscores, or hyphens"
```
cblecker: This replaces the original project ID pattern `^[a-z]([a-z0-9-]{4,28}[a-z0-9])$` with the generic GCP label pattern. Per [GCP documentation](https://cloud.google.com/resource-manager/docs/creating-managing-projects), project IDs allow only "lowercase letters, numbers, and hyphens" — **underscores are not valid** in project IDs.

The current regex `^[a-z]([_a-z0-9-]{0,61}[a-z0-9])?$`:
- Allows underscores (`_`) — invalid per GCP docs
- Has wrong length semantics (`{0,61}` vs `{4,28}`, partially mitigated by MinLength/MaxLength)
- Error message says "GCP label format" instead of "GCP project ID"

Suggest restoring the original pattern as an XValidation:

```suggestion
	// +kubebuilder:validation:XValidation:rule="self.matches('^[a-z]([a-z0-9-]{4,28}[a-z0-9])$')",message="project must be a valid GCP project ID: 6-30 lowercase letters, digits, or hyphens"
```

api/hypershift/v1beta1/gcp.go
```diff
@@ -136,19 +136,19 @@ type GCPPlatformSpec struct {
 	// +immutable
 	// +kubebuilder:validation:MinLength=1
 	// +kubebuilder:validation:MaxLength=63
-	// +kubebuilder:validation:Pattern=`^[a-z]+-[a-z]+[0-9]+$`
+	// +kubebuilder:validation:XValidation:rule="self.matches('^[a-z]([_a-z0-9-]{0,61}[a-z0-9])?$')",message="region must match GCP label format: lowercase letters, digits, underscores, or hyphens"
```
cblecker: This replaces the original region pattern `^[a-z]+-[a-z]+[0-9]+$` with the generic GCP label pattern. Per [GCP documentation](https://cloud.google.com/compute/docs/regions-zones), all GCP regions follow the `area-location#` format (e.g., `us-central1`, `europe-west12`, `northamerica-northeast1`).

The current regex `^[a-z]([_a-z0-9-]{0,61}[a-z0-9])?$`:
- Allows underscores — no GCP region contains underscores
- No longer requires a hyphen — every GCP region has at least one
- No longer requires trailing digits — every GCP region ends with digits
- Would accept invalid strings like "abc" or "us-west"
- Error message says "GCP label format" instead of "GCP region"

Suggest restoring the original pattern as an XValidation:

```suggestion
	// +kubebuilder:validation:XValidation:rule="self.matches('^[a-z]+-[a-z]+[0-9]+$')",message="region must be a valid GCP region (e.g., us-central1)"
```

cblecker (changes requested):
Overall this is well-structured work — the GA vs non-GA split, commit separation, and linter alignment are all solid. The `omitempty` additions on `+required` string fields (flagged by CodeRabbit) are correct per kube-api-linter SSA requirements since all affected fields also have `MinLength >= 1`.

However, there are 3 regex issues in the non-GA GCP type changes that need to be fixed before merge. All three have been verified against official GCP documentation:

1. **GCPResourceLabel.Value** (line 48) — XValidation drops empty string support. The original Pattern `^$|^[a-z]...` allowed empty strings; the new XValidation does not. GCP docs confirm label values CAN be empty.

2. **GCPPlatformSpec.Project** (line 123) — Uses generic label pattern instead of project ID pattern. Allows underscores which are invalid per GCP docs for project IDs.

3. **GCPPlatformSpec.Region** (line 139) — Uses generic label pattern instead of region pattern. No longer requires hyphen or trailing digits, accepting invalid region names.

All three appear to be copy-paste of the label key regex where field-specific patterns were needed. The original patterns were correct per GCP documentation and should be preserved as XValidation rules.

**Correct conversions** (for reference): GCPResourceReference.Name, GCPResourceLabel.Key, PoolID, ProviderID, ProjectNumber, all 5 service account emails, MachineType, Zone, Subnet, KMSKeyName, and all PSC patterns are all correctly converted from Pattern to XValidation.

api/hypershift/v1beta1/gcp.go
```diff
@@ -13,8 +13,8 @@ type GCPResourceReference struct {
 	// +required
 	// +kubebuilder:validation:MinLength=1
 	// +kubebuilder:validation:MaxLength=63
-	// +kubebuilder:validation:Pattern=`^[a-z]([-a-z0-9]*[a-z0-9])?$`
-	Name string `json:"name"`
+	// +kubebuilder:validation:XValidation:rule="self.matches('^[a-z]([-a-z0-9]*[a-z0-9])?$')",message="name must conform to GCP resource naming standards"
```
JoelSpeed: How does an end user understand GCP resource naming standards?


```suggestion
	// +kubebuilder:validation:XValidation:rule="self.matches('^[a-z]([-a-z0-9]*[a-z0-9])?$')",message="name must consist of only lowercase alphanumeric characters and hyphens, must start with an alphabetic character, and end with an alphanumeric character"
```

api/hypershift/v1beta1/gcp.go
```diff
@@ -33,9 +33,9 @@ type GCPResourceLabel struct {
 	// +required
 	// +kubebuilder:validation:MinLength=1
 	// +kubebuilder:validation:MaxLength=63
-	// +kubebuilder:validation:Pattern=`^[a-z]([_a-z0-9-]{0,61}[a-z0-9])?$`
+	// +kubebuilder:validation:XValidation:rule="self.matches('^[a-z]([_a-z0-9-]{0,61}[a-z0-9])?$')",message="key must match GCP label format: lowercase letters, digits, underscores, or hyphens"
```
JoelSpeed: You need to explain the start and end as well
```suggestion
	// +kubebuilder:validation:XValidation:rule="self.matches('^[a-z]([_a-z0-9-]{0,61}[a-z0-9])?$')",message="key must consist of only lowercase alphanumeric characters, hyphens and underscores, must start with an alphabetic character, and end with an alphanumeric character"
```

api/hypershift/v1beta1/gcp.go
JoelSpeed: This is an optional field, which means users don't actually need to set it. Currently, with the above suggestion, I can omit the field completely, or, I can explicitly send `""`. What behavioural difference does that make? We try to avoid being able to write the same thing twice in an API so it may be better to either make this required (explicit `""`) or disallow `""` and document `when omitted, the resource will be labelled with an empty value.

See also my other comments about how to format the validation message and apply similarly here

api/hypershift/v1beta1/gcp.go
```diff
@@ -161,6 +161,7 @@ type GCPPlatformSpec struct {
 	// +optional
 	// +listType=map
 	// +listMapKey=key
+	// +kubebuilder:validation:MinItems=0
```
JoelSpeed: This is generally considered bad practice. What is the behavioural difference between omitting this field and setting it to `[]` explicitly?

api/hypershift/v1beta1/gcp.go
```diff
@@ -267,9 +268,9 @@ type GCPServiceAccountsEmails struct {
 	//
 	// +required
 	// +immutable
-	// +kubebuilder:validation:Pattern=`^[a-z][a-z0-9-]{4,28}[a-z0-9]@[a-z][a-z0-9-]{4,28}[a-z0-9]\.iam\.gserviceaccount\.com$`
 	// +kubebuilder:validation:MinLength=37
 	// +kubebuilder:validation:MaxLength=100
+	// +kubebuilder:validation:XValidation:rule="self.matches('^[a-z][a-z0-9-]{4,28}[a-z0-9]@[a-z][a-z0-9-]{4,28}[a-z0-9]\\.iam\\.gserviceaccount\\.com$')",message="nodePool must be a valid GCP service account email"
```
JoelSpeed: This can be explained more thoroughly so that a user has a chance of understand what that means

api/hypershift/v1beta1/gcp.go
```diff
@@ -288,9 +289,9 @@ type GCPServiceAccountsEmails struct {
 	//
 	// +required
 	// +immutable
-	// +kubebuilder:validation:Pattern=`^[a-z][a-z0-9-]{4,28}[a-z0-9]@[a-z][a-z0-9-]{4,28}[a-z0-9]\.iam\.gserviceaccount\.com$`
 	// +kubebuilder:validation:MinLength=37
 	// +kubebuilder:validation:MaxLength=100
+	// +kubebuilder:validation:XValidation:rule="self.matches('^[a-z][a-z0-9-]{4,28}[a-z0-9]@[a-z][a-z0-9-]{4,28}[a-z0-9]\\.iam\\.gserviceaccount\\.com$')",message="controlPlane must be a valid GCP service account email"
```
JoelSpeed: Is it worth creating a type alias to contain this validation and be re-used in places where you need the same validation?

api/hypershift/v1beta1/gcp.go
```diff
@@ -417,6 +418,7 @@ type GCPNodePoolPlatform struct {
 	//   - A full resource URL: https://www.googleapis.com/compute/v1/projects/rhel-cloud/global/images/rhel-8-v20231010
 	//
 	// +optional
+	// +kubebuilder:validation:MinLength=1
```
JoelSpeed, line 417: This field doesn't need to be a pointer if it has a min length of 1

api/hypershift/v1beta1/gcp.go
```diff
@@ -444,6 +446,7 @@ type GCPNodePoolPlatform struct {
 	// +optional
 	// +listType=map
 	// +listMapKey=key
+	// +kubebuilder:validation:MinItems=0
```
JoelSpeed: As above, what is the difference between omit and explicit `[]`

api/hypershift/v1beta1/gcp.go
```diff
@@ -457,6 +460,7 @@ type GCPNodePoolPlatform struct {
 	//
 	// +optional
 	// +listType=set
+	// +kubebuilder:validation:MinItems=0
```
JoelSpeed: Difference between omit and explicit `[]`?

api/hypershift/v1beta1/gcp.go
```diff
@@ -503,15 +507,15 @@ type GCPBootDisk struct {
...
+	// +default="pd-balanced"
 	// +kubebuilder:validation:Enum=pd-standard;pd-ssd;pd-balanced
 	DiskType *string `json:"diskType,omitempty"`
 
 	// encryptionKey specifies customer-managed encryption key (CMEK) configuration.
 	// If not specified, Google-managed encryption keys are used.
 	//
 	// +optional
-	EncryptionKey *GCPDiskEncryptionKey `json:"encryptionKey,omitempty"`
+	EncryptionKey *GCPDiskEncryptionKey `json:"encryptionKey,omitzero"`
```
JoelSpeed, line 518: Does this need to be a pointer? What does `encryptionKey: {}` do? Vs omitting the field?

api/hypershift/v1beta1/gcp.go
```diff
@@ -550,7 +555,9 @@ type GCPNodeServiceAccount struct {
 	//
 	// +optional
 	// +listType=set
+	// +kubebuilder:validation:MinItems=0
```
JoelSpeed: Difference between omit and explicit `[]`?

api/hypershift/v1beta1/gcpprivateserviceconnect_types.go
```diff
@@ -42,12 +42,14 @@ type DNSZoneStatus struct {
 	// +required
 	// +kubebuilder:validation:MinLength=1
 	// +kubebuilder:validation:MaxLength=253
-	Name string `json:"name"`
+	Name string `json:"name,omitempty"`
 
 	// records lists the DNS records created in this zone
 	// +optional
 	// +listType=atomic
+	// +kubebuilder:validation:MinItems=0
```
JoelSpeed, line 50: Difference between omitted and `[]`?

api/hypershift/v1beta1/gcpprivateserviceconnect_types.go
```diff
@@ -59,30 +61,35 @@ type GCPPrivateServiceConnectSpec struct {
 	// This value must be a valid IPv4 or IPv6 address.
 	// +required
 	// +kubebuilder:validation:XValidation:rule="self.matches('^((\\\\d{1,3}\\\\.){3}\\\\d{1,3})$') || self.matches('^(([0-9a-fA-F]{1,4}:){7}[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,7}:|([0-9a-fA-F]{1,4}:){1,6}:[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,5}(:[0-9a-fA-F]{1,4}){1,2}|([0-9a-fA-F]{1,4}:){1,4}(:[0-9a-fA-F]{1,4}){1,3}|([0-9a-fA-F]{1,4}:){1,3}(:[0-9a-fA-F]{1,4}){1,4}|([0-9a-fA-F]{1,4}:){1,2}(:[0-9a-fA-F]{1,4}){1,5}|[0-9a-fA-F]{1,4}:((:[0-9a-fA-F]{1,4}){1,6})|:((:[0-9a-fA-F]{1,4}){1,7}|:))$')",message="loadBalancerIP must be a valid IPv4 or IPv6 address"
```
JoelSpeed: We have CEL libraries for validating IP and CIDR 👀

api/hypershift/v1beta1/gcpprivateserviceconnect_types.go
```diff
@@ -95,21 +102,24 @@ type GCPPrivateServiceConnectStatus struct {
 	// +listMapKey=type
 	// +patchMergeKey=type
 	// +patchStrategy=merge
+	// +kubebuilder:validation:MinItems=0
```
JoelSpeed, line 105: Difference between unset and `[]`?

api/hypershift/v1beta1/gcpprivateserviceconnect_types.go
```diff
@@ -118,11 +128,14 @@ type GCPPrivateServiceConnectStatus struct {
 	// This value must be a valid IPv4 or IPv6 address.
 	// +optional
 	// +kubebuilder:validation:XValidation:rule="self == '' || self.matches('^((\\\\d{1,3}\\\\.){3}\\\\d{1,3})$') || self.matches('^(([0-9a-fA-F]{1,4}:){7}[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,7}:|([0-9a-fA-F]{1,4}:){1,6}:[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,5}(:[0-9a-fA-F]{1,4}){1,2}|([0-9a-fA-F]{1,4}:){1,4}(:[0-9a-fA-F]{1,4}){1,3}|([0-9a-fA-F]{1,4}:){1,3}(:[0-9a-fA-F]{1,4}){1,4}|([0-9a-fA-F]{1,4}:){1,2}(:[0-9a-fA-F]{1,4}){1,5}|[0-9a-fA-F]{1,4}:((:[0-9a-fA-F]{1,4}){1,6})|:((:[0-9a-fA-F]{1,4}){1,7}|:))$')",message="endpointIP must be a valid IPv4 or IPv6 address"
+	// +kubebuilder:validation:MinLength=0
```
JoelSpeed: Why is a 0 length acceptable here? What is the difference between this being unset and explicitly `""`?

Note that at the moment, `""` won't round trip through a structured client, so it's impossible to leverage this functionality correctly if there is a behavioural difference

api/hypershift/v1beta1/gcpprivateserviceconnect_types.go
```diff
@@ -118,11 +128,14 @@ type GCPPrivateServiceConnectStatus struct {
...
 	// +kubebuilder:validation:MaxLength=45
 	EndpointIP string `json:"endpointIP,omitempty"`
 
 	// dnsZones contains DNS zone information created for this cluster
+	// +optional
 	// +listType=atomic
+	// +kubebuilder:validation:MinItems=0
```
JoelSpeed, line 139: Difference between omitted and `[]`?

JoelSpeed (commented):
For every CEL rule you've added to describe a pattern, please make sure that you actually describe the pattern in the message. Users won't know what a valid GCP project ID looks like, so explain it to them in terms of which characters are allowed and when (start and end), email even more so

For every list, it's generally not advised to distinguish between unset and explicit `[]`. For each, check if there's an intended behavioural difference between unset and an explicit empty list, if there is not, make the MinItems 1 or more

If there is a deliberate behavioural difference, please tag me and we can discuss how to proceed. You'll need at least to make the list `*[]T`

api/hypershift/v1beta1/gcp.go
Nirshal: Ty for the feedback, I'll apply the suggested pattern, but as message I'll use this one: "project must be 6 to 30 characters, contain only lowercase letters, digits, or hyphens, start with a letter, and not end with a hyphen", as suggested by @JoelSpeed 
Looking at the doc you linked, I noticed a rule that was not enforced before:
- Cannot contain restricted strings such as google and ssl.[...]
Do you think it is worth to delve further into this and find a list of restricted words to make the validation more strict?

api/hypershift/v1beta1/gcp.go
Nirshal: @cblecker I would like to implement @JoelSpeed suggestion on this, but of course I am aiming for general consensus here, is it ok for you?

api/hypershift/v1beta1/gcp.go
Nirshal: Ty for your feedback! I have been looking into this:
- on a serialization point of view, we have "omitempty" that makes nil  or [] equivalent because both are absent in the json
- hypershift-operator/controllers/nodepool/gcp.go:274 it is iterated with a "for range" so both nil or [] are ok
- hypershift-operator/controllers/hostedcluster/internal/platform/gcp/gcp.go:146 there is an if(len(...))>0 that is ok with both nil or [] so it's irrelevant here too.

For this reason, I'll set `+kubebuilder:validation:MinItems=1` to reinforce the semantic that if it is optional you either omit it or you must provide at least one item. I'll make sure to reflect this in the field documentation too.

api/hypershift/v1beta1/gcp.go
Nirshal: ty, I'll fix this with next commit

api/hypershift/v1beta1/gcp.go
Nirshal: I'll fix this with this message that also re-iterate on the number of characters: "name must be 1-63 characters, start with a lowercase letter, contain only lowercase letters, digits, or hyphens, and not end with a hyphen"

api/hypershift/v1beta1/gcp.go
cblecker: My take: we are trying to catch typos, copy/paste errors, and similar issues. I don't think we need to get into restricted strings, as those will be caught by google when someone is creating the GCP project. At this point, we can assume that the GCP project is already created, and we are just pointing hypershift to it

JoelSpeed: Yeah I would agree, lets not restrict on the restrict google/ssl strings here

api/hypershift/v1beta1/gcp.go
Nirshal: I'll fix the description with next commit

api/hypershift/v1beta1/gcp.go
cblecker: Then I'd suggest we go with this field being required, but an explicit empty value (`""`) being okay. The label is a kv pair, so this makes sense.

api/hypershift/v1beta1/gcp.go
Nirshal: I am considering to create a type alias `GCPServiceAccountEmail` as suggested, but the XValidation message would lose the field name (e.g. 'controlPlane must be...'), making it harder to identify which field has the error when multiple fail. Is there a way to include the field name in a type-level XValidation message, or is the Kubernetes error response path sufficient to disambiguate? 
I would include in the type alias the min and max length, since all the interested fields have `MinLength=37`, but I noticed that the Email field in `GCPNodeServiceAccount` has ¡MaxLength=254¡ (RFC 5321 generic email limit), while all 5 fields in `GCPServiceAccountsEmails` have `MaxLength=100`. Since the regex enforces the format `name@project.iam.gserviceaccount.com` where both name and project-id are max 30 chars, the theoretical maximum is 85 characters (30+1+30+24). Should we align MaxLength to 100 across all fields?

api/hypershift/v1beta1/gcp.go
Nirshal: Agree, I'll address this together with next comment introducing a type alias as you suggested.

api/hypershift/v1beta1/gcp.go
JoelSpeed: >  making it harder to identify which field has the error when multiple fail

The error output includes the field name in a prefix to the message, including the whole path to the field name

If the theoretical maximum is 85, then set the limit to 85

api/hypershift/v1beta1/gcp.go
Nirshal: Actually I just noticed reading [here](https://docs.cloud.google.com/compute/docs/naming-resources)  that the pattern is wrong because there is no need for the value to start with a lowercase letter. Also, the doc allows international characters but our regex does not manage that... I would at least change the pattern to relax the first letter requirement, if you agree.

api/hypershift/v1beta1/gcpprivateserviceconnect_types.go
Nirshal: I'll use "isIP()" CEL function, both here and for the EndpointIP at row 130, thanks for pointing out to this.

api/hypershift/v1beta1/gcp.go
Nirshal: The rationale behind this is that the pointer distinguishes "not specified" (nil → use default RHCOS image) from "specified" (must be non-empty). The empty string "" is not a valid image reference, so having two representations for "not specified" (nil and "") is avoided by MinLength=1. 
As an alternative, we could remove the pointer, make the field required, set MinLength=0, and treat "" as "use the default RHCOS image". This is the same pattern we applied to GCPResourceLabel.Value per @cblecker's suggestion. However, unlike label values where "" is a legitimate GCP value, here "" would purely mean "use the default", which feels less natural.
I'm leaning toward keeping this as it is, but of course I'll be happy to change if you prefer the other option.

api/hypershift/v1beta1/gcp.go
Nirshal: This is quite the same as row 164, I'll take the same approach.

api/hypershift/v1beta1/gcp.go
Nirshal: It's again the same situation as ResourceLabels, I'll apply the same fix

api/hypershift/v1beta1/gcp.go
Nirshal: It's again the same situation as ResourceLabels and NetworkTags, I'll apply the same fix

api/hypershift/v1beta1/gcpprivateserviceconnect_types.go
Nirshal: Regarding Records in DNSZoneStatus (api/hypershift/v1beta1/gcpprivateserviceconnect_types.go:47-54):

The controller always populates this field for every DNSZoneStatus entry (control-plane-operator/controllers/gcpprivateserviceconnect/psc_endpoint_controller.go:345-358), including cases where the list is explicitly empty — e.g. PrivateIngressCreatedRecords: []string{} (control-plane-operator/controllers/gcpprivateserviceconnect/dns.go:605).

Changing MinItems=0 to MinItems=1 would break the controller's current behavior. Should this field instead be +required (removing omitempty) since the controller always sets it? That would keep MinItems=0 valid — "always present, but can be an empty list". The concern is backward compatibility with existing resources that may not have Records set.

Would appreciate guidance on the preferred approach here.

api/hypershift/v1beta1/gcpprivateserviceconnect_types.go
Nirshal: Same situation as Records (see my [previous comment](https://github.com/openshift/hypershift/pull/7903/changes#r2924187179)): Conditions in GCPPrivateServiceConnectStatus (api/hypershift/v1beta1/gcpprivateserviceconnect_types.go:98-107) is a status field populated progressively by the controller via meta.SetStatusCondition (control-plane-operator/controllers/gcpprivateserviceconnect/psc_endpoint_controller.go:329,379,737,745,928).

The list starts empty when the resource is first created and conditions are added one at a time. Changing MinItems=0 to MinItems=1 would prevent the controller from saving the initial empty status. Additionally, Conditions []metav1.Condition is a standard Kubernetes pattern where MinItems=0 (or no MinItems at all) is the norm.

Same question applies: should we keep MinItems=0, remove it entirely (since 0 is the default), or make the field +required?

api/hypershift/v1beta1/gcp.go
Nirshal: Heads up: after making Value a required non-pointer string (as discussed), the kube-api-linter now flags it:


requiredfields: field GCPResourceLabel.Value has a valid zero value ("") and should be a pointer.
This is expected — the linter considers that a required field whose zero value ("") is semantically valid should be a pointer to distinguish "not set" from "empty". In our case we intentionally want "" to be a valid value (GCP allows empty label values), and since the field is required the "not set" case doesn't apply.

This would require an additional linter exclusion. Are we sure we want to go this route, or would you prefer a different approach?

api/hypershift/v1beta1/gcpprivateserviceconnect_types.go
Nirshal: With string (non-pointer) + omitempty, there is no difference between unset and "" — both are serialized identically (field omitted). There is no behavioral difference in the controller (control-plane-operator/controllers/gcpprivateserviceconnect/psc_endpoint_controller.go:313), which simply checks if gcpPSC.Status.EndpointIP == "".

In order to remove redundancies, I'll remove the self == '' || prefix from the XValidation, since with omitempty the empty string never reaches validation. I'll also change MinLength=0 to MinLength=1, which is more semantically accurate — and avoids having to add a purely formal linter exclusion for the minlength rule. I'll also update the doc comment to clarify that the field can be omitted, but if specified it must be a valid IP address.

api/hypershift/v1beta1/gcpprivateserviceconnect_types.go
Nirshal: Same situation as Records (see my [previous comment](https://github.com/openshift/hypershift/pull/7903/changes#r2924187179)): DNSZones in GCPPrivateServiceConnectStatus (api/hypershift/v1beta1/gcpprivateserviceconnect_types.go:136-141) is a status field. The controller checks if len(gcpPSC.Status.DNSZones) == 0 (control-plane-operator/controllers/gcpprivateserviceconnect/psc_endpoint_controller.go:419) for the initial state before DNS reconciliation, and then populates it with 3 zones (psc_endpoint_controller.go:345). With omitempty, nil and [] are serialized identically (field omitted), so there is no behavioral difference between the two.

Changing MinItems=0 to MinItems=1 would prevent the controller from saving the initial empty status. Should we keep MinItems=0 or make the field +required?

api/hypershift/v1beta1/gcp.go
Nirshal: The pointer is needed here to distinguish "not specified" (nil → Google-managed encryption) from "specified" (non-nil → CMEK). The controller uses if bootDiskConfig.EncryptionKey != nil (hypershift-operator/controllers/nodepool/gcp.go:256) to decide whether to configure customer-managed encryption.

Regarding encryptionKey: {} — since KMSKeyName is +required with MinLength=1, an empty struct would fail validation, so there is no ambiguity between omitting and specifying an empty value.

Regarding omitzero vs omitempty (also flagged by coderabbit): omitzero is required by the kube-api-linter for optional struct fields (.golangci.yml:55-58), so this is intentional.

Nirshal:
Pushed changes addressing some of the review feedbacks, see individual comment replies for details. A few items are pending your input. Thanks for your valuable feedback!

cmd/install/assets/hypershift-operator/zz_generated.crd-manifests/nodepools-TechPreviewNoUpgrade.crd.yaml
```diff
@@ -1178,22 +1198,32 @@ spec:
                                 - Logging writer
                                 - Monitoring metric writer
                                 - Storage object viewer (for pulling container images)
-                            maxLength: 254
-                            pattern: ^[a-z][a-z0-9-]{4,28}[a-z0-9]@[a-z][a-z0-9-]{4,28}[a-z0-9]\.iam\.gserviceaccount\.com$
+                            maxLength: 85
+                            minLength: 37
```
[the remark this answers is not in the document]
Nirshal: That's a false positive, Coderabbit forgot to count the "@" 🙂

cmd/install/assets/hypershift-operator/zz_generated.crd-manifests/gcpprivateserviceconnects-TechPreviewNoUpgrade.crd.yaml
```diff
@@ -67,35 +67,48 @@ spec:
...
                   loadBalancerIP is the IP address of the Internal Load Balancer
                   Populated by the observer from service status
                   This value must be a valid IPv4 or IPv6 address.
                 maxLength: 45
+                minLength: 7
                 type: string
                 x-kubernetes-validations:
                 - message: loadBalancerIP must be a valid IPv4 or IPv6 address
-                  rule: self.matches('^((\\d{1,3}\\.){3}\\d{1,3})$') || self.matches('^(([0-9a-fA-F]{1,4}:){7}[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,7}:|([0-9a-fA-F]{1,4}:){1,6}:[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,5}(:[0-9a-fA-F]{1,4}){1,2}|([0-9a-fA-F]{1,4}:){1,4}(:[0-9a-fA-F]{1,4}){1,3}|([0-9a-fA-F]{1,4}:){1,3}(:[0-9a-fA-F]{1,4}){1,4}|([0-9a-fA-F]{1,4}:){1,2}(:[0-9a-fA-F]{1,4}){1,5}|[0-9a-fA-F]{1,4}:((:[0-9a-fA-F]{1,4}){1,6})|:((:[0-9a-fA-F]{1,4}){1,7}|:))$')
+                  rule: self.isIP()
```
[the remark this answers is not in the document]
Nirshal: MinLength=7 is pre-existing and was not changed in this PR — only the XValidation rule was updated from a regex pattern to the built-in isIP() CEL function.

That said, this is a +required status field populated by the controller from the service load balancer status, not user input, so compressed IPv6 like ::1 would not realistically appear here.

api/hypershift/v1beta1/gcp.go
JoelSpeed: > Regarding encryptionKey: {} — since KMSKeyName is +required with MinLength=1, an empty struct would fail validation, so there is no ambiguity between omitting and specifying an empty value.

So this means there's no semantic difference. Drop the pointer here, you don't need it. 

Instead in the controller, compare the object to the empty object of the same kind to check whether it is configured or not

api/hypershift/v1beta1/gcp.go
JoelSpeed: It doesn't require a linter exclusion, it requires a pointer. If you follow the advice of the linter you'll have a correctly behaving structured client

If you ignore the linter, then a structured client doesn't actually get the benefit of the required tag. They will implicitly set the empty value when they may not have intended to do so

api/hypershift/v1beta1/gcp.go
JoelSpeed: Length constraints will trigger different validations, so I wouldn't include the length constraint in this error message

api/hypershift/v1beta1/gcp.go
JoelSpeed: You don't need to change the semantics of the field at all, you just need to drop the pointer

From a client perspective, they won't be able to persist an empty string, so your controller can observe the empty string as "client didn't set a value". If the value is non empty, we know the client did set a value.

We do not need to make the field required to remove the pointer

api/hypershift/v1beta1/gcpprivateserviceconnect_types.go
JoelSpeed: When we look at these serialization issues, I don't generally care what the controller is actually doing. I care about what the semantic differences are when the object is serialized.

So in this case, the question really is, do you want `records: []` to be printed into the serialized object when there are no records, or should the `records` key be omitted completely from that representation
(this is what I mean by a semantic difference btw, does the end user read some different meaning from this list being missing, vs deliberately the empty list)

If there is a genuine semantic difference, then make this `*[]T` with `omitempty` and `MinItems=0` is correct.
The linter may not agree with that statement though as I haven't implement this part of the functionality yet (this rule is different for CRDs vs native types)

api/hypershift/v1beta1/gcpprivateserviceconnect_types.go
JoelSpeed: There is no semantic difference for this list

Set this to MinItems 1 please

api/hypershift/v1beta1/gcpprivateserviceconnect_types.go
JoelSpeed: > Changing MinItems=0 to MinItems=1 would prevent the controller from saving the initial empty status.

This line is factually incorrect.

Please review this https://github.com/openshift/hypershift/pull/7903#discussion_r2932893765 and then reply here as well with your belief about the semantics of this field and whether `[]` has any difference in meaning to an end user, and then we can talk about the correct serialization

JoelSpeed:
I have removed the approve label from this PR

To other reviewers, please do not re-approve until I am happy with the state of this PR

There are currently many issues within this PR that need to be resolved before this PR can be merged and, given the nature of this PR, I think this needs particular attention from an API reviewer before approval

cblecker:
@Nirshal @JoelSpeed perhaps we can split this up into smaller chunks? I'd even be happy to take a personal pass at the GCP API bits specifically

JoelSpeed:
If there is a good way that we can split this so that we can get KAL enabled for new changes (this is what we do in o/api) and then come back to fill in older changes after, that would be beneficial IMO

api/.golangci.yml
```diff
@@ -43,2065 +111,3012 @@ linters:
...
+        path: hypershift/v1beta1/hostedcluster_types.go
+        text: 'defaultorrequired: field HostedClusterSpec.Etcd has conflicting markers'
+      - linters:
+          - kubeapilinter
+        path: hypershift/v1beta1/kubevirt.go
+        text: 'defaultorrequired: field KubevirtCachingStrategy.Type has conflicting markers'
+      - linters:
+          - kubeapilinter
+        path: hypershift/v1beta1/kubevirt.go
+        text: 'defaultorrequired: field KubevirtNodePoolPlatform.RootVolume has conflicting markers'
```
JoelSpeed, line 168: There is no way that the default can be effective for these cases. 

IMO you should drop the default tag since it has no effect. This might be good to do as a separate PR though to avoid creep

api/.golangci.yml
```diff
@@ -43,2065 +111,3012 @@ linters:
...
+        path: hypershift/v1beta1/powervs.go
+        text: 'defaults: field PowerVSNodePoolPlatform.ImageDeletePolicy should use \+default marker'
+      - linters:
+          - kubeapilinter
+        path: hypershift/v1beta1/powervs.go
+        text: 'defaults: field PowerVSNodePoolPlatform.SystemType should use \+default marker'
+      - linters:
+          - kubeapilinter
+        path: hypershift/v1beta1/powervs.go
+        text: 'defaults: field PowerVSNodePoolPlatform.ProcessorType should use \+default marker'
```
JoelSpeed, line 329: All of these can be fixed with `make lint-fix` without it being a breaking change, it shouldn't affect the generated content at all, separate commit?

api/.golangci.yml
```diff
@@ -43,2065 +111,3012 @@ linters:
...
+      - linters:
+          - kubeapilinter
+        path: hypershift/v1beta1/kubevirt.go
+        text: 'defaults: field KubevirtNodePoolPlatform.RootVolume has a default value but does not have omitempty'
+      # integers
       - linters:
           - kubeapilinter
         path: hypershift/v1beta1/hosted_controlplane.go
-        text: 'nobools: field HostedControlPlaneStatus.ExternalManagedControlPlane pointer should not use a bool. Use a string type with meaningful constant values as an enum.'
+        text: 'integers: field HostedControlPlaneStatus.NodeCount'
```
JoelSpeed, line 426: You haven't captured the useful bit of the error message here. Why do we need this exception?

api/.golangci.yml
JoelSpeed: Adding a non-zero minimum length should ratchet, so should be a safe change to make

api/.golangci.yml
```diff
@@ -43,2065 +111,3012 @@ linters:
...
+        text: 'minlength: field AWSEndpointServiceStatus.DNSZoneID'
+      - linters:
+          - kubeapilinter
+        path: hypershift/v1beta1/endpointservice_types.go
+        text: 'minlength: field AWSEndpointServiceStatus.SecurityGroupID'
+      - linters:
+          - kubeapilinter
+        path: hypershift/v1beta1/endpointservice_types.go
+        text: 'minlength: field AWSEndpointService.Status'
+      # MinProperties is not applicable on pointer/non-pointer structs with omitempty (codegen generates $ref, not inline object)
```
JoelSpeed, line 781: Huh? Can you expand on this? 

The `MinProperties` should be set on the struct definition itself, not the field, I think that's a marker scoping issue rather than anything else

api/.golangci.yml
```diff
@@ -43,2065 +111,3012 @@ linters:
...
-        text: 'requiredfields: field VPC should have the omitempty tag.'
+        path: hypershift/v1beta1/hosted_controlplane.go
+        text: 'requiredfields: field HostedControlPlaneStatus.Initialized should have the omitempty'
+      # ssatags
       - linters:
           - kubeapilinter
-        path: hypershift/v1beta1/powervs.go
-        text: 'requiredfields: field Zone has a valid zero value'
+        path: hypershift/v1beta1/hostedcluster_types.go
+        text: 'ssatags: ClusterNetworking.MachineNetwork should have a listType marker'
```
JoelSpeed, line 2966: All SSATags exceptions should be dropped. You can add `+listType=atomic` to each one without it being a breaking change

JoelSpeed:
What is the long term plan for all of the exceptions that are being added? Do you intend to follow up and resolve them, or will this list be as it is forever? If there is no intention to fix the issues in the long term, switching golangci-lint to check changes only since the last merged main might be an easier solution

CC @enxebre

api/hypershift/v1beta1/gcp.go
Nirshal: Sorry, I got a bit lost, my bad... let me try to summarize why my confusion arose:
- originally it was noted that the modification I made changed the validation behavior and did not allow empty string due to the new regex I used in XValidation;
- then I was also noted that ambiguity exists, because the user can obtain the same behavior (on a controller point of view) both omitting the field (because it's an optional and a pointer) and the controller treats  nil the same as `""` (`ptr.Deref(label.Value, "")`) then it was suggested to make the field `+required` to avoid that, but allow the empty value "";
- for some reason, which I attribute mainly to my lack of experience, I assumed it was unusual to have a required field combined with a pointer, now I see that it is not the case, I am sorry for that.
Then I'll proceed this way: I'll keep the pointer, change the optional to required and fix the validation (to admit empty string and also to not necessarily start with a lowercase letter, which it was never the case for the `Value` field, but it was just a requirement for the `Key`)

api/hypershift/v1beta1/gcp.go
Nirshal: Thanks for the explanation! I will drop the pointer in the next commit

api/hypershift/v1beta1/gcp.go
JoelSpeed: > for some reason, which I attribute mainly to my lack of experience, I assumed it was unusual to have a required field combined with a pointer

The guidance around serialization in K8s APIs has changed in the last ~year, so this is somewhat unfamiliar to a lot of people. A lot of work went into the linter so that it could recognise when pointers are required which boils down to "Is the zero value a valid user choice".

In this case, it is, so to make sure that a structured client is required to set the value, we need to actually force them to write something into the field, so we need a pointer. If they don't write anything, the field won't marshal, and they'll get a "field is required" error

If you ever have questions or confusion about things the linter recommends, poke the API review team in #forum-api-review and we can help you out

api/hypershift/v1beta1/gcp.go
Nirshal: I'll fix this with next commit, thanks!

JoelSpeed:
In terms of breaking this down, @enxebre and I agreed to break out the [updates to config](https://github.com/openshift/hypershift/pull/7982) so that we can start linting new APIs with the newer rules, and then we can start looking at fixes separately  (e.g. from this PR) once that has merged.

My suggestion for fixing the GCP parts (and I guess some of the older non breaking changes) would be that we take the existing config once it's merged:
* Remove the `--new-from-rev` argument from the makefile - this will then apply to all API changes, not just recent ones
* Edit the KAL config to disable all (`*`) linters and then enable 1x1 each linter and fix that specific linters issues
  * E.g. `disable: ['*'], enable: ['ssatags']` and then add `+listType=atomic` to all GA APIs, and choose appropriate list types for the GCP types

This should allow us to review and merge things in bitesize chunks I hope

I'm happy to be consulted on the expectation for each linter as we go through them all as well, just drop me a note and I can help work out what we should do for each

cblecker:
@Nirshal @JoelSpeed I attempted to split off just the GCP stuff, as we'd like to get any breaking changes in ASAP. Please have a look: https://github.com/openshift/hypershift/pull/8003

Nirshal:
@cblecker @JoelSpeed @enxebre this PR had 2 purposes:
- update the api linters configuration (scope the .golangci.yml) but that has been addressed in Joel's new pull [#7982](https://github.com/openshift/hypershift/pull/7982)
- address linters errors in the GCP API, but now that is in Christoph's PR [#8003](https://github.com/openshift/hypershift/pull/8003)
All other API's linters exceptions were out of scope, to avoid breaking changes.
Since I have uncommitted changes that I could push on this, but are mainly related to Joel's review comments, I am wondering if this particular PR still makes sense, and what the scope of the work should be once rebased on the the previously listed PRs above. Can you provide directions on the matter?

JoelSpeed:
There are a huge number of comments on this PR that aren't addressed in #8003 

To quote my review there
> Things I'm thinking we should be reviewing:
>
> * Lots of fields have very large maximum lengths that they probably don't need
> * Most fields don't document their validations
> * Are we sure we have the right validation patterns for GCP upstream fields - Recent experience suggests that possibly Claude may have come up with some of these regexes and may not have been precise (e.g. missing the fact you can't have consecutive - in most resource naming conventions on Azure)

IMO there's still value in assessing these points

Nirshal:
> There are a huge number of comments on this PR that aren't addressed in #8003
> 
> To quote my review there
> 
> > Things I'm thinking we should be reviewing:
> > 
> > * Lots of fields have very large maximum lengths that they probably don't need
> > * Most fields don't document their validations
> > * Are we sure we have the right validation patterns for GCP upstream fields - Recent experience suggests that possibly Claude may have come up with some of these regexes and may not have been precise (e.g. missing the fact you can't have consecutive - in most resource naming conventions on Azure)
> 
> IMO there's still value in assessing these points

Ok that make sense, it will be a bit chaotic though, because some of the changes I already made addresses the changes proposed by Christoph, so there will be conflicts... how do you suggest I proceed? Should I wait for the #8003 to merge and then rebase on that one and see what's left? Sorry for the trivial questions, I am new to the team and hypershift development so I appreciate any suggestion you may have 🙂

JoelSpeed:
I suspect waiting until Christoph's PR is merged is a good idea before continuing with the remaining items

cblecker:
@Nirshal I'm sorry if I stepped on toes -- I just wanted to get things correct in line with the work the GCP HCP is doing as quickly as possible, as breaking changes to things like the API may require refactors on other stuff we are doing and we wanted to maintain our project's velocity.

Nirshal:
> @Nirshal I'm sorry if I stepped on toes -- I just wanted to get things correct in line with the work the GCP HCP is doing as quickly as possible, as breaking changes to things like the API may require refactors on other stuff we are doing and we wanted to maintain our project's velocity.

@cblecker no worries, I just wanted to know how to move forward with this PR and if still had meaning, but Joel made a good point on that. I'll wait for #8003 to merge and then will rebase and fix the remaining comments. I understand the need for velocity is critical, so I'll provide any support you may need.

cblecker:
Thanks @Nirshal ! #8003 has now merged, so all the GCP related lint issues should be resolved

Nirshal:
> Thanks @Nirshal ! #8003 has now merged, so all the GCP related lint issues should be resolved

Thanks @cblecker, I'll resume working on this today's evening or at later tomorrow morning.

Nirshal:
Closing this PR as the work has been superseded. During review, it was split across multiple PRs by different contributors, and the branch is now 1000+ commits behind main with merge conflicts.

**Completed work:**

| PR | Author | Merged | What it did |
|-----|--------|--------|-------------|
| #7333 | @cblecker | 2025-12-04 | Removed unused exclusion rules |
| #7346 | @bryan-cox | 2025-12-16 | Updated kube-api-linter dependency |
| #7982 | @JoelSpeed | 2026-03-17 | Aligned KAL config with openshift/api, added --new-from-rev to lint only new changes |
| #8003 | @cblecker | 2026-03-20 | Fixed all GCP API type violations (split from this PR) |
| #8037 | @cristianoveiga | 2026-03-25 | Updated test expectations after GCP fixes |

**Current state on main:** 18 of 22 KAL linters are active. 489 per-field exclusions remain in `api/.golangci.yml`, covering GA API types. The `--new-from-rev` flag ensures new code is linted without exclusions.

**Remaining work from this PR's review (unresolved comments):**

I've created sub-tasks under [CNTRLPLANE-2049](https://issues.redhat.com/browse/CNTRLPLANE-2049) for the three items still open:
- Add MinLength validation to AgentPlatformNamespace, InstanceType, and InstanceProfile ([enxebre](https://github.com/openshift/hypershift/pull/7903#discussion_r2911028554), [CrystalChun](https://github.com/openshift/hypershift/pull/7903#discussion_r2911028554), [JoelSpeed](https://github.com/openshift/hypershift/pull/7903#discussion_r2911028554))
- Resolve ssatags violations by adding +listType=atomic markers ([JoelSpeed](https://github.com/openshift/hypershift/pull/7903#discussion_r2939663506))
- Remove ineffective defaults linter exclusions ([JoelSpeed](https://github.com/openshift/hypershift/pull/7903#discussion_r2939636863))

The broader linter-by-linter effort to reduce the remaining 489 exclusions ([as proposed by @JoelSpeed](https://github.com/openshift/hypershift/pull/7903#issuecomment-4073881246)) is tracked by the parent task.

Thanks to @JoelSpeed, @cblecker, @enxebre, and @CrystalChun for the thorough review.

PR #7907, opened 2026-03-10
OCPBUGS-77557: propagate additionalTrustBundle to AWS control plane components

sdminonne:
## Summary

- Add `DeploymentAddAWSCABundleVolume` helper that creates a combined CA bundle (system + user CAs) via an init container and sets `AWS_CA_BUNDLE` on the main container
- Wire trust bundle propagation into all AWS control plane components when `AdditionalTrustBundle` is set on the HostedControlPlane spec:
  - `aws-cloud-controller-manager`
  - `capi-provider`
  - `ingress-operator`
  - `karpenter`
  - `karpenter-operator`
  - `aws-node-termination-handler`
  - `kube-apiserver` AWS KMS sidecars (`aws-kms-active`, `aws-kms-backup`) when `SecretEncryption.KMS.Provider` is AWS
- Add unit tests for all components and a v2 e2e test for `aws-cloud-controller-manager`

## Problem

In isolated AWS environments (e.g., US-ISO regions), custom CA bundles specified via `HostedCluster.Spec.AdditionalTrustBundle` are not propagated to AWS control plane components. This causes TLS verification failures when these components call AWS API endpoints:

```
Post https://sts.us-iso-east-1.c2s.ic.gov: tls: failed to verify certificate:
x509: certificate signed by unknown authority
```

## Why not reuse `DeploymentAddTrustBundleVolume`?

The existing helper mounts a ConfigMap as a directory at `/etc/pki/tls/certs`, which **replaces** the entire system CA directory. This works for in-house components (CPO, ignition-server, OAPI) whose TLS needs are tightly controlled. However, the affected components are binaries that make HTTPS calls to standard AWS service endpoints (EC2, ELB, STS, SQS, KMS). The AWS SDK's default HTTP client loads the system CA store from `/etc/pki/tls/certs` to verify TLS certificates. Replacing that directory with a ConfigMap containing only the custom CA would cause the binary to lose the public root CAs (e.g., Amazon Trust Services), breaking connectivity to standard AWS API endpoints.

## Why `AWS_CA_BUNDLE` with a combined bundle?

The AWS SDK (both v1 and v2) reads `AWS_CA_BUNDLE` and uses it **instead of** the system CA bundle — it creates a new empty `x509.CertPool` and loads only the specified file. To avoid losing trust in standard AWS endpoints, an init container concatenates the system CAs (`/etc/pki/tls/certs/ca-bundle.crt`) with the user-provided CAs from `additionalTrustBundle` into a single combined PEM file. `AWS_CA_BUNDLE` points to this combined file, ensuring the AWS SDK trusts both system and custom CAs.

## KAS KMS sidecars

When secret encryption uses AWS KMS (`SecretEncryption.KMS.Provider == AWS`), the `aws-kms-active` and `aws-kms-backup` sidecar containers in the kube-apiserver deployment also need access to the combined CA bundle. These sidecars call AWS KMS endpoints to encrypt/decrypt data encryption keys. The `aws-kms-token-minter` sidecar is intentionally excluded as it does not make AWS API calls.

## Test plan

- [x] Unit tests verify volume, init container, mount, and env var presence when `AdditionalTrustBundle` is set
- [x] Unit tests verify no volume/env var when `AdditionalTrustBundle` is nil
- [x] Unit tests verify non-AWS platforms are unaffected (capi-provider, ingress-operator, karpenter-operator)
- [x] Unit tests verify `aws-kms-active` and `aws-kms-backup` get volume mount and `AWS_CA_BUNDLE` env var
- [x] Unit tests verify `aws-kms-token-minter` is not wired
- [x] Unit tests verify no KMS wiring when KMS containers are absent
- [x] v2 e2e test verifies `AWS_CA_BUNDLE` wiring on `aws-cloud-controller-manager` (add and remove)
- [x] `make test` passes
- [x] `make verify` passes

Fixes: https://issues.redhat.com/browse/OCPBUGS-77557


enxebre:
how about karpenter-aws and aws-node-termination-handler?

enxebre:
>The AWS SDK (both v1 and v2) reads AWS_CA_BUNDLE and appends those CAs to the system cert pool.

That statement seems to contradict the docs https://docs.aws.amazon.com/sdk-for-go/api/aws/session/
"Custom CA Bundle" section

"Path to a custom Credentials Authority (CA) bundle PEM file that the SDK will use instead of the default system's root CA bundle. Use this only if you want to replace the CA bundle the SDK uses for TLS requests."

sdminonne:
@coderabbitai review

sdminonne:
@enxebre PTAL

sdminonne:
@enxebre PTAL

bryan-cox:
@sdminonne are you still looking to take this PR forward?

control-plane-operator/controllers/hostedcontrolplane/v2/awsnodeterminationhandler/deployment_test.go
```diff
@@ -181,3 +192,91 @@ func TestGetTerminationHandlerQueueURL(t *testing.T) {
...
+						Type: hyperv1.AWSPlatform,
+						AWS: &hyperv1.AWSPlatformSpec{
+							Region: "us-east-1",
+						},
+					},
+					AdditionalTrustBundle: tc.additionalTrust,
+				},
+			}
+
+			cpContext := controlplanecomponent.WorkloadContext{
```
bryan-cox, line 227: Could you pull this block outside the loop instead of making it over and over?

control-plane-operator/controllers/hostedcontrolplane/v2/awsnodeterminationhandler/deployment.go
```diff
@@ -62,6 +62,10 @@ func adaptDeployment(cpContext component.WorkloadContext, deployment *appsv1.Dep
 		}
 	})
 
+	if hcp.Spec.AdditionalTrustBundle != nil {
+		podspec.DeploymentAddAWSCABundleVolume(hcp.Spec.AdditionalTrustBundle, deployment, cpContext.ReleaseImageProvider.GetImage(podspec.CPOImageName))
```
bryan-cox, line 66: I thought the cpContext had a copy of the hcp you could pass in. Do you know if that is true?

bryan-cox: Yeah you are actually using it here https://github.com/openshift/hypershift/pull/7907/changes#diff-db92db45ba677a590c2e7e4cd186a414d3496b6f62954e4f98218c44518e24cbR25. So maybe the HCP should be referenced from that.

support/podspec/volumes.go
```diff
@@ -32,6 +34,89 @@ func DeploymentAddTrustBundleVolume(trustBundleConfigMap *corev1.LocalObjectRefe
...
+// system CA bundle when AWS_CA_BUNDLE is set, rather than appending to it.
+//
+// This helper is used by both external third-party binaries (karpenter, aws-cloud-controller-manager,
+// aws-node-termination-handler) and internal HyperShift binaries (karpenter-operator, capi-provider,
+// ingress-operator). Even though internal binaries could handle CA loading in Go code, using the
+// same environment-variable-based approach keeps the fix uniform across all AWS components.
+//
+// The initContainerImage should be a RHEL-based image that has /bin/sh and cat available
+// (e.g. the control-plane-operator image).
+func DeploymentAddAWSCABundleVolume(trustBundleConfigMap *corev1.LocalObjectReference, deployment *appsv1.Deployment, initContainerImage string) {
```
bryan-cox: Could you just pass the cpContext in and drop the number of parameters down to two?

control-plane-operator/controllers/hostedcontrolplane/v2/cloud_controller_manager/aws/deployment_test.go
```diff
@@ -0,0 +1,108 @@
...
+			additionalTrust: &corev1.LocalObjectReference{Name: "user-ca-bundle"},
+			expectCABundle:  true,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			g := NewGomegaWithT(t)
+
+			hcp := &hyperv1.HostedControlPlane{
```
bryan-cox: could you move this block outside the for loop and then just change the AdditionalTrustBundle each time?

control-plane-operator/controllers/hostedcontrolplane/v2/cloud_controller_manager/aws/deployment_test.go
```diff
@@ -0,0 +1,108 @@
...
+				},
+				Spec: hyperv1.HostedControlPlaneSpec{
+					Platform: hyperv1.PlatformSpec{
+						Type: hyperv1.AWSPlatform,
+					},
+					AdditionalTrustBundle: tc.additionalTrust,
+				},
+			}
+
+			cpContext := controlplanecomponent.WorkloadContext{
```
bryan-cox, line 53: Could this be moved outside the for loop instead of being recreated each time?

control-plane-operator/controllers/hostedcontrolplane/v2/ingressoperator/deployment_test.go
```diff
@@ -77,3 +77,107 @@ func TestAdaptDeployment(t *testing.T) {
...
+			additionalTrust: &corev1.LocalObjectReference{Name: "user-ca-bundle"},
+			expectCABundle:  false,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			g := NewWithT(t)
+
+			hcp := &hyperv1.HostedControlPlane{
```
bryan-cox: similar comment as the other tests.

control-plane-operator/controllers/hostedcontrolplane/v2/ingressoperator/deployment_test.go
```diff
@@ -77,3 +77,107 @@ func TestAdaptDeployment(t *testing.T) {
...
+				},
+				Spec: hyperv1.HostedControlPlaneSpec{
+					Platform: hyperv1.PlatformSpec{
+						Type: tc.platformType,
+					},
+					AdditionalTrustBundle: tc.additionalTrust,
+				},
+			}
+
+			cpContext := component.WorkloadContext{
```
bryan-cox, line 123: similar comment as the other tests

control-plane-operator/controllers/hostedcontrolplane/v2/kas/deployment.go
```diff
@@ -130,6 +130,10 @@ func adaptDeployment(cpContext component.WorkloadContext, deployment *appsv1.Dep
 			if err := applyKMSConfig(&deployment.Spec.Template.Spec, secretEncryption, newKMSImages(hcp), hcp); err != nil {
 				return err
 			}
+			if secretEncryption.KMS != nil && secretEncryption.KMS.Provider == hyperv1.AWS && hcp.Spec.AdditionalTrustBundle != nil {
```
bryan-cox, line 133: Why would this not be in the block on L113?

control-plane-operator/controllers/hostedcontrolplane/v2/kas/deployment.go
```diff
@@ -164,6 +168,32 @@ func adaptDeployment(cpContext component.WorkloadContext, deployment *appsv1.Dep
...
+			ReadOnly:  true,
+		})
+		c.Env = append(c.Env, corev1.EnvVar{
+			Name:  "AWS_CA_BUNDLE",
+			Value: combinedCAMountPath + "/" + combinedCAFileName,
+		})
+	}
+
+	podspec.UpdateContainer("aws-kms-active", podSpec.Containers, wireCABundle)
+	podspec.UpdateContainer("aws-kms-backup", podSpec.Containers, wireCABundle)
```
bryan-cox: Do we always deploy the backup container if there is not a key set? 🤔 

If so, this would trip things up if there was no container I think.

support/podspec/volumes.go
```diff
@@ -32,6 +34,89 @@ func DeploymentAddTrustBundleVolume(trustBundleConfigMap *corev1.LocalObjectRefe
 	})
 }
 
+// DeploymentAddAWSCABundleVolume creates a combined CA bundle containing both the system CAs from
```
bryan-cox: Can we move this out to its own AWS platform file? This file to date has been platform agnostic.

test/e2e/nodepool_additionalTrustBundlePropagation_test.go
bryan-cox, line 1: Is this something that could be moved over to v2 e2e rather than adding new tests to v1 e2e?

support/podspec/volumes.go
```diff
@@ -32,6 +34,89 @@ func DeploymentAddTrustBundleVolume(trustBundleConfigMap *corev1.LocalObjectRefe
...
+// aws-node-termination-handler) and internal HyperShift binaries (karpenter-operator, capi-provider,
+// ingress-operator). Even though internal binaries could handle CA loading in Go code, using the
+// same environment-variable-based approach keeps the fix uniform across all AWS components.
+//
+// The initContainerImage should be a RHEL-based image that has /bin/sh and cat available
+// (e.g. the control-plane-operator image).
+func DeploymentAddAWSCABundleVolume(trustBundleConfigMap *corev1.LocalObjectReference, deployment *appsv1.Deployment, initContainerImage string) {
+	const (
+		userCAVolumeName     = "user-ca-bundle"
+		combinedCAVolumeName = "aws-ca-bundle"
```
bryan-cox: The volume name, mount path, and filename constants are duplicated in `kas/deployment.go:applyAWSCABundleToKMSContainers`. If either copy drifts, KMS sidecars silently break in isolated environments. Could you export the shared constants?

```go
const (
    AWSCABundleVolumeName = "aws-ca-bundle"
    AWSCABundleMountPath  = "/etc/pki/ca-trust/extracted/hypershift"
    AWSCABundleFileName   = "combined-ca-bundle.pem"
)
```

control-plane-operator/controllers/hostedcontrolplane/v2/kas/deployment.go
```diff
@@ -130,6 +130,10 @@ func adaptDeployment(cpContext component.WorkloadContext, deployment *appsv1.Dep
 			if err := applyKMSConfig(&deployment.Spec.Template.Spec, secretEncryption, newKMSImages(hcp), hcp); err != nil {
 				return err
 			}
+			if secretEncryption.KMS != nil && secretEncryption.KMS.Provider == hyperv1.AWS && hcp.Spec.AdditionalTrustBundle != nil {
+				podspec.DeploymentAddAWSCABundleVolume(hcp.Spec.AdditionalTrustBundle, deployment, cpContext.ReleaseImageProvider.GetImage(podspec.CPOImageName))
```
bryan-cox: `DeploymentAddAWSCABundleVolume` sets `AWS_CA_BUNDLE` on `Containers[0]`, which is `kube-apiserver` here. KAS doesn't use the AWS SDK — only the KMS sidecars do, and those are correctly wired by `applyAWSCABundleToKMSContainers` below. Could we split the helper so the volume/init-container setup is separate from the `Containers[0]` env var wiring? That way KAS only gets the volumes + init container, and the KMS sidecars get the env var via `applyAWSCABundleToKMSContainers`.

control-plane-operator/controllers/hostedcontrolplane/v2/awsnodeterminationhandler/deployment_test.go
```diff
@@ -9,9 +9,20 @@ import (
 	assets "github.com/openshift/hypershift/control-plane-operator/controllers/hostedcontrolplane/v2/assets"
 	controlplanecomponent "github.com/openshift/hypershift/support/controlplane-component"
 
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+type fakeReleaseProvider struct{}
```
bryan-cox: This struct is duplicated identically in 7 test files in this PR. There's already a shared `support/releaseinfo/fake.FakeReleaseProvider` used across 40+ tests in the codebase — could you use that instead?

control-plane-operator/controllers/hostedcontrolplane/v2/capi_provider/deployment.go
```diff
@@ -20,6 +22,10 @@ func (capi *CAPIProviderOptions) adaptDeployment(cpContext component.WorkloadCon
 
 	proxy.SetEnvVars(&deployment.Spec.Template.Spec.Containers[0].Env)
 
+	if cpContext.HCP.Spec.Platform.Type == hyperv1.AWSPlatform && cpContext.HCP.Spec.AdditionalTrustBundle != nil {
```
bryan-cox: nit: The other files in this PR (`awsnodeterminationhandler`, `karpenter`, `kas`) create a local `hcp := cpContext.HCP` and use `hcp.Spec.*`. For consistency, consider doing the same here and in `ingressoperator/deployment.go`.

support/podspec/volumes.go
```diff
@@ -32,6 +34,89 @@ func DeploymentAddTrustBundleVolume(trustBundleConfigMap *corev1.LocalObjectRefe
...
+		VolumeSource: corev1.VolumeSource{
+			EmptyDir: &corev1.EmptyDirVolumeSource{},
+		},
+	})
+
+	// Init container concatenates system CAs with user CAs into the combined bundle.
+	deployment.Spec.Template.Spec.InitContainers = append(deployment.Spec.Template.Spec.InitContainers, corev1.Container{
+		Name:  initContainerName,
+		Image: initContainerImage,
+		Command: []string{"/bin/sh", "-c",
```
bryan-cox: nit: If the system CA file is missing for some reason, this crashes the init container. A defensive fallback is cheap:
```sh
cat /etc/pki/tls/certs/ca-bundle.crt /user-ca/user-ca-bundle.pem > /etc/pki/ca-trust/extracted/hypershift/combined-ca-bundle.pem 2>/dev/null || cp /user-ca/user-ca-bundle.pem /etc/pki/ca-trust/extracted/hypershift/combined-ca-bundle.pem
```

bryan-cox (commented):
Thanks for the PR — the init container approach for combining system + user CAs is the right design given the AWS SDK's `AWS_CA_BUNDLE` replacement behavior. A few items to address:

support/podspec/volumes.go
sdminonne: Done — exported `AWSCABundleVolumeName`, `AWSCABundleMountPath`, and `AWSCABundleFileName` as shared constants. Both `DeploymentAddAWSCABundleVolume` and `applyAWSCABundleToKMSContainers` now use them.

control-plane-operator/controllers/hostedcontrolplane/v2/kas/deployment.go
sdminonne: Done — split the helper into `DeploymentAddAWSCABundleSetup` (volumes + init container only) and `ContainerAddAWSCABundle` (volume mount + env var). KAS now calls `DeploymentAddAWSCABundleSetup` and `applyAWSCABundleToKMSContainers` uses `podspec.ContainerAddAWSCABundle` to wire only the KMS sidecars. `DeploymentAddAWSCABundleVolume` is kept as a convenience wrapper for non-KAS callers (calls both functions for Containers[0]).

control-plane-operator/controllers/hostedcontrolplane/v2/awsnodeterminationhandler/deployment_test.go
sdminonne: Done — replaced all 6 copies with `testutil.FakeImageProvider()` from `support/testutil`.

support/podspec/volumes.go
sdminonne: `support/podspec` is a low-level utility package. `WorkloadContext` lives in `support/controlplane-component`, which imports `support/podspec`. Passing `cpContext` here would create a circular dependency (or at minimum a bad layering inversion). The current signature `(trustBundleConfigMap, deployment, initContainerImage)` keeps the package self-contained — all parameters are plain k8s types.

support/podspec/volumes.go
sdminonne: The existing `DeploymentAddTrustBundleVolume` in this file is also trust-bundle-specific (not truly platform-agnostic). The two functions share the same concern: trust bundle volume wiring. The file is ~140 lines — splitting one function into a separate file would scatter related code. Happy to revisit if the package grows with more platform-specific helpers, but for now I'd prefer keeping them together.

control-plane-operator/controllers/hostedcontrolplane/v2/kas/deployment.go
sdminonne: `podspec.UpdateContainer` is a no-op when the named container doesn't exist — it just iterates and matches by name (`containers.go:55-60`). So calling it on `"aws-kms-backup"` when there's no backup container is safe; it silently skips.

control-plane-operator/controllers/hostedcontrolplane/v2/kas/deployment.go
sdminonne: The platform switch at L113 handles pod identity webhooks — that's its concern. The KMS/secret-encryption block at L126+ is a separate concern with its own guard (`secretEncryption.KMS.Provider == hyperv1.AWS`). Nesting KMS handling inside the platform switch would conflate two independent concerns. The current placement correctly treats secret encryption as orthogonal to platform identity webhooks.

control-plane-operator/controllers/hostedcontrolplane/v2/awsnodeterminationhandler/deployment.go
sdminonne: Good point — updated to use `cpContext.HCP` consistently.

control-plane-operator/controllers/hostedcontrolplane/v2/awsnodeterminationhandler/deployment_test.go
sdminonne: Ack — will hoist the HCP/cpContext setup outside the loop.

control-plane-operator/controllers/hostedcontrolplane/v2/cloud_controller_manager/aws/deployment_test.go
sdminonne: Ack — will hoist outside the loop.

control-plane-operator/controllers/hostedcontrolplane/v2/cloud_controller_manager/aws/deployment_test.go
sdminonne: Ack.

control-plane-operator/controllers/hostedcontrolplane/v2/ingressoperator/deployment_test.go
sdminonne: Ack.

control-plane-operator/controllers/hostedcontrolplane/v2/ingressoperator/deployment_test.go
sdminonne: Ack.

control-plane-operator/controllers/hostedcontrolplane/v2/capi_provider/deployment.go
sdminonne: Ack — will add `hcp := cpContext.HCP` for consistency.

support/podspec/volumes.go
sdminonne: Ack — will add the defensive fallback.

test/e2e/nodepool_additionalTrustBundlePropagation_test.go
sdminonne: Will look into whether the v2 e2e framework supports what this test needs. If feasible, will migrate; otherwise will track as a follow-up.

support/podspec/volumes.go
bryan-cox: `DeploymentAddTrustBundleVolume` is platform-agnostic — it mounts a ConfigMap as a trusted-ca volume for any platform. The new functions are explicitly AWS-specific (`AWS_CA_BUNDLE`, AWS SDK concatenation strategy). The file went from ~35 lines of generic helpers to ~140 with ~90 lines of AWS-only code. I'd still prefer a `volumes_aws.go` to keep the separation clean.

control-plane-operator/controllers/hostedcontrolplane/v2/awsnodeterminationhandler/deployment_test.go
bryan-cox: This still looks like it's inside the loop. You can build the base HCP once outside and just set `hcp.Spec.AdditionalTrustBundle = tc.additionalTrust` inside each iteration.

control-plane-operator/controllers/hostedcontrolplane/v2/cloud_controller_manager/aws/deployment_test.go
bryan-cox: Same as above — still inside the loop.

control-plane-operator/controllers/hostedcontrolplane/v2/cloud_controller_manager/aws/deployment_test.go
bryan-cox: Same as above — still inside the loop.

control-plane-operator/controllers/hostedcontrolplane/v2/ingressoperator/deployment_test.go
bryan-cox: Same as above — still inside the loop.

control-plane-operator/controllers/hostedcontrolplane/v2/ingressoperator/deployment_test.go
bryan-cox: Same as above — still inside the loop.

control-plane-operator/controllers/hostedcontrolplane/v2/capi_provider/deployment.go
bryan-cox: `capi_provider` and `ingressoperator` still reference `cpContext.HCP.Spec.*` directly — the `hcp := cpContext.HCP` local isn't added yet.

support/podspec/volumes.go
bryan-cox: The defensive fallback still isn't here. If the system CA file is ever missing, this init container crashes and blocks pod startup in an isolated environment — exactly the environment this PR is designed to support. Cheap insurance.

test/e2e/nodepool_additionalTrustBundlePropagation_test.go
sdminonne: Done — removed the v1 test file and added the AWS_CA_BUNDLE wiring checks (add + remove) to the existing `NodePoolTrustBundleTest` in `test/e2e/v2/tests/nodepool_lifecycle_test.go`. Also switched from `Containers[0]` indexing to `podspec.FindContainer` and used the exported `podspec.AWSCABundleVolumeName`/`AWSCABundleMountPath`/`AWSCABundleFileName` constants.

bryan-cox:
Two minor nits from the latest push:

1. **karpenter and karpenteroperator tests**: The HCP object is still constructed inside the `for` loop in `karpenter/deployment_test.go` and `karpenteroperator/deployment_test.go`. The other tests (awsnodeterminationhandler, cloud_controller_manager/aws, ingressoperator) were updated to hoist it outside the loop — would be good to make these consistent.

2. **cloud_controller_manager/aws/deployment.go**: This file still uses `cpContext.HCP.Spec.*` directly instead of declaring a local `hcp := cpContext.HCP` like `capi_provider/deployment.go` and `ingressoperator/deployment.go` do. Minor, but would keep the pattern consistent.

bryan-cox:
Could you squash the commits? Otherwise lgtm

control-plane-operator/controllers/hostedcontrolplane/v2/capi_provider/deployment_test.go
```diff
@@ -231,3 +232,120 @@ func TestAdaptDeployment_WithNilAnnotations(t *testing.T) {
...
+			additionalTrust: &corev1.LocalObjectReference{Name: "user-ca-bundle"},
+			expectCABundle:  false,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			g := NewWithT(t)
+
+			hcp := &hyperv1.HostedControlPlane{
```
bryan-cox, line 267: This could be outside the look instead of recreating it each iteration.

control-plane-operator/controllers/hostedcontrolplane/v2/cloud_controller_manager/aws/deployment_test.go
bryan-cox: Still seems valid

control-plane-operator/controllers/hostedcontrolplane/v2/ingressoperator/deployment_test.go
bryan-cox: Comment still seems valid

test/e2e/nodepool_additionalTrustBundlePropagation_test.go
bryan-cox, line 1: why is this test being removed?

test/e2e/nodepool_test.go
bryan-cox, line 1: Why is this being removed?

support/podspec/volumes_aws.go
```diff
@@ -0,0 +1,111 @@
...
+				corev1.ResourceCPU:    resource.MustParse("10m"),
+				corev1.ResourceMemory: resource.MustParse("10Mi"),
+			},
+		},
+		SecurityContext: &corev1.SecurityContext{
+			AllowPrivilegeEscalation: ptr.To(false),
+			Capabilities: &corev1.Capabilities{
+				Drop: []corev1.Capability{"ALL"},
+			},
+		},
```
bryan-cox, line 78: nit: The init container sets `AllowPrivilegeEscalation: false` and drops all capabilities, but does not set `RunAsNonRoot: true`. The rest of the codebase enforces this via `enforceRestrictedSecurityContext` — this should be consistent.

support/podspec/volumes_aws.go
```diff
@@ -0,0 +1,111 @@
...
+// (e.g. the control-plane-operator image).
+func DeploymentAddAWSCABundleSetup(trustBundleConfigMap *corev1.LocalObjectReference, deployment *appsv1.Deployment, initContainerImage string) {
+	// Volume for user CAs from additionalTrustBundle ConfigMap.
+	deployment.Spec.Template.Spec.Volumes = append(deployment.Spec.Template.Spec.Volumes, corev1.Volume{
+		Name: userCAVolumeName,
+		VolumeSource: corev1.VolumeSource{
+			ConfigMap: &corev1.ConfigMapVolumeSource{
+				LocalObjectReference: *trustBundleConfigMap,
+				Items:                []corev1.KeyToPath{{Key: "ca-bundle.crt", Path: userCAFileName}},
+			},
```
bryan-cox, line 42: The combined CA bundle is produced by the init container at pod startup only. Unlike `DeploymentAddTrustBundleVolume` (which mounts the ConfigMap directly and gets Kubernetes auto-updates), this means a CA rotation requires a pod restart to take effect.

In practice this works because an HCP spec change triggers CPO reconciliation → deployment update → new pods. But if someone updates the ConfigMap contents without touching the HCP spec, pods will not pick up the change. Worth a doc comment on the function noting this behavior.

support/podspec/volumes_aws.go
```diff
@@ -0,0 +1,111 @@
...
+//
+// For deployments where only specific sidecar containers need AWS_CA_BUNDLE (e.g. KAS with
+// KMS sidecars), use DeploymentAddAWSCABundleSetup and ContainerAddAWSCABundle separately.
+//
+// The initContainerImage should be a RHEL-based image that has /bin/sh and cat available
+// (e.g. the control-plane-operator image).
+func DeploymentAddAWSCABundleVolume(trustBundleConfigMap *corev1.LocalObjectReference, deployment *appsv1.Deployment, initContainerImage string) {
+	DeploymentAddAWSCABundleSetup(trustBundleConfigMap, deployment, initContainerImage)
+	ContainerAddAWSCABundle(&deployment.Spec.Template.Spec.Containers[0])
+}
```
bryan-cox, line 111: nit: The doc comment says "wires it into Containers[0]" which is descriptive — consider adding a sentence like "Callers must ensure Containers[0] is the AWS SDK-using container." All current callers satisfy this, but it is a foot-gun for future use.

support/podspec/volumes_test.go
```diff
@@ -0,0 +1,150 @@
...
+			g.Expect(addedMount.ReadOnly).To(BeTrue())
+
+			// It should set AWS_CA_BUNDLE env var on the main container.
+			g.Expect(spec.Containers[0].Env).To(HaveLen(existingEnvCount + 1))
+			addedEnv := spec.Containers[0].Env[len(spec.Containers[0].Env)-1]
+			g.Expect(addedEnv.Name).To(Equal("AWS_CA_BUNDLE"))
+			g.Expect(addedEnv.Value).To(Equal("/etc/pki/ca-trust/extracted/hypershift/combined-ca-bundle.pem"))
+		})
+	}
+}
```
bryan-cox, line 150: `DeploymentAddAWSCABundleSetup` and `ContainerAddAWSCABundle` are only tested through the combined `DeploymentAddAWSCABundleVolume` wrapper. Since KAS uses them separately, consider adding isolated tests verifying:
- `DeploymentAddAWSCABundleSetup` adds volumes + init container but does NOT touch any container env vars or mounts
- `ContainerAddAWSCABundle` adds only the mount and env var to the specified container

control-plane-operator/controllers/hostedcontrolplane/v2/awsnodeterminationhandler/deployment.go
```diff
@@ -62,6 +62,10 @@ func adaptDeployment(cpContext component.WorkloadContext, deployment *appsv1.Dep
 		}
 	})
 
+	if hcp.Spec.AdditionalTrustBundle != nil {
```
bryan-cox, line 65: nit: This component is already gated to AWS by its predicate, so the missing platform check is correct. But a one-line comment like `// No platform check needed — component predicate restricts to AWS` would prevent future reviewers from flagging it as a bug. Same applies to `karpenter/deployment.go` and `cloud_controller_manager/aws/deployment.go`.

control-plane-operator/controllers/hostedcontrolplane/v2/awsnodeterminationhandler/deployment_test.go
```diff
@@ -181,3 +183,92 @@ func TestGetTerminationHandlerQueueURL(t *testing.T) {
...
+			} else {
+				g.Expect(volumes).ToNot(ContainElement(HaveField("Name", "user-ca-bundle")))
+				g.Expect(volumes).ToNot(ContainElement(HaveField("Name", "aws-ca-bundle")))
+				g.Expect(initContainers).ToNot(ContainElement(HaveField("Name", "setup-aws-ca-bundle")))
+				g.Expect(container.VolumeMounts).ToNot(ContainElement(HaveField("Name", "aws-ca-bundle")))
+				g.Expect(container.Env).ToNot(ContainElement(HaveField("Name", "AWS_CA_BUNDLE")))
+			}
+		})
+	}
+}
```
bryan-cox, line 274: The ~25-line CA bundle verification block (check volumes, init containers, mounts, env vars) is copy-pasted across 7 test files in this PR. Consider extracting a shared helper like `testutil.AssertAWSCABundleWiring(g, deployment, expectPresent)`. This is separate from the struct-hoisting feedback — the assertion logic itself is the duplication.

bryan-cox (commented):
Thanks for the thorough work — the init container concatenation approach is the right design given `AWS_CA_BUNDLE`'s replacement semantics, and the `Setup`/`ContainerAdd` split for KAS is clean.

Main item: the PR needs a rebase — `applyKMSConfig` signature has changed on main.

Beyond that, the suggestions below are mostly about documentation and test coverage for the split helpers. The fix itself is correct and the component coverage is complete.

PR #7908, opened 2026-03-10
CNTRLPLANE-3145: refactor(hostedcluster): segregate reconcile loop into error-collecting blocks

muraee:
## Summary

- Refactors `reconcile()` in the HostedCluster controller to use **categorized error handling** with critical/non-critical operations instead of sequential short-circuiting. Previously, any single failure among ~50 operations would block all subsequent work — e.g., a missing SSH key secret prevented CPO deployment and HCP creation.
- Introduces a `reconcileReport` struct (`reconcile_report.go`) that classifies operations as **critical** (blocks downstream Phase 8) or **nonCritical** (errors collected, never blocks). When critical operations fail, Phase 8 components are automatically skipped with clear reporting of what failed and what was blocked.
- Extracts inline code blocks into named methods and introduces wrapper methods (`reconcileOperatorDeployments`, `reconcileRBACAndPolicies`, `reconcileKubeconfigAndPasswordSync`, `reconcileAuxiliary`, `reconcilePlatformSpecific`) that collect errors independently.
- The `ReconciliationSucceeded` condition now reflects the structured report: when critical failures exist, the condition message surfaces which operations failed and which were blocked (e.g., `critical failures: [PullSecretSync]; blocked operations: [OperatorDeployments, RBACAndPolicies, ...]`).

### Error categorization

| Category | Behavior | Operations |
|----------|----------|------------|
| critical | Failures block Phase 8 components | PlatformCredentials, PullSecretSync, SecretEncryptionSync, CoreHCPChain |
| nonCritical | Errors collected, never blocks | SSHKeySync, AuditWebhookSync, AdditionalTrustBundle, all Phase 8 groups |

### Phase structure

| Phase | Behavior | Operations |
|-------|----------|------------|
| 0–5 | Short-circuit (prerequisites) | HCP get, deletion, platform defaults, status, finalizers, namespace, platform |
| 6a | Critical sync (error-collecting) | PlatformCredentials, PullSecretSync, SecretEncryptionSync |
| 6b | Non-critical sync (error-collecting, never blocked) | RestoredFromBackup, AuditWebhookSync, SSHKeySync, AdditionalTrustBundle, SA signing key, etcd MTLS, ETCDMemberRecovery, GlobalConfigSync |
| 7 | Core HCP chain (always runs regardless of 6a) | HCP object → CAPI InfraCR → CAPI Cluster |
| 8 | Components — **blocked if any critical failure** | KubeconfigAndPasswordSync, OperatorDeployments, RBACAndPolicies, PlatformOIDCAndCSI, MonitoringAndCLISecrets |

### Condition reporting

The `ReconciliationSucceeded` condition now reflects the structured error report:
- When critical failures exist, the condition message includes which operations failed and which were blocked
- When only non-critical failures exist, the condition reports the aggregate error as before
- Example condition message: `critical failures: [PullSecretSync]; blocked operations: [KubeconfigAndPasswordSync, OperatorDeployments, RBACAndPolicies, PlatformOIDCAndCSI, MonitoringAndCLISecrets]`

### Structured error aggregation

When critical failures exist, `aggregate()` returns only critical errors with blocked operation list — non-critical errors are suppressed since the user should fix the critical issue first:
```
critical error: failed to get pull secret...; blocked operations: [KubeconfigAndPasswordSync, OperatorDeployments, RBACAndPolicies, PlatformOIDCAndCSI, MonitoringAndCLISecrets]
```
When no critical failures exist, all errors are returned as-is.

### reconcileReport API

Two public methods on the report:
- `execute(name, category, func() error)` — always runs the operation and records the result
- `executeOrBlock(name, func() error)` — automatically checks `hasCriticalFailure()` and either runs the operation or records it as blocked

## Analysis

See `docs/design/hostedcluster-reconcile-segregation-analysis.md` for the full design.

## Test plan

- [x] All existing unit tests pass (`go test -count=1 -race ./hypershift-operator/controllers/hostedcluster/`)
- [x] `make lint` passes with 0 issues
- [x] New unit tests for `reconcileReport` methods (`TestReconcileReport`, `TestConditionMessage`, `TestAggregate`, `TestExecuteOrBlock`)
- [x] New unit tests for wrapper method isolation (`TestReconcileKubeconfigAndPasswordSync_*`, `TestReconcileRBACAndPolicies_*`)
- [x] New integration tests verifying blocking behavior:
  - Phase 6a critical failure → Phase 8 blocked, Phase 7 still runs
  - Phase 7 HCP creation failure → Phase 8 blocked
  - Phase 6b non-critical failure → nothing blocked


docs/design/hostedcluster-reconcile-segregation-analysis.md
```diff
@@ -0,0 +1,231 @@
...
+
+## Impact Summary
+
+**Before**: An error in any of ~50 operations blocks all subsequent operations. Examples:
+- SSH key secret missing -> CPO never deployed -> control plane never starts
+- Monitoring dashboard error -> karpenter operator never reconciled
+- AWS resource tags error -> HCP never updated
+
+**After (proposed)**: Only genuinely dependent operations block each other. Examples:
+- SSH key secret missing -> SSH key not synced (reported via condition), but CPO still
```
enxebre: As of today, I see ssh as dependency for the dataplane, as changing it would trigger a rollout, is this still the case?

docs/design/hostedcluster-reconcile-segregation-analysis.md
```diff
@@ -0,0 +1,231 @@
...
+  deploys, HCP still reconciles, CAPI still works
+- Monitoring dashboard error -> reported, but karpenter and network policies proceed normally
```
enxebre: can we include more examples and have test cases for all of them?

enxebre:
was there jira bug we can ref reporting the scenario where this was being problematic for managed?

docs/design/hostedcluster-reconcile-segregation-analysis.md
```diff
@@ -0,0 +1,231 @@
...
+  | R34: OpenShift trusted CAs                       |
+  |                                                  |
+  | *** LARGELY INDEPENDENT ***                      |
+  +-------------------------------------------------+
+```
+
+## Key Problems Identified
+
+1. **Secret sync operations block each other** (R13-R21): A missing SSH key secret prevents
+   pull secret sync, secret encryption setup, ETCD MTLS, and global config sync. These are
```
enxebre: why would we want to proceed with the CPO if the pull secret is invalid?

muraee: two reasons:
1. image could be already available in the cluster so an invalid pull-secret doesn't necessarily block the deployment 
2. the imagePull error would be reported in the CPO deployment making it easier to discover than going through the HO logs

docs/design/hostedcluster-reconcile-segregation-analysis.md
muraee: no behavior change, this will only make other controlPlane operations that are not affected by the missing ssh-key proceed.

muraee: this should help decopuling data-plane/control-plane upgrades we are working on. Otherwise, control-plane upgrade will not proceed if ssh-key (data-plane input) is missing

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -1389,552 +1388,812 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
...
+		r.reconcileRBACAndPolicies(ctx, log, createOrUpdate, hcluster, hcp,
+			controlPlanePKIOperatorSignsCSRs, controlPlaneOperatorAppliesManagementKASNetworkPolicyLabel,
+			releaseImageVersion)...)
+
+	if err := r.reconcilePlatformSpecific(ctx, log, hcluster, hcp, createOrUpdate); err != nil {
+		componentErrs = append(componentErrs, fmt.Errorf("failed to reconcile platform specific items: %w", err))
+	}
+
+	componentErrs = append(componentErrs,
+		r.reconcileAuxiliary(ctx, createOrUpdate, hcluster, hcp)...)
```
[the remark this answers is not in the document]
muraee: done

docs/design/hostedcluster-reconcile-segregation-analysis.md
muraee: done

docs/design/hostedcluster-reconcile-segregation-analysis.md
enxebre: >the imagePull error would be reported in the CPO deployment making it easier to discover than going through the HO logs

there should be a signal in the HC status or this is a UX bug.

enxebre:
>Previously, any single failure among ~50 operations would block all subsequent work — e.g., a missing SSH key secret prevented CPO deployment and HCP creation.

This is arguably the right behaviour for things that are API input. The issue is a UX one, that we don't fail on admission. Can we mitigate this with a validating policy?

I can think of the following guardrails:
- (Invalid input) - Invalid Input should ideally block on admission (CEL Or validating admission policy), otherwise we can aggregate errors of that same category but they should block further reconciliation on invalid config condition.
- (Critical runtime / Non-critical runtime) - Run time reconciling Errors should not necessarily block further reconciliation, i.e. we can aggregate when makes sense. But what we consider "critical" should block. "Critical" is something that we know prevents the HC form becoming operational and might produce a cascade failure storm, e.g. failing to reconcile the pull secret into the cp namespace.
- Deletion is never blocked 

So may be can restructure this PR to present buckets as Critical runtime / Non-critical runtime / Invalid input / Deletion. I think this mostly applies to your phases 6, 7, 8 but all of them should be compliant.

tuxerrante:
> Previously, any single failure among ~50 operations would block all subsequent work — e.g., a missing SSH key secret prevented CPO deployment and HCP creation.

> This is arguably the right behaviour for things that are API input. [...] So may be can restructure this PR to present buckets as Critical runtime / Non-critical runtime / Invalid input / Deletion.

Expanding on the Critical / Non-critical / Invalid-input / Deletion buckets — would it be feasible to also distinguish between **creation-time** and **steady-state** reconcile behavior?

During initial creation, the sequential fail-fast in Phase 7 ([`reconcileCoreHCPChain`](https://github.com/openshift/hypershift/pull/7908/changes#diff-b5cde720842b19b53da3f454d827e27327c4559b2e4849906dac728586f51eb3R1487)) is the right call — you can't deploy CPO against a non-existent HCP, and proceeding would just create noise. But once `hcp.Status.Phase == Available`, most Phase 8 components are idempotent no-ops that only verify existing state. Running them concurrently in steady state could reduce reconcile latency on dense management clusters (500+ HCs, 10 workers, every second saved compounds).

In terms of classification, the critical-path set is small enough to enumerate explicitly rather than infer:

```go
type ReconcileCategory int

const (
    CriticalRuntime    ReconcileCategory = iota // fail-fast, blocks everything
    NonCriticalRuntime                          // error-collecting, independent
    InvalidInput                                // aggregate but block on condition
    Deletion                                    // never blocked
)
```

The critical spine — without any of these the HC is non-functional:

| Category | Components | Rationale |
|----------|-----------|-----------|
| `CriticalRuntime` | `reconcileCoreHCPChain` (Phase 7: HCP → InfraCR → CAPI Cluster), CPO deployment, platform credentials (R11), pull secret sync (R13) | Hard dependencies — downstream components crash-loop or nil-pointer without these |
| `NonCriticalRuntime` | SSH key sync, monitoring dashboard, PKI RBAC, network policies, karpenter, kubeconfig/password sync | Can fail independently without making the control plane non-operational |
| `InvalidInput` | Validation gates (R7), unmanaged etcd validation (S15) | Should ideally be caught at admission via CEL / `ValidatingAdmissionPolicy`; when not, aggregate errors but set a blocking condition |
| `Deletion` | Phase 2 | Never blocked, as-is |

The Phase 7 → Phase 8 boundary in this PR already captures this partially — Phase 7 is sequential, Phase 8 is error-collecting. The refinement would be: within Phase 8, run the non-critical blocks **concurrently** (not just error-collecting but actually parallel via [`errgroup`](https://pkg.go.dev/golang.org/x/sync/errgroup)) when the HCP is already `Available`, since at that point the components are checking/updating disjoint resources:

```go
if hcp.Status.Phase == hyperv1.HostedControlPlaneAvailable {
    // Steady state: critical path sequential, rest concurrent
    g, ctx := errgroup.WithContext(ctx)
    for _, fn := range []func() error{
        func() error { return r.reconcileKubeconfigAndPasswordSync(ctx, ...) },
        func() error { return r.reconcileRBACAndPolicies(ctx, ...) },
        func() error { return r.reconcileOperatorDeployments(ctx, ...) },
        func() error { return r.reconcileAuxiliaryAndMonitoring(ctx, ...) },
    } {
        g.Go(fn)
    }
    componentErrs = append(componentErrs, g.Wait())
} else {
    // Creation: sequential error-collecting (current Phase 8 behavior)
    componentErrs = append(componentErrs, r.reconcileKubeconfigAndPasswordSync(ctx, ...)...)
    // ...
}
```

**Question:** has there been any measurement of reconcile latency on a dense cluster (e.g., 200+ HCs on one MC) comparing the current sequential loop vs this PR's error-collecting phases? That data would help quantify whether concurrent Phase 8 is worth the added complexity, or whether error-collecting without concurrency is already sufficient.

csrwng (commented):
## Ship observability first, then change behavior

The refactoring approach is sound — categorizing operations as critical/non-critical and collecting errors instead of short-circuiting is a clear improvement. However, this PR changes the control flow of the most critical reconciler in the system, and the new failure modes (parallel error paths, reordered operations) are exactly the kind that don't surface in unit tests or standard e2e — they manifest under partial failures in production.

### Step 0: Observability before behavior change

Before changing any error handling or ordering, ship a PR that adds structured logging or metrics to the existing sequential reconcile loop that records:
- Which operations fail and how often
- What downstream operations *would have* continued running under the new model
- Whether the current ordering assumptions actually matter in practice

This "dry-run" data from a production release cycle would validate the critical vs. non-critical categorization with real failure data, rather than guessing which operations are safe to unblock. It also gives us a baseline to compare against after the behavior change lands.

### Then: incremental rollout

This PR bundles three distinct changes into one shot, which creates a large blast radius:

1. **Extracting inline blocks into named methods** (pure refactor)
2. **Introducing `reconcileReport` and wiring it up** (new framework)
3. **Reclassifying operations as non-critical and reordering them** (behavior change)

Suggested split:

**PR 1 — Extract methods (zero behavior change).** Move inline code blocks into named methods, keeping the exact same sequential short-circuit order. This is safe to review, easy to verify (identical behavior), and reduces the diff for subsequent PRs.

**PR 2 — Introduce `reconcileReport`, classify everything as `critical`.** Wire up the report framework but keep all operations as `critical` so behavior is identical to today — every error still blocks downstream work. This validates the framework without changing semantics.

**PR 3+ — Reclassify operations as `nonCritical` one group at a time.** Move SSH key sync, audit webhook, etc. to `nonCritical` incrementally, with production validation between each change. Each PR is small, reviewable, and independently revertable.

### Other notes

- **Ordering changes need per-operation justification**: The PR reorders several operations (e.g., CLI Secrets moved from first to last, RestoredFromBackup shifted relative to pull secret sync). Each change should have a brief rationale explaining why nothing downstream depends on the old position.
- **Feature gate**: Consider a flag (env var or annotation) to switch between old and new reconcile paths during the rollout period, so issues can be mitigated without a rollback.

support/config/constants.go
```diff
@@ -48,7 +48,8 @@ const (
 	EnableCVOManagementClusterMetricsAccessEnvVar = "ENABLE_CVO_MANAGEMENT_CLUSTER_METRICS_ACCESS"
 	CVOPrometheusURLEnvVar                        = "CVO_PROMETHEUS_URL"
 
-	EnableEtcdRecoveryEnvVar = "ENABLE_ETCD_RECOVERY"
+	EnableEtcdRecoveryEnvVar             = "ENABLE_ETCD_RECOVERY"
+	EnableErrorCollectingReconcileEnvVar = "HYPERSHIFT_ERROR_COLLECTING_RECONCILE"
```
enxebre: I would expect we default to this transparently, and we made it easy for consumers i.e. managed services to opt-out if needed

hypershift-operator/controllers/hostedcluster/reconcile_report.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+// shouldBlock returns true if subsequent dependent operations should be skipped.
+// In legacy mode, any failure blocks. Otherwise only critical failures block.
+func (r *reconcileReport) shouldBlock() bool {
+	if r.legacy {
+		return r.hasAnyFailure()
+	}
+	return r.hasCriticalFailure()
+}
```
csrwng, line 76: nit: Consider using `sets.Set[string]` from `k8s.io/apimachinery/pkg/util/sets` instead of `map[string]bool` here and in `blockedNames()`. It would clean up the dedup pattern and `sets.List()` returns sorted results, which would make condition messages deterministic regardless of operation registration order.

csrwng (commented):
Overall this is a well-designed refactoring that addresses a real pain point — sequential short-circuiting in a ~50-operation reconcile loop. The `reconcileReport` abstraction is clean, the error categorization is sensible, the legacy mode flag is a smart migration strategy, and the test coverage (both unit and integration) is solid.

A few nits and one potential bug (orphaned `Status().Update()` in KubeVirt case) below.

**Pre-existing nit (not in diff):** Line 375 has a typo in `"ReconciliatonSucceeded"` — missing an 'i', should be `"ReconciliationSucceeded"`.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -1348,45 +1321,495 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
...
+
+	if r.EnableEtcdRecovery &&
+		hcluster.Spec.Etcd.ManagementType == hyperv1.Managed &&
+		hcluster.Spec.ControllerAvailabilityPolicy == hyperv1.HighlyAvailable {
+		report.execute("ETCDMemberRecovery", nonCritical, func() error {
+			requeueAfter, etcdErr := r.reconcileETCDMemberRecovery(ctx, hcluster, createOrUpdate)
+			if etcdErr != nil {
+				return fmt.Errorf("failed to perform etcd member recovery: %w", etcdErr)
+			}
+			report.requeueAfter = requeueAfter
```
csrwng: nit: Mutating `report.requeueAfter` from inside closures is a subtle side-channel, and the two call sites are inconsistent — this one does an unconditional assignment, while `PublicEndpointExposed` (line 1525) does a conditional minimum. Consider adding a `requestRequeue(d *time.Duration)` method to `reconcileReport` that encapsulates the "take the minimum" logic, so both sites become `report.requestRequeue(d)` and the comparison lives in one place.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -1348,45 +1321,495 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
...
+		if chainErr != nil {
+			return chainErr
+		}
+		if hcp == nil {
+			return fmt.Errorf("HCP object is nil")
+		}
+		return nil
+	})
+
+	report.executeOrBlock("KubeconfigAndPasswordSync", func() error {
```
csrwng, line 1511: nit: The ordering here is load-bearing — `KubeconfigAndPasswordSync`, `PlatformOIDCAndCSI`, and `MonitoringAndCLISecrets` are evaluated before `ReleaseImageVersion` so they can run even when the release image is unavailable. But this isn't obvious from reading the code. Consider adding phase sub-comments to clarify the intent, e.g.:

```
// Phase 8a: Components that don't depend on release image version.
// Evaluated before ReleaseImageVersion so they run even when the
// release image is unavailable.
```

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -1348,45 +1321,495 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
...
+
+	// Persist status updates
+	meta.SetStatusCondition(&hcluster.Status.Conditions, *freshCondition)
+	if err := r.Client.Status().Update(ctx, hcluster); err != nil {
+		return fmt.Errorf("failed to update status %v: %w", string(hyperv1.HostedClusterRestoredFromBackup), err)
+	}
+	return nil
+}
+
+// reconcilePullSecretSync syncs the pull secret from the HostedCluster namespace
```
csrwng: nit: Stale godoc — this says `reconcilePullSecretSync syncs the pull secret...` but the function below is `reconcileControlPlaneNamespace`.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -1968,129 +2258,90 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
...
+// reconcilePlatformSpecific handles platform-specific reconciliation including
+// KubeVirt CSI RBAC, AWS OIDC documents, and Azure SecretProviderClass resources.
+func (r *HostedClusterReconciler) reconcilePlatformSpecific(
+	ctx context.Context, log logr.Logger, hcluster *hyperv1.HostedCluster,
+	hcp *hyperv1.HostedControlPlane, createOrUpdate upsert.CreateOrUpdateFN,
+) error {
 	switch hcluster.Spec.Platform.Type {
 	case hyperv1.KubevirtPlatform:
 		if hcluster.Spec.Platform.Kubevirt != nil && hcluster.Spec.Platform.Kubevirt.Credentials != nil {
 			if err := r.Client.Status().Update(ctx, hcluster); err != nil {
```
csrwng: This `Status().Update()` appears orphaned — there's no visible status mutation preceding it within this method or in the operations that run before it in the new framework ordering. In the original code, it sat after `reconcileNetworkPolicies` and `reconcileControlPlanePKIOperatorRBAC`, but neither of those touch `hcluster.Status`, and the earlier status changes (KubeConfig, Version, CVO conditions) are already flushed at line 1263. This may have been a no-op even before the refactor.

Additionally, the `apierrors.IsConflict` check at line 2312 is now dead code — the original conflict branch returned `ctrl.Result{Requeue: true}, nil` (silent requeue), but the refactoring changed both branches to return the same `fmt.Errorf(...)`. If this status update is still needed, the conflict handling should be restored; if not, the entire block (lines 2310-2317) can be removed.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -1348,45 +1321,495 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
...
+	if autoNodeProgressing {
+		autoNodeRequeue := 15 * time.Second
+		if result.RequeueAfter == 0 || autoNodeRequeue < result.RequeueAfter {
+			result.RequeueAfter = autoNodeRequeue
+		}
+	}
+
+	if errs := report.allErrors(); len(errs) == 0 {
+		log.Info("successfully reconciled")
+	} else if msg := report.conditionMessage(); msg != "" {
```
csrwng: `conditionMessage()` is only used for logging here — the actual `ReconciliationSucceeded` condition message comes from `report.aggregate().Error()` at line 382. The method name is misleading. Consider either:
- Renaming it to something like `logSummary()` to reflect its actual usage, or
- Actually using it for the condition message, which would give a more structured, stable format (operation names vs. raw error text) that's better for monitoring/alerting — while keeping `aggregate().Error()` for detailed logging.

docs/design/hostedcluster-reconcile-segregation-analysis.md
```diff
@@ -0,0 +1,116 @@
...
+)
+
+type reconcileReport struct {
+    results      []operationResult
+    requeueAfter *time.Duration
+}
+```
+
+Key methods:
+- `recordOp(name, category, err)` — record a single operation result
```
csrwng: The method signatures listed here (`recordOp`, `recordBlocked`, `recordErrors`, `executeOrBlockMulti`) don't match the actual implementation in `reconcile_report.go`, which uses `execute()`, `executeOrBlock()`, `hasCriticalFailure()`, `aggregate()`, and `conditionMessage()`. Please update to match the implemented API.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
muraee: Done. Fixed the stale godoc to match `reconcileControlPlaneNamespace`.

---

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
muraee: Done. Removed the orphaned `Status().Update()` block — no status field was mutated before it, and the conflict branch was dead code (both branches returned the same error).

---

docs/design/hostedcluster-reconcile-segregation-analysis.md
muraee: Done. Updated the design doc method signatures to match the implemented API (`execute`, `executeOrBlock`, `hasCriticalFailure`, `aggregate`, `logSummary`).

---

hypershift-operator/controllers/hostedcluster/reconcile_report.go
muraee: Done. Switched `criticalFailureNames()` and `blockedNames()` to use `sets.New[string]()` + `sets.List()` for cleaner dedup and deterministic ordering.

---

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
muraee: Done. Added `requestRequeue(d *time.Duration)` method that encapsulates the min-comparison logic. Both call sites now use it.

---

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
muraee: Done. Added phase sub-comments — Phase 8a (version-independent) and Phase 8b (release-image-dependent) to clarify the load-bearing ordering.

---

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
muraee: Done. Renamed `conditionMessage()` to `logSummary()` to reflect its actual usage (structured logging, not condition message).

---

support/config/constants.go
muraee: Addressed. The env var was inverted — it's now `HYPERSHIFT_RECONCILE_LEGACY=1` (opt-out), so the new behavior is the default. The legacy reconciler is a full copy in `reconcile_legacy.go` for rollback safety.

---

muraee:
@csrwng Thanks for the thorough review! All nits addressed:
- Stale godoc fixed
- Orphaned `Status().Update()` removed
- Design doc method signatures updated
- `sets.Set[string]` for deterministic ordering
- `requestRequeue()` method for consistent requeue logic
- Phase 8a/8b sub-comments for ordering clarity
- `conditionMessage()` → `logSummary()`

Re: the `ReconciliatonSucceeded` typo — that's a pre-existing Reason string that may be consumed by external tooling (OCM, etc.). Changing it could be a breaking change, so leaving it for a separate PR with proper deprecation.

---

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -1348,45 +1323,496 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
...
+	if err != nil {
+		return hcp, fmt.Errorf("failed to determine if AWS node termination handler is needed: %w", err)
+	}
+
+	hcp = controlplaneoperator.HostedControlPlane(controlPlaneNamespace, hcluster.Name)
+	_, err = createOrUpdate(ctx, r.Client, hcp, func() error {
+		return reconcileHostedControlPlane(hcp, hcluster, isAutoscalingNeeded, isAWSNodeTerminationHandlerNeeded,
+			annotationsForCertRenewal(log,
+				hcp,
+				shouldCheckForStaleCerts(hcluster, defaultToControlPlaneV2),
```
csrwng, line 1607: Bug: when `CPOImageAndNamespace` fails (pull secret unavailable), `controlPlaneOperatorImageLabels` is nil, so `defaultToControlPlaneV2` is `false`. This causes `shouldCheckForStaleCerts` to return `true` for **CPOv2 clusters** — the opposite of what it should do.

The legacy path avoided this entirely by passing a no-op cert annotation function (`func() (map[string]string, error) { return nil, nil }`) during pull-secret-unavailable recovery (see `reconcile_legacy.go:335-336`).

All current CPO images have the `v2-isdefault=true` label (set in `Dockerfile`), and any cluster that's ever been healthy has `HasBeenAvailableAnnotation=true`. So the stale cert check would incorrectly run on every production CPOv2 cluster that transiently loses its pull secret. In the worst case, if the KAS serving cert secret and the live endpoint cert don't match (e.g. during a CPOv2-managed cert rotation), this sets `RestartDateAnnotation` on the HCP — triggering an **unnecessary control plane pod restart**.

Suggested fix: when `controlPlaneOperatorImageLabels` is nil, default `defaultToControlPlaneV2` to `true` (safe — it only disables the stale cert check, which shouldn't run when we can't determine the CPO version). Or pass a no-op cert annotation function like legacy did when pull secret is unavailable.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -1423,537 +1847,406 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
...
+// from the HCP namespace to the HostedCluster namespace, or deletes it if not needed.
+func (r *HostedClusterReconciler) reconcileCustomKubeconfigSync(
+	ctx context.Context, hcluster *hyperv1.HostedCluster, hcp *hyperv1.HostedControlPlane,
+	createOrUpdate upsert.CreateOrUpdateFN, cpoSupportsKASCustomKubeconfig bool,
+) error {
+	if !cpoSupportsKASCustomKubeconfig {
+		return nil
+	}
+	if len(hcp.Spec.KubeAPIServerDNSName) > 0 {
+		_, err := r.reconcileCustomExternalKubeconfig(ctx, createOrUpdate, hcp, hcluster)
```
csrwng: Bug: the `*time.Duration` requeue return value from `reconcileCustomExternalKubeconfig` is discarded here. Legacy captured it and stored it in `requeueAfter` (`reconcile_legacy.go:1571-1577`).

This requeue (30 seconds) fires when the CPO has advertised a custom kubeconfig secret name in `hcp.Status.CustomKubeconfig` but the secret doesn't exist yet — during initial provisioning or retarget. Without it, the controller won't poll for the secret and instead relies on watch events, which could delay custom kubeconfig availability.

```suggestion
		requeue, err := r.reconcileCustomExternalKubeconfig(ctx, createOrUpdate, hcp, hcluster)
```

The requeue value would also need to be propagated up to `report.requestRequeue`. One option: accept the `*reconcileReport` as a parameter, or return `(*time.Duration, error)` from this method and let the caller feed it to the report.

csrwng (commented):
Reviewed the behavioral changes between the legacy and new reconcile paths. Two items worth addressing before merge.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
muraee: Done. When `controlPlaneOperatorImageLabels` is nil (pull secret outage), `defaultToControlPlaneV2` now defaults to `true` — all current CPO images set `v2-isdefault=true`, so this prevents the stale cert check from incorrectly running on CPOv2 clusters.

---

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
muraee: Done. Propagated the requeue from `reconcileCustomExternalKubeconfig` through `reconcileCustomKubeconfigSync` → `reconcileKubeconfigAndPasswordSync` → `report.requestRequeue()`.

Checked the source: the KAS v2 component creates the custom kubeconfig secret without `SetHostedClusterAnnotation()`, so the watch won't catch its creation — the 30s requeue is needed.

---

PR #7916, opened 2026-03-11
AUTOSCALE-558: Expose KubeletConfig on OpenShiftEC2NodeClass

jkyros:

This: 
- Exposes `spec.kubelet` on `OpenShiftEC2NodeClass`, but DOES NOT reconcile it to ec2nodeclass directly like the other fields
- Converts it into ignition, merges it with the  existing default taint KubeletConfig, and adds it to the in-memory nodepool's list of  MachineConfig object references
  -  Much like highlander, there can be only one (KubeletConfig) in the MachineConfig pool 
 
Because: 
- We're using the Custom AMIFamily, the `spec.Kubelet` from the ec2nodeclass does NOT get reconciled into userdata by Karpenter, so we have to stuff the KubeletConfig into the ignition/MachineConfig ourselves. 
- Users need to be able to configure KubeletConfig on their nodes 

Fixes [AUTOSCALE-558](https://issues.redhat.com/browse/AUTOSCALE-558)

- I moved the base taint configMap out of the hostedcluster controller and into the Karpenter operator.
  - The thinking was it probably belonged to karpenter-operator anyway, and if I could access it as a "constant" I didn't need to request its contents from the cluster every reconcile to merge with KubeletConfig
- The e2e for KubeletConfig might be a little extra with the fixture, but I forgot that the MCO splits `systemReserved` into that other env file and I thought I broke it because it was missing -- wrote that to figure it out 

## Checklist:

## Release Notes

* **New Features**
  * Added kubelet configuration support for OpenShiftEC2NodeClass, enabling customization of kubelet parameters including pod density, resource reservations, eviction policies, image garbage collection settings, and CPU CFS quota.

* **Tests**
  * Added comprehensive test coverage for kubelet configuration reconciliation and end-to-end validation.


karpenter-operator/controllers/karpenterignition/karpenterignition_controller.go
```diff
@@ -387,6 +400,47 @@ func detectVersionSkew(hostedCluster *hyperv1.HostedCluster, version string) err
...
+			cm.Labels = map[string]string{}
+		}
+		cm.Labels[karpenterutil.KarpenterNodeClassKubeletConfigLabel] = "true"
+		cm.Data = map[string]string{
+			"config": manifest,
+		}
+		return nil
+	})
+	return err
+}
```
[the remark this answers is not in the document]
jkyros: Finalizer added if you want to look again.

jkyros:
This was failing because we didn't have https://github.com/openshift/hypershift/pull/7940 yet 
/test e2e-aws-autonode

jkyros:
@coderabbitai resume

jkyros:
Rebased, test fixes 
 /test e2e-aws-autonode

api/karpenter/v1beta1/karpenter_types.go
```diff
@@ -5,6 +5,51 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// KubeletConfiguration configures kubelet settings for nodes provisioned by this NodeClass.
+// These settings are injected into the node's ignition configuration via MachineConfig.
+type KubeletConfiguration struct {
+	// ClusterDNS is a list of IP addresses for the cluster DNS server.
+	// +optional
+	ClusterDNS []string `json:"clusterDNS,omitempty"`
```
jkyros: The mcc normally refuses to set clusterDNS because it's scary. We could potentially omit it from our API (which would skew from upstream) or just put guardrails around it, but I just wanted to mention itt: https://github.com/openshift/machine-config-operator/blob/4b32d3f8d98549400ec029c7c5cb23b44e045d10/pkg/controller/kubelet-config/helpers.go#L371

test/e2e/karpenter_test.go
```diff
@@ -700,6 +718,198 @@ func TestKarpenter(t *testing.T) {
...
+			nc := &hyperkarpenterv1.OpenshiftEC2NodeClass{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "kubelet-config-test",
+				},
+				Spec: hyperkarpenterv1.OpenshiftEC2NodeClassSpec{
+					Kubelet: &hyperkarpenterv1.KubeletConfiguration{
+						MaxPods:        ptr.To[int32](203),
+						PodsPerCore:    ptr.To[int32](11),
+						SystemReserved: map[string]string{"cpu": "510m", "memory": "521Mi"},
+						KubeReserved:   map[string]string{"cpu": "510m", "memory": "521Mi"},
```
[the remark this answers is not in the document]
jkyros: I understand why you are concerned, but did you look at the test fixture that does the grepping? They end up in different files and it checks the whole label.

jkyros:
/test e2e-aws-autonode
EDIT: did I use my quota or what, lol, it just won't run

api/karpenter/v1beta1/karpenter_types.go
```diff
@@ -5,6 +5,51 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// KubeletConfiguration configures kubelet settings for nodes provisioned by this NodeClass.
+// These settings are injected into the node's ignition configuration via MachineConfig.
+type KubeletConfiguration struct {
```
robpblake: @jkyros I'm pretty sure we'll need to expose `podPidsLimit` here as well. That is currently the only option available to customers using the `OCM` API, so I would expect them to need it here as well.

maxcao13:
/test e2e-aws-autonode
/test e2e-aws-techpreview

EDIT: github bot didn't was scaled to zero during night and couldn't respond? :stuck_out_tongue:

api/karpenter/v1beta1/karpenter_types.go
maxcao13: If OpenShift standalone already deems this scary, I would lean towards not including it. Since we can control the API pretty easily here, I think it's easier to add later, than to potentially have to guardrail on standalone HCP as well?

api/karpenter/v1beta1/karpenter_types.go
maxcao13: This will have to be an openshiftec2nodeclass specific field, the upstream ec2nodeclass doesn't support this yet, but it's in "priority/backlog". `https://github.com/aws/karpenter-provider-aws/issues/7982`

Not that this matters too much to us, since we don't propagate the fields to ec2nodeclass anyways, but just noting here that the API with the new field will look slightly different to customers. If upstream eventually decides to support this, I assume we would want the field naming to be similar to not be confusing.

EDIT: With that other PR, I guess naming doesn't really matter as much as I think :-)

maxcao13: This needs to be aware of https://github.com/openshift/hypershift/pull/7952 and vice versa.

jkyros: I will align this with that once we get that figured out. :smile:

jkyros: I think https://github.com/openshift/hypershift/pull/7952 is going to decide the direction here. If we're skewing from upstream already with that, I think this boils down to "what list of fields do we want exposed here", and we're deciding to more-or-less ignore upstream because our fields are already going to be weird and OpenShift-ified, right?

jkyros: I added `podPodsLimit` and made it sync the fields Karpenter cares about back to `ec2nodeclass` (it apparently uses them for scheduling) . 

I put the "christmas tree" approach where it syncs from the upstream on top of it so we can see it. If we don't like it, I'll peel it off.

jkyros:
Rebased
/test e2e-aws-autonode

jkyros:
Now that this is fixed, we can: 
/test e2e-aws-techpreview

hack/tools/kubelet-config-sync/main.go
```diff
@@ -0,0 +1,473 @@
...
+	"SyncFrequency":       "internal sync frequency",
+	"FileCheckFrequency":  "file check frequency",
+	"HTTPCheckFrequency":  "http check frequency",
+	"StaticPodURL":        "static pod URL",
+	"StaticPodURLHeader":  "static pod URL header",
+	"Address":             "kubelet bind address",
+	"Port":                "kubelet port",
+	"ReadOnlyPort":        "read-only port",
+
+	// TLS
```
jkyros, line 35: I set my "ast atrocity" that generates this from the kubernetes kubeletconfig struct up for deny-list (because I figured we'd want everything _except_ for dangerous ones), but we could also allow-list, or expose them all and then reject them via VAP/webhook/in the controller). I just wanted to vet the generation approach. 

I'm sure we could spend too long bikeshedding which ones to disable (aside from `ClusterDNS`) :smile:

jkyros:
closing in favor of https://github.com/openshift/hypershift/pull/8192

PR #7868, opened 2026-03-05
OCPBUGS-74960: prevent resource leak on deletion and handle DependencyViolation

sdminonne:
## Summary

- When `getClients` fails during deletion (e.g., after an operator restart), the controller now returns an error instead of logging and falling through to finalizer removal, which would permanently orphan AWS resources (security groups, VPC endpoints, DNS records)
- On deletion, the controller now performs best-effort client initialization by listing `HostedControlPlane` resources in the namespace. After a controller restart the `clientBuilder` is uninitialized; if the HCP still exists, `initializeWithHCP` is called so that `getClients` can succeed and cleanup can proceed
- Adds `DependencyViolation` error handling to the `deleteSecurityGroup` function — when AWS returns `DependencyViolation` during security group ingress/egress revocation or deletion, the controller returns a sentinel error that the caller translates into a controlled requeue (5s delay), allowing AWS to finish VPC endpoint cleanup before retrying
- Extracts `awsClientProvider` interface from `clientBuilder` to enable mock injection in tests
- Documents the remaining SharedVPC leak scenario: when the operator restarts during deletion and the HCP has already been deleted, the SharedVPC role ARNs (needed for cross-account AWS access) are lost. The fix preserves the finalizer, but retries will never succeed. A proper fix requires persisting the SharedVPC role ARNs in the AWSEndpointService status

## Test plan

- [x] Unit tests added for deletion reconciliation (`TestReconcileDeletion`): successful cleanup, empty status, VPC endpoint failure, DependencyViolation requeue
- [x] Unit test for best-effort HCP initialization during deletion: verifies `initializeWithHCP` is called when the HCP exists in the namespace
- [x] Unit test reproducing the controller-restart bug (`TestReconcileDeletion_AfterControllerRestart`): verifies error is returned and finalizer is preserved when no HCP exists
- [x] Unit tests for `deleteSecurityGroup` covering all `DependencyViolation` paths (ingress, egress, delete), SG not found, empty describe results, no ingress/egress rules, other AWS errors
- [x] Unit tests documenting the SharedVPC leak scenario (`TestReconcileDeletionSharedVPC`): uninitialized client after restart, initialized client without role ARNs
- [x] All new test cases pass
- [x] Package builds successfully
- [ ] Verify in a real cluster that orphaned security groups are eventually cleaned up on VPC endpoint deletion

Fixes: https://issues.redhat.com/browse/OCPBUGS-74960

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #7929, opened 2026-03-11
NO-JIRA: fix(aws): improve endpoint and security group teardown cleanup

olucasfreitas:
This PR fixes AWS teardown paths that can leak security groups and block VPC deletion after cluster uninstall.

It addresses three concrete cleanup issues:
- Keeps `AWSEndpointService` finalizers until cleanup can actually run, instead of skipping cleanup when AWS clients cannot be initialized during deletion.
- Handles interface endpoint deletion as asynchronous and requeues until endpoint deletion is fully observed before continuing SG cleanup.
- Treats security group `DependencyViolation` as retryable teardown progress and requeues instead of allowing teardown to continue with leaked SGs.

It also fixes a finalizer handling bug in `HostedCluster` deletion where CPO finalizers could be removed even with valid credentials after the grace period.

This is a run at helping fix the [OCPBUGS-74960](https://issues.redhat.com/browse/OCPBUGS-74960) bug witch already has another [PR](https://github.com/openshift/hypershift/pull/7868) for it, the intentions is to help even further to mitigate the issues with the cleanup

- Opened in draft so CI can be run and reviewed before marking ready.
- Includes focused unit tests for new retry/finalizer behavior.

## Checklist:

## Release Notes

* **Bug Fixes**
  * Improved reliability of AWS VPC endpoint and security group deletion with automatic retry mechanisms.
  * Enhanced finalizer handling for AWS endpoint services to prevent stuck cleanup operations.
  * Better error tolerance during resource cleanup for transient AWS API failures.

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
@@ -374,7 +374,7 @@ func (r *HostedControlPlaneReconciler) reconcileDeletion(ctx context.Context, ho
 	if shouldCleanupCloudResources(r.Log, hostedControlPlane) {
 		if code, destroyErr := r.destroyAWSDefaultSecurityGroup(ctx, hostedControlPlane); destroyErr != nil {
 			condition.Message = "failed to delete AWS default security group"
-			if code == "DependencyViolation" {
+			if code == supportawsutil.DependencyViolation {
 				condition.Message = destroyErr.Error()
 			}
 			condition.Reason = hyperv1.AWSErrorReason
@@ -388,8 +388,9 @@ func (r *HostedControlPlaneReconciler) reconcileDeletion(ctx context.Context, ho
 			switch code {
 			case "UnauthorizedOperation":
 				r.Log.Error(destroyErr, "Skipping AWS default security group deletion because of unauthorized operation.")
-			case "DependencyViolation":
-				r.Log.Error(destroyErr, "Skipping AWS default security group deletion because of dependency violation.")
+			case supportawsutil.DependencyViolation:
+				r.Log.Info("AWS default security group has dependencies, will retry", "error", destroyErr.Error())
+				return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
 			default:
 				return ctrl.Result{}, fmt.Errorf("failed to delete AWS default security group: %w", destroyErr)
 			}
@@ -2876,9 +2877,11 @@ func (r *HostedControlPlaneReconciler) destroyAWSDefaultSecurityGroup(ctx contex
 			IpPermissions: sg.IpPermissions,
 		}); err != nil {
 			code := supportawsutil.AWSErrorCode(err)
-			log.Error(err, "failed to revoke security group ingress permissions", "SecurityGroupID", aws.ToString(sg.GroupId), "code", code)
-
-			return code, fmt.Errorf("failed to revoke security group ingress rules: %s", code)
+			if !supportawsutil.ShouldIgnoreRevokePermissionError(code) {
+				log.Error(err, "failed to revoke security group ingress permissions", "SecurityGroupID", aws.ToString(sg.GroupId), "code", code)
+				return code, fmt.Errorf("failed to revoke security group ingress rules: %s", code)
+			}
+			log.Info("security group ingress permission already revoked", "SecurityGroupID", aws.ToString(sg.GroupId))
 		}
 	}
 
@@ -2888,9 +2891,11 @@ func (r *HostedControlPlaneReconciler) destroyAWSDefaultSecurityGroup(ctx contex
 			IpPermissions: sg.IpPermissionsEgress,
 		}); err != nil {
 			code := supportawsutil.AWSErrorCode(err)
-			log.Error(err, "failed to revoke security group egress permissions", "SecurityGroupID", aws.ToString(sg.GroupId), "code", code)
-
-			return code, fmt.Errorf("failed to revoke security group egress rules: %s", code)
+			if !supportawsutil.ShouldIgnoreRevokePermissionError(code) {
+				log.Error(err, "failed to revoke security group egress permissions", "SecurityGroupID", aws.ToString(sg.GroupId), "code", code)
+				return code, fmt.Errorf("failed to revoke security group egress rules: %s", code)
+			}
+			log.Info("security group egress permission already revoked", "SecurityGroupID", aws.ToString(sg.GroupId))
 		}
 	}
 
@@ -2898,8 +2903,11 @@ func (r *HostedControlPlaneReconciler) destroyAWSDefaultSecurityGroup(ctx contex
 		GroupId: sg.GroupId,
 	}); err != nil {
 		code := supportawsutil.AWSErrorCode(err)
+		if supportawsutil.ShouldIgnoreSGNotFound(code) {
+			log.Info("security group already deleted", "SecurityGroupID", aws.ToString(sg.GroupId))
+			return "", nil
+		}
 		log.Error(err, "failed to delete security group", "SecurityGroupID", aws.ToString(sg.GroupId), "code", code)
-
 		return code, fmt.Errorf("failed to delete security group %s: %s", aws.ToString(sg.GroupId), code)
 	}
 
--- a/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
+++ b/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
@@ -3302,7 +3302,10 @@ func (r *HostedClusterReconciler) deleteNodePools(ctx context.Context, c client.
 }
 
 // deleteAWSEndpointServices loops over AWSEndpointServiceList items and sends a delete request for each.
-// If the HC has no valid aws credentials it removes the CPO finalizer for each AWSEndpointService.
+// For items already being deleted, the CPO finalizer is preserved during the grace period when
+// credentials are valid or unknown, giving CPO time to clean up AWS resources. After the grace
+// period expires (or immediately when credentials are invalid), the finalizer is removed so
+// deletion can proceed.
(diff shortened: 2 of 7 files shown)


bryan-cox:
Thanks for the PR! We need a Jira for this per our contributor guidelines.

olucasfreitas:
@bryan-cox there is this issue here https://issues.redhat.com/browse/OCPBUGS-74960 that already tries to address this, my PR can maybe be viewed more as a nice to have or different approach, should I reference this ? do I reach out to someone to create this ticket ? do I create the ticket myself ?

control-plane-operator/controllers/awsprivatelink/awsprivatelink_controller.go
```diff
@@ -398,9 +398,20 @@ func (r *AWSEndpointServiceReconciler) Reconcile(ctx context.Context, req ctrl.R
 			return ctrl.Result{}, nil
 		}
 
+		hcp, err := r.hostedControlPlane(ctx, req.Namespace)
```
jparrill: The deletion path on `main` was recently refactored (lines 414-435) to already do best-effort `initializeWithHCP` from an `hcpList` lookup. This PR adds a second, slightly different initialization via `hostedControlPlane()` + `initializeWithHCP()` before `getClients()`. After rebase these will conflict — you'll need to reconcile both approaches into one.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -3187,29 +3187,34 @@ func (r *HostedClusterReconciler) deleteNodePools(ctx context.Context, c client.
...
 	log := ctrl.LoggerFrom(ctx)
 	var awsEndpointServiceList hyperv1.AWSEndpointServiceList
 	if err := c.List(ctx, &awsEndpointServiceList, &client.ListOptions{Namespace: namespace}); err != nil && !apierrors.IsNotFound(err) {
 		return false, fmt.Errorf("error listing awsendpointservices in namespace %s: %w", namespace, err)
 	}
+	credentialStatus := platformaws.GetCredentialStatus(hc)
 	for _, ep := range awsEndpointServiceList.Items {
 		if ep.DeletionTimestamp != nil {
-			if platformaws.GetCredentialStatus(hc) == platformaws.CredentialStatusValid && time.Since(ep.DeletionTimestamp.Time) < awsEndpointDeletionGracePeriod {
+			if credentialStatus != platformaws.CredentialStatusInvalid {
```
jparrill: Semantic change worth calling out in the PR description: before this change, `Unknown` + expired grace period → remove finalizer (orphan). After this change, `Unknown` → always keep finalizer (never orphan). This is more conservative and correct, but could block deletion indefinitely if credential conditions never resolve. Worth documenting the intentional behavior change.

control-plane-operator/controllers/awsprivatelink/awsprivatelink_controller.go
```diff
@@ -1070,17 +1079,59 @@ func (r *AWSEndpointServiceReconciler) delete(ctx context.Context, awsEndpointSe
 	return true, nil
 }
 
-func (r *AWSEndpointServiceReconciler) deleteSecurityGroup(ctx context.Context, ec2Client ec2iface.EC2API, sgID string) error {
+func ensureVPCEndpointDeleted(ctx context.Context, ec2Client ec2iface.EC2API, endpointID string) (bool, error) {
+	deleteOutput, err := ec2Client.DeleteVpcEndpointsWithContext(ctx, &ec2.DeleteVpcEndpointsInput{
```
jparrill: This calls `Delete` then immediately `Describe`. Since endpoint deletion is async, `Describe` will find it still alive and the caller requeues — which calls `Delete` again on the next reconcile. Consider checking the endpoint state first and only calling `Delete` if it's not already in `deleting` state, to avoid redundant API calls.

control-plane-operator/controllers/awsprivatelink/awsprivatelink_controller_test.go
```diff
@@ -206,3 +214,342 @@ func TestDiffPermissions(t *testing.T) {
...
+		},
+	}
+
+	for _, tc := range tests {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			deleted, err := ensureVPCEndpointDeleted(t.Context(), tc.client, endpointID)
+			if tc.expectError && err == nil {
+				t.Fatal("expected error, got nil")
```
jparrill: Tests in this file already import gomega (`. "github.com/onsi/gomega"`). Please use gomega assertions (`g.Expect(err).ToNot(HaveOccurred())`) instead of `t.Fatal`/`t.Fatalf` with manual comparisons. Also, follow the repo's test conventions: Gherkin-style names ("When... it should...") — which you already do in the table names, good — but also use table-driven format with `g := NewWithT(t)` and gomega matchers throughout. Same applies to the other new test files in this PR.

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -2324,6 +2331,18 @@ func shouldCleanupCloudResources(log logr.Logger, hcp *hyperv1.HostedControlPlan
 	return hcp.Annotations[hyperv1.CleanupCloudResourcesAnnotation] == "true"
 }
 
+func shouldRetryDefaultSecurityGroupDeletion(code string) bool {
```
jparrill: This is identical to `shouldRetrySecurityGroupDeletion` in `awsprivatelink_controller.go` (same for the `shouldIgnore*` variants). Consider extracting to a shared package (e.g., `support/awsutil`) to avoid duplication.

jparrill (commented):

control-plane-operator/controllers/awsprivatelink/awsprivatelink_controller.go
olucasfreitas: Thanks for the heads-up. The branch has been fully rebased onto current `main`, so the older `hostedControlPlane()`-gated deletion flow was dropped entirely. The current best-effort `initializeWithHCP` logic from `main` (introduced by PR #7868) is preserved as-is — this PR no longer touches `awsprivatelink_controller.go`.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
olucasfreitas: Good call. Updated the implementation to use a bounded fallback: when `CredentialStatusUnknown`, the CPO finalizer is preserved during the grace period (same 10-minute window as `Valid`), then falls back to removal if status never resolves. This avoids the original immediate-orphan behavior without creating an indefinite deletion wedge. The function comment and log messages have been updated to match the actual logic.

control-plane-operator/controllers/awsprivatelink/awsprivatelink_controller.go
olucasfreitas: Agreed on the redundant API calls concern. After rebasing, this PR no longer modifies `awsprivatelink_controller.go` at all — the current `main` behavior (from PR #7868) is preserved as-is to avoid reintroducing conflicts. The Delete-then-Describe churn is a pre-existing pattern in `main` that could be optimized in a follow-up PR by checking endpoint state before calling Delete.

control-plane-operator/controllers/awsprivatelink/awsprivatelink_controller_test.go
olucasfreitas: Fixed. The older `awsprivatelink` test changes were dropped during rebase (those paths are already well-tested on `main` from PR #7868). All remaining new tests use gomega matchers with `g := NewWithT(t)` and table-driven format with "When... it should..." naming throughout.

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
olucasfreitas: Done. Shared AWS error-code constants (`InvalidGroupNotFound`, `InvalidPermissionNotFound`) are now in `support/awsutil/errorcode.go`, with helper predicates `ShouldIgnoreSGNotFound` and `ShouldIgnoreRevokePermissionError` in `support/awsutil/sg_delete.go`. The HCP controller's `destroyAWSDefaultSecurityGroup` uses these directly instead of inline string comparisons. Only helpers that are actually called from production code are included.

olucasfreitas:
## Addressing review feedback

The branch has been fully rebased onto current `main` and the scope narrowed to only the changes that are still needed after PR #7868 was merged.

### Changes since last review

**HCP default-sg teardown (`hostedcontrolplane_controller.go`):**
- `DependencyViolation` now triggers a `RequeueAfter: 5s` instead of being silently skipped, preventing `*-default-sg` security groups from being orphaned.
- `InvalidGroup.NotFound` from `DeleteSecurityGroup` is treated as success (race between describe and delete).
- `InvalidPermission.NotFound` from ingress/egress revoke calls is treated as non-fatal.

**Endpoint service finalizer logic (`hostedcluster_controller.go`):**
- `CredentialStatusUnknown` now preserves the CPO finalizer during the grace period (same 10-minute window), then falls back to removal if status never resolves. Previously `Unknown` was treated identically to `Invalid`, causing premature finalizer removal.

**Shared helpers (`support/awsutil/`):**
- Extracted `InvalidGroupNotFound` and `InvalidPermissionNotFound` constants plus `ShouldIgnoreSGNotFound` and `ShouldIgnoreRevokePermissionError` predicates.
- No unused helpers left behind.

**Tests:**
- All tests use gomega matchers with table-driven format.
- Added OIDC-invalid credential branch coverage per CodeRabbit's suggestion.
- Added `unknown-within-grace` and `unknown-expired` test cases.

**Dropped from scope:**
- All `awsprivatelink_controller.go` changes (already handled by PR #7868 on `main`).

olucasfreitas:
@jparrill Sorry it took me a while to reply your message, now that I got some time I would love to get back to this 

when you can, please take a look here at the new changes

PR #7934, opened 2026-03-11
OCPBUGS-78297: use ClusterImagePolicy v1 instead of v1alpha1

QiWang19:
Migrate the ClusterImagePolicy API from v1alpha1 to v1. As v1alpha1 was 
a TechPreview API, it is being dropped in favor of the v1(GA) version.

Rendering v1alpha1 ClusterImagePolicy via the Machine Config Operator (MCO) 
become a no-op after the upgrade to the v1 API. It is safe to 
remove the v1alpha1 definition at this stage of the lifecycle.

Fixes: https://issues.redhat.com/browse/OCPBUGS-78297 

## Checklist:

## Summary by CodeRabbit

* **Refactor**
  * Switched image policy handling from the legacy alpha API to the stable API to improve platform stability and compatibility.
  * Removed dependency on the legacy alpha configuration during initialization, simplifying startup behavior and reducing potential integration issues.

### DIFF
--- a/hypershift-operator/controllers/nodepool/config.go
+++ b/hypershift-operator/controllers/nodepool/config.go
@@ -18,7 +18,6 @@ import (
 	supportutil "github.com/openshift/hypershift/support/util"
 
 	configv1 "github.com/openshift/api/config/v1"
-	configv1alpha1 "github.com/openshift/api/config/v1alpha1"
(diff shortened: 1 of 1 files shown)


enxebre:
Does it really make sense to have this at the NodePool level? this might be a good opportunity to move this into HC

PR #7979, opened 2026-03-16
AUTOSCALE-490: increase autonode test coverage

maxcao13:

```
Improves test coverage of the TestKarpenter suite:
// – validate the CRDs are installed
// – validate the default class is created and has expected values 
// – validate admin can't modify fields owned by the service, e.g. ami.
```

Fixes https://redhat.atlassian.net/browse/AUTOSCALE-490

This shouldn't increase the time it takes for the test to finish by a large amount. This will followed up with https://redhat.atlassian.net/browse/AUTOSCALE-577 after pencils-down to consolidate tests to improve execution time.

## Checklist:

## Summary by CodeRabbit

* **Tests**
  * Added CRD-level installation checks for Karpenter/Openshift Karpenter resources in guest clusters.
  * Added polling validations for default node-class resources, ensuring selector terms and discovery tags are populated from cluster infra ID.
  * Added downstream validations ensuring AMI selector terms, AMI family, user data, and service-owned tags are set.
  * Enhanced authorization tests to confirm GET, delete, and update operations are properly rejected.

### DIFF
--- a/test/e2e/karpenter_test.go
+++ b/test/e2e/karpenter_test.go
@@ -35,6 +35,7 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	policyv1 "k8s.io/api/policy/v1"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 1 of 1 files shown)


maxcao13:
I think the job is just taking too long, and we reached a timeout.

Will hopefully be solved by this story: https://redhat.atlassian.net/browse/AUTOSCALE-577
Will be working on it this sprint.

i'll wait for https://github.com/openshift/hypershift/pull/8198 to get in

PR #7996, opened 2026-03-18
CNTRLPLANE-2939: Coordinate CRD lifecycle with Cluster CAPI Operator

muraee:
When a management cluster has the Cluster CAPI Operator (CCAPIO) installed, HyperShift's hypershift install command installs CAPI CRDs that may conflict with CRDs already managed by CCAPIO. This PR adds coordination logic to the install flow:

Dry-run validation: All CRDs are validated via server-side dry-run before applying, catching webhook rejections and schema conflicts early — regardless of CCAPIO presence
CCAPIO detection: Uses the discovery API to check if the ClusterAPI kind is served under operator.openshift.io/v1alpha1
CRD ownership signaling: Gets-or-creates the ClusterAPI config singleton and populates spec.unmanagedCustomResourceDefinitions with HyperShift's CAPI CRD names (groups ending in .cluster.x-k8s.io), telling CCAPIO to skip these CRDs
The coordination only mutates the ClusterAPI config after dry-run passes, so there are no side effects if CRDs cannot be applied.

### Install flow (after this PR):
```
generate manifests
  → dry-run validate all CRDs
  → if ClusterAPI API available: ensure unmanaged CRDs in config
  → apply CRDs
  → wait for CRDs to be established (if requested)
  → apply remaining objects
```

### Vendor bump
Bumps github.com/openshift/api from v0.0.0-20260120150926-4c643a652d54 to v0.0.0-20260227165130-5a7add616a90 to vendor the ClusterAPI config type (operator.openshift.io/v1alpha1). The selected commit remains compatible with k8s.io v0.34.x.

Fixes https://issues.redhat.com/browse/CNTRLPLANE-2939

The unmanagedCustomResourceDefinitions field is append-only (CEL enforced) — once a CRD name is added, it cannot be removed
In the pre-upgrade scenario (4.N-1), CCAPIO webhooks are not running so dry-run passes trivially — this is expected and correct
Helm chart rendering (hyperShiftOperatorManifests with nil client) is unaffected — CCAPIO coordination only runs in InstallHyperShiftOperator

## Summary by CodeRabbit

* **New Features**
  * Installation performs CRD dry-run validation and coordinates with ClusterAPI to mark certain CRDs as unmanaged when applicable.

* **Chores**
  * Updated OpenShift API dependency version.
  * Removed NetworkDiagnosticsConfig from feature gates in generated CRD manifests.

* **Tests**
  * Added tests for ClusterAPI detection and unmanaged-CRD coordination.

### DIFF
--- a/cmd/install/install.go
+++ b/cmd/install/install.go
@@ -37,6 +37,7 @@ import (
 
 	configv1 "github.com/openshift/api/config/v1"
 	imageapi "github.com/openshift/api/image/v1"
+	operatorv1alpha1 "github.com/openshift/api/operator/v1alpha1"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -49,6 +50,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/apimachinery/pkg/util/yaml"
+	"k8s.io/client-go/discovery"
 	"k8s.io/utils/ptr"
 	"k8s.io/utils/set"
 
@@ -423,6 +425,36 @@ func InstallHyperShiftOperator(ctx context.Context, out io.Writer, opts Options)
 	crds, objects, err := hyperShiftOperatorManifests(ctx, client, opts)
 	if err != nil {
 		return err
+	}
+
+	// Validate all CRDs via dry-run before applying
+	if err := dryRunValidateCRDs(ctx, out, crds); err != nil {
+		return err
+	}
+
+	// Coordinate with Cluster CAPI Operator if the ClusterAPI API is available.
+	// This is done after dry-run so the ClusterAPI config is not mutated if CRDs
+	// cannot be applied.
+	config, err := util.GetConfig()
+	if err != nil {
+		return fmt.Errorf("failed to get kubernetes config: %w", err)
+	}
+	discoveryClient, err := discovery.NewDiscoveryClientForConfig(config)
+	if err != nil {
+		return fmt.Errorf("failed to create discovery client: %w", err)
+	}
+	registered, err := isClusterAPIRegistered(discoveryClient)
+	if err != nil {
+		return err
+	}
+	if registered {
+		fmt.Fprintf(out, "ClusterAPI API detected, coordinating with Cluster CAPI Operator\n")
+		if err := ensureUnmanagedCRDs(ctx, out, client, crds); err != nil {
+			return err
+		}
+		if err := waitForCAPIOperatorSync(ctx, out, client); err != nil {
+			return err
+		}
 	}
 
 	err = apply(ctx, out, crds)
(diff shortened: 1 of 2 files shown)


api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/hostedclusters.hypershift.openshift.io/HCPEtcdBackup.yaml
```diff
@@ -575,11 +575,14 @@ spec:
                               ciphers:
                                 description: |-
                                   ciphers is used to specify the cipher algorithms that are negotiated
-                                  during the TLS handshake.  Operators may remove entries their operands
-                                  do not support.  For example, to use DES-CBC3-SHA  (yaml):
```
enxebre: where this change coming from?

enxebre:
we'll want to test this via https://github.com/openshift/hypershift/pull/8016 as well
/lgtm
/hold
pending https://github.com/openshift/hypershift/pull/7996#discussion_r2965238884 resolution

JoelSpeed:
> In the pre-upgrade scenario (4.N-1), CCAPIO webhooks are not running so dry-run passes trivially — this is expected and correct

I think eventually we will have upgrade checks in place so that the webhooks will warn about the future version being incompatible, but otherwise, PR description is accurate

JoelSpeed:
/lgtm

> In the pre-upgrade scenario (4.N-1), CCAPIO webhooks are not running so dry-run passes trivially — this is expected and correct

I think this has made us realise we need to adjust the way we are setting up CompatibilityRequirements, but that's fine, I'll make a note and we can make this concept work and be safe so that your initial dry runs are effective

api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/hostedclusters.hypershift.openshift.io/HCPEtcdBackup.yaml
JoelSpeed: Recent change to o/api that is picked up in the vendor bump

muraee: as @JoelSpeed said. This is expected

enxebre:
@muraee can we rebase?
/hold cancel

gaol:
I did a round of tests and attached the test report to the Jira comments.

basically what I did were:
* test on 4.21 nightly cluster on AWS which does not have ClusterAPI CRD yet, so the coordinate does not happen, all works as before, HO installed without problems.

* test on 4.22 nightly cluster on AWS, enable the clusterapi featuregate and it will create a `ClusterAPI` CR, then `hypershift install` will do the coordinations.

* The hypershift operator pods are in running status after installation, the `UnmanagedCustomResourceDefinitions` contains all `*.cluster.x-k8s.io` CRD names, and the CRDs have hypershift as the field owner.

I did not test the hosted cluster creation yet after the hypershift operator installation succeeded (mgmt cluster expired ...).

If the pr needs the verified tag today, I can tag :) , or I can do the hosted cluster creation next Tuesday. (PTO tomorrow and holidays)

@muraee

cmd/install/install.go
```diff
@@ -898,6 +927,127 @@ func isAzurePlatformEnabled(platformsToInstall []string) bool {
...
+			Spec: &operatorv1alpha1.ClusterAPISpec{
+				UnmanagedCustomResourceDefinitions: capiCRDNames.SortedList(),
+			},
+		}
+		if err := client.Create(ctx, clusterAPI); err != nil {
+			return fmt.Errorf("failed to create ClusterAPI config: %w", err)
+		}
+		fmt.Fprintf(out, "Created ClusterAPI config with %d unmanaged CRDs\n", capiCRDNames.Len())
+		return nil
+	}
```
mdbooth: Please don't do this. This should never happen, so if it did it would indicate an unknown state. Best to just report an error.

cmd/install/install.go
```diff
@@ -898,6 +927,127 @@ func isAzurePlatformEnabled(platformsToInstall []string) bool {
...
+	if merged.Len() == existing.Len() {
+		// No new entries to add
+		return nil
+	}
+
+	if clusterAPI.Spec == nil {
+		clusterAPI.Spec = &operatorv1alpha1.ClusterAPISpec{}
+	}
+	clusterAPI.Spec.UnmanagedCustomResourceDefinitions = merged.SortedList()
+	if err := client.Update(ctx, clusterAPI); err != nil {
```
mdbooth: I think CodeRabbit's comment below is legitimate. I recommend you don't merge the lists and just use SSA for this. Then you also don't have to handle conflicts.

cmd/install/install.go
```diff
@@ -413,6 +415,33 @@ func InstallHyperShiftOperator(ctx context.Context, out io.Writer, opts Options)
...
+	}
+	registered, err := isClusterAPIRegistered(discoveryClient)
+	if err != nil {
+		return err
+	}
+	if registered {
+		fmt.Fprintf(out, "ClusterAPI API detected, coordinating with Cluster CAPI Operator\n")
+		if err := ensureUnmanagedCRDs(ctx, out, client, crds); err != nil {
+			return err
+		}
```
mdbooth, line 456: This isn't safe, yet. You also need to wait for CAPI Operator to stop managing the CRDs. You need to check:
1. status.revisionObservedGeneration == clusterAPI.observedGeneration
2. status.desiredRevision == status.currentRevision

But we forgot to add status.revisionObservedGeneration 😬 I'll prioritise adding that, but until we do there's a somewhat complicated workaround: you can check in the revision list that currentRevision's unmanagedCustomResourceDefinitions contains all the ones you added.

mdbooth (commented):
This is mostly there, thanks. Once you're also waiting for the config change to have been actioned it'll be there.

From our side, in addition to the missing status field I mentioned, note that we also haven't actually integrated CRD Compatibility into the installer yet. We're hopefully very close, but it won't current do anything. Until we do, even after you've marked the CRDs unmanaged we're still going to be trying to manage them. This is going to result in a controller fight.

cmd/install/install.go
muraee: are you saying we shouldn't create the singleton object?

cmd/install/install.go
muraee: PR adding  `status.revisionObservedGeneration` https://github.com/openshift/api/pull/2786

muraee: updated

cmd/install/install.go
```diff
@@ -927,6 +959,133 @@ func isAzurePlatformEnabled(platformsToInstall []string) bool {
...
+// HyperShift's CAPI CRD names listed in unmanagedCustomResourceDefinitions. This tells
+// the Cluster CAPI Operator to skip these CRDs during its own installation.
+// Server-Side Apply is used so the API server handles create-or-update atomically and
+// merges set entries across field owners without conflicts.
+func ensureUnmanagedCRDs(ctx context.Context, out io.Writer, client crclient.Client, capiCRDs []crclient.Object) error {
+	// Collect CAPI CRD names (groups ending in .cluster.x-k8s.io)
+	capiCRDNames := set.New[string]()
+	for _, crd := range capiCRDs {
+		name := crd.GetName()
+		if strings.HasSuffix(name, ".cluster.x-k8s.io") {
```
mdbooth, line 1005: I suspect this suffix selection will be insufficient. I'm thinking specifically of ASO on Azure (not currently deployed by CAPIO, but probably will be eventually) and ORC on OpenStack (currently deployed by CAPIO), but there could be others.

Something which can be addressed when it comes up, doesn't have to be a blocker today.

cmd/install/install.go
```diff
@@ -927,6 +959,133 @@ func isAzurePlatformEnabled(platformsToInstall []string) bool {
...
+		Spec: &operatorv1alpha1.ClusterAPISpec{
+			UnmanagedCustomResourceDefinitions: capiCRDNames.SortedList(),
+		},
+	}
+
+	var buf bytes.Buffer
+	if err := hyperapi.YamlSerializer.Encode(clusterAPI, &buf); err != nil {
+		return fmt.Errorf("failed to encode ClusterAPI config: %w", err)
+	}
+	if err := client.Patch(ctx, clusterAPI, crclient.RawPatch(types.ApplyPatchType, buf.Bytes()),
```
mdbooth: FYI this is not safe in general due to the vagiaries of serialisation in Go. Suggest you use an ApplyConfiguration instead which is specifically designed for this. You'll find them in openshift/client-go.

cmd/install/install.go
```diff
@@ -927,6 +959,133 @@ func isAzurePlatformEnabled(platformsToInstall []string) bool {
...
+	fmt.Fprintf(out, "Waiting for Cluster CAPI Operator to sync...\n")
+	return wait.PollUntilContextCancel(waitCtx, 2*time.Second, true, func(ctx context.Context) (bool, error) {
+		clusterAPI := &operatorv1alpha1.ClusterAPI{}
+		if err := c.Get(ctx, crclient.ObjectKey{Name: "cluster"}, clusterAPI); err != nil {
+			return false, fmt.Errorf("failed to get ClusterAPI config: %w", err)
+		}
+
+		if clusterAPI.Status.ObservedRevisionGeneration < clusterAPI.Generation {
+			return false, nil
+		}
```
mdbooth: This is not correct, yet: you're asserting that the ClusterAPI object you've read here was up to date with itself, but you're not asserting that it's up to date with the one you just wrote. i.e. The first time this executes you could (and if you're using an informer almost definitely will, if you're not using an informer still might) read an old version of ClusterAPI *before* you updated it. This old ClusterAPI will almost certainly pass both of these checks, but it's still out of date.

cmd/install/install.go
```diff
@@ -927,6 +959,133 @@ func isAzurePlatformEnabled(platformsToInstall []string) bool {
...
+
+	var buf bytes.Buffer
+	if err := hyperapi.YamlSerializer.Encode(clusterAPI, &buf); err != nil {
+		return fmt.Errorf("failed to encode ClusterAPI config: %w", err)
+	}
+	if err := client.Patch(ctx, clusterAPI, crclient.RawPatch(types.ApplyPatchType, buf.Bytes()),
+		crclient.ForceOwnership, crclient.FieldOwner("hypershift"),
+	); err != nil {
+		return fmt.Errorf("failed to apply ClusterAPI config: %w", err)
+	}
```
mdbooth, line 1042: At this point your clusterAPI object has been updated to be current with whatever was written in the PATCH. The `metadata.generation` from this object is the one you need to compare against in `waitForCAPIOperatorSync`.

cmd/install/install.go
muraee: Agreed, this will likely need expanding for ASO/ORC and others. I've left it as-is for now since those aren't deployed by CAPIO yet — we can extend the filter when the need arises.

cmd/install/install.go
muraee: Good call. ApplyConfiguration types for `operator/v1alpha1` don't exist in openshift/client-go yet, so I've switched to constructing the SSA patch payload as a `map[string]any` with `encoding/json.Marshal`. This ensures only the fields we intend to own are included in the patch, avoiding zero-value serialization issues.

cmd/install/install.go
muraee: Fixed. `ensureUnmanagedCRDs` now returns the `metadata.generation` from the SSA patch response, and `waitForCAPIOperatorSync` compares `status.observedRevisionGeneration` against that known generation rather than against the polled object's own generation. Added a test case for the stale-read scenario.

cmd/install/install.go
muraee: Done — `ensureUnmanagedCRDs` now returns `clusterAPI.Generation` from the patch response object and the caller passes it through to `waitForCAPIOperatorSync`.

mdbooth (commented):
/lgtm

👍

PR #8013, opened 2026-03-19
OTA-1764: feat(cvo): add TLS cipher suites and minimum version flags

DavidHurta:

Add new flags to the hosted CVO to comply with the [centralized TLS configuration in HyperShift](https://github.com/openshift/enhancements/blob/master/enhancements/security/centralized-tls-config.md#hypershift--hosted-control-planes). These flags are to override the CVO's internal TLS profile, which is used for its metrics server. The flags were added with https://github.com/openshift/cluster-version-operator/pull/1338.

Fixes #[OTA-1764](https://redhat.atlassian.net/browse/OTA-1764)

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Enhanced hosted control plane deployment TLS configuration by honoring the specified minimum TLS version and TLS cipher suites.
  * Ensures the correct TLS CLI arguments are applied (including modern TLS profile behavior).

* **Tests**
  * Added table-driven test coverage to verify TLS-related deployment arguments are added or omitted correctly for different TLS configuration combinations.

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/v2/cvo/deployment.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/v2/cvo/deployment.go
@@ -109,10 +109,11 @@ func (cvo *clusterVersionOperator) adaptDeployment(cpContext component.WorkloadC
 			Value: dataPlaneReleaseImage,
 		})
 
-		if tlsMinVersion := config.MinTLSVersion(configuration.GetTLSSecurityProfile()); tlsMinVersion != "" {
+		tlsProfile := configuration.GetTLSSecurityProfile()
+		if tlsMinVersion := config.MinTLSVersion(tlsProfile); tlsMinVersion != "" {
 			c.Args = append(c.Args, fmt.Sprintf("--tls-min-version=%s", tlsMinVersion))
 		}
-		if cipherSuites := config.CipherSuites(configuration.GetTLSSecurityProfile()); len(cipherSuites) != 0 {
+		if cipherSuites := config.CipherSuites(tlsProfile); len(cipherSuites) != 0 {
 			c.Args = append(c.Args, fmt.Sprintf("--tls-cipher-suites=%s", strings.Join(cipherSuites, ",")))
 		}
 
--- a/control-plane-operator/controllers/hostedcontrolplane/v2/cvo/deployment.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/v2/cvo/deployment.go
@@ -102,11 +102,19 @@ func (cvo *clusterVersionOperator) adaptDeployment(cpContext component.WorkloadC
 		})
 	})
 
+	configuration := cpContext.HCP.Spec.Configuration
 	podspec.UpdateContainer(ComponentName, deployment.Spec.Template.Spec.Containers, func(c *corev1.Container) {
 		podspec.UpsertEnvVar(c, corev1.EnvVar{
 			Name:  "RELEASE_IMAGE",
 			Value: dataPlaneReleaseImage,
 		})
+
+		if tlsMinVersion := config.MinTLSVersion(configuration.GetTLSSecurityProfile()); tlsMinVersion != "" {
+			c.Args = append(c.Args, fmt.Sprintf("--tls-min-version=%s", tlsMinVersion))
+		}
+		if cipherSuites := config.CipherSuites(configuration.GetTLSSecurityProfile()); len(cipherSuites) != 0 {
+			c.Args = append(c.Args, fmt.Sprintf("--tls-cipher-suites=%s", strings.Join(cipherSuites, ",")))
+		}
 
 		if updateService := cpContext.HCP.Spec.UpdateService; updateService != "" {
 			c.Args = append(c.Args, "--update-service", string(updateService))
--- a/control-plane-operator/controllers/hostedcontrolplane/testdata/cluster-version-operator/AROSwift/zz_fixture_TestControlPlaneComponents_cluster_version_operator_deployment.yaml
+++ b/control-plane-operator/controllers/hostedcontrolplane/testdata/cluster-version-operator/AROSwift/zz_fixture_TestControlPlaneComponents_cluster_version_operator_deployment.yaml
@@ -76,6 +76,8 @@ spec:
         - --serving-key-file=/etc/kubernetes/certs/server/tls.key
         - --hypershift=true
         - --v=4
+        - --tls-min-version=VersionTLS12
+        - --tls-cipher-suites=TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256,TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256
         command:
         - cluster-version-operator
         env:
--- a/control-plane-operator/controllers/hostedcontrolplane/testdata/cluster-version-operator/GCP/zz_fixture_TestControlPlaneComponents_cluster_version_operator_deployment.yaml
+++ b/control-plane-operator/controllers/hostedcontrolplane/testdata/cluster-version-operator/GCP/zz_fixture_TestControlPlaneComponents_cluster_version_operator_deployment.yaml
@@ -76,6 +76,8 @@ spec:
         - --serving-key-file=/etc/kubernetes/certs/server/tls.key
         - --hypershift=true
         - --v=4
+        - --tls-min-version=VersionTLS12
+        - --tls-cipher-suites=TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256,TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256
         command:
         - cluster-version-operator
         env:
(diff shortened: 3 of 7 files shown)


control-plane-operator/controllers/hostedcontrolplane/v2/cvo/deployment.go
```diff
@@ -101,12 +101,20 @@ func (cvo *clusterVersionOperator) adaptDeployment(cpContext component.WorkloadC
 		})
 	})
 
+	configuration := cpContext.HCP.Spec.Configuration
```
hongkailiu, line 105: I would do the `nil` checking first.
Same on line 111 and line 114.

DavidHurta: Thanks for the review!

The `GetTLSSecurityProfile` function does check for nil dereference:

https://github.com/openshift/hypershift/blob/7ed7eb9d6f4e331efdb697a6cca88f428a4b322d/vendor/github.com/openshift/hypershift/api/hypershift/v1beta1/clusterconfig.go#L17-L22

As does the `CipherSuites` function:

https://github.com/openshift/hypershift/blob/6aea3a7b9d2c3f30418a28d0e3c8552c9bc239d9/support/config/cipher.go#L61-L74

As does the `MinTLSVersion` function:

https://github.com/openshift/hypershift/blob/6aea3a7b9d2c3f30418a28d0e3c8552c9bc239d9/support/config/cipher.go#L33-L43

Other component packages are relying on the callee to check for nils for this case as well, instead of passing the responsibility on to the caller. Thus, this is consistent with the code base.

https://github.com/openshift/hypershift/blob/b9aa6acc9fb97b9cee4237dfd2094dd825b5431e/control-plane-operator/controllers/hostedcontrolplane/v2/machine_approver/deployment.go#L17-L27

https://github.com/openshift/hypershift/blob/ac88b5c1c04e334ce198154903e365d28d43cb53/control-plane-operator/controllers/hostedcontrolplane/v2/clusterpolicy/config.go#L46-L48

https://github.com/openshift/hypershift/blob/b9aa6acc9fb97b9cee4237dfd2094dd825b5431e/control-plane-operator/controllers/hostedcontrolplane/v2/kube_scheduler/deployment.go#L22-L29

https://github.com/openshift/hypershift/blob/f6e4761ceb268dc5bf3ac8593e9259f12c621eba/control-plane-operator/controllers/hostedcontrolplane/v2/oauth_apiserver/deployment.go#L48-L53

and more.

hongkailiu: :resolved

I worried about "Could any of these be `nil`?"

* `cpContext.HCP`
* `cpContext.HCP.Spec.Configuration`

But I got it now:

https://github.com/openshift/hypershift/blob/4571f71f92e1302fcac7587d00cfe8f7d85783ad/control-plane-operator/controllers/hostedcontrolplane/v2/cvo/deployment.go#L51-L53

The above lines are ahead of us. So `cpContext.HCP` cannot be `nil` when Line 104 is executed. Thus it is safe to use `cpContext.HCP.Spec.Configuration`. There are many other places in the same function to use `cpContext.HCP.Spec` and `cpContext.HCP` is not `nil` seems taken granted.

But `cpContext.HCP.Spec.Configuration` could be `nil` and so could `configuration`. 

Then I thought that Line 111 and Line 114 were not safe to call `configuration.GetTLSSecurityProfile()` when `configuration` is `nil`. However I was wrong about it. I totally forgot that "A nil struct pointer can call a method in Go" which is very uncommon in object-oriented languages. ^_^

control-plane-operator/controllers/hostedcontrolplane/v2/cvo/deployment.go
```diff
@@ -102,12 +102,20 @@ func (cvo *clusterVersionOperator) adaptDeployment(cpContext component.WorkloadC
...
 			Name:  "RELEASE_IMAGE",
 			Value: dataPlaneReleaseImage,
 		})
 
+		if tlsMinVersion := config.MinTLSVersion(configuration.GetTLSSecurityProfile()); tlsMinVersion != "" {
+			c.Args = append(c.Args, fmt.Sprintf("--tls-min-version=%s", tlsMinVersion))
+		}
+		if cipherSuites := config.CipherSuites(configuration.GetTLSSecurityProfile()); len(cipherSuites) != 0 {
+			c.Args = append(c.Args, fmt.Sprintf("--tls-cipher-suites=%s", strings.Join(cipherSuites, ",")))
+		}
```
cblecker: The logic looks correct — follows the same pattern as kube-scheduler, machine-approver, and other v2 components.

One thing I noticed: all 5 test fixtures exercise the same TLS code path (nil Configuration → Intermediate default), so they all produce identical `--tls-min-version=VersionTLS12` + cipher suites output. Might be worth adding a fixture variant (or unit test in `deployment_test.go`) that sets a Modern TLS profile — that's the interesting edge case where `CipherSuites()` returns empty and `--tls-cipher-suites` should be omitted. The `oauth_apiserver` component has this kind of coverage as a reference.

Not blocking — the underlying `config.MinTLSVersion()` / `config.CipherSuites()` functions are well-tested in `support/config/cipher_test.go`, so the risk is low. Just would be nice to have coverage at the component level too.

cblecker (commented):
Code looks good — correct, nil-safe, and consistent with how other v2 components handle TLS configuration. Left one suggestion about test coverage.

control-plane-operator/controllers/hostedcontrolplane/v2/cvo/deployment.go
```diff
@@ -102,12 +102,20 @@ func (cvo *clusterVersionOperator) adaptDeployment(cpContext component.WorkloadC
...
 	})
 
+	configuration := cpContext.HCP.Spec.Configuration
 	podspec.UpdateContainer(ComponentName, deployment.Spec.Template.Spec.Containers, func(c *corev1.Container) {
 		podspec.UpsertEnvVar(c, corev1.EnvVar{
 			Name:  "RELEASE_IMAGE",
 			Value: dataPlaneReleaseImage,
 		})
 
+		if tlsMinVersion := config.MinTLSVersion(configuration.GetTLSSecurityProfile()); tlsMinVersion != "" {
```
jparrill: nit: `configuration.GetTLSSecurityProfile()` is called twice — once for min version, once for cipher suites. It's a trivial accessor so no real performance concern, but extracting it to a variable would be cleaner:

```go
tlsProfile := configuration.GetTLSSecurityProfile()
if tlsMinVersion := config.MinTLSVersion(tlsProfile); tlsMinVersion != "" {
    ...
}
if cipherSuites := config.CipherSuites(tlsProfile); len(cipherSuites) != 0 {
    ...
}
```

That said, `machine_approver`, `kube_scheduler`, and `kcm` all do the same double-call, so if you'd rather stay consistent with the existing pattern that's fine too. Your call.

control-plane-operator/controllers/hostedcontrolplane/v2/cvo/deployment.go
```diff
@@ -102,12 +102,20 @@ func (cvo *clusterVersionOperator) adaptDeployment(cpContext component.WorkloadC
...
+			c.Args = append(c.Args, fmt.Sprintf("--tls-min-version=%s", tlsMinVersion))
+		}
+		if cipherSuites := config.CipherSuites(configuration.GetTLSSecurityProfile()); len(cipherSuites) != 0 {
```
jparrill: question: in `oauth_apiserver/deployment.go:52`, `--tls-min-version` is passed unconditionally (no `!= ""` guard), because `MinTLSVersion()` never returns an empty string thanks to the Intermediate fallback. Here you use the conditional guard, which is what `machine_approver` and `kube_scheduler` do. Both are correct since the fallback to Intermediate always yields `"VersionTLS12"`, but I wanted to confirm: is the intent to be more defensive so the CVO doesn't get an empty TLS flag if the profile somehow returns empty? If so, makes sense — it's safer than `oauth_apiserver`'s approach.

jparrill (commented):

Additionally, the Codecov failure is expected — `adaptDeployment` gained new lines (the TLS flag logic) that aren't directly exercised by unit tests. The coverage actually comes from the fixture-based `TestControlPlaneComponents` snapshot tests, but Codecov can't always attribute YAML output validation back to the specific source lines inside closures that produced it. The behavior IS tested — it's a reporting gap, not a coverage gap.

control-plane-operator/controllers/hostedcontrolplane/v2/cvo/deployment.go
DavidHurta: That's a good point; thank you.

I have added unit tests to verify the functionality for more combinations than the fixture testing verifies in https://github.com/openshift/hypershift/pull/8013/commits/d3064dbe60a139fa9780a2a5215599faf91ce6b9

control-plane-operator/controllers/hostedcontrolplane/v2/cvo/deployment.go
DavidHurta: Thank you for ensuring clean code!

Now done in https://github.com/openshift/hypershift/pull/8013/commits/b07e7035ddc38bf40e1374e8cf2d61846cff48c6

control-plane-operator/controllers/hostedcontrolplane/v2/cvo/deployment.go
DavidHurta: Yes, exactly, the intent is to be more defensive.

control-plane-operator/controllers/hostedcontrolplane/v2/cvo/deployment.go
cblecker: Thanks for adding the test cases! However, they all use `TLSProfileCustomType` with manually specified (or empty) cipher lists. That exercises the `securityProfile.Custom.Ciphers` branch in `CipherSuites()`, not the `configv1.TLSProfiles[securityProfile.Type].Ciphers` lookup path.

The Modern profile specifically is worth testing because its ciphers (`TLS_AES_128_GCM_SHA256`, etc.) are all TLS 1.3 ciphers that are commented out of `openSSLToIANACiphersMap` in `cipher.go`, so `CipherSuites()` returns empty through a different mechanism than Custom-with-empty-ciphers. A single test case with `Type: configv1.TLSProfileModernType` (no Custom field) expecting `--tls-min-version=VersionTLS13` and no `--tls-cipher-suites` flag would cover that.

DavidHurta: Thank you for the feedback! I initially used custom profile test cases to directly control the `Ciphers` and `minTLSVersion` field values, making it straightforward to exercise flag presence/absence.

I've now added a modern profile test case that exercises the `configv1.TLSProfiles[securityProfile.Type].Ciphers` lookup path and verifies the TLS 1.3 cipher filtering behavior you described.

cblecker (approved):
LGTM - the Modern profile test case exercises the `configv1.TLSProfiles` lookup path and TLS 1.3 cipher filtering as requested. Implementation follows established v2 component patterns cleanly.

DavidHurta:
/verified by CI

Unit tests ensure that flags are applied if applicable.

PR #8016, opened 2026-03-19
CNTRLPLANE-3037,OCPBUGS-78837: Introduce envtest for integration tests

clebs:
This PR introduces envtest and replicates the existing `TestOnCreateAPIUX` e2e test (but not replacing it yet).

Shifting part of the tests away from e2e into integration tests that do not require a full environment will help reduce the burden on our CI.

## Checklist:


test/envtest/create_cluster_test.go
```diff
@@ -0,0 +1,2334 @@
...
+
+	// Find the CRD manifests directory
+	_, thisFile, _, _ := runtime.Caller(0)
+	repoRoot := filepath.Join(filepath.Dir(thisFile), "..", "..")
+	crdDir := filepath.Join(repoRoot, "cmd", "install", "assets", "hypershift-operator", "zz_generated.crd-manifests")
+
+	// Start envtest with HyperShift CRDs
+	testEnv := &envtest.Environment{
+		CRDDirectoryPaths: []string{crdDir},
+		// Use TechPreviewNoUpgrade CRDs to include GCP validations
```
enxebre: >		// Use TechPreviewNoUpgrade CRDs to include GCP validations

where is this enforced? the path contains also default CRD version. I assume last one would win and it's circumstantially TechPreview? which is fragile.

Makefile
```diff
@@ -288,15 +295,25 @@ delegating_client:
 	mv ./cmd/infra/aws/delegating_client.txt ./cmd/infra/aws/delegating_client.go
 	$(GO) fmt ./cmd/infra/aws/delegating_client.go
 
+# Build setup-envtest from tools folder
+$(SETUP_ENVTEST): $(TOOLS_DIR)/go.mod
+	cd $(TOOLS_DIR); $(GO) build -tags=tools -o $(SETUP_ENVTEST) sigs.k8s.io/controller-runtime/tools/setup-envtest
+
+.PHONY: setup-envtest
+setup-envtest: $(SETUP_ENVTEST) ## Setup envtest binaries (etcd, kube-apiserver)
```
enxebre, line 303: I think we'll want to make this plumbing more flexible so we can test with different kas API versions

enxebre: also we'll want to also use ocp releases https://raw.githubusercontent.com/openshift/api/master/envtest-releases.yaml
https://github.com/openshift/api/blob/1f2fa3f09f4ea0d6c2b5a09ba8608b70c41f616f/tests/Makefile#L21-L28

test/envtest/create_cluster_test.go
enxebre: we'll also want to install all the capi CRDs.
Basically what --install command  with tech preview would apply.

Makefile
JoelSpeed: +1 Alberto

I don't care too much about running individual validation tests on older API servers, but I do think we need to at least take all CRDs and install them on the oldest OCP KAS that managed services are currently using (4.15 to my knowledge) and then as we start to enforce a skew policy and they upgrade to newer releases, we can bump that and hopefully get to a point where it is N-3 consistently going forward

JoelSpeed:
In openshift/api we made the choice to leverage YAML based testing for our integration tests. With the structured testing you currently have, you have no visibility into the marshalled content that is actually being sent to the API server. This can mask issues in your APIs (e.g. not having omitempty on a required field) so you may want to either us unstructured or switch to YAML as we do

You'll also want to look at the possibility of testing ratcheting validations. Over time you may want to make validation tighter, and with that you need a way to install older, invalid schemas and check the ratcheting.

It might be worth looking over [our test suite](https://github.com/openshift/api/tree/master/tests) for inspiration

Makefile
enxebre: we need ability to test against oldest ocp (4.15) and oldest vanilla kube (to be confirmed).

enxebre:
Let's adjust the PR to cover https://github.com/openshift/hypershift/pull/8016#discussion_r2965323972
Get that gating on CI.
Then follow up to address https://github.com/openshift/hypershift/pull/8016#issuecomment-4096917204

Makefile
JoelSpeed: We currently have 4.17+ OCP KAS in the index, shouldn't be too hard to push 4.15 and 4.16

Upstream [goes back to 1.28](https://github.com/kubernetes-sigs/controller-tools/blob/main/envtest-releases.yaml) (which is equiv to 4.15) so hopefully the minimum vanilla is at least that

test/envtest/create_cluster_test.go
clebs: I took the simplest approach here as first try. Now looking closer at the e2e I see I missed the whole filtering logic on assets.

I will use that package the same way the original test does when installing so we can select the feature set to test. Maybe for now with an env var defaulting to `TechPreviewNoUpgrade`.

Makefile
```diff
@@ -31,6 +31,13 @@ GITLINT := $(TOOLS_BIN_DIR)/$(GITLINT_DIST_DIR)/$(GITLINT_BIN)-bin
 
 PROMTOOL=$(abspath $(TOOLS_BIN_DIR)/promtool)
 
+# Setup envtest for running tests that require a Kubernetes API server
+# SETUP_ENVTEST_VER is the version of setup-envtest to use, matching the version in hack/tools/go.mod
+KUBEBUILDER_ENVTEST_KUBERNETES_VERSION ?= 1.34.0
```
JoelSpeed, line 36: 4.22 is 1.35, pretty sure we have 1.35 binaries published that you can use

clebs: I did take 1.34 because that is what hypershift is currently built with. If I should follow OCP k8s version instead I will look into this as part of the above effort to make the setup parametric.

test/envtest/create_cluster_test.go
```diff
@@ -0,0 +1,2358 @@
...
+	}
+}
+
+// hypershiftCRDs returns the HyperShift CRDs for the given feature set,
+// using the same annotation-based filtering as `hypershift install`.
+func hypershiftCRDs(featureSet string) []*apiextensionsv1.CustomResourceDefinition {
+	crdObjects := installassets.CustomResourceDefinitions(
+		func(path string, crd *apiextensionsv1.CustomResourceDefinition) bool {
+			if !strings.Contains(path, "hypershift-operator/") {
+				return false
```
enxebre, line 73: this excludes all the API CRDs which we would want to include as well

jparrill:
/retitle [CNTRLPLANE-3037](https://redhat.atlassian.net/browse/CNTRLPLANE-3037),OCPBUGS-78837: Introduce envtest for integration tests

jparrill:
The jira above was created to track this issue [OCPBUGS-78837](https://redhat.atlassian.net/browse/OCPBUGS-78837). Discussed on slack that this would be detected using this testEnv.

enxebre:
superseded by https://github.com/openshift/hypershift/pull/8034

clebs:
/close

An alternative PR merged #8089

PR #8010, opened 2026-03-19
CNTRLPLANE-2678: Add fetch-etcd-certs CPO subcommand for HCPEtcdBackup

jparrill:
## Summary

- Adds a new `fetch-etcd-certs` subcommand to the CPO binary that fetches `etcd-client-tls` Secret and `etcd-ca` ConfigMap from an HCP namespace and writes TLS certificates to disk
- Registers the command in CPO `commandFor()` (symlink dispatch) and `defaultCommand()` (subcommand dispatch)
- Adds symlink in `Dockerfile.dev` for dev image builds
- This subcommand will run as InitContainer 1 in the HCPEtcdBackup backup Job

This is the **OCP payload component** of [CNTRLPLANE-2678](https://redhat.atlassian.net/browse/CNTRLPLANE-2678) (HCPEtcdBackup Controller). The full controller implementation (reconciler, Job construction, NetworkPolicy, retention) will follow in subsequent PRs. This is shipped first as it affects the CPO binary in the OCP release payload.

## Related PRs
- https://github.com/openshift/hypershift/pull/8017
- https://github.com/openshift/hypershift/pull/8040

## Jira
- [CNTRLPLANE-2678](https://redhat.atlassian.net/browse/CNTRLPLANE-2678) - HCPEtcdBackup Controller: Reconciler, Job Construction, NetworkPolicy, and Retention
- Parent epic: [CNTRLPLANE-2676](https://redhat.atlassian.net/browse/CNTRLPLANE-2676) - HCPEtcdBackup CRD for OADP Integration

## Test plan

- [x] Unit tests: 7 table-driven test cases covering happy path, missing resources, missing keys, directory creation, file permissions (0600)
- [x] Command structure test: validates flag registration and required flag enforcement
- [x] Manual test: verified against a live dev cluster — fetched real etcd TLS certs and validated with `openssl`
- [x] Build verification: `go build ./etcd-backup/... ./control-plane-operator/...` passes

### How to test manually

```bash
# 1. Build the CPO binary
make control-plane-operator

# 2. Run against a cluster with a HostedCluster deployed
KUBECONFIG=<path-to-mgmt-kubeconfig> ./bin/control-plane-operator fetch-etcd-certs \
  --hcp-namespace=<hcp-namespace> \
  --output-dir=/tmp/etcd-certs-test

# 3. Verify the certs were written and are valid
openssl x509 -in /tmp/etcd-certs-test/etcd-client.crt -noout -subject -issuer
[...]

PR #8017, opened 2026-03-19
CNTRLPLANE-2684: CPO etcd-upload subcommand for cloud storage upload

jparrill:
## Summary

- Add `etcd-upload` subcommand to the control-plane-operator for uploading etcd snapshots to S3 and Azure Blob Storage
- Support conditional writes (IfNoneMatch) to prevent overwriting existing backups
- Support server-side encryption: SSE-KMS for S3, encryption scopes for Azure
- Azure authentication: `client-secret` (JSON file) and `managed-identity` (msi-dataplane via SecretProviderClass for ARO HCP), with `DefaultAzureCredential` fallback
- Restructure OADP integration tests under `test/integration/oadp/` with `cli/` and `upload/` subdirectories
- Fix Agent platform resource expectations in OADP CLI tests

## Epic

[CNTRLPLANE-2676](https://issues.redhat.com/browse/CNTRLPLANE-2676) - HCPEtcdBackup CRD for OADP Integration

## Dependencies

- [CNTRLPLANE-2677](https://redhat.atlassian.net/browse/CNTRLPLANE-2677) (API types + feature gate) - already merged

## Related PRs:
- https://github.com/openshift/hypershift/pull/8010
- https://github.com/openshift/hypershift/pull/8040

## For the reviewers

To run the upload integration tests you need:
- `make control-plane-operator` (builds the CPO binary)
- Authenticated `aws` CLI and/or `az` CLI

```bash
./test/integration/oadp/run.sh upload aws    # S3 only
./test/integration/oadp/run.sh upload azure  # Azure Blob only
./test/integration/oadp/run.sh upload        # Both (auto-detects)
```

The script creates all resources (buckets, KMS keys, SPs, storage accounts) and cleans them up on exit.

## Test plan

- [x] Unit tests for S3 uploader (5 cases)
- [x] Unit tests for Azure Blob uploader (6 cases)
- [x] Unit tests for Azure credential auth paths (4 cases)
- [x] Unit tests for CLI flags and factory (5 cases)
- [x] OADP CLI integration tests pass (all platforms)
- [x] Integration tests with real S3 bucket (`./test/integration/oadp/run.sh upload`)
- [x] Integration tests with real Azure Blob Storage (`./test/integration/oadp/run.sh upload`)

## Summary by CodeRabbit

* **New Features**
  * Added an etcd-upload CLI to upload snapshots to S3 or Azure Blob with key-prefix, optional KMS/encryption-scope, and multiple auth modes; exposes an etcd-upload subcommand.

* **Tests**
  * Unit tests for S3 and Azure uploaders and CLI behavior.
  * Integration tests and a runnable integration test script for real AWS/Azure uploads.

* **Chores**
  * Updated .gitignore to ignore .env and added Azure storage SDK dependency; integration test package reorganizations.


etcd-upload/s3_uploader.go
```diff
@@ -0,0 +1,82 @@
...
+		Body:        f,
+		IfNoneMatch: aws.String("*"),
+	}
+
+	if u.kmsKeyARN != "" {
+		input.ServerSideEncryption = types.ServerSideEncryptionAwsKms
+		input.SSEKMSKeyId = aws.String(u.kmsKeyARN)
+	}
+
+	if _, err := u.client.PutObject(ctx, input); err != nil {
```
sdminonne, line 80: I'm reporting here the output of our common friend:  any reason for not using `transfermanager.New(s3Client)` as [etcd-backup](https://github.com/openshift/hypershift/blob/7b71c50e19e0136091b52f685b5526ef7a2c9679/etcd-backup/etcdbackup.go#L129)

etcd-upload/s3_uploader.go
```diff
@@ -0,0 +1,82 @@
...
+		return nil, fmt.Errorf("--bucket is required for S3 storage type")
+	}
+	if region == "" {
+		return nil, fmt.Errorf("--region is required for S3 storage type")
+	}
+	if credentialsFile == "" {
+		return nil, fmt.Errorf("--credentials-file is required for S3 storage type")
+	}
+
+	cfg, err := config.LoadDefaultConfig(context.Background(),
```
sdminonne: context.Background() ... Gues the NewS3Uploader should take the caller context caller... (any timeout set)?
Unsure about this. Feel free to push back but happy to hear about it.
TY!

etcd-upload/s3_uploader.go
```diff
@@ -0,0 +1,82 @@
...
+		config.WithRegion(region),
+		config.WithSharedCredentialsFiles([]string{credentialsFile}),
```
sdminonne: is `credentialsFile` mandatory? Can't we fallaback to role-based? I think we do [elsewhere](https://github.com/openshift/hypershift/blob/7b71c50e19e0136091b52f685b5526ef7a2c9679/etcd-backup/etcdbackup.go#L110-L114) Am I wrong?

etcd-upload/etcdupload.go
```diff
@@ -0,0 +1,95 @@
...
+
+func NewStartCommand() *cobra.Command {
+	opts := options{}
+
+	cmd := &cobra.Command{
+		Use:          "etcd-upload",
+		Short:        "Upload an etcd snapshot to cloud storage",
+		SilenceUsage: true,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			ctx, cancel := signal.NotifyContext(cmd.Context(), os.Interrupt, syscall.SIGINT)
```
sdminonne: os.Interrupt and syscall.SIGINT aren't the same thing? Guess syscall.SIGTERM is needed.

sdminonne (changes requested):
Drop some questions and raise a couple of flags.

etcd-upload/s3_uploader.go
jparrill: Done! Migrated to `transfermanager.New(s3Client)` — consistent with `etcd-backup`. This gives us automatic multipart upload for files >5GB, parallel part uploads, and partial retry on failure.

Note: `transfermanager.PutObjectInput` doesn't support `IfNoneMatch` (conditional writes), but since we use timestamp-based key naming (`{prefix}/{unix_seconds}.db`), collisions are practically impossible. Worth the trade-off for the multipart benefits.

etcd-upload/s3_uploader.go
jparrill: Good catch, done! `NewS3Uploader` now takes `ctx context.Context` as its first parameter. The caller passes the context from `signal.NotifyContext` so cancellation propagates properly on SIGINT/SIGTERM.

etcd-upload/s3_uploader.go
jparrill: You're right, done\! `credentialsFile` is now optional. When empty, it falls back to the default AWS credential chain (environment variables, shared config, EC2 instance profile, IRSA, etc.) — same pattern as `etcd-backup`.

etcd-upload/etcdupload.go
jparrill: Correct, `os.Interrupt` is just `syscall.SIGINT` on Unix. Fixed — now using `syscall.SIGINT, syscall.SIGTERM` so the pod shuts down cleanly on both signals.

jparrill:
Addressed reviewer comments + rebase to catch Konflux fixes

test/integration/oadp/run.sh
```diff
@@ -0,0 +1,389 @@
...
+        err "Failed to create KMS key: ${key_json}"
+    fi
+    log "KMS key created: ${CREATED_AWS_KMS_KEY_ID}"
+
+    # Export env vars for tests
+    export ETCD_UPLOAD_TEST_AWS_BUCKET="${CREATED_AWS_BUCKET}"
+    export ETCD_UPLOAD_TEST_AWS_REGION="${CREATED_AWS_REGION}"
+    export ETCD_UPLOAD_TEST_AWS_CREDENTIALS_FILE="${CREATED_AWS_CREDS_FILE}"
+    export ETCD_UPLOAD_TEST_AWS_KMS_KEY_ARN="${key_arn}"
+}
```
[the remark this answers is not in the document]
jparrill: This is an integration test runner, not production code. We intentionally always create an explicit credentials file to test the `--credentials-file` flag path end-to-end. In IRSA/instance-profile environments these manual tests aren't run — the controller uses the default credential chain directly via `NewS3Uploader` with an empty credentials file. No change needed here.

test/integration/oadp/run.sh
```diff
@@ -0,0 +1,389 @@
...
+
+    # Create temporary service principal with Storage Blob Data Contributor role
+    local scope="/subscriptions/${AZURE_SUBSCRIPTION_ID}/resourceGroups/${CREATED_AZURE_RG}/providers/Microsoft.Storage/storageAccounts/${CREATED_AZURE_STORAGE_ACCOUNT}"
+    log "Creating temporary service principal..."
+    local sp_json
+    sp_json=$(az ad sp create-for-rbac \
+        --name "etcd-upload-test-${SUFFIX}" \
+        --role "Storage Blob Data Contributor" \
+        --scopes "${scope}" \
+        -o json 2>/dev/null)
```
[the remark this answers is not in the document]
jparrill: Good catch! `az ad sp delete` only removes the service principal but leaves the app registration behind. Added `az ad app delete --id` after each `az ad sp delete` in both the main teardown and the safety-net loop. Fixed in the latest push.

etcd-upload/azure_uploader.go
```diff
@@ -0,0 +1,138 @@
...
+	client          AzureBlobUploadAPI
+}
+
+// AzureBlobUploadAPI defines the Azure Blob client interface used by the uploader.
+type AzureBlobUploadAPI interface {
+	UploadFile(ctx context.Context, containerName string, blobName string, file *os.File, o *azblob.UploadFileOptions) (azblob.UploadFileResponse, error)
+}
+
+// azureCredentialsFile represents the Azure credentials JSON file format.
+type azureCredentialsFile struct {
```
bryan-cox, line 38: @jparrill @tony-schndr - what credential is this in ARO HCP? Like where does it come from? How is it plumbed here? Can you help me understand the architecture flow a bit better please?

jparrill: Good question, from my side:

- This PR delivers the etcd-upload CLI subcommand - the credential plumbing is not wired yet. That's the controller's responsibility (CNTRLPLANE-2678), which will create the backup Jobs and mount the appropriate credentials.
- For ARO HCP, we know the CPO itself authenticates via `msi-dataplane.NewUserAssignedIdentityCredential` with a certificate mounted through a SecretProviderClass + CSI driver. The etcd-upload command currently supports ClientSecretCredential (via JSON file) and DefaultAzureCredential as fallback - neither covers the ARO HCP auth path. This is a known gap that will need to be addressed when implementing the controller:
  1. A managed (I assume new because should be provided by ARO ppl) identity for the etcd-upload Job
  2. A SecretProviderClass for the Job pod to mount the cert from Key Vault
  3. Either adapting azure_uploader.go to support msi-dataplane credentials (in this PR, adding support to managed identities as auth-type)

- For self-managed Azure, the CPO currently has no Azure credentials at all, so this is a broader gap that predates this work.

etcd-upload/etcdupload.go
```diff
@@ -0,0 +1,95 @@
...
+	_ = cmd.MarkFlagRequired("key-prefix")
+
+	return cmd
+}
+
+func run(ctx context.Context, opts options) error {
+	if _, err := os.Stat(opts.snapshotPath); err != nil {
+		return fmt.Errorf("snapshot file not accessible: %w", err)
+	}
+
```
muraee, line 75: should add a validation step to validate that S3 options are set when `--storage-type` is S3 and Azure options are set when `AzureBlob`

etcd-upload/azure_uploader.go
tony-schndr: As I understood in the proposal since HO was running the job in the HO namespace.  Therefore upload could use the managed identity that is attached to HO.  We would add RBAC so that HO can access the storage account.  @jparrill is my understanding correct?

ref: https://github.com/jparrill/enhancements/blob/eb0772d8ee369e83bbad6aa1d7710e6ce8145b34/enhancements/hypershift/etcd-backup-crd-for-oadp-integration.md?plain=1#L79-L82

etcd-upload/etcdupload.go
jparrill: The validation already exists inside the provider-specific constructors: `NewS3Uploader` validates `bucket` and `region` ([s3_uploader.go:33-38](https://github.com/openshift/hypershift/pull/8017/files#diff-s3_uploader.go)), and `NewAzureBlobUploader` validates `container` and `storage-account` ([azure_uploader.go:52-57](https://github.com/openshift/hypershift/pull/8017/files#diff-azure_uploader.go)). These are called from `newUploader()` before any upload attempt, so the user gets a clear error immediately. Adding a separate validation step in `run()` would duplicate the same checks.

etcd-upload/azure_uploader.go
jparrill: Yes, your understanding is correct. The Job runs in the HO namespace, so it could leverage the HO's identity. We've just added `--auth-type managed-identity` support in this PR, which uses `msi-dataplane.NewUserAssignedIdentityCredential` to consume the certificate mounted via SecretProviderClass — so the etcd-upload binary is ready for this path.

However, the HO identity today does **not** have storage permissions in either platform, so we'll need to extend them as part of the controller work (CNTRLPLANE-2678):

**ARO HCP (Azure Blob Storage):** We'd need to assign `Storage Blob Data Contributor` role to the HO's managed identity, scoped to the target storage account/container.

**ROSA HCP (S3):** The HO's OIDC S3 credentials only cover OIDC bucket operations. We'd need to add S3 permissions for the backup bucket: `s3:PutObject`, `s3:CreateMultipartUpload`, `s3:UploadPart`, `s3:CompleteMultipartUpload`, `s3:AbortMultipartUpload` (required by the transfer manager for multipart uploads), plus `kms:GenerateDataKey` and `kms:Decrypt` if using SSE-KMS.

Whether we extend the existing HO identity or create a dedicated one for etcd backups is a decision for CNTRLPLANE-2678.

bryan-cox: > Therefore upload could use the managed identity that is attached to HO

@tony-schndr @jparrill there is no such managed identity today, is that correct?

jparrill: Right, for the etcd backup job in ARO HCP, we'd need to create a new managed identity with Storage Blob Data Contributor role + a SecretProviderClass to mount the credential, following the same pattern as we do with CPO.

muraee:
lgtm

jparrill:
rebased due to `make verify` change from a non-related topic.

etcd-upload/azure_uploader.go
jparrill: @tony-schndr WDTY?

tony-schndr: I think this is fine, I don't see a problem with adding the managed identity.

sdminonne (approved):
Much better now!
 TY @jparrill !

jparrill:
Integration tests:
```
Σ hypershift-CNTRLPLANE-2684 git:(CNTRLPLANE-2684) ./test/integration/oadp/run.sh upload
==> Using AWS region: us-west-2
==> Creating temporary S3 bucket: etcd-upload-test-602175...
==> Created temporary credentials file: /tmp/etcd-upload-aws-creds.pIULAS
==> Creating temporary KMS key...
==> KMS key created: 3e05e83c-c252-45cf-ac18-ddeaf815b131
==> Using Azure subscription: 5f99720c-6823-4792-8a28-69efb0719eea
==> Creating temporary resource group: etcd-upload-test-602175...
==> Creating temporary storage account: etcdtest602175...
==> Creating temporary container: etcd-upload-test...
==> Creating temporary service principal...
==> Service principal created: 9b99e121-8d27-4088-8255-84181d887402
==> Creating temporary encryption scope: etcd-upload-test-scope...
==> Waiting for Azure RBAC propagation (up to 200s)...
==> RBAC not ready yet, retrying in 20s (attempt 1/10)...
==> RBAC not ready yet, retrying in 20s (attempt 2/10)...
==> RBAC not ready yet, retrying in 20s (attempt 3/10)...
https://etcdtest602175.blob.core.windows.net/etcd-upload-test/.rbac-probe/1774602291.VzMAUy
==> RBAC propagation confirmed (attempt 4/10)
==> Running etcd-upload integration tests...
=== RUN   TestEtcdUploadS3
=== RUN   TestEtcdUploadS3/When_uploading_to_S3_without_KMS_it_should_succeed_and_return_an_S3_URL
=== RUN   TestEtcdUploadS3/When_uploading_to_S3_with_KMS_it_should_succeed_and_return_an_S3_URL
--- PASS: TestEtcdUploadS3 (2.42s)
    --- PASS: TestEtcdUploadS3/When_uploading_to_S3_without_KMS_it_should_succeed_and_return_an_S3_URL (1.19s)
    --- PASS: TestEtcdUploadS3/When_uploading_to_S3_with_KMS_it_should_succeed_and_return_an_S3_URL (1.23s)
=== RUN   TestEtcdUploadAzureBlob
=== RUN   TestEtcdUploadAzureBlob/When_uploading_to_Azure_Blob_without_encryption_it_should_succeed_and_return_an_HTTPS_URL
=== RUN   TestEtcdUploadAzureBlob/When_uploading_to_Azure_Blob_with_encryption_scope_it_should_succeed_and_return_an_HTTPS_URL
--- PASS: TestEtcdUploadAzureBlob (3.55s)
    --- PASS: TestEtcdUploadAzureBlob/When_uploading_to_Azure_Blob_without_encryption_it_should_succeed_and_return_an_HTTPS_URL (1.69s)
    --- PASS: TestEtcdUploadAzureBlob/When_uploading_to_Azure_Blob_with_encryption_scope_it_should_succeed_and_return_an_HTTPS_URL (1.86s)
PASS
ok      github.com/openshift/hypershift/test/integration/oadp/upload    6.407s
==> Cleaning up test resources...
==> Scheduling KMS key 3e05e83c-c252-45cf-ac18-ddeaf815b131 for deletion (7 days)...
==> Deleting S3 bucket etcd-upload-test-602175 and its contents...
delete: s3://etcd-upload-test-602175/integration-test/with-kms/1774602295.db
delete: s3://etcd-upload-test-602175/integration-test/no-kms/1774602294.db
remove_bucket: etcd-upload-test-602175
==> Deleting service principal 9b99e121-8d27-4088-8255-84181d887402...
==> Deleting application registration 9b99e121-8d27-4088-8255-84181d887402...
==> Deleting resource group etcd-upload-test-602175 (cascades to all resources)...
==> Cleanup complete.
```

etcd-upload/etcdupload.go
```diff
@@ -0,0 +1,97 @@
...
+		SilenceUsage: true,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			ctx, cancel := signal.NotifyContext(cmd.Context(), syscall.SIGINT, syscall.SIGTERM)
+			defer cancel()
+
+			return run(ctx, opts)
+		},
+	}
+
+	cmd.Flags().StringVar(&opts.snapshotPath, "snapshot-path", "", "path to the etcd snapshot file to upload")
```
mgencur, line 48: Can we more visibly split the options that are for AWS and Azure. Not sure, maybe something like in the test here: https://github.com/openshift/hypershift/pull/8017/changes#diff-21de2e0953073bd89e1f4e73612f5b8b8eda9730ba1a287c66054e7e337f81e6R22

etcd-upload/etcdupload.go
```diff
@@ -0,0 +1,97 @@
...
+
+	fmt.Println(result.URL)
+	return nil
+}
+
+func newUploader(ctx context.Context, opts options) (Uploader, error) {
+	switch opts.storageType {
+	case "S3":
+		return NewS3Uploader(ctx, opts.bucket, opts.region, opts.credentialsFile, opts.kmsKeyARN)
+	case "AzureBlob":
```
mgencur, line 97: Will there be more uploader types over time?

jparrill: humm maybe I would say GCP at some point? not near...

etcd-upload/etcdupload.go
mgencur: If the link doesn't work, I mean what is in upload_test.go

jparrill: Yes, let me do it

mgencur:
/lgtm

(as there are minor changes requested by me since last time it was lgtm-ed)

bryan-cox:
> https://github.com/openshift/hypershift/pull/8017#issuecomment-4149921454

@jparrill this is going to keep failing. If you want to lgtm the PR I mentioned in Slack [here](https://redhat-internal.slack.com/archives/G01QS0P2F6W/p1774667783854999), I think we can get that test suite back up.

jparrill:
@bryan-cox done

PR #8040, opened 2026-03-23
CNTRLPLANE-2678: Adapt etcd-backup command and add HO namespace manifests

jparrill:
## Summary

- Adapts the `etcd-backup` CPO subcommand to work when the backup Job runs from the HO namespace, aligning cert paths with the `fetch-etcd-certs` subcommand from PR #8010.
- Removes S3 upload logic from `etcd-backup` (upload responsibility moves to a separate `etcd-upload` step).
- Adds new manifest functions for HO namespace deployment: `EtcdBackupJob`, `EtcdBackupJobServiceAccount`, `EtcdBackupJobRole`, `EtcdBackupJobRoleBinding`, `EtcdBackupNetworkPolicy`.
- Preserves backward compatibility with existing `EtcdBackupCronJob` and `EtcdBackupServiceAccount`.
- Adds integration tests validating both the new HO namespace flow (cross-namespace RBAC + NetworkPolicy + fetch-etcd-certs) and the legacy HCP namespace flow (direct volume mounts).

## Dependencies
- Depends on PR #8010 (`fetch-etcd-certs` subcommand)

## Related PRs
- https://github.com/openshift/hypershift/pull/8017

## Test plan

- [x] Unit tests for `etcd-backup` command flags and defaults
- [x] Unit tests for all new manifest functions
- [x] Integration tests validated against a live management cluster
  - [x] HO namespace flow: Job with fetch-etcd-certs + etcdctl snapshot + etcdutl verify
  - [x] HCP namespace flow (legacy): Job with direct volume mounts + etcdctl snapshot + etcdutl verify
- [x] `make verify` passes

## How to run integration tests

```bash
KUBECONFIG=/path/to/management-cluster/kubeconfig \
ETCD_BACKUP_TEST_HCP_NAMESPACE=clusters-my-hcp \
  go test -tags integration -v -timeout 10m ./test/integration/oadp/backup/...
```

Ref: [CNTRLPLANE-2678](https://redhat.atlassian.net/browse/CNTRLPLANE-2678)

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/manifests/etcd.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/manifests/etcd.go
@@ -1,9 +1,12 @@
 package manifests
 
 import (
+	"fmt"
+
 	appsv1 "k8s.io/api/apps/v1"
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
+	networkingv1 "k8s.io/api/networking/v1"
 	policyv1 "k8s.io/api/policy/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -115,6 +118,71 @@ func EtcdBackupCronJob(hcpNamespace string) *batchv1.CronJob {
 					},
 				},
 			},
+		},
+	}
+}
+
+// EtcdBackupJob returns a Job manifest for running an etcd backup from the
+// HyperShift Operator namespace. The controller populates the full PodSpec
+// with init containers (fetch-etcd-certs, etcd-backup) and main container
+// (etcd-upload).
+func EtcdBackupJob(ns string, hcpName string) *batchv1.Job {
+	return &batchv1.Job{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("etcd-backup-%s", hcpName),
+			Namespace: ns,
+			Labels: map[string]string{
+				"app":                         "etcd-backup",
+				"hypershift.openshift.io/hcp": hcpName,
+			},
+		},
+	}
+}
+
+// EtcdBackupJobServiceAccount returns a ServiceAccount for etcd backup Jobs
+// running in the given namespace. This SA is used by Jobs in the HO namespace.
+func EtcdBackupJobServiceAccount(ns string) *corev1.ServiceAccount {
+	return &corev1.ServiceAccount{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "etcd-backup-job",
+			Namespace: ns,
+		},
+	}
+}
+
+// EtcdBackupJobRole returns a Role in the HCP namespace granting read access
+// to etcd TLS secrets and CA configmaps needed by the fetch-etcd-certs init
+// container.
+func EtcdBackupJobRole(hcpNamespace string) *rbacv1.Role {
+	return &rbacv1.Role{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "etcd-backup-job",
+			Namespace: hcpNamespace,
+		},
+	}
+}
+
+// EtcdBackupJobRoleBinding binds the etcd-backup-job ServiceAccount (in the HO
+// namespace) to the Role in the HCP namespace, enabling cross-namespace access
+// to etcd TLS resources.
+func EtcdBackupJobRoleBinding(hcpNamespace string) *rbacv1.RoleBinding {
+	return &rbacv1.RoleBinding{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "etcd-backup-job",
+			Namespace: hcpNamespace,
+		},
+	}
+}
+
+// EtcdBackupNetworkPolicy returns a NetworkPolicy in the HCP namespace that
+// allows ingress from etcd backup Job pods (in the HO namespace) to etcd on
+// port 2379. This policy is created before the Job and cleaned up after it
+// completes.
+func EtcdBackupNetworkPolicy(hcpNamespace string) *networkingv1.NetworkPolicy {
+	return &networkingv1.NetworkPolicy{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "allow-etcd-backup",
+			Namespace: hcpNamespace,
 		},
 	}
 }
--- a/control-plane-operator/controllers/hostedcontrolplane/manifests/etcd_test.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/manifests/etcd_test.go
@@ -1,0 +1,81 @@
+package manifests
+
+import (
+	"testing"
+
+	. "github.com/onsi/gomega"
+)
+
+func TestEtcdBackupJobManifests(t *testing.T) {
+	tests := []struct {
+		name     string
+		testFunc func(g Gomega)
+	}{
+		{
+			name: "When creating an EtcdBackupJob it should set the correct name and namespace with HCP labels",
+			testFunc: func(g Gomega) {
+				job := EtcdBackupJob("ho-namespace", "my-hcp")
+				g.Expect(job.Name).To(Equal("etcd-backup-my-hcp"))
+				g.Expect(job.Namespace).To(Equal("ho-namespace"))
+				g.Expect(job.Labels).To(HaveKeyWithValue("app", "etcd-backup"))
+				g.Expect(job.Labels).To(HaveKeyWithValue("hypershift.openshift.io/hcp", "my-hcp"))
+			},
+		},
+		{
+			name: "When creating an EtcdBackupJobServiceAccount it should use the correct name and namespace",
+			testFunc: func(g Gomega) {
+				sa := EtcdBackupJobServiceAccount("ho-namespace")
+				g.Expect(sa.Name).To(Equal("etcd-backup-job"))
+				g.Expect(sa.Namespace).To(Equal("ho-namespace"))
+			},
+		},
+		{
+			name: "When creating an EtcdBackupJobRole it should target the HCP namespace",
+			testFunc: func(g Gomega) {
+				role := EtcdBackupJobRole("clusters-my-hcp")
+				g.Expect(role.Name).To(Equal("etcd-backup-job"))
+				g.Expect(role.Namespace).To(Equal("clusters-my-hcp"))
+			},
+		},
+		{
+			name: "When creating an EtcdBackupJobRoleBinding it should target the HCP namespace",
+			testFunc: func(g Gomega) {
+				rb := EtcdBackupJobRoleBinding("clusters-my-hcp")
+				g.Expect(rb.Name).To(Equal("etcd-backup-job"))
+				g.Expect(rb.Namespace).To(Equal("clusters-my-hcp"))
+			},
+		},
+		{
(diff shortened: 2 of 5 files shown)


jparrill:
/hold

This PR depends on #8010 (`fetch-etcd-certs` subcommand). Do not merge until that PR is merged.

control-plane-operator/controllers/hostedcontrolplane/manifests/etcd_test.go
```diff
@@ -0,0 +1,81 @@
...
+
+func TestEtcdBackupJobManifests(t *testing.T) {
+	tests := []struct {
+		name     string
+		testFunc func(g Gomega)
+	}{
+		{
+			name: "When creating an EtcdBackupJob it should set the correct name and namespace with HCP labels",
+			testFunc: func(g Gomega) {
+				job := EtcdBackupJob("ho-namespace", "my-hcp")
```
sdminonne, line 17: May we randomize a little bit the `ho-namespace` `my-hcp` string here or inside `EtcdBackupJob` with `Time.Now() or whatever mechanism. The day we re-try this it will fail miserably :)

test/integration/oadp/backup/etcd_backup_test.go
```diff
@@ -0,0 +1,553 @@
...
+	}
+}
+
+func newClient(t *testing.T) crclient.Client {
+	t.Helper()
+	scheme := runtime.NewScheme()
+	if err := clientgoscheme.AddToScheme(scheme); err != nil {
+		t.Fatalf("failed to add client-go scheme: %v", err)
+	}
+	if err := networkingv1.AddToScheme(scheme); err != nil {
```
sdminonne: Needed?

test/integration/oadp/backup/etcd_backup_test.go
```diff
@@ -0,0 +1,553 @@
...
+	"k8s.io/utils/ptr"
+
+	crclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+)
+
+const defaultHONamespace = "hypershift"
```
sdminonne: `nit: use an env var  ETCD_BACKUP_TEST_HO_NAMESPACE

sdminonne (approved):
Only minor stuff. Please @jparrill comment and eventually follow-up then it's good to me.

control-plane-operator/controllers/hostedcontrolplane/manifests/etcd_test.go
jparrill: These are unit tests for the manifest functions — they only validate the returned Go struct in memory, no resources are created in a cluster. The strings are deterministic input/output assertions, so there's no collision risk. Randomizing them would make the assertions harder to read without any benefit.

test/integration/oadp/backup/etcd_backup_test.go
jparrill: Yes, it's needed. The HO namespace test creates a `NetworkPolicy` (type `networkingv1.NetworkPolicy`) to allow cross-namespace etcd access. Without registering the networking scheme, the client can't create or read that resource.

test/integration/oadp/backup/etcd_backup_test.go
jparrill: Done. Added `ETCD_BACKUP_TEST_HO_NAMESPACE` env var with fallback to `"hypershift"`. Honestly the HO namespace has never changed throughout the product's lifecycle, but it doesn't hurt to make it configurable as long as the default remains `"hypershift"`.

jparrill:
Addressed reviewer comments + rebase to catch Konflux fixes

jparrill:
/hold cancel

Dep PR merged.
Also, rebased due to the change on `make verify` target.

control-plane-operator/controllers/hostedcontrolplane/manifests/etcd.go
```diff
@@ -118,3 +121,68 @@ func EtcdBackupCronJob(hcpNamespace string) *batchv1.CronJob {
...
 }
+
+// EtcdBackupJob returns a Job manifest for running an etcd backup from the
+// HyperShift Operator namespace. The controller populates the full PodSpec
+// with init containers (fetch-etcd-certs, etcd-backup) and main container
+// (etcd-upload).
+func EtcdBackupJob(ns string, hcpName string) *batchv1.Job {
+	return &batchv1.Job{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("etcd-backup-%s", hcpName),
```
sdminonne, line 132: HCP name in  `ROSA` can be pretty long. kubename is 253 length char here we're using 13 chars just for prefix.
I would double check with @lucasponce or @andreadecorte or anyone of that gang :)

test/integration/oadp/backup/etcd_backup_test.go
sdminonne: networking should be already in scheme at line 80.
-> https://github.com/openshift/hypershift/blob/main/vendor/k8s.io/client-go/kubernetes/scheme/register.go#L122

sdminonne: doesn't harm but I'd remove it

etcd-backup/etcdbackup_test.go
```diff
@@ -2,12 +2,41 @@ package etcdbackup
...
 	"github.com/aws/aws-sdk-go-v2/aws"
 )
 
+func TestNewStartCommand(t *testing.T) {
+	g := NewWithT(t)
+	cmd := NewStartCommand()
+
+	g.Expect(cmd.Use).To(Equal("etcd-backup"))
+
+	for _, flag := range []string{"backup-dir", "etcd-endpoint", "etcd-client-cert", "etcd-client-key", "etcd-ca-cert"} {
```
sdminonne: Is this somehow misaligned with https://github.com/openshift/hypershift/blob/main/etcd-backup/etcdbackup.go#L72-L75 ?

sdminonne (commented):
@jparrill may you double check?

etcd-backup/etcdbackup_test.go
jparrill: Fixed the three comments.

control-plane-operator/controllers/hostedcontrolplane/manifests/etcd.go
jparrill: Makes sense the comment. In my integration tests:
- HC Name: `integration-test`
- Job Name: `etcd-backup-integration-test`
- Pod Name: `etcd-backup-integration-test-gwscb`

CleanShot 2026-03-26 at 12 24 30

Let me know if it's too long to change it accordingly.

andreadecorte: we set the HostedCluster.Name to <= 15 chars, so we should be fine here.

test/integration/oadp/backup/etcd_backup_test.go
mgencur: It's interesting how many times we use word "hypershift" throughout the code base but we don't have a single const for it (until now):
```
ᐅ find . -name "*.go" | xargs grep "\"hypershift\"" | wc -l
57
ᐅ find . -name "*.go" | xargs grep "\"hypershift\"" | grep const | wc -l
0
```

test/integration/oadp/backup/etcd_backup_test.go
```diff
@@ -0,0 +1,556 @@
...
+	if err != nil {
+		t.Fatalf("failed to get etcd StatefulSet: %v", err)
+	}
+	for _, c := range sts.Spec.Template.Spec.Containers {
+		if c.Name == "etcd" {
+			return c.Image
+		}
+	}
+	t.Fatal("etcd container not found in StatefulSet")
+	panic("unreachable")
```
mgencur: Just for my understanding, why is this line needed? `t.Fatal` should really ensure we don't reach this point.

test/integration/oadp/backup/etcd_backup_test.go
```diff
@@ -0,0 +1,556 @@
...
+}
+
+// getCPOImage reads the CPO container image from the running control-plane-operator
+// deployment in the HCP namespace.
+func getCPOImage(ctx context.Context, t *testing.T, k8sClient crclient.Client, hcpNamespace string) string {
+	t.Helper()
+	dep := &appsv1.Deployment{}
+	err := k8sClient.Get(ctx, crclient.ObjectKey{Name: "control-plane-operator", Namespace: hcpNamespace}, dep)
```
mgencur: Same as with the hypershift namespace: There should be a const for this "control-plane-operator" as it's used 44 times in the codebase. But not in this PR.

test/integration/oadp/backup/etcd_backup_test.go
```diff
@@ -0,0 +1,556 @@
...
+	for _, obj := range objects {
+		if err := k8sClient.Delete(ctx, obj); crclient.IgnoreNotFound(err) != nil {
+			t.Logf("warning: failed to delete %T %s/%s: %v", obj, obj.GetNamespace(), obj.GetName(), err)
+		}
+	}
+}
+
+// waitForJobCompletion polls the given Job until it completes or fails.
+func waitForJobCompletion(ctx context.Context, k8sClient crclient.Client, job *batchv1.Job) error {
+	return wait.PollUntilContextTimeout(ctx, 5*time.Second, 5*time.Minute, true, func(ctx context.Context) (bool, error) {
```
mgencur: Is there any const for these timeouts and intervals that we could reuse?

test/integration/oadp/backup/etcd_backup_test.go
```diff
@@ -0,0 +1,556 @@
...
+				return false, fmt.Errorf("job failed: %s", cond.Message)
+			}
+		}
+		return false, nil
+	})
+}
+
+// etcdEndpointURL returns the etcd client service endpoint for a given HCP namespace.
+func etcdEndpointURL(hcpNamespace string) string {
+	return fmt.Sprintf("https://etcd-client.%s.svc.cluster.local:2379", hcpNamespace)
```
mgencur: Could we put a const in constants.go for the port?

We already have:
* https://github.com/openshift/hypershift/blob/main/support/config/constants.go#L30
* https://github.com/openshift/hypershift/blob/main/control-plane-operator/controllers/hostedcontrolplane/v2/kas/params.go#L88
Would be good to get this from one place

mgencur (commented):
Looks good, just a few minor things.

test/integration/oadp/backup/etcd_backup_test.go
jparrill: As discussed, this is just to satisfy the compiler. Changing this to `return ""` maybe clearer :).

gaol:
I have tested the changes in this PR, it looks good to me, the test plan and report is attached in the linked Jira comment.

One query about the `etcd-backup` subcommand are :
* The `etcd-backup` does the backup and upload altogether, what will the `etcd-upload` introduced in #8017 do ?
* Which image should the `etcd-backup` command runs on ? it needs both `control-plane-operator` and `etcdctl` available at the same time.
* 
@jparrill

gaol:
/verified by @gaol

I will tag the `verified` label to move it forward, I don't see the changes itself leading any issue.

PR #8079, opened 2026-03-26
CNTRLPLANE-2078: add unit tests for AWS Route53 and IAM

LiangquanLi930:
add unit tests for AWS Route53 and IAM

Fixes [CNTRLPLANE-2056](https://redhat.atlassian.net/browse/CNTRLPLANE-2056)

## Checklist:

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive Route53 tests for public/private hosted zone lookup, SOA updates, wildcard record cleanup, DNS destroy semantics, and aggregated error handling.
  * Added private-link Route53 tests for record find/create/delete behaviors and API-error propagation.
  * Added extensive IAM tests covering role and instance-profile create/destroy flows, policy attach/detach, OIDC provider handling, shared VPC role teardown, and related error paths.

### DIFF
--- a/cmd/infra/aws/destroy_iam_test.go
+++ b/cmd/infra/aws/destroy_iam_test.go
@@ -3,8 +3,9 @@ package aws
 import (
 	"context"
 	"errors"
-	"strings"
 	"testing"
+
+	. "github.com/onsi/gomega"
 
 	"github.com/openshift/hypershift/support/awsapi"
 
@@ -61,7 +62,7 @@ func TestDestroyOIDCRole(t *testing.T) {
 		{
 			name: "When the role does not exist it should return false without error",
 			setupMock: func(m *awsapi.MockIAMAPI) {
-				m.EXPECT().GetRole(gomock.Any(), gomock.Any(), gomock.Any()).
+				m.EXPECT().GetRole(gomock.Any(), &iam.GetRoleInput{RoleName: aws.String(roleName)}, gomock.Any()).
 					Return(nil, noSuchEntity())
 			},
 			expectRemoved: false,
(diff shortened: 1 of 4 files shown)


cmd/infra/aws/route53_test.go
```diff
@@ -0,0 +1,565 @@
...
+
+			o := &DestroyInfraOptions{
+				BaseDomain: testBaseDomain,
+				Name:       testCluster,
+				Log:        logr.Discard(),
+			}
+			errs := o.DestroyDNS(context.Background(), mockR53)
+
+			nonNilErrs := 0
+			for _, e := range errs {
```
sdminonne: Not sure I'm understanding this.
Instead of check the length you should check the actual errors. 

If for whatever reason and please exaplan me 'cause I don't get it 
we do need to check the number then check len(errs) and it's 1 check if it's nil and if it is not check the proper one.

cmd/infra/aws/destroy_iam_test.go
```diff
@@ -0,0 +1,568 @@
...
+
+// destroyIAMOptions returns a DestroyIAMOptions with the shared test infraID.
+func destroyIAMOptions() *DestroyIAMOptions {
+	return &DestroyIAMOptions{
+		InfraID: testInfraID,
+		Log:     logr.Discard(),
+	}
+}
+
+func TestDestroyOIDCRole(t *testing.T) {
```
sdminonne, line 51: This is assuming InOrder calls right? No way to add `gomock.InOrder()` ?

cmd/infra/aws/destroy_iam_test.go
```diff
@@ -0,0 +1,568 @@
...
+		setupVPCOwnerMock     func(*awsapi.MockIAMAPI)
+		expectError           bool
+		errorContains         string
+	}{
+		{
+			name:                  "When PrivateZonesInClusterAccount is false ingress role should use vpcOwnerClient",
+			privateZonesInCluster: false,
+			setupIAMMock:          func(_ *awsapi.MockIAMAPI) {},
+			setupVPCOwnerMock: func(m *awsapi.MockIAMAPI) {
+				// ingress role not found on vpcOwner
```
sdminonne: Instead of commenting `ingress role` there's no way to replace `gomock.Any` with the proper value? And check the order at the same time?

cmd/infra/aws/destroy_iam_test.go
```diff
@@ -0,0 +1,568 @@
...
+				m.EXPECT().GetRole(gomock.Any(), gomock.Any(), gomock.Any()).
+					Return(nil, noSuchEntity())
+				// control-plane role not found on vpcOwner
+				m.EXPECT().GetRole(gomock.Any(), gomock.Any(), gomock.Any()).
```
sdminonne: Same here, instead of commeting, check the value and the order

sdminonne (changes requested):
First pass done. Please follow-up

cmd/infra/aws/destroy_iam_test.go
```diff
@@ -0,0 +1,568 @@
...
+
+			o := destroyIAMOptions()
+			removed, err := o.DestroyOIDCRole(context.Background(), mockIAM, "openshift-ingress")
+
+			if tt.expectError {
+				if err == nil {
+					t.Fatal("expected error, got nil")
+				}
+				if tt.errorContains != "" && !strings.Contains(err.Error(), tt.errorContains) {
+					t.Errorf("expected error containing %q, got: %v", tt.errorContains, err)
```
jparrill: Use gomega instead of if + t.error or t.fatal.

cmd/infra/aws/destroy_iam_test.go
```diff
@@ -0,0 +1,568 @@
...
+			} else {
+				if err != nil {
+					t.Errorf("expected no error, got: %v", err)
+				}
+			}
+		})
+	}
+}
+
+func TestDestroyOIDCResources(t *testing.T) {
```
jparrill, line 386: Humm, it's there other tests covering `CreateOIDCResources`, `CreateSharedOIDCRole`, `DestroyOIDCRoleWithRetry`?

cmd/infra/aws/destroy_iam_test.go
```diff
@@ -0,0 +1,568 @@
...
+			setupMock: func(m *awsapi.MockIAMAPI) {
+				m.EXPECT().ListOpenIDConnectProviders(gomock.Any(), gomock.Any(), gomock.Any()).
+					Return(&iam.ListOpenIDConnectProvidersOutput{
+						OpenIDConnectProviderList: []iamtypes.OpenIDConnectProviderListEntry{
+							{Arn: aws.String(otherARN)},
+						},
+					}, nil)
+				// All role lookups return not found — covers shared-role + 9 component roles
+				m.EXPECT().GetRole(gomock.Any(), gomock.Any(), gomock.Any()).
+					Return(nil, noSuchEntity()).AnyTimes()
```
jparrill: This could mask bugs, this is not saying how many times the GetRole() function is called.

jparrill (commented):
Dropped some comments, thanks!

cmd/infra/aws/destroy_iam_test.go
sdminonne: Good point @jparrill . Replacing gomock.Any() with actual value will permit to track the proper call.

cmd/infra/aws/destroy_iam_test.go
LiangquanLi930: Good catch. `DestroyOIDCRoleWithRetry` is a thin retry wrapper around `DestroyOIDCRole` (which is fully covered by `TestDestroyOIDCRole`), and it's also exercised indirectly through `TestDestroyOIDCResources` since `DestroyOIDCResources` calls it. `CreateOIDCResources` and `CreateSharedOIDCRole` are not currently covered — I'll track those as follow-up work in a separate PR.

cmd/infra/aws/route53_test.go
LiangquanLi930: Fixed. Replaced `expectErrCount int` with `expectError bool` + `errorContains string` in both `TestDestroyDNS` and `TestDestroyPrivateZones`. The assertion now collects non-nil errors and checks the actual error message via gomega's `ContainSubstring`, rather than just counting.

cmd/infra/aws/destroy_iam_test.go
LiangquanLi930: Fixed. All multi-step mock sequences in `TestDestroyOIDCRole` (and throughout the file) now use `gomock.InOrder()` to explicitly enforce call ordering.

cmd/infra/aws/destroy_iam_test.go
sdminonne: @LiangquanLi930 may you let me know what you think about this? I think gomock.Any should be replaced with the proper value

LiangquanLi930:
@jparrill @enxebre When you have time, could you help to approve this PR?

cmd/infra/aws/route53_test.go
```diff
@@ -0,0 +1,581 @@
...
+			if tt.useCtx != nil {
+				ctx = tt.useCtx()
+			}
+
+			o := &CreateInfraOptions{Region: "us-east-1"}
+			id, err := o.CreatePrivateZone(ctx, logr.Discard(), mockR53, tt.zoneName, tt.vpcID, tt.authorizeAssoc, mockVPCOwner, tt.initialVPC)
+
+			if tt.expectError {
+				if err == nil {
+					t.Fatal("expected error, got nil")
```
jparrill, line 301: Replace the t.Fatal/t.Errorf by gomega assertions.

cmd/infra/aws/destroy_iam_test.go
```diff
@@ -0,0 +1,602 @@
...
+		name          string
+		setupMock     func(*awsapi.MockIAMAPI)
+		expectError   bool
+		errorContains string
+	}{
+		{
+			name: "When OIDC provider matches infraID and shared role exists it should delete the provider and return early",
+			setupMock: func(m *awsapi.MockIAMAPI) {
+				gomock.InOrder(
+					m.EXPECT().ListOpenIDConnectProviders(gomock.Any(), gomock.Any(), gomock.Any()).
```
jparrill, line 405: The gomock.Any() matchers for GetRole, ListAttachedRolePolicies, ListRolePolicies, and DeleteRole inputs should be replaced with specific *iam.XxxInput structs, as was already done in TestDestroySharedVPCRoles. For example:

```
m.EXPECT().GetRole(gomock.Any(), &iam.GetRoleInput{
    RoleName: aws.String(testInfraID + "-shared-role"),
}, gomock.Any())
```

And for the .Times(10) cases, iterating over the actual role names instead:

```
for _, name := range []string{
    "shared-role", "openshift-ingress", "openshift-image-registry",
    "aws-ebs-csi-driver-controller", "cloud-controller", "node-pool",
    "control-plane-operator", "cloud-network-config-controller",
    "kms-provider", "karpenter",
} {
    m.EXPECT().GetRole(gomock.Any(), &iam.GetRoleInput{
        RoleName: aws.String(testInfraID + "-" + name),
    }, gomock.Any()).Return(nil, noSuchEntity())
}
```

This ensures the test validates which roles are being checked, not just how many times GetRole is called. This was already flagged by @sdminonne and fixed in TestDestroySharedVPCRoles but remains open here.

jparrill (commented):
I've dropped some comments for a follow up PR not to get those in by this one.

LiangquanLi930:
@jparrill Hmmm, Could you help to add ok-to-test label?

PR #8088, opened 2026-03-26
CNTRLPLANE-3070: Support KMS on self-managed Azure without affecting ARO HCP

bryan-cox:

Enables Azure Key Vault KMS encryption (etcd encryption at rest) for self-managed Azure HyperShift clusters using workload identity federation, without breaking the existing ARO HCP (managed Azure) KMS path that uses managed identities with CSI secret store volumes.

### Key Changes

**API**: Added `SelfManagedKMS` field (type `SelfManagedAzureKMS`) to `AzureKMSSpec` with a `ClientID` for the workload identity that has `Key Vault Crypto User` role on the Key Vault. CEL validation rules enforce mutual exclusivity between `kms` (managed) and `selfManagedKMS` (self-managed), and immutability once set.

**Control Plane Operator**:
- Azure KMS provider now supports both ARO HCP and self-managed paths
- ARO HCP: CSI secret store volume for managed identity (unchanged)
- Self-managed: token-minter sidecar + emptyDir cloud token volume for workload identity federation
- KMS cloud config secret sets federated identity fields (`UseWorkloadIdentityExtension`) for self-managed

**HyperShift Operator**:
- KMS config secret reconciliation branches on managed vs self-managed, returning an error when neither credential is configured
- Credential reconciliation for self-managed Azure workload identities

**CLI**:
- `create cluster azure`: Only sets `ManagedIdentity` KMS creds for managed Azure; self-managed uses `AzureKMSSpec.SelfManagedKMS`
- `create iam azure`: Creates KMS workload identity with federated credential for `kms-provider` service account

**E2E**: Updated `TestCreateClusterCustomConfig` to handle self-managed Azure KMS assertions. Added envtest coverage for CEL validation rules (mutual exclusivity, immutability).

**Documentation**: Added KMS encryption section to self-managed Azure cluster guide with Key Vault setup and workload identity federation instructions.

Fixes [CNTRLPLANE-3070](https://redhat.atlassian.net/browse/CNTRLPLANE-3070)

The self-managed Azure KMS authentication pattern follows the same approach used by Cloud Controller Manager (CCM) and Azure CSI storage drivers, which already support self-managed Azure with workload identity federation.

The token-minter sidecar mints OIDC tokens for the `kms-provider` service account in `kube-system` namespace, matching the pattern used by AWS KMS.

`kms` and `selfManagedKMS` are mutually exclusive and immutable once set — switching between managed and self-managed KMS auth after cluster creation is not supported.

## Checklist:


api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -527,6 +527,7 @@ type Capabilities struct {
...
 // +kubebuilder:validation:XValidation:rule="self.networking.networkType == 'OVNKubernetes' || !has(self.operatorConfiguration) || !has(self.operatorConfiguration.clusterNetworkOperator) || !has(self.operatorConfiguration.clusterNetworkOperator.ovnKubernetesConfig)", message="ovnKubernetesConfig is forbidden when networkType is not OVNKubernetes"
+// +kubebuilder:validation:XValidation:rule="!has(self.secretEncryption) || !has(self.secretEncryption.kms) || self.secretEncryption.kms.provider != 'Azure' || self.platform.type != 'Azure' || !has(self.platform.azure) || self.platform.azure.azureAuthenticationConfig.azureAuthenticationConfigType != 'WorkloadIdentities' || has(self.platform.azure.azureAuthenticationConfig.workloadIdentities.kms)",message="workloadIdentities.kms is required when Azure KMS encryption is configured with WorkloadIdentities authentication"
```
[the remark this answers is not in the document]
bryan-cox: Fixed. Removed this CEL rule entirely — it was overly restrictive. The AzureKMSSpec-level rules already enforce mutual exclusivity and require at least one of kms or selfManagedKMS to be set. The top-level rule incorrectly blocked managed Azure (ARO HCP) clusters that use WorkloadIdentities auth with the managed kms field.

cmd/cluster/azure/create.go
```diff
@@ -383,21 +383,32 @@ func (o *CreateOptions) ApplyPlatformSpecifics(cluster *hyperv1.HostedCluster) e
...
+			azureKMSSpec.SelfManagedKMS = hyperv1.SelfManagedAzureKMS{
+				ClientID: hyperv1.AzureClientID(o.infra.KMSClientID),
+			}
+		} else {
+			// Managed Azure (ARO HCP) with managed identities
+			azureKMSSpec.KMS = hyperv1.ManagedIdentity{
+				CredentialsSecretName: o.KMSUserAssignedCredsSecretName,
+				ObjectEncoding:        ObjectEncoding,
+			}
+		}
```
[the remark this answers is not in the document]
bryan-cox: Fixed. Now explicitly checks `o.infra.WorkloadIdentities != nil` instead of relying solely on `KMSClientID`. Returns an actionable error if a self-managed cluster is missing the KMS identity.

cmd/infra/azure/identities.go
```diff
@@ -332,11 +357,20 @@ func (i *IdentityManager) createFederatedIdentityCredential(ctx context.Context,
...
 // CreateWorkloadIdentitiesFromIAMOptions creates all managed identities and federated credentials
 // for workload identity using CreateIAMOptions. This is used by the standalone IAM create command.
-func (i *IdentityManager) CreateWorkloadIdentitiesFromIAMOptions(ctx context.Context, l logr.Logger, opts *CreateIAMOptions, resourceGroupName string) (*hyperv1.AzureWorkloadIdentities, error) {
-	workloadIdentities := &hyperv1.AzureWorkloadIdentities{}
-	definitions := GetWorkloadIdentityDefinitions(opts.Name, "")
+func (i *IdentityManager) CreateWorkloadIdentitiesFromIAMOptions(ctx context.Context, l logr.Logger, opts *CreateIAMOptions, resourceGroupName string) (*IAMOutput, error) {
+	output := &IAMOutput{}
+	definitions := GetWorkloadIdentityDefinitions(opts.Name, WorkloadIdentityOptions{
+		IncludeKMS: opts.EnableKMS,
+	})
```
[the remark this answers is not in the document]
bryan-cox: Not applicable. `CreateWorkloadIdentitiesFromIAMOptions` is the standalone `hypershift create iam azure` command — it intentionally creates identities for all topologies since they're reusable across clusters. `CreateIAMOptions` doesn't have a `Topology` field because topology filtering happens at cluster creation time, not identity creation time. The empty `Topology` correctly includes the CPO identity per the comment at identities.go:169.

control-plane-operator/controllers/hostedcontrolplane/v2/kas/secretencryption_test.go
```diff
@@ -154,6 +155,125 @@ func TestReconcileKMSEncryptionConfigAWS(t *testing.T) {
...
+		})
+	}
+}
+
+func generateExpectedAzureEncryptionConfig(apiVersion string) *v1.EncryptionConfiguration {
+	activeKeyHash, _ := util.HashStruct(hyperv1.AzureKMSKey{
+		KeyVaultName: "test-vault",
+		KeyName:      "test-key",
+		KeyVersion:   "test-version",
+	})
```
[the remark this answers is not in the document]
bryan-cox: Fixed. Added `testing.TB` parameter with `t.Helper()` and `t.Fatalf` on error. Also added the v1 Azure encryption config test case for parity with the AWS tests.

docs/content/how-to/azure/azure-workload-identity-setup.md
```diff
@@ -62,6 +62,23 @@ This creates 7 managed identities with federated credentials for:
...
+```bash
+hypershift create iam azure \
+    --name $CLUSTER_NAME \
+    --infra-id $INFRA_ID \
+    --azure-creds $AZURE_CREDS \
+    --resource-group-name $PERSISTENT_RG_NAME \
+    --oidc-issuer-url $OIDC_ISSUER_URL \
+    --output-file workload-identities.json \
+    --enable-kms
+```
```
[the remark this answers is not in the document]
bryan-cox: Fixed. Added `--location $LOCATION` to the KMS example command.

docs/content/how-to/azure/create-self-managed-azure-cluster.md
```diff
@@ -267,6 +267,133 @@ ${HYPERSHIFT_BINARY_PATH}/hypershift create nodepool azure \
...
+```bash
+hypershift create iam azure \
+    --name "$CLUSTER_NAME" \
+    --infra-id "$INFRA_ID" \
+    --azure-creds "$AZURE_CREDS" \
+    --resource-group-name "$PERSISTENT_RG_NAME" \
+    --oidc-issuer-url "$OIDC_ISSUER_URL" \
+    --output-file ./workload-identities.json \
+    --enable-kms
+```
```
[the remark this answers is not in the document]
bryan-cox: Fixed. Added `--location "$LOCATION"` to the IAM creation example.

control-plane-operator/controllers/hostedcontrolplane/v2/kas/kms/azure_test.go
```diff
@@ -0,0 +1,951 @@
...
+		},
+		{
+			name: "When managed the KMS container should NOT have workload identity env vars",
+			check: func(g Gomega, podConfig *KMSPodConfig) {
+				for _, c := range podConfig.Containers {
+					if c.Name == "azure-kms-provider-active" {
+						for _, e := range c.Env {
+							g.Expect(e.Name).NotTo(Equal("AZURE_CLIENT_ID"),
+								"expected managed KMS container to NOT have AZURE_CLIENT_ID env var")
+						}
```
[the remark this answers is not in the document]
bryan-cox: Fixed. Replaced `t.Fatal` with `g.Expect(false).To(BeTrue(), ...)` in all closure locations where `g` is available.

control-plane-operator/controllers/hostedcontrolplane/v2/kas/kms/azure_test.go
```diff
@@ -0,0 +1,951 @@
...
+		for _, c := range podConfig.Containers {
+			if c.Name == "azure-kms-provider-backup" {
+				found = true
+				envMap := map[string]string{}
+				for _, e := range c.Env {
+					envMap[e.Name] = e.Value
+				}
+				g.Expect(envMap).To(HaveKeyWithValue("AZURE_CLIENT_ID", "test-client-id"))
+				g.Expect(envMap).To(HaveKeyWithValue("AZURE_TENANT_ID", "test-tenant-id"))
+				break
```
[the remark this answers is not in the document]
bryan-cox: Fixed. Added `AZURE_FEDERATED_TOKEN_FILE` assertion to the backup container test.

control-plane-operator/controllers/hostedcontrolplane/v2/kas/kms/azure.go
```diff
@@ -58,17 +59,36 @@ var (
...
+	TenantID         string
+	TokenMinterImage string
+}
+
+func NewAzureKMSProvider(kmsSpec *hyperv1.AzureKMSSpec, image string, opts AzureKMSProviderOptions) (*azureKMSProvider, error) {
 	if kmsSpec == nil {
 		return nil, fmt.Errorf("azure kms metadata not specified")
 	}
+	if opts.IsSelfManaged && (opts.KMSClientID == "" || opts.TenantID == "") {
+		return nil, fmt.Errorf("kmsClientID and tenantID are required for self-managed Azure KMS")
```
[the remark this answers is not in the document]
bryan-cox: Fixed. Added `TokenMinterImage` validation in `NewAzureKMSProvider` — returns an error when self-managed mode is used without a token minter image.

docs/content/how-to/azure/create-self-managed-azure-cluster.md
```diff
@@ -267,6 +267,134 @@ ${HYPERSHIFT_BINARY_PATH}/hypershift create nodepool azure \
...
+```bash
+hypershift create iam azure \
+    --name "$CLUSTER_NAME" \
+    --infra-id "$INFRA_ID" \
+    --azure-creds "$AZURE_CREDS" \
+    --location "$LOCATION" \
+    --resource-group-name "$PERSISTENT_RG_NAME" \
+    --oidc-issuer-url "$OIDC_ISSUER_URL" \
+    --output-file ./workload-identities.json \
+    --enable-kms
```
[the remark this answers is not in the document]
bryan-cox: Fixed. Added a note clarifying that `INFRA_ID` was set during the Azure Workload Identity Setup guide, with a link to the referenced doc.

hypershift-operator/controllers/hostedcluster/internal/platform/azure/azure_test.go
```diff
@@ -411,3 +413,116 @@ func TestReconcileCredentials(t *testing.T) {
...
+			},
+		},
+		{
+			name: "When self-managed Azure without workload identities it should not set identity fields",
+			hc:   baseHC(),
+			validate: func(g Gomega, cfg azurecloud.AzureConfig) {
+				g.Expect(cfg.UseWorkloadIdentityExtension).To(BeFalse())
+				g.Expect(cfg.AADClientID).To(BeEmpty())
+				g.Expect(cfg.AADMSIDataPlaneIdentityPath).To(BeEmpty())
+			},
```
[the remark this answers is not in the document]
bryan-cox: Not applicable. This is a unit test for the cloud config generation function (`generateAzureCloudConfig`), not an API validation test. The function must handle all inputs gracefully, including cases where neither `kms` nor `selfManagedKMS` is set. This test verifies the fallback behavior — that no identity fields are set when no KMS identity is configured. The API validation layer (CRD CEL rules) is what enforces that at least one must be set; this unit test validates the config generation logic in isolation.

bryan-cox:
I verified this works and the report verification report is located here - https://redhat.atlassian.net/browse/CNTRLPLANE-3070?focusedCommentId=16980328

control-plane-operator/controllers/hostedcontrolplane/v2/kas/kms/azure.go
```diff
@@ -200,6 +255,37 @@ func (p *azureKMSProvider) buildKASContainerAzureKMS(kmsKey hyperv1.AzureKMSKey,
 	}
 }
 
+func (p *azureKMSProvider) buildKASContainerAzureKMSTokenMinter() func(*corev1.Container) {
```
Nirshal, line 258: I noticed that here the token-minter is built inline rather than using the `TokenMinterContainerOptions` framework from `support/controlplane-component/token-minter-container.go`. I understand this is probably because the KAS pod has a different container injection pattern than the standalone cpov2 components, so the framework might not be directly applicable here.

That said, I was wondering about two things the framework provides to all other token-minters that seem missing here — is there a reason they're not needed in this case?

1. **Startup probe** — the framework adds a `test -f <token-path>` probe so dependent containers know the token file exists before they start. Here the KMS provider seems to rely on its own internal retries instead. Is that sufficient, or would adding a startup probe be a worthwhile safety net?
2. **Native sidecar support** — on K8s >= 1.29, the framework configures the token-minter as an init container with `RestartPolicy=Always`, which guarantees ordering and automatic restart. I'm not sure if this matters for the KAS pod specifically — maybe the ordering is already handled differently there?

Totally possible I'm missing some context about how the KAS pod lifecycle works and why these aren't needed here.

bryan-cox: Good observation. You're right that the `TokenMinterContainerOptions` framework isn't directly applicable here — the KAS KMS providers use a different assembly pattern via `KMSPodConfig`, which only carries `Containers` and `Volumes` (no init containers). The `NativeSidecarContainersEnabled` capability is also not plumbed through to this code path.

This is consistent with the existing AWS KMS token-minter (`aws.go:249`), which is built the same way and also lacks both the startup probe and native sidecar support.

Both are valid improvements — I've filed [CNTRLPLANE-3484](https://issues.redhat.com/browse/CNTRLPLANE-3484) to track adding startup probe and native sidecar support to all KMS token-minters uniformly.

Nirshal: Makes totally sense, thanks for the explanation, also, it's a good idea to prevent this PR to further grow, delegating to a new issue.

bryan-cox:
/verified by @bryan-cox 

Test Results Verification Report here - https://bryan-cox.github.io/architectural-artifact-sharing/test-verification-report/index.html

api/hypershift/v1beta1/azure.go
```diff
@@ -821,9 +824,19 @@ type AzureKMSSpec struct {
...
-	KMS ManagedIdentity `json:"kms"`
+	// +optional
+	KMS ManagedIdentity `json:"kms,omitzero"`
+
+	// selfManagedKMS contains the workload identity used to authenticate with Azure KMS
+	// on self-managed Azure clusters. The identity authenticates via federated tokens
+	// using a token-minter sidecar.
+	// kms and selfManagedKMS are mutually exclusive.
+	//
+	// +optional
```
enxebre, line 838: I'd rather don't introduce this naming pattern until we decide if we want to introduce a semantic for product and have a detailed plan for it.
Can we name this after the targeted auth mechanism e.g. kmsFederatedTokenIdentity

bryan-cox: Done. Renamed to `kmsFederatedTokenIdentity` / `AzureKMSFederatedTokenIdentity` across the board — API types, CEL rules, envtests, CLI, controllers, and docs.

cmd/install/assets/crds/hypershift-operator/tests/hostedclusters.hypershift.openshift.io/stable.hostedclusters.kms.testsuite.yaml
```diff
@@ -0,0 +1,477 @@
...
+          servicePublishingStrategy:
+            type: Route
+            route: {}
+        - service: Ignition
+          servicePublishingStrategy:
+            type: Route
+            route: {}
+
+  onUpdate:
+  - name: When switching from kms to kmsFederatedTokenIdentity it should fail
```
enxebre: can we have the opposite covered as well? from kmsFederatedTokenIdentity to kms

api/hypershift/v1beta1/azure.go
```diff
@@ -810,6 +810,9 @@ const (
 // AzureKMSSpec defines metadata about the configuration of the Azure KMS Secret Encryption provider using Azure key vault
 //
 // +kubebuilder:validation:XValidation:rule="!has(self.backupKey) || self.backupKey.keyVaultName == self.activeKey.keyVaultName",message="backupKey.keyVaultName must match activeKey.keyVaultName; both keys must reside in the same Key Vault"
+// +kubebuilder:validation:XValidation:rule="!(has(self.kms) && has(self.kmsFederatedTokenIdentity))",message="kms and kmsFederatedTokenIdentity are mutually exclusive"
+// +kubebuilder:validation:XValidation:rule="has(self.kms) || has(self.kmsFederatedTokenIdentity)",message="one of kms or kmsFederatedTokenIdentity must be set"
+// +kubebuilder:validation:XValidation:rule="has(self.kms) == has(oldSelf.kms)",message="kms is immutable, cannot switch between kms and kmsFederatedTokenIdentity"
```
enxebre: "kms is immutable" this is a bit confusing providing it matches the single field name

support/azureutil/azureutil.go
```diff
@@ -270,6 +270,14 @@ func IsSelfManagedAzure(platform hyperv1.PlatformType) bool {
 	return platform == hyperv1.AzurePlatform && !IsAroHCP()
 }
 
+// IsSelfManagedAzureWithWorkloadIdentity returns true if the platform is self-managed Azure
```
enxebre, line 273: what's the appropriate term FederatedTokenIdentity or workloadIdentity? seems we are using both to refer to the same thing?

cmd/install/assets/crds/hypershift-operator/tests/hostedclusters.hypershift.openshift.io/stable.hostedclusters.kms.testsuite.yaml
bryan-cox: Done. Added the reverse test case: "When switching from kmsFederatedTokenIdentity to kms it should fail" — mirrors the existing kms-to-kmsFederatedTokenIdentity test.

---

api/hypershift/v1beta1/azure.go
bryan-cox: Done. Updated the CEL message to "the KMS authentication mode is immutable once set" to avoid ambiguity with the `kms` field name.

---

support/azureutil/azureutil.go
bryan-cox: They refer to different layers — WorkloadIdentities is the "what" and FederatedTokenIdentity is the "how":

- **WorkloadIdentities** (`AzurePlatformSpec.AzureAuthenticationConfig.WorkloadIdentities`) is the platform-level auth config — it declares which Azure client IDs each operator uses (disk, network, ingress, etc.). It answers "what identity does each component use?"

- **KMSFederatedTokenIdentity** (`AzureKMSSpec.KMSFederatedTokenIdentity`) is the specific mechanism by which the KMS sidecar authenticates to Azure Key Vault using projected service account tokens federated against Entra ID. It answers "how does the KMS plugin authenticate?"

The `IsSelfManagedAzureWithWorkloadIdentity` function checks the parent platform-level config, while the `KMSFederatedTokenIdentity` field is the child-level detail specific to KMS.

---

enxebre: seems to me we are expressing intent for the same auth mechanism with different names

```
type WorkloadIdentity struct {
    // clientID is client ID of a federated managed identity used in workload identity authentication
    //
    // +required
    ClientID AzureClientID `json:"clientID"`
}
```

```
// AzureKMSFederatedTokenIdentity contains the identity used to authenticate with Azure KMS
// via federated tokens.
type AzureKMSFederatedTokenIdentity struct {
	// clientID is the client ID of a federated managed identity used for Azure Key Vault KMS encryption.
	// This identity must have "Key Vault Crypto User" role on the Key Vault containing the encryption key.
	//
	// +required
	ClientID AzureClientID `json:"clientID,omitempty"`
}
```

aren't this both intent for "here's a client ID for a federated managed identity that will authenticate via workload identity (token-minter)."?
If so, shouldn't AzureKMSSpec just have a new entry pointing to the existing WorkloadIdentity struct?

bryan-cox: Correction to my earlier reply — you're right, they express the same thing. Both WorkloadIdentity and the now-deleted AzureKMSFederatedTokenIdentity were structurally identical: a single ClientID for a federated managed identity that authenticates via workload identity.

Updated to reuse the existing WorkloadIdentity type directly:

KMSFederatedTokenIdentity WorkloadIdentity `json:"kmsFederatedTokenIdentity,omitzero"`
The AzureKMSFederatedTokenIdentity type has been deleted.

AI-assisted response via Claude Code

bryan-cox:
/verified by @bryan-cox 

I verified this works and the report verification report is located here - https://redhat.atlassian.net/browse/CNTRLPLANE-3070?focusedCommentId=16980328

api/hypershift/v1beta1/azure.go
```diff
@@ -821,9 +824,19 @@ type AzureKMSSpec struct {
 	BackupKey *AzureKMSKey `json:"backupKey,omitempty"`
 
 	// kms is a pre-existing managed identity used to authenticate with Azure KMS.
+	// This is used for managed Azure (ARO HCP) clusters.
+	// kms and workloadIdentity are mutually exclusive.
 	//
-	// +required
-	KMS ManagedIdentity `json:"kms"`
+	// +optional
+	KMS ManagedIdentity `json:"kms,omitzero"`
```
muraee, line 831: I know this field was added before, but why is it called `kms`? with the new field, it should be called `managedIdentity` or smth

muraee: `.spec.secretEncryption.kms.azure.kms` doesn't sound correct.

bryan-cox:

bryan-cox:

api/hypershift/v1beta1/azure.go
bryan-cox: I'm not sure how we can change this now since ARO HCP is in flight other than introducing a new field and trying to deprecated this.

PR #8111, opened 2026-03-30
CNTRLPLANE-3115: Add OCP 4.15/4.16 to envtest integration tests

JoelSpeed:

We understand that there are management clusters running HyperShift Operator with openshift versions as old as 4.15.
We must ensure we add integration testing going all the way back to 4.15 to catch any potential breaking API changes that we might make.

Fixes 

## Checklist:


JoelSpeed:
/close

Having double checked with all of our hosted cluster providers, we don't actually need to be testing below 4.18. We won't need to move forward with these versions

PR #8123, opened 2026-03-30
OCPBUGS-78145: Pin restricted-v2 SCC on all HCP control plane pod templates

dpateriya:
## Summary

- Add a default `openshift.io/required-scc: restricted-v2` annotation in the shared `setAnnotations()` method (`support/controlplane-component/defaults.go`) so every v2 Deployment/StatefulSet control plane pod gets SCC-pinned automatically.
- The annotation is only set when not already present, preserving any explicit overrides (e.g. `privileged` for CSI workloads).
- Remove the now-redundant explicit annotation from `cluster-storage-operator/deployment.yaml` asset since the framework default covers it.

## Problem

HCP control plane pods run in namespaces with PSA `enforce: restricted`, but lack the `openshift.io/required-scc` annotation. If a permissive SCC (e.g. `anyuid`) is granted to `system:authenticated` on the management cluster, the SCC admission controller selects it for unpinned pods, causing PSA violations and permanent `FailedCreate`.

## Test plan

- [x] Two new unit tests in `defaults_test.go`: default annotation applied, and explicit override preserved
- [x] Regenerated all `testdata/` golden fixtures — every component now includes the annotation
- [x] All existing tests pass (`go test ./control-plane-operator/controllers/hostedcontrolplane/...`)
- [x] Local CodeRabbit review: **no findings**

Bug: https://issues.redhat.com/browse/OCPBUGS-78145

## Summary by CodeRabbit

* **Security**
  * Control plane workloads now get a default security constraint annotation ("restricted-v2") when none or only whitespace is present; explicitly provided values are preserved.
* **Bug Fix / Cleanup**
  * Removed a redundant per-component annotation so the system relies on a single centralized default.
* **Tests**
  * Added tests verifying defaulting, preservation, and normalization of the security annotation.


dpateriya:
/test e2e-gke


<br class="Apple-interchange-newline">

PR #8136, opened 2026-03-31
[release-4.21] OCPBUGS-81489: Add MTU override for OVN-Kubernetes on hosted clusters

jparrill:
## Summary

Backport of #7942 to `release-4.21`.

- Adds an optional, immutable `mtu` field to `OVNKubernetesConfig` API type for hosted clusters
- Adds `--ovn-kubernetes-mtu` CLI flag to `hypershift create cluster`
- Propagates the MTU value to the `operator.openshift.io/v1 Network` resource via `ReconcileNetworkOperator`
- Needed for non-commercial AWS regions that don't support the default 9001 MTU

Squashed commits from #7942:
* `feat(api): add MTU field to OVNKubernetesConfig for hosted clusters`
* `chore(generated): regenerate CRDs, clients, vendor, and docs`
* `feat(cli): add --ovn-kubernetes-mtu flag to hypershift create cluster`
* `chore(generated): add TestCreateCluster fixture for OVNKubernetesMTU`

Ref: https://issues.redhat.com/browse/OCPBUGS-77078
Ref: https://issues.redhat.com/browse/OCPBUGS-81489

## Test plan

- [ ] CI passes on release-4.21
- [ ] Verify CRDs are correctly generated for the target branch
- [ ] Verify `--ovn-kubernetes-mtu` flag works with `hypershift create cluster`


api/hypershift/v1beta1/operator.go
```diff
@@ -62,13 +62,31 @@ type ClusterNetworkOperatorSpec struct {
...
+	// requiring this field to be explicitly set to a lower value.
+	// The maximum is 9216, which is the standard jumbo frame upper limit
+	// supported by datacenter and cloud network interfaces.
+	// The minimum is 576, which is the minimum IPv4 MTU per RFC 791.
+	// This field is immutable once set.
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="mtu is immutable once set"
+	// +kubebuilder:validation:Minimum=576
+	// +kubebuilder:validation:Maximum=9216
+	// +optional
+	MTU *int32 `json:"mtu,omitempty"`
```
muraee: why is this a pointer?

jparrill (commented):
Pointer is technically unnecessary here since `Minimum=576` makes the zero value invalid — there's no ambiguity between "unset" and "explicitly zero." However, `kubeapilinter` on this branch requires pointers for optional scalar fields — changing to plain `int32` fails `make verify`. On main this was relaxed but release-4.21 still enforces it.

api/hypershift/v1beta1/operator.go
jparrill: Pointer is technically unnecessary here since Minimum=576 makes the zero value invalid — there's no ambiguity between "unset" and "explicitly zero." However, kubeapilinter on this branch requires pointers for optional scalar fields - changing to plain int32 fails make verify. On main this was relaxed but release-4.21 still enforces it.

```
hypershift/v1beta1/operator.go:89:2: optionalfields: field MTU should be a pointer. (kubeapilinter)
      MTU int32 `json:"mtu,omitempty"`
      ^
```

PR #8166, opened 2026-04-06
CNTRLPLANE-3160: Drop AutoNodeKarpenter feature gate and promote EC2NodeClass to v1

enxebre:

Promotes the AutoNode/Karpenter feature to stable/GA by:

1. **Dropping the `AutoNodeKarpenter` feature gate** — the `autoNode` field on HostedCluster and HostedControlPlane is no longer gated behind TechPreviewNoUpgrade
2. **Promoting OpenshiftEC2NodeClass API from v1beta1 to v1** — signals API stability
3. **Adding envtest validation test suites** for both HostedCluster autoNode CEL rules and OpenshiftEC2NodeClass v1 CEL rules
4. **Removing the TECH_PREVIEW_NO_UPGRADE skip** from the karpenter e2e test

No runtime/controller code changes — all controllers use `IsKarpenterEnabled()` which checks spec values, not the feature gate.

Fixes https://issues.redhat.com/browse/CNTRLPLANE-3160

- The karpenter envtest tests are self-contained under `karpenter-operator/controllers/karpenter/assets/` with their own `zz_generated.crd-manifests/` and `tests/` directories. The only integration point is adding a second path to `LoadTestSuiteSpecs` in `test/envtest/suite_test.go`.
- The hypershift install path (`cmd/install/install.go`) is unaffected — the karpenter CRD is not placed in the hypershift-operator install assets.
- The v1beta1 → v1 migration has no struct changes, only the API version label changes. The karpenter-operator controls the CRD lifecycle so there is no multi-version concern.

## Checklist:

🤖 Generated with [Claude Code](https://claude.ai/claude-code) via `/jira:solve [CNTRLPLANE-3160](https://redhat.atlassian.net/browse/CNTRLPLANE-3160)`

## Summary by CodeRabbit

* **New Features**
  * Karpenter API graduates to stable v1.
  * Added comprehensive validation tests for OpenShift EC2 node classes.
  * New documentation describing control plane agent and component reconciliation frameworks.

* **Chores**
  * Switched references to stable Karpenter v1 across components and tests.
  * Normalized AutoNode configuration shape across APIs and tooling.
  * CI change-detection updated to include additional Karpenter test assets.
  * Makefile CRD output processing improved.

* **Bug Fixes**
  * Cleared/updated AutoNodeKarpenter entries in feature-gate configurations.


api/hypershift/v1beta1/hosted_controlplane.go
```diff
@@ -230,9 +230,8 @@ type HostedControlPlaneSpec struct {
 	// using the specified provisioner (e.g. Karpenter) instead of requiring
 	// manual NodePool management.
 	//
-	// +openshift:enable:FeatureGate=AutoNodeKarpenter
 	// +optional
-	AutoNode *AutoNode `json:"autoNode,omitempty"`
+	AutoNode AutoNode `json:"autoNode,omitzero"`
```
jparrill, line 234: Just to know more, is this change expected by promoting a featureGate or this is due to the new api linter?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -1418,7 +1417,7 @@ type KarpenterConfig struct {
 	//
 	// +optional
 	// +unionMember
-	AWS *KarpenterAWSConfig `json:"aws,omitempty"`
+	AWS KarpenterAWSConfig `json:"aws,omitzero"`
```
jparrill, line 1420: Same comment as above

api/.golangci.yml
```diff
@@ -202,11 +202,6 @@ linters:
           - kubeapilinter
         path: hypershift/v1beta1/openstack.go
         text: 'arrayofstruct: OpenStackPlatformSpec.Subnets is an array of structs, but the struct has no required fields. At least one field should be marked as required to prevent ambiguous YAML configurations'
-      - linters:
-          - kubeapilinter
-        path: karpenter/v1beta1/karpenter_types.go
```
jparrill, line 207: Should we add `v1` to this path?

cmd/install/assets/crds/hypershift-operator/tests/hostedclusters.hypershift.openshift.io/stable.hostedclusters.karpenter.testsuite.yaml
```diff
@@ -0,0 +1,261 @@
+apiVersion: apiextensions.k8s.io/v1
```
jparrill, line 1: It's there any docs to know how to add more testCases to this suite/framework?

jparrill (commented):
Dropped some comments, mostly to learn. 🙏

api/hypershift/v1beta1/hosted_controlplane.go
enxebre: is expected that any API follow best practices, specially if it's GA. The linter enforces those best practices.

cmd/install/assets/crds/hypershift-operator/tests/hostedclusters.hypershift.openshift.io/stable.hostedclusters.karpenter.testsuite.yaml
enxebre: https://github.com/openshift/hypershift/tree/main/test/envtest

api/.golangci.yml
enxebre: the exception is not actually needed, the struct already has // +kubebuilder:validation:MinProperties=1

enxebre: this triggered an offline discussion for me with Joel. This would still hit the linter on a new field or a field change. "Not having a specific requirement but at least one" is not an ideal choice when the struct belongs to a slice. The reasoning is that with all fields optional, a user could accidentally write two separate items thinking they're writing one (or the other way around). I think we'll accepted that here.

api/karpenter/v1/doc.go
```diff
@@ -1,4 +1,4 @@
 // +k8s:deepcopy-gen=package,register
 // +groupName=karpenter.hypershift.openshift.io
 // +k8s:openapi-gen=true
-package v1beta1
+package v1
```
JoelSpeed, line 4: By doing this as a rename, rather than creating a duplicate, anyone currently vendoring HyperShift cannot safely have an intermediate step where they update their HyperShift dependency, are able to compile, and then move over to the v1 version of this API.

We would normally expect to create a v1 dir alongside the v1alpha1 (or beta in this case)

A secondary benefit of that is then we get a full diff of the API being promoted and can give it a more thorough API review

api/.golangci.yml
JoelSpeed: To put the slice thing into an example
```
- foo: some
- bar: thing
```
vs
```
- foo: some
  bar: thing
```
Is a very subtle difference that you might glance over, but has a very different meaning to the API.

This has caught people out in security contexts before with pretty nasty results

test/e2e/karpenter_test.go
```diff
@@ -827,9 +824,7 @@ func TestKarpenter(t *testing.T) {
 
 			// AutoNode.Provisioner.Karpenter.AWS is required at the API level when karpenter
 			// is configured, so this should never happen/never be nil for a valid karpenter cluster.
-			if hostedCluster.Spec.AutoNode == nil ||
-				hostedCluster.Spec.AutoNode.Provisioner.Karpenter == nil ||
-				hostedCluster.Spec.AutoNode.Provisioner.Karpenter.AWS == nil {
+			if hostedCluster.Spec.AutoNode.Provisioner.Karpenter.AWS == (hyperv1.KarpenterAWSConfig{}) {
```
muraee: comparing to an empty struct doesn't seems right. Should we check `hostedCluster.Spec.AutoNode.Provisioner.Karpenter.Platform == AWS` instead?

JoelSpeed: Agree, use a discriminator field if it exists

PR #8198, opened 2026-04-09
AUTOSCALE-577: reduce karpenter e2e execution time

maxcao13:
Reduces Karpenter e2e test execution time in `TestKarpenter` by parallelizing independent subtests and eliminating unnecessary overhead:
1. **Convert unstructured YAML to typed Go structs** — Replace `unstructured.Unstructured` with `*karpenterv1.NodePool` and `*appsv1.Deployment`, giving compile-time type safety and removing map-based field access.
2. **Replace YAML file loading with in-memory constructors** — Introduce `baseNodePool()` and `testWorkload()` that build K8s resources as Go structs directly, removing the runtime YAML/embed dependency. Each subtest creates its own isolated objects instead of sharing and mutating top-level state.
3. **Parallelize independent provisioning subtests** — Wrap instance-profile, version-field, version-skew, capacity-reservation, and arbitrary-subnet tests in a parallel group so they run concurrently while preserving sequential ordering with tests before (ARM64, plumbing, upgrade) and after (AutoNode lifecycle, PDB blocking).
4. **Remove unnecessary deprovisioning waits and reduce replicas** — Drop wait-for-zero-nodes from parallel subtests (cleanup defers handle deletion, and isolated NodePools mean no cross-test interference). Reduce the PDB blocking test replicas from 3 to 1 and rename it to reflect its actual purpose.
Fixes https://issues.redhat.com/browse/AUTOSCALE-606
Fixes https://issues.redhat.com/browse/AUTOSCALE-577
- The `sigs.k8s.io/yaml` import is dropped from this file entirely.
- The parallel group uses Go's standard `t.Run` + `t.Parallel()` pattern (same as `chaos_test.go`). The wrapping `t.Run("Parallel provisioning tests", ...)` ensures all parallel children complete before the sequential tests after them execute.
- Cursor vibecoded
## Checklist:

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end testing for control plane upgrades with Karpenter node management on AWS, covering node readiness validation, OS image verification, and pod rescheduling scenarios.

* **Chores**
  * Removed outdated test fixture assets.


maxcao13:
We are getting rate limited?

/test e2e-aws-autonode

maxcao13:
Alright, that's pretty good that we cut it down to 51m9s.

The control plane upgrade drift test does still take ~30 minutes. According to the epic we should separate it out.

I think the plan should be:

- Move control plane upgrade drift out to a separate hypershift e2e file (called `karpenter_control_plane_upgrade_test.go` or something.
- `e2e-aws-autonode` should include that also in the `CI_TEST_RUN` regex in the release repo

This will match how it works for the regular TestControlPlaneUpgrade test -- it's a separate test, jobs can target it separately, and `e2e-aws` still runs it.

I've moved the karpenter drift test out in latest commit. So as of right now in the state of this PR,`e2e-aws-autonode` should not include the separate test, but `e2e-aws-techpreview` should. I'll follow this up on release repo so that the autonode test runs both (which will be in parallel).

test/e2e/karpenter_test.go
```diff
@@ -202,1093 +177,918 @@ func TestKarpenter(t *testing.T) {
...
-			g.Expect(err).NotTo(HaveOccurred())
-
-			latestReleaseImage, err := releaseProvider.Lookup(ctx, globalOpts.LatestReleaseImage, pullSecret)
-			g.Expect(err).NotTo(HaveOccurred())
+		t.Run("Parallel provisioning tests", func(t *testing.T) {
+			t.Run("Instance profile annotation propagation", func(t *testing.T) {
+				t.Parallel()
+				// Get the current HostedCluster
+				err := mgtClient.Get(ctx, crclient.ObjectKeyFromObject(hostedCluster), hostedCluster)
+				g.Expect(err).NotTo(HaveOccurred())
```
enxebre: This subtest uses the parent's g (bound to the parent *testing.T). If an assertion fails, it calls Fatal on the parent's t from a goroutine, which panics.

test/e2e/karpenter_test.go
```diff
@@ -202,1093 +177,918 @@ func TestKarpenter(t *testing.T) {
...
+				t.Logf("Node provisioned with correct kubelet version (v%s) for NodeClass version %s", expectedKubeletVersion, nodeClassVersion)
+
+				// Verify MetadataOptions propagated to the actual EC2 instance
+				t.Log("Verifying MetadataOptions on EC2 instance via DescribeInstances")
+				ec2client := ec2Client(clusterOpts.AWSPlatform.Credentials.AWSCredentialsFile, clusterOpts.AWSPlatform.Region)
+				for _, node := range nodes {
+					providerID := node.Spec.ProviderID
+					g.Expect(providerID).NotTo(BeEmpty(), "node should have a providerID")
+
+					parts := strings.Split(providerID, "/")
```
enxebre: nit: we might want to dry this logic which is used several times

test/e2e/karpenter_test.go
```diff
@@ -57,49 +58,23 @@ func TestKarpenter(t *testing.T) {
...
-			"node.kubernetes.io/instance-type": "t3.large",
-			"karpenter.sh/nodepool":            karpenterNodePool.GetName(),
-		}
-
 		t.Run("Test ARM64 instance provisioning", func(t *testing.T) {
 			if !globalOpts.ConfigurableClusterOptions.AWSMultiArch && !globalOpts.ConfigurableClusterOptions.AzureMultiArch {
 				t.Skip("test only supported on multi-arch clusters")
 			}
+
+			armNodePool := baseNodePool("arm-nodepool")
```
enxebre: why is this not parallel as well?

test/e2e/karpenter_test.go
```diff
@@ -202,1093 +177,918 @@ func TestKarpenter(t *testing.T) {
...
+				workerInstanceProfile := hostedCluster.Spec.InfraID + "-worker"
 
-			replicas := 1
-			workLoads.Object["spec"].(map[string]interface{})["replicas"] = replicas
+				// Apply the annotation to the HostedCluster
+				err = e2eutil.UpdateObject(t, ctx, mgtClient, hostedCluster, func(obj *hyperv1.HostedCluster) {
+					if obj.Annotations == nil {
+						obj.Annotations = make(map[string]string)
+					}
+					obj.Annotations[hyperv1.AWSKarpenterDefaultInstanceProfile] = workerInstanceProfile
```
enxebre: why is this annotation not an API input?

test/e2e/karpenter_test.go
```diff
@@ -202,1093 +177,918 @@ func TestKarpenter(t *testing.T) {
...
+				g.Eventually(func(g Gomega) {
+					ec2NodeClassList := &awskarpenterv1.EC2NodeClassList{}
+					err := guestClient.List(ctx, ec2NodeClassList)
+					g.Expect(err).NotTo(HaveOccurred())
+					g.Expect(ec2NodeClassList.Items).NotTo(BeEmpty())
+
+					// Find the default EC2NodeClass
+					var defaultEC2NodeClass *awskarpenterv1.EC2NodeClass
+					for i := range ec2NodeClassList.Items {
+						if ec2NodeClassList.Items[i].Name == "default" {
```
enxebre: why are we listing and check, instead of just getting?

test/e2e/karpenter_test.go
```diff
@@ -202,1093 +177,918 @@ func TestKarpenter(t *testing.T) {
...
+							defaultEC2NodeClass = &ec2NodeClassList.Items[i]
+							break
+						}
+					}
+					g.Expect(defaultEC2NodeClass).NotTo(BeNil(), "default EC2NodeClass should exist")
+					g.Expect(defaultEC2NodeClass.Spec.InstanceProfile).NotTo(BeNil(), "InstanceProfile should be set")
+					g.Expect(*defaultEC2NodeClass.Spec.InstanceProfile).To(Equal(workerInstanceProfile))
```
enxebre: won't this change provoke drift for other tests using the class?

test/e2e/karpenter_test.go
```diff
@@ -1371,23 +1171,36 @@ func TestKarpenter(t *testing.T) {
...
 			g.Expect(guestClient.Create(ctx, karpenterNodePool)).To(Succeed())
 			t.Logf("Created Karpenter NodePool")
 			g.Expect(guestClient.Create(ctx, workLoads)).To(Succeed())
-			t.Logf("Created workloads")
+			t.Logf("Created workloads with 2 replicas")
+
+			_ = e2eutil.WaitForReadyNodesByLabels(t, ctx, guestClient, hostedCluster.Spec.Platform.Type, 2, nodeLabels)
+			t.Logf("Both nodes ready, scaling workload to 1 replica to verify deprovisioning and consolidation")
+
+			err := e2eutil.UpdateObject(t, ctx, guestClient, workLoads, func(obj *appsv1.Deployment) {
```
enxebre: is there any other parallel update call that could make the latest one fail because of stale resourceVersion?

enxebre:
>Alright, that's pretty good that we cut it down to 51m9s.
The control plane upgrade drift test does still take ~30 minutes. According to the epic we should separate it out.
I think the plan should be:
Move control plane upgrade drift out to a separate hypershift e2e file (called karpenter_control_plane_upgrade_test.go or something.
e2e-aws-autonode should include that also in the CI_TEST_RUN regex in the release repo
This will match how it works for the regular TestControlPlaneUpgrade test -- it's a separate test, jobs can target it separately, and e2e-aws still runs it.
I've moved the karpenter drift test out in latest commit. So as of right now in the state of this PR,e2e-aws-autonode should not include the separate test, but e2e-aws-techpreview should. I'll follow this up on release repo so that the autonode test runs both (which will be in parallel).

the downside is we increase the number of HCs/infra per run. I think that's better than the the timing impact of running the drift test sequentially though. Additional HC infra would be mitigated if we move to shared infra. Also we could follow up to move the no cp drift karpenter validations into the existing TestCreateClusterCustomConfig.
cc @sjenning

enxebre:
https://github.com/openshift/hypershift/pull/8198#issuecomment-4222036006

Might this be becasuse NillableDuration zero-value serializes as "Never", disabling consolidation? can we have this unit tested?

test/e2e/karpenter_test.go
```diff
@@ -202,1093 +177,918 @@ func TestKarpenter(t *testing.T) {
...
-			t.Logf("Starting Karpenter control plane upgrade. FromImage: %s, toImage: %s", globalOpts.PreviousReleaseImage, globalOpts.LatestReleaseImage)
-
-			releaseProvider := &releaseinfo.RegistryClientProvider{}
-			pullSecret, err := os.ReadFile(clusterOpts.PullSecretFile)
-			g.Expect(err).NotTo(HaveOccurred())
-
-			latestReleaseImage, err := releaseProvider.Lookup(ctx, globalOpts.LatestReleaseImage, pullSecret)
-			g.Expect(err).NotTo(HaveOccurred())
+		t.Run("Parallel provisioning tests", func(t *testing.T) {
+			t.Run("Instance profile annotation propagation", func(t *testing.T) {
```
muraee: can we please split each sub-test into it's own function which is called here. This is very hard to follow and review

test/e2e/karpenter_test.go
maxcao13: This was back in december when we were looking to unblock the customer as quickly as possible, and we decided to introduce this annotation for SREs to apply on the default nodeclass, rather than allow customer to set it themselves which was also a point of contention back then. https://redhat.atlassian.net/browse/AUTOSCALE-473

I believe we did this because user's can't actually change the default nodeclass on any other fields, so we thought it wouldn't make sense for users to only be able to change this one, especially since default should work out of the box.

test/e2e/karpenter_test.go
maxcao13: Yeah you're right, I think it's not failing or anything right now because the only other parallel test `that will/is` using the default nodeclass `will be/is` the ARM one. I will just make all the parallel tests create their own nodeclass, thanks!

EDIT: Actually, I think that test is deliberately is testing upon the default. I'll make a comment saying that you should create your own nodeclass in the parallel tests, and change the arm one.

maxcao13:
> Might this be becasuse NillableDuration zero-value serializes as "Never", disabling consolidation? can we have this unit tested?

Yeah I think you are right good catch. This is in the hostedcluster in the latest autonode e2e dump:
```yaml
spec:
  disruption:
    budgets:
    - nodes: 10%
    consolidateAfter: Never
    consolidationPolicy: WhenEmptyOrUnderutilized
```

IMO, this seems to be a regression since the upstream behaviour before should be that `consolidateAfter` is defaulted to `"0s"` if you don't specify it.

Could we open an ocpbug/autoscale card and fix it there instead along with the unit? For now, I can hardcode `consolidateAfter` in the test to make it pass.

EDIT: Never mind, not a bug. It was due to switching over from unmarshaling json from disk, to using the go types directly, which bypasses the zero-value issue.

test/e2e/karpenter_test.go
maxcao13: done, although i'm not sure if that made it easier to read. I tried to keep changes self-contained in commits to make it easier to follow.

maxcao13:
/test e2e-aws-autonode
/test e2e-aws-techpreview

Can't really iterate on my own cluster because of the network blocks, so let's see what happens :-)

test/e2e/karpenter_control_plane_upgrade_test.go
```diff
@@ -0,0 +1,170 @@
...
+						*hc.Status.AutoNode.NodeCount, *hc.Status.AutoNode.NodeClaimCount), nil
+				},
+			},
+			e2eutil.WithTimeout(5*time.Minute),
+		)
+
+		g.Expect(guestClient.Delete(ctx, karpenterNodePool)).To(Succeed())
+		t.Logf("Deleted Karpenter NodePool")
+		g.Expect(guestClient.Delete(ctx, workLoads)).To(Succeed())
+		t.Logf("Delete workloads")
```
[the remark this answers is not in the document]
maxcao13: I'll resolve this in https://github.com/openshift/hypershift/pull/7979, i don't want to have to ping for another review on this and go through retests which are not free.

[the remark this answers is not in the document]
maxcao13: sorry i meant 7979

PR #8199, opened 2026-04-10
CNTRLPLANE-3214: add envtest suites for HCPEtcdBackup CRD validation

jparrill:
## Summary

- Add comprehensive CRD validation test suites for the HCPEtcdBackup API under `cmd/install/assets/crds/hypershift-operator/tests/hcpetcdbackups.hypershift.openshift.io/`
- Covers all CEL validation rules: storage union discrimination, field format regex, enum validation, spec immutability, status field validation, and encryption metadata constraints
- Tests run against both TechPreviewNoUpgrade and CustomNoUpgrade CRD variants via the envtest harness, filtered by `featureGates: [HCPEtcdBackup]`
- Fix flaky `WaitForCRDs` timeout in the envtest framework (10s → 30s) to prevent `context deadline exceeded` failures in CI

## Details

Three test suite files:

| File | Tests | Coverage |
|------|-------|----------|
| `validation.testsuite.yaml` | 27 onCreate | Storage union, field regex (bucket, region, keyPrefix, kmsKeyARN, container, storageAccount, encryptionKeyURL), enum, MaxLength boundaries (bucket/63, container/63, storageAccount/24), SecretReference |
| `immutability.testsuite.yaml` | 11 onUpdate | Spec `self == oldSelf`: bucket/container/keyPrefix/credentials mutations, storageType change, encryption key additions, unchanged-spec pass |
| `status.testsuite.yaml` | 9 onUpdate | snapshotURL scheme validation, encryptionMetadata MinProperties/MaxProperties, encryption metadata field format |

Closes: [CNTRLPLANE-3214](https://issues.redhat.com/browse/CNTRLPLANE-3214)
Follow-up from: https://github.com/openshift/hypershift/pull/8179#discussion_r3060277697

## Test plan

- [x] `make test-envtest-kube` — all specs pass across K8s 1.31–1.35
- [x] `make test-envtest-kube ENVTEST_KUBE_VERSIONS="1.35.0"` — 3 consecutive runs without flakiness
- [ ] CI envtest jobs pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #8200, opened 2026-04-10
CNTRLPLANE-3215: fix envtest GHA workflow path filters for CRD test suites

bryan-cox:

Fixes the path filters in the envtest GitHub Actions workflows (`envtest-kube.yaml` and `envtest-ocp.yaml`).

The workflows were configured to trigger on changes to `cmd/install/assets/hypershift-operator/tests/**`, but the actual CRD test suite files live under `cmd/install/assets/crds/hypershift-operator/tests/**` (missing the `crds/` path segment). This caused the envtest workflows to never trigger when CRD test suites were added or modified — as seen in PR #8199.

Fixes [CNTRLPLANE-3215](https://redhat.atlassian.net/browse/CNTRLPLANE-3215)

The fix is a simple path correction in both workflow files — adding the missing `crds/` directory segment to the path filter.

## Checklist:

## Summary by CodeRabbit

* **Chores**
  * Updated continuous integration workflow triggers to monitor test asset changes from a reorganized location, ensuring automated testing runs appropriately.

### DIFF
--- a/.github/workflows/envtest-kube.yaml
+++ b/.github/workflows/envtest-kube.yaml
@@ -6,13 +6,13 @@ on:
     paths:
       - 'api/**'
       - 'test/envtest/**'
-      - 'cmd/install/assets/hypershift-operator/tests/**'
+      - 'cmd/install/assets/crds/hypershift-operator/tests/**'
   pull_request:
     branches: [main]
     paths:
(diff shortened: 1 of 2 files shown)


jparrill (commented):
/lgtm

PR #8204, opened 2026-04-10
CNTRLPLANE-3222: Migrate self-managed Azure e2e tests to v2 Ginkgo framework

bryan-cox:
## Summary

- Migrate self-managed Azure e2e tests from the v1 framework (`*testing.T`, `t.Run()`) to v2 (Ginkgo v2 `It()` blocks) so each test case produces a separate JUnit XML entry for individual Sippy reporting
- Extract `Validate*` helpers from `Ensure*` wrappers accepting `testing.TB` for cross-framework compatibility
- Refactor `EventuallyObject`/`EventuallyObjects` to accept `testing.TB` (backward-compatible since `*testing.T` implements `testing.TB`)
- Add `GetGuestClient()`/`SetupGuestClient()` to `TestContext` for guest cluster validation

### New test specs (9 total across 3 label groups)

**`self-managed-azure-public`** (3 specs):
- Workload identity webhook mutation
- KAS allowed CIDRs
- Ingress Operator configuration

**`self-managed-azure-private`** (4 specs, `Ordered`):
- Internal LB annotation on private-router Service
- AzurePrivateLinkService CR with PLS alias
- Private Endpoint IP populated
- Private DNS Zone ID populated

**`self-managed-azure-oauth-lb`** (2 specs):
- OAuth Service as LoadBalancer with external IP
- OAuth token flow through LoadBalancer endpoint

### CI workflow (openshift/release PR to follow)

The test binary is invoked 3 times with `--ginkgo.label-filter` against different guest clusters (public, private, OAuth LB), each producing its own JUnit XML.

## Test plan

- [x] `make e2e` — all test binaries compile including `bin/test-e2e-self-managed-azure`
- [x] `bin/test-e2e-self-managed-azure --ginkgo.dry-run --ginkgo.v` — lists 9 specs
- [x] Label filtering works: `--ginkgo.label-filter="self-managed-azure-public"` runs 3 specs, skips 6
- [x] `E2E_SHOW_ENV_HELP=1` prints Azure-specific env vars
- [x] `go build -tags e2e ./test/e2e/...` — v1 tests still compile (backward compatible)
- [ ] Full CI run against real Azure clusters (pending openshift/release PR)

Ref: [CNTRLPLANE-3222](https://redhat.atlassian.net/browse/CNTRLPLANE-3222)

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive Azure test suite: workload identity webhook, private topology checks, and OAuth via load balancer.
  * Refactored tests to centralize validation logic and improve polling/timeout robustness.
* **Chores**
  * Broadened test helper interfaces for wider reuse across test contexts.
  * Registered two optional Azure environment variables for private network/subscription scenarios.

### DIFF
--- a/test/e2e/util/eventually.go
+++ b/test/e2e/util/eventually.go
@@ -82,7 +82,7 @@ func WithFilteredConditionDump(matchers ...Condition) EventuallyOption {
 }
 
 // EventuallyObject polls until the predicate is fulfilled on the object.
-func EventuallyObject[T client.Object](t *testing.T, ctx context.Context, objective string, getter func(context.Context) (T, error), predicates []Predicate[T], options ...EventuallyOption) {
+func EventuallyObject[T client.Object](t testing.TB, ctx context.Context, objective string, getter func(context.Context) (T, error), predicates []Predicate[T], options ...EventuallyOption) {
 	t.Helper()
 	opts := defaultOptions()
 	for _, option := range options {
@@ -209,7 +209,7 @@ func summarizePredicteResults(results []predicateResult) bool {
 	return done
 }
 
-func printStatus[T client.Object](t *testing.T, lastTimestamp time.Time, object T, done bool, reasons []string) {
+func printStatus[T client.Object](t testing.TB, lastTimestamp time.Time, object T, done bool, reasons []string) {
 	if len(reasons) == 0 {
 		return
 	}
@@ -236,7 +236,7 @@ type predicateResult struct {
 }
 
 // EventuallyObjects polls until the predicate is fulfilled on each of a set of objects.
-func EventuallyObjects[T client.Object](t *testing.T, ctx context.Context, objective string, getter func(context.Context) ([]T, error), groupPredicates []Predicate[[]T], predicates []Predicate[T], options ...EventuallyOption) {
+func EventuallyObjects[T client.Object](t testing.TB, ctx context.Context, objective string, getter func(context.Context) ([]T, error), groupPredicates []Predicate[[]T], predicates []Predicate[T], options ...EventuallyOption) {
 	t.Helper()
 	opts := defaultOptions()
 	for _, option := range options {
@@ -376,7 +376,7 @@ type predicateReasons struct {
 	reasons []string
 }
 
-func printCollectionStatus[T client.Object](t *testing.T, lastTimestamp time.Time, done bool, reasons map[types.NamespacedName]predicateReasons) {
+func printCollectionStatus[T client.Object](t testing.TB, lastTimestamp time.Time, done bool, reasons map[types.NamespacedName]predicateReasons) {
 	prefix := ""
 	if !done {
 		prefix = "in"
@@ -522,7 +522,7 @@ func adaptConditions(in []metav1.Condition) []Condition {
 }
 
 // EventuallyNotFound polls until the object is not found (deleted).
-func EventuallyNotFound[T client.Object](t *testing.T, ctx context.Context, c client.Client, obj T, options ...EventuallyOption) {
+func EventuallyNotFound[T client.Object](t testing.TB, ctx context.Context, c client.Client, obj T, options ...EventuallyOption) {
 	t.Helper()
 	opts := defaultOptions()
 	for _, option := range options {
--- a/test/e2e/util/eventually.go
+++ b/test/e2e/util/eventually.go
@@ -82,7 +82,7 @@ func WithFilteredConditionDump(matchers ...Condition) EventuallyOption {
 }
 
 // EventuallyObject polls until the predicate is fulfilled on the object.
-func EventuallyObject[T client.Object](t *testing.T, ctx context.Context, objective string, getter func(context.Context) (T, error), predicates []Predicate[T], options ...EventuallyOption) {
+func EventuallyObject[T client.Object](t testing.TB, ctx context.Context, objective string, getter func(context.Context) (T, error), predicates []Predicate[T], options ...EventuallyOption) {
 	t.Helper()
 	opts := defaultOptions()
 	for _, option := range options {
@@ -209,7 +209,7 @@ func summarizePredicteResults(results []predicateResult) bool {
 	return done
 }
 
-func printStatus[T client.Object](t *testing.T, lastTimestamp time.Time, object T, done bool, reasons []string) {
+func printStatus[T client.Object](t testing.TB, lastTimestamp time.Time, object T, done bool, reasons []string) {
 	if len(reasons) == 0 {
 		return
 	}
@@ -236,7 +236,7 @@ type predicateResult struct {
 }
 
 // EventuallyObjects polls until the predicate is fulfilled on each of a set of objects.
-func EventuallyObjects[T client.Object](t *testing.T, ctx context.Context, objective string, getter func(context.Context) ([]T, error), groupPredicates []Predicate[[]T], predicates []Predicate[T], options ...EventuallyOption) {
+func EventuallyObjects[T client.Object](t testing.TB, ctx context.Context, objective string, getter func(context.Context) ([]T, error), groupPredicates []Predicate[[]T], predicates []Predicate[T], options ...EventuallyOption) {
 	t.Helper()
 	opts := defaultOptions()
 	for _, option := range options {
@@ -376,7 +376,7 @@ type predicateReasons struct {
 	reasons []string
 }
 
-func printCollectionStatus[T client.Object](t *testing.T, lastTimestamp time.Time, done bool, reasons map[types.NamespacedName]predicateReasons) {
+func printCollectionStatus[T client.Object](t testing.TB, lastTimestamp time.Time, done bool, reasons map[types.NamespacedName]predicateReasons) {
 	prefix := ""
 	if !done {
 		prefix = "in"
--- a/test/e2e/util/util.go
+++ b/test/e2e/util/util.go
@@ -168,7 +168,7 @@ func InitGuestClients(ctx context.Context, t *testing.T, g Gomega, mgtClient crc
 	}
 }
 
-func UpdateObject[T crclient.Object](t *testing.T, ctx context.Context, client crclient.Client, original T, mutate func(obj T)) error {
+func UpdateObject[T crclient.Object](t testing.TB, ctx context.Context, client crclient.Client, original T, mutate func(obj T)) error {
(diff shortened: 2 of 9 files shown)


test/e2e/util/util.go
```diff
@@ -3475,23 +3475,27 @@ func EnsureDefaultSecurityGroupTags(t *testing.T, ctx context.Context, client cr
...
+	kubeClient, err := kubeclient.NewForConfig(guestConfig)
+	g.Expect(err).NotTo(HaveOccurred())
+
+	// ensure that kube-apiserver is not reachable from anywhere
+	ensureAPIServerAllowedCIDRs(ctx, t, g, mgmtClient, kubeClient, hc, []string{"0.0.0.0/32"}, false)
+	// ensure kube-apiserver is reachable when allowed CIDRs allow access from everywhere
+	// This is useful for testing purposes, as it allows us to access the kube-apiserver from any IP
+	// In a production environment, this should be restricted to specific CIDRs
+	ensureAPIServerAllowedCIDRs(ctx, t, g, mgmtClient, kubeClient, hc, append([]string{"0.0.0.0/0"}, generateTestCIDRs250()...), true)
+}
```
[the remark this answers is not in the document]
bryan-cox: Fixed. Added save/defer/restore pattern to capture the original `APIServer` networking state and restore it after the test completes.

test/e2e/v2/selfmanagedazure/private_topology_test.go
```diff
@@ -0,0 +1,177 @@
...
+					alias := pls.Status.PrivateLinkServiceAlias
+					if alias == "" {
+						return false, "PLS alias is empty", nil
+					}
+					return true, fmt.Sprintf("PLS alias is %q", alias), nil
+				},
+			},
+			e2eutil.WithTimeout(15*time.Minute),
+			e2eutil.WithInterval(15*time.Second),
+		)
```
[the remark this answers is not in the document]
bryan-cox: Fixed. Moved readiness checks into the collection predicate using an any-match pattern — the test now succeeds as soon as at least one `AzurePrivateLinkService` has the expected field populated, rather than requiring all items to match. Applied to all three PLS checks (alias, endpoint IP, DNS zone).

bryan-cox:
Addressed all CodeRabbit review feedback in 161bb4a4e0:

**Major fixes:**
- **CIDR restore** (`util.go`): `ValidateKubeAPIServerAllowedCIDRs` now saves original `APIServer` networking state and defers restoration after test
- **PLS any-match** (`hosted_cluster_azure_test.go`): All three `EventuallyObjects` PLS checks now use collection-level any-match predicates instead of per-item predicates that require all items to pass
- **OAuth polling** (`oauth.go`): Added `request.Clone(ctx)` for per-attempt context binding and `httpClient.Timeout = 15s` to prevent hung requests outliving the poll

**Nitpick fixes:**
- **Unused `mgmtClient`** (`util_ingress_operator_configuration.go`): Removed from `ValidateIngressOperatorConfiguration` and `EnsureIngressOperatorConfiguration` signatures + updated all callers
- **Namespace cleanup** (`azure.go`): Added `defer guestClient.Delete()` for the test namespace in `ValidateAzureWorkloadIdentityWebhookMutation`
- **Empty kubeconfig** (`test_context.go`): `GetHostedClusterClient` now checks `len(kubeconfigData) == 0` in addition to key presence

**Not applicable (files deleted):**
- `suite_test.go` comments (Ginkgo-managed context, env var pair-validation): The `selfmanagedazure/` package was deleted — tests now live in the shared v2 binary at `test/e2e/v2/tests/hosted_cluster_azure_test.go`

test/e2e/v2/tests/hosted_cluster_azure_test.go
```diff
@@ -0,0 +1,328 @@
...
+			if hc == nil || hc.Spec.Platform.Type != hyperv1.AzurePlatform {
+				Skip("Azure public cluster tests are only for Azure platform")
+			}
+		})
+
+		It("should mutate pods with workload identity federated credentials", func() {
+			e2eutil.GinkgoAtLeast(e2eutil.Version422)
+			testCtx := getTestCtx()
+			hostedClusterClient := testCtx.GetHostedClusterClient()
+			Expect(hostedClusterClient).NotTo(BeNil(), "hosted cluster client is nil; HostedCluster may not have KubeConfig status set")
```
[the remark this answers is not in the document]
bryan-cox: Fixed in fd50296d64. Added `WaitForGuestKubeConfig()` before `GetHostedClusterClient()` in the workload identity and ingress specs to ensure `Status.KubeConfig` is populated before the `sync.Once` fires. Removed the unused `GetHostedClusterClient()` call from the CIDR spec entirely since it builds its own client from the kubeconfig data returned by `WaitForGuestKubeConfig()`.

jparrill (commented):

test/e2e/v2/tests/hosted_cluster_azure_test.go
```diff
@@ -0,0 +1,328 @@
...
+		It("should have Azure internal LB annotation on private-router Service", func() {
+			ctx := testCtx.Context
+			e2eutil.EventuallyObject(GinkgoTB(), ctx, "private-router Service has Azure internal LB annotation",
+				func(ctx context.Context) (*corev1.Service, error) {
+					svc := &corev1.Service{
+						ObjectMeta: metav1.ObjectMeta{
+							Namespace: controlPlaneNamespace,
+							Name:      "private-router",
+						},
+					}
```
jparrill, line 118: Nit: the v1 tests use both `groupPredicates` (at least 1 CR exists) and per-item `predicates` (each PLS has the field). Here only `groupPredicates` are used with `nil` per-item predicates, which changes the semantics: v2 passes if *any* PLS has the field, v1 required *all*. Also loses per-CR diagnostic logging on timeout.

Consider keeping the two-predicate-slice pattern from v1 for stricter validation and better CI failure triage.

test/e2e/v2/tests/hosted_cluster_azure_test.go
```diff
@@ -0,0 +1,328 @@
...
+					if err != nil {
+						return nil, err
+					}
+					items := make([]*hyperv1.AzurePrivateLinkService, len(plsList.Items))
+					for i := range plsList.Items {
+						items[i] = &plsList.Items[i]
+					}
+					return items, nil
+				},
+				[]e2eutil.Predicate[[]*hyperv1.AzurePrivateLinkService]{
```
jparrill, line 192: Nit: this duplicates `support/util.ServicePublishingStrategyByTypeByHC(hc, hyperv1.OAuthServer)` which already returns the strategy.

Also, the PLS list getter closure above is copy-pasted 3 times (lines 118-128, 152-162, 186-196) — consider extracting a local `listPLS` helper.

test/e2e/v2/tests/hosted_cluster_azure_test.go
```diff
@@ -0,0 +1,328 @@
...
+			hc := testCtx.GetHostedCluster()
+			kubeconfigData := e2eutil.WaitForGuestKubeConfig(GinkgoTB(), testCtx.Context, testCtx.MgmtClient, hc)
+			restConfig, err := clientcmd.RESTConfigFromKubeConfig(kubeconfigData)
+			Expect(err).NotTo(HaveOccurred(), "failed to create hosted cluster REST config")
+
+			e2eutil.ValidateKubeAPIServerAllowedCIDRs(GinkgoTB(), testCtx.Context, testCtx.MgmtClient, restConfig, hc)
+		})
+
+		It("should have Ingress Operator configuration applied", func() {
+			e2eutil.GinkgoAtLeast(e2eutil.Version421)
```
jparrill, line 75: Nit: the four private topology specs are independent assertions — none depends on prior spec state. `Ordered` prevents parallel execution and causes cascade skips on failure. The other two Context groups in this file use `BeforeEach` without `Ordered`. Consider matching that pattern here too.

test/e2e/util/eventually.go
```diff
@@ -376,7 +376,7 @@ type predicateReasons struct {
 	reasons []string
 }
 
-func printCollectionStatus[T client.Object](t *testing.T, lastTimestamp time.Time, done bool, reasons map[types.NamespacedName]predicateReasons) {
+func printCollectionStatus[T client.Object](t testing.TB, lastTimestamp time.Time, done bool, reasons map[types.NamespacedName]predicateReasons) {
```
jparrill, line 379: Nit: `EventuallyObject`, `EventuallyObjects`, `printStatus`, and `printCollectionStatus` were all widened to `testing.TB`, but `EventuallyNotFound` (line 525) was missed. Not blocking (no v2 caller today), but worth updating for consistency.

test/e2e/v2/tests/hosted_cluster_azure_test.go
bryan-cox: Good catch. Extracted a `listPLS` helper to deduplicate the 3 closures. On the group vs per-item predicates: `EventuallyObjects` with group predicates is intentional here — there's only ever one PLS CR per private-router Service, so "any" and "all" are equivalent. Used group predicates because the assertion is about the collection state (at least one exists with the field set), not about validating every individual item.

---

test/e2e/v2/tests/hosted_cluster_azure_test.go
bryan-cox: Done. Replaced `hasOAuthLoadBalancerStrategy` with `supportutil.ServicePublishingStrategyByTypeByHC(hc, hyperv1.OAuthServer)`. Also extracted the PLS list closure into a `listPLS` helper as noted in the other comment.

---

test/e2e/v2/tests/hosted_cluster_azure_test.go
bryan-cox: The `Ordered` + `BeforeAll` pattern is intentional here. The PLS specs have an implicit causal dependency: PLS alias must exist before PE IP can be populated, and PE IP before DNS zone ID. `Ordered` ensures Ginkgo runs them in definition order and short-circuits on the first failure, which is the right behavior — if PLS alias never appears, waiting 15 minutes each for PE IP and DNS zone provides no signal. The shared `testCtx`/`controlPlaneNamespace` variables also rely on `BeforeAll` which requires `Ordered`.

---

test/e2e/util/eventually.go
bryan-cox: Done. Widened `EventuallyNotFound` to `testing.TB` for consistency with the other functions.

---

PR #7793, opened 2026-02-24
WIP: feat(azure): enable ACR image pulls via managed identity on worker nodes

bryan-cox:
## Summary

- Add `AzureImageRegistryCredentials` API type with `ManagedIdentity` (ARM resource ID) and `Registries` (ACR hostnames) fields on `AzureNodePoolPlatform`
- Assign user-assigned managed identity to worker VMSS via CAPZ `AzureMachineTemplate` so nodes can authenticate to ACR via IMDS
- Generate a MachineConfig that configures kubelet's `acr-credential-provider` plugin with credential provider config, azure.json, and a systemd drop-in for kubelet flags
- Integrate MachineConfig into the NodePool config pipeline so changes trigger hash-based rolling node updates

## Details

Enables ARO HCP worker nodes to pull container images from Azure Container Registry using a user-assigned managed identity and the kubelet credential provider plugin, eliminating the need for static pull secrets.

### API

New optional field on `NodePool.Spec.Platform.Azure`:

```yaml
imageRegistryCredentials:
  managedIdentity: /subscriptions/.../providers/Microsoft.ManagedIdentity/userAssignedIdentities/my-acr-identity
  registries:
  - myregistry.azurecr.io
```

CEL validation enforces ARM resource ID format, length constraints, and registry list bounds.

### Node Artifacts

The MachineConfig places three artifacts on worker nodes:
1. `/etc/kubernetes/acr-credential-provider-config.yaml` — `CredentialProviderConfig` with user registries + wildcard patterns
2. `/etc/kubernetes/acr-azure.json` — Azure auth config with tenant ID, subscription ID, and managed identity
3. `kubelet.service` drop-in — sets `--image-credential-provider-config` and `--image-credential-provider-bin-dir` flags

## Test plan

- [x] Unit tests for CAPZ identity assignment (2 tests)
- [x] Unit tests for MachineConfig generation (12 tests)
- [x] Unit tests for config pipeline integration and hash-based rollout (8 tests)
- [ ] E2E: Create NodePool with `imageRegistryCredentials`, verify VMSS has identity, verify nodes pull from ACR without pull secrets

## JIRA

[ARO-24037](https://issues.redhat.com/browse/ARO-24037)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
[...]

PR #8205, opened 2026-04-10
feat(azure): enable ACR image pulls via managed identity on worker nodes

twolff-gh:
Followup to https://github.com/openshift/hypershift/pull/7793

Add support for configuring kubelet's image credential provider to
authenticate to Azure Container Registry (ACR) using a user-assigned
managed identity. This eliminates the need for static pull secrets
by leveraging IMDS-based token acquisition on worker nodes.

## Summary by CodeRabbit

* **New Features**
  * Azure node pools can include image registry credentials (managed identity + registries) to enable ACR authentication on worker nodes.
  * When provided, worker VMs are assigned the configured identity and a MachineConfig is generated to install the ACR credential provider and supporting config files.

* **Bug Fixes**
  * Serialization compatibility between versions preserved for Azure node pool platform and registry credential data.

* **Tests**
  * Added comprehensive tests covering config generation, ignition payloads, validation, and JSON/YAML serialization.


api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/nodepools.hypershift.openshift.io/AAA_ungated.yaml
```diff
@@ -915,6 +915,45 @@ spec:
...
+
+                              Format: /subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.ManagedIdentity/userAssignedIdentities/{identityName}
+                            maxLength: 512
+                            minLength: 1
+                            pattern: ^/subscriptions/[^/]+/resourceGroups/[^/]+/providers/Microsoft\.ManagedIdentity/userAssignedIdentities/[^/]+$
+                            type: string
+                          registries:
+                            description: |-
+                              registries is the list of ACR registries that the credential provider should authenticate
+                              to. Each entry should be a fully qualified ACR hostname (e.g., "myregistry.azurecr.io").
```
bennerv: wildcards work here as well.  may want to call it out.

api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/nodepools.hypershift.openshift.io/AAA_ungated.yaml
```diff
@@ -915,6 +915,45 @@ spec:
...
+                              The credential provider will use the managed identity to obtain short-lived tokens for
```
bennerv: set a default?  There's a default already in https://github.com/openshift/machine-config-operator/blob/main/templates/common/azure/files/etc-kubernetes-credential-providers-acr-credential-provider.yaml

api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/nodepools.hypershift.openshift.io/AAA_ungated.yaml
```diff
@@ -915,6 +915,45 @@ spec:
...
+                              maxLength: 256
+                              minLength: 1
+                              type: string
+                            maxItems: 20
+                            minItems: 1
+                            type: array
+                            x-kubernetes-list-type: set
+                        required:
+                        - managedIdentity
+                        - registries
```
bennerv: make this optional?  use the default? does it have to be required (specifically `registries`)

api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/nodepools.hypershift.openshift.io/AAA_ungated.yaml
```diff
@@ -915,6 +915,45 @@ spec:
...
                             ?  true : !has(self.azureMarketplace)'
+                      imageRegistryCredentials:
+                        description: |-
+                          imageRegistryCredentials specifies configuration for enabling kubelet's image credential
+                          provider to authenticate to Azure Container Registry (ACR) using a managed identity.
+                          When configured, worker nodes will use the acr-credential-provider plugin to obtain
+                          short-lived tokens for ACR image pulls instead of relying on static pull secrets.
+                          Changing this field will trigger a node rollout.
+                        properties:
+                          managedIdentity:
```
bennerv, line 928: for standalone or other cases, do we want to support system-assigned identity?  If so, we wouldn't have a customer pass these in at creation time.

api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/nodepools.hypershift.openshift.io/AAA_ungated.yaml
twolff-gh: Sounds good. Docs update?

api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/nodepools.hypershift.openshift.io/AAA_ungated.yaml
twolff-gh: I think it makes sense to be optional. I'll confirm

api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/nodepools.hypershift.openshift.io/AAA_ungated.yaml
twolff-gh: For the scope of this feature, I believe it was only for user-assigned identities. I'll confirm.
If we needed to support system-assigned, I wonder if it would need to be populated differently / new field?

api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/nodepools.hypershift.openshift.io/AAA_ungated.yaml
twolff-gh: Removing the generateCredentialProviderConfig guard should cause the registries to be empty, and matchImages will only contain the acrDefaultMatchImages wildcards

hypershift-operator/controllers/nodepool/azure_acr.go

Do you think that works?

api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/nodepools.hypershift.openshift.io/AAA_ungated.yaml
twolff-gh: Description now calls out wildcard support with example (*.azurecr.io) and lists all four default patterns

api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/nodepools.hypershift.openshift.io/AAA_ungated.yaml
twolff-gh: Registries is now optional. no minItems. customer only needs to provide managedIdentity. Wildcards handle the rest

api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/nodepools.hypershift.openshift.io/AAA_ungated.yaml
twolff-gh: Updated. Now, when registries is omitted, the CredentialProviderConfig uses default wildcards

api/.golangci.yml
```diff
@@ -2059,6 +2059,17 @@ linters:
         path: hypershift/v1beta1/hosted_controlplane.go
         text: 'statusoptional: status field "Ready" must be marked as optional, not required'
 
+      # ImageRegistryCredentials uses pointer for nil-check convention consistency with controller code
```
bryan-cox: If kubeapilinter was catching this, I'd recommend changing it to what it wanted over overriding it here.

api/hypershift/v1beta1/azure.go
```diff
@@ -109,6 +109,48 @@ type AzureNodePoolPlatform struct {
...
+	//
+	// The identity must have the AcrPull role granted on the target ACR registry(ies).
+	//
+	// Format: /subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.ManagedIdentity/userAssignedIdentities/{identityName}
+	//
+	// +required
+	// +kubebuilder:validation:XValidation:rule="self.matches('^/subscriptions/[^/]+/resourceGroups/[^/]+/providers/Microsoft\\\\.ManagedIdentity/userAssignedIdentities/[^/]+$')",message="managedIdentity must be a valid ARM resource ID for a user-assigned managed identity"
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=512
+	ManagedIdentity string `json:"managedIdentity,omitempty"`
```
bryan-cox: We already have a ManagedIdentity type that we should use here instead of string. I believe many of the kubebuilder validations here are already applied there.

api/hypershift/v1beta1/azure_test.go
```diff
@@ -0,0 +1,162 @@
...
+	VMSize           string              `json:"vmSize"`
+	Image            AzureVMImage        `json:"image"`
+	OSDisk           AzureNodePoolOSDisk `json:"osDisk"`
+	AvailabilityZone string              `json:"availabilityZone,omitempty"`
+	EncryptionAtHost string              `json:"encryptionAtHost,omitempty"`
+	SubnetID         string              `json:"subnetID"`
+	Diagnostics      *Diagnostics        `json:"diagnostics,omitempty"`
+}
+
+func TestAzureNodePoolPlatformSerializationCompatibility(t *testing.T) {
```
bryan-cox: We can drop the unit tests here for envtests instead.

hypershift-operator/controllers/nodepool/azure_acr.go
```diff
@@ -0,0 +1,212 @@
...
+	acrCredentialProviderConfigPath = "/etc/kubernetes/credential-providers/acr-credential-provider.yaml"
+	acrAzureJSONPath                = "/etc/kubernetes/acr-azure.json"
+	acrCredentialProviderBinName    = "acr-credential-provider"
+	acrMachineConfigName            = "50-acr-credential-provider"
+)
+
+// acrDefaultMatchImages lists the default wildcard patterns for Azure Container Registry
+// endpoints across all Azure clouds. These patterns are always appended to the user-specified
+// registries to ensure the credential provider covers all standard ACR endpoints.
+var acrDefaultMatchImages = []string{
```
bryan-cox, line 34: Can these not be defaulted in the api for the registries?

api/hypershift/v1beta1/azure.go
```diff
@@ -109,6 +109,48 @@ type AzureNodePoolPlatform struct {
...
+	// *.azurecr.cn, *.azurecr.de, *.azurecr.us). When omitted, only the default wildcard
+	// patterns are used, which cover all standard Azure Container Registry endpoints.
+	//
+	// +optional
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=20
+	// +kubebuilder:validation:items:MinLength=1
+	// +kubebuilder:validation:items:MaxLength=256
+	// +listType=set
+	Registries []string `json:"registries,omitempty"`
```
bryan-cox: How do we know the max length is 256?

api/hypershift/v1beta1/azure.go
```diff
@@ -109,6 +109,48 @@ type AzureNodePoolPlatform struct {
...
+	// +kubebuilder:validation:MaxLength=512
+	ManagedIdentity string `json:"managedIdentity,omitempty"`
+
+	// registries is an optional list of ACR registries that the credential provider should
+	// authenticate to. Each entry can be a fully qualified ACR hostname (e.g.,
+	// "myregistry.azurecr.io") or a wildcard pattern (e.g., "*.azurecr.io") to match
+	// multiple registries. When specified, these registries are added to the credential
+	// provider's matchImages list alongside the default wildcard patterns (*.azurecr.io,
+	// *.azurecr.cn, *.azurecr.de, *.azurecr.us). When omitted, only the default wildcard
+	// patterns are used, which cover all standard Azure Container Registry endpoints.
```
bryan-cox: Can we call out a customer only gets 16 of the 20 spots since 4 are used for the wildcards?

api/hypershift/v1beta1/azure.go
```diff
@@ -109,6 +109,48 @@ type AzureNodePoolPlatform struct {
...
 	// +optional
 	Diagnostics *Diagnostics `json:"diagnostics,omitempty"`
+
+	// imageRegistryCredentials specifies configuration for enabling kubelet's image credential
+	// provider to authenticate to Azure Container Registry (ACR) using a managed identity.
+	// When configured, worker nodes will use the acr-credential-provider plugin to obtain
+	// short-lived tokens for ACR image pulls instead of relying on static pull secrets.
+	// Changing this field will trigger a node rollout.
+	// +optional
+	ImageRegistryCredentials *AzureImageRegistryCredentials `json:"imageRegistryCredentials,omitempty,omitzero"`
```
bryan-cox: I don't think this needs both omitempty, omitzero. What does the kubeapilinter say?

hypershift-operator/controllers/nodepool/azure.go
```diff
@@ -246,6 +246,13 @@ func azureMachineTemplateSpec(nodePool *hyperv1.NodePool) (*capiazure.AzureMachi
 
 	azureMachineTemplate.Template.Spec.SSHPublicKey = dummySSHKey
 
+	if nodePool.Spec.Platform.Azure.ImageRegistryCredentials != nil {
+		azureMachineTemplate.Template.Spec.Identity = capiazure.VMIdentityUserAssigned
+		azureMachineTemplate.Template.Spec.UserAssignedIdentities = []capiazure.UserAssignedIdentity{{
+			ProviderID: "azure://" + nodePool.Spec.Platform.Azure.ImageRegistryCredentials.ManagedIdentity,
```
bryan-cox: Use CAPZ's ProviderIDPrefix constant instead of hardcoded "azure://" string for clarity and to make the CAPZ convention dependency explicit.

hypershift-operator/controllers/nodepool/azure.go
```diff
@@ -246,6 +246,13 @@ func azureMachineTemplateSpec(nodePool *hyperv1.NodePool) (*capiazure.AzureMachi
 
 	azureMachineTemplate.Template.Spec.SSHPublicKey = dummySSHKey
 
+	if nodePool.Spec.Platform.Azure.ImageRegistryCredentials != nil {
```
bryan-cox: Add comment documenting that identity assignment currently replaces (not appends), and should be refactored to append if additional identities are needed in the future.

hypershift-operator/controllers/nodepool/azure_acr.go
```diff
@@ -0,0 +1,212 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal azure config: %w", err)
+	}
+	return data, nil
+}
+
+// buildACRIgnitionConfig assembles the ignition v3.2.0 Config containing the
+// credential provider config file and azure.json file.
+func buildACRIgnitionConfig(credentialProviderConfig, azureJSON []byte) ([]byte, error) {
+	fileMode := 0644
```
bryan-cox: Consider tightening acr-azure.json file permissions to 0600 (only kubelet needs to read it).

hypershift-operator/controllers/nodepool/azure_acr_test.go
```diff
@@ -0,0 +1,565 @@
...
+}
+
+func TestGetACRCredentialProviderConfig(t *testing.T) {
+	testCases := []struct {
+		name        string
+		nodePool    *hyperv1.NodePool
+		hc          *hyperv1.HostedCluster
+		expectNil   bool
+		expectErr   bool
+		expectErrMs string
```
bryan-cox:
```suggestion
		expectErrMsg string
```

api/hypershift/v1beta1/azure.go
everettraven: From what I can tell, the zero value is not accepted here because this has required children fields.

That means that this shouldn't need to be a pointer because there is no need to distinguish between unset and set to the zero value and that it can just use `omitzero`.

api/hypershift/v1beta1/azure.go
```diff
@@ -109,6 +109,48 @@ type AzureNodePoolPlatform struct {
 	// If not specified, then Boot diagnostics will be disabled.
 	// +optional
 	Diagnostics *Diagnostics `json:"diagnostics,omitempty"`
+
+	// imageRegistryCredentials specifies configuration for enabling kubelet's image credential
+	// provider to authenticate to Azure Container Registry (ACR) using a managed identity.
+	// When configured, worker nodes will use the acr-credential-provider plugin to obtain
+	// short-lived tokens for ACR image pulls instead of relying on static pull secrets.
+	// Changing this field will trigger a node rollout.
+	// +optional
```
everettraven, line 120: What happens if I don't configure this field?

api/hypershift/v1beta1/azure.go
everettraven: It sounds like this is additive. If the hard limit is 20 including the defaults, you should limit this to only what the user can actually provide.

api/hypershift/v1beta1/azure.go
```diff
@@ -109,6 +109,48 @@ type AzureNodePoolPlatform struct {
...
+	//
+	// +required
+	// +kubebuilder:validation:XValidation:rule="self.matches('^/subscriptions/[^/]+/resourceGroups/[^/]+/providers/Microsoft\\\\.ManagedIdentity/userAssignedIdentities/[^/]+$')",message="managedIdentity must be a valid ARM resource ID for a user-assigned managed identity"
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=512
+	ManagedIdentity string `json:"managedIdentity,omitempty"`
+
+	// registries is an optional list of ACR registries that the credential provider should
+	// authenticate to. Each entry can be a fully qualified ACR hostname (e.g.,
+	// "myregistry.azurecr.io") or a wildcard pattern (e.g., "*.azurecr.io") to match
```
everettraven: It sounds like there is a required format here for each entry. This should be enforced via validation.

api/hypershift/v1beta1/azure.go
```diff
@@ -109,6 +109,48 @@ type AzureNodePoolPlatform struct {
...
+	// provider's matchImages list alongside the default wildcard patterns (*.azurecr.io,
+	// *.azurecr.cn, *.azurecr.de, *.azurecr.us). When omitted, only the default wildcard
+	// patterns are used, which cover all standard Azure Container Registry endpoints.
+	//
+	// +optional
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=20
+	// +kubebuilder:validation:items:MinLength=1
+	// +kubebuilder:validation:items:MaxLength=256
+	// +listType=set
```
everettraven: The constraints enforced by these markers should be explicitly documented in the GoDoc for the field so that they are mentioned in the generated documentation that end users will see.

api/hypershift/v1beta1/azure.go
```diff
@@ -109,6 +109,48 @@ type AzureNodePoolPlatform struct {
...
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=512
+	ManagedIdentity string `json:"managedIdentity,omitempty"`
+
+	// registries is an optional list of ACR registries that the credential provider should
+	// authenticate to. Each entry can be a fully qualified ACR hostname (e.g.,
+	// "myregistry.azurecr.io") or a wildcard pattern (e.g., "*.azurecr.io") to match
+	// multiple registries. When specified, these registries are added to the credential
+	// provider's matchImages list alongside the default wildcard patterns (*.azurecr.io,
+	// *.azurecr.cn, *.azurecr.de, *.azurecr.us). When omitted, only the default wildcard
```
everettraven: Am I allowed to specify an entry that matches one of these default patterns?

api/hypershift/v1beta1/azure.go
everettraven: If this is supposed to be representative of a DNS subdomain, I believe the limit is 253 (https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names)

api/hypershift/v1beta1/azure.go
twolff-gh: Do you mean an item-level pattern / regex?

everettraven: Yes

api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/nodepools.hypershift.openshift.io/AAA_ungated.yaml
twolff-gh: Yes, user-assigned was the main ask, but system-assigned would likely follow in the future.
Do they make sense as separate work - user now and system as another PR after?

api/.golangci.yml
twolff-gh: Override removed. azure_test.go which needed it was deleted

api/hypershift/v1beta1/azure_test.go
twolff-gh: All / both of them good to drop? (and add env tests)

api/hypershift/v1beta1/azure.go
twolff-gh: I'm trying to track it down, but ManagedIdentity is think is different than what is needed here? Do you know or can you confirm?

hypershift-operator/controllers/nodepool/azure_acr.go
twolff-gh: From what I found, these come from MCO credential provider config, if changed upstream, API default could go stale. Does that sound accurate?

Can or should we change this at the MCO config?

api/hypershift/v1beta1/azure.go
twolff-gh: Changed to 253

api/hypershift/v1beta1/azure.go
twolff-gh: MaxItems=16 added. user gets 16 slots, 4 defaults added by controller.

api/hypershift/v1beta1/azure.go
twolff-gh: omitzero used here now

hypershift-operator/controllers/nodepool/azure.go
twolff-gh: Added

api/hypershift/v1beta1/azure.go
bryan-cox: It's in this file further down

api/hypershift/v1beta1/azure_test.go
bryan-cox: Good to drop the whole file and just have envtests 👍🏻

hypershift-operator/controllers/nodepool/azure.go
twolff-gh: Using ProviderIDPrefix

hypershift-operator/controllers/nodepool/azure_acr.go
twolff-gh: Done

api/hypershift/v1beta1/azure.go
twolff-gh: Added a pattern shat should fit

api/hypershift/v1beta1/azure.go
twolff-gh: Should be addressed now

api/hypershift/v1beta1/azure.go
twolff-gh: Added: When not configured, no additional image credential provider is set up and worker nodes use the default pull secret for image authentication.

api/hypershift/v1beta1/azure.go
twolff-gh: Controller duplicates silently. Duplicate is dropped.
Does this work?

everettraven: De-duplication is fine - should we prevent users from specifying the wildcard patterns that are applied by default though? It doesn't look like there would be any value to the users to be able to specify those aside from  unnecessarily consuming some of their 16 slots.

Do we anticipate these defaults potentially changing much in the future (i.e removing one of them)? If so, it could make sense to allow end-users to explicitly provide them to be more resilient to changes in the defaults.

api/hypershift/v1beta1/azure.go
```diff
@@ -109,6 +109,62 @@ type AzureNodePoolPlatform struct {
...
+	// the entries are added to the credential provider's matchImages list alongside the
+	// default wildcard patterns. When omitted, only the default wildcard patterns are used,
+	// which cover all standard Azure Container Registry endpoints.
+	//
+	// +optional
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=16
+	// +kubebuilder:validation:items:MinLength=1
+	// +kubebuilder:validation:items:MaxLength=253
+	// +kubebuilder:validation:items:Pattern=`^(\*\.)?[a-zA-Z0-9]([a-zA-Z0-9\-]*[a-zA-Z0-9])?(\.[a-zA-Z0-9]([a-zA-Z0-9\-]*[a-zA-Z0-9])?)+$`
```
everettraven: Instead of using the `Pattern` marker, we generally recommend using the `XValidation` marker to enforce the pattern via a CEL expression and supplying a human-readable error message on validation failure.

Using the pattern marker means that the end user is provided the literal regex as part of the validation failure instead of a user-friendly message that describes the constraints.

api/hypershift/v1beta1/azure.go
```diff
@@ -109,6 +109,62 @@ type AzureNodePoolPlatform struct {
...
+	// managedIdentity is the ARM resource ID of a user-assigned managed identity that will be
+	// assigned to worker node VMs/VMSS. The credential provider plugin running on each node
+	// will use this identity to authenticate to ACR via the Azure Instance Metadata Service (IMDS).
+	//
+	// The identity must have the AcrPull role granted on the target ACR registry(ies).
+	//
+	// Format: /subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.ManagedIdentity/userAssignedIdentities/{identityName}
+	//
+	// +required
+	ManagedIdentity AzureManagedIdentityResourceID `json:"managedIdentity,omitempty"`
```
everettraven: Include the length constraints for this field (enforced by the type) in the GoDoc here as well please

hypershift-operator/controllers/nodepool/azure_acr.go
```diff
@@ -0,0 +1,229 @@
...
+
+// generateACRAzureJSON produces the JSON content for the Azure authentication config
+// consumed by the acr-credential-provider binary.
+func generateACRAzureJSON(managedIdentity, cloud, tenantID, subscriptionID string) ([]byte, error) {
+	cfg := acrAzureConfig{
+		Cloud:                       cloud,
+		TenantID:                    tenantID,
+		SubscriptionID:              subscriptionID,
+		UseManagedIdentityExtension: true,
+		UserAssignedIdentityID:      managedIdentity,
```
bennerv, line 153: Note in [cloud provider config docs](https://cloud-provider-azure.sigs.k8s.io/install/configs/), it says this is the clientID of the managed identity, but in reality it also is supported as a resourceID.  

```
userAssignedIdentityID	The Client ID of the user assigned MSI which is assigned to the underlying VMs	Required for user-assigned managed identity.
```

https://github.com/kubernetes-sigs/cloud-provider-azure/blob/e188952e7463d5539ef6fca0e0098acbb497638d/pkg/azclient/auth_func.go#L80

api/hypershift/v1beta1/azure.go
bennerv: something like: 
```
type UserAssignedManagedIdentity struct {
	// clientID is the client ID of a managed identity associated with CredentialsSecretName. This field is optional and
	// mainly used for CI purposes.
	//
	// +optional
	ClientID AzureClientID `json:"clientID,omitempty"`

	ResourceID string 
```

api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/nodepools.hypershift.openshift.io/AAA_ungated.yaml
bennerv: if system-assigned identities are added per-VM, then this doesn't make sense from a feature perspective.  Maybe try and see what happens, then ignore it or implement it as you see fit.

api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/nodepools.hypershift.openshift.io/GCPPlatform.yaml
```diff
@@ -915,6 +915,57 @@ spec:
...
+                                managed identity (e.g., /subscriptions/{sub}/resourceGroups/{rg}/providers/Microsoft.ManagedIdentity/userAssignedIdentities/{name})
+                              rule: self.matches('^/subscriptions/[^/]+/resourceGroups/[^/]+/providers/Microsoft\\.ManagedIdentity/userAssignedIdentities/[^/]+$')
+                          registries:
+                            description: |-
+                              registries is an optional list of ACR registries that the credential provider should
+                              authenticate to. Each entry can be a fully qualified ACR hostname (e.g.,
+                              "myregistry.azurecr.io") or a wildcard pattern (e.g., "*.azurecr.io") to match
+                              multiple registries. A maximum of 16 registries may be specified; four default wildcard
+                              patterns (*.azurecr.io, *.azurecr.cn, *.azurecr.de, *.azurecr.us) are added
+                              automatically by the controller. Each entry must be a valid hostname or wildcard
```
bennerv: If four wildcard patterns are added by the controller, what's the point in the customer specifying any entries at all since these only work for `azurecr.XX`?  Any `*.azurecr.*` will already go through the acr-pull-credentials.

twolff-gh: It would make things a lot simpler. Removing that field.
Since the RBAC enables cross-tenant registry interaction, it would allow customers to use any ACR. So, giving them the option to specify one explicitly is kind of pointless? Unless I'm missing something?

Additionally, if it is an override option, it would just artificially be limiting it, when RBAC already secures/restricts another registry.

So, it make more sense to remove it and the unnecessary complexity it adds since the defaults cover all the ACR endpoints.

Does that look right? Customer wants to use the registries support, posts the user-assigned ID to the endpoint and that enables it?

twolff-gh: Also, I don't think ROSA has the registries field either.

bennerv: > Since the RBAC enables cross-tenant registry interaction

managed identities are single tenant, so if we needed to access an image registry cross-tenant it would require a pull secret

> So, it make more sense to remove it and the unnecessary complexity it adds since the defaults cover all the ACR endpoints.

I think so. The only reason why it wouldn't make sense is if the customer wanted to restrict the registry access to specific ones, but they can already do that with azure rbac as you mentioned.

api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/nodepools.hypershift.openshift.io/AAA_ungated.yaml
twolff-gh: From what I see, a cluster needs to be created before getting a principal id which some process would need permissions to access customer resources.

Does the customer trust the platform to write IAM policies on their container registry? (Don't know)

Seems like a followup item. I'll gladly look into

api/hypershift/v1beta1/azure.go
twolff-gh: Implemented.

api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/nodepools.hypershift.openshift.io/GCPPlatform.yaml
twolff-gh: Removing the registries field
Adds more complexity than its worth. And ROSA doesn't even do it, and it's not explicitly asked for
Needs are med without it

api/hypershift/v1beta1/azure.go
twolff-gh: Removed the registries field. It was in the original WIP, but doesnt make sense in this final version.
Adds more complexity than necessary
The default patterns copied in plus the RBAC for ids should be enough

api/hypershift/v1beta1/azure.go
twolff-gh: Removed the registries field. It was in the original WIP, but doesnt make sense in this final version.
Adds more complexity than necessary
The default patterns copied in plus the RBAC for ids should be enough

api/hypershift/v1beta1/azure.go
```diff
@@ -109,6 +109,59 @@ type AzureNodePoolPlatform struct {
...
+type AzureManagedIdentityResourceID string
+
+// UserAssignedManagedIdentity specifies a user-assigned managed identity for Azure resource
+// authentication. The resourceID is always required for VMSS identity attachment. The clientID
+// is optional and can be provided for the credential provider configuration; if omitted, the
+// resourceID is used instead (cloud-provider-azure accepts both formats).
+type UserAssignedManagedIdentity struct {
+	// resourceID is the ARM resource ID of the user-assigned managed identity.
+	//
+	// Format: /subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.ManagedIdentity/userAssignedIdentities/{identityName}
```
everettraven, line 141: Fully document the length constraints as well please

api/hypershift/v1beta1/azure.go
```diff
@@ -109,6 +109,59 @@ type AzureNodePoolPlatform struct {
...
+	//
+	// +required
+	ResourceID AzureManagedIdentityResourceID `json:"resourceID,omitempty"`
+
+	// clientID is the client ID of the user-assigned managed identity. This field is optional
+	// and is mainly used for CI purposes.
+	//
+	// +optional
+	ClientID AzureClientID `json:"clientID,omitempty"`
```
everettraven, line 150: Is there a use case where this field provides value to end users?

We shouldn't design the API surface for what we need for testing (unless I'm misunderstanding "CI purposes").

api/hypershift/v1beta1/azure.go
```diff
@@ -109,6 +109,52 @@ type AzureNodePoolPlatform struct {
...
+
+// AzureManagedIdentityResourceID is a full Azure resource ID for a user-assigned managed identity.
+// The expected format is:
+//
+//	/subscriptions/{subscriptionID}/resourceGroups/{resourceGroup}/providers/Microsoft.ManagedIdentity/userAssignedIdentities/{identityName}
+//
+// +kubebuilder:validation:MinLength=1
+// +kubebuilder:validation:MaxLength=512
+// +kubebuilder:validation:XValidation:rule="self.matches('^/subscriptions/[^/]+/resourceGroups/[^/]+/providers/Microsoft\\\\.ManagedIdentity/userAssignedIdentities/[^/]+$')",message="must be a valid ARM resource ID for a user-assigned managed identity (e.g., /subscriptions/{sub}/resourceGroups/{rg}/providers/Microsoft.ManagedIdentity/userAssignedIdentities/{name})"
+type AzureManagedIdentityResourceID string
```
everettraven, line 132: From another PR I reviewed recently that enforces this format, it looks like subscription ID and resource group can be further validated to be the correct format like:

```
// +kubebuilder:validation:XValidation:rule="oldSelf.hasValue() && oldSelf.value() == self || size(self.split('/')) == 9 && self.split('/')[2].matches('^[0-9a-fA-F]{8}-([0-9a-fA-F]{4}-){3}[0-9a-fA-F]{12}$')",message="the subscriptionId in the imageID must be a valid UUID. It should be 5 groups of hyphen separated hexadecimal characters in the form 8-4-4-4-12",optionalOldSelf=true
// +kubebuilder:validation:XValidation:rule=`oldSelf.hasValue() && oldSelf.value() == self || size(self.split('/')) == 9 && self.split('/')[4].matches('^[a-zA-Z0-9_\\(\\)\\.-]{1,90}$')`,message="the resourceGroupName should be between 1 and 90 characters, consisting only of alphanumeric characters, hyphens, underscores, periods and parenthesis",optionalOldSelf=true
```

Reference: https://github.com/openshift/hypershift/pull/8106/changes#diff-11b7c56a5faf0c5e2154fc35474456e2aac1b6ea8e22e5bbda6c3a3996d91afaR143-R144

PR #8193, opened 2026-04-09
CNTRLPLANE-3197: update latest supported version from 4.22 to 4.23

clebs:
## Summary
- Update LatestSupportedVersion from 4.22.0 to 4.23.0
- Add Kubernetes 1.35.0 mapping for OCP 4.22.0

## Changes
- **Version constants**: Updated supported version range to 4.14-4.23
- **Kubernetes mappings**: Added 4.22.0 → 1.35.0 mapping
- **Test updates**: Updated all tests to reflect new version constraints
- **NodePool tests**: Fixed version skew test with compatible versions

## Fixes
- [CNTRLPLANE-3197](https://issues.redhat.com/browse/CNTRLPLANE-3197)

## Test plan
- [x] Unit tests pass with updated version ranges
- [x] `hypershift version` command shows 4.23.0 as latest supported
- [ ] Version skew validation works correctly with new minimum versions
- [ ] All existing functionality preserved

## Summary by CodeRabbit

* **New Features**
  * Added official support for OpenShift 5.0 and a Kubernetes 1.35.0 compatibility mapping for newer OpenShift releases.

* **Tests**
  * Updated test fixtures and end-to-end version references to reflect the expanded supported-version window (now including 5.0 and 4.23 down through 4.14).

PR #8216, opened 2026-04-13
OCPBUGS-83285: revert 5.0 supported version from release-4.22

bryan-cox:
## Summary

- Reverts `LatestSupportedVersion` from `5.0.0` back to `4.22.0` on the `release-4.22` branch
- Removes `prevLatestSupportedMajor` variable and cross-major version logic in `Supported()` that were incorrectly cherry-picked from main via PR #8193
- Fixes unit tests and e2e version constants to match

## Context

PR https://github.com/openshift/hypershift/pull/8193 updated the latest supported version to 5.0 on main. When it was cherry-picked to `release-4.22`, it incorrectly brought along the 5.0 version bump and the cross-major version logic. On `release-4.22`, the latest supported version should remain `4.22.0`.

## Test plan

- [x] `go test ./support/supportedversion/...` passes
- [x] CodeRabbit review passes with no findings

Fixes: https://issues.redhat.com/browse/OCPBUGS-83285

### DIFF
--- a/support/supportedversion/version.go
+++ b/support/supportedversion/version.go
@@ -32,13 +32,9 @@ const (
 // NOTE: The .0 (z release) should be ignored. It's only here to support
 // semver parsing.
 var (
-	LatestSupportedVersion      = semver.MustParse("5.0.0")
+	LatestSupportedVersion      = semver.MustParse("4.22.0")
 	MinSupportedVersion         = semver.MustParse("4.14.0")
 	IBMCloudMinSupportedVersion = semver.MustParse("4.14.0")
-	// prevLatestSupportedMajor holds the value of the latest version before we updated to a new major.
-	// This value is only used internally to compute the list of supported versions when we have 2
-	// different major versions at the same time.
-	prevLatestSupportedMajor = semver.MustParse("4.23.0")
 )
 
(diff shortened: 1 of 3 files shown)


support/supportedversion/version.go
```diff
@@ -85,21 +81,8 @@ func GetMinSupportedVersion(hc *hyperv1.HostedCluster) semver.Version {
 
 func Supported() []string {
 	versions := []string{trimVersion(LatestSupportedVersion.String())}
-	if LatestSupportedVersion.Major > MinSupportedVersion.Major {
```
clebs, line 88: Should we not leave this logic in? when `latest supported` and `min supported` major versions are equal this will behave as it did before, so there will be no change in behaviour.

Otherwise we need to make sure to reintroduce this later.

bryan-cox: This reverted change only goes in `release-4.22`. The change is still present in `main`.

clebs: 🤦 of course...

bryan-cox:
/override "Red Hat Konflux / enterprise-contract-mce-217 / hypershift-release-mce-217"

PR #8219, opened 2026-04-13
CNTRLPLANE-3524: feat: add etcd data re-encryption after encryption key rotation

muraee:
## Summary

- Add a re-encryption controller in the HCCO that detects encryption key rotations and triggers `StorageVersionMigration` CRs to re-encrypt all existing etcd data with the new active key
- Track progress through a new `EtcdDataEncryptionUpToDate` condition on HCP/HostedCluster
- Vendor library-go's `KubeStorageVersionMigrator` and kube-storage-version-migrator informer/lister packages

### Problem
HyperShift supports encryption key rotation (setting a new active key + backup key), but has no mechanism to re-encrypt existing etcd data with the new key. Old data remains encrypted with the previous key indefinitely, violating compliance requirements (OCPSTRAT-2527, [OCPSTRAT-2540](https://redhat.atlassian.net/browse/OCPSTRAT-2540)).

### Solution

**Key change detection** is performed entirely within the HCCO re-encryption controller — no CPO changes needed. The controller computes a SHA-256 fingerprint of the active key identity from the HCP spec on each reconcile and compares it against a `hypershift.openshift.io/encryption-migrated-key-hash` annotation on the `kas-secret-encryption-config` secret. When the hashes differ, re-encryption is triggered.

**API** (`api/hypershift/v1beta1/hostedcluster_conditions.go`):
- `EtcdDataEncryptionUpToDate` condition type
- Reasons: `ReEncryptionInProgress`, `ReEncryptionCompleted`, `ReEncryptionFailed`, `WaitingForKASConvergence`

**HCCO** (`control-plane-operator/hostedclusterconfigoperator/controllers/reencryption/`):
- New `reencryption` controller using library-go's `KubeStorageVersionMigrator`
- `computeActiveKeyFingerprint()` — SHA-256 hash of active key identity (supports AWS KMS, Azure KMS, IBM Cloud KMS, AESCBC)
- On first deployment, initializes migrated hash to active hash (assumes data is already encrypted with current key)
- Watches HCP, encryption config secret, and KAS deployment on management cluster
- Waits for KAS deployment convergence before starting migrations
- Creates `StorageVersionMigration` CRs in the guest cluster for all encrypted resources (secrets, configmaps, routes, oauth tokens)
- Retries failed migrations after 5-minute interval via `PruneMigration`
- Sets `EtcdDataEncryptionUpToDate` condition on HCP with detailed progress messages

**HyperShift Operator** (`hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go`):
- Bubbles up `EtcdDataEncryptionUpToDate` condition from HCP to HostedCluster (only when encryption is configured and condition exists)

### Vendoring
- `library-go/pkg/operator/encryption/controllers/migrators/` — `KubeStorageVersionMigrator`, `Migrator` interface
- `kube-storage-version-migrator/pkg/clients/informer/` and `lister/` — SVM informer/lister

Ref: OCPSTRAT-2527, [OCPSTRAT-2540](https://redhat.atlassian.net/browse/OCPSTRAT-2540)
Enhancement: openshift/enhancements#1969

## Test plan

- [x] Unit tests for `computeActiveKeyFingerprint()` — all provider types, determinism, different keys produce different hashes, unsupported type returns error
- [x] Unit tests for re-encryption controller reconcile — no encryption, first deploy init, hashes match, KAS not converged, migrations in progress, all succeeded, failed with retry/prune, cache not synced
- [x] `go build ./...` passes
- [x] `go test -race` passes for all modified packages
- [ ] E2E: AWS KMS key rotation with re-encryption verification
- [ ] E2E: Azure KMS key rotation with re-encryption verification
- [ ] E2E: AESCBC key rotation with re-encryption verification

## Summary by CodeRabbit

* **New Features**
  * Added etcd data encryption status tracking and re-encryption progress monitoring to hosted clusters.
  * Extended encryption key status fields across hosted control planes and multiple KMS providers.
  * Introduced encryption migration history tracking and lifecycle state management.

* **Chores**
  * Updated storage version migrator dependency classification.


control-plane-operator/controllers/hostedcontrolplane/v2/kas/secretencryption.go
```diff
@@ -85,6 +94,23 @@ func adaptSecretEncryptionConfig(cpContext component.WorkloadContext, secret *co
...
+	}
+	if fingerprint != "" {
+		if secret.Annotations == nil {
+			secret.Annotations = map[string]string{}
+		}
+		existingHash := secret.Annotations[EncryptionActiveKeyHashAnnotation]
+		if existingHash != fingerprint {
+			secret.Annotations[EncryptionActiveKeyHashAnnotation] = fingerprint
+			secret.Annotations[EncryptionRekeyNeededAnnotation] = "true"
+		}
```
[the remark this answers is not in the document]
muraee: Done. Now always persists the fingerprint, and only sets `rekey-needed=true` when a previous hash exists and differs — avoiding spurious migrations on new or upgrading clusters.

---

control-plane-operator/hostedclusterconfigoperator/controllers/reencryption/reencryption.go
```diff
@@ -0,0 +1,288 @@
...
+	}
+
+	// All migrations succeeded — remove the rekey-needed annotation and set condition True.
+	log.Info("All re-encryption migrations completed successfully")
+	if err := r.removeRekeyAnnotation(ctx, encryptionSecret); err != nil {
+		return reconcile.Result{}, fmt.Errorf("failed to remove rekey-needed annotation: %w", err)
+	}
+	setReEncryptionCondition(hcp, metav1.ConditionTrue, hyperv1.ReEncryptionCompletedReason,
+		"All etcd data has been re-encrypted with the active encryption key")
+	return reconcile.Result{}, nil
```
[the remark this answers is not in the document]
muraee: Done. Reordered to set the condition before removing the annotation. If annotation removal fails, the next reconcile will retry (all migrations already complete) rather than leaving the condition stuck in InProgress.

---

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -894,6 +894,15 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
...
 	}
 
+	// Bubble up the EtcdDataEncryptionUpToDate condition from HCP when encryption is configured.
+	if hcp != nil && hcluster.Spec.SecretEncryption != nil {
+		cond := meta.FindStatusCondition(hcp.Status.Conditions, string(hyperv1.EtcdDataEncryptionUpToDate))
+		if cond != nil {
+			cond.ObservedGeneration = hcluster.Generation
+			meta.SetStatusCondition(&hcluster.Status.Conditions, *cond)
+		}
+	}
```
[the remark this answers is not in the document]
muraee: Done. Now removes the condition when encryption is disabled, HCP is nil, or HCP no longer reports the condition — preventing stale values.

---

control-plane-operator/hostedclusterconfigoperator/controllers/reencryption/reencryption.go
```diff
@@ -0,0 +1,288 @@
...
+	hcpName   string
+}
+
+func (r *reEncryptionReconciler) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
+	log := ctrl.LoggerFrom(ctx)
+
+	hcp := resourcemanifests.HostedControlPlane(r.namespace, r.hcpName)
+	if err := r.cpClient.Get(ctx, client.ObjectKeyFromObject(hcp), hcp); err != nil {
+		return reconcile.Result{}, fmt.Errorf("failed to get HCP %s: %w", req, err)
+	}
```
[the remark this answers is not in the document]
muraee: Done. The encryption config secret and KAS deployment Get calls now handle NotFound by logging and returning a requeue instead of a hard error, so the controller idles during bootstrap/teardown rather than entering error backoff.

---

control-plane-operator/hostedclusterconfigoperator/controllers/reencryption/reencryption.go
```diff
@@ -0,0 +1,288 @@
...
+
+	// Get the kas-secret-encryption-config secret.
+	encryptionSecret := manifests.KASSecretEncryptionConfigFile(r.namespace)
+	if err := r.cpClient.Get(ctx, client.ObjectKeyFromObject(encryptionSecret), encryptionSecret); err != nil {
+		return reconcile.Result{}, fmt.Errorf("failed to get encryption config secret: %w", err)
+	}
+
+	// Check if re-encryption is needed.
+	rekeyNeeded := encryptionSecret.Annotations[kas.EncryptionRekeyNeededAnnotation]
+	activeKeyHash := encryptionSecret.Annotations[kas.EncryptionActiveKeyHashAnnotation]
```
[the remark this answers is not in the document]
muraee: This is a valid concern for the A→B→A rotation scenario. However, this is handled correctly by the existing design:

1. **`EnsureMigration` detects stale CRs**: When the writeKey annotation on the existing StorageVersionMigration CR doesn't match, the migrator deletes and recreates the CR (see `kubestorageversionmigrator.go:70-76`). So a B→A rotation would find the old A migration, see a different writeKey (from the A→B rotation's hash), and create a fresh migration.

2. **The A→B→A case where hashes collide**: This requires two rotations happening before the first migration completes. The enhancement explicitly documents this as a "must not do" scenario — a second key rotation while re-encryption is in progress is unsupported and will be guarded by a webhook in a follow-up (the `rekey-needed` annotation being present signals an active migration).

3. **The fingerprint includes key identity, not just key name**: For AESCBC it includes the actual key data hash, and for KMS it includes the full key path/version, making accidental collisions extremely unlikely.

A monotonic counter would add complexity (persisted state that must survive restarts) for a scenario that's already documented as unsupported. Happy to revisit if the follow-up webhook doesn't materialize.

---

devguyio:
I now have all the information needed for a complete root cause analysis. Here is the report:

---

# Test Failure Analysis Complete

## Job Information
- **Job Type**: GitHub Actions (Gitlint)
- **Run URL**: https://github.com/openshift/hypershift/actions/runs/24346700175/job/71089453364
- **PR**: [#8219](https://github.com/openshift/hypershift/pull/8219) — *feat: etcd data re-encryption after encryption key rotation*
- **Branch**: `etcd-reencryption-on-key-rotation`

## Test Failure Analysis

### Error
```
Commit fab20778d3:
1: CT1 Title does not start with one of fix, feat, chore, docs, style, refactor, perf, test, revert, ci, build:
   "vendor: add library-go migrators and kube-storage-version-migrator informer/lister"
```
### Root Cause
<details> 
<summary>The first commit in the PR (`fab20778d3`) has a title that **violates the Conventional Commits specification** enforced by the repository's `.gitlint` configuration.</summary>

The commit title is:
```
vendor: add library-go migrators and kube-storage-version-migrator informer/lister
```

The `.gitlint` configuration at the repository root requires titles to follow Conventional Commits format, with the allowed type prefixes being:

```
types = fix,feat,chore,docs,style,refactor,perf,test,revert,ci,build
```

The prefix `vendor` is **not** in the allowed list. The gitlint `contrib-title-conventional-commits` rule (CT1) rejects any commit title that does not begin with one of these types followed by a colon.

The second commit (`d5e509743d`) passes validation — its title `feat: add etcd data re-encryption after encryption key rotation` correctly uses the `feat` prefix.

</details>

### Recommendations
<details>
<summary> These are some recommendations based on the failure analysis </summary>
1. **Reword the first commit's title** to use an allowed Conventional Commits prefix. The most appropriate prefix for a vendoring commit is `chore` or `build`:
   ```
   chore: vendor library-go migrators and kube-storage-version-migrator informer/lister
   ```
   or:
   ```
   build: vendor library-go migrators and kube-storage-version-migrator informer/lister
   ```

2. **Interactive rebase** to amend the commit message:
   ```bash
   git rebase -i HEAD~2
   # Change 'pick' to 'reword' for the first commit
   # Update the title to use chore: or build: prefix
   git push --force-with-lease
   ```
</details>

### Evidence
<details>
<summary> This is some collected evidence that helped reach the mentioned root cause</summary>

| Evidence | Detail |
|----------|--------|
| **Failing commit** | `fab20778d318dbac81809b5f18434bbfacd3716e` |
| **Commit title** | `vendor: add library-go migrators and kube-storage-version-migrator informer/lister` |
| **Gitlint rule violated** | `CT1` (contrib-title-conventional-commits) |
| **Invalid prefix used** | `vendor` |
| **Allowed prefixes** | `fix`, `feat`, `chore`, `docs`, `style`, `refactor`, `perf`, `test`, `revert`, `ci`, `build` |
| **Passing commit** | `d5e509743d` — title correctly uses `feat:` prefix |

</details>

muraee:
Re: @coderabbitai's [comment](https://github.com/openshift/hypershift/pull/8219/files#r3078729163) on AESCBC key validation:

Done. Added validation to return empty fingerprint when key data is missing or empty.

---

muraee:
Re: @coderabbitai's [comment](https://github.com/openshift/hypershift/pull/8219/files#r3078729155) on missing migrated-hash during rotation:

The `encryption-rekey-needed` annotation was intentionally removed in this revision. The controller now detects re-encryption need by comparing the active key fingerprint (computed from HCP spec) against `encryption-migrated-key-hash` on the secret — no boolean signal involved.

The scenario described (secret recreated during pending rotation) requires manual deletion of an operator-managed secret, which would also disrupt KAS itself. In that edge case, the controller initializes the migrated hash to the current active key, which is correct — the CPO will recreate the secret with the current encryption config, and KAS will use the current key going forward. If re-encryption for the previous key was incomplete, the next key rotation will trigger a full re-encryption.

---

control-plane-operator/hostedclusterconfigoperator/controllers/reencryption/reencryption.go
```diff
@@ -0,0 +1,393 @@
...
+		if err := c.Get(ctx, client.ObjectKeyFromObject(activeKeySecret), activeKeySecret); err != nil {
+			return "", fmt.Errorf("failed to get aescbc active key secret for fingerprint: %w", err)
+		}
+		keyData, ok := activeKeySecret.Data[hyperv1.AESCBCKeySecretKey]
+		if !ok || len(keyData) == 0 {
+			return "", nil
+		}
+		input = fmt.Sprintf("aescbc/%x", sha256.Sum256(keyData))
+	default:
+		return "", nil
```
cblecker: **critical: Silent empty return for unknown encryption types**

The outer `default` case silently returns `("", nil)` for any unrecognized `SecretEncryption.Type`. If a new encryption type is added to the API without updating this function, re-encryption will silently never trigger — no error, no log, no condition. The user gets no indication that their data isn't being re-encrypted.

Contrast with line 351 where an unsupported KMS *provider* correctly returns an error:
```go
return "", fmt.Errorf("unsupported KMS provider: %s", secretEncryption.KMS.Provider)
```

The outer default should follow the same pattern:
```go
default:
    return "", fmt.Errorf("unsupported encryption type: %s", secretEncryption.Type)
```

api/hypershift/v1beta1/hostedcluster_conditions.go
```diff
@@ -322,6 +329,11 @@ const (
 	AutoNodeNotConfiguredReason    = "AutoNodeNotConfigured"
 	AutoNodeProgressingReason      = "AutoNodeProgressing"
 	AutoNodeEvaluationFailedReason = "AutoNodeEvaluationFailed"
+
+	ReEncryptionInProgressReason = "ReEncryptionInProgress"
+	ReEncryptionCompletedReason  = "ReEncryptionCompleted"
+	ReEncryptionFailedReason     = "ReEncryptionFailed"
+	ReEncryptionWaitingForKAS    = "WaitingForKASConvergence"
```
cblecker: **important: Missing `Reason` suffix**

The other three re-encryption reason constants all use the `*Reason` suffix (`ReEncryptionInProgressReason`, `ReEncryptionCompletedReason`, `ReEncryptionFailedReason`), but this one omits it. The overwhelming majority of reason constants in this file follow this convention.

```suggestion
	ReEncryptionWaitingForKASReason = "WaitingForKASConvergence"
```

(All references would need updating too.)

control-plane-operator/hostedclusterconfigoperator/controllers/reencryption/reencryption.go
```diff
@@ -0,0 +1,393 @@
...
+	}
+	originalHCP := hcp.DeepCopy()
+
+	result, err := r.reconcile(ctx, hcp)
+
+	// Update HCP status if changed.
+	if !reflect.DeepEqual(hcp.Status, originalHCP.Status) {
+		log.Info("Updating HCP status with re-encryption condition")
+		if updateErr := r.cpClient.Status().Update(ctx, hcp); updateErr != nil {
+			return reconcile.Result{}, fmt.Errorf("failed to update HCP status: %w", updateErr)
```
cblecker, line 153: **important: Original reconcile error lost when status update also fails**

If `r.reconcile()` returns an error AND has mutated `hcp.Status`, and then `Status().Update()` also fails (e.g., conflict), the original reconcile error is silently dropped — only the status update error is returned. This makes debugging harder because the root cause error is lost.

Consider combining both errors:
```go
if updateErr := r.cpClient.Status().Update(ctx, hcp); updateErr != nil {
    if err != nil {
        return reconcile.Result{}, fmt.Errorf("failed to update HCP status: %w (original error: %v)", updateErr, err)
    }
    return reconcile.Result{}, fmt.Errorf("failed to update HCP status: %w", updateErr)
}
```

Or check the reconcile error first and return it immediately (the `hcpstatus` controller follows a similar pattern where it checks reconcile error before status update). In practice, the double-failure case is rare, but when it happens the lost error makes diagnosis very difficult.

control-plane-operator/hostedclusterconfigoperator/controllers/reencryption/reencryption.go
```diff
@@ -0,0 +1,393 @@
...
+)
+
+const (
+	ControllerName = "reencryption"
+
+	// EncryptionMigratedKeyHashAnnotation stores the fingerprint of the encryption
+	// key that etcd data was last re-encrypted with. Written by this controller
+	// after all StorageVersionMigrations complete. Compared against the active key
+	// fingerprint (computed from HCP spec) to decide whether re-encryption is needed.
+	EncryptionMigratedKeyHashAnnotation = "hypershift.openshift.io/encryption-migrated-key-hash"
```
cblecker, line 50: **important: Comment is factually incomplete**

The comment says this annotation is "Written by this controller after all StorageVersionMigrations complete," but the annotation is also written at line 192 during first-deployment initialization *without* any migrations running. A future maintainer reading only this comment would incorrectly conclude that the annotation's presence always implies a full re-encryption cycle completed.

Consider something like:
> Stores the fingerprint of the encryption key that etcd data is known to be encrypted with. Initialized on first deployment (assuming data is already encrypted with the current key), and updated after all StorageVersionMigrations complete following a key rotation.

control-plane-operator/hostedclusterconfigoperator/controllers/reencryption/reencryption.go
```diff
@@ -0,0 +1,393 @@
...
+	})
+	if err := c.Watch(source.Kind[client.Object](opts.CPCluster.GetCache(), &appsv1.Deployment{},
+		handler.EnqueueRequestsFromMapFunc(mapToHCP), deploymentPredicate)); err != nil {
+		return fmt.Errorf("failed to watch KAS deployment: %w", err)
+	}
+
+	// Register an event handler on the SVM informer so the migrator's cache stays
+	// in sync. The controller reconciles on a requeue interval while migrations are
+	// in progress, so we just need a no-op handler for cache sync.
+	if _, err := migratorInstance.AddEventHandler(cache.ResourceEventHandlerFuncs{}); err != nil {
```
cblecker: **suggestion: Comment should explain the nil-deref risk if this is removed**

The no-op `cache.ResourceEventHandlerFuncs{}` looks removable, but `AddEventHandler` actually initializes the `cacheSynced` function pointer inside `KubeStorageVersionMigrator` (see `kubestorageversionmigrator.go:47`). Without this call, `HasSynced()` at line 224 would panic with a nil pointer dereference.

Consider updating the comment:
```go
// Register an event handler on the SVM informer. This call is required
// because AddEventHandler initializes the migrator's internal cacheSynced
// function; without it, HasSynced() will panic. The handler itself is a
// no-op because the controller reconciles on a requeue interval while
// migrations are in progress.
```

control-plane-operator/hostedclusterconfigoperator/controllers/reencryption/reencryption.go
```diff
@@ -0,0 +1,393 @@
...
+	// Compare the active key hash against the last-migrated key hash.
+	migratedKeyHash := encryptionSecret.Annotations[EncryptionMigratedKeyHashAnnotation]
+	if migratedKeyHash == "" {
+		// First deployment or upgrade — no prior migration record.
+		// Initialize the migrated hash to the active hash without re-encrypting,
+		// since the data is already encrypted with the current key.
+		log.Info("Initializing migrated key hash (first deployment)", "activeKeyHash", activeKeyHash)
+		if err := r.setMigratedKeyHash(ctx, encryptionSecret, activeKeyHash); err != nil {
+			return reconcile.Result{}, fmt.Errorf("failed to initialize migrated key hash: %w", err)
+		}
```
cblecker, line 204: **suggestion: Silent re-initialization risk if annotation is deleted**

If the `EncryptionMigratedKeyHashAnnotation` is accidentally deleted (by a user, a bug, or a controller recreating the secret) after a key rotation has occurred but before re-encryption completed, the controller will treat this as "first deployment" and initialize the migrated hash to the active hash — silently skipping re-encryption of data still encrypted with the old key.

Consider logging at `Warning` level instead of `Info` here, so accidental annotation deletion is at least observable. Or adding a brief comment noting the assumption (that the annotation absence genuinely means first deployment, not data loss).

api/hypershift/v1beta1/hostedcluster_conditions.go
```diff
@@ -243,6 +243,13 @@ const (
 	// **False / AutoNodeProgressing** means AutoNode is being enabled or disabled — the operation is in progress.
 	// **False / AutoNodeNotConfigured** means AutoNode is not configured in the spec and all Karpenter components have been removed.
 	AutoNodeEnabled ConditionType = "AutoNodeEnabled"
+
+	// EtcdDataEncryptionUpToDate indicates whether all etcd data is encrypted
+	// with the currently active encryption key.
+	// **True** means all data has been confirmed encrypted with the active key.
+	// **False** means re-encryption is in progress or has failed.
+	// Absent: encryption is not configured or no key rotation has occurred.
```
cblecker: **suggestion: Condition doc could be slightly more precise**

Two minor points:
1. `**False**` can also mean `WaitingForKASConvergence` (a distinct pre-condition state), not just "in progress or failed." Consider: `**False** means the KAS deployment has not yet converged, re-encryption is in progress, or re-encryption has failed. Check the Reason field for specifics.`
2. `Absent:` uses different formatting than `**True**` and `**False**` above it. Consider `**Absent**` for consistency.

cblecker (commented):
## Review Summary

This is a solid PR overall — clean controller design, good test coverage, and the security-conscious approach to fingerprinting (SHA-256 of key identity, no raw material in annotations) is well done. A few things worth addressing before merge.

### Things to fix

The most important issue: `computeActiveKeyFingerprint` silently returns `("", nil)` for any unrecognized `SecretEncryption.Type` (`reencryption.go:371-372`). If someone adds a new encryption type to the API and forgets to update this function, re-encryption will just never trigger — no error, no log, nothing. This is especially surprising given that line 351 already correctly returns an error for unsupported KMS providers. The outer default case should do the same.

The PR description needs updating — it describes the CPO computing fingerprints and setting `EncryptionActiveKeyHashAnnotation` / `EncryptionRekeyNeededAnnotation` annotations, but the actual implementation has the HCCO computing fingerprints directly from the HCP spec using `EncryptionMigratedKeyHashAnnotation`. The description should match what was built.

`ReEncryptionWaitingForKAS` at `hostedcluster_conditions.go:336` is missing the `Reason` suffix that the other three constants have (`ReEncryptionInProgressReason`, `ReEncryptionCompletedReason`, `ReEncryptionFailedReason`). Should be `ReEncryptionWaitingForKASReason`.

In `reencryption.go:140-146`, if `r.reconcile()` fails and the subsequent `Status().Update()` also fails, the original reconcile error gets silently dropped. Only the status update error bubbles up, which will make debugging painful.

The comment on `EncryptionMigratedKeyHashAnnotation` (`reencryption.go:44-48`) says it's "Written by this controller after all StorageVersionMigrations complete," but it's also written at line 192 during first-deployment initialization with no migrations involved. Worth fixing so it doesn't mislead someone later.

### Worth considering

The no-op `AddEventHandler` at `reencryption.go:113-116` looks like dead code, but it actually initializes the `cacheSynced` function pointer inside `KubeStorageVersionMigrator` — without it `HasSynced()` would panic. The comment should explain *why* it's there so nobody helpfully removes it.

If the `EncryptionMigratedKeyHashAnnotation` gets accidentally deleted after a key rotation (`reencryption.go:187-194`), the controller treats it as a fresh first deployment and just stamps the current hash without re-encrypting. The log at line 191 says "first deployment" which would be confusing in that scenario. A warning-level log with a more descriptive message would help.

On the testing side: the `EtcdDataEncryptionUpToDate` condition bubble-up logic in `hostedcluster_controller.go:897-906` has three branches and is the user-facing surface of this feature, but it's untested. Similarly, `EnsureMigration` returning an infrastructure error is never exercised in tests since the mock always returns nil for the error. And there are no tests for NotFound scenarios (encryption config secret absent, KAS deployment absent, AESCBC key secret absent) — these are handled gracefully in the code but would benefit from coverage.

Minor: the condition documentation at `hostedcluster_conditions.go:250-251` is inconsistent — the `**False**` description omits the `WaitingForKASConvergence` reason, and `Absent:` uses different formatting than `**True**`/`**False**` above it.

### What's good

The controller design is clean — separating outer `Reconcile` for status orchestration from inner `reconcile` for business logic is a nice pattern. The test suite is thorough with 9 table-driven cases plus dedicated fingerprint and prune tests, all following project conventions (Gherkin naming, gomega, `WithStatusSubresource`). The condition bubble-up is actually more robust than existing patterns like `ValidAWSKMSConfig`/`ValidAzureKMSConfig` since it properly removes the condition when the HCP is nil or the condition is absent. Good reuse of existing utilities throughout.

everettraven (commented):
I agree with @cblecker 's comments related to GoDoc verbosity on the new condition type for clarifying what each state+reason pairing represents, but other than that this LGTM from an API perspective.

control-plane-operator/hostedclusterconfigoperator/controllers/reencryption/reencryption.go
muraee: Done. Changed the outer `default` to return an error for unsupported encryption types. Added a corresponding test case in `TestComputeActiveKeyFingerprint`.

---

api/hypershift/v1beta1/hostedcluster_conditions.go
muraee: Done. Renamed `ReEncryptionWaitingForKAS` to `ReEncryptionWaitingForKASReason` for consistency with the other three reason constants. Updated all references.

---

control-plane-operator/hostedclusterconfigoperator/controllers/reencryption/reencryption.go
muraee: Done. Now preserves the original reconcile error in the status update failure message: `fmt.Errorf("failed to update HCP status: %w (original error: %v)", updateErr, err)`.

---

control-plane-operator/hostedclusterconfigoperator/controllers/reencryption/reencryption.go
muraee: Done. Updated the annotation comment to describe the full lifecycle: initialized on first deployment assuming data is already encrypted, and updated after all StorageVersionMigrations complete following a key rotation. Also noted the comparison against the active key fingerprint.

---

control-plane-operator/hostedclusterconfigoperator/controllers/reencryption/reencryption.go
muraee: Done. Updated the comment to explain that `AddEventHandler` initializes the migrator's internal `cacheSynced` function pointer, and without it `HasSynced()` will panic with a nil dereference.

---

control-plane-operator/hostedclusterconfigoperator/controllers/reencryption/reencryption.go
muraee: Done. Changed to `log.V(1).Info(...)` so it appears in debug output, and added a caveat comment noting that if the annotation is accidentally deleted after a key rotation, this will skip re-encryption for that rotation but the next rotation will trigger a full re-encryption.

---

api/hypershift/v1beta1/hostedcluster_conditions.go
muraee: Done. Added `**Absent**` with bold formatting for consistency, and updated the `**False**` description to mention the WaitingForKASConvergence reason alongside in-progress and failed states.

---

everettraven:
Approved from API perspective.

/approve

muraee:
@coderabbitai review

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -2074,10 +2074,151 @@ type AESCBCSpec struct {
...
+	To EncryptionKeyReference `json:"to,omitzero"`
+	// state tracks the current phase of this rotation.
+	// +required
+	State EncryptionMigrationState `json:"state,omitempty"`
+	// startedTime is when the rotation was initiated.
+	// +required
+	StartedTime metav1.Time `json:"startedTime,omitempty"`
+	// completionTime is when the rotation finished. Not set while the rotation is in progress.
+	// +optional
+	CompletionTime *metav1.Time `json:"completionTime,omitempty"`
```
enxebre: why is this one a pointer?

api/hypershift/v1beta1/ibmcloud.go
```diff
@@ -44,6 +44,41 @@ type IBMCloudKMSKeyEntry struct {
 	KeyVersion int `json:"keyVersion"`
 }
 
+// IBMCloudKMSKeyStatus contains identity fields for an IBM Cloud KMS key list entry,
+// sufficient to reconstruct the KP_DATA_JSON entry for the backup key.
+// +k8s:deepcopy-gen=true
+type IBMCloudKMSKeyStatus struct {
+	// crkID is the Customer Root Key ID.
```
enxebre: is there any regexps we can include here for all this fields?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -2074,10 +2074,151 @@ type AESCBCSpec struct {
...
+
+const (
+	SecretEncryptionProviderAzure    SecretEncryptionProvider = "Azure"
+	SecretEncryptionProviderAWS      SecretEncryptionProvider = "AWS"
+	SecretEncryptionProviderIBMCloud SecretEncryptionProvider = "IBMCloud"
+	SecretEncryptionProviderAESCBC   SecretEncryptionProvider = "AESCBC"
+)
+
+// SecretEncryptionStatus tracks the state of secret encryption key rotation and re-encryption.
+// +k8s:deepcopy-gen=true
```
enxebre, line 2096: why is +k8s:deepcopy-gen needed?

api/hypershift/v1beta1/aws.go
```diff
@@ -1082,6 +1085,22 @@ type AWSKMSKeyEntry struct {
...
 }
 
+// AWSKMSKeyStatus contains identity fields for an AWS KMS key, sufficient to
+// reconstruct the backup sidecar container arguments.
+// +k8s:deepcopy-gen=true
+type AWSKMSKeyStatus struct {
+	// arn is the Amazon Resource Name of the KMS key.
+	// +required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=2048
```
enxebre, line 1085: please include regex here, there's precedents for arns

control-plane-operator/controllers/hostedcontrolplane/v2/cvo/deployment.go
```diff
@@ -284,6 +284,8 @@ func resourcesToRemove(platformType hyperv1.PlatformType) []client.Object {
 			&appsv1.Deployment{ObjectMeta: metav1.ObjectMeta{Name: "aws-ebs-csi-driver-operator", Namespace: "openshift-cluster-csi-drivers"}},
 			&appsv1.Deployment{ObjectMeta: metav1.ObjectMeta{Name: "aws-ebs-csi-driver-controller", Namespace: "openshift-cluster-csi-drivers"}},
 			&appsv1.Deployment{ObjectMeta: metav1.ObjectMeta{Name: "csi-snapshot-controller", Namespace: "openshift-cluster-storage-operator"}},
+			&appsv1.Deployment{ObjectMeta: metav1.ObjectMeta{Name: "kube-storage-version-migrator-operator", Namespace: "openshift-kube-storage-version-migrator-operator"}},
```
enxebre, line 287: is there a counter part PR to drop the ibmcloud profile annotation from this resource?

control-plane-operator/controllers/hostedcontrolplane/v2/assets/kube-storage-version-migrator/deployment.yaml
```diff
@@ -0,0 +1,56 @@
...
+          secretName: service-network-admin-kubeconfig
+      containers:
+      - name: migrator
+        image: kube-storage-version-migrator
+        command:
+          - migrator
+        args:
+          - '--alsologtostderr'
+          - '--v=2'
+          - '--kubeconfig=/etc/kubernetes/kubeconfig'
```
enxebre, line 33: which kubeconfig is this new component using?

api/hypershift/v1beta1/ibmcloud.go
```diff
@@ -44,6 +44,41 @@ type IBMCloudKMSKeyEntry struct {
...
+	// +kubebuilder:validation:MaxLength=255
+	CRKID string `json:"crkID,omitempty"`
+	// instanceID is the KMS instance ID.
+	// +required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=255
+	InstanceID string `json:"instanceID,omitempty"`
+	// keyVersion is the key version number.
+	// +required
+	// +kubebuilder:validation:Minimum=0
```
enxebre: what's the use case for KeyVersion=0?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -2074,10 +2074,151 @@ type AESCBCSpec struct {
...
+type EncryptionMigrationHistory struct {
+	// from is the key that data was migrated from (the previous active key).
+	// +required
+	From EncryptionKeyReference `json:"from,omitzero"`
+	// to is the key that data was migrated to (the target key).
+	// +required
+	To EncryptionKeyReference `json:"to,omitzero"`
+	// state tracks the current phase of this rotation.
+	// +required
+	State EncryptionMigrationState `json:"state,omitempty"`
```
enxebre, line 2209: should this be an enum?

api/hypershift/v1beta1/ibmcloud.go
muraee: addressed by reusing the existing spec field.

api/hypershift/v1beta1/hostedcluster_types.go
muraee: removed, we are reusing the existing spec field.

api/hypershift/v1beta1/aws.go
muraee: addressed by reusing the existing spec field.

control-plane-operator/controllers/hostedcontrolplane/v2/cvo/deployment.go
muraee: no, we want to drop it conditionally here only because IBM don't want have this extra compoenent running mgmt side. It will keep running in the data-plane for them

api/hypershift/v1beta1/ibmcloud.go
muraee: addressed by reusing the existing spec field.

control-plane-operator/controllers/hostedcontrolplane/v2/assets/kube-storage-version-migrator/deployment.yaml
muraee:
```yaml
volumes:
      - name: kubeconfig
        secret:
          secretName: service-network-admin-kubeconfig
```

api/hypershift/v1beta1/hostedcluster_types.go
muraee: it is. The enum validation is applied on the `EncryptionMigrationState` type definition.

api/hypershift/v1beta1/hostedcluster_types.go
muraee: fixed

api/hypershift/v1beta1/aws.go
```diff
@@ -1082,6 +1085,22 @@ type AWSKMSKeyEntry struct {
 	ARN string `json:"arn"`
 }
 
+// AWSKMSKeyStatus contains identity fields for an AWS KMS key, sufficient to
+// reconstruct the backup sidecar container arguments.
+// +k8s:deepcopy-gen=true
```
enxebre: why is // +k8s:deepcopy-gen=true needed?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -2074,10 +2074,150 @@ type AESCBCSpec struct {
...
+	SecretEncryptionProviderAzure    SecretEncryptionProvider = "Azure"
+	SecretEncryptionProviderAWS      SecretEncryptionProvider = "AWS"
+	SecretEncryptionProviderIBMCloud SecretEncryptionProvider = "IBMCloud"
+	SecretEncryptionProviderAESCBC   SecretEncryptionProvider = "AESCBC"
+)
+
+// SecretEncryptionStatus tracks the state of secret encryption key rotation and re-encryption.
+// +k8s:deepcopy-gen=true
+// +kubebuilder:validation:MinProperties=1
+type SecretEncryptionStatus struct {
```
enxebre, line 2098: should this be EtcdEncryptionStatus?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -2074,10 +2074,150 @@ type AESCBCSpec struct {
...
+	// history[0] is not Completed or Interrupted.
+	// +optional
+	// +listType=atomic
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=5
+	History []EncryptionMigrationHistory `json:"history,omitempty"`
+}
+
+// SecretEncryptionKeyStatus records the active key identity using the same types as the spec.
+// +k8s:deepcopy-gen=true
```
enxebre: // +k8s:deepcopy-gen=true?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -2074,10 +2074,150 @@ type AESCBCSpec struct {
...
+	From EncryptionKeyReference `json:"from,omitzero"`
+	// to is the key that data was migrated to (the target key).
+	// +required
+	To EncryptionKeyReference `json:"to,omitzero"`
+	// state tracks the current phase of this rotation.
+	// +required
+	State EncryptionMigrationState `json:"state,omitempty"`
+	// startedTime is when the rotation was initiated.
+	// +required
+	StartedTime metav1.Time `json:"startedTime,omitempty"`
```
enxebre: omitzero?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -2074,10 +2074,150 @@ type AESCBCSpec struct {
...
+	AESCBC AESCBCKeyStatus `json:"aescbc,omitzero"`
+}
+
+// AESCBCKeyStatus contains a reference to the AESCBC key secret and a SHA-256 hash
+// of its contents for fingerprinting.
+// +k8s:deepcopy-gen=true
+type AESCBCKeyStatus struct {
+	// secret is a reference to the secret containing the AESCBC key.
+	// +required
+	Secret corev1.LocalObjectReference `json:"secret,omitempty"`
```
enxebre: omitzero?

enxebre:
## Code Review: PR #8219

### Blocking Issues

**B1. Duplicate Azure KMS token minter container (copy-paste bug)**
`control-plane-operator/controllers/hostedcontrolplane/v2/kas/kms/azure.go:192-200`

The PR introduces a duplicate `if p.isSelfManaged` block that appends `kasContainerAzureKMSTokenMinter` twice. This produces a pod spec with duplicate container names, which Kubernetes rejects. The second block (lines 197-200) must be removed.

**B2. `IBMCloudKMSKeyEntry.KeyVersion` (int) with `omitempty` drops valid zero values**
`api/hypershift/v1beta1/ibmcloud.go:44`

`KeyVersion` has `+kubebuilder:validation:Minimum=0` meaning `0` is valid, but `omitempty` on `int` omits the field when it equals `0`. A user setting `keyVersion: 0` would have it silently dropped during serialization, then rejected as missing by CRD validation. This is a pre-existing issue but the PR widens the impact by reusing `IBMCloudKMSKeyEntry` in `SecretEncryptionKeyStatus`. Consider using a `*int` with `omitempty`, or keeping `int` without `omitempty` for this specific field.

---

### Suggestions

#### API

- `AESCBCKeyStatus.Secret` (`corev1.LocalObjectReference`) should use `omitzero` not `omitempty` — `omitempty` does nothing for non-pointer structs per repo conventions. (`hostedcluster_types.go:~2154`)
- `EncryptionMigrationHistory.StartedTime` (`metav1.Time`) should use `omitzero` not `omitempty` for consistency with `CompletionTime`. (`hostedcluster_types.go:~2215`)
- `ReEncryptionPersistentFailureReason` is defined but never used anywhere in the PR. Remove it or use it. (`hostedcluster_conditions.go:352`)

#### Controller Design

- VAP `FailurePolicy` is `Ignore` (`hypershift-operator/main.go:1031`). The rotation guard is advisory — if the webhook infrastructure is unhealthy, users can trigger concurrent rotations. Consider `Fail` since the controller handles it gracefully anyway.
- `deriveTargetKeyRole` returns a hard error when the encryption config Secret doesn't exist (`reencryption.go:~282`). Consider handling `IsNotFound` by returning `TargetKeyAbsent, nil` — during initial setup the Secret may not exist yet.
- `handleEncryptionNotConfigured` clears `TargetKey` and removes the condition but leaves `ActiveKey` untouched (`reencryption.go:102-106`). If encryption is later re-enabled, the stale `ActiveKey` could cause the controller to skip bootstrap and start a spurious rotation. Consider also clearing `ActiveKey`.
- Overly broad Deployment and Secret watches in `setup.go:80-90` — every Secret/Deployment change in the CP namespace triggers reconciliation. Filter by name (e.g., `kube-apiserver` Deployment, encryption-relevant Secrets only).

#### DRY

- Duplicate `isKASConverged` implementations in `kas/secretencryption.go:72-82` and `reencryption.go:445-457`. Extract to a shared helper in `support/`.
- EncryptionConfiguration wrapping boilerplate is repeated in `kms/aws.go`, `kms/azure.go`, `kms/ibmcloud.go`, and `kas/aescbc.go`. Consider a shared `NewEncryptionConfiguration(resources, providers)` helper.
- Requeue interval magic numbers (`10s`, `30s`, `60s`) repeated throughout `reencryption.go`. Define as named constants.

#### Idiomatic Go

- `handleMigratingPhase` discards context with `_ context.Context` (`reencryption.go:~218`). Accept as `ctx` for forward compatibility.
- `recordMigrationDuration` is called on every status patch, not just on state transitions. This inflates the histogram. Gate on transition from non-completed to completed.
- `FingerprintFromKeyStatus` returns `""` as sentinel for unknown/missing providers instead of an error. Two different empty-status keys would compare as equal. Consider returning an error for unknown providers.

#### Test Coverage

- `DecodeEncryptionConfiguration` has no direct unit tests (valid parsing, malformed input, empty bytes).
- `FingerprintFromKeyStatus` is missing IBMCloud and AESCBC provider test cases, plus all empty-field early-return branches.
- `KeyStatusFromSpec` is missing AWS KMS and IBMCloud paths.
- `ibmcloud.go` (kms package) has no test file — `NewIBMCloudKMSProvider`, `GenerateKMSEncryptionConfig`, `GenerateKMSPodConfig` are untested.
- No test for controller restart recovery (stale history state vs actual cluster state).

---

muraee:
Thanks for the review! Addressed all items in the latest push:

**Fixed:** B1 (duplicate token minter), B2 (reverted `omitempty` on `KeyVersion`), `omitzero` on struct fields, unused constant, `IsNotFound` handling, broad watches, requeue magic numbers, unused context param, histogram inflation, and added missing test coverage.

**Not changed:**
- **VAP `FailurePolicy: Ignore`** — `Fail` would block all HostedCluster updates (including unrelated ones) if the VAP infrastructure is unhealthy. The controller already handles concurrent rotations gracefully by queuing, so `Ignore` is the safer operational default.
- **`handleEncryptionNotConfigured` leaving `ActiveKey`** — this is correct. `ActiveKey` records what key the data is *actually* encrypted with. Clearing it would lose that information, causing bootstrap to run instead of a proper rotation if encryption is re-enabled. The broader question of "should removing encryption trigger decryption" is a separate feature.
- **EncryptionConfiguration wrapping boilerplate** — pre-existing code across KMS providers, not introduced by this PR.
- **Duplicate `isKASConverged`** — both are thin wrappers around the shared `podspec.IsDeploymentReady`. Extracting to a shared package would add more import complexity than it saves. Fixed the HCCO version to handle `IsNotFound` consistently with the CPO version.
- **`FingerprintFromKeyStatus` returning `""` for unknown providers** — invalid providers are validated at the API level, so this code path is unreachable in practice.

api/hypershift/v1beta1/hostedcluster_types.go
muraee: makes sense. However, this is more consistent with the spec field `SecretEncryption`. Happy to change it though.

api/hypershift/v1beta1/hostedcluster_types.go
muraee: removed

hypershift-operator/main.go
```diff
@@ -1013,3 +1018,71 @@ func reconcileDeprecationValidatingAdmissionPolicy(ctx context.Context, client c
...
+						admissionregistrationv1.Update,
+					},
+					Rule: admissionregistrationv1.Rule{
+						APIGroups:   []string{"hypershift.openshift.io"},
+						APIVersions: []string{"v1beta1"},
+						Resources:   []string{"hostedclusters"},
+					},
+				},
+			},
+		}
```
csrwng, line 1055: This VAP blocks spec changes during active re-encryption, but it doesn't prevent **removing** `spec.secretEncryption` entirely once encryption has been configured. The third disjunct `(!has(object.spec.secretEncryption) && !has(oldObject.spec.secretEncryption))` only covers the case where neither old nor new has the field — the transition from "has encryption" → "no encryption" is allowed when `EtcdDataEncryptionUpToDate` is not `False` (i.e., steady state).

If a user removes `spec.secretEncryption`, `handleEncryptionNotConfigured` in the HCCO just clears `targetKey` and removes the condition — no migration to decrypt data is triggered. The existing encrypted data in etcd would become unreadable if the encryption config is also removed from KAS.

Two options:
1. **Prevent removal**: Add a CEL XValidation rule or VAP condition that blocks removing `spec.secretEncryption` once `status.secretEncryption.activeKey` is populated.
2. **Support decryption**: Add a migration path that promotes the `identity` provider to write position, runs StorageVersionMigration to re-write all data as plaintext, and only then removes the encryption configuration.

control-plane-operator/hostedclusterconfigoperator/controllers/reencryption/setup.go
```diff
@@ -0,0 +1,111 @@
...
+	// Watch KAS Deployment in the CP cluster for convergence detection.
+	if err := c.Watch(source.Kind[crclient.Object](opts.CPCluster.GetCache(), &appsv1.Deployment{},
+		handler.EnqueueRequestsFromMapFunc(func(_ context.Context, obj crclient.Object) []reconcile.Request {
+			if obj.GetNamespace() == opts.Namespace && obj.GetName() == manifests.KASDeployment("").Name {
+				return hcpRequest
+			}
+			return nil
+		}))); err != nil {
+		return fmt.Errorf("failed to watch KAS Deployment: %w", err)
+	}
```
csrwng, line 87: The HCCO now watches and directly reads the KAS Deployment (here and in `reencryption.go:isKASConverged`) to determine convergence state. The KAS Deployment is owned and managed by the CPO — this introduces a cross-domain coupling where the HCCO needs to understand control plane component internals.

Consider having the CPO surface KAS convergence state in HCP status (e.g., a condition or a field in `status.secretEncryption`) and having the HCCO read that instead. This would:
- Keep the HCCO focused on guest-cluster operations and HCP status
- Avoid duplicating the convergence check (the CPO already checks this in `secretencryption.go:isKASConverged`)
- Ensure that if the convergence definition ever changes, only one controller needs updating

csrwng (commented):
Overall this is a well-designed feature with solid architecture. The two-stage key rotation lifecycle (ReadOnlyDeploy → WritePromote → Migrating → Completed) is sound, and deriving phase from observable state rather than stored history makes the HCCO controller self-healing. Test coverage is comprehensive across both the CPO and HCCO sides.

Two architectural items flagged inline for discussion.

hypershift-operator/main.go
muraee: Great point that I have been considering. This is an existing gap, that we don't have a decrypt mechanism today. If someone clears `spec.secretEncryption` the cluster would be unusable.

If we consider blocking this behavior, I would add this as CEL validation directly in the API rather in the VAP. Otherwise, we probably need to implement a new feature to support decryption. Both of those options are orthogonal to this effort imo and should be handled separately.

control-plane-operator/hostedclusterconfigoperator/controllers/reencryption/setup.go
muraee: > Keep the HCCO focused on guest-cluster operations and HCP status


Given this is just a read operation, I would argue that it's not really a cross-domain coupling. Introducing a new condition for this small check adds more friction imo considering the HCCO would still need read the HCP which is also owned by the CPO.

> Avoid duplicating the convergence check (the CPO already checks this in secretencryption.go:isKASConverged)

both those functions already reuse the same shared function `podspec.IsDeploymentReady`, they are just tiny wrappers around to pass the correct deployment name.

hypershift-operator/main.go
csrwng: If we consider decryption a separate feature, then we should prevent it in the API

csrwng:
/lgtm
/hold
It's fine to fix the API (to prevent clearing encryption) in a follow-up given that it's not a regression.
Remove the hold if doing it in a follow up.

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -2073,10 +2073,147 @@ type AESCBCSpec struct {
...
+type SecretEncryptionKeyStatus struct {
+	// provider identifies the encryption provider.
+	// +required
+	// +unionDiscriminator
+	// +kubebuilder:validation:Enum=Azure;AWS;IBMCloud;AESCBC
+	Provider SecretEncryptionProvider `json:"provider,omitempty"`
+	// azure holds the Azure KMS key identity fields.
+	// +optional
+	// +unionMember
+	Azure AzureKMSKey `json:"azure,omitzero"`
```
enxebre, line 2135: why does azure struct doesn't follow the same naming pattern than aws/ibm?

muraee: good question. Though, I am just reusing the same struct from the spec.

muraee:
/unhold
Added a CEL rule to prevent removing `spec.SecretEncryption`

ahitacat:
/verified by ahitacat
Verified this in ARO-HCP personal dev environment.
Updated the HO with the latest build `on-pr-adb78ff817a6f952966372c1c02a2c2583bab9a7`, deployed a cluster with that build added as an annotation 
`hypershift.openshift.io/control-plane-operator-image: quay.io/redhat-user-workloads/crt-redhat-acm-tenant/control-plane-operator-main:on-pr-f7463be2c9312501857bbc0a03e8fdc8f6d04406`

Then I have rotate the key, and update the cluster. I could see the `EtcdDataEncryptionUpToDate`

```
    - lastTransitionTime: "2026-06-18T08:46:51Z"
      message: All etcd data is encrypted with the current active key
      observedGeneration: 2
      reason: ReEncryptionCompleted
      status: "True"
      type: EtcdDataEncryptionUpToDate 
```
And the status.secretEncryption.activeKey is equal to the spected one:

```
status:
secretEncryption:
      activeKey:
        azure:
          keyName: etcd-data-kms-encryption-key
          keyVaultName: ahitacat-kv-f6fd38
          keyVersion: 07be9b60713849da9c5281f9929de468
        provider: Azure
      history:
      - completionTime: "2026-06-18T08:46:51Z"
        from:
          fingerprint: 9807dbc6d76fccecef2aa613b0fd720a9d50579fbcf11187fa21e631ea1d667b
          provider: Azure
        startedTime: "2026-06-18T08:33:34Z"
        state: Completed
        to:
          fingerprint: a53695a388eacec58235de28f2e0e0258e10380fa51e03b53e6c233f19da9f73
          provider: Azure
```
```
spec:
    secretEncryption:
      kms:
        azure:
          activeKey:
            keyName: etcd-data-kms-encryption-key
            keyVaultName: ahitacat-kv-f6fd38
            keyVersion: 07be9b60713849da9c5281f9929de468 

```
It seems ok enough to me to merge it.

PR #8228, opened 2026-04-14
RFE-8777 Hypershift operator should provide DeploymentConditions

cwilkers:

To enable integration with Siteconfig Operator, Hypershift clusters should present their deployment status in a way that is compatible with ClusterDeployments, namely hivev1 clusterdeploymentconditions

- Pulls in Hivev1 for the original definitions
- Adds ClusterDeploymentConditions field to status
- Adds functions to compute conditions
- Generates deepcopy methods and api manifests
- Updates hypershift vendor directory
- Added tests

## Checklist:
- [ x] Relevant issues have been referenced.
- [ x] This change includes docs. 
- [ x] This change includes unit tests.

Co-authored-by: Cursor AI <noreply@cursor.com>

## Summary by CodeRabbit

* **New Features**
  * HostedCluster status now exposes ClusterDeployment conditions (Provisioned, ProvisionFailed, ProvisionStopped, RequirementsMet) for clearer provisioning state and improved status accuracy during reconciliation and deletion.

* **Tests**
  * Added unit tests validating computation, preservation, and lookup of ClusterDeployment conditions across lifecycle scenarios.

* **Chores**
  * Updated linting config and module metadata to accommodate the new status fields and validations.


api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -2105,6 +2105,58 @@ func ToPayloadArch(arch string) PayloadArchType {
...
+	ClusterProvisionFailedType ClusterDeploymentConditionType = "ProvisionFailed"
+
+	// ClusterProvisionStoppedType indicates the cluster provision has been stopped.
+	// This matches Hive's hivev1.ProvisionStoppedCondition ("ProvisionStopped").
+	ClusterProvisionStoppedType ClusterDeploymentConditionType = "ProvisionStopped"
+
+	// ClusterRequirementsMetType indicates the cluster meets all requirements for provisioning.
+	// This matches Hive's hivev1.RequirementsMetCondition ("RequirementsMet").
+	ClusterRequirementsMetType ClusterDeploymentConditionType = "RequirementsMet"
+)
```
CrystalChun: Can we directly use Hive's conditions instead of re-defining it here?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -2105,6 +2105,58 @@ func ToPayloadArch(arch string) PayloadArchType {
...
+	// +optional
+	// +kubebuilder:validation:MinLength=0
+	// +kubebuilder:validation:MaxLength=1024
+	Reason string `json:"reason,omitempty"`
+	// message is a human-readable message indicating details about last transition.
+	// +optional
+	// +kubebuilder:validation:MinLength=0
+	// +kubebuilder:validation:MaxLength=32768
+	Message string `json:"message,omitempty"`
+}
```
CrystalChun: Does site config require all of these fields? Can it be trimmed down?
If not can we just extend kube's apimachinery meta Condition type instead of redefining it? https://pkg.go.dev/k8s.io/apimachinery/pkg/api/meta#pkg-overview

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -5286,3 +5306,233 @@ func computeAzurePLSCondition(azPLSList hyperv1.AzurePrivateLinkServiceList, con
...
+			if cvoCond != nil {
+				condition.Message = fmt.Sprintf("Cluster Version Operator is not available: %s", cvoCond.Message)
+			} else {
+				condition.Message = "Waiting for Cluster Version Operator to become available"
+			}
+		} else {
+			condition.Reason = "NotProvisioned"
+			condition.Message = "Cluster is not yet provisioned"
+		}
+	}
```
CrystalChun: If you change the clusterdeploymentcondition to extend the condition package, you can use the kube apimachinery meta package to update the conditions without having to define the functions to find the condition itself https://pkg.go.dev/k8s.io/apimachinery/pkg/api/meta#pkg-overview

api/hypershift/v1beta1/hostedcluster_types.go
cwilkers: We could go either way here. The four conditions involved seem relatively stable, so I don't believe there is a lot of risk that their definitions will drift, but that risk is never zero. Adding the dependency would also require adding it to the API which would affect downstream consumers of hypershift beyond siteconfig operator.

I suppose given the two scenarios, I would go with the hive conditions just because using the dependency would guarantee they will not become a source of additional unexpected maintenance within the repo.

api/hypershift/v1beta1/hostedcluster_types.go
cwilkers: This seems to be resolved by using the Hive API.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
cwilkers: This is a little beyond my familiarity with kube apimachinery, but the Claude AI says it still needs a `findClusterDeploymentCondition` helper because there are type incompatibilities between the Hive conditions and Hypershift's conditionsso the meta package would not work.

CrystalChun: Ah gotcha, could we do something else to simplify this? 
Here's a suggestion of building a helper function from Claude that could be used for all of these compute functions

```
// buildClusterDeploymentCondition creates a ClusterDeploymentCondition with proper transition time handling.
// If the condition already exists with the same status, it preserves the LastTransitionTime.
// Otherwise, it sets LastTransitionTime to the current probeTime.
func buildClusterDeploymentCondition(
	existingConditions []hivev1.ClusterDeploymentCondition,
	conditionType hivev1.ClusterDeploymentConditionType,
	status corev1.ConditionStatus,
	reason string,
	message string,
	probeTime metav1.Time,
) hivev1.ClusterDeploymentCondition {
	condition := hivev1.ClusterDeploymentCondition{
		Type:               conditionType,
		Status:             status,
		Reason:             reason,
		Message:            message,
		LastProbeTime:      probeTime,
		LastTransitionTime: probeTime,
	}

	// Preserve LastTransitionTime if status hasn't changed
	existingCondition := findClusterDeploymentCondition(existingConditions, conditionType)
	if existingCondition != nil && existingCondition.Status == status {
		condition.LastTransitionTime = existingCondition.LastTransitionTime
	}

	return condition
}
```

and its usage

```suggestion
// computeProvisionedCondition determines if the cluster is provisioned
func computeProvisionedCondition(hcluster *hyperv1.HostedCluster, probeTime metav1.Time) hivev1.ClusterDeploymentCondition {
	// Provisioned is True when the cluster has been successfully provisioned:
	// control plane endpoint and kubeconfig exist, and the Cluster Version Operator
	// reports available (ClusterVersionAvailable on HostedCluster mirrors CVO on HCP).
	hasInfra := hcluster.Status.ControlPlaneEndpoint.Host != "" && hcluster.Status.KubeConfig != nil
	cvoCond := meta.FindStatusCondition(hcluster.Status.Conditions, string(hyperv1.ClusterVersionAvailable))
	cvoAvailable := cvoCond != nil && cvoCond.Status == metav1.ConditionTrue
	isProvisioned := hasInfra && cvoAvailable

	if isProvisioned {
		return buildClusterDeploymentCondition(
			hcluster.Status.ClusterDeploymentConditions,
			hivev1.ProvisionedCondition,
			corev1.ConditionTrue,
			hivev1.ProvisionedReason,
			hivev1.ProvisionedMessage,
			probeTime,
		)
	}

	// Not provisioned - determine reason
	var reason, message string
	if hasInfra && !cvoAvailable {
		reason = "WaitingForClusterVersionOperator"
		if cvoCond != nil {
			message = fmt.Sprintf("Cluster Version Operator is not available: %s", cvoCond.Message)
		} else {
			message = "Waiting for Cluster Version Operator to become available"
		}
	} else {
		reason = "NotProvisioned"
		message = "Cluster is not yet provisioned"
	}

	return buildClusterDeploymentCondition(
		hcluster.Status.ClusterDeploymentConditions,
		hivev1.ProvisionedCondition,
		corev1.ConditionFalse,
		reason,
		message,
		probeTime,
	)
}
```

cwilkers: I handed Claude the buildClusterDeploymentCondition and it refactored the individual condition functions. Thanks for the example!

CrystalChun:
Nice, this looks good to me!

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -7,6 +7,7 @@ import (
 	"github.com/openshift/hypershift/api/util/ipnet"
 
 	configv1 "github.com/openshift/api/config/v1"
+	hivev1 "github.com/openshift/hive/apis/hive/v1"
```
JoelSpeed, line 10: It is generally an anti-pattern to import another API. What difference about the ClusterDeployment condition when compared with a standard metav1.Condition?

Why can consumers not just observe the existing conditions field?

cwilkers: The underlying work is to allow the siteconfig operator (SO) to update its own CR (ClusterInstance)'s status from the clusters it deploys. In the case of siteconfig's original cluster types, there is a ClusterDeployment CR whose status is defined by Hive and represents the start and stop of a deployment in a way that the SO was written around.
When I added HCP to siteconfig as a supported cluster type, the original idea was to have SO read the deployment status from the HC CR, but it required extra logic given how many different status conditions a hosted cluster can report. The maintenance of hypershift specific logic within SO was rejected as an anti-pattern, and it was suggested to put that logic under the hypershift operator where it would be better maintained.
At the start of this PR, it was synthesizing regular metav1.Conditions to match the names of Hivev1 DeploymentConditions, but that was questioned as a maintenance hassle for hypershift as the Hive API could diverge. Thus the final form of this PR where we import Hivev1.

JoelSpeed: > The maintenance of hypershift specific logic within SO was rejected as an anti-pattern

Got any links to where this was discussed?

enxebre:
I would expect this logic to live in the consumer side which is interested in these particular semantics or otherwise propose a generic enhancement to capture why these semantics are valuable and how they could impact any existing consumer.
Extending the core API to vendor and fit any consumer specifics doesn't scale well.

api/hypershift/v1beta1/hostedcluster_types.go
cwilkers: I have, https://github.com/stolostron/siteconfig/pull/353#pullrequestreview-3490490385 and in https://redhat.atlassian.net/browse/CNF-18296

cwilkers:
>  or otherwise propose a generic enhancement to capture why these semantics are valuable and how they could impact any existing consumer. 

To a degree, that's the idea behind using the Hivev1 deploymentConditions. it is a common interface between agent-based and image-based cluster installs and any component (like SiteConfig in this instance) that is interested in the overall cluster's deployment status.

cwilkers:
Closing as the approach has been rejected.

PR #8246, opened 2026-04-15
OCPBUGS-81686: fix(authentication): use v2 auth validation for CEL and expression support.

ShazaAldawamneh:
Fixes OCPBUGS-81686
The validation logic was using v1/kas which doesn't support CEL claim
validation rules or expression-based username mappings. This was causing
errors when users configured ExternalOIDC with these features.
Changes:
- Export GenerateAuthConfig in v2/kas/auth.go to match v1 signature
- Update support/validations/authentication.go to import v2/kas
- Add test case for CEL validation rules and username expressions

Fixes OCPBUGS-81686

## Checklist:
- [Y] Subject and description added to both, commit and PR.
- [Y] Relevant issues have been referenced.
- [Y] This change includes unit tests.

## Summary by CodeRabbit

* **Refactor**
  * Authentication config generation now builds from the cluster Authentication spec with an explicit namespace; validation flows updated to use the revised generation path for consistency.

* **Tests**
  * OIDC validation tests expanded with feature-gated cases covering CEL-based claim validation and CEL-driven username mapping, including success and syntax-error failure scenarios.

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth.go
@@ -8,7 +8,6 @@ import (
 	"net/url"
 	"strings"
 
-	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
 	"github.com/openshift/hypershift/control-plane-operator/featuregates"
 	component "github.com/openshift/hypershift/support/controlplane-component"
 	"github.com/openshift/hypershift/support/supportedversion"
@@ -37,7 +36,7 @@ func adaptAuthConfig(cpContext component.WorkloadContext, config *corev1.ConfigM
 		return nil
 	}
 
-	authConfig, err := generateAuthConfig(cpContext, cpContext.Client, cpContext.HCP)
+	authConfig, err := GenerateAuthConfig(cpContext, cpContext.HCP.Spec.Configuration.Authentication, cpContext.Client, cpContext.HCP.Namespace)
 	if err != nil {
 		return fmt.Errorf("failed to generate authentication config: %w", err)
 	}
@@ -59,7 +58,11 @@ func adaptAuthConfig(cpContext component.WorkloadContext, config *corev1.ConfigM
 	return nil
 }
 
-func generateAuthConfig(ctx context.Context, c crclient.Reader, hcp *hyperv1.HostedControlPlane) (*AuthenticationConfiguration, error) {
+func GenerateAuthConfig(ctx context.Context, spec *configv1.AuthenticationSpec, c crclient.Reader, namespace string) (*AuthenticationConfiguration, error) {
+	if spec == nil {
+		return nil, fmt.Errorf("authentication spec cannot be nil")
+	}
+
 	config := &AuthenticationConfiguration{
 		TypeMeta: metav1.TypeMeta{
 			Kind:       "AuthenticationConfiguration",
@@ -68,8 +71,8 @@ func generateAuthConfig(ctx context.Context, c crclient.Reader, hcp *hyperv1.Hos
 		JWT: []JWTAuthenticator{},
 	}
 
-	for _, provider := range hcp.Spec.Configuration.Authentication.OIDCProviders {
-		jwt, err := generateJWTForProvider(ctx, provider, c, hcp.Namespace)
+	for _, provider := range spec.OIDCProviders {
+		jwt, err := generateJWTForProvider(ctx, provider, c, namespace)
 		if err != nil {
 			return nil, fmt.Errorf("generating JWT authenticator for provider %q: %v", provider.Name, err)
 		}
--- a/control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth_test.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth_test.go
@@ -1,6 +1,7 @@
 package kas
 
 import (
+	"context"
 	"encoding/json"
 	"testing"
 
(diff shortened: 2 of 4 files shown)


support/validations/authentication.go
everettraven, line 1: Should we update the corresponding test file here as well to ensure that we are appropriately testing variations of known valid and invalid inputs?

control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth_test.go
```diff
@@ -1521,6 +1521,90 @@ func TestAdaptAuthConfig(t *testing.T) {
 			},
 			shouldError: true,
 		},
+		{
+			name:   "CEL claim validation rules with username expression should validate successfully",
```
everettraven: It looks like existing test cases already cover this case?

https://github.com/openshift/hypershift/blob/f059233ec8170a8320031fda397549f3f35bcbe4/control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth_test.go#L1222-L1286

support/validations/authentication.go
ShazaAldawamneh: make sense, I forgot about that.

control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth_test.go
ShazaAldawamneh: yeah that's correct, I didn't realise that

everettraven (approved):
/lgtm

everettraven (commented):
/lgtm

control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth.go
```diff
@@ -59,7 +58,7 @@ func adaptAuthConfig(cpContext component.WorkloadContext, config *corev1.ConfigM
 	return nil
 }
 
-func generateAuthConfig(ctx context.Context, c crclient.Reader, hcp *hyperv1.HostedControlPlane) (*AuthenticationConfiguration, error) {
+func GenerateAuthConfig(spec *configv1.AuthenticationSpec, ctx context.Context, c crclient.Reader, namespace string) (*AuthenticationConfiguration, error) {
```
jparrill: Fix context.Context parameter ordering in GenerateAuthConfig — must be first parameter per Go convention

control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth.go
```diff
@@ -68,8 +67,8 @@ func generateAuthConfig(ctx context.Context, c crclient.Reader, hcp *hyperv1.Hos
 		JWT: []JWTAuthenticator{},
 	}
 
-	for _, provider := range hcp.Spec.Configuration.Authentication.OIDCProviders {
-		jwt, err := generateJWTForProvider(ctx, provider, c, hcp.Namespace)
+	for _, provider := range spec.OIDCProviders {
+		jwt, err := generateJWTForProvider(ctx, provider, c, namespace)
```
jparrill: Add nil guard for spec parameter — v1 has it, v2 doesn't, exported function will panic

control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth.go
```diff
@@ -37,7 +36,7 @@ func adaptAuthConfig(cpContext component.WorkloadContext, config *corev1.ConfigM
 		return nil
 	}
 
-	authConfig, err := generateAuthConfig(cpContext, cpContext.Client, cpContext.HCP)
+	authConfig, err := GenerateAuthConfig(cpContext.HCP.Spec.Configuration.Authentication, cpContext, cpContext.Client, cpContext.HCP.Namespace)
```
jparrill: Add direct unit test with new exported signature

support/validations/authentication_test.go
```diff
@@ -105,10 +111,86 @@ func TestValidateAuthenticationSpec(t *testing.T) {
 			},
 			shouldError: true,
 		},
+		{
+			name: "valid OIDC with CEL validation and username expression",
```
jparrill: Use Gherkin naming convention for tests matrix “When…it should..."

jparrill (commented):

control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth_test.go
```diff
@@ -60,6 +61,184 @@ MlubpsoEK2bYQDZskgDGCHI=
 -----END CERTIFICATE-----
 `
 
+func TestGenerateAuthConfig(t *testing.T) {
```
everettraven, line 64: Do we need to add explicit testing for `GenerateAuthConfig`? It looks like the testing we would care about here would already be covered by `TestAdaptAuthConfig` below?

ShazaAldawamneh: @jparrill requested a direct unit test for the exported function in their review [comment](https://github.com/openshift/hypershift/pull/8246#discussion_r3167027113). 

I can see both perspectives:
- The functionality is already covered through TestAdaptAuthConfig
- However, since GenerateAuthConfig is now exported, testing it directly provides clearer API contract validation and easier debugging when the function fails

@jparrill @everettraven - Would you like me to keep the direct test or remove it in favor of the existing coverage through TestAdaptAuthConfig?

everettraven: I'm not that strongly opinionated here. If @jparrill feels strongly we need an explicit test here, fine by me.

PR #8254, opened 2026-04-15
OCPBUGS-83328: fix skipProxyForKAS to use standard NO_PROXY matching

csrwng:

Fixes `skipProxyForKAS` in the HAProxy ignition config generation to use standard NO_PROXY matching semantics instead of naive `strings.Contains` substring matching.

The previous implementation used `strings.Contains(noProxy, s)` to check if KAS addresses should bypass the proxy. This failed for standard NO_PROXY patterns such as:
- **Leading-dot domain suffixes**: `.example.com` should match `api.test.example.com`, but `strings.Contains` doesn't implement domain suffix matching
- **Bare parent domain matching**: `example.com` should match `sub.example.com`
- **CIDR containment**: `172.16.0.0/12` in noProxy should match IP `172.20.0.1`

As a result, nodes would incorrectly tunnel KAS traffic through the HTTP proxy via `kubernetes-default-proxy` (HTTP CONNECT) instead of using the direct HAProxy TCP path, causing TLS handshake timeouts and cluster operator failures.

The fix uses the already-vendored `golang.org/x/net/http/httpproxy` package which implements the full NO_PROXY specification, while preserving backward-compatible heuristics for the `"kubernetes"` keyword and explicit service/cluster network CIDR entries.

Fixes https://issues.redhat.com/browse/OCPBUGS-83328

- The fix is strictly additive: `skipProxyForKAS` returns `true` in more cases than before (all previous exact-match cases still work), so existing clusters are unaffected
- The `golang.org/x/net/http/httpproxy` package is already vendored; no new dependencies are introduced
- Backward-compatible heuristics preserved: "kubernetes" keyword match and explicit serviceNetwork/clusterNetwork CIDR entries still trigger proxy skip, even though the internal API address (172.20.0.1) is in a different IP range than these networks

## Checklist:

🤖 Generated with [Claude Code](https://claude.com/claude-code) via `/jira:solve OCPBUGS-83328`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved NO_PROXY handling for Kubernetes API Server traffic to honor standard proxy-bypass semantics (domain suffixes, CIDRs, exact IPs, ports, wildcards) while keeping legacy heuristics for compatibility.
* **Tests**
  * Added comprehensive tests covering many NO_PROXY formats and port-matching scenarios.
* **Chores / Test Fixtures**
  * Added an ignition test fixture provisioning scripts, HAProxy config, and a systemd unit to validate runtime behavior.

### DIFF
--- a/hypershift-operator/controllers/nodepool/apiserver-haproxy/haproxy.go
+++ b/hypershift-operator/controllers/nodepool/apiserver-haproxy/haproxy.go
@@ -8,7 +8,6 @@ import (
 	"html/template"
 	"net"
 	"net/url"
-	"slices"
 	"strconv"
 	"strings"
 
@@ -37,6 +36,7 @@ import (
 	"github.com/clarketm/json"
 	ignitionapi "github.com/coreos/ignition/v2/config/v3_2/types"
 	"github.com/vincent-petithory/dataurl"
+	"golang.org/x/net/http/httpproxy"
 )
 
 const (
@@ -275,10 +275,11 @@ func apiServerProxyConfig(haProxyImage, cpoImage, clusterID,
 		},
 	}
 
-	// Check if no proxy contains any address that should result in skipping the system proxy
-	skipProxyForKAS := slices.ContainsFunc([]string{externalAPIAddress, internalAPIAddress, "kubernetes", serviceNetwork, clusterNetwork}, func(s string) bool {
-		return strings.Contains(noProxy, s)
-	})
+	// Check if no proxy contains any address that should result in skipping the system proxy.
+	// This uses standard NO_PROXY matching semantics (domain suffix, CIDR containment)
+	// instead of naive substring matching, so that entries like ".example.com" correctly
+	// match "api.test.example.com".
+	skipProxyForKAS := shouldSkipProxyForKAS(noProxy, externalAPIAddress, externalAPIPort, internalAPIAddress, internalAPIPort, serviceNetwork, clusterNetwork)
 
 	if proxyAddr == "" || skipProxyForKAS {
 		filesToAdd = append(filesToAdd, []fileToAdd{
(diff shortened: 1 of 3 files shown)


hypershift-operator/controllers/nodepool/apiserver-haproxy/haproxy.go
```diff
@@ -275,10 +275,11 @@ func apiServerProxyConfig(haProxyImage, cpoImage, clusterID,
...
 
-	// Check if no proxy contains any address that should result in skipping the system proxy
-	skipProxyForKAS := slices.ContainsFunc([]string{externalAPIAddress, internalAPIAddress, "kubernetes", serviceNetwork, clusterNetwork}, func(s string) bool {
-		return strings.Contains(noProxy, s)
-	})
+	// Check if no proxy contains any address that should result in skipping the system proxy.
+	// This uses standard NO_PROXY matching semantics (domain suffix, CIDR containment)
+	// instead of naive substring matching, so that entries like ".example.com" correctly
+	// match "api.test.example.com".
+	skipProxyForKAS := shouldSkipProxyForKAS(noProxy, externalAPIAddress, internalAPIAddress, serviceNetwork, clusterNetwork)
```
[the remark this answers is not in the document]
csrwng: Done. Threaded `externalAPIPort` and `internalAPIPort` into `shouldSkipProxyForKAS` and now construct test URLs with `net.JoinHostPort` so port-qualified `NO_PROXY` entries like `api.example.com:6443` are matched correctly. Added 3 regression tests: port-qualified domain match, port-qualified IP match, and non-matching port.

---

hypershift-operator/controllers/nodepool/apiserver-haproxy/haproxy.go
```diff
@@ -336,6 +337,76 @@ func apiServerProxyConfig(haProxyImage, cpoImage, clusterID,
...
+		port int32
+	}{
+		{host: externalAPIAddress, port: externalAPIPort},
+		{host: internalAPIAddress, port: internalAPIPort},
+	} {
+		if target.host == "" {
+			continue
+		}
+		host := target.host
+		if target.port != 0 {
```
sdminonne, line 375: Guess `target.port` can never be 0 but it doesn't hurt :)

hypershift-operator/controllers/nodepool/apiserver-haproxy/haproxy.go
```diff
@@ -336,6 +337,76 @@ func apiServerProxyConfig(haProxyImage, cpoImage, clusterID,
...
+			host = net.JoinHostPort(target.host, strconv.Itoa(int(target.port)))
+		}
+		testURL := &url.URL{Scheme: "https", Host: host}
+		if proxyURL, _ := proxyFunc(testURL); proxyURL == nil {
```
sdminonne, line 379: swallowing errors makes me nervous but this is the proper way here 👍

sdminonne (commented):
Thanks nothing to say :)

PR #8262, opened 2026-04-16
CNTRLPLANE-3040: add regression tests for CAS pause/unpause replica clamping

jparrill:
## Summary
- Add unit and E2E regression tests for OCPBUGS-78152 / [CNTRLPLANE-3040](https://redhat.atlassian.net/browse/CNTRLPLANE-3040)
- Prevents regression of the bug where CAS accumulated replica decrements on paused MachineDeployments, causing catastrophic scale-down on unpause
- Extracts shared test helpers (`getFirstNodePool`, `waitForAutoscalerDeploymentReady`) and constants to reduce duplication across autoscaling tests

### Unit tests (`capi_test.go`)
`TestPauseUnpauseCycle` — 9 table-driven test cases:
- MachineDeployment (Replace) and MachineSet (InPlace) upgrade types
- Replica clamping: unchanged, decremented to min, below min, above max
- Non-AWS platform `min=0` → `effectiveMin=1` enforcement (KubeVirt)
- Verifies: pause annotation set/removed, replicas clamped within [min,max], autoscaler annotations preserved

### E2E test (`autoscaling_test.go`)
`testAutoscalerRespectsNodePoolPause` — placed between `testAutoscaling` and `testAutoscalingBalancing`:
1. Scale up NodePool to max with workload
2. Configure aggressive CAS timers (unneeded=60s, delayAfterAdd=30s)
3. Pause NodePool → verify MachineDeployment pause annotation
4. Delete workload → wait 3min for CAS to attempt scale-down
5. Verify replicas unchanged while paused
6. Unpause → verify stability (no catastrophic scale-down)
7. Clean up to neutral state

### DRY improvements
- `getFirstNodePool()` — replaces 5-line list+assert pattern in 3 test functions
- `waitForAutoscalerDeploymentReady()` — replaces 15-line EventuallyObject pattern in 2 test functions
- Constants: `aggressiveDelayAfterAddSeconds`, `aggressiveUnneededDurationSeconds`, `casScaleDownWaitDuration`
- Uses `capiv1.PausedAnnotation` constant instead of string literal

## Test plan
- [x] `go vet` passes on both packages
- [x] `TestPauseUnpauseCycle` — 9/9 cases PASS
- [x] E2E compiles cleanly (`go vet -tags e2e ./test/e2e/`)
- [x] E2E verified on live cluster: replicas stayed at 3 during 3-minute pause, unpause was clean
- [ ] CI e2e-autoscaling job passes

Refs: OCPBUGS-78152, [CNTRLPLANE-3040](https://redhat.atlassian.net/browse/CNTRLPLANE-3040)

## Summary by CodeRabbit

* **Tests**
  * Added a regression test for the NodePool pause → replica drift → unpause cycle, verifying pause propagation, autoscaler annotation preservation, and replica clamping on reconcile.
  * Added an end-to-end test confirming the cluster-autoscaler detects paused node groups, prevents scale-down while paused, and resumes normal scaling after unpause.
* **Refactor**
  * Consolidated autoscaling test helpers and shared timing constants to reduce duplication and stabilize timing-sensitive assertions.

### DIFF
--- a/hypershift-operator/controllers/nodepool/capi_test.go
+++ b/hypershift-operator/controllers/nodepool/capi_test.go
@@ -2943,6 +2943,291 @@ func TestPropagateLabelsAndTaintsToMachines(t *testing.T) {
 	}
 }
 
+// TestPauseUnpauseCycle is a regression test for the interaction between pausing
+// NodePools and the Cluster Autoscaler (OCPBUGS-78152 / CNTRLPLANE-3040).
+//
+// It verifies that:
+//   - When a MachineDeployment/MachineSet is paused, the pause annotation is set
+//   - When reconcileMachineDeployment/reconcileMachineSet runs after unpause, the
+//     pause annotation is removed and autoscaler annotations are preserved
+//   - setMachineDeploymentReplicas clamps replicas within [min, max] bounds regardless
+//     of external modifications (e.g. CAS decrementing replicas while paused)
+func TestPauseUnpauseCycle(t *testing.T) {
+	t.Parallel()
+	testCases := []struct {
+		name              string
+		upgradeType       hyperv1.UpgradeType
+		platformType      hyperv1.PlatformType
+		initialReplicas   int32
+		autoscalingMin    int32
+		autoscalingMax    int32
+		replicasAtUnpause int32
+		expectedReplicas  int32
+	}{
+		{
+			name:              "When a paused MachineDeployment is unpaused it should remove the pause annotation and preserve replicas",
+			upgradeType:       hyperv1.UpgradeTypeReplace,
+			platformType:      hyperv1.AWSPlatform,
+			initialReplicas:   4,
+			autoscalingMin:    1,
+			autoscalingMax:    4,
+			replicasAtUnpause: 4,
+			expectedReplicas:  4,
+		},
+		{
+			name:              "When a paused MachineDeployment has replicas decremented to min it should keep replicas at min on unpause",
+			upgradeType:       hyperv1.UpgradeTypeReplace,
+			platformType:      hyperv1.AWSPlatform,
+			initialReplicas:   4,
+			autoscalingMin:    1,
+			autoscalingMax:    4,
+			replicasAtUnpause: 1,
+			expectedReplicas:  1,
+		},
+		{
+			name:              "When a paused MachineDeployment has replicas decremented below min it should clamp to min on unpause",
+			upgradeType:       hyperv1.UpgradeTypeReplace,
+			platformType:      hyperv1.AWSPlatform,
+			initialReplicas:   4,
+			autoscalingMin:    2,
+			autoscalingMax:    4,
+			replicasAtUnpause: 1,
+			expectedReplicas:  2,
+		},
+		{
+			name:              "When a paused MachineDeployment has replicas incremented above max it should clamp to max on unpause",
+			upgradeType:       hyperv1.UpgradeTypeReplace,
+			platformType:      hyperv1.AWSPlatform,
+			initialReplicas:   4,
+			autoscalingMin:    1,
+			autoscalingMax:    4,
+			replicasAtUnpause: 6,
+			expectedReplicas:  4,
+		},
+		{
+			name:              "When a paused MachineDeployment on AWS with min 0 it should allow scale to zero on unpause",
+			upgradeType:       hyperv1.UpgradeTypeReplace,
+			platformType:      hyperv1.AWSPlatform,
+			initialReplicas:   4,
+			autoscalingMin:    0,
+			autoscalingMax:    4,
+			replicasAtUnpause: 0,
+			expectedReplicas:  0,
+		},
+		{
+			name:              "When a paused MachineDeployment on non-AWS with min 0 it should clamp to effective min 1 on unpause",
+			upgradeType:       hyperv1.UpgradeTypeReplace,
+			platformType:      hyperv1.KubevirtPlatform,
+			initialReplicas:   4,
+			autoscalingMin:    0,
+			autoscalingMax:    4,
+			replicasAtUnpause: 0,
+			expectedReplicas:  1,
+		},
+		{
+			name:              "When a paused MachineSet is unpaused it should remove the pause annotation and preserve replicas",
+			upgradeType:       hyperv1.UpgradeTypeInPlace,
+			platformType:      hyperv1.AWSPlatform,
+			initialReplicas:   4,
+			autoscalingMin:    1,
+			autoscalingMax:    4,
+			replicasAtUnpause: 4,
+			expectedReplicas:  4,
+		},
+		{
+			name:              "When a paused MachineSet has replicas decremented below min it should clamp to min on unpause",
+			upgradeType:       hyperv1.UpgradeTypeInPlace,
+			platformType:      hyperv1.AWSPlatform,
+			initialReplicas:   4,
+			autoscalingMin:    2,
+			autoscalingMax:    4,
+			replicasAtUnpause: 1,
+			expectedReplicas:  2,
+		},
+		{
+			name:              "When a paused MachineSet has replicas incremented above max it should clamp to max on unpause",
+			upgradeType:       hyperv1.UpgradeTypeInPlace,
+			platformType:      hyperv1.AWSPlatform,
+			initialReplicas:   4,
+			autoscalingMin:    1,
+			autoscalingMax:    4,
+			replicasAtUnpause: 6,
+			expectedReplicas:  4,
+		},
+		{
+			name:              "When a paused MachineSet on AWS with min 0 it should allow scale to zero on unpause",
+			upgradeType:       hyperv1.UpgradeTypeInPlace,
+			platformType:      hyperv1.AWSPlatform,
+			initialReplicas:   4,
+			autoscalingMin:    0,
+			autoscalingMax:    4,
+			replicasAtUnpause: 0,
+			expectedReplicas:  0,
+		},
+		{
+			name:              "When a paused MachineSet on non-AWS with min 0 it should clamp to effective min 1 on unpause",
+			upgradeType:       hyperv1.UpgradeTypeInPlace,
+			platformType:      hyperv1.KubevirtPlatform,
+			initialReplicas:   4,
+			autoscalingMin:    0,
+			autoscalingMax:    4,
+			replicasAtUnpause: 0,
+			expectedReplicas:  1,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			g := NewWithT(t)
+
+			nodePool := &hyperv1.NodePool{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "test-nodepool",
+					Namespace: "test-namespace",
+				},
+				Spec: hyperv1.NodePoolSpec{
+					ClusterName: "test-cluster",
+					AutoScaling: &hyperv1.NodePoolAutoScaling{
+						Min: ptr.To[int32](tc.autoscalingMin),
+						Max: tc.autoscalingMax,
+					},
+					Management: hyperv1.NodePoolManagement{
+						UpgradeType: tc.upgradeType,
+						Replace: &hyperv1.ReplaceUpgrade{
+							Strategy: hyperv1.UpgradeStrategyRollingUpdate,
+							RollingUpdate: &hyperv1.RollingUpdate{
+								MaxUnavailable: ptr.To(intstr.FromInt(0)),
+								MaxSurge:       ptr.To(intstr.FromInt(1)),
+							},
+						},
+					},
+					Platform: hyperv1.NodePoolPlatform{
+						Type: tc.platformType,
+					},
+				},
+			}
+			if tc.platformType == hyperv1.AWSPlatform {
+				nodePool.Spec.Platform.AWS = &hyperv1.AWSNodePoolPlatform{AMI: "test-ami"}
+			}
+
+			hostedCluster := &hyperv1.HostedCluster{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "test-cluster",
+					Namespace: "test-namespace",
+				},
+				Spec: hyperv1.HostedClusterSpec{
+					Platform: hyperv1.PlatformSpec{
+						Type: tc.platformType,
+					},
+				},
+			}
+			if tc.platformType == hyperv1.AWSPlatform {
+				hostedCluster.Spec.Platform.AWS = &hyperv1.AWSPlatformSpec{}
+			}
+
+			controlPlaneNamespace := "test-cp-namespace"
+			c := fake.NewClientBuilder().WithScheme(api.Scheme).Build()
+			capiObj := &CAPI{
+				Token: &Token{
+					ConfigGenerator: &ConfigGenerator{
+						nodePool:              nodePool,
+						hostedCluster:         hostedCluster,
+						controlplaneNamespace: controlPlaneNamespace,
+						Client:                c,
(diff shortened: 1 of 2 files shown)


test/e2e/autoscaling_test.go
```diff
@@ -387,6 +360,239 @@ func newWorkLoad(njobs int32, memoryRequest resource.Quantity, nodeSelector, ima
...
+		defer func() {
+			cascadeDelete := metav1.DeletePropagationForeground
+			_ = guestClient.Delete(ctx, workload, &crclient.DeleteOptions{
+				PropagationPolicy: &cascadeDelete,
+			})
+		}()
+
+		// Wait for max nodes.
+		_ = e2eutil.WaitForNReadyNodes(t, ctx, guestClient, max, hostedCluster.Spec.Platform.Type)
+		t.Logf("Scaled up to %d nodes", max)
```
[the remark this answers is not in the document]
jparrill: Fixed. The test now explicitly enables autoscaling on the NodePool via `retry.RetryOnConflict` before proceeding, making it fully self-contained and independent of prior subtests.

test/e2e/autoscaling_test.go
```diff
@@ -387,6 +360,239 @@ func newWorkLoad(njobs int32, memoryRequest resource.Quantity, nodeSelector, ima
...
+		g.Expect(replicasAfterUnpause).To(Equal(replicasBeforePause),
+			"MachineDeployment replicas should remain stable immediately after unpause")
+
+		// Verify nodes are still ready.
+		_ = e2eutil.WaitForNReadyNodes(t, ctx, guestClient, max, hostedCluster.Spec.Platform.Type)
+		t.Logf("All %d nodes still ready after unpause — no catastrophic scale-down detected", max)
+
+		// Step 9: Clean up — wait for scale-down to numNodes so next test starts clean.
+		_ = e2eutil.WaitForNReadyNodes(t, ctx, guestClient, numNodes, hostedCluster.Spec.Platform.Type)
+		t.Logf("Scaled down to %d nodes — clean state for next test", numNodes)
```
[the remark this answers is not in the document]
jparrill: Fixed. Removed the strict `replicasAfterUnpause == replicasBeforePause` assertion and `WaitForNReadyNodes(..., max, ...)` after unpause. The test now waits for clean convergence back to `numNodes`, since CAS may legitimately start normal scale-down immediately after the pause annotation is removed.

test/e2e/autoscaling_test.go
```diff
@@ -387,6 +360,247 @@ func newWorkLoad(njobs int32, memoryRequest resource.Quantity, nodeSelector, ima
...
+			PropagationPolicy: &cascadeDelete,
+		})
+		g.Expect(err).NotTo(HaveOccurred())
+		t.Logf("Deleted workload")
+
+		// Step 5: Wait for CAS unneeded duration + buffer to give CAS time to attempt scale-down.
+		// With unneeded=60s and delayAfterAdd=30s, we wait 3 minutes total.
+		// If the CAS fix is working, replicas should NOT change.
+		t.Logf("Waiting %s for CAS to attempt scale-down on paused MachineDeployment...", casScaleDownWaitDuration)
+		time.Sleep(casScaleDownWaitDuration)
```
devguyio: **Replace `time.Sleep` with CAS log polling**

The CAS fix ([openshift/kubernetes-autoscaler#409](https://github.com/openshift/kubernetes-autoscaler/pull/409)) logs at `klog.V(4)` when it skips a paused node group:
```
klog.V(4).Infof("discovered a paused node group: %s", ng.Debug())
```
The CAS deployment already runs with `--v=4` (`v2/assets/cluster-autoscaler/deployment.yaml:36`), so this line is emitted in production. There's existing precedent for reading pod logs in e2e tests (`test/e2e/util/util.go:825`).

Instead of sleeping 3 minutes and hoping CAS ran during that window, the test could poll the CAS pod logs for `"discovered a paused node group"` after pausing. This would be:
- **Faster** — seconds instead of 3 minutes
- **Deterministic** — confirms CAS actually saw the paused group and skipped it, rather than relying on absence of action during an arbitrary window
- **Stronger** — the current approach passes vacuously if CAS is slow to evaluate (API latency, slow node status propagation)

devguyio (commented):
Solid regression test coverage overall. The 9 table-driven unit cases systematically hit every branch of `setMachineDeploymentReplicas` and both MachineDeployment/MachineSet code paths. The DRY refactors are clean.

**⚠️ Consider merging `testAutoscaling` and `testAutoscalerRespectsNodePoolPause`**

As-is, adding `testAutoscalerRespectsNodePoolPause` as a separate subtest has significant cost:

- **~10 extra minutes of CI wall time** — a full scale-up/down cycle on real AWS infrastructure, duplicating what `testAutoscaling` already does
- **~10 extra minutes of EC2 spend per run** — provisioning and draining `max` nodes a second time
- **Doubles the flake surface** — two independent scale-up/down cycles means two chances for node provisioning timeouts, API throttling, or slow kubelet registration to cause spurious failures
- **Multiplied across every PR** — this runs on every PR that triggers the autoscaling e2e job

The two subtests follow very similar flows:

`testAutoscaling` (existing):
1. Enable autoscaling (min=numNodes, max=numNodes+2)
2. Create workload → scale up to max
3. Delete workload → verify scale-down to numNodes

`testAutoscalerRespectsNodePoolPause` (new):
1. Enable autoscaling (same min/max)
2. Configure aggressive CAS timers
3. Create workload → scale up to max
4. Pause → delete workload → verify NO scale-down
5. Unpause → verify scale-down to numNodes

Steps 1-3 and the final scale-down are essentially duplicated. A merged flow could be:

1. Enable autoscaling + configure aggressive CAS timers
2. Create workload → scale up to max *(proves scale-up works)*
3. Pause NodePool → delete workload → verify replicas unchanged *(proves pause blocks CAS)*
4. Unpause → verify scale-down to numNodes *(proves scale-down works after unpause)*

One scale-up/down cycle instead of two. Same coverage, tells a stronger story: "CAS scales up, respects pause, and resumes scale-down cleanly" as one coherent lifecycle.

The tradeoff is fault isolation — if the pause logic breaks, a merged test would also mask whether basic scale-up/down still works. That said, the unit tests already cover the controller logic thoroughly, so the e2e is really validating the CAS integration end-to-end where the merged flow makes more sense.

---
*Review generated with [Claude Code](https://claude.com/claude-code)*

test/e2e/autoscaling_test.go
```diff
@@ -387,6 +360,247 @@ func newWorkLoad(njobs int32, memoryRequest resource.Quantity, nodeSelector, ima
...
 }
 
+const (
+	// aggressiveDelayAfterAddSeconds is the CAS delay-after-add setting used in pause regression tests.
+	aggressiveDelayAfterAddSeconds int32 = 30
+	// aggressiveUnneededDurationSeconds is the CAS unneeded-duration setting used in pause regression tests.
+	aggressiveUnneededDurationSeconds int32 = 60
+	// casScaleDownWaitDuration is the time to wait for CAS to attempt scale-down during pause tests.
+	// Calculated as unneeded + delay-after-add + buffer.
+	casScaleDownWaitDuration = 3 * time.Minute
```
devguyio: **Nit:** The comment says "unneeded + delay-after-add + buffer" but the value is hardcoded. If the sleep stays, consider computing it from the other two constants to keep the relationship explicit:
```go
casScaleDownWaitDuration = time.Duration(aggressiveUnneededDurationSeconds+aggressiveDelayAfterAddSeconds)*time.Second + 90*time.Second
```

test/e2e/autoscaling_test.go
enxebre: is there any event cas is triggering that this can watch so we don't perpetuate watching logs either?

jparrill: Nope, looks like CAS does not emit Kubernetes Events for paused node groups. No RBAC for events, no EventRecorder in the CAS component.

test/e2e/autoscaling_test.go
jparrill: Fixed

hypershift-operator/controllers/nodepool/capi_test.go
bryan-cox, line 1: t.parallel() should be added

hypershift-operator/controllers/nodepool/capi_test.go
```diff
@@ -2331,6 +2331,270 @@ func TestPause(t *testing.T) {
...
+		{
+			name:              "When a paused MachineSet on non-AWS with min 0 it should clamp to effective min 1 on unpause",
+			upgradeType:       hyperv1.UpgradeTypeInPlace,
+			platformType:      hyperv1.KubevirtPlatform,
+			initialReplicas:   4,
+			autoscalingMin:    0,
+			autoscalingMax:    4,
+			replicasAtUnpause: 0,
+			expectedReplicas:  1,
+		},
```
bryan-cox, line 3076: Missing AWS min=0 unit test case

docs/mkdocs.yml
bryan-cox, line 1: These changes needed here?

test/e2e/autoscaling_test.go
```diff
@@ -387,6 +363,311 @@ func newWorkLoad(njobs int32, memoryRequest resource.Quantity, nodeSelector, ima
...
+const (
+	aggressiveDelayAfterAddSeconds    int32 = 30
+	aggressiveUnneededDurationSeconds int32 = 60
+	casScaleDownWaitDuration                = time.Duration(aggressiveUnneededDurationSeconds+aggressiveDelayAfterAddSeconds)*time.Second + 90*time.Second
+	casLogPollInterval                      = 5 * time.Second
+	casPausedLogMessage                     = "discovered a paused node group"
+)
+
+// getFirstNodePool lists NodePools in the given namespace and asserts exactly one exists.
+func getFirstNodePool(t *testing.T, ctx context.Context, mgtClient crclient.Client, namespace string) *hyperv1.NodePool {
```
bryan-cox: This kind of makes it sound like there is more than one nodepool. Maybe this could be getOnlyNodePool?

test/e2e/autoscaling_test.go
```diff
@@ -387,6 +363,311 @@ func newWorkLoad(njobs int32, memoryRequest resource.Quantity, nodeSelector, ima
...
+			if err := mgtClient.Get(ctx, crclient.ObjectKeyFromObject(nodepool), nodepool); err != nil {
+				return err
+			}
+			nodepool.Spec.PausedUntil = ptr.To("true")
+			return mgtClient.Update(ctx, nodepool)
+		})
+		g.Expect(err).NotTo(HaveOccurred(), "failed to pause NodePool")
+		t.Logf("Paused NodePool %s", nodepool.Name)
+		defer func() {
+			_ = retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
bryan-cox: error is silently thrown away

test/e2e/autoscaling_test.go
```diff
@@ -387,6 +363,311 @@ func newWorkLoad(njobs int32, memoryRequest resource.Quantity, nodeSelector, ima
...
+
+		// 50% - enough that the existing and new nodes will be used, not enough to have more than 1 pod per node.
+		workloadMemRequest := resource.MustParse(fmt.Sprintf("%v", 0.5*float32(bytes)))
+		workload := newWorkLoad(max, workloadMemRequest, "", globalOpts.LatestReleaseImage)
+		err = guestClient.Create(ctx, workload)
+		g.Expect(err).NotTo(HaveOccurred())
+		t.Logf("Created workload. Node: %s, memcapacity: %s", nodes[0].Name, memCapacity.String())
+		defer func() {
+			cascadeDelete := metav1.DeletePropagationForeground
+			_ = guestClient.Delete(ctx, workload, &crclient.DeleteOptions{
```
bryan-cox: error is silently thrown away

docs/mkdocs.yml
jparrill: Removed

hypershift-operator/controllers/nodepool/capi_test.go
jparrill: Fixed. Added `t.Parallel()`.

hypershift-operator/controllers/nodepool/capi_test.go
jparrill: Fixed. Added AWS min=0 test cases for both MachineDeployment and MachineSet — validates that AWS allows scale-to-zero (expectedReplicas=0).

test/e2e/autoscaling_test.go
jparrill: Fixed. Renamed to `getOnlyNodePool` across all call sites.

test/e2e/autoscaling_test.go
jparrill: Fixed. Error is now logged: `t.Logf("cleanup: failed to unpause NodePool %s: %v", ...)`.

test/e2e/autoscaling_test.go
jparrill: Fixed. Error is now logged: `t.Logf("cleanup: failed to delete workload: %v", ...)`.

jparrill:
Rebased

jparrill:
Rebased again

hypershift-operator/controllers/nodepool/capi_test.go
```diff
@@ -2943,6 +2943,291 @@ func TestPropagateLabelsAndTaintsToMachines(t *testing.T) {
...
 
+// TestPauseUnpauseCycle is a regression test for the interaction between pausing
+// NodePools and the Cluster Autoscaler (OCPBUGS-78152 / CNTRLPLANE-3040).
+//
+// It verifies that:
+//   - When a MachineDeployment/MachineSet is paused, the pause annotation is set
+//   - When reconcileMachineDeployment/reconcileMachineSet runs after unpause, the
+//     pause annotation is removed and autoscaler annotations are preserved
+//   - setMachineDeploymentReplicas clamps replicas within [min, max] bounds regardless
+//     of external modifications (e.g. CAS decrementing replicas while paused)
```
enxebre, line 2954: isn't all this really effectively the same than adding a few test cases for TestSetMachineDeploymentReplicas and TestSetMachineSetReplicas?

test/e2e/autoscaling_test.go
```diff
@@ -387,6 +363,316 @@ func newWorkLoad(njobs int32, memoryRequest resource.Quantity, nodeSelector, ima
 	return job
 }
 
+const (
+	aggressiveDelayAfterAddSeconds    int32 = 30
```
enxebre, line 367: is this setting aggressive config for existing tests?

hypershift-operator/controllers/nodepool/capi_test.go
jparrill: Fair point — the clamping logic itself is the same. The difference is that TestPauseUnpauseCycle exercises the full reconcile flow end-to-end: 

```
Pause annotation set on NodePool
        │
        ▼
External replica modification (CAS decrements while paused)
        │
        ▼
reconcileMachineDeployment / reconcileMachineSet
        │
        ▼
Pause annotation removed
        │
        ▼
Replicas clamped back to [min, max]
```

Adding cases to TestSetMachineDeploymentReplicas/TestSetMachineSetReplicas would cover the clamping in isolation, but wouldn't validate that all the pieces work together through the reconcile cycle. I think both are worth keeping — unit cases for the function, integration-style test for the flow.

test/e2e/autoscaling_test.go
jparrill: Right — from the point this test runs until the end of the autoscaling suite, the aggressive policy stays set on the HostedCluster. I could restore the original config after the test, but is it worth it? The e2e tests are passing with the aggressive timers and we're saving wall time on scale-down waits for the remaining subtests. What do you think?

PR #7405, opened 2025-12-17
OCPBUGS-62177: Fix certificate revocation validation across all KAS pods

sdminonne:
## What this PR does / why we need it:

The control-plane-pki-operator's `CertificateRevocationController` incorrectly reports certificate revocation conditions because it does not verify each individual kube-apiserver pod has reloaded its CA bundles. This means a certificate may still be usable for a short period after the PKI operator reports it revoked, depending on which KAS pod handles the request.

This PR introduces per-pod KAS verification during certificate revocation by:

1. **Enumerating all kube-apiserver pods** via label selector (`app=kube-apiserver`)
2. **Connecting directly to each pod's IP** to verify certificate trust/revocation status
3. **Only advancing the CRR condition** (`NewCertificatesTrusted` / `PreviousCertificatesRevoked`) once **all** KAS pods confirm the expected state

### Fixes applied on top of initial implementation

The initial per-pod verification had two bugs causing e2e test failures:

- **TLS ServerName mismatch**: Connecting directly to pod IPs caused TLS verification failures because pod IPs are not in the KAS serving certificate SANs. Fixed by explicitly setting `TLSClientConfig.ServerName = "kube-apiserver"` so Go's TLS library verifies against the service DNS name instead of the pod IP.

- **Terminating pods not filtered**: During KAS rolling updates, old pods with `DeletionTimestamp` set still report `Ready=True` during their graceful termination period. The controller would attempt to connect to these unresponsive pods and get stuck. Fixed by skipping pods where `DeletionTimestamp != nil`.

Based on @enxebre https://github.com/openshift/hypershift/pull/6926, addressing reviewer feedback.

## Which issue(s) this PR fixes:

Fixes https://issues.redhat.com/browse/OCPBUGS-62177

## Special notes for your reviewer:

- The `testCertificateAgainstAllKASPods` helper is used in both `ensureNewSignerCertificatePropagated` (checks new cert is trusted) and `ensureOldSignerCertificateRevoked` (checks old cert is rejected)
- Unit tests added for `isPodReady` and `testCertificateAgainstAllKASPods` pod filtering behavior
- The 5-second per-pod timeout prevents the revocation process from hanging if a pod is unresponsive

## Checklist:
- [x] Subject and description added to both, commit and PR.
- [x] Relevant issues have been referenced.
- [ ] This change includes docs.
- [x] This change includes unit tests.

🤖 Generated with [Claude Code](https://claude.com/claude-code) via `/jira:solve OCPBUGS-62177`

PR #8263, opened 2026-04-16
OCPBUGS-62177: verify cert revocation against all KAS pods

sdminonne:

Updates the `CertificateRevocationController` to verify certificate trust and revocation against **every individual KAS pod** rather than through the service load balancer. In HA deployments with 3 KAS replicas, the service endpoint load-balances to a single pod, so the check could hit a pod that had loaded the updated trust bundle while others hadn't, causing premature state transitions in the revocation flow.

### Changes

**control-plane-pki-operator** (`CertificateRevocationController`):
- Add `verifyCertificateAgainstAllKASPods` which enumerates ready KAS pods via label selector and connects to each pod IP directly (with SNI set to the KAS service name for TLS validation)
- Cross-check ready pod count against the KAS Deployment's expected replica count to prevent verifying only a subset during informer cache lag or pod restarts
- Add dual-cert cross-checks: in `ensureNewSignerCertificatePropagated`, verify the old signer is also still trusted (detects mid-reload); in `ensureOldSignerCertificateRevoked`, verify the new signer cert is still trusted before declaring the old one revoked, preventing false positives when a pod restart temporarily rejects all certs
- Wire a KAS pod informer so pod readiness changes trigger immediate reconciliation
- Extract `verifyCertificateTrusted` and `verifyCertificateRevoked` helpers for per-pod SelfSubjectReview checks

**control-plane-operator** (RBAC):
- Grant `list`/`watch` on pods and `get` on deployments to the `control-plane-pki-operator` role so it can enumerate KAS pods and verify replica counts (update role asset + 5 test fixtures)

**support/podspec**:
- Add `IsPodReady` and `ContainerPort` helpers for pod readiness checks and named port lookup

**e2e / integration**:
- Add `TestCreateClusterHABreakGlassCredentials` exercising the break-glass credential flow on a HighlyAvailable control plane (3 KAS replicas)
- Add per-request timeout to revocation SSR polling in the integration test to prevent indefinite hangs when KAS stalls the TLS handshake during trust bundle reload
- Fix `hasWorkerNodes` detection for private clusters by checking NodePool replica counts instead of assuming workers exist

### Context

This was previously fixed in PRs #7405 + #7744 but reverted in #7784 because the e2e test added alongside it had a ~50% flake rate. The controller fix itself was sound — the flakiness was in the e2e test wiring. This PR re-applies the controller logic with comprehensive unit tests and a different, more targeted HA e2e test.

Fixes https://issues.redhat.com/browse/OCPBUGS-62177

The dual-cert cross-checks are the key addition beyond the original fix. In `ensureOldSignerCertificateRevoked`, verifying the new cert is trusted before declaring the old one revoked prevents false positives when a KAS pod restart during revocation causes the old cert to become unreachable (connection refused), which looks like revocation but is actually the pod being down. Similarly, in `ensureNewSignerCertificatePropagated`, verifying the old cert is also still trusted prevents declaring propagation complete when a pod is mid-reload and happens to accept the new cert but hasn't finished loading the full trust bundle.

## Checklist:
- [x] This change includes e2e.

### DIFF
--- a/control-plane-pki-operator/certificaterevocationcontroller/certificaterevocationcontroller.go
+++ b/control-plane-pki-operator/certificaterevocationcontroller/certificaterevocationcontroller.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"math/big"
+	"net"
 	"sort"
 	"strconv"
 	"strings"
@@ -20,6 +21,8 @@ import (
 	hcpmanifests "github.com/openshift/hypershift/control-plane-operator/controllers/hostedcontrolplane/manifests"
 	"github.com/openshift/hypershift/control-plane-pki-operator/certificates"
 	"github.com/openshift/hypershift/control-plane-pki-operator/manifests"
+	"github.com/openshift/hypershift/support/config"
+	"github.com/openshift/hypershift/support/podspec"
 
 	"github.com/openshift/library-go/pkg/certs/cert-inspection/certgraphanalysis"
 	"github.com/openshift/library-go/pkg/certs/cert-inspection/certgraphapi"
@@ -54,9 +57,14 @@ type CertificateRevocationController struct {
 	getSecret    func(namespace, name string) (*corev1.Secret, error)
 	listSecrets  func(namespace string) ([]*corev1.Secret, error)
 	getConfigMap func(namespace, name string) (*corev1.ConfigMap, error)
+	listPods     func(namespace string, selector labels.Selector) ([]*corev1.Pod, error)
 
 	// for unit testing only
 	skipKASConnections bool
+	// overrideVerifyCertAgainstKASPods, when non-nil, replaces verifyCertificateAgainstAllKASPods
+	// for unit testing. This allows tests to control per-pod verification behavior without
+	// requiring real KAS pod connections.
+	overrideVerifyCertAgainstKASPods func(ctx context.Context, namespace string, adminCfg *rest.Config, certPEM, keyPEM []byte, verifyFunc func(ctx context.Context, client kubernetes.Interface) (bool, error)) (bool, error)
 }
 
 // TODO: we need some sort of time-based GC for completed CRRs
@@ -85,11 +93,15 @@ func NewCertificateRevocationController(
 		getConfigMap: func(namespace, name string) (*corev1.ConfigMap, error) {
 			return kubeInformersForNamespaces.InformersFor(namespace).Core().V1().ConfigMaps().Lister().ConfigMaps(namespace).Get(name)
 		},
+		listPods: func(namespace string, selector labels.Selector) ([]*corev1.Pod, error) {
+			return kubeInformersForNamespaces.InformersFor(namespace).Core().V1().Pods().Lister().Pods(namespace).List(selector)
+		},
 	}
 
 	crrInformer := hypershiftInformers.Certificates().V1alpha1().CertificateRevocationRequests().Informer()
 	secretInformer := kubeInformersForNamespaces.InformersFor(hostedControlPlane.Namespace).Core().V1().Secrets().Informer()
 	configMapInformer := kubeInformersForNamespaces.InformersFor(hostedControlPlane.Namespace).Core().V1().ConfigMaps().Informer()
+	podInformer := kubeInformersForNamespaces.InformersFor(hostedControlPlane.Namespace).Core().V1().Pods().Informer()
 	listCRRs := func(namespace string) ([]*certificatesv1alpha1.CertificateRevocationRequest, error) {
 		return hypershiftInformers.Certificates().V1alpha1().CertificateRevocationRequests().Lister().CertificateRevocationRequests(hostedControlPlane.Namespace).List(labels.Everything())
 	}
@@ -98,6 +110,10 @@ func NewCertificateRevocationController(
 		WithInformersQueueKeysFunc(enqueueCertificateRevocationRequest, crrInformer).
 		WithInformersQueueKeysFunc(enqueueSecret(listCRRs), secretInformer).
 		WithInformersQueueKeysFunc(enqueueConfigMap(listCRRs), configMapInformer).
+		// KAS pod readiness and PodIP gate certificate verification progress; wiring
+		// the pod informer ensures that pod transitions trigger immediate reconciliation
+		// instead of waiting for the next ResyncEvery or SyntheticRequeueError cycle.
+		WithInformersQueueKeysFunc(enqueueKASPod(listCRRs, hostedControlPlane.Namespace), podInformer).
 		WithSync(c.syncCertificateRevocationRequest).
 		ResyncEvery(time.Minute).
 		ToController("CertificateRevocationController", eventRecorder.WithComponentSuffix(c.fieldManager))
@@ -244,6 +260,20 @@ func enqueueConfigMap(listCRRs func(namespace string) ([]*certificatesv1alpha1.C
 			return nil
 		}
 		return enqueueForSigner(configMap.Namespace, signer, listCRRs)
+	}
+}
+
+func enqueueKASPod(listCRRs func(namespace string) ([]*certificatesv1alpha1.CertificateRevocationRequest, error), namespace string) func(obj runtime.Object) []string {
+	return func(obj runtime.Object) []string {
+		pod, ok := obj.(*corev1.Pod)
+		if !ok {
+			klog.ErrorS(fmt.Errorf("unexpected object of type %T, wanted %T", obj, &corev1.Pod{}), "could not determine queue key")
+			return nil
+		}
+		if !kasAppLabelSelector.Matches(labels.Set(pod.Labels)) {
+			return nil
+		}
+		return enqueueAll(namespace, listCRRs)
 	}
 }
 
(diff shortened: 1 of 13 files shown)


control-plane-pki-operator/certificaterevocationcontroller/certificaterevocationcontroller.go
```diff
@@ -535,6 +543,91 @@ func (c *CertificateRevocationController) generateNewSignerCertificate(ctx conte
...
+		podCfg := rest.AnonymousClientConfig(adminCfg)
+		podCfg.TLSClientConfig.CertData = certPEM
+		podCfg.TLSClientConfig.KeyData = keyPEM
+		podCfg.Host = fmt.Sprintf("https://%s", net.JoinHostPort(pod.Status.PodIP, strconv.Itoa(int(port))))
+		// we're connecting directly to a pod IP, so we can't verify the server certificate against
+		// the CA that signed it for the service hostname; set ServerName for SNI and skip verification
+		podCfg.TLSClientConfig.ServerName = hcpmanifests.KubeAPIServerServiceName
+		podCfg.TLSClientConfig.Insecure = true //nolint:gosec // connecting directly to a pod IP in the same namespace
+		podCfg.TLSClientConfig.CAData = nil
+		podCfg.TLSClientConfig.CAFile = ""
```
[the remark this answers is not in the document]
sdminonne: Fixed in c345009ea9. Removed `Insecure=true` and the CA-clearing lines. `rest.AnonymousClientConfig` already copies `CAFile`/`CAData` from the admin config, and setting `ServerName` to the KAS service hostname is sufficient for Go's TLS to verify the certificate against that name rather than the pod IP. CA verification is now preserved for pod-level connections.

control-plane-pki-operator/certificaterevocationcontroller/certificaterevocationcontroller_test.go
```diff
@@ -1076,3 +1084,315 @@ func compareActions() []cmp.Option {
...
+							Status: corev1.ConditionTrue,
+						}},
+					},
+				},
+			},
+			testFunc: func(_ context.Context, _ kubernetes.Interface) (bool, error) {
+				return true, nil
+			},
+			expectedResult: true,
+		},
```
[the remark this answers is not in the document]
sdminonne: Fixed in 0e87b3d339. Added a `callCount` variable that wraps `testFunc` with a counting decorator. Each test case now declares `expectedCalls` (2 for the all-ready-pods case, 1 for the returns-false case, 0 for cases where testFunc should not be called). The assertion `if callCount \!= testCase.expectedCalls` fires after the call, catching any regression where the function short-circuits to a single pod.

control-plane-pki-operator/certificaterevocationcontroller/certificaterevocationcontroller.go
```diff
@@ -535,6 +559,98 @@ func (c *CertificateRevocationController) generateNewSignerCertificate(ctx conte
...
+		podClient, err := kubernetes.NewForConfig(podCfg)
+		if err != nil {
+			return false, fmt.Errorf("couldn't create client for KAS pod %s/%s: %w", pod.Namespace, pod.Name, err)
+		}
+
+		podCtx, cancel := context.WithTimeout(ctx, perPodVerifyTimeout)
+		passed, err := verifyFunc(podCtx, podClient)
+		cancel()
+		if err != nil {
+			return false, err
```
jparrill: Add details on the error's return.

control-plane-pki-operator/certificaterevocationcontroller/certificaterevocationcontroller_test.go
```diff
@@ -1076,3 +1087,438 @@ func compareActions() []cmp.Option {
...
+			} else {
+				g.Expect(err).ToNot(HaveOccurred())
+			}
+			g.Expect(result).To(Equal(testCase.expectedResult))
+			g.Expect(callCount).To(Equal(testCase.expectedCalls))
+		})
+	}
+}
+
+func TestEnqueueKASPod(t *testing.T) {
```
jparrill, line 1359: Do you mind add an Error path for the test function?

control-plane-pki-operator/certificaterevocationcontroller/certificaterevocationcontroller.go
```diff
@@ -535,6 +559,98 @@ func (c *CertificateRevocationController) generateNewSignerCertificate(ctx conte
 	return false, nil, false, nil
 }
 
+func isPodReady(pod *corev1.Pod) bool {
```
jparrill: Looks like this is necessary for the implementation. Do you mind extract it and put it as a public support/util tool? (also the tests)

control-plane-pki-operator/certificaterevocationcontroller/certificaterevocationcontroller.go
```diff
@@ -247,6 +258,19 @@ func enqueueConfigMap(listCRRs func(namespace string) ([]*certificatesv1alpha1.C
 	}
 }
 
+func enqueueKASPod(listCRRs func(namespace string) ([]*certificatesv1alpha1.CertificateRevocationRequest, error), namespace string) func(obj runtime.Object) []string {
+	return func(obj runtime.Object) []string {
+		pod, ok := obj.(*corev1.Pod)
+		if !ok {
+			return nil
```
jparrill, line 271: Add klog.ErrorS for type assertion failure in enqueueKASPod for consistency with `enqueueSecret` and `enqueueConfigMap`

jparrill (commented):

Btw, do you have plans to create an E2E in a follow up PR? If not is ok, we would need at least to have a ticket tracking it for the future implementation.

control-plane-pki-operator/certificaterevocationcontroller/certificaterevocationcontroller.go
sdminonne: will do

sdminonne:
@jparrill

I addresse all the comments

About https://github.com/openshift/hypershift/pull/8263#pullrequestreview-4179567902 for the e2e we have to discuss as a team. 
This fix address specifically HA clusters and in our e2e only the requesing serving e2e has it.  It was very flaky.
The current implementation is exetensively covered by unit tests only.

See context of this PR for the quick explanation about history.

jparrill (commented):
/approve

sdminonne:
https://github.com/openshift/hypershift/pull/8263/changes/745cefb08eb4f8cb90be4808093139a8fffb5d9b#diff-1ab8a204348b5b2e9d5d5995903341902c62d82f84061b1b7cd9e0858702e6c1R152

and

```
 control_plane_pki_operator.go:139: validating that the client certificate provides the appropriate access
         control_plane_pki_operator.go:125: amending the existing kubeconfig to use break-glass client certificate credentials
         control_plane_pki_operator.go:142: issuing SSR to identify the subject we are given using the client certificate
         control_plane_pki_operator.go:162: ensuring that the SSR identifies the client certificate as having system:masters power and correct
     username
     === RUN   TestCreateClusterHABreakGlassCredentials/Main/break-glass-credentials/customer-break-glass/CSR_flow/revocation
         pki.go:76: loading certificate/key pair from disk for signer customer-break-glass, use $REGENERATE_PKI to generate new ones
         control_plane_pki_operator.go:262: creating CRR
     e2e-clusters-zf8ng-ha-break-glass-credentials-8jxvn/27p8q5lri4z8y5mo5y1j9wo44jvdb1g4ndydwybeyevv to trigger signer certificate revocation
     === NAME  TestCreateClusterRequestServingIsolation/Teardown
         hypershift_framework.go:574: Destroyed cluster. Namespace: e2e-clusters-jpzxx, name: request-serving-isolation-pxllz
         hypershift_framework.go:529: archiving
     /logs/artifacts/TestCreateClusterRequestServingIsolation/hostedcluster-request-serving-isolation-pxllz to
     /logs/artifacts/TestCreateClusterRequestServingIsolation/hostedcluster.tar.gz
     --
         control_plane_pki_operator.go:227: Successfully waited for CSR "2l6aojtbxuaf6k7p7j9ybiejevrtgd0wfnhp8dch5l32" to be approved and signed in
      3s
         control_plane_pki_operator.go:139: validating that the client certificate provides the appropriate access
         control_plane_pki_operator.go:125: amending the existing kubeconfig to use break-glass client certificate credentials
         control_plane_pki_operator.go:142: issuing SSR to identify the subject we are given using the client certificate
         control_plane_pki_operator.go:162: ensuring that the SSR identifies the client certificate as having system:masters power and correct
     username
     === RUN   TestCreateClusterHABreakGlassCredentials/Main/break-glass-credentials/sre-break-glass/CSR_flow/revocation
         pki.go:76: loading certificate/key pair from disk for signer sre-break-glass, use $REGENERATE_PKI to generate new ones
         control_plane_pki_operator.go:262: creating CRR
     e2e-clusters-zf8ng-ha-break-glass-credentials-8jxvn/17odmpcht0fxq10pp1iq7wcv8xbichoeh2a4ifp3uj01 to trigger signer certificate revocation
     === NAME  TestCreateCluster/Main/break-glass-credentials/sre-break-glass/CSR_flow/revocation
         control_plane_pki_operator.go:269: Successfully waited for CRR
     e2e-clusters-jdmht-create-cluster-4pvlz/2u0wgv97sblna6w4ipj6cj2b6uzsqg267zo8tk74cq6l to complete in 2m6s
         control_plane_pki_operator.go:282: creating a client using the a certificate from the revoked signer
     --
         util.go:376: Successfully waited for a successful connection to the guest API server in 0s
         util.go:1987: NodePool replicas: 1
     === RUN   TestCreateCluster/Main/EnsureGlobalPullSecret/Check_Replace_nodes_have_globalPS_label_from_CAPI_propagation
         util.go:2008: All 1 nodes have the globalPS label
     === RUN   TestCreateCluster/Main/EnsureGlobalPullSecret/Check_if_GlobalPullSecret_secret_is_in_the_right_place_at_Dataplane
     === NAME  TestCreateClusterHABreakGlassCredentials/Main/break-glass-credentials/customer-break-glass/CSR_flow/revocation
         control_plane_pki_operator.go:269: Successfully waited for CRR
     e2e-clusters-zf8ng-ha-break-glass-credentials-8jxvn/27p8q5lri4z8y5mo5y1j9wo44jvdb1g4ndydwybeyevv to complete in 2m0s
         control_plane_pki_operator.go:282: creating a client using the a certificate from the revoked signer
         control_plane_pki_operator.go:125: amending the existing kubeconfig to use break-glass client certificate credentials
         control_plane_pki_operator.go:285: issuing SSR to confirm that we're not authorized to contact the server
     === RUN   TestCreateCluster/Main/EnsureGlobalPullSecret/Check_if_the_DaemonSet_is_present_in_the_DataPlane
     --
         karpenter_test.go:881: Node provisioned by capacity-reservation-test NodePool is ready
         karpenter_test.go:887: Verifying EC2 instance i-0fdb6b5af6b17423a was launched into capacity reservation cr-012ebe62f2729af0e
         karpenter_test.go:891: Instance i-0fdb6b5af6b17423a correctly launched into capacity reservation cr-012ebe62f2729af0e
     === NAME  TestCreateCluster/Main/EnsureGlobalPullSecret/Wait_for_critical_DaemonSets_to_be_ready_-_second_check
         util.go:2183: DaemonSet global-pull-secret-syncer not ready: 2/3 pods ready
     === NAME  TestCreateClusterHABreakGlassCredentials/Main/break-glass-credentials/sre-break-glass/CSR_flow/revocation
         control_plane_pki_operator.go:269: Successfully waited for CRR
     e2e-clusters-zf8ng-ha-break-glass-credentials-8jxvn/17odmpcht0fxq10pp1iq7wcv8xbichoeh2a4ifp3uj01 to complete in 2m9s
         control_plane_pki_operator.go:282: creating a client using the a certificate from the revoked signer
         control_plane_pki_operator.go:125: amending the existing kubeconfig to use break-glass client certificate credentials
         control_plane_pki_operator.go:285: issuing SSR to confirm that we're not authorized to contact the server
     === RUN   TestCreateClusterHABreakGlassCredentials/EnsureHostedCluster
     --
                 --- PASS: TestCreateClusterHABreakGlassCredentials/Main/break-glass-credentials/independent_signers (135.12s)
                 --- PASS: TestCreateClusterHABreakGlassCredentials/Main/break-glass-credentials/customer-break-glass (126.13s)
                     --- PASS: TestCreateClusterHABreakGlassCredentials/Main/break-glass-credentials/customer-break-glass/direct_fetch (0.03s)
                     --- PASS: TestCreateClusterHABreakGlassCredentials/Main/break-glass-credentials/customer-break-glass/CSR_flow (126.10s)
                         --- PASS:
     TestCreateClusterHABreakGlassCredentials/Main/break-glass-credentials/customer-break-glass/CSR_flow/invalid_CN_flagged_in_status (3.01s)
                         --- PASS: TestCreateClusterHABreakGlassCredentials/Main/break-glass-credentials/customer-break-glass/CSR_flow/revocation
     (120.03s)
                 --- PASS: TestCreateClusterHABreakGlassCredentials/Main/break-glass-credentials/sre-break-glass (135.17s)
                     --- PASS: TestCreateClusterHABreakGlassCredentials/Main/break-glass-credentials/sre-break-glass/direct_fetch (0.02s)
                     --- PASS: TestCreateClusterHABreakGlassCredentials/Main/break-glass-credentials/sre-break-glass/CSR_flow (135.15s)
                         --- PASS:
     TestCreateClusterHABreakGlassCredentials/Main/break-glass-credentials/sre-break-glass/CSR_flow/invalid_CN_flagged_in_status (3.02s)
                         --- PASS: TestCreateClusterHABreakGlassCredentials/Main/break-glass-credentials/sre-break-glass/CSR_flow/revocation
     (129.04s)
         --- PASS: TestCreateClusterHABreakGlassCredentials/EnsureHostedCluster (2.76s)

```

sdminonne:
```
 The sole test failure (1 of 367 tests, 41 skipped) is                                                                                                                                               
    TestCreateClusterHABreakGlassCredentials. It failed during infrastructure                                                                                                                                  
    provisioning before the hosted cluster was ever created. The Azure                                                                                                                                         
    VNet creation returned a 400 error because the constructed VNet name                                                                                                                                       
    'ha-break-glass-credentials-7mbxh-ha-break-glass-credentials-7mbxh'                                                                                                                                        
    (65 chars) is invalid -- Azure VNet names have a 64-character limit.

    The root cause is a name-duplication bug in the Azure infra creation
    code. When Name == InfraID (both are 'ha-break-glass-credentials-7mbxh'),
    the VNet name is built as: name + "-" + infraID, which doubles the
    value into a 65-char string that exceeds Azure's 64-char VNet name limit.
```

Going to fix this

test/e2e/create_cluster_test.go
```diff
@@ -143,6 +145,65 @@ func TestCreateCluster(t *testing.T) {
 		Execute(&clusterOpts, globalOpts.Platform, globalOpts.ArtifactDir, "create-cluster", globalOpts.ServiceAccountSigningKey)
 }
 
+// TestCreateClusterHABreakGlassCredentials exercises the break-glass credential flow
+// on a HighlyAvailable control plane (3 KAS replicas). This validates that the
+// CertificateRevocationController correctly verifies certificate revocation against
+// each individual KAS pod rather than through the service load balancer.
+func TestCreateClusterHABreakGlassCredentials(t *testing.T) {
```
muraee, line 152: why do we need a new test cluster? the original TestCreateCluster runs in HA mode already

sdminonne: TBH  I don't think so, that's why in the previous PR (the reverted one) I modified the request serving e2e which was flaky.

`TestCreateCluster` does not run with an HA control plane — it uses the default ControlPlaneAvailabilityPolicy which is SingleReplica and it never overrides.               https://github.com/openshift/hypershift/blob/6b39d47ed43ca9b099209578cff39805c0943270/test/e2e/util/options.go#L217

TestCreateCluster only sets InfrastructureAvailabilityPolicy to HighlyAvailable when 3+ zones are available — that controls worker node placement across zones, not control plane replicas
https://github.com/openshift/hypershift/blob/6b39d47ed43ca9b099209578cff39805c0943270/test/e2e/create_cluster_test.go#L47

The new test explicitly sets ControlPlaneAvailabilityPolicy = HighlyAvailable to get 3 KAS replicas. 
https://github.com/openshift/hypershift/blob/6b39d47ed43ca9b099209578cff39805c0943270/test/e2e/util/options.go#L217

csrwng:
## E2E test feedback

The new `TestCreateClusterHABreakGlassCredentials` creates a full HA cluster with cloud infrastructure and worker nodes, but the break-glass credential flow only needs a functioning KAS — it doesn't require worker nodes or cloud-specific infrastructure.

Consider using `NonePlatform` with `NodePoolReplicas: 0` (similar to how `TestHAEtcdChaos` works). This would be significantly cheaper while still exercising the per-pod cert revocation path against 3 KAS replicas:

```go
clusterOpts := globalOpts.DefaultClusterOptions(t)
clusterOpts.ControlPlaneAvailabilityPolicy = string(hyperv1.HighlyAvailable)
clusterOpts.NodePoolReplicas = 0

// ...
}).Execute(&clusterOpts, hyperv1.NonePlatform, globalOpts.ArtifactDir, "ha-break-glass-creds", globalOpts.ServiceAccountSigningKey)
```

I verified that:
- `WaitForGuestClient` only waits for a successful `SelfSubjectReview` against the guest API, not for nodes
- `RunTestControlPlanePKIOperatorBreakGlassCredentials` doesn't skip `NonePlatform` (only KubeVirt and GCP are skipped)
- The test only needs management and guest API clients, both of which work without worker nodes

Alternatively, if you'd prefer to piggyback on an existing HA cluster to avoid creating one at all, `TestUpgradeControlPlane` already creates an HA cluster and could host this as a final subtest (with a skip when `DisablePKIReconciliation` is set). That would have the tradeoff of coupling the two tests, but would save the cost of an entire cluster.

sdminonne:
Thanks @csrwng I'm implemeting the zero-workers nonePlatform approach

sdminonne:
**Root Cause:** The test uses hyperv1.NonePlatform in an AKS environment that requires explicit route hostnames. Since `NonePlatform` **doesn't configure route hostnames in service publishing strategies,** the OAuth
   route was created without a spec.host field. The control-plane-operator's infrastructure readiness check requires valid hosts on routes, so it blocked indefinitely.

sdminonne:
@csrwng 

Thanks for the suggestion — I tried switching to NonePlatform but it doesn't work for this test in either CI environment:

  - AKS: NonePlatform doesn't configure route hostnames in service publishing strategies, so the OAuth route is created without spec.host. The CPO infrastructure readiness check requires valid hosts on routes and blocks
  indefinitely — the cluster never comes up.
  - AWS: NonePlatform causes the test framework to skip cluster validation, so the test races ahead before the HCCO finishes reconciling CA bundles (client-ca, total-client-ca). KAS doesn't trust the break-glass signer CA
  yet, causing Unauthorized on all SSR checks.

Anyway I modified a test avoiding creation of yet another hosted cluster.
Hope it's OK.

sdminonne:
` --- PASS: TestCreateClusterHABreakGlassCredentials/ValidateHostedCluster (670.23s)`

test/e2e/util/hypershift_framework.go
```diff
@@ -289,15 +289,32 @@ func (h *hypershiftTest) after(hostedCluster *hyperv1.HostedCluster, platform hy
...
+			hasWorkerNodes := false
 			if !netutil.IsPrivateHC(hostedCluster) {
 				guestClient := WaitForGuestClient(t, t.Context(), h.client, hostedCluster)
 				var nodeList corev1.NodeList
 				if err := guestClient.List(t.Context(), &nodeList); err != nil {
 					t.Errorf("failed to list nodes in guest cluster: %v", err)
 				}
 				hasWorkerNodes = len(nodeList.Items) > 0
+			} else {
+				// Private clusters are not reachable from the test runner;
```
sdminonne, line 301: @csrwng FYI

sdminonne:
https://redhat.atlassian.net/browse/CNTRLPLANE-3430

PR #8265, opened 2026-04-16
AUTOSCALE-615: include Karpenter node vCPUs in billing metric

maxcao13:

Add `VCPUs` field to `AutoNodeStatus`, computed by karpenter-operator from `NodeClaim` capacity and cross-referenced against live Node objects. The metrics collector seeds per-cluster vCPU count from this field before accumulating native NodePool vCPUs on top.

NodeClaims and karpenter NodePools exist directly in the guest cluster, so we cannot source this info directly from the hypershift nodepool metrics package using the existing code/clients, which only has access to the management cluster.

We need this because ROSA HCP uses the hypershift_cluster_vcpus metric for billing our customers. This metric is not correctly reflecting the vCPUs for Nodes that are provisoned by Karpenter. This will result in under-billing of customers that are using Karpenter to provision Nodes within their cluster.

Fixes https://redhat.atlassian.net/browse/AUTOSCALE-615

Alternatively, we could have karpenter-operator source its own billing metric and ask ROSA to aggregate using this new exposed metric, but considering that we already expose information about nodes and nodeclaims in the `AutoNodeStatus`, I think it makes sense to be consistent here.

Made-with: Cursor

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * HostedCluster status now includes an optional AutoNode vCPU count; metrics now seed and aggregate AutoNode plus native node vCPUs for improved billing and capacity visibility.
  * Karpenter now sums live NodeClaim CPU capacity when computing AutoNode vCPUs and reacts to NodeClaim changes that affect CPU or assignment.

* **Tests**
  * Added unit and e2e tests covering vCPU aggregation, error reporting, metric convergence, and billing/vCPU validation during provisioning and consolidation.


joshbranham:
For HyperShift NodePools we return `-1` if we hit an error calculating a vCPU count. We are not doing that here, as proposed, but I think that is fine since that error/calculation happens downstream in karpenter. 

/lgtm
/test e2e-aws-techpreview

maxcao13:
/hold

Gonna hold so that 8192 gets in first.

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -2207,6 +2207,14 @@ type AutoNodeStatus struct {
 	// +kubebuilder:validation:Minimum=0
 	// +optional
 	NodeClaimCount *int32 `json:"nodeClaimCount,omitempty"`
+
+	// vcpus is the total number of virtual CPUs across all Karpenter-managed nodes
```
JoelSpeed, line 2209: Why vcpus as opposed to vCPUs?

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -2207,6 +2207,14 @@ type AutoNodeStatus struct {
...
+	// that have registered and reported capacity. This is the sum of CPU capacity
+	// from each NodeClaim whose corresponding node exists (Status.NodeName is set).
+	// Used by the metrics collector for billing aggregation.
+	// +kubebuilder:validation:Minimum=0
```
JoelSpeed, line 2214: You can have no vCPUs? When would you have 0?

Can we set a sensible upper limit for this value?

maxcao13: We would have 0 presumably when there are no Karpenter nodes available to aggregate the vcpus from. 

I'm not sure what a sensible limit would be here, isn't it unbounded? This is the total number of vCPUS across all provisioned Karpenter nodes. There's a soft HCP max node limit of 500 I believe, but I'm not sure there would be reason to cap it since instance types will also vary in size.

api/hypershift/v1beta1/hostedcluster_types.go
maxcao13: No reason, I can change the comment here if that fits the API conventions.

maxcao13:
/hold

holding for comments from joel, but also these e2es are permafailing due to infra as well.

api/hypershift/v1beta1/hostedcluster_types.go
maxcao13: Never mind, there is a reason according to linting rules:
```
hypershift/v1beta1/hostedcluster_types.go:2211:2: commentstart: godoc for field AutoNodeStatus.VCPUs should start with 'vcpus ...' (kubeapilinter)
```

api/hypershift/v1beta1/hostedcluster_types.go
JoelSpeed: Ok, so you would want `0` to be displayed to the end user when no nodes are provisioned? In which case this is good as is

> I'm not sure what a sensible limit would be here, isn't it unbounded? This is the total number of vCPUS across all provisioned Karpenter nodes. There's a soft HCP max node limit of 500 I believe, but I'm not sure there would be reason to cap it since instance types will also vary in size.

Ok so in this case, there are several things to consider.

Could you ever see a cluster having more than 1,000,000 cores for example? (10k nodes with 1k CPU each?)

If yes, how large could it go? Could it be more than 2^32?

Are there any consumers of this API that might not be able to handle supremely large numbers?

joshbranham: I am not sure what the upper limit of vCPU would be, but yeah we have a soft cap of 500 nodes in ROSA HCP. If you can come up with a very high limit then that seems fine. Right now the underlying metric we use for billing is exposed by hypershift-operator as a prom metric. There should be no other consumers of this API that I am aware of, given it's driven purely from a service provider perspective with a need to track billing for NodePools in Hypershift and Karpenter.

maxcao13: Theoretically with the 500 soft cap, and the fact that the largest CPU-focused instance type supported by Karpenter has 192 vCPUs, we could have some arbitrary max limit of 500 * 192 * 2 (margin) = 192000. But I know Joel, you may or may not be looking at some etcd sharding stuff that could increase this node limit :smile:, plus instance type maximums can change as well. But even with the case with 10k nodes and 1k cpu each, 1,000,000 cores, it should never happen in the ROSA case and the only consumers of the API should be able to handle that. 

If we really need a number here to satisfy openshift API standards, my instinct is to just make the maximum the golang `math.MaxInt` number. Would that be fine?

EDIT: As per slack, setting it to 1,000,000 as a reasonable upper limit we will never get to.

maxcao13:
this has gone through so many rebases now, that it would be a disservice not to check if autonode tests still pass

/test e2e-aws-autonode

PR #8266, opened 2026-04-16
OCPBUGS-83397: Fix concurrent CRD reconciliation race

jkyros:
We were racing on our CRD globals:
- CreateOrUpdate mutates the CRD variable
- With enough concurrency, if we get a partial/incomplete object, it gets stuck that way and never gets fixed 
- This has been in there since the beginning, we didn't break it, we just never saw it until we hit high concurrency 
- I also added a concurrent access test that verifies the fix (I did test it without the fix, and it totally fails, so if we had that test, we would have caught it) 

Fixes [OCPBUGS-83397](https://redhat.atlassian.net/browse/OCPBUGS-83397)

## Checklist:

## Release Notes

* **Bug Fixes**
  * Improved Karpenter operator stability by fixing CRD object mutation issues that could occur during concurrent reconciliation operations, ensuring consistent state handling.

* **Tests**
  * Added comprehensive concurrent access tests for Karpenter operator's CRD reconciliation logic to prevent race conditions.

### DIFF
--- a/karpenter-operator/controllers/karpenter/karpenter_controller.go
+++ b/karpenter-operator/controllers/karpenter/karpenter_controller.go
@@ -376,11 +376,12 @@ func (r *Reconciler) reconcileCRDs(ctx context.Context, onlyCreate bool) error {
 	errs := []error{}
 	var op controllerutil.OperationResult
 	var err error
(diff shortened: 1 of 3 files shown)


enxebre:
thanks!
/approve

PR #8273, opened 2026-04-17
OCPBUGS-83730: Add *.apps wildcard to base domain Private DNS zone for Azure private clusters

bryan-cox:
## Summary

- Add `*.apps.<clusterName>` A record to the base domain Private DNS zone for Azure private HostedClusters with PrivateLink
- Include the new record in cleanup/deletion path
- Add e2e test that queries Azure DNS to validate the expected records exist

## Problem

On Azure private HostedClusters, the `AzurePrivateLinkServiceReconciler` creates a Private DNS zone for the base domain with `api-<cluster>` and `oauth-<cluster>` records, but no `*.apps` wildcard. Guest cluster services like the console operator fail to resolve `*.apps.<cluster>.<basedomain>` because:

1. The `*.apps` wildcard only exists in the `.hypershift.local` zone
2. The guest cluster's ingress domain is `apps.<cluster>.<basedomain>`, not `hypershift.local`
3. DNS resolution fails with `no such host`, causing console operator degradation and preventing ClusterVersion completion

## Fix

Add `*.apps.<clusterName>` to the records created by `recordNamesForCR()` for the private-router CR. On private clusters, all `*.apps` traffic flows through the private-router via the passthrough ingress route (`PrivateStrategyType`), so pointing the wildcard to the Private Endpoint IP is correct.

This matches the AWS approach where `*.apps` records are created in the private Route53 zone.

## Testing

- Unit tests updated and passing
- All 4 affected test cases updated to include the new `*.apps` record in both creation and deletion assertions
- Added `BaseDomainDNSRecordsExist` e2e subtest to `TestAzurePrivateTopology` that queries the Azure Private DNS RecordSets API to verify `api-<cluster>` and `*.apps.<cluster>` A records exist in the base domain zone. This would have caught the original bug — the v1 e2e only checked that the zone ID was populated in the CR status, not that it contained the correct records.

## JIRA

https://issues.redhat.com/browse/OCPBUGS-83730

## Summary by CodeRabbit

* **New Features**
  * private-router now manages both the API A record and the wildcard apps (*.apps) base-domain A record; oauth record handling remains conditional.

* **Tests**
  * Controller tests updated to expect creation and deletion of the additional *.apps A record.
  * Added an E2E subtest verifying the base-domain contains both api and *.apps A records.

### DIFF
--- a/control-plane-operator/controllers/azureprivatelinkservice/controller.go
+++ b/control-plane-operator/controllers/azureprivatelinkservice/controller.go
@@ -745,15 +745,17 @@ func (r *AzurePrivateLinkServiceReconciler) reconcileDNS(ctx context.Context, az
 }
 
 // reconcileBaseDomainDNS creates a Private DNS Zone for the cluster's base domain,
-// links it to the guest VNet, and creates A records for the API and/or OAuth hostnames.
-// This enables worker VMs to resolve api-<name>.<basedomain> and oauth-<name>.<basedomain>
-// to the Private Endpoint IP, which is required for the console OAuth flow and other
-// services that use the external API/OAuth hostnames from within the private network.
+// links it to the guest VNet, and creates A records for the API, *.apps, and/or OAuth
+// hostnames. This enables worker VMs to resolve api-<name>.<basedomain>,
+// *.apps.<name>.<basedomain>, and oauth-<name>.<basedomain> to the Private Endpoint IP,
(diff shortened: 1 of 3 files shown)


Nirshal (approved):
/lgtm

bryan-cox:

This is a known issue not caused by this PR.

PR #8279, opened 2026-04-18
NO-JIRA: chore: pass COMMIT_HASH build arg to support git worktrees

cblecker:
## Summary

- Adds `ARG COMMIT_HASH` to all Dockerfiles/Containerfiles so the commit hash can be injected at build time
- Updates `make docker-build` to pass `--build-arg COMMIT_HASH=$(COMMIT_HASH)`, resolving the hash on the host where git works correctly
- Updates `hack/container-run.sh` to pass `-e COMMIT_HASH=` for the same reason

In a git worktree, `.git` is a file containing a pointer to the main repo's object store rather than a directory. When container builds copy this pointer file into the build context, `git rev-parse HEAD` inside the container fails silently, producing binaries with `<unknown>` version info. This was previously attempted via `.dockerignore` (#7944) but reverted (#8201) because CI needs `.git` inside the container. This approach instead resolves the hash on the host and passes it in — when no arg is provided (CI), the existing behavior is fully preserved.

## Test plan

- [x] From a git worktree, run `make docker-build` and verify the image has correct version info, not `<unknown>`
- [x] From a regular clone, run `make docker-build` and verify no regression

## Summary by CodeRabbit

* **Chores**
  * Build and run workflows now capture and propagate the current commit hash into images and running containers. Images produced for CLI, control plane, operator, e2e and dev builds include commit metadata, and container run commands set the COMMIT_HASH environment variable for improved artifact identification, troubleshooting and traceability.

### DIFF
--- a/Containerfile.cli
+++ b/Containerfile.cli
@@ -1,4 +1,5 @@
 FROM brew.registry.redhat.io/rh-osbs/openshift-golang-builder:rhel_9_golang_1.25 AS builder
+ARG COMMIT_HASH
 
 WORKDIR /hypershift
 
--- a/Containerfile.control-plane
+++ b/Containerfile.control-plane
@@ -1,4 +1,5 @@
 FROM registry.access.redhat.com/ubi9/go-toolset:1.25.7-1773318690 AS builder
+ARG COMMIT_HASH
 
 WORKDIR /hypershift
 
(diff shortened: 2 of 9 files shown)


cblecker:
/hold
/cc @enxebre 

@enxebre This change might allow us to put the git tree into `.dockerignore` again?

enxebre:
>@enxebre This change might allow us to put the git tree into .dockerignore again?

 yes, although that would require to chase down all consumers builds.

Thanks!
/lgtm

PR #8287, opened 2026-04-20
CNTRLPLANE-3306: add ExternalOIDCWithUpstreamParity e2e tests

ehearne-redhat:

This change:

+ Adds extensive testing for the ExternalOIDCWithUpstreamParity feature gate in `control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth_test.go`.
  + Includes testing the discoveryUrl, userValidationRules, claimValidationRules, and claimMappings.

+ Tests the feature gate in a HyperShift cluster in `test/e2e/external_oidc_test.go`.

+ Provides basic auth config testing of the feature gate in `test/e2e/util/external_oidc.go` .

+ Configures the control plane operator feature sets in hypershift-operator.
  + This change is required as hypershift-operator re-uses control-plane-operator
      auth config validation code, which checks specifically for control-plane-operator
      feature set enablement.

      Without this change, if we try to enable feature set such as TechPreviewNoUpgrade,
      validation will fail as hypershift-operator does not configure control-plane-operator
      feature sets.

+ Refactors existing tests to use a specified AuthConfig so that when the feature graduates it is easier to add the appropriate fields.
  + Tests can now interact with the keycloak instance via API so they can create 
     new users/groups as needed for better testing coverage.

This change will allow us to test feature gates behind TechPreviewNoUpgrade and
others that are present in control-plane-operator, throughout HyperShift.

This should allow us to progress in promotion of the feature from TechPreview to GA.

https://redhat.atlassian.net/browse/CNTRLPLANE-3306.

Allows us to progress in promotion of the ExternalOIDCWithUpstreamParity feature from TechPreview to GA.

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Added External OIDC with upstream parity (feature-gated): supports differing discovery URLs, CEL-based claim→username/groups mappings, multiple claim validation rules, and user validation rules (reject system: usernames).

* **Tests**
  * Expanded unit and end-to-end tests covering discovery URL propagation, complex CEL claim mappings (nested/conditional/orValue/defaults/filter/map), claim and user validation behaviors, error case for empty CEL claim expression, and runtime username/groups resolution.


control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth_test.go
```diff
@@ -1521,6 +1521,1084 @@ func TestAdaptAuthConfig(t *testing.T) {
...
+								},
+							},
+						},
+					},
+				},
+			},
+			shouldError: false,
+		},
+		{
+			name:   "multiple userValidationRules - AND logic, all rules must pass",
```
everettraven, line 2108: Unless I'm missing something, this seems like a duplicate of "valid userValidationRule - multiple expressions ANDed together"

control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth_test.go
```diff
@@ -1521,6 +1521,1084 @@ func TestAdaptAuthConfig(t *testing.T) {
...
+							AudienceMatchPolicy: AudienceMatchPolicyMatchAny,
+							Audiences:           []string{"one", "two"},
+						},
+						ClaimMappings: ClaimMappings{
+							Username: PrefixedClaimOrExpression{
+								Prefix: ptr.To("https://test.com#"),
+								Claim:  "username",
+							},
+							Groups: PrefixedClaimOrExpression{
+								Expression: "claims.roles.orValue([]).filter(r, r.startsWith('openshift-')).map(r, r.substring(10))",
```
everettraven, line 2567: I don't think this usage of `orValue` is actually valid.

You would need to use an optional key notation like:

```suggestion
								Expression: "claims.?roles.orValue([]).filter(r, r.startsWith('openshift-')).map(r, r.substring(10))",
```

test/e2e/util/external_oidc.go
```diff
@@ -149,6 +149,39 @@ func (config *ExtOIDCConfig) GetAuthenticationConfig() *configv1.AuthenticationS
...
+	if featuregates.Gate().Enabled(featuregates.ExternalOIDCWithUpstreamParity) {
+		// Use CEL expression for username mapping instead of static claim
+		authnSpec.OIDCProviders[0].ClaimMappings.Username = configv1.UsernameClaimMapping{
+			Expression: "claims.email.split('@')[0]",
+		}
+
+		// Use CEL expression for groups mapping instead of static claim
+		authnSpec.OIDCProviders[0].ClaimMappings.Groups = configv1.PrefixedClaimMapping{
+			TokenClaimMapping: configv1.TokenClaimMapping{
+				Expression: "has(claims.groups) && type(claims.groups) == list ? claims.groups : []",
```
everettraven, line 161: This could probably be simplified out to:
```suggestion
				Expression: "claims.?groups.orValue([])",
```

test/e2e/util/external_oidc.go
```diff
@@ -149,6 +149,39 @@ func (config *ExtOIDCConfig) GetAuthenticationConfig() *configv1.AuthenticationS
 		)
 	}
 
+	if featuregates.Gate().Enabled(featuregates.ExternalOIDCWithUpstreamParity) {
+		// Use CEL expression for username mapping instead of static claim
+		authnSpec.OIDCProviders[0].ClaimMappings.Username = configv1.UsernameClaimMapping{
+			Expression: "claims.email.split('@')[0]",
```
everettraven, line 155: Because you are using the email claim here, you have to either:
- check that the `email_verified` claim is `true` here
- check that the `email_verified` claim is `true` in the claim validation rules

test/e2e/external_oidc_test.go
```diff
@@ -109,5 +109,70 @@ func TestExternalOIDC(t *testing.T) {
...
 		}
+
+		if featuregates.Gate().Enabled(featuregates.ExternalOIDCWithUpstreamParity) {
+			t.Run("[OCPFeatureGate:ExternalOIDCWithUpstreamParity] Test CEL username expression mapping", func(t *testing.T) {
+				g := NewWithT(t)
+				t.Logf("begin to test CEL username expression mapping")
+				g.Expect(selfSubjectReview.Status.UserInfo.Username).NotTo(BeEmpty())
+				// Username should be the email prefix (before @) since we configured expression: claims.email.split('@')[0]
+				g.Expect(selfSubjectReview.Status.UserInfo.Username).NotTo(ContainSubstring("@"))
+				t.Logf("CEL username expression successfully mapped to: %s", selfSubjectReview.Status.UserInfo.Username)
```
everettraven, line 120: Should also make sure the actual username matches what we expect?

test/e2e/external_oidc_test.go
```diff
@@ -109,5 +109,70 @@ func TestExternalOIDC(t *testing.T) {
...
+			})
+
+			t.Run("[OCPFeatureGate:ExternalOIDCWithUpstreamParity] Test CEL groups expression mapping", func(t *testing.T) {
+				g := NewWithT(t)
+				t.Logf("begin to test CEL groups expression mapping")
+				// Groups expression uses: has(claims.groups) && type(claims.groups) == list ? claims.groups : []
+				// If the token has groups, they should be present without prefix (no prefix in CEL expression)
```
everettraven, line 127: For this test case, do we guarantee that the user will have groups configured?

test/e2e/external_oidc_test.go
```diff
@@ -109,5 +109,70 @@ func TestExternalOIDC(t *testing.T) {
...
+				g.Expect(oidcProvider.ClaimValidationRules).Should(HaveLen(1))
+				g.Expect(oidcProvider.ClaimValidationRules[0].Type).Should(Equal(configv1.TokenValidationRuleTypeCEL))
+
+				// Verify user validation rules are set
+				g.Expect(oidcProvider.UserValidationRules).Should(HaveLen(1))
+				g.Expect(oidcProvider.UserValidationRules[0].Expression).Should(Equal("!user.username.startsWith('system:')"))
+
+				t.Logf("All ExternalOIDCWithUpstreamParity configurations properly set")
+			})
+		}
```
everettraven, line 176: Do we actually need these to be separate test cases? It seems like these are all inherently tested by the other two tests that were added here so it doesn't seem like these being explicit cases adds much value other than quantity of test cases.

Is there a way that we could add some negative test cases here similar to that of our proposed standalone openshift test cases like:
- token that fails claim validation results in unauthorized response from the selfsubjectaccessreview request
- token that fails user validation results in unauthorized response from the selfsubjectaccessreview request

control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth_test.go
ehearne-redhat: Nice catch - yes, I believe it is! I'll remove that.

control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth_test.go
ehearne-redhat: OK - and I believe this is because `roles` fall under `claims.extra` which is optional from what I can gather. 

https://github.com/openshift/enhancements/blob/26be7aa617234b2a994e767045e335fd320dc993/enhancements/authentication/adding-uid-and-extra-claim-mapping-configuration-options-for-external-oidc.md?plain=1#L137-L161

https://github.com/openshift/hypershift/blob/f2fd2ca7bd92975d15fb0f3b284288d2b9e9c10f/control-plane-operator/controllers/hostedcontrolplane/kas/auth_types.go#L192-L193

No worries - I will adjust that. Thanks for spotting that! Learning experience for myself. :)

test/e2e/util/external_oidc.go
ehearne-redhat: Ah yes, this is because it is an optional field. :) Will update! 

https://github.com/openshift/hypershift/blob/f2fd2ca7bd92975d15fb0f3b284288d2b9e9c10f/control-plane-operator/controllers/hostedcontrolplane/kas/auth_types.go#L162-L163

test/e2e/external_oidc_test.go
ehearne-redhat: Yes, that makes sense. I have looked into how to do that. We can use `userInfo.extra["claims.email"]` and string parsing to ensure that the username matches what we expect i.e. if email is `foo@bar.net` then username should be `foo` .

control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth_test.go
everettraven: Doesn't have anything to due with the optionality of the `extra` field in the API. It is part of the standard Common Expression Language (CEL) behavior.

Here is the original CEL spec proposal for the behavior - https://github.com/google/cel-spec/wiki/proposal-246 :)

test/e2e/external_oidc_test.go
ehearne-redhat: I believe so - I'm looking at this script here. 

https://github.com/openshift/release/blob/main/ci-operator/step-registry/idp/external-oidc/keycloak/server/idp-external-oidc-keycloak-server-commands.sh#L189-L245

I'm thinking it wouldn't be a bad idea to have more reference to this script since there is this line that references a variable from this script.

https://github.com/openshift/hypershift/blob/main/test/e2e/util/external_oidc.go#L233-L234

That way we know exactly where this is all coming from. WDYT?

test/e2e/external_oidc_test.go
ehearne-redhat: OK, I see what you mean. You're saying the last test referenced here combined the first two tests referenced. Yeah, that makes sense. I'll remove the last one. If you want them both combined, let me know. :) 

There probably is a way to add negative test cases. However, I'm unsure how possible it is from the hypershift side because it all depends on what this script adds in terms of valid/invalid users. 

https://github.com/openshift/release/blob/main/ci-operator/step-registry/idp/external-oidc/keycloak/server/idp-external-oidc-keycloak-server-commands.sh

As it currently stands, it adds/configures just valid users/groups. From what I understand, we would need to either add a new test or modify this script to configure these invalid users/groups so we can properly test extensively. 

I'll push up what I have done currently, and look to go through new review comments tomorrow if any. :)

test/e2e/external_oidc_test.go
```diff
@@ -109,5 +110,63 @@ func TestExternalOIDC(t *testing.T) {
...
+			t.Run("[OCPFeatureGate:ExternalOIDCWithUpstreamParity] Test CEL username expression mapping", func(t *testing.T) {
+				g := NewWithT(t)
+				t.Logf("begin to test CEL username expression mapping")
+				g.Expect(selfSubjectReview.Status.UserInfo.Username).NotTo(BeEmpty())
+				// Username should be the email prefix (before @) since we configured expression: claims.email.split('@')[0]
+				// doesn't contain substring
+				g.Expect(selfSubjectReview.Status.UserInfo.Username).NotTo(ContainSubstring("@"))
+				// equals the actual email prefix
+				// e2eutil.ExternalOIDCExtraKeyFoo --> claims.email expression
+				emailValues := selfSubjectReview.Status.UserInfo.Extra[e2eutil.ExternalOIDCExtraKeyFoo]
```
ehearne-redhat: @everettraven do we know if this field is feature gate dependent? 

So if `ExternalOIDCWithUIDAndExtraClaimMappings` feature gate was disabled in the future what impact does it have on accessing this field for email verification?

test/e2e/external_oidc_test.go
```diff
@@ -109,5 +110,63 @@ func TestExternalOIDC(t *testing.T) {
...
+				g.Expect(emailValues).NotTo(BeEmpty())
+				email := emailValues[0]
+				expectedUserName := strings.Split(email, "@")[0]
+				g.Expect(selfSubjectReview.Status.UserInfo.Username).Should(Equal(expectedUserName))
```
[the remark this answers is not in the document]
ehearne-redhat: @coderabbitai could you show me how this is true? I had a similar concern but I couldn't find proof.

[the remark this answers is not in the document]
ehearne-redhat: @coderabbitai Would adding this field to ExternalOIDCWithUpstreamParity feature gate check be okay?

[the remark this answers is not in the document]
ehearne-redhat: @coderabbitai OK. Thank you - I think it is important to the reviewer that we can ensure the username evaluates properly. Unless there is another route we can take to ensure this outcome?

control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth_test.go
```diff
@@ -1521,6 +1521,1004 @@ func TestAdaptAuthConfig(t *testing.T) {
...
+							AudienceMatchPolicy: AudienceMatchPolicyMatchAny,
+							Audiences:           []string{"one", "two"},
+						},
+						ClaimMappings: ClaimMappings{
+							Username: PrefixedClaimOrExpression{
+								Expression: "claims.email.split('@')[0]",
+							},
+							Groups: PrefixedClaimOrExpression{
+								Expression: "claims.groups.orValue([]).filter(g, g.startsWith('ocp-'))",
+							},
```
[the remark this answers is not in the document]
ehearne-redhat: @coderabbitai As it currently stands, the tests are passing even without the optional notation. In a previous iteration it was noted that `claims.roles.?orValue` was not evaluating and failing because list --> `([])` was not of type `any` . The optional notation needed to be removed for the test to run and pass. 

I did look into the concern of CEL potentially erroring out because of this optional notation not being present. From what I can tell in the evaluation code, this should not happen. If `orValue` is not optional it will return the LHS (whatever it can find in the list) of the expression to evaluate instead. 

https://github.com/ehearne-redhat/hypershift/blob/external-oidc-test-featuregate-externaloidcwithupstreamparity/vendor/github.com/google/cel-go/cel/library.go#L761-L774

So, the expression will always evaluate successfully at current. 

However, if we upgrade in the future, it looks like future versions of CEL will error out because it doesn't use optional notation. 

https://github.com/google/cel-go/blob/master/cel/library.go#L763-L780

So it is still a problem we need to solve now to prevent it being an issue in the future.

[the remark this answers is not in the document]
ehearne-redhat: @coderabbitai I did try using `has()` and it also threw the `any` error as before. I am thinking it is best to remove these test cases because while as they are now will work, in the future it will cause errors, and they will likely be removed anyway. 

Do you have other suggestions? Are these specific test cases integral to this PR?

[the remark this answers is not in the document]
ehearne-redhat: @coderabbit awesome suggestion! Using `dyn([])` works perfectly locally. Looks like we don't need to remove these tests at all.

test/e2e/external_oidc_test.go
ehearne-redhat: Since I have ensured the field can accessible on either feature gate enablement this shouldn't be an issue. :)

test/e2e/util/external_oidc.go
```diff
@@ -149,6 +156,57 @@ func (config *ExtOIDCConfig) GetAuthenticationConfig() *configv1.AuthenticationS
...
+		// If not, we will need to add extra mapping to access email for username
+		// verification later.
+		if !featuregates.Gate().Enabled(featuregates.ExternalOIDCWithUIDAndExtraClaimMappings) {
+			authnSpec.OIDCProviders[0].ClaimMappings.Extra = append(authnSpec.OIDCProviders[0].ClaimMappings.Extra,
+				configv1.ExtraMapping{
+					Key:             ExternalOIDCExtraKeyFoo,
+					ValueExpression: ExternalOIDCExtraKeyFooValueExpression,
+				},
+			)
+		}
```
everettraven: I'm not sure I'm following why this step here is necessary? They are two separate features and this looks like it already happens above if the gate is enabled?

test/e2e/util/external_oidc.go
```diff
@@ -173,8 +231,11 @@ func ValidateAuthenticationSpec(t *testing.T, ctx context.Context, client crclie
 	actualAuth := hostedCluster.Spec.Configuration.Authentication
 	g.Expect(actualAuth.OIDCProviders).NotTo(BeEmpty())
 
-	if featuregates.Gate().Enabled(featuregates.ExternalOIDCWithUIDAndExtraClaimMappings) {
+	if featuregates.Gate().Enabled(featuregates.ExternalOIDCWithUpstreamParity) || featuregates.Gate().Enabled(featuregates.ExternalOIDCWithUIDAndExtraClaimMappings) {
```
everettraven: This also doesn't seem necessary? The feature we are adding tests for doesn't have any reliance on the extra claim mapping.

test/e2e/external_oidc_test.go
everettraven: That feature gate should already be enabled by default for a couple releases and therefore should never be disabled in the future. We actually need to remove it in the near future.

I don't think we need to check the `UserInfo.Extra` field as part of these tests though because that is functionality from an entirely different feature.

test/e2e/external_oidc_test.go
everettraven: Is there a more deterministic way for us to have a set of expected output data based on the input data?

Relying on a separate feature to test the output of this feature feels like a bit of an unnecessary dependency if we have a way to be more deterministic.

test/e2e/external_oidc_test.go
everettraven: I wonder if we should do away with the automatic setup of users and groups within Keycloak via that script?

Or at the very least have a deterministic set of users/groups that get created so that we can have more deterministic test behaviors?

Ideally, our test cases are self-sufficient and fully encompass the scope of what needs to be done.

Maybe this is better suited as a follow up though to prevent the scope of this work from getting blown up.

test/e2e/external_oidc_test.go
everettraven: > OK, I see what you mean. You're saying the last test referenced here combined the first two tests referenced. Yeah, that makes sense. I'll remove the last one. If you want them both combined, let me know. :)

What I was meaning is that the first 2 test cases:
- "[OCPFeatureGate:ExternalOIDCWithUpstreamParity] Test CEL username expression mapping"
- "[OCPFeatureGate:ExternalOIDCWithUpstreamParity] Test CEL groups expression mapping"

Already inherently test what these last 2 tests are testing because you configure the claim validation and user validation rules for all the test cases.

I'm not sure it makes sense for us to retest this behavior.

>There probably is a way to add negative test cases. However, I'm unsure how possible it is from the hypershift side because it all depends on what this script adds in terms of valid/invalid users.
>
>https://github.com/openshift/release/blob/main/ci-operator/step-registry/idp/external-oidc/keycloak/server/idp-external-oidc-keycloak-server-commands.sh
>
>As it currently stands, it adds/configures just valid users/groups. From what I understand, we would need to either add a new test or modify this script to configure these invalid users/groups so we can properly test extensively.

Okay, I've gone a bit back and forth on this throughout this review a touch already but now I'm thinking that if that script is presenting a limitation in what we can or cannot test I think modifying it as necessary is important to ensure we are getting the right level of test coverage here.

Because of this, we should include any modifications that need to happen there within scope of the testing work for HyperShift.

test/e2e/external_oidc_test.go
ehearne-redhat: I did try to ponder on how we could do this. I wanted a different way to this but this was the one that jumped out to me. It was the only way I could see to ensure an absolute verification of the username. 

I will investigate further and see.

test/e2e/external_oidc_test.go
ehearne-redhat: We would need to look into it, definitely. From what I could tell, the users/groups had to be deployed in this way because of the nature of HyperShift itself. We can't do CRUD for the users/groups here because Keycloak is run externally here - we rely on it being set up alongside HyperShift, and it would be different to us having access to a cluster to just spin up keycloak/modify on the fly in our origin test suites. 

I think realistically we would need to modify that script so the environment can match our expectations. However, it would also need to match the other test expectations too since these changes would impact those tests too. 

I think we should definitely aim to figure that out sooner rather than later before we have too many tests that "commit" or conform to the script expectations.

test/e2e/external_oidc_test.go
ehearne-redhat: Oh OK - noted. :)

I'll look into how exactly this would work so that our tests and the other tests using this script wouldn't exactly need to be changed. 

It could be the case where we have an additional script but then we set the expectation of needing new script every time we test hypershift which would add burden. It might make sense to make the necessary changes to both tests if needed now so the burden isn't heavy later (we would hope :D) . 

I'll look into it shortly.

test/e2e/external_oidc_test.go
ehearne-redhat: Ah OK - thanks for clarifying! In that case we would need to use another mechanism as you rightfully pointed out in https://github.com/openshift/hypershift/pull/8287#discussion_r3147357501 .

test/e2e/util/external_oidc.go
ehearne-redhat: Yep - makes sense. I just needed that clarity from https://github.com/openshift/hypershift/pull/8287#discussion_r3147342122 . I'll work on removing that.

test/e2e/external_oidc_test.go
everettraven: >We can't do CRUD for the users/groups here because Keycloak is run externally here - we rely on it being set up alongside HyperShift, and it would be different to us having access to a cluster to just spin up keycloak/modify on the fly in our origin test suites

The running Keycloak instance has to be accessible to the HostedCluster that is created. Makes me think it would inherently be accessible to us as well. We would just need to make sure we know where we need to make the requests to configure it on the fly.

I wonder if we could even go so far as deploying Keycloak on the fly to have full control in the test setup like we do in the standalone openshift tests.

ehearne-redhat: I think you're right - we have access to admin creds from the script in `keycloak` ns. 

https://github.com/openshift/release/blob/main/ci-operator/step-registry/idp/external-oidc/keycloak/server/idp-external-oidc-keycloak-server-commands.sh#L30 

With these creds we could control the users/groups by making api calls via `ExtOIDCConfig.IssuerURL` . 

And on that same thread, we could actually deploy Keycloak here. We probably have everything we need. I could convert the script deployment into go code here. 

This way we would have full control over testing username concretely. 

I'll work on a potential way of doing this and re-ping when ready. Thanks for thinking this one out! :)

ehearne-redhat:
Will rebase + address commit lint errors once I get feedback from `e2e-aks` and `e2e-aws` about modifying keycloak instance's user/groups via their API.

ehearne-redhat:
/payload-job periodic-ci-openshift-hypershift-release-4.23-periodics-e2e-aws-external-oidc

ehearne-redhat:
/payload-job periodic-ci-openshift-hypershift-release-4.23-periodics-e2e-aws-external-oidc-techpreview

ehearne-redhat:
test e2e-azure-aks-external-oidc-techpreview
/test e2e-aws-external-oidc-techpreview

control-plane-operator/featuregates/featuregates.go
```diff
@@ -18,8 +18,8 @@ const (
 var (
 	allFeatures = featuregates.NewFeatureSetAwareFeatures()
 
-	externalOIDCWithUIDAndExtraClaimMappingsFeature = featuregates.NewFeature(ExternalOIDCWithUIDAndExtraClaimMappings, featuregates.WithEnableForFeatureSets(configv1.TechPreviewNoUpgrade, configv1.Default))
-	externalOIDCWithUpstreamParityFeature           = featuregates.NewFeature(ExternalOIDCWithUpstreamParity, featuregates.WithEnableForFeatureSets(configv1.TechPreviewNoUpgrade))
+	externalOIDCWithUIDAndExtraClaimMappingsFeature = featuregates.NewFeature(ExternalOIDCWithUIDAndExtraClaimMappings, featuregates.WithEnableForFeatureSets(configv1.TechPreviewNoUpgrade, configv1.DevPreviewNoUpgrade, configv1.Default))
+	externalOIDCWithUpstreamParityFeature           = featuregates.NewFeature(ExternalOIDCWithUpstreamParity, featuregates.WithEnableForFeatureSets(configv1.TechPreviewNoUpgrade, configv1.DevPreviewNoUpgrade))
```
everettraven: No need to add `DevPreviewNoUpgrade` here. HyperShift has no notion of a `DevPreviewNoUpgrade` feature set for its feature gates (i.e there would be no way for an end user to reasonably enable this using the installation mechanisms that exist today AFAIK)

hypershift-operator/featuregate/feature.go
```diff
@@ -34,6 +34,10 @@ const (
 	// alpha: v0.1.49
 	// beta: x.y.z
 	HCPEtcdBackup featuregate.Feature = "HCPEtcdBackup"
+
+	// ExternalOIDC Feature Gates
+	ExternalOIDCWithUIDAndExtraClaimMappings featuregate.Feature = "ExternalOIDCWithUIDAndExtraClaimMappings"
+	ExternalOIDCWithUpstreamParity           featuregate.Feature = "ExternalOIDCWithUpstreamParity"
```
everettraven: I don't think this is quite right. Adding the feature gates here doesn't mean that they inherently get picked up by the validation behavior that is failing because that validation behavior relies on the control plane operator (CPO) feature gates package.

Add the same gates to a different package doesn't influence that at all.

Instead, we need to initialize the CPO feature gates package alongside the hypershift operator feature gates package initialization in https://github.com/openshift/hypershift/blob/1eddaf8e5e593b440723b6166c88d6220850173c/hypershift-operator/main.go#L196-L206

This would probably look something like:
```go
	// Attempt to determine featureset prior to adding featuregate flags.
	// It is safe to get the empty string from this as the empty string is the default featureset.
	// This should _generally_ be safe to do here because any unknown featureset provided in the environment
	// variable should result in using the default featureset.
	featureSet := os.Getenv("HYPERSHIFT_FEATURESET")
	// Doing this call first should explicitly configure the default gate state
	// for the provided featureset. This makes it so that specifying feature gates
	// via the `--feature-gates` flag will always override whatever has been configured
	// by the featureset.
	featuregate.ConfigureFeatureSet(featureSet)
	featuregate.Gate().AddFlag(cmd.Flags())

   // NEW
   // 'cpofeaturegate' is an import alias to the control plane operator featuregate package
   cpofeaturegate.ConfigureFeatureSet(featureSet)
```

test/e2e/util/external_oidc.go
```diff
@@ -347,3 +415,917 @@ func GetClientConfigForKeycloakOIDCUser(clientCfg *rest.Config, authConfig *ExtO
...
 }
+
+// KeycloakAdminClient provides methods to interact with Keycloak Admin REST API
+type KeycloakAdminClient struct {
+	BaseURL    string
+	AdminToken string
+	HTTPClient *http.Client
+	AdminUser  string
+	AdminPass  string
+	CACertFile string
```
ehearne-redhat: Check if we actually need this ca cert because we don't use TLS when we specify via the API (we skip the check).

test/e2e/util/external_oidc.go
```diff
@@ -347,3 +415,917 @@ func GetClientConfigForKeycloakOIDCUser(clientCfg *rest.Config, authConfig *ExtO
 
 	return userClientConfig
 }
+
+// KeycloakAdminClient provides methods to interact with Keycloak Admin REST API
+type KeycloakAdminClient struct {
```
ehearne-redhat: move the keycloak api related code to a separate file to keep things clean

test/e2e/util/external_oidc.go
```diff
@@ -347,3 +415,917 @@ func GetClientConfigForKeycloakOIDCUser(clientCfg *rest.Config, authConfig *ExtO
...
+// TestResources tracks resources created during a test for cleanup
+type TestResources struct {
+	AdminClient *KeycloakAdminClient
+	UserIDs     []string
+	GroupIDs    []string
+	UserCreds   map[string]string // username -> password
+}
+
+// NewTestResources creates a new TestResources tracker
+func NewTestResources(adminClient *KeycloakAdminClient) *TestResources {
```
ehearne-redhat: add a check to see if admin token expires and if so, refresh?

test/e2e/external_oidc_test.go
```diff
@@ -109,5 +117,258 @@ func TestExternalOIDC(t *testing.T) {
...
+
+				// Verify: Username is email prefix (before @)
+				expectedUsername := strings.Split(testUser.Email, "@")[0]
+				g.Expect(testUser.SelfSubjectReview.Status.UserInfo.Username).Should(Equal(expectedUsername),
+					"username should be email prefix from CEL expression: claims.email.split('@')[0]")
+				g.Expect(testUser.SelfSubjectReview.Status.UserInfo.Username).NotTo(ContainSubstring("@"),
+					"username should not contain @ symbol")
+				t.Logf("CEL username expression correctly mapped '%s' to '%s'", testUser.Email, testUser.SelfSubjectReview.Status.UserInfo.Username)
+
+				// Edge case test: preferred_username vs email-derived username mismatch
```
ehearne-redhat: This part of the test - we can remove because it doesn't test our feature gate configurations. We are mostly concerned with happy mappings here and negative testing can be done in validations testing.

test/e2e/external_oidc_test.go
```diff
@@ -109,5 +117,258 @@ func TestExternalOIDC(t *testing.T) {
...
+				t.Logf("Created user with email='%s', mapped username will be '%s' (ID: %s)", forbiddenEmail, forbiddenUsername, forbiddenUserID)
+
+				// Try to authenticate - should FAIL due to user validation rule
+				err = testResources.TryAuthenticateUser(ctx, t, forbiddenUsername, forbiddenPassword, clientCfg, clusterOpts.ExtOIDCConfig)
+				g.Expect(err).To(HaveOccurred(), "authentication must fail when username contains 'forbidden'")
+				g.Expect(err.Error()).Should(ContainSubstring("Unauthorized"),
+					"forbidden user cannot authenticate as it violates user validation rule")
+				t.Logf("✓ User with 'forbidden' in username correctly rejected with error: %v", err)
+
+				// NOTE: We cannot test the negative case for the system: prefix rule via Keycloak
```
ehearne-redhat: We can modify the config in `external_oidc.go` to have a rule that CAN map to `system:` for username regardless of email e.g. for a specific email keyword.

test/e2e/util/external_oidc.go
```diff
@@ -48,6 +50,14 @@ const (
...
+	// CEL expressions for claim validation rules
+	ClaimValidationExprEmailExists   = "has(claims.email) && claims.email != ''"
+	ClaimValidationExprEmailVerified = "claims.email_verified == true"
+
+	// CEL expressions for user validation rules
+	UserValidationExprNoSystemPrefix  = "!user.username.startsWith('system:')"
+	UserValidationExprNoForbiddenWord = "!user.username.contains('forbidden')"
 )
 
 type ExtOIDCConfig struct {
```
ehearne-redhat, line 52: We need to look at having a bespoke config per test so this won't fail when the feature gate graduates to default.

everettraven:
I think we will also need to do something similar to what we did in openshift/origin related to having a unique authentication configuration to be used for each test case so that we can control what the inputs are to every test case. Here is a link to the function we implemented to do this in o/origin: https://github.com/openshift/origin/blob/428b9a0e36ff2c43b2a956b12d22141a1c88b950/test/extended/authentication/oidc.go#L771 

This might end up being a case where we will have to spin up multiple hypershift clusters, one for each configuration we would like to test. It looks like there is a way to get that done via modification of the cluster options passed into the `NewHypershiftTest().Execute()` method.

test/e2e/external_oidc_test.go
ehearne-redhat: actually fine to keep

test/e2e/external_oidc_test.go
ehearne-redhat: Tie this in with https://github.com/openshift/hypershift/pull/8287#discussion_r3274745419 as - if we can use bespoke config per test mentioned in https://github.com/openshift/hypershift/pull/8287#issuecomment-4499555986, we should be able to make this happen that way .

test/e2e/util/hypershift_framework.go
```diff
@@ -483,7 +483,11 @@ func (h *hypershiftTest) createHostedCluster(opts *PlatformAgnosticOptions, plat
 					if v.Spec.Configuration == nil {
 						v.Spec.Configuration = &hyperv1.ClusterConfiguration{}
 					}
-					v.Spec.Configuration.Authentication = opts.ExtOIDCConfig.GetAuthenticationConfig()
```
everettraven, line 486: Could you just have the `GetAuthenticationConfig()` method return the right one?

test/e2e/util/external_oidc.go
```diff
@@ -67,6 +75,9 @@ type ExtOIDCConfig struct {
 	// for oidcProviders.issuer.issuerCertificateAuthority
 	IssuerCAConfigmapName string
 	IssuerCABundleFile    string
+
+	// custom field for adding custom auth configuration
+	CustomAuthSpec *configv1.AuthenticationSpec
```
everettraven: We probably don't want to have a full override here where we have to define the entire spec including issuers, clients, etc. We just want the ability to modify a given baseline configuration for each test case.

test/e2e/external_oidc_test.go
ehearne-redhat: If current modifications pass, we will add this alongside migrating tests over to bespoke custom auth spec pattern.

hypershift-operator/main_test.go
```diff
@@ -0,0 +1,24 @@
+package main
+
+import (
+	"testing"
+
+	"github.com/spf13/cobra"
+)
+
+// When NewStartCommand is called, it should return a valid cobra command
+func TestNewStartCommand(t *testing.T) {
```
everettraven: Any particular reason to add this test?

test/e2e/util/external_oidc.go
```diff
@@ -156,9 +168,99 @@ func (config *ExtOIDCConfig) GetAuthenticationConfig() *configv1.AuthenticationS
...
+			{
+				Expression: UserValidationExprNoSystemPrefix,
+				Message:    "username cannot use reserved system: prefix",
+			},
+			{
+				Expression: UserValidationExprNoForbiddenWord,
+				Message:    "username cannot contain the word 'forbidden'",
+			},
+		}
+	}
```
everettraven, line 159: Now that we have a way to customize a generic baseline configuration, it would make more sense for these feature-gated configuration adjustments to be done as part of the already feature-gated test cases that we run.

This would help prevent confusion whenever looking at the test case implementations and having to remember/cross-check the default configurations defined here for different scenarios. Especially for the upstream parity case where it will always overwrite the default username configuration once it is promoted to default (i.e always enabled).

test/e2e/util/external_oidc.go
everettraven, line 139: More of a general thought - I'll leave it up to you if you'd like to take any action here - what if we made every test case explicitly define the claim mappings it cares about so we never have to do "overrides" here and this becomes a pure "stamp out the always common things" utility?

test/e2e/util/keycloak.go
```diff
@@ -0,0 +1,1041 @@
...
+	// However, Keycloak credentials are stored differently on both.
+
+	// On AWS, both admin username and password credentials are stored
+	// via a StatefulSet called 'keycloak' in the 'keycloak' namespace.
+
+	// On AKS, the admin username is stored in a config map called
+	// 'keycloak-env-vars' in 'keycloak' namespace via data.KC_BOOTSTAP_ADMIN_USERNAME,
+	// and the admin password is stored in a secret called 'keycloak'
+	// in the 'keycloak' namespace via data.admin-password .
+	// https://github.com/bitnami/charts/tree/main/bitnami/keycloak/templates
```
everettraven, line 226: This is outside of the scope of the current PR, but I think we should create a tech debt ticket to track unifying this into a single approach for deploying Keycloak across all platforms.

I don't know why we did this in the first place, but it seems unnecessary and is only going to cause confusion and drift in the future.

Part of that tech debt ticket should be to update this test logic to use the one approach we land on.

test/e2e/external_oidc_test.go
```diff
@@ -57,57 +102,108 @@ func TestExternalOIDC(t *testing.T) {
...
+		// Auth config uses: Static claim-based username/groups WITH prefixes (legacy behavior)
+
+		// Legacy prefixed mappings test only runs when ExternalOIDCWithUpstreamParity is NOT enabled
+		// Config: Username/Groups use static claims with prefixes (claim: "email", prefix: "prefix-")
+		// When ExternalOIDCWithUpstreamParity IS enabled, it overwrites with CEL expressions (no prefixes)
+		if !featuregates.Gate().Enabled(featuregates.ExternalOIDCWithUpstreamParity) {
+			t.Run("[OCPFeatureGate:ExternalOIDCWithUIDAndExtraClaimMappings] Test external OIDC userInfo username and groups", func(t *testing.T) {
+				e2eutil.TestLegacyPrefixedMappings(t, ctx, hostedCluster, kc, clientCfg, clusterOpts.ExtOIDCConfig, selfSubjectReview)
 			})
+		}
```
everettraven: This looks like a prime example of the confusing logic associated with having default overrides based on feature gate state.

This test case is still a valid test case that should run even when the ExternalOIDCWithUpstreamParity feature gate is enabled.

It has a different configuration it was intended to test and we should ensure that intended configuration is in place for this test instead of skipping it when the ExternalOIDCWithUpstreamParity feature gate is enabled.

test/e2e/util/external_oidc_test_helpers.go
everettraven, line 1: These helper test functions introduce a lot of indirection here that is making the test cases pretty difficult to follow.

Is there a particular reason why you felt we needed these helpers instead of having each test case implement the sub-tests it cares about?

test/e2e/external_oidc_test.go
```diff
@@ -57,57 +102,108 @@ func TestExternalOIDC(t *testing.T) {
...
+				t.Skipf("Could not setup Keycloak admin client: %v", err)
+			}
+
+			t.Run("[OCPFeatureGate:ExternalOIDCWithUpstreamParity] Test user validation rules (custom UID config)", func(t *testing.T) {
+				// This test works because validation rules are preserved from feature gate config
+				// Only UID and Extra mappings were customized
+				e2eutil.TestUserValidationRules(t, ctx, hostedCluster, kc, clientCfg, customClusterOpts.ExtOIDCConfig)
+			})
+		}).Execute(&customClusterOpts, globalOpts.Platform, globalOpts.ArtifactDir, "ext-oidc-uid-claim", globalOpts.ServiceAccountSigningKey)
+	}
```
everettraven: This looks like the closest thing to what I envisioned being done for _all_ the existing test cases.

Basically, do this for a configuration we want to test and then run all the checks we'd like to have for that configuration as separate `t.Run(...)` entries.

test/e2e/util/keycloak.go
```diff
@@ -0,0 +1,1041 @@
...
+	return kc, nil
+}
+
+// TestResources tracks resources created during a test for cleanup
+type TestResources struct {
+	AdminClient *KeycloakAdminClient
+	UserIDs     []string
+	GroupIDs    []string
+	UserCreds   map[string]string // username -> password
+}
```
everettraven: Is there a particular reason that we need this abstraction over just using the standard keycloak client functionality you've implemented above?

test/e2e/util/keycloak.go
```diff
@@ -0,0 +1,1041 @@
...
+	if resp.StatusCode != http.StatusCreated {
+		body, _ := io.ReadAll(resp.Body)
+		return fmt.Errorf("failed to create mapper, status: %d, body: %s", resp.StatusCode, string(body))
+	}
+
+	return nil
+}
+
+// SetupKeycloakTestEnvironment creates test users, groups, and protocol mappers
+func SetupKeycloakTestEnvironment(t *testing.T, ctx context.Context, config *ExtOIDCConfig, adminUser, adminPass string, numUsers int) (string, error) {
```
everettraven: Looks like this is now unused - remove it?

test/e2e/util/keycloak.go
```diff
@@ -0,0 +1,1041 @@
...
+	Username          string
+	Email             string
+	Password          string
+	UserID            string
+	GroupID           string
+	GroupName         string
+	KubeConfig        *rest.Config
+	AuthClient        kauthnv1typedclient.AuthenticationV1Interface
+	SelfSubjectReview *kauthnv1.SelfSubjectReview
+}
```
everettraven: Could you explain why this is necessary as well? I'm having a hard time following why this is necessary for the tests.

hypershift-operator/main_test.go
ehearne-redhat: I added this test to help codecov/patch pass, as it was detecting minuscule code coverage change because of the change in `main.go` where we needed to wire up `cpofeaturegate`. I can remove if it is a blocker. Probably won't stop approvers from (potentially) approving.

test/e2e/util/external_oidc.go
ehearne-redhat: Ah OK I see what you mean, that makes sense.

test/e2e/util/external_oidc.go
ehearne-redhat: Makes sense to me.

test/e2e/util/external_oidc_test_helpers.go
ehearne-redhat: I don't think there was a particular reason. I can work on removing this.

test/e2e/util/keycloak.go
ehearne-redhat: I will work on removing that.

test/e2e/util/keycloak.go
ehearne-redhat: It gives us all the user info and whether they are validated. It's a handy way to have all the info we need. It is used in `TestCELUsernameMapping()` and the other helpers which I will move back to prevent confusion. I believe I originally used this as a way to ensure users are validated because of edge cases I was working on at the time. 

We can look at streamlining it or moving tests to using simpler structs.

test/e2e/util/keycloak.go
ehearne-redhat: I believe the reason why it was done like this was because the HyperShift platform meant we needed a different approach to instantiating these users/groups. I believe it was assumed easier (and likely was) to run a script to instantiate into the keycloak instance for testing. However, as we saw, it left us in a position where we couldn't appropriately test the platform as extensively as we wanted. This approach is arguably better for us. :)

test/e2e/util/keycloak.go
ehearne-redhat: Actually - as it currently stands, we do use this for cleanup of resources after tests.

test/e2e/util/keycloak.go
ehearne-redhat: Re-reading this, you probably meant using two different cloud providers, which also makes sense.

everettraven: Yeah, we should probably have the same general deployment mechanism for keycloak - and thus the same admin config discovery process - across all cloud providers we decide to test against.

hypershift-operator/main_test.go
everettraven: It isn't really testing anything meaningful.

I don't think we should be adding unnecessary tests just to pass a code coverage check. I'd suggest either removing this test or making it a meaningful test. I don't know the hypershift folks' opinions on the code coverage check, but I'd push for an override if it is failing on this specifically.

test/e2e/util/keycloak.go
everettraven: What resources need to be cleaned up?

test/e2e/external_oidc_test.go
```diff
@@ -13,16 +12,158 @@ import (
...
 	"github.com/openshift/hypershift/control-plane-operator/featuregates"
 )
 
+// createTestUserWithGroup creates a Keycloak user with a group and password, returning the username, email, password, and group name.
+func createTestUserWithGroup(ctx context.Context, g Gomega, kc *e2eutil.KeycloakAdminClient, emailVerified bool) (username, email, password, groupName string) {
+	// create group
+	groupName = "cel-test-group" + "-" + e2eutil.GenerateRandomPassword(8)
+
+	// create user
+	username = "cel-test-user" + "-" + e2eutil.GenerateRandomPassword(8)
```
everettraven: Could probably just fully generate these at random instead of using a prefix?

test/e2e/external_oidc_test.go
```diff
@@ -39,75 +180,135 @@ func TestExternalOIDC(t *testing.T) {
 	clusterOpts.NodePoolReplicas = 1
 	clusterOpts.FeatureSet = string(configv1.Default)
 
-	if os.Getenv("TECH_PREVIEW_NO_UPGRADE") == "true" {
-		clusterOpts.FeatureSet = string(configv1.TechPreviewNoUpgrade)
-	}
```
everettraven, line 44: Why do we no longer need this?

test/e2e/external_oidc_test.go
```diff
@@ -39,75 +180,135 @@ func TestExternalOIDC(t *testing.T) {
...
+		if err != nil {
+			t.Skipf("Could not setup Keycloak admin client: %v", err)
+		}
+		// ExternalOIDCWithUIDAndExtraClaimMappings has graduated to Default feature set
+		// Auth config includes: UID expression + Extra claim mappings
+		// Auth config uses: Static claim-based username/groups WITH prefixes (legacy behavior)
+
+		// Legacy prefixed mappings test only runs when ExternalOIDCWithUpstreamParity is NOT enabled
+		// Config: Username/Groups use static claims with prefixes (claim: "email", prefix: "prefix-")
+		// When ExternalOIDCWithUpstreamParity IS enabled, it overwrites with CEL expressions (no prefixes)
```
everettraven: Stale comment?

test/e2e/external_oidc_test.go
```diff
@@ -39,75 +180,135 @@ func TestExternalOIDC(t *testing.T) {
...
 			g.Expect(hostedCluster.Spec.Configuration).NotTo(BeNil())
 			g.Expect(hostedCluster.Spec.Configuration.Authentication).NotTo(BeNil())
 			g.Expect(hostedCluster.Spec.Configuration.Authentication.OIDCProviders).NotTo(BeEmpty())
 			clientCfg := e2eutil.WaitForGuestRestConfig(t, ctx, mgtClient, hostedCluster)
-			e2eutil.ChangeClientForKeycloakExtOIDC(t, ctx, clientCfg, clusterOpts.ExtOIDCConfig)
-			t.Logf("successfully get oidc user client")
-		})
 
-		if featuregates.Gate().Enabled(featuregates.ExternalOIDCWithUIDAndExtraClaimMappings) {
-			t.Run("[OCPFeatureGate:ExternalOIDCWithUIDAndExtraClaimMappings] Test external OIDC userInfo username", func(t *testing.T) {
```
everettraven: As long as the feature gate exists, we should probably keep this gated.

Once we remove the feature gate we can remove the gating here.

test/e2e/external_oidc_test.go
```diff
@@ -39,75 +180,135 @@ func TestExternalOIDC(t *testing.T) {
...
+			g.Expect(selfSubjectReview.Status.UserInfo.UID).Should(ContainSubstring(e2eutil.ExternalOIDCUIDExpressionPrefix))
+			g.Expect(selfSubjectReview.Status.UserInfo.UID).Should(ContainSubstring(e2eutil.ExternalOIDCUIDExpressionSubfix))
+
+			t.Logf("begin to test external OIDC userInfo Extra")
+			g.Expect(selfSubjectReview.Status.UserInfo.Extra).NotTo(BeEmpty())
+			g.Expect(selfSubjectReview.Status.UserInfo.Extra).Should(HaveKey(e2eutil.ExternalOIDCExtraKeyBar))
+			g.Expect(selfSubjectReview.Status.UserInfo.Extra).Should(HaveKey(e2eutil.ExternalOIDCExtraKeyFoo))
+		})
+
+		t.Run("[OCPFeatureGate:ExternalOIDCWithUIDAndExtraClaimMappings] Test external OIDC: check cluster operator status using oauth client", func(t *testing.T) {
```
everettraven: For existing test cases, lets avoid renames for the time being. I don't recall whether or not there is any impacts to component readiness or other test tracking when renames happen like this. I'm sure there is probably a process for test name changes that I'd like to make sure we follow if we want to pursue renaming (looks like it is https://github.com/openshift-eng/ci-test-mapping#renaming-tests).

Let's track any test case renaming and consolidation you think is important here as part of the tech debt items we want to track.

test/e2e/external_oidc_test.go
```diff
@@ -13,16 +12,158 @@ import (
...
+				Temporary: false,
+			},
+		},
+	}
+
+	_, err := kc.CreateUser(ctx, user)
+	g.Expect(err).NotTo(HaveOccurred(), "should not error creating a user")
+
+	return username, email, password, groupName
+}
```
everettraven: Could probably consolidate these into a singular function and have the same utility like:
```go
// createTestUserWithGroupAndCustomUsername creates a Keycloak user with a custom username pattern, group, and password.
func createTestUserWithGroup(ctx context.Context, kc *e2eutil.KeycloakAdminClient, usernamePrefix string, emailVerified bool) (string, string, string, string, error) {
	// create group
	groupName := e2eutil.GenerateRandomPassword(8)

	// create user with custom username
	username := e2eutil.GenerateRandomPassword(8)
    if len(usernamePrefix) > 0 {
        username = usernamePrefix + username
    } 
	email := username + "@test.example.com"

	password := e2eutil.GenerateRandomPassword(16)

	user := e2eutil.KeycloakUser{
		Username:      username,
		Enabled:       true,
		FirstName:     username,
		LastName:      "Test",
		Email:         email,
		EmailVerified: emailVerified,
		Groups:        []string{groupName},
		Credentials: []e2eutil.KeycloakCredential{
			{
				Type:      "password",
				Value:     password,
				Temporary: false,
			},
		},
	}

	_, err := kc.CreateUser(ctx, user)
	if err != nil {
        return "", "", "", "", err
    } 

	return username, email, password, groupName, nil
}
```

A couple things worth noting here that I changed as part of this suggestion:
- Just return an error and handle it where it is called, no need to pass the Gomega object as a parameter.
- Don't use named return values unless absolutely necessary. They can be problematic for various reasons and aren't idiomatic Go.

test/e2e/external_oidc_test.go
```diff
@@ -13,16 +12,158 @@ import (
...
+				Value:     password,
+				Temporary: false,
+			},
+		},
+	}
+
+	_, err := kc.CreateUser(ctx, user)
+	g.Expect(err).NotTo(HaveOccurred(), "should not error creating a user")
+
+	return username, email, password, groupName
```
everettraven: For both this one and the one above, it looks like the `email` return value is always ignored - we can probably just drop this from the set of return values.

test/e2e/external_oidc_test.go
```diff
@@ -13,16 +12,158 @@ import (
...
+//	UserValidationRules:   no system: prefix, no 'forbidden' word
+//
+// Execution 3 (TechPreviewNoUpgrade + AuthConfigWithUpstreamParity + customizations):
+//
+//	Username: CEL expression (preserved from upstream parity)
+//	Groups:   CEL expression (preserved from upstream parity)
+//	UID:      Claim "sub" (OVERRIDDEN by AuthConfigUIDFromClaim)
+//	Extra:    1 mapping (REPLACED by AuthConfigMinimalExtraMappings)
+//	ClaimValidationRules:  (preserved from upstream parity)
+//	UserValidationRules:   (preserved from upstream parity)
```
everettraven: Is this comment still up to date?

Also, is it really necessary? My hope is that after the most recent set of changes, the tests themselves are fairly straightforward to read and understand what they are doing.

test/e2e/external_oidc_test.go
```diff
@@ -39,75 +180,135 @@ func TestExternalOIDC(t *testing.T) {
...
+		featuregates.ConfigureFeatureSet(upstreamParityOpts.FeatureSet)
+
+		e2eutil.NewHypershiftTest(t, ctx, func(t *testing.T, g Gomega, mgtClient crclient.Client, hostedCluster *hyperv1.HostedCluster) {
 			g.Expect(hostedCluster.Spec.Configuration).NotTo(BeNil())
 			g.Expect(hostedCluster.Spec.Configuration.Authentication).NotTo(BeNil())
 			g.Expect(hostedCluster.Spec.Configuration.Authentication.OIDCProviders).NotTo(BeEmpty())
 			clientCfg := e2eutil.WaitForGuestRestConfig(t, ctx, mgtClient, hostedCluster)
-			e2eutil.ChangeClientForKeycloakExtOIDC(t, ctx, clientCfg, clusterOpts.ExtOIDCConfig)
-			t.Logf("successfully get oidc user client")
-		})
```
everettraven, line 70: Similar to my other comment related to test renames, let's avoid removing this one for the time being.

If we feel this isn't a good test to carry moving forward, lets track it in the set of tech debt and follow https://github.com/openshift-eng/ci-test-mapping#removing-tests to remove it.

test/e2e/external_oidc_test.go
```diff
@@ -13,16 +12,158 @@ import (
...
+
+// verifySelfSubjectReview creates a SelfSubjectReview and validates the username and groups are correct.
+func verifySelfSubjectReview(ctx context.Context, g Gomega, testAuthClient *kauthnv1typedclient.AuthenticationV1Client, username, groupName string) *kauthnv1.SelfSubjectReview {
+	selfSubjectReview, err := testAuthClient.SelfSubjectReviews().Create(ctx, &kauthnv1.SelfSubjectReview{}, metav1.CreateOptions{})
+	g.Expect(err).NotTo(HaveOccurred(), "user should be authenticated + able to do SelfSubjectReview")
+	g.Expect(selfSubjectReview.Status.UserInfo.Username).To(Equal(username), "username should be mapped correctly")
+	g.Expect(selfSubjectReview.Status.UserInfo.Groups).To(HaveLen(2), "user should have groups system:authenticated and IdP group")
+	g.Expect(selfSubjectReview.Status.UserInfo.Groups).To(ContainElements("system:authenticated", groupName))
+	return selfSubjectReview
+}
```
everettraven: Any particular reason we need to abstract this into a function?

It looks like only one of the new tests use this and some older tests were updated to use it.

I'd prefer if these checks were done in the test definitions themselves so they can be modified for individual test cases as needed without having to modify it for all test cases.

test/e2e/util/external_oidc.go
```diff
@@ -139,24 +151,124 @@ func (config *ExtOIDCConfig) GetAuthenticationConfig() *configv1.AuthenticationS
 		},
 	}
 
-	if featuregates.Gate().Enabled(featuregates.ExternalOIDCWithUIDAndExtraClaimMappings) {
-		authnSpec.OIDCProviders[0].ClaimMappings.UID = &configv1.TokenClaimOrExpressionMapping{
-			Expression: fmt.Sprintf(`"%s" + claims.sub + "%s"`, ExternalOIDCUIDExpressionPrefix, ExternalOIDCUIDExpressionSubfix),
+	// Now that ExternalOIDCWithUIDAndExtraClaimMappings is promoted to default feature set,
+	// we can add to config with no feature gate checks.
```
everettraven: In the interest of leaving pre-existing things that need some TLC alone, let's keep this gated for the time being and track feature gate cleanup as part of tech debt.

test/e2e/util/external_oidc.go
```diff
@@ -139,24 +151,124 @@ func (config *ExtOIDCConfig) GetAuthenticationConfig() *configv1.AuthenticationS
...
+// When it should combine multiple auth spec customizations
+func AuthConfigCombined(customizers ...func(*configv1.AuthenticationSpec)) func(*configv1.AuthenticationSpec) {
+	return func(spec *configv1.AuthenticationSpec) {
+		for _, customize := range customizers {
+			if customize != nil {
+				customize(spec)
+			}
+		}
+	}
 }
```
everettraven: These all look like they are unused.

test/e2e/util/external_oidc.go
```diff
@@ -182,6 +294,9 @@ func ValidateAuthenticationSpec(t testing.TB, ctx context.Context, client crclie
 
 	if featuregates.Gate().Enabled(featuregates.ExternalOIDCWithUIDAndExtraClaimMappings) {
 		g.Expect(actualAuth.OIDCProviders[0].ClaimMappings.Extra).NotTo(BeEmpty())
+	}
+
+	if featuregates.Gate().Enabled(featuregates.ExternalOIDCWithUIDAndExtraClaimMappings) {
```
everettraven: Any particular reason for this change?

test/e2e/util/external_oidc.go
```diff
@@ -336,7 +451,7 @@ func GetClientConfigForKeycloakOIDCUser(clientCfg *rest.Config, authConfig *ExtO
 		"--issuer-url=" + authConfig.IssuerURL,
 		"--client-id=" + authConfig.CliClientID,
 		"--extra-scopes=email,profile",
-		"--callback-address=127.0.0.1:8080",
+		"--callback-address=127.0.0.1:0",
```
everettraven: Why did this need to be changed?

test/e2e/util/external_oidc.go
```diff
@@ -139,24 +151,124 @@ func (config *ExtOIDCConfig) GetAuthenticationConfig() *configv1.AuthenticationS
...
+
+// AuthConfigWithUpstreamParity applies ExternalOIDCWithUpstreamParity feature gate configuration.
+// This includes:
+//   - CEL expressions for username/groups (no prefixes)
+//   - Claim validation rules (email exists, email_verified)
+//   - User validation rules (no system: prefix, no 'forbidden' word)
+//
+// NOTE: When using AuthConfigCombined(), this should be applied FIRST, then other customizations
+// that override specific fields (e.g., AuthConfigUIDFromClaim(), AuthConfigMinimalExtraMappings()).
+func AuthConfigWithUpstreamParity() func(*configv1.AuthenticationSpec) {
```
everettraven: Looks like this is used once. Should we just define the function we return here where we currently use this and remove this utility function?

test/e2e/util/external_oidc.go
```diff
@@ -49,6 +49,14 @@ const (
...
 	ExternalOIDCExtraKeyFoo                = "extratest.openshift.com/foo"
 	ExternalOIDCExtraKeyFooValueExpression = "claims.email" // This is a variable, not a string literal
+
+	// CEL expressions for claim validation rules
+	ClaimValidationExprEmailExists   = "has(claims.email) && claims.email != ''"
+	ClaimValidationExprEmailVerified = "claims.email_verified == true"
+
+	// CEL expressions for user validation rules
+	UserValidationExprNoSystemPrefix  = "!user.username.startsWith('system:')"
+	UserValidationExprNoForbiddenWord = "!user.username.contains('forbidden')"
```
everettraven: Looks like all of these are used only once as well. No need to make these constants IMO.

test/e2e/external_oidc_test.go
ehearne-redhat: I was thinking it made sense to remove it since we had the config set up but even then, we need to pass the appropriate feature set.

test/e2e/external_oidc_test.go
ehearne-redhat: If I'm understanding the document you provided, that also means I cannot remove tests either. For now, I will reinstate the tests I removed. I'll keep names the same for already existing tests, and I presume since the added tests have not merged yet, it's fine to keep their refactoring.

test/e2e/external_oidc_test.go
ehearne-redhat: It seems to me that the majority of the tests do use this. It can definitely be modified/removed however.

test/e2e/util/external_oidc.go
ehearne-redhat: I don't think so - it might have been a remnant of moving some config changes around. I'll merge these two.

test/e2e/util/external_oidc.go
ehearne-redhat: I think I changed this because there was an issue with ports being in use. Using `:0` meant if a port was in use we could use another. I can test to see if it is still the case.

test/e2e/external_oidc_test.go
ehearne-redhat: I have opted to remove the function as I am seeing now that depending on the auth config used we can't rely on this function to verify for us. Some tests for example use mappings to map prefix + email to username and this function cannot check that for us.

test/e2e/util/external_oidc.go
```diff
@@ -321,10 +330,14 @@ func ChangeUserForKeycloakExtOIDC(t testing.TB, ctx context.Context, clientCfg *
...
 
-	selfSubjectReview, err := authClient.SelfSubjectReviews().Create(ctx, &kauthnv1.SelfSubjectReview{}, metav1.CreateOptions{})
-	g.Expect(err).NotTo(HaveOccurred())
+	if !skipAuthCheck {
+		selfSubjectReview, err := authClient.SelfSubjectReviews().Create(ctx, &kauthnv1.SelfSubjectReview{}, metav1.CreateOptions{})
+		g.Expect(err).NotTo(HaveOccurred())
+		t.Logf("Detected external OIDC cluster using Keycloak as the provider. The user is now %q", selfSubjectReview.Status.UserInfo.Username)
+	} else {
+		t.Logf("Skipping Auth Check as skipAuthCheck is set to true. Providing client config for user %q", clientConfigForExtOIDCUser.Username)
+	}
```
everettraven: Nit: Should we just remove this chunk of code and update all existing test cases to ensure they are explicitly creating and validating a selfsubjectreview?

everettraven (approved):
/lgtm

test/e2e/util/external_oidc.go
```diff
@@ -318,13 +317,9 @@ func ChangeUserForKeycloakExtOIDC(t testing.TB, ctx context.Context, clientCfg *
...
 	clientConfigForExtOIDCUser := GetClientConfigForKeycloakOIDCUser(clientCfg, authConfig, tokenCacheDir)
-	authClient, err := kauthnv1typedclient.NewForConfig(clientConfigForExtOIDCUser)
-	g.Expect(err).NotTo(HaveOccurred())
-
-	selfSubjectReview, err := authClient.SelfSubjectReviews().Create(ctx, &kauthnv1.SelfSubjectReview{}, metav1.CreateOptions{})
+	_, err = kauthnv1typedclient.NewForConfig(clientConfigForExtOIDCUser)
 	g.Expect(err).NotTo(HaveOccurred())
 
-	t.Logf("Detected external OIDC cluster using Keycloak as the provider. The user is now %q", selfSubjectReview.Status.UserInfo.Username)
 	return clientConfigForExtOIDCUser
```
everettraven: I think this could be simplified down to just:
```go
return GetClientConfigForKeycloakOIDCUser(clientCfg, authConfig, tokenCacheDir)
```

test/e2e/external_oidc_test.go
```diff
@@ -102,12 +220,146 @@ func TestExternalOIDC(t *testing.T) {
...
+					},
+				},
+			}
+
+			// Add user validation rules
+			spec.OIDCProviders[0].UserValidationRules = []configv1.TokenUserValidationRule{
+				{
+					Expression: "!user.username.startsWith('system:')",
+					Message:    "username cannot use reserved system: prefix",
+				},
```
everettraven: We never test this rule, so we can probably drop it.

test/e2e/external_oidc_test.go
```diff
@@ -102,12 +220,146 @@ func TestExternalOIDC(t *testing.T) {
...
+
+			// Add claim validation rules
+			spec.OIDCProviders[0].ClaimValidationRules = []configv1.TokenClaimValidationRule{
+				{
+					Type: configv1.TokenValidationRuleTypeCEL,
+					CEL: configv1.TokenClaimValidationCELRule{
+						Expression: "has(claims.email) && claims.email != ''",
+						Message:    "email claim must be present and non-empty",
+					},
+				},
```
everettraven: I think we also never test this rule, so it could probably be dropped.

everettraven (commented):
A few minor comments

everettraven (approved):
/lgtm

everettraven:
@ehearne-redhat This looks like it is ready for review from a hypershift approver to try and grab the `approved` label

control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth_test.go
```diff
@@ -1733,6 +1733,1004 @@ func TestAdaptAuthConfig(t *testing.T) {
 			},
 			shouldError: true,
 		},
+		{
+			name:   "valid discovery URL - different host from issuer URL",
```
enxebre: please use "when it should" convention

hypershift-operator/main.go
```diff
@@ -212,6 +210,9 @@ func NewStartCommand() *cobra.Command {
 	featuregate.ConfigureFeatureSet(featureSet)
 	featuregate.Gate().AddFlag(cmd.Flags())
 
+	// Configure feature set from CPO (needed to propagate feature gates like TechPreviewNoUpgrade)
+	cpofeaturegate.ConfigureFeatureSet(featureSet)
```
enxebre, line 214: why is this needed here?

ehearne-redhat: I'm finding it difficult to provide a clear explanation so here goes:

The auth config validation for both control-plane-operator and hypershift-operator use `generateJWTForProvider()` from `control-plane-operator/controllers/hostedcontrolplane/v2/kas/auth.go`. 

This function checks the control-plane-operator's feature gates, even in hypershift-operator . 

In hypershift-operator, we don't set control-plane-operator feature set. Therefore, when it tries to perform auth config validation, it does not set the user validation rules because the cpofeaturegate.FeatureSet (TPNU in our case) isn't set. 

```mermaid
  flowchart TB
      subgraph HO[HyperShift Operator]
          HO_VAL[Auth Config Validation<br/>support/validations/authentication.go]
          HO_FG[featureset.FeatureSet]
      end

      subgraph CPO[Control Plane Operator]
          CPO_VAL[Auth Config Validation<br/>v2/kas/auth.go:adaptAuthConfig]
          CPO_FG[cpofeaturegate.FeatureSet]
      end

      subgraph SHARED[Shared Validation Code]
          GEN[kas.GenerateAuthConfig<br/>kas.generateJWTForProvider]
          FG_CHECK{featuregates.Gate<br/>.Enabled?}
      end

      HO_VAL --> GEN
      CPO_VAL --> GEN
      GEN --> FG_CHECK
      FG_CHECK -->|reads| CPO_FG

      HO_FG -.->|must propagate to| CPO_FG

      style SHARED fill:#e1f5ff
      style FG_CHECK fill:#fff4e6
      style CPO_FG fill:#d4edda

      classDef problem fill:#ffcccc
```

It instead uses default featureset which does not have `ExternalOIDCWithUpstreamParity` feature gate and causes validation to fail (the mentioned feature gate is not enabled) because there is now conflicting auth config validations, so tests cannot run. 

By setting the cpofeaturegate.FeatureSet (TPNU in our case) here we ensure consistent auth config validation across HyperShift because now user validation rules are set in both auth config validations, and we can test our `ExternalOIDCWithUpstreamParity` feature gate.

Please let me know if it made sense. :D

test/e2e/external_oidc_test.go
```diff
@@ -102,12 +220,135 @@ func TestExternalOIDC(t *testing.T) {
...
+				{
+					Expression: "!user.username.contains('forbidden')",
+					Message:    "username cannot contain the word 'forbidden'",
+				},
+			}
+		}
+
+		featuregates.ConfigureFeatureSet(upstreamParityOpts.FeatureSet)
+
+		e2eutil.NewHypershiftTest(t, ctx, func(t *testing.T, g Gomega, mgtClient crclient.Client, hostedCluster *hyperv1.HostedCluster) {
```
enxebre: this will create a new HC, that has cost and duration impact.
Can we reuse the existing HC for this?

ehearne-redhat: I will give this a go.

ehearne-redhat: Seems like it works. :)

ehearne-redhat:
/verified by e2e-aws-external-oidc-techpreview, e2e-azure-aks-external-oidc-techpreview

PR #8296, opened 2026-04-21
OCPBUGS-63720: orphan machines when managed identity is invalid on clus…

patilsuraj767:
…ter deletion

Summary

When a customer deletes a managed identity required for cleaning up resources in the managed resource group, HyperShift will now skip deletion of those Azure machines and allow the ARO cluster service to handle cleanup using a special Azure 1P application identity.

Fixes: [OCPBUGS-63720](https://redhat.atlassian.net/browse/OCPBUGS-63720)

Changes

- Implement DeleteOrphanedMachines for the Azure platform to satisfy the OrphanDeleter interface by inspecting each AzureMachine's own status conditions for Ready=False with Reason=DeletionFailed, rather than relying on HostedCluster-level credential conditions
- Add a deletionFailedThreshold (10 minutes) based on the machine's DeletionTimestamp to avoid orphaning machines that hit transient cloud API failures
- Add hasDeletionFailedCondition helper to check individual AzureMachine status conditions
- Add unit tests covering all relevant scenarios

How it works

- DeleteOrphanedMachines inspects each AzureMachine individually rather than relying on HostedCluster-level conditions:
- If ManagedIdentities is nil → return early (not a managed-identity cluster)
- List all AzureMachine resources in the control plane namespace
- For each machine, check three conditions before orphaning:
- The machine has a non-zero DeletionTimestamp (it is being deleted)
- The DeletionTimestamp is older than 10 minutes (rules out transient failures)
- The machine has a Ready=False condition with Reason=DeletionFailed (the CAPZ provider explicitly failed to delete the VM, e.g., due to invalid or expired credentials)
- If all three conditions are met, strip the azuremachine.infrastructure.cluster.x-k8s.io finalizer and log that the machine was orphaned

Testing

Unit tests cover:
- ManagedIdentities is nil → no orphaning (early return)
- No machines in namespace → no-op
- Machine with stale DeletionTimestamp and DeletionFailed condition → finalizer removed
- Machine with recent DeletionTimestamp and DeletionFailed condition → finalizer not removed (within threshold)
- Machine with stale DeletionTimestamp but no DeletionFailed condition → finalizer not removed
- Machine not pending deletion → no modification regardless of conditions

### DIFF
--- a/hypershift-operator/controllers/hostedcluster/internal/platform/azure/azure.go
+++ b/hypershift-operator/controllers/hostedcluster/internal/platform/azure/azure.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"path"
 	"strings"
+	"time"
 
 	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
 	"github.com/openshift/hypershift/control-plane-operator/controllers/hostedcontrolplane/cloud/azure"
@@ -24,11 +25,14 @@ import (
 	rbacv1 "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/utils/ptr"
 
 	capiazure "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
 	capiv1 "sigs.k8s.io/cluster-api/api/core/v1beta1"
+	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	"github.com/blang/semver"
 )
@@ -353,6 +357,71 @@ func (a Azure) CAPIProviderPolicyRules() []rbacv1.PolicyRule {
 
 func (a Azure) DeleteCredentials(ctx context.Context, c client.Client, hcluster *hyperv1.HostedCluster, controlPlaneNamespace string) error {
 	return nil
+}
+
+// deletionFailedThreshold is the minimum duration a machine must have had a non-zero
+// DeletionTimestamp before it is considered permanently stuck and eligible for orphaning.
+// This guards against orphaning machines that hit transient failures (e.g. rate limiting).
+const deletionFailedThreshold = 10 * time.Minute
+
+// DeleteOrphanedMachines removes the finalizer from AzureMachines that are stuck in deletion
+// due to credential failures. This is detected by checking each AzureMachine's Status.Conditions
+// for a Ready=False condition with Reason=DeletionFailed. Orphaning the machine allows management
+// cluster cleanup to proceed without requiring valid cloud credentials.
+func (Azure) DeleteOrphanedMachines(ctx context.Context, c client.Client, hc *hyperv1.HostedCluster, controlPlaneNamespace string) error {
+	// This orphaning behavior is intended for managed-identity cleanup flow.
+	if hc.Spec.Platform.Azure.AzureAuthenticationConfig.ManagedIdentities == nil {
+		return nil
+	}
+
+	azureMachineList := capiazure.AzureMachineList{}
+	if err := c.List(ctx, &azureMachineList, client.InNamespace(controlPlaneNamespace)); err != nil {
+		return fmt.Errorf("failed to list AzureMachines in %s: %w", controlPlaneNamespace, err)
+	}
+
+	logger := ctrl.LoggerFrom(ctx)
+	var errs []error
+
+	for i := range azureMachineList.Items {
+		azureMachine := &azureMachineList.Items[i]
+		if azureMachine.DeletionTimestamp.IsZero() {
+			continue
+		}
+		if time.Since(azureMachine.DeletionTimestamp.Time) < deletionFailedThreshold {
+			continue
+		}
+		if !hasDeletionFailedCondition(azureMachine) {
+			continue
+		}
+		// Remove the AzureMachine finalizer to orphan the machine, leaving Azure
+		// infrastructure intact rather than attempting cloud API calls with invalid credentials.
+		if removed := controllerutil.RemoveFinalizer(azureMachine, capiazure.MachineFinalizer); !removed {
+			continue
+		}
+		if err := c.Update(ctx, azureMachine); err != nil {
+			errs = append(errs, fmt.Errorf("failed to orphan machine %s/%s: %w",
+				azureMachine.Namespace, azureMachine.Name, err))
+			continue
+		}
+		logger.Info("orphaning azuremachine stuck in deletion due to credential failure",
+			"machine", client.ObjectKeyFromObject(azureMachine))
+	}
+
+	return utilerrors.NewAggregate(errs)
+}
+
+// hasDeletionFailedCondition returns true if the AzureMachine has a Ready condition with
+// Status=False and Reason=DeletionFailed, indicating the cloud provider could not delete
+// the underlying VM (e.g., due to invalid or expired credentials).
+func hasDeletionFailedCondition(azureMachine *capiazure.AzureMachine) bool {
+	for _, condition := range azureMachine.Status.Conditions {
+		if condition.Type == capiv1.ReadyCondition &&
+			condition.Status == corev1.ConditionFalse &&
+			condition.Reason == capiazure.DeletionFailedReason {
+			return true
+		}
+	}
+	return false
 }
 
 func reconcileAzureCluster(azureCluster *capiazure.AzureCluster, hcluster *hyperv1.HostedCluster, apiEndpoint hyperv1.APIEndpoint, azureClusterIdentity *capiazure.AzureClusterIdentity, _ string) error {
--- a/hypershift-operator/controllers/hostedcluster/internal/platform/azure/azure_test.go
+++ b/hypershift-operator/controllers/hostedcluster/internal/platform/azure/azure_test.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"testing"
+	"time"
 
 	. "github.com/onsi/gomega"
 
@@ -17,6 +18,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	capiazure "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
+	capiv1 "sigs.k8s.io/cluster-api/api/core/v1beta1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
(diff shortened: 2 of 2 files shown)


machi1990:
@patilsuraj767  This seems to handle the case of missing missing MI when deleting Azure machines. I am thinking that there could be other cases as well that'll fail deletion because of a missing MI (e.g KMS) or because of a missing user brought resources e.g KSM, Vnet, Subnet, or the overall RG etc Are you planning to handle that as well?

patilsuraj767:
> @patilsuraj767 This seems to handle the case of missing missing MI when deleting Azure machines. I am thinking that there could be other cases as well that'll fail deletion because of a missing MI (e.g KMS) or because of a missing user brought resources e.g KSM, Vnet, Subnet, or the overall RG etc Are you planning to handle that as well?

@machi1990 We need to verify whether the HyperShift delete workflow fails when user-provided resources are missing during deletion. If the cluster deletion fails due to missing user-provided resources, that should be addressed in a separate PR. This change should be scoped only to missing/invalid managed identity handling.

machi1990:
> > @patilsuraj767 This seems to handle the case of missing missing MI when deleting Azure machines. I am thinking that there could be other cases as well that'll fail deletion because of a missing MI (e.g KMS) or because of a missing user brought resources e.g KSM, Vnet, Subnet, or the overall RG etc Are you planning to handle that as well?
> 
> @machi1990 We need to verify whether the HyperShift delete workflow fails when user-provided resources are missing during deletion. If the cluster deletion fails due to missing user-provided resources, that should be addressed in a separate PR. This change should be scoped only to missing/invalid managed identity handling.

Yes it does and these are the cases we've already seen e.g the recent cspr orphaned are because of these issues.
```
lastTransitionTime: "2026-04-27T15:24:17Z"
      message: |
        failed to get network security group info to verify its location: failed to get network security group: GET https://management.azure.com/subscriptions/<subId>/resourceGroups/pr-check-e2e-tests-resource-group-b796l/providers/Microsoft.Network/networkSecurityGroups/pr-check-e2e-tests-nsg-fh2x7
        --------------------------------------------------------------------------------
        RESPONSE 404: 404 Not Found
        ERROR CODE: ResourceNotFound
        --------------------------------------------------------------------------------
        {
          "error": {
            "code": "ResourceNotFound",
            "message": "The Resource 'Microsoft.Network/networkSecurityGroups/pr-check-e2e-tests-nsg-fh2x7' under resource group 'pr-check-e2e-tests-resource-group-b796l' was not found. For more details please go to https://aka.ms/ARMResourceNotFoundFix"
          }
        }
``` 

```
 message: |
        failed to encrypt data using KMS (key: etcd-data-kms-encryption-key/<keyId>): POST https://pr-check-key-vault-p9x9h.vault.azure.net/keys/etcd-data-kms-encryption-key/<keyId>/encrypt
        --------------------------------------------------------------------------------
        RESPONSE 404: 404 Not Found
        ERROR CODE: VaultNotFound
        --------------------------------------------------------------------------------
        {
          "error": {
            "code": "VaultNotFound",
            "message": "Azure Key Vault 'pr-check-key-vault-p9x9h' does not exist."
          }
        }
        --------------------------------------------------------------------------------
```

@patilsuraj767 okay if it needs to be a separate PR but let's make sure that this is addressed as part of your effort as its not only MI missing that can cause deletion stuck but other infra missing as well can cause deletion to be stuck and its critical we tackle it broadly

hypershift-operator/controllers/hostedcluster/internal/platform/azure/azure.go
```diff
@@ -355,6 +358,71 @@ func (a Azure) DeleteCredentials(ctx context.Context, c client.Client, hcluster
...
+		return fmt.Errorf("failed to list AzureMachines in %s: %w", controlPlaneNamespace, err)
+	}
+
+	logger := ctrl.LoggerFrom(ctx)
+	var errs []error
+
+	for i := range azureMachineList.Items {
+		azureMachine := &azureMachineList.Items[i]
+		if !azureMachine.DeletionTimestamp.IsZero() {
+			// Remove finalizers to orphan the machine
```
stevekuznetsov: Can't we be specific about which finalizer(s) we intend to remove?

hypershift-operator/controllers/hostedcluster/internal/platform/azure/azure.go
```diff
@@ -355,6 +358,71 @@ func (a Azure) DeleteCredentials(ctx context.Context, c client.Client, hcluster
...
+// Azure 1P application identity.
+func (Azure) DeleteOrphanedMachines(ctx context.Context, c client.Client, hc *hyperv1.HostedCluster, controlPlaneNamespace string) error {
+
+	// This orphaning behavior is intended for managed-identity cleanup flow.
+	if hc.Spec.Platform.Azure.AzureAuthenticationConfig.ManagedIdentities == nil {
+		return nil
+	}
+
+	// If credentials are valid, nothing to do - normal deletion will work
+	if ValidCredentials(hc) {
```
stevekuznetsov: Can we race whereby the CPO doesn't have time yet to post the invalid credential status, and we still start the deletion? Can we detect that case and still do the right thing?

hypershift-operator/controllers/hostedcluster/internal/platform/azure/azure.go
```diff
@@ -355,6 +358,71 @@ func (a Azure) DeleteCredentials(ctx context.Context, c client.Client, hcluster
 	return nil
 }
 
+// ValidCredentials checks if the Azure credentials are valid by examining the HostedCluster status conditions.
+// Since all Azure hosted clusters use managed identities, we check for any conditions indicating
+// credential failures.
+// Returns true if credentials are valid, false otherwise.
+func ValidCredentials(hc *hyperv1.HostedCluster) bool {
```
jparrill: AWS uses a tri-state (`Valid`/`Invalid`/`Unknown`) and orphans on both `Invalid` and `Unknown`. Here `Unknown` (no conditions yet) is treated as `valid` → no orphaning. If intentional, a comment explaining why would help.

hypershift-operator/controllers/hostedcluster/internal/platform/azure/azure.go
```diff
@@ -355,6 +358,71 @@ func (a Azure) DeleteCredentials(ctx context.Context, c client.Client, hcluster
...
+	// Check if any condition has failed due to invalid Azure credentials (set by the CPO).
+	for _, condition := range hc.Status.Conditions {
```
jparrill: Only `ValidAzureKMSConfig` uses `InvalidAzureCredentialsReason` today. Consider checking specific condition types (like AWS does with `ValidOIDCConfiguration` / `ValidAWSIdentityProvider`) instead of scanning all conditions.

hypershift-operator/controllers/hostedcluster/internal/platform/azure/azure_test.go
```diff
@@ -411,3 +412,292 @@ func TestReconcileCredentials(t *testing.T) {
...
+					},
+				},
+			},
+			azureMachines: []capiazure.AzureMachine{
+				{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:              "machine-1",
+						Namespace:         controlPlaneNamespace,
+						Finalizers:        []string{"azuremachine.infrastructure.cluster.x-k8s.io"},
+						DeletionTimestamp: &metav1.Time{},
```
jparrill: `&metav1.Time{}` is zero — `IsZero()` returns `true`, so this machine is not actually pending deletion. The test passes via the early return, but the field is misleading. Either remove it or use `time.Now()`.

hypershift-operator/controllers/hostedcluster/internal/platform/azure/azure_test.go
```diff
@@ -411,3 +412,292 @@ func TestReconcileCredentials(t *testing.T) {
...
+					},
+				},
+			},
+			expectedFinalizersRemoved: false,
+			expectedError:             false,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
```
jparrill, line 684: nit: missing `t.Parallel()` — `TestValidCredentials` has it, safe to add here since each subtest has its own `fakeClient`.

jparrill (commented):

hypershift-operator/controllers/hostedcluster/internal/platform/azure/azure.go
patilsuraj767: I have updated the logic to remove only the specific finalizer instead of removing all finalizers indiscriminately.

hypershift-operator/controllers/hostedcluster/internal/platform/azure/azure.go
patilsuraj767: I think it can self-heal, If credentials become invalid around the same time deletion is triggered, ValidCredentials() can return true because CPO didn't completed yet, resulting in AzureMachine provider not able to delete the machine. The machine is stuck in deletion with its DeletionTimestamp set but finalizer never removed.
But HO will keep requeuing because the cluster still exists. Eventually the CPO posts the condition on the HCP, the HO copies it to the HC, and the next call to DeleteOrphanedMachines correctly orphans. So the race is transient.

hypershift-operator/controllers/hostedcluster/internal/platform/azure/azure.go
patilsuraj767: You are right but after revisiting this, I think AzureKMSConfig is completely unrelated here.
ValidAzureKMSConfig validates that the Key Vault credential `spec.secretEncryption.kms.azure.kms.credentialsSecretName` can call Azure Key Vault's Encrypt API. This is purely an etcd encryption concern.

AzureMachine (CAPZ) relies on AzureClusterIdentity, which is derived from a different identity: `ManagedIdentities.ControlPlane.NodePoolManagement.CredentialsSecretName`

As a result, `ValidCredentials` will produce false negatives for the actual intent. In a scenario where the NodePoolManagement identity is revoked (which DeleteOrphanedMachines is meant to handle), ValidCredentials would still return true, unless KMS is also configured and failing. That means the orphan cleanup logic would never be triggered.

Do you know how HO or CPO report health of the ManagedIdentities?  specifically of `AzureAuthenticationConfiguration.ManagedIdentities.ControlPlane` and `AzureAuthenticationConfiguration.ManagedIdentities.DataPlane`

jparrill: Humm, makes sense, ValidAzureKMSConfig validates KMS credentials, not the NodePoolManagement identity CAPZ uses. Checked the codebase and there's no HC/HCP condition reporting the health of `ManagedIdentities.ControlPlane.NodePoolManagement`. Checking DeletionFailed directly on each AzureMachine is the right call here.

jparrill:
/retitle  OCPBUGS-63720: orphan machines when managed identity is invalid on clus…

patilsuraj767:

The test is permafailing and not related to this PR.

patilsuraj767:
/verified by @patilsuraj767 

Tested the PR in personal dev environment ARO-HCP.

sdminonne:
@patilsuraj767 may you modify the PR description (agents get confused pretty quickly)?
The actual orphaning strategy is pretty different. AWS uses a cluster-wide credential validity check and  removes all finalizers immediately. Azure uses per-machine condition inspection with a 10-minute grace period and removes only the specific CAPZ finalizer.
I know that somehow this is reported but I've hard time to define it as 'mirror'.  TY!

patilsuraj767:
> @patilsuraj767 may you modify the PR description (agents get confused pretty quickly)? The actual orphaning strategy is pretty different. AWS uses a cluster-wide credential validity check and removes all finalizers immediately. Azure uses per-machine condition inspection with a 10-minute grace period and removes only the specific CAPZ finalizer. I know that somehow this is reported but I've hard time to define it as 'mirror'. TY!

@sdminonne I have updated the PR description, removed the AWS reference from the description. 
Also, I have rebased the PR and fixed the gitlint.
Please take a look.

devguyio:
/uncc @devguyio 

I see @csrwng @jparrill and @sdminonne already reviewed.

hypershift-operator/controllers/hostedcluster/internal/platform/azure/azure.go
```diff
@@ -355,6 +359,71 @@ func (a Azure) DeleteCredentials(ctx context.Context, c client.Client, hcluster
 	return nil
 }
 
+// deletionFailedThreshold is the minimum duration a machine must have had a non-zero
+// DeletionTimestamp before it is considered permanently stuck and eligible for orphaning.
+// This guards against orphaning machines that hit transient failures (e.g. rate limiting).
+const deletionFailedThreshold = 10 * time.Minute
```
stevekuznetsov, line 365: Do we have a Kusto query showing our distribution of deletion times? If so, please leave it here so someone can re-run the analysis later, and specify what percentile 10 minutes is - p99? p99.99?

PR #8302, opened 2026-04-21
OCPBUGS-84011: fix(karpenter): watch VAP resources to reconcile on deletion

joshbranham:

The EC2NodeClass controller creates a ValidatingAdmissionPolicy and binding on the guest cluster but did not watch those resources. If they were deleted externally, the controller would not detect the deletion and recreate them until an unrelated watched resource changed.

Add watches for ValidatingAdmissionPolicy and
ValidatingAdmissionPolicyBinding filtered to the karpenter-managed resources so that deletion triggers immediate reconciliation.

Fixes OCPBUGS-84011

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Admission policy and binding changes in guest clusters now also trigger synchronization of EC2 node class configuration across hosted clusters.

* **Tests**
  * Added unit tests verifying that admission policy and binding events correctly map to and enqueue reconciliation for existing EC2 node classes.

### DIFF
--- a/karpenter-operator/controllers/nodeclass/ec2_nodeclass_controller.go
+++ b/karpenter-operator/controllers/nodeclass/ec2_nodeclass_controller.go
@@ -101,22 +101,8 @@ func (r *EC2NodeClassReconciler) SetupWithManager(ctx context.Context, mgr ctrl.
 			},
 		)).
 		Watches(&awskarpenterv1.EC2NodeClass{}, &handler.EnqueueRequestForObject{}).
-		Watches(&admissionv1.ValidatingAdmissionPolicy{}, handler.EnqueueRequestsFromMapFunc(
-			func(ctx context.Context, o client.Object) []ctrl.Request {
-				if o.GetName() != "karpenter.ec2nodeclass.hypershift.io" {
-					return nil
-				}
-				return r.mapToOpenShiftEC2NodeClasses(ctx, o)
-			},
-		)).
-		Watches(&admissionv1.ValidatingAdmissionPolicyBinding{}, handler.EnqueueRequestsFromMapFunc(
-			func(ctx context.Context, o client.Object) []ctrl.Request {
-				if o.GetName() != "karpenter-binding.ec2nodeclass.hypershift.io" {
-					return nil
-				}
-				return r.mapToOpenShiftEC2NodeClasses(ctx, o)
-			},
-		)).
+		Watches(&admissionv1.ValidatingAdmissionPolicy{}, handler.EnqueueRequestsFromMapFunc(r.mapVAPToOpenShiftEC2NodeClasses)).
+		Watches(&admissionv1.ValidatingAdmissionPolicyBinding{}, handler.EnqueueRequestsFromMapFunc(r.mapVAPBindingToOpenShiftEC2NodeClasses)).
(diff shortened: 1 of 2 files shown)


karpenter-operator/controllers/nodeclass/ec2_nodeclass_controller.go
```diff
@@ -101,6 +101,22 @@ func (r *EC2NodeClassReconciler) SetupWithManager(ctx context.Context, mgr ctrl.
 			},
 		)).
 		Watches(&awskarpenterv1.EC2NodeClass{}, &handler.EnqueueRequestForObject{}).
+		Watches(&admissionv1.ValidatingAdmissionPolicy{}, handler.EnqueueRequestsFromMapFunc(
```
maxcao13: Can we add a unit test that makes sure if we delete it, that it comes back immediately? I think the existing fake clients in the tests can handle this integration test type scenario: https://github.com/joshbranham/hypershift/blob/abf43172890503b7a48743e84c718fff0f226d65/karpenter-operator/controllers/nodeclass/ec2_nodeclass_controller_test.go#L1265

lgtme otherwise :-)

joshbranham: Good call! Added a simple delete + get 404 + reconcile adds it back. Lemme know if you had something else in mind 👍

joshbranham:
Waiting for https://github.com/openshift/release/pull/78516

karpenter-operator/controllers/nodeclass/ec2_nodeclass_controller.go
enxebre: to clarify, while the additional test you added is good to have, it doesn't cover this change, watchers are not involved in that unit test.

karpenter-operator/controllers/nodeclass/ec2_nodeclass_controller.go
```diff
@@ -101,6 +101,22 @@ func (r *EC2NodeClassReconciler) SetupWithManager(ctx context.Context, mgr ctrl.
...
+			func(ctx context.Context, o client.Object) []ctrl.Request {
+				if o.GetName() != "karpenter.ec2nodeclass.hypershift.io" {
+					return nil
```
enxebre: can we please unit test this new changes, [codecov/patch](https://github.com/openshift/hypershift/pull/8302/checks?check_run_id=73509927552)

karpenter-operator/controllers/nodeclass/ec2_nodeclass_controller.go
joshbranham: Right, imo, testing the controller-runtime watch behavior seems not neccesary, but if you'd prefer that I can figure out how to make that happen.

karpenter-operator/controllers/nodeclass/ec2_nodeclass_controller.go
joshbranham: I extracted the map funcs so we can unit test them standalone in https://github.com/openshift/hypershift/pull/8302/commits/da4e269ec0c1f97107c873b8083d249e181cd9de 👍

PR #8305, opened 2026-04-22
CNTRLPLANE-2539: Move generation of the CAPI Provider Role

cwilkers:
[CNTRLPLANE-2539](https://redhat.atlassian.net/browse/CNTRLPLANE-2539): Move generation of CAPI Provider Role from cli to operator

Supports work in ZTP deployment of HCP clusters by taking Role generation (which is a security risk in gitops workflows) out of the CLI or manual steps and into the operator. Now the operator will handle the role creation, and gitops based deployments do not require relaxing security restrictions to create Roles.

Fixes [CNTRLPLANE-2539](https://redhat.atlassian.net/browse/CNTRLPLANE-2539)

## Checklist:
- [ + ] Subject and description added to both, commit and PR.
- [ + ] Relevant issues have been referenced.
- [ ? ] This change includes docs. 
- [ + ] This change includes unit tests.

## AI Assistance

- Model Used: Claude with claude-sonnet-4@20250514 in CLI and in VS Code IDE integration
- Scope: Planning, Test writing, and Code
- Level: Code generation and Code assistance
- Human Review: Quick code review by @cwilkers, manual testing of multiple use cases on local cluster

## Summary by CodeRabbit

* **Refactor**
  * RBAC role emission removed from the legacy creation path; agent credential lifecycle now centralizes role management and only deletes a shared role when no remaining bindings reference it.

* **Tests**
  * Added and extended tests to verify shared role creation, idempotent reconciliation, preservation across multiple hosted clusters, and deletion once unreferenced.

### DIFF
--- a/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
+++ b/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
@@ -2895,10 +2895,8 @@ func (r *HostedClusterReconciler) reconcileCAPIProvider(cpContext controlplaneco
 		},
 	}
 	err = cpContext.Client.Get(cpContext, client.ObjectKeyFromObject(capiProviderDeployment), capiProviderDeployment)
-	if err != nil {
-		if !apierrors.IsNotFound(err) {
-			return fmt.Errorf("failed to fetch capi provider deployment: %w", err)
-		}
+	if err != nil && !apierrors.IsNotFound(err) {
+		return fmt.Errorf("failed to fetch capi provider deployment: %w", err)
 	}
 	if err == nil {
 		if capiProviderDeployment.Spec.Template.ObjectMeta.Labels[hyperv1.ControlPlaneComponentLabel] != "capi-provider" {
--- a/hypershift-operator/controllers/hostedcluster/internal/platform/agent/agent.go
+++ b/hypershift-operator/controllers/hostedcluster/internal/platform/agent/agent.go
@@ -6,6 +6,7 @@ import (
 	"os"
 
 	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
+	"github.com/openshift/hypershift/hypershift-operator/controllers/manifests/clusterapi"
 	"github.com/openshift/hypershift/hypershift-operator/controllers/manifests/ignitionserver"
 	"github.com/openshift/hypershift/support/images"
 	"github.com/openshift/hypershift/support/k8sutil"
@@ -29,6 +30,10 @@ const (
 	// TODO Pin to specific release
 	imageCAPAgent         = "quay.io/edge-infrastructure/cluster-api-provider-agent:latest"
 	CredentialsRBACPrefix = "cluster-api-agent"
+	CAPIProviderRoleName  = "capi-provider-role"
+	// capiProviderAgentRBACLabel marks Role/RoleBinding reconciled for the agent CAPI provider.
+	capiProviderAgentRBACLabelKey   = "hypershift.openshift.io/capi-provider-agent-rbac"
+	capiProviderAgentRBACLabelValue = "true"
 )
 
 type Agent struct{}
@@ -132,10 +137,53 @@ func (p Agent) CAPIProviderDeploymentSpec(hcluster *hyperv1.HostedCluster, _ *hy
 	return deploymentSpec, nil
 }
 
-// TODO add a new method to Platform interface?
 func (p Agent) ReconcileCredentials(ctx context.Context, c client.Client, createOrUpdate upsert.CreateOrUpdateFN,
 	hcluster *hyperv1.HostedCluster,
 	controlPlaneNamespace string) error {
+	return ReconcileCAPIProviderRole(ctx, c, createOrUpdate, hcluster, controlPlaneNamespace)
+}
+
+// setHypershiftMetadata ensures the object has HyperShift label and annotation set.
+func setHypershiftMetadata(obj client.Object, hcluster *hyperv1.HostedCluster) {
+	if obj.GetLabels() == nil {
+		obj.SetLabels(map[string]string{})
+	}
+	obj.GetLabels()[capiProviderAgentRBACLabelKey] = capiProviderAgentRBACLabelValue
+	if obj.GetAnnotations() == nil {
+		obj.SetAnnotations(map[string]string{})
+	}
+	obj.GetAnnotations()[k8sutil.HostedClusterAnnotation] = client.ObjectKeyFromObject(hcluster).String()
+}
+
+// ReconcileCAPIProviderRole creates the RBAC Role and RoleBinding in the agent
+// namespace so the CAPI provider can manage Agent resources. Each HostedCluster
+// creates its own Role (named per control plane namespace) to enable proper
+// lifecycle management and watch-based reconciliation.
+func ReconcileCAPIProviderRole(ctx context.Context, c client.Client, createOrUpdate upsert.CreateOrUpdateFN,
+	hcluster *hyperv1.HostedCluster,
+	controlPlaneNamespace string) error {
+
+	roleName := fmt.Sprintf("%s-%s", CAPIProviderRoleName, controlPlaneNamespace)
+	role := &rbacv1.Role{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: hcluster.Spec.Platform.Agent.AgentNamespace,
+			Name:      roleName,
+		},
+	}
+	_, err := createOrUpdate(ctx, c, role, func() error {
+		setHypershiftMetadata(role, hcluster)
+		role.Rules = []rbacv1.PolicyRule{
+			{
+				APIGroups: []string{"agent-install.openshift.io"},
+				Resources: []string{"agents"},
+				Verbs:     []string{"get", "list", "watch", "update", "patch"},
+			},
+		}
+		return nil
+	})
+	if err != nil {
+		return fmt.Errorf("failed to reconcile Agent Role: %w", err)
+	}
 
 	roleBinding := &rbacv1.RoleBinding{
 		ObjectMeta: metav1.ObjectMeta{
@@ -143,18 +191,19 @@ func (p Agent) ReconcileCredentials(ctx context.Context, c client.Client, create
 			Name:      fmt.Sprintf("%s-%s", CredentialsRBACPrefix, controlPlaneNamespace),
 		},
 	}
-	_, err := createOrUpdate(ctx, c, roleBinding, func() error {
+	_, err = createOrUpdate(ctx, c, roleBinding, func() error {
+		setHypershiftMetadata(roleBinding, hcluster)
 		roleBinding.Subjects = []rbacv1.Subject{
 			{
 				Kind:      "ServiceAccount",
-				Name:      "capi-provider",
+				Name:      clusterapi.CAPIProviderServiceAccount(controlPlaneNamespace).Name,
 				Namespace: controlPlaneNamespace,
 			},
 		}
 		roleBinding.RoleRef = rbacv1.RoleRef{
-			APIGroup: "rbac.authorization.k8s.io",
+			APIGroup: rbacv1.GroupName,
 			Kind:     "Role",
-			Name:     "capi-provider-role",
+			Name:     roleName,
 		}
 		return nil
 	})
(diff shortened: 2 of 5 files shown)


hypershift-operator/controllers/hostedcluster/internal/platform/agent/agent.go
```diff
@@ -210,5 +231,25 @@ func (Agent) DeleteCredentials(ctx context.Context, c client.Client,
 	if _, err := hyperutil.DeleteIfNeeded(ctx, c, &rbacv1.RoleBinding{ObjectMeta: metav1.ObjectMeta{Name: fmt.Sprintf("%s-%s", CredentialsRBACPrefix, controlPlaneNamespace), Namespace: hc.Spec.Platform.Agent.AgentNamespace}}); err != nil {
 		return fmt.Errorf("failed to clean up CAPI provider rolebinding: %w", err)
 	}
+	deletedBindingName := fmt.Sprintf("%s-%s", CredentialsRBACPrefix, controlPlaneNamespace)
```
clebs:
```suggestion
	deletedBindingName := fmt.Sprintf("%s-%s", CredentialsRBACPrefix, controlPlaneNamespace)
	if _, err := hyperutil.DeleteIfNeeded(ctx, c, &rbacv1.RoleBinding{ObjectMeta: metav1.ObjectMeta{Name: deletedBindingName, Namespace: hc.Spec.Platform.Agent.AgentNamespace}}); err != nil {
		return fmt.Errorf("failed to clean up CAPI provider rolebinding: %w", err)
	}
```

cwilkers: Nice catch, I just force-pushed the amended change as suggested.

jparrill:
Thanks for working on this @cwilkers — I understand the motivation for ZTP/gitops workflows. I have a few concerns about the current approach that I'd like to discuss before we settle on the right solution:

1. Role reconciliation on every loop iteration

Moving the Role creation into `ReconcileCredentials()` means every reconcile of every Agent-platform HostedCluster will issue at minimum a GET against the API server for a Role whose content isstatic and never changes. Before this PR, the Role was created once by the CLI and left alone. The self-healing benefit is real, but the cost of reconciling a static resource on every iterationseems disproportionate — especially at scale with many Agent-platform HCs.

Would it make sense to either (a) check for existence before calling createOrUpdate, or (b) move the Role creation to a one-time setup in the operator startup (e.g., alongside other platformbootstrapping)?

2. Race condition in shared agent namespace teardown

When two HostedClusters share the same agentNamespace and are deleted concurrently, the following race can occur:

1. HC1 and HC2 both enter DeleteCredentials
2. Both delete their respective RoleBindings
3. Both list remaining RoleBindings — each sees zero (or only the other's with a DeletionTimestamp)
4. Both decide to delete the shared capi-provider-role                                                                                                                                                 5. While HC1's CAPI provider is still running its Cluster finalizer cleanup (which needs access to agents resources), HC2's DeleteCredentials deletes the Role

The double-delete itself is safe (DeleteIfNeeded is idempotent), but the problem is the Role disappearing while a CAPI provider still needs it to release agents during teardown. The deletion order
in delete() (CAPI Cluster deleted → wait for completion → DeleteCredentials) protects the single-HC case, but not the concurrent multi-HC case.1. Role reconciliation on every loop

3. Some Minos items:
- The List in DeleteCredentials fetches all RoleBindings in the agent namespace (a user-controlled namespace that could have many unrelated bindings). Consider adding a HyperShift label to managedRoleBindings during reconciliation and filtering by that label during deletion.
- The Role grants Verbs: []string{"*"} on agents — I know this is unchanged from the CLI version, but since we're moving it to the operator this would be a good opportunity to scope it down to the   minimum required set (get, list, watch, update, patch).
- Consider using rbacv1.GroupName instead of the hardcoded "rbac.authorization.k8s.io" string.

I see your point on the ZTP use case but concerns 1 and 2 would need to be addressed before this can go forward. Let me know your toughts.

cwilkers:
With some planning help by Claude, I can move the Role creation out of the ReconcileCredentials function where it doesn't make that much sense, and into the reconcileCAPIProvider function where it can be gated to only create if the capi deployment does not exist yet. I'll need to test.

For the deletion race condition, would it be an option to simply not have the operator delete the Role? It's not as clean, but a NS deletion will clear it and it doesn't really grant access to anything we would worry about. The other option suggested by AI would be to add multiple finalizers to the Role, which seems clunky to me.

cwilkers:
The changes I just pushed fix the deletion race condition by creating a different role for each hosted cluster using a namespace. Now the Role may be cleaned when finalizing the HostedCluster as normal.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -2899,6 +2900,12 @@ func (r *HostedClusterReconciler) reconcileCAPIProvider(cpContext controlplaneco
 		if !apierrors.IsNotFound(err) {
 			return fmt.Errorf("failed to fetch capi provider deployment: %w", err)
 		}
+		// Deployment does not exist yet — perform first-time platform RBAC setup.
+		if hcluster.Spec.Platform.Type == hyperv1.AgentPlatform {
```
jparrill: This introduces a direct import of `platformagent` and a platform type check in the generic controller. Every other platform routes this kind of work through the `Platform` interface — specifically `ReconcileCredentials`, which already has exactly the right signature (`ctx`, `client`, `createOrUpdate`, `hcluster`, `controlPlaneNamespace`).

Also, this block is inside the `IsNotFound` branch, meaning the Role/RoleBinding are only created when the CAPI provider Deployment doesn't exist yet. Once it's running, RBAC is never reconciled again. If someone deletes the Role out-of-band, there's no self-healing until the Deployment is also deleted.

The simplest fix for both problems:
```go
func (p Agent) ReconcileCredentials(ctx context.Context, c client.Client,
    createOrUpdate upsert.CreateOrUpdateFN, hcluster *hyperv1.HostedCluster,
    controlPlaneNamespace string) error {
    return ReconcileCAPIProviderRole(ctx, c, createOrUpdate, hcluster, controlPlaneNamespace)
}
```
This restores continuous reconciliation (self-healing), goes through the interface, and lets you revert the `createOrUpdate` parameter threading + this type check + the `platformagent` import.

hypershift-operator/controllers/hostedcluster/internal/platform/agent/agent.go
```diff
@@ -132,18 +136,64 @@ func (p Agent) CAPIProviderDeploymentSpec(hcluster *hyperv1.HostedCluster, _ *hy
 	return deploymentSpec, nil
 }
 
-// TODO add a new method to Platform interface?
 func (p Agent) ReconcileCredentials(ctx context.Context, c client.Client, createOrUpdate upsert.CreateOrUpdateFN,
 	hcluster *hyperv1.HostedCluster,
 	controlPlaneNamespace string) error {
+	return nil
```
jparrill: This used to create the RoleBinding on every reconcile loop. Now it's a no-op, but it's still called every loop via `reconcilePlatformCredentialsWithStatus`. The `PlatformCredentialsFound` condition will always report True for Agent regardless of whether the RBAC actually exists.

If the intent is to keep `ReconcileCAPIProviderRole` as the implementation, this method should delegate to it rather than silently doing nothing. The old TODO comment about adding a Platform interface method was removed but the underlying concern wasn't addressed.

hypershift-operator/controllers/hostedcluster/internal/platform/agent/agent.go
```diff
@@ -132,18 +136,64 @@ func (p Agent) CAPIProviderDeploymentSpec(hcluster *hyperv1.HostedCluster, _ *hy
...
+
+	roleName := fmt.Sprintf("%s-%s", CAPIProviderRoleName, controlPlaneNamespace)
+	role := &rbacv1.Role{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: hcluster.Spec.Platform.Agent.AgentNamespace,
+			Name:      roleName,
+		},
+	}
+	_, err := createOrUpdate(ctx, c, role, func() error {
+		if role.Labels == nil {
```
jparrill: Nit: this 8-line block (nil-check labels, set label, nil-check annotations, set annotation) is duplicated verbatim for the RoleBinding below. Consider extracting a small helper to keep them in sync.

hypershift-operator/controllers/hostedcluster/internal/platform/agent/agent.go
```diff
@@ -132,18 +136,64 @@ func (p Agent) CAPIProviderDeploymentSpec(hcluster *hyperv1.HostedCluster, _ *hy
...
+	_, err = createOrUpdate(ctx, c, roleBinding, func() error {
+		if roleBinding.Labels == nil {
+			roleBinding.Labels = map[string]string{}
+		}
+		roleBinding.Labels[capiProviderAgentRBACLabelKey] = capiProviderAgentRBACLabelValue
+		if roleBinding.Annotations == nil {
+			roleBinding.Annotations = map[string]string{}
+		}
+		roleBinding.Annotations[k8sutil.HostedClusterAnnotation] = client.ObjectKeyFromObject(hcluster).String()
 		roleBinding.Subjects = []rbacv1.Subject{
```
jparrill, line 196: Nit: the SA name `"capi-provider"` on the next line is already defined in `manifests/clusterapi/manifests.go` via `CAPIProviderServiceAccount()`. Using the manifest constant instead of a string literal avoids a silent coupling that could break if the name ever changes upstream.

hypershift-operator/controllers/hostedcluster/internal/platform/agent/agent.go
```diff
@@ -204,11 +254,26 @@ func reconcileAgentCluster(agentCluster *agentv1.AgentCluster, ignEndpoint, cont
 	return nil
 }
 
+// DeleteCredentials removes this HostedCluster's CAPI provider Role and RoleBinding.
+// Each cluster has its own Role, so both can be safely deleted without affecting other clusters.
 func (Agent) DeleteCredentials(ctx context.Context, c client.Client,
```
jparrill, line 259: Worth noting: clusters created with the previous CLI have a shared `capi-provider-role` (no `-<cpNamespace>` suffix). After upgrading, this cleanup path only looks for `capi-provider-role-<cpNamespace>`, so the old shared Role will leak in the agent namespace. Might want to add cleanup for the legacy name, or at least document that it's expected.

hypershift-operator/controllers/hostedcluster/internal/platform/agent/agent.go
```diff
@@ -132,18 +136,64 @@ func (p Agent) CAPIProviderDeploymentSpec(hcluster *hyperv1.HostedCluster, _ *hy
...
+			role.Labels = map[string]string{}
+		}
+		role.Labels[capiProviderAgentRBACLabelKey] = capiProviderAgentRBACLabelValue
+		if role.Annotations == nil {
+			role.Annotations = map[string]string{}
+		}
+		role.Annotations[k8sutil.HostedClusterAnnotation] = client.ObjectKeyFromObject(hcluster).String()
+		role.Rules = []rbacv1.PolicyRule{
+			{
+				APIGroups: []string{"agent-install.openshift.io"},
```
jparrill, line 177: Nit: this string also appears in `cmd/install/assets/hypershift_operator.go`. No constant exists for it anywhere — might be worth defining one to keep them in sync.

hypershift-operator/controllers/hostedcluster/internal/platform/agent/agent_test.go
```diff
@@ -57,12 +61,154 @@ func TestReconcileCredentials(t *testing.T) {
...
+	g.Expect(err).ToNot(HaveOccurred())
+	g.Expect(role.Labels[capiProviderAgentRBACLabelKey]).To(Equal(capiProviderAgentRBACLabelValue))
+	g.Expect(role.Annotations[k8sutil.HostedClusterAnnotation]).To(Equal("clusters/test-cluster"))
+	g.Expect(role.Rules).To(HaveLen(1))
+	g.Expect(role.Rules[0].APIGroups).To(Equal([]string{"agent-install.openshift.io"}))
+	g.Expect(role.Rules[0].Resources).To(Equal([]string{"agents"}))
+	g.Expect(role.Rules[0].Verbs).To(Equal([]string{"get", "list", "watch", "update", "patch"}))
+}
+
+func TestReconcileCAPIProviderRole_WhenCalledMultipleTimes_ItShouldBeIdempotent(t *testing.T) {
```
jparrill: Per TESTING.md, this should be a `t.Run` subtest inside `TestReconcileCAPIProviderRole` rather than a separate top-level function — same for `TestDeleteCredentialsWithMultipleClusters` which fits as a subtest of `TestDeleteCredentials`. Use table-driven tests with Gherkin-style names ("When... it should..."). The naming convention on this one is good though.

Also, `t.Parallel()` could be added to all these tests — they use independent fake clients with no shared state.

hypershift-operator/controllers/hostedcluster/internal/platform/agent/agent_test.go
```diff
@@ -27,10 +28,13 @@ import (
 	"github.com/google/go-cmp/cmp"
 )
 
-func TestReconcileCredentials(t *testing.T) {
+func TestReconcileCAPIProviderRole(t *testing.T) {
 	g := NewGomegaWithT(t)
-	platform := &Agent{}
 	hostedCluster := &hyperv1.HostedCluster{
```
jparrill: Minor: the `HostedCluster` object is constructed almost identically 5 times across the test functions. A `newTestHostedCluster(name, agentNamespace string)` helper would reduce the boilerplate and make future spec changes easier.

jparrill:
I’ve seen you’ve addressed the comments thanks for that.

/approve

cwilkers:
/verified by @cwilkers

Created GitOps deployment without pre-existing role. ClusterInstance generated HostedCluster, and role is created quickly afterwards.
Deleted role was recreated quickly.

PR #8306, opened 2026-04-22
OCPBUGS-83514: fix inconsistent error return in endpoint service adoption

sdminonne:
## Summary
- When endpoint service adoption fails in `reconcileAWSEndpointServiceStatus`, the code was returning the outer `CreateVpcEndpointServiceConfiguration` error code (`"InvalidParameter"`) instead of the actual error from `findExistingVpcEndpointService`. This made debugging harder since the logged error and returned error were inconsistent.
- Return the adoption error wrapped with context so callers see the actual failure reason.

## Additional context
This PR addresses item 2 from OCPBUGS-83514. Item 1 (the `//go:generate` directive in `support/awsapi/iam.go`) was investigated and confirmed to be intentionally placed by the `delegatingclientgenerator` template — no code change needed.

Fixes: https://redhat.atlassian.net/browse/OCPBUGS-83514

Ref: [PR #7871 review comment](https://github.com/openshift/hypershift/pull/7871)

## Test plan
- [ ] Verify `make build` passes
- [ ] Verify `make verify` passes

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of VPC endpoint service creation errors: reconciler now preserves underlying failure details while annotating issues with the originating AWS error code, yielding clearer, actionable error messages when adoption fails.

* **Tests**
  * Added unit tests covering VPC endpoint adoption paths and various cloud API error scenarios to validate reconciliation outcomes and error reporting.

### DIFF
--- a/hypershift-operator/controllers/platform/aws/controller.go
+++ b/hypershift-operator/controllers/platform/aws/controller.go
@@ -581,7 +581,7 @@ func (r *AWSEndpointServiceReconciler) reconcileAWSEndpointServiceStatus(ctx con
 					serviceName, serviceID, err = findExistingVpcEndpointService(ctx, ec2Client, aws.ToString(lbARN))
 					if err != nil {
 						log.Info("existing endpoint service not found, adoption failed", "err", err)
-						return errors.New(apiErr.ErrorCode())
+						return fmt.Errorf("endpoint service adoption failed: %v", err)
 					}
 				} else {
 					return errors.New(apiErr.ErrorCode())
--- a/hypershift-operator/controllers/platform/aws/controller_test.go
+++ b/hypershift-operator/controllers/platform/aws/controller_test.go
@@ -20,6 +20,7 @@ import (
 	ec2types "github.com/aws/aws-sdk-go-v2/service/ec2/types"
 	"github.com/aws/aws-sdk-go-v2/service/elasticloadbalancingv2"
 	elbv2types "github.com/aws/aws-sdk-go-v2/service/elasticloadbalancingv2/types"
+	"github.com/aws/smithy-go"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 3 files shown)


hypershift-operator/controllers/platform/aws/controller.go
```diff
@@ -577,7 +577,7 @@ func (r *AWSEndpointServiceReconciler) reconcileAWSEndpointServiceStatus(ctx con
 					serviceName, serviceID, err = findExistingVpcEndpointService(ctx, ec2Client, aws.ToString(lbARN))
 					if err != nil {
 						log.Info("existing endpoint service not found, adoption failed", "err", err)
-						return errors.New(apiErr.ErrorCode())
+						return fmt.Errorf("endpoint service adoption failed: %w", err)
```
mehabhalodiya: One minor suggestion: The message "endpoint service adoption failed" is good, but consider whether it should also include the original apiErr.ErrorCode() for full context, e.g.:
```
return fmt.Errorf("endpoint service adoption failed (trigger: %s): %w", apiErr.ErrorCode(), err)
```
This way the AWS error code that triggered the adoption path is still visible in the error chain. Not a blocker, but useful for debugging.

hypershift-operator/controllers/platform/aws/controller_test.go
```diff
@@ -153,6 +154,121 @@ func TestReconcileAWSEndpointServiceStatus(t *testing.T) {
...
+			fakeClient := fake.NewClientBuilder().WithScheme(hyperapi.Scheme).WithObjects(infra).Build()
+
+			mockEC2 := awsapi.NewMockEC2API(mockCtrl)
+			mockEC2.EXPECT().CreateVpcEndpointServiceConfiguration(gomock.Any(), gomock.Any()).
+				Return(nil, tt.createErr)
+
+			// Set up adoption-related mocks only for InvalidParameter cases.
+			if tt.adoptionDescribeOut != nil || tt.adoptionDescribeErr != nil {
+				mockEC2.EXPECT().DescribeVpcEndpointServiceConfigurations(gomock.Any(), gomock.Any()).
+					Return(tt.adoptionDescribeOut, tt.adoptionDescribeErr)
```
mehabhalodiya, line 276: Looking at lines 241–244, the adoption describe mock is only set up when `tt.adoptionDescribeOut != nil || tt.adoptionDescribeErr != nil`. For the `AccessDenied` case, neither is set, so the describe call won't be mocked. That's correct only if the code path for non-`InvalidParameter` errors does NOT call describe — make sure that branch exits before reaching the adoption describe call. Worth double-checking in the controller logic.

mehabhalodiya (commented):
Overall:

✅ The core error fix is correct and well-targeted. The test suite is solid. One thing to follow up on before approving:

- Confirm the `AccessDenied` test case doesn't accidentally expect a describe mock call that won't be set up.

hypershift-operator/controllers/platform/aws/controller_test.go
```diff
@@ -153,6 +154,121 @@ func TestReconcileAWSEndpointServiceStatus(t *testing.T) {
 	}
 }
 
+func TestReconcileAWSEndpointServiceStatusCreationErrors(t *testing.T) {
```
jparrill, line 189: It's possible to merge with `TestReconcileAWSEndpointServiceStatus` test function?

jparrill (commented):
/approve

Dropped a comment, otherwise lgtm

hypershift-operator/controllers/platform/aws/controller_test.go
sdminonne: Hmmm, good question. Looking

hypershift-operator/controllers/platform/aws/controller_test.go
sdminonne: Thanks for checking @mehabhalodiya   
The test  does not set up a describe mock, and gomock would panic if an unexpected call were made.

hypershift-operator/controllers/platform/aws/controller_test.go
sdminonne: No, it can't mocking setup is pretty different

hypershift-operator/controllers/platform/aws/controller.go
sdminonne: Right, doing

sdminonne:
@mehabhalodiya thanks! PTAL

mehabhalodiya (approved):
/lgtm
/approve

Thank you!

sdminonne:
<markdown-accessiblity-table data-catalyst="">
/test verify-workflows
--


</markdown-accessiblity-table>
<p dir="auto"><a href="https://prow.ci.openshift.org/pr-history?org=openshift&amp;repo=hypershift&amp;pr=8306" rel="nofollow"></a></p>

hypershift-operator/controllers/platform/aws/controller.go
```diff
@@ -581,7 +581,7 @@ func (r *AWSEndpointServiceReconciler) reconcileAWSEndpointServiceStatus(ctx con
 					serviceName, serviceID, err = findExistingVpcEndpointService(ctx, ec2Client, aws.ToString(lbARN))
 					if err != nil {
 						log.Info("existing endpoint service not found, adoption failed", "err", err)
-						return errors.New(apiErr.ErrorCode())
+						return fmt.Errorf("endpoint service adoption failed (trigger: %s): %w", apiErr.ErrorCode(), err)
```
enxebre: this error is bubbled up in a condition message, if the message contains contains variable output (e.g. aws request id) the condition would flip in loop.
/hold

enxebre: feel free to unhold if that's not the case

sdminonne: Thanks going to double check!

sdminonne: @enxebre modified this and added the AGENTS.md for future interactions. May you have a look when you've time?

enxebre:
lgtm, will defer to others to tag in case they have more feedback

hypershift-operator/controllers/platform/aws/controller.go
```diff
@@ -581,7 +581,7 @@ func (r *AWSEndpointServiceReconciler) reconcileAWSEndpointServiceStatus(ctx con
 					serviceName, serviceID, err = findExistingVpcEndpointService(ctx, ec2Client, aws.ToString(lbARN))
 					if err != nil {
 						log.Info("existing endpoint service not found, adoption failed", "err", err)
-						return errors.New(apiErr.ErrorCode())
+						return fmt.Errorf("endpoint service adoption failed (trigger: %s): failed to find existing endpoint service", apiErr.ErrorCode())
```
muraee: shouldn't `err` be included here instead of `apiErr.ErrorCode()`?

sdminonne: Good point double checking

sdminonne: Thanks for this @muraee ! I looked into it and AFAIK the err comes from [serviceName, serviceID, err = findExistingVpcEndpointService(ctx, ec2Client, aws.ToString(lbARN))](https://github.com/openshift/hypershift/pull/8306/changes#diff-703bbd22baf7c689e315f5c561efed7afea4f14d0ade48d66af784206a0bc421R581) and that [err may contain the err ID from AWs ](https://github.com/openshift/hypershift/pull/8306/changes#diff-703bbd22baf7c689e315f5c561efed7afea4f14d0ade48d66af784206a0bc421R658) and this is may trigger re-conciliation since the err is bubble-up in condition. 
It's true that a static message doesn't add a lot of info but resync should be avoided. 
I don't have a strong opinion but I would keep as is
Thoughts?

muraee: [findExistingVpcEndpointService](https://github.com/openshift/hypershift/blob/main/hypershift-operator/controllers/platform/aws/controller.go#L656) returns parsses the aws error and returns only the ErrorCode. It doesn't include any err ID

sdminonne: Right, taking back and amending the code. TY!

hypershift-operator/controllers/platform/aws/controller.go
Nirshal, line 658: I have a doubt here: 
- this row returns the error in its raw form, without "sanitization". Failing the describe could still happens for non-api related issues like network error, right? Do those kind of errors can have a variable part that can again induce flapping? 
- if we apply the AGENTS file suggestion here, we log and then we return a %T of the error, but then at row 583 we log again the "sanitized" version, and that is misleading (at first glance they may seem 2 different problems).
- If that's not the case, and we want to keep the code as it is now, we should care to have a way to indicate to the agents that in this particular case we do not want to follow the rule indicated by AGENTS.md, otherwise they are going to flag this as an inconsistency and try to correct this.

hypershift-operator/controllers/platform/aws/controller.go
```diff
@@ -581,7 +581,7 @@ func (r *AWSEndpointServiceReconciler) reconcileAWSEndpointServiceStatus(ctx con
 					serviceName, serviceID, err = findExistingVpcEndpointService(ctx, ec2Client, aws.ToString(lbARN))
 					if err != nil {
 						log.Info("existing endpoint service not found, adoption failed", "err", err)
-						return errors.New(apiErr.ErrorCode())
+						return fmt.Errorf("endpoint service adoption failed: %v", err)
```
Nirshal, line 584: Reading the conversation, it makes sense to me to say that because we "sanitize" all errors before this point we are safe to use %v (although I found an exception to this at row 658), but the AGENTS.md indicates to always log the full error (and that is covered by row 583) and return the summarized version. This may trigger the AI to further change this row next time it is required to work on this file. If we think we want to keep it like this, should we find a way to flag the intention to keep an exception to the rule?

Nirshal (commented):
It seems fine to me, I just have a couple of doubts that are probably coming from my lack of experience, if you would be so nice to explain them to me, I would gladly approve the PR.

hypershift-operator/controllers/platform/aws/controller.go
sdminonne: My understanding is that only AWS errs need to be sanitized. AWS errs contains a variable ID. Same err but different ID. If we propagate it to the condition we may have an unwanted re-sync.
@muraee please confirm or educate me in case I'm wrong.

sdminonne (commented):
OK for me overall... I would add an happy path test to the existing unit tests

PR #6001, opened 2025-04-08
OCPBUGS-41853: Add documentation for API Server custom certificate in HCP

jparrill:
## What this PR does / why we need it
Included 2 new sections explaining how to configure a custom KAS certificate in a HostedCluster and also How to combine this with a custom KAS dns to be used into a kubeconfig. All the details regarding this last point, are explained in [OCPSTRAT-1516](https://issues.redhat.com/browse/OCPSTRAT-1516)

CleanShot 2025-04-08 at 13 02 19

## Which issue(s) this PR fixes
Fixes #[OCPBUGS-41853](https://issues.redhat.com/browse/OCPBUGS-41853)
Related:
- [OCPBUGS-53261](https://issues.redhat.com/browse/OCPBUGS-53261)
- PR: https://github.com/openshift/hypershift/pull/5875

Issue #3985, opened 2024-05-05
Unable to Register New Nodes in HostedCluster When Adding API Server Certificate


alfredtm:
Hello! When adding API server certificate to the HostedCluster

```
apiVersion: hypershift.openshift.io/v1beta1
kind: HostedCluster
metadata:
  name: test
  namespace: test
spec:
  configuration:
    apiServer:
      servingCerts:
        namedCertificates:
        - servingCertificate:
            name: tls-secret
``` 
The bootstrap-kubeconfig is no longer trusted

``` 
Unable to connect to the server: tls: failed to verify certificate: x509: certificate signed by unknown authority (possibly because of "crypto/rsa: verification error" while trying to verify candidate authority certificate "root-ca")
``` 
This seems to be a issue when adding new nodes (using kubevirt) to the cluster. They dont register/show up on the HostedCluster. 

When not adding certificate to the apiServer the nodes register fine.

alfredtm:
I resolved this by setting `servicePublishingStrategy.loadBalancer.hostname` to the apiServer's internal service URL. 

The certificate-authority-data in bootstrap-kubeconfig is then trusted.

alfredtm:
Reopening this issue 
We are experiencing issues using the apiserver internal service url at a later stage in our deployment. 

Is there any way to modify the bootstrap-kubeconfig?  

I also found this issue that seems to be the same issue I am experiencing
https://issues.redhat.com/browse/OCPBUGS-19067

KiqoCode:
We have the same issue..

omlet05:
Same issue here, does anyone solved it?

omlet05:
@jparrill  Same with a brand new cluster or an update to an existing one.

alfredtm:
@jparrill This happens with newly created clusters. And existing clusters when adding APIServerCertificate at a later stage.

jparrill:
Thanks for the info folks, that really helped. I will try to take a look to the issue.

KiqoCode:
Hello,

There is an open Jira with Redhat now I know which is planned to be fixed in 4.18 I think.
For now it seems custom API Certificate is not supported on HostedCluster. 
Documentation have also been updated I know:
https://docs.openshift.com/container-platform/4.16/security/certificates/api-server.html
"
In hosted control plane clusters, you cannot replace self-signed certificates from the API.
--
"

PR #8311, opened 2026-04-23
OCPBUGS-84555: fix(cpo): include serving cert CAs in bootstrap and external admin kubeconfigs

amasolov:
## Summary

When using Route-based API server publishing with `namedCertificates` (i.e. the KAS serving certificate is signed by an external/corporate CA), two kubeconfigs were generated with only the HyperShift internal `root-ca` in their `certificate-authority-data`:

1. **Bootstrap kubeconfig** (`adaptBootstrapKubeconfigSecret`): used by worker kubelets to bootstrap into the cluster. This caused `x509: certificate signed by unknown authority` failures during node join.
2. **External admin kubeconfig** (`adapExternalAdminKubeconfigSecret`): the default admin kubeconfig generated for clusters that do not set `kubeAPIServerDNSName`. Clients (oc/kubectl) using this kubeconfig would also fail TLS verification.

The custom admin kubeconfig (`adaptCustomAdminKubeconfigSecret`, enabled only when `kubeAPIServerDNSName` is set) already used `combineRootCAWithServingCerts` and worked correctly.

## Fix

Apply the same `combineRootCAWithServingCerts` call to both affected functions, aligning trust across all generated kubeconfigs. When no `namedCertificates` are configured, behaviour is unchanged (only `root-ca` is included).

## Test plan

- [x] `go build ./...` passes
- [x] `go vet ./control-plane-operator/controllers/hostedcontrolplane/v2/kas/...` passes
- [x] `go test ./control-plane-operator/controllers/hostedcontrolplane/v2/kas/...` passes
- Manually verified on a lab HostedCluster with Route-based publishing and corporate CA: worker node joins successfully after bootstrap kubeconfig includes the serving cert CA chain.

## Summary by CodeRabbit

* **Bug Fixes**
  * Admin and bootstrap kubeconfigs now include serving certificates in the combined root CA bundle, improving trust coverage.
  * More specific error messages when merging root CAs or retrieving certificate secrets for clearer diagnostics.
  * Kubeconfig creation now consistently ensures client certificate secrets are retrieved before generating and storing the kubeconfig.

* **Refactor**
  * Centralized kubeconfig generation to remove duplicated logic and streamline secret handling.

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/v2/kas/kubeconfig.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/v2/kas/kubeconfig.go
@@ -112,6 +112,20 @@ func adaptKASBootstrapContainerKubeconfigSecret(cpContext component.WorkloadCont
 	return nil
 }
 
+// generateKubeConfigWithServingCerts builds a kubeconfig that trusts both the root CA and any
+// additional serving certificate CAs from namedCertificates. It fetches the combined CA bundle
+// and the specified client certificate secret, then generates the kubeconfig bytes.
+func generateKubeConfigWithServingCerts(cpContext component.WorkloadContext, certSecret *corev1.Secret, url string) ([]byte, error) {
+	totalRootCA, err := combineRootCAWithServingCerts(cpContext)
+	if err != nil {
+		return nil, fmt.Errorf("failed to include serving certificates: %w", err)
+	}
+	if err := cpContext.Client.Get(cpContext, client.ObjectKeyFromObject(certSecret), certSecret); err != nil {
+		return nil, fmt.Errorf("failed to get client cert secret %s: %w", certSecret.Name, err)
+	}
+	return generateKubeConfig(totalRootCA, certSecret, url)
+}
+
 func adapExternalAdminKubeconfigSecret(cpContext component.WorkloadContext, secret *corev1.Secret) error {
 	if cpContext.HCP.Spec.KubeConfig != nil {
 		secret.Name = cpContext.HCP.Spec.KubeConfig.Name
@@ -122,7 +136,8 @@ func adapExternalAdminKubeconfigSecret(cpContext component.WorkloadContext, secr
 	if !util.IsPublicHCP(cpContext.HCP) && !util.IsRouteKAS(cpContext.HCP) {
 		url = internalURL(cpContext.InfraStatus, cpContext.HCP.Name)
 	}
-	kubeconfig, err := GenerateKubeConfig(cpContext, manifests.SystemAdminClientCertSecret(cpContext.HCP.Namespace), url)
+
+	kubeconfig, err := generateKubeConfigWithServingCerts(cpContext, manifests.SystemAdminClientCertSecret(cpContext.HCP.Namespace), url)
 	if err != nil {
 		return fmt.Errorf("failed to generate kubeconfig: %w", err)
 	}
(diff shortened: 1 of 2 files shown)


amasolov:
## Prior Art & Context

This bug was previously reported but only received a documentation fix:

- **Upstream issue:** #3985 (closed by bot inactivity, no code fix)
- **Jira:** [OCPBUGS-41853](https://issues.redhat.com/browse/OCPBUGS-41853) (closed as "Done" in 4.19.0, resolution was docs PR #6001 only)
- **KCS:** [7111089](https://access.redhat.com/solutions/7111089) (describes the symptom)

The earlier documentation workaround (KCS [7135847](https://access.redhat.com/solutions/7135847)) uses `kubeAPIServerDNSName` + `namedCertificates` to generate a trusted admin kubeconfig. That path works for admin access but does not fix the bootstrap kubeconfig used by worker kubelets, so **worker nodes still fail to join** when a corporate CA is the trust anchor for the Route-based KAS endpoint.

This PR fixes the root cause by having `adaptBootstrapKubeconfigSecret` call `combineRootCAWithServingCerts`, matching the existing pattern in `adaptCustomAdminKubeconfigSecret`.

amasolov:
Re: the `combineRootCAWithServingCerts` concern about appending leaf certificates rather than CA certificates.

This function pre-dates this PR (it was already used by `adaptCustomAdminKubeconfigSecret` on `main`). The design is intentional and correct in practice:

1. **Real-world deployments**: the `tls.crt` field in named certificate secrets conventionally contains the full PEM chain (leaf + intermediates + signing CA), as per OpenShift documentation for `namedCertificates`. Appending the full chain to the trust bundle gives clients the CA they need to verify the serving certificate.

2. **E2e tests**: the test generates a self-signed certificate (`IsCA: false`), which acts as its own trust anchor. Go's `x509.Verify` trusts it when it appears in the root pool regardless of the `IsCA` basic constraint.

I've added a clarifying comment to the function in this latest push to document this expectation.

control-plane-operator/controllers/hostedcontrolplane/v2/kas/kubeconfig.go
```diff
@@ -111,6 +111,20 @@ func adaptKASBootstrapContainerKubeconfigSecret(cpContext component.WorkloadCont
 	return nil
 }
 
+// generateKubeConfigWithServingCerts builds a kubeconfig that trusts both the root CA and any
+// additional serving certificate CAs from namedCertificates. It fetches the combined CA bundle
+// and the specified client certificate secret, then generates the kubeconfig bytes.
+func generateKubeConfigWithServingCerts(cpContext component.WorkloadContext, certSecret *corev1.Secret, url string) ([]byte, error) {
```
bryan-cox, line 118: Can we get a unit test for this please?

amasolov: @bryan-cox done, thank you for looking into this

control-plane-operator/controllers/hostedcontrolplane/v2/kas/kubeconfig_test.go
bryan-cox, line 1: Can you update the tests to gerkin format please?

amasolov: Done

bryan-cox (approved):
/lgtm

PR #8312, opened 2026-04-23
AROSLSRE-687: propagate HCP Azure resource ID annotation to control plane namespace

ashishmax31:
Add a new annotation constant HCPAzureResourceIDAnnotation that carries
the Azure resource ID set by Cluster Service on the HostedCluster CR.
The hostedcluster controller now propagates this annotation to the
 hosted control plane namespace for Azure platform clusters.

Fixes 
https://redhat.atlassian.net/browse/AROSLSRE-687

## Checklist:
- [x ] Relevant issues have been referenced.

## Summary by CodeRabbit

* **New Features**
  * Azure HostedClusters now propagate an Azure resource ID annotation into their hosted control plane namespace for Azure-based clusters; the annotation is added when present and removed when absent, preserving other namespace annotations.

* **Tests**
  * Added unit tests covering propagation, removal, non-Azure behavior, and preservation of existing namespace annotations.

### DIFF
--- a/api/hypershift/v1beta1/hostedcluster_types.go
+++ b/api/hypershift/v1beta1/hostedcluster_types.go
@@ -78,6 +78,11 @@ const (
 	// AWSKMSProviderImage is an annotation that allows the specification of the AWS kms provider image.
 	// Upstream code located at: https://github.com/kubernetes-sigs/aws-encryption-provider
 	AWSKMSProviderImage = "hypershift.openshift.io/aws-kms-provider-image"
+	// ManagedAzureResourceIDAnnotation is an annotation set by Cluster Service on the HostedCluster CR
+	// containing the Azure resource ID. It is propagated to the hosted control plane namespace.
+	// This annotation is consumed by ARO-HCP logging and observability components to correlate the
+	// HostedCluster with the corresponding Azure resources.
+	ManagedAzureResourceIDAnnotation = "azure.microsoft.com/hcp-cluster-azure-resource-id"
 	// IBMCloudKMSProviderImage is an annotation that allows the specification of the IBM Cloud kms provider image.
 	IBMCloudKMSProviderImage = "hypershift.openshift.io/ibmcloud-kms-provider-image"
 	// PortierisImageAnnotation is an annotation that allows the specification of the portieries component
--- a/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
+++ b/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
@@ -1426,6 +1426,8 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
 		// Enable observability operator monitoring
 		metrics.EnableOBOMonitoring(controlPlaneNamespace)
 
+		propagateAzureResourceIDAnnotation(hcluster, controlPlaneNamespace)
+
 		return nil
 	})
 	if err != nil {
@@ -5410,6 +5412,22 @@ func computeGCPPSCCondition(gcpPSCList hyperv1.GCPPrivateServiceConnectList, con
 		resourceConditions[i] = psc.Status.Conditions
 	}
 	return computeEndpointServiceCondition(resourceConditions, conditionType, hyperv1.GCPErrorReason, hyperv1.GCPSuccessReason, "GCPPrivateServiceConnect conditions not found")
+}
+
+// propagateAzureResourceIDAnnotation copies the Azure resource ID annotation set by Cluster Service
+// on the HostedCluster CR to the control plane namespace, or removes it if no longer present.
+// This annotation is consumed by ARO-HCP logging and observability components to correlate the
+// HostedCluster with the corresponding Azure resources.
+func propagateAzureResourceIDAnnotation(hcluster *hyperv1.HostedCluster, ns *corev1.Namespace) {
+	resourceID := hcluster.Annotations[hyperv1.ManagedAzureResourceIDAnnotation]
+	if azureutil.IsAroHCP() && resourceID != "" {
+		if ns.Annotations == nil {
+			ns.Annotations = make(map[string]string)
+		}
+		ns.Annotations[hyperv1.ManagedAzureResourceIDAnnotation] = resourceID
+	} else {
+		delete(ns.Annotations, hyperv1.ManagedAzureResourceIDAnnotation)
+	}
 }
 
 func computeAzurePLSCondition(azPLSList hyperv1.AzurePrivateLinkServiceList, conditionType hyperv1.ConditionType) metav1.Condition {
--- a/hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
+++ b/hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
@@ -7657,3 +7657,107 @@ func TestReconcileETCDMemberRecovery(t *testing.T) {
 		g.Expect(err.Error()).To(ContainSubstring("failed to get etcd statefulset"))
 	})
 }
+
+func TestPropagateAzureResourceIDAnnotation(t *testing.T) {
+	const testResourceID = "/subscriptions/00000000/resourceGroups/rg/providers/Microsoft.RedHatOpenShift/hcpOpenShiftClusters/cluster1"
+
+	tests := []struct {
+		name               string
+		isAroHCP           bool
+		hcluster           *hyperv1.HostedCluster
+		nsAnnotations      map[string]string
+		expectedAnnotation string
+	}{
+		{
+			name:     "When ARO-HCP cluster has resource ID annotation, it should propagate to namespace",
+			isAroHCP: true,
+			hcluster: &hyperv1.HostedCluster{
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations: map[string]string{
+						hyperv1.ManagedAzureResourceIDAnnotation: testResourceID,
+					},
+				},
+			},
+			nsAnnotations:      nil,
+			expectedAnnotation: testResourceID,
+		},
+		{
+			name:     "When ARO-HCP cluster has no resource ID annotation, it should remove it from namespace",
+			isAroHCP: true,
+			hcluster: &hyperv1.HostedCluster{
+				ObjectMeta: metav1.ObjectMeta{},
+			},
+			nsAnnotations: map[string]string{
+				hyperv1.ManagedAzureResourceIDAnnotation: testResourceID,
+			},
+			expectedAnnotation: "",
+		},
+		{
+			name: "When non-ARO-HCP cluster has resource ID annotation, it should not propagate to namespace",
+			hcluster: &hyperv1.HostedCluster{
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations: map[string]string{
+						hyperv1.ManagedAzureResourceIDAnnotation: testResourceID,
+					},
+				},
+			},
+			nsAnnotations:      nil,
+			expectedAnnotation: "",
+		},
+		{
+			name:     "When non-ARO-HCP cluster has resource ID on namespace, it should remove it",
+			hcluster: &hyperv1.HostedCluster{},
+			nsAnnotations: map[string]string{
+				hyperv1.ManagedAzureResourceIDAnnotation: testResourceID,
+			},
+			expectedAnnotation: "",
+		},
+		{
+			name:     "When ARO-HCP cluster has resource ID annotation, it should update existing namespace annotation",
+			isAroHCP: true,
+			hcluster: &hyperv1.HostedCluster{
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations: map[string]string{
+						hyperv1.ManagedAzureResourceIDAnnotation: testResourceID,
+					},
+				},
+			},
+			nsAnnotations: map[string]string{
+				hyperv1.ManagedAzureResourceIDAnnotation: "old-value",
+				"some-other-annotation":                  "preserved",
+			},
+			expectedAnnotation: testResourceID,
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			if tc.isAroHCP {
+				azureutil.SetAsAroHCPTest(t)
+			}
+			g := NewGomegaWithT(t)
+			ns := &corev1.Namespace{
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations: tc.nsAnnotations,
+				},
+			}
+
+			propagateAzureResourceIDAnnotation(tc.hcluster, ns)
+
+			if tc.expectedAnnotation != "" {
+				g.Expect(ns.Annotations).To(HaveKeyWithValue(
+					hyperv1.ManagedAzureResourceIDAnnotation, tc.expectedAnnotation,
+				))
+			} else {
+				g.Expect(ns.Annotations).ToNot(HaveKey(hyperv1.ManagedAzureResourceIDAnnotation))
+			}
+
+			// Verify other annotations are not disturbed
+			for k, v := range tc.nsAnnotations {
+				if k != hyperv1.ManagedAzureResourceIDAnnotation {
+					g.Expect(ns.Annotations).To(HaveKeyWithValue(k, v))
+				}
+			}
+		})
+	}
+}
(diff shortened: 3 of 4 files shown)


JoelSpeed:
/approve

For API

jparrill:
/retitle [AROSLSRE-687](https://redhat.atlassian.net/browse/AROSLSRE-687): propagate HCP Azure resource ID annotation to control plane namespace

hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
```diff
@@ -6529,3 +6529,104 @@ func TestComputeEndpointServiceCondition(t *testing.T) {
...
+
+			// Verify other annotations are not disturbed
+			for k, v := range tc.nsAnnotations {
+				if k != hyperv1.HCPAzureResourceIDAnnotation {
+					g.Expect(ns.Annotations).To(HaveKeyWithValue(k, v))
+				}
+			}
+		})
+	}
+}
```
jparrill, line 7763: Nit: missing `t.Parallel()` in subtests.

jparrill (commented):
Dropped a comment. Thanks!

jparrill:
You will need to force push in order to retest the env test for KAS.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
ashishmax31: done

jparrill (commented):
Dropped some comments. Thanks

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -5318,6 +5320,22 @@ func computeGCPPSCCondition(gcpPSCList hyperv1.GCPPrivateServiceConnectList, con
 	return computeEndpointServiceCondition(resourceConditions, conditionType, hyperv1.GCPErrorReason, hyperv1.GCPSuccessReason, "GCPPrivateServiceConnect conditions not found")
 }
 
+// propagateAzureResourceIDAnnotation copies the Azure resource ID annotation set by Cluster Service
```
jparrill, line 5417: The comment says what the function does but not who actually reads this annotation from the namespace. I searched the repo and couldn't find any consumer — is it ARO SRE tooling or something external? Worth adding a note so nobody accidentally removes it later thinking it's unused.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -1397,6 +1397,8 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
 		// Enable observability operator monitoring
 		metrics.EnableOBOMonitoring(controlPlaneNamespace)
 
+		propagateAzureResourceIDAnnotation(hcluster, controlPlaneNamespace)
```
jparrill, line 1429: Curious — why the namespace instead of the HostedControlPlane? All the other platform annotations go through `mirroredAnnotations` in `reconcileHostedControlPlaneAnnotations` (~line 2292). If the consumer really needs it on the namespace that's fine, but it'd be a one-liner to also add it to the HCP. Thoughts?

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -5318,6 +5320,22 @@ func computeGCPPSCCondition(gcpPSCList hyperv1.GCPPrivateServiceConnectList, con
 	return computeEndpointServiceCondition(resourceConditions, conditionType, hyperv1.GCPErrorReason, hyperv1.GCPSuccessReason, "GCPPrivateServiceConnect conditions not found")
 }
 
+// propagateAzureResourceIDAnnotation copies the Azure resource ID annotation set by Cluster Service
+// on the HostedCluster CR to the control plane namespace, or removes it if no longer present.
+func propagateAzureResourceIDAnnotation(hcluster *hyperv1.HostedCluster, ns *corev1.Namespace) {
+	if hcluster.Spec.Platform.Type != hyperv1.AzurePlatform {
```
jparrill: Tiny thing: if a namespace somehow already has this annotation but the cluster isn't Azure, the early return leaves it sitting there forever. Platform type is immutable in practice so it's not really a bug, but a `delete` before the return would be strictly cleaner. Up to you.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
```diff
@@ -6734,3 +6734,105 @@ func TestComputeEndpointServiceCondition(t *testing.T) {
...
+				"some-other-annotation":              "preserved",
+			},
+			expectedAnnotation: testResourceID,
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			g := NewWithT(t)
```
jparrill: nit: the rest of this file uses `NewGomegaWithT(t)` — would keep it consistent.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
```diff
@@ -6734,3 +6734,105 @@ func TestComputeEndpointServiceCondition(t *testing.T) {
...
+						hyperv1.HCPAzureResourceIDAnnotation: testResourceID,
+					},
+				},
+				Spec: hyperv1.HostedClusterSpec{
+					Platform: hyperv1.PlatformSpec{Type: hyperv1.AWSPlatform},
+				},
+			},
+			nsAnnotations:      nil,
+			expectedAnnotation: "",
+		},
```
jparrill, line 7706: Could you add a case where a non-Azure namespace already has the annotation? Right now the early return means it stays, and having a test for that makes the behavior explicit. Something like:
```go
{
    name: "When non-Azure cluster has resource ID on namespace, it should leave it untouched",
    hcluster: &hyperv1.HostedCluster{
        Spec: hyperv1.HostedClusterSpec{
            Platform: hyperv1.PlatformSpec{Type: hyperv1.AWSPlatform},
        },
    },
    nsAnnotations: map[string]string{
        hyperv1.HCPAzureResourceIDAnnotation: testResourceID,
    },
    expectedAnnotation: "",
},
```

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
ashishmax31: We need it in the namespace for fluent-bit to enrich the logs with the resourceID.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
ashishmax31: ack, will change

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
ashishmax31: ok i will add it. It will be consumed by our logging tooling to annotate the logs coming in from the hosted cluster with the azure resource id. These logs are then forwarded to downstream customer destinations.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
ashishmax31: Ok done

ashishmax31:
@jparrill Addressed the review comments. PTAL

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -78,6 +78,9 @@ const (
 	// AWSKMSProviderImage is an annotation that allows the specification of the AWS kms provider image.
 	// Upstream code located at: https://github.com/kubernetes-sigs/aws-encryption-provider
 	AWSKMSProviderImage = "hypershift.openshift.io/aws-kms-provider-image"
+	// HCPAzureResourceIDAnnotation is an annotation set by Cluster Service on the HostedCluster CR
+	// containing the Azure resource ID. It is propagated to the hosted control plane namespace.
+	HCPAzureResourceIDAnnotation = "azure.microsoft.com/hcp-cluster-azure-resource-id"
```
bryan-cox: Can we name this `ManagedAzureResourceIDAnnotation`? We also have self managed Azure HCP. I don't think we would ever use this for it.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -5368,6 +5320,26 @@ func computeGCPPSCCondition(gcpPSCList hyperv1.GCPPrivateServiceConnectList, con
 	return computeEndpointServiceCondition(resourceConditions, conditionType, hyperv1.GCPErrorReason, hyperv1.GCPSuccessReason, "GCPPrivateServiceConnect conditions not found")
 }
 
+// propagateAzureResourceIDAnnotation copies the Azure resource ID annotation set by Cluster Service
+// on the HostedCluster CR to the control plane namespace, or removes it if no longer present.
+// This annotation is consumed by ARO-HCP logging and observability components to correlate the
+// HostedCluster with the corresponding Azure resources.
+func propagateAzureResourceIDAnnotation(hcluster *hyperv1.HostedCluster, ns *corev1.Namespace) {
+	if hcluster.Spec.Platform.Type != hyperv1.AzurePlatform {
```
bryan-cox: We should use the IsAroHcp check here or whatever is replacing it. Actually why not just move that check up on L1400?

cc @muraee

ashishmax31: @jparrill asked to handle this case too: https://github.com/openshift/hypershift/pull/8312#discussion_r3279541034
> Tiny thing: if a namespace somehow already has this annotation but the cluster isn't Azure, the early return leaves it sitting there forever. Platform type is immutable in practice so it's not really a bug, but a `delete` before the return would be strictly cleaner. Up to you.


If we move the check up then it'll be ugly to add this check

ashishmax31: There is no `IsAroHcp` method on the HostedCluster

api/hypershift/v1beta1/hostedcluster_types.go
ashishmax31: Yes, will rename

ashishmax31: Done

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
bryan-cox: There is. We use it all over the place. Here's an example - https://github.com/openshift/hypershift/blob/e25a87a75adb04dc80e877afcec761774baae36d/hypershift-operator/controllers/hostedcluster/internal/platform/azure/azure.go#L169

ashishmax31: Done

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -5401,6 +5384,26 @@ func computeGCPPSCCondition(gcpPSCList hyperv1.GCPPrivateServiceConnectList, con
 	return computeEndpointServiceCondition(resourceConditions, conditionType, hyperv1.GCPErrorReason, hyperv1.GCPSuccessReason, "GCPPrivateServiceConnect conditions not found")
 }
 
+// propagateAzureResourceIDAnnotation copies the Azure resource ID annotation set by Cluster Service
+// on the HostedCluster CR to the control plane namespace, or removes it if no longer present.
+// This annotation is consumed by ARO-HCP logging and observability components to correlate the
+// HostedCluster with the corresponding Azure resources.
+func propagateAzureResourceIDAnnotation(hcluster *hyperv1.HostedCluster, ns *corev1.Namespace) {
```
bryan-cox, line 5421: Can't this be simplified to just see if it is ARO HCP and then either set it or delete the annotation?

ashishmax31: No.. we want to get it from the hostedcluster CR. Cluster service sets it on the CR. (the value of the annotation is something only CS knows)

bryan-cox: Yeah I know it comes from CS. Let me explain a little further. I meant simplified it to something like:

```
  func propagateAzureResourceIDAnnotation(hcluster *hyperv1.HostedCluster, ns *corev1.Namespace) {
        resourceID := hcluster.Annotations[hyperv1.ManagedAzureResourceIDAnnotation]
        if azureutil.IsAroHCP() && resourceID != "" {
                if ns.Annotations == nil {
                        ns.Annotations = make(map[string]string)
                }
                ns.Annotations[hyperv1.ManagedAzureResourceIDAnnotation] = resourceID
        } else {
                delete(ns.Annotations, hyperv1.ManagedAzureResourceIDAnnotation)
        }
  }
```

Same behavior, still reads the value from the HC CR — just one delete path instead of two and no early return. Does that make sense?

bryan-cox: Or maybe better yet, check before calling the function instead of every time.

```
if azureutil.IsAroHCP() {
     propagateAzureResourceIDAnnotation(hcluster, controlPlaneNamespace)
}
```

We need to be considerate that self managed Azure exists and does not need to run ARO HCP specific things.

ashishmax31: Ok done.. 
```
  func propagateAzureResourceIDAnnotation(hcluster *hyperv1.HostedCluster, ns *corev1.Namespace) {
        resourceID := hcluster.Annotations[hyperv1.ManagedAzureResourceIDAnnotation]
        if azureutil.IsAroHCP() && resourceID != "" {
                if ns.Annotations == nil {
                        ns.Annotations = make(map[string]string)
                }
                ns.Annotations[hyperv1.ManagedAzureResourceIDAnnotation] = resourceID
        } else {
                delete(ns.Annotations, hyperv1.ManagedAzureResourceIDAnnotation)
        }
  }
  ```
  I went with this.. That helps us unit test all the scenarios cleanly.

**Review by bryan-cox — APPROVED** [review]

**Comment by bryan-cox** [comment]

This job is permafailing and I pinged the kubevirt folks earlier here - https://redhat-internal.slack.com/archives/C0B71V5TDMW/p1781616231915999

File: api/hypershift/v1beta1/hostedcluster_types.go [hunk 10]
```diff
@@ -78,6 +78,9 @@ const (
 	// AWSKMSProviderImage is an annotation that allows the specification of the AWS kms provider image.
 	// Upstream code located at: https://github.com/kubernetes-sigs/aws-encryption-provider
 	AWSKMSProviderImage = "hypershift.openshift.io/aws-kms-provider-image"
+	// ManagedAzureResourceIDAnnotation is an annotation set by Cluster Service on the HostedCluster CR
+	// containing the Azure resource ID. It is propagated to the hosted control plane namespace.
```
enxebre, line 82: A resource ID for which resource? the arm cluster resource?

enxebre: can we please document why is propagated to the control plane namespace?

enxebre: I'm curious if we have considered using the .ClusterID to input this value?

ashishmax31: >A resource ID for which resource? the arm cluster resource?

Yes ARM cluster resource

ashishmax31: >can we please document why is propagated to the control plane namespace?

It is written down here
https://github.com/ashishmax31/hypershift/blob/ba9711af5865b885d5f72b0a931afea7362a0120/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go#L5417

ashishmax31: >I'm curious if we have considered using the .ClusterID to input this value?

i did not understand.

ashishmax31:
demo

bryan-cox (approved):
/lgtm

PR #8322, opened 2026-04-23
OCPBUGS-84251: fix(azure): detect and replace stale role assignments on cluster re-creation

bryan-cox:

When a self-managed Azure cluster is destroyed and re-created with the same infraID, role
assignments from the previous cluster persist as orphans because neither `destroy iam`,
`destroy infra`, nor `destroy cluster` cleans them up. The deterministic UUID naming
(based on infraID+component+scope) causes `assignRole` to find the stale assignment by name
and skip creation, even though it points to a deleted managed identity. This results in
**403 AuthorizationFailed** errors for components like the ingress operator that need RBAC
roles on external resource groups (e.g., the DNS zone resource group).

### Root cause

Two bugs working together:

1. **`assignRole` GET check doesn't verify principal**: The deterministic role assignment name
   (`uuid(infraID+component+scope)`) matches a stale/orphaned assignment from a previous cluster.
   The GET check finds it by name and skips creation without verifying the `principalID` matches
   the new identity.

2. **No destroy path cleans up role assignments**: When you destroy a cluster, the managed
   identities are deleted but their role assignments persist as orphans in Azure RBAC.

### The failure sequence

1. `create infra --infra-id=XXX` → creates role assignment `uuid(XXX-ingress-scope)` pointing to identity **A**
2. `destroy infra` / `destroy iam` / `destroy cluster` → deletes identity **A** but leaves orphaned role assignment
3. `create iam` (new attempt, same name) → creates NEW identity **B** (different objectID)
4. `create infra --infra-id=XXX --assign-identity-roles` →
   - LIST check: finds orphan for **A**, `principalID != B` → doesn't match → continues
   - GET check: `uuid(XXX-ingress-scope)` → FINDS orphan → "already exists" → **SKIPS**
5. Identity **B** has NO role on dnsZoneRG → **403 AuthorizationFailed**

### Fix

1. **`assignRole`** now verifies `principalID` when GET finds an existing assignment. If
   mismatched (stale orphan), it deletes the stale assignment and creates a new one.

2. **`destroy iam`** now calls `CleanupRoleAssignments` before destroying managed identities,
   preventing orphaned role assignments from accumulating.

3. **`destroy cluster azure`** now calls `CleanupRoleAssignments` before destroying infrastructure,
   so both destroy paths clean up role assignments.

4. Extracted `roleAssignmentClient` interface from the Azure SDK concrete client to enable
   unit testing of the stale detection logic.

### Verified with `az` CLI

Created a test identity, assigned a role with a deterministic UUID name, deleted the identity
(leaving the role assignment orphaned), created a new identity — confirmed the GET returns
the stale `principalId` which our fix now detects and replaces.

Fixes OCPBUGS-84251

- The `--dns-zone-rg-name` flag is optional on all destroy commands; if not provided, DNS zone
  scoped assignments simply get 404 during cleanup (handled gracefully).

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * CLI flag to specify DNS zone resource group for Azure destroy
  * Automatic pre-cleanup of Azure role assignments across related resource groups

* **Bug Fixes**
  * Detects and replaces stale or mismatched Azure role assignments deterministically
  * Cleanup failures are logged and treated as non-fatal so deletion continues

* **Tests**
  * New unit tests covering role-assignment creation, detection, deletion, and cleanup behaviors

### DIFF
--- a/cmd/cluster/azure/destroy.go
+++ b/cmd/cluster/azure/destroy.go
@@ -42,8 +42,10 @@ func NewDestroyCommand(opts *core.DestroyOptions) *cobra.Command {
 	cmd.Flags().StringVar(&opts.AzurePlatform.Location, "location", opts.AzurePlatform.Location, "Location for the cluster")
 	cmd.Flags().StringVar(&opts.AzurePlatform.ResourceGroupName, "resource-group-name", opts.AzurePlatform.ResourceGroupName, "The name of the resource group containing the HostedCluster infrastructure resources that need to be destroyed.")
 	cmd.Flags().BoolVar(&opts.AzurePlatform.PreserveResourceGroup, "preserve-resource-group", opts.AzurePlatform.PreserveResourceGroup, "When true, the managed/main resource group will not be deleted during cluster destroy. Only cluster-specific resources within the resource group will be cleaned up.")
+	cmd.Flags().StringVar(&opts.AzurePlatform.DNSZoneRGName, "dns-zone-rg-name", opts.AzurePlatform.DNSZoneRGName, util.DNSZoneRGNameDestroyDescription)
 
 	_ = cmd.MarkFlagRequired("azure-creds")
+	_ = cmd.MarkFlagRequired("dns-zone-rg-name")
 
 	logger := log.Log
 	cmd.Run = func(cmd *cobra.Command, args []string) {
@@ -148,6 +150,24 @@ func DestroyCluster(ctx context.Context, o *core.DestroyOptions) error {
 }
 
 func destroyPlatformSpecifics(ctx context.Context, o *core.DestroyOptions) error {
+	// Clean up role assignments before destroying infrastructure to avoid orphans.
+	// Match the create path resource-group names: {name}-nsg and {name}-vnet.
+	subscriptionID, azureCreds, err := util.SetupAzureCredentials(o.Log, nil, o.AzurePlatform.CredentialsFile)
+	if err != nil {
+		return fmt.Errorf("failed to setup Azure credentials: %w", err)
+	}
+
+	nsgRG := o.Name + "-nsg"
+	vnetRG := o.Name + "-vnet"
+
+	rbacManager := azureinfra.NewRBACManager(subscriptionID, azureCreds)
+	// assignCustomHCPRoles=false is safe: GetServicePrincipalScopes only uses the flag to select
+	// the role definition ID, not to modify the scopes list. Cleanup derives role assignment names
+	// from infraID + component + scope, so the role ID is irrelevant.
+	if err := rbacManager.CleanupRoleAssignments(ctx, o.Log, o.InfraID, o.AzurePlatform.ResourceGroupName, nsgRG, vnetRG, o.AzurePlatform.DNSZoneRGName, false); err != nil {
+		o.Log.Error(err, "Failed to clean up some role assignments, continuing with infrastructure deletion")
+	}
+
 	destroyInfraOptions := &azureinfra.DestroyInfraOptions{
 		Name:                  o.Name,
 		Location:              o.AzurePlatform.Location,
--- a/cmd/infra/azure/destroy_iam.go
+++ b/cmd/infra/azure/destroy_iam.go
@@ -32,12 +32,14 @@ func NewDestroyIAMCommand() *cobra.Command {
 	cmd.Flags().StringVar(&opts.Cloud, "cloud", opts.Cloud, util.CloudDescription)
 	cmd.Flags().StringVar(&opts.Name, "name", opts.Name, util.NameDescription)
 	cmd.Flags().StringVar(&opts.InfraID, "infra-id", opts.InfraID, util.InfraIDDescription)
+	cmd.Flags().StringVar(&opts.DNSZoneRG, "dns-zone-rg-name", opts.DNSZoneRG, util.DNSZoneRGNameDestroyDescription)
 
 	_ = cmd.MarkFlagRequired("workload-identities-file")
 	_ = cmd.MarkFlagRequired("azure-creds")
 	_ = cmd.MarkFlagRequired("resource-group-name")
 	_ = cmd.MarkFlagRequired("name")
 	_ = cmd.MarkFlagRequired("infra-id")
+	_ = cmd.MarkFlagRequired("dns-zone-rg-name")
 
 	l := log.Log
 	cmd.RunE = func(cmd *cobra.Command, args []string) error {
@@ -70,6 +72,7 @@ func BindDestroyIAMProductFlags(opts *DestroyIAMOptions, flags *pflag.FlagSet) {
 	flags.StringVar(&opts.Cloud, "cloud", opts.Cloud, util.CloudDescription)
 	flags.StringVar(&opts.Name, "name", opts.Name, util.NameDescription)
 	flags.StringVar(&opts.InfraID, "infra-id", opts.InfraID, util.InfraIDDescription)
+	flags.StringVar(&opts.DNSZoneRG, "dns-zone-rg-name", opts.DNSZoneRG, "The resource group name where the DNS zone resides (used to clean up role assignments)")
 }
 
 // Validate validates the DestroyIAMOptions
@@ -88,6 +91,9 @@ func (o *DestroyIAMOptions) Validate() error {
 	}
 	if o.InfraID == "" {
 		return fmt.Errorf("infra-id is required")
+	}
+	if o.DNSZoneRG == "" {
+		return fmt.Errorf("dns-zone-rg-name is required")
 	}
 	return nil
 }
@@ -109,6 +115,19 @@ func (o *DestroyIAMOptions) Run(ctx context.Context, l logr.Logger) error {
 		"clusterName", o.Name,
 		"infraID", o.InfraID,
 		"resourceGroup", o.ResourceGroupName)
+
+	// Clean up role assignments before destroying identities to avoid orphans.
+	// Match the create path resource-group names: {name}-nsg and {name}-vnet.
+	nsgRG := o.Name + "-nsg"
+	vnetRG := o.Name + "-vnet"
+
+	rbacManager := NewRBACManager(subscriptionID, azureCreds)
+	// assignCustomHCPRoles=false is safe: GetServicePrincipalScopes only uses the flag to select
+	// the role definition ID, not to modify the scopes list. Cleanup derives role assignment names
+	// from infraID + component + scope, so the role ID is irrelevant.
+	if err := rbacManager.CleanupRoleAssignments(ctx, l, o.InfraID, o.ResourceGroupName, nsgRG, vnetRG, o.DNSZoneRG, false); err != nil {
+		l.Error(err, "Failed to clean up some role assignments, continuing with identity deletion")
+	}
 
 	// Create the identity manager
 	identityManager := NewIdentityManager(subscriptionID, azureCreds, o.Cloud)
--- a/cmd/infra/azure/rbac.go
+++ b/cmd/infra/azure/rbac.go
@@ -19,14 +19,25 @@ import (
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
 	azureauth "github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/authorization/armauthorization/v2"
 
 	"k8s.io/utils/ptr"
+
+	"github.com/go-logr/logr"
 )
 
 const (
 	graphAPIEndpoint = "https://graph.microsoft.com/v1.0/servicePrincipals"
 )
+
+// roleAssignmentClient abstracts the Azure role assignment operations for testability.
+type roleAssignmentClient interface {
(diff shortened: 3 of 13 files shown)


cmd/infra/azure/destroy_iam.go
```diff
@@ -110,6 +112,16 @@ func (o *DestroyIAMOptions) Run(ctx context.Context, l logr.Logger) error {
 		"infraID", o.InfraID,
 		"resourceGroup", o.ResourceGroupName)
 
+	// Clean up role assignments before destroying identities to avoid orphans.
+	// The resource group names for NSG and VNet follow the convention: {name}-nsg-{infraID} and {name}-vnet-{infraID}.
+	nsgRG := o.Name + "-nsg-" + o.InfraID
+	vnetRG := o.Name + "-vnet-" + o.InfraID
```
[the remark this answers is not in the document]
bryan-cox: Done. Fixed RG names to `o.Name + "-nsg"` and `o.Name + "-vnet"` to match the create path in `create.go:174,205`. Good catch — the `-infraID` suffix was incorrect.

---

cmd/infra/azure/destroy_iam.go
```diff
@@ -110,6 +112,16 @@ func (o *DestroyIAMOptions) Run(ctx context.Context, l logr.Logger) error {
...
+
+	rbacManager := NewRBACManager(subscriptionID, azureCreds)
+	if err := rbacManager.CleanupRoleAssignments(ctx, l, o.InfraID, o.ResourceGroupName, nsgRG, vnetRG, o.DNSZoneRG, false); err != nil {
+		l.Error(err, "Failed to clean up some role assignments, continuing with identity deletion")
+	}
```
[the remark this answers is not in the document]
bryan-cox: Declining this change. The cleanup is intentionally best-effort — `CleanupRoleAssignments` aggregates errors and continues through all components. Aborting on partial failure would prevent infrastructure deletion entirely, leaving the user stuck. Since `deleteRoleAssignmentByName` already treats 404 as success, any remaining errors are likely transient (permissions, network) and the user can retry. The stale assignments are also handled on re-creation by the `assignRole` detect-and-replace logic, so this is a defense-in-depth approach, not the sole cleanup mechanism.

---

cmd/cluster/azure/destroy.go
```diff
@@ -148,6 +149,21 @@ func DestroyCluster(ctx context.Context, o *core.DestroyOptions) error {
...
+	if err != nil {
+		return fmt.Errorf("failed to setup Azure credentials: %w", err)
+	}
+
+	nsgRG := o.Name + "-nsg-" + o.InfraID
+	vnetRG := o.Name + "-vnet-" + o.InfraID
+
+	rbacManager := azureinfra.NewRBACManager(subscriptionID, azureCreds)
+	if err := rbacManager.CleanupRoleAssignments(ctx, o.Log, o.InfraID, o.AzurePlatform.ResourceGroupName, nsgRG, vnetRG, o.AzurePlatform.DNSZoneRGName, false); err != nil {
+		o.Log.Error(err, "Failed to clean up some role assignments, continuing with infrastructure deletion")
```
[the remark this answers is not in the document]
bryan-cox: This is a valid observation, but it is pre-existing — `NewRBACManager` and all `NewRoleAssignmentsClient` calls (including in the create path at `create.go:148`) already pass `nil` for client options. Adding cloud config support to `RBACManager` is a separate improvement that should be addressed in a follow-up PR across all call sites, not just the destroy path added here.

---

cmd/infra/azure/rbac.go
```diff
@@ -208,25 +223,39 @@ func (r *RBACManager) assignRole(ctx context.Context, infraID, component, assign
...
+			return nil
 		}
+		// Stale assignment — different principal owns this deterministic name.
+		log.Log.Info("Deleting stale role assignment with mismatched principal",
+			"role", role, "expectedPrincipal", assigneeID,
+			"stalePrincipal", ptr.Deref(existing.Properties.PrincipalID, "<nil>"),
+			"scope", scope)
+		if _, err := client.Delete(ctx, scope, roleAssignmentName, nil); err != nil {
+			return fmt.Errorf("failed to delete stale role assignment: %w", err)
+		}
```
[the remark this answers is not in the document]
bryan-cox: Done. Already addressed in a previous push — safe extraction of `stalePrincipal` with nil guards on both `Properties` and `PrincipalID`. Also added test cases for both nil scenarios.

---

cmd/infra/azure/rbac.go
```diff
@@ -208,25 +222,43 @@ func (r *RBACManager) assignRole(ctx context.Context, infraID, component, assign
...
-	if errors.As(err, &respErr) {
-		if respErr.StatusCode == http.StatusNotFound {
-			// proceed to create
-		} else if respErr.StatusCode == http.StatusForbidden || strings.EqualFold(respErr.ErrorCode, "AuthorizationFailed") {
-			log.Log.Info("Get not permitted; will attempt create and rely on 409 for idempotency.", "role", role, "assigneeID", assigneeID, "scope", scope)
-		} else {
-			return fmt.Errorf("failed checking role assignment existence: %w", err)
+		if existing.Properties != nil && existing.Properties.PrincipalID != nil && strings.EqualFold(*existing.Properties.PrincipalID, assigneeID) {
+			log.Log.Info("Skipping role assignment creation, role assignment already exists.", "role", role, "assigneeID", assigneeID, "scope", scope)
+			return nil
```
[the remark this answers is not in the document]
bryan-cox: Done. Added `RoleDefinitionID` validation to the GET match condition — now skips creation only when both principal and role definition match. Updated the corresponding test case.

---

cmd/cluster/azure/destroy.go
```diff
@@ -42,6 +42,7 @@ func NewDestroyCommand(opts *core.DestroyOptions) *cobra.Command {
 	cmd.Flags().StringVar(&opts.AzurePlatform.Location, "location", opts.AzurePlatform.Location, "Location for the cluster")
 	cmd.Flags().StringVar(&opts.AzurePlatform.ResourceGroupName, "resource-group-name", opts.AzurePlatform.ResourceGroupName, "The name of the resource group containing the HostedCluster infrastructure resources that need to be destroyed.")
 	cmd.Flags().BoolVar(&opts.AzurePlatform.PreserveResourceGroup, "preserve-resource-group", opts.AzurePlatform.PreserveResourceGroup, "When true, the managed/main resource group will not be deleted during cluster destroy. Only cluster-specific resources within the resource group will be cleaned up.")
+	cmd.Flags().StringVar(&opts.AzurePlatform.DNSZoneRGName, "dns-zone-rg-name", opts.AzurePlatform.DNSZoneRGName, util.DNSZoneRGNameDestroyDescription)
```
Nirshal, line 45: I would make this **required** instead than optional:
- there is not such an ecosystem of early adopters that justifies the risk to pollute resource groups if users are unaware/forgetful of the flag, just for retro-compatibility concerns;
- it is a symmetric behavior with respect to the create, so it make sense having to specify the flag for the destroy, if it is required for the create. 

Additionally, we should properly document the flag in the existing howto Azure self-managed documentation in `docs/content/how-to/azure/`

cmd/infra/azure/destroy_iam.go
```diff
@@ -70,6 +71,7 @@ func BindDestroyIAMProductFlags(opts *DestroyIAMOptions, flags *pflag.FlagSet) {
 	flags.StringVar(&opts.Cloud, "cloud", opts.Cloud, util.CloudDescription)
 	flags.StringVar(&opts.Name, "name", opts.Name, util.NameDescription)
 	flags.StringVar(&opts.InfraID, "infra-id", opts.InfraID, util.InfraIDDescription)
+	flags.StringVar(&opts.DNSZoneRG, "dns-zone-rg-name", opts.DNSZoneRG, "The resource group name where the DNS zone resides (used to clean up role assignments)")
```
Nirshal, line 75: See my previous comment

cmd/infra/azure/rbac_test.go
```diff
@@ -0,0 +1,510 @@
...
+			expectCreate: true,
+			expectDelete: false,
+		},
+		"When LIST page returns error it should return error": {
+			listErr:   internalServerError(),
+			expectErr: true,
+		},
+
+		// --- GET behaviors ---
+		"When GET finds assignment with matching principal and role it should skip creation": {
```
Nirshal, line 157: I was concerned about idempotency for `create iam` and the whole flow if for some reason the user forgetfully calls the command with the same infraId as a running cluster (ex: terminal history). I checked carefully and this is not the case today. Nonetheless consider adding a test case that verifies idempotency when create iam is re-run with the same infraID on an existing cluster — Azure CreateOrUpdate preserves the same principalID, so assignRole should skip without deleting. This guards against future changes breaking the idempotent behavior.

Nirshal (commented):
Thanks for the quick fix Bryan — the root cause matches exactly what I found during my testing.
                                                                
The key part is the hardening on the deterministic ID check in assignRole: verifying PrincipalID and RoleDefinitionID before skipping. I verified that this also works correctly in the idempotent case (re-running create iam + create infra on an existing cluster with the same infraID does not break it, because Azure `CreateOrUpdate` preserves the same principalID), but a test to be sure the behavior does not change in the future could be beneficial.

One thing I wanted to flag: `CleanupRoleAssignments` is called from both `destroy cluster` and `destroy iam`, doing the same work with the same parameters. This is a bit confusing, I think it would be cleaner if each destroy command had its own responsibility without overlap, but still, it's probably just cosmetic.  
See also my inline comments about docs that need update and the flag being optional, while for symmetry with the "create" it is better to be required IMHO.

cmd/cluster/azure/destroy.go
bryan-cox: Done. Made `--dns-zone-rg-name` required in all 3 destroy commands (`cmd/cluster/azure/destroy.go`, `cmd/infra/azure/destroy_iam.go`, `product-cli/cmd/cluster/azure/destroy.go`). Updated docs in `create-self-managed-azure-cluster.md`, `create-iam-separately.md`, and `deploy-azure-private-clusters.md` to include the flag in all destroy command examples.

---

cmd/infra/azure/destroy_iam.go
bryan-cox: Done. Applied the same change here — `--dns-zone-rg-name` is now required in `destroy_iam.go` (both `MarkFlagRequired` and `Validate()`). Docs updated as well.

---

cmd/infra/azure/rbac_test.go
bryan-cox: Good call — this is already covered by the existing test suite. The test case "When LIST finds matching assignment it should skip creation" (rbac_test.go:124) verifies that when the LIST pager returns a matching assignment (same principal, role, and scope), `assignRole` skips both delete and create. The "When GET finds assignment with matching principal and role it should skip creation" case (rbac_test.go:157) covers the GET fallback path. Together these ensure idempotency across both detection mechanisms.

---

bryan-cox:
@Nirshal Thanks for the thorough review and for validating the fix against your testing!

Re: the duplicate `CleanupRoleAssignments` in both `destroy cluster` and `destroy iam` — agreed it's a bit redundant. The reason is that these two commands can be run independently: a user might run `destroy cluster` without ever running `destroy iam` (or vice versa). Both need to handle cleanup to avoid orphaned role assignments regardless of which path the user takes. The calls are idempotent (404s from already-deleted assignments are silently ignored), so the overlap is harmless. I'll add a comment clarifying this in a follow-up if needed.

All inline comments addressed:
- `--dns-zone-rg-name` is now **required** in all 3 destroy commands (hypershift, hcp, and destroy-iam CLIs)
- Updated docs in `create-self-managed-azure-cluster.md`, `create-iam-separately.md`, and `deploy-azure-private-clusters.md`
- Idempotency is covered by existing tests (LIST match and GET match paths)

---

Nirshal:
Minor observation (not related to this PR, but noticed during testing): `product-cli/main.go` does not initialize the `controller-runtime` logger, while `main.go` (hypershift CLI) does it at line 43 with `ctrl.SetLogger(zap.New(...))`. This causes a noisy warning during `create infra` with the product-cli:

```
[controller-runtime] log.SetLogger(...) was never called; logs will not be displayed.
```

The `hypershift` CLI does not produce this warning because the logger is properly initialized. This is pre-existing and not introduced by this PR, but it might be worth adding the same logger setup to `product-cli/main.go` in a follow-up.

Nirshal:
/verified by @Nirshal  (see https://redhat.atlassian.net/browse/OCPBUGS-84251?focusedCommentId=16793903)

Nirshal:
/verified by @Nirshal

Verification comment: https://redhat.atlassian.net/browse/OCPBUGS-84251?focusedId=16793872&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-16793872

bryan-cox:
/verified by @Nirshal (see https://redhat.atlassian.net/browse/OCPBUGS-84251?focusedCommentId=16793903)

bryan-cox:
/verified by @Nirshal (see https://redhat.atlassian.net/browse/OCPBUGS-84251?focusedCommentId=16793903)

bryan-cox:
/verified by @Nirshal (see https://redhat.atlassian.net/browse/OCPBUGS-84251?focusedCommentId=16793903)

bryan-cox:

This only touches Azure CLI and docs.

PR #8328, opened 2026-04-24
docs: NO-JIRA: clarify serialization tag behaviour in api/AGENTS.md

enxebre:
## Summary
- Explain why `omitempty`/`omitzero` should be set on **every** field regardless of `+required` or `+optional`: the tag controls serialization (what goes on the wire), not validation (enforced at admission via the schema). Without the tag, zero-value fields serialize as explicit values, breaking defaulting, server-side apply, and strategic merge patch.
- Add upstream [Kubernetes CRD docs](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/) link alongside the downstream [OpenShift API conventions](https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md), with explicit note that downstream wins on conflicts.

## Test plan
- [ ] Docs-only change, no code impact

## Documentation
* Updated API convention guidance to rely on the linter as the enforcement source and clarified precedence between OpenShift and upstream Kubernetes guidance.
* Strengthened serialization guidance: fields must use explicit omit-style tagging to control wire representation, with notes on scalar vs struct usage, behavioral impacts when zero values are serialized, and tooling/runtime expectations.

### DIFF
--- a/api/AGENTS.md
+++ b/api/AGENTS.md
@@ -6,9 +6,7 @@ This module (`github.com/openshift/hypershift/api`) is consumed by external Go c
 
 These are non-obvious behaviors of the Kubernetes API machinery that affect how CRD types in `api/` must be written. These are not style preferences or conventions — they are fundamental facts and reasoning about how the system works.
 
-For conventions read [https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md](https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md)
-
-`make api-lint-fix` will enforce most conventions and best practices. **Do not suppress or work around its findings** — they exist because violations have caused production issues. If a finding seems wrong, understand why the rule exists before dismissing it.
+For conventions, always trust the kube-api-linter (`make api-lint-fix`). Do not question its findings and do not add exclusions unless explicitly told to by a human reviewer. The linter encodes the [OpenShift API conventions](https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md) (authoritative) and [upstream Kubernetes API conventions](https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api-conventions.md) (informational, downstream wins on conflicts). Do not fetch or read those documents — the linter already enforces them, and the actionable rules are captured in this file.
 
 ### API Versioning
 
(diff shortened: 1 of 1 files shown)


api/AGENTS.md
```diff
@@ -6,7 +6,12 @@ This module (`github.com/openshift/hypershift/api`) is consumed by external Go c
 
 These are non-obvious behaviors of the Kubernetes API machinery that affect how CRD types in `api/` must be written. These are not style preferences or conventions — they are fundamental facts and reasoning about how the system works.
 
-For conventions read [https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md](https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md)
+For conventions read:
+
+- **Downstream (authoritative):** [OpenShift API conventions](https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md)
+- **Upstream (informational):** [Kubernetes API conventions](https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api-conventions.md)
```
JoelSpeed: Both of these docs are huge, is telling the agent to read so much context going to cause context rot? Some of the serialization stuff you add here is already mentioned in these docs, but its in the ~middle. So I suspect the agent forgets it because we are giving it too much.

I think those docs need a lot of work to make them agent accessible 

Worth teaching the agent to trust the linter more? And save it from having to consume any conventions docs that cannot be linted?

api/AGENTS.md
```diff
@@ -29,7 +34,8 @@ ENVTEST_OCP_K8S_VERSIONS=1.35.0 make test-envtest-ocp # Run envtest for CEL vali
 
 ### Serialization
 
-- `omitempty` **does nothing for non-pointer structs.** Only `omitzero` correctly omits a struct field when it equals its zero value. This is a Go encoding/json behavior, not a Kubernetes convention.
+- **Always set `omitempty` or `omitzero` on every field, regardless of whether it is `+required` or `+optional`.** These tags control serialization, not validation. `+required` is a schema constraint enforced at admission time; the serialization tag controls what goes on the wire. Without a tag, a zero-value field serializes as an explicit value (e.g., `"pullSecret": {"name": ""}`), which makes the API server unable to distinguish "not set" from "explicitly set to empty." This breaks defaulting, server-side apply field ownership, and strategic merge patch — all of which rely on field absence to mean "don't touch this."
```
JoelSpeed: Be explicit about which tags?
```suggestion
- **Always set `omitempty` or `omitzero` on every field, regardless of whether it is `+required` or `+optional`.** `omitempty`/`omitzero` tags control serialization, not validation. `+required` is a schema constraint enforced at admission time; the serialization tag controls what goes on the wire. Without a tag, a zero-value field serializes as an explicit value (e.g., `"pullSecret": {"name": ""}`), which makes the API server unable to distinguish "not set" from "explicitly set to empty." This breaks defaulting, server-side apply field ownership, and strategic merge patch — all of which rely on field absence to mean "don't touch this."
```
The other element of this is that without omission, a structured client serializes the empty object as you've said, and this passes a `+required` check (based on key presence) without any check on the value. Most often folks add `+required` but expect that means non-empty, which it doesn't. So the API user can forget to add a value for the required field, it then passes the requiredness check, and the reader sees a required field with empty value which is unexpected

api/AGENTS.md
enxebre: >Worth teaching the agent to trust the linter more? And save it from having to consume any conventions docs that cannot be linted?

I see those as different goals. We can and do enforce the agent building blocks to run the linter, we can do that more deterministically with git/cc hooks, whatever. What I'm after here is to improve the agent ability to reason about api machinery and understanding why the linter does what it does, not to enforce running the linter

I agree on the signal/noise potential problem, and this has not given good results so far, so let's change it.

PR #8331, opened 2026-04-24
OCPBUGS-77827: fix(api): add missing has() guards to servingCerts CEL validation rule

rutvik23:
- Adds missing `has()` guards to the CEL validation rule that checks whether an APIServer  `loadBalancer.hostname` conflicts with `servingCerts.namedCertificates[].names`

- The rule was introduced in #6194 (OCPBUGS-56725) but assumed `servingCerts` and  `namedCertificates` would always be present. In practice:
  - On **create**: `servingCerts` is a non-pointer struct (serialized as `servingCerts: {}`),  
  but `namedCertificates` (omitempty slice) is absent → `no such key: namedCertificates`       
  - On **patch/upgrade**: Stored objects from older CRD versions may not have `servingCerts` at
   all → `no such key: servingCerts`                                                           
   - This blocks any user who sets `loadBalancer.hostname` on the APIServer service with          
  `configuration.apiServer` (e.g. `additionalCORSAllowedOrigins`) but without `servingCerts`.
  
### Design decision                                                                          
                                                                                 
 - CEL uses short-circuit evaluation within `&&` chains. Adding `has()` guards for `servingCerts`, `namedCertificates`,
    and `cert.names` before accessing them is the CEL pattern to handle optional fields safely. This is consistent with the existing `has()` guards already in the same rule for `loadBalancer`, `configuration`, and `apiServer`. 

Fixes [OCPBUGS-77827](https://redhat.atlassian.net/browse/OCPBUGS-77827)

- Three `has()` guards were added: `has(self.configuration.apiServer.servingCerts)`,         
  `has(self.configuration.apiServer.servingCerts.namedCertificates)`, and `has(cert.names)`.
- Backport to affected release branches (4.20+) likely needed. 

## Checklist:
- [x] This change includes envtest tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened validation for API server certificate settings to handle optional fields safely and avoid invalid load balancer hostname assignments, reducing configuration errors and improving reliability.


cmd/install/assets/crds/hypershift-operator/tests/hostedclusters.hypershift.openshift.io/stable.hostedclusters.services.testsuite.yaml
```diff
@@ -246,3 +246,46 @@ tests:
                 - anything
                 - kas.duplicated.hostname.com
     expectedError: "loadBalancer hostname cannot be in ClusterConfiguration.apiserver.servingCerts.namedCertificates"
+  - name: When loadBalancer hostname is set and apiServer has no servingCerts it should pass
```
jparrill, line 249: For now in the envtest, is only covered the first `has()`. Do you mind add two more test cases to independently exercise each `has()` guard:
- "**When loadBalancer hostname is set and apiServer has servingCerts but no namedCertificates it should pass**" -- set `servingCerts: {}` without namedCertificates. This exercises the `has(self.configuration.apiServer.servingCerts.namedCertificates)` guard.
- "**When loadBalancer hostname is set and namedCertificates entry has no names it should pass**" -- add a namedCertificates entry with only servingCertificate and no names field. This exercises the `has(cert.names)` guard.

jparrill (commented):
Thanks for the PR!. Dropped a comment

enxebre:
Thanks! I think this requires several fixes:

>On create: servingCerts is a non-pointer struct (serialized as servingCerts: {}),

Can you please report/PR against o/api to make sure 	ServingCerts APIServerServingCerts `json:"servingCerts"` uses omitzero.

Also it seems NamedCertificates should be required here:
```
type APIServerServingCerts struct {
	// namedCertificates references secrets containing the TLS cert info for serving secure traffic to specific hostnames.
	// If no named certificates are provided, or no named certificates match the server name as understood by a client,
	// the defaultServingCertificate will be used.
	// +optional
	// +listType=atomic
	// +kubebuilder:validation:MaxItems=32
	NamedCertificates []APIServerNamedServingCert `json:"namedCertificates,omitempty"`
}
```

Then in this repo, is there a real use case to allow loadBalancer.hostname with out configuration.apiServer.servingCerts.namedCertificates[0].Names? if there isn't, then this can be required by the cel rule

cc @JoelSpeed

JoelSpeed:
> Can you please report/PR against o/api to make sure ServingCerts APIServerServingCerts json:"servingCerts" uses omitzero.

This API is old, changing the serialization now is not an option. Adding omitzero to newer clients will cause conflicts between old and new clients as some would attempt to drop the empty `{}`, and some would try to add it back. In general we don't change the `omitzero`/`omitempty` once an API has reached v1

rutvik23:
> Then in this repo, is there a real use case to allow loadBalancer.hostname with out configuration.apiServer.servingCerts.namedCertificates[0].Names? if there isn't, then this can be required by the cel rule

It seems the users are allowed to set `loadBalancer.hostname` for DNS resolution (e.g., with external-dns creating a record for (kas.example.com) while using the default operator-managed serving certificates and the hostname just controls where DNS points. The affected user had `loadBalancer.hostname + configuration.apiServer` (for CORS/TLS/Audit profile) but no servingCerts and the validation failed when HC was patched.

jparrill (commented):
/lgtm

api/hypershift/v1beta1/hostedcluster_types.go
```diff
@@ -524,7 +524,7 @@ type Capabilities struct {
...
+// +kubebuilder:validation:XValidation:rule=`!self.services.exists(s, s.service == 'APIServer' && has(s.servicePublishingStrategy.loadBalancer) && s.servicePublishingStrategy.loadBalancer.hostname != "" && has(self.configuration) && has(self.configuration.apiServer) && has(self.configuration.apiServer.servingCerts) && has(self.configuration.apiServer.servingCerts.namedCertificates) && self.configuration.apiServer.servingCerts.namedCertificates.exists(cert, has(cert.names) && cert.names.exists(n, n == s.servicePublishingStrategy.loadBalancer.hostname)))`, message="APIServer loadBalancer hostname cannot be in ClusterConfiguration.apiserver.servingCerts.namedCertificates[]"
```
JoelSpeed, line 527: Have you considered using optional types here to remove all the `has` blocks
```suggestion
// +kubebuilder:validation:XValidation:rule=`!self.services.exists(s, s.service == 'APIServer' && has(s.servicePublishingStrategy.loadBalancer) && s.?servicePublishingStrategy.loadBalancer.hostname.orValue("") != "" && self.?configuration.apiServer.servingCerts.namedCertificates.orValue("[]").exists(cert, cert.?names.orValue("[]").exists(n, n == s.servicePublishingStrategy.loadBalancer.hostname)))`, message="APIServer loadBalancer hostname cannot be in ClusterConfiguration.apiserver.servingCerts.namedCertificates[]"
```

rutvik23: @JoelSpeed I tried using CEL optional types as you suggested; but then `make update` & CRD generation fails consistently on this cost-estimation error: `OpenAPIv3 schema, spec.validation.openAPIV3Schema: Forbidden: x-kubernetes-validations estimated rule cost total for entire OpenAPIv3 schema exceeds budget by factor of more than 100x (try simplifying the rule, or adding maxItems, maxProperties, and maxLength where arrays, maps, and strings are declared)]`  

- upon digging more in this context, it turned out that CEL optional types (.?/.orValue()) have significantly higher estimated cost in the Kubernetes CRD cost estimator compared to has() guards. Hence, the has() guards worked out properly in local testing & make update also didn't fail.
- The HostedCluster CRD is massive with many validation rules, so it seem to be having very little cost headroom. We've had similar issue in this repo earlier -> [NO-JIRA: fix(api): replace CEL url() validators with regex to fix CRD cost budget](https://github.com/openshift/hypershift/pull/8004)

So I think we're good to go with has() guards for now.

JoelSpeed: > have significantly higher estimated cost

Do you have any sources/links to explain this? Since it's a presence check, it should be fixed (low) cost 🤔

rutvik23: The root cause is likely in the `cel-go` cost estimator (cel-go/checker/cost.go); let me know if you agree with this hypothesis:                                                                                                                              
- `Path tracking lost with .?`: Regular field selects track schema paths via `costSelect → addPath()` (https://github.dev/google/cel-go/blob/master/checker/cost.go#L507-L528), which allows the estimator to look up maxItems from the CRD schema. However, `.? (optional select) is dispatched as a CallKind, and costCall` (https://github.dev/google/cel-go/blob/master/checker/cost.go#L531) has no path-building for the select_optional_field overload; so the schema path is lost after ?.    
- `orValue() returns no ResultSize`: It falls through to the generic O(1) case (https://github.dev/google/cel-go/blob/master/checker/cost.go#L866) with no result size metadata, so `sizeOrUnknown returns {Min: 0, Max: math.MaxUint64}`. 
- Unbounded size multiplied through nested exists(): The comprehension cost (https://github.dev/google/cel-go/blob/master/checker/cost.go#L688) computes `rangeCnt.MultiplyByCost(stepCost)`. With both `namedCertificates.orValue([]) and cert.?names.orValue([])` having `unbounded range` sizes, the nested loops produce `MaxUint64 * MaxUint64 → overflow`.  

### Comparison
Aspect                 | has() guards           | .?/.orValue()
---------------------- | ---------------------- | -------------------------------
AST kind               | SelectKind (test-only) | CallKind
Path tracking          | Preserved              | Lost — no path for select_optional_field
List size for exists() | Bounded by schema      | Max: math.MaxUint64
Nested loop cost       | bounded * bounded      | MaxUint64 * MaxUint64

JoelSpeed: Thanks for checking, this feels like a bug in the estimator to me, lets continue with what you have for now

jparrill (commented):
/lgtm

rutvik23:
/verified by HO [image](quay.io/rkshirsa/hypershift-operator:OCPBUGS-77827) local build

PR #8337, opened 2026-04-24
CNTRLPLANE-2262: Add Azure scale-from-zero support

jhjaggars:
Extend the existing scale-from-zero autoscaling framework to support Azure by implementing an Azure instance type provider that queries the Azure Resource SKUs API for VM size specifications and writing capacity annotations on MachineDeployments.

Changes:
- Add Azure instancetype.Provider using armcompute.ResourceSKUsClient
- Add AzureMachineTemplate case to scale_from_zero.go type switch
- Extend supportedScaleFromZeroPlatform() for Azure
- Extend reconcileScaleFromZeroAnnotations() for Azure
- Update autoscalerEnabledCondition() to accept Azure with min=0
- Update effectiveMin guard in capi.go to allow min=0 for Azure
- Add "azure" to supportedProviders in main.go and install.go
- Add Azure provider initialization with credential file parsing
- Update CRD CEL validation to allow min=0 for Azure platform
- Add unit tests for Azure provider and extended type switches

Fixes 

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Scale-from-zero autoscaling now supported on Azure as well as AWS; operator/CLI accept Azure as a provider and use Azure SKU data for instance-type info.

* **Bug Fixes**
  * Replica and autoscaler behavior updated so min=0 is honored for Azure where supported.

* **Documentation**
  * API and CRD docs updated to reflect Azure support for scale-from-zero.

* **Tests**
  * Added/updated tests covering Azure scale-from-zero, instance-type parsing, and annotation behavior.


everettraven (commented):
Seems like a pretty straightforward API change.

Approved from an API perspective.

/approve

jhjaggars:
/verified by @jhjaggars (on top of OCP ipi in Azure)

hypershift-operator/controllers/nodepool/capi.go
```diff
@@ -1081,7 +1082,7 @@ func setMachineSetReplicas(nodePool *hyperv1.NodePool, machineSet *capiv1.Machin
 		// NodePools from being permanently stuck at 0 replicas on platforms that don't support
 		// scale-from-zero metadata.
 		effectiveMin := ptr.Deref(nodePool.Spec.AutoScaling.Min, 0)
-		if effectiveMin == 0 && nodePool.Spec.Platform.Type != hyperv1.AWSPlatform {
+		if effectiveMin == 0 && nodePool.Spec.Platform.Type != hyperv1.AWSPlatform && nodePool.Spec.Platform.Type != scaleFromZeroPlatform {
```
enxebre: do we need to keep nodePool.Spec.Platform.Type != hyperv1.AWSPlatform check here?

hypershift-operator/controllers/nodepool/scale_from_zero.go
```diff
@@ -80,10 +81,8 @@ func setScaleFromZeroAnnotationsOnObject(ctx context.Context, provider instancet
 	case *infrav1.AWSMachineTemplate:
 		instanceType = template.Spec.Template.Spec.InstanceType
 		statusCapacity = template.Status.Capacity
-	// Future platform support can be added here:
-	// case *capiazure.AzureMachineTemplate:
-	//     instanceType = template.Spec.Template.Spec.VMSize
-	//     statusCapacity = template.Status.Capacity
+	case *capiazure.AzureMachineTemplate:
+		instanceType = template.Spec.Template.Spec.VMSize
```
enxebre, line 85: don't we need `statusCapacity = template.Status.Capacity` here?

hypershift-operator/controllers/nodepool/nodepool_controller.go
```diff
@@ -104,6 +104,7 @@ type NodePoolReconciler struct {
 	KubevirtInfraClients    kvinfra.KubevirtInfraClientMap
 	EC2Client               awsapi.EC2API
 	InstanceTypeProvider    instancetype.Provider
+	ScaleFromZeroPlatform   hyperv1.PlatformType
```
enxebre, line 108: I understand it's only currently possible have scaling for zero for a platform at a time. As it's not supported to have mixed nodepools.
Still I'd rather let this be a dynamic check on nodepools as before

jhjaggars: I combined your feedback into this commit: https://github.com/openshift/hypershift/pull/8337/commits/787a0ae107db6b931d4efecb984d66278373839a

jhjaggars:
/verified by @jhjaggars (on top of OCP ipi in Azure)

hypershift-operator/controllers/nodepool/instancetype/azure/provider.go
```diff
@@ -0,0 +1,168 @@
...
+
+	if p.cache == nil {
+		if err := p.loadSKUs(ctx); err != nil {
+			return nil, fmt.Errorf("failed to load Azure Resource SKUs: %w", err)
+		}
+	}
+
+	info, ok := p.cache[instanceType]
+	if !ok {
+		return nil, fmt.Errorf("VM size %q not found in Azure Resource SKUs for location %q", instanceType, p.location)
```
csrwng: nit: Once the SKU cache is loaded, it's never refreshed — new VM sizes added to the region after the operator starts will never be picked up (unlike the AWS provider which queries EC2 on every call). Consider adding a TTL or periodic refresh to the cache.

Additionally, if a VM size isn't in the cache (e.g., a HostedCluster in a different location than the configured `--scale-from-zero-creds` location uses a region-specific SKU), this error is permanent, but the caller retries every 30 seconds indefinitely. Consider distinguishing "cache load failed" (transient, worth retrying) from "VM size not in cache" (permanent, should surface clearly rather than retry-loop).

.gitignore
```diff
@@ -59,4 +59,7 @@ __pycache__/
 .pytest_cache/
 .venv/
 venv/
-*.egg
\ No newline at end of file
+*.egg
```
csrwng, line 62: nit: The `.claude/skills/dev/azure-sandbox-*` entries appear unrelated to Azure scale-from-zero — consider moving them to a separate PR.

csrwng (commented):
Nice work extending scale-from-zero to Azure. The data-driven `hasStatusCapacity` refactor is a good pattern. Two minor comments below.

PR #8338, opened 2026-04-26
OCPBUGS-84368: add safe-to-evict annotation and remove tolerations to fix autoscaler scale-down and drain loop

sdminonne:
### Problem

The `kas-connection-checker` deployment runs in the `kube-system` namespace without a PDB or safe-to-evict annotation. The cluster autoscaler treats pods in `kube-system` without a PDB as system-critical pods that [cannot be evicted during scale-down](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-types-of-pods-can-prevent-ca-from-removing-a-node), blocking node draining on underutilized nodes. See also the [autoscaler system drainability rule](https://github.com/kubernetes/autoscaler/blob/51bd425d1212ceead9a21f4d883923c61680e17e/cluster-autoscaler/simulator/drainability/rules/system/rule.go#L58-L65).

Additionally, the deployment used a blanket `{Effect: NoSchedule, Operator: Exists}` toleration that matched the cordon taint (`node.kubernetes.io/unschedulable:NoSchedule`), causing replacement pods to be scheduled back onto cordoned nodes during drain — creating a destructive evict-reschedule loop.

### Solution

1. **Add `cluster-autoscaler.kubernetes.io/safe-to-evict: "true"` annotation** to the pod template. This bypasses the autoscaler's [system drainability rule](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/simulator/drainability/rules/system/rule.go) for kube-system pods, allowing nodes to be scaled down. The annotation is preferred over a PDB because it doesn't block scale-to-zero operations.

2. **Remove all custom tolerations.** The previous blanket `NoSchedule` toleration matched the cordon taint, causing the evict-reschedule loop described above. In a HyperShift hosted cluster there are no master nodes in the guest cluster, and the checker doesn't need to run on infra nodes, so no `NoSchedule` tolerations are needed. Kubernetes provides default `NoExecute` tolerations (300s grace period) for `unreachable` and `not-ready` via the `DefaultTolerationSeconds` admission controller.

3. **Add `TopologySpreadConstraint`** with `ScheduleAnyway` to spread the 3 replicas across nodes. Using `ScheduleAnyway` (soft constraint) avoids blocking scheduling on small clusters (1-2 nodes).

4. **Preserve existing annotations and labels** during reconciliation instead of replacing the entire map. Explicitly delete the stale `openshift.io/required-scc` annotation from older HCCO versions (kube-system is exempt from SCC admission via the `systemNamespaces` allowlist in origin).

### Scope and impact

The tolerations are only removed from the `kas-connection-checker` Deployment in the guest cluster's `kube-system` namespace — no other workloads are affected. If all worker nodes carry custom `NoSchedule` taints and the checker pods cannot schedule, the `ControlPlaneConnectionAvailable` condition will report `False` (not `Unknown`) with reason `ControlPlaneConnectionKASAccessFailedReason`, because the ConfigMap never receives a `lastSucceeded` timestamp. This is an accurate signal: if the checker cannot run, data-plane-to-control-plane connectivity genuinely cannot be verified. In the normal case, the 3 replicas with `ScheduleAnyway` topology spread will schedule on any untainted nodes.

### Changes

- `control-plane-operator/.../resources/resources.go`: Add `safe-to-evict` annotation to pod template (merge, don't replace); remove stale `required-scc` annotation; remove all custom tolerations; add `TopologySpreadConstraint`; broaden log message from "deployment" to "resources".
- `control-plane-operator/.../resources/resources_test.go`: Update tests to validate the `safe-to-evict` annotation, assert no custom tolerations, verify topology spread constraint. Rename test to `TestReconcileKASConnectionChecker` per `TESTING.md` convention.
- `test/e2e/v2/tests/hosted_cluster_compliance_test.go`: Add `EnsureKASConnectionCheckerSpecTest` to the v2 e2e suite, gated behind `VersionAtLeast(Version423)`.
- `test/e2e/create_cluster_test.go`: Remove `EnsureKASConnectionCheckerSpec` call (moved to v2).
- `test/e2e/util/util.go`: Remove `EnsureKASConnectionCheckerSpec` helper (moved to v2).

## Test plan

- [x] `go test ./control-plane-operator/hostedclusterconfigoperator/controllers/resources/... -count=1` passes
- [x] `make verify` passes
- [ ] Verify on a live cluster that the annotation is present on kas-connection-checker pods
- [x] Verify cluster autoscaler can scale down nodes running kas-connection-checker pods

Fixes: https://redhat.atlassian.net/browse/OCPBUGS-84368


control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -1749,6 +1752,20 @@ done`, endpoint, manifests.KASConnectionCheckerConfigMapName, manifests.KASConne
...
+	pdb := manifests.KASConnectionCheckerPodDisruptionBudget()
+	if _, err := r.CreateOrUpdate(ctx, r.client, pdb, func() error {
+		pdb.Spec.Selector = &metav1.LabelSelector{
+			MatchLabels: map[string]string{
+				"app": manifests.KASConnectionCheckerName,
+			},
+		}
+		pdb.Spec.MinAvailable = ptr.To(intstr.FromInt32(1))
+		pdb.Spec.UnhealthyPodEvictionPolicy = ptr.To(policyv1.AlwaysAllow)
+		return nil
```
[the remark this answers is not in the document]
sdminonne: @coderabbitai ptal

sdminonne:
According to claude 

```
  Summary: This is a KNOWN FLAKY upstream Kubernetes e2e test, unrelated to the PR #8338 changes.
           The test creates a CRD with a custom ResourceQuota and waits for the quota status to
           reflect the custom resource lifecycle. It timed out after ~61 seconds with
           "context deadline exceeded" at resource_quota.go:683. PR #8338 only modifies
           kas-connection-checker (PDB + safe-to-evict annotation) and has no interaction
           with ResourceQuota or CRD lifecycle machinery. The test was not eligible for retries
           per the retry strategy. Overall suite: 1929 passed, 1 blocking fail, 1 informing fail,
           1 flaky, 2127 skipped.

  Evidence:
    - JUnit XML failure message:
      <failure>fail [k8s.io/kubernetes/test/e2e/apimachinery/resource_quota.go:683]: context deadline exceeded</failure>
    - Test output from JUnit system-out:
      "Creating a Custom Resource Definition" at 20:29:39
      "Truncated CRD plural name from 'e2e-test-e2e-resourcequota-975-6663-crds' to 'e2e-test-e2e-resourcequota-975'"
      "context deadline exceeded" at 20:30:40 (~61s later)
      "Found 0 events" in namespace e2e-resourcequota-975
    - Build log: "failed: (1m1s) 2026-04-26T20:30:40 '[sig-api-machinery] ResourceQuota should create
      a ResourceQuota and capture the life of a custom resource.'"
    - Retry log: "Test [...] not eligible for retries (strategy returned 0)"
```

I'm rerunning

enxebre:
Thanks! few questions:

>Add [cluster-autoscaler.kubernetes.io/safe-to-evict](http://cluster-autoscaler.kubernetes.io/safe-to-evict): "true" annotation to the kas-connection-checker pod template so the cluster autoscaler can evict pods despite system-node-critical priority class.

why is this annotation needed? does the autoscaler has any specific behaviour for pods with this priority class during draining?

>system-node-critical priority class, which prevents the cluster autoscaler from evicting its pods during scale-down.

where does autoscaler implement this?
wouldn't this mess with scale down draining just because of its hard tolerations?

enxebre:
Orthogonal to the PR but this comment seems to contradict that toleration? https://github.com/openshift/hypershift/blob/main/control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go#L1722-L1731

The reason this pod is blocking scale down, seems to be that it runs in the kube-system ns without a PDB.
https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-types-of-pods-can-prevent-ca-from-removing-a-node
[https://github.com/kubernetes/autoscaler/blob/51bd425d1212ceead9a21f4d883923c61680[…]/cluster-autoscaler/simulator/drainability/rules/system/rule.go](https://github.com/kubernetes/autoscaler/blob/51bd425d1212ceead9a21f4d883923c61680e17e/cluster-autoscaler/simulator/drainability/rules/system/rule.go#L58-L65)

We can solve this by either setting a PDB or adding the safe to evict annotation.
The fact that the pod is system-node-critical has no impact on autoscaler decisions (it impacts scheduling/preemption decisions). The PR desc and commit messages are misleading.

If we think the pod deserves a pdb, let's have it and not use the annotation as it's no needed. Otherwise let's have the annotation with no PDB. I'd vote the latter so this pod pdb doesn't block scale to zero operations. Let's please also make sure pr desc and commits captures the real rationale

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources_test.go
```diff
@@ -476,7 +493,6 @@ func withICS(hcp *hyperv1.HostedControlPlane) *hyperv1.HostedControlPlane {
 }
 
 func TestReconcileKubeadminPasswordHashSecret(t *testing.T) {
-	t.Parallel()
```
enxebre, line 497: why are all this parallel being dropped?

enxebre:
The PR desc still mentions cluster-autoscaler.kubernetes.io/safe-to-evict: "true" which is not used

Can we also update this to drop the toleration that contradicts the // comment https://github.com/openshift/hypershift/blob/main/control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go#L1722-L1731?

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources_test.go
sdminonne: Right, thanks. it was a bad merge. Restored them. Everywhere.

sdminonne:
comformance looks a flake

patjlm:
Any news on this one? This blocks scale down to zero (or unique nodepool removal)

sdminonne:
## Validation — All 3 Changes Verified on Live Cluster

Tested on a HyperShift hosted cluster (AWS us-east-1, 2 worker nodes) with `kas-connection-checker` deployed with all PR changes applied.

### 1. `safe-to-evict` Annotation Present

```
$ oc get pod -n kube-system -l app=kas-connection-checker -o jsonpath='{.items[0].metadata.annotations.cluster-autoscaler\.kubernetes\.io/safe-to-evict}'
true
```

All 3 replicas carry `cluster-autoscaler.kubernetes.io/safe-to-evict: "true"`.

### 2. No Custom Tolerations

Pods have no custom tolerations — only the default `NoExecute` tolerations for `not-ready` and `unreachable` (300s) added by the `DefaultTolerationSeconds` admission controller. The blanket `NoSchedule` toleration that caused the drain loop is gone.

### 3. TopologySpreadConstraint Active

3 replicas spread across 2 worker nodes with `ScheduleAnyway`:
```
kas-connection-checker-...-hqkbt   ip-10-0-137-53.ec2.internal
kas-connection-checker-...-l7vs7   ip-10-0-137-53.ec2.internal
kas-connection-checker-...-wp4nk   ip-10-0-139-101.ec2.internal
```

### 4. Autoscaler Scale-Down Test — PASSED

1. Enabled autoscaling (min:0, max:3) on the nodepool
2. Created a temporary workload to trigger scale-up from 2 → 3 nodes
3. Deleted the workload — 3rd node became idle
4. Autoscaler marked 3rd node as "unneeded"
5. **Scale-down from 3 → 2 succeeded** — kas-connection-checker pods did NOT block eviction
6. Autoscaler log confirmed: `lastScaleDownDeleteTime` updated

Without the `safe-to-evict` annotation, the autoscaler's [system drainability rule](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/simulator/drainability/rules/system/rule.go) would have blocked eviction of these `kube-system` pods.

### 5. Scale-to-Zero and Recovery Test — PASSED

1. Set nodepool `replicas: 0` — both nodes drained and all machines deleted
2. Node drain completed cleanly — kas-connection-checker pods evicted without hanging
3. Hosted cluster control plane remained healthy (KAS Available, Etcd QuorumAvailable)
4. Scaled back to `replicas: 2` — 2 new nodes came up, kas-connection-checker pods rescheduled and healthy
5. `kas-connection-checker` ConfigMap `lastSucceeded` updated after recovery

### 6. Functional Health

kas-connection-checker actively checking KAS connectivity throughout all tests:
```
$ oc get configmap kas-connection-checker -n kube-system -o jsonpath='{.data.lastSucceeded}'
2026-09-01T18:25:37Z
```

### Summary

All 3 changes from the PR work as intended:
- safe-to-evict annotation allows autoscaler to evict kas-connection-checker pods
- Removing the blanket NoSchedule toleration prevents the drain loop (pods don't reschedule onto cordoned nodes)
- TopologySpreadConstraint with ScheduleAnyway spreads replicas without blocking scheduling on small clusters

bryan-cox:
/verified by @sdminonne 

See https://github.com/openshift/hypershift/pull/8338#issuecomment-5498654097

sdminonne:
/verified by me in local dev cluster

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources_test.go
```diff
@@ -3209,7 +3210,7 @@ func getKASCheckerDeployment(t *testing.T, c client.Client) *appsv1.Deployment {
 	return dep
 }
 
-func TestReconcileKASConnectionCheckerDeployment(t *testing.T) {
+func TestReconcilerReconcileKASConnectionChecker(t *testing.T) {
```
bryan-cox: `TESTING.md` requires unit tests to use the `Test<FunctionName>` convention. Please rename this to `TestReconcileKASConnectionChecker`.

test/e2e/create_cluster_test.go
```diff
@@ -109,6 +109,7 @@ func TestCreateCluster(t *testing.T) {
 		e2eutil.EnsureCustomTolerations(t, ctx, mgtClient, hostedCluster)
 		e2eutil.EnsureAppLabel(t, ctx, mgtClient, hostedCluster)
 		e2eutil.EnsureFeatureGateStatus(t, ctx, guestClient)
+		e2eutil.EnsureKASConnectionCheckerSpec(t, ctx, guestClient, hostedCluster)
```
bryan-cox: Should this read-only hosted-cluster compliance check live in the v2 e2e suite instead? v2 already supports accessing the hosted cluster through `GetHostedClusterClient()`. If this is specifically intended to validate newly created clusters, keeping it here is reasonable.

bryan-cox (changes requested):
The implementation matches `OCPBUGS-84368`, and the focused tests pass. Please address the test naming and commit-footer items:

- `resources_test.go` should use `TestReconcileKASConnectionChecker` per `TESTING.md`.
- Add the repository-required `Signed-off-by` footer per `DEVELOPMENT.md`.

I also have a question about whether the read-only e2e assertion should move to the v2 suite.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -1798,8 +1798,12 @@ done`, endpoint, manifests.KASConnectionCheckerConfigMapName, manifests.KASConne
 		// admission, so the annotation would be inert. Worse, if that exemption ever
 		// changed, restricted-v2 (MustRunAsRange) would reject the explicit UID below
 		// because it falls outside the namespace uid-range, breaking the checker.
-		// Set to nil so the annotation is also cleared from pre-existing deployments.
-		deployment.Spec.Template.ObjectMeta.Annotations = nil
+		deployment.Spec.Template.ObjectMeta.Annotations = map[string]string{
```
csrwng: We should not clear existing annotations/labels... just ensure that the annotation we want exists.
I see that previously we were setting it to nil, but I don't think that's the best approach.
Other operators may want to add their own annotations to workloads and we should not wipe them out.

sdminonne: OK. Going to amend this

patjlm:
LGTM on my side, thanks for looking into this!

sdminonne:
## Manual Verification Results

Verified on a hosted cluster running **4.23 nightly** (`4.23.0-0.nightly-multi-2026-09-01-183453`) with a custom CPO image built from this branch (`quay.io/sminonne/hypershift-cpo:fix-kas-connection-checker-pdb-safe-to-evict-19398b2383b4`).

### 1. kas-connection-checker deployment spec

All three spec changes confirmed on the live hosted cluster:

```
$ oc get deployment kas-connection-checker -n kube-system -o jsonpath='{.spec.template.metadata.annotations}'
{"cluster-autoscaler.kubernetes.io/safe-to-evict":"true"}

$ oc get deployment kas-connection-checker -n kube-system -o jsonpath='{.spec.template.spec.tolerations}'
(empty — no custom tolerations)

$ oc get deployment kas-connection-checker -n kube-system -o jsonpath='{.spec.template.spec.topologySpreadConstraints}'
[{"labelSelector":{"matchLabels":{"app":"kas-connection-checker"}},"maxSkew":1,"topologyKey":"kubernetes.io/hostname","whenUnsatisfiable":"ScheduleAnyway"}]
```

- `safe-to-evict` annotation present
- No custom tolerations
- TopologySpreadConstraint with `ScheduleAnyway` policy
- `openshift.io/required-scc` annotation **not present** (intentionally — kube-system is in the SCC `systemNamespaces` allowlist)

### 2. Scale-down / drain validation

Scaled the NodePool from 2 replicas to 0 to trigger node drain:

- **First node drained and deleted successfully** — no drain loops or eviction failures
- **kas-connection-checker pods were evicted without blocking the drain** — all 3 pods moved to `Pending` state (no available nodes to reschedule onto)
- **Second node drain in progress** — only DaemonSet pods remain on the node during infrastructure cleanup

The `safe-to-evict` annotation ensures the cluster-autoscaler treats kas-connection-checker pods as evictable during scale-down, resolving the drain loop described in OCPBUGS-84368.

sdminonne:
/verified by @sdminonne in local dev cluster info added in comment

sdminonne:
/verified by @sdminonne in local dev cluster. Last commit touched only e2e

PR #8339, opened 2026-04-27
GCP-635: feat(gcp): make JWKS file optional when OIDC issuer URL is provided

ckandag:
Made-with: Cursor

When --oidc-issuer-url is specified for GCP IAM setup, the --oidc-jwks-file flag is no longer required. The WIF pool provider will use issuer URI discovery mode to fetch JWKS from the public OIDC endpoint instead of requiring inline JWKS JSON.

Fixes 
[GCP-635](https://redhat.atlassian.net/browse/GCP-635)

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * CLI now accepts either a JWKS file or an issuer URL; at least one is required. JWKS parsing runs only when a file is provided. JWKS comparisons normalize whitespace and treat empty inputs consistently.
* **Documentation**
  * Help text updated to clarify conditional/optional behavior of the OIDC-related flags and the issuer URL.
* **Tests**
  * Validation and JWKS tests added/updated to cover issuer-only cases, file/JSON errors, and semantic JWKS comparisons.

### DIFF
--- a/cmd/infra/gcp/iam.go
+++ b/cmd/infra/gcp/iam.go
@@ -234,15 +234,20 @@ func (c *IAMManager) CreateOIDCProvider(ctx context.Context) (string, string, er
 	c.logger.Info("Using OIDC issuer URI", "issuerURI", issuerURI)
 
 	providerAudience := c.formatProviderAudience()
+	oidc := &iam.Oidc{
+		AllowedAudiences: []string{defaultOIDCAudience},
+		IssuerUri:        issuerURI,
+		JwksJson:         jwksJson,
+	}
+	if jwksJson == "" {
+		oidc.ForceSendFields = []string{"JwksJson"}
+	}
+
 	provider := &iam.WorkloadIdentityPoolProvider{
 		Description: fmt.Sprintf("OIDC Provider for HyperShift cluster %s", c.infraID),
 		DisplayName: providerID,
 		Disabled:    false,
-		Oidc: &iam.Oidc{
-			AllowedAudiences: []string{defaultOIDCAudience},
-			IssuerUri:        issuerURI,
-			JwksJson:         jwksJson,
-		},
+		Oidc:        oidc,
 		AttributeMapping: map[string]string{
 			"google.subject": "assertion.sub",
 		},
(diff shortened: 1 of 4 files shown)


cmd/infra/gcp/iam.go
```diff
@@ -214,38 +214,42 @@ func (c *IAMManager) CreateOIDCProvider(ctx context.Context) (string, string, er
...
-		// JWKS is sufficient for the provider;
-		// Valid OIDC issuer URL option is only relevant when JWKS is not provided.
-		// In this case, the issuer URL will be derived from infraID if not provided.
 		Oidc: &iam.Oidc{
 			AllowedAudiences: []string{defaultOIDCAudience},
-			IssuerUri:        c.formatIssuerUri(),
-			JwksJson:         string(jwksData),
+			IssuerUri:        issuerURI,
+			JwksJson:         jwksJson,
 		},
```
cblecker: When no JWKS file is provided, `jwksJson` is `""`. The vendored GCP struct uses `json:"jwksJson,omitempty"`, so the empty string is silently dropped from the serialized request body. The update mask at line 883 includes `oidc.jwks_json`, but if the field isn't in the body the API won't clear an existing value — meaning a provider that previously had inline JWKS can never be transitioned to issuer-discovery mode. `ensureProviderUsable` would then detect a mismatch on every call and loop forever trying to apply the update.

Fix: set `ForceSendFields` on the `Oidc` struct when `jwksJson == ""` to force the empty string into the request body. There's existing precedent for this pattern in the same package:

- `networking.go:73` — `ForceSendFields: []string{"AutoCreateSubnetworks"}`
- `networking.go:312` — `router.ForceSendFields = []string{"Nats"}`

Something like:

```go
oidc := &iamv1.WorkloadIdentityPoolProvider_Oidc{
    AllowedAudiences: audiences,
    IssuerUri:        issuerURL,
    JwksJson:         jwksJson,
}
if jwksJson == "" {
    oidc.ForceSendFields = []string{"JwksJson"}
}
```

cblecker (commented):
One issue with how empty JWKS is handled during provider updates — see inline comment.

cmd/infra/gcp/iam_test.go
cblecker, line 1: The new `TestLoadAndValidateJWKS` and `TestCompareJWKS` functions use raw `t.Errorf`/`t.Fatalf` assertions. `AGENTS.md` prefers gomega for unit test assertions — worth adopting here (and the existing tests in this file) rather than extending the raw-assertion pattern further.

cblecker (commented):
Nit on test assertion style in the new test functions.

cmd/infra/gcp/iam_test.go
ckandag: fixed

PR #8342, opened 2026-04-27
fix(cpo): harden probe configs and retry logic for Azure self-managed CI stability

bryan-cox:

Fixes flaky `pull-ci-openshift-hypershift-main-e2e-azure-self-managed` CI tests (~25% pass rate) caused by container restarts triggering `EnsureNoCrashingPods` failures. Each commit addresses a distinct container restart root cause identified from CI failure logs.

**Fixes applied (all platforms unless noted):**

- **azure-workload-identity-webhook** (KAS sidecar, Azure-only): Liveness probe had implicit `failureThreshold=1` (K8s default), causing kills on any single slow response. Set explicit `failureThreshold=3`, `timeoutSeconds=5`. Added startup probe. This is critical because the webhook has a `MutatingWebhookConfiguration` with `FailurePolicy: Fail`, so killing it cascades to block all pod creations.
- **konnectivity-socks5-proxy**: Called `os.Exit(1)` on transient failures during `client.New()` and `NewKonnectivityDialer()` with no retry logic. Added retry loops (30 attempts × 10s backoff). Hoisted `GetConfigOrDie()` above the retry loop since it calls `os.Exit(1)` internally.
- **router** (private-router, all platforms): Had no startup probe, so the liveness probe (with default 1s timeout) could kill the container before HAProxy finished loading config. Added a startup probe (300s budget) and increased liveness timeout to 5s.
- **openshift-oauth-apiserver** (all platforms): Liveness and readiness probes had `timeoutSeconds=1`, too tight under resource contention with 9+ hosted clusters on 3 nodes. Increased to `timeoutSeconds=5`. Added startup probe for consistency.
- **openshift-apiserver** (all platforms): Readiness probe had `timeoutSeconds=1`, same issue. Increased to `timeoutSeconds=5`.

Fixes flaky `e2e-azure-self-managed` CI job

- The `TestControlPlaneComponents` fixture test has a pre-existing build failure on `main` (unrelated mock interface changes), so fixture files could not be regenerated. They will need updating once that build issue is resolved upstream.
- Commits 3-5 (router, oauth-apiserver, openshift-apiserver) affect all platforms, not just Azure. The improvements are valid everywhere.

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Increased liveness/readiness probe timeouts across components to reduce false failures and improve recovery.

* **New Features**
  * Added startup probes to gate initialization for API and routing components to avoid premature restarts.
  * Introduced bounded retry-and-log startup behavior for a critical networking proxy to improve resilience to transient failures.

* **Tests**
  * Extended unit test assertions to cover updated probe timing and failure thresholds.

### DIFF
--- a/konnectivity-socks5-proxy/main.go
+++ b/konnectivity-socks5-proxy/main.go
@@ -3,6 +3,7 @@ package konnectivitysocks5proxy
 import (
 	"fmt"
 	"os"
+	"time"
 
 	"github.com/openshift/hypershift/support/konnectivityproxy"
 	"github.com/openshift/hypershift/support/supportedversion"
@@ -16,6 +17,31 @@ import (
 	"github.com/spf13/cobra"
 	"go.uber.org/zap/zapcore"
 )
+
+const (
+	// startupMaxRetries is the maximum number of retry attempts during startup
+	// for creating the Kubernetes client and initializing the konnectivity dialer.
+	startupMaxRetries = 30
+	// startupRetryInterval is the time to wait between retry attempts.
+	startupRetryInterval = 10 * time.Second
+)
+
+// retryWithBackoff calls fn up to maxRetries times, sleeping retryInterval between
+// attempts. It returns the first nil error from fn, or the last error if all
+// attempts fail.
+func retryWithBackoff(maxRetries int, retryInterval time.Duration, fn func() error) error {
+	var lastErr error
+	for attempt := 1; attempt <= maxRetries; attempt++ {
+		lastErr = fn()
+		if lastErr == nil {
+			return nil
+		}
+		if attempt < maxRetries {
+			time.Sleep(retryInterval)
+		}
+	}
+	return lastErr
+}
 
 func NewStartCommand() *cobra.Command {
 	l := log.Log.WithName("konnectivity-socks5-proxy")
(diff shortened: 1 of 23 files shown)


konnectivity-socks5-proxy/main.go
```diff
@@ -51,19 +52,38 @@ func NewStartCommand() *cobra.Command {
...
+			c, err = client.New(ctrl.GetConfigOrDie(), client.Options{})
+			if err == nil {
+				break
+			}
+			l.Info("Failed to create client, retrying", "attempt", attempt, "error", err)
+			if attempt == 30 {
+				fmt.Fprintf(os.Stderr, "Error: cannot get client after %d attempts: %v", attempt, err)
+				os.Exit(1)
+			}
+			time.Sleep(10 * time.Second)
```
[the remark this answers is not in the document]
bryan-cox: This was addressed — `GetConfigOrDie()` is now hoisted above the retry loop (line 56). The config itself is always available (in-cluster config or mounted secret); the transient failure is API discovery inside `client.New()`, not config loading. Using `GetConfigOrDie()` is consistent with every other usage in the codebase and appropriate here since a missing kubeconfig is a permanent misconfiguration, not a transient error worth retrying.

konnectivity-socks5-proxy/main.go
```diff
@@ -51,19 +52,39 @@ func NewStartCommand() *cobra.Command {
...
+		cfg := ctrl.GetConfigOrDie()
+		var c client.Client
+		for attempt := 1; attempt <= 30; attempt++ {
+			var err error
+			c, err = client.New(cfg, client.Options{})
+			if err == nil {
+				break
+			}
+			l.Info("Failed to create client, retrying", "attempt", attempt, "error", err)
+			if attempt == 30 {
```
clebs: Should these values be picked up from the `startupProbe` above?
If not, should they be flags with defaults?
If also no, they should be well defined constants instead of magic numbers.

bryan-cox: Extracted to named constants `startupMaxRetries` and `startupRetryInterval`. These are startup-time retry values for API client creation, not related to the Kubernetes startup probe (which governs the container's health check lifecycle). Flags would be over-engineering since these values don't need runtime tuning.

konnectivity-socks5-proxy/main.go
```diff
@@ -51,19 +60,39 @@ func NewStartCommand() *cobra.Command {
...
+			if err == nil {
+				break
+			}
+			l.Info("Failed to initialize konnectivity dialer, retrying", "attempt", attempt, "error", err)
+			if attempt == startupMaxRetries {
+				fmt.Fprintf(os.Stderr, "Error: cannot initialize konnectivity dialer after %d attempts: %v", attempt, err)
+				os.Exit(1)
+			}
+			time.Sleep(startupRetryInterval)
 		}
```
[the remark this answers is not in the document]
bryan-cox: The retry is intentional. `NewKonnectivityDialer` reads TLS cert/key files from mounted secrets (lines 164-175 in `dialer.go`). During initial pod startup, there's a narrow window where the volume hasn't fully propagated yet — `os.ReadFile` fails with a transient file-not-found. The second attempt succeeds once the kubelet finishes the volume mount. This is the exact failure pattern we observed in CI logs causing the container restarts.

Validation errors (wrong flags) would be permanent, but those are caught during development — they don't change at runtime. The file reads are the transient case worth retrying.

konnectivity-socks5-proxy/main.go
```diff
@@ -17,6 +18,31 @@ import (
...
+	// for creating the Kubernetes client and initializing the konnectivity dialer.
+	startupMaxRetries = 30
+	// startupRetryInterval is the time to wait between retry attempts.
+	startupRetryInterval = 10 * time.Second
+)
+
+// retryWithBackoff calls fn up to maxRetries times, sleeping retryInterval between
+// attempts. It returns the first nil error from fn, or the last error if all
+// attempts fail.
+func retryWithBackoff(maxRetries int, retryInterval time.Duration, fn func() error) error {
```
clebs, line 32: We should use `k8s.io/client-go/util/retry`'s `OnError` utility that already provides this logic.

bryan-cox:
/close

I don't think we need this now

PR #8344, opened 2026-04-27
OCPBUGS-84509: Add AWS LB Controller cross-zone load balancing annotation

typeid:
## Summary

- HyperShift sets cross-zone load balancing on private NLBs using the in-tree AWS cloud provider annotation (`aws-load-balancer-cross-zone-load-balancing-enabled`). The AWS Load Balancer Controller (used by EKS Auto Mode) does not recognize this annotation, resulting in NLBs created without cross-zone load balancing.
- Adds the AWS Load Balancer Controller annotation (`aws-load-balancer-attributes: "load_balancing.cross_zone.enabled=true"`) alongside the existing one in all three locations: `ReconcileRouterService`, `ReconcileService`, and `ReconcilePrivateService`.

## Background

The in-tree AWS cloud provider and the AWS Load Balancer Controller use different Service annotations to configure the same NLB behavior. The in-tree provider uses `service.beta.kubernetes.io/aws-load-balancer-cross-zone-load-balancing-enabled` ([k8s docs](https://kubernetes.io/docs/concepts/services-networking/service/#aws-nlb-support)), while the AWS LB Controller uses `service.beta.kubernetes.io/aws-load-balancer-attributes` with the `load_balancing.cross_zone.enabled` attribute ([AWS LB Controller docs](https://kubernetes-sigs.github.io/aws-load-balancer-controller/latest/guide/service/annotations/#additive-annotations)). Both annotations can safely coexist — each controller ignores the other's annotations.

Cross-zone load balancing is required for PrivateLink connectivity. Without it, connections from a VPC Endpoint that land on an availability zone with no management cluster node are silently dropped ([AWS cross-zone docs](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/network-load-balancers.html#cross-zone-load-balancing)).

## Jira
https://issues.redhat.com/browse/OCPBUGS-84509

## Summary by CodeRabbit

* **New Features**
  * Enabled AWS cross-zone load balancing for control plane API server, router, and private services to improve traffic distribution across availability zones.

* **Tests**
  * Updated test validations to verify the AWS load balancer attribute configuration for cross-zone behavior.

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/infra/infra_test.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/infra/infra_test.go
@@ -1212,6 +1212,7 @@ func TestReconcileAPIServerService(t *testing.T) {
 	}
 	withCrossZoneAnnotation := func(svc *corev1.Service) {
 		svc.Annotations["service.beta.kubernetes.io/aws-load-balancer-cross-zone-load-balancing-enabled"] = "true"
+		svc.Annotations["service.beta.kubernetes.io/aws-load-balancer-attributes"] = "load_balancing.cross_zone.enabled=true"
 	}
 	withLoadBalancerSourceRanges := func(svc *corev1.Service) {
 		svc.Spec.LoadBalancerSourceRanges = allowCIDRString
@@ -1491,6 +1492,7 @@ func TestReconcileHCPRouterServices(t *testing.T) {
 	}
 	withCrossZoneAnnotation := func(svc *corev1.Service) {
 		svc.Annotations["service.beta.kubernetes.io/aws-load-balancer-cross-zone-load-balancing-enabled"] = "true"
+		svc.Annotations["service.beta.kubernetes.io/aws-load-balancer-attributes"] = "load_balancing.cross_zone.enabled=true"
 	}
 	tests := []struct {
 		name                         string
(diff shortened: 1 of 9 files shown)


typeid:
/verified by @typeid 

I was able to run this build on EKS using the AWS LB Controller: the LBs come up with cross-zone load balancing: on.

control-plane-operator/controllers/hostedcontrolplane/infra/infra_test.go
```diff
@@ -1491,6 +1492,7 @@ func TestReconcileHCPRouterServices(t *testing.T) {
 	}
 	withCrossZoneAnnotation := func(svc *corev1.Service) {
 		svc.Annotations["service.beta.kubernetes.io/aws-load-balancer-cross-zone-load-balancing-enabled"] = "true"
+		svc.Annotations["service.beta.kubernetes.io/aws-load-balancer-attributes"] = "load_balancing.cross_zone.enabled=true"
```
enxebre, line 1495: can we add a comment on why/who consumes each annotation, i.e. same than you already included in the PR desc so it's immediately clear for agents when figuring things out

typeid: Makes sense, done. :)

typeid:
/verified by @typeid

I was able to run this build on EKS using the AWS LB Controller: the LBs come up with cross-zone load balancing: on.

PR #8352, opened 2026-04-27
OCPBUGS-77268: reconcile HCP when pull secret is unavailable

rafael-azevedo:

When the pull secret is missing or invalid, the HostedCluster reconciler exits before reaching reconcileHostedControlPlane. This prevents RequestServingNodeAdditionalSelector and other spec fields from being propagated to the HCP, breaking HA scheduling for request-serving nodes.

Log the error and call reconcileHostedControlPlane before returning so pod placement decisions continue during a pull secret outage. The error is still returned to trigger a requeue for full reconciliation once the pull secret is restored.

JIRA: OCPBUGS-77268

Fixes https://issues.redhat.com/browse/OCPBUGS-77268

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * When a pull secret fetch fails, reconciliation now logs the error and triggers a recovery reconcile of the control plane so pod placement decisions (nodeSelector, autoscaling, node-termination handler) and certificate-related metadata are still processed.

* **Tests**
  * Added unit and parallel e2e tests simulating pull-secret unavailability to verify propagation of nodeSelector, annotations and release image to the control plane and recovery after restoration.

### DIFF
--- a/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
+++ b/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
@@ -655,7 +655,30 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
 
 	pullSecretBytes, err := hyperutil.GetPullSecretBytes(ctx, r.Client, hcluster)
 	if err != nil {
-		return ctrl.Result{}, err
+		log.Error(err, "failed to get pull secret")
+		// Ensure even with a bad pull secret we process pod placement decisions
+		// before returning the pull secret error for requeue.
+		hcp = controlplaneoperator.HostedControlPlane(controlPlaneNamespace.Name, hcluster.Name)
+		isAutoscalingNeeded, autoscaleErr := r.isAutoscalingNeeded(ctx, hcluster)
+		if autoscaleErr != nil {
+			log.Error(autoscaleErr, "failed to determine if autoscaler is needed during pull secret recovery, defaulting to true")
+			isAutoscalingNeeded = true
+		}
+		isAWSNodeTerminationHandlerNeeded, nthErr := r.isAWSNodeTerminationHandlerNeeded(ctx, hcluster)
+		if nthErr != nil {
+			log.Error(nthErr, "failed to determine if AWS node termination handler is needed during pull secret recovery, defaulting to true")
+			isAWSNodeTerminationHandlerNeeded = true
+		}
+		_, hcpErr := createOrUpdate(ctx, r.Client, hcp, func() error {
+			// Skip cert annotation resolution during pull secret recovery — it requires
+			// the pull secret to resolve the CPO image, which is unavailable here.
+			return reconcileHostedControlPlane(hcp, hcluster, isAutoscalingNeeded, isAWSNodeTerminationHandlerNeeded,
+				func() (map[string]string, error) { return nil, nil })
+		})
+		if hcpErr != nil {
+			log.Error(hcpErr, "failed to reconcile hostedcontrolplane during pull secret recovery")
+		}
+		return ctrl.Result{}, fmt.Errorf("failed to get pull secret: %w", err)
 	}
 
 	controlPlaneOperatorImage, err := hyperutil.GetControlPlaneOperatorImage(ctx, hcluster, releaseProvider, r.HypershiftOperatorImage, pullSecretBytes)
--- a/hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
+++ b/hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
@@ -256,6 +256,195 @@ func TestHasBeenAvailable(t *testing.T) {
 	}
 }
 
+func TestReconcileWhenPullSecretMissing(t *testing.T) {
+	t.Parallel()
+	g := NewWithT(t)
+
+	hcluster := &hyperv1.HostedCluster{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "test-cluster",
+			Namespace: "any",
+			Annotations: map[string]string{
+				hyperv1.RequestServingNodeAdditionalSelectorAnnotation: `{"node-role.kubernetes.io/worker": ""}`,
+			},
+		},
+		Spec: hyperv1.HostedClusterSpec{
+			ClusterID: "id",
+			InfraID:   "infra-id",
+			Networking: hyperv1.ClusterNetworking{
+				ClusterNetwork: []hyperv1.ClusterNetworkEntry{{CIDR: *ipnet.MustParseCIDR("172.16.1.0/24")}},
+			},
+			Services: []hyperv1.ServicePublishingStrategyMapping{
+				{
+					Service: hyperv1.Ignition,
+					ServicePublishingStrategy: hyperv1.ServicePublishingStrategy{
+						Type: hyperv1.LoadBalancer,
+					},
+				},
+			},
+			PullSecret: corev1.LocalObjectReference{
+				Name: "pull-secret",
+			},
+			Release: hyperv1.Release{Image: "quay.io/openshift-release-dev/ocp-release:4.15.0"},
+			Etcd:    hyperv1.EtcdSpec{ManagementType: hyperv1.Managed},
+			Platform: hyperv1.PlatformSpec{
+				Type: hyperv1.AWSPlatform,
+			},
+			NodeSelector: map[string]string{
+				"node-role.kubernetes.io/worker": "",
+			},
+		},
+	}
+
+	hcpNs := hcpmanifests.HostedControlPlaneNamespace(hcluster.Namespace, hcluster.Name)
+	hcp := controlplaneoperator.HostedControlPlane(hcpNs, hcluster.Name)
+
+	// Create the namespace so createOrUpdate can find it
+	ns := &corev1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{Name: hcpNs},
+	}
+
+	// No pull secret in the fake client — this is the scenario we're testing
+	objects := []crclient.Object{
+		hcp,
+		hcluster,
+		ns,
+	}
+
+	client := fake.NewClientBuilder().WithScheme(api.Scheme).WithObjects(objects...).WithStatusSubresource(hcluster).Build()
+
+	mockCtrl := gomock.NewController(t)
+	defer mockCtrl.Finish()
+	mockedProvider := releaseinfo.NewMockProviderWithOpenShiftImageRegistryOverrides(mockCtrl)
+
+	r := &HostedClusterReconciler{
+		Client:                        client,
+		Clock:                         clocktesting.NewFakeClock(time.Now()),
+		CertRotationScale:             24 * time.Hour,
+		createOrUpdate:                func(reconcile.Request) upsert.CreateOrUpdateFN { return ctrl.CreateOrUpdate },
+		ManagementClusterCapabilities: &fakecapabilities.FakeSupportNoCapabilities{},
+		RegistryProvider: fakeReleaseProvider{
+			releaseProvider: mockedProvider,
+			metadataProvider: fakeimagemetadataprovider.FakeRegistryClientImageMetadataProvider{
+				Result: &dockerv1client.DockerImageConfig{},
+			},
+		},
+		now: func() metav1.Time { return metav1.NewTime(time.Now()) },
+	}
+
+	ctx := t.Context()
+	_, err := r.Reconcile(ctx, ctrl.Request{NamespacedName: crclient.ObjectKeyFromObject(hcluster)})
+
+	// When the pull secret is missing, reconciliation should return an error for requeue
+	g.Expect(err).To(HaveOccurred(), "When pull secret is missing it should return an error")
+	g.Expect(err.Error()).To(ContainSubstring("failed to get pull secret"), "When pull secret is missing the error should mention pull secret")
+
+	// Verify reconcileHostedControlPlane still ran — the HCP should have the HostedCluster's spec fields
+	updatedHCP := controlplaneoperator.HostedControlPlane(hcpNs, hcluster.Name)
+	err = client.Get(ctx, crclient.ObjectKeyFromObject(updatedHCP), updatedHCP)
+	g.Expect(err).ToNot(HaveOccurred(), "When pull secret is missing it should still be able to get the HCP")
+
+	g.Expect(updatedHCP.Spec.NodeSelector).To(Equal(hcluster.Spec.NodeSelector),
+		"When pull secret is missing it should still propagate NodeSelector to HCP")
+	g.Expect(updatedHCP.Annotations).To(HaveKeyWithValue(
+		hyperv1.RequestServingNodeAdditionalSelectorAnnotation,
+		hcluster.Annotations[hyperv1.RequestServingNodeAdditionalSelectorAnnotation]),
+		"When pull secret is missing it should still propagate RequestServingNodeAdditionalSelector to HCP")
+	g.Expect(updatedHCP.Spec.ReleaseImage).To(Equal(hcluster.Spec.Release.Image),
+		"When pull secret is missing it should still propagate ReleaseImage to HCP")
+
+	// With no NodePools in the fake client, autoscaling is not needed so the
+	// autoscaler should be disabled. This validates the defaulting logic in the
+	// pull secret recovery path.
+	g.Expect(updatedHCP.Annotations).To(HaveKeyWithValue(
+		hyperv1.DisableClusterAutoscalerAnnotation, "true"),
+		"When pull secret is missing and no NodePools exist it should disable the cluster autoscaler")
+}
+
+func TestReconcileWhenPullSecretMissingAndNodePoolListFails(t *testing.T) {
+	t.Parallel()
+	g := NewWithT(t)
+
+	hcluster := &hyperv1.HostedCluster{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "test-cluster",
+			Namespace: "any",
+		},
+		Spec: hyperv1.HostedClusterSpec{
+			ClusterID: "id",
+			InfraID:   "infra-id",
+			Networking: hyperv1.ClusterNetworking{
+				ClusterNetwork: []hyperv1.ClusterNetworkEntry{{CIDR: *ipnet.MustParseCIDR("172.16.1.0/24")}},
+			},
+			Services: []hyperv1.ServicePublishingStrategyMapping{
+				{
+					Service: hyperv1.Ignition,
+					ServicePublishingStrategy: hyperv1.ServicePublishingStrategy{
+						Type: hyperv1.LoadBalancer,
+					},
+				},
+			},
+			PullSecret: corev1.LocalObjectReference{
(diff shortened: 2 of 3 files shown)


hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -667,6 +667,29 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
...
+			return reconcileHostedControlPlane(hcp, hcluster, isAutoscalingNeeded, isAWSNodeTerminationHandlerNeeded,
+				annotationsForCertRenewal(log,
+					hcp,
+					shouldCheckForStaleCerts(hcluster, false),
+					r.kasServingCertHashFromSecret(ctx, hcp),
+					r.kasServingCertHashFromEndpoint(kasHostAndPortFromHCP(hcp))))
+		})
+		if hcpErr != nil {
+			log.Error(hcpErr, "failed to reconcile hostedcontrolplane during pull secret recovery")
+		}
```
joshbranham, line 680: Is this essentially the same logic that happens later in the reconcile func if the pull secret had been valid? If so, should we make a private func to try and reuse the logic, or do you feel it's clearer this way?

rafael-azevedo: It's the same reconcileHostedControlPlane call, but in the degraded path we hardcode defaultToControlPlaneV2=false since CPO labels aren't available without the pull secret. I kept it inline to reduce adding complexity for one call. Happy to make a helper func if you think we should.

joshbranham: If it's different then seems fine to me to leave it here, we don't tend to constrain function complexity anyways afaict.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -667,6 +667,29 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
...
+			return ctrl.Result{}, fmt.Errorf("failed to determine if autoscaler is needed: %w", autoscaleErr)
+		}
+		isAWSNodeTerminationHandlerNeeded, nthErr := r.isAWSNodeTerminationHandlerNeeded(ctx, hcluster)
+		if nthErr != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to determine if AWS node termination handler is needed: %w", nthErr)
+		}
+		_, hcpErr := createOrUpdate(ctx, r.Client, hcp, func() error {
+			return reconcileHostedControlPlane(hcp, hcluster, isAutoscalingNeeded, isAWSNodeTerminationHandlerNeeded,
+				annotationsForCertRenewal(log,
+					hcp,
```
cblecker: I saw CodeRabbit flagged the hardcoded `false` here and it was closed — can you help me understand why this is safe for CPOv2 clusters?

When `defaultToControlPlaneV2` would normally be `true`, `shouldCheckForStaleCerts` returns `false` (skips the check). Hardcoding `false` here means the check runs for CPOv2 clusters that have been available, which could trigger `kasServingCertHashFromEndpoint` — a `tls.Dial` with no timeout — and potentially set `RestartDateAnnotation` on a hash mismatch.

If you've thought through this and it's fine in practice (e.g. because the lazy closure short-circuits before reaching the dial), a brief code comment explaining why the hardcoded `false` is safe here would help future readers. Otherwise, a noop cert annotation function would side-step the concern entirely.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -667,6 +667,29 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
 
 	pullSecretBytes, err := hyperutil.GetPullSecretBytes(ctx, r.Client, hcluster)
 	if err != nil {
+		log.Error(err, "failed to get pull secret")
+		// Ensure even with a bad pull secret we process pod placement decisions
+		// before returning the pull secret error for requeue.
```
cblecker, line 660: If either of these checks fails, the error returned here replaces the original pull-secret error — so the requeue happens on the wrong signal and you lose observability into the root cause. Since this is already a recovery path, would it make sense to log the error and default to a safe value (e.g. `isAutoscalingNeeded = true`) so the original pull-secret error is always what gets returned?

cblecker (commented):
The fix looks solid — propagating `reconcileHostedControlPlane` into the pull-secret error path so spec fields still reach the HCP is the right call. Two inline suggestions worth a look before merging.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
rafael-azevedo: This is a good catch. I am addressing this now. I had read it as hardcoding false would bypass the cert checks. I am just passing in an empty func now bypassing the cert stuff completely.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
rafael-azevedo: I've put in better logging as well as set set the `isAutoscalingNeeded = true`

hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
```diff
@@ -252,6 +252,103 @@ func TestHasBeenAvailable(t *testing.T) {
 	}
 }
 
+func TestReconcileWhenPullSecretMissing(t *testing.T) {
```
cblecker, line 259: nit: consider adding `t.Parallel()` here — almost every other test in this file uses it.

cblecker (approved):
Both of my previous comments are addressed — the recovery path now uses safe defaults with proper logging and a noop cert annotation function. The fix looks good.

cblecker:
/approve cancel
Didn't mean to approve this one. LGTM, but would love hypershift core team to also approve

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
cblecker: The noop function is the right call. Thanks for addressing this.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
cblecker: Looks good, thanks for addressing this.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
rafael-azevedo: added !

test/e2e/chaos_test.go
```diff
@@ -459,6 +459,171 @@ func testSingleMemberRecovery(parentCtx context.Context, client crclient.Client,
...
+				} else {
+					delete(hc.Annotations, hyperv1.RequestServingNodeAdditionalSelectorAnnotation)
+				}
+				_ = mgtClient.Update(ctx, hc)
+			}
+		})
+
+		// Corrupt the pull secret
+		t.Log("Corrupting pull secret...")
+		pullSecret.Data[".dockerconfigjson"] = []byte(`{"auths":"not-an-object"}`)
```
cblecker: I don't think this corruption payload actually exercises the new code path from this PR.

`GetPullSecretBytes` (support/util/util.go:449-461) only checks two things: (1) does the Secret exist, and (2) does it have a `.dockerconfigjson` key. It doesn't validate the content. Since this corruption preserves the key, `GetPullSecretBytes` succeeds and returns the corrupted bytes — the new degraded-mode `reconcileHostedControlPlane` is never entered.

From there, the corrupted bytes flow to `GetControlPlaneOperatorImage`. If `ControlPlaneOperatorImageAnnotation` is set on the HC (which it likely is in a real cluster), the function short-circuits at line 477 without ever using the bytes, and reconciliation proceeds through the normal path. The test passes, but it's testing the existing behavior, not the fix. If the annotation isn't set, `releaseProvider.Lookup` fails with the bad credentials, hitting a different early return at line 656 — again, before the new code runs.

To reliably trigger the `GetPullSecretBytes` failure path this PR adds, delete the key instead:
```go
delete(pullSecret.Data, ".dockerconfigjson")
```

This makes `GetPullSecretBytes` return `"expected .dockerconfigjson key in pull secret"` and actually enters the new recovery block. (The unit test gets this right by omitting the Secret entirely.)

test/e2e/chaos_test.go
```diff
@@ -459,6 +459,171 @@ func testSingleMemberRecovery(parentCtx context.Context, client crclient.Client,
...
+		// Ensure pull secret and test-owned keys are restored even if the test fails mid-way
+		t.Cleanup(func() {
+			t.Log("Cleanup: restoring pull secret and test-owned HostedCluster keys...")
+			secret := &corev1.Secret{}
+			if err := mgtClient.Get(ctx, crclient.ObjectKey{
+				Namespace: hostedCluster.Namespace,
+				Name:      hostedCluster.Spec.PullSecret.Name,
+			}, secret); err == nil {
+				secret.Data[".dockerconfigjson"] = originalData
+				_ = mgtClient.Update(ctx, secret)
```
cblecker: nit: If this Update fails (e.g. 409 conflict from a concurrent reconciler), the pull secret stays corrupted with no log trace, which could cause confusing failures in the framework's post-test validation. Same for the HC update on line 518.

Since you can't `t.Error` from a cleanup func, maybe just log it:
```go
if err := mgtClient.Update(ctx, secret); err != nil {
    t.Logf("WARNING: failed to restore pull secret during cleanup: %v", err)
}
```

cblecker (commented):
The fix itself looks correct — the degraded-mode `reconcileHostedControlPlane` with noop cert annotations and safe defaults (true for autoscaling/NTH) is well-designed, and the original pull-secret error is properly preserved for requeue. The unit test solidly validates the new code path.

Main concern is with the e2e test — the corruption payload doesn't actually trigger `GetPullSecretBytes` to fail, so it may not be exercising the code this PR adds. Details in the inline comment.

jparrill (commented):

test/e2e/chaos_test.go
jparrill: I don't think this corruption triggers the new code path. `GetPullSecretBytes` (support/util/util.go:449-461) only checks that the Secret exists and has the `.dockerconfigjson` key — it doesn't validate the content. So `{"auths":"not-an-object"}` passes `GetPullSecretBytes` successfully and the normal reconcile path runs, not the new recovery block.

To actually exercise the fix, you'd need to either delete the key:
```go
delete(pullSecret.Data, ".dockerconfigjson")
```
or delete the Secret entirely (which is what the unit test does correctly).

As written, the test passes but it's testing the existing behavior, not the fix.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -649,6 +649,27 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
...
+			isAutoscalingNeeded = true
+		}
+		isAWSNodeTerminationHandlerNeeded, nthErr := r.isAWSNodeTerminationHandlerNeeded(ctx, hcluster)
+		if nthErr != nil {
+			log.Error(nthErr, "failed to determine if AWS node termination handler is needed during pull secret recovery, defaulting to true")
+			isAWSNodeTerminationHandlerNeeded = true
+		}
+		_, hcpErr := createOrUpdate(ctx, r.Client, hcp, func() error {
+			return reconcileHostedControlPlane(hcp, hcluster, isAutoscalingNeeded, isAWSNodeTerminationHandlerNeeded,
+				func() (map[string]string, error) { return nil, nil })
```
jparrill, line 676: Worth a one-liner explaining why this is intentionally a no-op? Future readers might assume it's a bug. Something like:
```go
// Skip cert renewal during pull secret recovery — cert checks depend on
// a functioning control plane which may be degraded without the pull secret.
func() (map[string]string, error) { return nil, nil },
```

hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
```diff
@@ -252,6 +252,104 @@ func TestHasBeenAvailable(t *testing.T) {
...
+	g.Expect(err).ToNot(HaveOccurred(), "When pull secret is missing it should still be able to get the HCP")
+
+	g.Expect(updatedHCP.Spec.NodeSelector).To(Equal(hcluster.Spec.NodeSelector),
+		"When pull secret is missing it should still propagate NodeSelector to HCP")
+	g.Expect(updatedHCP.Annotations).To(HaveKeyWithValue(
+		hyperv1.RequestServingNodeAdditionalSelectorAnnotation,
+		hcluster.Annotations[hyperv1.RequestServingNodeAdditionalSelectorAnnotation]),
+		"When pull secret is missing it should still propagate RequestServingNodeAdditionalSelector to HCP")
+	g.Expect(updatedHCP.Spec.ReleaseImage).To(Equal(hcluster.Spec.Release.Image),
+		"When pull secret is missing it should still propagate ReleaseImage to HCP")
```
jparrill, line 354: The test verifies NodeSelector, annotation, and ReleaseImage propagation — which covers the core scenario well. Two things I think would round it out:

1. Assert the autoscaling/NTH annotation behavior: with no NodePools in the fake client, `isAutoscalingNeeded` returns `false`, so `DisableClusterAutoscalerAnnotation` should be `"true"` on the HCP. This locks down the defaulting logic.
2. Consider converting to table-driven to make it easy to add a sub-case where the client fails on List (to verify the "default to true" error path).

Not blocking, but the defaulting behavior is the main safety net of the fix and it'd be nice to have it pinned.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -649,6 +649,27 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
...
+		})
+		if hcpErr != nil {
+			log.Error(hcpErr, "failed to reconcile hostedcontrolplane during pull secret recovery")
+		}
 		return ctrl.Result{}, err
```
jparrill: Nit: the surrounding error returns all wrap with context (`fmt.Errorf("failed to get controlPlaneOperatorImage: %w", err)` at line 676, etc.) but this one returns the raw error. `GetPullSecretBytes` does its own wrapping internally so it's not broken, just inconsistent with the pattern in this function.

test/e2e/chaos_test.go
jparrill: Minor: if the cleanup `Update` gets a 409 conflict (possible with the reconciler running concurrently), the pull secret stays corrupted silently because of `_ = mgtClient.Update(...)`. Adding a `t.Logf` for the error case would help debug flakes:
```go
if err := mgtClient.Update(ctx, secret); err != nil {
    t.Logf("cleanup: failed to restore pull secret: %v", err)
}
```

rafael-azevedo:
@jparrill @cblecker Thanks I've addressed the comments !

cblecker (approved):
LGTM. Will leave the final sign off to @jparrill

test/e2e/chaos_test.go
```diff
@@ -459,6 +459,215 @@ func testSingleMemberRecovery(parentCtx context.Context, client crclient.Client,
...
+		g.Expect(originalData).NotTo(BeEmpty(), "pull secret should have .dockerconfigjson data")
+
+		// Save original values of test-owned keys for cleanup
+		err = mgtClient.Get(ctx, crclient.ObjectKeyFromObject(hostedCluster), hostedCluster)
+		g.Expect(err).NotTo(HaveOccurred(), "failed to get hostedcluster")
+		originalSelectorVal, hadSelector := hostedCluster.Spec.NodeSelector["ocpbugs-77268-test"]
+		originalAnnotationVal, hadAnnotation := hostedCluster.Annotations[hyperv1.RequestServingNodeAdditionalSelectorAnnotation]
+
+		// Ensure pull secret and test-owned keys are restored even if the test fails mid-way
+		t.Cleanup(func() {
```
cblecker, line 497: The `t.Cleanup` here and the manual restoration at the end of the test body (lines 616-640) both restore the same state. Do we need both? Would it work to consolidate into just `t.Cleanup` (adding the recovery `EventuallyObject` wait there too) to avoid doing the restoration twice on the happy path?

hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
```diff
@@ -252,6 +252,111 @@ func TestHasBeenAvailable(t *testing.T) {
...
+		},
+		now: func() metav1.Time { return metav1.NewTime(time.Now()) },
+	}
+
+	ctx := t.Context()
+	_, err := r.Reconcile(ctx, ctrl.Request{NamespacedName: crclient.ObjectKeyFromObject(hcluster)})
+
+	// When the pull secret is missing, reconciliation should return an error for requeue
+	g.Expect(err).To(HaveOccurred(), "When pull secret is missing it should return an error")
+	g.Expect(err.Error()).To(ContainSubstring("pull secret"), "When pull secret is missing the error should mention pull secret")
```
cblecker: nit: The controller wraps this as `"failed to get pull secret: %w"`. A tighter assertion would avoid false positives if a different code path happens to mention "pull secret" in its error:

```suggestion
	g.Expect(err.Error()).To(ContainSubstring("failed to get pull secret"), "When pull secret is missing the error should mention pull secret")
```

hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
```diff
@@ -252,6 +252,111 @@ func TestHasBeenAvailable(t *testing.T) {
...
+		hcluster.Annotations[hyperv1.RequestServingNodeAdditionalSelectorAnnotation]),
+		"When pull secret is missing it should still propagate RequestServingNodeAdditionalSelector to HCP")
+	g.Expect(updatedHCP.Spec.ReleaseImage).To(Equal(hcluster.Spec.Release.Image),
+		"When pull secret is missing it should still propagate ReleaseImage to HCP")
+
+	// With no NodePools in the fake client, autoscaling is not needed so the
+	// autoscaler should be disabled. This validates the defaulting logic in the
+	// pull secret recovery path.
+	g.Expect(updatedHCP.Annotations).To(HaveKeyWithValue(
+		hyperv1.DisableClusterAutoscalerAnnotation, "true"),
```
cblecker, line 360: This validates the happy path (no error, no NodePools → autoscaling not needed). But the recovery path also has error-defaulting logic that defaults `isAutoscalingNeeded = true` and `isAWSNodeTerminationHandlerNeeded = true` on error. Would it be worth adding a test case that exercises those defaults?

cblecker (commented):
Review of the test changes — a few suggestions on the e2e cleanup pattern and unit test assertions.

rafael-azevedo:
@jparrill Looks like everything is good to go. My tests are passing , but some failures unrelated to the PR possible flakes are occuring.

enxebre:
PTAL @muraee re https://github.com/openshift/hypershift/pull/7908

cblecker:
confirmed a clean rebase. reapplying..
/lgtm

PR #8355, opened 2026-04-28
NO-JIRA: Do not import external dependencies into API packages

JoelSpeed:

By including utility functionality in the API module, this pollutes the imports and causes issues for external contributors. For example, having the karpenter dependency pins external importers of the APIs to K8s 0.34 because of the reliance on an outdated openshift fork of karpenter.

Utils should live separately, in this case they were only used in one location so I've moved them directly to the call site package.

Please do not allow util functions in the API module again.

Fixes 

## Checklist:

## Summary by CodeRabbit

* **Chores**
  * Simplified the project's dependency set by removing numerous unused transitive packages to improve build clarity and maintainability.
  * Removed several compatibility overrides and replace directives to streamline dependency resolution and reduce potential conflicts.
  * Added a small lightweight text utility dependency to support development and testing tasks.


JoelSpeed:
/verified by @JoelSpeed 

I tested this locally and it appears to work for my purpose. Disconnects the karpenter link when I revendor this into ARO

PR #8358, opened 2026-04-28
OCPBUGS-38898: fix kubeadmin password hash reconciliation to compare password against stored hash

dhgautam99:

Previously, `ReconcileKubeadminPasswordHashSecret` would skip hash generation
whenever a non-empty `kubeadmin` hash already existed in the secret, without
verifying that the hash actually corresponded to the current password. This
meant that if the password secret was updated, the stale bcrypt hash would
persist and the new password would never take effect.

A concrete example is etcd restoration: after restoring an HCP cluster, a new
kubeadmin password is generated, but the restored secret still contains the old
bcrypt hash. Since the old hash is non-empty, the reconciler would short-circuit
and never regenerate, leaving the hash permanently out of sync with the new
password.

This PR fixes the reconciliation logic to use `bcrypt.CompareHashAndPassword`
to verify the existing hash matches the current password before short-circuiting.
If the comparison fails (password changed), a new hash is generated and stored.

Fixes OCPBUGS-38898

- The `password` variable read is moved before the early-return check so it is
  available for the `bcrypt.CompareHashAndPassword` call.
- No new dependencies are introduced; `bcrypt` was already imported.

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Admin password handling now verifies stored password hashes and only treats them as current when they match the provided plaintext; stale or invalid hashes are regenerated and updated to maintain reliable access.

* **Tests**
  * Expanded unit tests for admin password reconciliation to cover matching, stale, and empty-hash scenarios and to validate that reconciled hashes authenticate correctly against the source plaintext.

### DIFF
--- a/control-plane-operator/hostedclusterconfigoperator/controllers/resources/kubeadminpassword/reconcile.go
+++ b/control-plane-operator/hostedclusterconfigoperator/controllers/resources/kubeadminpassword/reconcile.go
@@ -9,13 +9,15 @@ import (
 )
 
 func ReconcileKubeadminPasswordHashSecret(secret *corev1.Secret, passwordSecret *corev1.Secret) error {
+	password := passwordSecret.Data["password"]
 	if secret.Data != nil {
 		hash, hasHash := secret.Data["kubeadmin"]
 		if hasHash && len(hash) > 0 {
-			return nil
+			if bcrypt.CompareHashAndPassword(hash, password) == nil {
+				return nil
+			}
 		}
 	}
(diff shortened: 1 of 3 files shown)


clebs:
@dhgautam99 you need the `verified` label to merge this.
Either assign to QE for validation or if there are tests that verify the correctness you can add the label yourself.

dhgautam99:
/verified by @dhgautam99 
In my local cluster testing

PR #8364, opened 2026-04-29
NO-JIRA: Add API dependency import restriction tool

JoelSpeed:

Follow up to #8355 

The more polluted the import list is, the harder it is for consumers to integrate HyperShift into their projects. For example, we used to have karpenter as a dependency, this meant that any HyperShift consumer also had karpenter as a dependency, and must now manage both the HyperShift dependency compatibility but also the karpenter dependency compatibility.

It is best practice not to include implementation in the API module and to keep the module strictly to APIs.

This PR add `make verify-api-deps` to prevent folks from adding new dependencies until/unless they convince an API reviewer that there is a need.

Fixes 

## Checklist:

## Summary by CodeRabbit

* **Chores**
  * Added a new verify-api-deps check to validate the API module’s direct dependencies against an approved allowlist; runs locally and as part of parallel verification in CI.
  * Included a standalone CLI invocation to perform the dependency allowlist check.

* **Documentation**
  * Added user-facing docs describing the allowlist format, default allowed modules (Kubernetes core and OpenShift API), how to run the check, and the workflow for requesting allowlist changes.

### DIFF
--- a/hack/tools/verify-api-deps/main.go
+++ b/hack/tools/verify-api-deps/main.go
@@ -21,8 +21,13 @@ func main() {
 }
 
 func verifyAPIDependencies() error {
-	// The API module is always at ./api relative to the repository root
-	apiModPath := "api"
+	// Find the repository root and locate the API module
+	repoRoot, err := findRepoRoot()
+	if err != nil {
+		return fmt.Errorf("failed to find repository root: %w", err)
+	}
+
+	apiModPath := filepath.Join(repoRoot, "api")
 
 	// Load allowed dependencies from the .imports_allowed file
 	allowedAPIModules, err := loadAllowedImports(apiModPath)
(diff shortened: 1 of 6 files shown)


hack/tools/verify-api-deps/main.go
```diff
@@ -0,0 +1,98 @@
...
+// Note: Indirect dependencies are automatically ignored by the verification logic.
+var allowedAPIModules = sets.New(
+	// Core Kubernetes API dependencies
+	"k8s.io/api",
+	"k8s.io/apimachinery",
+	"k8s.io/utils",
+
+	// OpenShift API dependencies
+	"github.com/openshift/api",
+)
```
JoelSpeed: Possibly this would be better as a file in the api folder so that it comes under the API OWNERS file rather than the root OWNERS file

enxebre:
can we please include a ref to this make target here https://github.com/openshift/hypershift/blob/main/api/AGENTS.md#api-versioning

JoelSpeed:
> can we please include a ref to this make target here https://github.com/openshift/hypershift/blob/main/api/AGENTS.md#api-versioning

It's already included in the verify, do you still want me to put it in separately as an early/quick check?

enxebre:
>It's already included in the verify, do you still want me to put it in separately as an early/quick check?

yes please, just like we have api-lint, api-lint-fix... in that list so it has a list of targeted items without needing to run verify which is more expensive all the time but only once.
we'll be putting evals for this soon.

PR #8365, opened 2026-04-29
OCPBUGS-77307: Generate KubeVirt nmstate network config conditionally

qinqon:

The MCO templates unconditionally render nmstate configuration files that disable IPv6 autoconf and set up the `fe80::1` ARP proxy gateway route. This is correct for the default pod network, where OVN-Kubernetes assigns IPv6 via DHCPv6 stateful. However, when a KubeVirt NodePool uses multus as the primary network (`AttachDefaultNetwork=false`), these configurations break SLAAC and prevent nodes from getting IPv6 addresses in dual-stack setups.

This PR makes the HyperShift nodepool controller generate an override MachineConfig that replaces the MCO-rendered nmstate files with no-op content, restoring standard IPv6 auto-configuration (SLAAC). The override is scoped to exactly the broken population — NodePools using multus as primary network on clusters whose networking includes IPv6:

- **Default-network NodePools**: nothing is generated. The MCO templates remain the source of truth, the NodePool config hash is unchanged, and upgrading the HyperShift operator does not trigger a fleet-wide rollout.
- **IPv4-only multus NodePools**: nothing is generated either. The stale nmstate files are asymptomatic there, and since cluster networking CIDRs are immutable those clusters can never become affected.
- **Multus NodePools on IPv6-enabled clusters**: the override MachineConfig is generated; the resulting NodePool rollout is the bug fix itself.

Fixes https://issues.redhat.com/browse/OCPBUGS-77307

No MCO changes are required: instead of moving ownership of the nmstate configuration into HyperShift (which would have changed the NodePool config hash for every KubeVirt NodePool and caused a fleet-wide rollout on operator upgrade), HyperShift only neutralizes the MCO-rendered files where they are wrong.

Unit tests in `config_test.go` (`TestGetPlatformConfigs`) assert that no platform config is generated for default-network and IPv4-only multus NodePools, guaranteeing the config hash — and therefore the fleet — is untouched by an operator upgrade.

The e2e `KubeVirtAdvancedMultinetTest` assertion is IP-family aware: on IPv4-only CI lanes it verifies the override is correctly *not* generated (MCO config still applied); on IPv6-enabled clusters it verifies the override neutralizes the MCO config.

**Test coverage note**: the KubeVirt CI lane (`e2e-kubevirt-aws-ovn-reduced`) is IPv4-only, so CI exercises the gating logic and the no-rollout guarantees (unit + e2e), but not the positive path end-to-end (override applied on nodes, SLAAC working). The positive path is covered by unit tests on the generated MachineConfig content and will be verified on a dual-stack environment as part of the OCPBUGS-77307 QE verification.

## Checklist:

### DIFF
--- a/hypershift-operator/controllers/nodepool/config.go
+++ b/hypershift-operator/controllers/nodepool/config.go
@@ -12,6 +12,7 @@ import (
 
 	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
 	cpomanifests "github.com/openshift/hypershift/control-plane-operator/controllers/hostedcontrolplane/manifests"
+	kvinfra "github.com/openshift/hypershift/hypershift-operator/controllers/nodepool/kubevirt"
 	"github.com/openshift/hypershift/support/api"
 	"github.com/openshift/hypershift/support/backwardcompat"
 	"github.com/openshift/hypershift/support/capabilities"
@@ -212,7 +213,53 @@ func (cg *ConfigGenerator) generateMCORawConfig(ctx context.Context, caps *hyper
 		configs = append(configs, nodeTuningGeneratedConfigs...)
 	}
 
+	// Generate platform-specific MachineConfigs.
+	platformConfigs, err := cg.getPlatformConfigs()
+	if err != nil {
+		return "", err
+	}
+	configs = append(configs, platformConfigs...)
+
 	return cg.parse(configs)
+}
+
+// getPlatformConfigs returns platform-specific MachineConfig ConfigMaps
+// based on the NodePool's platform configuration.
+func (cg *ConfigGenerator) getPlatformConfigs() ([]corev1.ConfigMap, error) {
+	var rawConfig string
+	var err error
+
+	switch cg.nodePool.Spec.Platform.Type {
+	case hyperv1.KubevirtPlatform:
+		rawConfig, err = cg.kubevirtPlatformConfig()
+	}
+	if err != nil {
+		return nil, fmt.Errorf("failed to generate platform config: %w", err)
+	}
+
+	if rawConfig == "" {
+		return nil, nil
+	}
+
+	return []corev1.ConfigMap{
+		{
+			Data: map[string]string{
+				TokenSecretConfigKey: rawConfig,
+			},
+		},
+	}, nil
+}
+
+// kubevirtPlatformConfig generates KubeVirt-specific MachineConfig content.
+// For NodePools using multus as primary network (AttachDefaultNetwork=false) on
+// clusters with IPv6 networking, it generates an override that replaces the
+// MCO-rendered nmstate files (which assume the default pod network) with no-op
+// content, allowing standard network auto-configuration (SLAAC) to work.
+// For every other NodePool nothing is generated, keeping the NodePool config
+// hash unchanged so that upgrading the HyperShift operator does not trigger a
+// rollout.
+func (cg *ConfigGenerator) kubevirtPlatformConfig() (string, error) {
+	return kvinfra.GenerateNetworkOverrideMachineConfig(cg.nodePool, cg.hostedCluster.Spec.Networking)
 }
 
 // getUserConfigs returns a slice with all the configMaps in nodePool.Spec.Config.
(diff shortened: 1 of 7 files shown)


qinqon:
Rebased commit 932164d on main and pushed.

qinqon:
CI check `codecov/project` failed on commit 4b54100 but appears unrelated to this PR's changes.

The codecov/project check is failing because overall project coverage dropped by 2.98% (from 36.69% to 33.72%) due to 426 files with indirect coverage changes unrelated to the 7 files in the diff. The Codecov report itself confirms "All modified and coverable lines are covered by tests." The coverage decrease is caused by base-branch drift across hundreds of unrelated files, not by any missing test coverage in the changed code. This is a known pattern where codecov's project-level coverage gate triggers on base-branch coverage calculation shifts.

qinqon:
Rebased commit ad3d238 on main and pushed.

qinqon:
Rebased commit 5ff7bc0 on main and pushed.

qinqon:
## CI Failure Analysis — commit 1947eb9

**Total failures**: 1 (1 unrelated)

---

<details>
<summary>⚠️ Unrelated: <code>codecov/project</code> — flaky test</summary>

### Error
````
```
HEAD has 1 upload less than BASE
| Flag               | BASE (522d041) | HEAD (1947eb9) |
|hypershift-operator |1               |0               |

Coverage Diff
  Coverage   47.17%   41.77%   -5.40%
  Files         338      252      -86
  Lines       40608    16401   -24207

All modified and coverable lines are covered by tests.
Report is 3 commits behind head on main.
```
````

### Root Cause
The codecov/project check failed because the HEAD commit is missing the `hypershift-operator` flag upload (0 uploads vs 1 in BASE). This causes Codecov to compare 252 files against 338 files, dropping 86 files and ~24,000 lines from the calculation. The report itself confirms "All modified and coverable lines are covered by tests" -- the failure is purely a missing upload artifact, not a real coverage regression.

### Recommendation
Retest to trigger a fresh coverage upload. The missing `hypershift-operator` upload is a transient CI issue. Codecov confirms all modified lines have coverage.

</details>

**Comment by qinqon** [comment]
/cc @orenc1 @nunnatsa 

Can you review ?

File: hypershift-operator/controllers/nodepool/config.go [hunk 1]
```diff
@@ -159,9 +160,64 @@ func (cg *ConfigGenerator) generateMCORawConfig(ctx context.Context, caps *hyper
 		configs = append(configs, nodeTuningGeneratedConfigs...)
 	}
 
+	// Generate platform-specific MachineConfigs.
+	platformConfigs, err := cg.getPlatformConfigs()
```
enxebre, line 217: this would cause a fleet wide nodepool rollout as you upgrade the HO

qinqon: We are going to need to change MC to fix the bug, maybe we can make it opt-in somehow or detect the specific scenario (hosted cluster with non cluster default network and ipv6)

qinqon: Reworked to address this — the HyperShift operator no longer generates anything for default-network NodePools: the MCO templates remain the source of truth, so the raw config and hash are byte-identical to before and upgrading the HO does not trigger any rollout. `TestGetPlatformConfigs` now asserts this hash-neutrality explicitly.

The override MachineConfig is only emitted for NodePools using multus as primary network **and** whose HostedCluster networking includes IPv6 — exactly the broken population, where the rollout is the fix itself. IPv4-only multus clusters are also left untouched (asymptomatic, and since networking CIDRs are immutable they can never become affected).

qinqon: @enxebre I have limit it to add-default-network=false and ipv6 do make sense now ? 

An alternative is to just document this so customers change their nodepoos with machine config deleting those files.

qinqon:
Rebased commit de2d742 on main and pushed (conflicts resolved).

---
*🤖 Generated by [oompa](https://github.com/qinqon/oompa). Use `/oompa` to talk with me.*

qinqon:
superseed by https://github.com/openshift/hypershift/pull/9381

PR #8366, opened 2026-04-29
NO-JIRA: Extend timeout for CRD removal during integration tests

JoelSpeed:

After adjusting the output and running this in CI, I could see that the failures were because the deletiontimestamp wasn't set. I've updated the logic to make this more robust - if it sees that the CRD isn't gone, it will attempt to delete it again.

Fixes 

## Checklist:

## Summary by CodeRabbit

* **Tests**
  * Improved CRD uninstallation verification: missing resources are treated as successful removal, polling now retries deletions until resources are confirmed removed, polling is more responsive (shorter interval), and failure messages are clearer and more descriptive.

### DIFF
--- a/test/envtest/generator.go
+++ b/test/envtest/generator.go
@@ -251,15 +251,28 @@ func GenerateCRDInstallTest(featureSet string) {
 		// Uninstall after validation and wait for full removal so that
 		// subsequent per-suite tests do not hit a stale CRD that is still
 		// being deleted by the API server.
-		Expect(envtest.UninstallCRDs(cfg, envtest.CRDInstallOptions{
-			CRDs: crds,
-		})).ToNot(HaveOccurred())
+		for _, crd := range crds {
+			Expect(k8sClient.Delete(ctx, crd)).To(SatisfyAny(Succeed(), WithTransform(apierrors.IsNotFound, BeTrue())))
+		}
+
 		for _, crd := range crds {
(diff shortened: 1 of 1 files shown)


enxebre:
I've hit timeout with 60s in the past, do we want to set 120?

JoelSpeed:
Ack, extended to 120s

test/envtest/generator.go
```diff
@@ -259,7 +259,7 @@ func GenerateCRDInstallTest(featureSet string) {
 			Eventually(func() bool {
 				err := k8sClient.Get(ctx, key, &apiextensionsv1.CustomResourceDefinition{})
 				return apierrors.IsNotFound(err)
-			}, "30s", "1s").Should(BeTrue(), fmt.Sprintf("CRD %s should be fully removed", crd.Name))
+			}, "120s", "1s").Should(BeTrue(), fmt.Sprintf("CRD %s should be fully removed", crd.Name))
```
jparrill: Please update the PR description to reflect the expectations.

jparrill (commented):
Dropped some comments . Thanks! EnvTests are failing too :)

enxebre:
failed again, 180?

JoelSpeed:
Something tells me that the "delete" isn't sticking - trying to repro locally with some more output when it does fail

JoelSpeed:
No luck reproducing locally, lets see if I can dump the CRD on failure and see what it says

PR #8368, opened 2026-04-29
CNTRLPLANE-3167: Auto-detect credential type for HCPEtcdBackup

jparrill:
## Summary

- Auto-detect credential mode (STS/IRSA for AWS, Workload Identity for Azure) from the credential Secret content in the HCPEtcdBackup controller
- Extract credential resolution logic into `credentials.go` with dedicated unit tests
- Add managed services credential documentation covering OADP plugin Secret flow, IAM/federated credential prerequisites, and troubleshooting

## Description

ROSA HCP and ARO HCP use short-lived federated credentials instead of long-lived static keys. This PR enables the HCPEtcdBackup controller to auto-detect the credential type from the Secret format:

- **AWS STS/IRSA**: When the `credentials` key contains a bare IAM role ARN (`arn:...`), the controller configures a projected SA token volume with `sts.amazonaws.com` audience and sets `AWS_ROLE_ARN`/`AWS_WEB_IDENTITY_TOKEN_FILE` env vars
- **Azure Workload Identity**: When the Secret has a `cloud` key with a non-empty `AZURE_CLIENT_ID=`, the controller adds the `azure.workload.identity/use=true` pod label and annotates the SA with `azure.workload.identity/client-id`
- Static credentials (AWS INI file, Azure client-secret, Azure managed-identity) continue to work as before

No API changes — detection is purely based on Secret content inspection.

### Related PRs

- Plugin side: [openshift/hypershift-oadp-plugin#247](https://github.com/openshift/hypershift-oadp-plugin/pull/247) — OADP plugin changes to copy credentials preserving the `cloud` key for Azure WI detection

## Jira

- [CNTRLPLANE-3167](https://issues.redhat.com/browse/CNTRLPLANE-3167)

## Test plan

- [x] Unit tests for `resolveAWSCredentials`, `resolveAzureCredentials`, `resolveCredentials`, `needsCredentialsFile`
- [x] Unit tests for `createBackupJob` with STS, Workload Identity, and managed-identity modes
- [x] Unit tests for `ensureServiceAccount` annotation management
- [ ] E2E: AWS STS backup on ROSA HCP cluster
- [ ] E2E: Azure Workload Identity backup on ARO HCP cluster

### DIFF
--- a/docs/content/how-to/disaster-recovery/etcd-snapshot-backup/index.md
+++ b/docs/content/how-to/disaster-recovery/etcd-snapshot-backup/index.md
@@ -159,3 +159,7 @@ Step-by-step description of the backup process: how the OADP plugin triggers the
 ### [Restore Flow](restore-flow.md)
 
 Step-by-step description of the restore process: how the OADP plugin injects the snapshot URL, how the Control Plane Operator restores etcd, and how the cluster recovers.
+
+### [Managed Services Credentials](managed-services-credentials.md)
+
+Credential configuration for managed platforms (ROSA HCP and ARO HCP) that use federated credentials (AWS STS/IRSA and Azure Workload Identity) instead of static keys. Covers auto-detection logic, Secret format requirements, and infrastructure prerequisites.
--- a/hypershift-operator/controllers/etcdbackup/credentials.go
+++ b/hypershift-operator/controllers/etcdbackup/credentials.go
@@ -1,0 +1,107 @@
+package etcdbackup
+
+import (
+	"encoding/json"
+	"strings"
+
+	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/klog/v2"
+)
+
+type credentialMode string
+
+const (
+	credentialModeAWSStatic             credentialMode = "aws-static"
+	credentialModeAWSSTS                credentialMode = "aws-sts"
+	credentialModeAzureClientSecret     credentialMode = "azure-client-secret"
+	credentialModeAzureWorkloadIdentity credentialMode = "azure-workload-identity"
+	credentialModeAzureManagedIdentity  credentialMode = "azure-managed-identity"
+)
+
+type resolvedCredentials struct {
+	Mode       credentialMode
+	SecretName string
+	RoleARN    string
+	ClientID   string
+}
+
+func (c resolvedCredentials) needsCredentialsFile() bool {
+	switch c.Mode {
+	case credentialModeAWSStatic, credentialModeAzureClientSecret, credentialModeAzureManagedIdentity:
+		return true
+	default:
+		return false
+	}
+}
+
+func resolveCredentials(storageType hyperv1.HCPEtcdBackupStorageType, secret *corev1.Secret) resolvedCredentials {
+	switch storageType {
+	case hyperv1.S3BackupStorage:
+		return resolveAWSCredentials(secret)
+	case hyperv1.AzureBlobBackupStorage:
+		return resolveAzureCredentials(secret)
+	default:
+		klog.Warningf("resolveCredentials: unknown storage type %q for secret %s, defaulting to AWS static credentials", storageType, secret.Name)
+		return resolvedCredentials{Mode: credentialModeAWSStatic, SecretName: secret.Name}
+	}
+}
+
+func resolveAWSCredentials(secret *corev1.Secret) resolvedCredentials {
+	creds := strings.TrimSpace(string(secret.Data["credentials"]))
+	if strings.HasPrefix(creds, "arn:") {
+		return resolvedCredentials{
+			Mode:       credentialModeAWSSTS,
+			SecretName: secret.Name,
+			RoleARN:    creds,
+		}
+	}
+	return resolvedCredentials{
+		Mode:       credentialModeAWSStatic,
+		SecretName: secret.Name,
+	}
+}
+
+func resolveAzureCredentials(secret *corev1.Secret) resolvedCredentials {
+	if cloudData, ok := secret.Data["cloud"]; ok {
+		var clientID string
+		for line := range strings.SplitSeq(string(cloudData), "\n") {
+			line = strings.TrimSpace(line)
+			if v, ok := strings.CutPrefix(line, "AZURE_CLIENT_ID="); ok {
+				clientID = strings.TrimSpace(v)
+				break
+			}
+		}
+		if clientID == "" {
+			return resolvedCredentials{
+				Mode:       credentialModeAzureManagedIdentity,
+				SecretName: secret.Name,
+			}
+		}
+		return resolvedCredentials{
+			Mode:       credentialModeAzureWorkloadIdentity,
+			SecretName: secret.Name,
+			ClientID:   clientID,
+		}
+	}
+
+	if credData, ok := secret.Data["credentials"]; ok {
+		var creds struct {
+			ClientSecret string `json:"clientSecret"`
+		}
+		if err := json.Unmarshal(credData, &creds); err != nil {
+			klog.Warningf("resolveAzureCredentials: failed to parse 'credentials' key in secret %s as JSON: %v, falling through to managed-identity mode", secret.Name, err)
+		} else if creds.ClientSecret != "" {
+			return resolvedCredentials{
+				Mode:       credentialModeAzureClientSecret,
+				SecretName: secret.Name,
+			}
+		}
+	}
+
+	return resolvedCredentials{
+		Mode:       credentialModeAzureManagedIdentity,
+		SecretName: secret.Name,
+	}
+}
--- a/hypershift-operator/controllers/etcdbackup/credentials.go
+++ b/hypershift-operator/controllers/etcdbackup/credentials.go
@@ -36,6 +36,25 @@ func (c resolvedCredentials) needsCredentialsFile() bool {
 	}
 }
 
+func (c resolvedCredentials) needsProjectedToken() bool {
+	return c.Mode == credentialModeAWSSTS
+}
+
+func (c resolvedCredentials) needsWorkloadIdentityLabel() bool {
+	return c.Mode == credentialModeAzureWorkloadIdentity
+}
+
+func (c resolvedCredentials) azureAuthType() string {
+	switch c.Mode {
+	case credentialModeAzureClientSecret:
+		return "client-secret"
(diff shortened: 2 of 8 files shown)


docs/content/how-to/disaster-recovery/etcd-snapshot-backup/index.md
```diff
@@ -159,3 +159,7 @@ Step-by-step description of the backup process: how the OADP plugin triggers the
 ### [Restore Flow](restore-flow.md)
 
 Step-by-step description of the restore process: how the OADP plugin injects the snapshot URL, how the Control Plane Operator restores etcd, and how the cluster recovers.
+
+### [Managed Services Credentials](managed-services-credentials.md)
```
cblecker, line 163: This new page also needs an entry in `docs/mkdocs.yml` under the "Etcd Snapshot Backup (Tech Preview)" nav section, otherwise it won't show up in the sidebar. Something like:
```yaml
- how-to/disaster-recovery/etcd-snapshot-backup/managed-services-credentials.md
```
after the `restore-flow.md` entry.

hypershift-operator/controllers/etcdbackup/credentials.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+func resolveCredentials(storageType hyperv1.HCPEtcdBackupStorageType, secret *corev1.Secret) resolvedCredentials {
+	switch storageType {
+	case hyperv1.S3BackupStorage:
+		return resolveAWSCredentials(secret)
+	case hyperv1.AzureBlobBackupStorage:
+		return resolveAzureCredentials(secret)
+	default:
+		return resolvedCredentials{Mode: credentialModeAWSStatic, SecretName: secret.Name}
```
cblecker, line 66: nit: this default case silently returns `credentialModeAWSStatic` for any unknown storage type. The callers (`getCredentialSecretName`, `buildUploadArgs`) already error on unsupported types so it's not a bug today, but if a new storage type gets added later this could mask issues. Worth considering returning a distinct unknown mode or logging a warning here.

cblecker (commented):
Nice separation of the credential resolution into its own file — the detection logic and test coverage look solid. Two small items inline.

hypershift-operator/controllers/etcdbackup/credentials.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+}
+
+func resolveAzureCredentials(secret *corev1.Secret) resolvedCredentials {
+	if cloudData, ok := secret.Data["cloud"]; ok {
+		var clientID string
+		for line := range strings.SplitSeq(string(cloudData), "\n") {
+			line = strings.TrimSpace(line)
+			if v, ok := strings.CutPrefix(line, "AZURE_CLIENT_ID="); ok {
+				clientID = v
```
cblecker: `resolveAWSCredentials` correctly trims whitespace from the credentials value with `strings.TrimSpace`, but this path doesn't trim the extracted clientID. If the secret data has Windows-style line endings (`\r\n`) or trailing whitespace, you'd end up with `clientID = "client-789\r"` which would be set as the SA annotation and silently fail against Azure AD.

Worth adding `clientID = strings.TrimSpace(v)` here for consistency with the AWS path.

hypershift-operator/controllers/etcdbackup/credentials.go
```diff
@@ -0,0 +1,103 @@
...
+			SecretName: secret.Name,
+			ClientID:   clientID,
+		}
+	}
+
+	if credData, ok := secret.Data["credentials"]; ok {
+		var creds struct {
+			ClientSecret string `json:"clientSecret"`
+		}
+		if err := json.Unmarshal(credData, &creds); err == nil && creds.ClientSecret != "" {
```
cblecker: If the `credentials` key contains malformed JSON (truncated secret, wrong encoding, a typo), this silently falls through to managed-identity mode with no logging or feedback. A user who provides a client-secret JSON file with a typo would get managed-identity auth instead, and the root cause would be hard to track down.

Even without changing the function signature, a log line here when `err != nil` would help operators diagnose misconfigured secrets.

hypershift-operator/controllers/etcdbackup/reconciler.go
```diff
@@ -552,14 +579,27 @@ func (r *HCPEtcdBackupReconciler) setEncryptionMetadata(backup *hyperv1.HCPEtcdB
...
 			Name:      jobServiceAccountName,
 			Namespace: r.OperatorNamespace,
 		},
 	}
 	_, err := controllerutil.CreateOrUpdate(ctx, r.Client, sa, func() error {
+		if creds.Mode == credentialModeAzureWorkloadIdentity {
+			if sa.Annotations == nil {
+				sa.Annotations = map[string]string{}
+			}
+			if sa.Annotations["azure.workload.identity/client-id"] == "" {
```
cblecker, line 595: This preserves an existing annotation, which makes sense for infrastructure-managed SAs. But it also means if the credential secret is rotated to a different Azure client ID, the SA annotation won't be updated — the backup job would continue using the old client ID.

Might be worth at least logging when the existing annotation value differs from the resolved `creds.ClientID`, so operators can tell why the "wrong" identity is being used.

cblecker (commented):
A few more things beyond the inline comments:

The mode-specific checks (`creds.Mode == credentialModeAWSSTS`, `creds.Mode == credentialModeAzureWorkloadIdentity`) are scattered across reconciler.go in four places. Moving these to behavioral methods on `resolvedCredentials` (like `needsProjectedToken()`, `needsWorkloadIdentityLabel()`, `azureAuthType()`) would keep all mode knowledge co-located in credentials.go and make the reconciler a pure consumer of queries. Not blocking, but would clean up nicely.

Also, there's no test that verifies the precedence order when both `cloud` and `credentials` keys are present in the same secret. The resolver checks `cloud` first, which is the right behavior, but a test documenting that priority would prevent future regressions.

docs/content/how-to/disaster-recovery/etcd-snapshot-backup/index.md
jparrill: Already added in commit `841e5116b`. Thanks for catching it — was fixed right after your comment.

hypershift-operator/controllers/etcdbackup/credentials.go
jparrill: Good call. Added `klog.Warningf` logging the unknown storage type and secret name before falling through to the default. This way a new storage type won't silently get AWS static behavior without any trace in the logs.

hypershift-operator/controllers/etcdbackup/credentials.go
jparrill: Nice catch — Windows-style `\r\n` in the `cloud` secret data would indeed corrupt the SA annotation. Added `strings.TrimSpace(v)` to match the AWS path.

hypershift-operator/controllers/etcdbackup/credentials.go
jparrill: Agreed. Restructured the conditional to log a `klog.Warningf` with the parse error and secret name when `json.Unmarshal` fails, before falling through to managed-identity. Should make misconfigured secrets much easier to diagnose.

hypershift-operator/controllers/etcdbackup/reconciler.go
jparrill: The SA annotation is reconciled on every backup attempt via `ensureServiceAccount` (called at line 250 on each reconcile). The `== ""` guard only preserves annotations set by external infrastructure (Helm/ARO), which is intentional — those take precedence over the resolved credential. A credential rotation in practice means a new HCPEtcdBackup CR rather than in-place secret mutation, so the annotation would be correctly set on the next reconcile cycle.

tony-schndr:
@jparrill I have validated image https://quay.io/repository/jparrill/hypershift?tab=tags&tag=[CNTRLPLANE-3167](https://redhat.atlassian.net/browse/CNTRLPLANE-3167)-v4 works in ARO-HCP personal development environments by:

1. Creating a Hosted Cluster
2. Creating a test config map on Hosted Cluster
3. Taking an velero backup using `etcdBackupMethod: etcdSnapshot`
4. Deleting the Hosted Cluster
5. Restoring it from the backup from step 3.
6. Validating the test config map is still present on the Hosted Cluster

Exact image used (mirrored from quay.io): arohcpsvcdev.azurecr.io/jparrill/hypershift@sha256:220f711095682c9844e85669abf807269b16fea34be6282c78e8f474b4ddfac6

jparrill:
@verified by @tony-schndr

jparrill:
For more info:
- This was manually verified by me on AWS Self-managed
- Verified by @tony-schndr in ARO environment

hypershift-operator/controllers/etcdbackup/credentials.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+}
+
+func resolveAzureCredentials(secret *corev1.Secret) resolvedCredentials {
+	if cloudData, ok := secret.Data["cloud"]; ok {
+		var clientID string
+		for line := range strings.SplitSeq(string(cloudData), "\n") {
+			line = strings.TrimSpace(line)
+			if v, ok := strings.CutPrefix(line, "AZURE_CLIENT_ID="); ok {
+				clientID = strings.TrimSpace(v)
```
sdminonne, line 91: 'nit but only paraoid survive. 
Are we sure about the secret `cloud` content?  What if AZURE_CLIENT_ID is not present? Or even worst 
`AZURE_CLIENT_ID= ` with a whitespace?

sdminonne (approved):
@jparrill just one nit if you can have a look and eventually answer me eventually with a simple, 'wont-do-thanks' :) 
Maybe a unit test may help on it.
Otherwise good job. LGTM

hypershift-operator/controllers/etcdbackup/credentials.go
jparrill: Good eye! Both cases are already handled:

- **`AZURE_CLIENT_ID` not present**: `clientID` stays `""`, and line 95 falls back to `credentialModeAzureManagedIdentity`.
- **`AZURE_CLIENT_ID= ` (whitespace)**: `strings.TrimSpace(v)` at line 91 cleans it to `""`, same fallback path.

This was added after cblecker's earlier review catching the `\r\n` edge case — the `TrimSpace` covers both trailing CRLF and standalone whitespace. Thanks for double-checking!

jparrill:
PR Rebased. Do you mind tagging @sdminonne ?

PR #8369, opened 2026-04-29
CNTRLPLANE-3318: Support targeted node scale-down for NodePools

jparrill:
## Summary

- Add `hypershift.openshift.io/scale-down` annotation constant to the API (`NodeScaleDownAnnotation`)
- Implement HCCO Node Controller logic to sync the annotation to CAPI's `cluster.x-k8s.io/delete-machine` on the corresponding Machine, giving it top priority for deletion during scale-down
- Removing the annotation from the Node causes the controller to remove `delete-machine` from the Machine (change-of-mind support)
- Add E2E coverage in the existing autoscaling test: annotate the oldest node, verify HCCO syncs, verify autoscaler deletes the annotated node
- Add user-facing documentation under `how-to/automated-machine-management/targeted-scale-down.md`

## Jira

[CNTRLPLANE-3318](https://issues.redhat.com/browse/CNTRLPLANE-3318)

## Test plan

- [x] Unit tests for `reconcileDeleteMachineAnnotation` (10 table-driven cases + patch error injection)
- [x] Unit tests for `getMachineForNode` (4 cases)
- [x] E2E: annotate oldest node → verify HCCO sets `delete-machine` on Machine → autoscaler scale-down → verify annotated node was deleted
- [x] `go build -tags e2e ./test/e2e/...` passes
- [x] `go vet -tags e2e ./test/e2e/...` passes
- [x] Integration test script validated on real cluster (4 tests: set annotation, remove annotation, strict value matching, full happy path)

## Summary by CodeRabbit

* **New Features**
  * Added an annotation users can apply to individual nodes to mark them for prioritized deletion during NodePool scale-down; the controller propagates that priority to the corresponding management Machine so scale-down prefers annotated nodes.

* **Documentation**
  * Added a how-to guide with usage examples, constraints (value must be exactly "true"), and operational notes on targeted scale-down.

* **Tests**
  * Expanded unit and e2e tests to cover setting/removing the annotation, error paths, and validating targeted scale-down behavior.

### DIFF
--- a/api/hypershift/v1beta1/nodepool_types.go
+++ b/api/hypershift/v1beta1/nodepool_types.go
@@ -29,6 +29,20 @@ const (
 	// NodePoolReleaseVersionAnnotation is set on userdata Secrets (Replace strategy) and on Machines (InPlace strategy)
 	// to track the OCP release version associated with each machine.
 	NodePoolReleaseVersionAnnotation = "hypershift.openshift.io/release-version"
+
+	// NodeScaleDownAnnotation is applied to Nodes in the hosted cluster by users.
+	// When set to exactly "true" (case-sensitive), the HCCO Node Controller sets
+	// cluster.x-k8s.io/delete-machine=yes on the corresponding Machine, giving it
+	// top priority for deletion during NodePool scale-down.
+	//
+	// This annotation is a deletion-priority hint, not a trigger: the annotated
+	// Machine is only removed when a scale-down actually occurs (e.g., the NodePool
+	// replica count is reduced or the cluster autoscaler decides to shrink the pool).
+	//
+	// Removing this annotation from the Node causes the controller to remove the
+	// delete-machine annotation from the Machine, allowing the user to change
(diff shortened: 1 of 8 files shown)


JoelSpeed (commented):
/approve

For API change

enxebre:
/hold

The current proposal in the PR just moves that annotation to Nodes. It doesn't enable any practical enhancement feature wise. This is a known approach and has limitations, e.g. what happens if you have multiple nodes annotated.

Different consumers are going to have different expectations. We can not just add annotations for every upcoming use case.
We need to consider a solution that scales and enables multiple business cases. Current specific narrow use case is unblocked by the existing workarounds (machine annotation / make node unhealthy).

We might want to consider having something CAPI upstream more flexible and scalable like
a scale down strategy and consider external-scale-down-priority on each Machine, enabling holistic ordering based on different criteria...

jparrill:
Hey @enxebre it's there any outcome from the upstream discussion regarding this topic/PR?

PR #8375, opened 2026-04-29
OCPBUGS-84551: fix(ingress): set FIPS_ENABLED env var on ingress operator

csrwng:
## Summary

- The ingress operator determines FIPS mode by reading `/proc/sys/crypto/fips_enabled` on the node where it runs. In hosted clusters, the ingress operator runs on the management cluster, which may have a different FIPS state than the hosted cluster. This causes the operator to deploy routers with incorrect cipher configuration when the FIPS states differ.
- Set `FIPS_ENABLED=true` on the ingress operator container when the hosted cluster has FIPS enabled, so the operator uses the correct cipher suite regardless of the management cluster's FIPS state.

## Test plan

- [ ] Verify unit tests pass: `go test ./control-plane-operator/controllers/hostedcontrolplane/v2/ingressoperator/ -v`
- [ ] Create a FIPS-enabled hosted cluster and verify the ingress operator deployment has `FIPS_ENABLED=true` set
- [ ] Create a non-FIPS hosted cluster and verify the ingress operator deployment does not have `FIPS_ENABLED` set
- [ ] Verify the ingress operator deploys routers with correct FIPS cipher configuration on a FIPS-enabled hosted cluster running on a non-FIPS management cluster

## Summary by CodeRabbit

* **New Features**
  * Ingress operator now enables FIPS mode when FIPS is turned on in the hosted control plane configuration (exposes FIPS_ENABLED to the operator).

* **Tests**
  * Added unit tests validating that the ingress operator receives the FIPS_ENABLED setting when FIPS is true and does not when false.

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/v2/ingressoperator/deployment.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/v2/ingressoperator/deployment.go
@@ -22,6 +22,12 @@ func adaptDeployment(cpContext component.WorkloadContext, deployment *appsv1.Dep
 			Name: "CANARY_IMAGE", Value: cpContext.UserReleaseImageProvider.GetImage("cluster-ingress-operator"),
 		})
 
+		if cpContext.HCP.Spec.FIPS {
+			podspec.UpsertEnvVar(c, corev1.EnvVar{
+				Name: "FIPS_ENABLED", Value: "true",
+			})
+		}
+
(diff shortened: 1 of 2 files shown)


csrwng:
/verified by @csrwng 

Manually tested a 5.0ci release payload with the control plane operator built from this PR. 

- Created a HostedCluster with .spec.fips=true, inspected the ingress-operator deployment in the control plane namespace. The ingress operator deployment included the new FIPS_ENABLED=true environment variable. 
- Created a HostedCluster with .spec.fips=false, inspected the ingress-operator deployment in the control plane namespace. The ingress operator did not include the FIPS_ENABLED=true environment variable.

PR #8350, opened 2026-04-27
WIP: e2e: add CPU resource request overrides for control plane components

csrwng:
## Summary
- Add `resource-request-override` annotations to e2e HostedCluster defaults so the scheduler sees realistic CPU demand and spreads hosted clusters across management nodes
- Addresses recurring Azure e2e failures where pods time out due to CPU starvation from over-packing

### Problem
Control plane pod CPU requests are far below actual usage:

| Deployment | Request | Peak Actual | Ratio |
|---|---|---|---|
| ignition-server | 10m | 960m | 96x |
| control-plane-operator | 10m | 885m | 89x |
| packageserver | 10m | 552m | 55x |
| catalog operators (x4) | 10m each | 365m each | 37x |
| kube-apiserver | 350m | 1199m | 3.4x |

The scheduler sees ~3,015m per cluster and packs 8 clusters onto a few nodes. Actual peak is ~8,000m per cluster, causing CPU starvation that delays CRD establishment and triggers `progressDeadlineSeconds` / test timeouts.

### Fix
Override CPU requests via annotations to ~4,700m per cluster total, based on observed averages from two Azure e2e runs. Values are set between average and peak usage.

## Test plan
- [ ] Run Azure self-managed e2e with this change and verify pods start within expected timeframes
- [ ] Verify hosted clusters are spread across more management nodes
- [ ] Confirm no regression on AWS e2e

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Tests**
  * Enhanced cluster creation test configuration with additional CPU resource override annotations for control-plane and operator components.

PR #8385, opened 2026-04-30
CNTRLPLANE-3351: e2e: add opt-in CPU resource request overrides for control plane components

bryan-cox:

Cherry-picks the resource request override annotations from #8350 (by @csrwng) but makes them **opt-in** via `E2E_RESOURCE_REQUEST_OVERRIDES=1` env var.

Control plane pod CPU requests are far below actual usage (e.g. ignition-server requests 10m but peaks at 960m). The scheduler sees ~3,015m per cluster and packs too many clusters onto a few nodes. Actual peak is ~8,000m per cluster, causing CPU starvation that delays CRD establishment and triggers test timeouts.

When enabled, overrides CPU requests for control plane pods to better reflect actual consumption (~4700m per cluster vs. ~3015m default), preventing scheduler over-packing. The env var lets us enable it selectively for affected jobs (e.g. Azure e2e) first, without inadvertently breaking other jobs that are running fine.

Fixes [CNTRLPLANE-3351](https://redhat.atlassian.net/browse/CNTRLPLANE-3351)

- Follows the same env var pattern as `TEST_CPO_OVERRIDE`
- CPU values are from @csrwng's analysis in #8350, based on observed averages from Azure e2e runs
- To enable for a Prow job, set `E2E_RESOURCE_REQUEST_OVERRIDES=1` in the job env

## Checklist:

## Summary by CodeRabbit

* **Tests**
  * Centralized default test annotations and made them environment-driven.
  * Default cleanup and validation annotations are preserved.
  * New opt-in behavior: setting an environment switch adds CPU resource-override annotations for control-plane and operator components during e2e runs.

### DIFF
--- a/test/e2e/util/options.go
+++ b/test/e2e/util/options.go
@@ -226,12 +226,9 @@ func (o *Options) DefaultClusterOptions(t *testing.T) PlatformAgnosticOptions {
 			ClusterCIDR:                      []string{"10.132.0.0/14"},
 			BeforeApply:                      o.BeforeApply,
 			Log:                              NewLogr(t),
-			Annotations: []string{
-				fmt.Sprintf("%s=true", hyperv1.CleanupCloudResourcesAnnotation),
-				fmt.Sprintf("%s=true", hyperv1.SkipReleaseImageValidation),
-			},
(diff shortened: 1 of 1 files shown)


bryan-cox:
/override "ci/prow/e2e-aws"

failure is not related to the PR

PR #8394, opened 2026-05-01
fix(e2e): wait for HCP propagation before checking KAS reachability

bryan-cox:
## Summary
- Fixes KAS AllowedCIDRs test timeout in v2 e2e by waiting for HCP to reflect the updated AllowedCIDRBlocks before checking KAS reachability
- The HO informer cache may not see the HC spec change immediately, causing the initial reconciliation to read stale data and the HCP createOrUpdate to be a no-op
- Polls the HCP and touches an HC annotation on each retry to force a new HO reconciliation with an up-to-date cache

## Test plan
- [ ] CI rehearsal on openshift/release PR #77674 passes all 9 Azure v2 self-managed tests


bryan-cox:
No longer needed — switching the CI config to build from the fork directly instead.

bryan-cox:
Reopening — pointing CI at the fork doesn't work (Prow GitHub app not installed on bryan-cox org). This fix needs to land on the upstream azure-self-managed-e2e branch.

bryan-cox:
Closing to regroup on the approach.

PR #8399, opened 2026-05-02
ARO-27293: slim and harden the shared-ingress HAProxy runtime

### LINKED ISSUE #8398: shared-ingress: move HAProxy image to pinned UBI micro runtime
## Summary
The shared-ingress HAProxy image currently uses the full UBI 10 base image even though the runtime payload is only HAProxy plus its required libraries.

That larger base image pulls in many packages that are not used by the shared-ingress deployment path, which increases image size and vulnerability scanner noise for an ARO HCP-specific, Red Hat-owned artifact.

## Problem
We want to reduce the shared-ingress image footprint and trim unused packages while keeping the implementation aligned with HyperShift and ARO HCP productization constraints:
- stay on Red Hat-owned, pinned images
- keep hermetic RPM inputs
- preserve the existing HAProxy runtime contract
- verify that HAProxy still starts correctly with the controller's mounted config and runtime socket paths

## Proposed direction
- move the final runtime from full UBI to pinned UBI micro
- keep a pinned UBI builder stage for RPM installation
- remove the unused socat package from the image inputs
- add a startup smoke test that exercises the same HAProxy command, mounts, and read-only root filesystem used by the controller

## Expected benefit
A local comparison of the baseline and proposed images shows a substantial reduction in size and scanner findings:
- baseline image size: 254682256 bytes
- proposed image size: 107109829 bytes
- baseline vulnerabilities: High 28, Medium 186, Low 139
- proposed vulnerabilities: High 1, Medium 46, Low 22

## Notes
This issue is being opened to document the approach for a >200 line change before submitting the corresponding PR, per the contributing guidance.

tuxerrante:

This PR reduces the shared-ingress HAProxy image footprint and runtime attack surface without leaving the Red Hat image supply chain.

The runtime image currently starts from the full `ubi10/ubi` base even though the production container only needs HAProxy and its runtime libraries. This PR switches that flow to a pinned multi-stage build that installs HAProxy into an installroot in a pinned `ubi10/ubi` builder and copies only that payload into a pinned `ubi10/ubi-micro` final image.

It also removes the unused `socat` RPM from the image inputs, hardens the HAProxy container security context in the shared-ingress deployment, and adds a portable smoke test that exercises the same startup contract the controller uses in-cluster.

### Problem this solves

The previous image pulled in many packages that are not needed by the shared-ingress runtime path. That inflated the image, increased scanner noise, and left avoidable packages such as Python- and Vim-related content in a critical ingress-facing image.

By trimming the image down to the HAProxy runtime payload we expect:
- lower image pull and extraction latency during rollout because the runtime image is about 52.6% smaller in this environment
- lower vulnerability surface and less scanner noise because unused packages are no longer shipped in the final image
- a more restrictive pod security posture via read-only rootfs, dropped capabilities, `allowPrivilegeEscalation: false`, and `RuntimeDefault` seccomp

### Local image comparison used for this change

- Baseline image: `registry.access.redhat.com/ubi10/ubi:10.0-1753787353`
  - size: `226035753` bytes
  - grype findings: `0 critical`, `29 high`, `193 medium`, `142 low`
- Proposed image built from this branch:
  - size: `107109829` bytes
  - grype findings: `0 critical`, `0 high`, `45 medium`, `22 low`

Fixes #8398

- This is a >200 line change, so I opened #8398 first to comply with the contributing guidance for larger PRs.
- The smoke test is intended to be useful on both Linux and macOS hosts. It uses the repo's container runtime detection and verifies HAProxy startup with the same mounted config path, writable runtime socket path, and read-only root filesystem used by the controller.
- I used `NO-JIRA` in the title because this change is linked to a GitHub issue rather than a Jira ticket.

## Validation

- `go test ./hypershift-operator/controllers/sharedingress`
- `make test-shared-ingress-smoke`
- `make pre-commit`
- `go test ./hypershift-operator/controllers/nodepool -run TestResolveHAProxyImage`

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Added a runnable smoke test and a Makefile target to invoke it.

* **Security**
  * Hardened the shared ingress router containers to run with reduced privileges and stricter runtime profiles.

* **Tests**
  * Added unit tests validating the router’s hardened security settings.

* **Chores**
  * Converted the shared ingress image to a multi-stage build and removed an unnecessary runtime package.

* **Documentation**
  * Updated build and local validation instructions for the shared ingress image.

### DIFF
--- a/hypershift-operator/controllers/sharedingress/router.go
+++ b/hypershift-operator/controllers/sharedingress/router.go
@@ -154,11 +154,18 @@ func buildHCPRouterContainerMain() func(*corev1.Container) {
 				Protocol:      corev1.ProtocolTCP,
 			},
 		}
+		// The shared-ingress router only binds unprivileged ports, so it does not
+		// need NET_BIND_SERVICE even after dropping all other capabilities.
 		c.SecurityContext = &corev1.SecurityContext{
+			AllowPrivilegeEscalation: ptr.To(false),
+			ReadOnlyRootFilesystem:   ptr.To(true),
 			Capabilities: &corev1.Capabilities{
-				Add: []corev1.Capability{
-					"NET_BIND_SERVICE",
+				Drop: []corev1.Capability{
+					"ALL",
 				},
+			},
+			SeccompProfile: &corev1.SeccompProfile{
+				Type: corev1.SeccompProfileTypeRuntimeDefault,
 			},
 		}
 		c.VolumeMounts = append(c.VolumeMounts,
@@ -191,6 +198,14 @@ func buildConfigGeneratorContainer(hypershiftOperatorImage string) func(*corev1.
 			"--haproxy-socket-path", "/var/run/haproxy/admin.sock",
 		}
 		c.Image = hypershiftOperatorImage
+		c.SecurityContext = &corev1.SecurityContext{
+			AllowPrivilegeEscalation: ptr.To(false),
+			Capabilities: &corev1.Capabilities{
+				Drop: []corev1.Capability{
+					"ALL",
+				},
+			},
+		}
 
 		c.VolumeMounts = append(c.VolumeMounts,
 			corev1.VolumeMount{
--- a/hypershift-operator/controllers/sharedingress/router_test.go
+++ b/hypershift-operator/controllers/sharedingress/router_test.go
@@ -1,10 +1,12 @@
 package sharedingress
 
 import (
-	"reflect"
 	"testing"
 
+	. "github.com/onsi/gomega"
+
 	"github.com/openshift/hypershift/support/config"
+	"github.com/openshift/hypershift/support/podspec"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -15,37 +17,15 @@ import (
 )
 
 func TestReconcileRouterDeployment(t *testing.T) {
-	type args struct {
-		deployment *appsv1.Deployment
-	}
 	tests := []struct {
-		name    string
-		args    args
-		wantErr bool
+		name   string
+		assert func(*WithT, *appsv1.Deployment)
 	}{
 		{
-			name: "Valid config map and deployment",
-			args: args{
-				deployment: &appsv1.Deployment{
-					ObjectMeta: metav1.ObjectMeta{
-						Name:      "test-deployment",
-						Namespace: "test-namespace",
-					},
-				},
-			},
-			wantErr: false,
-		},
-	}
-	for _, tt := range tests {
-		t.Run(tt.name, func(t *testing.T) {
-			err := ReconcileRouterDeployment(tt.args.deployment, "test-hypershift-operator-image")
-			if (err != nil) != tt.wantErr {
-				t.Errorf("ReconcileRouterDeployment() error = %v, wantErr %v", err, tt.wantErr)
-			}
-			if !tt.wantErr {
-				if *tt.args.deployment.Spec.Replicas != 2 {
-					t.Errorf("Expected replicas to be 2, got %d", *tt.args.deployment.Spec.Replicas)
-				}
+			name: "When reconciling a valid deployment it should set the default scheduling configuration",
+			assert: func(g *WithT, deployment *appsv1.Deployment) {
+				g.Expect(deployment.Spec.Replicas).ToNot(BeNil())
+				g.Expect(*deployment.Spec.Replicas).To(Equal(int32(2)))
 
 				expectedAffinity := &corev1.Affinity{
 					NodeAffinity: &corev1.NodeAffinity{
@@ -75,9 +55,7 @@ func TestReconcileRouterDeployment(t *testing.T) {
 						},
 					},
 				}
-				if !reflect.DeepEqual(tt.args.deployment.Spec.Template.Spec.Affinity, expectedAffinity) {
-					t.Errorf("Expected affinity to be %v, got %v", expectedAffinity, tt.args.deployment.Spec.Template.Spec.Affinity)
-				}
+				g.Expect(deployment.Spec.Template.Spec.Affinity).To(Equal(expectedAffinity))
 
 				expectedTolerations := []corev1.Toleration{
 					{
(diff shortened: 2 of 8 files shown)


tuxerrante:
Root cause for the failing shared-ingress Konflux build was the mismatch between the new empty-root `dnf --installroot /rootfs` transaction and the old layered-image RPM lock input.

I pushed a follow-up commit that:
- resolves `shared-ingress/rpms.lock.yaml` as a bare-root transaction
- adds the UBI BaseOS repo to `shared-ingress/rpms.in.yaml`
- disables weak dependencies in the RPM input so the lockfile matches the `Containerfile`

Local verification on the updated branch:
- `podman build -f shared-ingress/Containerfile -t localhost/hypershift-shared-ingress:fix ./shared-ingress`
- `make test-shared-ingress-smoke`

tuxerrante:
I pushed a second follow-up after reproducing the remaining installroot failure locally.

The key detail is that the `filesystem` RPM aborts the transaction if `/rootfs/lib64` exists as a directory instead of the merged-`/usr` symlink it expects. Precreating `/rootfs/lib64 -> usr/lib64` before `dnf --installroot /rootfs` makes the same transaction succeed locally.

Fresh local verification on the updated head:
- `podman build -f shared-ingress/Containerfile -t localhost/hypershift-shared-ingress:fix2 ./shared-ingress`
- `make test-shared-ingress-smoke`

tuxerrante:
I pushed another targeted follow-up after reproducing the Konflux-style RPM path locally.

What the local hermetic repro showed:
- the generated local RPM repos work, but `dnf --installroot` fails before the transaction when repo GPG checking is enabled
- the Hermeto-style `.repo` entries point at the local mirror and keep `gpgcheck=1`, but they do not provide an importable `gpgkey`
- the same offline install succeeds once DNF is invoked with `--nogpgcheck`

Fresh local verification on the updated head:
- `podman build -f shared-ingress/Containerfile -t localhost/hypershift-shared-ingress:fix3 ./shared-ingress`
- `make test-shared-ingress-smoke`
- offline repro against a local Hermeto-style RPM mirror built from `shared-ingress/rpms.lock.yaml`

tuxerrante:
I pushed one more small follow-up for the outstanding CodeRabbit nitpicks.

Included in this update:
- removed `--rm` from `shared-ingress/smoke-test.sh` so failed smoke-test runs preserve container logs until the EXIT trap cleans up
- added a minimal `SecurityContext` to the `config-generator` sidecar (`allowPrivilegeEscalation: false`, `capabilities.drop: ["ALL"]`)
- added unit coverage for the sidecar hardening

I intentionally did **not** add `RunAsNonRoot` to the HAProxy container in this follow-up. The current shared-ingress runtime image path still does not declare a non-root `USER`, so forcing `RunAsNonRoot` here would be a behavior change that could break pod startup rather than a no-risk hardening tweak.

Fresh local verification on the updated head:
- `go test ./hypershift-operator/controllers/sharedingress`
- `make test-shared-ingress-smoke`
- `make lint-fix`
- `make pre-commit` reached the repo verify step and stopped only because the tree was intentionally dirty with these staged changes

shared-ingress/Containerfile
```diff
@@ -1,7 +1,22 @@
-FROM registry.access.redhat.com/ubi10/ubi:10.0-1753787353
+FROM registry.access.redhat.com/ubi10/ubi@sha256:ceba546c5876f319e9dfa9beb4dd5ada6f33bd0f017bbaa4cb3974d626442a08 AS builder
 
-RUN dnf install -y haproxy-3.0.5-4.el10_1.1 socat-1.7.4.4-8.el10 \
-  && dnf clean all
+RUN rm -rf /rootfs \
+  && mkdir -p /rootfs/usr \
+  && ln -s usr/lib64 /rootfs/lib64 \
+  && dnf install -y \
+    --nogpgcheck \
```
jparrill, line 9: Why was this added?

hypershift-operator/controllers/sharedingress/router.go
```diff
@@ -155,11 +155,16 @@ func buildHCPRouterContainerMain() func(*corev1.Container) {
 			},
 		}
 		c.SecurityContext = &corev1.SecurityContext{
+			AllowPrivilegeEscalation: ptr.To(false),
+			ReadOnlyRootFilesystem:   ptr.To(true),
 			Capabilities: &corev1.Capabilities{
-				Add: []corev1.Capability{
-					"NET_BIND_SERVICE",
+				Drop: []corev1.Capability{
+					"ALL",
```
jparrill, line 164: Do you mind add a comment at the code why NET_BIND_SERVICE was removed?

hypershift-operator/controllers/sharedingress/router_test.go
```diff
@@ -93,6 +93,112 @@ func TestReconcileRouterDeployment(t *testing.T) {
 			}
 		})
 	}
+
+	t.Run("When reconciling router deployment it should harden the HAProxy container security context", func(t *testing.T) {
```
jparrill: Use table driven test format.

hypershift-operator/controllers/sharedingress/router_test.go
```diff
@@ -93,6 +93,112 @@ func TestReconcileRouterDeployment(t *testing.T) {
...
+				Namespace: "test-namespace",
+			},
+		}
+
+		if err := ReconcileRouterDeployment(deployment, "test-hypershift-operator-image"); err != nil {
+			t.Fatalf("ReconcileRouterDeployment() error = %v", err)
+		}
+
+		var routerContainer *corev1.Container
+		for i := range deployment.Spec.Template.Spec.Containers {
```
jparrill: Look for helpers at the support module to do this.

hypershift-operator/controllers/sharedingress/router_test.go
```diff
@@ -93,6 +93,112 @@ func TestReconcileRouterDeployment(t *testing.T) {
...
+				Namespace: "test-namespace",
+			},
+		}
+
+		if err := ReconcileRouterDeployment(deployment, "test-hypershift-operator-image"); err != nil {
+			t.Fatalf("ReconcileRouterDeployment() error = %v", err)
+		}
+
+		var configGeneratorContainer *corev1.Container
+		for i := range deployment.Spec.Template.Spec.Containers {
```
jparrill: Look for helpers at the support module to do this.

hypershift-operator/controllers/sharedingress/router_test.go
```diff
@@ -93,6 +93,112 @@ func TestReconcileRouterDeployment(t *testing.T) {
...
+			t.Fatal("expected capabilities to be set")
+		}
+
+		expectedDrop := []corev1.Capability{"ALL"}
+		if !reflect.DeepEqual(configGeneratorContainer.SecurityContext.Capabilities.Drop, expectedDrop) {
+			t.Fatalf("expected dropped capabilities %v, got %v", expectedDrop, configGeneratorContainer.SecurityContext.Capabilities.Drop)
+		}
+
+		if len(configGeneratorContainer.SecurityContext.Capabilities.Add) != 0 {
+			t.Fatalf("expected no added capabilities, got %v", configGeneratorContainer.SecurityContext.Capabilities.Add)
```
jparrill: Replace `t.Fatal*` and `reflect.DeepEqual` by omega convention. (Not just this entry + other ones in the PR)

jparrill (changes requested):

shared-ingress/Containerfile
tuxerrante: Added an inline note in `shared-ingress/Containerfile`. This flag is only needed for the Konflux/Hermeto hermetic path: the generated local repos are checksum-verified, but their `.repo` entries do not carry an importable `gpgkey`, so `dnf --installroot` aborts without `--nogpgcheck`.

hypershift-operator/controllers/sharedingress/router.go
tuxerrante: Fixed in 36c4903. I added an inline comment in `router.go`: the shared-ingress router now binds only 6443 and 8443, so it no longer needs `NET_BIND_SERVICE` after dropping the rest of the capabilities.

hypershift-operator/controllers/sharedingress/router_test.go
tuxerrante: Fixed in 36c4903. I switched the container lookup in this test to `support/podspec.FindContainer` instead of open-coding the loop.

hypershift-operator/controllers/sharedingress/router_test.go
tuxerrante: Fixed in 36c4903. I converted the shared-ingress tests touched here to `gomega` expectations and removed the `reflect.DeepEqual` / `t.Fatal*` assertions from this file.

hypershift-operator/controllers/sharedingress/router_test.go
tuxerrante: Fixed in 36c4903. `TestReconcileRouterDeployment` now uses table-driven subtests for both the scheduling assertions and the two security-context checks.

hypershift-operator/controllers/sharedingress/router_test.go
tuxerrante: Fixed in 36c4903. This lookup also uses `support/podspec.FindContainer` now.

test/integration/shared-ingress/smoke-test.sh
```diff
@@ -0,0 +1,110 @@
+#!/usr/bin/env bash
```
jparrill, line 1: The smoke test should live under `test/integration/` to follow the existing test directory structure. Shell-based integration tests that validate container contracts belong there, not in the component source directory.

jparrill:
This PR introduces non-trivial changes (image build restructuring, security context hardening, new smoke test, RPM dependency changes). It should be tracked with a Jira ticket rather than `NO-JIRA`. Please create a ticket and update the PR title accordingly.

test/integration/shared-ingress/smoke-test.sh
tuxerrante: Thanks for the feedback @jparrill! Done — the smoke test now lives under `test/integration/shared-ingress/smoke-test.sh` and the Makefile target + README references are updated accordingly.

tuxerrante:
Thanks for the thorough review @jparrill, all your feedback has been addressed and the branch is restructured into two clean commits.

One open question from the CodeRabbit review: it flagged the missing `RunAsNonRoot: true` in the private-router SecurityContext, which is required for full Kubernetes Restricted PSS compliance. However, the `ubi10/ubi-micro` base image doesn't set a `USER` directive, so HAProxy would default to UID 0 — adding `RunAsNonRoot: true` without also setting `RunAsUser` or adding a `USER` directive in the Containerfile could break pod startup.

Do you see any side effects or concerns with introducing `RunAsNonRoot: true` here? Specifically:
- Should we add a non-root `USER` in the Containerfile (the `haproxy` RPM creates a `haproxy` system user)?
- Or is it safer to defer this to a follow-up after validating HAProxy runs correctly as non-root with admin socket writes and config reads?

Happy to tackle it either way — just want your take on the risk.

PR #8404, opened 2026-05-04
NO-JIRA: fix(karpenter): resolve HCP karpenter finalizer when AutoNode is disabled

enxebre:
## Summary

When AutoNode is disabled before HostedCluster deletion, the karpenter-operator deployment is removed and cannot process the HCP karpenter finalizer. The `resolveKarpenterFinalizer` fallback previously early-returned when `IsKarpenterEnabled` was false, leaving the HCP stuck in terminating indefinitely.

### Root cause

The `resolveKarpenterFinalizer` function had a guard:
```go
if !karpenterutil.IsKarpenterEnabled(hc.Spec.AutoNode) {
    return nil
}
```
This skipped the fallback entirely when AutoNode was disabled, even though the HCP still had the finalizer from when Karpenter was previously enabled.

### Fix

Remove the early-return guard and instead use `IsKarpenterEnabled` to decide the cleanup strategy:
- **AutoNode enabled**: defer to the karpenter-operator for graceful cleanup — only force-remove the finalizer once the guest KAS is down
- **AutoNode disabled**: remove the finalizer immediately since no controller exists to handle it

### Test plan

- [x] Existing `TestResolveKarpenterFinalizer` cases pass (renamed to Gherkin)
- [x] New test: "When karpenter is disabled and HCP has a stale finalizer it should remove the finalizer immediately"
- [x] New test: "When karpenter is disabled and KAS is still available it should remove the finalizer immediately"

/cc @maxcao13 @jkyros

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved an issue where disabling node auto-scaling could cause HostedCluster deletion to hang indefinitely. The system now properly removes cleanup markers when auto-scaling is disabled, and only defers removal when enabled and the cluster is accessible.

### DIFF
--- a/hypershift-operator/controllers/hostedcluster/karpenter.go
+++ b/hypershift-operator/controllers/hostedcluster/karpenter.go
@@ -118,18 +118,16 @@ spec:
 }
 
 // resolveKarpenterFinalizer removes the karpenter finalizer from the HCP
-// when the karpenter-operator cannot do so itself because the guest KAS is down.
+// when the karpenter-operator cannot do so itself. This covers two scenarios:
 //
-// This breaks the deadlock where:
-//  1. The HostedCluster deletion waits for the HCP to be removed
-//  2. The HCP can't be removed because it has the karpenter finalizer
-//  3. The karpenter-operator can't remove the finalizer because the guest KAS is down
-//     and it depends on guest-side watches to function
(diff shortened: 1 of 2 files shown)


joshbranham:
/retitle NO-JIRA: fix(karpenter): resolve HCP karpenter finalizer when AutoNode is disabled

PR #8406, opened 2026-05-04
OCPBUGS-84577: clear stale EtcdRecoveryActive failure condition when etcd is healthy

vsolanki12:

When the etcd recovery job fails but etcd self-heals, the `EtcdRecoveryJobFailed` condition was never cleared. This caused the OpenShift Console to display a stale error message ("Error in Etcd Recovery job: the Etcd cluster requires manual intervention.") on the HostedCluster overview page, even when the cluster was fully healthy (`Available=True`, `Degraded=False`, `EtcdAvailable=True`).

This fix adds two checks in `reconcileETCDMemberRecovery`:
- When a failed recovery job exists but the etcd StatefulSet is fully available (3/3 replicas), clean up the job and clear the condition.
- When no failing etcd pods exist and etcd is healthy, clear any stale `EtcdRecoveryJobFailed` condition.

Fixes https://issues.redhat.com/browse/OCPBUGS-84577

- The etcd recovery feature is gated behind `ENABLE_ETCD_RECOVERY` env var and only applies to managed, highly-available etcd clusters.
- Both fix paths were verified on a live KubeVirt HCP cluster by simulating the stale condition and confirming it gets cleared.
- The `etcd_manual_intervention_required` metric (which reads this condition) will also correctly reset.

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Automatically cleans up etcd recovery resources and clears the recovery-active failure state when the etcd cluster reports full readiness, avoiding unnecessary manual-intervention alerts while still signaling failures when the cluster is unhealthy.
  * Clears stale recovery failure conditions when there is no failing etcd pod and the cluster is fully available.

* **Tests**
  * Added unit tests covering recovery success/failure paths, condition clearing, and cleanup behavior.

### DIFF
--- a/hypershift-operator/controllers/hostedcluster/etcd_recovery.go
+++ b/hypershift-operator/controllers/hostedcluster/etcd_recovery.go
@@ -29,6 +29,12 @@ import (
 	"github.com/go-logr/logr"
 )
 
+func isEtcdStatefulSetHealthy(sts *appsv1.StatefulSet) bool {
+	return sts.Spec.Replicas != nil &&
+		sts.Status.ReadyReplicas == *sts.Spec.Replicas &&
+		sts.Status.AvailableReplicas == *sts.Spec.Replicas
+}
+
 type etcdJobStatus struct {
 	exists     bool
 	finished   bool
@@ -62,12 +68,31 @@ func (r *HostedClusterReconciler) reconcileETCDMemberRecovery(ctx context.Contex
 // It returns (done, err) where done=true means the caller should return immediately
 // without falling through to detectAndTriggerEtcdRecovery.
 func (r *HostedClusterReconciler) handleExistingEtcdRecoveryJob(ctx context.Context, log logr.Logger, hcluster *hyperv1.HostedCluster, recoveryJob *batchv1.Job, jobStatus *etcdJobStatus) (bool, error) {
+	hcpNS := manifests.HostedControlPlaneNamespace(hcluster.Namespace, hcluster.Name)
+
 	if !jobStatus.finished {
 		log.Info("waiting for etcd recovery job to complete")
 		return true, nil
 	}
 
 	if !jobStatus.successful {
+		etcdStatefulSet := etcdrecoverymanifests.EtcdStatefulSet(hcpNS)
+		if err := r.Get(ctx, crclient.ObjectKeyFromObject(etcdStatefulSet), etcdStatefulSet); err != nil {
+			if !apierrors.IsNotFound(err) {
+				return false, fmt.Errorf("failed to get etcd statefulset: %w", err)
+			}
+		} else if isEtcdStatefulSetHealthy(etcdStatefulSet) {
+			log.Info("etcd recovered despite failed recovery job, cleaning up")
+			if err := r.cleanupEtcdRecoveryObjects(ctx, hcluster); err != nil {
+				return false, fmt.Errorf("failed to cleanup etcd recovery job: %w", err)
+			}
+			if err := r.setEtcdRecoveryCondition(ctx, hcluster, metav1.ConditionFalse, hyperv1.AsExpectedReason, "Etcd recovered despite failed recovery job."); err != nil {
+				return false, err
+			}
+			return true, nil
+		}
+
+		// Etcd is still unhealthy after the failed recovery job; report manual intervention needed.
 		if err := r.setEtcdRecoveryCondition(ctx, hcluster, metav1.ConditionFalse, hyperv1.EtcdRecoveryJobFailedReason, "Error in Etcd Recovery job: the Etcd cluster requires manual intervention."); err != nil {
 			return false, err
 		}
@@ -80,7 +105,7 @@ func (r *HostedClusterReconciler) handleExistingEtcdRecoveryJob(ctx context.Cont
 		return false, fmt.Errorf("failed to cleanup etcd recovery job: %w", err)
 	}
 
-	if err := r.setEtcdRecoveryCondition(ctx, hcluster, metav1.ConditionFalse, hyperv1.AsExpectedReason, "ETCD Recovery job succeeded."); err != nil {
+	if err := r.setEtcdRecoveryCondition(ctx, hcluster, metav1.ConditionFalse, hyperv1.AsExpectedReason, "Etcd recovery job succeeded."); err != nil {
 		return false, err
 	}
 
@@ -100,8 +125,11 @@ func (r *HostedClusterReconciler) setEtcdRecoveryCondition(ctx context.Context, 
 		LastTransitionTime: r.now(),
 	}
 
+	// Update the condition if the status or reason changed. The reason check
+	// is needed to transition from EtcdRecoveryJobFailed -> AsExpected when
+	// etcd self-heals (both use Status=False).
 	oldCondition := meta.FindStatusCondition(hcluster.Status.Conditions, string(hyperv1.EtcdRecoveryActive))
-	if oldCondition == nil || oldCondition.Status != condition.Status {
+	if oldCondition == nil || oldCondition.Status != condition.Status || oldCondition.Reason != condition.Reason {
 		meta.SetStatusCondition(&hcluster.Status.Conditions, condition)
 		if err := r.Client.Status().Update(ctx, hcluster); err != nil {
 			return fmt.Errorf("failed to update etcd recovery job condition: %w", err)
@@ -117,8 +145,9 @@ func (r *HostedClusterReconciler) detectAndTriggerEtcdRecovery(ctx context.Conte
 			log.Info("etcd statefulset does not yet exist")
 			return nil, nil
 		}
+		return nil, fmt.Errorf("failed to get etcd statefulset: %w", err)
 	}
-	fullyAvailable := etcdStatefulSet.Status.ReadyReplicas == 3 && etcdStatefulSet.Status.AvailableReplicas == 3
+	fullyAvailable := isEtcdStatefulSetHealthy(etcdStatefulSet)
 	if !fullyAvailable {
 		log.Info("etcd is not reporting fully available, need to watch")
 	}
@@ -134,6 +163,18 @@ func (r *HostedClusterReconciler) detectAndTriggerEtcdRecovery(ctx context.Conte
 		if !fullyAvailable {
 			return &requeueAfter, nil
 		}
+
+		// Only clear conditions with EtcdRecoveryJobFailedReason. Conditions set
+		// during active recovery (Status=True/Reason=AsExpected) are managed by
+		// handleExistingEtcdRecoveryJob when the job completes.
+		oldCondition := meta.FindStatusCondition(hcluster.Status.Conditions, string(hyperv1.EtcdRecoveryActive))
+		if oldCondition != nil && oldCondition.Reason == hyperv1.EtcdRecoveryJobFailedReason {
+			log.Info("etcd is healthy but EtcdRecoveryActive has stale failure condition, clearing it")
+			if err := r.setEtcdRecoveryCondition(ctx, hcluster, metav1.ConditionFalse, hyperv1.AsExpectedReason, "Etcd cluster is healthy."); err != nil {
+				return nil, err
+			}
+		}
+
 		return nil, nil
 	}
 
@@ -143,7 +184,7 @@ func (r *HostedClusterReconciler) detectAndTriggerEtcdRecovery(ctx context.Conte
 		return nil, err
 	}
 
-	if err := r.setEtcdRecoveryCondition(ctx, hcluster, metav1.ConditionTrue, hyperv1.AsExpectedReason, "ETCD Recovery job in progress."); err != nil {
+	if err := r.setEtcdRecoveryCondition(ctx, hcluster, metav1.ConditionTrue, hyperv1.AsExpectedReason, "Etcd recovery job in progress."); err != nil {
 		return nil, err
 	}
 
--- a/hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
+++ b/hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
@@ -30,6 +30,7 @@ import (
 	hcmetrics "github.com/openshift/hypershift/hypershift-operator/controllers/hostedcluster/metrics"
 	hcpmanifests "github.com/openshift/hypershift/hypershift-operator/controllers/manifests"
 	"github.com/openshift/hypershift/hypershift-operator/controllers/manifests/controlplaneoperator"
+	etcdrecoverymanifests "github.com/openshift/hypershift/hypershift-operator/controllers/manifests/etcdrecovery"
 	kvinfra "github.com/openshift/hypershift/kubevirtexternalinfra"
 	"github.com/openshift/hypershift/support/api"
 	"github.com/openshift/hypershift/support/azureutil"
@@ -50,9 +51,11 @@ import (
 	configv1 "github.com/openshift/api/config/v1"
 
 	appsv1 "k8s.io/api/apps/v1"
+	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	errors2 "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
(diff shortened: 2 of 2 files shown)


vsolanki12:
@coderabbitai full review

vsolanki12:
@coderabbitai full review

vsolanki12:
@coderabbitai full review

hypershift-operator/controllers/hostedcluster/etcd_recovery.go
```diff
@@ -56,6 +56,26 @@ func (r *HostedClusterReconciler) reconcileETCDMemberRecovery(ctx context.Contex
...
+				}
+			} else if etcdStatefulSet.Status.ReadyReplicas == 3 && etcdStatefulSet.Status.AvailableReplicas == 3 {
+				log.Info("etcd recovered despite failed recovery job, cleaning up")
+				if err := r.cleanupEtcdRecoveryObjects(ctx, hcluster); err != nil {
+					return nil, fmt.Errorf("failed to cleanup etcd recovery job: %w", err)
+				}
+				etcdRecoveryActiveCondition.Status = metav1.ConditionFalse
+				etcdRecoveryActiveCondition.Reason = hyperv1.AsExpectedReason
+				etcdRecoveryActiveCondition.LastTransitionTime = r.now()
+				meta.SetStatusCondition(&hcluster.Status.Conditions, etcdRecoveryActiveCondition)
```
sdminonne: What about `etcdRecoveryActiveCondition.Message`  in this path?

hypershift-operator/controllers/hostedcluster/etcd_recovery.go
```diff
@@ -137,6 +157,22 @@ func (r *HostedClusterReconciler) reconcileETCDMemberRecovery(ctx context.Contex
...
+
+		// Clear stale EtcdRecoveryJobFailed condition if etcd is healthy
+		oldCondition := meta.FindStatusCondition(hcluster.Status.Conditions, string(hyperv1.EtcdRecoveryActive))
+		if oldCondition != nil && oldCondition.Reason == hyperv1.EtcdRecoveryJobFailedReason {
+			log.Info("etcd is healthy but EtcdRecoveryActive has stale failure condition, clearing it")
+			meta.SetStatusCondition(&hcluster.Status.Conditions, metav1.Condition{
+				Type:               string(hyperv1.EtcdRecoveryActive),
+				Status:             metav1.ConditionFalse,
+				Reason:             hyperv1.AsExpectedReason,
+				ObservedGeneration: hcluster.Generation,
```
sdminonne: What about `etcdRecoveryActiveCondition.Message` in this path?

hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
```diff
@@ -6569,3 +6572,165 @@ func TestComputeEndpointServiceCondition(t *testing.T) {
...
 }
+
+func TestReconcileETCDMemberRecovery(t *testing.T) {
+	hcpNS := "clusters-test-hc"
+
+	healthyEtcdPods := func() []crclient.Object {
+		var pods []crclient.Object
+		for i := 0; i < 3; i++ {
+			pods = append(pods, &corev1.Pod{
+				ObjectMeta: metav1.ObjectMeta{
```
sdminonne, line 7155: May you add a case where `RestartCount` is > 0 and `Running` state? 
To cover case when POD recovers.

sdminonne (changes requested):
May you follow up?

hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
vsolanki12: Thanks, I have made all the changes suggested on the etcd_recovery.go & hostedcluster_controller_test.go file.

hypershift-operator/controllers/hostedcluster/etcd_recovery.go
vsolanki12: Thanks, I have made all the changes suggested on the etcd_recovery.go & hostedcluster_controller_test.go file.

hypershift-operator/controllers/hostedcluster/etcd_recovery.go
vsolanki12: Thanks, I have made all the changes suggested on the etcd_recovery.go & hostedcluster_controller_test.go file.

vsolanki12:
supported logs:
Before fix: 
`{
  "type": "EtcdRecoveryActive",
  "status": "False",
  "reason": "EtcdRecoveryJobFailed",
  "message": "Error in Etcd Recovery
  job: the Etcd cluster requires
  manual intervention.",
  "lastTransitionTime": "..."
}`

After fix:
`{
  "type": "EtcdRecoveryActive",
  "status": "False",
  "reason": "AsExpected",
  "lastTransitionTime": "..."
}`

hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
```diff
@@ -6734,3 +6737,195 @@ func TestComputeEndpointServiceCondition(t *testing.T) {
...
+		},
+	}
+
+	unhealthyStatefulSet := &appsv1.StatefulSet{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "etcd",
+			Namespace: hcpNS,
+		},
+		Status: appsv1.StatefulSetStatus{
+			ReadyReplicas:     2,
```
sdminonne: Is this unhealthy 'cause we hardcode etcd spec.replicas = 3? Mind adding the spec then?

hypershift-operator/controllers/hostedcluster/etcd_recovery.go
```diff
@@ -62,12 +62,30 @@ func (r *HostedClusterReconciler) reconcileETCDMemberRecovery(ctx context.Contex
...
 		return true, nil
 	}
 
 	if !jobStatus.successful {
+		etcdStatefulSet := etcdrecoverymanifests.EtcdStatefulSet(hcpNS)
+		if err := r.Get(ctx, crclient.ObjectKeyFromObject(etcdStatefulSet), etcdStatefulSet); err != nil {
+			if !apierrors.IsNotFound(err) {
+				return false, fmt.Errorf("failed to get etcd statefulset: %w", err)
+			}
+		} else if etcdStatefulSet.Status.ReadyReplicas == 3 && etcdStatefulSet.Status.AvailableReplicas == 3 {
```
sdminonne: this is hardcoding `etcdStatefulSet.Status.ReadyReplicas = 3` right?

sdminonne (changes requested):
Generally OK (aside that we're hard-coding helthy etcd repliacs = 3) Need to improve coverage (for example in case of errors).
Can we create instead a isETCDHealthy func?

hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
vsolanki12: Yes, made the changed as per suggestions.

hypershift-operator/controllers/hostedcluster/etcd_recovery.go
vsolanki12: Yes, made the changed as per suggestions.

hypershift-operator/controllers/hostedcluster/etcd_recovery.go
sdminonne: Mind using here the brand new `isEtcdStatefulSetHealthy`

sdminonne, line 143: Here we've a bug, I know you didn't modify this part but since we're here may you fix it?
In case err is different that `NotFound` we pass through and keep running (for example API transient error).

sdminonne (commented):
Thanks nice work! One follow-up only and one pre-existing bug :( 
Mind having a look?

hypershift-operator/controllers/hostedcluster/etcd_recovery.go
vsolanki12: Thanks I have updated and push in latest commit.

vsolanki12: Thanks I have updated and push in latest commit.

hypershift-operator/controllers/hostedcluster/etcd_recovery.go
```diff
@@ -62,12 +68,30 @@ func (r *HostedClusterReconciler) reconcileETCDMemberRecovery(ctx context.Contex
...
+		} else if isEtcdStatefulSetHealthy(etcdStatefulSet) {
+			log.Info("etcd recovered despite failed recovery job, cleaning up")
+			if err := r.cleanupEtcdRecoveryObjects(ctx, hcluster); err != nil {
+				return false, fmt.Errorf("failed to cleanup etcd recovery job: %w", err)
+			}
+			if err := r.setEtcdRecoveryCondition(ctx, hcluster, metav1.ConditionFalse, hyperv1.AsExpectedReason, "ETCD recovered despite failed recovery job."); err != nil {
+				return false, err
+			}
+			return true, nil
+		}
```
bryan-cox, line 93: **[blocking]** The fallthrough to this existing `setEtcdRecoveryCondition` call is the "etcd is still unhealthy" case, but it is easy to miss after the new health-check block above. A short comment would help:

```go
// Etcd is still unhealthy after the failed recovery job; report manual intervention needed.
```

hypershift-operator/controllers/hostedcluster/etcd_recovery.go
```diff
@@ -99,7 +123,7 @@ func (r *HostedClusterReconciler) setEtcdRecoveryCondition(ctx context.Context,
 	}
 
 	oldCondition := meta.FindStatusCondition(hcluster.Status.Conditions, string(hyperv1.EtcdRecoveryActive))
-	if oldCondition == nil || oldCondition.Status != condition.Status {
+	if oldCondition == nil || oldCondition.Status != condition.Status || oldCondition.Reason != condition.Reason {
```
bryan-cox, line 132: **[blocking]** This guard change is necessary for the fix (both old and new condition use `Status=False`), but the intent is non-obvious. Please add a comment explaining why `Reason` is now compared:

```go
// Update the condition if the status or reason changed. The reason check
// is needed to transition from EtcdRecoveryJobFailed -> AsExpected when
// etcd self-heals (both use Status=False).
```

hypershift-operator/controllers/hostedcluster/etcd_recovery.go
```diff
@@ -131,6 +156,15 @@ func (r *HostedClusterReconciler) detectAndTriggerEtcdRecovery(ctx context.Conte
 		if !fullyAvailable {
 			return &requeueAfter, nil
 		}
+
+		oldCondition := meta.FindStatusCondition(hcluster.Status.Conditions, string(hyperv1.EtcdRecoveryActive))
+		if oldCondition != nil && oldCondition.Reason == hyperv1.EtcdRecoveryJobFailedReason {
+			log.Info("etcd is healthy but EtcdRecoveryActive has stale failure condition, clearing it")
+			if err := r.setEtcdRecoveryCondition(ctx, hcluster, metav1.ConditionFalse, hyperv1.AsExpectedReason, "ETCD cluster is healthy."); err != nil {
+				return nil, err
+			}
```
bryan-cox, line 175: **[question]** This only clears stale conditions with `Reason == EtcdRecoveryJobFailedReason`. If someone manually deletes the failed recovery Job while recovery is active, the condition could be stuck at `Status=True/Reason=AsExpected` — that would not be caught here.

Is manual job deletion during active recovery a scenario worth guarding against? If not, a brief comment explaining the scope would be helpful.

bryan-cox (commented):
## Staff-Level Review

The fix correctly addresses the reported problem — stale `EtcdRecoveryJobFailed` condition persisting after etcd self-heals. The approach is architecturally sound, `isEtcdStatefulSetHealthy` is a good extraction, and the metric integration is correct. See inline comments for items to address before merge.

hypershift-operator/controllers/hostedcluster/etcd_recovery.go
```diff
@@ -62,12 +68,30 @@ func (r *HostedClusterReconciler) reconcileETCDMemberRecovery(ctx context.Contex
...
 	}
 
 	if !jobStatus.successful {
+		etcdStatefulSet := etcdrecoverymanifests.EtcdStatefulSet(hcpNS)
+		if err := r.Get(ctx, crclient.ObjectKeyFromObject(etcdStatefulSet), etcdStatefulSet); err != nil {
+			if !apierrors.IsNotFound(err) {
+				return false, fmt.Errorf("failed to get etcd statefulset: %w", err)
+			}
+		} else if isEtcdStatefulSetHealthy(etcdStatefulSet) {
+			log.Info("etcd recovered despite failed recovery job, cleaning up")
```
bryan-cox, line 85: **[nit]** New messages use "ETCD" (all-caps) while existing code at line 93 uses "Etcd" (title-case). Not blocking, but consider unifying the capitalization.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
```diff
@@ -6734,3 +6738,238 @@ func TestComputeEndpointServiceCondition(t *testing.T) {
...
+
+		_, err := r.reconcileETCDMemberRecovery(
+			ctrl.LoggerInto(t.Context(), zap.New(zap.UseDevMode(true))),
+			hcluster,
+			upsert.New(false).CreateOrUpdate,
+		)
+		g.Expect(err).To(HaveOccurred())
+		g.Expect(err.Error()).To(ContainSubstring("failed to get etcd statefulset"))
+	})
+}
```
bryan-cox, line 7442: **[blocking]** Missing test: the transient error test only exercises `detectAndTriggerEtcdRecovery` (no job seeded). Please add a test where a failed job exists AND the StatefulSet Get fails, to verify the new Get in `handleExistingEtcdRecoveryJob` (line 76) properly returns the error rather than falling through.

hypershift-operator/controllers/hostedcluster/etcd_recovery.go
vsolanki12: Added scope comment. Not a scenario worth guarding against active recovery conditions are managed by `handleExistingEtcdRecovery`Job.

hypershift-operator/controllers/hostedcluster/etcd_recovery.go
vsolanki12: Thank you @bryan-cox, 
Done, added the comment.

hypershift-operator/controllers/hostedcluster/etcd_recovery.go
vsolanki12: Thank you @bryan-cox, 
Done, added the comment.

hypershift-operator/controllers/hostedcluster/etcd_recovery.go
vsolanki12: Done, unified to "Etcd" throughout.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
vsolanki12: Done, added when failed job exists and StatefulSet Get fails with transient error it should return the error.

bryan-cox (approved):
/approve

bryan-cox:
/lgtm cancel

only meant to put approve

hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
```diff
@@ -7141,3 +7144,287 @@ func TestKasServingCertHashFromEndpoint(t *testing.T) {
...
+		objects          []crclient.Object
+		conditions       []metav1.Condition
+		expectedReason   string
+		conditionExists  bool
+		expectJobDeleted bool
+	}{
+		{
+			name:            "When etcd is healthy and stale EtcdRecoveryJobFailed condition exists it should clear the condition",
+			conditions:      []metav1.Condition{staleCondition},
+			objects:         append(healthyEtcdPods(), healthyStatefulSet),
```
sdminonne: `healthyStatefulSet` need to be initialized at every test instance (as `healthyEtcdPods` is doing)

hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
```diff
@@ -7141,3 +7144,287 @@ func TestKasServingCertHashFromEndpoint(t *testing.T) {
...
+			expectedReason:  hyperv1.AsExpectedReason,
+			conditionExists: true,
+		},
+		{
+			name:            "When etcd is healthy and no EtcdRecoveryActive condition exists it should not add one",
+			conditions:      []metav1.Condition{},
+			objects:         append(healthyEtcdPods(), healthyStatefulSet),
```
sdminonne: healthyStatefulSet need to be initialized at every test instance (as healthyEtcdPods is doing)

hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
```diff
@@ -7141,3 +7144,287 @@ func TestKasServingCertHashFromEndpoint(t *testing.T) {
...
+			conditions:       []metav1.Condition{staleCondition},
+			objects:          append(healthyEtcdPods(), healthyStatefulSet, failedJob),
+			expectedReason:   hyperv1.AsExpectedReason,
+			conditionExists:  true,
+			expectJobDeleted: true,
+		},
+		{
+			name:            "When etcd pods have restarted but recovered it should clear the stale condition",
+			conditions:      []metav1.Condition{staleCondition},
+			objects:         append(recoveredEtcdPods(), healthyStatefulSet),
```
sdminonne: healthyStatefulSet need to be initialized at every test instance (as healthyEtcdPods is doing)

hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
```diff
@@ -7141,3 +7144,287 @@ func TestKasServingCertHashFromEndpoint(t *testing.T) {
...
+			expectedReason:  hyperv1.AsExpectedReason,
+			conditionExists: true,
+		},
+		{
+			name:            "When failed job exists and etcd is still unhealthy it should keep the failure condition",
+			conditions:      []metav1.Condition{staleCondition},
+			objects:         append(healthyEtcdPods(), unhealthyStatefulSet, failedJob),
```
sdminonne: Same for `unhealthyStatefulSet` and `failedJob`... 
**TL/DR**: don't reuse mutable objects across unit tests

hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
```diff
@@ -7141,3 +7144,287 @@ func TestKasServingCertHashFromEndpoint(t *testing.T) {
...
+			expectJobDeleted: true,
+		},
+		{
+			name:            "When etcd pods have restarted but recovered it should clear the stale condition",
+			conditions:      []metav1.Condition{staleCondition},
+			objects:         append(recoveredEtcdPods(), healthyStatefulSet),
+			expectedReason:  hyperv1.AsExpectedReason,
+			conditionExists: true,
+		},
+		{
```
sdminonne, line 7274: If I'm not wrong we may need a test

"When failed job exists and etcd statefulset does not exist"

sdminonne (changes requested):
@vsolanki12 thanks for this. I addes some comments mainly in testing.
Mind follow-up?

hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
vsolanki12: Done, added this test case. It verifies the `NotFound` path in `handleExistingEtcdRecoveryJob` when the `StatefulSet` doesn't exist, it falls through to set `EtcdRecoveryJobFailedReason`.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
vsolanki12: Done, converted all shared mutable objects (healthyStatefulSet, unhealthyStatefulSet, staleCondition, failedJob) to closures returning fresh instances per test invocation.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
vsolanki12: Done, all mutable objects are now closures.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
vsolanki12: Done, all mutable objects are now closures.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
vsolanki12: Done, converted `healthyStatefulSet`, `unhealthyStatefulSet`, `staleCondition`, and failedJob to closures that return fresh instances per test, matching the `healthyEtcdPods()` pattern.

vsolanki12:
/verified by @vsolanki12 

**Before fix:**
```
{
  "type": "EtcdRecoveryActive",
  "status": "False",
  "reason": "EtcdRecoveryJobFailed",
  "message": "Error in Etcd Recovery
  job: the Etcd cluster requires
  manual intervention.",
  "lastTransitionTime": "..."
}
```

**After fix:**
```
{
  "type": "EtcdRecoveryActive",
  "status": "False",
  "reason": "AsExpected",
  "lastTransitionTime": "..."
}
```

PR #8411, opened 2026-05-04
GCP-412: add envtest cases for GCP imageRegistry CEL validation

cblecker:
## Summary

- Adds envtest YAML test cases for `imageRegistry` service account CEL validation in the GCP `HostedCluster` spec
- Covers three scenarios: cross-project validation (onCreate), immutability (onUpdate), and CEL ratcheting (onUpdate, both failure and pass cases)
- Mirrors the existing `networkSA` test patterns in the same file, using CEL rule index 4 (imageRegistry) instead of 5 (network)

## Test plan

- [x] `make test-envtest-ocp` — full envtest run across OCP K8s versions passes with new test cases
- [x] `make test` — no regressions in existing unit tests


cblecker:
@coderabbit resume

PR #8413, opened 2026-05-04
GCP-636: feat(gcp): support for managing GCP OIDC discovery documents

ckandag:

Add support for managing GCP OIDC discovery documents in GCS buckets
as part of the HostedCluster lifecycle. This enables Workload Identity
Federation (WIF) for GCP-hosted control planes by uploading OIDC
discovery and JWKS documents to a configurable GCS bucket.

Key changes:
- Add GCSAPI interface and GCS client using google.golang.org/api/storage/v1
  SDK, consistent with existing HyperShift patterns for GCP API access
- Implement reconcileGCPOIDCDocuments to upload .well-known/openid-configuration
  and keys.json to the OIDC bucket during HostedCluster reconciliation
- Implement cleanupGCPOIDCBucketData to remove documents on cluster deletion
  with proper finalizer management
- Inject GCSClient into HostedClusterReconciler with --gcp-oidc-storage-bucket
  flag to configure the target bucket
- Add comprehensive unit tests with fakeGCSClient mock

Signed-off-by: ckandaga <ckandaga@redhat.com>
Co-authored-by: Cursor <cursoragent@cursor.com>

Fixes : https://redhat.atlassian.net/browse/GCP-636 

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Store guest-cluster OIDC discovery documents in Google Cloud Storage with automated upload during reconciliation; cleanup on cluster deletion via a finalizer.
  * New CLI flag to configure the GCS bucket and automatic GCS client initialization.
  * Operator exposes a ValidOIDCConfiguration status condition reflecting GCP OIDC state.

* **Tests**
  * Added unit tests covering reconciliation, uploads, cleanup, finalizer behavior, and error scenarios for GCP OIDC operations.


ckandag:
Re: CodeRabbit suggestion to always call `reconcileGCPOIDCDocuments` even when `GCPOIDCStorageBucketName`/`GCSClient` are unset —

Intentionally keeping the guard. Unlike AWS where S3 OIDC is always expected, on GCP the `--gcp-oidc-storage-bucket-name` flag is optional. Not every GCP deployment needs OIDC document upload (e.g., self-managed clusters, dev environments without WIF). Surfacing `ValidOIDCConfiguration=False` when the operator was intentionally not configured for it would create false-alarm noise.

hypershift-operator/controllers/hostedcluster/gcp_oidc.go
```diff
@@ -0,0 +1,108 @@
...
+		}
+		return fmt.Errorf("failed to get controlplane service account signing key %q: %w", client.ObjectKeyFromObject(secret), err)
+	}
+
+	if _, ok := secret.Data[serviceSignerPublicKey]; !ok {
+		return fmt.Errorf("controlplane service account signing key secret %q missing required key %s", client.ObjectKeyFromObject(secret), serviceSignerPublicKey)
+	}
+
+	params := oidc.OIDCGeneratorParams{
+		IssuerURL: hcp.Spec.IssuerURL,
```
csrwng, line 64: **IssuerURL is not derived from the bucket name.** In the AWS path, `oidcDiscoveryURL(bucket, region, infraID)` computes the IssuerURL from the S3 bucket when not explicitly provided. There's no GCP equivalent — `hcp.Spec.IssuerURL` is used as-is with no validation that it corresponds to the GCS bucket's public URL (e.g., `https://storage.googleapis.com/<bucket>/<infraID>`). The OIDC documents uploaded here need to be reachable at the IssuerURL for Workload Identity Federation to work. Should the IssuerURL be computed from the bucket name, or at minimum validated to match?

hypershift-operator/controllers/hostedcluster/gcp_oidc.go
```diff
@@ -0,0 +1,108 @@
...
+)
+
+const (
+	gcpOIDCDocumentsFinalizer = "hypershift.io/gcp-oidc-discovery"
+)
+
+func (r *HostedClusterReconciler) reconcileGCPOIDCDocuments(ctx context.Context, log logr.Logger, hcluster *hyperv1.HostedCluster, hcp *hyperv1.HostedControlPlane) error {
+	if hcluster.Spec.ServiceAccountSigningKey != nil && hcluster.Spec.ServiceAccountSigningKey.Name != "" {
+		return nil
+	}
```
csrwng, line 29: **Missing finalizer-based short-circuit.** The AWS equivalent (`reconcileAWSOIDCDocuments`) skips re-uploading when the finalizer is already present, avoiding constant re-uploads on every reconcile loop. The GCP version re-uploads every time. This is both less efficient and inconsistent with the AWS pattern. Consider adding the same short-circuit:

```suggestion
	if controllerutil.ContainsFinalizer(hcluster, gcpOIDCDocumentsFinalizer) {
		return nil
	}

	if r.GCPOIDCStorageBucketName == "" || r.GCSClient == nil {
```

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -2112,6 +2116,32 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
 				}
 			}
```
csrwng, line 2078: **Silently skipping OIDC reconciliation when bucket is not configured.** The AWS path calls `reconcileAWSOIDCDocuments` unconditionally and returns a hard error if the bucket isn't set, surfacing `ValidOIDCConfiguration=False`. The GCP path silently skips the entire block, leaving no signal that OIDC is misconfigured. This also makes the guard inside `reconcileGCPOIDCDocuments` (lines 29-31) dead code.

Consider matching the AWS behavior — call `reconcileGCPOIDCDocuments` unconditionally and set `ValidOIDCConfiguration=False` when the bucket isn't configured.

csrwng (commented):
Reviewed the non-vendored files. Main concerns are around behavioral inconsistencies with the existing AWS OIDC pattern.

hypershift-operator/controllers/hostedcluster/gcp_oidc.go
ckandag: This is by design. The issuerURL will not be derived from the bucket name because GCS buckets in GCP will not be made publicly accessible or directly reachable via their native bucket URLs. Instead, a GCP CDN proxy will be deployed in front of the buckets to serve the docs.

For managed-service deployments, the issuerURL is correctly configured on a per-region basis during HostedCluster spec generation.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
ckandag: The idea was to not force a  bucket configuration and continue to support the current case where oidc is pre-configured and signing key is passed in during cluster creation.

hypershift-operator/controllers/hostedcluster/gcp_oidc.go
ckandag: fixed

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -2092,6 +2096,32 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
...
 		}
+	case hyperv1.GCPPlatform:
+		if r.GCPOIDCStorageBucketName != "" && r.GCSClient != nil {
+			if err := r.reconcileGCPOIDCDocuments(ctx, log, hcluster, hcp); err != nil {
+				meta.SetStatusCondition(&hcluster.Status.Conditions, metav1.Condition{
+					Type:               string(hyperv1.ValidOIDCConfiguration),
```
apahim: conditions duplicated with the AWS case above (line 2040), should this be extratected into a shared helper?

hypershift-operator/controllers/hostedcluster/gcp_oidc.go
```diff
@@ -0,0 +1,114 @@
...
+		objectName := hcluster.Spec.InfraID + path
+		if err := r.GCSClient.DeleteObject(ctx, r.GCPOIDCStorageBucketName, objectName); err != nil {
+			log.Error(err, "Failed to delete OIDC document from GCS", "object", objectName)
+			deleteErrors = append(deleteErrors, err)
+		}
+	}
+
+	if len(deleteErrors) > 0 {
+		return fmt.Errorf("failed to delete %d OIDC documents from GCS bucket %s", len(deleteErrors), r.GCPOIDCStorageBucketName)
+	}
```
apahim, line 114: we should log what the specific errors are here. somthing like:
```
  if err := errors.Join(deleteErrors...); err != nil {
      return fmt.Errorf("failed to delete OIDC documents from GCS bucket %s: %w", r.GCPOIDCStorageBucketName, err)
  }
```

hypershift-operator/controllers/hostedcluster/gcp_oidc.go
```diff
@@ -0,0 +1,114 @@
...
+
+	for path, generator := range oidcDocumentGenerators() {
+		bodyReader, err := generator(params)
+		if err != nil {
+			return fmt.Errorf("failed to generate OIDC document %s: %w", path, err)
+		}
+		objectName := hcluster.Spec.InfraID + path
+		if err := r.GCSClient.UploadObject(ctx, r.GCPOIDCStorageBucketName, objectName, bodyReader); err != nil {
+			return fmt.Errorf("failed to upload %s to gs://%s/%s: %w", path, r.GCPOIDCStorageBucketName, objectName, err)
```
apahim, line 75: nitpick: this might be throwing an error for the a different document each loop, as the range oidcDocumentGenerators() above is non-deterministic by design

apahim (approved):
/lgtm

a few nits, nothing critical

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
ckandag: refactored

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
ckandag: Moved the check into reconcileOIDCDocuments, aligning the behavior with AWS, and now throwing an error when neither a signing key is configured in the spec nor bucket configuration is provided for OIDC upload.

hypershift-operator/controllers/hostedcluster/gcp_oidc.go
ckandag: The error already includes the specific path/object that failed, so it's always clear which document hit the issue regardless of iteration order. There are also only two documents (.well-known/openid-configuration and keys.json), so in practice if one fails due to a GCS issue other will with the same issue

Addressed the other two comments but leaving this as is.

PR #8421, opened 2026-05-05
OCPBUGS-84303: fix(api): add IPv6 OVN join subnet config to prevent dual-stack routing collision

orenc1:
When a KubeVirt hosted cluster and its management cluster both use OVN-Kubernetes with dual-stack networking, they each default to fd98::/64 for the IPv6 join switch subnet. External IPv6 LoadBalancer traffic targeting VM pods is SNAT'd to the management cluster's join IP (e.g. fd98::2). Inside the VM, the guest cluster's OVN intercepts the response because it also owns fd98::/64, black-holing the packet.

This commit fixes the issue in two ways:

1. Automatic KubeVirt default: for KubeVirt hosted clusters with OVNKubernetes, the reconciler now sets IPv6.InternalJoinSubnet to fd99::/64 by default, avoiding the collision with the management cluster's fd98::/64. This mirrors the existing V4InternalSubnet override (100.66.0.0/16) already in place for IPv4.

2. User-facing API: adds OVNIPv6Config type to OVNKubernetesConfig, allowing explicit configuration of IPv6 internalJoinSubnet and internalTransitSwitchSubnet for any platform. This maps to the upstream operatorv1.IPv6OVNKubernetesConfig and includes IPv6 CIDR format validation via CEL rules.

Also extends CIDR overlap validation in the HostedCluster webhook to cover IPv6 OVN subnets, and adds envtest CRD validation cases.

Fixes  https://redhat.atlassian.net/browse/OCPBUGS-84303

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * OVN-Kubernetes now exposes IPv6 internal transit and join subnet settings and will default a KubeVirt join subnet when the cluster has IPv6 network(s); operator-provided IPv6 values can override the default.

* **Bug Fixes / Validation**
  * Validation added to prevent identical IPv6 join/transit subnets, enforce IPv6 CIDR/prefix format and allowed prefix range, and include OVN IPv6 (and KubeVirt defaults) in overlap checks.

* **Tests**
  * Expanded coverage for IPv6-only and mixed IPv4+IPv6 scenarios, defaulting, overrides, and overlap failures.


orenc1:
> Caution
> 
> Some comments are outside the diff and can’t be posted inline due to platform limitations.
> ⚠️ Outside diff range comments (1)
> 
> 🤖 Prompt for all review comments with AI agents
> 
> ℹ️ Review info

This is a pre-existing pattern — the IPv4 block uses the exact same additive-only approach (only copies non-empty values, never clears). The MTU field follows the same convention as well. Changing this behavior would alter semantics for all OVN config fields (IPv4, IPv6, MTU), which is out of scope for this bug fix.

Additionally, OVN internal join/transit subnets are effectively immutable after cluster creation — changing them at runtime would break OVN networking. The additive-only reconciliation pattern is appropriate here.

orenc1:
The fix is verified. with a different, non-conflicting OVN join IPv6 subnet for the hosted cluster, LB service is accessible:

## Verification: IPv6 LoadBalancer Fix (OVN Join Subnet Collision)
### Environment
- **Management cluster**: 3-node bare metal (cnvqe-064/065/066), dual-stack
- **Hosted cluster**: `dualstack-fix`, 3 KubeVirt worker nodes, dual-stack (OCP built from PR payload)
- **OVN IPv6 join subnet (guest)**: `fd99::/64` ✅ (previously collided at `fd98::/64`)
### Setup
1. Created a `hello-openshift` deployment in the hosted cluster (`lb-test` namespace)
2. Exposed it as a **dual-stack LoadBalancer** service (`RequireDualStack`)
3. KubeVirt cloud provider created a proxy service on the management cluster
4. MetalLB assigned:
   - IPv4 VIP: `10.46.255.14`
   - IPv6 VIP: `fc00:f853:ccd:e799::4`
### Results
| Source | Target | Result |
|--------|--------|--------|
| Management cluster pod → IPv4 LB (`10.46.255.14:8080`) | Hosted cluster service | ✅ **PASS** — `Hello OpenShift!` |
| Management cluster pod → IPv6 LB (`[fc00:f853:ccd:e799::4]:8080`) | Hosted cluster service | ✅ **PASS** — `Hello OpenShift!` (3/3 runs) |
| Management cluster node → IPv6 LB | Hosted cluster service | ✅ **PASS** — `Hello OpenShift!` |
### Key Confirmation
The guest cluster's OVN network operator config shows the fix is active:
```json
{
  "ipv6": {
    "internalJoinSubnet": "fd99::/64"
  }
}
```
/verified by @orenc1

jparrill (commented):

cmd/install/assets/crds/hypershift-operator/tests/hostedclusters.hypershift.openshift.io/stable.hostedclusters.networking.testsuite.yaml
```diff
@@ -619,3 +619,92 @@ tests:
...
+            type: Route
+            route: {}
+        - service: Konnectivity
+          servicePublishingStrategy:
+            type: Route
+            route: {}
+        - service: Ignition
+          servicePublishingStrategy:
+            type: Route
+            route: {}
```
jparrill, line 710: The positive cases look good, but I think we need some negative envtest cases too — the CEL regex escaping is notoriously fragile and without these we can't really be sure the rules actually reject bad input on a real API server. Could you add cases for:

- Same value for both `internalJoinSubnet` and `internalTransitSwitchSubnet` in IPv6 (should hit `"ipv6 internalJoinSubnet and internalTransitSwitchSubnet must not be the same"`)
- An invalid IPv6 CIDR like `"not-a-cidr"`
- A prefix length > 125 (e.g. `fd99::/126`)
- Empty `ipv6: {}` to exercise the `MinProperties=1` constraint

api/hypershift/v1beta1/operator.go
```diff
@@ -126,6 +136,46 @@ type OVNIPv4Config struct {
...
+	// specified for OVN-Kubernetes as well as other networks used on the host.
+	// When omitted, this means no opinion and the platform is left to choose a reasonable
+	// default which is subject to change over time.
+	// The current default subnet is fd97::/64.
+	// The subnet must be large enough to accommodate one IP per node in your cluster.
+	// The value must be in proper IPv6 CIDR format.
+	// Note that IPv6 dual addresses are not permitted.
+	// +kubebuilder:validation:MaxLength=48
+	// +kubebuilder:validation:MinLength=3
+	// +kubebuilder:validation:XValidation:rule="self.matches('^\\\\s*((([0-9A-Fa-f]{1,4}:){7}([0-9A-Fa-f]{1,4}|:))|(([0-9A-Fa-f]{1,4}:){6}(:[0-9A-Fa-f]{1,4}|:))|(([0-9A-Fa-f]{1,4}:){5}((:[0-9A-Fa-f]{1,4}){1,2}|:))|(([0-9A-Fa-f]{1,4}:){4}((:[0-9A-Fa-f]{1,4}){1,3}|:))|(([0-9A-Fa-f]{1,4}:){3}((:[0-9A-Fa-f]{1,4}){1,4}|:))|(([0-9A-Fa-f]{1,4}:){2}((:[0-9A-Fa-f]{1,4}){1,5}|:))|(([0-9A-Fa-f]{1,4}:){1}((:[0-9A-Fa-f]{1,4}){1,6}|:))|(::((:[0-9A-Fa-f]{1,4}){1,7}|:)))\\\\s*/([0-9]|[1-9][0-9]|1[0-1][0-9]|12[0-8])$')", message="Subnet must be in valid IPv6 CIDR format (e.g., fd97::/64)"
```
jparrill: Noticed the regex has `\\s*` which lets values like `"  fd99::/64"` pass CEL validation, but then `net.ParseCIDR()` chokes on them at runtime in the controller. The IPv4 regex doesn't allow whitespace — would be good to keep them consistent and drop the `\\s*` so users get a clear error at admission time rather than a cryptic one during reconciliation.

Also thought about suggesting the built-in `isCIDR(self) && cidr(self).ip().family() == 6` (which is what upstream openshift/api uses), but our envtests still run against k8s 1.30.3 (OCP 4.17) and `isCIDR()` was only added in 1.31. So the regex is necessary for now — maybe worth a TODO comment for when we drop 1.30.

api/hypershift/v1beta1/operator.go
```diff
@@ -126,6 +136,46 @@ type OVNIPv4Config struct {
 	InternalJoinSubnet string `json:"internalJoinSubnet,omitempty"`
 }
 
+// OVNIPv6Config contains IPv6-specific configuration options for OVN-Kubernetes.
+// https://github.com/openshift/api/blob/6d3c4e25a8d3aeb57ad61649d80c38cbd27d1cc8/operator/v1/types_network.go#L541-L570
+// +kubebuilder:validation:MinProperties=1
+type OVNIPv6Config struct {
```
jparrill, line 160: Per our `api/CLAUDE.md` conventions, new API types should have a round-trip serialization test to make sure N-1 code can handle data from the new version and vice versa. I know `omitzero` should make this safe in practice, but having the explicit test keeps us consistent with the guidelines and protects against regressions.

api/hypershift/v1beta1/operator.go
```diff
@@ -126,6 +136,46 @@ type OVNIPv4Config struct {
...
+	// default join subnet (fd98::/64).
+	// The subnet must be large enough to accommodate one IP per node in your cluster.
+	// The value must be in proper IPv6 CIDR format.
+	// Note that IPv6 dual addresses are not permitted.
+	// +kubebuilder:validation:MaxLength=48
+	// +kubebuilder:validation:MinLength=3
+	// +kubebuilder:validation:XValidation:rule="self.matches('^\\\\s*((([0-9A-Fa-f]{1,4}:){7}([0-9A-Fa-f]{1,4}|:))|(([0-9A-Fa-f]{1,4}:){6}(:[0-9A-Fa-f]{1,4}|:))|(([0-9A-Fa-f]{1,4}:){5}((:[0-9A-Fa-f]{1,4}){1,2}|:))|(([0-9A-Fa-f]{1,4}:){4}((:[0-9A-Fa-f]{1,4}){1,3}|:))|(([0-9A-Fa-f]{1,4}:){3}((:[0-9A-Fa-f]{1,4}){1,4}|:))|(([0-9A-Fa-f]{1,4}:){2}((:[0-9A-Fa-f]{1,4}){1,5}|:))|(([0-9A-Fa-f]{1,4}:){1}((:[0-9A-Fa-f]{1,4}){1,6}|:))|(::((:[0-9A-Fa-f]{1,4}){1,7}|:)))\\\\s*/([0-9]|[1-9][0-9]|1[0-1][0-9]|12[0-8])$')", message="Subnet must be in valid IPv6 CIDR format (e.g., fd98::/64)"
+	// +kubebuilder:validation:XValidation:rule="self.matches('^.*/[0-9]+$') && int(self.split('/')[1]) <= 125", message="subnet must be in the range /0 to /125 inclusive"
+	// +optional
+	InternalJoinSubnet string `json:"internalJoinSubnet,omitempty"`
```
jparrill, line 200: Not for this PR, but changing the OVN join or transit subnet on a running cluster would break networking pretty badly. Might be worth a follow-up to add `self == oldSelf` CEL immutability rules here. The IPv4 fields don't have them either (pre-existing gap), so we could add immutability to both families at once.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/network/reconcile.go
```diff
@@ -61,6 +67,12 @@ func ReconcileNetworkOperator(network *operatorv1.Network, networkType hyperv1.N
 			if network.Spec.DefaultNetwork.OVNKubernetesConfig.GenevePort == nil {
 				network.Spec.DefaultNetwork.OVNKubernetesConfig.GenevePort = &port
 			}
+			if network.Spec.DefaultNetwork.OVNKubernetesConfig.IPv6 == nil {
```
jparrill: This is going to set `IPv6.InternalJoinSubnet = fd99::/64` on every KubeVirt+OVN cluster, even if it's a pure IPv4 deployment. I'm not sure how the CNO handles getting an IPv6 join subnet config when there's no IPv6 networking configured — might be harmless, but worth checking. Could we guard this behind some kind of dual-stack detection (e.g. checking if any of the cluster/service/machine networks have an IPv6 CIDR), or have you already confirmed this is safe with the CNO folks?

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -4334,6 +4334,34 @@ func validateSliceNetworkCIDRs(hc *hyperv1.HostedCluster) field.ErrorList {
...
+		var ipv6JoinSubnet, ipv6TransitSubnet string
+		if hc.Spec.OperatorConfiguration != nil && hc.Spec.OperatorConfiguration.ClusterNetworkOperator != nil &&
+			hc.Spec.OperatorConfiguration.ClusterNetworkOperator.OVNKubernetesConfig != nil {
+			ipv6JoinSubnet = hc.Spec.OperatorConfiguration.ClusterNetworkOperator.OVNKubernetesConfig.IPv6.InternalJoinSubnet
+			ipv6TransitSubnet = hc.Spec.OperatorConfiguration.ClusterNetworkOperator.OVNKubernetesConfig.IPv6.InternalTransitSwitchSubnet
+		}
+		// The reconciler defaults KubeVirt IPv6 join subnet to fd99::/64;
+		// include the effective value so overlaps are caught at admission time.
+		if ipv6JoinSubnet == "" && hc.Spec.Platform.Type == hyperv1.KubevirtPlatform {
+			ipv6JoinSubnet = "fd99::/64"
```
jparrill: This same default lives as a proper constant `kubevirtDefaultV6InternalJoinSubnet` in `reconcile.go`, but here it's a bare string. Since HO and CPO are different binaries released independently, if the default ever changes we'd need to remember to update both. Maybe worth moving it to the `api/` module so both can reference it, or at least adding a local constant with a comment pointing to the other one?

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -4334,6 +4334,34 @@ func validateSliceNetworkCIDRs(hc *hyperv1.HostedCluster) field.ErrorList {
 			}
 		}
 	}
+
+	if hc.Spec.Networking.NetworkType == hyperv1.OVNKubernetes {
```
jparrill, line 4340: Nice that you're injecting the KubeVirt IPv6 default into the overlap validation — but the existing IPv4 path doesn't do the same thing. If someone creates a KubeVirt cluster with a machine network that overlaps `100.66.0.0/16`, the admission check won't catch it and it'll blow up at runtime. Since you're already here and the pattern is the same, could you add the IPv4 equivalent too? Something like:

```go
if ipv4JoinSubnet == "" && hc.Spec.Platform.Type == hyperv1.KubevirtPlatform {
    ipv4JoinSubnet = "100.66.0.0/16"
}
```

I think doing this in the same PR makes more sense than a follow-up since the IPv6 pattern is right there as a template.

api/hypershift/v1beta1/operator.go
jparrill: Hey, one more thought on this — I looked into whether we could use `isCIDR()` + `cidr()` instead of this regex, since that's what upstream `openshift/api` uses for the exact same `IPv6OVNKubernetesConfig` type. Turns out it's cheaper *and* catches more things:

| Check | Regex (380 chars) | `isCIDR()` + `cidr()` |
|---|---|---|
| Valid IPv6 syntax | ✅ | ✅ |
| CIDR prefix present (`/N`) | ✅ | ✅ |
| Prefix range 0–125 | ✅ (extra rule) | ✅ (`cidr(self).prefixLength() <= 125`) |
| IPv6 family (rejects IPv4) | ✅ (implicit) | ✅ (`cidr(self).ip().family() == 6`) |
| Host bits set after prefix (e.g. `fd99::1/64`) | ❌ passes | ✅ rejects |
| IPv4-mapped IPv6 (`::ffff:1.2.3.4/96`) | ❌ passes | ✅ rejects |
| Leading/trailing whitespace | ❌ passes (the `\s*` bug) | ✅ rejects |

CEL cost budget:

| Approach | Estimated cost | % of per-rule budget (10M) |
|---|---|---|
| `isCIDR()` + `cidr()` | ~15 | 0.00015% |
| Regex (380 chars) | ~476 | 0.00476% |

So it's 32x cheaper and strictly more correct. The two replacement rules would be:

```go
// +kubebuilder:validation:XValidation:rule="isCIDR(self) && cidr(self).ip().family() == 6", message="Subnet must be in valid IPv6 CIDR format"
// +kubebuilder:validation:XValidation:rule="isCIDR(self) && cidr(self).prefixLength() <= 125", message="subnet must be in the range /0 to /125 inclusive"
```

I checked availability — `isCIDR()` was backported to OCP 4.14+ (k8s 1.28+), and our lowest OCP envtest version is 1.30.3. Vanilla k8s envtests start at 1.31.0 where it's also available. So we're good across the board.

jparrill: TL;DR: Switch from regex to isCIDR() — it's what upstream openshift/api already uses for the same type, it's 32x cheaper on CEL budget, catches edge cases the regex misses (host bits, IPv4-mapped, whitespace), and it's available on all our envtest k8s versions. Please @enxebre @JoelSpeed correct me if I'm wrong

orenc1: ok, switching to using CEL validation with `isCIDR(self) && cidr(self)` instead of the regex :+1:

cmd/install/assets/crds/hypershift-operator/tests/hostedclusters.hypershift.openshift.io/stable.hostedclusters.networking.testsuite.yaml
orenc1: Added 4 negative test cases to `stable.hostedclusters.networking.testsuite.yaml`:

Same value for both `internalJoinSubnet` and `internalTransitSwitchSubnet` --> expects `"ipv6 internalJoinSubnet and internalTransitSwitchSubnet must not be the same"`
Invalid IPv6 CIDR (`"not-a-cidr"`) --> expects `"Subnet must be in valid IPv6 CIDR format"`
Prefix length > 125 (fd99::/126) --> expects `"subnet must be in the range /0 to /125 inclusive"`
Empty ipv6: {} --> expects `"should have at least 1 properties" (MinProperties=1)`

api/hypershift/v1beta1/operator.go
orenc1: we can use `isCIDR(self) && cidr(self).ip().family() == 6` and `isCIDR(self) && cidr(self).prefixLength() <= 125`

api/hypershift/v1beta1/operator.go
orenc1: There are no existing round-trip serialization test patterns in this repo's api/ module. The `omitzero` tag on the non-pointer `OVNIPv6Config` field handles the N-1 case correctly (zero value is omitted from serialization).

control-plane-operator/hostedclusterconfigoperator/controllers/resources/network/reconcile.go
orenc1: added a guard of `hasIPv6Network()` to add the IPv6 config.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
orenc1: Moved the fd99::/64 and 100.66.0.0/16 constants to `api/hypershift/v1beta1/operator.go` as exported constants (`KubevirtDefaultV6InternalJoinSubnet`, `KubevirtDefaultV4InternalSubnet`). Both the CPO reconciler and the HO controller now reference these from the shared API module.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
orenc1: Added the IPv4 equivalent in `validateSliceNetworkCIDRs`: when the platform is KubeVirt, network type is OVNKubernetes, and no explicit IPv4 join subnet is set, the KubeVirt default 100.66.0.0/16 is included in overlap validation. Added 2 new unit tests: one verifying overlap detection with the default, and one verifying non-overlapping networks succeed

api/hypershift/v1beta1/operator.go
everettraven: New optional fields that are correctly omitted (as this one is) will round trip successfully.

api/hypershift/v1beta1/operator.go
```diff
@@ -126,6 +150,46 @@ type OVNIPv4Config struct {
...
+	// When omitted, this means no opinion and the platform is left to choose a reasonable
+	// default which is subject to change over time.
+	// The current default subnet is fd97::/64.
+	// The subnet must be large enough to accommodate one IP per node in your cluster.
+	// The value must be in proper IPv6 CIDR format.
+	// Note that IPv6 dual addresses are not permitted.
+	// +kubebuilder:validation:MaxLength=48
+	// +kubebuilder:validation:MinLength=3
+	// +kubebuilder:validation:XValidation:rule="isCIDR(self) && cidr(self).ip().family() == 6", message="Subnet must be in valid IPv6 CIDR format (e.g., fd97::/64)"
+	// +kubebuilder:validation:XValidation:rule="isCIDR(self) && cidr(self).prefixLength() <= 125", message="subnet must be in the range /0 to /125 inclusive"
```
everettraven, line 177: Please explicitly document the constraints enforced by these markers in the GoDoc.

The generated documentation that end-users using something like `kubectl explain ...` would see does not include the marker contents. Without explicit documentation they will not be aware of these constraints without having to jump through more hoops.

everettraven, line 177: General note for any other reviewers - the CIDR CEL library was introduced in Kubernetes 1.30. From what I recall being told, we are look for N-4 compatibility from the kube version skew, so with 4.22 being based on 1.35 that means the oldest kube version we would end up running on would be 1.31. This should be OK.

api/hypershift/v1beta1/operator.go
```diff
@@ -126,6 +150,46 @@ type OVNIPv4Config struct {
...
+	// For KubeVirt hosted clusters, if this field is not set, HyperShift will
+	// automatically use fd99::/64 to avoid collisions with the management cluster's
+	// default join subnet (fd98::/64).
+	// The subnet must be large enough to accommodate one IP per node in your cluster.
+	// The value must be in proper IPv6 CIDR format.
+	// Note that IPv6 dual addresses are not permitted.
+	// +kubebuilder:validation:MaxLength=48
+	// +kubebuilder:validation:MinLength=3
+	// +kubebuilder:validation:XValidation:rule="isCIDR(self) && cidr(self).ip().family() == 6", message="Subnet must be in valid IPv6 CIDR format (e.g., fd98::/64)"
+	// +kubebuilder:validation:XValidation:rule="isCIDR(self) && cidr(self).prefixLength() <= 125", message="subnet must be in the range /0 to /125 inclusive"
```
everettraven, line 197: Similar comment re: GoDoc.

api/hypershift/v1beta1/operator.go
everettraven: If we know ahead of time that making a change to the networking configuration on a running cluster would be bad, let's go ahead and prevent it where we can now.

Just a note that immutability rules must be done on the closest parent field that is required, otherwise the field isn't actually immutable because it can be removed by removing an optional parent (or the field itself) in the chain and re-applying with a different value.

cmd/install/assets/crds/hypershift-operator/tests/hostedclusters.hypershift.openshift.io/stable.hostedclusters.networking.testsuite.yaml
```diff
@@ -619,3 +619,268 @@ tests:
...
+    expectedError: "ipv6 internalJoinSubnet and internalTransitSwitchSubnet must not be the same"
+
+  - name: When ovnKubernetesConfig ipv6 internalJoinSubnet has invalid CIDR it should fail
```
everettraven, line 757: Replicate this test for `internalTrasitSwitchSubnet`?

cmd/install/assets/crds/hypershift-operator/tests/hostedclusters.hypershift.openshift.io/stable.hostedclusters.networking.testsuite.yaml
```diff
@@ -619,3 +619,268 @@ tests:
...
+          servicePublishingStrategy:
+            type: Route
+            route: {}
+        - service: Ignition
+          servicePublishingStrategy:
+            type: Route
+            route: {}
+    expectedError: "Subnet must be in valid IPv6 CIDR format"
+
+  - name: When ovnKubernetesConfig ipv6 internalJoinSubnet has prefix length greater than 125 it should fail
```
everettraven, line 845: Replicate this test for `internalTrasitSwitchSubnet`?

everettraven (changes requested):
API structure looks good to me overall. Just a handful of comments.

cmd/install/assets/crds/hypershift-operator/tests/hostedclusters.hypershift.openshift.io/stable.hostedclusters.networking.testsuite.yaml
orenc1: envtest case has been added, expecting range of /0 to /125 inclusive

cmd/install/assets/crds/hypershift-operator/tests/hostedclusters.hypershift.openshift.io/stable.hostedclusters.networking.testsuite.yaml
orenc1: envtest case has been added, expecting valid IPv6 CIDR format

api/hypershift/v1beta1/operator.go
orenc1: added.

api/hypershift/v1beta1/operator.go
orenc1: explicit documentation to both `InternalTransitSwitchSubnet` and `InternalJoinSubnet` has been added, clarifying the constraints, including the immutability.

api/hypershift/v1beta1/operator.go
orenc1: ok, added three layers of protection:

`self == oldSelf` on each field (prevents value changes)
`!has(oldSelf.ipv6) || has(self.ipv6)` on `OVNKubernetesConfig` (prevents removing the ipv6 parent to bypass field immutability)
`!has(oldSelf.ipv6) || !has(oldSelf.ipv6.internalJoinSubnet) || (has(self.ipv6) && has(self.ipv6.internalJoinSubnet))` (and same for transit subnet) to prevent removing individual fields through the optional parent chain

api/hypershift/v1beta1/operator.go
```diff
@@ -62,7 +76,11 @@ type ClusterNetworkOperatorSpec struct {
...
+// +kubebuilder:validation:XValidation:rule="!has(self.ipv6) || !has(self.ipv6.internalJoinSubnet) || !has(self.ipv6.internalTransitSwitchSubnet) || self.ipv6.internalJoinSubnet != self.ipv6.internalTransitSwitchSubnet", message="ipv6 internalJoinSubnet and internalTransitSwitchSubnet must not be the same"
 // +kubebuilder:validation:XValidation:rule="!has(oldSelf.mtu) || has(self.mtu)",message="mtu is immutable once set and cannot be removed"
+// +kubebuilder:validation:XValidation:rule="!has(oldSelf.ipv6) || has(self.ipv6)", message="ipv6 is immutable once set and cannot be removed"
+// +kubebuilder:validation:XValidation:rule="!has(oldSelf.ipv6) || !has(oldSelf.ipv6.internalJoinSubnet) || (has(self.ipv6) && has(self.ipv6.internalJoinSubnet))", message="ipv6.internalJoinSubnet cannot be removed once set"
+// +kubebuilder:validation:XValidation:rule="!has(oldSelf.ipv6) || !has(oldSelf.ipv6.internalTransitSwitchSubnet) || (has(self.ipv6) && has(self.ipv6.internalTransitSwitchSubnet))", message="ipv6.internalTransitSwitchSubnet cannot be removed once set"
```
everettraven, line 84: For these immutability rules, what happens if the parent `ovnKubernetesConfig` field is set, removed, and re-applied with new values?

orenc1: you're right, good catch. if we remove the entire `ovnKubernetesConfig` field, then re-apply with different values, the immutability rules could be bypassed.
to mitigate that, i'm adding a validation rule of "cannot be removed once set" on `ClusterNetworkOperatorSpec` for `ovnKubernetesConfig`.

everettraven: It still looks like you could just remove the `ClusterNetworkOperatorSpec` field based on: https://github.com/openshift/hypershift/blob/f76be88237e1ae0c1c4df2e34d8b0f9eb8ec7953/api/hypershift/v1beta1/hostedcluster_types.go#L2372-L2375

If you want true immutability you need to set the rule all the way up to the first required field, which looks like it means on the root `HostedCluster` type because even `spec` is optional: https://github.com/openshift/hypershift/blob/f76be88237e1ae0c1c4df2e34d8b0f9eb8ec7953/api/hypershift/v1beta1/hostedcluster_types.go#L2411-L2413

I'd make sure you pay particular attention to how immutability applies as you go up the chain of fields to make sure you don't make something that is OK to mutate immutable.

orenc1: I see. how do you suggest to proceed? should I modify the validation rules of the higher fields in the chain, up to the root? I feel it's beyond the scope of this PR.

everettraven: I don't have a strong preference. If this significantly increases the scope of this work to achieve this, I'm OK with best effort immutability for now with a more concrete immutability story for the near-ish future.

everettraven (approved):
Aside from the immutability rules being able to be circumvented, this looks good to me.

Not blocking on concrete immutability rules to prevent unnecessary scope increase for this PR. It has best-effort immutability rules in place which is good enough for me for now.

/approve

PR #8422, opened 2026-05-05
*: proposed follow-ups after cluster deletion investigation

stevekuznetsov:
control-plane-operator: signal that we're giving up on deletion

When the CPO decides to give up on deleting cloud resources, it emits a
log but otherwise does not expose this behavior through conditions,
making it hard to know that this happened.

Signed-off-by: Steve Kuznetsov <stekuznetsov@microsoft.com>

---

hypershift-operator: emit deletion progress condition

Instead of putting deletion progress into logs, where it's hard to
notice and spammy, present a user-facing condition that tracks the
phases. While doing that, improve the Namespace deletion code to be more
verbose. No top-level Namespace deletion condition seems to exist, so
cherry-pick a few we know are useful to highlight. We can always add
more later.

Signed-off-by: Steve Kuznetsov <stekuznetsov@microsoft.com>

---

*: remove low-quality logs

These log messages provide no clear information and are emitted
unconditionally, leading any log period from their controllers to be
full of them.

Signed-off-by: Steve Kuznetsov <stekuznetsov@microsoft.com>

---

## Summary by CodeRabbit

* **New Features**
  * Added an explicit "deleting" HostedCluster status condition and richer deletion progress updates that report each deletion phase and include namespace phase and blocker details.

* **Bug Fixes**
  * Cloud resource deletion timeouts are surfaced as a distinct status to stop indefinite retries and reflect timeout state.

* **Chores**
  * Removed redundant informational log messages to reduce control-plane log noise.

### DIFF
--- a/api/hypershift/v1beta1/hostedcluster_conditions.go
+++ b/api/hypershift/v1beta1/hostedcluster_conditions.go
@@ -252,6 +252,12 @@ const (
 	// **False / AutoNodeProgressing** means AutoNode is being enabled or disabled — the operation is in progress.
 	// **False / AutoNodeNotConfigured** means AutoNode is not configured in the spec and all Karpenter components have been removed.
 	AutoNodeEnabled ConditionType = "AutoNodeEnabled"
+
+	// HostedClusterDeleting indicates whether the HostedCluster is being deleted and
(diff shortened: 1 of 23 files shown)


csrwng:
/close

opened separate PRs for each commit:
#8426
#8427
#8428

PR #8424, opened 2026-05-05
ARO-26764: narrow HostedCluster primary watch to actionable metadata updates

tuxerrante:

The `HostedCluster` controller currently receives reconcile requests from its primary watch for updates that do not represent new desired state, especially status-only churn. That makes the controller noisier than it needs to be and makes it harder to reason about which reconciliations are actually driven by user intent versus controller-written status.

This change narrows only the primary `HostedCluster` watch so that it continues to reconcile on meaningful inputs, while avoiding self-induced reconciles from status-only updates.

The key design point is that this cannot be implemented as a pure generation-only filter. For `HostedCluster`, some metadata-only updates are already real control inputs for reconciliation and must continue to enqueue even when the object generation does not change. Examples include specific annotations consumed by reconciliation, mirrored `api.openshift.com/*` labels, scope transitions, and deletion start.

To keep the behavior correct while reducing noise, this PR introduces a dedicated primary predicate that allows:

- generation-changing updates
- deletion timestamp transitions
- actionable annotation changes
- actionable mirrored label changes

and filters out status-only updates from the primary watch.

The change is intentionally narrow:

- child-resource watches are left unchanged
- the API / CRD surface is unchanged
- the existing `ObservedGeneration` status model remains in place
- the reconcile loop still re-reads the current `HostedCluster`, so stale queued requests continue to act on the latest available object state

Alongside the watch change, this PR fixes two same-area behaviors that became important once metadata-only updates were treated more explicitly:

1. `ValidReleaseImage` is re-evaluated when `SkipReleaseImageValidation` is added or removed without a generation bump, so a previously cached `True` condition cannot remain stale.
2. mirrored `api.openshift.com/*` labels are actively removed from `HostedControlPlane` when they are removed from the `HostedCluster`, instead of only being copied on add/update.

Overall, the goal is to reduce unnecessary reconciliations without weakening any existing metadata-driven control path.

Fixes [ARO-26764](https://redhat.atlassian.net/browse/ARO-26764)

## Notes:

This PR is easier to review if you read it as a queueing-boundary change rather than as a reconcile algorithm rewrite.

The new predicate does **not** try to guarantee one reconcile per generation. Instead, it makes the enqueue boundary stricter and more explicit. The "latest object wins" behavior still comes from the existing reconcile entrypoint, which fetches the current `HostedCluster` by `NamespacedName` when the request is processed. That means older queued requests still reconcile against the latest stored object state, not the old event payload.

Because of that, the review focus should be:

- whether the allowlist of actionable metadata inputs is complete
- whether status-only updates are now correctly filtered out from the primary watch
- whether same-generation metadata changes still propagate correctly
- whether the blast radius is appropriately contained to the shared `HostedCluster` primary watch and adjacent regression fixes

Expected impact:

- fewer self-induced `HostedCluster` reconciliations from status-only updates
- no intended behavior change for spec updates
- no intended behavior change for metadata-only inputs that already affect reconciliation
- HCP-wide effect because this path is shared across Hosted Control Plane distributions, not specific to one platform

## Test plan:

- [x] `make verify`
- [x] predicate unit tests cover generation changes, status-only updates, actionable annotations, actionable labels, deletion timestamp changes, and scope transitions
- [x] falsification tests verify that stale queued requests still reconcile the latest `HostedCluster` generation and latest actionable metadata
- [x] regression tests cover same-generation `SkipReleaseImageValidation` toggles
- [x] regression tests cover removal of mirrored `api.openshift.com/*` labels from `HostedControlPlane`

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Added generation-aware, skip-aware release image validation with a dedicated “ReleaseImageValidationSkipped” reason.
  * Recognizes and mirrors OCM-prefixed labels (`api.openshift.com/`) to the control plane.

* **Bug Fixes**
  * HostedControlPlane mirroring now uses prefix-driven actionable cleanup (removes stale actionable entries) and improved mirrored annotation sync.
  * Reconciliation predicates more precisely trigger updates based on generation, actionable changes, and deletion timestamp changes.

* **Tests**
  * Added/expanded unit tests covering predicate behavior, stale label clearing, condition ObservedGeneration stamping, and release-image validation decision logic.


jparrill (commented):
Dropped some comments. Thanks

hypershift-operator/controllers/hostedcluster/hostedcluster_predicates.go
```diff
@@ -0,0 +1,204 @@
...
+	hyperv1.ClusterAPIOpenStackProviderImage,
+	hyperv1.OpenStackResourceControllerImage,
+	hyperv1.SkipReleaseImageValidation,
+	hyperv1.SkipKASConflicSANValidation,
+	hyperv1.SkipControlPlaneNamespaceDeletionAnnotation,
+	k8sutil.HostedClustersScopeAnnotation,
+}
+
+var hostedClusterActionableLabelPrefixes = []string{
+	"api.openshift.com",
```
jparrill: Nit: nice that you extracted this from the inline hardcoded string. It might be worth promoting it to a constant in the API package at some point (there's already `LimitedSupportLabel = "api.openshift.com/limited-support"` over there), but that's pre-existing and doesn't need to block this PR.

hypershift-operator/controllers/hostedcluster/hostedcluster_predicates.go
```diff
@@ -0,0 +1,204 @@
...
+		if strings.HasPrefix(key, prefix) {
+			prefixed[key] = value
+		}
+	}
+	return prefixed
+}
+
+func hostedClusterActionableLabelChanged(oldLabels, newLabels map[string]string) bool {
+	for _, prefix := range hostedClusterActionableLabelPrefixes {
+		if annotationPrefixChanged(oldLabels, newLabels, prefix) {
```
jparrill: Nit: `annotationPrefixChanged` works fine here since both annotations and labels are `map[string]string`, but reading `annotationPrefixChanged(oldLabels, newLabels, ...)` made me double-take. Maybe rename it to something map-generic like `prefixedKeyChanged`? Or even just renaming the parameters to `old`/`new` instead of `oldAnnotations`/`newAnnotations` would help.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -1188,13 +1188,19 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
...
 		// (hopefully they are cached).  Skip if we have already observed for
 		// this generation.
-		if condition == nil || condition.ObservedGeneration != hcluster.Generation || condition.Status != metav1.ConditionTrue {
+		if shouldValidateReleaseImage(hcluster, condition) {
 			condition := metav1.Condition{
 				Type:               string(hyperv1.ValidReleaseImage),
 				ObservedGeneration: hcluster.Generation,
 			}
+			skipReleaseImageValidation := hasSkipReleaseImageValidationAnnotation(hcluster)
 			err := r.validateReleaseImage(ctx, hcluster, releaseProvider)
```
jparrill: When `skipReleaseImageValidation` is true, `r.validateReleaseImage()` still runs here but its result gets thrown away by the switch — the method short-circuits internally on the same annotation check. Not a bug, but it took me a minute to convince myself it was intentional. Would it be cleaner to skip the call entirely?

```go
var err error
if !skipReleaseImageValidation {
    err = r.validateReleaseImage(ctx, hcluster, releaseProvider)
}
```

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -147,6 +146,7 @@ const (
 	previouslySyncedRestartDateAnnotation = "hypershift.openshift.io/previous-restart-date"
 	kasServingCertHashAnnotation          = "hypershift.openshift.io/kas-serving-cert-hash"
 	referencedResourceAnnotationPrefix    = "referenced-resource.hypershift.openshift.io/"
+	releaseImageValidationSkippedReason   = "ReleaseImageValidationSkipped"
```
jparrill: Suggestion: the other reason values for the `ValidReleaseImage` condition (`AsExpectedReason`, `InvalidImageReason`) live in `api/hypershift/v1beta1/hostedcluster_conditions.go`. Having this one as a controller-private constant creates an asymmetry — external consumers (ARO-HCP, OCM) that inspect conditions programmatically would need to hardcode the string. Consider promoting it to the API package alongside the existing reasons. Not blocking for this PR, but worth a follow-up.

hypershift-operator/controllers/hostedcluster/hostedcluster_predicates_test.go
```diff
@@ -0,0 +1,383 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        "example",
+			Namespace:   "clusters",
+			Generation:  generation,
+			Annotations: annotations,
+		},
+	}
+}
+
+func ptrToTime(value metav1.Time) *metav1.Time {
```
jparrill: Nit: `k8s.io/utils/ptr` is already imported in this package (`hostedcluster_controller.go`). You could use `ptr.To(metav1.Now())` here and drop this helper.

hypershift-operator/controllers/hostedcluster/hostedcluster_predicates_test.go
```diff
@@ -0,0 +1,383 @@
...
+			}
+		})
+	}
+}
+
+func TestHostedClusterPrimaryPredicate_WhenHostedClusterUpdatesItShouldFilterMeaningfulChanges(t *testing.T) {
+	t.Setenv(k8sutil.EnableHostedClustersAnnotationScopingEnv, "")
+	t.Setenv(k8sutil.HostedClustersScopeAnnotationEnv, "")
+
+	predicate := hostedClusterPrimaryPredicate(fake.NewClientBuilder().WithScheme(api.Scheme).Build())
```
jparrill: Heads up — `predicate` here shadows the `sigs.k8s.io/controller-runtime/pkg/predicate` package import. Not a problem right now since you don't reference the package after this line, but it could bite if someone adds a case that needs `predicate.GenerationChangedPredicate{}` or similar. Maybe rename to `pred`?

hypershift-operator/controllers/hostedcluster/hostedcluster_predicates.go
tuxerrante: Thanks — agreed it belongs in the API package. Tracking as a follow-up to keep this PR scoped to the watch change. The predicate already uses the extracted `hostedClusterActionableLabelPrefixes` slice, so there's a single source of truth within the controller; promoting it to a constant next to `LimitedSupportLabel` is the natural next step.

hypershift-operator/controllers/hostedcluster/hostedcluster_predicates.go
tuxerrante: Good catch — renamed to `prefixedKeyChanged` and `prefixedEntries` with generic parameter names (`old`/`new`, `m`) in 98cfe794e.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
tuxerrante: Done in 258cb70d0 — the call is now guarded with `if !skipReleaseImageValidation` so it's skipped entirely when the annotation is present.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
tuxerrante: Agreed — the asymmetry is real and ARO-HCP/OCM would benefit from having it in the API package. Tracking as a follow-up to keep this PR scoped to the watch change.

hypershift-operator/controllers/hostedcluster/hostedcluster_predicates_test.go
tuxerrante: Done in 705cfef34 — switched to `ptr.To(metav1.Now())` and removed the `ptrToTime` helper.

hypershift-operator/controllers/hostedcluster/hostedcluster_predicates_test.go
tuxerrante: Good point — renamed to `pred` in 705cfef34 to avoid the shadow.

hypershift-operator/controllers/hostedcluster/hostedcluster_predicates.go
tuxerrante: Follow-up tracked as [ARO-27290](https://issues.redhat.com/browse/ARO-27290).

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
tuxerrante: Follow-up tracked as [ARO-27291](https://issues.redhat.com/browse/ARO-27291).

tuxerrante: Update: addressed in #8577 (stacked PR) — promotes `releaseImageValidationSkippedReason` to `hyperv1.ReleaseImageValidationSkippedReason` in the API package alongside `AsExpectedReason` and `InvalidImageReason`.

hypershift-operator/controllers/hostedcluster/hostedcluster_predicates.go
tuxerrante: Correction: folded directly into this PR instead of a separate follow-up. See commit 2eb06732d — `hostedClusterActionableLabelPrefixes` now uses `hyperv1.OCMLabelPrefix` from the API package. ARO-27290 closed.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
tuxerrante: Correction: folded directly into this PR instead of a separate follow-up. See commit e0e1fc87d — `releaseImageValidationSkippedReason` is now `hyperv1.ReleaseImageValidationSkippedReason` in the API package. ARO-27291 closed.

hypershift-operator/controllers/hostedcluster/hostedcluster_predicates.go
tuxerrante: Commit link correction (GitHub resolves short SHAs against upstream, not the fork): [2eb06732d](https://github.com/tuxerrante/hypershift/commit/2eb06732d) — `hostedClusterActionableLabelPrefixes` now uses `hyperv1.OCMLabelPrefix`.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
tuxerrante: Commit link correction: [e0e1fc87d](https://github.com/tuxerrante/hypershift/commit/e0e1fc87d) — `releaseImageValidationSkippedReason` is now `hyperv1.ReleaseImageValidationSkippedReason`.

tuxerrante:
Force-pushed after rebasing this PR branch on latest `main`.

Kindly asking for another review after the rebase:

For traceability, this is the exact `range-diff` command run against this branch commit ranges (pre-rebase tip vs rebased tip):

```bash
git range-diff \
  e09cc2d208a4691f875441670e5cdc4173c374fb..b1377f914aab62af49c42c2cc9abc52699c6b917 \
  c35f662c8cf546db3ddcfeffb686ce6f3fbfde5a..bdf1bd82cb0690ee287e19b45c3d6dd69598e209
```

`range-diff` summary:
- 9/11 commits are equivalent (`=`)
- 2 commits were conflict-resolved (`!`) in `hostedcluster_controller.go`:
  - [`38d184062`](https://github.com/tuxerrante/hypershift/commit/38d1840629372bdab0c8622c0c329760e8fb6084) -> [`937f05a20`](https://github.com/openshift/hypershift/pull/8424/commits/937f05a202d643bcf2b4b5a6f10c286f1d06333f)
  - [`1280cf1af`](https://github.com/tuxerrante/hypershift/commit/1280cf1af4a9dddc522b6d248d3fa10aab5aa4c9) -> [`3f4856bf1`](https://github.com/openshift/hypershift/pull/8424/commits/3f4856bf1ab8f3cb03f157975fbd38c7c2eeff03)

Rebased commit stack in this PR:
- [`bdf1bd82c`](https://github.com/openshift/hypershift/pull/8424/commits/bdf1bd82cb0690ee287e19b45c3d6dd69598e209)
- [`6ab6edd79`](https://github.com/openshift/hypershift/pull/8424/commits/6ab6edd79f247e8f6d8961b3cb6c1db909d8ce20)
- [`fde93ef4d`](https://github.com/openshift/hypershift/pull/8424/commits/fde93ef4da0fdd89750f7bcbb98e7d79d3a321ee)
- [`0c00f8c36`](https://github.com/openshift/hypershift/pull/8424/commits/0c00f8c36e54dfb03af11f5702d9ae2de0ad0d0f)
- [`2112e3e74`](https://github.com/openshift/hypershift/pull/8424/commits/2112e3e74b01a7edfb1d307fb56b9ca6b12b8f5f)
- [`dad9e1749`](https://github.com/openshift/hypershift/pull/8424/commits/dad9e17491a03c18d677518c06dd62efa94a9011)
- [`9cba6d643`](https://github.com/openshift/hypershift/pull/8424/commits/9cba6d6433d6ec54f5a2fa3a5fd5b8f4d44ba36f)
- [`3f4856bf1`](https://github.com/openshift/hypershift/pull/8424/commits/3f4856bf1ab8f3cb03f157975fbd38c7c2eeff03)
- [`6a0143539`](https://github.com/openshift/hypershift/pull/8424/commits/6a0143539f5dfed1e20119d7c5831302aef5f4ee)
- [`a95d876d4`](https://github.com/openshift/hypershift/pull/8424/commits/a95d876d48bfe1178bce4ec6bf9534b878e2a6f2)
- [`937f05a20`](https://github.com/openshift/hypershift/pull/8424/commits/937f05a202d643bcf2b4b5a6f10c286f1d06333f)

Validation rerun after rebase:
- `make lint-fix`
- `go test ./hypershift-operator/controllers/hostedcluster/...`
- `cd api && go test ./...`
- `make verify`

@JoelSpeed 
@jparrill

tuxerrante:
/remove-lifecycle stale

Yes, we can!

PR #8425, opened 2026-05-05
ARO-26557: Fix etcd bootstrap race when replicas join after quorum

patriksuba:
Parallel etcd pods and delayed PVCs allow a majority to bootstrap before every replica starts. A late replica with an empty data dir must use member add and existing initial-cluster-state instead of new, or it can form a divergent cluster identity.

Change the reset-member init to detect that case (member list shows at least quorum-min members but not this hostname) and add the member before starting etcd.

Fixes [ARO-26557](https://redhat.atlassian.net/browse/ARO-26557)

## Checklist:

## Summary by CodeRabbit

* **Improvements**
  * Init logic for etcd membership now exposes expected-member and quorum values to startup; pods will attempt dynamic joins only when observed members meet the quorum and are below the expected size, refuse to grow past expected size, and log or skip attempts when quorum/env vars are insufficient.

* **Tests**
  * Added unit tests validating quorum-size calculation and quorum-check behavior for dynamic join decisions.


patriksuba:
Closing in favor of [OCPBUGS-109628](https://redhat.atlassian.net/browse/OCPBUGS-109628)

PR #8427, opened 2026-05-05
CNTRLPLANE-3383: HO: Add HostedClusterDeleting condition to track deletion progress

csrwng:
## Summary

- Add new `HostedClusterDeleting` condition type with phase-specific reasons (`WaitingForNodePoolDeletion`, `WaitingForCAPIClusterDeletion`, `WaitingForEndpointServiceDeletion`, `WaitingForPrivateConnectDeletion`, `WaitingForControlPlaneDeletion`, `WaitingForNamespaceDeletion`, `DeletionCompleted`)
- Initialize the condition to `False/AsExpected` during reconciliation if not present
- Update the condition at each deletion phase in the HO `delete()` method via a `setDeletionProgress` helper
- Enrich namespace deletion reporting by inspecting `NamespaceContentRemaining`, `NamespaceFinalizersRemaining`, and `NamespaceDeletionContentFailure` conditions

## Context

HostedCluster deletion progress is currently only visible through HO log messages. This makes it difficult for users, OCM, and SRE tooling to determine which phase of deletion a cluster is in or detect stuck deletions without log access.

Identified during a debugging session investigating cluster deletion issues in the managed Azure service (ARO HCP).

## Jira

https://issues.redhat.com/browse/CNTRLPLANE-3383

## Test plan

- [ ] Verify `HostedClusterDeleting` condition is initialized to `False/AsExpected` on clusters without it
- [ ] Verify condition transitions to `True` with phase-appropriate reason during deletion
- [ ] Verify namespace deletion condition message includes phase and blocking details
- [ ] Verify condition is only updated when reason or message changes
- [ ] `make test` passes
- [ ] `make build` passes

## Summary by CodeRabbit

* **New Features**
  * Granular HostedCluster deletion status with phase-specific progress (node pools, CAPI/control plane, endpoint service, private connect, namespace) and an explicit "deletion completed" state.
  * HostedClusters now explicitly report a steady "not being deleted" status when applicable.

* **Bug Fixes**
  * Deletion-status updates are idempotent and avoid redundant status churn.

* **Tests**
  * New tests for steady-state deletion condition, idempotency, per-phase reporting, namespace blocking details, and final completion.

### DIFF
--- a/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
+++ b/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
@@ -24,6 +24,7 @@ import (
 	"net/netip"
 	"os"
 	"reflect"
+	"sort"
 	"strconv"
 	"strings"
 	"time"
@@ -497,6 +498,14 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
 		}
 	}
 
+	// Keep the steady-state non-deleting condition current.
+	if hcluster.DeletionTimestamp.IsZero() {
+		if err := r.reconcileDeletingConditionSteadyState(ctx, hcluster); err != nil {
+			return ctrl.Result{}, err
+		}
+	}
+
+	// If deleted, clean up and return early.
 	if !hcluster.DeletionTimestamp.IsZero() {
 		// This new condition is necessary for OCM personnel to report any cloud dangling objects to the user.
 		// The grace period is customizable using an annotation called HCDestroyGracePeriodAnnotation. It's a time.Duration annotation.
@@ -3736,10 +3745,38 @@ func deleteControlPlaneOperatorRBAC(ctx context.Context, c client.Client, rbacNa
 	return nil
 }
 
+func (r *HostedClusterReconciler) reconcileDeletingConditionSteadyState(ctx context.Context, hc *hyperv1.HostedCluster) error {
+	if updated := meta.SetStatusCondition(&hc.Status.Conditions, metav1.Condition{
+		Type:               string(hyperv1.HostedClusterDeleting),
+		Status:             metav1.ConditionFalse,
+		Reason:             hyperv1.AsExpectedReason,
+		Message:            "HostedCluster is not being deleted",
+		ObservedGeneration: hc.Generation,
+	}); updated {
+		if err := r.Client.Status().Update(ctx, hc); err != nil {
+			return fmt.Errorf("failed to reconcile HostedClusterDeleting condition: %w", err)
+		}
+	}
+	return nil
+}
+
 //nolint:gocyclo
 func (r *HostedClusterReconciler) delete(ctx context.Context, hc *hyperv1.HostedCluster) (bool, error) {
 	controlPlaneNamespace := manifests.HostedControlPlaneNamespace(hc.Namespace, hc.Name)
 	log := ctrl.LoggerFrom(ctx)
+
+	setDeletionProgress := func(reason, message string) error {
+		if updated := meta.SetStatusCondition(&hc.Status.Conditions, metav1.Condition{
+			Type:               string(hyperv1.HostedClusterDeleting),
+			Status:             metav1.ConditionTrue,
+			Reason:             reason,
+			Message:            message,
+			ObservedGeneration: hc.Generation,
+		}); !updated {
+			return nil
+		}
+		return r.Client.Status().Update(ctx, hc)
+	}
 
 	// Unpause CAPI cluster to allow deletion to proceed
 	if err := pauseCAPICluster(ctx, r.Client, hc, false); err != nil {
@@ -3774,6 +3811,24 @@ func (r *HostedClusterReconciler) delete(ctx context.Context, hc *hyperv1.Hosted
 		return false, err
 	}
 
+	remainingNodePools, err := listNodePools(ctx, r.Client, hc.Namespace, hc.Name)
+	if err != nil {
+		return false, fmt.Errorf("failed to list NodePools: %w", err)
+	}
+	if len(remainingNodePools) > 0 {
+		npNames := make([]string, len(remainingNodePools))
+		for i := range remainingNodePools {
+			npNames[i] = remainingNodePools[i].Name
+		}
+		sort.Strings(npNames)
+		log.Info("Waiting for NodePool deletion", "remaining", len(remainingNodePools), "nodepools", npNames)
+		if err := setDeletionProgress(hyperv1.DeletionWaitingForNodePoolDeletionReason,
+			fmt.Sprintf("Waiting for %d NodePool(s) to be deleted: %s", len(remainingNodePools), strings.Join(npNames, ", "))); err != nil {
+			return false, fmt.Errorf("failed to update deletion progress: %w", err)
+		}
+		return false, nil
+	}
+
 	p, err := platform.GetPlatform(ctx, hc, nil, "", nil)
 	if err != nil {
 		return false, err
@@ -3797,6 +3852,10 @@ func (r *HostedClusterReconciler) delete(ctx context.Context, hc *hyperv1.Hosted
 
 		if exists {
 			log.Info("Waiting for cluster deletion", "clusterName", hc.Spec.InfraID, "controlPlaneNamespace", controlPlaneNamespace)
+			if err := setDeletionProgress(hyperv1.DeletionWaitingForCAPIClusterDeletionReason,
+				fmt.Sprintf("Waiting for CAPI cluster %s/%s to be deleted", controlPlaneNamespace, hc.Spec.InfraID)); err != nil {
+				return false, fmt.Errorf("failed to update deletion progress: %w", err)
+			}
 			return false, nil
 		} else {
 			// once infra is deleted remove finalizers.
@@ -3836,6 +3895,10 @@ func (r *HostedClusterReconciler) delete(ctx context.Context, hc *hyperv1.Hosted
 		}
 		if exists {
 			log.Info("Waiting for awsendpointservice deletion", "controlPlaneNamespace", controlPlaneNamespace)
+			if err := setDeletionProgress(hyperv1.DeletionWaitingForEndpointServiceDeletionReason,
+				fmt.Sprintf("Waiting for AWS endpoint services in %s to be deleted", controlPlaneNamespace)); err != nil {
+				return false, fmt.Errorf("failed to update deletion progress: %w", err)
+			}
 			return false, nil
 		}
 	}
@@ -3847,6 +3910,10 @@ func (r *HostedClusterReconciler) delete(ctx context.Context, hc *hyperv1.Hosted
 		}
 		if exists {
 			log.Info("Waiting for gcpprivateserviceconnect deletion", "controlPlaneNamespace", controlPlaneNamespace)
+			if err := setDeletionProgress(hyperv1.DeletionWaitingForPrivateConnectDeletionReason,
+				fmt.Sprintf("Waiting for GCP Private Service Connect resources in %s to be deleted", controlPlaneNamespace)); err != nil {
+				return false, fmt.Errorf("failed to update deletion progress: %w", err)
+			}
 			return false, nil
 		}
 	}
@@ -3902,6 +3969,10 @@ func (r *HostedClusterReconciler) delete(ctx context.Context, hc *hyperv1.Hosted
 	}
 	if exists {
 		log.Info("Waiting for hostedcontrolplane deletion", "controlPlaneNamespace", controlPlaneNamespace)
+		if err := setDeletionProgress(hyperv1.DeletionWaitingForControlPlaneDeletionReason,
+			fmt.Sprintf("Waiting for HostedControlPlane %s/%s to be deleted", controlPlaneNamespace, hc.Name)); err != nil {
+			return false, fmt.Errorf("failed to update deletion progress: %w", err)
+		}
 		return false, nil
 	}
 
@@ -3916,6 +3987,9 @@ func (r *HostedClusterReconciler) delete(ctx context.Context, hc *hyperv1.Hosted
 	r.KubevirtInfraClients.Delete(hc.Spec.InfraID)
 
 	if skipNSDeletion := hc.Annotations[hyperv1.SkipControlPlaneNamespaceDeletionAnnotation]; skipNSDeletion == "true" {
+		if err := setDeletionProgress(hyperv1.DeletionCompletedReason, "Deletion completed (namespace deletion skipped)"); err != nil {
+			return false, fmt.Errorf("failed to update deletion progress: %w", err)
+		}
 		return true, nil
 	}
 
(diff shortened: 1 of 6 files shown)


hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -491,6 +491,20 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
...
+			Type:               string(hyperv1.HostedClusterDeleting),
+			Status:             metav1.ConditionFalse,
+			Reason:             hyperv1.AsExpectedReason,
+			Message:            "HostedCluster is not being deleted",
+			ObservedGeneration: hcluster.Generation,
+		})
+		if err := r.Client.Status().Update(ctx, hcluster); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to initialize HostedClusterDeleting condition: %w", err)
+		}
+	}
```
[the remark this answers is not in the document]
csrwng: Fixed in 7c3dde10c. The condition is now reconciled continuously on every non-deleting reconcile, gated on `DeletionTimestamp.IsZero()`. The update is skipped when status, reason, message, and ObservedGeneration all match the desired steady-state value.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -3394,6 +3424,10 @@ func (r *HostedClusterReconciler) delete(ctx context.Context, hc *hyperv1.Hosted
 
 		if exists {
 			log.Info("Waiting for cluster deletion", "clusterName", hc.Spec.InfraID, "controlPlaneNamespace", controlPlaneNamespace)
+			if err := setDeletionProgress(hyperv1.DeletionWaitingForCAPIClusterDeletionReason,
+				fmt.Sprintf("Waiting for CAPI cluster %s/%s to be deleted", controlPlaneNamespace, hc.Spec.InfraID)); err != nil {
+				return false, fmt.Errorf("failed to update deletion progress: %w", err)
+			}
```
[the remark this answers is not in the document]
csrwng: Fixed in 7c3dde10c. After `deleteNodePools`, the code now calls `listNodePools` and if any remain (still terminating with finalizers), sets `WaitingForNodePoolDeletion` with a count message and returns early. The CAPI cluster phase is only entered once all NodePools are gone. Also added unit tests covering both the NodePool teardown phase and the steady-state condition reconciliation.

jparrill (commented):

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -3371,6 +3409,19 @@ func (r *HostedClusterReconciler) delete(ctx context.Context, hc *hyperv1.Hosted
 		return false, err
 	}
 
+	remainingNodePools, err := listNodePools(ctx, r.Client, hc.Namespace, hc.Name)
+	if err != nil {
+		return false, fmt.Errorf("failed to list NodePools: %w", err)
+	}
+	if len(remainingNodePools) > 0 {
```
jparrill, line 3818: **Behavioral change beyond observability**: This introduces a new blocking gate — `delete()` now waits for all NodePools to be fully removed before proceeding to CAPI cluster deletion. Previously, `deleteNodePools()` triggered deletion and continued immediately.

This is arguably a correctness improvement (the old flow was racing), but:
- Please call this out explicitly in the PR description — it's not just observability.
- If a NodePool has a stuck finalizer, deletion now hangs indefinitely at `WaitingForNodePoolDeletion`. Consider listing the stuck NodePool names in the message (not just count) for faster SRE debugging.
- Consider a follow-up for timeout/degradation after extended waits.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -3338,6 +3360,22 @@ func (r *HostedClusterReconciler) delete(ctx context.Context, hc *hyperv1.Hosted
...
+	setDeletionProgress := func(reason, message string) error {
+		condition := metav1.Condition{
+			Type:               string(hyperv1.HostedClusterDeleting),
+			Status:             metav1.ConditionTrue,
+			Reason:             reason,
+			Message:            message,
+			ObservedGeneration: hc.Generation,
+		}
+		old := meta.FindStatusCondition(hc.Status.Conditions, string(hyperv1.HostedClusterDeleting))
+		if old != nil && old.Reason == condition.Reason && old.Message == condition.Message {
```
jparrill: nit: The idempotency check here only compares `Reason` and `Message`, while the equivalent check in `reconcile()` (~line 504) compares 4 fields (`Status`, `Reason`, `Message`, `ObservedGeneration`). If `hc.Generation` changes during deletion (unlikely but possible), this will skip updating `ObservedGeneration`.

Consider using `meta.SetStatusCondition` return value pattern (already used at ~line 410 for `ValidAWSIdentityProvider`) to unify both paths and avoid subtle divergence:
```go
if updated := meta.SetStatusCondition(&hcluster.Status.Conditions, desired); updated {
    if err := r.Client.Status().Update(ctx, hcluster); err != nil { ... }
}
```

hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
```diff
@@ -6569,3 +6570,347 @@ func TestComputeEndpointServiceCondition(t *testing.T) {
...
+				},
+			}
+
+			fakeClient := fake.NewClientBuilder().
+				WithScheme(api.Scheme).
+				WithObjects(hc).
+				WithStatusSubresource(hc).
+				Build()
+
+			// Directly exercise the condition-reconciliation logic from reconcile().
```
jparrill: This test reconstructs the condition-building and comparison logic inline rather than calling the actual production code path. If `reconcile()` changes (e.g., a new field is added to the comparison, or the block is accidentally removed), this test will still pass because it tests a copy, not the real path.

Suggestion: Extract the steady-state condition logic into a named method (e.g., `reconcileDeletingConditionSteadyState(ctx, hc)`) that can be unit-tested independently. Or accept as pragmatic for v1 and track a follow-up to improve test fidelity.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -3497,10 +3567,45 @@ func (r *HostedClusterReconciler) delete(ctx context.Context, hc *hyperv1.Hosted
...
+		}
+
+		log.Info(message, "controlPlaneNamespace", controlPlaneNamespace)
+		if err := setDeletionProgress(hyperv1.DeletionWaitingForNamespaceDeletionReason, message); err != nil {
+			return false, fmt.Errorf("failed to update deletion progress: %w", err)
+		}
 		return false, nil
 	}
 
+	if err := setDeletionProgress(hyperv1.DeletionCompletedReason, "Deletion completed"); err != nil {
```
jparrill, line 4045: nit: `DeletionCompleted` is set just before `delete()` returns `true`, and the caller immediately removes the finalizer → object is GC'd. Without grace period, this condition exists for milliseconds and may never be observed by external consumers (OCM, SRE dashboards). With the grace period annotation, "Deletion completed" is misleading since the object is still around.

Worth a one-liner in the godoc clarifying that "completed" means resource teardown, not full object removal.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -3497,10 +3567,45 @@ func (r *HostedClusterReconciler) delete(ctx context.Context, hc *hyperv1.Hosted
...
+							"conditionType", cond.Type,
+							"reason", cond.Reason,
+							"message", cond.Message,
+						)
+					}
+				}
+			}
+			if len(details) > 0 {
+				message = fmt.Sprintf("Waiting for namespace %s to be deleted (phase: %s): %s",
+					controlPlaneNamespace, ns.Status.Phase, strings.Join(details, "; "))
```
jparrill, line 4030: Two things here:
1. The message concatenates all blocking namespace conditions via `strings.Join(details, "; ")`. In pathological cases this could produce a very long condition message. Consider truncating to ~1024 chars.
2. This is the most complex new logic path (fetching namespace, inspecting 3 condition types, assembling diagnostic messages) and has no test coverage — that's likely contributing to the `codecov/patch` failure. Edge cases to consider: multiple blocking conditions, conditions with `Status: False`, namespace Get failures.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
```diff
@@ -50,6 +50,7 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/apimachinery/pkg/api/meta"
```
jparrill: nit: Verify this import ordering passes `gci` — `lint` and `verify` checks are currently failing and this could be the cause (`meta` sorts after `errors` alphabetically but appears before `errors2` here).

hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
```diff
@@ -6569,3 +6570,347 @@ func TestComputeEndpointServiceCondition(t *testing.T) {
...
+				current.Reason != desired.Reason ||
+				current.Message != desired.Message ||
+				current.ObservedGeneration != desired.ObservedGeneration
+
+			g.Expect(needsUpdate).To(Equal(tc.expectStatusUpdate))
+
+			if needsUpdate {
+				meta.SetStatusCondition(&hc.Status.Conditions, desired)
+				g.Expect(fakeClient.Status().Update(ctx, hc)).To(Succeed())
+			}
```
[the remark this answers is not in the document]
csrwng: Addressed — extracted the steady-state condition logic into a named `reconcileDeletingConditionSteadyState()` method on the reconciler. The test now calls that production method directly instead of reconstructing the logic inline.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
```diff
@@ -6569,3 +6570,347 @@ func TestComputeEndpointServiceCondition(t *testing.T) {
...
+				g.Expect(cond.ObservedGeneration).To(Equal(int64(1)))
+			} else {
+				// No NodePools case: it proceeds past the NodePool phase into later
+				// deletion stages. We verify it did not set WaitingForNodePoolDeletion.
+				updatedHC := &hyperv1.HostedCluster{}
+				g.Expect(fakeClient.Get(ctx, types.NamespacedName{Namespace: "clusters", Name: "test-cluster"}, updatedHC)).To(Succeed())
+				cond := meta.FindStatusCondition(updatedHC.Status.Conditions, string(hyperv1.HostedClusterDeleting))
+				if cond != nil {
+					g.Expect(cond.Reason).ToNot(Equal(hyperv1.DeletionWaitingForNodePoolDeletionReason))
+				}
```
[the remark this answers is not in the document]
csrwng: Fixed — added `g.Expect(err).ToNot(HaveOccurred())` assertion in the no-NodePool branch.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
csrwng: Good catch. The deletion message now lists stuck NodePool names (not just count) for faster SRE debugging, e.g. `"Waiting for 2 NodePool(s) to be deleted: np-1, np-2"`. Will call out the behavioral change in the PR description. Timeout/degradation tracking is a good follow-up idea — will track separately.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
csrwng: Fixed — both `setDeletionProgress` and `reconcileDeletingConditionSteadyState` now use the `meta.SetStatusCondition` return value pattern, consistent with the `ValidAWSIdentityProvider` block at ~line 410. No more manual field comparison divergence.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
csrwng: Addressed — extracted the logic into a named `reconcileDeletingConditionSteadyState(ctx, hc)` method. The test now calls the production method directly and also validates ResourceVersion to confirm whether a status update actually occurred.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
csrwng: Added godoc on `DeletionCompletedReason` clarifying that it indicates all hosted cluster resources have been torn down, and the HostedCluster object itself may still exist briefly until the finalizer is removed.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
csrwng: Both addressed:
1. Added truncation to 1024 chars for the namespace condition message.
2. Added `TestDeleteNamespaceTeardownPhase` with two sub-tests: one with blocking conditions (`NamespaceContentRemaining` + `NamespaceFinalizersRemaining`) that verifies they appear in the message, and one without blocking conditions that verifies only the phase is reported.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
csrwng: Fixed — swapped `meta` and `errors2` into correct alphabetical order. Verified `make lint-fix` passes with 0 issues.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -3368,6 +3403,23 @@ func (r *HostedClusterReconciler) delete(ctx context.Context, hc *hyperv1.Hosted
...
+		return false, fmt.Errorf("failed to list NodePools: %w", err)
+	}
+	if len(remainingNodePools) > 0 {
+		npNames := make([]string, len(remainingNodePools))
+		for i := range remainingNodePools {
+			npNames[i] = remainingNodePools[i].Name
+		}
+		log.Info("Waiting for NodePool deletion", "remaining", len(remainingNodePools), "nodepools", npNames)
+		if err := setDeletionProgress(hyperv1.DeletionWaitingForNodePoolDeletionReason,
+			fmt.Sprintf("Waiting for %d NodePool(s) to be deleted: %s", len(remainingNodePools), strings.Join(npNames, ", "))); err != nil {
```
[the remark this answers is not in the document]
csrwng: Fixed — added `sort.Strings(npNames)` before composing the log and condition message to ensure deterministic ordering across reconciles.

jparrill (approved):
All 6 review comments from May 11 have been addressed. Clean implementation. /lgtm

clebs:
@csrwng this LGTM overall, I will tag once unit test issue is fixed and PR is rebased.

PR #8436, opened 2026-05-06
OCPBUGS-84955: Add render-sensitive flag to `hypershift install render` 

clebs:

Introduce the --render-sensitive flag to the hypershift install render command to avoid leaking secrets.
This is now needed because the new webhookcerts controller introduces secrets that should not be rendered by default.

Fixes [OCPBUGS-84955](https://redhat.atlassian.net/browse/OCPBUGS-84955)

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Added a --render-sensitive option to control whether sensitive resources (e.g., Secrets) are emitted when rendering manifests; sensitive objects are excluded by default.
  * Template rendering now requires explicit --render-sensitive to include sensitive data; omitting it returns a clear error.

* **Tests**
  * Added tests validating inclusion/exclusion of sensitive resources and the template-without-flag failure.

* **Chores**
  * Exposed a public option to enable render-sensitive behavior.

### DIFF
--- a/cmd/install/install.go
+++ b/cmd/install/install.go
@@ -155,6 +155,7 @@ type Options struct {
 	ScaleFromZeroCreds                        string
 	ScaleFromZeroCredentialsSecret            string
 	ScaleFromZeroCredentialsSecretKey         string
+	RenderSensitive                           bool
 }
 
 func (o *Options) Validate() error {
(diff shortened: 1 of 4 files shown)


patjlm (approved):

Nirshal (approved):
LGTM - clean change, safe default, good test coverage.

PR #8439, opened 2026-05-06
OCPBUGS-77856: fix: use NodePort for HCP router Service on non-cloud platforms

vsolanki12:

On non-cloud platforms (Agent, KubeVirt, OpenStack, None), the HCP router Service was unconditionally created as LoadBalancer. Since these platforms lack a cloud LB controller, the Service stays pending forever, blocking InfrastructureReady and preventing the hosted control plane from becoming available.

Three code paths are fixed:
- `ReconcileRouterService`: use NodePort for non-cloud platforms instead of unconditional LoadBalancer
- `reconcileRouterServiceStatus`: add NodePort guard before calling `CollectLBMessageIfNotProvisioned`
- `kas.ReconcileServiceStatus`: add NodePort guard in the Route case

Fixes [OCPBUGS-77856](https://issues.redhat.com/browse/OCPBUGS-77856)

- PowerVS is intentionally left as LoadBalancer since it has cloud LB support via IBM Cloud VPC.
- `LoadBalancerSourceRanges` on NodePort services is harmless — Kubernetes ignores this field for non-LoadBalancer types.
- The `etcd-upload/s3_uploader_mock.go` build failure in CI is a pre-existing issue on main, unrelated to this PR.

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Router service now defaults to NodePort on non-cloud platforms and LoadBalancer on cloud platforms, and preserves an explicitly set service type.
  * CIDR source-range restrictions are applied only for external services that resolve to LoadBalancer.
  * Route publishing correctly returns the service ClusterIP/port for NodePort-backed services and empty/zero when unassigned.

* **Tests**
  * Added tests for platform-dependent NodePort vs LoadBalancer behavior and NodePort route publishing scenarios.

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/infra/infra.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/infra/infra.go
@@ -698,7 +698,8 @@ func (r *Reconciler) reconcileExternalRouterServiceStatus(ctx context.Context, h
 	return r.reconcileRouterServiceStatus(ctx, manifests.RouterPublicService(hcp.Namespace), events.NewMessageCollector(ctx, r.Client))
 }
 
-func (r *Reconciler) reconcileRouterServiceStatus(ctx context.Context, svc *corev1.Service, messageCollector events.MessageCollector) (host string, needed bool, message string, err error) {
+func (r *Reconciler) reconcileRouterServiceStatus(ctx context.Context, svc *corev1.Service, messageCollector events.MessageCollector) (host string,
+	needed bool, message string, err error) {
 	needed = true
 	if err = r.Client.Get(ctx, client.ObjectKeyFromObject(svc), svc); err != nil {
 		if apierrors.IsNotFound(err) {
@@ -708,6 +709,14 @@ func (r *Reconciler) reconcileRouterServiceStatus(ctx context.Context, svc *core
 		err = fmt.Errorf("failed to get router service (%s): %w", svc.Name, err)
 		return
 	}
+
+	if svc.Spec.Type == corev1.ServiceTypeNodePort {
+		if len(svc.Spec.Ports) > 0 && svc.Spec.Ports[0].NodePort != 0 {
+			host = svc.Spec.ClusterIP
+		}
+		return
+	}
+
 	if message, err = k8sutil.CollectLBMessageIfNotProvisioned(svc, messageCollector); err != nil || message != "" {
 		return
 	}
--- a/control-plane-operator/controllers/hostedcontrolplane/ingress/router.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/ingress/router.go
@@ -64,7 +64,14 @@ func ReconcileRouterService(svc *corev1.Service, internal, crossZoneLoadBalancin
 	for k, v := range hcpRouterLabels() {
 		svc.Labels[k] = v
 	}
-	svc.Spec.Type = corev1.ServiceTypeLoadBalancer
+	if svc.Spec.Type == "" {
+		switch hcp.Spec.Platform.Type {
+		case hyperv1.AgentPlatform, hyperv1.KubevirtPlatform, hyperv1.NonePlatform:
+			svc.Spec.Type = corev1.ServiceTypeNodePort
+		default:
+			svc.Spec.Type = corev1.ServiceTypeLoadBalancer
+		}
+	}
 	svc.Spec.Selector = hcpRouterLabels()
 	foundHTTPS := false
 
@@ -89,7 +96,7 @@ func ReconcileRouterService(svc *corev1.Service, internal, crossZoneLoadBalancin
 	// Apply LoadBalancerSourceRanges for external router services to restrict CIDR access
 	// Only apply for external (non-internal) services and when not running on ARO HCP
 	allowedCIDRBlocks := netutil.AllowedCIDRBlocks(hcp)
-	if !internal && !azureutil.IsAroHCP() {
+	if !internal && !azureutil.IsAroHCP() && svc.Spec.Type == corev1.ServiceTypeLoadBalancer {
 		svc.Spec.LoadBalancerSourceRanges = allowedCIDRBlocks
 	}
 
--- a/control-plane-operator/controllers/hostedcontrolplane/kas/service.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/kas/service.go
@@ -174,10 +174,21 @@ func ReconcileServiceStatus(svc *corev1.Service, strategy *hyperv1.ServicePublis
 		port = svc.Spec.Ports[0].NodePort
 		host = strategy.NodePort.Address
 	case hyperv1.Route:
+		if svc.Spec.Type == corev1.ServiceTypeNodePort {
+			if len(svc.Spec.Ports) > 0 && svc.Spec.Ports[0].NodePort != 0 {
+				if strategy.Route != nil {
+					host = strategy.Route.Hostname
+					port = 443
+				}
+			}
+			return
+		}
 		if message, err := k8sutil.CollectLBMessageIfNotProvisioned(svc, messageCollector); err != nil || message != "" {
 			return host, port, message, err
 		}
-		host = strategy.Route.Hostname
+		if strategy.Route != nil {
+			host = strategy.Route.Hostname
+		}
 		port = 443
 	}
 	return
--- a/control-plane-operator/controllers/hostedcontrolplane/infra/infra_test.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/infra/infra_test.go
@@ -1738,6 +1738,41 @@ func TestReconcileRouterServiceStatus(t *testing.T) {
 			},
 			expectedHost: "1.2.3.4",
 		},
+		{
+			name: "When NodePort service with assigned port it should return ClusterIP as host",
+			svc: &corev1.Service{
+				ObjectMeta: metav1.ObjectMeta{Name: svcName, Namespace: namespace},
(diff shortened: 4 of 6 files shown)


control-plane-operator/controllers/hostedcontrolplane/kas/service.go
```diff
@@ -167,6 +167,13 @@ func ReconcileServiceStatus(svc *corev1.Service, strategy *hyperv1.ServicePublis
 		port = svc.Spec.Ports[0].NodePort
 		host = strategy.NodePort.Address
 	case hyperv1.Route:
+		if svc.Spec.Type == corev1.ServiceTypeNodePort {
+			if len(svc.Spec.Ports) > 0 && svc.Spec.Ports[0].NodePort != 0 {
+				host = strategy.Route.Hostname
+				port = 443
```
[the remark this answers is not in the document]
vsolanki12: How about the other factors, example Line 180 (host = strategy.Route.Hostname) is the existing code — it also accesses strategy.Route.Hostname without a nil check. If we add a nil guard  on line 172 but the existing line 180 doesn't have one, we'd be inconsistent, and the existing code would still panic in that scenario.
In practice, when strategy.Type == hyperv1.Route, strategy.Route should always be populated — that's an API contract. Adding a nil check only in our new block while the original code doesn't have one doesn't add real safety.

what do you think in this prospective as well.

jparrill:
Please @vsolanki12 do you mind address the reviews from Coderrabit bot?

control-plane-operator/controllers/hostedcontrolplane/kas/service.go
```diff
@@ -167,6 +167,13 @@ func ReconcileServiceStatus(svc *corev1.Service, strategy *hyperv1.ServicePublis
 		port = svc.Spec.Ports[0].NodePort
 		host = strategy.NodePort.Address
 	case hyperv1.Route:
+		if svc.Spec.Type == corev1.ServiceTypeNodePort {
+			if len(svc.Spec.Ports) > 0 && svc.Spec.Ports[0].NodePort != 0 {
+				host = strategy.Route.Hostname
```
jparrill: Hey, `strategy.Route` is an optional pointer — if it's ever nil here we'd get a panic. I noticed that `ReconcileKonnectivityServerServiceStatus` in this same file does check for nil before accessing `.Hostname`. Might be worth adding the same guard here just to be safe:
```go
if strategy.Route != nil {
    host = strategy.Route.Hostname
    port = 443
}
```
(line 180 below has the same pre-existing issue, but that's a separate thing)

control-plane-operator/controllers/hostedcontrolplane/ingress/router.go
```diff
@@ -64,7 +64,12 @@ func ReconcileRouterService(svc *corev1.Service, internal, crossZoneLoadBalancin
 	for k, v := range hcpRouterLabels() {
 		svc.Labels[k] = v
 	}
-	svc.Spec.Type = corev1.ServiceTypeLoadBalancer
+	switch hcp.Spec.Platform.Type {
+	case hyperv1.AgentPlatform, hyperv1.KubevirtPlatform, hyperv1.OpenStackPlatform, hyperv1.NonePlatform:
+		svc.Spec.Type = corev1.ServiceTypeNodePort
```
jparrill: I might be wrong here, but doesn't OpenStack have LB support through Octavia? If there are existing HyperShift-on-OpenStack deployments where the LB actually works today, switching them to NodePort would break things.

Have you checked with the OpenStack folks whether their management clusters typically have a cloud-controller-manager configured? If they do, we might want to leave OpenStack out of this list.

Also, I see the same four-platform grouping in `support/netutil/visibility.go:LabelHCPRoutes()`. Not necessarily for this PR, but eventually it could be worth extracting a shared helper so the two lists don't drift apart when someone adds a new platform.

- [Slack thread](https://redhat-external.slack.com/archives/C01C8502FMM/p1778574422307059)

control-plane-operator/controllers/hostedcontrolplane/ingress/router.go
```diff
@@ -64,7 +64,12 @@ func ReconcileRouterService(svc *corev1.Service, internal, crossZoneLoadBalancin
...
+	default:
+		svc.Spec.Type = corev1.ServiceTypeLoadBalancer
+	}
 	svc.Spec.Selector = hcpRouterLabels()
 	foundHTTPS := false
```
jparrill, line 77: Small thing — now that some services will be NodePort, `LoadBalancerSourceRanges` (line 98) gets set on them too. Kubernetes ignores it for non-LB types so it won't break anything, but it's a bit confusing if someone inspects the service. Maybe worth gating this on `svc.Spec.Type == corev1.ServiceTypeLoadBalancer`?

jparrill:

control-plane-operator/controllers/hostedcontrolplane/ingress/router.go
mandre: OpenStack indeed supports LB through Octavia, but because OpenStack being OpenStack, you have no guarantee the service is enabled. This shouldn't be a problem for RH OpenStack, where Octavia is generally enabled.
So, perhaps the logic for OpenStack should be updated to dynamically check for an Octavia endoint?

control-plane-operator/controllers/hostedcontrolplane/kas/service.go
vsolanki12: Thank you, Added the nil check on both line 172 and line 180 as per the suggestion.

control-plane-operator/controllers/hostedcontrolplane/ingress/router.go
vsolanki12: Thank you @jparrill @mandre for highlighting this. Currently, I have removed OpenStack from the NodePort list. it will keep using LoadBalancer as before. Since Octavia availability varies across OpenStack deployments. Might be it need a separate PR since it needs runtime infrastructure checks.

control-plane-operator/controllers/hostedcontrolplane/ingress/router.go
vsolanki12: Thanks for highlighting this, I have made the changes as per suggestion.

control-plane-operator/controllers/hostedcontrolplane/ingress/router.go
```diff
@@ -64,7 +64,12 @@ func ReconcileRouterService(svc *corev1.Service, internal, crossZoneLoadBalancin
 	for k, v := range hcpRouterLabels() {
 		svc.Labels[k] = v
 	}
-	svc.Spec.Type = corev1.ServiceTypeLoadBalancer
+	switch hcp.Spec.Platform.Type {
+	case hyperv1.AgentPlatform, hyperv1.KubevirtPlatform, hyperv1.NonePlatform:
+		svc.Spec.Type = corev1.ServiceTypeNodePort
+	default:
+		svc.Spec.Type = corev1.ServiceTypeLoadBalancer
```
jparrill: nit: this platform list will need updating if a new non-cloud platform is added. Not blocking, just noting it's not centralized with `LabelHCPRoutes()` in `support/netutil/visibility.go`.

control-plane-operator/controllers/hostedcontrolplane/ingress/router.go
```diff
@@ -64,7 +64,12 @@ func ReconcileRouterService(svc *corev1.Service, internal, crossZoneLoadBalancin
 	for k, v := range hcpRouterLabels() {
 		svc.Labels[k] = v
 	}
-	svc.Spec.Type = corev1.ServiceTypeLoadBalancer
+	switch hcp.Spec.Platform.Type {
```
csrwng: My main concern with this change is that we will change the Service type for an existing HostedCluster. That could be unexpected on an upgrade. My preference would be to set the service type if creating the service. If an existing service, leave the type alone.

vsolanki12: Thank you @csrwng , wrapped the type assignment in a `svc.Spec.Type == ""` check so it only applies on initial creation. Existing services will keep their current type on upgrade. Added a test to verify an existing LoadBalancer service on a non-cloud platform stays LoadBalancer after reconcile.

control-plane-operator/controllers/hostedcontrolplane/infra/infra.go
```diff
@@ -708,6 +709,14 @@ func (r *Reconciler) reconcileRouterServiceStatus(ctx context.Context, svc *core
 		err = fmt.Errorf("failed to get router service (%s): %w", svc.Name, err)
 		return
 	}
+
+	if svc.Spec.Type == corev1.ServiceTypeNodePort {
+		if len(svc.Spec.Ports) > 0 && svc.Spec.Ports[0].NodePort != 0 {
+			host = svc.Spec.ClusterIP
```
csrwng, line 715: This doesn't seem right. The host will not be reachable from outside the mgmt cluster. It seems we would need to either use the node ip of where the router pod is running or an external name that the user specifies (for which we don't have a spot in the API)

vsolanki12:
Closing this PR — during review csrwng identified that the platform auto-detection approach has a fundamental limitation: the ClusterIP used as the router host for NodePort services is not externally reachable from outside the management cluster.

A proper solution requires an API change to add `Router` as a new `ServiceType` in `spec.services[]`, allowing operators to specify NodePort with an externally reachable address. This is now tracked in [CNTRLPLANE-3527](https://redhat.atlassian.net/browse/CNTRLPLANE-3527).

The enhancement proposal and a new implementation PR will follow from that Epic.

PR #8445, opened 2026-05-06
GCP-410: fix(gcp): add registry SA WIF binding for image registry GCS access

cblecker:
## Summary

- The image registry **server** pod runs as the `registry` ServiceAccount in `openshift-image-registry`, not as `cluster-image-registry-operator`. Without a WIF binding for `registry`, its projected token exchange fails with `iam.serviceAccounts.getAccessToken denied` when accessing GCS storage.
- Rename `k8sServiceAccount` (singular object) to `k8sServiceAccounts` (array) in `iam-bindings.json` and update the Go struct/loop to support multiple WIF bindings per GSA.
- Add `registry` SA to the `image-registry` entry, matching upstream `openshift/cluster-image-registry-operator`'s `CredentialsRequest` which lists both `cluster-image-registry-operator` and `registry` under `serviceAccountNames`.
- Wrap `createServiceAccount` with the existing `retryWithExponentialBackoff` helper and classify HTTP 429 responses as transient in `isTransientIAMError`, so rate limit errors during infra setup are retried rather than failing immediately. Increase `iamPropagationTimeout` to 120s to give the backoff enough room to succeed after the GCP per-minute SA creation quota resets.

## Test plan

- [x] `go build ./cmd/infra/gcp/...` passes
- [x] `go test ./cmd/infra/gcp/...` passes (new tests verify `image-registry` has 2 K8s SA entries and that 429 is treated as a transient error)
- [x] `go vet ./cmd/infra/gcp/...` passes
- [ ] Create a GCP HCP cluster and verify the image registry pod can access GCS without `getAccessToken denied` errors
- [ ] Verify `e2e-v2-gke` no longer fails intermittently with `Service accounts created per minute per project` 429 errors

## Summary by CodeRabbit

* **New Features**
  * Support for multiple Kubernetes service accounts per GCP service account.

* **Improvements**
  * Increased IAM propagation timeout from 60s to 120s.
  * Enhanced handling of rate limiting in IAM operations.
  * Improved retry logic for service account creation and binding operations.

### DIFF
--- a/cmd/infra/gcp/iam-bindings.json
+++ b/cmd/infra/gcp/iam-bindings.json
@@ -9,10 +9,9 @@
         "roles/compute.networkAdmin",
         "roles/iam.serviceAccountUser"
       ],
-      "k8sServiceAccount": {
-        "namespace": "kube-system",
-        "name": "capi-gcp-controller-manager"
-      }
+      "k8sServiceAccounts": [
+        {"namespace": "kube-system", "name": "capi-gcp-controller-manager"}
+      ]
     },
     {
       "name": "ctrlplane-op",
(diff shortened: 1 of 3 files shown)


cblecker:
/verified later by @cblecker 

Verifying this by the e2e test after it merges

cblecker:
/verified later @cblecker

Verifying this by the e2e test after it merges

cblecker:
/override codecov/patch

The calls to GCP (and other cloud providers) aren't mocked. This is intentional.

PR #8448, opened 2026-05-06
[openshift-4.18] OCPBUGS-80278: Bump google.golang.org/grpc to v1.79.3

germanparente:

fix CVE-2026-33186 openshift4/ose-hypershift-rhel9: gRPC-Go: Authorization bypass due to improper HTTP/2 path validation [openshift-4.18]


jparrill:
/close

Closing in favor of #8515. We cannot update all grpc-go versions to a single specific version — it depends on many factors beyond grpc-go itself (transitive dependencies, Go version compatibility, API stability across modules). The replacement PR addresses the CVE through a more targeted approach.

PR #8449, opened 2026-05-06
[openshift-4.17] OCPBUGS-80103: Bump google.golang.org/grpc to v1.79.3

germanparente:

fix for CVE-2026-33186


jparrill:
/close

Closing in favor of #8516. We cannot update all grpc-go versions to a single specific version — it depends on many factors beyond grpc-go itself (transitive dependencies, Go version compatibility, API stability across modules). The replacement PR addresses the CVE through a more targeted approach.

PR #8458, opened 2026-05-07
OCPBUGS-85243: Set aws-load-balancer-scheme on public HCP router service

typeid:
## Summary

- Set `service.beta.kubernetes.io/aws-load-balancer-scheme: internet-facing` on the public HCP router service so it creates an internet-facing NLB on both OpenShift (in-tree cloud provider) and EKS (AWS Load Balancer Controller) management clusters
- Clean up mutually exclusive annotations (`aws-load-balancer-internal` vs `aws-load-balancer-scheme`) when switching between internal and external

## Problem

The AWS Load Balancer Controller (used on EKS) defaults to `scheme=internal` when `aws-load-balancer-scheme` is not set ([docs](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.4/guide/service/annotations/)). The in-tree AWS cloud provider (used on OpenShift) defaults to `internet-facing` ([docs](https://cloud-provider-aws.sigs.k8s.io/service_controller/)).

For `PublicAndPrivate` clusters using `type: Route` with explicit hostnames, HyperShift deploys a dedicated HCP router with a public LoadBalancer service. Without the scheme annotation, this NLB is created as internal on EKS, making KAS and OAuth unreachable from outside the VPC.

## Test plan

- [x] Existing unit tests updated to verify `internet-facing` annotation on public services
- [x] New tests for external scheme and internal-to-external switching
- [x] Verified internal services do NOT get the scheme annotation
- [x] Deploy PublicAndPrivate HCP on EKS and verify public router NLB is internet-facing

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed AWS load balancer annotation handling so external routers are marked internet-facing, internal routers use the internal-LB annotation, and conflicting scheme annotations are removed during transitions.

* **Tests**
  * Added tests covering external vs internal AWS annotation behavior and mode-switching scenarios; updated fixtures to reflect the expected scheme for public/private routers.

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/ingress/router.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/ingress/router.go
@@ -29,6 +29,15 @@ func ReconcileRouterService(svc *corev1.Service, internal, crossZoneLoadBalancin
 		svc.Annotations["service.beta.kubernetes.io/aws-load-balancer-type"] = "nlb"
 		if internal {
 			svc.Annotations["service.beta.kubernetes.io/aws-load-balancer-internal"] = "true"
+			delete(svc.Annotations, "service.beta.kubernetes.io/aws-load-balancer-scheme")
+		} else {
+			delete(svc.Annotations, "service.beta.kubernetes.io/aws-load-balancer-internal")
+			// The AWS Load Balancer Controller (used on EKS) defaults to scheme=internal:
+			//   https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.4/guide/service/annotations/
+			// The in-tree AWS cloud provider (used on OpenShift) defaults to internet-facing:
+			//   https://cloud-provider-aws.sigs.k8s.io/service_controller/
+			// Set the annotation explicitly so the public router works on both.
+			svc.Annotations["service.beta.kubernetes.io/aws-load-balancer-scheme"] = "internet-facing"
 		}
 		if crossZoneLoadBalancingEnabled {
 			// In-tree AWS cloud provider annotation for cross-zone load balancing (OpenShift management clusters).
--- a/control-plane-operator/controllers/hostedcontrolplane/ingress/router_test.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/ingress/router_test.go
@@ -3,6 +3,8 @@ package ingress
 import (
 	"testing"
 
+	. "github.com/onsi/gomega"
+
 	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
 	"github.com/openshift/hypershift/support/netutil"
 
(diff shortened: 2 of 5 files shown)


control-plane-operator/controllers/hostedcontrolplane/ingress/router.go
```diff
@@ -29,6 +29,11 @@ func ReconcileRouterService(svc *corev1.Service, internal, crossZoneLoadBalancin
 		svc.Annotations["service.beta.kubernetes.io/aws-load-balancer-type"] = "nlb"
 		if internal {
 			svc.Annotations["service.beta.kubernetes.io/aws-load-balancer-internal"] = "true"
+		} else {
+			// The AWS Load Balancer Controller (used on EKS) defaults to scheme=internal.
+			// The in-tree AWS cloud provider (used on OpenShift) defaults to internet-facing.
+			// Set the annotation explicitly so the public router works on both.
+			svc.Annotations["service.beta.kubernetes.io/aws-load-balancer-scheme"] = "internet-facing"
 		}
```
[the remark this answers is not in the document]
typeid: The caller reconcileHCPRouterServices creates two separate Service objects: RouterPublicService (always called with internal=false) and PrivateRouterService (always called with internal=true). A given Service object never flips between internal and external — they're distinct resources with different names. If one is no longer needed, it gets deleted entirely via DeleteIfNeeded, not re-reconciled with the opposite internal flag.
So there's no code path where a single Service would accumulate both aws-load-balancer-internal and aws-load-balancer-scheme annotations from successive reconciliations with different internal values. The stale-annotation concern is theoretical only.

control-plane-operator/controllers/hostedcontrolplane/ingress/router.go
```diff
@@ -29,6 +29,13 @@ func ReconcileRouterService(svc *corev1.Service, internal, crossZoneLoadBalancin
 		svc.Annotations["service.beta.kubernetes.io/aws-load-balancer-type"] = "nlb"
 		if internal {
 			svc.Annotations["service.beta.kubernetes.io/aws-load-balancer-internal"] = "true"
+		} else {
+			// The AWS Load Balancer Controller (used on EKS) defaults to scheme=internal:
+			//   https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.4/guide/service/annotations/
+			// The in-tree AWS cloud provider (used on OpenShift) defaults to internet-facing:
+			//   https://cloud-provider-aws.sigs.k8s.io/service_controller/
+			// Set the annotation explicitly so the public router works on both.
+			svc.Annotations["service.beta.kubernetes.io/aws-load-balancer-scheme"] = "internet-facing"
```
jparrill, line 40: **suggestion (non-blocking):** The PR description mentions "clean up mutually exclusive annotations" but the production code only adds -- it never deletes the conflicting annotation. While the current architecture uses separate Service objects for public/private (so the transition scenario can't happen today), defensive cleanup costs nothing and follows the pattern already used in `oauth/service.go:77`:
```go
if internal {
	svc.Annotations["service.beta.kubernetes.io/aws-load-balancer-internal"] = "true"
	delete(svc.Annotations, "service.beta.kubernetes.io/aws-load-balancer-scheme")
} else {
	svc.Annotations["service.beta.kubernetes.io/aws-load-balancer-scheme"] = "internet-facing"
	delete(svc.Annotations, "service.beta.kubernetes.io/aws-load-balancer-internal")
}
```
This would also make the code match what the `infra_test.go` fixture helper already expects (the `privateService` helper does `delete(s.Annotations, "...scheme")`).

control-plane-operator/controllers/hostedcontrolplane/ingress/router_test.go
```diff
@@ -56,6 +58,25 @@ func TestReconcileRouterServiceAnnotations(t *testing.T) {
 	}
 }
 
+// When reconciling an external (non-internal) AWS router service
+// it should set aws-load-balancer-scheme to internet-facing.
+func TestReconcileRouterService_WhenExternalAWS_ItShouldSetInternetFacingScheme(t *testing.T) {
```
jparrill, line 63: **nit:** Good test -- gomega + Gherkin naming, positive and negative assertions. Consider adding a complementary subtest that pre-populates the conflicting annotation to verify cleanup:
```go
t.Run("When switching from internal to external it should remove the internal annotation", func(t *testing.T) {
	g := NewGomegaWithT(t)
	svc := &corev1.Service{}
	svc.Annotations = map[string]string{
		"service.beta.kubernetes.io/aws-load-balancer-internal": "true",
	}
	err := ReconcileRouterService(svc, false, true, hcp)
	g.Expect(svc.Annotations).To(HaveKeyWithValue(
		"service.beta.kubernetes.io/aws-load-balancer-scheme", "internet-facing"))
	g.Expect(svc.Annotations).ToNot(HaveKey(
		"service.beta.kubernetes.io/aws-load-balancer-internal"))
})
```
This would pair with the defensive `delete()` calls suggested in `router.go`.

jparrill (commented):
Review with two suggestions -- both non-blocking. The fix is correct and safe for merge as-is.

control-plane-operator/controllers/hostedcontrolplane/ingress/router.go
typeid: Makes sense, thanks! Updated.

control-plane-operator/controllers/hostedcontrolplane/ingress/router_test.go
typeid: Added a test case for the switching back and forth.

control-plane-operator/controllers/hostedcontrolplane/ingress/router.go
sdminonne: +1 thanks both

sdminonne (approved):
To be honest I would definitively put in the same test `TestReconcileRouterService_WhenExternalAWS_ItShouldSetInternetFacingScheme` and `TestReconcileRouterService_WhenSwitchingMode_ItShouldRemoveConflictingAnnotation` but I really don't want to block on this

PR #8469, opened 2026-05-08
CNTRLPLANE-3371: Fix AllowedCIDRs e2e test for Route-based KAS

bryan-cox:
## What

Fixes the `ValidateKubeAPIServerAllowedCIDRs` e2e test so it passes on v2 Azure self-managed clusters where KAS uses Route publishing strategy (via `--external-dns-domain`).

## Why

The test was skipped in v2 CI (`--ginkgo.skip="KAS allowed CIDRs"`) because it always failed. Both v1 and v2 Azure self-managed use Route strategy for KAS, but v1 passes while v2 fails due to a difference in cluster lifecycle timing combined with HTTP/2 connection reuse.

### Root cause: HTTP/2 connection reuse

The test reuses a single `kubeclient.Clientset` across all `ServerVersion()` poll iterations. Go's HTTP/2 transport multiplexes all requests over a single persistent TCP connection. If the first poll succeeds before Azure NSG rules take effect, all subsequent polls reuse that connection and never observe the expected failure.

**Why v1 passes but v2 fails**: In v1, the cluster is created fresh inside `TestCreateCluster`, so the CPO is in its initial reconciliation burst — the router service's `LoadBalancerSourceRanges` and corresponding Azure NSG rules are updated before the first `ServerVersion()` call. In v2, the cluster is pre-created and shared across tests, so the CPO is in steady-state with longer reconciliation intervals. The first `ServerVersion()` call succeeds before the NSG rules catch up, and HTTP/2 holds that connection open for all subsequent polls.

### Additional fix: missing downstream service wait

The test waits for `AllowedCIDRBlocks` to propagate from the HostedCluster to the HostedControlPlane, but does not wait for the CPO to reconcile the downstream LoadBalancer service's `LoadBalancerSourceRanges`. This is a race condition that exists in both v1 and v2 — v1 just happens to win the race due to CPO being in active reconciliation. Adding an explicit wait makes the test correct rather than relying on timing.

## Changes

**`test/e2e/util/util.go`** — single file, three changes:

1. **`ensureAPIServerAllowedCIDRs` signature**: `*kubeclient.Clientset` → `*rest.Config` to enable fresh client creation per poll
2. **Fresh kubeclient per poll**: Each `ServerVersion()` iteration creates a new client via `kubeclient.NewForConfig(rest.CopyConfig(guestConfig))`, preventing HTTP/2 connection reuse.
3. **Strategy-aware service wait**: New `allowedCIDRsTargetService()` helper determines the correct LB service based on APIServer publishing strategy (Route → `router`, LoadBalancer → platform-specific KAS LB). An `Eventually` block waits for the service's `LoadBalancerSourceRanges` to match before checking KAS reachability.

## Test Plan

- [x] `go build -tags e2e ./test/e2e/...` — compiles
- [x] `go build -tags e2ev2 ./test/e2e/v2/...` — compiles
- [x] `go vet -tags e2e ./test/e2e/...` — passes
- [ ] Re-run v2 rehearsal on [openshift/release#79048](https://github.com/openshift/release/pull/79048) after merge

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved API server CIDR validation to wait for downstream load‑balancer updates, verify reachability against the appropriate downstream service per publishing strategy/platform, and recreate connections to ensure new network rules are enforced.
  * Made Azure workload‑identity webhook mutation verification more robust by retrying delete/recreate and revalidating pod mutation.

* **Tests**
  * Added tests covering downstream service selection across platforms and publishing strategies and strengthened reachability checks.

* **Chores**
  * Reordered Azure post‑create validation steps for more reliable verification.

### DIFF
--- a/test/e2e/create_cluster_test.go
+++ b/test/e2e/create_cluster_test.go
@@ -116,8 +116,11 @@ func TestCreateCluster(t *testing.T) {
 		e2eutil.EnsureDefaultSecurityGroupTags(t, ctx, mgtClient, hostedCluster, clusterOpts)
 
 		if globalOpts.Platform == hyperv1.AzurePlatform {
-			e2eutil.EnsureKubeAPIServerAllowedCIDRs(t, ctx, mgtClient, guestConfig, hostedCluster)
+			// WI webhook must run before AllowedCIDRs. AllowedCIDRs blocks and restores
+			// all KAS traffic; the webhook sidecar (FailurePolicy: Ignore) may not be
+			// ready during recovery, silently skipping mutation on pods created in that window.
 			e2eutil.EnsureAzureWorkloadIdentityWebhookMutation(t, ctx, guestClient)
+			e2eutil.EnsureKubeAPIServerAllowedCIDRs(t, ctx, mgtClient, guestConfig, hostedCluster)
 		}
 
 		e2eutil.EnsureGlobalPullSecret(t, ctx, mgtClient, hostedCluster, globalOpts.AdditionalPullSecretFile)
--- a/test/e2e/util/azure.go
+++ b/test/e2e/util/azure.go
@@ -10,6 +10,7 @@ import (
 	. "github.com/onsi/gomega"
 
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/utils/ptr"
@@ -38,7 +39,7 @@ func ValidateAzureWorkloadIdentityWebhookMutation(t testing.TB, ctx context.Cont
 	}
 	g.Expect(guestClient.Create(ctx, serviceAccount)).To(Succeed(), "failed to create test service account")
 
-	pod := &corev1.Pod{
+	podTemplate := &corev1.Pod{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "azure-wi-webhook-test-pod",
 			Namespace: nsName,
@@ -70,34 +71,33 @@ func ValidateAzureWorkloadIdentityWebhookMutation(t testing.TB, ctx context.Cont
 			RestartPolicy: corev1.RestartPolicyNever,
 		},
 	}
-	g.Expect(guestClient.Create(ctx, pod)).To(Succeed(), "failed to create pod for webhook mutation test")
 
-	EventuallyObject(
-		t,
-		ctx,
-		"Azure workload identity webhook to mutate test pod",
-		func(ctx context.Context) (*corev1.Pod, error) {
-			mutatedPod := &corev1.Pod{}
-			err := guestClient.Get(ctx, types.NamespacedName{Name: pod.Name, Namespace: pod.Namespace}, mutatedPod)
-			return mutatedPod, err
-		},
-		[]Predicate[*corev1.Pod]{
-			func(mutatedPod *corev1.Pod) (bool, string, error) {
-				if hasProjectedTokenVolume(mutatedPod.Spec.Volumes) {
-					return true, "", nil
-				}
-				return false, "expected projected service account token volume to be injected", nil
-			},
-			func(mutatedPod *corev1.Pod) (bool, string, error) {
-				if hasAzureFederatedTokenEnv(mutatedPod.Spec.Containers) {
-					return true, "", nil
-				}
-				return false, "expected AZURE_FEDERATED_TOKEN_FILE env var in pod containers", nil
-			},
-		},
-		WithTimeout(3*time.Minute),
-		WithInterval(5*time.Second),
-	)
+	// The WI webhook is a MutatingAdmissionWebhook with FailurePolicy: Ignore
+	// that only fires on pod CREATE. If the webhook sidecar isn't ready when
+	// the pod is created, the pod is admitted without mutation and no amount
+	// of GET polling can recover. Delete and recreate each iteration to
+	// re-trigger admission until the webhook is ready.
+	g.Eventually(func(g Gomega) {
+		existing := &corev1.Pod{}
+		err := guestClient.Get(ctx, types.NamespacedName{Name: podTemplate.Name, Namespace: podTemplate.Namespace}, existing)
+		if err == nil {
+			g.Expect(guestClient.Delete(ctx, existing)).To(Succeed(), "failed to delete pod for retry")
+			g.Eventually(func() bool {
+				err := guestClient.Get(ctx, types.NamespacedName{Name: podTemplate.Name, Namespace: podTemplate.Namespace}, &corev1.Pod{})
+				return apierrors.IsNotFound(err)
+			}).WithContext(ctx).WithTimeout(30*time.Second).WithPolling(time.Second).Should(BeTrue(), "pod should be deleted before retry")
+		} else {
+			g.Expect(apierrors.IsNotFound(err)).To(BeTrue(), "unexpected error getting existing pod: %v", err)
+		}
+
+		fresh := podTemplate.DeepCopy()
+		g.Expect(guestClient.Create(ctx, fresh)).To(Succeed(), "failed to create pod for webhook mutation test")
+
+		mutated := &corev1.Pod{}
+		g.Expect(guestClient.Get(ctx, types.NamespacedName{Name: fresh.Name, Namespace: fresh.Namespace}, mutated)).To(Succeed())
+		g.Expect(hasProjectedTokenVolume(mutated.Spec.Volumes)).To(BeTrue(), "expected projected service account token volume to be injected")
+		g.Expect(hasAzureFederatedTokenEnv(mutated.Spec.Containers)).To(BeTrue(), "expected AZURE_FEDERATED_TOKEN_FILE env var in pod containers")
+	}).WithContext(ctx).WithTimeout(3 * time.Minute).WithPolling(10 * time.Second).Should(Succeed())
 }
 
 func EnsureAzureWorkloadIdentityWebhookMutation(t *testing.T, ctx context.Context, guestClient crclient.Client) {
--- a/test/e2e/util/util.go
+++ b/test/e2e/util/util.go
@@ -3602,17 +3602,23 @@ func ValidateKubeAPIServerAllowedCIDRs(t testing.TB, ctx context.Context, mgmtCl
 			}
 		})
 		g.Expect(err).NotTo(HaveOccurred(), "failed to restore HostedCluster API server CIDRs")
+
+		// Verify KAS is reachable on the original transport before returning. The
+		// AllowedCIDRs test uses cfg.Dial to create isolated transports, but subsequent
+		// tests share the original guestConfig's transport. Without this wait, the next
+		// test may start before Azure LB propagation completes the CIDR restoration.
+		g.Eventually(func(g Gomega) {
+			client, err := kubeclient.NewForConfig(guestConfig)
+			g.Expect(err).ToNot(HaveOccurred(), "failed to create kubeclient for transport recovery")
+			_, err = client.ServerVersion()
+			g.Expect(err).ToNot(HaveOccurred(), "KAS should be reachable on original transport after CIDR cleanup")
+		}).WithContext(ctx).WithTimeout(5 * time.Minute).WithPolling(10 * time.Second).Should(Succeed())
 	}()
 
-	kubeClient, err := kubeclient.NewForConfig(guestConfig)
-	g.Expect(err).NotTo(HaveOccurred())
-
 	// ensure that kube-apiserver is not reachable from anywhere
-	ensureAPIServerAllowedCIDRs(ctx, t, g, mgmtClient, kubeClient, hc, []string{"0.0.0.0/32"}, false)
+	ensureAPIServerAllowedCIDRs(ctx, t, g, mgmtClient, guestConfig, hc, []string{"0.0.0.0/32"}, false)
 	// ensure kube-apiserver is reachable when allowed CIDRs allow access from everywhere
-	// This is useful for testing purposes, as it allows us to access the kube-apiserver from any IP
-	// In a production environment, this should be restricted to specific CIDRs
-	ensureAPIServerAllowedCIDRs(ctx, t, g, mgmtClient, kubeClient, hc, append([]string{"0.0.0.0/0"}, generateTestCIDRs250()...), true)
+	ensureAPIServerAllowedCIDRs(ctx, t, g, mgmtClient, guestConfig, hc, append([]string{"0.0.0.0/0"}, generateTestCIDRs250()...), true)
 }
 
 func EnsureKubeAPIServerAllowedCIDRs(t *testing.T, ctx context.Context, mgmtClient crclient.Client, guestConfig *rest.Config, hc *hyperv1.HostedCluster) {
@@ -3621,7 +3627,7 @@ func EnsureKubeAPIServerAllowedCIDRs(t *testing.T, ctx context.Context, mgmtClie
 	})
 }
 
-func ensureAPIServerAllowedCIDRs(ctx context.Context, t testing.TB, g Gomega, mgmtClient crclient.Client, guestClient *kubeclient.Clientset, hc *hyperv1.HostedCluster, allowedCIDRs []string, shouldBeReachable bool) {
+func ensureAPIServerAllowedCIDRs(ctx context.Context, t testing.TB, g Gomega, mgmtClient crclient.Client, guestConfig *rest.Config, hc *hyperv1.HostedCluster, allowedCIDRs []string, shouldBeReachable bool) {
 	expectedCIDRs := make([]hyperv1.CIDRBlock, len(allowedCIDRs))
 	for i, cidr := range allowedCIDRs {
 		expectedCIDRs[i] = hyperv1.CIDRBlock(cidr)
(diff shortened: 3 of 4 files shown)


test/e2e/util/util.go
```diff
@@ -3571,6 +3592,35 @@ func ensureAPIServerAllowedCIDRs(ctx context.Context, t testing.TB, g Gomega, mg
...
+	if !netutil.IsPublicHC(hc) {
+		return nil
+	}
+	strategy := netutil.ServicePublishingStrategyByTypeByHC(hc, hyperv1.APIServer)
+	if strategy == nil {
+		return nil
+	}
+	switch strategy.Type {
+	case hyperv1.Route:
+		if azureutil.IsAroHCP() && !netutil.IsPrivateHC(hc) {
```
cblecker: The compound condition here doesn't quite match how the CPO makes this decision. The CPO uses `IsAroHCP()` as a standalone check when handling router services (`infra.go:459` — deletes `RouterPublicService` for ARO HCP unconditionally). ARO HCP never has a public router LB service with `LoadBalancerSourceRanges`, since Swift handles connectivity.

The issue is with `PublicAndPrivate` topology: `IsPublicHC` returns true (so the top guard passes), but `IsPrivateHC` also returns true (via the topology check), making `!IsPrivateHC` false. The guard doesn't fire and we return `RouterPublicService` — a service the CPO actively deletes for ARO HCP. The downstream `Eventually` would then time out waiting for `LoadBalancerSourceRanges` on a service that doesn't exist.

Simplifying to just `azureutil.IsAroHCP()` matches the CPO's logic:

```go
if azureutil.IsAroHCP() {
    return nil
}
```

test/e2e/util/util.go
```diff
@@ -3561,8 +3556,34 @@ func ensureAPIServerAllowedCIDRs(ctx context.Context, t testing.TB, g Gomega, mg
 			"HCP AllowedCIDRBlocks should match the HostedCluster spec")
 	}).WithContext(ctx).WithTimeout(time.Minute * 3).WithPolling(time.Second * 5).Should(Succeed())
 
+	// Wait for the CPO to reconcile the downstream service with the expected LoadBalancerSourceRanges.
+	// The target service depends on the APIServer publishing strategy:
+	// - Route: the "router" LB service carries the CIDRs
+	// - LoadBalancer: the KAS LB service itself carries the CIDRs
+	targetSvc := allowedCIDRsTargetService(hc, hcpNamespace)
+	if targetSvc != nil {
```
cblecker, line 3674: When `allowedCIDRsTargetService` returns nil, the service wait is silently skipped — there's no log output indicating this happened or why. Given that a nil return means the test goes straight to the reachability check without ensuring the CPO has propagated the CIDRs, it'd be useful to at least log it for CI debugging. Something like:

```go
} else {
    t.Logf("No downstream LB service identified for this cluster configuration; skipping LoadBalancerSourceRanges wait")
}
```

This also helps distinguish "we deliberately skipped" from "the function has a bug that returned nil unexpectedly" when investigating future flakes.

test/e2e/util/util.go
```diff
@@ -3571,6 +3592,35 @@ func ensureAPIServerAllowedCIDRs(ctx context.Context, t testing.TB, g Gomega, mg
 	}).WithContext(ctx).WithTimeout(time.Minute * 3).WithPolling(time.Second * 5).Should(Succeed())
 }
 
+// allowedCIDRsTargetService returns the LoadBalancer service that enforces AllowedCIDRBlocks
+// based on the HostedCluster's APIServer publishing strategy. Returns nil when no LB service
+// carries source ranges (private clusters, NodePort, ARO HCP).
+// Mirrors service selection in CPO: infra.go:reconcileAPIServerService, kas/service.go:ReconcileService.
```
cblecker: This citation is incomplete — it only covers the LoadBalancer path. For the Route case (which is the main path this PR is fixing), the relevant CPO code is `infra.go:reconcileHCPRouterServices` → `ingress/router.go:ReconcileRouterService`. The cited `kas/service.go:ReconcileService` sets `LoadBalancerSourceRanges` only in the `LoadBalancer` case, not for Route.

These file-level references are also going to get stale as the CPO migrates to the v2 component framework. Something like this would age better:

```go
// Mirrors CPO's API server and router service reconciliation logic.
```

test/e2e/util/util.go
```diff
@@ -3571,6 +3592,35 @@ func ensureAPIServerAllowedCIDRs(ctx context.Context, t testing.TB, g Gomega, mg
...
+func allowedCIDRsTargetService(hc *hyperv1.HostedCluster, hcpNamespace string) *corev1.Service {
```
cblecker, line 3715: nit: This function has 6+ branches mirroring CPO service selection logic. A table-driven unit test would be pretty cheap — the only external dependency is `IsAroHCP` which reads an env var, so you can stub it with `t.Setenv`. That would give some regression protection if the CPO's service selection ever drifts. Not blocking at all, just worth considering as a follow-up.

cblecker (commented):
The root cause analysis here is solid — the HTTP/2 connection reuse explanation is clear and the fix (fresh client per poll + waiting for the downstream service to reflect the updated source ranges before testing reachability) is the right approach rather than just a timing band-aid. The `allowedCIDRsTargetService` helper is a nice encapsulation of the CPO service selection logic.

A few comments inline, the most notable being a potential issue with the ARO HCP guard in the Route case.

test/e2e/util/util.go
bryan-cox: Done. Simplified to just `azureutil.IsAroHCP()` — good catch on the `PublicAndPrivate` topology case where both `IsPublicHC` and `IsPrivateHC` return true.

---

test/e2e/util/util.go
bryan-cox: Done. Added an else branch logging when the service wait is skipped.

---

test/e2e/util/util.go
bryan-cox: Done. Simplified to a generic reference that won't go stale with the v2 component migration.

---

test/e2e/util/util.go
bryan-cox: Done. Added a table-driven unit test for \`allowedCIDRsTargetService\` covering all branches: Route (AWS, Azure self-managed, ARO HCP), LoadBalancer (Azure, Azure via management annotation, AWS), private cluster, NodePort, and missing strategy. Uses \`t.Setenv\` to stub \`IsAroHCP\`.

---

test/e2e/util/util.go
```diff
@@ -3561,8 +3556,36 @@ func ensureAPIServerAllowedCIDRs(ctx context.Context, t testing.TB, g Gomega, mg
...
+		t.Log("No downstream LB service identified for this cluster configuration; skipping LoadBalancerSourceRanges wait")
+	}
+
+	// Create a fresh kubeclient per poll to avoid HTTP/2 connection reuse. Go's HTTP/2
+	// transport multiplexes requests over a single persistent TCP connection. If a prior
+	// successful request established a connection before NSG rules took effect, all
+	// subsequent requests reuse that connection and bypass the restriction.
 	g.Eventually(func(g Gomega) {
-		_, err = guestClient.ServerVersion()
+		freshClient, err := kubeclient.NewForConfig(rest.CopyConfig(guestConfig))
```
cblecker: Unfortunately `rest.CopyConfig` doesn't actually give you a new HTTP transport here. I traced through the client-go source: `kubeclient.NewForConfig` → `rest.HTTPClientFor` → `transport.New()`, which calls `tlsCache.get()` when `config.Transport` is nil. The cache key (`tlsCacheKey` in `transport/cache.go`) is built from the TLS data values — `string(c.TLS.CAData)`, `string(c.TLS.CertData)`, `string(c.TLS.KeyData)`, etc. — not pointers. `CopyConfig` copies the same byte content, and since `Dial` is nil, `TransportConfig()` leaves `DialHolder` nil too. Both the original config and the copy produce identical cache keys, so the cache returns the same `*http.Transport` instance with its existing HTTP/2 connection pool.

The simplest fix is to set `Dial` on the copied config before creating the client. `TransportConfig()` wraps a non-nil `Dial` in a new `&DialHolder{}` each time, making the pointer unique and busting the cache:

```go
cfg := rest.CopyConfig(guestConfig)
cfg.Dial = (&net.Dialer{Timeout: 30 * time.Second, KeepAlive: 30 * time.Second}).DialContext
freshClient, err := kubeclient.NewForConfig(cfg)
```

test/e2e/util/util.go
```diff
@@ -3561,8 +3556,36 @@ func ensureAPIServerAllowedCIDRs(ctx context.Context, t testing.TB, g Gomega, mg
...
+			g.Expect(actualSourceRanges).To(Equal(expectedSourceRanges),
+				"service %s/%s LoadBalancerSourceRanges should match expected CIDRs", targetSvc.Namespace, targetSvc.Name)
+		}).WithContext(ctx).WithTimeout(time.Minute * 3).WithPolling(time.Second * 5).Should(Succeed())
+	} else {
+		t.Log("No downstream LB service identified for this cluster configuration; skipping LoadBalancerSourceRanges wait")
+	}
+
+	// Create a fresh kubeclient per poll to avoid HTTP/2 connection reuse. Go's HTTP/2
+	// transport multiplexes requests over a single persistent TCP connection. If a prior
+	// successful request established a connection before NSG rules took effect, all
```
cblecker: nit: "NSG rules" is Azure-specific — this function runs on AWS (security groups) and GCP (firewall rules) too. Something like "network restrictions" or "load balancer source-range enforcement" would be accurate across platforms.

test/e2e/util/util_test.go
```diff
@@ -7,9 +7,136 @@ import (
...
+						Azure: ptr.To(hyperv1.AzurePlatformSpec{Topology: hyperv1.AzureTopologyPrivate}),
+					},
+					Services: []hyperv1.ServicePublishingStrategyMapping{{
+						Service:                   hyperv1.APIServer,
+						ServicePublishingStrategy: hyperv1.ServicePublishingStrategy{Type: hyperv1.Route},
+					}},
+				},
+			},
+			wantNil: true,
+		},
```
cblecker, line 98: These two test cases (NodePort and no-strategy) don't actually reach the branch they claim to test. Both create an `AWSPlatform` HostedCluster without setting `Platform.AWS`, so `IsPublicHC` evaluates `ptr.Deref(nil, AWSPlatformSpec{}).EndpointAccess == ""` — which matches neither `Public` nor `PublicAndPrivate` — and returns `false`. The function exits at the `!IsPublicHC(hc)` guard before the strategy switch is ever reached.

Using the `publicHC` helper fixes this (it correctly sets `EndpointAccess: hyperv1.Public` for AWS):

```go
{
    name:    "When NodePort strategy it should return nil",
    hc:      publicHC(hyperv1.AWSPlatform, hyperv1.NodePort),
    wantNil: true,
},
{
    name: "When no APIServer strategy it should return nil",
    hc: func() *hyperv1.HostedCluster {
        hc := publicHC(hyperv1.AWSPlatform, hyperv1.Route)
        hc.Spec.Services = nil
        return hc
    }(),
    wantNil: true,
},
```

The tests still return `wantNil: true` either way, but a bug in the `default` switch case or the nil-strategy guard wouldn't be caught as-is.

test/e2e/util/util_test.go
```diff
@@ -7,9 +7,136 @@ import (
...
+					Platform: hyperv1.PlatformSpec{Type: hyperv1.AWSPlatform},
+				},
+			},
+			wantNil: true,
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			g := NewWithT(t)
```
cblecker, line 117: nit: `azureutil.SetAsAroHCPTest(t)` already does exactly this — might as well use the helper for consistency with how the rest of the codebase sets up ARO HCP test environments.

cblecker (changes requested):
Second round overall looks good — all the previous feedback was addressed. A few new things came up on closer look, one of which is a real bug in the HTTP/2 fix.

test/e2e/util/util.go
bryan-cox: Done. Set `cfg.Dial` to create a unique `*transport.DialHolder` pointer per iteration, busting the TLS transport cache.

---

test/e2e/util/util.go
bryan-cox: Done. Replaced "NSG rules" with "load balancer source-range restrictions".

---

test/e2e/util/util_test.go
bryan-cox: Done. NodePort now uses `publicHC(hyperv1.AWSPlatform, hyperv1.NodePort)` so it passes `IsPublicHC` and exercises the switch default. No-strategy case uses `publicHC` with `hc.Spec.Services = nil` so it reaches the `strategy == nil` guard.

---

test/e2e/util/util_test.go
bryan-cox: Done. Switched to `azureutil.SetAsAroHCPTest(t)`.

---

cblecker (approved):
All round-2 feedback addressed correctly — the `cfg.Dial` fix properly busts the TLS transport cache, the test branch fixes reach the right code paths, and the nits are cleaned up.

/lgtm

cblecker:
@coderabbitai resume

enxebre:
cc @muraee

bryan-cox:
Retesting once more — all three Azure failures appear to be infrastructure issues unrelated to this PR's changes:

- **e2e-azure-self-managed**: Controller-runtime cache sync timeout; tests never ran.
- **e2e-aks-4-22**: TLS handshake timeouts across all tests, not specific to any single test.
- **e2e-aks**: `EnsureAzureWorkloadIdentityWebhookMutation` failed, which is a different test from the AllowedCIDRs changes in this PR. The AllowedCIDRs cleanup restores to a more permissive state (nil) before this test runs.

test/e2e/util/util.go
```diff
@@ -3660,8 +3567,40 @@ func ensureAPIServerAllowedCIDRs(ctx context.Context, t testing.TB, g Gomega, mg
...
+		}).WithContext(ctx).WithTimeout(time.Minute * 3).WithPolling(time.Second * 5).Should(Succeed())
+	} else {
+		t.Log("No downstream LB service identified for this cluster configuration; skipping LoadBalancerSourceRanges wait")
+	}
+
+	// Create a fresh kubeclient per poll to avoid HTTP/2 connection reuse. Go's HTTP/2
+	// transport multiplexes requests over a single persistent TCP connection. If a prior
+	// successful request established a connection before load balancer source-range
+	// restrictions took effect, all subsequent requests reuse that connection and bypass
+	// the restriction. Setting cfg.Dial ensures a unique TLS transport cache key per
```
muraee: this doesn't make sense, once a successful connection is made we would quit the polling loop anyway.

creating a fresh connection per `ensureAPIServerAllowedCIDRs()` call is sufficient, not per poll

test/e2e/create_cluster_test.go
```diff
@@ -116,8 +114,8 @@ func TestCreateCluster(t *testing.T) {
 		e2eutil.EnsureDefaultSecurityGroupTags(t, ctx, mgtClient, hostedCluster, clusterOpts)
 
 		if globalOpts.Platform == hyperv1.AzurePlatform {
-			e2eutil.EnsureKubeAPIServerAllowedCIDRs(t, ctx, mgtClient, guestConfig, hostedCluster)
 			e2eutil.EnsureAzureWorkloadIdentityWebhookMutation(t, ctx, guestClient)
+			e2eutil.EnsureKubeAPIServerAllowedCIDRs(t, ctx, mgtClient, guestConfig, hostedCluster)
```
muraee, line 123: why this change?

test/e2e/util/util.go
bryan-cox: For `shouldBeReachable=false`, we need to verify that KAS becomes unreachable after the LB source-range block propagates. If we create the client once, the first poll may succeed before the block takes effect, and Go's HTTP/2 transport will multiplex all subsequent requests over that same TCP connection — the test would never observe the block.

Creating a fresh `rest.Config` with a new `cfg.Dial` per iteration gives a unique key in client-go's TLS transport cache, forcing a new TCP connection each time. For `shouldBeReachable=true` a single client would work, but keeping it symmetric is simpler than a conditional for negligible cost (~36 TLS handshakes over 3 min).

---

test/e2e/create_cluster_test.go
bryan-cox: After AllowedCIDRs blocks all KAS traffic (`0.0.0.0/32`) and then restores it, the KAS pod's WI webhook sidecar may still be recovering. Since the webhook has `FailurePolicy: Ignore`, pods created during recovery are silently admitted without WI mutation, causing the WI test to time out waiting for projected token volumes. Running WI first avoids this window.

---

test/e2e/util/util.go
muraee: can we please update the comment to be more clear.

test/e2e/create_cluster_test.go
muraee: please add a comment explaining the order, so contributors know not to change it later.

test/e2e/util/util.go
bryan-cox: Done. Rewrote the comment to explain what HTTP/2 transport reuse does and why `cfg.Dial` busts the cache.

---

test/e2e/create_cluster_test.go
bryan-cox: Done. Added a comment explaining that WI must run first because AllowedCIDRs blocks/restores KAS traffic and the webhook sidecar (FailurePolicy: Ignore) may not be ready during recovery.

---

test/e2e/util/azure.go
```diff
@@ -70,34 +71,31 @@ func ValidateAzureWorkloadIdentityWebhookMutation(t testing.TB, ctx context.Cont
...
+	g.Eventually(func(g Gomega) {
+		existing := &corev1.Pod{}
+		err := guestClient.Get(ctx, types.NamespacedName{Name: podTemplate.Name, Namespace: podTemplate.Namespace}, existing)
+		if err == nil {
+			g.Expect(guestClient.Delete(ctx, existing)).To(Succeed(), "failed to delete pod for retry")
+			g.Eventually(func() bool {
+				err := guestClient.Get(ctx, types.NamespacedName{Name: podTemplate.Name, Namespace: podTemplate.Namespace}, &corev1.Pod{})
+				return apierrors.IsNotFound(err)
+			}).WithContext(ctx).WithTimeout(30 * time.Second).WithPolling(time.Second).Should(BeTrue(), "pod should be deleted before retry")
+		}
```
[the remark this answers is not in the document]
bryan-cox: Done. Added an else branch that asserts `IsNotFound` on unexpected errors, failing fast instead of proceeding to create.

---

cblecker:
@coderabbitai resume

test/e2e/util/util_test.go
```diff
@@ -7,9 +7,129 @@ import (
...
+			if tc.wantNil {
+				g.Expect(svc).To(BeNil())
+			} else {
+				g.Expect(svc).ToNot(BeNil())
+				g.Expect(svc.Name).To(Equal(tc.wantName))
+				g.Expect(svc.Namespace).To(Equal(ns))
+			}
+		})
+	}
+}
```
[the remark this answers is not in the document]
bryan-cox: Can't parallelize this test — the ARO HCP subtest uses `azureutil.SetAsAroHCPTest(t)` which calls `t.Setenv`. Go panics when `t.Setenv` is used in a test (or subtest of a test) that called `t.Parallel()`, since env vars are process-global. Confirmed by running with `-race`: `panic: testing: test using t.Setenv can not use t.Parallel`.

---

bryan-cox:

bryan-cox:

bryan-cox:
/test e2e-aws

one more time

PR #8471, opened 2026-05-09
WIP: feat(HCCO): stop reconciling over componentRoutes and appsDomain on guest Ingress config

joshbranham:
The HCCO was overwriting the entire Ingress.Spec on every reconciliation, which prevented customers from setting componentRoutes or appsDomain directly on the guest cluster. Preserve these fields so the Ingress Operator can act on customer-configured custom routes (e.g. Console, Downloads) and alternative apps domains.

Fixes 

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserve guest cluster component routes and apps domain when reconciling ingress; default domain when missing to prevent unintended overwrite.

* **New Features**
  * Add an ingress-focused validating admission policy to restrict unauthorized ingress modifications while allowing permitted updates.

* **Tests**
  * Add unit tests covering ingress reconciliation and the ingress admission policy; update test fixtures to include the new policy and binding.


support/globalconfig/ingress.go
```diff
@@ -19,12 +19,22 @@ func IngressConfig() *configv1.Ingress {
...
+
 	if hcp.Spec.Configuration != nil && hcp.Spec.Configuration.Ingress != nil {
 		cfg.Spec = *hcp.Spec.Configuration.Ingress
 	}
 	if cfg.Spec.Domain == "" {
 		cfg.Spec.Domain = IngressDomain(hcp)
 	}
+
+	cfg.Spec.ComponentRoutes = existingComponentRoutes
+	cfg.Spec.AppsDomain = existingAppsDomain
```
apahim, line 37: Having parts of the Ingress managed by hypershift and parts of it managed by customer dissolves the ownership boudaries.

Have we considered not defining an ingress at all in the hostedcluster CR and let the customer manage the entirety of it?

Or, have we considered the other extreme or it, which is to let customers add this extra configuration using the customer facing api of the managed service, OCM in this case, which will propagate the costumer desired state into the HostedCluster CR?

joshbranham: To set some background before answering your points below, this is a POC of a pragmatic interim solution to solve a long withstanding feature gap in ROSA between Classic and HCP. This proposal has been brought to HyperShift as a fast way to enable customers to brand the console themselves as they would in ROSA Classic or OCP itself.

> Have we considered not defining an ingress at all in the hostedcluster CR and let the customer manage the entirety of it?

As a service provider, today, we want to provide the console to the customer as part of the service, which requires us (unfortunately) to care about Ingress on the dataplane. There were discussions this morning in the HCP Office Hours where we discussed an RFE to support the console on the management cluster, which would enable "giving up" on dataplane ingress.

Maybe there is an interim solution here where HyperShift configures the ingress `config` day 0 and then never reconciles it again, similar to how they handle the default `ingresscontroller`?

> Or, have we considered the other extreme or it, which is to let customers add this extra configuration using the customer facing api of the managed service, OCM in this case, which will propagate the costumer desired state into the HostedCluster CR?

This is the full fledged solution, which requires significant development time and has considerations for all clients that interact with OCM as well. We are just entertaining a middle ground right now, where a field that a customer would otherwise own can be managed by them without the need for a service provider to build scaffolding around it to the HostedCluster.

I've got some more context in private Jira's and GDocs, feel free to ping me on Slack if you want to chat more 👍

PR #8472, opened 2026-05-09
OCPBUGS-86616,ARO-24037: feat(azure): add ACR pull identity to worker cloud.conf and VMSS

twolff-gh:

Adds an optional ImageRegistryCredentials field to AzurePlatformSpec that configures worker nodes to pull images from Azure Container Registry using a user-assigned managed identity.

When set, CPO writes userAssignedIdentityID into the worker cloud.conf and the NodePool controller attaches the MI to the VMSS via CAPZ.

This gives kubelet's ACR credential provider the identity it needs to authenticate without image pull secrets.

Fixes [ARO-24037](https://redhat.atlassian.net/browse/ARO-24037)

cloud-provider-azure accepts both ARM resource ID and client ID formats in userAssignedIdentityID — see [auth_func.go#L80](https://github.com/kubernetes-sigs/cloud-provider-azure/blob/e188952e7463d5539ef6fca0e0098acbb497638d/pkg/azclient/auth_func.go#L80)

We use the ARM resource ID since CAPZ also needs it for VMSS identity attachment.

Only the worker cloud.conf (ConfigMap) is modified — the CP cloud.conf (Secret) is unchanged.

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Azure platform spec can optionally specify Image Registry Credentials using a user-assigned managed identity for pulling from Azure Container Registry.

* **Behavior Changes**
  * Control-plane cloud config includes the configured user-assigned identity.
  * Worker machine templates are configured to use the specified user-assigned identity when present.

* **Tests**
  * Added tests for config generation, secret handling (no credential leakage), machine-template wiring, and JSON compatibility.


api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/hostedclusters.hypershift.openshift.io/AAA_ungated.yaml
```diff
@@ -5123,6 +5123,34 @@ spec:
                         - AzureGermanCloud
                         - AzureStackCloud
                         type: string
+                      imageRegistryCredentials:
```
bennerv: the openshift image registry is a component too.  Do we care to distinguish naming and do something like `azureContainerRegistryCredentials` to be more explicit?

api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/hostedclusters.hypershift.openshift.io/AAA_ungated.yaml
```diff
@@ -5123,6 +5123,34 @@ spec:
...
+                        properties:
+                          managedIdentity:
+                            description: managedIdentity identifies the user-assigned
+                              managed identity used for ACR image pulls.
+                            properties:
+                              resourceID:
+                                description: resourceID is the ARM resource ID of
+                                  the user-assigned managed identity.
+                                maxLength: 512
+                                minLength: 1
```
bennerv: Do we have min lengths for resourceIDs?  at minimum it's `/subscriptions/<subscription-guid>/resourceGroups/<min-resource-group-length>/providers/Microsoft.ManagedIdentity/userAssignedIdentities/<userAssignedIdentityName>`

api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/hostedclusters.hypershift.openshift.io/AAA_ungated.yaml
```diff
@@ -5123,6 +5123,34 @@ spec:
...
+                          to ACR without image pull secrets.
+                        properties:
+                          managedIdentity:
+                            description: managedIdentity identifies the user-assigned
+                              managed identity used for ACR image pulls.
+                            properties:
+                              resourceID:
+                                description: resourceID is the ARM resource ID of
+                                  the user-assigned managed identity.
+                                maxLength: 512
```
bennerv: where's this max length come from?

api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/hostedclusters.hypershift.openshift.io/AAA_ungated.yaml
twolff-gh: azureContainerRegistryCredentials works better

api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/hostedclusters.hypershift.openshift.io/AAA_ungated.yaml
twolff-gh: minLength: 131 — fixed segments (91) + min RG (1) + min identity name (3) + separators
maxLength: 345 — fixed segments (91) + max RG (90) + max identity name (128) + separators

https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules

api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/hostedclusters.hypershift.openshift.io/AAA_ungated.yaml
twolff-gh: I set a basic default.

The actual max is lower

twolff-gh: 345

api/hypershift/v1beta1/azure.go
```diff
@@ -546,6 +555,35 @@ type AzureResourceManagedIdentities struct {
 // +kubebuilder:validation:Pattern=`^[0-9a-fA-F]{8}-([0-9a-fA-F]{4}-){3}[0-9a-fA-F]{12}$`
 type AzureClientID string
 
+// AzureManagedIdentityResourceID is an ARM resource ID for a user-assigned managed identity.
+//
+// The expected format is:
+//
+//	/subscriptions/{subscriptionID}/resourceGroups/{resourceGroupName}/providers/Microsoft.ManagedIdentity/userAssignedIdentities/{identityName}
```
JoelSpeed: Have you considered requesting the subscription ID, resource group name and identity name as individual fields and constructing this string from those? UX would likely be better for the end user that way no?

api/hypershift/v1beta1/azure.go
```diff
@@ -546,6 +555,35 @@ type AzureResourceManagedIdentities struct {
...
+//
+// +kubebuilder:validation:XValidation:rule="self.lowerAscii().matches('^/subscriptions/[^/]+/resourcegroups/[^/]+/providers/microsoft\\\\.managedidentity/userassignedidentities/[^/]+$')",message="must be an ARM resource ID in the format /subscriptions/{subscriptionID}/resourceGroups/{resourceGroupName}/providers/Microsoft.ManagedIdentity/userAssignedIdentities/{identityName}"
+// +kubebuilder:validation:MinLength=131
+// +kubebuilder:validation:MaxLength=345
+type AzureManagedIdentityResourceID string
+
+// UserAssignedManagedIdentity identifies a user-assigned managed identity by its ARM resource ID.
+type UserAssignedManagedIdentity struct {
+	// resourceID is the ARM resource ID of the user-assigned managed identity.
```
JoelSpeed: You need to explain the format of the field and all validations so that users can see what value to expect to populate here

api/hypershift/v1beta1/azure_test.go
```diff
@@ -0,0 +1,102 @@
...
+		{
+			name: "When AzureContainerRegistryCredentials is zero value it should be omitted from JSON",
+			current: AzurePlatformSpec{
+				Location:       "eastus",
+				SubscriptionID: "sub-1",
+				TenantID:       "tenant-1",
+			},
+			expectFieldInJSON: false,
+		},
+	}
```
JoelSpeed: What are we testing here? This feels like we are testing the go standard library?

api/hypershift/v1beta1/azure.go
bryan-cox: +1 - we already have the subscription and resource group name elsewhere. Why not just pass in the identity name and build the string you need in the controller.

hypershift-operator/controllers/nodepool/azure.go
```diff
@@ -255,7 +267,7 @@ func (c *CAPI) azureMachineTemplate(_ context.Context, templateNameGenerator fun
 		return nil, fmt.Errorf("failed to apply Azure image defaults: %w", err)
 	}
 
-	spec, err := azureMachineTemplateSpec(c.nodePool)
+	spec, err := azureMachineTemplateSpec(c.nodePool, c.hostedCluster)
```
bryan-cox: Why not just pass the name in here rather than the whole HC?

api/hypershift/v1beta1/azure.go
bryan-cox, line 1: Any changes to the API should come with backing envtests testing the changes.

api/hypershift/v1beta1/azure_test.go
twolff-gh: True. Removed

api/hypershift/v1beta1/azure.go
bennerv: > +1 - we already have the subscription and resource group name elsewhere. Why not just pass in the identity name and build the string you need in the controller.

I don't think we should assume the subscription is the same subscription the cluster is in.  The same with resource group.  

As long as the capz identity has access to attach that identity to the virtual machine at creation time, the only real restriction is that the uamsi is in the same tenant.

hypershift-operator/controllers/nodepool/azure.go
twolff-gh: Makes sense. Will change

api/hypershift/v1beta1/azure.go
twolff-gh: Adding them now

api/hypershift/v1beta1/azure.go
bryan-cox: ^ that tribal knowledge should be explicitly documented in the API.

hypershift-operator/controllers/nodepool/azure.go
```diff
@@ -244,6 +245,16 @@ func azureMachineTemplateSpec(nodePool *hyperv1.NodePool) (*capiazure.AzureMachi
...
 
+	if hostedCluster != nil && hostedCluster.Spec.Platform.Azure != nil && hostedCluster.Spec.Platform.Azure.AcrImagePullManagedIdentityID != nil {
+		azureMachineTemplate.Template.Spec.Identity = capiazure.VMIdentityUserAssigned
+		azureMachineTemplate.Template.Spec.UserAssignedIdentities = append(
+			azureMachineTemplate.Template.Spec.UserAssignedIdentities,
+			capiazure.UserAssignedIdentity{
+				ProviderID: capzutil.ProviderIDPrefix + *hostedCluster.Spec.Platform.Azure.AcrImagePullManagedIdentityID,
+			},
+		)
+	}
```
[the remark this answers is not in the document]
twolff-gh: Wrong

field ProviderID string `json:"providerID"`
ProviderID is the identification ID of the user-assigned Identity, the format of an identity is: 'azure:///subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.ManagedIdentity/userAssignedIdentities/{identityName}'

CAPZ v1.21 does not require raw ARM resource IDs. The vendored type doc at
types.go:587-589 explicitly states the format is azure:///subscriptions/.... The
validator (ParseResourceID) and converter (sanitized) both use strings.TrimPrefix
to strip the prefix before passing to Azure SDK — they tolerate both formats but
the documented contract includes the prefix. Our code follows the contract.
Phase B confirmed it works at runtime.

api/hypershift/v1beta1/azure.go
```diff
@@ -546,6 +555,35 @@ type AzureResourceManagedIdentities struct {
...
+// AzureManagedIdentityResourceID is an ARM resource ID for a user-assigned managed identity.
+//
+// The expected format is:
+//
+//	/subscriptions/{subscriptionID}/resourceGroups/{resourceGroupName}/providers/Microsoft.ManagedIdentity/userAssignedIdentities/{identityName}
+//
+// +kubebuilder:validation:XValidation:rule="self.lowerAscii().matches('^/subscriptions/[^/]+/resourcegroups/[^/]+/providers/microsoft\\\\.managedidentity/userassignedidentities/[^/]+$')",message="must be an ARM resource ID in the format /subscriptions/{subscriptionID}/resourceGroups/{resourceGroupName}/providers/Microsoft.ManagedIdentity/userAssignedIdentities/{identityName}"
+// +kubebuilder:validation:MinLength=129
+// +kubebuilder:validation:MaxLength=241
+type AzureManagedIdentityResourceID string
```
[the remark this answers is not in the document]
twolff-gh: // managedIdentityResourceID is the ARM resource ID of a user-assigned managed identity
	// that will be attached to worker VM scale sets for authenticating image pulls from
	// Azure Container Registry.
	//
	// Requirements:
	//   - The identity must have the AcrPull role assigned on the target ACR(s).
	//   - The identity must be in the same Azure AD tenant as the HostedCluster.
	//   - The identity may reside in any subscription or resource group within that tenant.
	//
	// The expected format is (case-insensitive for the fixed path segments):
	//
	//   /subscriptions/{subscriptionID}/resourceGroups/{resourceGroupName}/providers/Microsoft.ManagedIdentity/userAssignedIdentities/{identityName}
	//
	// Length bounds (131–345) reflect the minimum and maximum possible lengths of the ARM
	// path given Azure's naming rules for resource groups (1–90 chars) and managed identity
	// names (3–128 chars).
	//
	// The value can be obtained via: az identity show --query id -o tsv
	//
	// +required
	// +kubebuilder:validation:XValidation:rule="self.lowerAscii().matches('^/subscriptions/[^/]+/resourcegroups/[^/]+/providers/microsoft\\\\.managedidentity/userassignedidentities/[^/]+$')",message="must be a user-assigned managed identity ARM resource ID in the format /subscriptions/{subscriptionID}/resourceGroups/{resourceGroupName}/providers/Microsoft.ManagedIdentity/userAssignedIdentities/{identityName}"
	// +kubebuilder:validation:MinLength=131
	// +kubebuilder:validation:MaxLength=345

api/hypershift/v1beta1/azure.go
twolff-gh: Updated. Let me know if it is missing anything now

api/hypershift/v1beta1/azure.go
twolff-gh: Updated the API. Let me know if it needs anything else

control-plane-operator/controllers/hostedcontrolplane/v2/cloud_controller_manager/azure/config.go
```diff
@@ -199,4 +203,5 @@ type AzureConfig struct {
 	ClusterServiceLoadBalancerHealthProbeMode       string `json:"clusterServiceLoadBalancerHealthProbeMode"`
 	ClusterServiceSharedLoadBalancerHealthProbePath string `json:"clusterServiceSharedLoadBalancerHealthProbePath,omitempty"`
 	ClusterServiceSharedLoadBalancerHealthProbePort int32  `json:"clusterServiceSharedLoadBalancerHealthProbePort,omitempty"`
+	UserAssignedIdentityID                          string `json:"userAssignedIdentityID,omitempty"`
```
twolff-gh, line 206: @bennerv @bryan-cox @JoelSpeed 

Maybe Im thinking about this to hard, lmk

The UserAssignedIdentityID is essentially undocumented behavior (takes either clientID or resourceID, and detects /SUBSCRIPTIONS/ in the string), but it is accepted behavior (original PR, not explicitly, but it was approved -- nobody mentioned anything about the behavior):

https://github.com/kubernetes-sigs/cloud-provider-azure/pull/8596

Should I add a comment to the struct, or do we need something more for this? Or maybe not in Hypershift, but someplace else? Or both?

May not be a big deal since its behavior is consistent all the way through. I just dont like the naming of the field, but we cant do anything about it. Unless I request upstream

I was going to add:
	// UserAssignedIdentityID accepts both a client ID (UUID) and an ARM resource ID.
	// cloud-provider-azure detects the format via a "/SUBSCRIPTIONS/" prefix check:
	// https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/pkg/azclient/auth_func.go

Flow:

1. cloud-provider-azure, https://github.com/kubernetes-sigs/cloud-provider-azure/blob/6f25da06da2dc0364ec45a035b4506f8be8d0dc7/pkg/azclient/auth_func.go#L80, as a credOptions.ID
2. azure-sdk-for-go, https://github.com/Azure/azure-sdk-for-go/blob/a148d76ad858a37cb131f150cba41ecbf78c7ee9/sdk/azidentity/managed_identity_credential.go#L93, ManagedIdentityCredentialOptions.ID is ManagedIDKind interface
3. azure-sdk-go, https://github.com/Azure/azure-sdk-for-go/blob/a148d76ad858a37cb131f150cba41ecbf78c7ee9/sdk/azidentity/managed_identity_client.go#L128, passed to managedidentity.New
4. microsoft-authentication-library-for-go, https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/e5d7af6beeb5a8e1fdf47fb3f978dc107f8d5e11/apps/managedidentity/managedidentity.go#L168, proper query param
5. Then token

twolff-gh: **Undocumented in that auth_func implicitly constructs the proper options for the azure and msal packages internally. No docstrings, etc

twolff-gh: Thinking too hard. Closing this out

api/hypershift/v1beta1/azure.go
```diff
@@ -546,6 +555,37 @@ type AzureResourceManagedIdentities struct {
 // +kubebuilder:validation:Pattern=`^[0-9a-fA-F]{8}-([0-9a-fA-F]{4}-){3}[0-9a-fA-F]{12}$`
 type AzureClientID string
 
+// AzureContainerRegistryCredentials configures authentication for worker nodes
+// pulling images from Azure Container Registry (ACR) using a user-assigned managed identity.
+// The identity does not need to be in the same subscription or resource group as the
+// HostedCluster, but it must be in the same Azure AD tenant. The management cluster's
+// CAPZ identity must have Microsoft.ManagedIdentity/userAssignedIdentities/*/assign/action
+// on the identity's scope to attach it to worker VM scale sets at creation time.
+type AzureContainerRegistryCredentials struct {
```
JoelSpeed: This appears to have taken a step backwards since the last review? Why did we lose the intermediate ManagedIdentity struct? Why did we lose the type alias with the validations assigned?

Also, on the thread where I mentioned splitting the subscriptionID, resourceGroupName and identityName into three separate fields, there was no actual response to say why we shouldn't do that

It seemed like @bryan-cox was in favour of that? 

The current field has a hard UX for the end user, when most of the string is static, lets build that for them

bryan-cox: Yeah, that's how we did it in other places for ARO HCP. However, I am amenable to keeping it like this if that is what the ARO HCP team wants.

This part of the API is not external as far as I know, and the only consumers are the ARO HCP folks themselves—unless that is incorrect, @bennerv?

twolff-gh: > This appears to have taken a step backwards since the last review? Why did we lose the intermediate ManagedIdentity struct? Why did we lose the type alias with the validations assigned?

> Also, on the thread where I mentioned splitting the subscriptionID, resourceGroupName and identityName into three separate fields, there was no actual response to say why we shouldn't do that

I had attempted to accommodate by iterating on the PR locally and ended up walking back those changes. The fields got eaten during that. I'll be restoring them.

twolff-gh: > there was no actual response to say why we shouldn't do that

Apologies, I conflated the responses.

> I mentioned splitting the subscriptionID, resourceGroupName and identityName

There are several places where the ARM API takes a resource ID directly - VNets, NSGs, others.

In this case it makes sense because it gets passed down to CAPZ that uses it directly. Decomposing it and then recomposing makes sense when they are independently meaningful to the API, but here I think they make more sense attached to the context of the full string being used as-is down further.

I'd argue that here it makes it harder for the customer because they have access to the full string directly with the proper sub and rg for the managed ID, and they would instead need to copy/paste the components out of the string or use code to present the fields properly. I don't know how most customers do this now though - copy/paste?

twolff-gh: But I'll need clarification from @bennerv

JoelSpeed: Ok, that seems like a reasonable argument. If you think the most common use case is that they can copy/paste the whole string out and don't need to form it themselves, then keep as you had with the single field

twolff-gh: @JoelSpeed 

From an API surface support vs UX side, we could do both potentially:
https://learn.microsoft.com/en-us/azure/openshift/howto-create-openshift-cluster?pivots=aro-deploy-az-cli#create-the-cluster

The CLI here ^ resolves both - full and name only. But that's a CLI experience vs API. Are they different ?

We could add both fields - mutually exclusive, and the resource ID name only resolves against the current cluster sub and rg? As long as the API surface is documented appropriately, they could do either. Is that reasonable?

api/hypershift/v1beta1/azure.go
```diff
@@ -469,6 +469,15 @@ type AzurePlatformSpec struct {
 	// +kubebuilder:validation:MaxLength=255
 	TenantID string `json:"tenantID"`
 
+	// azureContainerRegistryCredentials configures authentication for worker nodes pulling images
+	// from Azure Container Registry (ACR) using a user-assigned managed identity.
+	// When set, the identity is attached to worker VM scale sets and its resource ID is written
+	// into the worker cloud provider config so kubelet's ACR credential provider can authenticate
+	// without image pull secrets.
+	//
```
bryan-cox, line 477: suggestion: This field should have the `+rollout` marker like other fields that trigger NodePool rollouts (e.g. `pullSecret`, `sshKey`, `image` on `HostedClusterSpec`). When this changes, the CAPZ machine template spec changes, which produces a new template name and triggers a rollout.

```suggestion
	// azureContainerRegistryCredentials configures authentication for worker nodes pulling images
	// from Azure Container Registry (ACR) using a user-assigned managed identity.
	// When set, the identity is attached to worker VM scale sets and its resource ID is written
	// into the worker cloud provider config so kubelet's ACR credential provider can authenticate
	// without image pull secrets.
	// Changing this value will trigger a rollout for all existing NodePools in the cluster.
	//
	// +rollout
	// +optional
	AzureContainerRegistryCredentials AzureContainerRegistryCredentials `json:"azureContainerRegistryCredentials,omitzero"`
```

control-plane-operator/controllers/hostedcontrolplane/v2/cloud_controller_manager/azure/config_test.go
```diff
@@ -77,6 +77,61 @@ func newTestHCP(annotations map[string]string) *hyperv1.HostedControlPlane {
 	}
 }
 
+func TestConfigWithAcrMI(t *testing.T) {
+	hcp := newTestHCP(nil)
+	hcp.Namespace = "HCP_NAMESPACE"
```
bryan-cox: suggestion: These tests use `t.Fatalf` for assertions, but project convention prefers Gomega (`g.Expect`). For example:

```go
g := NewGomegaWithT(t)
```

The existing `TestConfigErrorStates` in this file also uses `t.Fatalf`, so this is consistent within the file — but since you're adding new tests, it'd be a good opportunity to align with the broader project preference.

hypershift-operator/controllers/nodepool/azure.go
```diff
@@ -244,6 +245,16 @@ func azureMachineTemplateSpec(nodePool *hyperv1.NodePool) (*capiazure.AzureMachi
 		}
 	}
 
+	if acrIdentityResourceID != "" {
+		azureMachineTemplate.Template.Spec.Identity = capiazure.VMIdentityUserAssigned
+		azureMachineTemplate.Template.Spec.UserAssignedIdentities = append(
+			azureMachineTemplate.Template.Spec.UserAssignedIdentities,
+			capiazure.UserAssignedIdentity{
+				ProviderID: capzutil.ProviderIDPrefix + acrIdentityResourceID,
+			},
```
bryan-cox, line 254: suggestion: Consider adding a test case where `UserAssignedIdentities` already contains entries before the ACR identity is appended. The current `append` is correct, but if another code path later populates this slice, a test proving the append (vs. overwrite) behavior would catch regressions early.

hypershift-operator/controllers/nodepool/azure.go
```diff
@@ -244,6 +245,16 @@ func azureMachineTemplateSpec(nodePool *hyperv1.NodePool) (*capiazure.AzureMachi
 		}
 	}
 
+	if acrIdentityResourceID != "" {
+		azureMachineTemplate.Template.Spec.Identity = capiazure.VMIdentityUserAssigned
+		azureMachineTemplate.Template.Spec.UserAssignedIdentities = append(
+			azureMachineTemplate.Template.Spec.UserAssignedIdentities,
+			capiazure.UserAssignedIdentity{
```
bryan-cox, line 252: praise: The `ProviderIDPrefix + acrIdentityResourceID` construction is correct — the result `azure:///subscriptions/...` (three slashes) matches CAPZ's documented format at `types.go:587`. CodeRabbit flagged this as wrong but the suggestion to remove the prefix was incorrect.

control-plane-operator/controllers/hostedcontrolplane/v2/cloud_controller_manager/azure/config_test.go
twolff-gh: Done. incoming

api/hypershift/v1beta1/azure.go
twolff-gh: Done. Incoming

hypershift-operator/controllers/nodepool/azure.go
twolff-gh: I could decompose this to a func and make that testable, the only reason I wouldn't do it now is since azureMachineTemplateSpec is building from nothing (not modifying a spec with those fields populated), there wouldn't be a case where we would run into the append vs overwrite path.

If someone adds that path later, I would expect azureMachineTemplateSpec to change in some way potentially to enable the behavior, and then it should follow that the behavior be modified then.

As it is, UserAssignedIdentities is always nil slice when the append is called. Nothing gets a chance to be overwritten.

Or, I could just decompose to a helper and make it testable directly now to preemptively guard. Easy enough. :upside_down_face:

api/hypershift/v1beta1/azure.go
twolff-gh: Per Slack. Sticking with the full resource ID only to keep the API clean and the behavior consistent

hypershift-operator/controllers/nodepool/azure.go
twolff-gh: @bryan-cox

bryan-cox: Makes sense. I'm good to leave it as is.

bryan-cox (approved):

bennerv (commented):
/lgtm

bryan-cox:
/retitle OCPBUGS-86616,[ARO-24037](https://redhat.atlassian.net/browse/ARO-24037) feat(azure): add ACR pull identity to worker cloud.conf and VMSS

bryan-cox:
/retitle OCPBUGS-86616,[ARO-24037](https://redhat.atlassian.net/browse/ARO-24037): feat(azure): add ACR pull identity to worker cloud.conf and VMSS

api/hypershift/v1beta1/azure.go
```diff
@@ -469,6 +469,17 @@ type AzurePlatformSpec struct {
...
+	// azureContainerRegistryCredentials configures authentication for worker nodes pulling images
+	// from Azure Container Registry (ACR) using a user-assigned managed identity.
+	// When set, the identity is attached to worker VM scale sets and its resource ID is written
+	// into the worker cloud provider config so kubelet's ACR credential provider can authenticate
+	// without image pull secrets.
+	// Changing this value will trigger a rollout for all existing NodePools in the cluster.
+	//
+	// +rollout
+	// +optional
+	AzureContainerRegistryCredentials AzureContainerRegistryCredentials `json:"azureContainerRegistryCredentials,omitzero"`
```
JoelSpeed: In the future, might there be other container registry options? Worth adding this as a two level thing to allow future siblings?

```
containerRegistry:
  credentials:
    ...
```

api/hypershift/v1beta1/azure.go
JoelSpeed: Needs to be explained in this godoc so that it ends up in the generated docs the user would see with kubectl explain.

Please expand this comment on the field to explain the valid input

api/hypershift/v1beta1/azure.go
```diff
@@ -546,6 +557,35 @@ type AzureResourceManagedIdentities struct {
...
+}
+
+// AzureContainerRegistryCredentials configures authentication for worker nodes
+// pulling images from Azure Container Registry (ACR) using a user-assigned managed identity.
+// The identity does not need to be in the same subscription or resource group as the
+// HostedCluster, but it must be in the same Azure AD tenant. The management cluster's
+// CAPZ identity must have Microsoft.ManagedIdentity/userAssignedIdentities/*/assign/action
+// on the identity's scope to attach it to worker VM scale sets at creation time.
+type AzureContainerRegistryCredentials struct {
+	// managedIdentity identifies the user-assigned managed identity used for ACR image pulls.
```
JoelSpeed: Is there likely to be alternatives to managedIdentity in the future? Wondering if we should set this up as a discriminated union to start with and then that'll allow easier expansion in the future

api/hypershift/v1beta1/azure.go
twolff-gh: Good question. What credential type would live here that isn't already a pull secret and isn't a managed identity variant?
Or are you thinking of other options?

api/hypershift/v1beta1/azure.go
twolff-gh: Everything that isnt a managed identity is either a pull secret or a completely different config path.
I dont see anything landing here as a second union variant.
There wouldnt be a meaningfully different auth type that would share this field.

- Service principal: pull secret
- Projected SA tokens: kubelet credential provider config, not this field
- System assigned MI: same struct, resourceID becomes optional

What do you think?

api/hypershift/v1beta1/azure.go
JoelSpeed: Perhaps there might be some TLS or other container registry configuration that might be needed. Proxy information? Not sure tbh, just something we could do now to make life easier in a possible future

api/hypershift/v1beta1/azure.go
JoelSpeed: That seems reasonable with todays view of credentials, but I'm still hesitant that we might want to make this structure more future proof in case we later want sibling options for how to use the service principal for example

api/hypershift/v1beta1/azure.go
twolff-gh: I really dont think there would be anything in the future, but Ill go check

Unless @bennerv, do you know if we would ever need anything else here? Do we need to change the API surface?

bryan-cox:
@twolff-gh since you already lost lgtm, would you mind rebasing the branch please? 😄 

That should fix the codecov/project error you are seeing.

twolff-gh:
> @twolff-gh since you already lost lgtm, would you mind rebasing the branch please? 😄
> 
> That should fix the codecov/project error you are seeing.

Will do!

api/hypershift/v1beta1/azure.go
bennerv: I like the advice from Joel.  I'd do what he suggests to future proof ourselves.

twolff-gh: Sounds good

@JoelSpeed adding it now

api/hypershift/v1beta1/azure.go
twolff-gh: Makes sense. WOrking on it now

api/hypershift/v1beta1/azure.go
twolff-gh: @JoelSpeed Added.

api/hypershift/v1beta1/azure.go
twolff-gh: Done. You can review again when you can. Thanks!

JoelSpeed (commented):
/approve for API

Thanks @twolff-gh

control-plane-operator/controllers/hostedcontrolplane/v2/cloud_controller_manager/azure/config_test.go
```diff
@@ -77,6 +79,64 @@ func newTestHCP(annotations map[string]string) *hyperv1.HostedControlPlane {
 	}
 }
 
+func TestConfigWithAcrMI(t *testing.T) {
```
enxebre: nit: please let this be a test case of the exsiting TestConfig

api/hypershift/v1beta1/azure.go
```diff
@@ -546,6 +556,67 @@ type AzureResourceManagedIdentities struct {
...
+}
+
+// AzureContainerRegistryCredentialType identifies the type of credential used for ACR image pulls.
+//
+// +kubebuilder:validation:Enum=ManagedIdentity
+type AzureContainerRegistryCredentialType string
+
+const (
+	// AzureContainerRegistryCredentialManagedIdentity uses a user-assigned managed identity for ACR authentication.
+	AzureContainerRegistryCredentialManagedIdentity AzureContainerRegistryCredentialType = "ManagedIdentity"
```
enxebre, line 599: is this intentionally not using the existing AzureAuthenticationType?

enxebre:
how does this relate to UseManagedIdentityExtension input in the config file? should if influence it?

enxebre:
dropped some questions, lgtm overall

api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/hostedclusters.hypershift.openshift.io/AAA_ungated.yaml
```diff
@@ -5211,6 +5211,61 @@ spec:
                         - AzureGermanCloud
                         - AzureStackCloud
                         type: string
+                      containerRegistry:
+                        description: |-
+                          containerRegistry configures how worker nodes authenticate to Azure Container Registry (ACR).
+                          When set, the managed identity is attached to worker VM scale sets and its resource ID is
```
bennerv: nit: 
> worker VM scale sets

Technically they're deployed as independent VMs.

api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/hostedclusters.hypershift.openshift.io/AAA_ungated.yaml
```diff
@@ -5211,6 +5211,61 @@ spec:
...
+                          authenticate without image pull secrets.
+                          Changing this value will trigger a rollout for all existing NodePools in the cluster.
+                        properties:
+                          credentials:
+                            description: |-
+                              credentials configures authentication for worker nodes pulling images from ACR
+                              using a user-assigned managed identity.
+                              The identity does not need to be in the same subscription or resource group as the
+                              HostedCluster, but it must be in the same Azure AD tenant. The management cluster's
+                              CAPZ identity must have Microsoft.ManagedIdentity/userAssignedIdentities/*/assign/action
```
bennerv, line 5228: do we have this permission in our built-in role today so customers can assign it at the managed identity level? 

Looks like not, make sure we have a card to update the built-in role: https://dev.azure.com/msazure/One/_git/AD-RBAC-RoleDefinitions?path=/MICROSOFT.REDHATOPENSHIFT/PROD/Azure%20Red%20Hat%20OpenShift%20Hosted%20Control%20Planes%20Cluster%20API%20Provider.json

^ ARO HCP specific.

bennerv (commented):
two super minor things.

api/hypershift/v1beta1/azure.go
bennerv: The naming of these things are fun.  `AzureAuthenticationTypeManagedIdentities` -> corresponds with an Azure-specific technology we usually call MSI or Managed Identities for Azure Resource.  

Since we're a first party service, we can fetch credentials for user-assigned or system-assigned managed identities.  This is what the `AzureAuthenticationTypeManagedIdentities AzureAuthenticationType = "ManagedIdentities"` corresponds to. 

For ACR managed identity, this attaches a managed identity to the VMs created, and then the VMs use the instance metadata endpoint to authenticate as the user-assigned managed identity to Azure and pull images from ACR. 

So based on our current definition of them, they're not interchangeable.

api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/hostedclusters.hypershift.openshift.io/AAA_ungated.yaml
twolff-gh: https://redhat.atlassian.net/browse/ARO-27663

api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/hostedclusters.hypershift.openshift.io/AAA_ungated.yaml
twolff-gh: Fixed. Will push

control-plane-operator/controllers/hostedcontrolplane/v2/cloud_controller_manager/azure/config_test.go
twolff-gh: Fixed. Will push

api/hypershift/v1beta1/zz_generated.featuregated-crd-manifests/hostedclusters.hypershift.openshift.io/AAA_ungated.yaml
twolff-gh: Is this an RP update then?

twolff-gh:
> how does this relate to UseManagedIdentityExtension input in the config file? should if influence it?

Theyre related but I dont think we need to change UseManagedIdentityExtension here.

Its already set true on the ARO HCP worker cloud.conf. userAssignedIdentityID is read by the kubelet credential provider for ACR token acquisition. It needs UseManagedIdentityExtension=true as a prereq, which is already set.

What do you think? Hopefully this helped :pray: 

@enxebre

control-plane-operator/controllers/hostedcontrolplane/v2/cloud_controller_manager/azure/config.go
```diff
@@ -28,6 +28,10 @@ func adaptConfig(cpContext component.WorkloadContext, cm *corev1.ConfigMap) erro
 		return err
 	}
 
+	if cpContext.HCP.Spec.Platform.Azure != nil && cpContext.HCP.Spec.Platform.Azure.ContainerRegistry.Credentials.ManagedIdentity.ResourceID != "" {
```
machi1990, line 31: What happens when the user wants to unset it? i.e opt out of the functionality?

hypershift-operator/controllers/nodepool/azure.go
```diff
@@ -244,6 +245,16 @@ func azureMachineTemplateSpec(nodePool *hyperv1.NodePool) (*capiazure.AzureMachi
 		}
 	}
 
+	if acrIdentityResourceID != "" {
```
machi1990, line 248: Same question as https://github.com/openshift/hypershift/pull/8472/changes#r3386583683

How does the user say that "I want to disable this feature by setting the field to empty"

twolff-gh: It would be to set to empty

https://docs.google.com/document/d/124DWiXThqpuh1uIGqoVWqwTovQE5RDlLh9iwy_W1bA0/edit?tab=t.0#heading=h.s8clv8jkiao

machi1990: Ah perfect. With this check here, I missed the part where the unsetting was occurring.

control-plane-operator/controllers/hostedcontrolplane/v2/cloud_controller_manager/azure/config.go
twolff-gh: Confirmed again today - the UX for Day1/2

Day 1 — Create cluster (PUT): Include properties.platform.containerRegistry with the managed identity resource ID. Omit it entirely to create without — most clusters will start this way.

Day 2 — Enable (PATCH): Send a PATCH adding containerRegistry with the identity. Triggers a nodepool rollout — new VMs get the identity attached, cloud.conf gets userAssignedIdentityID. Customer sees UpdatingConfig: True on the nodepool during rollout.

Day 2 — Disable (PATCH): Send "containerRegistry": null in the PATCH body. Per JSON merge-patch (RFC 7396), null removes the field. Triggers another rollout — identity removed from VMs, userAssignedIdentityID removed from cloud.conf.

https://redhat.atlassian.net/browse/ARO-24037

twolff-gh: Lmk if I need to address anything :pray:

api/hypershift/v1beta1/azure.go
twolff-gh: Anything left for this?

control-plane-operator/controllers/hostedcontrolplane/v2/cloud_controller_manager/azure/config.go
machi1990: Thank for for comfirming

bryan-cox:
/hold 

@twolff-gh wants to write up how he tested this and share it first.

PR #8480, opened 2026-05-11
OCPBUGS-85351: fix(azure): prevent externalDNSDomain from shadowing cluster apps domain

bryan-cox:
## Summary

Prevents a misconfiguration where `--external-dns-domain` equals `{name}.{baseDomain}` on a private Azure HostedCluster, which causes the PLS controller to create a Private DNS zone that shadows `*.apps` DNS resolution.

This is validated at three layers:
- **Webhook**: Rejects HostedCluster creation when the external DNS domain would shadow the cluster domain (skipped on ARO HCP, which handles DNS differently)
- **CLI**: Validates `--external-dns-domain` before cluster creation
- **Controller**: Detects shadowing and sets `AzurePrivateDNSAvailable=False` with reason `BaseDomainShadowsClusterDomain`, skipping zone creation entirely

### Problem

When `--external-dns-domain` equals `{name}.{baseDomain}`, the PLS controller's `baseDomainFromServices()` derives a zone name that collides with the cluster's apps domain. The resulting Private DNS zone becomes authoritative for **all** queries under `{name}.{baseDomain}` on the VNet, including `*.apps.{name}.{baseDomain}`. Since the zone only has `api` and `oauth` A records, apps queries return NXDOMAIN — breaking console and all ingress.

### Why the Controller Cannot Self-Heal (Two Routers Problem)

We initially considered adding a `*.apps` wildcard A record to the shadowing zone, but this doesn't work due to the "Two Routers Problem":

1. **Management-plane router** (`private-router`): An HAProxy pod in the HCP namespace, fronted by an internal load balancer and exposed to the guest VNet via Private Link. HAProxy uses SNI-based routing with ACLs **only** for `.hypershift.local` hostnames (KAS, ignition, konnectivity, OAuth). Any hostname that doesn't match falls through to `default_backend kube_api`.

2. **Data-plane router** (`router-default`): The OpenShift ingress controller running on worker nodes, serving `*.apps.{name}.{baseDomain}` with the correct wildcard ingress certificate.

Adding `*.apps → PE IP` would route apps traffic to `private-router`, where it falls through to KAS (no matching SNI ACL) and the client gets a **TLS certificate mismatch** — KAS cert instead of ingress wildcard cert. This changes the failure from NXDOMAIN to a cert error, but neither is usable.

The correct target (data-plane `router-default` IP) is **not discoverable** from the PLS controller:
- The controller runs in the control plane and has no client to the guest cluster
- No HCP status field reports the guest cluster's ingress IP
- HCCO does not propagate this information back

### Why Prevention Is the Only Fix

All three inputs that determine whether shadowing occurs are **immutable**:
- `metadata.name` — Kubernetes-immutable
- `spec.dns.baseDomain` — CEL: `oldSelf == "" || self == oldSelf`
- `spec.services` (contains the hostnames derived from `--external-dns-domain`) — CEL: `self.services == oldSelf.services`

A user cannot correct this misconfiguration without destroying and recreating the cluster. This makes the webhook and CLI validation critical — they prevent the problem before it becomes unfixable.

### Why a Webhook Instead of CEL

`--external-dns-domain` is not a first-class API field. It is encoded into individual service hostnames in `spec.services[].servicePublishingStrategy.route.hostname` and `.loadBalancer.hostname`. Detecting the shadowing condition in CEL would require extracting the domain suffix from each hostname via string manipulation, then comparing it against `metadata.name + "." + spec.dns.baseDomain` — this is awkward and fragile in CEL expressions.

The webhook provides a cleaner implementation and better error messages. Per maintainer guidance, it is acceptable as a UX improvement (defense in depth) alongside the controller guard.

## Test plan

- [x] `TestValidateCreateAzureHostedCluster` — 8 BDT cases covering exact match, parent domain, case-insensitive, LoadBalancer strategy, nil platform, non-Azure, no hostnames
- [x] `TestValidateCreateAzureHostedCluster_WhenRunningOnAROHCP_ItShouldSkipShadowingValidation` — verifies webhook skips validation on managed Azure
- [x] `TestValidateExternalDNSDomain` — 8 BDT cases covering exact match, parent domain, case-insensitive, non-dot-boundary suffix, empty inputs
- [x] `TestBaseDomainShadowsClusterDomain` — 8 detection test cases
- [x] `TestReconcile_WhenBaseDomainShadowsClusterDomain_ItShouldSetDegradedConditionAndSkipZoneCreation` — full reconcile test verifying degraded condition is set and no zone is created
- [x] `TestReconcile_WhenBaseDomainDiffersFromClusterDomain_ItShouldCreateZoneNormally` — positive test confirming normal path still works


cmd/cluster/azure/create.go
```diff
@@ -222,6 +222,10 @@ func (o *RawCreateOptions) Validate(ctx context.Context, _ *core.CreateOptions)
 		}
 	}
 
+	if err := validateExternalDNSDomain(opts); err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
bryan-cox: Stale — the CLI validation was dropped entirely in the latest force-push. `create.go` has no changes vs main.

control-plane-operator/controllers/azureprivatelinkservice/controller.go
```diff
@@ -334,8 +334,25 @@ func (r *AzurePrivateLinkServiceReconciler) Reconcile(ctx context.Context, req c
...
+			meta.SetStatusCondition(&azPLS.Status.Conditions, metav1.Condition{
+				Type:               string(hyperv1.AzurePrivateDNSAvailable),
+				Status:             metav1.ConditionFalse,
+				Reason:             "DNSZoneShadowing",
+				Message:            fmt.Sprintf("Base domain DNS zone %q would shadow *.apps DNS resolution because it matches {clusterName}.{baseDomain}; zone creation skipped to prevent NXDOMAIN for ingress", azPLS.Spec.BaseDomain),
+				ObservedGeneration: azPLS.Generation,
+			})
+			if err := r.Status().Patch(ctx, azPLS, patch); err != nil {
+				return ctrl.Result{}, fmt.Errorf("failed to update DNS shadowing condition: %w", err)
+			}
```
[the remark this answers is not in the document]
bryan-cox: Stale — the skip-and-degrade approach was replaced with the wildcard approach in the latest force-push. `reconcileBaseDomainDNS` now adds a `*.apps` wildcard A record to Zone 3 when shadowing is detected, so the zone is still created with all necessary records (api, oauth, and \*.apps). No zone is skipped, and existing broken clusters self-heal on next reconcile.

control-plane-operator/controllers/azureprivatelinkservice/controller.go
```diff
@@ -744,6 +744,13 @@ func (r *AzurePrivateLinkServiceReconciler) reconcileDNS(ctx context.Context, az
 	}, log)
 }
 
+func baseDomainShadowsClusterDomain(baseDomain, clusterName, hcpBaseDomain string) bool {
+	if baseDomain == "" || clusterName == "" || hcpBaseDomain == "" {
+		return false
+	}
+	return strings.EqualFold(baseDomain, clusterName+"."+hcpBaseDomain)
```
Nirshal: Shouldn't we consider other shadowing scenarios like `baseDomain == hcpBaseDomain`? For instance: `--external-dns-domain acm-dev04.azure.devcluster.openshift.com` with `--base-domain acm-dev04.azure.devcluster.openshift.com`. Then the private zone will be `acm-dev04.azure.devcluster.openshift.com` that will be a suffix of `*.apps.hcp-two.acm-dev04.azure.devcluster.openshift.com` -> shadowing. 
Also, if `baseDomain` is a parent of `hcpBaseDomain` (but this is unlikely to happen, I think). 
I would suggest something like this:



```suggestion
clusterDomain := strings.ToLower(clusterName + "." + hcpBaseDomain)
return strings.HasSuffix(clusterDomain, strings.ToLower(baseDomain))
```

bryan-cox: Good catch — the `baseDomain == hcpBaseDomain` case is a valid shadowing scenario I missed. Adopted the `HasSuffix` approach with a dot-boundary guard to avoid false positives on non-domain-boundary suffixes (e.g., `ample.com` matching `example.com`):

```go
clusterDomain := strings.ToLower(clusterName + "." + hcpBaseDomain)
baseDomainLower := strings.ToLower(baseDomain)
if clusterDomain == baseDomainLower {
    return true
}
return strings.HasSuffix(clusterDomain, "."+baseDomainLower)
```

Added 3 new test cases: `baseDomain == hcpBaseDomain` (your scenario), non-domain-boundary suffix (`ample.com`), and multi-level parent domain.

---

hypershift-operator/controllers/hostedcluster/hostedcluster_webhook.go
```diff
@@ -205,6 +210,64 @@ func (v hostedClusterValidator) validateCreateKubevirtHostedCluster(_ context.Co
 	return nil, nil
 }
 
+// CEL cannot express this: array iteration + string split + cross-field suffix matching exceeds the cost budget.
```
enxebre: This validation can be expressed in CEL — it doesn't need a webhook. The claim that "CEL cannot express this" is inaccurate.

The following CEL rule on `HostedClusterSpec` catches the case where a service hostname's domain suffix matches or is a parent of `baseDomain`:

```
self.platform.type != "Azure" || self.dns.baseDomain == "" || !self.services.exists(s, (has(s.servicePublishingStrategy.route) && has(s.servicePublishingStrategy.route.hostname) && s.servicePublishingStrategy.route.hostname.contains('.') && ('.' + self.dns.baseDomain).endsWith('.' + s.servicePublishingStrategy.route.hostname.substring(s.servicePublishingStrategy.route.hostname.indexOf('.') + 1))) || (has(s.servicePublishingStrategy.loadBalancer) && has(s.servicePublishingStrategy.loadBalancer.hostname) && s.servicePublishingStrategy.loadBalancer.hostname.contains('.') && ('.' + self.dns.baseDomain).endsWith('.' + s.servicePublishingStrategy.loadBalancer.hostname.substring(s.servicePublishingStrategy.loadBalancer.hostname.indexOf('.') + 1))))
```

Key technique: `('.' + baseDomain).endsWith('.' + domainSuffix)` handles both exact match and parent match in one expression (the prepended dot ensures dot-boundary matching).

Verified: `make api` passes and envtests pass across k8s 1.30–1.35.

Note: this checks against `baseDomain` (available on `HostedClusterSpec`), not `clusterName.baseDomain` (which requires `metadata.name`, only accessible at the root `HostedCluster` type where controller-gen doesn't emit XValidation). The controller guard still catches the narrower exact-match case.

Per `api/CLAUDE.md`: "Use CEL rules, not webhooks, for new validation."

bryan-cox: Done. Replaced the webhook validation with a CEL rule on `HostedClusterSpec` and added envtest coverage for it. Also removed the `validateCreateAzureHostedCluster` method and its tests.

The CEL rule uses your suggested technique: `('.' + self.dns.baseDomain).endsWith('.' + hostname.substring(hostname.indexOf('.') + 1))` to handle both exact-match and parent-domain overlap with dot-boundary safety.

---

JoelSpeed:
/approve

For API change

PR #8486, opened 2026-05-12
OCPBUGS-56376: Honor spec.configuration.apiServer.encryption in HostedCluster

vismishr:

When a user sets `spec.configuration.apiServer.encryption.type: aescbc` on a
HostedCluster, the field is copied to the HostedControlPlane but never consumed
by the CPO's KAS component. Only `spec.secretEncryption` is wired into the KAS
deployment, making the `apiServer.encryption` config a silent no-op. 

This PR adds `reconcileAPIServerEncryptionConfig()` which auto-generates a
32-byte AESCBC key and populates `spec.secretEncryption` when
`configuration.apiServer.encryption.type` is `aescbc` and `secretEncryption`
is not already set. If `secretEncryption` is already configured, it takes
precedence (override power). This follows the same pattern used by KubeVirt
platform defaults.

Also adds validation in `validateOCPConfigurations()` for unsupported encryption
types: `aesgcm` returns an error (not supported in HyperShift), and `KMS`
returns an error when `secretEncryption` is not configured with platform-specific
KMS settings.

Fixes https://redhat.atlassian.net/browse/OCPBUGS-56376

- The auto-generation logic reuses the exact same pattern as KubeVirt's
  `reconcileKubevirtPlatformDefaultSettings()` (same secret name, same
  `createOrUpdate` idempotency, same owner ref).
- The new function runs before `reconcilePlatformDefaultSettings()` so it is
  platform-agnostic. For KubeVirt clusters, if the new code generates a key,
  the existing KubeVirt code sees `secretEncryption` already set and skips.
- Existing clusters with the misconfiguration will get encryption enabled on
  the next reconcile — this is the intended behavior and what customers expect.

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Auto-generates a 32-byte opaque AESCBC etcd encryption secret, sets it as the active key, and updates secretEncryption to AESCBC when API server encryption is AESCBC and secretEncryption is unset; preserves existing secretEncryption.

* **Validation**
  * Rejects AESGCM as unsupported. Requires secretEncryption to be configured for KMS and at least one platform-specific KMS provider (AWS/Azure/IBM Cloud).

* **Tests**
  * Added unit tests for reconciliation and validation of API server encryption.

### DIFF
--- a/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
+++ b/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
@@ -4883,54 +4883,8 @@ func (r *HostedClusterReconciler) reconcileKubevirtPlatformDefaultSettings(ctx c
 		}
 	}
 
-	if hc.Spec.SecretEncryption == nil ||
-		len(hc.Spec.SecretEncryption.Type) == 0 ||
-		(hc.Spec.SecretEncryption.Type == hyperv1.AESCBC &&
-			(hc.Spec.SecretEncryption.AESCBC == nil || len(hc.Spec.SecretEncryption.AESCBC.ActiveKey.Name) == 0)) {
-
-		logger.Info("no etcd encryption key configuration found; adding", "hostedCluster name", hc.Name, "hostedCluster namespace", hc.Namespace)
-		etcdEncSec := &corev1.Secret{
-			ObjectMeta: metav1.ObjectMeta{
-				Namespace: hc.Namespace,
-				Name:      hc.Name + etcdEncKeyPostfix,
-			},
-		}
-
-		_, err := createOrUpdate(ctx, r.Client, etcdEncSec, func() error {
-			// don't override existing key just in case something weird happened
-			_, exists := etcdEncSec.Data[hyperv1.AESCBCKeySecretKey]
-			if exists {
-				return nil
-			}
-
-			generatedKey := make([]byte, 32)
-			_, err := rand.Read(generatedKey)
-			if err != nil {
-				return fmt.Errorf("failed to generate the etcd encryption key; %w", err)
-			}
-
-			if etcdEncSec.Data == nil {
-				etcdEncSec.Data = map[string][]byte{}
-			}
-			etcdEncSec.Data[hyperv1.AESCBCKeySecretKey] = generatedKey
-			etcdEncSec.Type = corev1.SecretTypeOpaque
-
-			ownerRef := config.OwnerRefFrom(hc)
-			ownerRef.ApplyTo(etcdEncSec)
-			return nil
-		})
-		if err != nil {
-			return fmt.Errorf("failed to create ETCD SecretEncryption key for KubeVirt platform HostedCluster: %w", err)
-		}
-
-		hc.Spec.SecretEncryption = &hyperv1.SecretEncryptionSpec{
-			Type: hyperv1.AESCBC,
-			AESCBC: &hyperv1.AESCBCSpec{
-				ActiveKey: corev1.LocalObjectReference{
-					Name: etcdEncSec.Name,
-				},
-			},
-		}
+	if err := r.ensureAESCBCEncryptionKey(ctx, hc, createOrUpdate, logger); err != nil {
+		return fmt.Errorf("KubeVirt platform: %w", err)
 	}
 
 	// Reconcile management infrastructure annotation
@@ -4967,11 +4921,28 @@ func (r *HostedClusterReconciler) reconcileAPIServerEncryptionConfig(ctx context
 	if hc.Spec.Configuration.APIServer.Encryption.Type != configv1.EncryptionTypeAESCBC {
 		return nil
 	}
-	if hc.Spec.SecretEncryption != nil && len(hc.Spec.SecretEncryption.Type) > 0 {
+	if hc.Spec.SecretEncryption != nil && len(hc.Spec.SecretEncryption.Type) > 0 &&
+		hc.Spec.SecretEncryption.Type != hyperv1.AESCBC {
+		logger.Info("configuration.apiServer.encryption.type is aescbc but secretEncryption is already set with a different type; secretEncryption takes precedence",
+			"apiServerEncryptionType", hc.Spec.Configuration.APIServer.Encryption.Type,
+			"secretEncryptionType", hc.Spec.SecretEncryption.Type,
+		)
+		return nil
+	}
+	if err := r.ensureAESCBCEncryptionKey(ctx, hc, createOrUpdate, logger); err != nil {
+		return fmt.Errorf("apiServer encryption config: %w", err)
+	}
+	return nil
+}
+
+func (r *HostedClusterReconciler) ensureAESCBCEncryptionKey(ctx context.Context, hc *hyperv1.HostedCluster, createOrUpdate upsert.CreateOrUpdateFN, logger logr.Logger) error {
+	if hc.Spec.SecretEncryption != nil && len(hc.Spec.SecretEncryption.Type) > 0 &&
+		!(hc.Spec.SecretEncryption.Type == hyperv1.AESCBC &&
+			(hc.Spec.SecretEncryption.AESCBC == nil || len(hc.Spec.SecretEncryption.AESCBC.ActiveKey.Name) == 0)) {
 		return nil
 	}
 
-	logger.Info("configuration.apiServer.encryption.type is aescbc but secretEncryption is not set; auto-generating key", "hostedCluster", client.ObjectKeyFromObject(hc))
+	logger.Info("no etcd encryption key configuration found; auto-generating", "hostedCluster", client.ObjectKeyFromObject(hc))
 	etcdEncSec := &corev1.Secret{
 		ObjectMeta: metav1.ObjectMeta{
 			Namespace: hc.Namespace,
@@ -4980,16 +4951,18 @@ func (r *HostedClusterReconciler) reconcileAPIServerEncryptionConfig(ctx context
 	}
 
 	_, err := createOrUpdate(ctx, r.Client, etcdEncSec, func() error {
+		_, exists := etcdEncSec.Data[hyperv1.AESCBCKeySecretKey]
+		if exists {
+			return nil
+		}
+		generatedKey := make([]byte, 32)
+		if _, err := rand.Read(generatedKey); err != nil {
+			return fmt.Errorf("failed to generate the etcd encryption key: %w", err)
+		}
 		if etcdEncSec.Data == nil {
 			etcdEncSec.Data = map[string][]byte{}
 		}
-		if _, exists := etcdEncSec.Data[hyperv1.AESCBCKeySecretKey]; !exists {
-			generatedKey := make([]byte, 32)
-			if _, err := rand.Read(generatedKey); err != nil {
-				return fmt.Errorf("failed to generate the etcd encryption key: %w", err)
-			}
-			etcdEncSec.Data[hyperv1.AESCBCKeySecretKey] = generatedKey
-		}
+		etcdEncSec.Data[hyperv1.AESCBCKeySecretKey] = generatedKey
 		etcdEncSec.Type = corev1.SecretTypeOpaque
 
 		ownerRef := config.OwnerRefFrom(hc)
@@ -4997,7 +4970,7 @@ func (r *HostedClusterReconciler) reconcileAPIServerEncryptionConfig(ctx context
 		return nil
 	})
 	if err != nil {
-		return fmt.Errorf("failed to create ETCD SecretEncryption key from apiServer encryption config: %w", err)
+		return fmt.Errorf("failed to create ETCD SecretEncryption key: %w", err)
 	}
 
 	hc.Spec.SecretEncryption = &hyperv1.SecretEncryptionSpec{
@@ -5008,7 +4981,6 @@ func (r *HostedClusterReconciler) reconcileAPIServerEncryptionConfig(ctx context
 			},
 		},
 	}
-
 	return nil
 }
 
--- a/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
+++ b/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
@@ -633,6 +633,11 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
 
 	createOrUpdate := r.createOrUpdate(req)
 
+	// Reconcile apiServer encryption configuration
+	if err := r.reconcileAPIServerEncryptionConfig(ctx, hcluster, createOrUpdate, log); err != nil {
+		return ctrl.Result{}, err
+	}
+
 	// Reconcile platform defaults
 	if err := r.reconcilePlatformDefaultSettings(ctx, hcluster, createOrUpdate, log); err != nil {
 		return ctrl.Result{}, err
@@ -4057,6 +4062,25 @@ func (r *HostedClusterReconciler) validateOCPConfigurations(ctx context.Context,
 		}
 	}
 
+	if hc.Spec.Configuration != nil && hc.Spec.Configuration.APIServer != nil {
+		encType := hc.Spec.Configuration.APIServer.Encryption.Type
+		encPath := field.NewPath("spec", "configuration", "apiServer", "encryption", "type")
+		switch encType {
+		case configv1.EncryptionTypeAESGCM:
+			errs = append(errs, field.Invalid(encPath, encType, "aesgcm encryption is not supported; use spec.configuration.apiServer.encryption.type aescbc or configure spec.secretEncryption with type aescbc or kms"))
+		case configv1.EncryptionTypeKMS:
+			if hc.Spec.SecretEncryption == nil ||
+				hc.Spec.SecretEncryption.Type != hyperv1.KMS ||
+				hc.Spec.SecretEncryption.KMS == nil {
+				errs = append(errs, field.Invalid(encPath, encType, "KMS encryption requires configuring spec.secretEncryption with platform-specific KMS settings"))
+			} else if hc.Spec.SecretEncryption.KMS.AWS == nil &&
+				hc.Spec.SecretEncryption.KMS.IBMCloud == nil &&
+				hc.Spec.SecretEncryption.KMS.Azure == nil {
+				errs = append(errs, field.Invalid(encPath, encType, "KMS encryption requires at least one platform-specific provider (aws, azure, or ibmcloud) in spec.secretEncryption.kms"))
+			}
+		}
+	}
+
 	return errs.ToAggregate()
 }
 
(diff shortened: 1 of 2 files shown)


vismishr:
@coderabbitai review

vismishr:
@coderabbitai full review

vismishr:
@coderabbitai full review

vismishr:
@coderabbitai full review

hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
```diff
@@ -6569,3 +6569,321 @@ func TestComputeEndpointServiceCondition(t *testing.T) {
...
+							},
+						},
+					},
+					SecretEncryption: &hyperv1.SecretEncryptionSpec{
+						Type: hyperv1.AESCBC,
+						AESCBC: &hyperv1.AESCBCSpec{
+							ActiveKey: corev1.LocalObjectReference{
+								Name: "custom-key",
+							},
+						},
```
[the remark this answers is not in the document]
vismishr: @coderabbitai full review

vismishr:
@coderabbitai full review

vismishr:
@coderabbitai full review

vismishr:
@coderabbitai full review

vismishr:
@coderabbitai full review

vismishr:
@coderabbitai full review

vismishr:
@coderabbitai full review

vismishr:
@coderabbitai full review

vismishr:
@coderabbitai full review

vismishr:
@coderabbitai full review

vismishr:
@coderabbitai full review

devguyio (commented):
## Code Review Report

### PR Info
| Field | Value |
|-------|-------|
| **Title** | OCPBUGS-56376: Honor spec.configuration.apiServer.encryption in HostedCluster |
| **Author** | @vismishr |
| **Base** | `main` ← `fix/honor-apiserver-encryption-config` |
| **Files** | 2 changed (+457) |

### Files Reviewed
| File | Type |
|------|------|
| `hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go` | Source (+80) |
| `hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go` | Test (+377) |

### Unit Test Coverage

Tests cover the main paths well — aescbc key generation, no-ops for nil/identity/already-set cases, and validation error cases. However, there are gaps:

- **Missing partial-configuration test** (see inline): No test covers `secretEncryption.type = aescbc` with `aescbc == nil` or `activeKey.name == ""`. This maps directly to the guard condition bug.
- **Missing idempotency test**: No test verifies that a second call to `reconcileAPIServerEncryptionConfig` (after a key was already generated) is a clean no-op.
- **Missing envtest tests**: If validation moves to CEL (see SOLID section), it needs envtest coverage under `cmd/install/assets/crds/hypershift-operator/tests/hostedclusters.hypershift.openshift.io/`. The framework is YAML-driven and runs across Kubernetes 1.31–1.35 automatically.

### Idiomatic Go Code

- **Bug — Guard condition weaker than KubeVirt equivalent** (see inline at line 4970): The guard that skips reconciliation when `secretEncryption` is set misses the broken state where `type = aescbc` but `aescbc` is nil or `activeKey.name` is empty. The existing KubeVirt code at line 4886-4889 handles this — this function should match.
- **UX — Silent skip on type mismatch** (see inline at line 4972): When the user sets `apiServer.encryption.type: aescbc` but `secretEncryption` is already configured with a different type (e.g., `kms`), the function silently skips. A log warning for the mismatch would help users understand that their OCP-level config is being ignored.

### DRY Compliance

- **Code duplication with KubeVirt** (see inline at line 4963): Lines 4974-5010 (`reconcileAPIServerEncryptionConfig`) are nearly identical to lines 4891-4933 (`reconcileKubevirtPlatformDefaultSettings`). Same secret name pattern, same 32-byte key generation, same `createOrUpdate` call, same `SecretEncryptionSpec` struct assignment. The duplication has already caused a divergence in guard logic.

### SOLID Compliance

- **SRP — Validation in wrong layer** (see inline at line 4065): The new validation (reject `aesgcm`, require KMS config) is purely declarative — no runtime state, API calls, or secret lookups. This belongs as CEL XValidation markers on the API types (admission-time rejection) rather than Go-side controller validation (post-persist status conditions). The codebase already has ~20 CEL rules on `HostedClusterSpec` and a TODO at line 3698 flagging the migration away from Go-side validation.

### Build Verification

Skipped — review based on diff analysis.

### Overall Verdict

**PASS WITH RECOMMENDATIONS**

The core functionality is correct and solves a real user problem. The issues below should be addressed before merging.

### Required Actions

1. **Guard condition bug**: Match the KubeVirt guard to handle partial `secretEncryption` configuration (line 4970)
2. **CEL validation**: Move declarative validation to CEL XValidation markers on the API types, with envtest coverage
3. **Extract shared helper**: Deduplicate key generation logic between this function and `reconcileKubevirtPlatformDefaultSettings`

### Recommended Improvements

1. Add a log warning when `apiServer.encryption.type` conflicts with existing `secretEncryption.type`
2. Add test case for partial `secretEncryption` (type set but AESCBC config missing)
3. Add idempotency test (second reconcile call is a no-op)

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -4057,6 +4062,25 @@ func (r *HostedClusterReconciler) validateOCPConfigurations(ctx context.Context,
...
+				hc.Spec.SecretEncryption.Type != hyperv1.KMS ||
+				hc.Spec.SecretEncryption.KMS == nil {
+				errs = append(errs, field.Invalid(encPath, encType, "KMS encryption requires configuring spec.secretEncryption with platform-specific KMS settings"))
+			} else if hc.Spec.SecretEncryption.KMS.AWS == nil &&
+				hc.Spec.SecretEncryption.KMS.IBMCloud == nil &&
+				hc.Spec.SecretEncryption.KMS.Azure == nil {
+				errs = append(errs, field.Invalid(encPath, encType, "KMS encryption requires at least one platform-specific provider (aws, azure, or ibmcloud) in spec.secretEncryption.kms"))
+			}
+		}
+	}
```
devguyio, line 4082: **SOLID / Architecture** — This validation is purely declarative (no runtime state needed) and would be better expressed as CEL XValidation markers on `HostedClusterSpec` in `api/hypershift/v1beta1/hostedcluster_types.go`.

The codebase already has ~20 CEL rules on `HostedClusterSpec`, including cross-field rules referencing `self.configuration.apiServer` (line 527). There's also a TODO at line 3698 flagging the migration away from Go-side validation.

CEL would reject invalid input at admission time instead of letting the object persist and surfacing errors later as status conditions. For example:

```go
// +kubebuilder:validation:XValidation:rule="!has(self.configuration) || !has(self.configuration.apiServer) || self.configuration.apiServer.encryption.type != 'aesgcm'",message="aesgcm encryption is not supported; use aescbc or configure secretEncryption with kms"
```

The KMS cross-field check would be similar, validating `self.secretEncryption` when `self.configuration.apiServer.encryption.type == 'KMS'`.

If the validation moves to CEL, it should also be tested via the envtest framework — see `cmd/install/assets/crds/hypershift-operator/tests/hostedclusters.hypershift.openshift.io/stable.hostedclusters.validation.testsuite.yaml` for the format. A new `stable.hostedclusters.encryption.testsuite.yaml` would cover admission-time rejection of unsupported encryption types and missing KMS configuration, tested across multiple Kubernetes versions automatically.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -4932,6 +4956,62 @@ func (r *HostedClusterReconciler) reconcileKubevirtPlatformDefaultSettings(ctx c
...
+// sets encryption.type to aescbc without configuring secretEncryption directly.
+// If secretEncryption is already set, it takes precedence (override power).
+func (r *HostedClusterReconciler) reconcileAPIServerEncryptionConfig(ctx context.Context, hc *hyperv1.HostedCluster, createOrUpdate upsert.CreateOrUpdateFN, logger logr.Logger) error {
+	if hc.Spec.Configuration == nil || hc.Spec.Configuration.APIServer == nil {
+		return nil
+	}
+	if hc.Spec.Configuration.APIServer.Encryption.Type != configv1.EncryptionTypeAESCBC {
+		return nil
+	}
+	if hc.Spec.SecretEncryption != nil && len(hc.Spec.SecretEncryption.Type) > 0 {
```
devguyio: **Bug** — The guard here is weaker than the existing KubeVirt equivalent at line 4886-4889.

The KubeVirt code also catches the case where `secretEncryption.type == AESCBC` but `aescbc == nil` or `activeKey.name == ""` — a partially-configured state that leaves encryption broken. This function would skip that case and not generate a key.

Should the guard match KubeVirt's behavior? e.g.:

```go
if hc.Spec.SecretEncryption != nil && len(hc.Spec.SecretEncryption.Type) > 0 &&
    !(hc.Spec.SecretEncryption.Type == hyperv1.AESCBC &&
        (hc.Spec.SecretEncryption.AESCBC == nil || len(hc.Spec.SecretEncryption.AESCBC.ActiveKey.Name) == 0)) {
    return nil
}
```

Or better, extract the shared logic into a helper to avoid the divergence entirely (see next comment).

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -4932,6 +4956,62 @@ func (r *HostedClusterReconciler) reconcileKubevirtPlatformDefaultSettings(ctx c
 	return nil
 }
 
+// reconcileAPIServerEncryptionConfig honors spec.configuration.apiServer.encryption
+// by auto-generating an AESCBC key and populating spec.secretEncryption when the user
+// sets encryption.type to aescbc without configuring secretEncryption directly.
+// If secretEncryption is already set, it takes precedence (override power).
+func (r *HostedClusterReconciler) reconcileAPIServerEncryptionConfig(ctx context.Context, hc *hyperv1.HostedCluster, createOrUpdate upsert.CreateOrUpdateFN, logger logr.Logger) error {
```
devguyio, line 4917: **DRY** — This function duplicates the key generation and secretEncryption population logic from `reconcileKubevirtPlatformDefaultSettings` (lines 4891-4933). Same secret name pattern, same 32-byte key generation, same `createOrUpdate` call, same `SecretEncryptionSpec` struct assignment.

The duplication has already caused a divergence — the guard condition here (line 4970) is weaker than the KubeVirt version (line 4886-4889).

Consider extracting the shared logic into a helper, e.g.:

```go
func (r *HostedClusterReconciler) ensureAESCBCEncryptionKey(
    ctx context.Context,
    hc *hyperv1.HostedCluster,
    createOrUpdate upsert.CreateOrUpdateFN,
    logger logr.Logger,
) error
```

Both `reconcileAPIServerEncryptionConfig` and `reconcileKubevirtPlatformDefaultSettings` would call it after their respective trigger checks. This eliminates the duplication and prevents future guard divergence.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -4932,6 +4956,62 @@ func (r *HostedClusterReconciler) reconcileKubevirtPlatformDefaultSettings(ctx c
...
+	if hc.Spec.Configuration == nil || hc.Spec.Configuration.APIServer == nil {
+		return nil
+	}
+	if hc.Spec.Configuration.APIServer.Encryption.Type != configv1.EncryptionTypeAESCBC {
+		return nil
+	}
+	if hc.Spec.SecretEncryption != nil && len(hc.Spec.SecretEncryption.Type) > 0 {
+		return nil
+	}
```
devguyio, line 4943: **nit / UX** — When `apiServer.encryption.type` is set to `aescbc` but `secretEncryption` is already configured with a different type (e.g., `kms`), this silently skips. The user has explicitly set an encryption config that's being ignored.

Consider adding a log warning here for the mismatch case, e.g.:

```go
if hc.Spec.SecretEncryption != nil && len(hc.Spec.SecretEncryption.Type) > 0 {
    if hc.Spec.SecretEncryption.Type != hyperv1.AESCBC {
        logger.Info(
            "configuration.apiServer.encryption.type is aescbc but secretEncryption is already set with a different type; secretEncryption takes precedence",
            "apiServerEncryptionType", hc.Spec.Configuration.APIServer.Encryption.Type,
            "secretEncryptionType", hc.Spec.SecretEncryption.Type,
        )
    }
    return nil
}
```

hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
```diff
@@ -6569,3 +6569,380 @@ func TestComputeEndpointServiceCondition(t *testing.T) {
 		})
 	}
 }
+
+func TestReconcileAPIServerEncryptionConfig(t *testing.T) {
```
devguyio, line 6573: **Test gap** — Two missing test cases for this suite:

1. **Partial `secretEncryption`**: `secretEncryption` is set with `type: aescbc` but `aescbc` is nil or `activeKey.name` is empty. This is the partial-configuration scenario from the guard condition discussion — the test would document the expected behavior for that edge case.

2. **Idempotency**: Run `reconcileAPIServerEncryptionConfig` twice and verify the second call doesn't regenerate the key or modify the secret. The `createOrUpdate` machinery handles this, but an explicit test documents the contract.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
vismishr: Addressed in f2f6156fa.

Extracted the AESCBC key generation into a shared `ensureAESCBCEncryptionKey()` helper that uses the KubeVirt-equivalent guard condition. The helper's guard is:

```go
if hc.Spec.SecretEncryption != nil && len(hc.Spec.SecretEncryption.Type) > 0 &&
    !(hc.Spec.SecretEncryption.Type == hyperv1.AESCBC &&
        (hc.Spec.SecretEncryption.AESCBC == nil || len(hc.Spec.SecretEncryption.AESCBC.ActiveKey.Name) == 0)) {
    return nil
}
```

This now correctly catches the partially-configured state where `secretEncryption.type == AESCBC` but `aescbc` is nil or `activeKey.name` is empty — instead of silently skipping, it proceeds with key generation, matching the existing KubeVirt behavior at L4886-4889.

In `reconcileAPIServerEncryptionConfig`, the flow before calling the helper is:
1. Early return if `Configuration` or `APIServer` is nil
2. Early return if `apiServer.encryption.type != aescbc`
3. If `secretEncryption` is set with a non-AESCBC type (e.g. `kms`), log a warning and return (see UX comment)
4. Otherwise delegate to `ensureAESCBCEncryptionKey` which handles: nil secretEncryption, empty type, and the partial AESCBC case

Added two test cases covering this fix:
- `"When secretEncryption type is aescbc but aescbc config is nil, it should generate key and set secretEncryption"`
- `"When secretEncryption type is aescbc but activeKey name is empty, it should generate key and set secretEncryption"`

Both verify that key generation proceeds and `secretEncryption` is correctly populated despite the partial configuration.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
vismishr: Addressed in f2f6156fa.

Extracted the shared AESCBC key generation logic into `ensureAESCBCEncryptionKey(ctx, hc, createOrUpdate, logger)` — matching the exact signature you suggested. Both callers now delegate to this single helper:

**`reconcileAPIServerEncryptionConfig`** handles its own trigger checks (nil Configuration, non-AESCBC type, type mismatch warning) and then calls:
```go
if err := r.ensureAESCBCEncryptionKey(ctx, hc, createOrUpdate, logger); err != nil {
    return fmt.Errorf("apiServer encryption config: %w", err)
}
```

**`reconcileKubevirtPlatformDefaultSettings`** had its entire 48-line AESCBC block (L4886-4934) replaced with:
```go
if err := r.ensureAESCBCEncryptionKey(ctx, hc, createOrUpdate, logger); err != nil {
    return fmt.Errorf("KubeVirt platform: %w", err)
}
```

The rest of `reconcileKubevirtPlatformDefaultSettings` (DNS baseDomain autodetection, management infrastructure annotation reconciliation) is unchanged.

The helper contains the KubeVirt-equivalent guard condition, so both code paths now share the same stronger guard that catches partially-configured AESCBC state. This eliminates the divergence you identified — any future guard fix applies to both callers automatically.

Error context is preserved via Go-convention error wrapping at each call site rather than a parameter, so logs still distinguish whether the failure originated from apiServer encryption config or KubeVirt platform defaults.

All existing `TestKubevirtETCDEncKey` tests (7 cases, exercising the full `r.Reconcile()` loop) continue to pass — the KubeVirt test cases don't set `configuration.apiServer`, so `reconcileAPIServerEncryptionConfig` returns nil early and doesn't interfere.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
vismishr: Addressed in f2f6156fa.

Added a log warning in `reconcileAPIServerEncryptionConfig` that fires when the user sets `apiServer.encryption.type=aescbc` but `secretEncryption` is already configured with a different type. The check runs before the `ensureAESCBCEncryptionKey` call:

```go
if hc.Spec.SecretEncryption != nil && len(hc.Spec.SecretEncryption.Type) > 0 &&
    hc.Spec.SecretEncryption.Type != hyperv1.AESCBC {
    logger.Info(
        "configuration.apiServer.encryption.type is aescbc but secretEncryption is already set with a different type; secretEncryption takes precedence",
        "apiServerEncryptionType", hc.Spec.Configuration.APIServer.Encryption.Type,
        "secretEncryptionType", hc.Spec.SecretEncryption.Type,
    )
    return nil
}
```

This gives operators visibility into the override — the log includes both the apiServer encryption type and the existing secretEncryption type so the mismatch is immediately clear. The function still returns nil (secretEncryption takes precedence as designed), but now the user has a signal that their `apiServer.encryption` config is being ignored.

Added a test case `"When apiServer encryption is aescbc but secretEncryption is kms, it should not override and preserve secretEncryption"` that verifies:
- No secret is created (key generation is skipped)
- The existing KMS secretEncryption is preserved untouched
- The reconciler returns no error

hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
vismishr: Both test gaps addressed in f2f6156fa.

**1. Partial `secretEncryption` tests:**

Added two test cases covering the partially-configured state:

- `"When secretEncryption type is aescbc but aescbc config is nil, it should generate key and set secretEncryption"` — sets `secretEncryption: {type: aescbc}` with no AESCBC block. Verifies that `ensureAESCBCEncryptionKey` detects the incomplete config via the guard condition `hc.Spec.SecretEncryption.AESCBC == nil`, proceeds with key generation, creates the 32-byte secret, and populates the full `secretEncryption` spec.

- `"When secretEncryption type is aescbc but activeKey name is empty, it should generate key and set secretEncryption"` — sets `secretEncryption: {type: aescbc, aescbc: {activeKey: {name: ""}}}`. Verifies the guard catches `len(hc.Spec.SecretEncryption.AESCBC.ActiveKey.Name) == 0` and generates the key rather than leaving encryption broken with an empty reference.

Both cases assert `expectSecretCreated: true` (secret exists with 32-byte key) and `expectSecretEncryptionSet: true` (secretEncryption populated with the generated key name).

**2. Idempotency test:**

Added `TestReconcileAPIServerEncryptionConfig_Idempotency` as a separate test function:

1. Creates a HostedCluster with `apiServer.encryption.type=aescbc` and nil `secretEncryption`
2. Calls `reconcileAPIServerEncryptionConfig` — key is generated, secret is created
3. Captures the generated key bytes via `copy(firstKey, secret.Data[hyperv1.AESCBCKeySecretKey])`
4. Calls `reconcileAPIServerEncryptionConfig` a second time
5. Asserts: no error, and `secret.Data[hyperv1.AESCBCKeySecretKey]` equals `firstKey` (byte-for-byte identical)

This works because after the first call, `hc.Spec.SecretEncryption` is fully populated (type=AESCBC, AESCBC.ActiveKey.Name set), so the second call hits the `ensureAESCBCEncryptionKey` guard and returns nil without touching the secret. The `createOrUpdate` inner function also has an early return when the key already exists (`if exists { return nil }`), providing a second layer of idempotency protection.

PR #8490, opened 2026-05-12
NO-JIRA: Fix Azure private/topology CEL validation rules

enxebre:
## Description

Fix two CEL validation gaps on Azure `AzurePlatformSpec` and `AzurePrivateSpec`:

1. **`private` was settable without `topology`**: The existing rule `!has(self.topology) || (...)` short-circuited to `true` when `topology` was omitted, allowing `private` to be set without `topology`. Fixed to `has(self.topology) && (...) ? has(self.private) : !has(self.private)` — now correctly forbids `private` when `topology` is absent or `Public`.

2. **`privateLink` struct not required when `type` is `PrivateLink`**: Only the negative constraint existed (forbid `privateLink` when type is not PrivateLink). Added `self.type != 'PrivateLink' || has(self.privateLink)` to require the struct.

## Changes

- `api/hypershift/v1beta1/azure.go`: Fix topology/private CEL rule, add privateLink requirement rule
- Envtest cases: added "private without topology should fail", "private with Public topology should fail", "PrivateLink without privateLink config should fail"; fixed existing tests to include `privateLink` struct

## Test plan

- [x] `make test-envtest-ocp` — all 630 specs pass

## Summary by CodeRabbit

* **Bug Fixes**
  * Azure topology validation tightened: private configuration is now consistently required for Private or PublicAndPrivate topologies and forbidden otherwise, preventing ambiguous or missing private settings.
  * PrivateLink validation strengthened: PrivateLink-specific settings are now required when PrivateLink is selected and disallowed for other types, reducing misconfiguration risk.

### DIFF
--- a/api/hypershift/v1beta1/azure.go
+++ b/api/hypershift/v1beta1/azure.go
@@ -668,8 +668,7 @@ const (
 // mechanism. Currently only PrivateLink is supported; additional mechanisms (e.g., Swift) may
 // be added in the future.
 //
-// +kubebuilder:validation:XValidation:rule="self.type != 'PrivateLink' ? !has(self.privateLink) : true",message="privateLink is forbidden when type is not PrivateLink"
-// +kubebuilder:validation:XValidation:rule="self.type != 'PrivateLink' || has(self.privateLink)",message="privateLink is required when type is PrivateLink"
+// +kubebuilder:validation:XValidation:rule="self.type == 'PrivateLink' ? has(self.privateLink) : !has(self.privateLink)",message="privateLink is required when type is PrivateLink, and forbidden otherwise"
 // +union
 type AzurePrivateSpec struct {
 	// type specifies the private connectivity mechanism used for the hosted cluster's API server.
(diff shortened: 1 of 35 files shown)


bryan-cox (approved):
/lgtm

bryan-cox (approved):
/lgtm

api/hypershift/v1beta1/azure.go
```diff
@@ -669,6 +669,7 @@ const (
 // be added in the future.
 //
 // +kubebuilder:validation:XValidation:rule="self.type != 'PrivateLink' ? !has(self.privateLink) : true",message="privateLink is forbidden when type is not PrivateLink"
+// +kubebuilder:validation:XValidation:rule="self.type != 'PrivateLink' || has(self.privateLink)",message="privateLink is required when type is PrivateLink"
 // +union
```
muraee, line 672: can't those 2 rules be combined?
```
// +kubebuilder:validation:XValidation:rule="(self.type == 'PrivateLink') == has(self.privateLink)",message="privateLink must be set if and only if type is PrivateLink"
```

enxebre: sgtm, @JoelSpeed is there any preference / convention for this? what's the impact on budget?

JoelSpeed: The one we typically use for this is
```
// +kubebuilder:validation:XValidation:rule="self.type == 'PrivateLink' ? has(self.privateLink) : !has(self.privateLink)",message="privateLink is required when type is PrivateLink, and forbidden otherwise"
```

PR #8494, opened 2026-05-12
CNTRLPLANE-3329: Replace actions/cache with EFS-backed build cache in unit tests

vismishr:

`actions/cache` adds ~2-3.5 min overhead per shard (upload/download) on every
CI run. This PR replaces the `actions/cache` step in `test-reusable.yaml` with
a conditional copy from the EFS-backed PV mount at `/cache/go-build`.

The cache is copied to a local tmpdir (`/tmp/go-build-cache`) at job start,
which is already set as `GOCACHE` in the workflow. If the EFS mount is not
present (e.g. before the Helm values PR is merged, or during mount issues),
the step silently skips and CI runs without a cache — same as today.

- Depends on PR #8493 (Helm values update to mount the PVC on runner pods).
  However, because the cache step is conditional (`if [ -d /cache/go-build ]`),
  this PR can be merged in any order without breaking CI.
- The `GOCACHE` and `GOMODCACHE` env vars at the job level are unchanged.
- Follow-up PR will extend EFS caching to lint, verify, and envtest workflows.

## Checklist:

## Summary by CodeRabbit

* **Chores**
  * CI now pre-warms a local Go build cache via a dedicated warming step before running tests, replacing the previous direct cache restore to improve reliability and speed of unit-test runs.
  * Test sharding and test-reporting behavior remain unchanged; overall effect is faster, more stable CI unit-test execution.

### DIFF
--- a/.github/actions/warm-go-cache/action.yaml
+++ b/.github/actions/warm-go-cache/action.yaml
@@ -8,6 +8,6 @@ runs:
         echo "GOCACHE=/tmp/go-build-cache" >> "$GITHUB_ENV"
         if [ -d /cache/go-build ]; then
           mkdir -p /tmp/go-build-cache && \
-          cp -a /cache/go-build/. /tmp/go-build-cache/ || \
+          timeout 120 cp -a /cache/go-build/. /tmp/go-build-cache/ || \
           echo "Warning: failed to copy EFS cache, proceeding without cache"
         fi
--- a/.github/actions/warm-go-cache/action.yaml
+++ b/.github/actions/warm-go-cache/action.yaml
@@ -1,0 +1,13 @@
+name: 'Warm Go build cache'
+description: 'Set GOCACHE and optionally warm from EFS-backed PV'
+runs:
+  using: composite
+  steps:
+    - shell: bash
+      run: |
+        echo "GOCACHE=/tmp/go-build-cache" >> "$GITHUB_ENV"
+        if [ -d /cache/go-build ]; then
+          mkdir -p /tmp/go-build-cache && \
+          cp -a /cache/go-build/. /tmp/go-build-cache/ || \
+          echo "Warning: failed to copy EFS cache, proceeding without cache"
+        fi
--- a/.github/workflows/test-reusable.yaml
+++ b/.github/workflows/test-reusable.yaml
@@ -86,12 +86,11 @@ jobs:
         with:
           go-version-file: go.mod
           cache: false
-      - name: Restore Go build cache
-        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
-        with:
-          path: /tmp/go-build-cache
-          key: go-build-${{ matrix.shard }}-${{ hashFiles('go.mod') }}-${{ github.sha }}
-          restore-keys: go-build-${{ matrix.shard }}-${{ hashFiles('go.mod') }}-
+      - name: Warm Go build cache from EFS
+        run: |
+          if [ -d /cache/go-build ]; then
+            cp -a /cache/go-build /tmp/go-build-cache
+          fi
       - name: Run tests
         run: make test-shard TEST_PACKAGES="${{ matrix.packages }}" COVER_PROFILE="cover-${{ matrix.shard }}.out"
       - name: Upload to Codecov
--- a/.github/workflows/test-reusable.yaml
+++ b/.github/workflows/test-reusable.yaml
@@ -73,7 +73,6 @@ jobs:
           - shard: other
             packages: ./karpenter-operator/... ./control-plane-pki-operator/... ./contrib/... ./ignition-server/... ./pkg/... ./dnsresolver/... ./product-cli/... ./client/... ./test/integration/... ./test/e2e/util/... ./test/util/... ./availability-prober/... ./konnectivity-socks5-proxy/... ./konnectivity-https-proxy/... ./kubernetes-default-proxy/... ./kubevirtexternalinfra/... ./etcd-defrag/... ./etcd-backup/... ./etcd-recovery/... ./etcd-upload/... ./kas-bootstrap/... ./sharedingress-config-generator/... ./sync-fg-configmap/... ./sync-global-pullsecret/... ./token-minter/...
     env:
-      GOCACHE: /tmp/go-build-cache
       GOMODCACHE: /tmp/go-mod-cache
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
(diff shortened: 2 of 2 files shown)


.github/workflows/test-reusable.yaml
```diff
@@ -86,12 +86,11 @@ jobs:
...
-        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
-        with:
-          path: /tmp/go-build-cache
-          key: go-build-${{ matrix.shard }}-${{ hashFiles('go.mod') }}-${{ github.sha }}
-          restore-keys: go-build-${{ matrix.shard }}-${{ hashFiles('go.mod') }}-
+      - name: Warm Go build cache from EFS
+        run: |
+          if [ -d /cache/go-build ]; then
+            cp -a /cache/go-build /tmp/go-build-cache
+          fi
```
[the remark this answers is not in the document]
vismishr: Fixed — updated to `mkdir -p /tmp/go-build-cache && cp -a /cache/go-build/. /tmp/go-build-cache/` so contents are copied regardless of whether the destination exists.

vismishr:
@coderabbitai full review

.github/workflows/test-reusable.yaml
```diff
@@ -86,12 +86,12 @@ jobs:
...
-        with:
-          path: /tmp/go-build-cache
-          key: go-build-${{ matrix.shard }}-${{ hashFiles('go.mod') }}-${{ github.sha }}
-          restore-keys: go-build-${{ matrix.shard }}-${{ hashFiles('go.mod') }}-
+      - name: Warm Go build cache from EFS
+        run: |
+          if [ -d /cache/go-build ]; then
+            mkdir -p /tmp/go-build-cache
+            cp -a /cache/go-build/. /tmp/go-build-cache/
+          fi
```
cblecker: +1 to CodeRabbit's suggestion here — if the EFS mount exists but the copy fails (stale NFS handle, disk full in `/tmp`, I/O timeout), this will take down the whole job. Since the stated intent is graceful degradation, wrapping the copy so failures are non-fatal would close that gap:

```yaml
      - name: Warm Go build cache from EFS
        run: |
          if [ -d /cache/go-build ]; then
            mkdir -p /tmp/go-build-cache && \
            cp -a /cache/go-build/. /tmp/go-build-cache/ || \
            echo "Warning: failed to copy EFS cache, proceeding without cache"
          fi
```

Same applies to all 4 files in #8495.

.github/workflows/test-reusable.yaml
```diff
@@ -86,12 +86,12 @@ jobs:
         with:
           go-version-file: go.mod
           cache: false
-      - name: Restore Go build cache
-        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
-        with:
-          path: /tmp/go-build-cache
-          key: go-build-${{ matrix.shard }}-${{ hashFiles('go.mod') }}-${{ github.sha }}
-          restore-keys: go-build-${{ matrix.shard }}-${{ hashFiles('go.mod') }}-
+      - name: Warm Go build cache from EFS
```
cblecker: Looking at this together with #8495 — this exact block (plus the `GOCACHE` env var) ends up copy-pasted into 5 workflow files. Worth extracting into a composite action so there's a single source of truth?

Something like `.github/actions/warm-go-cache/action.yaml`:

```yaml
name: 'Warm Go build cache'
description: 'Set GOCACHE and optionally warm from EFS-backed PV'
runs:
  using: composite
  steps:
    - shell: bash
      run: |
        echo "GOCACHE=/tmp/go-build-cache" >> "$GITHUB_ENV"
        if [ -d /cache/go-build ]; then
          mkdir -p /tmp/go-build-cache && \
          cp -a /cache/go-build/. /tmp/go-build-cache/ || \
          echo "Warning: failed to copy EFS cache, proceeding without cache"
        fi
```

Then each workflow just needs `- uses: ./.github/actions/warm-go-cache` after checkout, and you can drop the job-level `GOCACHE` env vars since `$GITHUB_ENV` makes it available to all subsequent steps. Keeps the error handling centralized too.

.github/workflows/test-reusable.yaml
vismishr: Done — added the error handling in the composite action so it applies to all workflows. Copy failures now log a warning and the job continues without cache.

.github/workflows/test-reusable.yaml
vismishr: Good idea — extracted into `.github/actions/warm-go-cache/action.yaml` as a composite action. It sets `GOCACHE` via `$GITHUB_ENV` and includes the error handling, so each workflow just needs `- uses: ./.github/actions/warm-go-cache` after checkout. Updated both this PR and #8495.

vismishr:
@coderabbitai full review

.github/actions/warm-go-cache/action.yaml
```diff
@@ -0,0 +1,13 @@
...
+description: 'Set GOCACHE and optionally warm from EFS-backed PV'
+runs:
+  using: composite
+  steps:
+    - shell: bash
+      run: |
+        echo "GOCACHE=/tmp/go-build-cache" >> "$GITHUB_ENV"
+        if [ -d /cache/go-build ]; then
+          mkdir -p /tmp/go-build-cache && \
+          cp -a /cache/go-build/. /tmp/go-build-cache/ || \
```
bryan-cox: **suggestion:** Consider wrapping the copy in a `timeout` to guard against stale NFS handles on the EFS mount. If the mount hangs, this `cp` could block until the job-level `timeout-minutes` kills the entire job.

```suggestion
          timeout 120 cp -a /cache/go-build/. /tmp/go-build-cache/ || \
```

.github/workflows/test-reusable.yaml
```diff
@@ -86,12 +85,7 @@ jobs:
         with:
           go-version-file: go.mod
           cache: false
-      - name: Restore Go build cache
-        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
-        with:
-          path: /tmp/go-build-cache
-          key: go-build-${{ matrix.shard }}-${{ hashFiles('go.mod') }}-${{ github.sha }}
-          restore-keys: go-build-${{ matrix.shard }}-${{ hashFiles('go.mod') }}-
+      - uses: ./.github/actions/warm-go-cache
```
bryan-cox, line 89: nit: With `GOCACHE` removed from the job-level `env:` block and now set inside this composite action via `$GITHUB_ENV`, the variable is no longer visible at a glance in the workflow file. If someone adds a Go step before `warm-go-cache` (or reorders steps), it would silently use Go's default cache path instead.

Consider adding a brief comment here or keeping `GOCACHE` in the job-level env as a declarative default that the action can override.

bryan-cox (commented):
Good approach — replacing `actions/cache` with an EFS-backed local copy is a solid optimization for self-hosted runners. A few inline suggestions below.

.github/actions/warm-go-cache/action.yaml
vismishr: Good call — added `timeout 120` to the `cp` in the composite action. Updated in both this PR and #8495.

.github/workflows/test-reusable.yaml
vismishr: Makes sense — restored `GOCACHE: /tmp/go-build-cache` in the job-level `env:` block here and in all 4 workflow files in #8495. The composite action still sets it via `$GITHUB_ENV` as a runtime override, but the declarative default ensures it's always visible and safe against step reordering.

bryan-cox (approved):
/lgtm

PR #8495, opened 2026-05-12
CNTRLPLANE-3329: Extend EFS-backed build cache to lint, verify, and envtest workflows

vismishr:

Extends the EFS-backed Go build cache to all Go-based CI workflows beyond
unit tests. Each workflow now includes a conditional step that copies the
shared cache from `/cache/go-build` to a local tmpdir, setting `GOCACHE`
to use it. This eliminates cold compilation overhead across all CI jobs.

Workflows updated:
- `lint-reusable.yaml`
- `verify-reusable.yaml`
- `envtest-ocp-reusable.yaml`
- `envtest-kube-reusable.yaml`

If the EFS mount is not present, the step silently skips and CI runs
without a cache — no functional change from today's behavior.

- Depends on PR #8493 (Helm values — mount PVC on runner pods) and
  PR #8494 (unit test workflow update). Can be merged in any order due
  to the conditional cache step.
- `docs-build-reusable.yaml` is not included — it uses Python/mkdocs
  and does not benefit from Go build cache.
- The same conditional pattern (`if [ -d /cache/go-build ]`) is used
  consistently across all workflows.

## Checklist:

## Summary by CodeRabbit

* **Chores**
  * Added a reusable job-local Go build cache warming step and set the job cache location to /tmp/go-build-cache.
  * Integrated the cache-warming step into verification, lint, and envtest (Kubernetes and OCP) workflows.
  * Updated cache-warming failure output to emit a GitHub Actions-formatted warning when pre-seeding from shared storage fails.

### DIFF
--- a/.github/actions/warm-go-cache/action.yaml
+++ b/.github/actions/warm-go-cache/action.yaml
@@ -9,5 +9,5 @@ runs:
         if [ -d /cache/go-build ]; then
           mkdir -p /tmp/go-build-cache && \
           timeout 120 cp -a /cache/go-build/. /tmp/go-build-cache/ || \
-          echo "Warning: failed to copy EFS cache, proceeding without cache"
+          echo "::warning::Failed to copy EFS cache, proceeding without cache"
         fi
--- a/.github/workflows/verify-reusable.yaml
+++ b/.github/workflows/verify-reusable.yaml
@@ -11,6 +11,8 @@ jobs:
     name: Verify
     runs-on: arc-runner-set
     timeout-minutes: 60
+    env:
+      GOCACHE: /tmp/go-build-cache
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
--- a/.github/workflows/verify-reusable.yaml
+++ b/.github/workflows/verify-reusable.yaml
@@ -11,10 +11,17 @@ jobs:
     name: Verify
     runs-on: arc-runner-set
     timeout-minutes: 60
+    env:
+      GOCACHE: /tmp/go-build-cache
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           persist-credentials: false
+      - name: Warm Go build cache from EFS
+        run: |
+          if [ -d /cache/go-build ]; then
+            cp -a /cache/go-build /tmp/go-build-cache
+          fi
       - run: make generate update
       - run: make staticcheck
       - run: make fmt
--- a/.github/workflows/verify-reusable.yaml
+++ b/.github/workflows/verify-reusable.yaml
@@ -20,7 +20,8 @@ jobs:
       - name: Warm Go build cache from EFS
         run: |
           if [ -d /cache/go-build ]; then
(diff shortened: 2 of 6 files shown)


vismishr:
@coderabbitai full review

vismishr:
@coderabbitai full review

.github/workflows/verify-reusable.yaml
```diff
@@ -11,10 +11,13 @@ jobs:
     name: Verify
     runs-on: arc-runner-set
     timeout-minutes: 60
+    env:
+      GOCACHE: /tmp/go-build-cache
     steps:
```
bryan-cox, line 14: nit: `GOCACHE` is also set to this same value inside the `warm-go-cache` composite action (line 8 of `action.yaml` writes it to `$GITHUB_ENV`). Having it in both places means if the path ever changes, you'd need to update it here and in all the other workflows plus the action.

Consider picking one source of truth:
- **Option A**: Remove `GOCACHE` from the composite action and keep these job-level `env` blocks (makes the cache location visible at the job level).
- **Option B**: Remove the job-level `env` from all workflows and rely solely on the composite action's `$GITHUB_ENV` export (single place to change).

Same applies to `envtest-kube-reusable.yaml`, `envtest-ocp-reusable.yaml`, `lint-reusable.yaml`, and `test-reusable.yaml`.

.github/actions/warm-go-cache/action.yaml
```diff
@@ -9,5 +9,5 @@ runs:
         if [ -d /cache/go-build ]; then
           mkdir -p /tmp/go-build-cache && \
           timeout 120 cp -a /cache/go-build/. /tmp/go-build-cache/ || \
-          echo "Warning: failed to copy EFS cache, proceeding without cache"
+          echo "::warning::Failed to copy EFS cache, proceeding without cache"
```
bryan-cox, line 12: Nice — `::warning::` surfaces this as a GitHub Actions annotation in the workflow summary UI instead of burying it in the logs. Good change.

bryan-cox (commented):
GHA review: one suggestion on GOCACHE source of truth. The rest looks clean — correct step ordering, consistent pattern, no security issues.

.github/workflows/verify-reusable.yaml
vismishr: Thanks Bryan — went with **Option B** (single source of truth in the composite action).

### What changed

Removed the `GOCACHE: /tmp/go-build-cache` job-level `env` declaration from all 5 workflows:

- `verify-reusable.yaml`
- `lint-reusable.yaml`
- `envtest-ocp-reusable.yaml`
- `envtest-kube-reusable.yaml`
- `test-reusable.yaml`

`GOCACHE` is now only set in `.github/actions/warm-go-cache/action.yaml` via `$GITHUB_ENV` (line 8). If the cache path ever needs to change, it's a one-line update in the composite action — no workflow files to touch.

For `test-reusable.yaml`, the job-level `env` block was kept since it still declares `GOMODCACHE` — only the `GOCACHE` line was removed.

Commit: `65666ea` — `ci(workflows): remove duplicate GOCACHE from job-level env blocks`

bryan-cox (approved):
/lgtm

PR #8497, opened 2026-05-12
OCPBUGS-81544: requeue when AutoNodeEnabled is progressing

maxcao13:

The HostedCluster reconciler reads `ControlPlaneComponent` status to set the
`AutoNodeEnabled` condition, but it does not watch CPC resources. When AutoNode
is enabled or disabled, the condition gets stuck at `AutoNodeProgressing` until
an unrelated resource change happens to trigger reconciliation. In CI this
causes the `TestKarpenter/Main/AutoNode_enable/disable_lifecycle` e2e test to
time out — the karpenter CPC reports `RolloutComplete=True` within ~1 minute
but the HC condition is never updated.

This adds a 15-second requeue when `reconcileAutoNodeEnabledCondition` reports
a progressing state, so the HC reconciler polls until the transition completes.
This is a targeted fix that avoids adding a broad CPC watch (which would fire
for all ~30-40 CPCs on every status change across every HostedCluster).

Fixes periodic `TestKarpenter/Main/AutoNode_enable/disable_lifecycle` failures
in `periodic-ci-openshift-hypershift-release-5.0-periodics-e2e-aws-ovn`.

The alternative considered was adding `ControlPlaneComponent` to the HC
reconciler's `managedResources()` watch list. That would provide immediate
reactivity but at the cost of watching all CPCs across all HCP namespaces
(~30-40 per HostedCluster), when only 2 karpenter CPCs matter during
enable/disable transitions. A bounded 15s poll during transitions is the
better tradeoff.

## Checklist:

Assisted-by: Cursor Agent

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **Improvements**
  * Enhanced AutoNode reconciliation with optimized requeue timing, ensuring the system polls for automatic node management progress more frequently (every 15 seconds maximum).

* **Tests**
  * Updated unit and end-to-end tests to validate AutoNode progress tracking and reconciliation behavior.

### DIFF
--- a/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
+++ b/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
@@ -847,8 +847,8 @@ func (r *HostedClusterReconciler) reconcile(ctx context.Context, req ctrl.Reques
 	propagateControlPlaneVersion(hcluster, hcp)
 
 	// Set the AutoNodeEnabled condition reflecting both spec intent and actual component rollout progress.
-	meta.SetStatusCondition(&hcluster.Status.Conditions,
-		r.reconcileAutoNodeEnabledCondition(ctx, hcluster, controlPlaneNamespace.Name))
+	autoNodeCondition, autoNodeProgressing := r.reconcileAutoNodeEnabledCondition(ctx, hcluster, controlPlaneNamespace.Name)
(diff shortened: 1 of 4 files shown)


maxcao13:
Added a commit to fix some data racing because we ran into concurrent map writes related to the shared hostedcluster pointer object.

/test e2e-aws-autonode
/test e2e-aws

maxcao13:
cc @enxebre @jkyros

PR #8504, opened 2026-05-13
OCPBUGS-85580: Fix webhook TLS failure after service-ca to self-managed cert migration

joshbranham:
## Summary

- During upgrade from service-ca managed certs to self-managed certs, the `WebhookCertReconciler` removed the `serving-cert-secret-name` annotation from the operator Service but did not remove the `service.beta.openshift.io/inject-cabundle` annotation from the `MutatingWebhookConfiguration`
- This caused the service-ca operator to continuously overwrite the caBundle with its own CA (`openshift-service-serving-signer`), while the operator was serving with a certificate signed by the self-managed CA (`hypershift-webhook-ca`), resulting in `x509: certificate signed by unknown authority` errors
- Adds `removeInjectCABundleAnnotation()` to strip the annotation from both Mutating and Validating webhook configurations during the service-ca migration path

## Test plan

- [x] Unit test added covering annotation removal from both MutatingWebhookConfiguration and ValidatingWebhookConfiguration
- [x] All existing webhook cert tests continue to pass
- [x] Verified workaround on live management cluster — removing the annotation and restarting the operator resolved the TLS errors

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved webhook certificate handling during upgrades to avoid CA bundle being overwritten by external certificate management, ensuring webhook CA bundles remain correct.

* **Tests**
  * Added and extended tests covering webhook configuration updates and the upgrade scenario that removes conflicting annotations from webhook resources.

### DIFF
--- a/hypershift-operator/controllers/webhookcerts/webhookcerts_controller_test.go
+++ b/hypershift-operator/controllers/webhookcerts/webhookcerts_controller_test.go
@@ -92,7 +92,7 @@ func TestReconcile(t *testing.T) {
 		crd := &apiextensionsv1.CustomResourceDefinition{
 			ObjectMeta: metav1.ObjectMeta{Name: "hostedclusters.hypershift.openshift.io"},
 			Spec: apiextensionsv1.CustomResourceDefinitionSpec{
-				Group: "hypershift.openshift.io",
+				Group: webhookConfigName,
 				Names: apiextensionsv1.CustomResourceDefinitionNames{
 					Plural:   "hostedclusters",
 					Singular: "hostedcluster",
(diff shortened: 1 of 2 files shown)


hypershift-operator/controllers/webhookcerts/webhookcerts_controller_test.go
```diff
@@ -314,6 +314,57 @@ func TestReconcile(t *testing.T) {
 		g.Expect(caSecret.Data).To(HaveKey(certs.CASignerCertMapKey))
 	})
 
+	t.Run("When upgrading from service-ca it should remove inject-cabundle annotation from webhook configurations", func(t *testing.T) {
+		g := NewWithT(t)
+
+		mwc := &admissionregistrationv1.MutatingWebhookConfiguration{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: webhookConfigName,
```
cblecker, line 322: nit: Since you introduced `webhookConfigName`, it might be worth updating the existing tests in this file to use it too — lines 184, 194, 213, and 218 still hardcode `"hypershift.openshift.io"`. No big deal either way.

cblecker (commented):
Fix looks correct — the annotation removal is well-placed within `removeServiceCAResources` and the ordering with respect to `patchWebhookConfigsCABundle` is sound. Nice catch on this one.

/lgtm

hypershift-operator/controllers/webhookcerts/webhookcerts_controller_test.go
joshbranham: Good call, silly robot missed that. Fixed, needs another lgtm if you are up for it.

PR #8507, opened 2026-05-13
OCPBUGS-85584: Fix stale api.openshift.com labels on HostedControlPlane

dustman9000:
## Summary

Fixes stale `api.openshift.com/*` labels on HostedControlPlane CRs that persist after being removed from the HostedCluster.

## Problem

The label sync in `reconcileHostedControlPlane` ([line 2426-2432](https://github.com/openshift/hypershift/blob/main/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go#L2426-L2432)) only copies labels from HC to HCP. It never removes HCP labels that were deleted from HC.

When OCM/CS removes `api.openshift.com/limited-support` from the HostedCluster (via ManifestWork when LS is cleared), the label stays on the HostedControlPlane forever.

## Reproduction (stage, 2026-05-13)

Reproduced on stage cluster `test-hcp-stable-420` (2q4jijtpbjlrtjavc29s8n54bnah70kl), MC `hs-mc-13vnfh7o0`:

1. Cluster had 1 LS reason. Both HC and HCP had `api.openshift.com/limited-support: "true"`. Correct.
2. Removed LS reason via `ocm delete`. OCM confirmed 0 LS reasons.
3. Checked HC: label cleared (CS removed via ManifestWork). Checked HCP: label still `"true"`.
4. After 60 seconds: HC still cleared, HCP still `"true"`. **Label removal never propagated.**

Also confirmed on production clusters `dai-test-hcp` and `clairp02ue1` (0 LS reasons in OCM, stale label on HCP).

## Impact

Route Monitor Operator checks this label on the HCP to skip probe creation for LS clusters. Stale labels cause RMO to permanently skip synthetic monitoring probe creation for fully supported clusters. ~60% of clusters on some RHOBS cells have no synthetic probes due to this bug.

## Fix

Before copying HC labels to HCP, remove any `api.openshift.com/*` labels from HCP that no longer exist on HC. Includes 4 unit tests covering label copy, stale removal, non-OCM label preservation, and nil HC labels.

## Jira

- [OCPBUGS-85584](https://redhat.atlassian.net/browse/OCPBUGS-85584)
- [SREP-4859](https://redhat.atlassian.net/browse/SREP-4859) (downstream impact)

## Summary by CodeRabbit

* **Bug Fixes**
  * Label synchronization for OpenShift API labels now removes stale labels from the target resource and tightens matching to avoid partial/prefix collisions, while still copying current labels.

* **Tests**
  * Added coverage validating label add, stale-label removal, preservation of unrelated labels, and behavior when source labels are absent.

### DIFF
--- a/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
+++ b/hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
@@ -135,6 +135,8 @@ const (
 	useRestrictedPodSecurityLabel                              = "io.openshift.hypershift.restricted-psa"
 	defaultToControlPlaneV2Label                               = "io.openshift.hypershift.control-plane-operator.v2-isdefault"
 
+	apiOpenShiftComLabelPrefix = "api.openshift.com/"
+
 	etcdEncKeyPostfix = "-etcd-encryption-key"
 
 	jobHostedClusterNameLabel      = "hypershift.openshift.io/cluster-name"
@@ -2423,10 +2425,17 @@ func reconcileHostedControlPlane(hcp *hyperv1.HostedControlPlane, hcluster *hype
 	if hcp.Labels == nil {
 		hcp.Labels = make(map[string]string)
 	}
-	// All labels on the HostedCluster with this special prefix are copied
-	// Those are labels set by OCM
+	// These labels are managed by OCM. Delete-then-copy ensures removals
+	// on the HostedCluster (e.g., clearing limited-support) propagate to the HCP.
+	for key := range hcp.Labels {
+		if strings.HasPrefix(key, apiOpenShiftComLabelPrefix) {
+			if _, exists := hcluster.Labels[key]; !exists {
+				delete(hcp.Labels, key)
+			}
+		}
+	}
 	for key, val := range hcluster.Labels {
-		if strings.HasPrefix(key, "api.openshift.com") {
+		if strings.HasPrefix(key, apiOpenShiftComLabelPrefix) {
 			hcp.Labels[key] = val
 		}
 	}
--- a/hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
+++ b/hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
@@ -310,6 +310,64 @@ func TestReconcileHostedControlPlaneAdditionalTrustBundle(t *testing.T) {
 	}
 }
 
+func TestReconcileHostedControlPlaneLabelSync(t *testing.T) {
+	t.Parallel()
+	tests := []struct {
+		name           string
+		hcLabels       map[string]string
+		hcpLabels      map[string]string
+		expectedLabels map[string]string
+	}{
+		{
+			name:           "When HC has api.openshift.com labels, it should copy them to HCP",
+			hcLabels:       map[string]string{"api.openshift.com/limited-support": "true", "api.openshift.com/name": "test"},
+			hcpLabels:      map[string]string{},
+			expectedLabels: map[string]string{"api.openshift.com/limited-support": "true", "api.openshift.com/name": "test"},
+		},
+		{
+			name:           "When HC removes an api.openshift.com label, it should remove the stale label from HCP",
+			hcLabels:       map[string]string{"api.openshift.com/name": "test"},
+			hcpLabels:      map[string]string{"api.openshift.com/limited-support": "true", "api.openshift.com/name": "old"},
+			expectedLabels: map[string]string{"api.openshift.com/name": "test"},
+		},
+		{
+			name:           "When HC has no api.openshift.com labels, it should preserve non-api.openshift.com labels on HCP",
+			hcLabels:       map[string]string{},
+			hcpLabels:      map[string]string{"api.openshift.com/limited-support": "true", "cluster.x-k8s.io/cluster-name": "keep-me"},
+			expectedLabels: map[string]string{"cluster.x-k8s.io/cluster-name": "keep-me"},
+		},
+		{
+			name:           "When HC labels are nil, it should remove all api.openshift.com labels from HCP",
+			hcLabels:       nil,
+			hcpLabels:      map[string]string{"api.openshift.com/limited-support": "true"},
+			expectedLabels: map[string]string{},
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			g := NewGomegaWithT(t)
+			hc := &hyperv1.HostedCluster{
+				ObjectMeta: metav1.ObjectMeta{Labels: test.hcLabels},
+			}
+			hcp := &hyperv1.HostedControlPlane{
+				ObjectMeta: metav1.ObjectMeta{Labels: test.hcpLabels},
+			}
+			err := reconcileHostedControlPlane(hcp, hc, false, false, func() (map[string]string, error) { return nil, nil })
+			g.Expect(err).ToNot(HaveOccurred())
+
+			for key, val := range test.expectedLabels {
+				g.Expect(hcp.Labels).To(HaveKeyWithValue(key, val))
+			}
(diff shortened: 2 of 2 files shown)


hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
```diff
@@ -310,6 +310,68 @@ func TestReconcileHostedControlPlaneAdditionalTrustBundle(t *testing.T) {
...
+			if err != nil {
+				t.Fatal(err)
+			}
+			for key, val := range test.expectedLabels {
+				if hcp.Labels[key] != val {
+					t.Errorf("expected label %s=%s, got %s", key, val, hcp.Labels[key])
+				}
+			}
+			for key := range hcp.Labels {
+				if strings.HasPrefix(key, "api.openshift.com") {
```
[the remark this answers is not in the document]
dustman9000: Good catch. The existing copy loop (line 2435) also uses `"api.openshift.com"` without the trailing slash, so our deletion loop matches the same prefix for consistency. Changing one without the other could cause labels to be deleted but not re-copied (or vice versa). I will keep them consistent for now since there are no known labels that would match `api.openshift.com` but not `api.openshift.com/`.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -2423,8 +2423,18 @@ func reconcileHostedControlPlane(hcp *hyperv1.HostedControlPlane, hcluster *hype
...
+	for key := range hcp.Labels {
+		if strings.HasPrefix(key, "api.openshift.com") {
+			if _, exists := hcluster.Labels[key]; !exists {
+				delete(hcp.Labels, key)
+			}
+		}
+	}
 	for key, val := range hcluster.Labels {
 		if strings.HasPrefix(key, "api.openshift.com") {
 			hcp.Labels[key] = val
```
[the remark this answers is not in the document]
dustman9000: Good catch. The existing copy loop (line 2435) also uses "api.openshift.com" without the trailing slash, so our deletion loop matches the same prefix for consistency. Changing one without the other could cause labels to be deleted but not re-copied (or vice versa). I will keep them consistent for now since there are no known labels that would match api.openshift.com but not api.openshift.com/.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -2423,8 +2423,18 @@ func reconcileHostedControlPlane(hcp *hyperv1.HostedControlPlane, hcluster *hype
...
 	}
-	// All labels on the HostedCluster with this special prefix are copied
-	// Those are labels set by OCM
+	// Sync api.openshift.com/* labels from HostedCluster to HostedControlPlane.
+	// These labels are managed by OCM (e.g., limited-support, environment, name).
+	// First remove any api.openshift.com/* labels from HCP that no longer exist
+	// on the HC, then copy current values. This ensures label removals (e.g.,
+	// when limited support is cleared) propagate to the HCP.
+	for key := range hcp.Labels {
+		if strings.HasPrefix(key, "api.openshift.com") {
```
cblecker: nit: Since you're already touching both loops, would you mind pulling the prefix out into a `const` with a trailing slash? Something like:

```go
const apiOpenShiftComLabelPrefix = "api.openshift.com/"
```

Without the trailing slash, this technically matches anything that starts with `api.openshift.com` — including a hypothetical `api.openshift.company/foo`. A shared const fixes that and avoids the duplicated string literal across both loops.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
```diff
@@ -310,6 +310,68 @@ func TestReconcileHostedControlPlaneAdditionalTrustBundle(t *testing.T) {
...
+func TestReconcileHostedControlPlaneLabelSync(t *testing.T) {
+	t.Parallel()
+	tests := []struct {
+		name           string
+		hcLabels       map[string]string
+		hcpLabels      map[string]string
+		expectedLabels map[string]string
+	}{
+		{
+			name:           "copies api.openshift.com labels from HC to HCP",
```
cblecker: nit: The project convention for test case names is Gherkin-style — "When... it should...". Would you mind renaming these? Something like:

- `"When HC has api.openshift.com labels, it should copy them to HCP"`
- `"When HC removes an api.openshift.com label, it should remove the stale label from HCP"`
- `"When HCP has non-api.openshift.com labels, it should preserve them"`
- `"When HC labels are nil, it should remove all api.openshift.com labels from HCP"`

hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
```diff
@@ -310,6 +310,68 @@ func TestReconcileHostedControlPlaneAdditionalTrustBundle(t *testing.T) {
...
+		t.Run(test.name, func(t *testing.T) {
+			hc := &hyperv1.HostedCluster{
+				ObjectMeta: metav1.ObjectMeta{Labels: test.hcLabels},
+			}
+			hcp := &hyperv1.HostedControlPlane{
+				ObjectMeta: metav1.ObjectMeta{Labels: test.hcpLabels},
+			}
+			err := reconcileHostedControlPlane(hcp, hc, false, false, func() (map[string]string, error) { return nil, nil })
+			if err != nil {
+				t.Fatal(err)
```
cblecker: nit: The rest of this file uses gomega — would be nice to stay consistent here. Switching to `g.Expect(err).ToNot(HaveOccurred())` and `g.Expect(hcp.Labels).To(Equal(test.expectedLabels))` would also let you collapse the two verification loops below into a single assertion, which is simpler and catches unexpected mutations to any label — not just the `api.openshift.com` ones.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
```diff
@@ -2423,8 +2423,18 @@ func reconcileHostedControlPlane(hcp *hyperv1.HostedControlPlane, hcluster *hype
 	if hcp.Labels == nil {
 		hcp.Labels = make(map[string]string)
 	}
-	// All labels on the HostedCluster with this special prefix are copied
-	// Those are labels set by OCM
+	// Sync api.openshift.com/* labels from HostedCluster to HostedControlPlane.
```
cblecker: nit: Most of this comment describes what the code does, which the two loops already make clear. The useful parts are that OCM owns these labels and why delete-before-copy is needed. Maybe condense to:

```go
// These labels are managed by OCM. Delete-then-copy ensures removals
// on the HostedCluster propagate; a copy-only loop leaves stale labels.
```

cblecker (commented):
Fix looks correct — delete-then-copy is the right approach, and the nil map case is handled cleanly. A few minor style suggestions inline.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
dustman9000: Done, thanks.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
dustman9000: Done, thanks.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller_test.go
dustman9000: Done, thanks.

hypershift-operator/controllers/hostedcluster/hostedcluster_controller.go
dustman9000: Done, thanks.

clcollins:
@cblecker - I think this needs an `/approve` too, if you've got a sec?

PR #8508, opened 2026-05-13
NO-JIRA: build: use uv to run python tools with pip fallback

cblecker:
## Summary

- When `uv` is detected on `PATH`, codespell, gitlint, and pyyaml are invoked via `uv run --with <pkg>==<ver>` instead of being installed into `hack/tools/bin/` with `pip install --target`
- When `uv` is not available, the existing pip-based install targets are used unchanged (full fallback, no behaviour change for CI or contributors without uv)
- All conditional logic is confined to a single `ifneq` variable block; the `verify-codespell`, `run-gitlint`, and `verify-docs-nav` targets themselves have zero conditionals

## Test plan

- [ ] With uv on PATH: `make verify-codespell`, `make run-gitlint`, `make verify-docs-nav` — confirm each uses `uv run`
- [ ] Without uv: `UV= make verify-codespell` (and the other two) — confirm pip fallback still works
- [ ] `make -j verify-parallel` — confirm parallel execution works with both paths
- [ ] Custom override: `UV=/custom/uv make verify-codespell` — confirm override is respected

## Summary by CodeRabbit

* **Chores**
  * Updated build system to support the `uv` package manager with automatic fallback to traditional tooling when unavailable.
  * Enhanced verification commands for documentation navigation, spell-checking, and linting to use dynamic tool detection.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,6 +35,25 @@ PYYAML_VER := 6.0.3
 PYYAML_DIST_DIR := pyyaml_dist
 PYYAML_STAMP := $(TOOLS_BIN_DIR)/$(PYYAML_DIST_DIR)/.installed
 
+# Python tool runner: prefer uv when available, fall back to pip
+UV ?= $(shell command -v uv 2>/dev/null)
+
+ifneq ($(UV),)
+CODESPELL_CMD  = $(UV) run --with codespell==$(CODESPELL_VER) codespell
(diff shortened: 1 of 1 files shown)


cblecker:
This isn't really needed anymore as #8269 made it so manually sourcing venv or installing it into the system was no longer required.

PR #8509, opened 2026-05-13
OCPBUGS-85585: tighten registry override matching to strict longest-prefix across release-image consumers

raelga:
## Summary
The `RegistryMirrorProviderDecorator` used by `Lookup()` rewrites every
ImageStream tag with a bare `strings.Replace(image, source, target, 1)`. This
matches anywhere in the image reference, so:
- it can incorrectly rewrite substrings inside hostnames (e.g. `quay.io`
  inside `quay.io.example.com`),
- it does not reliably rewrite all component images, leading to CPO
  sub-resources (e.g. the `availability-prober` init container injected into
  `capi-provider` and other deployments) still pointing at the original
  registry.

Downstream, this breaks clusters where a ValidatingAdmissionPolicy in Deny
mode only allows images from approved registries: HCP creation hangs
indefinitely.

## Root cause
`support/releaseinfo/registry_mirror_provider.go` ran an unbounded
`strings.Replace` per `(source, target)` pair across the entire image string.
Match boundaries were neither anchored to the start of the reference nor to a
path separator, and iteration order over the override map was undefined, so
which override "won" depended on Go's randomized map iteration.

## Fix
Extract a shared `registryoverride.Replace(image, overrides)` helper at
`support/util/registryoverride/` and use it from both:

1. `RegistryMirrorProviderDecorator.Lookup()` (the actual bug fix), and
2. `imageprovider.NewWithRegistryOverrides()` (defense-in-depth: callers that
   construct a provider directly still get the same semantics).

The helper:
- matches strictly: `image == source` or `image` has prefix `source + "/"`
  (no substring/false-host matches),
- picks the **longest matching source prefix** (deterministic across map
  iteration order),
- is a no-op when overrides are empty or no source matches.

`imageprovider.New(releaseImage)` now delegates to
`NewWithRegistryOverrides(releaseImage, nil)`, and the controller wiring at
`hostedcontrolplane_controller.go:1106` keeps the explicit
`NewWithRegistryOverrides(..., r.ReleaseProvider.GetRegistryOverrides())`
call so the override path is exercised even when callers don't go through
`Lookup()` first.

## Impact
- Component images in CPO sub-resources (init containers in `capi-provider`,
  ignition server, etc.) now reliably honor `--registry-overrides`.
- VAPs in Deny mode no longer block HCP creation in air-gapped / mirrored
  environments.
- Side effect: false-positive substring rewrites in the decorator (a
  pre-existing latent bug) are also fixed.

## Test plan
- Unit tests for `registryoverride.Replace` covering 12 cases incl. boundary,
  empty overrides, longest-prefix, no mutation of the overrides map.
- `imageprovider_test.go` covers no-overrides, no-match, single-match,
  multi-override, longest-prefix-wins, **mutation safety** (release-image and
  overrides untouched), and **idempotency** (no double-apply when image was
  already mirrored).
- `hostedcontrolplane_controller_test.go` (`TestEventHandling`,
  `TestNonReadyInfraTriggersRequeueAfter`) now exercise the override path
  with non-empty overrides through full reconcile.
- E2E with `--registry-overrides` + VAP in Deny mode (manual).

## Related
- [OCPBUGS-85585](https://redhat.atlassian.net/browse/OCPBUGS-85585)
- [AROSLSRE-829](https://redhat.atlassian.net/browse/AROSLSRE-829)
- [OCPSTRAT-2437](https://redhat.atlassian.net/browse/OCPSTRAT-2437)

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
@@ -1108,7 +1108,7 @@ func (r *HostedControlPlaneReconciler) update(ctx context.Context, hostedControl
 	}
 
 	userReleaseImageProvider := imageprovider.New(userReleaseImage)
-	releaseImageProvider := imageprovider.New(releaseImage)
+	releaseImageProvider := imageprovider.NewWithRegistryOverrides(releaseImage, r.ReleaseProvider.GetRegistryOverrides())
 
 	var errs []error
 	if err := r.reconcileCPOV2(ctx, hostedControlPlane, infraStatus, releaseImageProvider, userReleaseImageProvider); err != nil {
--- a/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller_test.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller_test.go
@@ -731,6 +731,8 @@ func TestEventHandling(t *testing.T) {
 	mockedProviderWithOpenshiftImageRegistryOverrides.EXPECT().
 		Lookup(gomock.Any(), gomock.Any(), gomock.Any()).
 		Return(testutils.InitReleaseImageOrDie("4.15.0"), nil).AnyTimes()
+	mockedProviderWithOpenshiftImageRegistryOverrides.EXPECT().
+		GetRegistryOverrides().Return(map[string]string{}).AnyTimes()
 	mockEC2 := awsapi.NewMockEC2API(mockCtrl)
 	mockEC2.EXPECT().DescribeVpcEndpoints(gomock.Any(), gomock.Any()).Return(&ec2.DescribeVpcEndpointsOutput{}, fmt.Errorf("not ready")).AnyTimes()
 
@@ -814,6 +816,8 @@ func TestNonReadyInfraTriggersRequeueAfter(t *testing.T) {
 	mockedProviderWithOpenshiftImageRegistryOverrides.EXPECT().
 		Lookup(gomock.Any(), gomock.Any(), gomock.Any()).
 		Return(testutils.InitReleaseImageOrDie("4.15.0"), nil).AnyTimes()
+	mockedProviderWithOpenshiftImageRegistryOverrides.EXPECT().
+		GetRegistryOverrides().Return(map[string]string{}).AnyTimes()
 	mockEC2 := awsapi.NewMockEC2API(mockCtrl)
 	mockEC2.EXPECT().DescribeVpcEndpoints(gomock.Any(), gomock.Any()).Return(&ec2.DescribeVpcEndpointsOutput{}, fmt.Errorf("not ready")).AnyTimes()
 	hcp := sampleHCP(t)
--- a/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller_test.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller_test.go
@@ -732,7 +732,7 @@ func TestEventHandling(t *testing.T) {
 		Lookup(gomock.Any(), gomock.Any(), gomock.Any()).
 		Return(testutils.InitReleaseImageOrDie("4.15.0"), nil).AnyTimes()
 	mockedProviderWithOpenshiftImageRegistryOverrides.EXPECT().
-		GetRegistryOverrides().Return(map[string]string{}).AnyTimes()
+		GetRegistryOverrides().Return(map[string]string{"registry": "override"}).AnyTimes()
 	mockEC2 := awsapi.NewMockEC2API(mockCtrl)
 	mockEC2.EXPECT().DescribeVpcEndpoints(gomock.Any(), gomock.Any()).Return(&ec2.DescribeVpcEndpointsOutput{}, fmt.Errorf("not ready")).AnyTimes()
 
@@ -817,7 +817,7 @@ func TestNonReadyInfraTriggersRequeueAfter(t *testing.T) {
 		Lookup(gomock.Any(), gomock.Any(), gomock.Any()).
 		Return(testutils.InitReleaseImageOrDie("4.15.0"), nil).AnyTimes()
 	mockedProviderWithOpenshiftImageRegistryOverrides.EXPECT().
-		GetRegistryOverrides().Return(map[string]string{}).AnyTimes()
+		GetRegistryOverrides().Return(map[string]string{"registry": "override"}).AnyTimes()
 	mockEC2 := awsapi.NewMockEC2API(mockCtrl)
 	mockEC2.EXPECT().DescribeVpcEndpoints(gomock.Any(), gomock.Any()).Return(&ec2.DescribeVpcEndpointsOutput{}, fmt.Errorf("not ready")).AnyTimes()
 	hcp := sampleHCP(t)
--- a/control-plane-operator/controllers/hostedcontrolplane/imageprovider/imageprovider.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/imageprovider/imageprovider.go
@@ -1,9 +1,10 @@
 package imageprovider
 
 import (
-	"strings"
+	"maps"
 
 	"github.com/openshift/hypershift/support/releaseinfo"
+	"github.com/openshift/hypershift/support/util/registryoverride"
 )
 
 //go:generate ../../../../hack/tools/bin/mockgen -source=imageprovider.go -package=imageprovider -destination=imageprovider_mock.go
@@ -27,11 +28,7 @@ type SimpleReleaseImageProvider struct {
 }
 
 func New(releaseImage *releaseinfo.ReleaseImage) *SimpleReleaseImageProvider {
-	return &SimpleReleaseImageProvider{
-		componentsImages: releaseImage.ComponentImages(),
-		missingImages:    make([]string, 0),
-		ReleaseImage:     releaseImage,
-	}
+	return NewWithRegistryOverrides(releaseImage, nil)
 }
 
 func NewFromImages(componentsImages map[string]string) *SimpleReleaseImageProvider {
@@ -66,21 +63,13 @@ func (p *SimpleReleaseImageProvider) ComponentImages() map[string]string {
 // NewWithRegistryOverrides creates a SimpleReleaseImageProvider that applies
 // registry overrides to all component images. This ensures init containers
 // and other sub-resources created by CPO use the overridden image references.
+//
+// The returned provider owns a private copy of releaseImage.ComponentImages()
+// so callers can safely mutate it.
 func NewWithRegistryOverrides(releaseImage *releaseinfo.ReleaseImage, registryOverrides map[string]string) *SimpleReleaseImageProvider {
-	originalImages := releaseImage.ComponentImages()
-	images := make(map[string]string, len(originalImages))
-	for k, v := range originalImages {
-		images[k] = v
-	}
-	if len(registryOverrides) > 0 {
-		for key, image := range images {
-			for source, target := range registryOverrides {
-				if image == source || strings.HasPrefix(image, source+"/") {
-					images[key] = strings.Replace(image, source, target, 1)
-					break
-				}
-			}
-		}
+	images := maps.Clone(releaseImage.ComponentImages())
+	for key, image := range images {
+		images[key] = registryoverride.Replace(image, registryOverrides)
 	}
 	return &SimpleReleaseImageProvider{
 		componentsImages: images,
--- a/control-plane-operator/controllers/hostedcontrolplane/imageprovider/imageprovider.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/imageprovider/imageprovider.go
@@ -1,6 +1,10 @@
 package imageprovider
 
-import "github.com/openshift/hypershift/support/releaseinfo"
+import (
+	"strings"
+
+	"github.com/openshift/hypershift/support/releaseinfo"
+)
 
 //go:generate ../../../../hack/tools/bin/mockgen -source=imageprovider.go -package=imageprovider -destination=imageprovider_mock.go
 
@@ -57,4 +61,30 @@ func (p *SimpleReleaseImageProvider) ImageExist(key string) (string, bool) {
 
 func (p *SimpleReleaseImageProvider) ComponentImages() map[string]string {
 	return p.componentsImages
+}
+
+// NewWithRegistryOverrides creates a SimpleReleaseImageProvider that applies
+// registry overrides to all component images. This ensures init containers
+// and other sub-resources created by CPO use the overridden image references.
+func NewWithRegistryOverrides(releaseImage *releaseinfo.ReleaseImage, registryOverrides map[string]string) *SimpleReleaseImageProvider {
+	originalImages := releaseImage.ComponentImages()
+	images := make(map[string]string, len(originalImages))
+	for k, v := range originalImages {
+		images[k] = v
+	}
+	if len(registryOverrides) > 0 {
+		for key, image := range images {
+			for source, target := range registryOverrides {
+				if image == source || strings.HasPrefix(image, source+"/") {
+					images[key] = strings.Replace(image, source, target, 1)
+					break
+				}
+			}
+		}
+	}
+	return &SimpleReleaseImageProvider{
+		componentsImages: images,
+		missingImages:    make([]string, 0),
+		ReleaseImage:     releaseImage,
+	}
 }
--- a/control-plane-operator/controllers/hostedcontrolplane/imageprovider/imageprovider_test.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/imageprovider/imageprovider_test.go
@@ -4,6 +4,13 @@ import (
 	"testing"
 
 	. "github.com/onsi/gomega"
+
+	"github.com/openshift/hypershift/support/releaseinfo"
+
+	imageapi "github.com/openshift/api/image/v1"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 func TestNewFromImages(t *testing.T) {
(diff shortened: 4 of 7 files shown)


control-plane-operator/controllers/hostedcontrolplane/imageprovider/imageprovider.go
```diff
@@ -58,3 +62,25 @@ func (p *SimpleReleaseImageProvider) ImageExist(key string) (string, bool) {
 func (p *SimpleReleaseImageProvider) ComponentImages() map[string]string {
 	return p.componentsImages
 }
+
+// NewWithRegistryOverrides creates a SimpleReleaseImageProvider that applies
+// registry overrides to all component images. This ensures init containers
+// and other sub-resources created by CPO use the overridden image references.
+func NewWithRegistryOverrides(releaseImage *releaseinfo.ReleaseImage, registryOverrides map[string]string) *SimpleReleaseImageProvider {
+	images := releaseImage.ComponentImages()
```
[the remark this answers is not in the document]
raelga: Fixed in 8fd79a4a8. Map is now cloned before modification.

control-plane-operator/controllers/hostedcontrolplane/imageprovider/imageprovider.go
```diff
@@ -58,3 +62,25 @@ func (p *SimpleReleaseImageProvider) ImageExist(key string) (string, bool) {
...
+	if len(registryOverrides) > 0 {
+		for key, image := range images {
+			for source, target := range registryOverrides {
+				if strings.HasPrefix(image, source) {
+					images[key] = strings.Replace(image, source, target, 1)
+					break
```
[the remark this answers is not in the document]
raelga: Fixed in 8fd79a4a8. Now uses source+"/" prefix matching to prevent subdomain false matches.

control-plane-operator/controllers/hostedcontrolplane/imageprovider/imageprovider_test.go
```diff
@@ -174,3 +180,98 @@ func TestComponentImages(t *testing.T) {
...
+		}
+
+		provider := NewWithRegistryOverrides(releaseImage, overrides)
+
+		g.Expect(provider.GetImage("availability-prober")).To(Equal(
+			"acr.example.com/quay-cache/redhat-user-workloads/crt-redhat-acm-tenant/cpo:latest"))
+		g.Expect(provider.GetImage("etcd")).To(Equal(
+			"acr.example.com/gcr-cache/etcd-development/etcd:v3.5"))
+	})
+}
```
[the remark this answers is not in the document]
raelga: Fixed in 8fd79a4a8. Added the subdomain edge case test.

jparrill:
/retitle OCPBUGS-85585: apply registry overrides to component images in CPO sub-resources

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
```diff
@@ -1103,7 +1103,7 @@ func (r *HostedControlPlaneReconciler) update(ctx context.Context, hostedControl
 	}
 
 	userReleaseImageProvider := imageprovider.New(userReleaseImage)
-	releaseImageProvider := imageprovider.New(releaseImage)
+	releaseImageProvider := imageprovider.NewWithRegistryOverrides(releaseImage, r.ReleaseProvider.GetRegistryOverrides())
```
jparrill, line 1111: Hey, I traced the image flow and I think the `releaseImage` here already has overrides applied. When `r.ReleaseProvider.Lookup()` runs at line 1070, it goes through the `RegistryMirrorProviderDecorator` (wired up at `main.go:308-314`) which does `strings.Replace` on all ImageStream tags — including the static ones like `availability-prober` from `StaticProviderDecorator`.

So by this point, `releaseImage.ComponentImages()` should already return overridden refs. Calling `NewWithRegistryOverrides` again would try to match the original prefix (e.g. `quay.io`) but it's already been replaced, so it's effectively a no-op.

Do you have logs or a specific cluster config where you saw the un-overridden image? That would help nail down if the gap is really here or maybe further up the chain — e.g. `ControlPlaneOperatorImageAnnotation` bypassing the release provider entirely (`support/util/util.go:477`).

control-plane-operator/controllers/hostedcontrolplane/imageprovider/imageprovider.go
```diff
@@ -58,3 +62,29 @@ func (p *SimpleReleaseImageProvider) ImageExist(key string) (string, bool) {
...
+func NewWithRegistryOverrides(releaseImage *releaseinfo.ReleaseImage, registryOverrides map[string]string) *SimpleReleaseImageProvider {
+	originalImages := releaseImage.ComponentImages()
+	images := make(map[string]string, len(originalImages))
+	for k, v := range originalImages {
+		images[k] = v
+	}
+	if len(registryOverrides) > 0 {
+		for key, image := range images {
+			for source, target := range registryOverrides {
+				if image == source || strings.HasPrefix(image, source+"/") {
```
jparrill: Nice — this prefix guard is actually safer than what `RegistryMirrorProviderDecorator` does at `registry_mirror_provider.go:38`, which is a bare `strings.Replace` that would incorrectly match `quay.io` inside `quay.io.evil.com`.

One concern though: now we have two different matching algorithms for the same operation in the same pipeline. Worth extracting this into a shared helper in `support/util/` and fixing the decorator too — either here or as a follow-up. WDYT?

control-plane-operator/controllers/hostedcontrolplane/imageprovider/imageprovider.go
```diff
@@ -58,3 +62,29 @@ func (p *SimpleReleaseImageProvider) ImageExist(key string) (string, bool) {
 func (p *SimpleReleaseImageProvider) ComponentImages() map[string]string {
 	return p.componentsImages
 }
+
+// NewWithRegistryOverrides creates a SimpleReleaseImageProvider that applies
+// registry overrides to all component images. This ensures init containers
+// and other sub-resources created by CPO use the overridden image references.
+func NewWithRegistryOverrides(releaseImage *releaseinfo.ReleaseImage, registryOverrides map[string]string) *SimpleReleaseImageProvider {
```
jparrill, line 69: Minor: since this shares the same struct construction as `New()`, you could have `New()` just delegate:
```go
func New(releaseImage *releaseinfo.ReleaseImage) *SimpleReleaseImageProvider {
    return NewWithRegistryOverrides(releaseImage, nil)
}
```
Keeps initialization in one place.

control-plane-operator/controllers/hostedcontrolplane/imageprovider/imageprovider.go
```diff
@@ -58,3 +62,29 @@ func (p *SimpleReleaseImageProvider) ImageExist(key string) (string, bool) {
...
+	originalImages := releaseImage.ComponentImages()
+	images := make(map[string]string, len(originalImages))
+	for k, v := range originalImages {
```
jparrill: Tiny nit: since we're on Go 1.25+, this could be `maps.Clone(releaseImage.ComponentImages())`. Totally optional.

control-plane-operator/controllers/hostedcontrolplane/imageprovider/imageprovider_test.go
```diff
@@ -174,3 +181,115 @@ func TestComponentImages(t *testing.T) {
...
 	})
 }
+
+func newTestReleaseImage(images map[string]string) *releaseinfo.ReleaseImage {
+	tags := make([]imageapi.TagReference, 0, len(images))
+	for name, image := range images {
+		tags = append(tags, imageapi.TagReference{
+			Name: name,
+			From: &corev1.ObjectReference{Name: image},
+		})
```
jparrill, line 192: Tests look solid! Two cases I think would be worth adding:

1. **Mutation safety** — make sure the original `ReleaseImage` isn't modified (matters because `ReleaseImage` gets cached in the decorator chain):
```go
t.Run("When overrides are applied, it should not mutate the original release image", func(t *testing.T) {
    t.Parallel()
    g := NewWithT(t)
    releaseImage := newTestReleaseImage(map[string]string{"etcd": "quay.io/etcd:v3"})
    NewWithRegistryOverrides(releaseImage, map[string]string{"quay.io": "mirror.io"})
    g.Expect(releaseImage.ComponentImages()["etcd"]).To(Equal("quay.io/etcd:v3"))
})
```

2. **Double-override idempotency** — this is actually the real runtime scenario since `RegistryMirrorProviderDecorator.Lookup()` already applies overrides before we get here:
```go
t.Run("When images are already overridden, it should not double-apply", func(t *testing.T) {
    t.Parallel()
    g := NewWithT(t)
    releaseImage := newTestReleaseImage(map[string]string{
        "availability-prober": "mirror.example.com/quay-cache/ns/image:tag",
    })
    overrides := map[string]string{"quay.io": "mirror.example.com/quay-cache"}
    provider := NewWithRegistryOverrides(releaseImage, overrides)
    g.Expect(provider.GetImage("availability-prober")).To(Equal(
        "mirror.example.com/quay-cache/ns/image:tag"))
})
```

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller_test.go
```diff
@@ -727,6 +727,8 @@ func TestEventHandling(t *testing.T) {
 	mockedProviderWithOpenshiftImageRegistryOverrides.EXPECT().
 		Lookup(gomock.Any(), gomock.Any(), gomock.Any()).
 		Return(testutils.InitReleaseImageOrDie("4.15.0"), nil).AnyTimes()
+	mockedProviderWithOpenshiftImageRegistryOverrides.EXPECT().
+		GetRegistryOverrides().Return(map[string]string{}).AnyTimes()
```
jparrill: These mocks return empty overrides, so they only hit the no-op path. It'd be great to have at least one test with actual non-empty overrides to exercise the real logic end-to-end from the controller level.

jparrill (commented):

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
raelga: You're right that `Lookup()` already runs the decorator before this line, so by the time we hit `imageprovider.New(...)` the images should be rewritten in principle. The reason they weren't, reliably, is in the decorator itself: `RegistryMirrorProviderDecorator` was using a bare `strings.Replace(image, source, target, 1)` with no anchoring, no path boundary, and undefined iteration order across overrides, so it would silently miss images depending on map order and could even rewrite substrings inside hostnames.

I've pushed a new series that pulls the matching logic into [`support/util/registryoverride.Replace`](https://github.com/openshift/hypershift/pull/8509/commits/51b45f929) and uses it from both the decorator ([the actual fix](https://github.com/openshift/hypershift/pull/8509/commits/006059459)) and `NewWithRegistryOverrides` ([defense-in-depth](https://github.com/openshift/hypershift/pull/8509/commits/f0b92a6a5)). The controller-line change stays as-is but is no longer load-bearing on its own; see the updated PR description for the reframed root cause.

I haven't found any path that bypasses the release provider (`ControlPlaneOperatorImageAnnotation` only sets one specific image and goes through `support/util/util.go:477` which is independent of the override pipeline), so the decorator was the gap.

control-plane-operator/controllers/hostedcontrolplane/imageprovider/imageprovider.go
raelga: Done: [extracted to `registryoverride.Replace`](https://github.com/openshift/hypershift/pull/8509/commits/51b45f929) and the [decorator now uses it too](https://github.com/openshift/hypershift/pull/8509/commits/006059459).

One semantic I want to flag explicitly: the helper picks the **longest matching source prefix** rather than first-match. Reason: the original loose `strings.Replace` matched any substring (so longest-prefix is unambiguously safer), and first-match would be non-deterministic across Go map iteration order. Happy to switch to first-source-with-defined-order if you'd rather avoid that semantic, but longest-prefix-wins seems both safer and more intuitive for things like `quay.io` vs `quay.io/openshift-release-dev/`.

control-plane-operator/controllers/hostedcontrolplane/imageprovider/imageprovider.go
raelga: Done in [`f0b92a6a5`](https://github.com/openshift/hypershift/pull/8509/commits/f0b92a6a5): `New(releaseImage)` is now a one-liner over `NewWithRegistryOverrides(releaseImage, nil)`. `maps.Clone(nil)` returns `nil` and the override loop no-ops cleanly.

control-plane-operator/controllers/hostedcontrolplane/imageprovider/imageprovider.go
raelga: Done in [`f0b92a6a5`](https://github.com/openshift/hypershift/pull/8509/commits/f0b92a6a5): `maps.Clone(releaseImage.ComponentImages())`. (Also using `maps.Clone` in the new tests where I needed a defensive copy of the overrides map for mutation-safety assertions.)

control-plane-operator/controllers/hostedcontrolplane/imageprovider/imageprovider_test.go
raelga: Done in [`dde0356f8`](https://github.com/openshift/hypershift/pull/8509/commits/dde0356f8): three new sub-tests inside `TestNewWithRegistryOverrides`:

1. **`longest matching source prefix wins over shorter one`**: reinforces the semantic flagged in [#3242084370](https://github.com/openshift/hypershift/pull/8509#discussion_r3242084370).
2. **`overrides map and source images must not be mutated`**: asserts `releaseImage.ComponentImages()` returns the original quay.io ref and the input overrides map is unchanged after the call.
3. **`re-applying the same overrides to an already-overridden provider is a no-op`**: passes the rewritten ComponentImages back through `NewWithRegistryOverrides` with the same overrides; nothing rewrites a second time because the strict prefix no longer matches.

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller_test.go
raelga: Done in [`62a565b70`](https://github.com/openshift/hypershift/pull/8509/commits/62a565b70): flipped both PR-added mocks (`TestEventHandling`, `TestNonReadyInfraTriggersRequeueAfter`) from `map[string]string{}` to `map[string]string{"registry": "override"}` (matching the convention already used at lines 176 and 1040 of the same file) so the full reconcile path actually walks the override-application logic at `hostedcontrolplane_controller.go:1106` (and the propagation through `configoperatorv2.NewComponent` and the ignitionserver `--registry-overrides` flag).

The release-image fixture (`InitReleaseImageOrDie("4.15.0")`) has no images matching the override key, so the existing semantic assertions of these tests are unchanged: this is purely added code-path coverage. A stronger end-to-end assertion would need a custom release-image fixture with `quay.io/...` tags, which I can add as a follow-up if you'd like, but felt out of scope for the bug fix.

raelga:
The two failures (`ci/prow/security` and `ci/prow/verify-deps`) are unrelated to this PR. Both show `Pod scheduling timeout` in their Prow status, which is a transient build-cluster infrastructure issue: the test pods never got scheduled, so the jobs did not even start executing the code under test.

Re-running:

/retest

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller_test.go
raelga: https://github.com/openshift/hypershift/pull/8509#issuecomment-4622257287

control-plane-operator/controllers/hostedcontrolplane/imageprovider/imageprovider_test.go
raelga: LGTM'd in https://github.com/openshift/hypershift/pull/8509#issuecomment-4622257287

control-plane-operator/controllers/hostedcontrolplane/imageprovider/imageprovider.go
raelga: LGTM'd in https://github.com/openshift/hypershift/pull/8509#issuecomment-4622257287

control-plane-operator/controllers/hostedcontrolplane/imageprovider/imageprovider.go
raelga: LGTM'd in https://github.com/openshift/hypershift/pull/8509#issuecomment-4622257287

control-plane-operator/controllers/hostedcontrolplane/imageprovider/imageprovider.go
raelga: LGTM'd in https://github.com/openshift/hypershift/pull/8509#issuecomment-4622257287

control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go
raelga: LGTM'd in https://github.com/openshift/hypershift/pull/8509#issuecomment-4622257287

raelga:
Rebased onto `main` (head now `55b6fda6eb`).

The only failing checks were the two Konflux **Enterprise Contract** gates, and they were failing purely because the branch was stale — not because of any code in this PR:

- This branch forked from `main` on 2026-05-13 and was ~389 commits behind.
- Konflux builds each PR with the `.tekton/` pipeline files committed on the branch, which pin every build task to a specific task-bundle **digest**. Enterprise Contract verifies those digests are in the currently approved set.
- After this branch was cut, `main` updated those pins (version bumps for `clamav-scan`, `rpms-signature-scan`, `build-image-index`, ~17 digest re-pins, new `enable-package-registry-proxy` param). The branch still referenced the old, now-unapproved digests → ~60 EC policy violations.
- This PR touches **zero `.tekton/` files**, so there was nothing in the diff to fix. The `control-plane-operator` EC check already passed, and up-to-date PRs show these same checks resolving to neutral — confirming the cause was branch staleness, not the code.

The rebase replayed the 6 commits cleanly (no conflicts); `.tekton/` now matches `main`, and the affected unit tests (`registryoverride`, `releaseinfo`, `imageprovider`, and the hostedcontrolplane reconcile tests) still pass. EC is re-running against the approved bundle set.

PR #8520, opened 2026-05-14
fix: OCPBUGS-81675: hcco: merge additionalTrustBundle into proxy trusted CA bundle

dpateriya:
## Problem

When `HostedCluster.spec.additionalTrustBundle` is set, HCCO syncs it into `openshift-config/user-ca-bundle` on the guest cluster. However, that ConfigMap is **not** referenced by the guest cluster's `proxy.config.openshift.io/cluster` object, so CNO never merges it into `openshift-config-managed/trusted-ca-bundle`.

Any component that reads `trusted-ca-bundle` to build its TLS trust pool (such as CVO when contacting a custom `spec.updateService`) cannot verify certificates signed by the internal CA that was provided only via `additionalTrustBundle`. The failure surface is:

```
Unable to retrieve available updates: tls: failed to verify certificate: x509: certificate signed by unknown authority
```

This affects users running an internal Cincinnati-compatible update service (e.g. OSUS) behind a certificate signed by an internal CA that is provided via `additionalTrustBundle`.

## Fix

Two changes in HCCO (`reconciler` in `resources.go`):

**Change 1 — `reconcileProxyTrustedCAConfigMap`**

When both `proxy.trustedCA` and `additionalTrustBundle` are set, merge the `additionalTrustBundle` CA data into the same destination ConfigMap (in `openshift-config`) alongside the proxy CA data. CNO then merges both into `openshift-config-managed/trusted-ca-bundle`.

**Change 2 — Proxy object reconciliation in `reconcileConfig`**

When `proxy.trustedCA` is not set but `additionalTrustBundle` is set, point the guest cluster's `proxy.config.openshift.io/cluster` object's `spec.trustedCA.Name` at `"user-ca-bundle"` (which `reconcileUserCertCABundle` already writes). CNO then picks it up and merges it into `trusted-ca-bundle`.

## Why here (HCCO) and not in upstream CVO

A previous PR ([openshift/cluster-version-operator#1370](https://github.com/openshift/cluster-version-operator/pull/1370)) attempted to fix this in CVO gated on `--hypershift`. Reviewer @enxebre correctly pointed out that CPO ships with the OCP payload at the same cadence as CVO, so there is no release-speed advantage, and the CVO fix would be unnecessary tech debt. This HCCO fix benefits **all** consumers of `trusted-ca-bundle` (CVO, MCO, etc.) rather than just CVO.

## Testing

```
go test ./control-plane-operator/hostedclusterconfigoperator/controllers/resources/ -run "TestReconcileProxyTrustedCAConfigMap|TestReconcileUserCertCABundle" -v
```

New `TestReconcileProxyTrustedCAConfigMap` table-driven test covers 4 cases:
- neither `proxy.trustedCA` nor `additionalTrustBundle` set → no dest ConfigMap created
- only `proxy.trustedCA` set → dest ConfigMap contains proxy CA data only
- both set → dest ConfigMap contains merged CA data from both sources
- only `additionalTrustBundle` set → `reconcileProxyTrustedCAConfigMap` returns early (handled by Change 2)

Fixes: https://issues.redhat.com/browse/OCPBUGS-81675

## Summary by CodeRabbit

* **Bug Fixes**
  * Proxy trusted CA handling now sets the proxy to use the user CA bundle when an additional trust bundle is present, avoids removing that bundle during cleanup, and correctly creates/merges CA data so concatenated bundles preserve newline formatting.

* **Tests**
  * Added unit tests validating trusted CA creation, error cases, precedence behavior, and correct merging between proxy trusted CA and additional trust bundles.

### DIFF
--- a/control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
+++ b/control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
@@ -47,6 +47,7 @@ import (
 	hyperapi "github.com/openshift/hypershift/support/api"
 	"github.com/openshift/hypershift/support/azureutil"
 	"github.com/openshift/hypershift/support/capabilities"
+	"github.com/openshift/hypershift/support/certs"
 	"github.com/openshift/hypershift/support/config"
 	"github.com/openshift/hypershift/support/globalconfig"
 	"github.com/openshift/hypershift/support/releaseinfo"
@@ -1071,6 +1072,13 @@ func (r *reconciler) reconcileConfig(ctx context.Context, hcp *hyperv1.HostedCon
 	proxy := globalconfig.ProxyConfig()
 	if _, err := r.CreateOrUpdate(ctx, r.client, proxy, func() error {
 		globalconfig.ReconcileInClusterProxyConfig(proxy, hcp.Spec.Configuration)
+		// When no proxy.trustedCA is set but additionalTrustBundle is, point the
+		// Proxy object at openshift-config/user-ca-bundle (already written by
+		// reconcileUserCertCABundle) so CNO merges it into
+		// openshift-config-managed/trusted-ca-bundle.
+		if proxy.Spec.TrustedCA.Name == "" && hcp.Spec.AdditionalTrustBundle != nil {
+			proxy.Spec.TrustedCA.Name = manifests.UserCABundle().Name
+		}
 		return nil
 	}); err != nil {
 		errs = append(errs, fmt.Errorf("failed to reconcile proxy config: %w", err))
@@ -1131,19 +1139,23 @@ func (r *reconciler) reconcileProxyTrustedCAConfigMap(ctx context.Context, hcp *
 
 	currentConfigMapRef := proxy.Spec.TrustedCA.Name
 	if currentConfigMapRef != "" && currentConfigMapRef != configMapRef {
-		// cleanup old configMaps
-		cm := &corev1.ConfigMap{}
-		cm.Name = currentConfigMapRef
+		// Do not clean up user-ca-bundle when it is being managed via additionalTrustBundle;
+		// reconcileConfig sets proxy.Spec.TrustedCA.Name to user-ca-bundle in that case.
+		if !(currentConfigMapRef == manifests.UserCABundle().Name && hcp.Spec.AdditionalTrustBundle != nil) {
+			// cleanup old configMaps
+			cm := &corev1.ConfigMap{}
+			cm.Name = currentConfigMapRef
 
-		// log and ignore deletion errors, should not disrupt normal workflow
-		cm.Namespace = hcp.Namespace
-		if err := r.cpClient.Delete(ctx, cm); err != nil {
-			log.Error(err, "failed to delete configmap", "name", cm.Name, "namespace", cm.Namespace)
-		}
+			// log and ignore deletion errors, should not disrupt normal workflow
+			cm.Namespace = hcp.Namespace
+			if err := r.cpClient.Delete(ctx, cm); err != nil {
+				log.Error(err, "failed to delete configmap", "name", cm.Name, "namespace", cm.Namespace)
+			}
 
-		cm.Namespace = manifests.ProxyTrustedCAConfigMap("").Namespace
-		if err := r.client.Delete(ctx, cm); err != nil {
-			log.Error(err, "failed to delete configmap in hosted cluster", "name", cm.Name, "namespace", cm.Namespace)
+			cm.Namespace = manifests.ProxyTrustedCAConfigMap("").Namespace
+			if err := r.client.Delete(ctx, cm); err != nil {
+				log.Error(err, "failed to delete configmap in hosted cluster", "name", cm.Name, "namespace", cm.Namespace)
+			}
 		}
 	}
 
@@ -1158,7 +1170,26 @@ func (r *reconciler) reconcileProxyTrustedCAConfigMap(ctx context.Context, hcp *
 
 	destCM := manifests.ProxyTrustedCAConfigMap(sourceCM.Name)
 	if _, err := r.CreateOrUpdate(ctx, r.client, destCM, func() error {
-		destCM.Data = sourceCM.Data
+		destCM.Data = make(map[string]string)
+		for k, v := range sourceCM.Data {
+			destCM.Data[k] = v
+		}
+		// Also merge additionalTrustBundle so that CNO writes both CAs into
+		// openshift-config-managed/trusted-ca-bundle, benefiting all consumers
+		// (including CVO's update service TLS verification).
+		if hcp.Spec.AdditionalTrustBundle != nil {
+			atbCM := &corev1.ConfigMap{}
+			if err := r.cpClient.Get(ctx, client.ObjectKey{Namespace: hcp.Namespace, Name: hcp.Spec.AdditionalTrustBundle.Name}, atbCM); err != nil {
+				return fmt.Errorf("failed to get additionalTrustBundle configmap %s/%s: %w", hcp.Namespace, hcp.Spec.AdditionalTrustBundle.Name, err)
+			}
+			if v := atbCM.Data[certs.UserCABundleMapKey]; v != "" {
+				existing := destCM.Data[certs.UserCABundleMapKey]
+				if existing != "" && !strings.HasSuffix(existing, "\n") {
+					existing += "\n"
+				}
+				destCM.Data[certs.UserCABundleMapKey] = existing + v
+			}
+		}
 		return nil
 	}); err != nil {
 		return fmt.Errorf("failed to reconcile referenced TrustedCA config map %s/%s: %w", destCM.Namespace, destCM.Name, err)
--- a/control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources_test.go
+++ b/control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources_test.go
@@ -637,6 +637,238 @@ func TestReconcileUserCertCABundle(t *testing.T) {
 	}
 }
 
+func TestReconcileProxyTrustedCAConfigMap(t *testing.T) {
+	const (
+		testNamespace = "master-cluster1"
+		testHCPName   = "cluster1"
+		proxyCAName   = "my-proxy-ca"
+	)
+
+	proxyCAData := "-----BEGIN CERTIFICATE-----\nMIIBfakeProxyCACert==\n-----END CERTIFICATE-----\n"
+	atbCAData := "-----BEGIN CERTIFICATE-----\nMIIBfakeAdditionalTrustBundleCert==\n-----END CERTIFICATE-----\n"
+
+	proxyCAConfigMap := func() *corev1.ConfigMap {
+		return &corev1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{Name: proxyCAName, Namespace: testNamespace},
+			Data:       map[string]string{"ca-bundle.crt": proxyCAData},
+		}
+	}
+	additionalTrustBundleCM := func() *corev1.ConfigMap {
+		return &corev1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{Name: "user-ca-bundle", Namespace: testNamespace},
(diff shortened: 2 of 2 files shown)


dpateriya:
## Verified

Tested manually on a HyperShift hosted cluster with `spec.additionalTrustBundle` set.

**Before the fix:**
- `proxy.config.openshift.io/cluster` on the guest had `spec.trustedCA.name: ""` 
- `openshift-config-managed/trusted-ca-bundle` did not contain the internal CA
- CVO failed with `x509: certificate signed by unknown authority` when contacting `spec.updateService`

**After patching HCCO with this fix:**
- `proxy.config.openshift.io/cluster` is now updated with `spec.trustedCA.name: "user-ca-bundle"`
- CNO picks up `user-ca-bundle` and merges it into `openshift-config-managed/trusted-ca-bundle`
- CVO reads `trusted-ca-bundle` for TLS verification and can now trust the internal CA

**Validation:**
```
$ oc get proxy cluster -o jsonpath='{.spec.trustedCA.name}' --kubeconfig hosted-kubeconfig
user-ca-bundle
```

The full data flow is now:

```
hc.spec.additionalTrustBundle
        │
        ▼
HCCO writes "openshift-config/user-ca-bundle" (already existed)
        │
        ▼  ← YOUR FIX: point proxy.spec.trustedCA.Name at "user-ca-bundle"
        │
CNO sees proxy.spec.trustedCA.Name is set
        │
        ▼
CNO merges it into "openshift-config-managed/trusted-ca-bundle"
        │
        ▼
CVO reads trusted-ca-bundle → TLS now trusts the internal CA → updateService works

```



/verified by me

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -1071,6 +1071,13 @@ func (r *reconciler) reconcileConfig(ctx context.Context, hcp *hyperv1.HostedCon
 	proxy := globalconfig.ProxyConfig()
 	if _, err := r.CreateOrUpdate(ctx, r.client, proxy, func() error {
 		globalconfig.ReconcileInClusterProxyConfig(proxy, hcp.Spec.Configuration)
+		// When no proxy.trustedCA is set but additionalTrustBundle is, point the
+		// Proxy object at openshift-config/user-ca-bundle (already written by
+		// reconcileUserCertCABundle) so CNO merges it into
+		// openshift-config-managed/trusted-ca-bundle.
+		if proxy.Spec.TrustedCA.Name == "" && hcp.Spec.AdditionalTrustBundle != nil {
```
jparrill, line 1079: Hey, I think there might be an ordering issue here that could cause a destructive loop.

`reconcileProxyTrustedCAConfigMap` runs **before** this block (line 1119 vs ~1123 within `reconcileConfig`). On the second reconcile, it reads the proxy object from the guest cluster and sees `currentConfigMapRef = "user-ca-bundle"` (set by this code on the previous loop). But since `configMapRef` is `""` (no proxy trustedCA configured), the cleanup block at lines 1185-1200 evaluates `currentConfigMapRef != configMapRef` as true and **deletes** `user-ca-bundle` from both the guest cluster and the control plane namespace.

If the `additionalTrustBundle` source ConfigMap in the management namespace is also named `user-ca-bundle` (which is the standard name from `cpomanifests.UserCAConfigMap`), this deletes the source data on the management side. That's a pretty dangerous side effect.

I think the cleanup block needs a guard, something like:

```go
if !(currentConfigMapRef == manifests.UserCABundle().Name && hcp.Spec.AdditionalTrustBundle != nil) {
    // existing cleanup logic
}
```

Or alternatively, handle the ATB-only case directly inside `reconcileProxyTrustedCAConfigMap` instead of relying on this synthetic mutation + `reconcileUserCertCABundle` running later.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -1071,6 +1071,13 @@ func (r *reconciler) reconcileConfig(ctx context.Context, hcp *hyperv1.HostedCon
...
+			proxy.Spec.TrustedCA.Name = "user-ca-bundle"
```
jparrill: nit: `"user-ca-bundle"` is already defined in `manifests.UserCABundle().Name` (same package, `manifests/pki.go:18`). Would be good to use that to avoid the magic string.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -1158,7 +1165,22 @@ func (r *reconciler) reconcileProxyTrustedCAConfigMap(ctx context.Context, hcp *
...
+		// Also merge additionalTrustBundle so that CNO writes both CAs into
+		// openshift-config-managed/trusted-ca-bundle, benefiting all consumers
+		// (including CVO's update service TLS verification).
+		if hcp.Spec.AdditionalTrustBundle != nil {
+			atbCM := &corev1.ConfigMap{}
+			if err := r.cpClient.Get(ctx, client.ObjectKey{Namespace: hcp.Namespace, Name: hcp.Spec.AdditionalTrustBundle.Name}, atbCM); err != nil {
+				return fmt.Errorf("failed to get additionalTrustBundle configmap %s/%s: %w", hcp.Namespace, hcp.Spec.AdditionalTrustBundle.Name, err)
+			}
+			if v := atbCM.Data["ca-bundle.crt"]; v != "" {
+				destCM.Data["ca-bundle.crt"] += v
```
jparrill: This concatenation could produce an invalid PEM bundle if the proxy CA data doesn't end with a trailing newline. You'd get something like:

```
-----END CERTIFICATE----------BEGIN CERTIFICATE-----
```

which most TLS stacks will silently fail to parse.

Something like this would be safer:

```go
existing := destCM.Data[certs.UserCABundleMapKey]
if existing != "" && !strings.HasSuffix(existing, "\n") {
    existing += "\n"
}
destCM.Data[certs.UserCABundleMapKey] = existing + v
```

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -1158,7 +1165,22 @@ func (r *reconciler) reconcileProxyTrustedCAConfigMap(ctx context.Context, hcp *
...
+		}
+		// Also merge additionalTrustBundle so that CNO writes both CAs into
+		// openshift-config-managed/trusted-ca-bundle, benefiting all consumers
+		// (including CVO's update service TLS verification).
+		if hcp.Spec.AdditionalTrustBundle != nil {
+			atbCM := &corev1.ConfigMap{}
+			if err := r.cpClient.Get(ctx, client.ObjectKey{Namespace: hcp.Namespace, Name: hcp.Spec.AdditionalTrustBundle.Name}, atbCM); err != nil {
+				return fmt.Errorf("failed to get additionalTrustBundle configmap %s/%s: %w", hcp.Namespace, hcp.Spec.AdditionalTrustBundle.Name, err)
+			}
+			if v := atbCM.Data["ca-bundle.crt"]; v != "" {
```
jparrill: nit: There's a constant for this — `certs.UserCABundleMapKey` in `support/certs/tls.go:50`. It's already used in `hostedcontrolplane_controller.go` for the same purpose.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources_test.go
```diff
@@ -637,6 +637,124 @@ func TestReconcileUserCertCABundle(t *testing.T) {
 	}
 }
 
+func TestReconcileProxyTrustedCAConfigMap(t *testing.T) {
```
jparrill, line 640: The test covers `reconcileProxyTrustedCAConfigMap` well, but I don't see any coverage for the new logic in `reconcileConfig` that sets `proxy.Spec.TrustedCA.Name = "user-ca-bundle"`. That's the critical piece that tells CNO to pick up the trust bundle — would be great to have tests for:

- ATB set, no proxy config → `proxy.Spec.TrustedCA.Name` becomes `"user-ca-bundle"`
- Both ATB and proxy trustedCA set → existing name is preserved (not overwritten)
- Neither set → name stays empty

Also, it would be worth adding an error case where the ATB ConfigMap doesn't exist on the management cluster (realistic scenario during provisioning or if there's a typo in the name).

A couple of extra nits:
- HyperShift convention is Gherkin-style test names ("When... it should..."), e.g. `"When both are set, it should merge both CA bundles"`
- These subtests look safe for `t.Parallel()` since each one creates independent fake clients

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources_test.go
```diff
@@ -637,6 +637,124 @@ func TestReconcileUserCertCABundle(t *testing.T) {
...
+					Configuration: &hyperv1.ClusterConfiguration{
+						Proxy: &configv1.ProxySpec{TrustedCA: configv1.ConfigMapNameReference{Name: proxyCAName}},
+					},
+				},
+			},
+			mgmtObjects:      []client.Object{proxyCAConfigMap(), additionalTrustBundleCM()},
+			guestObjects:     []client.Object{guestProxyObj()},
+			wantErr:          false,
+			wantDestCMExists: true,
+			wantDestCMData:   proxyCAData + atbCAData,
```
jparrill, line 712: The test uses plain strings here (`"proxy-cert-data"`, `"additional-trust-bundle-cert-data"`), which means it can't catch the PEM newline separator issue. If you used realistic PEM data (with `-----BEGIN/END CERTIFICATE-----` blocks), this assertion would fail with the current concatenation and surface the bug.

jparrill (commented):

jparrill:
> /verified by me

@dpateriya the `/verified` label should come from someone other than the PR author — ideally someone from QE or another engineer who has independently tested the fix in a real cluster. Self-verification doesn't give us the confidence that the change actually works end-to-end in a live environment, especially for a trust bundle fix like this where the failure mode (silent TLS rejection) is easy to miss in unit tests alone.

Could someone from QE pick this up and verify it on a hosted cluster with `additionalTrustBundle` + a custom update service behind an internal CA?

dpateriya:
The initial `/verified` label was added based on my own testing in a development environment, and I acknowledge that independent QE verification is needed before this can be considered fully validated.

**What was tested (dev environment):**
- Patched HCCO image deployed to a HyperShift hosted cluster
- Confirmed `proxy.config.openshift.io/cluster` gets `spec.trustedCA.name: user-ca-bundle` when only `hc.spec.additionalTrustBundle` is set
- Confirmed CNO subsequently creates `openshift-config-managed/trusted-ca-bundle` with the CA data

**Request:** Could someone from QE independently verify this fix covers the expected scenarios (ATB-only, both ATB+proxy.trustedCA, removal/cleanup)?

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources_test.go
dpateriya: addressed

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources_test.go
dpateriya: addressed

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
dpateriya: addressed

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
dpateriya: addressed

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
dpateriya: done

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
dpateriya: fixed

dpateriya:
@jparrill could you please review this PR once you LGTM, I will review this in a real env.

PR #8522, opened 2026-05-14
OCPBUGS-85151: Re-enable serviceaccount-pull-secrets controller when registry managementState changes from Removed

vsolanki12:

When the image registry `managementState` is set to `Removed`, HCCO disables the `serviceaccount-pull-secrets` controller in openshift-controller-manager by adding a `-openshift.io/serviceaccount-pull-secrets` override to the OCM ConfigMap. However, the logic was one-way — there was no code path to re-enable the controller when `managementState` changed back to `Managed`. Combined with CPO's preservation of existing controller overrides (PR #8072), this caused the controller to remain permanently disabled.

This change makes the logic bidirectional: when `managementState` is no longer `Removed` and the controller is currently disabled, it clears the controller override to re-enable it.

Fixes https://issues.redhat.com/browse/OCPBUGS-85151
  
Tested manually on a self-managed HyperShift cluster (AWS, release 4.22.0-rc.3). Verified:
1. Setting `managementState: Removed` disables the controller (log: "disabling serviceaccount-pull-secrets controller")
2. Setting `managementState: Managed` re-enables the controller (log: "re-enabling serviceaccount-pull-secrets controller", controllers reset to null)
3. CPO preservation logic correctly handles the cleared override

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Service account pull-secrets controller is now toggled off/on precisely based on image registry management state using a disable-token, restores wildcard behavior when needed, skips unnecessary config writes, and preserves IBM Cloud/Azure behavior.

* **Tests**
  * Added unit and reconciliation tests validating detection/removal of the disable-token and correct controller enable/disable behavior across state transitions and platforms.

### DIFF
--- a/control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
+++ b/control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net"
 	"reflect"
+	"slices"
 	"sort"
 	"strings"
 	"sync"
@@ -131,6 +132,8 @@ exec /bin/azure-cloud-node-manager \
   --enable-deprecated-beta-topology-labels \
   --wait-routes=false
 `
+
+var disabledServiceAccountPullSecretsController = fmt.Sprintf("-%s", openshiftcpv1.OpenShiftServiceAccountPullSecretsController)
 
 type reconciler struct {
 	client         client.Client
@@ -629,8 +632,7 @@ func (r *reconciler) reconcileRegistryAndIngress(ctx context.Context, hcp *hyper
 			}
 
 			// TODO: remove this when ROSA HCP stops setting the managementState to Removed to disable the Image Registry
-			if registryConfig.Spec.ManagementState == operatorv1.Removed && r.platformType != hyperv1.IBMCloudPlatform && r.platformType != hyperv1.AzurePlatform {
-				log.Info("imageregistry operator managementstate is removed, disabling openshift-controller-manager controllers and cleaning up resources")
+			if r.platformType != hyperv1.IBMCloudPlatform && r.platformType != hyperv1.AzurePlatform {
 				ocmConfigMap := cpomanifests.OpenShiftControllerManagerConfig(r.hcpNamespace)
 				if _, err := r.CreateOrUpdate(ctx, r.cpClient, ocmConfigMap, func() error {
 					if ocmConfigMap.Data == nil {
@@ -638,12 +640,28 @@ func (r *reconciler) reconcileRegistryAndIngress(ctx context.Context, hcp *hyper
 					}
 					config := &openshiftcpv1.OpenShiftControllerManagerConfig{}
 					if configStr, exists := ocmConfigMap.Data[ocm.ConfigKey]; exists && len(configStr) > 0 {
-						err := k8sutil.DeserializeResource(configStr, config, api.Scheme)
-						if err != nil {
+						if err := k8sutil.DeserializeResource(configStr, config, api.Scheme); err != nil {
 							return fmt.Errorf("unable to decode existing openshift controller manager configuration: %w", err)
 						}
 					}
-					config.Controllers = []string{"*", fmt.Sprintf("-%s", openshiftcpv1.OpenShiftServiceAccountPullSecretsController)}
+					if registryConfig.Spec.ManagementState == operatorv1.Removed {
+						if isServiceAccountPullSecretsControllerDisabled(config.Controllers) {
+							// Already disabled; returning nil without mutation causes CreateOrUpdate to skip the update.
+							return nil
+						}
+						log.Info("imageregistry operator managementstate is removed, disabling serviceaccount-pull-secrets controller")
+						if len(config.Controllers) == 0 {
+							config.Controllers = []string{"*", disabledServiceAccountPullSecretsController}
+						} else {
+							config.Controllers = append(config.Controllers, disabledServiceAccountPullSecretsController)
+						}
+					} else if isServiceAccountPullSecretsControllerDisabled(config.Controllers) {
+						log.Info("imageregistry operator managementstate is no longer removed, re-enabling serviceaccount-pull-secrets controller")
+						config.Controllers = removeDisabledServiceAccountPullSecretsController(config.Controllers)
+					} else {
+						// No change needed; returning nil without mutation causes CreateOrUpdate to skip the update.
+						return nil
+					}
 					configStr, err := k8sutil.SerializeResource(config, api.Scheme)
 					if err != nil {
 						return fmt.Errorf("failed to serialize openshift controller manager configuration: %w", err)
@@ -3650,4 +3668,21 @@ func imageRegistryPlatformWithPVC(platform hyperv1.PlatformType) bool {
 	default:
 		return false
 	}
+}
+
+func isServiceAccountPullSecretsControllerDisabled(controllers []string) bool {
+	return slices.Contains(controllers, disabledServiceAccountPullSecretsController)
+}
+
+func removeDisabledServiceAccountPullSecretsController(controllers []string) []string {
+	filtered := make([]string, 0, len(controllers))
+	for _, c := range controllers {
+		if c != disabledServiceAccountPullSecretsController {
+			filtered = append(filtered, c)
+		}
+	}
+	if len(filtered) == 0 {
+		return []string{"*"}
+	}
+	return filtered
 }
--- a/control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources_test.go
+++ b/control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources_test.go
@@ -14,6 +14,7 @@ import (
 
 	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
 	cpomanifests "github.com/openshift/hypershift/control-plane-operator/controllers/hostedcontrolplane/manifests"
+	"github.com/openshift/hypershift/control-plane-operator/controllers/hostedcontrolplane/ocm"
 	"github.com/openshift/hypershift/control-plane-operator/hostedclusterconfigoperator/api"
 	"github.com/openshift/hypershift/control-plane-operator/hostedclusterconfigoperator/controllers/resources/kas"
 	"github.com/openshift/hypershift/control-plane-operator/hostedclusterconfigoperator/controllers/resources/manifests"
@@ -21,6 +22,7 @@ import (
 	"github.com/openshift/hypershift/hypershift-operator/controllers/nodepool"
 	"github.com/openshift/hypershift/support/azureutil"
 	"github.com/openshift/hypershift/support/globalconfig"
+	"github.com/openshift/hypershift/support/k8sutil"
 	"github.com/openshift/hypershift/support/netutil"
 	"github.com/openshift/hypershift/support/releaseinfo"
 	fakereleaseprovider "github.com/openshift/hypershift/support/releaseinfo/fake"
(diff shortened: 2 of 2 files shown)


control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -629,21 +629,27 @@ func (r *reconciler) reconcileRegistryAndIngress(ctx context.Context, hcp *hyper
...
 						}
 					}
-					config.Controllers = []string{"*", fmt.Sprintf("-%s", openshiftcpv1.OpenShiftServiceAccountPullSecretsController)}
+					if registryConfig.Spec.ManagementState == operatorv1.Removed {
+						log.Info("imageregistry operator managementstate is removed, disabling serviceaccount-pull-secrets controller")
+						config.Controllers = []string{"*", fmt.Sprintf("-%s", openshiftcpv1.OpenShiftServiceAccountPullSecretsController)}
+					} else if isServiceAccountPullSecretsControllerDisabled(config.Controllers) {
+						log.Info("imageregistry operator managementstate is no longer removed, re-enabling serviceaccount-pull-secrets controller")
+						config.Controllers = nil
+					} else {
```
bryan-cox, line 661: **[Medium]** Consider using `[]string{"*"}` instead of `nil` here.

The `Controllers` field is tagged `json:"controllers"` (no `omitempty`), so `nil` serializes to `"controllers": null`. Whether OCM interprets `null` as "use default (all controllers)" vs "empty list (run nothing)" is ambiguous.

Using `[]string{"*"}` is explicit and unambiguous — it says "run all on-by-default controllers," which is the documented default and the intent here.

This also makes the CPO interaction in `adaptConfig` deterministic: with `nil`, CPO's `len(existingControllers) > 0` is false so it enters neither branch and relies on the base config default. With `["*"]`, CPO explicitly preserves it via the `else if` branch.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -629,21 +629,27 @@ func (r *reconciler) reconcileRegistryAndIngress(ctx context.Context, hcp *hyper
...
+						return nil
+					}
```
bryan-cox, line 664: **[Low — readability]** This `return nil` without mutation is correct — `CreateOrUpdate` deep-compares before/after and skips the API write when nothing changed. But the intent is non-obvious to readers unfamiliar with the upsert framework. A short comment would help:

```go
} else {
	// No change needed; returning nil without mutation causes CreateOrUpdate to skip the update.
	return nil
}
```

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -3651,3 +3657,14 @@ func imageRegistryPlatformWithPVC(platform hyperv1.PlatformType) bool {
 		return false
 	}
 }
+
+// isServiceAccountPullSecretsControllerDisabled checks if the ServiceAccountPullSecretsController is disabled.
+func isServiceAccountPullSecretsControllerDisabled(controllers []string) bool {
+	disabled := fmt.Sprintf("-%s", openshiftcpv1.OpenShiftServiceAccountPullSecretsController)
```
bryan-cox: **[Low — DRY]** This `fmt.Sprintf("-%s", openshiftcpv1.OpenShiftServiceAccountPullSecretsController)` also appears at line ~647 in the reconcile logic. Consider extracting to a package-level var:

```go
var disabledServiceAccountPullSecretsController = fmt.Sprintf("-%s", openshiftcpv1.OpenShiftServiceAccountPullSecretsController)
```

Then use it in both the slice literal and this helper. The duplication is low-risk since it's derived from a typed constant, but extracting it makes the relationship between the two usages explicit.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources_test.go
```diff
@@ -3646,3 +3646,45 @@ func TestCleanupLegacyResources(t *testing.T) {
 		})
 	}
 }
```
bryan-cox, line 3652: **[Medium — test gap]** The helper function tests are well-structured (table-driven, Gherkin names, gomega) — nice work. However, the core behavioral change — the bidirectional toggling in `reconcileRegistryAndIngress` — has no test coverage. The re-enable path (`Removed` → `Managed`) is the entire point of this PR.

Consider adding a test that:
1. Sets up a reconciler with a `cpClient` containing an OCM ConfigMap where the controller is already disabled (`Controllers: ["*", "-openshift.io/serviceaccount-pull-secrets"]`)
2. Sets `registryConfig.Spec.ManagementState = Managed`
3. Calls the reconcile path
4. Asserts `config.Controllers` is reset (to `["*"]` or nil depending on chosen approach)

Also covers the idempotent no-op case: `ManagementState = Managed` with controller already enabled → no ConfigMap update.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources_test.go
```diff
@@ -3646,3 +3646,45 @@ func TestCleanupLegacyResources(t *testing.T) {
...
+		{
+			name:        "When only the wildcard is present, it should return false",
+			controllers: []string{"*"},
+			expected:    false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			g := NewWithT(t)
```
bryan-cox, line 3955: **[Low]** Nit: subtests are independent and could benefit from `t.Parallel()` inside each `t.Run` callback, consistent with Go best practices for table-driven tests.

bryan-cox (commented):
Pre-commit review (Go + HyperShift profile). Overall: **PASS WITH RECOMMENDATIONS** — the fix is correct and does not introduce new architectural violations. See inline comments for specific suggestions.

csrwng (commented):
Thanks for the fix — the bidirectional logic is a good improvement. Left a couple of comments.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
```diff
@@ -629,21 +629,27 @@ func (r *reconciler) reconcileRegistryAndIngress(ctx context.Context, hcp *hyper
...
 							return fmt.Errorf("unable to decode existing openshift controller manager configuration: %w", err)
 						}
 					}
-					config.Controllers = []string{"*", fmt.Sprintf("-%s", openshiftcpv1.OpenShiftServiceAccountPullSecretsController)}
+					if registryConfig.Spec.ManagementState == operatorv1.Removed {
+						log.Info("imageregistry operator managementstate is removed, disabling serviceaccount-pull-secrets controller")
+						config.Controllers = []string{"*", fmt.Sprintf("-%s", openshiftcpv1.OpenShiftServiceAccountPullSecretsController)}
+					} else if isServiceAccountPullSecretsControllerDisabled(config.Controllers) {
+						log.Info("imageregistry operator managementstate is no longer removed, re-enabling serviceaccount-pull-secrets controller")
+						config.Controllers = nil
```
csrwng: Setting `config.Controllers = nil` is fragile — it would wipe out any other controller overrides that might exist in the list, not just the `-openshift.io/serviceaccount-pull-secrets` entry. Instead, consider removing only the specific disable entry (e.g., `-openshift.io/serviceaccount-pull-secrets`) from the slice, and then set `Controllers = nil` only if the remaining list is just `["*"]` (or empty).

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources_test.go
```diff
@@ -3646,3 +3646,45 @@ func TestCleanupLegacyResources(t *testing.T) {
 		})
 	}
 }
+
+func TestIsServiceAccountPullSecretsControllerDisabled(t *testing.T) {
```
csrwng, line 3918: nit: The helper function is well-tested, but it would be good to also have reconciliation-level tests covering the full `reconcileRegistryAndIngress` flow — e.g., verifying that when `managementState` transitions from `Removed` to `Managed`, the OCM ConfigMap's `Controllers` field is correctly updated, and conversely that it stays untouched when the state hasn't changed.

vsolanki12: Done. Added `TestReconcileRegistryAndIngress_ServiceAccountPullSecretsController` with 4 test cases: disable when Removed, re-enable when Managed, no-op when already enabled, and platform exclusion for IBMCloud.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
vsolanki12: Thank you @csrwng, 
Done. Now removes only the `-openshift.io/serviceaccount-pull-secrets` entry from the slice, preserving any other controller overrides. Falls back to `[]string{"*"}` if the remaining list is empty.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources_test.go
vsolanki12: Done. Added `t.Parallel()` inside each `t.Run` callback.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources_test.go
vsolanki12: Done. Added `TestReconcileRegistryAndIngress_ServiceAccountPullSecretsController` with 4 test cases: disable when Removed, re-enable when Managed, no-op when already enabled, and platform exclusion for IBMCloud.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
vsolanki12: Extracted `var disabledServiceAccountPullSecretsController` and used it in both the reconcile logic and the helper function.

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
vsolanki12: Done added. `// No change needed; returning nil without mutation causes CreateOrUpdate to skip the update.`

control-plane-operator/hostedclusterconfigoperator/controllers/resources/resources.go
vsolanki12: Thanks @bryan-cox , 
Done. Changed to selective removal of the disabled entry, falling back to `[]string{"*"}` when the remaining list is empty.

vsolanki12:
@coderabbitai full review

vsolanki12:
Before fix:
```
$ oc get cm openshift-controller-manager-config -n clusters-vsolanki-hcp-test-2 -o jsonpath='{.data.config\.yaml}' | yq '.controllers'
- "*"
- "-openshift.io/serviceaccount-pull-secrets"
```

After fix:
```
$ oc get cm openshift-controller-manager-config -n clusters-vsolanki-hcp-test-2 -o jsonpath='{.data.config\.yaml}' | yq '.controllers'
- "*"
```

vsolanki12:
**Before Fix:**
```
$ oc get cm openshift-controller-manager-config -n clusters-vsolanki-hcp-test-2 -o jsonpath='{.data.config\.yaml}' | yq '.controllers'
- "*"
- "-openshift.io/serviceaccount-pull-secrets"
```

**After fix:**
```
$ oc get cm openshift-controller-manager-config -n clusters-vsolanki-hcp-test-2 -o jsonpath='{.data.config\.yaml}' | yq '.controllers'
- "*"
```

PR #8536, opened 2026-05-18
OCPBUGS-86020: OCPBUGS-62177: [release 4-20] verify cert revocation against all KAS pods

sdminonne:
Updates the CertificateRevocationController to verify certificate trust and revocation against every individual KAS pod rather than through the service load balancer. In HA deployments with 3 KAS replicas, the service endpoint load-balances to a single pod, so the check could hit a pod that had loaded the updated trust bundle while others hadn't, causing premature state transitions in the revocation flow.

Fixes https://redhat.atlassian.net/browse/OCPBUGS-86020

The dual-cert cross-checks are the key addition beyond the original fix. In ensureOldSignerCertificateRevoked, verifying the new cert is trusted before declaring the old one revoked prevents false positives when a KAS pod restart during revocation causes the old cert to become unreachable (connection refused), which looks like revocation but is actually the pod being down. Similarly, in ensureNewSignerCertificatePropagated, verifying the old cert is also still trusted prevents declaring propagation complete when a pod is mid-reload and happens to accept the new cert but hasn't finished loading the full trust bundle.

## Checklist:


sdminonne:
```
TestCreateClusterHABreakGlassCredentials passed. All subtests passed (no <failure> tags), including the key ones:                                                                                                              
                                                                                                                                                                                                                               
  - Main/break-glass-credentials/customer-break-glass/CSR_flow/revocation — 153s, passed                                                                                                                                         
  - Main/break-glass-credentials/sre-break-glass/CSR_flow/revocation — 171s, passed
  - Main/break-glass-credentials/independent_signers — 174s, passed                                                                                                                                                              
  - ValidateHostedCluster — 433s, passed                                                                                                                                                                                       
  - Teardown — 274s, passed
  - TestCreateClusterHABreakGlassCredentials (top-level) — 1625s (27 min), passed
```

PR #8537, opened 2026-05-18
CNTRLPLANE-3250,CNTRLPLANE-430: API-driven Azure topology and private connectivity (Phase 1)

muraee:
## Summary

- Extend `AzurePrivateType` enum with `Swift` variant and add `AzureSwiftSpec` struct with immutable `podNetworkInstance` field
- Add per-cluster API-driven helpers (`UseSwiftNetworkingHCP/HC`, `UseSharedIngressHCP/HC`, `IsAroHCPByHCP/HC`) that derive behavior from `Topology` and `Private` API fields instead of `isAroHCP()` env-var checks
- Replace all per-cluster `isAroHCP()` / `UseSharedIngress()` calls in visibility, router, shared ingress, infra, nodepool, and CPO v2 component controllers with API-driven helpers
- Add `PublicEndpointExposed` condition type for tracking public endpoint exposure state
- Phase 1 backward compatibility: legacy annotation fallback in `UseSwiftNetworkingHCP()` for unmigrated clusters

This is Phase 1 of the [API-driven Azure topology and private connectivity enhancement](https://github.com/openshift/enhancements/blob/master/enhancements/hypershift/api-driven-azure-topology-and-private-connectivity.md). Phase 2 (ARO-RP data migration) and Phase 3 (legacy cleanup) are out of scope.

### API (`api/hypershift/v1beta1/azure.go`)
- `AzurePrivateType`: `Enum=PrivateLink;Swift`
- `AzureSwiftSpec`: `podNetworkInstance` (required, immutable, max 255 chars)
- `AzurePrivateSpec`: struct-level type immutability (`!has(oldSelf.type) || self.type == oldSelf.type`) enabling Phase 2 migration, positive/negative union member rules

### Helper Functions (`support/netutil/visibility.go`, `support/azureutil/azureutil.go`)
- `UseSwiftNetworkingHCP/HC()` — API field first, env var + annotation fallback
- `UseSharedIngressHCP/HC()` — `UseSwiftNetworking && IsPublic`
- `IsAroHCPByHCP/HC()` — per-cluster ARO wrapper for component predicates
- `IsPrivateHCP/HC()` — Phase 1 fallback for `Topology=""` + Swift annotation
- `LabelHCPRoutes()` — `UseSharedIngress || UseSwiftNetworking`

### Controller Changes
- **Config-generator**: cluster-level `UseSharedIngressHC()` filter (security-critical — prevents Private cluster endpoints from leaking into public HAProxy config)
- **CPO v2 components** (CNO, CCM, storage, registry, ingress, KAS): `IsAroHCPByHCP(cpContext.HCP)` replaces env-var predicates
- **Infra**: `UseSwiftNetworkingHCP()` for ClusterIP vs LB router service decision
- **NodePool HAProxy, network policies, KAS healthcheck**: per-cluster `UseSharedIngressHCP/HC()`

### Not Changed (management-cluster-level, kept as env-var)
- `hypershift-operator/main.go` — controller startup
- `control-plane-operator/main.go` — controller startup
- `cmd/cluster/core/dump.go` — dump command

## Test plan
- [x] Unit tests: visibility helpers, serialization compatibility
- [x] Envtest CEL validation: 10 test cases (5 onCreate, 5 onUpdate) for Swift union rules
- [ ] E2E: existing ARO HCP tests pass (CI)
- [ ] Manual: verify `UseSharedIngressHC()` correctly filters Private+Swift clusters from HAProxy config

## Summary by CodeRabbit

* **New Features**
  * Azure Swift networking added as an alternative to PrivateLink.
  * New PublicEndpointExposed condition to track public API reachability via shared ingress.

* **Improvements**
  * Per-cluster Swift/shared-ingress and ARO detection for more accurate behavior in mixed environments.
  * Routing, ingress, router and storage flows now respect Swift/shared-ingress decisions.

* **Behavior/Validation**
  * Azure private spec supports a Swift member with exclusive-selection and immutability rules.

* **Tests**
  * Added serialization compatibility and public-endpoint condition tests.


api/hypershift/v1beta1/azure.go
```diff
@@ -686,6 +695,13 @@ type AzurePrivateSpec struct {
 	// +optional
 	// +unionMember
 	PrivateLink AzurePrivateLinkSpec `json:"privateLink,omitzero"`
+
+	// swift configures Azure Swift pod networking for private API server access.
+	// This field is required when type is "Swift" and must not be set otherwise.
+	//
```
enxebre, line 701: can we document that this requires management cluster setup, not handled by hypershift automatically

api/hypershift/v1beta1/azure.go
```diff
@@ -716,6 +732,24 @@ type AzurePrivateLinkSpec struct {
 	AdditionalAllowedSubscriptions []AzureSubscriptionID `json:"additionalAllowedSubscriptions,omitempty"`
 }
 
+// AzureSwiftSpec configures Azure Swift pod networking for private API server access.
+// Swift assigns a private IP from the customer VNet directly to the hosted cluster's
+// router pods, providing private connectivity without a separate Private Link Service.
+//
+// +kubebuilder:validation:XValidation:rule="self.podNetworkInstance == oldSelf.podNetworkInstance",message="podNetworkInstance is immutable"
+type AzureSwiftSpec struct {
+	// podNetworkInstance is the name of the Azure Swift pod network instance that
```
enxebre: can we document the field is immutable

support/netutil/visibility.go
```diff
@@ -148,9 +205,10 @@ func IsPublicHC(hc *hyperv1.HostedCluster) bool {
 // Returns true when routes should be labeled for HCP router; false when routes should
 // use the management cluster router.
 func LabelHCPRoutes(hcp *hyperv1.HostedControlPlane) bool {
-	// When shared ingress is active (e.g., ARO HCP), all routes must be labeled
-	// so the SharedIngressReconciler can discover and admit them.
-	if isAroHCP() {
+	// When shared ingress or Swift networking is active, all routes must be
```
enxebre, line 235: this is needed for the sharedingress router and/or for the internal per HC "swift" router, correct?

control-plane-operator/controllers/hostedcontrolplane/v2/router/util/util.go
```diff
@@ -17,11 +16,8 @@ func UseHCPRouter(hcp *hyperv1.HostedControlPlane) bool {
 	if hcp.Spec.Platform.Type == hyperv1.IBMCloudPlatform {
 		return false
 	}
-	// For ARO HCP, the dedicated HCP router is only needed when the cluster is private
-	// (Swift enabled). LabelHCPRoutes returns true for all ARO to support the
-	// SharedIngressReconciler, but that doesn't mean a dedicated router deployment is needed.
-	if azureutil.IsAroHCP() {
-		return netutil.IsPrivateHCP(hcp)
+	if netutil.UseSwiftNetworkingHCP(hcp) {
```
enxebre: do we really need this special handling here?
I think we should just let l25 decide and drop this one

sharedingress-config-generator/config.go
```diff
@@ -85,6 +85,10 @@ func generateRouterConfig(ctx context.Context, client crclient.Client, w io.Writ
 		if !hc.DeletionTimestamp.IsZero() {
 			continue
 		}
+		// Skip clusters that don't use shared ingress (Private topology or non-Swift).
+		if !netutil.UseSharedIngressHC(&hc) {
```
enxebre, line 89: why does this change doesn't come with test cases changes to cover it?

control-plane-operator/controllers/hostedcontrolplane/infra/infra_test.go
```diff
@@ -568,6 +574,12 @@ func TestReconcileInfrastructure(t *testing.T) {
 				hcp.Annotations = map[string]string{
 					hyperv1.SwiftPodNetworkInstanceAnnotation: "swift-network-instance",
```
enxebre, line 606: can we have separate test cases, one exercising the annotation, another one the API, another one both?

control-plane-operator/controllers/hostedcontrolplane/infra/infra_test.go
```diff
@@ -1571,16 +1583,44 @@ func TestReconcileHCPRouterServices(t *testing.T) {
 			expectedServices:             nil,
 		},
 		{
-			name:                         "When ARO is enabled it should not create any services",
+			name:                         "When ARO is enabled it should create only a ClusterIP private router service",
```
enxebre: why is this changing?

control-plane-operator/controllers/hostedcontrolplane/infra/infra_test.go
```diff
@@ -1617,6 +1657,14 @@ func TestReconcileHCPRouterServices(t *testing.T) {
 				hcp.Annotations = map[string]string{
 					hyperv1.SwiftPodNetworkInstanceAnnotation: "swift-network-instance",
 				}
+				hcp.Spec.Platform.Azure = &hyperv1.AzurePlatformSpec{
```
enxebre, line 1735: can we have separate test cases for all our possible states i.e. only annotation, only api, both

control-plane-operator/controllers/hostedcontrolplane/infra/infra_test.go
```diff
@@ -1803,6 +1851,12 @@ func TestReconcileInternalRouterServiceStatus(t *testing.T) {
 						Type: hyperv1.AzurePlatform,
 						Azure: &hyperv1.AzurePlatformSpec{
 							Location: "eastus",
+							Private: hyperv1.AzurePrivateSpec{
```
enxebre, line 1980: can we have separate test cases for all our possible states i.e. only annotation, only api, both

support/netutil/visibility.go
```diff
@@ -14,25 +15,83 @@ func isAroHCP() bool {
...
 }
 
-func IsPrivateHCP(hcp *hyperv1.HostedControlPlane) bool {
-	// ARO always have swift enabled.
-	// We still check the annotation to keep CI working.
-	// TODO(alberto): Remove this once CI has swift enabled.
+// UseSwiftNetworkingHCP returns true when the HCP uses Azure Swift pod networking.
+// Checks the Private.Type API field first; falls back to env var + annotation
+// for unmigrated clusters during the Phase 1 migration window.
+func UseSwiftNetworkingHCP(hcp *hyperv1.HostedControlPlane) bool {
```
enxebre, line 21: does this func has unit coverage?

support/netutil/visibility.go
```diff
@@ -14,25 +15,83 @@ func isAroHCP() bool {
...
+		klog.V(2).Info("Using legacy annotation fallback for Swift networking detection on HostedCluster", "namespace", hc.Namespace, "name", hc.Name)
+		return true
+	}
+	return false
+}
+
+// UseSharedIngressHCP returns true when this specific HCP should use the
+// management cluster's shared ingress (HAProxy) for public endpoints.
+// Only Swift clusters with public topology use shared ingress.
+func UseSharedIngressHCP(hcp *hyperv1.HostedControlPlane) bool {
```
enxebre, line 82: does this func has unit coverage?

support/netutil/visibility.go
```diff
@@ -14,25 +15,83 @@ func isAroHCP() bool {
...
+// UseSharedIngressHC returns true when this specific HostedCluster should use
+// the management cluster's shared ingress for public endpoints.
+func UseSharedIngressHC(hc *hyperv1.HostedCluster) bool {
+	return UseSwiftNetworkingHC(hc) && IsPublicHC(hc)
+}
+
+// SwiftPodNetworkInstanceHCP returns the Swift pod network instance name for
+// the HCP. Checks the API field first; falls back to the annotation for
+// unmigrated clusters.
+func SwiftPodNetworkInstanceHCP(hcp *hyperv1.HostedControlPlane) string {
```
enxebre, line 95: does this func has unit coverage?

api/hypershift/v1beta1/azure.go
```diff
@@ -653,31 +653,40 @@ const (
...
 
 // AzurePrivateSpec configures private connectivity to an Azure hosted cluster's API server.
 // It is a discriminated union keyed on the type field, which selects the private connectivity
-// mechanism. Currently only PrivateLink is supported; additional mechanisms (e.g., Swift) may
-// be added in the future.
+// mechanism.
 //
-// +kubebuilder:validation:XValidation:rule="self.type != 'PrivateLink' ? !has(self.privateLink) : true",message="privateLink is forbidden when type is not PrivateLink"
+// +kubebuilder:validation:XValidation:rule="!has(oldSelf.type) || self.type == oldSelf.type",message="type is immutable"
+// +kubebuilder:validation:XValidation:rule="self.type != 'Swift' || !has(self.privateLink)",message="privateLink must not be set when type is Swift"
```
enxebre: instead of all these, can we please align with https://github.com/openshift/hypershift/pull/8490#discussion_r3227109842

muraee: Done. Consolidated 4 separate rules into 2 ternary rules per JoelSpeed's pattern:
- `self.type == 'PrivateLink' ? has(self.privateLink) : !has(self.privateLink)`
- `self.type == 'Swift' ? has(self.swift) : !has(self.swift)`

---

api/hypershift/v1beta1/azure.go
muraee: Done. Added documentation that Swift networking requires the management cluster to be pre-configured with Azure Swift support, not provisioned by HyperShift automatically.

---

api/hypershift/v1beta1/azure.go
muraee: Done. Updated godoc to explicitly state "This field is immutable once set" in the podNetworkInstance documentation.

---

control-plane-operator/controllers/hostedcontrolplane/v2/router/util/util.go
muraee: Done. Removed the UseSwiftNetworkingHCP special case — line 25 (`IsPrivateHCP || LabelHCPRoutes`) already covers all Swift cases correctly:
- Swift + PublicAndPrivate/Private: IsPrivateHCP returns true
- Swift + empty topology: IsPrivateHCP falls back to UseSwiftNetworking
- Swift + Public: LabelHCPRoutes returns true via UseSwiftNetworking

---

sharedingress-config-generator/config.go
muraee: The config_test.go test cases were updated in the \`test(azure): update tests to use Swift API fields\` commit — the test HCs now set Swift API fields so they pass the UseSharedIngressHC filter. The second test case (\"When ARO Swift is enabled\") specifically tests that Private clusters skip the KAS backend.

---

support/netutil/visibility.go
muraee: Both. Routes are labeled for two consumers:
1. **SharedIngressReconciler** — discovers labeled routes to populate HAProxy config and admit them (set RouterCanonicalHostname for ExternalDNS)
2. **Per-HC HCP router** — the dedicated router deployment serves labeled routes for private access via Swift

For PublicAndPrivate+Swift clusters, both consumers use the labels. For Private+Swift, only the HCP router uses them (SharedIngressReconciler's config-generator filters the cluster out).

---

control-plane-operator/controllers/hostedcontrolplane/infra/infra_test.go
muraee: Done. Split into three separate test cases:
- **AnnotationFallback**: env var + annotation only, no API fields (legacy path)
- **APIField**: Private.Type=Swift only, no annotation (new path)
- **Both**: annotation AND API field set (coexistence during migration)

Applied the same pattern to TestReconcileHCPRouterServices and TestReconcileInternalRouterServiceStatus.

---

control-plane-operator/controllers/hostedcontrolplane/infra/infra_test.go
muraee: The test name was changed to better describe the tested behavior and now has three separate variants (annotation-only, API-only, both) rather than a single mixed test.

---

support/netutil/visibility.go
muraee: Yes, added TestUseSwiftNetworkingHCP and TestUseSwiftNetworkingHC covering: API field set, annotation fallback, neither, non-Azure platform, and PrivateLink type.

---

support/netutil/visibility.go
muraee: Yes, added TestUseSharedIngressHCP and TestUseSharedIngressHC covering: Swift+PublicAndPrivate (true), Swift+Private (false), Swift+empty topology (true), non-Swift (false).

---

support/netutil/visibility.go
muraee: Yes, added TestSwiftPodNetworkInstanceHCP covering: API field value, annotation fallback value, neither (empty string), non-Azure platform (empty string).

---

control-plane-operator/controllers/hostedcontrolplane/infra/infra_test.go
muraee: Done. Split into annotation-only, API-only, and both variants. (Also addresses the same feedback at line 1854.)

---

control-plane-operator/controllers/hostedcontrolplane/infra/infra_test.go
muraee: Done. Same three-variant pattern applied here.

---

enxebre:
do we need any changes for nodepool_controller.go:1050 and haproxy.go:160,178

control-plane-operator/controllers/hostedcontrolplane/v2/router/util/util.go
enxebre, line 27: this will return true for LabelHCPRoutes when shared ingress but not private i.e. CI.
That will deploy a noop router? Is that intended?

muraee: Good catch. Yes, in CI where `MANAGED_SERVICE=ARO-HCP` is set but Swift PNI is NOT configured (no annotation, no API field), `UseSharedIngressHC/HCP()` returns `false` for all clusters because `UseSwiftNetworkingHC/HCP()` returns `false`.

**Impact in CI without Swift PNI:**

- **Management-cluster level** (`UseSharedIngress()` in `main.go`): Still env-var based, unchanged. The SharedIngressReconciler IS started and deploys HAProxy infrastructure.
- **Per-cluster level** (`UseSharedIngressHC/HCP()`): Returns `false`. This means:
  - Config-generator skips all clusters → HAProxy config has zero backends
  - Worker node HAProxy (`haproxy.go:160`): `UseSharedIngressHC() && !IsPrivateHC()` = `false` → workers don't use shared ingress LB
  - Network policies: no shared ingress policy created

This IS a behavioral change from the old code. Previously:
- Old: `UseSharedIngress()` (env var) = `true` AND `!IsPrivateHC()` = `true` → workers DID use shared ingress LB
- New: `UseSharedIngressHC()` = `false` → workers DON'T use shared ingress LB

So in CI without Swift PNI, the shared ingress infrastructure is deployed but idle — no cluster routes through it. CI would need to set the Swift PNI (via annotation or API field) for shared ingress to be exercised.

Regarding the router question: `UseHCPRouter()` = `IsPrivateHCP() || LabelHCPRoutes()`. In CI without Swift, both return `false` (empty topology, no Swift fallback, LB KAS) → no router deployed. No noop router.

---

saschagrunert (commented):
## API Review

**Test coverage:** 5 onCreate + 5 onUpdate CEL validation tests, plus unit tests for visibility helpers, serialization compatibility, and the PublicEndpointExposed condition. No ratcheting tests.

### Findings

| # | Severity | Finding |
|---|----------|---------|
| 1 | must-fix | Validation tightening (privateLink now required with type=PrivateLink) needs ratcheting test |
| 2 | must-fix | Em dashes in API field godocs |
| 3 | should-fix | `podNetworkInstance` lacks pattern/format validation |
| 4 | should-fix | `IsAroHCPByHCP` name conflates "is ARO" with "uses Swift" |
| 5 | should-fix | `UseSharedIngressHCP` fallback is broader than prior behavior |

Union pattern, serialization (`omitzero` on struct members), immutability rules, and the security-critical shared ingress filtering all look correct.

api/hypershift/v1beta1/azure.go
```diff
@@ -653,31 +653,38 @@ const (
...
 
 // AzurePrivateSpec configures private connectivity to an Azure hosted cluster's API server.
 // It is a discriminated union keyed on the type field, which selects the private connectivity
-// mechanism. Currently only PrivateLink is supported; additional mechanisms (e.g., Swift) may
-// be added in the future.
+// mechanism.
 //
-// +kubebuilder:validation:XValidation:rule="self.type != 'PrivateLink' ? !has(self.privateLink) : true",message="privateLink is forbidden when type is not PrivateLink"
+// +kubebuilder:validation:XValidation:rule="!has(oldSelf.type) || self.type == oldSelf.type",message="type is immutable"
+// +kubebuilder:validation:XValidation:rule="self.type == 'PrivateLink' ? has(self.privateLink) : !has(self.privateLink)",message="privateLink is required when type is PrivateLink, and forbidden otherwise"
```
saschagrunert, line 677: This validation tightening needs a ratcheting test.

The old rule only forbade `privateLink` when type was not `PrivateLink`. This new rule also requires it when type is `PrivateLink`. That is a validation tightening: on `main` today, `type: PrivateLink` without a `privateLink` field is valid. After this PR, it is not (confirmed by the 3 existing tests that had to be patched to add `privateLink.natSubnetID`).

Since this CEL rule does not reference `oldSelf`, it qualifies for automatic ratcheting (k8s >= 1.30), so existing objects should survive updates that don't touch the `private` block. But a ratcheting test using `initialCRDPatches` should be added to prove this. The test should create an object with `type: PrivateLink` without `privateLink` (bypassing validation via the patch), then verify an unrelated spec update (e.g., changing `baseDomain`) succeeds.

saschagrunert (commented):
Gave it an initial shadow review, see inline comments for more details.

api/hypershift/v1beta1/azure.go
```diff
@@ -686,6 +693,15 @@ type AzurePrivateSpec struct {
 	// +optional
 	// +unionMember
 	PrivateLink AzurePrivateLinkSpec `json:"privateLink,omitzero"`
+
+	// swift configures Azure Swift pod networking for private API server access.
+	// Swift networking requires the management cluster to be pre-configured with
+	// Azure Swift support — this is not provisioned by HyperShift automatically.
```
saschagrunert: nit: em dash (U+2014) in API field comments. These end up in `oc explain` output and generated docs, and some terminals/renderers don't handle them well.

```suggestion
	// Azure Swift support; this is not provisioned by HyperShift automatically.
```

api/hypershift/v1beta1/azure.go
```diff
@@ -716,6 +732,24 @@ type AzurePrivateLinkSpec struct {
...
+// AzureSwiftSpec configures Azure Swift pod networking for private API server access.
+// Swift assigns a private IP from the customer VNet directly to the hosted cluster's
+// router pods, providing private connectivity without a separate Private Link Service.
+//
+// +kubebuilder:validation:XValidation:rule="self.podNetworkInstance == oldSelf.podNetworkInstance",message="podNetworkInstance is immutable"
+type AzureSwiftSpec struct {
+	// podNetworkInstance is the name of the Azure Swift pod network instance that
+	// provides private connectivity to the hosted cluster's router pods. This value
+	// is assigned by the management service.
+	// This field is immutable once set — changing it would disrupt Swift networking
```
saschagrunert: Same em dash issue.

```suggestion
	// This field is immutable once set. Changing it would disrupt Swift networking
```

api/hypershift/v1beta1/azure.go
```diff
@@ -716,6 +732,24 @@ type AzurePrivateLinkSpec struct {
...
+	// by reassigning the router pod's private IP.
+	//
+	// +required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=255
+	PodNetworkInstance string `json:"podNetworkInstance,omitempty"`
```
saschagrunert, line 752: `MaxLength=255` with no character set restriction is very permissive. Azure resource names typically follow constrained patterns. Consider adding a CEL rule that restricts the allowed character set with a human-readable error message, or document explicitly in the godoc that the format is unconstrained.

support/azureutil/azureutil.go
```diff
@@ -260,11 +261,25 @@ func IsPrivateKeyVault(hcp *hyperv1.HostedControlPlane) bool {
...
+// Use this only for management-cluster-level decisions where no HC/HCP context is available.
+// For per-cluster decisions, use IsAroHCPByHCP or IsAroHCPByHC instead.
 func IsAroHCP() bool {
 	return os.Getenv("MANAGED_SERVICE") == hyperv1.AroHCP
 }
 
+// IsAroHCPByHCP returns true when this HCP belongs to an ARO-managed cluster.
+// API-driven: uses Private.Type == Swift as the per-cluster indicator for ARO
+// management. Includes legacy fallback for unmigrated clusters during Phase 1.
+func IsAroHCPByHCP(hcp *hyperv1.HostedControlPlane) bool {
```
saschagrunert, line 274: `IsAroHCPByHCP` delegates to `UseSwiftNetworkingHCP`, equating "is ARO" with "uses Swift." Every caller actually wants "uses managed networking," not "is ARO." If a non-ARO cluster ever uses Swift, or an ARO cluster uses PrivateLink during Phase 2 migration before Swift is configured, this breaks.

Consider renaming to reflect what it actually checks, or at minimum add a godoc note that Phase 3 cleanup should address this conflation.

support/netutil/visibility.go
```diff
@@ -14,25 +15,90 @@ func isAroHCP() bool {
...
+
+// UseSharedIngressHCP returns true when this specific HCP should use the
+// management cluster's shared ingress (HAProxy) for public endpoints.
+func UseSharedIngressHCP(hcp *hyperv1.HostedControlPlane) bool {
+	if UseSwiftNetworkingHCP(hcp) && IsPublicHCP(hcp) {
+		return true
+	}
+	// Fallback: ARO management cluster with public Azure topology.
+	// Covers CI and unmigrated clusters without Swift PNI configured.
+	return isAroHCP() && hcp.Spec.Platform.Type == hyperv1.AzurePlatform && IsPublicHCP(hcp)
```
saschagrunert: This fallback causes ALL public Azure HCPs on any ARO management cluster to use shared ingress, even without Swift configured and without the legacy annotation. The old `UseSharedIngress()` was a global env-var check, but it only ran in ARO codepaths that already had the annotation set. This is broader than the prior behavior.

Is this intentional (e.g., to cover CI clusters that lack both the API field and the annotation)? If so, a comment explaining why would help. If not, consider also checking the legacy annotation in the fallback path to match the old behavior.

api/hypershift/v1beta1/azure.go
muraee: thanks, this is actually just copying what's being done in this PR https://github.com/openshift/hypershift/pull/8490
tests are added there.

support/netutil/visibility.go
muraee: Yes, this is intentional — it covers CI clusters that have `MANAGED_SERVICE=ARO-HCP` set but lack both the Swift API field and the legacy annotation. Without this fallback, CI clusters would not use shared ingress at all (worker nodes couldn't reach KAS), as confirmed by the e2e-aks failures we hit before adding it.

The old `UseSharedIngress()` was indeed a global check (`return isAroHCP()`), but it was used directly in per-cluster codepaths (e.g., `haproxy.go:160`, `infra.go:530`, `network_policies.go:80`) without requiring the annotation — only `IsPrivateHCP()` checked the annotation. So the effective old behavior was: all public Azure clusters on ARO management clusters used shared ingress.

I'll add a comment explaining the rationale.

---

api/hypershift/v1beta1/azure.go
saschagrunert: nit: em dash (U+2014). These end up in `oc explain` output and generated docs, and some terminals do not render them well.

```suggestion
	// Azure Swift support; this is not provisioned by HyperShift automatically.
```

api/hypershift/v1beta1/azure.go
```diff
@@ -716,6 +732,24 @@ type AzurePrivateLinkSpec struct {
...
+// Swift assigns a private IP from the customer VNet directly to the hosted cluster's
+// router pods, providing private connectivity without a separate Private Link Service.
+//
+// +kubebuilder:validation:XValidation:rule="self.podNetworkInstance == oldSelf.podNetworkInstance",message="podNetworkInstance is immutable"
+type AzureSwiftSpec struct {
+	// podNetworkInstance is the name of the Azure Swift pod network instance that
+	// provides private connectivity to the hosted cluster's router pods. This value
+	// is assigned by the management service.
+	// This field is immutable once set — changing it would disrupt Swift networking
+	// by reassigning the router pod's private IP.
```
saschagrunert: Two things on this block:

1. Em dash (same as above).
2. The godoc does not mention the MinLength=1 / MaxLength=255 constraints. Per API conventions, all validation markers must be documented in the field comment so users can discover limits via `oc explain` without reading the source.

```suggestion
	// This field is immutable once set. Changing it would disrupt Swift networking
	// by reassigning the router pod's private IP.
	// The value must be between 1 and 255 characters.
```

api/hypershift/v1beta1/azure.go
saschagrunert, line 752: MaxLength=255 with no character set restriction is very permissive. Azure PNI names likely follow constrained naming rules. Consider adding a CEL rule that restricts the allowed characters with a human-readable error message (per the api-review guide: "Use CEL for patterns over `+kubebuilder:validation:Pattern`; explain the validation in sentences, do not dump regex for end users"). If the format really is unconstrained, document that explicitly in the godoc so reviewers do not keep asking.

saschagrunert (commented):
Withdrawing my previous ratcheting concern: PR #8490 (now merged) introduced the PrivateLink ternary change. This PR only adds the Swift ternary (new field, nothing to ratchet) and loosens the type immutability rule. No validation tightening here.

Three remaining items on the API side, all on new code in this PR.

api/hypershift/v1beta1/azure.go
muraee: done

api/hypershift/v1beta1/azure.go
muraee: done

api/hypershift/v1beta1/azure.go
muraee: updated the validation

zgalor:
## Testing Report — Swift Private Type

**Tested by**: zgalor
**Date**: 2026-06-08
**Environment**: ARO-HCP personal dev env (`pers-usw3zgal-mgmt-1`, westus3)

### Setup

- Built HyperShift operator image from PR commit `189a75a9` and deployed to management cluster
- Built CPO image from PR branch (`Dockerfile.dev`) and patched CPO deployments with it
- Updated HyperShift CRDs (`HostedCluster`, `HostedControlPlane`) from PR branch
- Installed `secret-sync-controller` (CRD + controller) — was missing from personal dev env
- Updated CS code to set `Private.Type=Swift` with `swift.podNetworkInstance` field

### Test Clusters

| Cluster | api.listening | Topology | Private.Type | Swift.PodNetworkInstance |
|---------|--------------|----------|-------------|------------------------|
| zg-public-3 | external | PublicAndPrivate | Swift | pni-... |
| zg-private-3 | internal | Private | Swift | pni-... |

### Verified

1. **CRD validation accepts `Swift` type** — HostedCluster CRs with `private.type: Swift` and `swift.podNetworkInstance` pass CRD validation and are applied successfully.

2. **Topology mapping is correct** — `api.listening=external` maps to `PublicAndPrivate`, `api.listening=internal` maps to `Private`.

3. **Union discriminator enforced** — CRD correctly requires `swift` sub-field when `type=Swift` and rejects it without. Error: `swift is required when type is Swift, and forbidden otherwise`.

4. **Konnectivity route auto-population works with PR's CPO** — When using the CPO built from this PR, the konnectivity and ignition route hostnames are correctly auto-populated (e.g. `konnectivity-server.apps.<cluster>.hypershift.local`). Note: initial testing with the stock CPO image (without PR changes) showed this not working — the PR's CPO is required.

5. **Public cluster (`PublicAndPrivate`) fully working** — InfrastructureReady=True, etcd up, KAS (3 replicas) running, control plane available.

6. **KAS visibility on public cluster** — KAS reachable via public endpoint: `curl https://api-int.zg-public-3.k4w4.usw3zgal.hcp.osadev.cloud/healthz` returns `HTTP 200 ok`.

### Issue: `Private` Topology Cluster Blocked on private-router

The `Private` topology cluster is stuck on `InfrastructureReady: False` with message:
> `private-router load balancer is not provisioned; 2m since creation.`

**Details:**
- The CPO creates a `private-router` service as `ClusterIP` for the private cluster
- The infrastructure status check then expects this to be a `LoadBalancer` with an ingress IP, and fails because there's no `status.loadBalancer.ingress`
- This blocks the entire control plane bootstrapping for `Private` topology clusters

**Why it only affects `Private` topology:**
- `PublicAndPrivate` topology → `IsPublicHCP=true` → `UseSharedIngressHCP=true` → skips internal router check entirely
- `Private` topology → `IsPublicHCP=false` → `UseSharedIngressHCP=false` → goes through the private router path, which creates the service and then fails the LB status check

**Analysis:**
- `reconcileInternalRouterServiceStatus` has a guard: `if !netutil.IsPrivateHCP(hcp) || netutil.UseSwiftNetworkingHCP(hcp) || ...` that should skip for Swift clusters
- But `reconcileHCPRouterServices` appears to still create the `private-router` service for Swift private clusters
- There seems to be a mismatch: the service is created but the status check that's supposed to be skipped for Swift is still running, or a different code path is checking it

**Not yet verified (blocked by above):**
- KAS visibility on `Private` topology cluster — unable to test because control plane doesn't come up

### Additional Setup Notes

- The HyperShift CRDs need to be applied from the PR branch — deploying only the operator image is not sufficient since the old CRDs don't include the `Swift` enum.
- The CPO image must also be built from the PR branch. Use `Dockerfile.dev` and override via `hypershift.openshift.io/control-plane-operator-image` annotation on the HostedCluster, or patch the CPO deployment directly.
- The `SecretSync` CRD (`secret-sync.x-k8s.io/v1alpha1`) must be installed for ManifestWork resources to apply successfully.
- After installing new CRDs, the klusterlet work agent may need a pod restart to re-discover the new API resources.

control-plane-operator/controllers/hostedcontrolplane/infra/infra_test.go
```diff
@@ -535,20 +535,28 @@ func TestReconcileInfrastructure(t *testing.T) {
...
+				hcp.Spec.Platform.Azure.AzureAuthenticationConfig = hyperv1.AzureAuthenticationConfiguration{
+					AzureAuthenticationConfigType: hyperv1.AzureAuthenticationTypeManagedIdentities,
+				}
+				return hcp
+			}(),
 			setupEnv: func(t *testing.T) {
 				t.Setenv("MANAGED_SERVICE", hyperv1.AroHCP)
 			},
 			expectError: false,
-			// For ARO with shared ingress:
```
enxebre, line 548: why are these removed? aren't they true anymore?

control-plane-operator/controllers/hostedcontrolplane/infra/testdata/zz_fixture_TestReconcileInfrastructure_ARO_Route_Swift_Private.yaml
```diff
@@ -0,0 +1,280 @@
...
+      namespace: test-namespace
+    spec:
+      ports:
+      - name: https
+        port: 443
+        protocol: TCP
+        targetPort: https
+      selector:
+        app: private-router
+      type: ClusterIP
```
enxebre, line 281: shouldn't this be a LB?

enxebre: nah, ignore that. No LB in swift path

enxebre:
Other than the regression lgtm overall
TestReconcileInternalRouterServiceStatus needs adjustment to catch the regression
TestReconcileInfrastructure needs to account for only private scenarios

control-plane-operator/controllers/hostedcontrolplane/infra/infra_test.go
muraee: restored

zgalor:
## Re-test After Fix — `e2aa9ffd8`

Retested with the latest commit (`e2aa9ffd8 fix(cpo): handle Swift private topology in infra reconciliation`) on fresh clusters using both the HyperShift operator and CPO built from the PR branch.

### Results

| Cluster | Topology | InfraReady | Etcd | KAS | KAS Public Reachability |
|---------|----------|-----------|------|-----|------------------------|
| zg-public-4 | PublicAndPrivate | True | True | True | `HTTP 200 ok` |
| zg-private-4 | Private | True | True | True | Connection refused (expected) |

### Verified

1. **Private-router issue fixed** — `Private` topology cluster now reaches `InfrastructureReady: True` without manual intervention.
2. **Konnectivity routes auto-populated** — Both clusters get `konnectivity-server.apps.<cluster>.hypershift.local` set by the CPO automatically.
3. **KAS visibility correct** — Public cluster KAS reachable via shared ingress (`HTTP 200`). Private cluster KAS not reachable from public network (connection refused).
4. **Route structure correct for private cluster** — No `kube-apiserver` public route created; only `kube-apiserver-private` and `kube-apiserver-internal` (hypershift.local) routes exist.

LGTM from the CS testing side.

PR #8540, opened 2026-05-19
AUTOSCALE-685: Stop controllers fighting over HCP Status() updates 

jkyros:
- We have multiple controllers writing to the HCP's status object with `Update()` calls, which replaces the entire object 
- Depending on what the controller performing the Update() understands and sees, and when it performs the update, that potentially blows away fields like `vcpus` from the autonode status that we added here: https://github.com/openshift/hypershift/pull/8265/ 
- This just replaces the Status().Update() calls with calls to Patch calls that use client.MergeFrom() so that the full object updates don't wipe other controllers' status update
- Testing this explicitly is unpleasant without some kind of refactor, the parts controllers that write the status aren't set up for easy testing, but it's observably possible the issue can happen because of the use of `Update()` calls across multiple controllers on the same object 

Fixes 

## Checklist:

## Summary by CodeRabbit

* **Refactor**
  * Status-condition updates across control-plane, ignition server, and backup controllers now use server-side apply for more reliable, conflict-resistant status synchronization.

* **New Features**
  * Added reusable SSA condition helpers to build, merge, and preserve condition fields consistently.
  * Controllers now use a typed Hypershift client to apply status updates.

* **Chores / Tests**
  * Tests updated to validate status condition changes via the typed Hypershift API surface.
* **Chores**
  * Startup wiring updated to construct and inject the Hypershift client; codegen post-processing adjusted for apply-configuration generation.

### DIFF
--- a/client/applyconfiguration/hypershift/v1beta1/clusterversionstatus.go
+++ b/client/applyconfiguration/hypershift/v1beta1/clusterversionstatus.go
@@ -27,7 +27,7 @@ type ClusterVersionStatusApplyConfiguration struct {
 	Desired            *v1.Release            `json:"desired,omitempty"`
 	History            []v1.UpdateHistory     `json:"history,omitempty"`
 	ObservedGeneration *int64                 `json:"observedGeneration,omitempty"`
-	AvailableUpdates   []v1.Release           `json:"availableUpdates,omitempty"`
+	AvailableUpdates   []v1.Release           `json:"availableUpdates,omitzero"`
 	ConditionalUpdates []v1.ConditionalUpdate `json:"conditionalUpdates,omitempty"`
 }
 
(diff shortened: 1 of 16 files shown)


jkyros:
Codecov is probably rightfully mad, but yuck, I'm going to have to rearrange it to actually cover this :disappointed: 
/test e2e-aws
/test e2e-aws-autonode
/test e2e-aws-upgrade-hypershift-operator

jkyros:
Yeah, that's not good enough, that looks like it totally ruins conditions by...replacing the whole array? Because...of course it would.  Going to have to server-side apply or something.

enxebre:
maybe also client.MergeFromWithOptions(originalHostedControlPlane, client.MergeFromWithOptimisticLock{}) helps

jkyros:
That didn't look like us, that looked like e2e-aws is in rough shape 
/test e2e-aws-autonode

jkyros:
SSA is too complicated with the applyConfiguration serialization. I wanted to see if this worked, but I'm going to go back to Patch/MergeFrom and oplocks, which will cause a bunch of retries but should otherwise be safe. This feels gross. 
/test e2e-aws-autonode

jkyros:
Closing in favor of https://github.com/openshift/hypershift/pull/8562

PR #8543, opened 2026-05-19
OCPBUGS-85778: Fix NodePool reconciliation failure when updating mirrored immutable ConfigMaps

vsolanki12:

When multiple NodePools reference the same kubelet ConfigMap and the source ConfigMap is modified after mirroring, the nodepool-controller's NTO reconciliation fails because mirrored ConfigMaps are created with `immutable: true`. Kubernetes rejects in-place updates to immutable ConfigMaps, causing NodePools to get stuck in `UpdatingConfig=True` with:

```
failed to reconcile mirrored ConfigMap: data: Forbidden: field is immutable when `immutable` is set
```

This PR:
- Removes the `immutable: true` flag from `mutateMirroredConfig` so newly created mirrored ConfigMaps are mutable
- Adds `deleteImmutableConfigMapIfNeeded` to `reconcileMirroredConfigs` to handle migration — existing immutable mirrored ConfigMaps are deleted before `CreateOrUpdate` recreates them as mutable
- This mirrors the parallel fix already applied on the CPO/HCCO side (commit bbd8f63d9, PR #6726) which fixed Stage 2 (HCP namespace → guest cluster), while this PR fixes Stage 1 (user namespace → HCP namespace)

Fixes https://issues.redhat.com/browse/OCPBUGS-85778

- Tested on a live KubeVirt HCP cluster (OCP 4.21.9 management, 4.18.30 guest) with two NodePools sharing the same `custom-kubelet` ConfigMap
- Bug reproduced: both NodePools stuck in `UpdatingConfig=True` with immutable field error
- After deploying patched operator image: mirrored ConfigMaps recreated as mutable, reconciliation succeeded, no new errors

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented reconciliation failures by removing immutable ConfigMaps and recreating them as mutable so updates apply reliably.
  * Ensured mirrored ConfigMaps are set mutable during reconciliation to allow safe updates.

* **Tests**
  * Updated tests to expect mutable mirrored ConfigMaps and to validate recreation behavior when immutable configs are encountered.

### DIFF
--- a/hypershift-operator/controllers/nodepool/nto.go
+++ b/hypershift-operator/controllers/nodepool/nto.go
@@ -116,6 +116,10 @@ func (r *NodePoolReconciler) reconcileMirroredConfigs(ctx context.Context, logr 
 				Name:      netutil.ShortenName(mirroredConfig.Name, nodePool.Name, validation.LabelValueMaxLength),
 				Namespace: controlPlaneNamespace},
 		}
+		if err := r.deleteImmutableConfigMapIfNeeded(ctx, logr, cm, nodePool.Name); err != nil {
+			return err
+		}
+		cm.SetResourceVersion("")
 		if result, err := r.CreateOrUpdate(ctx, r.Client, cm, func() error {
 			return mutateMirroredConfig(cm, mirroredConfig, nodePool)
 		}); err != nil {
@@ -193,7 +197,7 @@ func reconcilePerformanceProfileConfigMap(performanceProfileConfigMap *corev1.Co
 }
 
 func mutateMirroredConfig(cm *corev1.ConfigMap, mirroredConfig *MirrorConfig, nodePool *hyperv1.NodePool) error {
-	cm.Immutable = ptr.To(true)
+	cm.Immutable = ptr.To(false)
 	if cm.Annotations == nil {
 		cm.Annotations = make(map[string]string)
 	}
@@ -205,6 +209,21 @@ func mutateMirroredConfig(cm *corev1.ConfigMap, mirroredConfig *MirrorConfig, no
 	cm.Labels = labels.Merge(cm.Labels, mirroredConfig.Labels)
 	cm.Data = mirroredConfig.Data
 	return nil
+}
+
+func (r *NodePoolReconciler) deleteImmutableConfigMapIfNeeded(ctx context.Context, log logr.Logger, cm *corev1.ConfigMap, nodePoolName string) error {
+	_, err := k8sutil.DeleteIfNeededWithPredicate(ctx, r.Client, cm, func(existing *corev1.ConfigMap) bool {
+		if existing.Labels[NTOMirroredConfigLabel] != "true" || existing.Labels[hyperv1.NodePoolLabel] != nodePoolName {
+			return false
+		}
+		if existing.Immutable != nil && *existing.Immutable {
+			log.Info("deleting immutable mirrored ConfigMap to recreate as mutable",
+				"configMap", client.ObjectKeyFromObject(existing).String())
+			return true
+		}
+		return false
+	})
+	return err
 }
 
 func (r *NodePoolReconciler) getTuningConfig(ctx context.Context,
--- a/hypershift-operator/controllers/nodepool/nto_test.go
+++ b/hypershift-operator/controllers/nodepool/nto_test.go
@@ -553,7 +553,7 @@ func TestReconcileMirroredConfigs(t *testing.T) {
 			existingConfigsInHcpNs: nil,
 			expectedMirroredConfigs: []corev1.ConfigMap{
 				{
-					Immutable: ptr.To(true),
+					Immutable: ptr.To(false),
 					ObjectMeta: metav1.ObjectMeta{
 						Name:      netutil.ShortenName("foo", npName, validation.LabelValueMaxLength),
 						Namespace: hcpNamespace,
@@ -602,7 +602,7 @@ func TestReconcileMirroredConfigs(t *testing.T) {
 			},
 			expectedMirroredConfigs: []corev1.ConfigMap{
 				{
-					Immutable: ptr.To(true),
+					Immutable: ptr.To(false),
 					ObjectMeta: metav1.ObjectMeta{
 						Name:      netutil.ShortenName("foo", npName, validation.LabelValueMaxLength),
 						Namespace: hcpNamespace,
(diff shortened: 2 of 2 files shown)


hypershift-operator/controllers/nodepool/nto.go
```diff
@@ -207,6 +210,23 @@ func mutateMirroredConfig(cm *corev1.ConfigMap, mirroredConfig *MirrorConfig, no
...
+		if apierrors.IsNotFound(err) {
+			return nil
+		}
+		return fmt.Errorf("failed to get ConfigMap %s: %w", client.ObjectKeyFromObject(cm).String(), err)
+	}
+	if existingCM.Immutable != nil && *existingCM.Immutable {
+		log.Info("deleting immutable mirrored ConfigMap to recreate as mutable", "configMap", client.ObjectKeyFromObject(existingCM).String())
+		if _, err := k8sutil.DeleteIfNeeded(ctx, r.Client, existingCM); err != nil {
+			return fmt.Errorf("failed to delete immutable ConfigMap %s: %w", client.ObjectKeyFromObject(existingCM).String(), err)
+		}
```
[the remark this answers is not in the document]
vsolanki12: Applied the suggested label guard. `deleteImmutableConfigMapIfNeeded` now checks `NTOMirroredConfigLabel` and `nodePoolAnnotation` labels before deleting.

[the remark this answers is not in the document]
vsolanki12: @coderabbitai review

hypershift-operator/controllers/nodepool/nto.go
```diff
@@ -207,6 +210,26 @@ func mutateMirroredConfig(cm *corev1.ConfigMap, mirroredConfig *MirrorConfig, no
...
+		if apierrors.IsNotFound(err) {
+			return nil
+		}
+		return fmt.Errorf("failed to get ConfigMap %s: %w", client.ObjectKeyFromObject(cm).String(), err)
+	}
+	if existingCM.Labels[NTOMirroredConfigLabel] != "true" || existingCM.Labels[nodePoolAnnotation] != nodePoolName {
+		return nil
+	}
+	if existingCM.Immutable != nil && *existingCM.Immutable {
+		log.Info("deleting immutable mirrored ConfigMap to recreate as mutable", "configMap", client.ObjectKeyFromObject(existingCM).String())
```
jparrill: There's a small race window here that concerns me. Between the delete and the `CreateOrUpdate` right after, the mirrored CM doesn't exist in the HCP namespace. HCCO watches these CMs via `KubeletConfigConfigMapLabel` (`resources.go:308`), so if it reconciles during that gap it'll list `want` from the HCP namespace (CM missing), diff against `have` in the guest (`openshift-config-managed`), and delete the guest-side copy (`resources.go:2999-3001`). If NTO in the guest sees that KubeletConfig disappear, it could regenerate MachineConfigs without it and MCO would roll nodes.

The window is narrow and it's a one-time migration, so it's unlikely — but a spurious node rollout in production is not great. One option: add a guard in HCCO's `reconcileKubeletConfig` so it doesn't delete a guest CM that has `NTOMirroredConfigLabel` and a still-existing NodePool, even if the source is transiently absent.

hypershift-operator/controllers/nodepool/nto.go
```diff
@@ -116,6 +116,9 @@ func (r *NodePoolReconciler) reconcileMirroredConfigs(ctx context.Context, logr
 				Name:      netutil.ShortenName(mirroredConfig.Name, nodePool.Name, validation.LabelValueMaxLength),
 				Namespace: controlPlaneNamespace},
 		}
+		if err := r.deleteImmutableConfigMapIfNeeded(ctx, logr, cm, nodePool.Name); err != nil {
```
jparrill, line 119: What happens if this delete succeeds but `CreateOrUpdate` fails right after (transient API error, network blip)? The CM is gone from the HCP namespace until the next reconcile. That widens the race window from the comment above — instead of milliseconds, it could be seconds depending on the backoff. Might be worth noting that this is acceptable given controller-runtime requeues quickly, but it's something to be aware of.

jparrill (commented):

A couple of things I noticed while tracing the code that are **not from this PR** but are related:

**HCCO `deleteImmutableConfigMapIfNeeded` lacks ownership check** (`resources.go:3010`): The HCCO version only checks `Immutable == true` before deleting — it doesn't verify any ownership labels. Your HO version is actually safer because it checks `NTOMirroredConfigLabel` and `nodePoolAnnotation`. Might be worth backporting that ownership check to the HCCO side too.

**ValidatingAdmissionPolicy uses wrong API group** (`admissionpolicies.go:176`): `configv1.SchemeGroupVersion.Group` is `"config.openshift.io"`, but ConfigMaps are core API (`""`). This means the VAP that's supposed to protect mirrored CMs from unauthorized deletes in the guest never actually matches. Worth a separate fix.

hypershift-operator/controllers/nodepool/nto.go
vsolanki12: Agreed the window is narrow and one-time. The delete and CreateOrUpdate happen in the same reconcile iteration with no blocking calls between them, so the gap is sub-millisecond in practice. Filed [OCPBUGS-86949](https://issues.redhat.com/browse/OCPBUGS-86949) to track the HCCO-side hardening — adding a guard in `reconcileKubeletConfig` against transient source absence, and backporting the ownership label check to the HCCO's `deleteImmutableConfigMapIfNeeded`.

hypershift-operator/controllers/nodepool/nto.go
vsolanki12: Acknowledged. If CreateOrUpdate fails after delete, controller-runtime requeues within seconds and the CM gets recreated on the next pass. This is the standard controller reconciliation pattern. The HCCO-side guard tracked in [OCPBUGS-86949](https://issues.redhat.com/browse/OCPBUGS-86949) would also mitigate this scenario.

hypershift-operator/controllers/nodepool/nto.go
```diff
@@ -207,6 +210,26 @@ func mutateMirroredConfig(cm *corev1.ConfigMap, mirroredConfig *MirrorConfig, no
...
+		return nil
+	}
+	if existingCM.Immutable != nil && *existingCM.Immutable {
+		log.Info("deleting immutable mirrored ConfigMap to recreate as mutable", "configMap", client.ObjectKeyFromObject(existingCM).String())
+		if _, err := k8sutil.DeleteIfNeeded(ctx, r.Client, existingCM); err != nil {
+			return fmt.Errorf("failed to delete immutable ConfigMap %s: %w", client.ObjectKeyFromObject(existingCM).String(), err)
+		}
+	}
+	return nil
+}
```
jparrill, line 227: There's already a `k8sutil.DeleteIfNeededWithPredicate` that does exactly this — Get, check a predicate, delete if it matches. It also handles the deletion-timestamp edge case that this version skips, and it avoids the double-Get (your function does a Get, then `DeleteIfNeeded` inside does another one on the same object). The whole function could be:
```go
func (r *NodePoolReconciler) deleteImmutableConfigMapIfNeeded(ctx context.Context, log logr.Logger, cm *corev1.ConfigMap, nodePoolName string) error {
    _, err := k8sutil.DeleteIfNeededWithPredicate(ctx, r.Client, cm, func(existing *corev1.ConfigMap) bool {
        if existing.Labels[NTOMirroredConfigLabel] != "true" || existing.Labels[hyperv1.NodePoolLabel] != nodePoolName {
            return false
        }
        if existing.Immutable != nil && *existing.Immutable {
            log.Info("deleting immutable mirrored ConfigMap to recreate as mutable",
                "configMap", client.ObjectKeyFromObject(existing).String())
            return true
        }
        return false
    })
    return err
}
```

hypershift-operator/controllers/nodepool/nto.go
```diff
@@ -207,6 +210,26 @@ func mutateMirroredConfig(cm *corev1.ConfigMap, mirroredConfig *MirrorConfig, no
...
 
+func (r *NodePoolReconciler) deleteImmutableConfigMapIfNeeded(ctx context.Context, log logr.Logger, cm *corev1.ConfigMap, nodePoolName string) error {
+	existingCM := &corev1.ConfigMap{}
+	if err := r.Client.Get(ctx, client.ObjectKeyFromObject(cm), existingCM); err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil
+		}
+		return fmt.Errorf("failed to get ConfigMap %s: %w", client.ObjectKeyFromObject(cm).String(), err)
+	}
+	if existingCM.Labels[NTOMirroredConfigLabel] != "true" || existingCM.Labels[nodePoolAnnotation] != nodePoolName {
```
jparrill: nit: The rest of `nto.go` uses `hyperv1.NodePoolLabel` for label access (lines 59, 100, 203, 343). Using `nodePoolAnnotation` here works (same string), but it reads as if you're checking an annotation when you're actually checking a label. Suggest `hyperv1.NodePoolLabel` for consistency.

hypershift-operator/controllers/nodepool/nto_test.go
```diff
@@ -652,6 +652,44 @@ func TestReconcileMirroredConfigs(t *testing.T) {
...
+						},
+					},
+					Data: map[string]string{
+						TokenSecretConfigKey: kubeletConfig1,
+					},
+				},
+			},
+		},
+		{
+			name:                  "When an immutable mirrored ConfigMap exists and data changes, it should delete and recreate as mutable",
```
jparrill, line 672: The label check at `nto.go:221` (`existingCM.Labels[nodePoolAnnotation] != nodePoolName`) is the safety mechanism that prevents one NodePool's reconciler from deleting another NodePool's ConfigMap — which is the exact multi-NodePool scenario that caused this bug. But there's no test for it. Could you add a case with an immutable CM labeled with a different NodePool name (e.g., `"other-nodepool"`) and verify it is NOT deleted?

hypershift-operator/controllers/nodepool/nto_test.go
```diff
@@ -652,6 +652,44 @@ func TestReconcileMirroredConfigs(t *testing.T) {
...
+						Data: map[string]string{
+							TokenSecretConfigKey: kubeletConfig1,
+						},
+					},
+					Labels: map[string]string{
+						KubeletConfigConfigMapLabel: "true",
+					},
+				},
+			},
+			existingConfigsInHcpNs: []client.Object{
```
jparrill, line 691: nit: A test case where the existing CM has `Immutable: ptr.To(false)` (or nil) and verifying nothing happens would prevent future regressions — e.g., someone removing the nil/false check and causing unnecessary deletes.

jparrill (commented):

hypershift-operator/controllers/nodepool/nto_test.go
vsolanki12: Added, When an existing mirrored `ConfigMap` is already mutable, it should not be deleted". Verifies the CM is left untouched when Immutable is nil.

hypershift-operator/controllers/nodepool/nto_test.go
vsolanki12: Added,  "When an immutable mirrored ConfigMap belongs to a different NodePool, it should not be deleted". It creates an immutable CM labeled with "other-nodepool" and asserts it's preserved after reconciliation.

hypershift-operator/controllers/nodepool/nto.go
vsolanki12: updated to `hyperv1.NodePoolLabel` for consistency with the rest of nto.go

hypershift-operator/controllers/nodepool/nto.go
vsolanki12: Thank you @jparrill Done,  refactored to use `DeleteIfNeededWithPredicate` exactly as suggested. Eliminates the double Get and picks up the deletion timestamp guard

vsolanki12:
/verified by @vsolanki12 

**Before fix:**
```
$ oc describe nodepool ocpbugs-85778 -n clusters | tail -10
Events:
  Type     Reason          Age   From                 Message
  ----     ------          ----  ----                 -------
  Warning  ReconcileError  15m   nodepool-controller  failed to reconcile NTO:
    failed to mirror configs: failed to reconcile mirrored
    clusters-ocpbugs-85778/custom-kubelet-ocpbugs-85778 ConfigMap:
    ConfigMap "custom-kubelet-ocpbugs-85778" is invalid:
    data: Forbidden: field is immutable when `immutable` is set
```

**After fix:**
```
$ oc get cm -n clusters-ocpbugs-85778 custom-kubelet-ocpbugs-85778 -o jsonpath='{.immutable}'
(empty — no longer immutable)

$ oc get cm -n clusters-ocpbugs-85778 custom-kubelet-ocpbugs-85778-b -o jsonpath='{.immutable}'
(empty — no longer immutable)

2. No more reconciliation errors:

$ oc describe nodepool ocpbugs-85778-b -n clusters | tail -10
Events:
  Type     Reason          Age                   From                 Message
  ----     ------          ----                  ----                 -------
  Warning  ReconcileError  100m (x2 over 100m)   nodepool-controller  ...

3. Operator logs clean:

$ oc logs -n hypershift deployment/operator --tail=50 | grep -i "immutable\|mirrored"
```

PR #8545, opened 2026-05-19
OCPBUGS-84955: bootstrap serving certs at hypershift operator startup

clebs:

The webhook server requires TLS certs on disk before it can start listening. Previously, certs were only created via `hypershift install` manifests, meaning the operator would fail to start if the serving cert secret was not yet present like it is the case when running `hyperhsift install render` without the `--render-sensitive` flag.

This is used in a scenario where the rendered manifests are pushed to a gitops workflow and, therefore, secrets can not be rendered.

By adding bootstrap cert generation at startup, the operator is self-sufficient: if the secret exists, the volume mount delivers certs normally; if it is missing or empty, certs are generated, persisted, and written to disk. The secret volume is now marked optional so the pod can start without it.

Followup to [OCPBUGS-84955](https://redhat.atlassian.net/browse/OCPBUGS-84955)

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Operator can bootstrap webhook CA and serving certificates at startup when a certificate directory is configured; installer no longer embeds or propagates webhook CA/certs into rendered assets.
  * Webhook configurations no longer include an embedded CA bundle.

* **Bug Fixes**
  * Webhook serving-certificate secret is now optional so the operator can start without pre-provisioned certs.

* **Tests**
  * Added tests covering certificate bootstrapping, creation, regeneration, and existing-secret behaviors.

### DIFF
--- a/cmd/install/install.go
+++ b/cmd/install/install.go
@@ -30,7 +30,6 @@ import (
 	crdassets "github.com/openshift/hypershift/cmd/install/assets/crds"
 	"github.com/openshift/hypershift/cmd/util"
 	"github.com/openshift/hypershift/hypershift-operator/controllers/sharedingress"
-	"github.com/openshift/hypershift/hypershift-operator/controllers/webhookcerts"
 	hyperapi "github.com/openshift/hypershift/support/api"
 	"github.com/openshift/hypershift/support/config"
 	"github.com/openshift/hypershift/support/metrics"
@@ -756,22 +755,10 @@ func hyperShiftOperatorManifests(ctx context.Context, client crclient.Client, op
 	operatorServiceAccount, rbacObjs := setupRBAC(opts, operatorNamespace)
 	objects = append(objects, rbacObjs...)
 
-	// Generate self-managed webhook CA and serving cert when any webhook is enabled.
-	var webhookCABundle []byte
-	if opts.EnableDefaultingWebhook || opts.EnableConversionWebhook || opts.EnableValidatingWebhook || opts.EnableAuditLogPersistence {
-		caSecret, servingSecret, caBundle, err := webhookcerts.GenerateInitialWebhookCerts(operatorNamespace.Name, assets.HypershiftOperatorName)
-		if err != nil {
-			return nil, nil, fmt.Errorf("failed to generate webhook certs: %w", err)
-		}
-		objects = append(objects, caSecret, servingSecret)
-		webhookCABundle = caBundle
-	}
-
 	if opts.EnableDefaultingWebhook || opts.EnableAuditLogPersistence {
 		mutatingWebhookConfiguration := assets.HyperShiftMutatingWebhookConfiguration{
 			Namespace:                 operatorNamespace,
 			EnableAuditLogPersistence: opts.EnableAuditLogPersistence,
-			CABundle:                  webhookCABundle,
 		}.Build()
 		objects = append(objects, mutatingWebhookConfiguration)
 	}
@@ -779,7 +766,6 @@ func hyperShiftOperatorManifests(ctx context.Context, client crclient.Client, op
 	if opts.EnableValidatingWebhook {
 		validatingWebhookConfiguration := assets.HyperShiftValidatingWebhookConfiguration{
 			Namespace: operatorNamespace.Name,
-			CABundle:  webhookCABundle,
 		}.Build()
 		objects = append(objects, validatingWebhookConfiguration)
 	}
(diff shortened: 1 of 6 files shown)


hypershift-operator/controllers/webhookcerts/webhookcerts_controller.go
```diff
@@ -364,3 +366,62 @@ func GenerateInitialWebhookCerts(namespace, serviceName string) (*corev1.Secret,
...
+	}
+	if createErr := c.Create(ctx, servingSecret); createErr != nil {
+		if !apierrors.IsAlreadyExists(createErr) {
+			return fmt.Errorf("failed to create serving cert secret: %w", createErr)
+		}
+		existingSecret.Data = servingSecret.Data
+		if err := c.Update(ctx, existingSecret); err != nil {
+			return fmt.Errorf("failed to update serving cert secret: %w", err)
+		}
+	}
```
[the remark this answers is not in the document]
clebs: This logic runs exactly once on operator startup, there should be no other processes creating the secret at this time.

hypershift-operator/controllers/webhookcerts/webhookcerts_controller.go
```diff
@@ -364,3 +366,62 @@ func GenerateInitialWebhookCerts(namespace, serviceName string) (*corev1.Secret,
...
+
+	log.Info("Webhook certificates bootstrapped, writing to disk")
+	return writeCertFiles(certDir, servingSecret.Data[corev1.TLSCertKey], servingSecret.Data[corev1.TLSPrivateKeyKey])
+}
+
+func writeCertFiles(certDir string, cert, key []byte) error {
```
muraee: why do we need to write to disk as well?

muraee:
please also remove the code from the CLI generating those secrets.
https://github.com/openshift/hypershift/blob/ca519ed78d72076b75acb982771a68035410c7ce/cmd/install/install.go#L759-L768

hypershift-operator/controllers/webhookcerts/webhookcerts_controller.go
clebs: Because when secrets don't exist yet, volume mount has nothing to deliver. Pod started with Optional: true on the volume, so k8s mounts an empty dir. `EnsureWebhookCerts` creates the secrets via API, but
Kubernetes doesn't hot-reload secret volumes into an already-running container mid-startup. The webhook server needs certs on disk now to bind the TLS listener.

On next pod restart, secret exists and it works as it normally does therefore EnsureWebhookCerts returns early (no-op path).

clebs:
@muraee I left the `hypershift install` path there to not disrupt the other use cases, but will remove it and run some tests.

hypershift-operator/controllers/webhookcerts/webhookcerts_controller.go
muraee: controller-runtime `certWatcher` used by the webhook-server should take care of reloading the certs when the secret exist or changed (it uses fsnotify).
Have you tested without this?

clebs: I have not, I will test it out.

clebs: @muraee I have removed the writing to disk form the bootstrapping as well as the bootstrapping in the CLI.
All installation paths work, PTAL!

cmd/install/install.go
```diff
@@ -831,7 +817,7 @@ func hyperShiftOperatorManifests(ctx context.Context, client crclient.Client, op
 		objects = append(objects, sharedIngressObjs...)
 	}
 
-	crds, err = setupCRDs(ctx, client, opts, operatorNamespace, operatorService, webhookCABundle)
+	crds, err = setupCRDs(ctx, client, opts, operatorNamespace, operatorService)
 	if err != nil {
 		return nil, nil, err
 	}
```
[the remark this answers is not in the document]
clebs: This is not an issue: the secret bootstrapping is now part of the hypershift operator startup sequence which happens after the install command (both render and regular) finish.
That will ensure the bundles are in sync, either with new or existing secrets, when the webhookcerts controller starts.

hypershift-operator/controllers/webhookcerts/webhookcerts_controller.go
```diff
@@ -364,3 +364,73 @@ func GenerateInitialWebhookCerts(namespace, serviceName string) (*corev1.Secret,
...
+	if err != nil {
+		return fmt.Errorf("failed to generate webhook certs: %w", err)
+	}
+
+	if createErr := c.Create(ctx, caSecret); createErr != nil {
+		if !apierrors.IsAlreadyExists(createErr) {
+			return fmt.Errorf("failed to create CA secret: %w", createErr)
+		}
+		// CA already exists — fetch it and regenerate the serving cert so it
+		// is signed by the persisted CA, not the transient one we just created.
```
muraee: I think this is bit too much, if either of the 2 secrets is missing/empty, we should just replace both of them, unless there is a strong reason not to.

Also, you can use controller-runtime `controllerutil.CreateOrPatch` instead of checking errors on create then update.

clebs: Ok, was not sure if replacing an existing CA would have side effects. But since we then patch the caBundles it should be fine.

Added the changes.

PR #8552, opened 2026-05-20
OCPBUGS-86238: set limits for aro.openshift.io/swift-nic in request overrides for ARO swift

enxebre:
## Summary

- When using `resource-request-override.hypershift.openshift.io/` to override extended resources like `aro.openshift.io/swift-nic`, the pod fails API server admission because Kubernetes requires `limits == requests` for extended resources
- Adds `applyNonOvercommitableResourceLimits()` to automatically set limits equal to requests for `aro.openshift.io/swift-nic` during request overrides
- Applies to both init containers and regular containers

## Jira

https://issues.redhat.com/browse/OCPBUGS-86238

## Test plan

- [x] Unit tests for `applyNonOvercommitableResourceLimits` (3 cases: swift-nic only, standard resources, mixed)
- [x] Unit tests for `applyRequestsOverrides` integration (5 cases: cpu/memory, swift-nic, mixed, init containers, different deployment)
- [x] `make verify` passes (codespell, lint, build, tests)

## Summary by CodeRabbit

* **Bug Fixes**
  * Resource override handling now also sets matching limits for a specific extended resource so overrides apply consistently to init and regular containers.

* **Tests**
  * New unit tests validate override behavior for standard resources, the extended resource, mixed cases, and annotation targeting.

* **Chores**
  * CI pipeline tasks updated to enable the package registry proxy during dependency prefetch.

### DIFF
--- a/support/controlplane-component/defaults.go
+++ b/support/controlplane-component/defaults.go
@@ -427,12 +427,29 @@ func (c *controlPlaneWorkload[T]) applyRequestsOverrides(podTemplate *corev1.Pod
 	for i, c := range podTemplate.Spec.InitContainers {
 		if res, ok := requestsOverrides[c.Name]; ok {
 			maps.Copy(podTemplate.Spec.InitContainers[i].Resources.Requests, res)
+			applyNonOvercommitableResourceLimits(&podTemplate.Spec.InitContainers[i], res)
 		}
 	}
 	for i, c := range podTemplate.Spec.Containers {
 		if res, ok := requestsOverrides[c.Name]; ok {
 			maps.Copy(podTemplate.Spec.Containers[i].Resources.Requests, res)
+			applyNonOvercommitableResourceLimits(&podTemplate.Spec.Containers[i], res)
 		}
+	}
+}
+
+const aroSwiftNICResource corev1.ResourceName = "aro.openshift.io/swift-nic"
+
+// applyNonOvercommitableResourceLimits sets limits equal to requests for extended
+// resources that cannot be overcommitted, specifically "aro.openshift.io/swift-nic".
+// The API server requires limits == requests for these resources.
+// https://github.com/kubernetes/kubernetes/blob/621e250502ddeeab8274836e88b506c0c4f57232/pkg/apis/core/validation/validation.go#L7975-L7976
+func applyNonOvercommitableResourceLimits(container *corev1.Container, overrides corev1.ResourceList) {
+	if quantity, ok := overrides[aroSwiftNICResource]; ok {
+		if container.Resources.Limits == nil {
+			container.Resources.Limits = corev1.ResourceList{}
+		}
+		container.Resources.Limits[aroSwiftNICResource] = quantity
 	}
 }
 
--- a/support/controlplane-component/defaults_test.go
+++ b/support/controlplane-component/defaults_test.go
@@ -10,6 +10,7 @@ import (
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	runtime "k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/utils/ptr"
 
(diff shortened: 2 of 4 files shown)


support/controlplane-component/defaults.go
```diff
@@ -427,11 +427,30 @@ func (c *controlPlaneWorkload[T]) applyRequestsOverrides(podTemplate *corev1.Pod
...
+			applyNonOvercommitableResourceLimits(&podTemplate.Spec.InitContainers[i], res)
 		}
 	}
 	for i, c := range podTemplate.Spec.Containers {
 		if res, ok := requestsOverrides[c.Name]; ok {
 			maps.Copy(podTemplate.Spec.Containers[i].Resources.Requests, res)
+			applyNonOvercommitableResourceLimits(&podTemplate.Spec.Containers[i], res)
+		}
+	}
+}
```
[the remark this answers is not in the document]
enxebre: I prefer to keep this PR narrowed and follow up for that in a different PR

support/controlplane-component/defaults_test.go
```diff
@@ -285,6 +286,240 @@ func generateResources() (map[string]*corev1.Secret, map[string]*corev1.ConfigMa
...
+				g.Expect(podTemplate.Spec.Containers).To(Equal(test.expectedContainers))
+			}
+			if test.expectedInitContainers != nil {
+				g.Expect(podTemplate.Spec.InitContainers).To(Equal(test.expectedInitContainers))
+			}
+		})
+	}
+}
+
+func TestApplyNonOvercommitableResourceLimits(t *testing.T) {
```
muraee, line 469: isn't this already covered in `TestApplyRequestsOverrides`

support/controlplane-component/defaults.go
```diff
@@ -427,11 +427,30 @@ func (c *controlPlaneWorkload[T]) applyRequestsOverrides(podTemplate *corev1.Pod
...
+
+const aroSwiftNICResource corev1.ResourceName = "aro.openshift.io/swift-nic"
+
+// applyNonOvercommitableResourceLimits sets limits equal to requests for extended
+// resources that cannot be overcommitted, specifically "aro.openshift.io/swift-nic".
+// The API server requires limits == requests for these resources.
+// https://github.com/kubernetes/kubernetes/blob/621e250502ddeeab8274836e88b506c0c4f57232/pkg/apis/core/validation/validation.go#L7975-L7976
+func applyNonOvercommitableResourceLimits(container *corev1.Container, overrides corev1.ResourceList) {
+	for resourceName, quantity := range overrides {
```
jparrill: Nit: if we keep the single-resource approach, a direct map lookup is cleaner than iterating the whole map:

```go
if quantity, ok := overrides[aroSwiftNICResource]; ok {
	if container.Resources.Limits == nil {
		container.Resources.Limits = corev1.ResourceList{}
	}
	container.Resources.Limits[aroSwiftNICResource] = quantity
}
```

This becomes moot if we go with a generic `isExtendedResource` approach (where the loop is the right pattern).

support/controlplane-component/defaults_test.go
```diff
@@ -285,6 +286,240 @@ func generateResources() (map[string]*corev1.Secret, map[string]*corev1.ConfigMa
...
+		name                   string
+		annotations            map[string]string
+		containers             []corev1.Container
+		initContainers         []corev1.Container
+		expectedContainers     []corev1.Container
+		expectedInitContainers []corev1.Container
+	}{
+		{
+			name: "When overriding cpu and memory it should only update requests",
+			annotations: map[string]string{
```
jparrill, line 300: Minor: these annotation keys are hardcoded as string literals, but the production code uses `hyperv1.ResourceRequestOverrideAnnotationPrefix`. Using the constant here would prevent drift if the prefix ever changes:

```go
hyperv1.ResourceRequestOverrideAnnotationPrefix + "/router.router": "cpu=500m,memory=1Gi",
```

jparrill (commented):
Dropped some minor nits. Otherwise lgtm

support/controlplane-component/defaults_test.go
enxebre: this is intentionally using the literal, I want this tests to break if the value of the constant is changed, to ensure that is a conscious action.

support/controlplane-component/defaults_test.go
enxebre: yes, I still want to cover this function in isolation as a consumable piece of code.

PR #8571, opened 2026-05-21
CNTRLPLANE-3329: Enable fuse-overlayfs in ARC runner pods via user namespaces

celebdor:
## Summary

Enables the fuse-overlayfs mount in ARC runner pods so the EFS-backed Go build cache can be used as a zero-copy overlay instead of a full `cp -a` at each job start. PR #8568 added fuse-overlayfs to the runner image and the warm-go-cache action, but the mount was failing because `/dev/fuse` was not exposed and the SCC blocked the `mount(2)` syscall.

Changes:
- **`values.yaml`**: Adds the CRI-O annotation to expose `/dev/fuse`, enables pod-level user namespaces (`hostUsers: false`), runs as uid 0 inside the user namespace (maps to unprivileged host uid), and explicitly requests `CAP_SYS_ADMIN` on the container for the mount syscall
- **`restricted-v3-fuse-scc.yaml`**: Custom SCC based on `restricted-v3` with minimal changes — allows `SYS_ADMIN` in `allowedCapabilities`, drops only `MKNOD`/`NET_RAW` instead of `ALL`, permits uid 0, and requires `userNamespaceLevel: RequirePodLevel` so these privileges only exist inside the user namespace

### Why user namespaces?

`fuse-overlayfs` needs either `fusermount3` (setuid, blocked by `allowPrivilegeEscalation: false`) or `CAP_SYS_ADMIN` for `mount(2)`. Running in a user namespace (`hostUsers: false`) means uid 0 and `CAP_SYS_ADMIN` inside the pod map to an unprivileged host identity — no actual privilege escalation on the host.

### Pre-deploy step

The runner service account must be granted the SCC before helm upgrade:
```
oc adm policy add-scc-to-user restricted-v3-fuse \
  -z arc-runner-set-gha-rs-no-permission -n arc-runners
```

## Test plan

- [x] Verified fuse-overlayfs mount works with this SCC + user namespace config on the cluster
- [ ] Deploy via helm upgrade and confirm runner pods start with the new SCC
- [ ] Confirm fuse-overlayfs mount succeeds in CI job logs (no "falling back to copy" warning)
- [ ] Compare job durations before/after using `contrib/ci/gha-cache-timing.sh`

## Summary by CodeRabbit

* **Chores**
  * Enhanced security configuration for GitHub Actions runners to enable unprivileged FUSE mounts in user namespaces: added SYS_ADMIN capability for the runner container, broadened UID/GID ranges for pod-level user mapping, and configured device access for /dev/fuse.
  * Added pod-level securityContext (runAsUser) and tightened capability drops to maintain controlled escalation while preserving required functionality.

### DIFF
--- a/hack/github-actions-runner/restricted-v3-fuse-scc.yaml
+++ b/hack/github-actions-runner/restricted-v3-fuse-scc.yaml
@@ -1,0 +1,59 @@
+apiVersion: security.openshift.io/v1
+kind: SecurityContextConstraints
+metadata:
+  name: restricted-v3-fuse
+  annotations:
+    kubernetes.io/description: >-
+      Based on restricted-v3 with minimal changes to enable unprivileged
+      FUSE mounts: allows uid 0 and CAP_SYS_ADMIN inside user
+      namespaces. userNamespaceLevel: RequirePodLevel ensures these
+      privileges only exist inside the user namespace and map to
+      unprivileged host identities.
+# Identical to restricted-v3
+allowHostDirVolumePlugin: false
+allowHostIPC: false
+allowHostNetwork: false
+allowHostPID: false
+allowHostPorts: false
+allowPrivilegeEscalation: false
+allowPrivilegedContainer: false
+readOnlyRootFilesystem: false
+priority: null
+seLinuxContext:
+  type: MustRunAs
+seccompProfiles:
+- runtime/default
+userNamespaceLevel: RequirePodLevel
+volumes:
+- configMap
+- csi
+- downwardAPI
+- emptyDir
+- ephemeral
+- image
+- persistentVolumeClaim
+- projected
+- secret
+users: []
+groups: []
+# Changed from restricted-v3: allow SYS_ADMIN for FUSE mounts
+allowedCapabilities:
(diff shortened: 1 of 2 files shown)


bryan-cox (approved):
/lgtm

celebdor:
## Closing in favor of GOCACHEPROG approach

This PR attempted to enable `fuse-overlayfs` in runner pods via user namespaces (`hostUsers: false`) and a custom SCC. However, `hostUsers: false` requires **idmapped mounts** on every volume, and the EFS PVC (NFS-backed) does not support `mount_setattr` on RHEL 9's kernel 5.14:

```
Error: container create failed: mount_setattr `/cache/go-build`
(maybe the file system used doesn't support idmap mounts on this kernel?): Invalid argument
```

NFS idmapped mount support was added upstream in kernel 6.3 and has not been backported to RHEL 9.

**Alternative**: Use Go 1.24+'s `GOCACHEPROG` protocol instead. A small Go program reads from the read-only EFS cache directly (standard filesystem reads, no FUSE) and writes new entries to local temp storage. Zero copy, no special SCC, no user namespaces — works with the default `restricted-v2`.

PR #8576, opened 2026-05-22
CNTRLPLANE-3329: Use GOCACHEPROG for zero-copy EFS build cache

celebdor:
## Summary

Replace fuse-overlayfs with Go 1.24+'s `GOCACHEPROG` protocol for serving the EFS-backed Go build cache to CI jobs. The previous approach (#8571) required user namespaces (`hostUsers: false`) which failed because EFS (NFS) does not support idmapped mounts on RHEL 9's kernel 5.14.

`gocacheprog` is a small Go program (~180 lines, stdlib only) that implements the GOCACHEPROG JSON-over-stdin/stdout protocol:
- **GET**: reads from the read-only EFS PVC first, then the local writable directory
- **PUT**: writes to the local writable directory only
- **Zero copy**: no FUSE, no overlay mounts, just direct filesystem reads from the EFS PVC
- **No special permissions**: works with the default `restricted-v2` SCC

Changes:
- `contrib/ci/gocacheprog/` — new GOCACHEPROG binary (Go module, stdlib only)
- `Dockerfile.github-actions-runner` — builds and installs `gocacheprog` into the runner image
- `.github/actions/warm-go-cache/action.yaml` — sets `GOCACHEPROG` env var instead of `GOCACHE`; falls back to default Go cache if `gocacheprog` binary is not present

Supersedes #8571 (closed — user namespaces + EFS incompatible on this kernel).

## Test plan

- [x] Cold build with GOCACHEPROG: `go build ./support/api/` populates writable cache
- [x] Warm rebuild: 0.28s, all cache hits from writable dir
- [x] Read-only source (simulates EFS): 0.22s, zero new entries in writable dir
- [x] `go test` works with GOCACHEPROG
- [ ] Konflux builds the runner image with gocacheprog
- [ ] Deploy new runner image and verify CI jobs use GOCACHEPROG (no "gocacheprog not found" warning)
- [ ] Compare job durations before/after with `contrib/ci/gha-cache-timing.sh`

## Summary by CodeRabbit

* **New Features**
  * Added a lightweight helper binary to serve a read-only Go build cache with a writable overlay for CI.

* **Chores**
  * CI action updated to use the helper when available and warn/fallback when absent.
  * CI runner image now builds and installs the helper and ensures it’s included in build context.

* **Tests**
  * Added comprehensive tests validating cache operations and concurrency.

### DIFF
--- a/.github/actions/warm-go-cache/action.yaml
+++ b/.github/actions/warm-go-cache/action.yaml
@@ -1,22 +1,17 @@
 name: 'Warm Go build cache'
-description: 'Set GOCACHE via fuse-overlayfs over the EFS-backed PV or a writable fallback'
+description: 'Set GOCACHEPROG to use EFS-backed read-only cache with local writable overlay'
 runs:
   using: composite
   steps:
     - shell: bash
       run: |
         mkdir -p /tmp/go-build-cache
-        mounted=false
-        if [ -d /cache/go-build ] && command -v fuse-overlayfs >/dev/null 2>&1 && [ -e /dev/fuse ]; then
-          mkdir -p /tmp/go-cache-upper /tmp/go-cache-work
-          if fuse-overlayfs -o lowerdir=/cache/go-build,upperdir=/tmp/go-cache-upper,workdir=/tmp/go-cache-work /tmp/go-build-cache; then
-            mounted=true
-          else
-            echo "::warning::fuse-overlayfs mount failed, falling back to copy"
-          fi
+        args="--rw /tmp/go-build-cache"
+        if [ -d /cache/go-build ]; then
+          args="--ro /cache/go-build $args"
         fi
-        if [ "$mounted" = "false" ] && [ -d /cache/go-build ]; then
-          timeout 120 cp -a /cache/go-build/. /tmp/go-build-cache/ || \
-            echo "::warning::Failed to copy EFS cache, proceeding without cache"
+        if command -v gocacheprog >/dev/null 2>&1; then
+          echo "GOCACHEPROG=gocacheprog $args" >> "$GITHUB_ENV"
+        else
+          echo "::warning::gocacheprog not found, using default cache"
         fi
-        echo "GOCACHE=/tmp/go-build-cache" >> "$GITHUB_ENV"
--- a/.github/actions/warm-go-cache/action.yaml
+++ b/.github/actions/warm-go-cache/action.yaml
@@ -5,12 +5,12 @@ runs:
   steps:
     - shell: bash
       run: |
-        mkdir -p /tmp/go-build-cache
-        args="--rw /tmp/go-build-cache"
-        if [ -d /cache/go-build ]; then
-          args="--ro /cache/go-build $args"
-        fi
         if command -v gocacheprog >/dev/null 2>&1; then
+          mkdir -p /tmp/go-build-cache
+          args="--rw /tmp/go-build-cache"
+          if [ -d /cache/go-build ]; then
+            args="--ro /cache/go-build $args"
+          fi
           echo "GOCACHEPROG=gocacheprog $args" >> "$GITHUB_ENV"
         else
           echo "::warning::gocacheprog not found, using default cache"
--- a/contrib/ci/gocacheprog/main.go
+++ b/contrib/ci/gocacheprog/main.go
@@ -1,0 +1,182 @@
+// gocacheprog implements the GOCACHEPROG protocol (Go 1.24+) to serve
+// a read-only Go build cache with a writable overlay. GET requests are
+// served from a writable local directory first, then from a read-only
+// shared directory (e.g. an EFS-backed PVC). PUT requests always write
+// to the local directory. This eliminates the need to copy the shared
+// cache at job start.
+package main
+
+import (
+	"encoding/hex"
+	"encoding/json"
+	"flag"
+	"fmt"
+	"io"
+	"log"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+)
+
+type request struct {
+	ID       int64  `json:"ID"`
+	Command  string `json:"Command"`
+	ActionID []byte `json:"ActionID,omitempty"`
+	OutputID []byte `json:"OutputID,omitempty"`
+	Body     []byte `json:"-"`
+	BodySize int64  `json:"BodySize,omitempty"`
+}
+
+type response struct {
+	ID            int64      `json:"ID"`
+	Err           string     `json:"Err,omitempty"`
+	KnownCommands []string   `json:"KnownCommands,omitempty"`
+	Miss          bool       `json:"Miss,omitempty"`
+	OutputID      []byte     `json:"OutputID,omitempty"`
+	Size          int64      `json:"Size,omitempty"`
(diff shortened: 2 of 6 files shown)


bryan-cox (commented):
GitHub Actions review — 7 findings (6 inline + 1 general).

**Not inline** (outside diff hunk): `Dockerfile.github-actions-runner` line 14 still installs `fuse-overlayfs` via `apt-get`, but the action.yaml no longer uses it. Remove it from the `apt-get install` list to reduce image size and attack surface.

contrib/ci/gocacheprog/main.go
```diff
@@ -0,0 +1,182 @@
...
+			}
+		}
+
+		go func() {
+			res := handleRequest(&req, *roDir, *rwDir)
+			mu.Lock()
+			je.Encode(res)
+			mu.Unlock()
+		}()
+	}
```
bryan-cox, line 84: When stdin hits EOF (line 64), `main()` returns immediately while goroutines spawned here may still be in-flight. Their responses are silently dropped.

Add a `sync.WaitGroup` to drain in-flight handlers before returning:

```go
var wg sync.WaitGroup
for {
    var req request
    if err := jd.Decode(&req); err != nil {
        if err == io.EOF {
            break
        }
        log.Fatalf("gocacheprog: decode request: %v", err)
    }
    // ... body decode ...
    wg.Add(1)
    go func() {
        defer wg.Done()
        res := handleRequest(&req, *roDir, *rwDir)
        mu.Lock()
        je.Encode(res)
        mu.Unlock()
    }()
}
wg.Wait()
```

In practice Go may not wait for trailing responses, but this is the correct pattern and prevents data races on `os.Stdout` during exit.

contrib/ci/gocacheprog/main.go
```diff
@@ -0,0 +1,182 @@
...
+			if err == io.EOF {
+				return
+			}
+			log.Fatalf("gocacheprog: decode request: %v", err)
+		}
+
+		if req.Command == "put" && req.BodySize > 0 {
+			if err := jd.Decode(&req.Body); err != nil {
+				log.Fatalf("gocacheprog: decode body: %v", err)
+			}
```
bryan-cox, line 73: The GOCACHEPROG protocol sends the body as raw bytes after the JSON request line, not as a separate JSON value. Using `jd.Decode(&req.Body)` expects a JSON-encoded token (base64 string). If this has been tested end-to-end and works, Go must be encoding it as JSON — but worth double-checking against the Go 1.24 GOCACHEPROG spec to make sure this isn't silently truncating or corrupting cache entries.

contrib/ci/gocacheprog/main.go
```diff
@@ -0,0 +1,182 @@
...
+	h := hex.EncodeToString(id)
+	return filepath.Join(dir, h[:2], h+"-a")
+}
+
+// outputFile returns the path to a Go cache data file.
+// Format: <dir>/<first-byte-hex>/<full-hex-outputID>-d
+func outputFile(dir string, id []byte) string {
+	h := hex.EncodeToString(id)
+	return filepath.Join(dir, h[:2], h+"-d")
+}
```
bryan-cox, line 119: If `id` is an empty byte slice, `hex.EncodeToString(id)` returns `""` and `h[:2]` panics with a slice bounds error. Go shouldn't send empty IDs, but for a standalone binary a guard is cheap:

```go
func actionFile(dir string, id []byte) string {
    if len(id) == 0 {
        return ""
    }
    h := hex.EncodeToString(id)
    return filepath.Join(dir, h[:2], h+"-a")
}
```

Same for `outputFile`.

contrib/ci/gocacheprog/main.go
```diff
@@ -0,0 +1,182 @@
...
+
+func handlePut(req *request, rwDir string) response {
+	dPath := outputFile(rwDir, req.OutputID)
+	os.MkdirAll(filepath.Dir(dPath), 0o777)
+	if err := os.WriteFile(dPath, req.Body, 0o666); err != nil {
+		return response{ID: req.ID, Err: err.Error()}
+	}
+
+	aPath := actionFile(rwDir, req.ActionID)
+	os.MkdirAll(filepath.Dir(aPath), 0o777)
```
bryan-cox: Both `os.MkdirAll` calls (lines 163 and 169) ignore errors. If directory creation fails (permissions, disk full), the subsequent `os.WriteFile` fails with a confusing "no such file or directory". Check the error:

```go
if err := os.MkdirAll(filepath.Dir(dPath), 0o777); err != nil {
    return response{ID: req.ID, Err: err.Error()}
}
```

Same for `aPath` on line 169.

contrib/ci/gocacheprog/main.go
```diff
@@ -0,0 +1,182 @@
...
+	entry := fmt.Sprintf("v1 %s %s %d %d\n",
+		hex.EncodeToString(req.ActionID),
+		hex.EncodeToString(req.OutputID),
+		len(req.Body),
+		time.Now().UnixNano(),
+	)
```
bryan-cox, line 189: The size in the action entry uses `len(req.Body)` but the protocol provides `req.BodySize` as the authoritative value. These could differ if the body was partially read or if there's a type mismatch (int vs int64). Use `req.BodySize` for consistency:

```go
entry := fmt.Sprintf("v1 %s %s %d %d\n",
    hex.EncodeToString(req.ActionID),
    hex.EncodeToString(req.OutputID),
    req.BodySize,
    time.Now().UnixNano(),
)
```

.github/actions/warm-go-cache/action.yaml
```diff
@@ -1,22 +1,17 @@
 name: 'Warm Go build cache'
-description: 'Set GOCACHE via fuse-overlayfs over the EFS-backed PV or a writable fallback'
+description: 'Set GOCACHEPROG to use EFS-backed read-only cache with local writable overlay'
 runs:
   using: composite
   steps:
     - shell: bash
       run: |
```
bryan-cox, line 7: Nit: `mkdir -p /tmp/go-build-cache` is only needed when `gocacheprog` is found (it becomes the `--rw` dir). In the fallback path (no `gocacheprog`), the directory is created but never used. Consider moving it inside the `if command -v gocacheprog` block.

contrib/ci/gocacheprog/main.go
celebdor: Good catch to double-check. I verified this end-to-end (cold build, warm rebuild, read-only source) and it works correctly.

The Go toolchain sends the body as a **JSON-encoded base64 string** on a separate line after the request JSON — this is how `encoding/json` marshals `[]byte` fields. The reference implementation ([bradfitz/go-tool-cache](https://github.com/bradfitz/go-tool-cache/blob/main/cmd/go-cacher/cacher.go)) also reads the body with `jd.Decode(&req.ObjectID)` / `jd.Decode(&bodyb)` using `json.Decoder`, which expects JSON tokens (i.e., base64 strings for byte slices). So `jd.Decode(&req.Body)` correctly decodes the base64 JSON string into `[]byte`.

The key confusion is that the [internal spec](https://pkg.go.dev/cmd/go/internal/cacheprog) says "body bytes" but the wire format is JSON-all-the-way — each body is a JSON value (base64 string) on its own line, not raw bytes on the stream.

PR #8579, opened 2026-05-22
OCPBUGS-86661: Konnectivity retry proxy connection on timeout

YamunadeviShanmugam:
This PR fixes the intermittent E2E test failures where the konnectivity-proxy-socks5 sidecar container crashes and restarts during cluster bootstrap. Fix includes retry loop, native Kubernetes startup probes, and explicit component dependency ordering.
Proxy Internal Retry Logic
- Added a new retry infrastructure with exponential backoff to handle transient connection errors safely. The container process now stays alive while waiting for the network tunnel to become available.

Native Startup Probes
konnectivity-container.go: Added a native TCP startup probe to all Konnectivity sidecar containers. This gives the sidecars a safe 60-second window to initialize during cluster bootstrap before any readiness or liveness checks trigger.

Component Dependency Ordering
Deployment graph is updated with dependency check. The openshift-oauth-apiserver pods will now explicitly wait for the konnectivity-agent to be completely Available and RolloutComplete before they begin scheduling.

Test Fixtures are updated accordingly 

## Summary by CodeRabbit

- **New Features**
  - Added automatic recovery for temporary Konnectivity connection and startup failures.
  - Added graceful shutdown handling for the SOCKS5 proxy.
  - Added startup health checks for Konnectivity containers, including support for custom serving ports.
  - Integrated Konnectivity as a dependency of the OAuth API server.

- **Bug Fixes**
  - Improved error handling to distinguish retryable failures from permanent errors.

- **Tests**
  - Expanded coverage for retries, shutdown, health checks, configuration, and component initialization.

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/v2/oauth_apiserver/component.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/v2/oauth_apiserver/component.go
@@ -2,6 +2,7 @@ package oapi
 
 import (
 	kasv2 "github.com/openshift/hypershift/control-plane-operator/controllers/hostedcontrolplane/v2/kas"
+	konnectivityagent "github.com/openshift/hypershift/control-plane-operator/controllers/hostedcontrolplane/v2/konnectivity_agent"
 	oapiv2 "github.com/openshift/hypershift/control-plane-operator/controllers/hostedcontrolplane/v2/oapi"
 	component "github.com/openshift/hypershift/support/controlplane-component"
 	"github.com/openshift/hypershift/support/podspec"
@@ -46,7 +47,7 @@ func NewComponent() component.ControlPlaneComponent {
 			"pdb.yaml",
 			component.AdaptPodDisruptionBudget(),
 		).
-		WithDependencies(oapiv2.ComponentName).
+		WithDependencies(oapiv2.ComponentName, konnectivityagent.ComponentName).
 		InjectAvailabilityProberContainer(podspec.AvailabilityProberOpts{}).
 		InjectKonnectivityContainer(component.KonnectivityContainerOptions{
 			Mode: component.Socks5,
--- a/control-plane-operator/controllers/hostedcontrolplane/v2/oauth_apiserver/component_test.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/v2/oauth_apiserver/component_test.go
@@ -135,3 +135,12 @@ func TestPredicate(t *testing.T) {
 		})
 	}
 }
+
+func TestNewComponent(t *testing.T) {
+	t.Run("When creating component it should build successfully", func(t *testing.T) {
+		g := NewWithT(t)
+		component := NewComponent()
+		g.Expect(component).ToNot(BeNil())
+		g.Expect(component.Name()).To(Equal(ComponentName))
+	})
+}
--- a/konnectivity-socks5-proxy/guard.go
+++ b/konnectivity-socks5-proxy/guard.go
@@ -1,0 +1,266 @@
+package konnectivitysocks5proxy
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net"
+	"os"
+	"strings"
+	"syscall"
+	"time"
+
+	"github.com/openshift/hypershift/support/konnectivityproxy"
+
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/armon/go-socks5"
+	"github.com/go-logr/logr"
+)
+
+const (
+	defaultKonnectivityHost = "konnectivity-server-local"
+	defaultKonnectivityPort = 8090
+	konnectivityDialTimeout = 5 * time.Second
+)
+
+// coreGuardBackoffConfig holds exponential backoff parameters for konnectivity bootstrap and serve retries.
+type coreGuardBackoffConfig struct {
+	initialDelay time.Duration
+	factor       float64
+	jitter       float64
+	steps        int
+	cap          time.Duration
+}
+
+func (c coreGuardBackoffConfig) waitBackoff() wait.Backoff {
+	return wait.Backoff{
+		Duration: c.initialDelay,
+		Factor:   c.factor,
+		Jitter:   c.jitter,
+		Steps:    c.steps,
+		Cap:      c.cap,
+	}
+}
+
+var (
+	coreGuardBackoff = coreGuardBackoffConfig{
+		initialDelay: 1 * time.Second,
+		factor:       2.0,
+		jitter:       0.1,
+		steps:        5,
+		cap:          30 * time.Second,
+	}
+	dialKonnectivityServer     = dialKonnectivityServerTCP
+	bootstrapKonnectivityFn    = bootstrapKonnectivity
+	tryBootstrapKonnectivityFn = tryBootstrapKonnectivity
+	getConfigFn                = ctrl.GetConfig
+	newClientFn                = client.New
+)
+
+// runWithCoreGuard keeps the process alive while konnectivity infrastructure becomes
+// reachable, retrying transient failures with exponential backoff instead of exiting.
+func runWithCoreGuard(ctx context.Context, log logr.Logger, opts konnectivityproxy.Options, servingPort uint32, serve func(dialer konnectivityproxy.ProxyDialer) error) error {
+	delay := coreGuardBackoff.waitBackoff().DelayFunc()
+	for {
+		if err := ctx.Err(); err != nil {
+			return err
+		}
+
+		dialer, err := bootstrapKonnectivityFn(ctx, log, opts)
+		if err != nil {
+			return err
+		}
+
+		log.Info("Konnectivity socks5 proxy bootstrap complete, starting server", "port", servingPort)
+		if err := serve(dialer); err != nil {
+			if !isTransientKonnectivityError(err) {
+				return fmt.Errorf("socks5 server exited: %w", err)
+			}
+			log.Error(err, "socks5 server stopped due to transient error, retrying")
+			sleepDuration := delay()
+			select {
+			case <-ctx.Done():
+				return ctx.Err()
+			case <-time.After(sleepDuration):
+				continue
+			}
+		}
+		return nil
+	}
+}
+
+// serveWithGracefulShutdown creates and runs a socks5 server with proper graceful shutdown handling.
+// It returns nil on clean shutdown (context cancellation), or an error if the server fails.
+func serveWithGracefulShutdown(ctx context.Context, dialer konnectivityproxy.ProxyDialer, servingPort uint32, log logr.Logger) error {
+	conf := &socks5.Config{
+		Dial:     dialer.DialContext,
+		Resolver: dialer,
+	}
+	server, err := socks5.New(conf)
+	if err != nil {
+		return fmt.Errorf("cannot create socks5 server: %w", err)
+	}
+
+	// Create listener explicitly for graceful shutdown
+	lc := &net.ListenConfig{}
+	listener, err := lc.Listen(ctx, "tcp", fmt.Sprintf(":%d", servingPort))
+	if err != nil {
+		return fmt.Errorf("cannot create listener: %w", err)
+	}
+
+	// Run server in goroutine
+	serveDone := make(chan error, 1)
+	go func() {
+		serveDone <- server.Serve(listener)
+	}()
+
+	// Wait for either context cancellation or serve error
+	select {
+	case <-ctx.Done():
+		// Graceful shutdown: close listener to stop accepting new connections
+		if err := listener.Close(); err != nil {
+			log.Error(err, "failed to close listener during shutdown")
+		}
+		// Wait for Serve to finish
+		<-serveDone
+		// Treat context cancellation as clean shutdown
+		return nil
+	case err := <-serveDone:
+		// Server stopped, return the error
+		return err
+	}
+}
+
+func bootstrapKonnectivity(ctx context.Context, log logr.Logger, opts konnectivityproxy.Options) (konnectivityproxy.ProxyDialer, error) {
+	// Create kube client once - client creation failures are permanent (missing files, bad config)
+	// and should not be retried. Only konnectivity server availability is retried.
+	cfg, err := getConfigFn()
+	if err != nil {
+		return nil, fmt.Errorf("cannot get client config: %w", err)
+	}
+
+	kubeClient, err := newClientFn(cfg, client.Options{})
+	if err != nil {
+		return nil, fmt.Errorf("cannot get client: %w", err)
+	}
+
+	opts.Client = kubeClient
+
+	delay := coreGuardBackoff.waitBackoff().DelayFunc()
+	attempt := 0
+
+	for {
+		if err := ctx.Err(); err != nil {
+			return nil, err
+		}
+		attempt++
+
+		d, err := tryBootstrapKonnectivityFn(ctx, opts)
+		if err == nil {
+			return d, nil
+		}
+		if !isTransientKonnectivityError(err) {
+			return nil, fmt.Errorf("konnectivity bootstrap failed: %w", err)
+		}
+
+		log.Error(err, "transient konnectivity bootstrap failure, retrying", "attempt", attempt)
+		sleepDuration := delay()
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		case <-time.After(sleepDuration):
+			// Continue retry loop
+		}
+	}
+}
+
+func tryBootstrapKonnectivity(ctx context.Context, opts konnectivityproxy.Options) (konnectivityproxy.ProxyDialer, error) {
+	dialer, err := konnectivityproxy.NewKonnectivityDialer(opts)
+	if err != nil {
+		return nil, fmt.Errorf("cannot initialize konnectivity dialer: %w", err)
+	}
+
+	host := opts.KonnectivityHost
+	if host == "" {
+		host = defaultKonnectivityHost
+	}
+	port := opts.KonnectivityPort
+	if port == 0 {
+		port = defaultKonnectivityPort
+	}
+
+	if err := dialKonnectivityServer(ctx, host, port); err != nil {
+		return nil, fmt.Errorf("cannot dial konnectivity server at %s:%d: %w", host, port, err)
+	}
+
+	return dialer, nil
+}
+
+func dialKonnectivityServerTCP(ctx context.Context, host string, port uint32) error {
+	address := net.JoinHostPort(host, fmt.Sprintf("%d", port))
+	conn, err := (&net.Dialer{Timeout: konnectivityDialTimeout}).DialContext(ctx, "tcp", address)
+	if err != nil {
+		return err
+	}
+	_ = conn.Close()
+	return nil
+}
+
+func isTransientKonnectivityError(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	if errors.Is(err, context.DeadlineExceeded) {
+		return true
+	}
+
+	var netErr net.Error
+	if errors.As(err, &netErr) && netErr.Timeout() {
+		return true
+	}
+
+	var opErr *net.OpError
+	if errors.As(err, &opErr) {
+		if isTransientKonnectivityError(opErr.Err) {
+			return true
+		}
+	}
+
+	var syscallErr syscall.Errno
+	if errors.As(err, &syscallErr) {
+		switch syscallErr {
+		case syscall.ECONNREFUSED, syscall.ECONNRESET, syscall.ETIMEDOUT,
+			syscall.EHOSTUNREACH, syscall.ENETUNREACH, syscall.EPIPE:
+			return true
+		}
+	}
+
+	// Configuration errors such as missing certificate files should fail fast, not retry.
+	if errors.Is(err, os.ErrNotExist) {
+		return false
+	}
+
+	msg := err.Error()
+	transientFragments := []string{
+		"connection refused",
+		"connection reset",
+		"i/o timeout",
+		"context deadline exceeded",
+		"no route to host",
+		"network is unreachable",
+		"operation timed out",
+		"TLS handshake timeout",
+	}
+	for _, fragment := range transientFragments {
+		if strings.Contains(msg, fragment) {
+			return true
+		}
+	}
+
+	return false
+}
--- a/konnectivity-socks5-proxy/guard_test.go
+++ b/konnectivity-socks5-proxy/guard_test.go
@@ -1,0 +1,503 @@
+package konnectivitysocks5proxy
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net"
+	"os"
+	"syscall"
+	"testing"
+	"time"
+
+	. "github.com/onsi/gomega"
+
+	"github.com/openshift/hypershift/support/konnectivityproxy"
+
+	"k8s.io/client-go/rest"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/go-logr/logr"
+)
+
+// mockProxyDialer implements konnectivityproxy.ProxyDialer interface
+type mockProxyDialer struct{}
+
+func (m *mockProxyDialer) DialContext(ctx context.Context, network, address string) (net.Conn, error) {
+	return nil, nil
+}
+
+func (m *mockProxyDialer) Dial(network, address string) (net.Conn, error) {
+	return nil, nil
+}
+
+func (m *mockProxyDialer) Resolve(ctx context.Context, name string) (context.Context, net.IP, error) {
+	return ctx, net.ParseIP("127.0.0.1"), nil
+}
+
+func (m *mockProxyDialer) IsCloudAPI(host string) bool {
+	return false
+}
+
+func waitBackoffForTest() coreGuardBackoffConfig {
+	return coreGuardBackoffConfig{
+		initialDelay: 10 * time.Millisecond,
+		factor:       2.0,
+		cap:          50 * time.Millisecond,
+	}
+}
+
+// setupRunWithCoreGuardTest saves package-level variables and restores them via t.Cleanup.
+// Use this helper for tests that mock runWithCoreGuard behavior.
+func setupRunWithCoreGuardTest(t *testing.T) {
+	t.Helper()
+	originalBackoff := coreGuardBackoff
+	originalBootstrap := bootstrapKonnectivityFn
+
+	coreGuardBackoff = waitBackoffForTest()
+
+	t.Cleanup(func() {
+		coreGuardBackoff = originalBackoff
+		bootstrapKonnectivityFn = originalBootstrap
+	})
+}
+
+// setupBootstrapKonnectivityTest saves package-level variables and restores them via t.Cleanup.
+// Use this helper for tests that mock bootstrapKonnectivity behavior.
+func setupBootstrapKonnectivityTest(t *testing.T) {
+	t.Helper()
+	originalBackoff := coreGuardBackoff
+	originalTry := tryBootstrapKonnectivityFn
+	originalGetConfig := getConfigFn
+	originalNewClient := newClientFn
+
+	coreGuardBackoff = waitBackoffForTest()
+
+	// Stub getConfigFn and newClientFn to avoid environment dependency
+	getConfigFn = func() (*rest.Config, error) {
+		return &rest.Config{}, nil
+	}
+	newClientFn = func(config *rest.Config, options client.Options) (client.Client, error) {
+		return nil, nil // Return nil client - not used since tryBootstrapKonnectivityFn is mocked
+	}
+
+	t.Cleanup(func() {
+		coreGuardBackoff = originalBackoff
+		tryBootstrapKonnectivityFn = originalTry
+		getConfigFn = originalGetConfig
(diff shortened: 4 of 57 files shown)


konnectivity-socks5-proxy/guard.go
```diff
@@ -0,0 +1,200 @@
...
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		case <-time.After(sleepDuration):
+			// Continue retry loop
+		}
+	}
+}
+
+func tryBootstrapKonnectivity(ctx context.Context, opts konnectivityproxy.Options) (konnectivityproxy.ProxyDialer, error) {
+	cfg, err := ctrl.GetConfig()
```
jparrill: This creates a new REST config + kube client on every retry — new transports, TLS handshakes, disk reads each time. Could leak file descriptors under rapid retries.

Hoist client creation to `bootstrapKonnectivity` and pass it in. If the client becomes stale (token expiry, cert rotation), catch the specific auth error and recreate lazily — but the default should be reuse.

konnectivity-socks5-proxy/guard.go
```diff
@@ -0,0 +1,200 @@
...
+	if host == "" {
+		host = defaultKonnectivityHost
+	}
+	port := opts.KonnectivityPort
+	if port == 0 {
+		port = defaultKonnectivityPort
+	}
+
+	if err := dialKonnectivityServer(ctx, host, port); err != nil {
+		return nil, err
```
jparrill: This error is returned bare while all others in this function are wrapped. Wrap it for consistency and include the address for debugging:
```go
return nil, fmt.Errorf("cannot dial konnectivity server at %s:%d: %w", host, port, err)
```

konnectivity-socks5-proxy/guard.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+	dialKonnectivityServer     = dialKonnectivityServerTCP
+	bootstrapKonnectivityFn    = bootstrapKonnectivity
+	tryBootstrapKonnectivityFn = tryBootstrapKonnectivity
+)
+
+// runWithCoreGuard keeps the process alive while konnectivity infrastructure becomes
+// reachable, retrying transient failures with exponential backoff instead of exiting.
+func runWithCoreGuard(ctx context.Context, log logr.Logger, opts konnectivityproxy.Options, servingPort uint32, serve func(dialer konnectivityproxy.ProxyDialer) error) error {
+	delay := coreGuardBackoff.DelayFunc()
```
jparrill: The delay is never reset. If serve runs for hours then hits a transient blip, the backoff picks up at the cap instead of starting fresh. Intentional? If not, reset `delay` at the top of the retry branch.

konnectivity-socks5-proxy/guard.go
```diff
@@ -0,0 +1,200 @@
...
+
+var (
+	coreGuardBackoff = wait.Backoff{
+		Duration: 1 * time.Second,
+		Factor:   2.0,
+		Jitter:   0.1,
+		Steps:    5,
+		Cap:      30 * time.Second,
+	}
+	dialKonnectivityServer     = dialKonnectivityServerTCP
```
jparrill, line 57: These globals make tests non-parallelizable forever. A struct with these as fields would be more idiomatic. Not blocking for a bug fix, but worth considering if this grows.

konnectivity-socks5-proxy/main.go
```diff
@@ -51,32 +51,49 @@ func NewStartCommand() *cobra.Command {
...
+			serveDone := make(chan error, 1)
+			go func() {
+				serveDone <- server.Serve(listener)
+			}()
+
+			// Wait for either context cancellation or serve error
+			select {
+			case <-ctx.Done():
+				// Graceful shutdown: close listener to stop accepting new connections
+				_ = listener.Close()
```
jparrill: If this fails, `<-serveDone` below hangs. At least log the error.

support/controlplane-component/konnectivity-container.go
```diff
@@ -180,6 +181,22 @@ func (opts KonnectivityContainerOptions) buildContainer(hcp *hyperv1.HostedContr
 		kubeconfingVolumeName = "kubeconfig"
 	}
 
+	var servingPort int32
```
jparrill: Both branches are identical except for the field name. Simplify with `var servingPort int32 = 8090` as default and only override in each case. Also `8090` appears in 4 places across the codebase without a shared constant.

control-plane-operator/controllers/hostedcontrolplane/v2/oauth_apiserver/component.go
```diff
@@ -46,7 +47,7 @@ func NewComponent() component.ControlPlaneComponent {
 			"pdb.yaml",
 			component.AdaptPodDisruptionBudget(),
 		).
-		WithDependencies(oapiv2.ComponentName).
+		WithDependencies(oapiv2.ComponentName, konnectivityagent.ComponentName).
```
jparrill, line 50: 7 other components also inject konnectivity sidecars and none declare this dependency. What makes oauth-apiserver special here? If retry + startup probe is enough (and it's applied globally), should this be removed? If it's genuinely needed, should the others get it too? A comment explaining the rationale would help.

konnectivity-socks5-proxy/guard_test.go
```diff
@@ -0,0 +1,573 @@
+package konnectivitysocks5proxy
```
jparrill, line 1: These tests mutate package-level vars — `t.Parallel()` can never be used. Worth a comment so future contributors don't add it.

konnectivity-socks5-proxy/guard_test.go
```diff
@@ -0,0 +1,573 @@
...
+		g.Expect(konnectivityDialTimeout).To(Equal(5 * time.Second))
+	})
+}
+
+func TestRunWithCoreGuard(t *testing.T) {
+	t.Run("When serve succeeds immediately, it should return nil", func(t *testing.T) {
+		g := NewGomegaWithT(t)
+
+		// Save originals
+		originalBackoff := coreGuardBackoff
```
jparrill: This block is repeated ~9 times. A `t.Cleanup`-based helper would cut the noise significantly.

konnectivity-socks5-proxy/guard_test.go
```diff
@@ -0,0 +1,573 @@
...
+
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		defer cancel()
+
+		g.Expect(retryDialKonnectivityServer(ctx, logr.Discard(), "127.0.0.1", port)).To(Succeed())
+		g.Expect(attempts).To(BeNumerically(">=", 2))
+	})
+}
+
+func retryDialKonnectivityServer(ctx context.Context, log logr.Logger, host string, port uint32) error {
```
jparrill: This reimplements the production retry loop. `TestBootstrapKonnectivity` already covers retry-then-succeed — can this be removed or go through `bootstrapKonnectivity` directly?

konnectivity-socks5-proxy/main_test.go
```diff
@@ -0,0 +1,160 @@
...
+	})
+
+	t.Run("When validating graceful shutdown behavior in serve function", func(t *testing.T) {
+		g := NewGomegaWithT(t)
+
+		// This test verifies that the serve function respects context cancellation
+		// by checking that it returns nil (clean shutdown) when context is canceled
+
+		// Create the serve function (same pattern as in main.go)
+		serveFunc := func(ctx context.Context, servingPort uint32) error {
```
jparrill: This copy-pastes the serve function from `main.go`. If the production code changes, this test won't catch it. Either extract the serve logic into a named function, or rely on `TestRunWithCoreGuard` which already covers this path.

support/controlplane-component/konnectivity-container_test.go
```diff
@@ -0,0 +1,208 @@
...
+		g.Expect(podSpec.Containers).To(HaveLen(2))
+		g.Expect(podSpec.Containers[0].Name).To(Equal("konnectivity-proxy-https"))
+		g.Expect(podSpec.Containers[0].StartupProbe).ToNot(BeNil())
+		g.Expect(podSpec.Containers[1].Name).To(Equal("konnectivity-proxy-socks5"))
+		g.Expect(podSpec.Containers[1].StartupProbe).ToNot(BeNil())
+	})
+}
+
+// fakeImageProvider implements ReleaseImageProvider for testing
+type fakeImageProvider struct{}
```
jparrill: `testutil.FakeImageProvider()` already exists and is used in this package's other tests. Reuse it.

jparrill (commented):
Dropped some comments. Thanks

Apart from those comments inline, you will need to:
- Create a bug in OCPBUGS project
- `/retitle OCPBUGS-XXXXX: Konnectivity retry proxy connection on timeout`
- `/jira refresh`

konnectivity-socks5-proxy/guard.go
YamunadeviShanmugam: Addressed via 
index 6568b84c2..a2b6b0553 100644
--- a/konnectivity-socks5-proxy/guard.go
+++ b/konnectivity-socks5-proxy/guard.go
@@ -128,7 +128,7 @@ func tryBootstrapKonnectivity(ctx context.Context, opts konnectivityproxy.Option
        }
 
        if err := dialKonnectivityServer(ctx, host, port); err != nil {
-               return nil, err
+               return nil, fmt.Errorf("cannot dial konnectivity server at %s:%d: %w", host, port, err)
        }
 
        return dialer, nil

support/controlplane-component/konnectivity-container_test.go
YamunadeviShanmugam: addressed

konnectivity-socks5-proxy/guard.go
```diff
@@ -0,0 +1,200 @@
...
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/go-logr/logr"
+)
+
+const (
+	defaultKonnectivityHost = "konnectivity-server-local"
```
enxebre, line 25: can you point to the exact root cause that is causing the restarts?
doesn't DialContext already handles transient failures?
in addition to the component dep, do we really need anything else that something like
```
                err = wait.ExponentialBackoff(
                        wait.Backoff{Duration: time.Second, Factor: 2, Jitter: 0.1, Steps: 10, Cap: 15 * time.Second},
                        func() (bool, error) {
                                if ctx.Err() != nil {
                                        return false, ctx.Err()
                                }
                                d, err := konnectivityproxy.NewKonnectivityDialer(opts)
                                if err != nil {
                                        l.Error(err, "konnectivity dialer init failed, retrying")
                                        return false, nil // retry
                                }
                                dialer = d
                                return true, nil // done
                        },
                )
```

support/controlplane-component/konnectivity-container.go
YamunadeviShanmugam: Addressed

konnectivity-socks5-proxy/main.go
YamunadeviShanmugam: Addressed

konnectivity-socks5-proxy/guard.go
YamunadeviShanmugam: addressed it by updating it to a struct

konnectivity-socks5-proxy/guard_test.go
YamunadeviShanmugam: Agreed, removing it is a duplicate test covering the same path, removed it

konnectivity-socks5-proxy/guard.go
YamunadeviShanmugam: Addressed by resetting the delay at the top of retry branch

konnectivity-socks5-proxy/guard_test.go
YamunadeviShanmugam: Noted for further contributions

konnectivity-socks5-proxy/main_test.go
YamunadeviShanmugam: addressed

konnectivity-socks5-proxy/guard_test.go
YamunadeviShanmugam: Addressed

konnectivity-socks5-proxy/guard.go
YamunadeviShanmugam: Addressed

konnectivity-socks5-proxy/guard.go
YamunadeviShanmugam: @enxebre **Doesn't DialContext Handle Transient Failures?**  
 No - DialContext is called lazily, not during bootstrap. 
NewKonnectivityDialer only reads cert files and creates a struct; it doesn't connect. Connection happens when the main container makes its first socks5 request. 

Without the dial check during bootstrap: 
(1) NewKonnectivityDialer succeeds, 
(2) socks5 server starts listening, 
(3) but konnectivity-server is unreachable, so  all requests fail. 
The dial check provides pre-flight verification that connectivity works before declaring the sidecar ready. It's fail-fast feedback, not request-time retry.

**Do We Need More Than Component Dep +  Backoff?** 
NewKonnectivityDialer doesn't connect; it only reads cert files from disk. Retrying it handles volume-mount races but doesn't verify connectivity. Without the dial check, there's no connectivity verification during bootstrap. It will not solve the network connectivity issue
```

CI failure 
Container konnectivity-proxy-socks5 in pod openshift-oauth-apiserver has a restartCount > 0 (1)
Error: context deadline exceeded
Summary
All 10 test failures originate from 3 EnsureNoCrashingPods sub-test failures (the remaining 7 are cascading parent test failures). In 3 of 6 hosted clusters (TestAutoscaling, TestAzurePrivateTopology, TestNodePool/HostedCluster2), the konnectivity-proxy-socks5 sidecar container in openshift-oauth-apiserver pods hit a context deadline exceeded error during initial startup and restarted once. The EnsureNoCrashingPods check uses defaultCrashToleration=0 for Azure platform, so even a single restart causes failure. This is a pre-existing, non-deterministic infrastructure flake — 3 of 6 clusters hit it while the other 3 (including TestCreateCluster, TestAzureOAuthLoadBalancer, TestNodePool/HostedCluster0) had zero restarts and passed.

Root Cause
The konnectivity-proxy-socks5 container (/usr/bin/control-plane-operator konnectivity-socks5-proxy run --resolve-from-guest-cluster-dns=true) is a SOCKS5 proxy sidecar injected into hosted cluster control plane pods. During hosted cluster startup, this container races to establish connectivity to the Konnectivity server on the management cluster.

What happened chronologically (identical across all 3 failing clusters):

08:50:19-23Z — Pod scheduled and init containers started
08:51:26-55Z — Main containers (openshift-oauth-apiserver, audit-logs, kas-readiness-check) started
08:53:56-58Z — konnectivity-proxy-socks5 failed with Error: context deadline exceeded (~2 minutes after pod scheduling, before konnectivity container ever started successfully)
08:54:05-06Z — Container restarted and started successfully
The context deadline exceeded indicates the container has an internal startup timeout. During initial cluster bootstrap, the Konnectivity server may not yet be ready to accept connections. The konnectivity-proxy-socks5 container times out waiting for the server, exits with an error, and CRI-O restarts it. On the second attempt, the server is ready, and the container succeeds.

Why some clusters pass and others don't: The passing clusters happened to have their Konnectivity server ready before the socks5 proxy's internal deadline expired (passing cluster's konnectivity started at 08:51:03Z vs failing cluster's initial attempt window of 08:51:55Z–08:53:58Z). This is purely a timing race during bootstrap.
```

YamunadeviShanmugam:
/retitle OCPBUGS-86661: Konnectivity retry proxy connection on timeout

jparrill:
## Re-review: dependency scope and bootstrap impact

I've verified the latest push addresses most of my previous review comments (struct conversion, delay reset, error wrapping, client hoisting, test helpers). Good work there.

However, the dependency scope has grown significantly since my first review, and I want to flag this explicitly:

### 8 components now depend on `konnectivity-agent`

The original bug (OCPBUGS-86661) was about the konnectivity-socks5-proxy sidecar crashing during bootstrap. The retry logic + startup probe should handle that at the container level. But this PR also adds `WithDependencies(konnectivityagent.ComponentName)` to **8 components**:

- `cno`
- `ingress-operator`
- `openshift-apiserver` (oapi)
- `oauth`
- `oauth-apiserver`
- `catalog-operator`
- `olm-operator`
- `packageserver`

This means none of these components will start scheduling until konnectivity-agent reports `Available` + `RolloutComplete`. That's a significant change to the control plane bootstrap graph — these components are now serialized behind konnectivity-agent.

**Questions:**
1. Has the impact on bootstrap time been measured? Even a few seconds of added latency per component compounds across 8.
2. If the retry loop + startup probe is sufficient to handle the sidecar race (which it should be — that's the whole point of the guard), do we actually need the component-level dependency at all? The dependency is a belt-and-suspenders approach that also constrains upgrades/rollouts.
3. Previously only `oauth-apiserver` had this dependency. What changed to justify all 8? The answer "all of them inject a konnectivity sidecar" is true but was also true before this bug — if they worked without the dependency before (modulo the sidecar restart), and the retry/probe now handles the restart, what's the net justification?

I'd like to see data or at minimum a clear rationale before approving this change to the deployment graph.

jparrill:
@enxebre — ping on this PR. Your question from May 26 about the root cause and whether this needs more than dependency + simple `wait.ExponentialBackoff` is still the key design question.

Yamuna responded explaining that `DialContext` is lazy (not called during bootstrap), so the sidecar can start listening but fail all proxied requests if konnectivity-server isn't reachable yet. That justifies the dial check in the guard. But the scope has expanded: there are now 3 layers of defense (retry guard, startup probe, AND component dependency on konnectivity-agent for 8 components), which feels over-engineered for a transient bootstrap race.

Could you take a look at the current state and share whether:
1. You're OK with the approach, or prefer something simpler
2. The 8-component dependency addition is acceptable or should be scoped down

This PR has been open for ~2 weeks with no approvals, so it'd be good to unblock or redirect. Thanks!

YamunadeviShanmugam:
@jparrill : I looked closer at the changes, and I agree with you
we only need to keep WithDependencies(konnectivity-agent) on oauth-apiserver. I'm reverting it for the other 7 components (cno, ingress-operator, openshift-apiserver, oauth, catalog-operator, olm-operator, and packageserver).
oauth-apiserver is a special case as itneeds the dependency because data plane webhooks have to reach back into it to function. It's a hard requirement, not a timing workaround. For other 7 components it is not needed.

control-plane-operator/controllers/hostedcontrolplane/v2/oauth_apiserver/component.go
YamunadeviShanmugam: we only need to keep WithDependencies(konnectivity-agent) on oauth-apiserver. 
oauth-apiserver is a special case as itneeds the dependency because data plane webhooks have to reach back into it to function. For other 7 components it is not needed.

jparrill (commented):
/lgtm

control-plane-operator/controllers/hostedcontrolplane/v2/olm/catalog_operator/component_test.go
```diff
@@ -0,0 +1,16 @@
+package catalogoperator
+
```
enxebre: how's this change related to this PR?

control-plane-operator/controllers/hostedcontrolplane/v2/olm/olm_operator/component_test.go
```diff
@@ -0,0 +1,16 @@
...
+import (
+	"testing"
+
+	. "github.com/onsi/gomega"
+)
+
+func TestNewComponent(t *testing.T) {
+	t.Run("When creating component it should build successfully", func(t *testing.T) {
+		g := NewWithT(t)
+		component := NewComponent()
```
enxebre: how's this change related to this PR?

control-plane-operator/controllers/hostedcontrolplane/v2/oauth_apiserver/component_test.go
```diff
@@ -135,3 +135,12 @@ func TestPredicate(t *testing.T) {
 		})
 	}
 }
+
+func TestNewComponent(t *testing.T) {
+	t.Run("When creating component it should build successfully", func(t *testing.T) {
```
enxebre, line 140: how's this change related to this PR?

control-plane-operator/controllers/hostedcontrolplane/v2/olm/catalog_operator/component_test.go
YamunadeviShanmugam: @enxebre : It was added to meet the unit test coverage when I added fix for other components which is later removed as part of discussion, left it as it is increasing unit test coverage

control-plane-operator/controllers/hostedcontrolplane/v2/olm/olm_operator/component_test.go
YamunadeviShanmugam: It was added to meet the unit test coverage when I added fix for other components which is later removed as part of discussion, left it as it is increasing unit test coverage

control-plane-operator/controllers/hostedcontrolplane/v2/oauth_apiserver/component_test.go
YamunadeviShanmugam: It was added to meet the unit test coverage when I added fix for other components which is later removed as part of discussion, left it as it is increasing unit test coverage

control-plane-operator/controllers/hostedcontrolplane/v2/oauth_apiserver/component.go
enxebre: which dataplane webhooks are you referring to?
Also how does agent dep helps here anyway providing this proxy needs to dials to the server?

control-plane-operator/controllers/hostedcontrolplane/v2/cno/component_test.go
```diff
@@ -69,3 +69,12 @@ func TestPlatformHasCloudNetworkConfigController(t *testing.T) {
 		})
 	}
 }
+
+func TestNewComponent(t *testing.T) {
+	t.Run("When creating component it should build successfully", func(t *testing.T) {
+		g := NewWithT(t)
```
enxebre: how's this change related to this PR?

control-plane-operator/controllers/hostedcontrolplane/v2/ingressoperator/component_test.go
```diff
@@ -80,3 +80,12 @@ func TestIsIngressCapabilityEnabled(t *testing.T) {
 		})
 	}
 }
+
+func TestNewComponent(t *testing.T) {
+	t.Run("When creating component it should build successfully", func(t *testing.T) {
+		g := NewWithT(t)
+		component := NewComponent()
+		g.Expect(component).ToNot(BeNil())
```
enxebre: how's this change related to this PR?

control-plane-operator/controllers/hostedcontrolplane/v2/oapi/component_test.go
```diff
@@ -0,0 +1,16 @@
...
+
+import (
+	"testing"
+
+	. "github.com/onsi/gomega"
+)
+
+func TestNewComponent(t *testing.T) {
+	t.Run("When creating component it should build successfully", func(t *testing.T) {
+		g := NewWithT(t)
```
enxebre: how's this change related to this PR?

konnectivity-socks5-proxy/guard.go
```diff
@@ -0,0 +1,266 @@
...
+	tryBootstrapKonnectivityFn = tryBootstrapKonnectivity
+	getConfigFn                = ctrl.GetConfig
+	newClientFn                = client.New
+)
+
+// runWithCoreGuard keeps the process alive while konnectivity infrastructure becomes
+// reachable, retrying transient failures with exponential backoff instead of exiting.
+func runWithCoreGuard(ctx context.Context, log logr.Logger, opts konnectivityproxy.Options, servingPort uint32, serve func(dialer konnectivityproxy.ProxyDialer) error) error {
+	for {
+		delay := coreGuardBackoff.waitBackoff().DelayFunc()
```
enxebre: this refreshes the delay time every time serve() fails effectively not implementing backoff, is that intended?

enxebre:
is there any evidence on what produced the "The context deadline exceeded" mentioned in the original ticket, and why that "The context deadline exceeded" result in a container restart? So it's clear how this additional complexity solve that provlem?

control-plane-operator/controllers/hostedcontrolplane/v2/oauth_apiserver/component_test.go
enxebre: please remove anything that's not related to the problem this PR is trying to solve

control-plane-operator/controllers/hostedcontrolplane/v2/olm/catalog_operator/component_test.go
YamunadeviShanmugam: Removed

control-plane-operator/controllers/hostedcontrolplane/v2/olm/olm_operator/component_test.go
YamunadeviShanmugam: Removed

control-plane-operator/controllers/hostedcontrolplane/v2/oauth_apiserver/component_test.go
YamunadeviShanmugam: addressed, files not related to the PR are removed

control-plane-operator/controllers/hostedcontrolplane/v2/cno/component_test.go
YamunadeviShanmugam: Addressed

control-plane-operator/controllers/hostedcontrolplane/v2/ingressoperator/component_test.go
YamunadeviShanmugam: Addressed

control-plane-operator/controllers/hostedcontrolplane/v2/oapi/component_test.go
YamunadeviShanmugam: addrssed

konnectivity-socks5-proxy/guard.go
YamunadeviShanmugam: Addressed

YamunadeviShanmugam:
> is there any evidence on what produced the "The context deadline exceeded" mentioned in the original ticket, and why that "The context deadline exceeded" result in a container restart? So it's clear how this additional complexity solve that provlem?


The container failed with Error: context deadline exceeded ~2 minutes after pod scheduling, before the konnectivity server was ready. This happened during the initial bootstrap phase when the container was trying to connect to the konnectivity server that didn't exist yet. As the old code had no retry logic, any error during startup or serving caused the process to exit - When the process exits, CRI-O/kubelet sees it as a crash and restarts the container. 

With the fix the container will not crash and wait for the konnectivity server to be ready(not crashing on any transient errors)

YamunadeviShanmugam:
@enxebre @jparrill : Addressed review comments, mandatory checks are passing, e2e-aws sample pipeline is passing, PTAL

clebs:
@YamunadeviShanmugam thanks for your PR! Overall it looks good, but it is hard to get an overview.

Could you rebase your branch instead of adding a merge commit on top and squash your 2 commits with the same message?
Thanks.

YamunadeviShanmugam:
@clebs : Squashed all commits to single commit, thanks for looking into this!

clebs:
Is this comment from @enxebre resolved? https://github.com/openshift/hypershift/pull/8579/changes#r3504381520

control-plane-operator/controllers/hostedcontrolplane/v2/oauth_apiserver/component.go
YamunadeviShanmugam: The dataplane webhooks being referred to are the ValidatingAdmissionWebhook and MutatingAdmissionWebhook registrations that oauth-apiserver creates. These webhooks are registered with the dataplane apiserver, and when the dataplane apiserver tries to call them, it needs a network path back to the control plane where oauth-apiserver runs.
 The konnectivity-agent dependency helps because:
  1. Webhook registration: oauth-apiserver registers admission webhooks with the dataplane apiserver
  2. Network requirement: When the dataplane apiserver tries to invoke these webhooks during request processing, it needs network connectivity back to the control plane
  3. Dependency ensures ordering: By declaring WithDependencies(konnectivity-agent), we ensure konnectivity-agent (and its socks5 proxy) is running before oauth-apiserver is marked as ready
  4. Webhook invocations work: With the proxy up, the dataplane apiserver can route webhook calls through the konnectivity tunnel back to oauth-apiserver
 Without this dependency, oauth-apiserver could be marked ready before the proxy is up, and then webhook invocations from the dataplane would fail, causing dataplane apiserver operations to hang or error.

YamunadeviShanmugam:
> Is this comment from @enxebre resolved? https://github.com/openshift/hypershift/pull/8579/changes#r3504381520

Yes @clebs

YamunadeviShanmugam:
@clebs @jparrill @cblecker : PTAL , PR has gone a rebase with main.

cblecker:
@coderabbit resume

PR #8584, opened 2026-05-25
CNTRLPLANE-3277: Add Azure OAuth LoadBalancer private topology e2e test

Nirshal:

Adds the `AzureOAuthLoadBalancerPrivateTest` to the v2 Ginkgo e2e suite, validating
that the OAuth server works correctly when published via LoadBalancer on an Azure
Private topology cluster.

The test verifies:
- The `oauth-openshift` Service is created as type LoadBalancer with an allocated endpoint
- The Service carries the Azure internal LoadBalancer annotation (confirming it's an ILB, not public)
- The OAuth token flow (kubeadmin login + htpasswd IDP setup + token request) works through the private endpoint

Changes to `lifecycle/azure.go`:
- Adds a dedicated `oauth-lb-private` ClusterSpec combining `--endpoint-access=Private`
  with `--oauth-publishing-strategy=LoadBalancer`
- Restores the `private` ClusterSpec to match main (no OAuth LoadBalancer override),
  resolving the conflict with `AzureEndpointAccessTransitionTest` (PR #8718) which
  expects Route-based OAuth on the private cluster
- Removes the `private-and-oauth` SequentialGroup — with separate clusters, there is
  no shared-state ordering concern (the sequential group was the original compromise
  per CB-2 review feedback to avoid an extra cluster; Bryan's endpoint access transition
  test makes the dedicated cluster necessary)
- Moves `private` and `oauth-lb-private` TestGroups to Parallel

**Note:** The OAuth token flow test requires the test runner to have network connectivity
to the Azure VNet (e.g., running in the management cluster or via VPN/peering).

Fixes https://redhat.atlassian.net/browse/CNTRLPLANE-3277

- **Depends on #9194** ([CNTRLPLANE-3978](https://redhat.atlassian.net/browse/CNTRLPLANE-3978)): the CPO fix for the AzurePrivateLinkService finalizer race condition. Without it, deleting the `oauth-lb-private` cluster (which creates 2 AzurePrivateLinkService CRs) orphans Azure PE resources and blocks `destroy-guests`. This PR must be merged after #9194.
- This builds on top of #8527 ([CNTRLPLANE-3222](https://redhat.atlassian.net/browse/CNTRLPLANE-3222)) which ported the v1 lifecycle tests to v2 and established the Azure self-managed test infrastructure.
- The original approach shared the `private` cluster variant by adding `--oauth-publishing-strategy=LoadBalancer` at creation time and using a SequentialGroup. This conflicted with `AzureEndpointAccessTransitionTest` (#8718) which expects Route-based OAuth on the private cluster. Since `ServicePublishingStrategy` is immutable after cluster creation (CEL validation), the dedicated `oauth-lb-private` cluster is the only viable solution.
- `AzureOAuthLoadBalancerPrivateTest` mutates the HostedCluster (creates htpasswd IDP, triggers kubeadmin secret deletion). Cleanup is handled via `t.Cleanup` (not `DeferCleanup`, since `oauth.go` doesn't import Ginkgo and has a v1 caller).
- No companion PR in `openshift/release` is needed — the v2 framework `create-guests`/`run-tests` binaries read `ClusterSpecs` and `TestMatrix` directly from `lifecycle/azure.go`.

## Checklist:
- [ ] This change includes unit tests. No unit tests needed — changes are e2e test code and CI cluster configuration only, with no new business logic.

## Summary by CodeRabbit

- **Tests**
  - Expanded Azure end-to-end coverage for private clusters using OAuth with internal LoadBalancers.
  - Added validation for OAuth and guest API access through secure port-forward connections.
  - Added coverage for the combined private endpoint and OAuth LoadBalancer configuration.
  - Added checks for expected initial node counts across Azure cluster variants.
  - Improved test reliability by separating LoadBalancer endpoint allocation from readiness checks and restoring authentication configuration after validation.

### DIFF
--- a/test/e2e/util/oauth.go
+++ b/test/e2e/util/oauth.go
@@ -8,6 +8,7 @@ import (
 	"net"
 	"net/http"
 	"net/url"
+	"sync"
 	"testing"
 	"time"
 
@@ -577,15 +578,17 @@ func EnsureOAuthWithIdentityProviderViaLoadBalancer(t *testing.T, ctx context.Co
 // podPortForward holds the state of an active port-forward to a pod.
 type podPortForward struct {
 	stopCh    chan struct{}
+	stop      func()
 	errCh     <-chan error
 	localPort uint16
 	podName   string
 }
 
 // establishPodPortForward finds a running pod matching the given labels and creates a
-// SPDY port-forward tunnel to port 6443. It returns a podPortForward whose stopCh the
-// caller must close (typically via t.Cleanup). If the pod restarts during the caller's
-// healthcheck, the caller can detect it via errCh and call this function again.
+// SPDY port-forward tunnel to port 6443. Cleanup is registered via t.Cleanup using an
+// idempotent stop function. If the pod restarts during the caller's healthcheck, the
+// caller can detect it via errCh and call this function again, using pf.stop() to close
+// the previous tunnel.
 func establishPodPortForward(
 	t testing.TB, ctx context.Context,
 	mgmtClient crclient.Client,
@@ -634,14 +637,18 @@ func establishPodPortForward(
 	fw, err := portforward.New(dialer, []string{"0:6443"}, stopCh, readyCh, io.Discard, io.Discard)
 	g.Expect(err).ToNot(HaveOccurred(), "failed to create port forwarder")
 
+	stopOnce := sync.Once{}
+	stop := func() { stopOnce.Do(func() { close(stopCh) }) }
+	t.Cleanup(stop)
+
 	errCh := make(chan error, 1)
 	go func() { errCh <- fw.ForwardPorts() }()
 	select {
 	case <-readyCh:
 	case err = <-errCh:
-		g.Expect(err).ToNot(HaveOccurred(), "%s port-forward failed to start", componentName)
+		t.Fatalf("%s port-forward exited before becoming ready: %v", componentName, err)
 	case <-ctx.Done():
-		close(stopCh)
+		stop()
 		t.Fatalf("context canceled while waiting for %s port-forward readiness: %v", componentName, ctx.Err())
 	}
 
@@ -652,6 +659,7 @@ func establishPodPortForward(
 
 	return &podPortForward{
 		stopCh:    stopCh,
+		stop:      stop,
 		errCh:     errCh,
 		localPort: ports[0].Local,
 		podName:   pod.Name,
@@ -685,7 +693,6 @@ func SetupOAuthPortForwardTransport(
 	podLabels := crclient.MatchingLabels{hyperv1.ControlPlaneComponentLabel: "oauth-openshift"}
 
 	pf := establishPodPortForward(t, ctx, mgmtClient, kubeClient, mgmtConfig, hcpNamespace, podLabels, "oauth-openshift")
-	t.Cleanup(func() { close(pf.stopCh) })
 
 	guestConfig, err := guestRestConfig(t, ctx, mgmtClient, hostedCluster)
 	g.Expect(err).ToNot(HaveOccurred(), "failed to get guest REST config")
@@ -709,9 +716,8 @@ func SetupOAuthPortForwardTransport(
 		select {
 		case pfErr := <-pf.errCh:
 			t.Logf("OAuth port-forward to %s broke (%v), re-establishing", pf.podName, pfErr)
-			oldStop := pf.stopCh
+			pf.stop()
 			pf = establishPodPortForward(t, ctx, mgmtClient, kubeClient, mgmtConfig, hcpNamespace, podLabels, "oauth-openshift")
-			close(oldStop)
 			localAddr = fmt.Sprintf("localhost:%d", pf.localPort)
 			oauthTransport.CloseIdleConnections()
 			return false, nil
@@ -766,7 +772,6 @@ func SetupGuestKASPortForwardConfig(
 	podLabels := crclient.MatchingLabels{"app": "kube-apiserver", hyperv1.ControlPlaneComponentLabel: "kube-apiserver"}
 
 	pf := establishPodPortForward(t, ctx, mgmtClient, kubeClient, mgmtConfig, hcpNamespace, podLabels, "kube-apiserver")
-	t.Cleanup(func() { close(pf.stopCh) })
 
 	guestConfig, err := guestRestConfig(t, ctx, mgmtClient, hostedCluster)
 	g.Expect(err).ToNot(HaveOccurred(), "failed to get guest REST config")
@@ -786,9 +791,8 @@ func SetupGuestKASPortForwardConfig(
 		select {
 		case pfErr := <-pf.errCh:
 			t.Logf("KAS port-forward to %s broke (%v), re-establishing", pf.podName, pfErr)
-			oldStop := pf.stopCh
+			pf.stop()
 			pf = establishPodPortForward(t, ctx, mgmtClient, kubeClient, mgmtConfig, hcpNamespace, podLabels, "kube-apiserver")
-			close(oldStop)
 			guestConfig.Host = fmt.Sprintf("https://localhost:%d", pf.localPort)
 			testClient, err = kubernetes.NewForConfig(guestConfig)
 			g.Expect(err).ToNot(HaveOccurred(), "failed to create test client after KAS port-forward re-establishment")
--- a/test/e2e/util/oauth.go
+++ b/test/e2e/util/oauth.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"encoding/base64"
 	"fmt"
+	"io"
+	"net"
 	"net/http"
 	"net/url"
 	"testing"
@@ -29,11 +31,46 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
 	restclient "k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/tools/portforward"
+	"k8s.io/client-go/transport/spdy"
 
 	crclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
+
+type oauthTokenConfig struct {
+	transport        http.RoundTripper
+	transportFactory func() http.RoundTripper
+	guestConfig      *restclient.Config
+}
+
+// OAuthTokenOption configures how OAuth token requests are performed.
+type OAuthTokenOption func(*oauthTokenConfig)
+
+// WithTransport overrides the default HTTP transport used for OAuth requests.
+// Use this with SetupOAuthPortForwardTransport to route requests through a
+// port-forward tunnel to the OAuth server in private topology clusters.
+func WithTransport(rt http.RoundTripper) OAuthTokenOption {
+	return func(c *oauthTokenConfig) { c.transport = rt }
+}
+
+// WithGuestConfig overrides the guest cluster REST config used for API calls
+// (e.g. GetUserForToken) in ValidateOAuthIdentityProviderFlow. Use this with
+// SetupGuestKASPortForwardConfig to route guest KAS calls through a port-forward
+// tunnel in private topology clusters where the KAS is behind an ILB.
+func WithGuestConfig(cfg *restclient.Config) OAuthTokenOption {
+	return func(c *oauthTokenConfig) { c.guestConfig = cfg }
+}
+
+// WithTransportFactory provides a function that creates a fresh HTTP transport
+// by establishing a new port-forward to the current OAuth pod. This is called
+// after IDP config changes trigger an OAuth server rollout, because the old
+// port-forward dies with the replaced pod.
+func WithTransportFactory(f func() http.RoundTripper) OAuthTokenOption {
+	return func(c *oauthTokenConfig) { c.transportFactory = f }
+}
 
 func EnsureOAuthWithIdentityProvider(t *testing.T, ctx context.Context, client crclient.Client, hostedCluster *hyperv1.HostedCluster) {
 	t.Run("EnsureOAuthWithIdentityProvider", func(t *testing.T) {
@@ -111,8 +148,13 @@ func WaitForOAuthToken(t *testing.T, ctx context.Context, oauthRoute *routev1.Ro
 
 // WaitForOAuthTokenByHost performs the OAuth token request flow against the given host.
 // This supports both Route-based and LoadBalancer-based OAuth endpoints.
-func WaitForOAuthTokenByHost(t testing.TB, ctx context.Context, oauthHost string, restConfig *restclient.Config, username, password string) string {
+func WaitForOAuthTokenByHost(t testing.TB, ctx context.Context, oauthHost string, restConfig *restclient.Config, username, password string, opts ...OAuthTokenOption) string {
 	g := NewWithT(t)
+
+	cfg := &oauthTokenConfig{}
+	for _, o := range opts {
+		o(cfg)
+	}
 
 	oauthClient := configmanifests.OAuthServerChallengingClient().Name
 	tokenReqUrl := fmt.Sprintf("https://%s/oauth/authorize?response_type=token&client_id=%s", oauthHost, oauthClient)
@@ -122,8 +164,13 @@ func WaitForOAuthTokenByHost(t testing.TB, ctx context.Context, oauthHost string
 	request.Header.Set("Authorization", getBasicHeader(username, password))
 	request.Header.Set("X-CSRF-Token", "1")
 
-	transport, err := restclient.TransportFor(restclient.AnonymousClientConfig(restConfig))
-	g.Expect(err).ToNot(HaveOccurred(), "error getting transport")
+	var transport http.RoundTripper
+	if cfg.transport != nil {
+		transport = cfg.transport
+	} else {
+		transport, err = restclient.TransportFor(restclient.AnonymousClientConfig(restConfig))
+		g.Expect(err).ToNot(HaveOccurred(), "error getting transport")
+	}
 
 	httpClient := &http.Client{
 		Transport: transport,
@@ -306,16 +353,15 @@ func validateClusterPreIDP(t *testing.T, ctx context.Context, client crclient.Cl
 
 }
 
-// WaitForOAuthLoadBalancerReady waits for the oauth-openshift LoadBalancer Service to have an
-// external endpoint allocated and for the /healthz endpoint to return HTTP 200.
-// It returns the OAuth hostname from the HostedCluster's service publishing strategy,
-// which matches the TLS certificate SANs, rather than the raw LoadBalancer IP.
-func WaitForOAuthLoadBalancerReady(t testing.TB, ctx context.Context, client crclient.Client, restConfig *restclient.Config, hostedCluster *hyperv1.HostedCluster) string {
+// WaitForOAuthLoadBalancerEndpoint waits for the oauth-openshift LoadBalancer Service to have an
+// endpoint allocated and returns the OAuth hostname from the HostedCluster's service publishing
+// strategy. This hostname matches the TLS certificate SANs and is the one ExternalDNS creates
+// a DNS record for. Unlike WaitForOAuthLoadBalancerReady, this does not perform a direct
+// /healthz check, making it suitable for private topology clusters where the LoadBalancer
+// endpoint is not directly reachable from the test runner.
+func WaitForOAuthLoadBalancerEndpoint(t testing.TB, ctx context.Context, client crclient.Client, hostedCluster *hyperv1.HostedCluster) string {
 	g := NewWithT(t)
 
-	// Get the OAuth hostname from the HostedCluster's service publishing strategy.
-	// This is the hostname that the TLS certificate is issued for and that ExternalDNS
-	// creates a DNS record for, so it must be used for TLS connections.
 	oauthStrategy := netutil.ServicePublishingStrategyByTypeByHC(hostedCluster, hyperv1.OAuthServer)
 	g.Expect(oauthStrategy).ToNot(BeNil(), "OAuth service publishing strategy not found in HostedCluster spec")
 	g.Expect(oauthStrategy.LoadBalancer).ToNot(BeNil(), "OAuth LoadBalancer strategy not found")
@@ -323,7 +369,6 @@ func WaitForOAuthLoadBalancerReady(t testing.TB, ctx context.Context, client crc
 	g.Expect(oauthHost).ToNot(BeEmpty(), "OAuth LoadBalancer hostname is empty")
 	t.Logf("OAuth hostname from HostedCluster spec: %s", oauthHost)
 
-	// Wait for the LoadBalancer to get an external endpoint (confirms LB is provisioned)
 	hcpNamespace := manifests.HostedControlPlaneNamespace(hostedCluster.Namespace, hostedCluster.Name)
 	svc := hcpmanifests.OauthServerService(hcpNamespace)
 	err := wait.PollUntilContextTimeout(ctx, 5*time.Second, 10*time.Minute, true, func(ctx context.Context) (done bool, err error) {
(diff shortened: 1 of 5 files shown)


Nirshal:
**Re: CodeRabbit suggestion to change `BeforeAll` to `BeforeEach`** — intentionally keeping `BeforeAll` here.

This `Context` uses the `Ordered` decorator with shared state (`testCtx`, `hc`, `controlPlaneNamespace`) across three sequential `It` blocks. `BeforeAll` runs once and initializes the shared variables; `BeforeEach` would redundantly re-fetch the test context and re-run the guards before every `It`.

This matches the existing pattern in `AzurePrivateTopologyTest` (same file, line 91) which also uses `Ordered` + `BeforeAll` for the same reason.

test/e2e/v2/tests/hosted_cluster_azure_test.go
```diff
@@ -286,11 +286,109 @@ func AzureOAuthLoadBalancerTest(getTestCtx internal.TestContextGetter) {
...
+					},
+				},
+				e2eutil.WithTimeout(10*time.Minute),
+			)
+		})
+
+		It("should complete OAuth token flow through LoadBalancer endpoint", func() {
+			ctx := testCtx.Context
+			e2eutil.ValidateOAuthWithIdentityProviderViaLoadBalancer(GinkgoTB(), ctx, testCtx.MgmtClient, hc)
+		})
```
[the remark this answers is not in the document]
Nirshal: The mutation happens inside the shared helper `ValidateOAuthWithIdentityProviderViaLoadBalancer` (in `test/e2e/util/oauth.go`), which is also used by the existing `AzureOAuthLoadBalancerTest` (line 284, same file) without cleanup. Adding cleanup in the caller would mean duplicating restore logic that belongs in the helper itself.

This is a pre-existing concern, not introduced by this PR. Noted in the PR description as a known tradeoff. If we address it, it should be done in the shared helper as a follow-up.

cblecker, line 756: The CI run showed this test can't pass as-is — `WaitForOAuthLoadBalancerReady` (in `test/e2e/util/oauth.go:350-375`) does a direct TLS dial to `https://<oauth-hostname>/healthz`. For private topology that hostname resolves to the Azure internal LB IP (e.g., `10.0.0.22`), which is unreachable from the CI build cluster.

Interestingly the OAuth token flow itself _succeeds_ because it routes through the guest kubeconfig transport — it's specifically the raw healthcheck that fails. So the underlying validation logic works; it's just the connectivity check that's incompatible with private topology.

The fix needs to happen in the shared helper. A few approaches:
1. Skip `WaitForOAuthLoadBalancerReady` entirely for private topology and rely on the token flow validation alone — it already exercises the full OAuth stack end-to-end.
2. Use the guest kubeconfig transport for the healthcheck too, the same way the token flow step does.
3. Route it through a pod on the management cluster that has VNet connectivity.

This needs to be addressed in the helper before this can land.

test/e2e/v2/tests/hosted_cluster_azure_test.go
```diff
@@ -286,11 +286,109 @@ func AzureOAuthLoadBalancerTest(getTestCtx internal.TestContextGetter) {
...
+
 // RegisterHostedClusterAzureTests registers all Azure-specific hosted cluster tests.
 func RegisterHostedClusterAzureTests(getTestCtx internal.TestContextGetter) {
 	AzurePublicClusterTest(getTestCtx)
 	AzurePrivateTopologyTest(getTestCtx)
 	AzureOAuthLoadBalancerTest(getTestCtx)
+	AzureOAuthLoadBalancerPrivateTest(getTestCtx)
 }
 
 var _ = Describe("Hosted Cluster Azure", Label("hosted-cluster-azure"), func() {
```
cblecker: The public variant (`AzureOAuthLoadBalancerTest`) works cleanly because it runs on a dedicated cluster (`cluster-name-oauth-lb`) created specifically for it — mutation is safe there since nothing else runs on that cluster. This private variant is trying to replicate that pattern but without the dedicated cluster, sharing `cluster-name-private` with the read-only `AzurePrivateTopologyTest` instead.

Registering this last is a reasonable mitigation, but declaration order between `Context` blocks isn't enforced by Ginkgo's `Ordered` decorator — that only applies within a single `Context`. If the registration order changes, or a new test gets inserted between the two, the read-only private topology tests could run after the HC has been mutated (htpasswd IDP created, kubeadmin secret deleted).

The cleanest fix that matches how the public variant is structured would be a dedicated `cluster-name-oauth-lb-private` cluster. I understand the reviewer feedback to reduce cluster count, so if a dedicated cluster is off the table, the other option is to make this a sequential step in the test matrix (after the read-only private tests), the way `etcd-chaos` runs sequentially after `upgrade` on the upgrade cluster. That would give the same isolation guarantee without an extra cluster.

cblecker (changes requested):
The test structure itself is solid — good use of `Ordered` + `BeforeAll`, thorough skip guards, and the internal LB annotation check is a meaningful addition for private topology coverage.

Two things to address before this can merge:

**1. Healthcheck unreachable from CI (blocking):** `WaitForOAuthLoadBalancerReady` dials the OAuth hostname directly via TLS, which resolves to the internal LB IP for private topology — unreachable from the build cluster. The CI run confirmed this (job 2059278516129632256, `oauth.go:375`, 300s timeout). The token flow step worked; it's only the raw healthcheck that fails. Fix needs to happen in the shared helper.

**2. Test isolation on the shared cluster:** The public `AzureOAuthLoadBalancerTest` works because it has a dedicated cluster. This private variant shares a cluster with read-only tests and relies on declaration order for isolation. Either a dedicated cluster, or a sequential matrix step, would give the same guarantee the public variant has.

Minor nit: the Service type + endpoint predicates in the first `It` block are nearly identical to the public variant (~30 lines). Not blocking, but worth a shared helper if more LB variants get added.

test/e2e/v2/tests/hosted_cluster_azure_test.go
Nirshal: Addressed in c2740440d1. Extracted `WaitForOAuthLoadBalancerEndpoint` from `WaitForOAuthLoadBalancerReady` — it waits for the LB endpoint without the direct `/healthz` TLS dial. The private test now calls `WaitForOAuthLoadBalancerEndpoint` + `ValidateOAuthIdentityProviderFlow` (also extracted), skipping the health check entirely while still validating the full OAuth IDP token flow end-to-end.

`WaitForOAuthLoadBalancerReady` is unchanged in behavior — it now composes `WaitForOAuthLoadBalancerEndpoint` + the health check, so the public test path is unaffected.

test/e2e/v2/tests/hosted_cluster_azure_test.go
Nirshal: Addressed in c2740440d1. Moved the `private` test group from `Parallel` to `Sequential` in the test matrix, following the same pattern as `upgrade-and-chaos`. The new `private-and-oauth` sequential group runs read-only private topology tests first, then the mutating OAuth LB test second — guaranteeing isolation regardless of registration order.

Nirshal:
CI failure on `e2e-azure-v2-self-managed` was an infrastructure issue — the `hypershift-install` step timed out after ~5m due to API server rate limiting (`client rate limiter Wait returned an error: context deadline exceeded`). The HyperShift operator never deployed, so no test clusters were created and none of our test code ran. Retriggering.

/test e2e-azure-v2-self-managed

cblecker (commented):
Both previous blocking items are addressed — thank you for the updates! One minor annotation fix needed for CI tooling.

test/e2e/v2/tests/hosted_cluster_azure_test.go
```diff
@@ -286,11 +286,110 @@ func AzureOAuthLoadBalancerTest(getTestCtx internal.TestContextGetter) {
...
+// publishing validation in a private topology cluster.
+// These tests verify that:
+//   - The oauth-openshift Service is created as a LoadBalancer with an allocated endpoint
+//   - The Service carries the Azure internal LoadBalancer annotation
+//   - The OAuth token flow (kubeadmin + htpasswd IDP) works through that endpoint
+//
+// The OAuth token flow test requires the test runner to have network connectivity
+// to the Azure VNet (e.g., running in the management cluster or via VPN/peering).
+func AzureOAuthLoadBalancerPrivateTest(getTestCtx internal.TestContextGetter) {
+	Context("Azure OAuth LoadBalancer in Private Topology", Label("Azure", "self-managed-azure-oauth-lb-private"), Ordered, func() {
```
cblecker: The Context block is missing the `[Feature:AzureOAuth]` annotation that the other tests in this file have (e.g., line 230). Per the v2 AGENTS.md section 19, this is needed for Sippy Component Readiness mapping. Should be:
```go
Context("[Feature:AzureOAuth] Azure OAuth LoadBalancer in Private Topology", Label("Azure", "self-managed-azure-oauth-lb-private"), Ordered, func() {
```

cblecker (commented):
nit: missing `[Feature:AzureOAuth]` annotation for Sippy mapping

test/e2e/v2/tests/hosted_cluster_azure_test.go
Nirshal: I totally missed that, thanks a lot for finding it!

test/e2e/util/oauth.go
```diff
@@ -378,16 +386,16 @@ func WaitForOAuthLoadBalancerReady(t testing.TB, ctx context.Context, client crc
 	return oauthHost
 }
 
-func ValidateOAuthWithIdentityProviderViaLoadBalancer(t testing.TB, ctx context.Context, client crclient.Client, hostedCluster *hyperv1.HostedCluster) {
+// ValidateOAuthIdentityProviderFlow validates the full OAuth identity provider flow against
+// a LoadBalancer endpoint: kubeadmin login, htpasswd IDP setup, testuser login, and
+// kubeadmin secret removal. It takes the oauthHost to use for token requests, allowing
+// callers to control whether the host was obtained via WaitForOAuthLoadBalancerReady
+// (with health check) or WaitForOAuthLoadBalancerEndpoint (without).
```
cblecker: nit: The doc comment says "validates the full OAuth identity provider flow against a LoadBalancer endpoint" but the function is actually endpoint-agnostic — it just takes an `oauthHost` string and performs OAuth token requests against it. The host could come from any source. Consider:

```go
// ValidateOAuthIdentityProviderFlow validates the full OAuth identity provider flow
// against the given oauthHost: kubeadmin login, htpasswd IDP setup, testuser login,
// and kubeadmin secret removal. ...
```

test/e2e/util/oauth.go
```diff
@@ -435,20 +442,27 @@ func ValidateOAuthWithIdentityProviderViaLoadBalancer(t testing.TB, ctx context.
...
 
 	user, err = GetUserForToken(guestConfig, accessToken)
 	g.Expect(err).ToNot(HaveOccurred())
 	g.Expect(user.Name).To(Equal("testuser"))
 
-	// Validate kubeadmin secret was removed after IDP was added
 	validateClusterPostIDP(t, ctx, client, hostedCluster)
 }
 
+func ValidateOAuthWithIdentityProviderViaLoadBalancer(t testing.TB, ctx context.Context, client crclient.Client, hostedCluster *hyperv1.HostedCluster) {
```
cblecker, line 559: Two things on this wrapper:

1. **Missing doc comment**: After the refactor, this exported function lost its documentation. A brief comment like `// ValidateOAuthWithIdentityProviderViaLoadBalancer combines WaitForOAuthLoadBalancerReady (with /healthz check) and ValidateOAuthIdentityProviderFlow. For private topology clusters where /healthz is unreachable, call those functions separately.` would help callers choose between this and the lower-level functions.

2. **Double `guestRestConfig` call**: This function calls `guestRestConfig` at line 459 to pass to `WaitForOAuthLoadBalancerReady`, then `ValidateOAuthIdentityProviderFlow` calls `guestRestConfig` again internally at line 396. Minor inefficiency — could pass it through or restructure slightly. Not a blocker.

test/e2e/v2/tests/hosted_cluster_azure_test.go
```diff
@@ -286,11 +286,110 @@ func AzureOAuthLoadBalancerTest(getTestCtx internal.TestContextGetter) {
...
+						}
+						host := ingress.IP
+						if host == "" {
+							host = ingress.Hostname
+						}
+						return true, fmt.Sprintf("oauth-openshift LoadBalancer has an allocated endpoint: %s", host), nil
+					},
+				},
+				e2eutil.WithTimeout(10*time.Minute),
+			)
```
cblecker, line 721: nit: The two predicates here (Service type check + LB endpoint check) are nearly identical to the ones in `AzureOAuthLoadBalancerTest` (lines 254-275). The only textual difference is the success message string. Consider extracting a shared `oauthServiceLBPredicates()` helper that returns `[]e2eutil.Predicate[*corev1.Service]` — both tests could use it, reducing maintenance burden. Not blocking, just a refactoring opportunity.

cblecker (commented):
The refactoring looks good overall — clean separation of `WaitForOAuthLoadBalancerEndpoint` from `WaitForOAuthLoadBalancerReady`, and the `SequentialGroup` restructuring properly addresses the shared-cluster ordering concern from the previous review round.

A few inline suggestions on docs and duplication, plus one that couldn't be placed inline:

**`oauth.go:336,343` — Log messages say "external endpoint" but function now serves private topology too**

Since `WaitForOAuthLoadBalancerEndpoint` is designed for both public and private clusters (the doc comment explicitly says so), the log messages referencing "external endpoint" are misleading for internal LBs. Also, line 343's format string `"%s%s"` concatenates `ingress.IP` and `ingress.Hostname` without a separator — in practice only one is populated, but the conditional logic from the test file predicates would be cleaner here too.

test/e2e/util/oauth.go
Nirshal: Fixed in c2be1dc — rephrased the doc comment to reflect the endpoint-agnostic design, as you suggested. Also added a note that the function mutates cluster state (htpasswd Secret, OAuth config, kubeadmin removal) so callers know it belongs in lifecycle tests only.

test/e2e/util/oauth.go
Nirshal: Fixed in c2be1dc:

1. **Doc comment**: Added — describes the composition of `WaitForOAuthLoadBalancerReady` + `ValidateOAuthIdentityProviderFlow` and notes that private topology callers should use the lower-level functions directly.

2. **Double `guestRestConfig`**: Leaving as-is to avoid changing the pre-existing signature of `ValidateOAuthIdentityProviderFlow`. Threading `restConfig` through would change the contract for all callers. Agreed it is minor.

test/e2e/v2/tests/hosted_cluster_azure_test.go
Nirshal: Fixed in c2be1dc — extracted `oauthServiceLBPredicates()` as you suggested. Both the public and private tests now use the shared helper. The duplication was introduced by us (blame confirmed it), so good catch.

Nirshal:
Re: the review feedback on `oauth.go:336,343` (log messages saying "external endpoint" + fragile `"%s%s"` format):

Fixed in c2be1dc — removed "external" from log messages since `WaitForOAuthLoadBalancerEndpoint` now serves both public and private topologies. Replaced the `"%s%s"` concatenation with explicit `(IP=%s, Hostname=%s)` format using `<none>` for empty fields, following the `"<none>"` placeholder pattern used elsewhere in the codebase (e.g., `create-guests/main.go`, k8s apiserver audit).

test/e2e/util/oauth.go
```diff
@@ -378,16 +394,18 @@ func WaitForOAuthLoadBalancerReady(t testing.TB, ctx context.Context, client crc
...
 }
 
-func ValidateOAuthWithIdentityProviderViaLoadBalancer(t testing.TB, ctx context.Context, client crclient.Client, hostedCluster *hyperv1.HostedCluster) {
+// ValidateOAuthIdentityProviderFlow validates the full OAuth identity provider flow
+// against the given oauthHost: kubeadmin login, htpasswd IDP setup, testuser login,
+// and kubeadmin secret removal. The oauthHost can come from any source, such as
+// WaitForOAuthLoadBalancerReady (with health check) or WaitForOAuthLoadBalancerEndpoint
+// (endpoint only, suitable for private topology).
+// This function mutates cluster state (creates htpasswd Secret, patches OAuth config,
+// deletes kubeadmin secret) and should only be used in lifecycle tests.
```
bryan-cox: **suggestion (AGENTS.md Standard 12 — docstring accuracy):** This line says the function "deletes kubeadmin secret," but the function only *validates* that the controller removed it (via `validateClusterPostIDP`). The controller deletes it in response to the IDP being added — the function doesn't delete it directly.

Consider:
```suggestion
// validates kubeadmin secret removal) and should only be used in lifecycle tests.
```

test/e2e/util/oauth.go
```diff
@@ -399,7 +417,6 @@ func ValidateOAuthWithIdentityProviderViaLoadBalancer(t testing.TB, ctx context.
 	g.Expect(err).ToNot(HaveOccurred())
 	g.Expect(user.Name).To(Equal("kube:admin"))
 
-	// Set up htpasswd identity provider
 	secret := corev1.Secret{
```
bryan-cox, line 479: **important (AGENTS.md Standard 9 — state mutation and cleanup):** This function creates an `htpasswd` Secret here and later mutates the HostedCluster's identity provider list (via `UpdateObject`), but neither mutation has a `DeferCleanup`. AGENTS.md Standard 9 says: *"When creating a resource, register `DeferCleanup` immediately after `Create` — not later in the test flow."*

I realize this is inherited from the pre-existing `ValidateOAuthWithIdentityProviderViaLoadBalancer`, but since you're extracting a new function here, it's a good opportunity to fix it. Something like:

```go
g.Expect(client.Create(ctx, &secret)).To(Succeed())
DeferCleanup(func() {
    err := client.Delete(ctx, &secret)
    if err != nil && !apierrors.IsNotFound(err) {
        g.Expect(err).NotTo(HaveOccurred(), "cleanup: failed to delete htpasswd secret")
    }
})
```

Same for the OAuth config patch — capture the original `IdentityProviders` slice before mutation and defer its restoration.

If you'd rather keep this scoped and address cleanup separately, that's fine too — but please file a follow-up.

bryan-cox (commented):
Review of AGENTS.md compliance and correctness. Two findings — one important, one suggestion. Overall the PR is well-structured; the refactoring is clean and backward compatible.

test/e2e/util/oauth.go
Nirshal: Fixed in 9f15d56 — good catch, the function validates that the controller removed the kubeadmin secret (via `validateClusterPostIDP`), it does not delete it directly.

test/e2e/util/oauth.go
Nirshal: Agreed — we'll address this in this PR rather than deferring to a follow-up.

One implementation detail: `ValidateOAuthIdentityProviderFlow` takes `testing.TB` and `oauth.go` doesn't import Ginkgo (only Gomega). The function is also still called from the v1 test in `azure_oauth_lb_test.go:82` (via `EnsureOAuthWithIdentityProviderViaLoadBalancer`), which has no Ginkgo context. So we can't use `DeferCleanup` directly — we'll use `t.Cleanup()` instead, which is part of the `testing.TB` interface and works for both v1 (`*testing.T`) and v2 (`GinkgoTB()` maps it to Ginkgo's `DeferCleanup` under the hood).

We'll handle both mutations:
1. Secret creation → `t.Cleanup` to delete (with `IsNotFound` guard)
2. OAuth config patch → capture original before mutation, `t.Cleanup` to restore

Nirshal: Implemented in aaa0e53 — we added `t.Cleanup` for both mutations:

1. **htpasswd Secret**: `t.Cleanup` with `context.Background()` and `IsNotFound` guard
2. **OAuth config**: fresh `client.Get` to capture original state before mutation, then `t.Cleanup` with `context.WithTimeout(context.Background(), 2m)` to restore it

We used `context.Background()` in cleanup closures (instead of the caller's `ctx`) so cleanup still works if the test times out.

LIFO ordering ensures OAuth config is restored first (removing the IDP reference), then the Secret is deleted.

We're heading out on PTO — if you'd prefer a different approach (e.g. `DeferCleanup` with a Ginkgo import), happy to adjust on Monday.

bryan-cox (commented):
/approve

Nirshal:
@cblecker @bryan-cox

Heads up on a significant change in this revision. The original approach shared the `private` cluster variant by adding `--oauth-publishing-strategy=LoadBalancer` at creation time and using a `SequentialGroup` to ensure test ordering.

This conflicts with Bryan's `AzureEndpointAccessTransitionTest` introduced in #8718, which runs on the same `private` cluster and expects Route-based OAuth (it looks for the `oauth-private-external` Service). Since `ServicePublishingStrategy` is immutable after cluster creation (CEL validation on `spec.services`), sequential ordering can't resolve this: the cluster is already created with the wrong publishing strategy for Bryan's test.

After discussing with Bryan, we agreed the only viable solution is a dedicated `oauth-lb-private` cluster variant. This revision:
- Restores the `private` ClusterSpec to match main (no OAuth LoadBalancer override)
- Adds a new `oauth-lb-private` ClusterSpec combining `--endpoint-access=Private` + `--oauth-publishing-strategy=LoadBalancer`
- Removes the `private-and-oauth` SequentialGroup, since with separate clusters there's no shared-state concern
- Moves both TestGroups to Parallel

All previous review feedback has been addressed and preserved, no regressions introduced.

Nirshal:
The test was never run, there was an ImagePullBackOff on the test's container, the `quay-proxy.ci.openshift.org/openshift/ci@sha256:d2cdb6f7...` image was not found. I'll relaunch it. 
/retest

Nirshal:
/hold 
still working on it

Nirshal:
/test e2e-azure-v2-self-managed 

since I suspect last time it was a flake or it was not clear anyway, because of an HO restart that made the process hit the timeout.

test/e2e/util/oauth.go
```diff
@@ -458,6 +574,225 @@ func EnsureOAuthWithIdentityProviderViaLoadBalancer(t *testing.T, ctx context.Co
...
+	fw, err := portforward.New(dialer, []string{"0:6443"}, stopCh, readyCh, io.Discard, io.Discard)
+	g.Expect(err).ToNot(HaveOccurred(), "failed to create port forwarder")
+
+	errCh := make(chan error, 1)
+	go func() { errCh <- fw.ForwardPorts() }()
+	select {
+	case <-readyCh:
+	case err = <-errCh:
+		g.Expect(err).ToNot(HaveOccurred(), "port-forward failed to start")
+	}
```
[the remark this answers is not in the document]
Nirshal: Good catch. Moved `t.Cleanup(close(stopCh))` immediately after `portforward.New` in both `SetupOAuthPortForwardTransport` and `SetupGuestKASPortForwardConfig`, so the forwarder goroutine is always cleaned up even if a subsequent assertion aborts the helper.

Also applied the nitpick from the review summary: replaced the single `List` call in `SetupGuestKASPortForwardConfig` with polling + `DeletionTimestamp` filter, matching the pattern in `SetupOAuthPortForwardTransport`.

test/e2e/util/oauth.go
```diff
@@ -458,6 +574,230 @@ func EnsureOAuthWithIdentityProviderViaLoadBalancer(t *testing.T, ctx context.Co
...
+		close(stopCh)
+	})
+
+	errCh := make(chan error, 1)
+	go func() { errCh <- fw.ForwardPorts() }()
+	select {
+	case <-readyCh:
+	case err = <-errCh:
+		g.Expect(err).ToNot(HaveOccurred(), "port-forward failed to start")
+	}
```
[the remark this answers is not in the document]
Nirshal: Good catch. Added a `case <-ctx.Done()` branch to both `SetupOAuthPortForwardTransport` and `SetupGuestKASPortForwardConfig` select blocks, so the test fails immediately with a diagnostic message when the context is cancelled instead of blocking until the suite timeout.

Nirshal:
Regarding the nitpick about rebuilding the option list from `cfg` instead of enumerating fields (oauth.go:538-544): I'm going to skip this one. There are currently only two options (`WithTransport` and `WithGuestConfig`), and the proposed refactor using `func(c *oauthTokenConfig) { *c = resolved }` trades explicit readability for a pattern that's harder to follow. If a third option is added, it'll be easy to extend the existing block. Not worth the indirection for now.

Nirshal:
Test infrastructure failure: the job failed during the `create-management-cluster` pre-step (after 31m56s), before any test code was executed. All subsequent steps (`dump`, `destroy-guests`, `hypershift-k8sgpt`) failed with `cluster-manifests.json: no such file or directory` because the management cluster was never created. This is unrelated to the PR changes.

Nirshal:
The test failed with: Failed to download metadata for repo 'packages-microsoft-com-prod': 
repomd.xml GPG signature verification error: Bad GPG signature
I'll retrigger.

cblecker:
@coderabbit resume

test/e2e/util/oauth.go
```diff
@@ -458,6 +574,241 @@ func EnsureOAuthWithIdentityProviderViaLoadBalancer(t *testing.T, ctx context.Co
...
+	case err = <-errCh:
+		g.Expect(err).ToNot(HaveOccurred(), "%s port-forward failed to start", componentName)
+	case <-ctx.Done():
+		close(stopCh)
+		t.Fatalf("context canceled while waiting for %s port-forward readiness: %v", componentName, ctx.Err())
+	}
+
+	ports, err := fw.GetPorts()
+	g.Expect(err).ToNot(HaveOccurred(), "failed to get forwarded ports")
+	g.Expect(ports).ToNot(BeEmpty(), "no forwarded ports found")
```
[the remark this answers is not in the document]
Nirshal: Good catch. Fixed: `establishPodPortForward` now registers cleanup via `t.Cleanup` using a `sync.Once`-wrapped stop function immediately after creating `stopCh`, so early assertion failures can no longer leak the forwarder goroutine. The readiness select uses `t.Fatalf` to fail unconditionally when `errCh` fires (including nil error). Both callers drop their redundant `t.Cleanup` registrations and use the idempotent `pf.stop()` when closing the previous tunnel during re-establishment.

PR #8585, opened 2026-05-26
OCPBUGS-85351: Revert "fix(azure): prevent externalDNSDomain from shadowing cluster apps domain"

Nirshal:
## Summary

- The fix was originally introduced in https://github.com/openshift/hypershift/pull/8480
- It is being reverted in this PR because the CEL validation rejects legitimate ARO-HCP HostedCluster configurations. ARO-HCP uses service hostnames (e.g., `api-int.{name}.domain`) that share a parent domain with the cluster baseDomain (`aro.{name}.domain`), which the validation incorrectly flags as DNS zone shadowing.
- This revert unblocks ARO-HCP E2E tests that are failing after bumping to the latest HyperShift operator. The shadowing prevention will need to be re-implemented with ARO-HCP's DNS configuration pattern taken into account.

## References

- Original PR: #8480
- Breaking commit: be26321400
- Report: https://redhat-external.slack.com/archives/C075PHEFZKQ/p1779791610616139

/cc @bryan-cox @enxebre


bryan-cox:
/approve
/lgtm

This regressed ARO HCP.

enxebre:
can we please follow up to: 
- Replicate the same dns pattern in the CI azure e2e
- Include that case in the envtest suite

/approve

PR #8589, opened 2026-05-26
NO-JIRA: fix(karpenter): start node cleanup when CAPI Cluster is deleted

maxcao13:

The karpenter-operator previously only began deleting Karpenter NodePools and NodeClaims when the HCP received its DeletionTimestamp. In the HostedCluster deletion sequence, the HCP is deleted much later than regular HyperShift NodePools — after CAPI Cluster deletion, AWS endpoint service cleanup, and other steps complete. This meant Karpenter nodes were torn down sequentially after regular CAPI nodes instead of in parallel.

Watch the CAPI Cluster object on the management side and treat its DeletionTimestamp as an early deletion signal. The CAPI Cluster is deleted at the same point regular NodePools are deleted, so Karpenter cleanup now runs concurrently with CAPI machine teardown.

The karpenter finalizer on the HCP is only removed once the HCP itself has a DeletionTimestamp, preserving the existing gate for HCP deletion.

Assisted by cursor agent with claude 4.6 opus.

no-jira

Alternatively, we could have the hcp deletion flow set a special hcp annotation for the karpenter-operator to consume, but I chose to go this route to reduce internal APIs. We also don't have a standalone lifecycle object for AutoNode (whereas there is one that exists for CAPI machines -- CAPI cluster object) -- we only have the autonode field in the HCP/HC. This will speed up deletion by a good amount on autonode enabled clusters (since we no longer wait until capi object is deleted to cleanup karpenter anymore), so I'm hoping this will also solve some teardown timeout flakes.

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved deletion cleanup coordination so hosted control planes and related clusters clean up resources reliably during complex deletion sequences.

* **Chores**
  * Broadened role permissions to allow access to cluster resources required for cleanup operations.

* **Tests**
  * Added and updated tests covering additional cluster deletion and cleanup scenarios, including cases where related cluster objects are deleted independently.

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/v2/assets/karpenter-operator/role.yaml
+++ b/control-plane-operator/controllers/hostedcontrolplane/v2/assets/karpenter-operator/role.yaml
@@ -84,6 +84,14 @@ rules:
       - "patch"
       - "delete"
   - apiGroups:
+      - "cluster.x-k8s.io"
+    resources:
+      - "clusters"
+    verbs:
+      - "get"
+      - "list"
+      - "watch"
+  - apiGroups:
       - ""
     resources:
       - "pods"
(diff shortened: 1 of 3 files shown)


maxcao13:
/test e2e-aws-autonode

infra

PR #13, opened 2021-02-03
Add support for hostedCluster to delete all nodePools

enxebre:
Currently when deleting a hostedCluster it won't delete any owned nodePool causing a race between CAPI deleting the underlying machineSets and the nodePool CR wanting to recreate them.

PR #8590, opened 2026-05-26
GCP-688: Port remaining v1 TestCreateCluster gaps to v2 e2e

cblecker:
## Summary

Ports the remaining 6 v1 `TestCreateCluster` validation gaps to v2 e2e tests, following up on PR #8511 which ported 14 validations.

**Non-lifecycle tests** (read-only verification, no cluster mutation):
- **EnsureNodeCommunication** (`hosted_cluster_node_communication_test.go`): Verifies konnectivity-agent pods exist in the hosted cluster and that pod logs are retrievable, confirming the konnectivity tunnel works
- **ValidateIngressOperatorConfiguration** (`hosted_cluster_ingress_test.go`): Platform-agnostic test that validates `EndpointPublishingStrategy` configured on the HostedCluster is reflected in the hosted cluster's IngressController (replaces Azure-only version with platform-agnostic coverage, requires 4.21+)
- **VerifyCPOOverrideImage** (`hosted_cluster_cpo_test.go`): Checks that when a CPO override image is configured for the platform/version, the running control-plane-operator pod uses that image (self-skips when no override is configured)

**Lifecycle tests** (mutate cluster state with cleanup):
- **EnsureDefaultSecurityGroupTags** (`hosted_cluster_aws_test.go`): AWS-only, adds a day-2 tag to the HostedCluster spec and verifies it's applied to the default worker security group via AWS API (requires 4.20+)
- **EnsureAWSCCMWithCustomizations** (`hosted_cluster_aws_test.go`): AWS-only, validates NLBSecurityGroupMode=Managed in aws-cloud-config ConfigMap and creates a LoadBalancer service to verify managed security groups are attached to the NLB (requires 4.23+)
- **EnsureGlobalPullSecret** (`hosted_cluster_pull_secret_test.go`): Full pull secret lifecycle — creates additional pull secret, verifies propagation through GlobalPullSecret pipeline, validates DaemonSet readiness, verifies in-place management-cluster pull secret propagation without rollout, and confirms cleanup on deletion (AWS/Azure, requires 4.19+)

## Test plan

- [x] `make e2ev2` compiles successfully
- [x] `make lint` passes with 0 issues
- [x] Pre-existing unit test failures are unrelated to these changes
- [x] Run v2 suite on GCP — non-lifecycle tests pass, AWS-only tests skip
- [x] Run v2 suite on AWS — lifecycle and non-lifecycle AWS tests pass
- [ ] Run v2 suite on Azure — global pull secret lifecycle test passes

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end tests for hosted clusters: AWS security‑group tagging and load‑balancer validation, control‑plane operator image override checks, ingress operator endpoint behavior, node communication/log access verification, and global pull‑secret lifecycle scenarios.
* **Documentation**
  * Updated e2e framework guidelines: package organization, mandatory cleanup patterns, idempotent resource handling, and improved assertion and testing best practices.
* **Chores**
  * Added optional environment variable to configure an additional pull-secret file for e2e runs.

### DIFF
--- a/test/e2e/v2/tests/hosted_cluster_aws_test.go
+++ b/test/e2e/v2/tests/hosted_cluster_aws_test.go
@@ -33,6 +33,7 @@ import (
 	"github.com/openshift/hypershift/test/e2e/v2/internal"
 
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	crclient "sigs.k8s.io/controller-runtime/pkg/client"
@@ -91,6 +92,8 @@ func EnsureDefaultSecurityGroupTagsTest(getTestCtx internal.TestContextGetter) {
 			day2TagKey := "test-day2-tag"
 			day2TagValue := "test-day2-value"
 
+			originalTags := append([]hyperv1.AWSResourceTag(nil), hc.Spec.Platform.AWS.ResourceTags...)
+
 			err = e2eutil.UpdateObject(GinkgoTB(), tc.Context, tc.MgmtClient, hc, func(obj *hyperv1.HostedCluster) {
 				obj.Spec.Platform.AWS.ResourceTags = append(obj.Spec.Platform.AWS.ResourceTags, hyperv1.AWSResourceTag{
 					Key:   day2TagKey,
@@ -98,6 +101,14 @@ func EnsureDefaultSecurityGroupTagsTest(getTestCtx internal.TestContextGetter) {
 				})
 			})
 			Expect(err).NotTo(HaveOccurred(), "failed to update HostedCluster with day-2 tag")
+			DeferCleanup(func() {
+				err := e2eutil.UpdateObject(GinkgoTB(), tc.Context, tc.MgmtClient, hc, func(obj *hyperv1.HostedCluster) {
+					obj.Spec.Platform.AWS.ResourceTags = append([]hyperv1.AWSResourceTag(nil), originalTags...)
+				})
+				if err != nil && !apierrors.IsNotFound(err) {
+					Expect(err).NotTo(HaveOccurred(), "cleanup: failed to restore HostedCluster AWS resource tags")
+				}
+			})
 
 			Eventually(func(g Gomega) {
 				sg, err := e2eutil.GetDefaultSecurityGroup(tc.Context, awsCredsFile, region, sgID)
@@ -159,9 +170,12 @@ func AWSCCMWithCustomizationsTest(getTestCtx internal.TestContextGetter) {
 					},
 				}
 				Expect(hcClient.Create(tc.Context, testNS)).To(Succeed(), "failed to create test namespace")
-				defer func() {
-					_ = hcClient.Delete(tc.Context, testNS)
-				}()
+				DeferCleanup(func() {
+					err := hcClient.Delete(tc.Context, testNS)
+					if err != nil && !apierrors.IsNotFound(err) {
+						Expect(err).NotTo(HaveOccurred(), "cleanup: failed to delete test namespace %s", testNS.Name)
+					}
+				})
 
 				testSvc := &corev1.Service{
 					ObjectMeta: metav1.ObjectMeta{
--- a/test/e2e/v2/tests/hosted_cluster_aws_test.go
+++ b/test/e2e/v2/tests/hosted_cluster_aws_test.go
@@ -1,0 +1,265 @@
+//go:build e2ev2
+
+/*
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package tests
+
+import (
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	ec2types "github.com/aws/aws-sdk-go-v2/service/ec2/types"
+	elbv2 "github.com/aws/aws-sdk-go-v2/service/elasticloadbalancingv2"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	hyperv1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
+	awsutil "github.com/openshift/hypershift/cmd/infra/aws/util"
+	e2eutil "github.com/openshift/hypershift/test/e2e/util"
+	"github.com/openshift/hypershift/test/e2e/v2/internal"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	crclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func RegisterHostedClusterAWSTests(getTestCtx internal.TestContextGetter) {
+	EnsureDefaultSecurityGroupTagsTest(getTestCtx)
+	AWSCCMWithCustomizationsTest(getTestCtx)
+}
+
+func EnsureDefaultSecurityGroupTagsTest(getTestCtx internal.TestContextGetter) {
+	When("a day-2 resource tag is added to the HostedCluster spec", func() {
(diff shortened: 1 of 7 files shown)


bryan-cox (commented):
V1 vs V2 coverage review — checked all 6 ported tests against their v1 originals. 4/6 are full parity or improved. Two have coverage gaps noted inline.

bryan-cox (commented):
V1 vs V2 coverage review — checked all 6 ported tests against their v1 originals. 4/6 are full parity or improved. Two have coverage gaps noted inline.

test/e2e/v2/tests/hosted_cluster_pull_secret_test.go
```diff
@@ -0,0 +1,365 @@
...
+			g.Expect(daemonSet.Status.DesiredNumberScheduled).To(BeNumerically(">=", nodeCount),
+				"DaemonSet %s desired=%d < minExpected=%d", ds.name, daemonSet.Status.DesiredNumberScheduled, nodeCount)
+			g.Expect(daemonSet.Status.NumberReady).To(Equal(daemonSet.Status.DesiredNumberScheduled),
+				"DaemonSet %s not fully ready: %d/%d", ds.name, daemonSet.Status.NumberReady, daemonSet.Status.DesiredNumberScheduled)
+		}, 20*time.Minute, 10*time.Second).Should(Succeed(), "DaemonSet %s/%s did not become ready", ds.namespace, ds.name)
+	}
+}
+
+func init() {
+	internal.RegisterEnvVar(
```
bryan-cox: suggestion: AGENTS.md Standard #8 says to register all env vars in the central catalog (`internal/env_vars.go`). This is the only env var across all v2 test files registered in an `init()` inside the test file itself. Move `E2E_ADDITIONAL_PULL_SECRET_FILE` to `internal/env_vars.go` for consistency.

test/e2e/v2/tests/hosted_cluster_pull_secret_test.go
```diff
@@ -0,0 +1,365 @@
...
+			Eventually(func(g Gomega) {
+				ds := &appsv1.DaemonSet{}
+				g.Expect(hcClient.Get(tc.Context, crclient.ObjectKey{
+					Name:      hccomanifests.GlobalPullSecretDSName,
+					Namespace: hccomanifests.GlobalPullSecretNamespace,
+				}, ds)).To(Succeed())
+				g.Expect(ds.Status.DesiredNumberScheduled).To(BeNumerically(">=", nodeCount))
+				g.Expect(ds.Status.NumberReady).To(Equal(ds.Status.DesiredNumberScheduled))
+			}, 20*time.Minute, 10*time.Second).Should(Succeed())
+		})
```
bryan-cox, line 188: question: The v1 `EnsureGlobalPullSecret` includes three validation steps that are not ported here:

1. `VerifyKubeletConfigWithDaemonSet` (called twice in v1) - deploys a verifier DaemonSet that mounts the host filesystem and compares on-disk `/var/lib/kubelet/config.json` against the cluster pull secret.
2. Negative pod pull test - creates a pod with unauthorized creds and asserts image pull fails.
3. Positive pod pull test - creates a pod after applying valid creds and asserts image pull succeeds.

Together these prove the global pull secret pipeline works at the node level, not just at the Kubernetes object level. Was the omission intentional, or should these be added as a follow-up?

test/e2e/v2/tests/hosted_cluster_ingress_test.go
```diff
@@ -0,0 +1,96 @@
...
+func ValidateIngressOperatorConfigurationTest(getTestCtx internal.TestContextGetter) {
+	When("hosted cluster has IngressOperator EndpointPublishingStrategy configured", func() {
+		It("should reflect the custom strategy in the hosted cluster IngressController", func() {
+			tc := getTestCtx()
+			if e2eutil.IsLessThan(e2eutil.Version421) {
+				Skip("Ingress operator configuration requires version >= 4.21")
+			}
+			hc := tc.GetHostedCluster()
+
+			if hc.Spec.OperatorConfiguration == nil ||
```
bryan-cox, line 46: question: V1 (`util_ingress_operator_configuration.go:25-27`) fails the test if `OperatorConfiguration`, `IngressOperator`, or `EndpointPublishingStrategy` is nil. V2 skips instead. This means a regression where the HC is created without the expected ingress configuration would pass silently. Was the change from fail-to-skip intentional?

test/e2e/v2/tests/hosted_cluster_aws_test.go
```diff
@@ -0,0 +1,279 @@
...
+				]
+			}`, sgID)
+
+			Expect(hc.Spec.Platform.AWS.RolesRef.ControlPlaneOperatorARN).NotTo(BeEmpty(),
+				"HostedCluster should have ControlPlaneOperatorARN set")
+
+			cleanup, err := e2eutil.PutRolePolicy(tc.Context, awsCredsFile, region,
+				hc.Spec.Platform.AWS.RolesRef.ControlPlaneOperatorARN, tagsPolicy)
+			Expect(err).NotTo(HaveOccurred(), "failed to put role policy for tagging default security group")
+			defer func() {
```
bryan-cox: suggestion: This uses `defer` instead of `DeferCleanup`. If Ginkgo aborts the spec (timeout, interrupt), `defer` is not guaranteed to run — Ginkgo can panic-recover above the call site. The tag restoration on line 104 correctly uses `DeferCleanup`. Should be consistent:

```go
DeferCleanup(func() {
    Expect(cleanup()).To(Succeed(), "failed to cleanup role policy for tagging default security group")
})
```

test/e2e/v2/tests/hosted_cluster_pull_secret_test.go
cblecker: Good catch. Moved the `RegisterEnvVar` call to `internal/env_vars.go` per Standard #8.

test/e2e/v2/tests/hosted_cluster_pull_secret_test.go
cblecker: The omission was intentional. The v2 test validates the full Kubernetes API-level pipeline: management-cluster secret propagation, GlobalPullSecret creation/update/deletion, and DaemonSet readiness (ovnkube-node, global-pull-secret-syncer, konnectivity-agent). The DaemonSet readiness checks serve as implicit proof that images can be pulled and pods scheduled successfully.

The three omitted v1 steps — `VerifyKubeletConfigWithDaemonSet` (on-disk kubelet config.json verification), negative pod pull test, and positive pod pull test — are defense-in-depth at the node/filesystem and container-runtime levels. They're valuable but represent a separate concern from the controller pipeline being validated here.

test/e2e/v2/tests/hosted_cluster_ingress_test.go
cblecker: The change from Fail to Skip was intentional, following CLAUDE.md Standard #13: non-lifecycle tests must only verify existing state and `Skip()` when preconditions are absent.

In v1, the test ran against a freshly-created cluster with specific ingress configuration, so `Fail()` was appropriate — the configuration was guaranteed to be present. In v2's model, `OperatorConfiguration.IngressOperator` is an optional field. Not all hosted clusters have it configured (e.g., clusters without internal load balancing). The test verifies the configuration if present and gracefully skips otherwise.

The test is non-lifecycle (read-only, no `lifecycle` label, no cluster mutation), so Standard #13 applies.

test/e2e/v2/tests/hosted_cluster_aws_test.go
cblecker: Fixed — changed to `DeferCleanup` for consistency with line 104 and Ginkgo abort/timeout safety.

test/e2e/v2/tests/hosted_cluster_ingress_test.go
```diff
@@ -0,0 +1,96 @@
...
+				hc.Spec.OperatorConfiguration.IngressOperator == nil ||
+				hc.Spec.OperatorConfiguration.IngressOperator.EndpointPublishingStrategy == nil {
+				Skip("HostedCluster does not have IngressOperator EndpointPublishingStrategy configured")
+			}
+
+			expectedStrategy := hc.Spec.OperatorConfiguration.IngressOperator.EndpointPublishingStrategy
+			Expect(expectedStrategy.Type).To(Equal(operatorv1.LoadBalancerServiceStrategyType),
```
csrwng: I would consider these pre-conditions and not necessarily test failures. If the strategy is not LoadBalancer or LoadBalancer does not have the internal scope, then skip the test.

csrwng (commented):

test/e2e/v2/tests/hosted_cluster_ingress_test.go
cblecker: Good catch. Removed them entirely rather than converting to Skip — these were artifacts of the v1 lifecycle model where the test itself injected that exact configuration via a `BeforeApply` hook. In v2 the test runs against a pre-existing cluster it doesn't control, so they were both redundant and unnecessarily restrictive. The `Eventually` block already validates propagation dynamically using `expectedStrategy`, so the test now works for any configured strategy type.

PR #8597, opened 2026-05-27
OCPBUGS-43444: Allow kubevirt-csi storageclass default to be changed by user

orenc1:
When creating an HCP cluster on OpenShift Virtualization (KubeVirt), the `kubevirt-csi-infra-default` StorageClass is automatically set as the default via the `storageclass.kubernetes.io/is-default-class` annotation. The `reconcileDefaultTenantStorageClass` function unconditionally overwrites the entire annotations map with `is-default-class: "true"` on every reconciliation cycle. This prevents users from setting the annotation to "false" to designate a different CSI driver's StorageClass as the default — any manual change is immediately reverted by the operator.

This behavior is inconsistent with non-HCP OCP clusters where users can freely toggle the default StorageClass, and has caused issues where applications pick the `kubevirt-csi` StorageClass via the "default" designation when a different one is intended.

This commit changes `reconcileDefaultTenantStorageClass` to:
- Initialize the annotations map only when nil, preserving any pre-existing annotations instead of clobbering them.
- Only set `is-default-class` to "true" when the annotation does not already exist. If the annotation is present (regardless of value), the user's choice is preserved across reconciliation cycles.

Unit tests are added covering:
- Fresh StorageClass (no annotation) gets is-default-class: "true"
- User-set annotation of "false" is preserved
- User-set annotation of "true" is preserved
- Pre-existing unrelated annotations are not clobbered

Fixes: https://redhat.atlassian.net/browse/OCPBUGS-43444

## Checklist:

## Release Notes

* **Bug Fixes**
  * Fixed StorageClass annotation handling to preserve existing annotations instead of overwriting them during reconciliation.

* **Tests**
  * Added comprehensive test coverage for StorageClass configuration management, including annotation preservation scenarios.

### DIFF
--- a/control-plane-operator/controllers/hostedcontrolplane/csi/kubevirt/kubevirt.go
+++ b/control-plane-operator/controllers/hostedcontrolplane/csi/kubevirt/kubevirt.go
@@ -135,9 +135,17 @@ func reconcileCustomTenantStorageClass(sc *storagev1.StorageClass, infraSCName s
 }
 
 func reconcileDefaultTenantStorageClass(sc *storagev1.StorageClass) error {
-	sc.Annotations = map[string]string{
-		"storageclass.kubernetes.io/is-default-class": "true",
+	if sc.Annotations == nil {
+		sc.Annotations = map[string]string{}
 	}
+
+	// Only set the default storage class when the annotation doesn't exist.
+	// This allows users to set is-default-class=false if they wish to
+	// install their own CSI drivers and choose a different default.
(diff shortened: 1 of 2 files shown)


jparrill (commented):
/lgtm

orenc1:
CI infrastructure issue.
/retest-required

orenc1:
there was an outage in quay.io
/retest-required

orenc1:
/jira backport release-4.22,release-4.21,release-4.20,release-4.19,release-4.18

PR #8601, opened 2026-05-27
OCPBUGS-89360: NO-JIRA: build(operator): drop hypershift-no-cgo from operator container images

yiraeChristineKim:
Remove the QE-only non-CGO hypershift CLI from production operator images built via Containerfile.operator and Dockerfile. The non-FIPS hypershift-no-cgo binary causes Enterprise Contract FIPS violations in Konflux builds for MCE.
The FIPS-compliant hypershift binary (CGO_ENABLED=1) remains as the shipped CLI. Also remove the hypershift-no-cgo Makefile target and update the build-ho-image skill docs to match.

```
multicluster-engine-hypershift-operator,/usr/bin/hcp,go binary is not CGO_ENABLED
  multicluster-engine-hypershift-operator,/usr/bin/hypershift-no-cgo,go binary does not contain required symbol(s)
  multicluster-engine-managedcluster-import-controller-container,/etc/redhat-release,operating system is not FIPS certified
```
Conversation
Ref: https://redhat-internal.slack.com/archives/C058TF9K37Z/p1779803011756959

Fixes 

## Checklist:

## Summary by CodeRabbit

* **Chores**
  * Optimized operator container image by streamlining build artifacts and removing unnecessary binaries from the final runtime image.

### DIFF
--- a/Containerfile.operator
+++ b/Containerfile.operator
@@ -6,15 +6,11 @@ WORKDIR /hypershift
 COPY --chown=default . .
 
 RUN make hypershift \
-  && make hypershift-no-cgo \
   && make hypershift-operator \
-  && make product-cli \
   && make karpenter-operator
 
 FROM registry.access.redhat.com/ubi9/ubi-minimal:9.7-1777857961
 COPY --from=builder /hypershift/bin/hypershift \
-  /hypershift/bin/hypershift-no-cgo \
-  /hypershift/bin/hcp \
   /hypershift/bin/hypershift-operator \
   /hypershift/bin/karpenter-operator \
   /usr/bin/


Containerfile.operator
```diff
@@ -6,15 +6,11 @@ WORKDIR /hypershift
 COPY --chown=default . .
 
 RUN make hypershift \
-  && make hypershift-no-cgo \
   && make hypershift-operator \
-  && make product-cli \
```
mgencur, line 11: Why removing product-cli? Are we supposed to remove only `hypershift-no-cgo` ?

Containerfile.operator
```diff
@@ -6,15 +6,11 @@ WORKDIR /hypershift
...
   && make karpenter-operator
 
 FROM registry.access.redhat.com/ubi9/ubi-minimal:9.7-1777857961
 COPY --from=builder /hypershift/bin/hypershift \
-  /hypershift/bin/hypershift-no-cgo \
-  /hypershift/bin/hcp \
```
mgencur, line 17: Why remove `hcp` ? Are we supposed to remove only `hypershift-no-cgo` ?

Dockerfile
```diff
@@ -6,15 +6,11 @@ WORKDIR /hypershift
 COPY . .
 
 RUN make hypershift \
-  && make hypershift-no-cgo \
   && make hypershift-operator \
-  && make product-cli \
```
mgencur, line 11: Why `product-cli` ? Are we supposed to remove only `hypershift-no-cgo` ?

Dockerfile.dev
```diff
@@ -12,7 +12,6 @@ RUN make build
 
 FROM registry.access.redhat.com/ubi9:latest
 COPY --from=builder /hypershift/bin/hypershift \
-                    /hypershift/bin/hcp \
```
mgencur, line 15: Are we supposed to remove only `hypershift-no-cgo` ?

Dockerfile
```diff
@@ -6,15 +6,11 @@ WORKDIR /hypershift
...
   && make karpenter-operator
 
 FROM registry.access.redhat.com/ubi9:latest
 COPY --from=builder /hypershift/bin/hypershift \
-                    /hypershift/bin/hypershift-no-cgo \
-                    /hypershift/bin/hcp \
```
mgencur, line 17: Are we supposed to remove only `hypershift-no-cgo` ?

mgencur (changes requested):

Containerfile.operator
yiraeChristineKim: We build product-cli using https://github.com/openshift/hypershift/blob/main/Containerfile.cli

Containerfile.operator
yiraeChristineKim: We build product-cli using https://github.com/openshift/hypershift/blob/main/Containerfile.cli

Dockerfile
yiraeChristineKim: We build product-cli using https://github.com/openshift/hypershift/blob/main/Containerfile.cli

Dockerfile.dev
yiraeChristineKim: We build product-cli using https://github.com/openshift/hypershift/blob/main/Containerfile.cli

Dockerfile
yiraeChristineKim: We build product-cli using https://github.com/openshift/hypershift/blob/main/Containerfile.cli

Containerfile.operator
mgencur: I'm sorry but I don't understand. Can you shed more light on this? The fact that you build product-cli using that container file doesn't explain to me why you're removing `make product-cli`.

mgencur: I mean. This looks like completely unrelated to `hypershift-no-cgo`. I don't see any relation. So maybe this should be sent as a separate PR with a better description?

rokej: @mgencur Removing the product-cli seems unrelated but this PR is about removing non-FIPS compliant binaries that are not necessary to run the hypershift operator from the container image and the product-cli is another non-FIPS compliant binary. We deliver the product-cli in a separate CLI download container image and make it available from the cluster for users to download and use.

Dockerfile
rokej: See my comment https://github.com/openshift/hypershift/pull/8601#discussion_r3356576011

csrwng:
@yiraeChristineKim @rokej do we need to remove the statically compiled binary from all images? My understanding is that we'd only need to do it for the Containerfile(s) built in konflux.

yiraeChristineKim:
> @yiraeChristineKim @rokej do we need to remove the statically compiled binary from all images? My understanding is that we'd only need to do it for the Containerfile(s) built in konflux.

We are only using the CLI binary built from Containerfile.cli. Are you using a CLI binary built from any container file other than Containerfile.cli?

rokej:
@csrwng I missed it. Yes, the binaries should be removed only from the konflux container file which is `Containerfile.operator` only. cc @yiraeChristineKim

yiraeChristineKim:
@rokej @csrwng  Can you review this again

yiraeChristineKim:
@mgencur Could you please review this? It's urgent.

yiraeChristineKim:
@csrwng  Can I override these tests?

csrwng:
/override ci/prow/e2e-aks
/override ci/prow/e2e-aws

The changed containerfile does not impact these tests.

csrwng:

csrwng:
/override "Red Hat Konflux / hypershift-release-mce-50-on-pull-request"

Both konflux tests finished successfully, but for whatever reason the result was not reported back to github

yiraeChristineKim:
/jira backport release-4.22,release-4.21,release-4.20,release-4.19,release-4.18,release-4.17,release-4.16

PR #8602, opened 2026-05-27
CNTRLPLANE-3434: add ho-release-gate pipeline for nightly promotion

Nirshal:
## Summary

Adds the Tekton pipeline and PipelineRun template for the HyperShift Operator nightly release gating pipeline, triggered via IntegrationTestScenario (ITS).

- Part of [CNTRLPLANE-3434](https://redhat.atlassian.net/browse/CNTRLPLANE-3434) / [OCPSTRAT-3250](https://redhat.atlassian.net/browse/OCPSTRAT-3250)
- Stale promotion alerting: [CNTRLPLANE-3451](https://redhat.atlassian.net/browse/CNTRLPLANE-3451)
- Enhancement PR: [openshift/enhancements#2016](https://github.com/openshift/enhancements/pull/2016)
- Pipeline validates HO images via Prow e2e tests before promoting them to a verified Quay repo
- ARO HCP is the pilot platform; ROSA HCP and GCP HCP follow later using the same template
- **Depends on [openshift/release#81877](https://github.com/openshift/release/pull/81877)** for the `OVERRIDE_HYPERSHIFT_OPERATOR_IMAGE` step parameter and Gangway transport variable

## Pipeline flow

```
CronJob (3:15 UTC nightly)
  |-- Resolves latest Konflux Snapshot (auto-released, push build)
  |-- Labels Snapshot with test.appstudio.openshift.io/run=<its-name>

Integration Service
  |-- Detects labeled Snapshot, creates PipelineRun from template
  |-- Injects SNAPSHOT JSON payload + ITS params into PipelineRun

PipelineRun (ho-release-gate):
  extract-image     - Parses SNAPSHOT JSON to extract HO image and commit SHA.
                      Gets snapshot name from PipelineRun label.
  run-e2e           - Triggers Prow periodic e2e jobs via gangway API.
                      Two categories: blocking (gate fails if any fail) and
                      informing (reported but do not block the gate).
                      45min initial delay, then polls every 10 min with 30s stagger.
                      Overrides: MULTISTAGE_PARAM_OVERRIDE_OVERRIDE_HYPERSHIFT_OPERATOR_IMAGE
                                 (bypasses ci-operator ImageStream race condition),
                                 OVERRIDE_IMAGE_HYPERSHIFT_TESTS.
                      Always exits 0 - writes per-job results JSON (with type field).
  evaluate-results  - Reads results, applies AND logic on blocking tests only.
                      Informing failures logged as warnings.
                      Always exits 0 - writes gate-passed result (true/false).
  create-release    - If gate passed: creates Release CR referencing the ReleasePlan
                      and Snapshot, writes release name to result.
                      If gate failed: writes "N/A" to result, exits 1
                      (marks pipeline as Failed in UI).
  finally:
    notify-slack       - Fires on every run (pass or fail).
                         Three visual states:
                           Green (#2E7D32): gate passed, Release created.
                           Orange (#F57C00): gate passed, Release creation failed.
                           Red (#D32F2F): gate failed.
                         Gate-label in header, grouped results (blocking + informing).
                         On gate failure: checks for stale promotion streak via
                         KubeArchive. If streak >= threshold, sends stale alert
                         instead of normal failure notification.
    notify-slack-error - Fallback for catastrophic failures (DAG failure).
                         Zero result dependencies, fires when create-release
                         did not run. Gate-label in header.
                         Also checks for stale promotion streak via KubeArchive.
                         Sends stale alert or generic error notification.

Release (on gate pass):
  create-release task creates Release CR explicitly.
  ReleasePlan targets rhtap-releng-tenant (managed workspace).
  Managed pipeline pushes image to verified Quay repo.
```

## Stale promotion alerting ([CNTRLPLANE-3451](https://redhat.atlassian.net/browse/CNTRLPLANE-3451))

When the gate fails, both `notify-slack` and `notify-slack-error` query the
KubeArchive REST API for historical PipelineRuns matching the ITS scenario label.
The consecutive failure streak is measured in days (difference between now and the
oldest consecutive failed run, inclusive of today).

If `streak_days >= stale-threshold-days`, the normal notification is replaced with
a stale promotion alert containing:

- Header with gate label and streak duration
- Fields: Gate, Threshold, Streak, Current PipelineRun
- History of up to 10 most recent failed runs with clickable PipelineRun links,
  dates, and failure reasons (Failed, PipelineRunTimeout, CouldntGetPipeline, etc.)
- Truncation summary for streaks longer than 10 runs
- Footer with investigation guidance

The stale check is safe to skip: if KubeArchive is unreachable or returns no data,
the normal notification is sent instead.

## Notification reliability

All tasks use **default-first initialization**: each task writes safe default values
to its result paths at the start of its script, before any logic. If a task starts
but crashes mid-execution (e.g., OOMKilled), results are still initialized.

Two mutually exclusive finally tasks ensure a notification is always sent:
- `notify-slack`: detailed notification, depends on task results (skipped if results uninitialized)
- `notify-slack-error`: generic fallback, zero result dependencies, fires only when `create-release` was skipped by DAG failure

Both finally tasks perform the stale promotion check independently, so a stale alert
is sent regardless of which notification path fires.

## Files

| File | Description |
|------|-------------|
| `.tekton/pipelines/ho-release-gate.yaml` | Pipeline definition |
| `.tekton/pipelines/ho-release-gate-run.yaml` | PipelineRun template (referenced by ITS via git resolver) |
| `.tekton/lib/ho_release_gate.py` | Pipeline orchestration: image extraction, job triggering/polling, gate evaluation, notifications, stale alerting |
| `.tekton/lib/kubearchive_utils.py` | KubeArchive REST API: fetch historical PipelineRuns |
| `.tekton/lib/prow_utils.py` | Gangway API: trigger jobs, resolve URLs, poll status |
| `.tekton/lib/slack_utils.py` | Slack webhook: send messages, build Block Kit payloads |
| `.tekton/lib/http_utils.py` | Low-level HTTP wrapper with retry logic (stdlib only) |
| `.tekton/lib/README.md` | Library documentation: architecture, constraints, runtime environment |
| `.tekton/lib/tests/` | Unit tests for all modules (unittest + unittest.mock, stdlib only) |
| `.tekton/lib/mock/` | Mock utilities for pipeline integration testing |

## Pipeline parameters

| Parameter | Source | Description |
|-----------|--------|-------------|
| `SNAPSHOT` | Integration Service (auto-injected) | JSON payload of the Konflux Snapshot |
| `e2e-blocking-job-names` | ITS params | JSON array of blocking Prow periodic job names |
| `e2e-informing-job-names` | ITS params | JSON array of informing Prow periodic job names |
| `gate-label` | ITS params | Label identifying which service gate (e.g. ARO HCP) |
| `release-plan-name` | ITS params | Name of the ReleasePlan to reference when creating the Release CR |
| `stale-threshold-days` | ITS params | Number of consecutive failure days before triggering stale alert (default: 3) |

## Resources

All resources live in namespace `crt-redhat-acm-tenant` unless noted.

| Resource | Name | Managed via |
|----------|------|-------------|
| Pipeline + PipelineRun template | `.tekton/pipelines/ho-release-gate*.yaml` | This PR |
| ReleasePlanAdmission | `redhat-hypershift-operator-ho-release-gate-aro-hcp` (in `rhtap-releng-tenant`) | GitOps ([!18934](https://gitlab.cee.redhat.com/releng/konflux-release-data/-/merge_requests/18934) - Merged) |
| ReleasePlan | `hypershift-operator-ho-release-gate-aro-hcp` | GitOps ([!18938](https://gitlab.cee.redhat.com/releng/konflux-release-data/-/merge_requests/18938) - Merged) |
| CronJob | `hypershift-operator-nightly-promotion` | GitOps ([!18938](https://gitlab.cee.redhat.com/releng/konflux-release-data/-/merge_requests/18938) - Merged) |
| IntegrationTestScenario | `hypershift-ho-release-gate-aro-hcp` | GitOps ([!19261](https://gitlab.cee.redhat.com/releng/konflux-release-data/-/merge_requests/19261) - Merged) |
| ServiceAccount | `nightly-promotion-sa` | GitOps ([!18934](https://gitlab.cee.redhat.com/releng/konflux-release-data/-/merge_requests/18934) - Merged) |
| RBAC (snapshot labeling) | `nightly-promotion-sa` -> `konflux-tester-internalbot-actions` | GitOps ([!19870](https://gitlab.cee.redhat.com/releng/konflux-release-data/-/merge_requests/19870) - Merged) |
| RBAC (release creation) | `nightly-promotion-sa` -> `konflux-releaser-bot-actions` via taskRunSpecs | GitOps ([!19870](https://gitlab.cee.redhat.com/releng/konflux-release-data/-/merge_requests/19870) - Merged) |
| Secret | `gangway-token` (Prow auth) | Manual (not in GitOps) |
| Secret | `slack-webhook` (Slack notifications) | Manual (not in GitOps) |

## Release mechanism

The pipeline uses **explicit Release CR creation** (not auto-release):

1. The `create-release` task creates a `Release` object referencing the ReleasePlan (name passed via ITS param `release-plan-name`) and the tested Snapshot
2. The ReleasePlan targets `rhtap-releng-tenant` (managed workspace)
3. The RPA (`redhat-hypershift-operator-ho-release-gate-aro-hcp`) picks up the Release
4. The managed pipeline (`rh-push-to-external-registry`) pushes the validated image to the verified Quay repo with platform-prefixed tags (`aro-hcp-latest`, `aro-hcp-latest-{{ timestamp }}`, `aro-hcp-{{ git_sha }}`, etc.)

Target repo: `quay.io/redhat-services-prod/crt-redhat-acm-tenant/hypershift/hypershift-operator-verified`

> **Note:** Auto-release was disabled ([!19296](https://gitlab.cee.redhat.com/releng/konflux-release-data/-/merge_requests/19296) - Merged) because with `contexts: disabled` on the ITS, Integration Service was auto-releasing every normal push snapshot through our ReleasePlan, bypassing the release gate entirely.

## Remaining work

| Item | Status | Notes |
|------|--------|-------|
| HO image override fix | **Blocked** | Depends on [openshift/release#81877](https://github.com/openshift/release/pull/81877) (adds `OVERRIDE_HYPERSHIFT_OPERATOR_IMAGE` step parameter + Gangway transport variable to `hypershift-install`) |
| Pipeline source migration | On merge | ITS resolverRef will point to `openshift/hypershift:main` once this PR merges ([!19913](https://gitlab.cee.redhat.com/releng/konflux-release-data/-/merge_requests/19913) - Draft) |
| Regression analysis | Future | Component-readiness tracking (deads2k requirement) |

## Related

- Enhancement PR: [openshift/enhancements#2016](https://github.com/openshift/enhancements/pull/2016)
- Documentation PR: [openshift/hypershift#8947](https://github.com/openshift/hypershift/pull/8947) - to be merged AFTER this PR
- **HO image override**: [openshift/release#81877](https://github.com/openshift/release/pull/81877) - adds `OVERRIDE_HYPERSHIFT_OPERATOR_IMAGE` parameter to `hypershift-install` step (dependency)
- GitOps MR (RPA + SA): [konflux-release-data!18934](https://gitlab.cee.redhat.com/releng/konflux-release-data/-/merge_requests/18934) - Merged
- GitOps MR (ReleasePlan + CronJob): [konflux-release-data!18938](https://gitlab.cee.redhat.com/releng/konflux-release-data/-/merge_requests/18938) - Merged
- GitOps MR (ITS + params + RBAC): [konflux-release-data!19261](https://gitlab.cee.redhat.com/releng/konflux-release-data/-/merge_requests/19261) - Merged
- GitOps MR (disable auto-release): [konflux-release-data!19296](https://gitlab.cee.redhat.com/releng/konflux-release-data/-/merge_requests/19296) - Merged
- GitOps MR (RBAC cleanup): [konflux-release-data!19870](https://gitlab.cee.redhat.com/releng/konflux-release-data/-/merge_requests/19870) - Merged
- GitOps MR (cleanup + upstream): [konflux-release-data!19913](https://gitlab.cee.redhat.com/releng/konflux-release-data/-/merge_requests/19913) - Draft
- Gangway token PR: [openshift/release#80420](https://github.com/openshift/release/pull/80420) - Merged
- Epic: [CNTRLPLANE-3434](https://redhat.atlassian.net/browse/CNTRLPLANE-3434)
- Feature: [OCPSTRAT-3250](https://redhat.atlassian.net/browse/OCPSTRAT-3250)
- Stale alerting: [CNTRLPLANE-3451](https://redhat.atlassian.net/browse/CNTRLPLANE-3451)

## Summary by CodeRabbit

* **New Features**
  * Added a new Tekton release gate pipeline that parses the provided snapshot, runs configured Gangway Prow jobs, and computes a gate verdict.
  * Release creation is skipped when blocking checks don’t pass.
  * Added Slack notifications for normal results, errors, and stale-promotion conditions; notifications run even on failures.
* **Documentation**
  * Documented the shared stdlib-only release-gate helper library and how to run its tests.
* **Tests**
  * Added unit test coverage for HTTP retries, Prow/Gangway helpers, gate evaluation, Slack payloads, and stale detection logic.

### DIFF
--- a/.tekton/lib/ho_release_gate.py
+++ b/.tekton/lib/ho_release_gate.py
@@ -629,7 +629,8 @@ def build_stale_notification(streak, gate_label, threshold_days,
 
 def check_and_build_stale_payload(token, its_scenario, threshold_days,
                                   current_run_name, konflux_base_url,
-                                  namespace, gate_label):
+                                  namespace, gate_label,
+                                  kubearchive_api_base):
     """Check for a stale promotion streak and build the alert payload.
 
     Queries KubeArchive for historical PipelineRuns matching the ITS
@@ -648,6 +649,7 @@ def check_and_build_stale_payload(token, its_scenario, threshold_days,
         konflux_base_url: Base URL of the Konflux UI for building links.
         namespace: Kubernetes namespace for the KubeArchive query.
         gate_label: Human-readable gate label (e.g. "ARO HCP").
+        kubearchive_api_base: Base URL of the KubeArchive API server.
 
     Returns:
         Dict payload for send_slack_message if stale (streak days >=
@@ -661,7 +663,8 @@ def check_and_build_stale_payload(token, its_scenario, threshold_days,
     print(f"  Label selector:  {label_selector}", flush=True)
     print(f"  Threshold:       {threshold_days} day(s)", flush=True)
 
-    runs = fetch_pipelineruns(token, namespace, label_selector)
+    runs = fetch_pipelineruns(token, namespace, label_selector,
+                              kubearchive_api_base)
     runs = [r for r in runs if r["name"] != current_run_name]
     if not runs:
         print("  No historical runs found - skipping stale check",
--- a/.tekton/lib/ho_release_gate.py
+++ b/.tekton/lib/ho_release_gate.py
@@ -680,8 +680,13 @@ def check_and_build_stale_payload(token, its_scenario, threshold_days,
               " - skipping stale check", flush=True)
         return None
 
-    oldest_dt = datetime.fromisoformat(
-        oldest_created.replace("Z", "+00:00"))
+    try:
+        oldest_dt = datetime.fromisoformat(
+            oldest_created.replace("Z", "+00:00"))
+    except ValueError:
+        print(f"  WARN: cannot parse timestamp '{oldest_created}'"
+              " - skipping stale check", flush=True)
+        return None
     streak_days = (datetime.now(timezone.utc) - oldest_dt).days + 1
 
     print(f"  Failure streak: {len(streak)} run(s)"
--- a/.tekton/lib/ho_release_gate.py
+++ b/.tekton/lib/ho_release_gate.py
@@ -1,0 +1,701 @@
+"""Pipeline-specific logic for the HO Release Gate."""
+
+import json
+import re
+import time
+from datetime import datetime, timezone
+
+from kubearchive_utils import fetch_pipelineruns, build_pipelinerun_url
+from prow_utils import (trigger_prow_job, resolve_prow_url,
+                        get_prow_job_status, short_name)
+from slack_utils import build_slack_payload, mrkdwn_section, fields_section, divider
+
+__all__ = ["extract_component_image",
+           "trigger_all_jobs", "resolve_all_urls", "poll_until_complete",
+           "print_run_summary", "build_results_json", "evaluate_gate",
+           "build_gate_notification", "build_error_notification",
+           "check_failure_streak", "build_stale_notification",
+           "check_and_build_stale_payload"]
+
+
+def extract_component_image(snapshot_json_str,
+                            component_name="hypershift-operator-main"):
+    """Extract container image reference from a Konflux Snapshot JSON.
+
+    Parses the Snapshot, locates the named component, and returns its
+    container image along with the sha256 digest and git revision.
+
+    Args:
+        snapshot_json_str: Raw JSON string of the Konflux Snapshot.
+        component_name: Name of the component to extract (default:
+            hypershift-operator-main).
+
+    Returns:
+        Dict with keys: image, digest, revision.
+
+    Raises:
+        ValueError: If the JSON is invalid, the component is not found,
+            or the containerImage field is missing/empty.
+    """
+    try:
+        snapshot = json.loads(snapshot_json_str)
+    except (json.JSONDecodeError, TypeError) as e:
+        raise ValueError(f"SNAPSHOT JSON is not valid: {e}")
+
+    for comp in snapshot.get("components", []):
+        if comp.get("name") == component_name:
+            image = comp.get("containerImage", "")
+            if not image:
+                raise ValueError(
+                    f"Component '{component_name}' has no containerImage")
+
+            match = re.search(r"sha256:[a-f0-9]+", image)
+            digest = match.group(0) if match else "N/A"
+
+            revision = (comp.get("source", {})
+                        .get("git", {})
+                        .get("revision", "N/A"))
+
+            return {"image": image, "digest": digest, "revision": revision}
+
+    raise ValueError(f"Component '{component_name}' not found in SNAPSHOT")
+
+
+def trigger_all_jobs(blocking_names, informing_names, gangway_url, token,
+                     env_overrides, trigger_delay=60, rate_limit_backoff=120,
+                     max_retries=3):
+    """Trigger all blocking and informing Prow jobs sequentially.
+
+    Iterates through blocking jobs first, then informing, with a delay
+    between each trigger to respect Gangway rate limits. Raises
+    ValueError if blocking_names is empty.
+
+    Args:
+        blocking_names: List of blocking Prow periodic job name strings.
+            At least one is required.
+        informing_names: List of informing Prow periodic job name strings.
+            May be empty.
+        gangway_url: Base URL of the Gangway executions endpoint.
+        token: Bearer token for Gangway authentication.
+        env_overrides: Dict of environment variables injected into each
+            Prow job's PodSpec (e.g. HO image override).
+        trigger_delay: Seconds to wait between triggering consecutive jobs.
+        rate_limit_backoff: Seconds to wait on 429/5xx before retrying
+            a single trigger (passed through to trigger_prow_job).
+        max_retries: Maximum trigger attempts per job on retryable errors.
+
+    Returns:
+        List of job dicts, each with keys: name, type, job_id, url, result.
+        Jobs that failed to trigger have job_id=None and result="error".
+    """
+    if not blocking_names:
+        raise ValueError("e2e-blocking-job-names is empty"
+                         " - at least one blocking job is required")
+
+    num_blocking = len(blocking_names)
+    num_informing = len(informing_names)
+    print(f"Blocking: {num_blocking} | Informing: {num_informing}"
+          f" | Total: {num_blocking + num_informing}", flush=True)
+    print(f"Trigger delay: {trigger_delay}s"
+          f" | Rate limit backoff: {rate_limit_backoff}s"
+          f" | Max retries: {max_retries}", flush=True)
+
+    jobs = []
+
+    print(f"\n=== run-e2e: triggering {num_blocking} blocking job(s) ===",
+          flush=True)
+    for i, name in enumerate(blocking_names):
+        job = _trigger_single(name, "blocking", gangway_url, token,
+                              env_overrides, max_retries, rate_limit_backoff)
+        jobs.append(job)
+        if i < num_blocking - 1 or num_informing > 0:
+            time.sleep(trigger_delay)
+
+    if num_informing > 0:
+        print(f"\n=== run-e2e: triggering {num_informing} informing job(s) ===",
+              flush=True)
+        for i, name in enumerate(informing_names):
+            job = _trigger_single(name, "informing", gangway_url, token,
+                                  env_overrides, max_retries, rate_limit_backoff)
+            jobs.append(job)
+            if i < num_informing - 1:
+                time.sleep(trigger_delay)
+
+    return jobs
+
+
+def _trigger_single(name, job_type, gangway_url, token, env_overrides,
+                    max_retries, backoff):
+    """Trigger a single Prow job and return a job dict.
+
+    Args:
+        name: Full Prow periodic job name.
+        job_type: "blocking" or "informing" (used for logging and result type).
+        gangway_url: Base URL of the Gangway executions endpoint.
+        token: Bearer token for Gangway authentication.
+        env_overrides: Dict of environment variables for the Prow job.
+        max_retries: Maximum trigger attempts on retryable errors.
+        backoff: Seconds to wait between retry attempts.
+
+    Returns:
+        Job dict with keys: name, type, job_id, url, result.
+    """
+    sn = short_name(name)
+    print(f"\n--- [{job_type}] Triggering [{sn}] ---", flush=True)
+    print(f"  Full name: {name}", flush=True)
+
+    job_id = trigger_prow_job(gangway_url, token, name, env_overrides,
+                              max_retries=max_retries, backoff=backoff)
+
+    job = {
+        "name": name,
+        "type": job_type,
+        "job_id": job_id,
+        "url": "",
+        "result": "pending" if job_id else "error"
+    }
+
+    if job_id:
+        print(f"  Job ID: {job_id}", flush=True)
+    else:
+        print(f"  ERROR: Trigger failed after {max_retries} attempts", flush=True)
+
+    return job
+
+
+def resolve_all_urls(jobs, gangway_url, token, max_attempts=10, delay=15):
+    """Resolve Prow UI URLs for all successfully triggered jobs.
+
+    Skips jobs with job_id=None (trigger failures). Updates each job
+    dict in-place with the resolved URL.
+
+    Args:
+        jobs: List of job dicts (as returned by trigger_all_jobs).
+        gangway_url: Base URL of the Gangway executions endpoint.
+        token: Bearer token for Gangway authentication.
+        max_attempts: Maximum polling attempts per job for URL resolution.
+        delay: Seconds to wait between polling attempts per job.
+    """
+    print("\n=== run-e2e: resolving Prow URLs ===", flush=True)
+    for job in jobs:
+        if job["job_id"] is None:
+            continue
+        url = resolve_prow_url(gangway_url, token, job["job_id"],
+                               max_attempts=max_attempts, delay=delay)
+        job["url"] = url
+        sn = short_name(job["name"])
+        if url:
+            print(f"  [{job['type']}] [{sn}] {url}", flush=True)
+        else:
+            print(f"  [{job['type']}] [{sn}] WARN: could not resolve URL",
+                  flush=True)
+
+
+def poll_until_complete(jobs, gangway_url, token, initial_delay=2700,
+                        poll_interval=600, poll_stagger=30, timeout=14400):
+    """Poll all pending jobs until every job completes or the timeout expires.
+
+    Waits initial_delay seconds before the first poll cycle. Each cycle
+    iterates over pending jobs with poll_stagger seconds between individual
+    status checks. Rate-limited responses (429/5xx) skip the job for the
+    current cycle. Jobs still pending after the timeout are marked as error.
+
+    Args:
+        jobs: List of job dicts (as returned by trigger_all_jobs).
+        gangway_url: Base URL of the Gangway executions endpoint.
+        token: Bearer token for Gangway authentication.
+        initial_delay: Seconds to wait before the first poll cycle
+            (gives Prow time to schedule and start the jobs).
+        poll_interval: Seconds to wait between poll cycles.
+        poll_stagger: Seconds to wait between polling individual jobs
+            within a single cycle (avoids Gangway rate limits).
+        timeout: Total seconds (including initial_delay) after which
+            remaining pending jobs are marked as error.
+    """
+    print(f"\nPoll interval: {poll_interval}s | Timeout: {timeout}s", flush=True)
+    print(f"Initial delay: {initial_delay}s | Poll stagger: {poll_stagger}s",
+          flush=True)
+
+    print(f"\n=== run-e2e: waiting {initial_delay}s before first poll ===",
+          flush=True)
+    start_time = time.monotonic()
+    time.sleep(initial_delay)
+
+    print("=== run-e2e: polling all jobs ===", flush=True)
+    poll_count = 0
+
+    while (time.monotonic() - start_time) < timeout:
+        pending = [j for j in jobs if j["result"] == "pending"]
+        if not pending:
+            break
+
+        poll_count += 1
+        elapsed = int(time.monotonic() - start_time)
+        ts = datetime.now(timezone.utc).strftime("%Y-%m-%dT%H:%M:%SZ")
+        print(f"\n[{ts}] Poll #{poll_count} ({elapsed}s)"
+              f" - {len(pending)} job(s) pending", flush=True)
+
+        poll_idx = 0
+        for job in jobs:
+            if job["result"] != "pending" or job["job_id"] is None:
+                continue
+
+            if poll_idx > 0:
+                time.sleep(poll_stagger)
+            poll_idx += 1
+
+            sn = short_name(job["name"])
+            mapped, raw = get_prow_job_status(gangway_url, token, job["job_id"])
+
+            if mapped == "rate_limited":
+                print(f"  [{job['type']}] [{sn}] HTTP {raw.split()[1] if ' ' in raw else raw}"
+                      f" - rate limited, will retry next cycle", flush=True)
+                continue
+            elif mapped == "error":
+                print(f"  [{job['type']}] [{sn}] {raw}", flush=True)
+                job["result"] = "error"
+                continue
+
+            print(f"  [{job['type']}] [{sn}] {raw}", flush=True)
+
+            if mapped == "passed":
+                job["result"] = "passed"
+                print(f"  [{job['type']}] [{sn}] PASSED", flush=True)
+            elif mapped == "failed":
+                job["result"] = "failed"
+                print(f"  [{job['type']}] [{sn}] FAILED ({raw})", flush=True)
+
+        pending_after = sum(1 for j in jobs if j["result"] == "pending")
+        if pending_after == 0:
+            break
+
+        time.sleep(poll_interval)
+
+    for job in jobs:
+        if job["result"] == "pending":
+            job["result"] = "error"
+            sn = short_name(job["name"])
+            print(f"  [{job['type']}] [{sn}] TIMEOUT", flush=True)
+
+
+def print_run_summary(jobs):
+    """Print a short results summary to stdout after polling completes.
+
+    Groups results by type (blocking, informing) and prints each job's
+    short name, result, and Prow URL.
+
+    Args:
+        jobs: List of job dicts (as returned by trigger_all_jobs).
+    """
+    print("\n=== run-e2e: results summary ===\n", flush=True)
+    blocking = [j for j in jobs if j["type"] == "blocking"]
+    informing = [j for j in jobs if j["type"] == "informing"]
+
+    print("--- BLOCKING TESTS ---", flush=True)
+    for j in blocking:
+        sn = short_name(j["name"])
+        print(f"  {sn}: {j['result']} ({j['url']})", flush=True)
+
+    if informing:
+        print("\n--- INFORMING TESTS ---", flush=True)
+        for j in informing:
+            sn = short_name(j["name"])
+            print(f"  {sn}: {j['result']} ({j['url']})", flush=True)
+
+
+def build_results_json(jobs):
+    """Serialize job results into a compact JSON string for Tekton results.
+
+    The output format is a JSON array of objects with keys:
+    job (full name), result, url, type. Uses compact separators
+    to minimize size (Tekton results have a 4KB limit).
+
+    Args:
+        jobs: List of job dicts (as returned by trigger_all_jobs).
+
+    Returns:
+        Compact JSON string, e.g. '[{"job":"...","result":"passed",...}]'.
+    """
+    results = []
+    for job in jobs:
+        results.append({
+            "job": job["name"],
+            "result": job["result"],
+            "url": job["url"],
+            "type": job["type"]
+        })
+    return json.dumps(results, separators=(",", ":"))
+
+
+def evaluate_gate(results_json_str):
+    """Evaluate the release gate verdict and print a formatted summary.
+
+    Gate logic: all blocking tests must have result=="passed" (AND).
+    Informing tests are reported but do not affect the verdict.
+    Prints a formatted table of results and the verdict to stdout.
+
+    Args:
+        results_json_str: JSON string as produced by build_results_json.
+
+    Returns:
+        True if the gate passed (all blocking tests passed), False otherwise.
+        Returns False if results_json_str is not valid JSON.
+    """
+    try:
+        results = json.loads(results_json_str)
+    except (json.JSONDecodeError, TypeError):
+        print("ERROR: results JSON is invalid or truncated, failing gate",
+              flush=True)
+        return False
+
+    blocking = [r for r in results if r["type"] == "blocking"]
+    informing = [r for r in results if r["type"] == "informing"]
+
+    if not blocking:
+        print("ERROR: no blocking tests found in results"
+              " - at least one blocking test is required", flush=True)
+        return False
+
+    print(f"--- BLOCKING TESTS ({len(blocking)}) ---", flush=True)
+    print(f"{'JOB':<50} {'RESULT':<10} URL", flush=True)
+    print(f"{'---':<50} {'------':<10} ---", flush=True)
+
+    gate_passed = True
+    for r in blocking:
+        job = short_name(r["job"])
+        print(f"{job:<50} {r['result']:<10} {r.get('url', '')}", flush=True)
+        if r["result"] != "passed":
+            gate_passed = False
+
+    if informing:
+        print(f"\n--- INFORMING TESTS ({len(informing)}) ---", flush=True)
+        print(f"{'JOB':<50} {'RESULT':<10} URL", flush=True)
+        print(f"{'---':<50} {'------':<10} ---", flush=True)
+        warnings = 0
+        for r in informing:
+            job = short_name(r["job"])
+            print(f"{job:<50} {r['result']:<10} {r.get('url', '')}", flush=True)
+            if r["result"] != "passed":
+                warnings += 1
+        if warnings > 0:
+            print(f"\nWARN: {warnings} informing test(s) failed"
+                  " (does not affect gate verdict)", flush=True)
+
+    verdict = "true" if gate_passed else "false"
+    print(f"\nGate verdict: {verdict} ({len(blocking)} blocking, AND logic)",
+          flush=True)
+
+    if gate_passed:
+        print("\n=== GATE PASSED - create-release will create the Release CR ===",
+              flush=True)
+    else:
+        print("\n=== GATE FAILED - image will NOT be promoted ===", flush=True)
+
+    return gate_passed
+
+
+def build_gate_notification(gate_passed, gate_label, results_json_str,
+                            release_name, snapshot, pipeline_run,
+                            konflux_base_url):
+    """Build the Slack Block Kit payload for a gate result notification.
+
+    Handles three visual states:
+    - Green (#2E7D32): gate passed and Release CR was created.
+    - Orange (#F57C00): gate passed but Release creation failed.
+    - Red (#D32F2F): gate failed (one or more blocking tests failed).
+
+    Args:
+        gate_passed: Boolean gate verdict from evaluate_gate.
+        gate_label: Human-readable gate label (e.g. "ARO HCP").
+        results_json_str: JSON string as produced by build_results_json.
+        release_name: Name of the created Release CR, or "N/A" if none.
+        snapshot: Konflux Snapshot name.
+        pipeline_run: Tekton PipelineRun name.
+        konflux_base_url: Base URL of the Konflux UI for building links.
+
+    Returns:
+        Dict payload ready to be passed to send_slack_message.
+    """
+    if release_name and release_name != "N/A":
+        release_field = (f"<{konflux_base_url}/releases/{release_name}/"
+                         f"|{release_name}>")
+    else:
+        release_field = "N/A"
+
+    if gate_passed and (release_name == "N/A" or not release_name):
+        color = "#F57C00"
+        header = (f":warning: *HO Release Gate [{gate_label}] PASSED"
+                  f" - Release creation failed*")
+    elif gate_passed:
+        color = "#2E7D32"
+        header = f":white_check_mark: *HO Release Gate [{gate_label}] PASSED*"
+    else:
+        color = "#D32F2F"
+        header = f":x: *HO Release Gate [{gate_label}] FAILED*"
+
+    result_lines = _format_test_results(results_json_str)
+
+    pr_link = (f"<{konflux_base_url}/pipelineruns/{pipeline_run}/"
+               f"|{pipeline_run}>")
+
+    blocks = [
+        mrkdwn_section(header),
+        fields_section([
+            f"*Gate:*\n{gate_label}",
+            f"*Snapshot:*\n{snapshot}",
+            f"*PipelineRun:*\n{pr_link}",
+            f"*Release:*\n{release_field}"
+        ]),
+        divider(),
+        mrkdwn_section(result_lines)
+    ]
+
+    return build_slack_payload(color, blocks)
+
+
+def _format_test_results(results_json_str):
+    """Format test results as Slack mrkdwn text for the notification body.
+
+    Groups results by type (blocking, informing) with emoji indicators
+    and Prow links.
+
+    Args:
+        results_json_str: JSON string as produced by build_results_json.
+
+    Returns:
+        Slack mrkdwn string with newline-separated result lines.
+    """
+    try:
+        results = json.loads(results_json_str)
+    except (json.JSONDecodeError, TypeError):
+        return ("\n:x: Pipeline failed before tests started."
+                " Check PipelineRun logs.")
+
+    def _format_group(group, header):
+        if not group:
+            return []
+        parts = [f"\n*{header}:*"]
+        for r in group:
+            job = short_name(r["job"])
+            emoji = (":white_check_mark:" if r["result"] == "passed"
+                     else ":red_circle:")
+            url = r.get("url", "")
+            if url:
+                parts.append(f"{emoji} `{job}` - {r['result']} (<{url}|Prow>)")
+            else:
+                parts.append(f"{emoji} `{job}` - {r['result']}")
+        return parts
+
+    parts = []
+    parts.extend(_format_group(
+        [r for r in results if r["type"] == "blocking"], "Blocking Tests"))
+    parts.extend(_format_group(
+        [r for r in results if r["type"] == "informing"], "Informing Tests"))
+
+    return "\n".join(parts) if parts else "\n:x: No test results available."
+
+
+def build_error_notification(pipeline_run, konflux_base_url, gate_label):
+    """Build the Slack Block Kit payload for a pipeline DAG failure.
+
+    Used by the notify-slack-error finally task, which fires only when
+    create-release was never reached (status=None). Always red (#D32F2F).
+
+    Args:
+        pipeline_run: Tekton PipelineRun name.
+        konflux_base_url: Base URL of the Konflux UI for building links.
+        gate_label: Human-readable gate label (e.g. "ARO HCP").
+
+    Returns:
+        Dict payload ready to be passed to send_slack_message.
+    """
+    pr_link = (f"<{konflux_base_url}/pipelineruns/{pipeline_run}/"
+               f"|{pipeline_run}>")
+
+    blocks = [
+        mrkdwn_section(
+            f":x: *HO Release Gate [{gate_label}] PIPELINE ERROR*\n"
+            "Pipeline failed before reaching the evaluation phase."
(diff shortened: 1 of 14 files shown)


.tekton/pipelines/ho-release-gate.yaml
```diff
@@ -0,0 +1,148 @@
...
+        script: |
+          #!/bin/bash
+          set -euo pipefail
+          echo "Snapshot: $(params.snapshot-name)"
+          echo "HO image: $(params.ho-image)"
+
+          if [[ -z "$(params.ho-image)" ]]; then
+            echo "ERROR: ho-image parameter is empty"
+            exit 1
+          fi
```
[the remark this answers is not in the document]
Nirshal: Addressed: added early validation for snapshot-name in extract-image. If the label is empty the task now exits immediately with a clear error, before triggering any e2e jobs.

Nirshal:
**Addressing CodeRabbit nitpick: pin container image versions**

All 7 occurrences of `quay.io/konflux-ci/appstudio-utils:latest` are now pinned using the `tag+digest` format (`latest@sha256:32cda04...`), following the Konflux recommended practice.

MintMaker (built on Renovate) will automatically open PRs to bump the digest when the `:latest` tag is updated in the registry. MintMaker scans all YAML files under `.tekton/` weekly (Saturdays after 5 AM UTC), including custom pipelines like ours.

Nirshal:
@coderabbitai resume

.tekton/lib/ho_release_gate.py
```diff
@@ -0,0 +1,701 @@
...
+              f" ({threshold_days}d) - normal notification",
+              flush=True)
+        return None
+
+    print(f"  Streak ({streak_days}d) >= threshold"
+          f" ({threshold_days}d) - STALE ALERT", flush=True)
+
+    return build_stale_notification(
+        streak, gate_label, threshold_days,
+        current_run_name, konflux_base_url, streak_days)
```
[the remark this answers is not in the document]
Nirshal: Fixed in f0baa9a: added try/except ValueError around fromisoformat, logs a WARN and returns None (skips stale check gracefully, same pattern as the empty-timestamp guard above).

.tekton/lib/kubearchive_utils.py
```diff
@@ -0,0 +1,98 @@
...
+        reason = last_cond.get("reason", "Unknown")
+        cond_status = last_cond.get("status", "False")
+
+        runs.append({"name": name, "created": created,
+                     "status": cond_status == "True", "reason": reason})
+
+    runs.sort(key=lambda r: r["created"], reverse=True)
+
+    print(f"  KubeArchive: found {len(runs)} PipelineRun(s)", flush=True)
+    return runs
```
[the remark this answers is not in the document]
Nirshal: Fixed in f0baa9a: added creationTimestampAfter (60-day lookback) and explicit limit (120) to the KubeArchive query. KubeArchive does not guarantee server-side sort order, so without a time filter the first page could contain arbitrarily old runs. The lookback window and limit are defined as documented constants (STALE_LOOKBACK_DAYS, STALE_QUERY_LIMIT) in kubearchive_utils.py. Stale promotion is a condition that should be resolved within days, so 60 days is deliberately oversized to never cut off relevant data.

.tekton/lib/mock/test_util_mock.py
```diff
@@ -0,0 +1,382 @@
...
+    enough days to trigger the stale promotion alert. The number
+    of days is derived from the STALE_THRESHOLD_DAYS environment
+    variable, which is already set in the pipeline tasks as part
+    of the stale alert configuration. The generated streak covers
+    threshold_days + 1 days to guarantee the stale condition is met
+    (streak_days >= threshold_days).
+
+    Each generated run is spaced one day apart, with the most recent
+    run placed yesterday (relative to now) and the oldest run placed
+    threshold_days ago. All runs have status=False and reason="Failed".
```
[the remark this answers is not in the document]
Nirshal: Fixed in f0baa9a: updated docstring to say "threshold_days + 1 days ago" which matches the actual code behavior (num_runs = threshold + 1, days_ago ranges from 1 to num_runs).

.tekton/lib/prow_utils.py
```diff
@@ -0,0 +1,181 @@
...
+    )
+
+    if status == 429 or (500 <= status < 600):
+        return "rate_limited", f"HTTP {status}"
+
+    if status != 200:
+        msg = f"HTTP {status} - auth error or unexpected failure"
+        if body:
+            print(f"  Raw response: {body}", flush=True)
+        return "error", msg
```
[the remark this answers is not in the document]
Nirshal: Fixed in f0baa9a: added `status == 0` to the retryable condition in `get_prow_job_status`. status=0 comes from `http_request` when `URLError` is caught (connection refused, DNS failure, timeout) -- transient network issues that should be retried on the next poll cycle rather than treated as terminal errors. Added test `test_connection_failure_is_retryable` to cover this.

.tekton/lib/prow_utils.py
```diff
@@ -0,0 +1,181 @@
...
+        "Content-Type": "application/json"
+    }
+
+    def should_retry(status):
+        return status == 0 or status == 429 or 500 <= status < 600
+
+    status, body = http_request_with_retry(
+        gangway_url, method="POST", headers=headers, body=payload,
+        retries=max_retries, backoff=backoff, should_retry=should_retry
+    )
```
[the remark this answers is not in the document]
Nirshal: Gangway does not support idempotency keys. The priority here is ensuring every test is triggered: if a blocking test fails to trigger, all other tests already running in the same pipeline become useless since the gate cannot produce a valid verdict. A sporadic duplicate Prow run is a minor side effect compared to a pipeline that cannot evaluate its gate. In practice, transient 5xx from the gateway proxy are the most common trigger failure mode.

.tekton/lib/tests/test_prow_utils.py
```diff
@@ -0,0 +1,250 @@
...
+        mock_req.return_value = (200, json.dumps({"id": "1"}))
+
+        trigger_prow_job(
+            "https://gangway.example.com/v1/executions",
+            "tok", "periodic-job", {"K": "V"}, max_retries=5, backoff=60)
+
+        _, kwargs = mock_req.call_args
+        body = kwargs.get("body") or mock_req.call_args[0][3] if len(mock_req.call_args[0]) > 3 else kwargs.get("body")
+        self.assertEqual(kwargs["retries"], 5)
+        self.assertEqual(kwargs["backoff"], 60)
```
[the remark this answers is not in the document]
Nirshal: Fixed in f0baa9a: the test now verifies the actual POST body (job_name, pod_spec_options.envs) and the retry parameters (retries, backoff) passed to http_request_with_retry.

.tekton/lib/mock/__init__.py
bryan-cox, line 1: What are these changes?

.tekton/pipelines/ho-release-gate-run.yaml
```diff
@@ -0,0 +1,29 @@
...
+        resources:
+          requests:
+            storage: 10Mi
+  pipelineRef:
+    resolver: git
+    params:
+      - name: url
+        value: https://github.com/openshift/hypershift
+      - name: revision
+        value: ho-release-gate-pipeline
```
bryan-cox: **[blocking]** This references the feature branch `ho-release-gate-pipeline`. Once merged and the branch is deleted, the git resolver will fail to find the Pipeline definition.

```suggestion
        value: main
```

If the ITS configuration overrides this value at runtime, please add a comment here stating that.

.tekton/pipelines/ho-release-gate.yaml
```diff
@@ -0,0 +1,510 @@
...
+      steps:
+      - name: clone
+        image: quay.io/konflux-ci/appstudio-utils:latest@sha256:32cda04122ee0a5d5b4307e121d192d1265c3bff727c9d894f36b6772da53d8c
+        script: |
+          #!/bin/bash
+          set -euo pipefail
+          echo "=== clone-lib ==="
+          echo "Running as: $(oc whoami)"
+          REPO_URL="https://github.com/openshift/hypershift.git"
+          BRANCH="ho-release-gate-pipeline"
```
bryan-cox: **[blocking]** Same issue as the PipelineRun template — this sparse checkout of `.tekton/lib/` will break once the feature branch is deleted post-merge.

```suggestion
          BRANCH="main"
```

.tekton/pipelines/ho-release-gate.yaml
```diff
@@ -0,0 +1,510 @@
...
+                                       build_results_json)
+
+          print("=== run-e2e ===", flush=True)
+          whoami = subprocess.run(["oc", "whoami"], capture_output=True, text=True)
+          print(f"Running as: {whoami.stdout.strip()}", flush=True)
+
+          with open("$(results.results-json.path)", "w") as f:
+              f.write("[]")
+
+          gangway_url = "https://gangway-ci.apps.ci.l2s4.p1.openshiftapps.com/v1/executions"
```
bryan-cox: **[suggestion]** This Gangway URL is hardcoded. The CI cluster has migrated before (`app.ci` → `build0x`). Consider making this a pipeline parameter with this as the default, so a cluster move doesn't require a code change.

.tekton/pipelines/ho-release-gate.yaml
```diff
@@ -0,0 +1,510 @@
...
+              print("ERROR: e2e-blocking-job-names is empty - at least one blocking job is required", flush=True)
+              sys.exit(1)
+
+          print(f"HO image:       {ho_image}", flush=True)
+          print(f"Blocking jobs:  {blocking}", flush=True)
+          print(f"Informing jobs: {informing}", flush=True)
+
+          env_overrides = {
+              "OVERRIDE_IMAGE_HYPERSHIFT_OPERATOR": ho_image,
+              "OVERRIDE_IMAGE_HYPERSHIFT_TESTS": "registry.ci.openshift.org/hypershift/hypershift-tests:latest"
```
bryan-cox, line 203: **[suggestion]** Using `:latest` for the test image means the gate runs whatever test binary happens to be latest, which may not correspond to the HO image being validated. If the Snapshot contains a test image component, it would be more correct to extract and use that. If `:latest` is intentional (e.g., tests are always forward-compatible), a comment explaining why would help future readers.

.tekton/pipelines/ho-release-gate.yaml
```diff
@@ -0,0 +1,510 @@
...
+                                       check_and_build_stale_payload)
+
+          snapshot = os.environ["SNAPSHOT_NAME"]
+          gate_passed = os.environ["GATE_PASSED"] == "true"
+          gate_label = os.environ["GATE_LABEL"]
+          release_name = os.environ["RELEASE_NAME"]
+          pipeline_run = "$(context.pipelineRun.name)"
+          results_json = os.environ["RESULTS_JSON"]
+          webhook_url = os.environ["SLACK_WEBHOOK_URL"]
+          konflux_base = "https://konflux-ui.apps.stone-prd-rh01.pg1f.p1.openshiftapps.com/ns/crt-redhat-acm-tenant/applications/hypershift-operator"
```
bryan-cox, line 402: **[suggestion]** Since ARO HCP is the pilot and ROSA/GCP HCP follow using the same template, consider extracting the namespace (`crt-redhat-acm-tenant`) and Konflux app URL as pipeline parameters. This avoids forking the pipeline when onboarding new platforms — their ITS definitions can supply different values.

.tekton/pipelines/ho-release-gate.yaml
```diff
@@ -0,0 +1,510 @@
...
+                  payload = build_gate_notification(
+                      gate_passed, gate_label, results_json,
+                      release_name, snapshot, pipeline_run, konflux_base)
+          else:
+              payload = build_gate_notification(
+                  gate_passed, gate_label, results_json,
+                  release_name, snapshot, pipeline_run, konflux_base)
+
+          success = send_slack_message(webhook_url, payload)
+          if not success:
```
bryan-cox, line 436: **[suggestion]** When the Slack notification fails, this task still exits 0, making the failure invisible in PipelineRun status. Consider `sys.exit(1)` so the PipelineRun shows a partial failure, or write a result indicating notification status. I understand the rationale of not blocking pipeline completion from a `finally` task, but silent notification failure means operators won't know alerts are broken.

.tekton/pipelines/ho-release-gate.yaml
```diff
@@ -0,0 +1,510 @@
...
+          token = token_result.stdout.strip()
+          threshold = int(os.environ["STALE_THRESHOLD_DAYS"])
+          stale = check_and_build_stale_payload(
+              token, its_scenario, threshold, pipeline_run,
+              konflux_base, "crt-redhat-acm-tenant", gate_label)
+          payload = stale if stale else build_error_notification(
+              pipeline_run, konflux_base, gate_label)
+
+          success = send_slack_message(webhook_url, payload)
+          if not success:
```
bryan-cox, line 525: **[question]** The PR description says `notify-slack` and `notify-slack-error` are "mutually exclusive." Is there a scenario where *neither* fires? A short comment in the YAML explaining the mutual exclusivity logic (which task fires under which condition) would help future maintainers.

.tekton/lib/ho_release_gate.py
```diff
@@ -0,0 +1,706 @@
...
+            sn = short_name(j["name"])
+            print(f"  {sn}: {j['result']} ({j['url']})", flush=True)
+
+
+def build_results_json(jobs):
+    """Serialize job results into a compact JSON string for Tekton results.
+
+    The output format is a JSON array of objects with keys:
+    job (full name), result, url, type. Uses compact separators
+    to minimize size (Tekton results have a 4KB limit).
```
bryan-cox, line 311: **[suggestion]** With multiple blocking + informing jobs, each carrying a full Prow URL (~150 chars), the result could exceed 4KB. Consider either:
1. Adding a size check with truncation fallback
2. Writing results to a workspace file instead of a Tekton result
3. Truncating URLs in the JSON (full URLs are already printed to stdout)

bryan-cox (commented):
## Review: `feat(ci): add ho-release-gate pipeline for nightly promotion`

Well-structured Tekton pipeline with solid Python library design. Two blocking issues (both the same class: feature branch references that need updating to `main`), several hardening suggestions, and a few questions. See inline comments.

| Category | Count |
|----------|-------|
| Blocking | 2 |
| Suggestions | 6 |
| Nits | 1 |
| Questions | 2 |
| Praise | 3 |

**Praise:**
- Excellent defensive design with default-first result initialization — prevents silent failures on OOMKill
- Comprehensive test suite with well-structured edge cases
- stdlib-only constraint is well-justified with clear upgrade path documented in README

Once the two branch references are corrected, this should be in good shape for a second pass.

.tekton/lib/kubearchive_utils.py
```diff
@@ -0,0 +1,114 @@
...
+
+import json
+from datetime import datetime, timezone, timedelta
+from urllib.parse import quote
+
+from http_utils import http_request_with_retry
+
+__all__ = ["fetch_pipelineruns", "build_pipelinerun_url"]
+
+KUBEARCHIVE_API_BASE = (
```
bryan-cox: **[suggestion]** This couples the library to a specific production cluster. Consider making it configurable via env var for staging/alternative environments:

```python
KUBEARCHIVE_API_BASE = os.environ.get(
    "KUBEARCHIVE_API_BASE",
    "https://kubearchive-api-server-product-kubearchive"
    ".apps.stone-prd-rh01.pg1f.p1.openshiftapps.com"
)
```

.tekton/pipelines/ho-release-gate.yaml
```diff
@@ -0,0 +1,510 @@
...
+
+          if not gate_passed:
+              its_scenario = os.environ["ITS_SCENARIO"]
+              token_result = subprocess.run(
+                  ["oc", "whoami", "-t"], capture_output=True, text=True)
+              token = token_result.stdout.strip()
+              threshold = int(os.environ["STALE_THRESHOLD_DAYS"])
+              stale = check_and_build_stale_payload(
+                  token, its_scenario, threshold, pipeline_run,
+                  konflux_base, "crt-redhat-acm-tenant", gate_label)
```
bryan-cox: **[question]** Since `generateName` is used with `oc create` in the `create-release` task, re-triggering the pipeline for the same Snapshot creates a duplicate Release CR. Is there deduplication upstream (ReleasePlanAdmission, managed pipeline), or should there be a check here?

.tekton/lib/mock/__init__.py
Nirshal: The `mock/` directory contains `test_util_mock.py`, a utility module that generates synthetic test data for end-to-end pipeline integration testing (fake Snapshots, simulated Gangway/Prow responses, stale streak generators). The `__init__.py` makes it a Python package so the utilities can be imported. These are not unit tests (those are in `tests/`) but helpers for validating the full pipeline flow. I left them in the PR because I found them very useful to test some corner cases, and I thought they might have value for future maintenance. That said, working with these requires temporarily altering the Tekton pipeline, which is something possible only on an open PR (which still requires the ITS to point to the PR branch instead of main), so I am open to removing these and archiving them locally on my computer, just in case I need them in the future. Waiting for feedback from you on this.

.tekton/pipelines/ho-release-gate-run.yaml
Nirshal: Fixed in `6028dd6`. Changed to `openshift/hypershift` + `main`.

.tekton/pipelines/ho-release-gate.yaml
Nirshal: Fixed in `6028dd6`. Changed to `openshift/hypershift` + `main`.

.tekton/pipelines/ho-release-gate.yaml
Nirshal: Agreed. The Gangway URL is hardcoded in a single place (the `run-e2e` inline script). I will extract it as a pipeline parameter with the current value as default, so a cluster migration only requires updating the ITS definition.

.tekton/pipelines/ho-release-gate.yaml
Nirshal: This is a known limitation that I already flagged to the team (see [Slack thread](https://redhat-internal.slack.com/archives/C09NZ92TA3C/p1780588165452839?thread_ts=1779890741.385209&cid=C09NZ92TA3C)). The Konflux Snapshot only contains the `hypershift-operator` component. The test image (`hypershift-tests`) is built by the OpenShift CI pipeline, not by Konflux, so it is not part of the Snapshot and there is no straightforward way to extract the matching version from another source. Using `:latest` is intentional given this constraint. I will add a comment in the code explaining why.

.tekton/pipelines/ho-release-gate.yaml
Nirshal: The namespace `crt-redhat-acm-tenant` is the Konflux tenant where all resources live (Snapshots, ReleasePlans, Release CRs) -- it is not specific to ARO HCP. When onboarding ROSA or GCP, each will have its own ITS definition and ReleasePlan, but they will all live in the same tenant namespace. The pipeline is already designed to be reusable across managed services through its parameters (`gate-label`, `release-plan-name`, `e2e-blocking-job-names`, etc.), so no forking is needed. The Konflux app URL is only used for building links in Slack notifications and is also tenant-level, not service-specific.

.tekton/pipelines/ho-release-gate.yaml
Nirshal: I intentionally chose to exit 0 on notification failure. If `notify-slack` exits non-zero, the PipelineRun is marked as Failed in Tekton and KubeArchive. When an operator notices that notifications are missing and starts investigating, they would see all PipelineRuns marked as Failed and would have to open each one to determine whether the failure was a real gate failure or just a broken notification. By keeping the exit code tied to the gate verdict only, the PipelineRun status directly answers "did the release succeed?" without ambiguity. At that point the operator already knows notifications are broken and can check the task logs of the latest run to understand why.

.tekton/pipelines/ho-release-gate.yaml
Nirshal: I agree that this is a bit counter-intuitive. The `notify-slack-error` is meant to be a "catch-all" before the release task is reached, so every error before reaching that point will be reported by that. The reason why we also have error handling in the other finally task is that after `create-release`, more information about the errors becomes available, so the message can be more helpful. The mutual exclusivity exists and relies on Tekton's task status propagation: (1) If the pipeline reaches `evaluate-results`, then `create-release` runs (Succeeded or Failed), `notify-slack` fires (it receives all resolved params), and `notify-slack-error` is skipped (create-release.status != None). (2) If the pipeline fails before `evaluate-results` (e.g. extract-image or run-e2e crash), `create-release` is skipped (status=None), its result params are unresolved so `notify-slack` cannot run, and `notify-slack-error` fires instead. There is no scenario where neither fires. I will add a YAML comment block above the `finally` section explaining this logic.

.tekton/lib/ho_release_gate.py
Nirshal: Good catch, this is a real risk. I will switch to your option 2: write the results JSON to a workspace file instead of a Tekton result. The `shared` workspace is already mounted on all tasks (we use it for the Python libraries), so both `evaluate-results` and `notify-slack` can read from it directly. I will also remove the Tekton result declaration to avoid confusion.

.tekton/lib/kubearchive_utils.py
Nirshal: Agreed. I will make it configurable via env var with the current value as default, as you suggested.

.tekton/pipelines/ho-release-gate.yaml
Nirshal: There is no deduplication at the Release CR creation level, so a duplicate CR will be created. However, the managed release pipeline is designed to be idempotent. Early in execution, the `filter-already-released-advisory-images` task checks if the container image digests from the Snapshot have already been pushed to the target repository. If so, it sets a `skip_release` flag that causes Tekton to skip all downstream tasks (push-snapshot, rh-sign-image, create-pyxis-image, create-advisory, etc.). Individual tasks also check their own domains separately as an additional safety net. In practice, a duplicate Release CR will start the pipeline but it will short-circuit after the initial setup tasks without altering the target registry or creating duplicate advisories. If you want, we could add a pre-creation check, but this will add complexity and could also not be straightforward, because we could have several Release CRs for the same Snapshot when we add more managed services (like ROSA) and in that case it would be further difficult and prone to errors to correctly check.

.tekton/pipelines/ho-release-gate.yaml
```diff
@@ -0,0 +1,537 @@
...
+
+          snapshot = os.environ["SNAPSHOT_NAME"]
+          gate_passed = os.environ["GATE_PASSED"] == "true"
+          gate_label = os.environ["GATE_LABEL"]
+          release_name = os.environ["RELEASE_NAME"]
+          pipeline_run = "$(context.pipelineRun.name)"
+          with open("$(workspaces.shared.path)/results.json") as f:
+              results_json = f.read()
+          webhook_url = os.environ["SLACK_WEBHOOK_URL"]
+          konflux_base = "https://konflux-ui.apps.stone-prd-rh01.pg1f.p1.openshiftapps.com/ns/crt-redhat-acm-tenant/applications/hypershift-operator"
```
bryan-cox, line 402: **Still open from round 1 — hardcoded Konflux URL + namespace**

`konflux_base` and the namespace `crt-redhat-acm-tenant` are baked into the pipeline YAML. If the tenant or Konflux instance changes, every pipeline that inlines this needs an update.

Consider lifting both into pipeline parameters (with the current values as defaults) so they can be overridden at the PipelineRun level without touching the Pipeline definition.

.tekton/pipelines/ho-release-gate.yaml
```diff
@@ -0,0 +1,537 @@
...
+                      gate_passed, gate_label, results_json,
+                      release_name, snapshot, pipeline_run, konflux_base)
+          else:
+              payload = build_gate_notification(
+                  gate_passed, gate_label, results_json,
+                  release_name, snapshot, pipeline_run, konflux_base)
+
+          success = send_slack_message(webhook_url, payload)
+          if not success:
+              print("ERROR: Slack notification failed after all retries", flush=True)
```
bryan-cox, line 437: **Still open from round 1 — silent `exit 0` on Slack failure**

When Slack notification fails, the task prints the error but still exits 0. Downstream consumers (humans, dashboards) have no way to tell that notification was lost.

Consider writing a task result (e.g. `$(results.slack-status.path)`) with `"failed"` so callers can detect it, even if the task itself should not block the pipeline.

.tekton/pipelines/ho-release-gate.yaml
```diff
@@ -0,0 +1,537 @@
...
+          whoami = subprocess.run(["oc", "whoami"], capture_output=True, text=True)
+          print(f"Running as: {whoami.stdout.strip()}", flush=True)
+          print(f"Snapshot:     {snapshot}", flush=True)
+          print(f"PipelineRun:  {pipeline_run}", flush=True)
+          print(f"Gate passed:  {gate_passed}", flush=True)
+          print(f"Gate label:   {gate_label}", flush=True)
+          print(f"Release:      {release_name}", flush=True)
+
+          if not gate_passed:
+              its_scenario = os.environ["ITS_SCENARIO"]
```
bryan-cox, line 414: **New — duplicated stale-check block**

The ~15-line stale-check block (`its_scenario = …` through `send_slack_message(stale_payload, …)`) is copy-pasted between `notify-slack` and `notify-slack-error`. This is the same logic with the same env vars — a classic Duplicated Code smell.

Extract it into a shared shell function or a separate Task so changes only need to happen in one place.

.tekton/pipelines/ho-release-gate.yaml
```diff
@@ -0,0 +1,537 @@
...
+
+          # :latest for hypershift-tests is intentional: the test image is built
+          # by OpenShift CI, not Konflux, so it is not part of the Snapshot and
+          # there is no straightforward way to extract a matching version.
+          env_overrides = {
+              "MULTISTAGE_PARAM_OVERRIDE_OVERRIDE_HYPERSHIFT_OPERATOR_IMAGE": ho_image,
+              "OVERRIDE_IMAGE_HYPERSHIFT_TESTS": "registry.ci.openshift.org/hypershift/hypershift-tests:latest"
+          }
+
+          jobs = trigger_all_jobs(blocking, informing, gangway_url, token, env_overrides)
```
bryan-cox, line 206: **New — `run-e2e` can exit non-zero despite spec "always exits 0"**

`trigger_all_jobs()` raises `ValueError` on missing env vars, and this `script:` block has no try/except around it. If `GANGWAY_URL` or a job-list var is unset, the task exits non-zero and the pipeline fails hard — violating the spec requirement that run-e2e always exits 0, deferring pass/fail to evaluate-results.

Wrap the call in a try/except that catches `ValueError`, logs it, and writes a sentinel to `$(results.*)` so evaluate-results can report the real cause.

.tekton/pipelines/ho-release-gate.yaml
```diff
@@ -0,0 +1,537 @@
...
+            exit 1
+          fi
+
+          echo "=== create-release ==="
+          echo "Running as: $(oc whoami)"
+          echo "Snapshot:     ${SNAPSHOT}"
+          echo "ReleasePlan:  ${RELEASE_PLAN}"
+
+          set +e
+          RELEASE_NAME=$(oc create -f - -o jsonpath='{.metadata.name}' <<EOFRELEASE
```
bryan-cox, line 303: **Still open from round 1 — duplicate Release CR on re-trigger**

If the pipeline is re-triggered for the same snapshot, `oc create -f -` will attempt to create a Release CR that may already exist, causing a failure. Is there deduplication logic elsewhere (e.g. a unique name derived from the snapshot SHA), or should this use `oc apply` / create-if-not-exists?

.tekton/lib/mock/test_util_mock.py
```diff
@@ -0,0 +1,383 @@
...
+          f" (STALE_THRESHOLD_DAYS={threshold})", flush=True)
+    for r in runs:
+        print(f"  [MOCK] {r['name']}"
+              f" | {r['created']}"
+              f" | {r['status']}", flush=True)
+
+    return runs
+
+
+def fetch_pipelineruns_mock_stale_long(token, namespace, label_selector):
```
bryan-cox, line 321: nit: `fetch_pipelineruns_mock_stale_long` is exported in `__all__` but never referenced in any test file. Dead code — either add test coverage that uses it or remove it.

bryan-cox (commented):
Second-round review. 6 inline comments — 4 carry forward from the first round (still open), 2 are new findings.

.tekton/pipelines/ho-release-gate.yaml
Nirshal: The duplicated block is the parameter preparation for `check_and_build_stale_payload()`, which is where the actual shared logic lives (already extracted in `ho_release_gate.py`). The env var reads and token acquisition are the glue that feeds values into that function.

Extracting this glue into a helper would create a script-like function that reads from `os.environ` and calls `subprocess` internally. This would be difficult to test in isolation (it's not a pure function, it's effectively a script), and it would break the design pattern we follow throughout the pipeline: testable, reusable functions in the Python library, called by thin inline scripts in the YAML. Adding a helper in between would introduce a layer of indirection (scripts calling scripts calling functions) without reducing complexity.

Since the two finally tasks run as separate Tekton pods with no shared execution context, each must independently prepare its own parameters. I think the current factoring is the right one, but happy to chat about it if you see it differently.

.tekton/lib/mock/__init__.py
bryan-cox: You're right — the mock package is a self-contained toolkit for manual integration testing, not a unit test suite. `fetch_pipelineruns_mock_stale_long` is documented with usage instructions in the module docstring and is part of the mock API surface. "Not called from `tests/`" doesn't make it dead code here. Withdrawing.

.tekton/pipelines/ho-release-gate.yaml
bryan-cox: Makes sense — the namespace and URL are tenant-level constants shared across all managed services, not service-specific values. The pipeline is already parameterized at the right level (`gate-label`, `release-plan-name`, `e2e-blocking-job-names`). Extracting these would add complexity for a scenario that won't happen. Withdrawing.

.tekton/pipelines/ho-release-gate.yaml
bryan-cox: Good rationale. PipelineRun status = gate verdict is the right invariant for a `finally` task — mixing in notification health would make triage harder, not easier. Withdrawing.

.tekton/pipelines/ho-release-gate.yaml
bryan-cox: I misread the failure mode — `generateName` + `oc create` produces a new CR each time, so there's no collision. And the managed release pipeline's `filter-already-released-advisory-images` task handles idempotency downstream. No dedup needed here. Withdrawing.

bryan-cox:
Replying to Alessandro's four responses from the first round — he's right on all of them, withdrawing those items.

**Mock dead code (`fetch_pipelineruns_mock_stale_long`):** The `mock/` package is a self-contained toolkit for manual integration testing, not a unit test suite. The function is documented with usage instructions in the module docstring and is part of the mock API surface. "Not called from `tests/`" doesn't make it dead code here. Withdrawing.

**Konflux URL + namespace:** The namespace and URL are tenant-level constants shared across all managed services, not service-specific values. The pipeline is already parameterized at the right level (`gate-label`, `release-plan-name`, `e2e-blocking-job-names`). Withdrawing.

**Silent exit 0 on Slack failure:** PipelineRun status = gate verdict is the right invariant for a `finally` task — mixing in notification health would make triage harder, not easier. Withdrawing.

**Duplicate Release CR:** I misread the failure mode — `generateName` + `oc create` produces a new CR each time, so there's no collision. The managed release pipeline's `filter-already-released-advisory-images` handles idempotency downstream. Withdrawing.

The two new findings from round 2 (duplicated stale-check block and `run-e2e` ValueError path) still stand as independent items. Sorry for the noise on the re-opened ones.

.tekton/lib/mock/__init__.py
bryan-cox: You're right — the mock package is a self-contained toolkit for manual integration testing, not a unit test suite. `fetch_pipelineruns_mock_stale_long` is documented with usage instructions in the module docstring and is part of the mock API surface. "Not called from `tests/`" doesn't make it dead code here. Withdrawing.

.tekton/pipelines/ho-release-gate.yaml
bryan-cox: Makes sense — the namespace and URL are tenant-level constants shared across all managed services, not service-specific values. The pipeline is already parameterized at the right level (`gate-label`, `release-plan-name`, `e2e-blocking-job-names`). Withdrawing.

.tekton/pipelines/ho-release-gate.yaml
bryan-cox: Good rationale. PipelineRun status = gate verdict is the right invariant for a `finally` task — mixing in notification health would make triage harder, not easier. Withdrawing.

.tekton/pipelines/ho-release-gate.yaml
bryan-cox: I misread the failure mode — `generateName` + `oc create` produces a new CR each time, so there's no collision. The managed release pipeline's `filter-already-released-advisory-images` handles idempotency downstream. Withdrawing.

.tekton/pipelines/ho-release-gate.yaml
bryan-cox: test reply

.tekton/pipelines/ho-release-gate.yaml
bryan-cox: Alessandro's response is correct — the namespace and URL are tenant-level constants shared across all managed services, not service-specific. The pipeline is already parameterized at the right level. Withdrawing.

.tekton/pipelines/ho-release-gate.yaml
Nirshal: The pipeline has two distinct error categories by design, each with its own reporting path:

1. **Test failures** (normal operation): one or more Prow jobs fail. `run-e2e` exits 0, writes structured results to the workspace, `evaluate-results` determines the gate verdict, and `notify-slack` sends a detailed notification with per-job status, URLs, and stale promotion alerts. This is the path optimized for actionable detail.

2. **Infrastructure errors** (exceptional): a task crashes due to misconfiguration, transient cluster issues, API outages, etc. `notify-slack-error` fires and alerts the team that the pipeline itself broke. Troubleshooting goes through PipelineRun logs, which are the only place that can provide full context for unexpected failures.

These two paths are intentionally separate. Adding error handling inside `run-e2e` to catch infrastructure errors and route them through `notify-slack` would blur this boundary: we would have two error reporting paths for infrastructure failures (one partial in `run-e2e`, one generic in `notify-slack-error`), and the partial one could never be exhaustive because it only covers one task out of four. A crash in `clone-lib`, `extract-image`, or `evaluate-results` would still go through `notify-slack-error` regardless. The result would be added complexity without added coverage.

On the specific scenario (missing env vars): `GANGWAY_URL`, `HO_IMAGE`, `BLOCKING_JSON`, and `INFORMING_JSON` are all injected via explicit Tekton param bindings in the task definition, and `GANGWAY_TOKEN` comes from a `secretKeyRef`. If any of these are missing, Tekton fails the pod scheduling before the script runs - the `os.environ` reads are never reached.

.tekton/pipelines/ho-release-gate.yaml
bryan-cox: Alessandro's rationale holds. PipelineRun status = gate verdict is the right invariant for a finally task — mixing in notification health would make triage harder, not easier. Withdrawing

.tekton/pipelines/ho-release-gate.yaml
bryan-cox: Misread the failure mode — generateName + oc create produces a new CR each time, no collision. The managed release pipeline's filter-already-released-advisory-images handles idempotency downstream. Withdrawing.

.tekton/lib/mock/test_util_mock.py
bryan-cox: The mock/ package is a self-contained toolkit for manual integration testing, not a unit test suite. The function is documented with usage instructions in the module docstring. "Not called from tests/" doesn't make it dead code. Withdrawing.

bryan-cox (approved):

PR #8615, opened 2026-05-28
OCPBUGS-57453: fix(cpo): use Status.DNSZoneID before querying AWS Route53

PoornimaSingour:
After a CPO pod restart, the **AWSEndpointService** reconciler called **Route53 ListHostedZones** to look up the DNS zone ID on every reconciliation, even though **Status.DNSZoneID** already had the correct value persisted in the CR. At scale with many HostedClusters failing, this could hit AWS rate limits.

Added a three-tier lookup for the zone ID: 

- in-memory cache first, 
- then the persisted CR status, 
- and only fall back to the AWS API when both are empty (first-ever reconciliation). 

When reading from status, the in-memory cache is also populated so subsequent reconciles within the same pod lifetime use the fastest path.

Fixes  : https://redhat.atlassian.net/browse/OCPBUGS-57453

## Checklist:

## Summary by CodeRabbit

* **Resilience / Bug Fixes**
  * Improved reconciliation during AWS throttling: throttling errors are logged and retried after a fixed 2-minute backoff instead of failing immediately.
  * Enhanced DNS hosted-zone selection by preferring cached values, then using the status-provided value, and only looking up when both are missing.
  * If Route53 reports a missing hosted zone, cached and status-based hosted-zone identifiers are cleared to allow recovery.

* **Tests**
  * Added unit tests covering hosted-zone resolution, DNS record creation, throttling detection, and recovery behavior.
  * Extended Route53 record tests to verify the original AWS error type is preserved.

### DIFF
--- a/control-plane-operator/controllers/awsprivatelink/awsprivatelink_controller.go
+++ b/control-plane-operator/controllers/awsprivatelink/awsprivatelink_controller.go
@@ -789,15 +789,19 @@ func (r *AWSEndpointServiceReconciler) reconcileEndpointDNSRecords(ctx context.C
 
 	zn := zoneName(hcp.Name)
 	var zoneID string
-	if r.awsClientBuilder.getLocalHostedZoneID() == "" {
+	if localZoneID := r.awsClientBuilder.getLocalHostedZoneID(); localZoneID != "" {
+		zoneID = localZoneID
+	} else if awsEndpointService.Status.DNSZoneID != "" {
+		zoneID = awsEndpointService.Status.DNSZoneID
+		r.awsClientBuilder.setLocalHostedZoneID(zoneID)
+		log.Info("using DNSZoneID from status", "zoneID", zoneID)
+	} else {
 		var err error
 		zoneID, err = lookupZoneID(ctx, route53Client, zn)
 		if err != nil {
 			return nil, "", err
 		}
 		r.awsClientBuilder.setLocalHostedZoneID(zoneID)
-	} else {
-		zoneID = r.awsClientBuilder.getLocalHostedZoneID()
 	}
 
 	var fqdns []string
--- a/control-plane-operator/controllers/awsprivatelink/awsprivatelink_controller.go
+++ b/control-plane-operator/controllers/awsprivatelink/awsprivatelink_controller.go
@@ -57,6 +57,7 @@ import (
 const (
 	defaultResync               = 10 * time.Hour
 	externalPrivateServiceLabel = "hypershift.openshift.io/external-private-service"
+	throttleRequeueDelay        = 2 * time.Minute
 )
 
 // PrivateServiceObserver watches a given Service type LB and reconciles
@@ -527,6 +528,10 @@ func (r *AWSEndpointServiceReconciler) Reconcile(ctx context.Context, req ctrl.R
 				return ctrl.Result{}, err
 			}
 		}
+		if isAWSThrottleError(err) {
+			log.Info("AWS rate limit hit, backing off", "requeueAfter", throttleRequeueDelay)
+			return ctrl.Result{RequeueAfter: throttleRequeueDelay}, nil
+		}
 		return ctrl.Result{}, err
 	}
 
@@ -546,6 +551,14 @@ func (r *AWSEndpointServiceReconciler) Reconcile(ctx context.Context, req ctrl.R
 	// always requeue to catch and report out of band changes in AWS
 	// NOTICE: if the RequeueAfter interval is short enough, it could result in hitting some AWS request limits.
 	return ctrl.Result{RequeueAfter: 5 * time.Minute}, nil
+}
+
+func isAWSThrottleError(err error) bool {
+	switch supportawsutil.AWSErrorCode(err) {
+	case "Throttling", "ThrottlingException", "RequestLimitExceeded", "TooManyRequestsException":
+		return true
+	}
+	return false
 }
 
 func hasAWSConfig(platform *hyperv1.PlatformSpec) bool {
@@ -810,6 +823,12 @@ func (r *AWSEndpointServiceReconciler) reconcileEndpointDNSRecords(ctx context.C
 		fqdns = append(fqdns, fqdn)
 		err := CreateRecord(ctx, route53Client, zoneID, fqdn, aws.ToString(endpointDNSEntries[0].DnsName), route53types.RRTypeCname)
 		if err != nil {
+			var noSuchZone *route53types.NoSuchHostedZone
+			if errors.As(err, &noSuchZone) {
+				r.awsClientBuilder.setLocalHostedZoneID("")
+				awsEndpointService.Status.DNSZoneID = ""
+				log.Info("hosted zone not found, clearing cached DNSZoneID", "zoneID", zoneID)
+			}
 			return nil, "", err
 		}
 		log.Info("DNS record created", "fqdn", fqdn)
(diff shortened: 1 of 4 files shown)


control-plane-operator/controllers/awsprivatelink/awsprivatelink_controller.go
```diff
@@ -789,15 +789,19 @@ func (r *AWSEndpointServiceReconciler) reconcileEndpointDNSRecords(ctx context.C
...
 	zn := zoneName(hcp.Name)
 	var zoneID string
-	if r.awsClientBuilder.getLocalHostedZoneID() == "" {
+	if localZoneID := r.awsClientBuilder.getLocalHostedZoneID(); localZoneID != "" {
+		zoneID = localZoneID
+	} else if awsEndpointService.Status.DNSZoneID != "" {
+		zoneID = awsEndpointService.Status.DNSZoneID
+		r.awsClientBuilder.setLocalHostedZoneID(zoneID)
+		log.Info("using DNSZoneID from status", "zoneID", zoneID)
+	} else {
```
clebs, line 811:
```suggestion
	zoneID := r.awsClientBuilder.getLocalHostedZoneID()
	if zoneID == "" && awsEndpointService.Status.DNSZoneID != "" {
		zoneID = awsEndpointService.Status.DNSZoneID
		r.awsClientBuilder.setLocalHostedZoneID(zoneID)
		log.Info("using DNSZoneID from status", "zoneID", zoneID)
	} else {
```

clebs (commented):
@PoornimaSingour left a small suggestion, otherwise LGTM!

control-plane-operator/controllers/awsprivatelink/awsprivatelink_controller.go
PoornimaSingour: @clebs Thanks for the suggestion.
 I think the 3-way if/else if/else is actually needed here. With the 2-way version, when the in-memory cache is already populated (zoneID != ""), the condition zoneID == "" is false, so we'd fall into the else block and call lookupZoneID() — hitting the Route53 ListHostedZones API on every reconcile even though we already have the value cached. That's the exact problem this PR is trying to fix. The three branches map to three distinct cases:

-   Cache hit → use it, no I/O
-   Cache miss, but status has it → restore cache from status, no API call
-   Neither → look it up via Route53

  Lets me know your views in the same.

clebs: Oh yes, you are right. I missed that last branch on my suggestion. The correct code would be:
```
zoneID := r.awsClientBuilder.getLocalHostedZoneID()
if zoneID == "" {
    awsEndpointService.Status.DNSZoneID != "" {
        zoneID = awsEndpointService.Status.DNSZoneID
	r.awsClientBuilder.setLocalHostedZoneID(zoneID)
	log.Info("using DNSZoneID from status", "zoneID", zoneID)
    } else {
        // rest of the else block here
    }
}   
```

But I digress, it does not differ much from what you already have, take the version you find more understandable.

PoornimaSingour: @clebs Thanks for the follow-up and the corrected version.

Both are logically equivalent. I'll keep the flat `if/else if/else` since the three tiers are mutually exclusive and this way all three cases are visible at the same level.

 Appreciate the review!

csrwng (commented):
The 3-tier lookup (cache → status → API) is a good approach for reducing unnecessary Route53 calls after pod restarts. Two concerns below about error recovery and throttling resilience.

control-plane-operator/controllers/awsprivatelink/awsprivatelink_controller.go
```diff
@@ -789,15 +789,19 @@ func (r *AWSEndpointServiceReconciler) reconcileEndpointDNSRecords(ctx context.C
 
 	zn := zoneName(hcp.Name)
 	var zoneID string
-	if r.awsClientBuilder.getLocalHostedZoneID() == "" {
+	if localZoneID := r.awsClientBuilder.getLocalHostedZoneID(); localZoneID != "" {
+		zoneID = localZoneID
+	} else if awsEndpointService.Status.DNSZoneID != "" {
```
csrwng, line 807: If the hosted zone is ever deleted and recreated (getting a new zone ID), the stale value persisted in `Status.DNSZoneID` will be used indefinitely — even across pod restarts — causing `CreateRecord` to fail repeatedly with `NoSuchHostedZone`.

The Jira bug itself notes: *"Not only rely in the Status because it could expire."*

Suggestion: in the `CreateRecord` error handling block below (line ~812), check for a stale zone and clear both caches so the next reconcile recovers via a fresh API lookup:

```go
err := CreateRecord(ctx, route53Client, zoneID, fqdn, ...)
if err != nil {
    if err.Error() == "NoSuchHostedZone" {
        r.awsClientBuilder.setLocalHostedZoneID("")
        awsEndpointService.Status.DNSZoneID = ""
        log.Info("hosted zone not found, clearing cached DNSZoneID", "zoneID", zoneID)
    }
    return nil, "", err
}
```

The cleared `Status.DNSZoneID` will be persisted by the status update in the Reconcile error path (line ~526), so the next reconcile will fall through to `lookupZoneID`.

control-plane-operator/controllers/awsprivatelink/awsprivatelink_controller.go
```diff
@@ -789,15 +789,19 @@ func (r *AWSEndpointServiceReconciler) reconcileEndpointDNSRecords(ctx context.C
...
+		zoneID = awsEndpointService.Status.DNSZoneID
+		r.awsClientBuilder.setLocalHostedZoneID(zoneID)
+		log.Info("using DNSZoneID from status", "zoneID", zoneID)
+	} else {
 		var err error
 		zoneID, err = lookupZoneID(ctx, route53Client, zn)
 		if err != nil {
```
csrwng, line 814: The original incident (OHSS-44302) involved a death spiral: Route53 rate limiting caused `lookupZoneID` to fail, which left the cache empty, which caused more `lookupZoneID` calls on the next reconcile, making the rate limiting worse. The current controller-runtime backoff maxes out at 30 seconds (configured at controller setup) — too short to recover from Route53 rate limiting.

Consider checking for throttling errors (e.g., `err.Error() == "Throttling"` or `"RequestLimitExceeded"`) and returning a longer `RequeueAfter` from the Reconcile method — something like 1-2 minutes — to give Route53 time to recover.

`CreateRecord` already extracts the AWS error code via `smithy.APIError.ErrorCode()` and returns it as the error string, so you can match against known throttle codes. The same applies to errors from `lookupZoneID` here — though note that `lookupZoneID` returns the raw AWS error (not code-only), so you may want to use `awsutil.AWSErrorCode(err)` or a similar check at the Reconcile level.

control-plane-operator/controllers/awsprivatelink/awsprivatelink_controller.go
PoornimaSingour: @csrwng , Thanks for the suggestion. Addressed in the latest commit.

control-plane-operator/controllers/awsprivatelink/awsprivatelink_controller.go
PoornimaSingour: @csrwng, Thank you !!! Addressed — added 2-minute RequeueAfter on throttle errors using supportawsutil.AWSErrorCode(err) at the Reconcile level, catching throttle errors
from both CreateRecord and lookupZoneID uniformly. Added unit tests for isAWSThrottleError.

PoornimaSingour:
/verified by me 

test results uploaded here : https://drive.google.com/drive/folders/1E3ZrhfYAxWVlFmSu_8tjWmnfYJhAbNC0?usp=drive_link

PR #8616, opened 2026-05-28
NO-JIRA: feat(skills): add validate-pr-override-images skill

enxebre:
## Summary
- Add a Claude Code skill and GitHub Action to validate that CPO override images in a PR actually contain the fix PRs they claim to include
- `validate-overrides.sh` — orchestrates the full validation: parses a structured contract from the PR description, extracts override images per branch from the diff, and validates each image contains the claimed PRs using `git merge-base --is-ancestor`
- `verify-pr-in-image.sh` — inspects a container image via skopeo, reads its `vcs-ref` label, and checks whether a given PR merge commit is an ancestor of the image's build commit
- `SKILL.md` — Claude Code skill wrapper that invokes `validate-overrides.sh`
- GitHub Action workflow triggers on PRs modifying `overrides.yaml` and runs the validation automatically

Depends on https://github.com/openshift/hypershift/pull/8627 to add skopeo and gh CLI to the ARC runner image.

## PR description contract

PR authors modifying CPO overrides must include lines in the PR description like:
```
branch: 4.19 wants: https://github.com/openshift/hypershift/pull/1234
branch: 4.20 wants: https://github.com/openshift/hypershift/pull/1234, https://github.com/openshift/hypershift/pull/5678
```

The workflow fails if no such lines are found.

## Usage
```
/validate-pr-override-images <PR-URL-or-number>
```

Or run the script directly:
```bash
.claude/skills/validate-pr-override-images/validate-overrides.sh <pr-number>
```

## Test plan
- [x] `verify-pr-in-image.sh` returns PASS for a known-good image+PR combination
- [x] `verify-pr-in-image.sh` returns FAIL (exit 1) for a PR not in the image
- [x] `validate-overrides.sh 8616` parses contract from this PR description and validates images
- [x] Skill appears in Claude Code available skills list
- [x] GitHub Action workflow passes with correct contract
- [x] GitHub Action workflow fails with incorrect contract

### DIFF
--- a/.claude/skills/validate-pr-override-images/SKILL.md
+++ b/.claude/skills/validate-pr-override-images/SKILL.md
@@ -1,0 +1,38 @@
+---
+description: Validates that CPO override images in a PR actually contain the PRs they claim to include
+argument-hint: "<PR-URL-or-number>"
+---
+
+## Name
+validate-pr-override-images
+
+## Synopsis
+```text
+/validate-pr-override-images <PR-URL-or-number>
+```
+
+## Description
+Validates that CPO override images in a PR actually contain the claimed fix PRs.
+
+The PR description must include a structured contract:
+```
+branch: 4.20 wants: https://github.com/openshift/hypershift/pull/8593
+branch: 4.21 wants: https://github.com/openshift/hypershift/pull/8593, https://github.com/openshift/hypershift/pull/8565
+```
+
+Prerequisites:
+- `skopeo` must be installed (`brew install skopeo` on macOS)
+- The local git repo must have the relevant release branches fetched
+- Images must be accessible from quay.io
+
+## Implementation
+
+Extract the PR number from the argument, then run:
+```bash
+.claude/skills/validate-pr-override-images/validate-overrides.sh <pr-number>
+```
+
+Report the output to the user.
+
+## Arguments
+- `$1`: PR URL (e.g., `https://github.com/openshift/hypershift/pull/8610`) or PR number (e.g., `8610`)
--- a/.claude/skills/validate-pr-override-images/validate-overrides.sh
+++ b/.claude/skills/validate-pr-override-images/validate-overrides.sh
@@ -1,0 +1,179 @@
+#!/bin/bash
+# validate-overrides.sh
+# Parses a PR description for the override contract (branch: X.Y wants: PR-links),
+# extracts override images from the diff, and validates each image contains the claimed PRs.
+# Usage: ./validate-overrides.sh <pr-number> [repo]
+
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="$(cd "$SCRIPT_DIR/../../.." && pwd)"
+
+if [[ $# -lt 1 || $# -gt 2 ]]; then
+  echo "Usage: $0 <pr-number> [repo]" >&2
+  exit 2
+fi
+
+PR="$1"
+GH_REPO="${2:-openshift/hypershift}"
+
+echo "=== Validating CPO override images for PR #${PR} ==="
+echo ""
+
+# Step 1: Parse PR description for the contract
+echo "--- Step 1: Parsing PR description ---"
+BODY=$(gh pr view "$PR" --repo "$GH_REPO" --json body -q .body | tr -d '\r')
+
+BRANCH_LIST=""
+FOUND_LINES=0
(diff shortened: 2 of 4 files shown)


jparrill:
Hey @enxebre, thanks for putting this together — the idea of validating override images is solid and I've been thinking about this gap too.

My concern is that as a Claude skill this relies on someone remembering to run it. If we forget, we're back to square one. Also the script has some fragilities: it needs podman running locally, depends on merge commit message format (squash merges or different cherry-pick conventions would break it), and git fetch --all can be slow.

What if instead of a manual skill we make this a CI check that runs automatically on every PR touching overrides.yaml? We already have the pattern for this:

1. Unit tests in overrides_test.go — validate structural invariants with zero network: no duplicate versions per platform, valid image ref format (@sha256: digest), sane testing config (at most one platform with runTests: true). These catch copy-paste errors at make test time.
2. A verify tool + GHA workflow — following the `verify-api-deps` / `cpo-container-sync` pattern, add a `verify-cpo-overrides` target that checks image existence via registry API (HEAD on the manifest, no full pull needed). Wire it into a workflow triggered on paths: `['hypershift-operator/controlplaneoperator-overrides/**']`.

This way every override PR gets validated automatically — no manual step, no podman dependency, and it catches the most common mistakes (wrong digest, image doesn't exist, duplicate versions). 

WDYT?

enxebre:
> target that checks image existence via registry API (HEAD on the manifest, no full pull needed). Wire it into a workflow triggered on paths: ['hypershift-operator/controlplaneoperator-overrides/**']

These are meant to be building blocks for CI automation.
The value of the script is validating that intended git backports ARE within an container image
The value of the skill is judgement to figure what the intended backports for each image override based on a PR desc. 

You could possibly do the second step deterministically as part of the script without llm, agreeing on a contractual consistent PR desc for overrides changes.

enxebre:
>You could possibly do the second step deterministically as part of the script without llm, agreeing on a contractual consistent PR desc for overrides changes.

I just pushed an absolutely untested commit so you can see the idea
https://github.com/openshift/hypershift/pull/8616/changes/05a351cf7ac7ee5a5283bc4ed5ac459d1e7fa638

if doing that deterministically results too convoluted, we can still run the skill from GH actions, it's just a building block

/hold

.claude/skills/validate-pr-override-images/SKILL.md
```diff
@@ -0,0 +1,90 @@
...
+
+It works by:
+1. Fetching the PR description and diff
+2. Extracting claimed fix PR numbers from the description (e.g., `#8593, #8565, #8564`), ignoring the PR's own number
+3. Extracting unique override images from added `cpoImage:` lines in the diff, deduplicated by digest
+4. For each unique image, running `verify-pr-in-image.sh` (located in this skill's directory) against each claimed PR
+5. Reporting a pass/fail summary per image and per PR
+
+Prerequisites:
+- `podman` must be running (`podman machine start` if needed)
```
celebdor: Shouldn't it also be pointed to a valid pull secret? some overrides can be to actually released newer Z streams, for which skopeo would need the pull secret

celebdor (changes requested):

.claude/skills/validate-pr-override-images/SKILL.md
```diff
@@ -0,0 +1,90 @@
...
+Parse added lines matching the `cpoImage:` pattern. Deduplicate — many version entries share the same image digest.
+
+### Step 4: Verify each image contains the claimed PRs
+For each unique image and each claimed PR, run:
+```bash
+.claude/skills/validate-pr-override-images/verify-pr-in-image.sh <image> <pr-number> <repo-root>
+```
+
+The script:
+- Pulls the image via `podman`
```
jparrill: I've tested skopeo command in local and it not perform a pull of the image (podman pull docker://xxxxx) to grab the details, so this needs to be adjusted.

.claude/skills/validate-pr-override-images/SKILL.md
jparrill: I would say that this makes sense in darwin environments, but in a regular Linux env this is not necessary, only the service running if applies.

jparrill (commented):
Dropped some more comments. Thanks!

.claude/skills/validate-pr-override-images/SKILL.md
celebdor: Right, the advantage of using skopeo is that it does NOT make the pull

